PR #299, opened 2023-08-31
fetch git tags once to resolve go module versions

taylormadore:
STONEBLD-1271

Improves the performance of gomod requests when
the number of modules in the repository is large
and/or the git history is extensive. Git tags will only be fetched once per request and the results
are cached in a new ModuleVersionResolver class

### DIFF
--- a/cachi2/core/package_managers/gomod.py
+++ b/cachi2/core/package_managers/gomod.py
@@ -855,22 +855,38 @@ class ModuleVersionResolver:
         # If the module is version v2 or higher, the major version of the module is included as /vN at
         # the end of the module path. If the module is version v0 or v1, the major version is omitted
         # from the module path.
-        module_major_version = None
         match = re.match(r"(?:.+/v)(?P<major_version>\d+)$", module_name)
-        if match:
-            module_major_version = int(match.groupdict()["major_version"])
+        module_major_version = int(match.group("major_version")) if match else None
 
-        if module_major_version:
-            major_versions_to_try: tuple[int, ...] = (module_major_version,)
-        else:
-            # Prefer v1.x.x tags but fallback to v0.x.x tags if both are present
-            major_versions_to_try = (1, 0)
+        # If no match, prefer v1.x.x tags but fallback to v0.x.x tags if both are present
+        major_versions_to_try = (module_major_version,) if module_major_version else (1, 0)
 
         if app_dir.path == app_dir.root:
             subpath = None
         else:
             subpath = app_dir.path.relative_to(app_dir.root).as_posix()
 
+        tag_on_commit = self._get_highest_semver_tag_on_current_commit(
+            major_versions_to_try, subpath
+        )
+        if tag_on_commit:
+            return tag_on_commit
+
+        log.debug("No semantic version tag was found on the commit %s", self._commit.hexsha)
+        pseudo_version = self._get_highest_reachable_semver_tag(major_versions_to_try, subpath)
+        if pseudo_version:
+            return pseudo_version
+
+        log.debug("No valid semantic version tag was found")
+        # Fall-back to a vX.0.0-yyyymmddhhmmss-abcdefabcdef pseudo-version
+        return self._get_golang_pseudo_version(
+            module_major_version=module_major_version, subpath=subpath
+        )
+
+    def _get_highest_semver_tag_on_current_commit(
+        self, major_versions_to_try: tuple[int, ...], subpath: Optional[str]
+    ) -> Optional[str]:
+        """Return the highest semver tag on the current commit."""
         for major_version in major_versions_to_try:
             # Get the highest semantic version tag on the commit with a matching major version
             tag_on_commit = self._get_highest_semver_tag(major_version, subpath=subpath)
@@ -888,8 +904,12 @@ class ModuleVersionResolver:
                 tag_on_commit.name if not subpath else tag_on_commit.name.replace(f"{subpath}/", "")
             )
 
-        log.debug("No semantic version tag was found on the commit %s", self._commit.hexsha)
+        return None
 
+    def _get_highest_reachable_semver_tag(
+        self, major_versions_to_try: tuple[int, ...], subpath: Optional[str]
+    ) -> Optional[str]:
+        """Return the pseudo-version using the highest reachable semver tag as a base."""
         # This logic is based on:
         # https://github.com/golang/go/blob/a23f9afd9899160b525dbc10d01045d9a3f072a0/src/cmd/go/internal/modfetch/coderepo.go#L511-L521
         for major_version in major_versions_to_try:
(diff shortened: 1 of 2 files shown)


cachi2/core/package_managers/gomod.py
```diff
@@ -660,228 +669,273 @@ def _should_vendor_deps(
...
-    repo = git.Repo(app_dir.root)
-    if update_tags:
+class ModuleVersionResolver:
+    """Resolves the versions of Go modules in a git repository."""
+
+    def __init__(self, repo_path: RootedPath):
+        """Initialize a ModuleVersionResolver for the repo located at repo_path."""
+        self._repo = git.Repo(repo_path)
+        self._commit = self._repo.commit(self._repo.rev_parse("HEAD").hexsha)
+        self._fetch_remote_tags()
```
chmeliik, line 679: I'm generally not a fan of `__init__`s that do too much, especially network IO.

And in this case, `_fetch_remote_tags` also has implications for build reproducibility, since the same repo+ref can get a different main module version on the next rebuild. Which is fine, it's always been that way, but I'd like to keep it explicit (it was at least a param before).

Maybe something like this
```python
version_resolver = ModuleVersionResolver(repo_path)
version_resolver.fetch_new_tags()

# or
version_resolver = ModuleVersionResolver.for_repo(repo_path, fetch_new_tags=True)
```

chmeliik (commented):
+1 on the approach :+1: 

Just 1 at-first-sight comment

cachi2/core/package_managers/gomod.py
brunoapimentel: +1 to removing the network call from `__init__`.

On the other hand, from a pure design perspective, one could instantiate the object and call `get_golang_version` without ever calling `fetch_remote_tags`, which would be a problem if done by accident.

Still, I can't think of a better way of handling this, maybe logging a warning if that happens? I don't think it's a major issue though, so probably just leave it as it is.

tests/unit/package_managers/test_gomod.py
```diff
@@ -1364,3 +1391,43 @@ def test_get_repository_name(mock_git_repo: Any, input_url: str) -> None:
...
+    remote_repo.create_tag("v1.0.0")
+
+    return remote_repo_path, local_repo_path
+
+
+def test_fetch_tags(repo_remote_with_tag: tuple[RootedPath, RootedPath]) -> None:
+    _, local_repo_path = repo_remote_with_tag
+    version_resolver = ModuleVersionResolver(local_repo_path)
+    version_resolver.fetch_remote_tags()
+    assert version_resolver._commit_tags == ["v1.0.0"]
```
brunoapimentel, line 1422: Maybe add a scenario in which we could also check the `_all_tags` attribute?

cachi2/core/package_managers/gomod.py
```diff
@@ -660,228 +670,272 @@ def _should_vendor_deps(
...
+            log.debug(
+                "Using the semantic version tag of %s for commit %s",
+                tag_on_commit.name,
+                self._commit.hexsha,
+            )
+
+            # We want to preserve the version in the "v0.0.0" format, so the subpath is not needed
+            return (
+                tag_on_commit.name if not subpath else tag_on_commit.name.replace(f"{subpath}/", "")
+            )
```
brunoapimentel, line 792: Since we're also doing a refactoring in this PR, another improvement suggestion is to break this piece here into a separate function (`_get_highest_semver_tag_on_current_commit`?), as well as the other piece that is currently L798-817 (`_get_highest_reachable_semver_tag`?). 

I always find it hard to read this function here, maybe breaking it a little would make it a little easier on the eyes. But again, just a suggestion, ignore it if you don't like it.

taylormadore:
Rebased on main

cachi2/core/package_managers/gomod.py
taylormadore: I'll take a look at this today :+1:

tests/unit/package_managers/test_gomod.py
```diff
@@ -1485,3 +1506,55 @@ def test_get_repository_name(mock_git_repo: Any, input_url: str) -> None:
...
+
+    return remote_repo_path, local_repo_path
+
+
+def test_fetch_tags(repo_remote_with_tag: tuple[RootedPath, RootedPath]) -> None:
+    _, local_repo_path = repo_remote_with_tag
+    version_resolver = ModuleVersionResolver(local_repo_path)
+    version_resolver.fetch_remote_tags()
+    assert version_resolver._commit_tags == ["v2.0.0"]
+    assert version_resolver._all_tags == ["v1.0.0", "v2.0.0"]
```
chmeliik, line 1549: ~~Can we also check that both properties are `[]` before we fetch the tags?~~

I think we can't because the properties are cached. That makes my suggestion to split `fetch_remote_tags()` from `__init__` more questionable than I thought :sweat_smile: 

Maybe fetch_remote_tags could reset the cached properties, or it really shouldn't be possible to create a VersionResolver without fetching tags

cachi2/core/package_managers/gomod.py
```diff
@@ -241,8 +242,9 @@ def _create_modules_from_parsed_data(
     main_module_dir: RootedPath,
     parsed_modules: Iterable[ParsedModule],
     modules_in_go_sum: frozenset[ModuleID],
+    version_resolver: "ModuleVersionResolver",
 ) -> list[Module]:
-    def _create_module(module: ParsedModule) -> Module:
+    def _create_module(version_resolver: "ModuleVersionResolver", module: ParsedModule) -> Module:
```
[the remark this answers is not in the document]
taylormadore: Fixed :+1:

tests/unit/package_managers/test_gomod.py
taylormadore: Added a classmethod. Let me know what you think :pray:

cachi2/core/package_managers/gomod.py
taylormadore: Good call, I think it's more readable now

cachi2/core/package_managers/gomod.py
```diff
@@ -28,6 +29,7 @@
 import pydantic
 import semver
 from packageurl import PackageURL
+from typing_extensions import Self
```
chmeliik, line 32: this import should be in an `if TYPE_CHECKING` block because typing-extensions are a dev dependency, not a runtime dependency

(they are a runtime dependency of typer and pydantic, but we shouldn't runtime-depend on indirect dependencies)

chmeliik (approved):
LGTM :rocket: 

just one detail about importing typing_extensions at runtime

cachi2/core/package_managers/gomod.py
chmeliik: (and Self will need to be "Self")

chmeliik (approved):

lkolacek (approved):
Looks great, Thanks Taylor :+1:

PR #301, opened 2023-09-01
Report missing hashes for npm sbom component

ejegrova:


tests/unit/package_managers/test_npm.py
```diff
@@ -977,6 +1081,7 @@ def test_resolve_npm_validation(
                     "purl": f"pkg:npm/foo@1.0.0?vcs_url={MOCK_REPO_VCS_URL}",
                     "bundled": False,
                     "dev": False,
+                    "missing_hash_in_file": [],
```
chmeliik, line 1084: Hmm, now that I'm looking at it, I think updating these tests would have made more sense. `test_get_sbom_components` doesn't really cover any real cases, these ones do. Removing the `integrity` in a few cases and keeping it in others should be enough, probably?

tests/unit/package_managers/test_npm.py
```diff
@@ -667,6 +762,7 @@ def test_get_purl_integrity_handling(
                     "version": "1.0.0",
                     "bundled": True,
                     "dev": False,
+                    "missing_hash_in_file": [],
```
chmeliik, line 765: It would also be good to test missing hashes here (the _generate_component_list function)

cachi2/core/package_managers/npm.py
```diff
@@ -45,6 +45,7 @@ class NpmComponentInfo(TypedDict):
     version: str
     dev: bool
     bundled: bool
+    missing_hash_in_file: list
```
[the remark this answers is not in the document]
ejegrova: fixed

tests/unit/package_managers/test_npm.py
```diff
@@ -897,7 +1025,6 @@ def test_resolve_npm_validation(
                     "bar": {
                         "version": "2.0.0",
                         "resolved": "https://registry.npmjs.org/bar/-/bar-2.0.0.tgz",
-                        "integrity": "sha512-JCB8C6SnDoQf",
```
chmeliik, line 900: Why is `"missing_hash_in_file": None` for this test even though the integrity is missing?

tests/unit/package_managers/test_npm.py
```diff
@@ -428,20 +429,112 @@ def test_get_packages(
...
+                            "version": "1.0.0",
+                            "resolved": "https://github.com/foo",
+                        },
+                    ),
+                ],
+                [
+                    {
+                        "name": "foo",
+                        "missing_hash_in_file": Path("path/to/node_modules/foo/package-lock.json"),
+                    },
```
chmeliik, line 450: I still find these tests confusing. `/path/to/node_modules/foo` is the path to the installed `foo` dependency. Why is missing_hash_in_file = `path/to/node_modules/foo/package-lock.json`?

tests/unit/package_managers/test_npm.py
ejegrova: I thought it's the correct behavior since it's not missing in `node_modules/bar` [here](https://github.com/containerbuildsystem/cachi2/pull/301/files/95fa45da3379071b12ad11e5446874a490ad8d8b#diff-628fe82bed25383be81718c4bf663b93c2aad092de52f56b3bb9303f24197391L893).

tests/unit/package_managers/test_npm.py
ejegrova: yup, that's nonsense, should I keep the changes? I think we don't need this anymore since it's tested in `test_resolve_npm` .

tests/unit/package_managers/test_npm.py
chmeliik: Ah, right, yeah that's correct. If that's intentional (having integrity in `packages` but not in `dependencies`), could you add a comment about that?

tests/unit/package_managers/test_npm.py
chmeliik: Does test_resolve_npm cover git dependencies (we don't want to report missing_hash) and https dependencies (we do)? If yes, that's enough IMO

tests/unit/package_managers/test_npm.py
ejegrova: yup, I did it on purpose, added a comment

tests/unit/package_managers/test_npm.py
ejegrova: there was some missing, added a test case

tests/unit/package_managers/test_npm.py
brunoapimentel: If the opposite happened (integrity is present in the `dependencies` key, but missing from `packages`), then we would still report it as missing, right? IIUC, we are completely ignoring the dependencies key in the PackageLock class.

Do we want to change one of the test cases here to cover this scenario as well?

tests/integration/test_data/npm_multiple_packages_lockfile3/bom.json
```diff
@@ -73,6 +73,22 @@
       "type": "library",
       "version": "1.1.1"
     },
+    {
+      "name": "assertion-error",
```
brunoapimentel, line 77: Do we have any dependency that is also present in first_pkg or second_pkg? It would be a good way to test we're only reporting missing hashes in `third_pkg`.

(`is-positive` appears in all pkgs, but since it is a git dependency, it does not have hashes)

tests/unit/package_managers/test_npm.py
lkolacek: Good point Bruno, I believe this scenario is worth it. +1

lkolacek:
LGTM  :+1:

taylormadore (approved):

lkolacek (approved):

tests/integration/test_data/npm_multiple_packages_lockfile3/bom.json
ejegrova: added sax as a dependency to `third_pkg`, this one is also in second_pkg

brunoapimentel (approved):

slimreaper35 (approved):

PR #309, opened 2023-09-06
Yarn high-level code

brunoapimentel:
This is an outline of how the code to resolve Yarn will look like. The goals we're aiming for are filing smaller and more focused implementation stories, being able to work on them in parallel and to agree upon a general direction on how the code should look like.

I think it's important to emphasize the code here is meant as a general guideline, and that the person that is implementing the separate stories is free to change whatever he/she wants from it.

## Design principles
Some design principles I tried to follow:

- Offload the main logic to functions, use model classes only to represent the relevant data and offer minor utility methods. (anemic models, as opposed to domain rich models)
- Keep the models immutable whenever possible
- Try to keep the functions small and focused
- Try to make use of Python modules to break the large code base into smaller, more readable pieces
- Try to make the Python modules self-contained, exposing only a small and concise API

### DIFF
--- a/cachi2/core/package_managers/yarn/__init__.py
+++ b/cachi2/core/package_managers/yarn/__init__.py
@@ -1,0 +1,3 @@
+from cachi2.core.package_managers.yarn.main import fetch_yarn_source
+
+__all__ = ["fetch_yarn_source"]
--- a/cachi2/core/package_managers/yarn/locators.py
+++ b/cachi2/core/package_managers/yarn/locators.py
@@ -1,0 +1,115 @@
+from typing import NamedTuple, Union
+
+
+class FileLocator(NamedTuple):
+    """Locator that handles file, portal and link protocols.
+
+    Sample locator string:
+    name@protocol:path-to-local-file-or-folder::locator=project-name%40workspace%3Apath
+    """
+
+    @classmethod
+    def from_string(cls, locator: str) -> "FileLocator":
+        """Parse a Locator from a string."""
+        return NotImplemented
+
+    def __str__(self) -> str:
+        return NotImplemented
+
+
+class HttpsLocator(NamedTuple):
+    """Locator that handles remote files fetched by HTTPS.
+
+    Sample locator string:
+    name@https://domain.com/url/to/file.tar.gz
+    """
+
+    @classmethod
+    def from_string(cls, locator: str) -> "HttpsLocator":
+        """Parse a Locator from a string."""
+        return NotImplemented
+
+    def __str__(self) -> str:
+        return NotImplemented
+
+
+class NpmLocator(NamedTuple):
+    """Locator that handles registry dependencies.
+
+    Sample locator strings:
+    name@npm:version
+    @scope/name@npm:version
+    """
+
+    @property
+    def is_scoped(self) -> bool:
+        """Check if this package is scoped.
+
+        A scoped package follows the naming format of @scope/name.
+        """
+        return NotImplemented
+
+    def scope(self) -> str:
+        """Return the scope of this package."""
+        return NotImplemented
+
+    @classmethod
+    def from_string(cls, locator: str) -> "NpmLocator":
+        """Parse a Locator from a string."""
+        return NotImplemented
+
+    def __str__(self) -> str:
+        return NotImplemented
+
+
+class PatchLocator(NamedTuple):
+    """Locator that handles patched dependencies.
+
+    Sample locator string:
+    name@patch:patched-dependency@npm%3A0.0.1#./path/to/patch::version=1.0.0&hash=abc123&locator=package-name%40workspace%3A
+    name@patch:patched-dependency@npm%3A0.0.1#built-in<compat/patch-name>::version=1.0.0&hash=abc123 (built-in patch)
+    """
+
+    @classmethod
+    def from_string(cls, locator: str) -> "PatchLocator":
+        """Parse a Locator from a string."""
+        return NotImplemented
+
+    def __str__(self) -> str:
+        return NotImplemented
+
+
+class WorkspaceLocator(NamedTuple):
+    """Locator that handles workspace dependencies.
+
+    Sample locator string:
+    name@workspace:path/to/dir
+    """
+
+    @classmethod
+    def from_string(cls, locator: str) -> "WorkspaceLocator":
+        """Parse a Locator from a string."""
+        return NotImplemented
+
+    def __str__(self) -> str:
+        return NotImplemented
+
+
+Locator = Union[
+    FileLocator,
+    HttpsLocator,
+    NpmLocator,
+    PatchLocator,
+    WorkspaceLocator,
+]
+
+
+def parse_locator(locator: str) -> Locator:
+    """Parse a Locator object based on a locator string from the 'yarn info' command.
+
+    :raises PackageRejected: if a locator can't be parsed, or if it contains an protocol that is
+        not supported.
+    """
+    # we should raise a different type of error for unsparseable/unknown locators, and banned
+    # locators (such as ones that resolve to a Git dependency or containing the exec protocol)
+    return NotImplemented
--- a/cachi2/core/package_managers/yarn/main.py
+++ b/cachi2/core/package_managers/yarn/main.py
@@ -1,0 +1,102 @@
+import logging
+
+from cachi2.core.models.input import Request
+from cachi2.core.models.output import Component, EnvironmentVariable, RequestOutput
+from cachi2.core.package_managers.yarn.project import Project
+from cachi2.core.package_managers.yarn.resolver import (
+    create_component_from_package,
+    resolve_packages,
+)
+from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+
+def fetch_yarn_source(request: Request) -> RequestOutput:
+    """Process all the yarn source directories in a request."""
+    components = []
+
+    for package in request.yarn_packages:
+        path = request.source_dir.join_within_root(package.path)
+        project = Project.from_source_dir(path)
+        components.extend(_resolve_yarn_project(project, request.output_dir))
+
+    return RequestOutput.from_obj_list(
+        components, _generate_environment_variables(), project_files=[]
+    )
+
+
+def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Component]:
+    """Process a request for a single yarn source directory.
+
+    :param project: the directory to be processed.
+    :param output_dir: the directory where the prefetched dependencies will be placed.
+    """
+    _configure_yarn_version(project)
+
+    try:
+        _set_yarnrc_configuration(project, output_dir)
+        packages = resolve_packages(project.source_dir)
+
+        if project.is_zero_installs:
+            _check_yarn_cache(project.source_dir)
+        else:
+            _fetch_dependencies(project.source_dir, output_dir)
+    finally:
+        _undo_changes(project)
+
+    return [create_component_from_package(package, project) for package in packages]
+
+
+def _configure_yarn_version(project: Project) -> None:
+    """Resolve the yarn version and set it in the package.json file if needed.
+
+    :raises PackageRejected: in case the yarn version can't be determined, or if there is a
+        mismatch between the version in package.json and yarnrc.yml.
+    """
+    pass
+
+
+def _set_yarnrc_configuration(project: Project, output_dir: RootedPath) -> None:
+    """Set all the necessary configuration in yarnrc for the project processing.
+
+    :param project: the configuration changes dependending on if the project uses the zero-installs
+        or the regular workflow.
+    :param output_dir: in case the dependencies need to be fetched, this is where they will be
+        downloaded to.
+    """
+    # the plugins should be disabled here regardless of the project workflow.
+    pass
+
+
+def _check_yarn_cache(source_dir: RootedPath) -> None:
+    """Check the contents of the yarn cache using 'yarn install'.
+
+    :param source_dir: the directory in which the yarn command will be called.
+    :raises SubprocessCallError: if the 'yarn install' command fails.
+    """
+    # the yarn commands can be called by using the core.utils.run_cmd function
+    pass
+
+
+def _fetch_dependencies(source_dir: RootedPath, output_dir: RootedPath) -> None:
+    """Fetch dependencies using 'yarn install'.
+
+    :param source_dir: the directory in which the yarn command will be called.
+    :param output_dir: the directory where the yarn dependencies will be downloaded to.
+    :raises SubprocessCallError: if the 'yarn install' command fails.
+    """
+    # the output_dir is where the globalFolder must be set to
+    # the yarn commands can be called by using the core.utils.run_cmd function
+    pass
+
+
+def _undo_changes(project: Project) -> None:
+    """Undo any changes that were made to the files during the request's processing."""
+    # restore the disabled plugins here, as well as undo any additional changes
+    pass
+
+
+def _generate_environment_variables() -> list[EnvironmentVariable]:
+    """Generate environment variables that will be used for building the project."""
+    return NotImplemented
--- a/cachi2/core/package_managers/yarn/project.py
+++ b/cachi2/core/package_managers/yarn/project.py
@@ -1,0 +1,128 @@
+"""
+Parse the relevant files of a yarn project.
+
+It also provides basic utility functions. The main logic to resolve and prefetch the dependencies
+should be implemented in other modules.
+"""
+from typing import Any, NamedTuple, Optional
+
+from cachi2.core.rooted_path import RootedPath
+
+
+class YarnRc:
+    """A yarnrc file.
+
+    This class abstracts the underlying attributes and only exposes what
+    is relevant for the request processing.
+    """
+
+    def __init__(self, path: RootedPath, data: dict[str, Any]) -> None:
+        """Initialize a YarnRc object.
+
+        :param path: the path to the yarnrc file, relative to the request source dir.
+        :param data: the raw data for the yarnrc file.
+        """
+        self._path = path
+        self._data = data
+
+    @property
+    def cache_path(self) -> str:
+        """Get the configured location for the yarn cache folder.
+
+        Fallback to the default path in case the configuration key is missing.
+        """
+        return NotImplemented
+
+    @property
+    def registry_server(self) -> str:
+        """Get the globally configured registry server.
+
+        Fallback to the default server in case the configuration key is missing.
+        """
+        return NotImplemented
+
+    @property
+    def yarn_path(self) -> Optional[str]:
+        """Path to the yarn script present in this directory."""
+        return NotImplemented
+
+    @property
+    def yarn_version(self) -> Optional[str]:
+        """Yarn version used in this project.
+
+        Extracted from the contents of the yarnPath configuration, if it is present.
+        """
+        return NotImplemented
+
+    def registry_server_for_scope(self, scope: str) -> str:
+        """Get the configured registry server for a scoped package.
+
+        Fallback to the global defined registry server if there's no configuration for this specific
+        scope.
+
+        See: https://v3.yarnpkg.com/configuration/yarnrc#npmScopes
+        """
+        return NotImplemented
+
+    @classmethod
+    def from_file(cls, file_path: RootedPath) -> "YarnRc":
+        """Parse the content of a yarnrc file."""
+        return NotImplemented
+
+
+class PackageJson:
+    """A package.json file.
+
+    This class abstracts the underlying attributes and only exposes what
+    is relevant for the request processing.
+    """
+
+    def __init__(self, path: RootedPath, data: dict[str, Any]) -> None:
+        """Initialize a PackageJson object.
+
+        :param path: the path to the package.json file.
+        :param data: the raw data for the package.json file.
+        """
+        self._path = path
+        self._data = data
+
+    @property
+    def package_manager(self) -> Optional[str]:
+        """Get the package manager string."""
+        return NotImplemented
+
+    @classmethod
+    def from_file(cls, file_path: RootedPath) -> "PackageJson":
+        """Parse the content of a package.json file."""
+        return NotImplemented
+
+
+class Project(NamedTuple):
+    """A directory containing yarn sources."""
+
+    source_dir: RootedPath
+    yarn_rc: YarnRc
+    package_json: PackageJson
+
+    @property
+    def is_zero_installs(self) -> bool:
+        """If a project is using the zero-installs workflow or not.
+
+        This is determined by the existence of a non-empty yarn cache folder. For more details on
+        zero-installs, see: https://v3.yarnpkg.com/features/zero-installs.
+        """
+        return NotImplemented
+
+    @property
+    def yarn_cache(self) -> RootedPath:
+        """The path to the yarn cache folder.
+
+        The cache location is affected by the cacheFolder configuration in yarnrc. See:
+        https://v3.yarnpkg.com/configuration/yarnrc#cacheFolder.
+        """
+        return NotImplemented
+
+    @classmethod
+    def from_source_dir(cls, source_dir: RootedPath) -> "Project":
+        """Create a Project from a sources directory path."""
+        return NotImplemented
--- a/cachi2/core/package_managers/yarn/resolver.py
+++ b/cachi2/core/package_managers/yarn/resolver.py
@@ -1,0 +1,112 @@
+"""
+Resolve the dependency list for a yarn project.
+
+It also performs the necessary validations to avoid allowing an invalid project to keep being
+processed.
+"""
+import logging
+from typing import NamedTuple
+
+from cachi2.core.models.sbom import Component
+from cachi2.core.package_managers.yarn.locators import Locator
+from cachi2.core.package_managers.yarn.project import Optional, Project
+from cachi2.core.package_managers.yarn.utils import run_yarn_cmd
+from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+
+class Cache(NamedTuple):
+    """Cache information for a package."""
+
+    checksum: str
+    path: str
+
+
+class Package(NamedTuple):
+    """A package listed by the yarn info command.
+
+    See the output for 'yarn info -AR --json --cache'.
+
+    Note that the attributes present in the json vary a little depending on the type of
+    dependency.
+    """
+
+    # cache seems to be optional for portal, link and workspace protocols.
+    # 'yarn info' returns a cache object with empty keys.
+    cache: Optional[Cache]
+    locator: Locator
+    version: str
+
+    @classmethod
+    def from_info_string(cls, info: str) -> "Package":
+        """Create a Package from the output of yarn info."""
+        # this should use the locators.parse_locator function
+        return NotImplemented
+
+
+def resolve_packages(source_dir: RootedPath) -> list[Package]:
+    """Fetch and parse package data from the 'yarn info' output.
+
+    This function also performs a validation to ensure that the current yarn project can be
+    processed.
+
+    :raises PackageRejected: if the validation fails.
+    :raises SubprocessCallError: if the 'yarn info' command fails.
+    """
+    result = run_yarn_cmd([], source_dir, {})
+
+    # the result is not a valid json list, but a sequence of json objects separated by line breaks
+    packages = [Package.from_info_string(info) for info in result.splitlines()]
+
+    _vet_git_dependencies(packages)
+
+    return packages
+
+
+def create_component_from_package(package: Package, project: Project) -> Component:
+    """Create a SBOM component from a yarn Package."""
+    # if the SBOM generation code grows too much, it may be a good idea to split it into a dedicated
+    # module.
+
+    name = _resolve_package_name(package)
+
+    return Component(
+        name=name, version=package.version, purl=_generate_purl_for_package(package, name, project)
+    )
+
+
+def _vet_git_dependencies(packages: list[Package]) -> None:
+    """Stop the request processing if a Git dependency is found.
+
+    Git dependencies will cause the execution of existing js lifecycle scripts when 'yarn install'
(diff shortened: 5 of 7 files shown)


cachi2/core/package_managers/yarn/project.py
```diff
@@ -0,0 +1,129 @@
...
+    @classmethod
+    def from_file(cls, file_path: RootedPath) -> "PackageJson":
+        """Parse the content of a package.json file."""
+        with file_path.path.open("r") as file:
+            data = json.load(file)
+
+        return PackageJson(file_path, data)
+
+
+class Project(NamedTuple):
```
brunoapimentel, line 108: We are not calling a "main input package" a project anywhere else, so I struggled a little with this name here. But I kept it for now, let me know if you have a better suggestion for it.

chmeliik:
> * Offload the main logic to functions, use model classes only to represent the relevant data and offer minor utility methods. (anemic models, as opposed to domain rich models)
> * Keep the models immutable whenever possible

:heart:

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    Note that the attributes present in the json vary a little depending on the
+    type of dependency.
+    """
+
+    locator: str
+    version: str
+    dependencies: list[Dependency]
```
chmeliik, line 56: I don't really think we will have a use for the `dependencies` unless we decide to implement dev-dependency detection

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,197 @@
...
+    @property
+    def name_in_locator(self) -> str:
+        """Name of the package as parsed from its locator string."""
+        parts = self.locator.split("@")
+
+        # aliased package, starts with @
+        if parts[0] == "":
+            return "@" + parts[1]
+
+        return parts[0]
```
chmeliik, line 80: I'd be careful about this, it's only useful for registry dependencies

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,197 @@
...
+
+    if package.protocol in ["file"]:
+        return _resolve_remote_file_dependency_name(package)
+
+    if package.protocol in ["portal", "link", "workspace"]:
+        return _resolve_local_file_dependency_name(package)
+
+    if package.protocol in ["git", "github"]:
+        return _resolve_git_dependency_name(package)
+
```
chmeliik, line 151: Ah, I see. We don't need to custom-fetch them, we can just call `yarn install` and read the package.jsons in the zip files (`yarn info` tells you where those zip files are)

chmeliik (commented):
Overall direction LGTM, love the Project module+model :rocket:

cachi2/core/package_managers/yarn/main.py
```diff
@@ -0,0 +1,77 @@
...
+
+def fetch_yarn_source(request: Request) -> RequestOutput:
+    """Process a request for a Yarn source directory."""
+    project = Project.from_source_dir(request.source_dir)
+    _configure_yarn_version(project)
+
+    try:
+        _prepare_for_prefetch(project.yarn_rc, request.output_dir)
+        components = resolve_dependency_info(project.source_dir)
+        _fetch_dependencies(project)
```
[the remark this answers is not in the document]
chmeliik: related to https://github.com/containerbuildsystem/cachi2/pull/309#discussion_r1318317509

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,197 @@
...
+    name = _resolve_package_name(package)
+
+    return Component(
+        name=name,
+        version=package.version,
+        purl=_generate_purl_for_package(package, name),
+    )
+
+
+def _generate_purl_for_package(package: Package, name: str) -> str:
```
chmeliik, line 191: purl generation will need to respect YarnRc ([npmRegistryServer](https://yarnpkg.com/configuration/yarnrc#npmRegistryServer) and [npmScopes](https://yarnpkg.com/configuration/yarnrc#npmScopes))

cachi2/core/package_managers/yarn/main.py
```diff
@@ -0,0 +1,77 @@
...
+from cachi2.core.package_managers.yarn.resolver import resolve_dependency_info
+from cachi2.core.package_managers.yarn.utils import run_yarn_cmd
+from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+
+def fetch_yarn_source(request: Request) -> RequestOutput:
+    """Process a request for a Yarn source directory."""
+    project = Project.from_source_dir(request.source_dir)
```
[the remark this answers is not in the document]
brunoapimentel: Fixed.

cachi2/core/package_managers/yarn/main.py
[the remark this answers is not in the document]
brunoapimentel: I split the resolving of components into two parts: 
- resolve the `yarn info` output and return the list of Packages
- fetch the dependencies
- use the list of Packages to generate the components (while resolving their real names)

cachi2/core/package_managers/yarn/main.py
```diff
@@ -0,0 +1,77 @@
...
+    except Exception:
+        _restore_files(project)
+        raise
+
+    _restore_files(project)
+
+    return RequestOutput.from_obj_list(
+        components,
+        _generate_environment_variables(),
+        _generate_project_files(),
```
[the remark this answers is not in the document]
brunoapimentel: Removed.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -0,0 +1,129 @@
...
+from typing import Any, NamedTuple, Optional, TypedDict
+
+from yaml import safe_load
+
+from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+
+_YARNRC_VERSION_REGEX = re.compile("yarn-[0-9].[0-9].[0-9].cjs")
```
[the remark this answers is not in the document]
brunoapimentel: I removed it to avoid making the PR large and long. We can deal with it during the actual implementation stories.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -0,0 +1,129 @@
...
+        """Determined by the existence of a non-empty .yarn/cache folder."""
+        return self.yarn_cache.path.exists()  # && not empty
+
+    @classmethod
+    def from_source_dir(cls, source_dir: RootedPath) -> "Project":
+        """Create a Project from a sources directory path."""
+        # Handle errors: file exists? malformed?
+        yarn_rc = YarnRc.from_file(source_dir.join_within_root(".yarnrc.yml"))
+        package_json = PackageJson.from_file(source_dir.join_within_root("package.json"))
+        yarn_cache = source_dir.join_within_root(".yarn/cache")
```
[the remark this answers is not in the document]
brunoapimentel: > is it always .yml? Does yarn also look for .yaml?
Yes. Apparently yarn ignores a `.yarnrc.yaml` file completely.

> yarn_cache needs to respect YarnRc.cache_path
Fixed.

cachi2/core/package_managers/yarn/resolver.py
brunoapimentel: Removed.

cachi2/core/package_managers/yarn/resolver.py
brunoapimentel: Fixed it. See https://github.com/containerbuildsystem/cachi2/pull/309#discussion_r1318218711.

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,197 @@
...
+from typing import NamedTuple, TypedDict
+
+from cachi2.core.models.sbom import Component
+from cachi2.core.package_managers.yarn.utils import run_yarn_cmd
+from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+
+_PACKAGE_PROTOCOL_REGEX = re.compile("@[a-z]+:")
```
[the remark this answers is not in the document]
brunoapimentel: Removed this, and it should be handled by the Locator models now.

cachi2/core/package_managers/yarn/locators.py
```diff
@@ -0,0 +1,236 @@
+from typing import NamedTuple, Protocol
+
+from cachi2.core.errors import PackageRejected
+
+
+class Locator(Protocol):
```
brunoapimentel, line 6: I'm mostly fiddling with Python ~interfaces~ protocols here, I'm not sure I like them here yet.

brunoapimentel: I decided to leave the purl generation out of the locator models though, since they would rely on external attributes (actual name from package.json, main module version for local files/worskpaces).

Another reason is that I'm seeing these as mostly parsers for the string locators, so they shouldn't have knowledge about purl generation.

cachi2/core/package_managers/yarn/resolver.py
brunoapimentel: I changed the YarnRc class to have this info, but I still need to improve the purl generation part.

brunoapimentel:
Next iteration of this draft. I would mostly focus on:
- Locator models
- Changes to the `fetch_yarn_source` function (yarn/main.py)

I still think the code in `resolver.py` will have some changes, specially in the purl handling part (wondering if we should factor that into a `sbom.py` module).

cachi2/core/package_managers/yarn/resolver.py
chmeliik: We can leave that for later, IMO for now the important thing is to give _generate_purl_for_package a yarn_rc param and explain why it needs it

cachi2/core/package_managers/yarn/project.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @property
+    def is_zero_installs(self) -> bool:
+        """Determined by the existence of a non-empty .yarn/cache folder."""
+        return self.yarn_cache.path.exists()  # && not empty
+
+    @classmethod
+    def from_source_dir(cls, source_dir: RootedPath) -> "Project":
+        """Create a Project from a sources directory path."""
+        # Handle errors: file exists? malformed?
```
chmeliik, line 148: TODO x2

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,197 @@
...
+    See 'yarn info' output. Example:
+
+    {
+        "descriptor": "yargs@npm:^17.7.2",
+        "locator": "yargs@npm:17.7.2"
+    }
+    """
+
+    descriptor: str
+    locator: str
```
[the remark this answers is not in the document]
chmeliik: > I think we should try to settle on the class attributes though

I don't know if we can without reading every Yarnberry Resolver

cachi2/core/package_managers/yarn/locators.py
chmeliik: I'm not sure if the Locator Protocol is useful at all

`name` is only useful for one particular Locator type, `protocol` probably isn't useful for any and `from_string` doesn't need to be part of it since parse_locator will work either way

I think a good old Union would the best. We have a fixed set of known locators, so the type might as well reflect that (as opposed to a protocol which can have infinitely many implementers, so you can't do exhaustiveness checking on it)

cachi2/core/package_managers/yarn/locators.py
```diff
@@ -0,0 +1,236 @@
...
+    if "@patch:" in locator:
+        return PatchLocator.from_string(locator)
+
+    if "@workspace:" in locator:
+        return WorkspaceLocator.from_string(locator)
+
+    raise PackageRejected(
+        reason=f'Locator "{locator}" couldn\'t be parsed.',
+        solution="Cachi2 only supports the default Yarn protocols: https://v3.yarnpkg.com/features/protocols.",
+    )
```
chmeliik, line 236: All the parsing methods and parse_locator should be TODO or removed as well IMO

I would prefer removed to avoid leading the implementer down the wrong path (we really, really should check how yarnberry parses them)

cachi2/core/package_managers/yarn/project.py
```diff
@@ -0,0 +1,129 @@
...
+class Plugin(TypedDict):
+    """A plugin defined in a yarnrc.yml file.
+
+    Example:
+    - path: .yarn/plugins/@yarnpkg/plugin-exec.cjs
+      spec: "@yarnpkg/plugin-exec"
+    """
+
+    path: str
+    spec: str
```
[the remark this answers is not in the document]
chmeliik: bump so it's not forgotten

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,116 @@
...
+    def from_info_string(cls, info: str) -> "Package":
+        """Create a Package from the output of yarn info."""
+        data = json.loads(info)
+        cache = data["children"]["Cache"]
+
+        return Package(
+            cache=Cache(checksum=cache["Checksum"], path=cache["Path"]),
+            locator=parse_locator(data["value"]),
+            version=data["children"]["Version"],
+        )
```
chmeliik, line 49: TODO: is children.Version, children.Cache.Checksum, children.Cache.Path always present and non-null?

cachi2/core/package_managers/yarn/locators.py
```diff
@@ -0,0 +1,236 @@
...
+    name: str = "n/a"
+    protocol: str = "exec"
+
+    @classmethod
+    def from_string(cls, locator: str) -> "ExecLocator":
+        """Parse a Locator from a string."""
+        raise PackageRejected(
+            reason="The exec protocol is not supported by Cachi2, since it would result in arbitrary code execution.",
+            solution="All dependencies that make use of the exec protocol must be removed from the package.json file.",
+        )
```
chmeliik, line 47: I was wondering why `resolve_dependency_info` has a check for git but not for exec and then found the check here.

Is this the right place for it?

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,116 @@
...
+def _resolve_package_name(package: Package) -> str:
+    """Resolve the package's real name based on its protocol.
+
+    Non-registry deps might have names in their locator string that
+    differ from the one in their package.json file.
+
+    Look at the package.json name for every non-registry dependency.
+    """
+    if package.locator.protocol == "npm":
+        return package.locator.name
```
chmeliik, line 100: `if isinstance(package.locator, NpmLocator):` is probably better (otherwise you have to give `protocol` and `name` to every locator type even when it doesn't make sense)

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,116 @@
...
+    return package.locator.name
+
+
+def _generate_purl_for_package(package: Package, name: str) -> str:
+    """Create a purl for a package based on its protocol."""
+    if package.locator.protocol == "npm":
+        return f"pkg:npm/{package.locator.name}@{package.version}"
+    # add specific implementations for every supported protocol
+    else:
+        return f"pkg:npm/{name}@{package.version}?url={package.locator}"
```
chmeliik, line 116: TODO: normalization
TODO: use the PackageUrl class

chmeliik:
I think we already have too many implementation details here, no more please :smile:

brunoapimentel:
> I think we already have too many implementation details here, no more please 😄

I pushed a WIP commit so you can see the changes I'm doing. I initially took a more radical approach and removed pretty much all implementation details and even some lower-level functions, but I thought it was probably more confusing that way.

LMK if the WIP is good enough. I can add a commit with the more radical approach if you want.

cachi2/core/package_managers/yarn/resolver.py
brunoapimentel: The initial reasoning was that a Locator could have more than one protocol (GitLocator: https and ssh, for instance). I thought this would affect the purl generation, but it seems that is not the case. 

My latest changes already went for you approach (although I used `type(package.locator)`, which is a worse way of dealing with it).

cachi2/core/package_managers/yarn/locators.py
brunoapimentel: I added a comment to explain my reasoning in the WIP commit. I think the error message should differentiate locators that we can't parse vs the ones we ban.

Not 100% happy of having it here either. Alternatively, we could parse them all and only then ban the ones we forbid in a `_vet_protocols` function.

chmeliik: But +1 to having a specific error message for the "known but won't even try to parse" locator types

chmeliik:
> LMK if the WIP is good enough. I can add a commit with the more radical approach if you want.

That seems fine

It's more TODOs than I even expected :sweat_smile: . What would we be checking for the "check if this approach is sufficient" ones?

cachi2/core/package_managers/yarn/main.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def _restore_files(project: Project) -> None:
+    """Restore the files that were changed during the prefetch."""
+    pass
+
+
+def _generate_environment_variables() -> list[EnvironmentVariable]:
+    """Generate environment variables that will be used for building the repository that is being processed."""
+    return NotImplemented
```
chmeliik, line 84: Would we want to set YARN_GLOBAL_MIRROR even if all the projects are zero-installs and $output_dir/deps/yarnberry is empty? (If that's how we handle zero-installs)? I guess it doesn't really hurt anything

cachi2/core/package_managers/yarn/main.py
```diff
@@ -0,0 +1,84 @@
...
+    pass
+
+
+def _fetch_dependencies(project: Project) -> None:
+    """Fetch dependencies using 'yarn install'.
+
+    :raises SubprocessCallError: if the 'yarn install' command fails.
+    """
+    run_yarn_cmd(["yarn", "--help"], project.source_dir)
+    pass
```
chmeliik, line 74: Maybe call out that for zero-installs, we're not fetching but verifying (although verifying does involve fetching because of `--check-cache`)

chmeliik (commented):
I find it easier to focus on the adherence to the design now that the implementations are gone :+1: 

In the overall workflow, we're ignoring zero-installs and plugin handling a bit

cachi2/core/package_managers/yarn/main.py
```diff
@@ -0,0 +1,84 @@
...
+    """Resolve the yarn version and set it in the package.json file if needed.
+
+    :raises PackageRejected: in case the yarn version can't be determined, or if
+    there is a mismatch between the version in package.json and yarnrc.yml.
+    """
+    pass
+
+
+def _prepare_for_prefetch(yarnrc: YarnRc, output_dir: RootedPath) -> None:
+    """Change the configuration in yarnrc to perform the prefetch.
```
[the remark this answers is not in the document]
brunoapimentel: The Yarn docs [state](https://v3.yarnpkg.com/configuration/yarnrc) that "...note that most settings can also be defined through environment variables (at least for the simpler ones; arrays and objects aren't supported yet)."

So I'd prefer to only set environment variables, but my fear is that we will end up with a mixed approach, since plugins is an array and can't be set via env vars. If we do all the configuration by changing yarnrc though, it should be enough.

Wdyt? We can probably start with "plugins in yarnrc, everthing else in env vars", and change it if more things need to be changed in yarnrc.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -0,0 +1,84 @@
...
+
+    :param output_dir: the directory where the prefetched dependencies will be placed.
```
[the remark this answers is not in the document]
brunoapimentel: Yeah, that is a good point. I think only setting it for no-zero-installs is the correct choice.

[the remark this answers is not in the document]
brunoapimentel: IIUIC, in the case of zero-installs, we won't prefetch anything, the content used for the build will be what is commited in the repo already, right? If the project has zero-installs configuration, but doesn't have the correct yarn cache commited, the request will fail.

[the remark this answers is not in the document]
chmeliik: That's the intention. But `yarn install --check-cache` will probably populate the globalMirror, so if we're not careful we will also inject that

cachi2/core/package_managers/yarn/main.py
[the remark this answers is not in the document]
chmeliik: That does sound good, though I wonder how far down you will have to pass the env vars dict before you can pass it to `run_yarn_cmd`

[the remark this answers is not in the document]
chmeliik: Maybe a common place (like a function?) for handling env vars would be good, otherwise they're liable to become a mess (they'll need to be used in `resolver.resolve_dependency_info`, `main._fetch_dependencies` and `main._check_yarn_cache` at least)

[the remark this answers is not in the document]
chmeliik: > Do we need to use the env vars in resolve_dependency_info, though?
> 
> IIUC, the only configuration that would affect it is enableGlobalCache, and we won't be setting it. Is there any other config that affects yarn info?

* globalFolder - you need `yarn info` to tell you what the cachePath is
* enableMirror - same
* ignorePath
* and, to keep things sane, everything you're going to set for `yarn install`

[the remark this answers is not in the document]
brunoapimentel: globalFolder will only change the cachePath if we also set enableGlobalCache, and I think we won't, right?

I do agree on the "keep things sane" bit anyways.

cachi2/core/package_managers/yarn/main.py
brunoapimentel: Do you mean YARN_ENABLE_MIRROR? It probably doesn't hurt (we will probably mount the ouput_dir in RHTAP regardless), but I think it is also easy enough to pass the project variable here and set it accordingly.

cachi2/core/package_managers/yarn/main.py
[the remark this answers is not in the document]
chmeliik: You're right that globalFolder and enableMirror probably don't matter unless you also set enableGobalCache, but I don't know if we'll be setting that or not

cachi2/core/package_managers/yarn/main.py
chmeliik: Both YARN_GLOBAL_MIRROR and YARN_ENABLE_MIRROR

We will mount the output dir at /cachi2, /cachi2/deps/yarnberry/ will be empty or nonexistent. We may or may not set `YARN_GLOBAL_MIRROR=/cachi2/deps/yarnberry` and `YARN_ENABLE_MIRROR=true`

Depending on what the user does in the Dockerfile, they might end up writing to / reading from the global mirror. Yarn should do the right thing when reading from an empty/non-existent global mirror. If Yarn writes something to the global mirror, that should be fine too - by that point, RHTAP will hopefully have copied the things it needs for the source container.

chmeliik: But anyway, from a high-level code design perspective, the only thing that would need to change is passing the Project :+1:

chmeliik: (And making sure that `Project.is_zero_installs` does the right thing even if you've already called `yarn install`)

chmeliik:
> This seems to make the code a little simpler since we don't need to have
a way to share the env vars among the different yarn cli invocations.
We'd need to double check if setting everything in yarnrc is enough.

The .yarnrc.yml option does seem simpler.

Or at least it hides the complexity, since we share configuration by modifying a file in a directory and running all the CLI commands from that directory. Since we can't avoid modifying that file (plugins), we might as well use it for everything

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,101 @@
...
+    return packages
+
+
+def create_component_from_package(package: Package, yarn_rc: YarnRc) -> Component:
+    """Create a SBOM component from a yarn Package."""
+    name = _resolve_package_name(package)
+
+    return Component(
+        name=name, version=package.version, purl=_generate_purl_for_package(package, name, yarn_rc)
+    )
```
chmeliik, line 70: Just remembered we're forgetting the `pedigree.patches` for patch: dependencies, but that will require changes to the Component model as well

And from the code "interface" side of things, that probably just means we need to pass `Project` or at least its `source_dir` to this function so that it can handle file paths (in fact, I think we need that for all file-like dependencies)

cachi2/core/package_managers/yarn/main.py
```diff
@@ -35,13 +34,16 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Comp
     _configure_yarn_version(project)
 
     try:
-        _set_yarnrc_configuration(project, output_dir)
-        packages = resolve_dependency_info(project.source_dir)
+        _disable_plugins(project.yarn_rc)
+
+        cli = YarnCli(project.source_dir, _generate_yarn_env(project, output_dir))
+
+        packages = resolve_dependency_info(cli)
```
ben-alkov, line 41: This just looks weird to me (i.e. passing in a CLI object)...

ben-alkov: So, in other words, I agree with Taylor and Adam that using the rc file makes more sense :grin:

cachi2/core/package_managers/yarn/resolver.py
brunoapimentel: I added the project as a parameter to this function, and to the _generate_purl_for_package function (we'd probably need to split the latter into several functions, but I'll leave this to the implementation).

chmeliik (commented):
LGTM :+1: 

only nitpicks left from the "actually looking at the details" pass :smile:

cachi2/core/package_managers/yarn/main.py
```diff
@@ -0,0 +1,105 @@
...
+
+        if project.is_zero_installs:
+            _check_yarn_cache(project.source_dir)
+        else:
+            _fetch_dependencies(project.source_dir, output_dir)
+    except Exception:
+        _undo_changes(project)
+        raise
+
+    _undo_changes(project)
```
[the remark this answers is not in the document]
brunoapimentel: Stupid me had no clue a `try... finally` block was valid syntax... I though it needed to be `try... except... finally`.

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -0,0 +1,110 @@
...
+from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+
+class Cache(NamedTuple):
+    """Cache information for a package."""
+
+    checksum: str
+    path: str
```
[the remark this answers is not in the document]
brunoapimentel: Added a commentary to remind us of this, and made `Package.cache` optional instead (seemed nicer :grin: )

cachi2/core/package_managers/yarn/__init__.py
```diff
@@ -0,0 +1 @@
+from cachi2.core.package_managers.yarn.main import fetch_yarn_source  # noqa
```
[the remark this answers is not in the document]
brunoapimentel: I was wondering about this. But `mypy` complains that:
`cachi2/core/resolver.py:15: error: Module has no attribute "fetch_yarn_source"  [attr-defined]`

cachi2/core/package_managers/yarn/resolver.py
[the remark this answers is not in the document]
chmeliik: true, there's probably never a case where only one would be null :+1:

cachi2/core/package_managers/yarn/__init__.py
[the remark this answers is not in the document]
chmeliik: I don't see it :thinking: 

```diff
diff --git a/cachi2/core/package_managers/yarn/__init__.py b/cachi2/core/package_managers/yarn/__init__.py
index cd66705..e2374b6 100644
--- a/cachi2/core/package_managers/yarn/__init__.py
+++ b/cachi2/core/package_managers/yarn/__init__.py
@@ -1 +1,4 @@
-from cachi2.core.package_managers.yarn.main import fetch_yarn_source  # noqa
+from cachi2.core.package_managers.yarn.main import fetch_yarn_source
+
+
+__all__ = ["fetch_yarn_source"]
diff --git a/cachi2/core/resolver.py b/cachi2/core/resolver.py
index 7bbc8a9..ce3f99d 100644
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -4,7 +4,7 @@ from typing import Callable
 from cachi2.core.errors import UnsupportedFeature
 from cachi2.core.models.input import PackageManagerType, Request
 from cachi2.core.models.output import RequestOutput
-from cachi2.core.package_managers import gomod, npm, pip
+from cachi2.core.package_managers import gomod, npm, pip, yarn
 
 Handler = Callable[[Request], RequestOutput]
 
@@ -12,6 +12,7 @@ _package_managers: dict[PackageManagerType, Handler] = {
     "gomod": gomod.fetch_gomod_source,
     "npm": npm.fetch_npm_source,
     "pip": pip.fetch_pip_source,
+    "yarn": yarn.fetch_yarn_source,
 }
```

```bash
$ tox -e mypy
Success: no issues found in 65 source files
```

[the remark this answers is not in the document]
chmeliik: (I meant that the way to re-export is to import the thing _and_ add the `__all__` declaration)

chmeliik (approved):
LGTM :rocket:

cachi2/core/package_managers/yarn/__init__.py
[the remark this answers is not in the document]
brunoapimentel: Yeah, I did remove the import, so that was the mistake.

taylormadore (approved):
Looks amazing :tada:

lkolacek (approved):
It's outstanding! :blush: :rocket:

PR #312, opened 2023-09-08
cachi2 yarn: test cases

slimreaper35:
STONEBLD-1686

### DIFF
--- a/tests/integration/test_yarn.py
+++ b/tests/integration/test_yarn.py
@@ -1,0 +1,61 @@
+import logging
+from pathlib import Path
+
+import pytest
+
+from . import utils
+
+log = logging.getLogger(__name__)
+
+
+@pytest.mark.parametrize(
+    "test_params",
+    [
+        pytest.param(
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-yarn-berry.git",
+                ref="4b77c878694e656e7d0ceac6771131273f98f5df",
+                packages=({"path": ".", "type": "yarn"},),
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="All dependencies fetched successfully",
+            ),
+            marks=pytest.mark.xfail,  # temporary
+            id="yarn_zero_installs",
+        ),
+        pytest.param(
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-yarn-berry.git",
+                ref="452d7f530513206de63af402536e9ff736c2aa79",
+                packages=({"path": ".", "type": "yarn"},),
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="All dependencies fetched successfully",
+            ),
+            marks=pytest.mark.xfail,  # temporary
+            id="yarn_no_zero_installs",
+        ),
+    ],
+)
+def test_yarn_packages(
+    test_params: utils.TestParameters,
+    cachi2_image: utils.ContainerImage,
+    tmp_path: Path,
+    test_data_dir: Path,
+    request: pytest.FixtureRequest,
+) -> None:
+    """
+    Test fetched dependencies for yarn 2+.
+
+    :param test_params: Test case arguments
+    :param tmp_path: Temp directory for pytest
+    """
+    test_case = request.node.callspec.id
+
+    source_folder = utils.clone_repository(
+        test_params.repo, test_params.ref, f"{test_case}-source", tmp_path
+    )
+
+    utils.fetch_deps_and_check_output(
+        tmp_path, test_case, test_params, source_folder, test_data_dir, cachi2_image
+    )


slimreaper35:
Temporary repository for testing is [here](https://github.com/slimreaper35/yarnberry-tmp). I will clone that to cachito-testing.

slimreaper35:
Testing repo is [here](https://github.com/cachito-testing/cachi2-yarn-berry).

tests/integration/test_yarn.py
```diff
@@ -0,0 +1,55 @@
...
+        ),
+    ],
+)
+def test_yarn_packages(
+    test_params: utils.TestParameters,
+    cachi2_image: utils.ContainerImage,
+    tmp_path: Path,
+    test_data_dir: Path,
+    request: pytest.FixtureRequest,
+) -> None:
```
ben-alkov, line 46: A docstring suggestion would be nice

ben-alkov:
@slimreaper35; Is this a W.I.P.? i.e. will you be adding more test scenarios?

slimreaper35:
Probably not, there is another [story](https://issues.redhat.com/browse/STONEBLD-1687) for that, which should cover e2e test as well.

brunoapimentel:
https://github.com/cachito-testing/cachi2-yarn-berry/blob/offline-cache/package.json#L23

According to the design, we will ban any dependency with the `exec` protocol, so this would make any request targeting this branch fail. We will probably want to cover these "fail" scenarios, the question is if we want it in this PR.

This will likely also happen to Git dependencies, but I think we should keep them until this is settled.

tests/integration/test_yarn.py
```diff
@@ -0,0 +1,55 @@
...
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-yarn-berry.git",
+                ref="a37898af6d5f2fa4335e179444fad1ca7ad74613",
+                packages=({"path": ".", "type": "yarn"},),
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="All dependencies fetched successfully",
+            ),
+            marks=pytest.mark.xfail,  # temporary
+            id="zero-installs",
```
chmeliik, line 36: Let's follow the same `id` naming as other integration tests: `yarn_{test_case}` or `yarnberry_{test_case}`. This makes it so that you can run all the yarn tests with `tox -e integration -- -k yarn`

tests/integration/test_yarn.py
```diff
@@ -0,0 +1,55 @@
...
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-yarn-berry.git",
+                ref="76131443f7472916b316954c957259ee4cee97a7",
+                packages=({"path": ".", "type": "yarn"},),
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="All dependencies fetched successfully",
+            ),
+            marks=pytest.mark.xfail,  # temporary
+            id="offline-cache",
```
chmeliik, line 24: This isn't an [offline cache](https://v3.yarnpkg.com/features/offline-cache) test case

The offline cache is the .yarn/cache directory which contains all the dependencies. The second test case has the offline cache, this one doesn't. (The second one also has the .pnp.cjs file. Offline cache + .pnp.cjs = zero-installs.)

tests/integration/test_yarn.py
```diff
@@ -0,0 +1,55 @@
...
+log = logging.getLogger(__name__)
+
+
+@pytest.mark.parametrize(
+    "test_params",
+    [
+        pytest.param(
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-yarn-berry.git",
+                ref="76131443f7472916b316954c957259ee4cee97a7",
```
chmeliik, line 17: In the READMEs, could you explain which features the test cases cover?

chmeliik: The commit history might be helpful for that https://github.com/cachito-testing/cachi2-yarn-berry/commits/offline-cache

chmeliik (approved):
LGTM, just the description in https://github.com/cachito-testing/cachi2-yarn-berry/tree/no-zero-installs looks incomplete

> This integration test is to ensure that the regular workflow of Yarnberry.

slimreaper35:
updated commit hash

brunoapimentel:
Typo in "workflow": https://github.com/cachito-testing/cachi2-yarn-berry/blob/no-zero-installs/README.md?plain=1#L1

chmeliik:
> "the-answer" does not specify a version in its package.json [file](https://github.com/cachito-testing/cachi2-yarn-berry/blob/zero-installs/packages/the-answer/package.json), which means that `yarn info -AR --json --cache` resolves it as `0.0.0`.
> 
> Is this an intentional test case? (I think it should be 😅)

I did the bare minimum when preparing the tests, and for yarn the that does not include setting a `version` (so yeah, I think we do want to test that, because users will likely do the bare minimum as well)

chmeliik:
> Maybe it would be nice to have in the README of the test repo a brief note on all scenarios we're trying to cover here, just like we have for the [protocols](https://github.com/cachito-testing/cachi2-yarn-berry/blob/no-zero-installs/README.md?plain=1#L6). Some other things I noticed we also have in the test repo:
> 
>     * dev, optional and peer dependencies
> 
>     * workspaces
> 
>     * scoped dependencies
> 
>     * [pre|post]install scripts? (I see them in the no-zero-installs package.json)
> 
> 
> @chmeliik anything else worth noting?

Maybe these?

* https://github.com/cachito-testing/cachi2-yarn-berry/commit/4326fbbde1b1770e752138a9347e13fcfaabc9be
  * dependency not relevant for the current platform, we want to fetch it anyway
* https://github.com/cachito-testing/cachi2-yarn-berry/commit/4066baa6ff91ce6d4491370479d8b8c23ed7dd37
  * dependency that has to be compiled (the commit message says postinstall script, I think that's not accurate - I think yarnberry knows that compilation is needed based on the presence of [binding.gyp](https://github.com/tree-sitter/tree-sitter-json/blob/master/binding.gyp))

ejegrova (approved):

lkolacek:
Everything looks great :+1:, but I have 1 (maybe stupid) question. Project requested yarn 3.6.1 and `yarn install` on node18 container changed it to 3.6.3. Is this expected behavior? I would expect corepack to use 3.6.1 version which seems to be valid. Haven't found any useful info in yarn install log about this change.

lkolacek:
> Everything looks great 👍, but I have 1 (maybe stupid) question. Project requested yarn 3.6.1 and `yarn install` on node18 container changed it to 3.6.3. Is this expected behavior? I would expect corepack to use 3.6.1 version which seems to be valid. Haven't found any useful info in yarn install log about this change.

My bad, I updated yarn version via corepack.

lkolacek (approved):

slimreaper35:
So, what is the decision :smile: on README.md in the [repo](https://github.com/cachito-testing/cachi2-yarn-berry/tree/no-zero-installs) @everyone ?

chmeliik:
> So, what is the decision 😄 on README.md in the [repo](https://github.com/cachito-testing/cachi2-yarn-berry/tree/no-zero-installs) @everyone ?

Let's add the things we suggested (you can probably omit "workspaces", those are already mentioned in Protocols

slimreaper35:
@everyone PTAL - [repo](https://github.com/cachito-testing/cachi2-yarn-berry/tree/no-zero-installs) 
---
(rebased and updated commit hashes)

chmeliik (approved):

brunoapimentel (approved):

PR #255, opened 2023-07-25
pip wheels

slimreaper35:
JIRA: STONEBLD-1332

# Maintainers will complete the following section

- [x] Commit messages are descriptive enough
- [x] Code coverage from testing does not decrease and new code is covered
- n/a Docs updated (if applicable)
- n/a Docs links in the code are still valid (if docs were updated)

PR #319, opened 2023-09-18
cachi2 wheels: allow dependencies with only wheel distributions

slimreaper35:
If a user enables `allow_binary option`, reject only those distributions which have no source and no
wheel distributions

STONEBLD-1641

### DIFF
--- a/cachi2/core/package_managers/pip.py
+++ b/cachi2/core/package_managers/pip.py
@@ -14,17 +14,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from os import PathLike
 from pathlib import Path
-from typing import (
-    IO,
-    TYPE_CHECKING,
-    Any,
-    Iterable,
-    Iterator,
-    Literal,
-    Optional,
-    Union,
-    no_type_check,
-)
+from typing import IO, TYPE_CHECKING, Any, Iterable, Iterator, Optional, Union, cast, no_type_check
 
 import tomli
 from packageurl import PackageURL
@@ -1424,6 +1414,19 @@ def _download_dependencies(
             if allow_binary:
                 to_download.extend(w for w in wheels if not w.path.exists())
 
+            if source is None:
+                # at least one wheel exists -> report in the SBOM
+                downloaded.append(
+                    {
+                        "package": req.package,
+                        "version": req.version_specs[0][1],
+                        "kind": req.kind,
+                        "hash_verified": require_hashes,
+                        "requirement_file": str(requirements_file.file_path.subpath_from_root),
+                    }
+                )
+                continue
+
             download_binary_file(source.url, source.path, auth=None)
             _check_metadata_in_sdist(source.path)
             download_info = source.download_info
@@ -1657,7 +1660,7 @@ class DistributionPackageInfo:
 
     name: str
     version: str
-    package_type: Literal["sdist", "wheel"]
+    package_type: str
     path: Path
     url: str
     is_yanked: bool
@@ -1721,7 +1724,7 @@ class DistributionPackageInfo:
 
 def _process_package_distributions(
     requirement: PipRequirement, pip_deps_dir: RootedPath, allow_binary: bool = False
-) -> tuple[DistributionPackageInfo, list[DistributionPackageInfo]]:
+) -> tuple[Optional[DistributionPackageInfo], list[DistributionPackageInfo]]:
     name = requirement.package
     version = requirement.version_specs[0][1]
     normalized_version = canonicalize_version(version)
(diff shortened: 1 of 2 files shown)


cachi2/core/package_managers/pip.py
```diff
@@ -1424,6 +1424,9 @@ def _download_dependencies(
             if allow_binary:
                 to_download.extend(w for w in wheels if not w.path.exists())
 
+            if source is None:
+                continue
```
chmeliik, line 1428: Just remembered that this won't be that simple because of https://github.com/containerbuildsystem/cachi2/pull/255#discussion_r1305581494

We must report the downloaded dependency if we download at least 1 archive for it, we can't omit it just because it doesn't have sdists

cachi2/core/package_managers/pip.py
```diff
@@ -1424,6 +1424,19 @@ def _download_dependencies(
...
+                downloaded.append(
+                    {
+                        "package": req.package,
+                        "version": req.version_specs[0][1],
+                        "kind": req.kind,
+                        "hash_verified": False,
+                        "requirement_file": str(requirements_file.file_path.subpath_from_root),
+                    }
+                )
+                continue
```
slimreaper35, line 1438: That's kind of ugly but since there are more stories for cachi2-wheels, which unify handling both distributions, it should be temporary, or not ?

chmeliik: Hopefully yes :smile:

cachi2/core/package_managers/pip.py
```diff
@@ -1424,6 +1424,19 @@ def _download_dependencies(
...
                 to_download.extend(w for w in wheels if not w.path.exists())
 
+            if source is None:
+                # at least one wheel exists -> report in the SBOM
+                downloaded.append(
+                    {
+                        "package": req.package,
+                        "version": req.version_specs[0][1],
+                        "kind": req.kind,
+                        "hash_verified": False,
```
chmeliik, line 1434: Why is hash_verified a constant False?

cachi2/core/package_managers/pip.py
```diff
@@ -1768,22 +1796,7 @@ def _process_package_distributions(
...
-        if p.version
-        and canonicalize_version(p.version) == normalized_version
-        and p.package_type == "sdist"
-    ]
-
-    if not sdists:
+    if len(sdists) == 0 and len(wheels) == 0:
         raise PackageRejected(
             f"No sdists found for package {name}=={version}",
             solution=(
```
chmeliik, line 1802: This error message only works for allow_binary=False

slimreaper35:
[repo](https://github.com/slimreaper35/cachi2-pip-without-sdists) for manual testing

cachi2/core/package_managers/pip.py
```diff
@@ -1795,17 +1808,21 @@ def _process_package_distributions(
...
-            ),
-        )
+    if len(sdists) == 0 and len(wheels) > 0:
+        log.warning("No sdists found for package %s==%s", name, version)
+        best_sdist = None
+    else:
+        best_sdist = max(sdists, key=_sdist_preference)
+        if best_sdist.is_yanked:
+            raise PackageRejected(
+                f"All sdists for package {name}=={version} are yanked",
```
chmeliik, line 1818: Hmm, being yanked is another thing we handle for sdists but not for wheels.

It might be hard to come up with any sane way to handle yanked-ness in an allow_binary context. Maybe we should just log a warning and download yanked things anyway (for consistency: even when allow_binary=False). That's how 'yanked' is supposed to work when the version is pinned, after all. https://peps.python.org/pep-0592/#installers

chmeliik: But that's probably something for a follow-up story

chmeliik:
After adding a hash to your requirements.txt:
```
tensorflow==2.13.0 --hash=sha256:b2978b39e8b3919059b5fd9e28508d50a77965d06ed0b537ed71c97de22dabdf
```
tensorflow is still reported with missing_hash
```json
    {
      "name": "tensorflow",
      "purl": "pkg:pypi/tensorflow@2.13.0",
      "version": "2.13.0",
      "properties": [
        {
          "name": "cachi2:found_by",
          "value": "cachi2"
        },
        {
          "name": "cachi2:missing_hash:in_file",
          "value": "requirements.txt"
        }
      ],
      "type": "library"
    }
```

cachi2/core/package_managers/pip.py
```diff
@@ -1765,47 +1795,45 @@ def _process_package_distributions(
...
-        if len(filtered) > 0 and len(wheels) == 0:
+        if len(filtered) == 0 and len(sdists) == 0:
+            raise PackageRejected(
+                f"No archives found for package {name}=={version}",
+                solution=(
+                    "It seems that this version does not exist or isn't published.\n"
+                    "You may be able to specify the dependency directly via a URL instead, "
+                    "for example the wheel for a GitHub release."
+                ),
+                docs=PIP_NO_SDIST_DOC,
```
chmeliik, line 1806: Do GitHub releases of Python projects usually include wheel files? I don't think so

The "solution" here would probably be to double-check if the name and version that the user specified really exists on PyPI

cachi2/core/package_managers/pip.py
```diff
@@ -1765,47 +1795,45 @@ def _process_package_distributions(
...
+            )
+        elif len(filtered) == 0:
+            log.warning("No wheel distributions found for package %s==%s", name, version)
+        elif len(wheels) == 0:
             log.warning("All %s wheel distributions were filtered out", name)
```
chmeliik, line 1811: What is `filtered` and what is `wheels`? From the log messages, it looks like `filtered` is the unfiltered list of wheels and `wheels` is the filtered list? :sweat_smile:

cachi2/core/package_managers/pip.py
```diff
@@ -1765,47 +1795,45 @@ def _process_package_distributions(
             else:
                 log.info("Filtering out %s due to checksum mismatch", wheel.filename)
 
-        if len(filtered) > 0 and len(wheels) == 0:
+        if len(filtered) == 0 and len(sdists) == 0:
+            raise PackageRejected(
+                f"No archives found for package {name}=={version}",
```
chmeliik, line 1800: Why are we checking the `filtered` list, rather than the `wheels` list? Isn't it possible that `filtered` is non-empty and `wheels` is empty? We should fail if we don't have any archives for a dependency

chmeliik: > We should fail if we don't have any archives for a dependency

And keep in mind that "don't have any archives" can also happen after we download the wheels and delete the ones that don't hash-check

slimreaper35:
I refactored almost the whole function `_process_package_distributions` , to make it (hopefully) easier to understand. Currently:

1. Find all packages on PyPI
2. Filter those packages
3. Log the warnings based on the results, or possibly raise exception

Did not update unit tests yet.

cachi2/core/package_managers/pip.py
lkolacek: Should we specifically mention in one of these stories to refactor this?

chmeliik:
> I refactored almost the whole function `_process_package_distributions` , to make it (hopefully) easier to understand. Currently:
> 
>     1. Find all packages on PyPI
> 
>     2. Filter those packages
> 
>     3. Log the warnings based on the results, or possibly raise exception
> 
> 
> Did not update unit tests yet.

Nice, for me it's a lot easier to understand now :+1:

cachi2/core/package_managers/pip.py
slimreaper35: Does @anyone have any suggestions for this?

cachi2/core/package_managers/pip.py
```diff
@@ -1734,78 +1737,70 @@ def _process_package_distributions(
...
+            )
+    else:
+        log.warning("No source distributions found for package %s==%s", name, version)
+        best_sdist = None
+
+        if len(wheels) == 0:
+            if allow_binary:
+                solution = "Try to specify the dependency directly via a URL instead,\nfor example the tarball for a GitHub release."
+            else:
+                solution = "Please check that the package exists on PyPI.\nConsider using option allow_binary to allow wheels."
```
chmeliik, line 1798: IMO these are reversed

If allow_binary is true and there are 0 sdists and 0 wheels, the user probably just messed up the package name or version. There doesn't need to be any complicated solution or docs, just ask the user to double-check.

If allow_binary is false, then we're only looking for sdists, not wheels. That's very unusual from a user's perspective and we need to explain what went wrong. We should also include a link to the docs, like the original error message: https://github.com/containerbuildsystem/cachi2/pull/319/files?diff=split&w=0#diff-cfda53cfc3e45684cd3ac813162ad9ee144594345ff841a69250cc88de960dbcL1787-L1795 (I think we can keep the original error message)

cachi2/core/package_managers/pip.py
chmeliik: yes we should :slightly_smiling_face: 

probably in the one story for refactoring

chmeliik: linked this thread there

cachi2/core/package_managers/pip.py
```diff
@@ -1734,78 +1737,79 @@ def _process_package_distributions(
...
+            )
+    else:
+        log.warning("No source distributions found for package %s==%s", name, version)
+        best_sdist = None
+
+        if len(wheels) == 0:
+            if allow_binary:
+                solution = (
+                    "Please check that the package exists on PyPI or that the name and version are correct.\n"
+                    "Consider using option allow_binary to allow wheel distributions."
```
chmeliik, line 1798: This is in the `allow_binary == True` context, so the suggestion to "Consider using allow_binary" doesn't apply

tests/unit/package_managers/test_pip.py
```diff
@@ -2686,23 +2686,41 @@ def test_process_non_existing_package_distributions(
             == f"PyPI query failed: No details about project '{package_name}' available at URL"
         )
 
+    @pytest.mark.parametrize("allow_binary", (True, False))
     @mock.patch.object(pypi_simple.PyPISimple, "get_project_page")
     def test_process_existing_package_without_distributions(
```
chmeliik, line 2691: Is there a test for what happens when there are 0 sdists but there are some wheels?

chmeliik (approved):
LGTM :+1: 

I think there's a corner case where we delete all the wheels after downloading because of checksum mismatches so there are 0 archives left for a package, but that should be extremely rare. Probably not worth handling.

chmeliik:
marking docs as N/A since there's a separate docs story

ben-alkov (approved):
LGTM

cachi2/core/package_managers/pip.py
brunoapimentel: So, in the scenario that a package has wheels but all sdists are yanked, the request would still fail, right?

Should we file the follow-up story right away so we don't forget about it?

brunoapimentel (approved):

PR #328, opened 2023-09-26
Pydantic: This isn't just an upgrade. It's a breakthrough.

slimreaper35:


slimreaper35:
squashed commits

tests/unit/models/test_input.py
```diff
@@ -67,55 +63,55 @@ class TestPackageInput:
...
             ),
         ],
     )
     def test_invalid_packages(self, input_data: dict[str, Any], expect_error: str) -> None:
-        with pytest.raises(pydantic.ValidationError, match=expect_error):
-            # doesn't pass type check: https://github.com/pydantic/pydantic/issues/1847
-            pydantic.parse_obj_as(PackageInput, input_data)  # type: ignore
+        with pytest.raises((pydantic.ValidationError, TypeError), match=expect_error):
+            adapter = pydantic.TypeAdapter(PackageInput)
+            adapter.validate_python(input_data)
```
chmeliik, line 114: When does it raise TypeError? It shouldn't raise TypeError :thinking:

chmeliik (commented):
I've noticed that we have quite a lot of duplication in the error messages now, e.g.

```shell
$ cachi2 fetch-deps '{"type": "pip", "path": "foo"}'        
...
packages
  Value error, package path does not exist (or is not a directory): foo (error=package path does not exist (or is not a directory): foo)

$ cachi2 fetch-deps '{"type": "cabal"}'        
...
packages -> 0
  Input tag 'cabal' found using 'type' does not match any of the expected tags: 'gomod', 'npm', 'pip', 'yarn' (discriminator='type'; tag=cabal; expected_tags='gomod', 'npm', 'pip', 'yarn')
```

Maybe we can stop adding the `context` values to the error messages
https://github.com/containerbuildsystem/cachi2/blob/e1ad4be28a0408ebc0f9399f0e84c0b7c8aad5f7/cachi2/core/models/input.py#L37-L41

tests/unit/models/test_input.py
slimreaper35: As far as I know Pydantic 2 catches some errors and not others by default because it wants to give users the flexibility to handle validation errors in their own way. It catches only ValueError, Assertion and custom ValidationError.

If I changed TypeError to for example KeyError or RuntimeError, still the same result.
Thats why I had change it [here](https://github.com/containerbuildsystem/cachi2/pull/328/files#diff-4b875352bcb272a1d35c2b56db986573fc40bd9c2a1d0d008c9d15c1270dd3a0) as well, because of:

```
@staticmethod
def _validate(value: Any) -> "RootedPath":
    if not isinstance(value, (str, PathLike)):
        raise TypeError(f"expected str or os.PathLike, got {type(value).__name__}")
    return RootedPath(path=value)
```

But maybe there is some workaround here :man_artist:

chmeliik: Ah, I see. Let's raise ValueError instead then

chmeliik:
> I've noticed that we have quite a lot of duplication in the error messages now, e.g.
> 
> ```shell
> $ cachi2 fetch-deps '{"type": "pip", "path": "foo"}'        
> ...
> Error: InvalidInput: 1 validation error for user input
> packages
>   Value error, package path does not exist (or is not a directory): foo (error=package path does not exist (or is not a directory): foo)
> 
> $ cachi2 fetch-deps '{"type": "cabal"}'        
> ...
> Error: InvalidInput: 1 validation error for user input
> packages -> 0
>   Input tag 'cabal' found using 'type' does not match any of the expected tags: 'gomod', 'npm', 'pip', 'yarn' (discriminator='type'; tag=cabal; expected_tags='gomod', 'npm', 'pip', 'yarn')
> ```
> 
> Maybe we can stop adding the `context` values to the error messages
> 
> https://github.com/containerbuildsystem/cachi2/blob/e1ad4be28a0408ebc0f9399f0e84c0b7c8aad5f7/cachi2/core/models/input.py#L37-L41

What about the error messages? It looks like the message itself contains all the relevant info now, by adding the context values we just duplicate the info

(would be good to fix it in a separate commit)

chmeliik (approved):
LGTM :+1: 

The second commit message could include an explanation for why we're removing the context

ben-alkov (approved):
LGTM

cachi2/core/models/input.py
```diff
@@ -121,30 +119,32 @@ class Request(pydantic.BaseModel):
...
+    @pydantic.field_validator("packages")
     def _unique_packages(cls, packages: list[PackageInput]) -> list[PackageInput]:
         """De-duplicate the packages to be processed."""
         return unique(packages, by=lambda pkg: (pkg.type, pkg.path))
 
-    @pydantic.validator("packages", each_item=True)
-    def _check_package_paths(cls, package: PackageInput, values: dict) -> PackageInput:
+    @pydantic.field_validator("packages")
+    def _check_packages_paths(
+        cls, packages: list[PackageInput], info: pydantic.ValidationInfo
```
brunoapimentel, line 129: Alternatively, we could use [pydantic.functional_validators.AfterValidator](https://docs.pydantic.dev/latest/concepts/validators/#generic-validated-collections) to still apply the validation for `each_item`. What I don't like about it is that we'd need to move this function to outside the class, diverging from the rest of validator functions. It does look better without the loop, on the other hand. So take this as a nitpick/suggestion.

Here's how I tested it locally:
```
diff --git a/cachi2/core/models/input.py b/cachi2/core/models/input.py
index 198a158..84d866b 100644
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -7,6 +7,8 @@ from cachi2.core.errors import InvalidInput
 from cachi2.core.models.validators import check_sane_relpath, unique
 from cachi2.core.rooted_path import PathOutsideRoot, RootedPath
 
+from pydantic.functional_validators import AfterValidator
+
 if TYPE_CHECKING:
     from pydantic.error_wrappers import ErrorDict
 
@@ -111,12 +113,29 @@ PackageInput = Annotated[
 ]
 
 
+def _check_packages_paths(package: PackageInput, info: pydantic.ValidationInfo) -> list[PackageInput]:
+    """Check that package paths are existing subdirectories."""
+    source_dir: RootedPath = info.data.get("source_dir", None)
+    if source_dir is not None:
+        try:
+            abspath = source_dir.join_within_root(package.path)
+        except PathOutsideRoot:
+            raise ValueError(
+                f"package path (a symlink?) leads outside source directory: {package.path}"
+            )
+        if not abspath.path.is_dir():
+            raise ValueError(
+                f"package path does not exist (or is not a directory): {package.path}"
+            )
+    return package
+
+
 class Request(pydantic.BaseModel):
     """Holds all data needed for the processing of a single request."""
 
     source_dir: RootedPath
     output_dir: RootedPath
-    packages: list[PackageInput]
+    packages: list[Annotated[PackageInput, AfterValidator(_check_packages_paths)]]
     flags: frozenset[Flag] = frozenset()
 
     @pydantic.field_validator("packages")
@@ -124,26 +143,6 @@ class Request(pydantic.BaseModel):
         """De-duplicate the packages to be processed."""
         return unique(packages, by=lambda pkg: (pkg.type, pkg.path))
 
-    @pydantic.field_validator("packages")
-    def _check_packages_paths(
-        cls, packages: list[PackageInput], info: pydantic.ValidationInfo
-    ) -> list[PackageInput]:
-        """Check that package paths are existing subdirectories."""
-        source_dir: RootedPath = info.data.get("source_dir", None)
-        if source_dir is not None:
-            for p in packages:
-                try:
-                    abspath = source_dir.join_within_root(p.path)
-                except PathOutsideRoot:
-                    raise ValueError(
-                        f"package path (a symlink?) leads outside source directory: {p.path}"
-                    )
-                if not abspath.path.is_dir():
-                    raise ValueError(
-                        f"package path does not exist (or is not a directory): {p.path}"
-                    )
-        return packages
-
     @pydantic.field_validator("packages")
     def _packages_not_empty(cls, packages: list[PackageInput]) -> list[PackageInput]:
         """Check that the packages list is not empty."""
```

brunoapimentel (approved):

PR #338, opened 2023-10-04
Implement classes to parse Yarn files

ejegrova:
Implement the classes stablished by the high-level code design (#309) to parse Yarn files (i.e. package.json and yarnrc.yml). 

### DIFF
--- a/cachi2/core/package_managers/yarn/main.py
+++ b/cachi2/core/package_managers/yarn/main.py
@@ -61,7 +61,11 @@ def _configure_yarn_version(project: Project) -> None:
         if the yarn version can't be determined from either yarnPath or packageManager
         if there is a mismatch between the yarn version specified by yarnPath and PackageManager
     """
-    yarn_path_version = get_semver_from_yarn_path(project.yarn_rc.yarn_path)
+    if project.yarn_rc:
+        yarn_path_version = get_semver_from_yarn_path(project.yarn_rc.yarn_path)
+    else:
+        yarn_path_version = None
+
     package_manager_version = get_semver_from_package_manager(project.package_json.package_manager)
 
     if not yarn_path_version and not package_manager_version:
--- a/cachi2/core/package_managers/yarn/main.py
+++ b/cachi2/core/package_managers/yarn/main.py
@@ -38,6 +38,8 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Comp
     :param project: the directory to be processed.
     :param output_dir: the directory where the prefetched dependencies will be placed.
     """
+    log.info(f"Fetching the yarn dependencies at the subpath {output_dir.subpath_from_root}")
+
     _configure_yarn_version(project)
 
     try:
--- a/cachi2/core/package_managers/yarn/project.py
+++ b/cachi2/core/package_managers/yarn/project.py
@@ -11,11 +11,16 @@ from pathlib import Path
 from typing import Any, NamedTuple, Optional
 
 import semver
+import yaml
 
-from cachi2.core.errors import UnexpectedFormat
+from cachi2.core.errors import PackageRejected, UnexpectedFormat
 from cachi2.core.rooted_path import RootedPath
 
 log = logging.getLogger(__name__)
+
+
+DEFAULT_CACHE_FOLDER = "./.yarn/cache"
+DEFAULT_REGISTRY = "https://registry.yarnpkg.com"
 
 
 class YarnRc:
@@ -35,12 +40,12 @@ class YarnRc:
         self._data = data
 
     @property
-    def cache_path(self) -> str:
+    def cache_folder(self) -> str:
         """Get the configured location for the yarn cache folder.
 
         Fallback to the default path in case the configuration key is missing.
         """
-        return NotImplemented
+        return self._data.get("cacheFolder", DEFAULT_CACHE_FOLDER)
 
     @property
     def registry_server(self) -> str:
@@ -48,20 +53,12 @@ class YarnRc:
 
         Fallback to the default server in case the configuration key is missing.
         """
-        return NotImplemented
+        return self._data.get("npmRegistryServer", DEFAULT_REGISTRY)
 
     @property
     def yarn_path(self) -> Optional[str]:
         """Path to the yarn script present in this directory."""
-        return NotImplemented
-
-    @property
-    def yarn_version(self) -> Optional[str]:
-        """Yarn version used in this project.
-
-        Extracted from the contents of the yarnPath configuration, if it is present.
-        """
-        return NotImplemented
+        return self._data.get("yarnPath")
 
     def registry_server_for_scope(self, scope: str) -> str:
         """Get the configured registry server for a scoped package.
@@ -71,12 +68,29 @@ class YarnRc:
 
         See: https://v3.yarnpkg.com/configuration/yarnrc#npmScopes
         """
-        return NotImplemented
+        registry = self._data.get("npmScopes", {}).get(scope, {}).get("npmRegistryServer")
+
+        return registry or self.registry_server
 
     @classmethod
     def from_file(cls, file_path: RootedPath) -> "YarnRc":
         """Parse the content of a yarnrc file."""
-        return NotImplemented
+        try:
+            with file_path.path.open("r") as f:
+                yarnrc_data = yaml.safe_load(f)
+        except yaml.YAMLError as e:
+            raise PackageRejected(
+                f"Can't parse the {file_path.subpath_from_root} file. Parser error: {e}",
+                solution=(
+                    "The yarnrc file must contain valid YAML. "
+                    "Refer to the parser error and fix the contents of the file."
+                ),
+            )
+
+        if yarnrc_data is None:
+            yarnrc_data = {}
+
+        return cls(file_path, yarnrc_data)
 
 
 class PackageJson:
@@ -98,7 +112,7 @@ class PackageJson:
     @property
     def package_manager(self) -> Optional[str]:
         """Get the package manager string."""
-        return NotImplemented
+        return self._data.get("packageManager")
 
     @package_manager.setter
     def package_manager(self, package_manager: str) -> None:
@@ -108,7 +122,27 @@ class PackageJson:
     @classmethod
     def from_file(cls, file_path: RootedPath) -> "PackageJson":
         """Parse the content of a package.json file."""
-        return NotImplemented
+        try:
+            with file_path.path.open("r") as f:
+                package_json_data = json.load(f)
+        except FileNotFoundError:
+            raise PackageRejected(
+                "The package.json file must be present for the yarn package manager",
+                solution=(
+                    "Please double-check that you have specified the correct path "
+                    "to the package directory containing this file"
+                ),
+            )
+        except json.decoder.JSONDecodeError as e:
+            raise PackageRejected(
+                f"Can't parse the {file_path.subpath_from_root} file. {e}",
+                solution=(
+                    "The package.json file must contain valid JSON. "
+                    "Refer to the parser error and fix the contents of the file."
+                ),
+            )
+
+        return cls(file_path, package_json_data)
 
     def write_to_file(self) -> None:
         """Write the data to the package.json file."""
@@ -121,7 +155,7 @@ class Project(NamedTuple):
     """A directory containing yarn sources."""
 
     source_dir: RootedPath
-    yarn_rc: YarnRc
+    yarn_rc: Optional[YarnRc]
     package_json: PackageJson
 
     @property
@@ -131,7 +165,12 @@ class Project(NamedTuple):
         This is determined by the existence of a non-empty yarn cache folder. For more details on
         zero-installs, see: https://v3.yarnpkg.com/features/zero-installs.
         """
-        return False
+        dir = self.yarn_cache
+
+        if not dir.path.is_dir():
+            return False
+
+        return any(file.suffix == ".zip" for file in dir.path.iterdir())
 
     @property
     def yarn_cache(self) -> RootedPath:
@@ -140,12 +179,23 @@ class Project(NamedTuple):
         The cache location is affected by the cacheFolder configuration in yarnrc. See:
         https://v3.yarnpkg.com/configuration/yarnrc#cacheFolder.
         """
-        return NotImplemented
+        if self.yarn_rc:
+            return self.source_dir.join_within_root(self.yarn_rc.cache_folder)
+
+        return self.source_dir.join_within_root(DEFAULT_CACHE_FOLDER)
 
     @classmethod
     def from_source_dir(cls, source_dir: RootedPath) -> "Project":
         """Create a Project from a sources directory path."""
-        return cls(source_dir, NotImplemented, NotImplemented)
+        yarn_rc_path = source_dir.join_within_root(".yarnrc.yml")
+
+        if yarn_rc_path.path.exists():
+            yarn_rc = YarnRc.from_file(source_dir.join_within_root(".yarnrc.yml"))
+        else:
+            yarn_rc = None
+
+        package_json = PackageJson.from_file(source_dir.join_within_root("package.json"))
+        return cls(source_dir, yarn_rc, package_json)
 
 
 def get_semver_from_yarn_path(yarn_path: Optional[str]) -> Optional[semver.version.Version]:
--- a/tests/unit/package_managers/yarn/test_project.py
+++ b/tests/unit/package_managers/yarn/test_project.py
@@ -4,11 +4,199 @@ from typing import Optional
 import pytest
 import semver
 
-from cachi2.core.errors import UnexpectedFormat
+from cachi2.core.errors import PackageRejected, UnexpectedFormat
 from cachi2.core.package_managers.yarn.project import (
+    PackageJson,
+    Project,
+    YarnRc,
     get_semver_from_package_manager,
     get_semver_from_yarn_path,
 )
+from cachi2.core.rooted_path import PathOutsideRoot, RootedPath
+
+VALID_YARNRC_FILE = """
+plugins:
+  - path: .yarn/plugins/@yarnpkg/plugin-typescript.cjs
+    spec: "@yarnpkg/plugin-typescript"
+  - path: .yarn/plugins/@yarnpkg/plugin-exec.cjs
+    spec: "@yarnpkg/plugin-exec"
+
+yarnPath: .custom/path/yarn-3.6.1.cjs
+
+enableInlineBuilds: true
+
+supportedArchitectures:
+  os:
+    - "linux"
+
+cacheFolder: ./.custom/cache
+
+npmRegistryServer: https://registry.alternative.com
+
+npmScopes:
+    foobar:
+        npmRegistryServer: https://registry.foobar.com
+"""
+
+EMPTY_YML_FILE = ""
+
+INVALID_YML = "this: is: not: valid: yaml"
+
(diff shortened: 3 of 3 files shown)


cachi2/core/package_managers/yarn/project.py
```diff
@@ -120,9 +148,12 @@ def yarn_cache(self) -> RootedPath:
...
         """
-        return NotImplemented
+        return self.source_dir.join_within_root(self.yarn_rc.cache_path)
 
     @classmethod
     def from_source_dir(cls, source_dir: RootedPath) -> "Project":
         """Create a Project from a sources directory path."""
-        return NotImplemented
+        yarn_rc = YarnRc.from_file(source_dir.join_within_root(".yarnrc.yml"))
+        package_json = PackageJson.from_file(source_dir.join_within_root("package.json"))
```
brunoapimentel, line 157: The [docs](https://v3.yarnpkg.com/configuration/yarnrc) state that:
```
While Yarn will automatically find them in the parent directories, they should usually be kept at the root of your project (often your repository)
```

So that is one thing we need to account for when covering the path/symlink validation story.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -120,9 +148,12 @@ def yarn_cache(self) -> RootedPath:
...
         https://v3.yarnpkg.com/configuration/yarnrc#cacheFolder.
         """
-        return NotImplemented
+        return self.source_dir.join_within_root(self.yarn_rc.cache_path)
 
     @classmethod
     def from_source_dir(cls, source_dir: RootedPath) -> "Project":
         """Create a Project from a sources directory path."""
-        return NotImplemented
+        yarn_rc = YarnRc.from_file(source_dir.join_within_root(".yarnrc.yml"))
```
brunoapimentel, line 156: The yarnrc is not mandatory, so we need to account for cases in which it's absent. Calling `yarn init` in a folder does not autogenerate it.

brunoapimentel: What if package.json is missing? Are we going to raise a `PackageRejected`?

cachi2/core/package_managers/yarn/project.py
```diff
@@ -31,43 +39,56 @@ def cache_path(self) -> str:
...
+        return self._data.get("npmRegistryServer", "https://registry.yarnpkg.com")
 
     @property
     def yarn_path(self) -> Optional[str]:
         """Path to the yarn script present in this directory."""
-        return NotImplemented
+        try:
+            return self._data.get("yarnPath")
+        except KeyError:
+            log.warning(f"No yarnPath in {self._path}")
```
brunoapimentel, line 58: Since yarnPath is not a mandatory key, I think logging a warning here may be misleading.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -31,43 +39,56 @@ def cache_path(self) -> str:
...
     @property
     def yarn_version(self) -> Optional[str]:
         """Yarn version used in this project.
 
         Extracted from the contents of the yarnPath configuration, if it is present.
         """
-        return NotImplemented
+        try:
+            yarn_path = self._data.get("yarnPath")
+            return re.search(r"([0-9]+\.[0-9]+\.[0-9]+)", yarn_path).group(0)
```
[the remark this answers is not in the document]
chmeliik: +1

rather than trying to match npm semver, how about we just take this part of the filename as version?
```
some/path/yarn-(.*).cjs
               ^^^^
```

cachi2/core/package_managers/yarn/project.py
chmeliik: `source_dir.join_within_root("package.json")` does the validation right?

cachi2/core/package_managers/yarn/project.py
chmeliik: > Calling yarn init in a folder does not autogenerate it.

Calling `yarn init -2` does, but I think you're right that it's not mandatory

cachi2/core/package_managers/yarn/project.py
chmeliik: +1

also `self._data.get()` will never raise a KeyError, IMO you can remove the whole try/except

cachi2/core/package_managers/yarn/project.py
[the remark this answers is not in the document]
ejegrova: nice :+1:  changed the regex

cachi2/core/package_managers/yarn/project.py
ejegrova: removed

cachi2/core/package_managers/yarn/project.py
```diff
@@ -31,43 +40,48 @@ def cache_path(self) -> str:
...
 
     @property
     def yarn_version(self) -> Optional[str]:
         """Yarn version used in this project.
 
         Extracted from the contents of the yarnPath configuration, if it is present.
         """
-        return NotImplemented
+        yarn_path = self.yarn_path
+        return re.search(r"yarn-(.*).cjs", yarn_path).group(1)
```
taylormadore, line 65: Won't we hit an issue here if self.yarn_path is None?

cachi2/core/package_managers/yarn/project.py
ejegrova: Raising `PackageRejected` in PackageJson classmethod.

cachi2/core/package_managers/yarn/project.py
ejegrova: yup

chmeliik: And also if the pattern doesn't match and re.search returns None (mypy caught both)

chmeliik: TBH I feel like this logic doesn't really belong in YarnRC but in the code for setting yarn version

cachi2/core/package_managers/yarn/project.py
```diff
@@ -111,7 +141,10 @@ def is_zero_installs(self) -> bool:
         This is determined by the existence of a non-empty yarn cache folder. For more details on
         zero-installs, see: https://v3.yarnpkg.com/features/zero-installs.
         """
-        return NotImplemented
+        dir = self.yarn_cache
+        return Path.is_dir(dir.path) and any(
+            ".zip" in file.suffix for file in Path.iterdir(dir.path)
+        )
```
chmeliik, line 147:
```suggestion
        return dir.path.is_dir() and any(
            file.suffix == ".zip" for file in dir.path.iterdir()
        )
```

cachi2/core/package_managers/yarn/project.py
ejegrova: ok, I'm gonna remove `yarn_version` property

cachi2/core/package_managers/yarn/project.py
brunoapimentel: I was thinking that a parent dir package.json could be accessed by yarn itself when we invoked any of the yarn commands. But with the validation to enforce that a valid package.json file exists in every project, this will not happen.

brunoapimentel:
I am taking over this PR since Eliška's out for some time. Added unit tests and tried to cover more corner cases. I still need to do more testing with real scenarios, but it should be ready to review as is.

tests/unit/package_managers/yarn/test_project.py
```diff
@@ -0,0 +1,148 @@
...
+
+    return PackageJson.from_file(path)
+
+
+def test_parse_package_json(rooted_tmp_path: RootedPath) -> None:
+    package_json = _prepare_package_json_file(rooted_tmp_path, VALID_PACKAGE_JSON_FILE)
+    assert package_json.package_manager == "yarn@3.6.1"
+
+
+def test_paser_empty_package_json_file(rooted_tmp_path: RootedPath) -> None:
```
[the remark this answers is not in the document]
brunoapimentel: Fixed.

taylormadore (approved):
LGTM :+1:

cachi2/core/package_managers/yarn/project.py
```diff
@@ -62,12 +65,26 @@ def registry_server_for_scope(self, scope: str) -> str:
...
+        return registry or self.registry_server
 
     @classmethod
     def from_file(cls, file_path: RootedPath) -> "YarnRc":
         """Parse the content of a yarnrc file."""
-        return NotImplemented
+        try:
+            with file_path.path.open("r") as f:
+                yarnrc_data = yaml.safe_load(f)
+        except yaml.error.MarkedYAMLError as e:
```
chmeliik, line 78: Should it be specifically MarkedYAMLError? There's also a generic yaml.YAMLError

cachi2/core/package_managers/yarn/project.py
```diff
@@ -62,12 +65,26 @@ def registry_server_for_scope(self, scope: str) -> str:
...
+            raise PackageRejected(
+                "Can't parse the .yarnrc.yml file",
+                solution=f"Check the contents of the file. Parser error: {e}",
```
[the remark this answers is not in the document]
chmeliik: > Might want to tell the user where the .yarnc.yml file is (relative to the repo root)

Or just log which path we're processing when we start processing it. There will probably be more cases like this

cachi2/core/package_managers/yarn/project.py
```diff
@@ -4,10 +4,21 @@
...
 
+import yaml
+
+from cachi2.core.errors import PackageRejected
 from cachi2.core.rooted_path import RootedPath
 
+log = logging.getLogger(__name__)
+
+
+DEFAULT_CACHE_FOLDER = "./.yarn/cache"
```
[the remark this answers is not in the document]
chmeliik: Good point, for the outside-RHTAP use case a user could indeed be setting `YARN_` environment variables

Even the path to .yarnrc.yml itself is configurable via env var (but this would _only_ work outside RHTAP, where the user can't set env vars) https://v3.yarnpkg.com/configuration/yarnrc#rcFilename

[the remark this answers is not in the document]
chmeliik: > Even the path to .yarnrc.yml itself is configurable via env var (but this would _only_ work outside RHTAP, where the user can't set env vars I think) https://v3.yarnpkg.com/configuration/yarnrc#rcFilename

Or maybe they can, I never know which part of the definition the user can edit or not https://tekton.dev/docs/pipelines/tasks/#specifying-a-step-template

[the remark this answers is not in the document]
eskultety: > Good point, for the outside-RHTAP use case a user could indeed be setting `YARN_` environment variables
> 
> Even the path to .yarnrc.yml itself is configurable via env var (but this would _only_ work outside RHTAP, where the user can't set env vars I think) https://v3.yarnpkg.com/configuration/yarnrc#rcFilename

Aand as it turns out, in this particular case `YARN_CACHE_FOLDER` is a valid env variable.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -26,33 +37,25 @@ def __init__(self, path: RootedPath, data: dict[str, Any]) -> None:
         self._data = data
 
     @property
-    def cache_path(self) -> str:
+    def cache_folder(self) -> str:
```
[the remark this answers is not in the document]
eskultety: > Stupid nitpick: Is "folder" an official term used by Yarn itself? Technically, folder can only store files, not subdirectories, so maybe `cache_dir` or `cachedir` might a more appropriate name. Looking at the code base, there is a precedent in gomod, so you can happily ignore this, it's just any Python primitive used for the actual underlying operation in this context would most likely use the word 'dir' in its name :).

Turns out, Yarn indeed calls it a _cacheFolder_, so disregard my comment please.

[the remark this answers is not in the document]
brunoapimentel: Yeah, the reason I changed it here was to match the actual configuration parameter.

cachi2/core/package_managers/yarn/project.py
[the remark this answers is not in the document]
brunoapimentel: If we always set the `env` parameter when [calling](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/yarn/utils.py#L14) the subprocess, we will override any environment variable that is set in the user's env. So I think the solution here is being sure this parameter is at least set to an empty dict.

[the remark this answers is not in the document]
brunoapimentel: If we do want to expand our use case to also consider the environment variables that are already set, then we probably will need to have a separate parsing of `YARN_*` environment variables and always consider precedence for them instead of what's in yarnrc (see the early work in what we will need to support [here](https://github.com/brunoapimentel/cachi2/blob/set-yarn-config/cachi2/core/package_managers/yarn/project.py#L48-L115)).

I think it's definitely a valid use case, but we should probably handle that in a separate PR. In any case, the variable here is meant to represent the fallback value for when `cacheFolder` is not set in .yarnrc, so I think it is still correct.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -4,10 +4,21 @@
...
+import yaml
+
+from cachi2.core.errors import PackageRejected
 from cachi2.core.rooted_path import RootedPath
 
+log = logging.getLogger(__name__)
+
+
+DEFAULT_CACHE_FOLDER = "./.yarn/cache"
+DEFAULT_REGISTRY = "https://registry.yarnpkg.com"
```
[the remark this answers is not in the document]
brunoapimentel: I think it can be set with `YARN_NPM_REGISTRY_SERVER`.

cachi2/core/package_managers/yarn/project.py
brunoapimentel: Changed it to YAMLError.

cachi2/core/package_managers/yarn/project.py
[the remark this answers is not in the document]
brunoapimentel: I did both. I'm not 100% happy with the error log output, but I think all the necessary information is there. 

We can probably revisit them with that logging story we have (if we ever get to it).

tests/unit/package_managers/yarn/test_project.py
```diff
@@ -0,0 +1,168 @@
...
+    assert project.yarn_rc._path == rooted_tmp_path.join_within_root(".yarnrc.yml")
+    assert project.package_json._path == rooted_tmp_path.join_within_root("package.json")
+
+
+def test_parse_project_folder_without_yarnrc(rooted_tmp_path: RootedPath) -> None:
+    _prepare_package_json_file(rooted_tmp_path, VALID_PACKAGE_JSON_FILE)
+
+    project = Project.from_source_dir(rooted_tmp_path)
+
+    assert not project.is_zero_installs
```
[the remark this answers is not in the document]
brunoapimentel: You're right. Added a new scenario for this test.

cachi2/core/package_managers/yarn/project.py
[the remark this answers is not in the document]
chmeliik: Yeah, we either need to document it as unsupported or support it.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -32,6 +32,8 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Comp
     :param project: the directory to be processed.
     :param output_dir: the directory where the prefetched dependencies will be placed.
     """
+    log.info(f"Fetching the yarn dependencies at the subpath {output_dir.subpath_from_root}")
```
chmeliik, line 35: nitpick: unrelated to this commit

chmeliik:
Marking docs as n/a because we have a docs story

lkolacek (approved):
LGTM :+1:

cachi2/core/package_managers/yarn/project.py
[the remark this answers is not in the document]
brunoapimentel: I'll file a story so we can discuss what's the best approach.

cachi2/core/package_managers/yarn/main.py
brunoapimentel: I was wondering if we should start adding the logs on the main functions as the work goes or if we should have a separate story for it. In any case, since we already have the path for the file in that [error message](https://github.com/containerbuildsystem/cachi2/pull/338/files#diff-cbc68c73831a63c3d995893ed401dc07bf92cd8e8654b498177c465e8584ef18), I think I'll just drop this line here.

chmeliik: Yeah, just do it in a separate commit please since it's not related to project classes

brunoapimentel: I mean, not add the log line at all in this PR, and leave it for later. Unless you think it's nice to add it now, then I can add it in separate commit.

chmeliik: Ah, right. Yeah let's do it now, unless you can make sure that it will, at some point, happen :smile:

brunoapimentel: Ok, kept it in a separate commit.

brunoapimentel:
Split the log message into a separate commit and added ids to the unit test scenarios so they can be more readable in the pytest output.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -4,10 +4,21 @@
 It also provides basic utility functions. The main logic to resolve and prefetch the dependencies
 should be implemented in other modules.
 """
+import json
+import logging
```
eskultety, line 8: importing logging and initializing the logger needs to be done in the next commit where you actually make use of the logging interface.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -111,7 +151,8 @@ def is_zero_installs(self) -> bool:
         This is determined by the existence of a non-empty yarn cache folder. For more details on
         zero-installs, see: https://v3.yarnpkg.com/features/zero-installs.
         """
-        return False
+        dir = self.yarn_cache
+        return dir.path.is_dir() and any(file.suffix == ".zip" for file in dir.path.iterdir())
```
eskultety, line 155: nitpick: I feel like there's too much going on on this line. Would you mind creating an intermediary variable for the comprehension and then using that inside this boolean expression? IMO it would make it a bit more readable. Having said that I know there's a lot of precedent in the code base so you can happily ignore my comment based on that, but I think despite Python allowing a lot in terms of single-line expressions/constructions, it IMO hinders readability in general.

eskultety (commented):
A couple of nitpicks, nothing major though. Overall, the code looks fine to me, however, I can't provide comments on the general architecture/design wrt/ Yarn internals.

slimreaper35 (approved):

cachi2/core/package_managers/yarn/project.py
brunoapimentel: I changed the "is there any zip file" piece to a nested function, that way it's simple to figure out what's happening by reading the function name.

Other alternative is decomposing this to an if block, since the second piece is dependent on the is_dir check. Let me know if you prefer it that way.

eskultety: > Other alternative is decomposing this to an if block, since the second piece is dependent on the is_dir check. Let me know if you prefer it that way.

Call me old-fashioned, but ^this to me sounds like the more readable approach.

brunoapimentel: Ok, changed it to an if block.

eskultety (approved):

PR #339, opened 2023-10-05
Make it harder to omit a locator test by mistake

chmeliik:
The test parametrization zips the input with the expected output. Zip stops when the shortest iterable ends. That means if the input and output lists have different lengths, cases can be omitted accidentally.

zip(..., strict=True) would solve this on python>=3.10

We support python 3.9, so instead use zip_longest, which will "pad" the shorter list with None

### DIFF
--- a/tests/unit/package_managers/yarn/test_locators.py
+++ b/tests/unit/package_managers/yarn/test_locators.py
@@ -1,4 +1,5 @@
 import re
+from itertools import zip_longest
 from pathlib import Path
 
 import pytest
@@ -410,7 +411,7 @@ PARSED_SUPPORTED_LOCATORS = [
     "locator_str, expect_parsed_locator, expect_parsed_reference",
     [
         (locator_str, parsed_locator, parsed_reference)
-        for locator_str, (parsed_locator, parsed_reference) in zip(
+        for locator_str, (parsed_locator, parsed_reference) in zip_longest(
             ALL_LOCATORS, PARSED_LOCATORS_AND_REFERENCES
         )
     ],
@@ -457,7 +458,7 @@ def test_unexpected_reference_format() -> None:
 
 
 @pytest.mark.parametrize(
-    "locator_str, expect_locator", zip(SUPPORTED_LOCATORS, PARSED_SUPPORTED_LOCATORS)
+    "locator_str, expect_locator", zip_longest(SUPPORTED_LOCATORS, PARSED_SUPPORTED_LOCATORS)
 )
 def test_parse_locator(locator_str: str, expect_locator: Locator) -> None:
     assert parse_locator(locator_str) == expect_locator


chmeliik:
PTAL, just a tiny test fix to avoid shooting ourselves in the foot later :smile: 

TBH the unit tests for the locators could use a bit more cleanup (probably moving the input and expected output closer together), but I don't know if that's worth it at this point

taylormadore (approved):

slimreaper35 (approved):

brunoapimentel (approved):

brunoapimentel:
Btw, what's holding us from dropping the support to Python 3.9? Some RHEL default version?

chmeliik:
> Btw, what's holding us from dropping the support to Python 3.9? Some RHEL default version?

Cachi2 is also a standalone CLI tool, not just the container. Python 3.9 is still supported (3.8 as well, but we never supported that) https://devguide.python.org/versions/

We don't have any policy about dropping support for python version (unless they're EOL, then it should be fairly non-controversial)

eskultety:
> > Btw, what's holding us from dropping the support to Python 3.9? Some RHEL default version?
> 
> Cachi2 is also a standalone CLI tool, not just the container. Python 3.9 is still supported (3.8 as well, but we never supported that) https://devguide.python.org/versions/
> 
> We don't have any policy about dropping support for python versions (unless they're EOL, then it should be fairly non-controversial)

Should we make an official support policy statement we can refer to in the future?

eskultety:
> > Should we make an official support policy statement we can refer to in the future?
> 
> We should, yeah. What should it be?
> 
> Let's say
> 
> upstream_python = latest release minus 2 minor versions (3.12 - 0.2 = 3.10) rhel_python = the latest available in the second-latest RHEL (RHEL8 => python 3.11)
> 
> And we commit to supporting `min(upstream_python, rhel_python)`
> 
> moving to 3.10 would be pretty awesome, [pattern matching](https://peps.python.org/pep-0636/) ❤️

I thinks ^this suggestion a bit too progressive and instead we should be more conservative. The matter is also a bit more complicated than just looking at Python versions, because since this is an upstream project, people may want to contribute from a plethora of distributions (still supported though) other than RHEL. I'm not saying one should run an ancient distro and expect technically bleeding edge software to work out of the box, I'm just saying that LTS versions of distros are very popular. With that in mind, using something like [repology](https://repology.org/project/python/versions) can tell us that e.g. on Debian 11 which is scheduled for EOL in 2024 only has 3.9 available.
Using the above as an argument, I can immediately think of 2 options (there are definitely more ways):

1. Instead of focusing solely on Python, we derive the minimum supported Python version which hasn't been EOL'd yet according to availability on the distro (e.g. RHEL still defaults to 3.6 on platform Python, despite being EOL'd and so they have extra python3.8+ packages...on the other hand Debian 11 has only 3.9 as stated above) by making a platform support policy statement instead of a Python version one. We could further tweak that statement so that we don't lock ourselves into supporting all kinds of LTS releases indefinitely and instead say that we only support e.g. the 2 latest major releases of a given selection of distros up until 2 years after the release of the last major release or when the vendor itself drops support, whichever comes sooner.

   ^This here is not a thought of mine (giving credit where it's due), we did exactly that in projects I have worked on [[1](https://libvirt.org/platforms.html)][[2](https://www.qemu.org/docs/master/about/build-platforms.html)][[3](https://libosinfo.org/build_platforms.html)] 

2. We won't care about distros at all and simply remain conservative and state that we support all versions of Python which haven't been officially EOL'd yet with the exception of 3.8 because we don't support that one already

chmeliik:
>     2. We won't care about distros at all and simply remain conservative and state that we support all versions of Python which haven't been officially EOL'd yet with the exception of 3.8 because we don't support that one already

That sounds reasonable. If a more complex policy doesn't let us support fewer python versions, it's IMO not worth it.

eskultety:
> > ```
> > 2. We won't care about distros at all and simply remain conservative and state that we support all versions of Python which haven't been officially EOL'd yet with the exception of 3.8 because we don't support that one already
> > ```
> 
> That sounds reasonable. If a more complex policy doesn't let us support fewer python versions, it's IMO not worth it.

Well, just like 1., 2. has drawbacks also - wide range of versions leaves us open to various deprecation issues which we'd have to work around and deal with for quite some time. An example of which would be importlibs, where they introduced a couple of APIs in 3.7 to aid in locating package data as a replacement for setuptools' `pkg_resources`:

- [importlib.resources.path](https://docs.python.org/3.7/library/importlib.html?highlight=importlib%20resources#importlib.resources.path)
- [importlib.resources.open_text](https://docs.python.org/3.7/library/importlib.html?highlight=importlib%20resources#importlib.resources.open_text)
- [importlib.resources.read_text](https://docs.python.org/3.7/library/importlib.html?highlight=importlib%20resources#importlib.resources.read_text)

Only to deprecate all of them in 3.11 in favour of new 3.9 APIs:
- [importlib.resources.files](https://docs.python.org/3.11/library/importlib.resources.html#importlib.resources.files)
- [importlib.resources.as_file](https://docs.python.org/3.11/library/importlib.resources.html#importlib.resources.as_file)

So e.g. for 3.8 which hasn't been EOL'd yet you'd need to maintain a fallback logic until you're ready to bump the minimal version to 3.9. So ^this is just an example of a risk we'd be taking with the more conservative approach I suggested in 2. and hence it's up for a broader discussion if we want to make a formal statement going that direction. See, I told you the matter is "a bit" more complicated :)

chmeliik:
> See, I told you the matter is "a bit" more complicated :)

Oh yeah, it always is :smile: 

Probably deserves an investigation story?

PR #342, opened 2023-10-06
cachi2 yarn: prefetch dependencies for regular workflow

slimreaper35:

### DIFF
--- a/cachi2/core/package_managers/yarn/main.py
+++ b/cachi2/core/package_managers/yarn/main.py
@@ -7,6 +7,7 @@ from cachi2.core.package_managers.yarn.resolver import (
     create_component_from_package,
     resolve_packages,
 )
+from cachi2.core.package_managers.yarn.utils import run_yarn_cmd
 from cachi2.core.rooted_path import RootedPath
 
 log = logging.getLogger(__name__)
@@ -73,7 +74,7 @@ def _check_yarn_cache(source_dir: RootedPath) -> None:
     """Check the contents of the yarn cache using 'yarn install'.
 
     :param source_dir: the directory in which the yarn command will be called.
-    :raises SubprocessCallError: if the 'yarn install' command fails.
+    :raises YarnCommandError: if the 'yarn install' command fails.
     """
     # the yarn commands can be called by using the core.utils.run_cmd function
     pass
@@ -84,11 +85,12 @@ def _fetch_dependencies(source_dir: RootedPath, output_dir: RootedPath) -> None:
 
     :param source_dir: the directory in which the yarn command will be called.
     :param output_dir: the directory where the yarn dependencies will be downloaded to.
-    :raises SubprocessCallError: if the 'yarn install' command fails.
+    :raises YarnCommandError: if the 'yarn install' command fails.
     """
-    # the output_dir is where the globalFolder must be set to
-    # the yarn commands can be called by using the core.utils.run_cmd function
-    pass
+    cachi2_output = output_dir.join_within_root("deps", "yarn")
+
+    args = ["install", "--mode", "skip-build"]
+    run_yarn_cmd(args, source_dir, {"YARN_GLOBAL_FOLDER": str(cachi2_output)})
 
 
 def _undo_changes(project: Project) -> None:
--- a/cachi2/core/errors.py
+++ b/cachi2/core/errors.py
@@ -112,3 +112,17 @@ class GoModError(Cachi2Error):
         The output of the failing go command should provide more details, please check the logs.
         """
     ).strip()
+
+
+class YarnCommandError(Cachi2Error):
+    """The 'yarn' command used while processing a Yarn project returned an error."""
+
+    default_solution = textwrap.dedent(
+        """
+        The cause of the failure could be:
+        - something is broken in Cachi2
+        - something is wrong with your yarn project
+        - communication with an external service failed (please try again)
+        The output of the failing yarn command should provide more details, please check the logs.
+        """
+    ).strip()
--- a/cachi2/core/package_managers/yarn/resolver.py
+++ b/cachi2/core/package_managers/yarn/resolver.py
@@ -52,7 +52,7 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
     processed.
 
     :raises PackageRejected: if the validation fails.
-    :raises SubprocessCallError: if the 'yarn info' command fails.
+    :raises YarnCommandError: if the 'yarn info' command fails.
     """
     under_development = True
     if under_development:
--- a/cachi2/core/package_managers/yarn/utils.py
+++ b/cachi2/core/package_managers/yarn/utils.py
@@ -1,3 +1,6 @@
+import subprocess  # nosec
+
+from cachi2.core.errors import YarnCommandError
 from cachi2.core.rooted_path import RootedPath
 from cachi2.core.utils import run_cmd
 
@@ -9,7 +12,9 @@ def run_yarn_cmd(cmd: list[str], source_dir: RootedPath, env: dict[str, str]) ->
     :param source_dir: the directory in the repository containing the yarn source files.
     :param env: environment variables to be set during the command's execution.
 
-    :raises SubprocessCallError: if the command fails.
+    :raises YarnCommandError: if the command fails.
     """
-    run_cmd(cmd=["yarn", *cmd], params={"cwd": source_dir, "env": env})
-    return NotImplemented
+    try:
+        return run_cmd(cmd=["yarn", *cmd], params={"cwd": source_dir, "env": env})
+    except subprocess.CalledProcessError:
+        raise YarnCommandError(f"Yarn command failed: {' '.join(cmd)}")
--- a/cachi2/core/utils.py
+++ b/cachi2/core/utils.py
@@ -17,7 +17,7 @@ def run_cmd(cmd: Any, params: dict) -> str:
     :param dict params: keyword parameters for command execution
     :returns: the command output
     :rtype: str
-    :raises SubprocessCallError: if the command fails
+    :raises CalledProcessError: if the command fails
     """
     params.setdefault("capture_output", True)
     params.setdefault("universal_newlines", True)
--- a/tests/unit/package_managers/yarn/test_main.py
+++ b/tests/unit/package_managers/yarn/test_main.py
@@ -1,0 +1,26 @@
+from unittest import mock
+
+import pytest
+
+from cachi2.core.errors import YarnCommandError
+from cachi2.core.package_managers.yarn.main import _fetch_dependencies
+from cachi2.core.rooted_path import RootedPath
+
+
+@mock.patch("cachi2.core.package_managers.yarn.main.run_yarn_cmd")
+def test_fetch_dependencies(mock_yarn_cmd: mock.Mock, rooted_tmp_path: RootedPath) -> None:
+    source_dir = rooted_tmp_path
+    output_dir = rooted_tmp_path.join_within_root("cachi2-output")
+
+    mock_yarn_cmd.side_effect = YarnCommandError("berryscary")
+
+    with pytest.raises(YarnCommandError) as exc_info:
+        _fetch_dependencies(source_dir, output_dir)
+
+    mock_yarn_cmd.assert_called_once_with(
+        ["install", "--mode", "skip-build"],
+        source_dir,
+        {"YARN_GLOBAL_FOLDER": str(output_dir.join_within_root("deps", "yarn"))},
+    )
+
+    assert str(exc_info.value) == "berryscary"
--- a/tests/unit/test_errors.py
+++ b/tests/unit/test_errors.py
@@ -54,6 +54,21 @@ def test_gomod_error_friendly_msg() -> None:
     assert err.friendly_msg() == expect_msg
 
 
+def test_yarn_command_error_friendly_msg() -> None:
+    err = errors.YarnCommandError("Some yarn command failed")
+    expect_msg = dedent(
+        """
+        Some yarn command failed
+          The cause of the failure could be:
+          - something is broken in Cachi2
+          - something is wrong with your yarn project
+          - communication with an external service failed (please try again)
+          The output of the failing yarn command should provide more details, please check the logs.
+        """
+    ).strip()
+    assert err.friendly_msg() == expect_msg
+
+
 def test_fetch_error_friendly_msg() -> None:
     err = errors.FetchError("Failed to fetch something")
     expect_msg = dedent(


cachi2/core/package_managers/yarn/main.py
```diff
@@ -1,4 +1,5 @@
 import logging
```
eskultety, line 1: **Commenting on the commit message:**
You're missing any kind of commit message. Unless being a trivial change, the change should be described and explained in the commit message so that an 'external' contributor understands why we actually need the change and why the commit did it the way it had done it (unless it's truly a trivial change). There's no abstract accompanying the PR, there's only a vague reference in your branch naming hinting at STONEBLD-1788 which doesn't tell much external contributors.
I would suggest describing in the commit message what "regular workflow" is; Note that many people look and search primarily within commit history first before scrolling through the list of pull requests.

Also, naming the commit subject as "Fetch dependencies for regular workflow" doesn't say much about what the commit is about at all unless diving straight into the code, because the subject should IMO follow some logical structure, .e.g. 'package_managers: yarn: Fetch dependencies ...' which immediately hints at which part of the code the commit is touching.

slimreaper35: I think the docs should handle proper descriptions. Also, I would expect that any external contributor would dive into the code at first and then start to look up the commit history or not ?

eskultety: which docs? If you're referring to the yarnberry docs then you should at least link it and not expect people go figure out the corresponding section on their own.

Sure, but imagine a block of code changing a 100x. How would you know the complete history and reasoning behind a change if there's no commit message explaning it? The thing is unfortunately, you can't make assumptions on what people will or will not do, that's why we have commit history and commit messages -> for proper bookkeeping and reminding us of past decisions.

slimreaper35: The docs for yarn do not exist yet. I could add a link to yarn documentation, but again I would expect an external contributor to have some knowledge about that. :smile: 

Anyway I updated the commit messages to be more verbose.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -86,9 +88,11 @@ def _fetch_dependencies(source_dir: RootedPath, output_dir: RootedPath) -> None:
...
     :raises CalledProcessError: if the 'yarn install' command fails.
     """
-    # the output_dir is where the globalFolder must be set to
-    # the yarn commands can be called by using the core.utils.run_cmd function
-    pass
+    args = ["install", "--mode", "skip-build"]
+    try:
+        run_yarn_cmd(args, source_dir, {"YARN_GLOBAL_FOLDER": str(output_dir)})
+    except subprocess.CalledProcessError:
+        raise
```
brunoapimentel, line 95: Gomod has a specific [error](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/errors.py#L99) type for subprocess errors? Do we want a similar one for Yarn?

brunoapimentel (commented):
I also think we should unit test the `_fetch_dependencies` function, even though we'll likely end up with a very basic test that essentially guarantees that the function parameters are being handled correctly.

brunoapimentel:
Also, since this seems to be more straightforward than I imagined, wdyt about also checking the cache for the zero-installs workflow in this PR (STONEBLD-1789)? It also boils down to calling `yarn install`.

cachi2/core/package_managers/yarn/main.py
eskultety: IMO that would mainly depend on whether cachi2 wants to wrap different Yarn error codes which we'd have to pull from subprocess, IOW looking at the Gomod error, it's just a generic Cachi2Error which prints out some string and optionally provides a string formatted solution, so technically a generic Cachi2Error would be enough, but since there's already an existing precedent, you have a strong point.

chmeliik: Yeah, at least we want to re-raise this as some subclass of Cachi2Error, a new one for Yarn probably makes sense

cachi2/core/package_managers/yarn/main.py
```diff
@@ -84,11 +86,13 @@ def _fetch_dependencies(source_dir: RootedPath, output_dir: RootedPath) -> None:
...
     :param output_dir: the directory where the yarn dependencies will be downloaded to.
-    :raises SubprocessCallError: if the 'yarn install' command fails.
+    :raises CalledProcessError: if the 'yarn install' command fails.
     """
-    # the output_dir is where the globalFolder must be set to
-    # the yarn commands can be called by using the core.utils.run_cmd function
-    pass
+    args = ["install", "--mode", "skip-build"]
+    try:
+        run_yarn_cmd(args, source_dir, {"YARN_GLOBAL_FOLDER": str(output_dir)})
```
chmeliik, line 93: The GLOBAL_FOLDER should be output_dir/deps/yarn or output_dir/deps/yarnberry

It's probably fine to call it yarn, when (if) the time comes to implement Yarn v1 we can download those to deps/yarn-classic

cachi2/core/package_managers/yarn/main.py
```diff
@@ -86,9 +89,13 @@ def _fetch_dependencies(source_dir: RootedPath, output_dir: RootedPath) -> None:
...
-    # the output_dir is where the globalFolder must be set to
-    # the yarn commands can be called by using the core.utils.run_cmd function
-    pass
+    cachi2_output = output_dir.join_within_root("deps", "yarnberry")
+
+    args = ["install", "--mode", "skip-build"]
+    try:
+        run_yarn_cmd(args, source_dir, {"YARN_GLOBAL_FOLDER": str(cachi2_output)})
+    except subprocess.CalledProcessError:
+        raise YarnBerryError(reason="Failed to fetch dependencies using 'yarn install'")
```
chmeliik, line 98: IMO we should re-raise the CalledProcessError as YarnBerryError already in run_yarn_cmd

Otherwise every piece of code that uses run_yarn_cmd will have to do this

brunoapimentel:
We should also make the [run_yarn_cmd](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/yarn/utils.py#L15) function fully functional. It is currently still returning a `NotImplemented`.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -84,11 +86,15 @@ def _fetch_dependencies(source_dir: RootedPath, output_dir: RootedPath) -> None:
...
+    :raises YarnBerryError: if the 'yarn install' command fails.
     """
-    # the output_dir is where the globalFolder must be set to
-    # the yarn commands can be called by using the core.utils.run_cmd function
-    pass
+    cachi2_output = output_dir.join_within_root("deps", "yarnberry")
+
+    args = ["install", "--mode", "skip-build"]
+    try:
+        run_yarn_cmd(args, source_dir, {"YARN_GLOBAL_FOLDER": str(cachi2_output)})
```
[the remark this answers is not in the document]
chmeliik: I figure run_cmd should always do that https://github.com/containerbuildsystem/cachi2/pull/348/commits/99ce092e85edd965fd061bd822b68a2eeaf56805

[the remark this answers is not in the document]
brunoapimentel: Oh right, I overlooked that. So I guess it's just a matter of finishing `run_yarn_cmd` (this [comment](https://github.com/containerbuildsystem/cachi2/pull/342#issuecomment-1755628754)), and maybe it doesn't even need to return anything?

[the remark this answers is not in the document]
chmeliik: It does, I'll need the `yarn info` output :slightly_smiling_face:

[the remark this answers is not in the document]
chmeliik: btw you didn't overlook it, I'm changing that in my PR :smile:

cachi2/core/package_managers/yarn/main.py
```diff
@@ -84,11 +86,15 @@ def _fetch_dependencies(source_dir: RootedPath, output_dir: RootedPath) -> None:
...
+    except YarnBerryError:
+        raise
```
chmeliik, line 97: you can remove the try/except block now

```python
try:
    do_a_thing()
except ...:
    raise
```
is the same as just `do_a_thing()`

cachi2/core/package_managers/yarn/main.py
```diff
@@ -84,11 +86,15 @@ def _fetch_dependencies(source_dir: RootedPath, output_dir: RootedPath) -> None:
 
     :param source_dir: the directory in which the yarn command will be called.
     :param output_dir: the directory where the yarn dependencies will be downloaded to.
-    :raises CalledProcessError: if the 'yarn install' command fails.
+    :raises YarnBerryError: if the 'yarn install' command fails.
     """
-    # the output_dir is where the globalFolder must be set to
-    # the yarn commands can be called by using the core.utils.run_cmd function
-    pass
+    cachi2_output = output_dir.join_within_root("deps", "yarnberry")
```
chmeliik, line 91: Any opinions on the file path? It's not _technically_ public API since the output directory structure should be considered opaque, but users will most likely see these paths somewhere, e.g. source containers once RHTAP implements them

If Yarn Berry is _the_ version of Yarn for us, we should probably just call it deps/yarn

cachi2/core/package_managers/yarn/main.py
```diff
@@ -73,7 +73,7 @@ def _check_yarn_cache(source_dir: RootedPath) -> None:
     """Check the contents of the yarn cache using 'yarn install'.
 
     :param source_dir: the directory in which the yarn command will be called.
-    :raises SubprocessCallError: if the 'yarn install' command fails.
+    :raises CalledProcessError: if the 'yarn install' command fails.
```
chmeliik, line 78: In the next commit, please change all occurrences of `:raises CalledProcessError:` in `cachi2/core/package_managers/yarn/*` to `:raises YarnBerryError:` :pray: 

All the methods will be using the run_yarn_cmd function, which is changed in the next commit to raise the more specific error

cachi2/core/package_managers/yarn/main.py
eskultety: > If Yarn Berry is the version of Yarn for us, we should probably just call it deps/yarn

Without having literally any knowledge of Yarn, if there are multiple versions/flavours of Yarn packaging you already do support or may support in the future `deps/yarn` will bite us in the back for sure. We should either refer to it by the official name (some lowercase name variant) or yarn + version if those are a thing by which it could be determined whether it's Yarn berry or something else.

eskultety: As for users, if it's not part of the official public API, it doesn't matter at all as the paths are bound to change in that case, so whatever they do on them/expect from them is irrelevant in that case, so we're free to alter it to our liking. Of course the question then is whether we should make these deps paths officially part of the API so we guarantee them - which we should do if we refer to them in any shape or form in the docs... just my 2c

chmeliik: There's yarn v1: https://classic.yarnpkg.com/en/docs
And yarn v2+: https://yarnpkg.com/

The official name for both is Yarn, but they might as well be two entirely different package managers. V1 is sometimes referred to as Yarn Classic, v2+ as Yarn Berry.

Yarn v1 is supported in Cachito but not in Cachi2, v2+ is not supported in Cachito and will be in Cachi2. There is _some_ hope that we wouldn't have to implement v1 in Cachi2, but I'm not among the optimists :smile: 

In Cachito Yarn v1 deps went to `deps/yarn`, but I don't think we need to keep backwards compatibility with those, they never mattered to any system or user AFAIK.

chmeliik: > Of course the question then is whether we should make these deps paths officially part of the API so we guarantee them - which we should do if we refer to them in any shape or form in the docs... just my 2c

True, we should probably consider changes in the path as breaking changes

eskultety: > I think we should stick with `deps/yarn`. The reasoning is that we'd better stick to a single term throughout the upstream project, either `yarn` or `yarnberry`, and it feels like we are going for the former. It is how it is commonly referred to, AFAICT (simply `yarn`).

I don't disagree but in case the Yarn community decides to go full Google mode and create another fork to supersede yarn berry and make it 'the yarn to go to', then we'd get ourselves into a pickle, especially if we decide to make the mentioned path part of the API. But again, maybe I'm just being overly cautious and thinking too far ahead, ultimately, the decision is a collective one, so if the consensus is that `deps/yarn` is okay, then it's fine by me.

chmeliik: Ultimately, for the user, what is the impact of whatever we choose?

* slight confusion over why the file path includes `yarn-with-a-suffix` rather than `yarn` (if I ever look at those)

Anything else?

chmeliik: There's a separate matter of what do we call the package manager in the cachi2 input (now this very much _is_ public API)

`cachi2 fetch-deps yarn`?

`cachi2 fetch-deps yarnberry`?

But maybe cachi2 should be smart, accept only `yarn` and auto-detect which yarn it is. We had a story in the backlog to figure this out, but deferred because yarn v1 support may not be needed.

chmeliik: TLDR: `deps/yarn` is fine by me :smile:

chmeliik (commented):
LGTM apart from the output directory name :+1:

cachi2/core/package_managers/yarn/main.py
slimreaper35: Did not expect such a long discussion about the file path :smile: 

Should I rename YarnBerryError to YarnError to keep it consistent?

taylormadore: > Should I rename YarnBerryError to YarnError to keep it consistent?

To me, that makes sense :+1:

chmeliik: > Should I rename YarnBerryError to YarnError to keep it consistent?

Yeah, I think dropping Berry from the name makes sense. The error message is probably generic enough to be usable for all flavours of yarn anyway. Maybe call it YarnCommandError though, YarnError seems too generic

taylormadore (approved):
LGTM :+1:

chmeliik:
For future reference, here's what I imagine a commit message for a change like this could look like

```
Use 'yarn install --mode=skip-build' to download dependencies for
projects that do not use zero-installs. That is, projects which do not
contain a cacheFolder directory (default .yarn/cache/) with the
dependencies already downloaded. Such projects will be handled
separately.

While running the command, set the YARN_GLOBAL_FOLDER environment
variable to '$cachi2_output_dir/deps/yarn' to make sure yarn downloads
the dependencies into this directory.

The --mode=skip-build option is required to avoid executing the
preinstall, install and postinstall scripts if the project has any.
Such scripts are arbitrary code, which we should not execute.
```

Though some parts of this probably wouldn't be necessary if they were code comments

The current commit message also says YarnBerryError rather than YarnCommandError

slimreaper35:
Maybe we could start using any commit message template or pre-commit or something that is automated.

I scrolled down commit history a little bit and I rarely saw such a verbose commit message (except for your commits 😄 ). Mostly it was a JIRA ticket number and even for larger PRs than this one.

WDYT ?

eskultety:
> For future reference, here's what I imagine a commit message for a change like this could look like

To take effect, this needs to be defined in some kind of contributor guidelines which we currently don't have, but yeah, totally in agreement.

eskultety:
> Maybe we could start using any commit message template or pre-commit or something that is automated.

You'd have to use AI to provide any meaningful gist for the commit subject, we're not there yet :). Writing good commit messages is a skill worth having rather than trying to automate it. Think of it as a way of explaining your changes to a real person.

> 
> I scrolled down commit history a little bit and I rarely saw such a verbose commit message (except for your commits 😄 ). Mostly it was a JIRA ticket number and even for larger PRs than this one.

The fact that the commit history is scarce in detailed information in this case must not serve as a precedent to adopt an improvement in the future. Projects and communities evolve gradually :).

slimreaper35:
I am sorry, I meant that the check of the commit message would automated at least at some form. I think it would be fine for consistency of all commits. Since check in the PR template is quite subjective.

Second point, I did not want to sound it like an excuse, but I feel like I am the only one who is blamed for commit messages.

chmeliik:
> I am sorry, I meant that the check of the commit message would automated at least at some form. I think it would be fine for consistency of all commits. Since check in the PR template is quite subjective.
> 
> Second point, I did not want to sound it like an excuse, but I feel like I am the only one who is blamed for commit messages.

I don't think verbosity is always needed, but the message does need to describe what the change does and explain things that are not obvious from the code or comments. For more complex changes, the message should help the reviewer understand what they're getting into.

It absolutely is subjective as you said, and we're not always good at enforcing the checklist. Sorry if it looks like you're the only one getting called out for non-descriptive commits :disappointed: (I'm pretty sure others do as well)

Using a commit message template could be a good idea :+1:

chmeliik (approved):
LGTM :+1:

slimreaper35:
One more thing...In this PR I should split the one commit into probably two (implement YarnCommandError and implement fetching), but usually I make changes along the discussion (which is completely fine), and then I would have to rearrange commits as well, which is not always straightforward so it leads to one or more big commits where I have to explain everything. I wish I could do it better from the beginning. Thats the skill and I want to have. I think it would not lead to those kind of discussions.

But anyaway, I agree that something like commit template is a good idea. Thats my final words. :drop-the-mike:

chmeliik:
> One more thing...In this PR I should split the one commit into probably two (implement YarnCommandError and implement fetching), but usually I make changes along the discussion (which is completely fine), and then I would have to rearrange commits as well, which is not always straightforward so it leads to one or more big commits where I have to explain everything. I wish I could do it better from the beginning. Thats the skill I want to have. I think it would not lead to those kind of discussions.

Do you use `git rebase --interactive`? That's what I do to re-arrange commits, insert a new commit between old ones, edit an older commit etc.

> But anyaway, I agree that something like commit template is a good idea. Thats my final words. :drop-the-mike:

Can you look into that, or bring it up in a team discussion?

PR #343, opened 2023-10-09
ensure packageManager is correctly set for yarn

taylormadore:
This change depends on https://github.com/containerbuildsystem/cachi2/pull/338

STONEBLD-1776

Since we will rely on corepack to download and configure
the correct version of yarn for a request, ensure that
packageManager is defined in package.json.

Use either yarnPath from .yarnrc, packageManager, or a
combination of the two to set the correct yarn version.

Raise exceptions for the following cases:
  - packageManager is already set, but we can't parse it
  - neither yarnPath or packageManager is set (or parseable)
  - the versions specified by yarnPath and packageManager
    are different

### DIFF
--- a/cachi2/core/package_managers/yarn/main.py
+++ b/cachi2/core/package_managers/yarn/main.py
@@ -1,8 +1,13 @@
 import logging
 
+from cachi2.core.errors import PackageRejected
 from cachi2.core.models.input import Request
 from cachi2.core.models.output import Component, EnvironmentVariable, RequestOutput
-from cachi2.core.package_managers.yarn.project import Project
+from cachi2.core.package_managers.yarn.project import (
+    Project,
+    get_semver_from_package_manager,
+    get_semver_from_yarn_path,
+)
 from cachi2.core.package_managers.yarn.resolver import (
     create_component_from_package,
     resolve_packages,
@@ -52,10 +57,41 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Comp
 def _configure_yarn_version(project: Project) -> None:
     """Resolve the yarn version and set it in the package.json file if needed.
 
-    :raises PackageRejected: in case the yarn version can't be determined, or if there is a
-        mismatch between the version in package.json and yarnrc.yml.
+    :raises PackageRejected:
+        if the yarn version can't be determined from either yarnPath or packageManager
+        if there is a mismatch between the yarn version specified by yarnPath and PackageManager
     """
-    pass
+    yarn_path_version = get_semver_from_yarn_path(project.yarn_rc.yarn_path)
+    package_manager_version = get_semver_from_package_manager(project.package_json.package_manager)
+
+    if not yarn_path_version and not package_manager_version:
+        raise PackageRejected(
+            "Unable to determine the yarn version to use to process the request",
+            solution=(
+                "Ensure that either yarnPath is defined in .yarnrc.yml or that packageManager "
+                "is defined in package.json"
+            ),
+        )
+
+    if (
+        yarn_path_version
+        and package_manager_version
+        and yarn_path_version != package_manager_version
+    ):
+        raise PackageRejected(
+            (
+                f"Mismatch between the yarn versions specified by yarnPath (yarn@{yarn_path_version}) "
+                f"and packageManager (yarn@{package_manager_version})"
+            ),
+            solution=(
+                "Ensure that the versions of yarn specified by yarnPath in .yarnrc.yml and "
+                "packageManager in package.json agree"
+            ),
+        )
+
+    if not package_manager_version:
+        project.package_json.package_manager = f"yarn@{yarn_path_version}"
+        project.package_json.write_to_file()
 
 
 def _set_yarnrc_configuration(project: Project, output_dir: RootedPath) -> None:
--- a/cachi2/core/package_managers/yarn/project.py
+++ b/cachi2/core/package_managers/yarn/project.py
@@ -4,9 +4,18 @@ Parse the relevant files of a yarn project.
 It also provides basic utility functions. The main logic to resolve and prefetch the dependencies
 should be implemented in other modules.
 """
+import json
+import logging
+import re
+from pathlib import Path
 from typing import Any, NamedTuple, Optional
 
+import semver
+
+from cachi2.core.errors import UnexpectedFormat
 from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
 
 
 class YarnRc:
@@ -91,10 +100,21 @@ class PackageJson:
         """Get the package manager string."""
         return NotImplemented
 
+    @package_manager.setter
+    def package_manager(self, package_manager: str) -> None:
+        """Set the package manager string."""
+        self._data["packageManager"] = package_manager
+
     @classmethod
     def from_file(cls, file_path: RootedPath) -> "PackageJson":
         """Parse the content of a package.json file."""
         return NotImplemented
+
+    def write_to_file(self) -> None:
+        """Write the data to the package.json file."""
+        with self._path.path.open("w") as f:
+            json.dump(self._data, f, indent=2)
+            f.write("\n")
 
 
 class Project(NamedTuple):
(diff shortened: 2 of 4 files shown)


cachi2/core/package_managers/yarn/project.py
```diff
@@ -91,11 +97,21 @@ def package_manager(self) -> Optional[str]:
...
+    def package_manager(self, package_manager: str) -> None:
+        """Set the package manager string."""
+        self._data["packageManager"] = package_manager
+
     @classmethod
     def from_file(cls, file_path: RootedPath) -> "PackageJson":
         """Parse the content of a package.json file."""
         return NotImplemented
 
+    def to_file(self) -> None:
```
brunoapimentel, line 110: Would it make sense to call `yarn set version` instead of writing the file manually?

cachi2/core/package_managers/yarn/project.py
```diff
@@ -91,11 +97,21 @@ def package_manager(self) -> Optional[str]:
         """Get the package manager string."""
         return NotImplemented
 
+    @package_manager.setter
+    def package_manager(self, package_manager: str) -> None:
```
brunoapimentel, line 101: Not sure how far we want to take the concept of immutability, but an alternative to this setter is adding a utility method such as [this](https://github.com/brunoapimentel/cachi2/commit/6879b1e5ea56cdaf0fa341cfd7ce51df347b244d#diff-482c5a124171fdd52977104dd909f1a274ec6debcd844ea4f60ba03f37d21de5R89-R93).

cachi2/core/package_managers/yarn/project.py
```diff
@@ -126,3 +142,46 @@ def yarn_cache(self) -> RootedPath:
     def from_source_dir(cls, source_dir: RootedPath) -> "Project":
         """Create a Project from a sources directory path."""
         return NotImplemented
+
+
+def get_semver_from_yarn_path(yarn_path: str) -> Optional[semver.version.Version]:
```
brunoapimentel, line 147: Nitpick: if we're keeping these methods in the `project.py` file, maybe they can be methods in `YarnRc` and `PackageJson`? 

I feel they are a little more than just utility methods, since they have some yarn business logic copied into them, but since we're still doing the bulk of the processing in `main.py`, we're still keeping this module's ethos IMO.

Alternatively, they could be private methods in main.py.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -126,3 +142,46 @@ def yarn_cache(self) -> RootedPath:
...
+    name, version = match.groups()
+    if name != "yarn":
+        raise UnexpectedFormat("packageManager in package.json must be yarn")
+
+    try:
+        return semver.version.Version.parse(version)
+    except ValueError as e:
+        raise UnexpectedFormat(
+            f"{version} is not a valid semver for packageManager in package.json"
+        ) from e
```
chmeliik, line 187: Is semver.version.Version able to parse every valid npm semver version?

If not, we can leave the validation to corepack

cachi2/core/package_managers/yarn/project.py
chmeliik: I think `yarn set version` does a lot more than what we want, like downloading the `yarn` release that you specify

cachi2/core/package_managers/yarn/project.py
taylormadore: I tested the usual suspects, but I can do a more extensive verification. The main reason for this check was to ensure that we can do a comparison with the yarnPath semver (if necessary) later in _configure_yarn_version

cachi2/core/package_managers/yarn/project.py
brunoapimentel: If it downloads the yarn release, would it be able to replace Corepack?

cachi2/core/package_managers/yarn/project.py
brunoapimentel: That's a good point. We could disassociate the concept of `_path` from the class, and have all in-memory representations of the file that could be written to any path. But at this point, I think let's just make it mutable. :sweat_smile:

cachi2/core/package_managers/yarn/project.py
chmeliik: It just saves the `.cjs` file in`yarnPath` AFAIK (but proably does other things as well, this was just an example)

chmeliik (commented):
Nice work :tada: 

I didn't realize how important it was to compare the versions as semver.VersionInfo until I saw the +sha case

cachi2/core/package_managers/yarn/main.py
```diff
@@ -51,10 +56,32 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Comp
...
+        and yarn_path_version != package_manager_version
+    ):
+        raise PackageRejected(
+            "Unable to determine the yarn version to use to process the request",
+            solution="Ensure that the yarnPath version in .yarnrc and the packageManager version in package.json agree",
+        )
+
+    if not package_manager_version:
+        project.package_json.package_manager = f"yarn@{yarn_path_version}"
+        project.package_json.to_file()
```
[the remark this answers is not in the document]
taylormadore: Good point, I'll change it to write_to_file if that sounds good :+1:

cachi2/core/package_managers/yarn/project.py
```diff
@@ -126,3 +143,54 @@ def yarn_cache(self) -> RootedPath:
...
+    yarn_spec_pattern = re.compile(r"^yarn-(.+)\.cjs$")
+    match = yarn_spec_pattern.match(Path(yarn_path).name)
+    if not match:
+        return None
+
+    yarn_version = match.group(1)
+    try:
+        return semver.version.Version.parse(yarn_version)
+    except ValueError:
+        return None
```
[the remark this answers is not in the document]
taylormadore: Yeah, I agree. Some reaction/notification seems appropriate here. I suppose I lean towards a warning rather than a fail because even though we can't parse yarnPath, packageManager may still be usable.

Does that sound good?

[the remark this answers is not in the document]
chmeliik: The concern is that the yarn version specified in `packageManager` may behave differently than the one in `yarnPath`. But yeah, a warning is probably enough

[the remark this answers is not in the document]
taylormadore: Added :+1:

cachi2/core/package_managers/yarn/project.py
brunoapimentel: Well, since it seems no one has a strong preference either way, I'll leave my comment as suggestions, and go with whatever you feel it's best (including the current code).

chmeliik (approved):
LGTM :+1:

tests/unit/package_managers/yarn/test_main.py
```diff
@@ -0,0 +1,93 @@
...
+    yarn_path_version: Optional[semver.version.Version],
+    package_manager_version: Optional[semver.version.Version],
+) -> None:
+    mock_project = mock.Mock()
+    mock_yarn_path_semver.return_value = yarn_path_version
+    mock_package_manager_semver.return_value = package_manager_version
+
+    _configure_yarn_version(mock_project)
+
+    if package_manager_version is None:
```
brunoapimentel, line 45: Small detail, but should we assert that the `package_manager_version` stays the same and the file was not written in case `package_manager_version is None` resolves to False? Otherwise, the test is only making sure that the function gets executed without errors for 3 of the scenarios.

brunoapimentel (approved):
Left a minor comment, but otherwise LGTM

tests/unit/package_managers/yarn/test_main.py
taylormadore: Added something for this :+1: 

I think the right way to do this would be an assertion that the setter for the package_manager property was not called, but I played around with it for a while and couldn't figure out how to make it work :man_facepalming: 

If anyone else knows how to do it, let me know

taylormadore:
Rebased on main

tests/unit/package_managers/yarn/test_main.py
chmeliik: This LGTM

PR #347, opened 2023-10-10
Makefile changes to improve (slightly) local development experience

eskultety:

- n/a Code coverage from testing does not decrease and new code is covered
---

- optimizes the `clean` target
- drops the `virtualenv` dependency
- drops the hard 3.9 dependency on the local virtual environments

The last 2 patches are a bit controversial in how they address creation of the virtual environments:
- venv doesn't default to python3.9 anymore because whatever the default python3 on the local system is should do just fine
- force only the `pip-compile` target to Python3.9 as that is kind of a hard dependency that would otherwise potentially cause a regression

I'm open to other suggestions for the ^above as unarguably it's not a particularly nice approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,15 +5,22 @@ TOX_ARGS ?=
 .PHONY: clean
 all: venv
 
+define make_venv
+	$(PYTHON_BIN) -m venv $(VENV)
+	$(VENV)/bin/pip install --upgrade pip
+endef
+
 clean:
 	rm -rf dist venv .tox *.egg-info *.log*
 
+
+venv: PYTHON_BIN := $(shell which $(PYTHON_VERSION_VENV))
+venv: VENV := venv
 venv:
-	$(shell which $(PYTHON_VERSION_VENV)) -m venv venv
-	venv/bin/pip install --upgrade pip
-	venv/bin/pip install -r requirements.txt
-	venv/bin/pip install tox
-	venv/bin/pip install -e .
+	$(call make_venv)
+	$(VENV)/bin/pip install -r requirements.txt
+	$(VENV)/bin/pip install tox
+	$(VENV)/bin/pip install -e .
 
 test: venv
 	venv/bin/tox
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PYTHON_VERSION_VENV ?= python3.9
+PYTHON_VERSION_VENV ?= python3
 TOX_ENVLIST ?= py39
 TOX_ARGS ?=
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -15,19 +15,19 @@ venv:
 	venv/bin/pip install tox
 	venv/bin/pip install -e .
 
-test:
+test: venv
```
chmeliik, line 23: Should we un-PHONY the venv so that it doesn't get re-created on every `make test` etc.?

Makefile
```diff
@@ -7,7 +7,7 @@ all: venv
 clean:
 	rm -rf dist venv .tox *.egg-info *.log*
 
-.PHONY: venv
+.PHONY: all clean venv
```
chmeliik, line 10: nitpick: if we're going to have one .PHONY, can it be at the top of the file?

alternatively, I think you can attach a .PHONY to each target

```makefile
.PHONY: clean
clean:
    ...

.PHONY: venv
venv:
    ...
```

README.md
```diff
@@ -128,7 +128,7 @@ when downloading Go modules. See [Go environment variables](https://go.dev/ref/m
 
 ## Development
 
-### Virtualenv
+### Virtua environment
```
chmeliik, line 131: typo in Virtual

Makefile
```diff
@@ -1,4 +1,4 @@
-PYTHON_VERSION_VENV ?= python3.9
+PYTHON_VERSION_VENV ?= python3
```
chmeliik, line 1: Personally I will keep using py3.9 for development until we drop support for it (but I'm fine with that not being the default, I can set it locally)

(I prefer earlier feedback that what I'm using doesn't exist in 3.9 as opposed to finding out only once I run the unit tests)

chmeliik: And I found out that the issue in f1915e55e7ee115a5aed0ebd8581b070bb608f5b was in fact caused by tox misconfiguration. The way we have it configured, we don't run tests for py3.9, py3.10 and py3.11, we run tests for [whatever version of python you use to invoke `tox`] 3 times

I can fix that in a separate PR

Makefile
chmeliik: There's also the pip install steps, but yeah those also just check that everything is already installed and then re-install cachi2

All in all it's reasonably fast

Makefile
eskultety: go ahead if you want, but I have those changes on a separate branch as the changes were not  strictly related to the Makefile

Makefile
eskultety: I tried your suggestion anyway, but for some reason even if I drop venv from .PHONY make will still run the target every single time, despite the directory being present. I suspect the `--make-venv` target which doesn't have any dependency and also doesn't create a file with the same name which makes this intermediary target phony too.
So, I may have found a cleaner solution:

```
define make_venv
    $(PYTHON_BIN) -m venv $(VENV) 
    $(VENV)/bin/pip install --upgrade pip                                 
endef

...
venv:
    $(call make_venv)
...
````

By using a function instead of an intermediary target, we don't need to worry users being able to execute it. I'll update the PR accordingly.

By using a function I could also test dropping 'venv' from .PHONY. It did work, but note that the only way to re-run the 'venv' target in that case is if the directory is touched or removed, so if the contents get corrupt the target won't re-run, but I guess that's expected.

Makefile
chmeliik: ah great, I'll leave it to you then :slightly_smiling_face:

eskultety:
Since v1:
- replaced the ugly `--make_venv` pseudoprivate target with a Makefile function
- dropped 'venv' from .PHONY skipping creation and development re-installs of cachi2 if the virtual env exists already
- moved .PHONY definition to the top of the file

chmeliik (approved):
LGTM :+1:

chmeliik:
/ok-to-test

(I _think_ that's the command to approve the RHTAP pipeline)

brunoapimentel (approved):
LGTM!

Thanks for this, it was very informative. I realize now I don't know the first thing about Makefiles :sweat_smile:

brunoapimentel:
@eskultety Do you have merge permission? If not, can you rebase this and I'll merge it for you?

eskultety:
> @eskultety Do you have merge permission? If not, can you rebase this and I'll merge it for you?

I don't. Rebased.

PR #348, opened 2023-10-10
Yarnberry: parse yarn info output

chmeliik:
STONEBLD-1778

Implement the resolver.resolve_packages function.

Parse the output of `yarn info --all --recursive --cache --json`, pick
out the relevant attributes, post-process them and return them as
Packages.

Alter the Package class slightly:
- make version Optional, use None if yarn info reports 0.0.0-use.local
- make checksum and cache_path both individually Optional, there are
  cases where checksum is null but cache_path isn't
- move checksum and cache_path directly to the Package class
- make locator a string (and rename to raw_locator), because parsed
  Locators don't make a good identifier for error messages (they omit
  unnecessary attributes, so a string representation of a Locator still
  wouldn't be grep-able in the yarn.lock file)
- add parsed_locator property

Remove the unnecessary vet_git_dependencies function, instead validate
that all locators are supported (git locators are not).

- N/A Docs updated (if applicable)
- N/A Docs links in the code are still valid (if docs were updated)

### DIFF
--- a/cachi2/core/utils.py
+++ b/cachi2/core/utils.py
@@ -2,7 +2,7 @@ import json
 import logging
 import re
 import subprocess  # nosec
-from typing import Any, Iterator
+from typing import Any, Iterator, Optional
 
 from cachi2.core.config import get_config
 
@@ -31,10 +31,20 @@ def run_cmd(cmd: Any, params: dict) -> str:
     try:
         response.check_returncode()
     except subprocess.CalledProcessError:
-        log.error('The command "%s" failed with: %s', " ".join(cmd), response.stderr)
+        log.error('The command "%s" failed', " ".join(cmd))
+        _log_error_output("STDERR", response.stderr)
+        if not response.stderr:
+            _log_error_output("STDOUT", response.stdout)
         raise
 
     return response.stdout
+
+
+def _log_error_output(out_or_err: str, output: Optional[str]) -> None:
+    if output:
+        log.error("%s:\n%s", out_or_err, output.rstrip())
+    else:
+        log.error("%s: <empty>", out_or_err)
 
 
 def load_json_stream(s: str) -> Iterator:
--- a/hack/mock-unittest-data/yarn.sh
+++ b/hack/mock-unittest-data/yarn.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+
+cat << banner-end
+--------------------------------------------------------------------------------
+Generating mock data for yarn unit tests
+--------------------------------------------------------------------------------
+banner-end
+
+tmpdir=$(dirname "$(mktemp --dry-run)")
+
+git clone https://github.com/cachito-testing/cachi2-yarn-berry \
+    --depth=1 \
+    --single-branch \
+    --branch=zero-installs \
+    "$tmpdir/cachi2-yarn-berry"
+trap 'rm -rf "$tmpdir/cachi2-yarn-berry"' EXIT
+
+cd "$tmpdir/cachi2-yarn-berry"
+
+yarn info --all --recursive --cache --json |
+    # filter out unsupported protocols
+    jq 'select(.value | test("commit=") | not)' > yarninfo.json
+
+# - take 1 or 2 examples of each supported protocol
+# - make sure to include the curious case where
+#   checksum is null but path isn't
+jq -s < yarninfo.json '
+    map(select(.value | test("@npm:")))[0],
+    map(select(.value | test("@workspace:")))[0,1],
+    map(select(.value | test("@patch:")))[0,1],
+    map(select(.value | test("@file:")))[0,1],
+    map(select(.value | test("@portal:")))[0],
+    map(select(.value | test("@link:")))[0],
+    map(select(.value | test("@https:.*tar.gz")))[0],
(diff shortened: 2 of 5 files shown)


chmeliik:
Slightly improved the run_cmd error logging and added unit tests for it

chmeliik:
cc @eskultety

We'll need to get you added to the containerbuildsystem org (you can ask @mkosiarc, but not sure if you still want to create that separate GH account)

hack/mock-unittest-data/yarn.sh
```diff
@@ -0,0 +1,51 @@
...
+cat << banner-end
+--------------------------------------------------------------------------------
+Generating mock data for yarn unit tests
+--------------------------------------------------------------------------------
+banner-end
+
+tmpdir=$(dirname "$(mktemp --dry-run)")
+
+git clone https://github.com/cachito-testing/cachi2-yarn-berry \
+    --branch=zero-installs \
```
chmeliik, line 13: I incorporated the changes from https://github.com/cachito-testing/cachi2-yarn-berry/pull/2

Will update if necessary once (if) that is merged

chmeliik:
Now using more sensible ordering for the keys in the mocked data (the unchanged `yarn info` ordering)

eskultety:
> cc @eskultety
> 
> We'll need to get you added to the containerbuildsystem org (you can ask @mkosiarc, but not sure if you still want to create that separate GH account)

Okay. Nah, tweaked my mail filters for the notifications already instead. Adding me to the org surely will make things easier for me as I'll be able to drop the filters then, but I won't be creating a new GH account for now just for this.

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def mock_yarn_info_output(yarn_info_outputs: list[dict[str, Any]]) -> str:
+    yarn_info_string_output = "\n".join(
+        json.dumps(obj, separators=(",", ":")) for obj in yarn_info_outputs
+    )
+    return yarn_info_string_output + "\n"
+
+
+# re-generate using hack/mock-unittest-data/yarn.sh
```
lkolacek, line 20: Do we want to manually fix expected data once they are changed? Would it make sense to make this update part of generate-test-data in Makefile?

lkolacek:
LGTM :+1:

tests/unit/package_managers/yarn/test_resolver.py
chmeliik: IMO not really.
- `make generate-test-data` generates the expected output for integration tests.
- `hack/mock-unittest-data/yarn.sh` and `hack/mock-unittest-data/gomod.sh` generate the _input_ for _unit_ tests

There should rarely be a need to update those inputs, but the scripts are there in case you need them

lkolacek: OK agree, we are instructed in a comment that such script to update data is included here, it's enough. :)

taylormadore (approved):
LGTM :+1:

chmeliik:
Rebased on main, resolved conflicts

eskultety (commented):
can't comment on the class attributes and implementation yet, so I have to leave those bits to others.

cachi2/core/utils.py
```diff
@@ -31,12 +31,21 @@ def run_cmd(cmd: Any, params: dict) -> str:
     try:
         response.check_returncode()
     except subprocess.CalledProcessError:
-        log.error('The command "%s" failed with: %s', " ".join(cmd), response.stderr)
+        log.error('The command "%s" failed', " ".join(cmd))
+        _log_error_output("STDOUT", response.stdout)
+        _log_error_output("STDERR", response.stderr)
```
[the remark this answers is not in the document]
chmeliik: :+1: fallback on empty stderr sounds better

chmeliik:
Changed the logging of failed subprocesses to only log stdout if stderr is empty

added --depth=1 and --single-branch to the yarn.sh mocking script

hack/mock-unittest-data/yarn.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
chmeliik: Good call, it is more readable. I'll fix that in a follow-up PR if you don't mind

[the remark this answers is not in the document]
eskultety: Since this change is related to generating test data sets and hence not crucial to the core logic being added (from functionality POV) I'm totally fine if you submit a follow-up PR for this.
I would have used a different argument if the new file were crucial to the core functionality though :) .

hack/mock-unittest-data/yarn.sh
```diff
@@ -0,0 +1,57 @@
...
+--------------------------------------------------------------------------------
+Generating mock data for yarn unit tests
+--------------------------------------------------------------------------------
+banner-end
+
+tmpdir=$(dirname "$(mktemp --dry-run)")
+
+git clone https://github.com/cachito-testing/cachi2-yarn-berry \
+    --depth=1 \
+    --single-branch \
```
eskultety, line 14: `--depth` implies `--single-branch` so technically you don't need it. However, explicit is usually better than implicit, so let's keep it this way.

eskultety:
LGTM, still though, someone else needs to grant the actual approval.

PR #352, opened 2023-10-11
Yarnberry: resolve package names and versions

chmeliik:
STONEBLD-1790

The entries returned by 'yarn info' do not necessarily contain the
correct names and versions of packages.

NAMES:
    Yarnberry lets the user specify an arbitrary name whenever the user
    adds any dependency other than an npm package or a workspace. The
    locator returned by 'yarn info' uses the user-specified name.

VERSIONS:
    When reporting local (soft-link, i.e. workspace, portal, link)
    packages, 'yarn info' uses '0.0.0-use.local' as the version.

Resolve the actual names and versions of packages from their
package.json files when necessary.

HARD-LINK:
* npm: trust the 'yarn info' data, it seems reliable
* Patch, File, Https: trust the version, resolve the name from the
  package.json in the cache archive
* Patch without a cache archive: try to resolve the name of the
  original package

SOFT-LINK:
* Workspace: trust the name, resolve the version from package.json
* Portal: resolve the name and version from package.json
* Link: if they happen to have a package.json, treat them the same as
  portal packages, otherwise use the name from the locator and don't
  report any version

### DIFF
--- a/cachi2/core/package_managers/yarn/resolver.py
+++ b/cachi2/core/package_managers/yarn/resolver.py
@@ -4,19 +4,39 @@ Resolve the dependency list for a yarn project.
 It also performs the necessary validations to avoid allowing an invalid project to keep being
 processed.
 """
+import json
 import logging
+import zipfile
 from dataclasses import dataclass
 from functools import cached_property
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Union
 
 import pydantic
 from packageurl import PackageURL
 
-from cachi2.core.errors import UnsupportedFeature
+from cachi2.core.errors import PackageRejected, UnsupportedFeature
 from cachi2.core.models.sbom import Component
-from cachi2.core.package_managers.yarn.locators import Locator, parse_locator
+from cachi2.core.package_managers.yarn.locators import (
+    FileLocator,
+    HttpsLocator,
+    LinkLocator,
+    Locator,
+    NpmLocator,
+    PatchLocator,
+    PortalLocator,
+    WorkspaceLocator,
+    parse_locator,
+)
 from cachi2.core.package_managers.yarn.project import Optional, Project
 from cachi2.core.package_managers.yarn.utils import run_yarn_cmd
 from cachi2.core.rooted_path import RootedPath
+
+if TYPE_CHECKING:
+    # Import conditionally so that we don't have to introduce a runtime dependency on
+    # typing-extensions. This is only imported when a type-checker is running.
+    # In python 3.11, it can be imported directly from the stdlib 'typing' module.
+    from typing_extensions import assert_never
 
 log = logging.getLogger(__name__)
 
@@ -151,6 +171,10 @@ class _ResolvedPackage:
     raw_locator: str
 
 
+class _CouldNotResolve(ValueError):
+    """_ComponentResolver failed to resolve the name or version of a package."""
+
+
 class _ComponentResolver:
     def __init__(self, project: Project, output_dir: RootedPath) -> None:
         self._project = project
@@ -158,7 +182,17 @@ class _ComponentResolver:
 
     def get_component(self, package: Package) -> Component:
         """Create an SBOM component for a yarn Package."""
-        resolved_package = self._resolve_package(package)
+        try:
+            resolved_package = self._resolve_package(package)
+        except _CouldNotResolve as e:
+            raise PackageRejected(
+                f"Failed to resolve the name and version for {package.raw_locator}: {e}",
+                solution=(
+                    "Please try running 'yarn install' to see if yarn makes any changes.\n"
+                    "If yarn succeeds and doesn't make any changes, please report this Cachi2 bug."
+                ),
+            ) from e
+
         return Component(
             name=resolved_package.name,
             version=resolved_package.version,
@@ -167,12 +201,118 @@ class _ComponentResolver:
 
     def _resolve_package(self, package: Package) -> _ResolvedPackage:
         """Resolve the real name and version of the package."""
-        return _ResolvedPackage(
-            locator=package.parsed_locator,
-            name="placeholder",
-            version=package.version,
-            raw_locator=package.raw_locator,
-        )
+
+        def log_for_locator(msg: str, *args: Any, level: int = logging.DEBUG) -> None:
+            log.log(level, f"%s: {msg}", package.raw_locator, *args)
+
+        locator = package.parsed_locator
+
+        if isinstance(locator, NpmLocator):
+            # npm dependencies have reliable names and versions in yarn info output
+            name = self._scoped_name(locator)
+            version = package.version
+        elif isinstance(locator, WorkspaceLocator):
+            packjson = self._project_subpath(locator.relpath, "package.json")
+            log_for_locator("reading package version from %s", packjson.subpath_from_root)
+            # workspace dependencies have reliable names but report '0.0.0-use.local' as the version
+            name = self._scoped_name(locator)
+            _, version = self._read_name_version_from_packjson(packjson)
+        elif isinstance(locator, (PatchLocator, FileLocator, HttpsLocator)):
+            if not package.cache_path:
+                raise _CouldNotResolve(
+                    "expected a zip archive in the cache but 'yarn info' says there is none",
+                )
+            cache_path = self._cache_path_as_rooted(package.cache_path)
+            if not cache_path.path.exists():
+                raise _CouldNotResolve(
+                    f"cache archive does not exist: {cache_path.subpath_from_root}"
+                )
+            log_for_locator("reading package name from %s", cache_path.subpath_from_root)
+            # patch, file and https dependencies have reliable versions but unreliable names
+            name = self._read_name_from_cache(cache_path)
+            version = package.version
+        elif isinstance(locator, (PortalLocator, LinkLocator)):
+            parent_locator = locator.locator
+            packjson = self._project_subpath(
+                parent_locator.relpath, locator.relpath, "package.json"
+            )
+            if isinstance(locator, LinkLocator) and not packjson.path.exists():
+                # if a link dependency doesn't have a package.json, we have to rely on the locator
+                name = self._scoped_name(locator)
+                version = None
+            else:
+                # otherwise, take both the name and the version from package.json
+                # (name is unreliable, version is '0.0.0-use.local')
+                log_for_locator(
+                    "reading package name and version from %s", packjson.subpath_from_root
+                )
+                name, version = self._read_name_version_from_packjson(packjson)
+        else:
+            # This line can never be reached assuming type-checker checks are passing
+            # https://typing.readthedocs.io/en/latest/source/unreachable.html#assert-never-and-exhaustiveness-checking
+            assert_never(locator)
+
+        return _ResolvedPackage(locator, name, version, package.raw_locator)
+
+    def _read_name_from_cache(self, cache_path: RootedPath) -> str:
+        with zipfile.ZipFile(cache_path) as zf:
+            packjson_paths = (
+                filename
+                for filename in zf.namelist()
+                # node_modules/@scope/name/package.json
+                # node_modules/name/package.json
+                if (path := Path(filename)).parts[0] == "node_modules"
+                and len(path.parts) in (3, 4)
+                and path.parts[-1] == "package.json"
+            )
+            packjson_path = next(packjson_paths, None)
+            if packjson_path is None:
+                raise _CouldNotResolve(f"{cache_path.subpath_from_root}: no package.json")
+
+            packjson_content = zf.read(packjson_path)
+
+        try:
+            packjson = json.loads(packjson_content)
+        except json.JSONDecodeError as e:
+            raise _CouldNotResolve(
+                f"{cache_path.subpath_from_root}::{packjson_path}: invalid JSON ({e})"
+            ) from e
+
+        if not (name := packjson.get("name")):
+            raise _CouldNotResolve(
+                f"{cache_path.subpath_from_root}::{packjson_path}: no 'name' attribute"
+            )
+
+        return name
+
+    def _scoped_name(self, locator: Union[NpmLocator, WorkspaceLocator, LinkLocator]) -> str:
+        if locator.scope:
+            return f"@{locator.scope}/{locator.name}"
+        return locator.name
+
+    def _read_name_version_from_packjson(
+        self, packjson_path: RootedPath
+    ) -> tuple[str, Optional[str]]:
+        try:
+            packjson = json.loads(packjson_path.path.read_text())
+        except FileNotFoundError as e:
+            raise _CouldNotResolve(f"missing {packjson_path.subpath_from_root}") from e
+        except json.JSONDecodeError as e:
+            raise _CouldNotResolve(f"{packjson_path.subpath_from_root}: invalid JSON ({e})") from e
+
+        if not (name := packjson.get("name")):
+            raise _CouldNotResolve(f"{packjson_path.subpath_from_root}: no 'name' attribute")
+
+        return name, packjson.get("version")
+
+    def _project_subpath(self, *parts: Union[str, Path]) -> RootedPath:
+        return self._project.source_dir.join_within_root(*parts)
+
+    def _cache_path_as_rooted(self, cache_path: str) -> RootedPath:
+        if Path(cache_path).is_relative_to(self._project.source_dir):
+            return self._project_subpath(cache_path)
+        else:
+            return self._output_dir.join_within_root(cache_path)
 
 
 def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> str:
--- a/cachi2/core/package_managers/yarn/resolver.py
+++ b/cachi2/core/package_managers/yarn/resolver.py
@@ -124,16 +124,18 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
     return packages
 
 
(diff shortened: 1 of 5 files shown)


cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -4,113 +4,353 @@
...
+            else:
+                raise _CouldNotResolve(
+                    "the 'yarn info' output does not include either an existing zip archive "
+                    "or the original unpatched package",
+                )
+            version = package.version
+        else:
+            assert_never(locator)
+
+        return _ResolvedPackage(locator, name, version, package.raw_locator)
```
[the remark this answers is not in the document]
chmeliik: CodeQL pls. Do you even `[assert_never](https://typing.readthedocs.io/en/latest/source/unreachable.html#assert-never-and-exhaustiveness-checking)`

chmeliik:
This is the first step towards SBOM generation, PTAL. Only the first 4 commits make sense at the moment, unless you want to see the results of the passing integration tests.

Does the approach make sense?

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -124,44 +141,217 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
...
+                log_for_locator("resolving the name of the original package")
+                name = self._resolve_package(orig_package).name
+            else:
+                raise _CouldNotResolve(
+                    "the 'yarn info' output does not include either an existing zip archive "
+                    "or the original unpatched package",
+                )
+            version = package.version
+        else:
+            assert_never(locator)
```
ben-alkov, line 276: My concern with asserts in non-test code is *always* "The current code generator emits no code for an assert statement when optimization is requested at compile time" (e.g. command line option [-O]) - asserts are simply remove.
*We* probably wouldn't run it that way, but someone else might, and years from now, when none of us are actively working on cachi2 any more, someone inside RH mght even think it's a good idea to run with `-O`...

chmeliik: It's not really an assert, it's a type-checker trick to check that this branch can never be reached at runtime: https://typing.readthedocs.io/en/latest/source/unreachable.html#assert-never-and-exhaustiveness-checking

If the `mypy` check is passing, then this code is unreachable. If cachi2 somehow does reach this branch, it will fail with a `NameError: name 'assert_never' is not defined`, because we only import it `if TYPE_CHECKING`

In python 3.11 and later, you can import `assert_never` directly from `typing` and don't have to import it conditionally from `typing_extensions`. Then it would fail with an `AssertionError: Expected code to be unreachable, but got: ...` regardless of any optimizations (the function doesn't use `assert`, it raises the error directly)

ben-alkov: 😵‍💫

ben-alkov: Weird. Using a *type* checker for that feels like [crossing the streams](https://quotegeek.com/quotes-from-movies/ghostbusters/206/). OTOH, I wouldn't know how else to automagically check something like that...

tl;dr: TIL :grin:

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -124,44 +141,217 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
...
+            packjson = json.loads(packjson_path.path.read_text())
+        except FileNotFoundError as e:
+            raise _CouldNotResolve(f"missing {packjson_path.subpath_from_root}") from e
+        except json.JSONDecodeError as e:
+            raise _CouldNotResolve(f"{packjson_path.subpath_from_root}: invalid JSON ({e})") from e
+
+        if not (name := packjson.get("name")):
+            raise _CouldNotResolve(f"{packjson_path.subpath_from_root}: no 'name' attribute")
+
+        return name, packjson.get("version")
```
brunoapimentel, line 324: So we're assuming that, if version does not exist, we'll just leave it empty? This can potentially be the case of the main package, since it is treated as a workspace.

I'm just wondering if we want to use some approach involving the commit hash for such cases (similar to gomod). But pseudoversions are not a thing for npm/yarn, right?

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -124,44 +141,217 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
...
 
-    Look at the package.json name for every non-registry dependency.
-    """
-    return "placeholder"
+
+class _CouldNotResolve(ValueError):
+    """_ComponentResolver failed to resolve the name or version of a package."""
+
+
+class _ComponentResolver:
```
brunoapimentel, line 176: This is a nitpick, but probably worth mentioning it in the early review: I don't quite like the idea of having a class to perform some immediate operation. I think that having attributes gives the idea that we're keeping state for a longer time, but in this case here, it is simply to fulfill all the operations to `get_component`, and then the object can be discarded.

I do like the nesting that the class brings, though, since it creates a clear boundary to show that all methods inside are related to this specific feature. I tried to work around this dilemma [here](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/gomod.py#L244-L289) by using a closure and nested functions, so I wouldn't need to instantiate an object.

Again, I feel it's more of a nitpick, so just sharing my thoughts.

brunoapimentel:
Somewhat unrelated, but I noticed that when we ban a dependency based on the protocol, we don't specify where exactly it is being imported from:
```
2023-10-17 19:24:55,527 ERROR Cachi2 does not support Git or Exec dependencies for Yarn Berry: is-positive@https://github.com/kevva/is-positive.git#commit=f0c336c9587d43d97932395b1d2798c298dad40b
```
In this case, this dependency is in a [workspace](https://github.com/cachito-testing/cachi2-yarn-berry/blob/zero-installs/packages/the-answer/package.json#L7).

brunoapimentel (commented):
Overall approach LGTM. I did some testing to validate the approach we use for resolving name and version for each type of protocol, and it seems correct to me.

cachi2/core/package_managers/yarn/resolver.py
chmeliik: > So we're assuming that, if version does not exist, we'll just leave it empty? This can potentially be the case of the main package, since it is treated as a workspace.

Yeah. It can be the case for the main package and pretty much any dependency other than npm dependencies.

> I'm just wondering if we want to use some approach involving the commit hash for such cases (similar to gomod). But pseudoversions are not a thing for npm/yarn, right?

In the purl yes, in the version no. I don't see a reason to make up versions for packages that don't have one

chmeliik: The name and version should be those that make sense within the package manager's ecosystem. If a CVE is found in `some-package@1.2.3`, it should be possible to identify that lib in our SBOMs regardless of where the package came from (false positives possible of course).

The purl should then make it possible to tell where the package came from. E.g. if it's the "main package" of the repository you're building, it would be `pkg:npm/some-package@1.2.3?vcs_url=https://github.com/some-org/some-package@abcdef1234...`

If `some-package` doesn't have a version, using `abcdef1234...` as the version won't really achieve anything. We would have to know what `<insert popular CVE databases>` do when packages don't have versions

cachi2/core/package_managers/yarn/resolver.py
chmeliik: If you inline the function, it becomes:

```python
if isinstance(locator, ...):
   ...
else:
    locator: Never = locator
    raise AssertionError(f"Expected code to be unreachable, but got: {locator!r}")
```

Remove the line with Never and you're not using the type checker, you're raising an exception as usual. Keep the line and you're doing the same thing + you can statically check that you covered all possible cases

chmeliik: It gets a bit weirder if you remember that we're importing assert_never conditionally. Maybe it would help to inline the check instead of using assert_never? Or re-implement assert_never as a util and import it from our own code?

chmeliik:
Equivalent rewrite to try and make CodeQL happy. It doesn't report the issue when I [run it locally](https://docs.github.com/en/code-security/codeql-cli/getting-started-with-the-codeql-cli), IDK (but then again, I probably did it wrong)

cachi2/core/package_managers/yarn/resolver.py
chmeliik: > I tried to work around this dilemma [here](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/gomod.py#L244-L289) by using a closure and nested functions, so I wouldn't need to instantiate an object.

Probably not a good idea in this case https://github.com/containerbuildsystem/cachi2/pull/352/commits/d18e686931cc14376d76890cb72b0478ed23eb2f

chmeliik: I could un-nest them and pass around the `package_mapping`, `project` and `output_dir` as params where needed if that's better

chmeliik:
> Somewhat unrelated, but I noticed that when we ban a dependency based on the protocol, we don't specify where exactly it is being imported from:
> 
> ```
> 2023-10-17 19:24:55,527 ERROR Cachi2 does not support Git or Exec dependencies for Yarn Berry: is-positive@https://github.com/kevva/is-positive.git#commit=f0c336c9587d43d97932395b1d2798c298dad40b
> ```
> 
> In this case, this dependency is in a [workspace](https://github.com/cachito-testing/cachi2-yarn-berry/blob/zero-installs/packages/the-answer/package.json#L7).

You could do a graph search based on the `children.Dependencies[].locator` and print the whole import path of each unsupported dependency, if you want to file a story for that

chmeliik:
Undid https://github.com/containerbuildsystem/cachi2/commit/d18e686931cc14376d76890cb72b0478ed23eb2f and split the commits, overall no change (except the undoing of  that commit)

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -124,44 +141,211 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
     return packages
 
 
-def create_component_from_package(package: Package, project: Project) -> Component:
```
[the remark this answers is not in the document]
chmeliik: Yeah that's fair. I split it a bit differently, but it should be better than before

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -120,7 +120,7 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
     # --all: report dependencies of all workspaces, not just the active workspace
```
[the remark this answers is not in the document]
ben-alkov: My sense is that one should avoid `os.*` in modern Python if there's an alternative built-in available (e.g. `shutil` here)

brunoapimentel:
> You could do a graph search based on the `children.Dependencies[].locator` and print the shortest import path of each unsupported dependency, if you want to file a story for that

Nevermind, I think at this point it'll just stay there until it's forgotten.

cachi2/core/package_managers/yarn/resolver.py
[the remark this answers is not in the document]
brunoapimentel: Just close this as false positive and let's move on with it :smile:

cachi2/core/package_managers/yarn/resolver.py
brunoapimentel: Great point. I forgot that we'd have the `vcs_url` part in the purl, that will make very clear what the actual package is, so this solves my initial concern. As for the CVE part, I guess you're right, it'll be dependend on packages that need to have versions.

cachi2/core/package_managers/yarn/resolver.py
brunoapimentel: I thought [it](https://github.com/containerbuildsystem/cachi2/commit/d18e686931cc14376d76890cb72b0478ed23eb2f) looked good at first sight, but looking at it again, I think it's easy to forget what `project` or `output_dir` are, since it's a lengthy implementation. That's the only downside I see to it, really.

All in all, I think we're still better off with the class :grin:

brunoapimentel: And since we're nitpicking this piece, why do we have [_scoped_name](https://github.com/containerbuildsystem/cachi2/commit/d18e686931cc14376d76890cb72b0478ed23eb2f) as a standalone function? Do we want to allow uses of it ouside of the _ComponentResolver?

cachi2/core/package_managers/yarn/resolver.py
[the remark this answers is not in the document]
eskultety: @ben-alkov sure, but here it wasn't about direct os.* usage, it's what `subprocess.run->subprocess.Popen` will ultimately resolve to and in this case we had to dig in the docs all the way on the POSIX syscall level to figure out what's going on.

cachi2/core/package_managers/yarn/resolver.py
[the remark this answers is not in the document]
eskultety: Sure, you're more familiar with the code that I am, so I completely trust your judgement on the split :).

chmeliik:
Added an explanation for assert_never, still have to fix the subprocess bits

cachi2/core/package_managers/yarn/resolver.py
chmeliik: Added

cachi2/core/package_managers/yarn/resolver.py
chmeliik: Probably not, moved it into the class

cachi2/core/package_managers/yarn/utils.py
```diff
@@ -1,19 +1,26 @@
...
     :param source_dir: the directory in the repository containing the yarn source files.
-    :param env: environment variables to be set during the command's execution.
-
+    :param env: environment variables to be set during the command's execution
+        by default, the subprocess is spawned with a clean environment
+        (other than PATH, which the subprocess inherits from the current process)
     :raises YarnCommandError: if the command fails.
     """
+    env = env or {}
+    env = env | {"PATH": os.environ.get("PATH", "")}
```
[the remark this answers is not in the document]
chmeliik: I've extracted the subprocess fixes into a separate PR: #366 

Addressed there

cachi2/core/package_managers/yarn/resolver.py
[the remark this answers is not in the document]
chmeliik: Addressed in #366

chmeliik:
Dropped the subprocess changes, extracted into #366 so that this PR can focus only on the name+version resolving

Dropped the TEMP changes, finalized unit tests

chmeliik:
Rebased on main, ready for review

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -124,44 +144,212 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
...
+                    "the 'yarn info' output does not include either an existing zip archive "
+                    "or the original unpatched package",
+                )
+            version = package.version
+        else:
+            # This line can never be reached assuming type-checker checks are passing
+            # https://typing.readthedocs.io/en/latest/source/unreachable.html#assert-never-and-exhaustiveness-checking
+            assert_never(locator)
+
+        return _ResolvedPackage(locator, name, version, package.raw_locator)
```
ben-alkov, line 277: Seeing this changed to a single `return` makes the Functional Programmer in me very happy :D

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -1,15 +1,28 @@
...
+from cachi2.core.errors import PackageRejected, UnsupportedFeature
+from cachi2.core.models.sbom import Component
 from cachi2.core.package_managers.yarn.locators import parse_locator
-from cachi2.core.package_managers.yarn.resolver import Package, resolve_packages
+from cachi2.core.package_managers.yarn.project import PackageJson, Project, YarnRc
+from cachi2.core.package_managers.yarn.resolver import Package, create_components, resolve_packages
 from cachi2.core.rooted_path import RootedPath
 
 
+def setup_module() -> None:
```
brunoapimentel, line 18: Just for my education: what test is actually disabling the logs?

brunoapimentel (commented):
Just a very minor nitpick about the commit messages: I'd make all of them just `Yarn: `, since that's the slightly more consistent name we've been calling it. There's quite a few `Yarnberry` commits at this point, so this ship sailed long ago in any case :man_shrugging:

brunoapimentel (approved):
LGTM :rocket:

tests/unit/package_managers/yarn/test_resolver.py
chmeliik: By "bisecting" the test file, I found out that it's something in test_cli.py. But I don't know what specifically and why :disappointed:

cachi2/core/package_managers/yarn/resolver.py
ben-alkov: I just realized that CodeQL's earlier problem with that local variable`version` being "used before initialize" is because that "single exit" pattern *requires* explicitly declaring (e.g. `version = None` or maybe the correct type, probably around [here](https://github.com/chmeliik/cachi2/blob/bd4db5f7c5942f23fcb9610dd54e62cf5e9697ea/cachi2/core/package_managers/yarn/resolver.py#L208)) anything you return which *isn't* already assigned to in the "top level" scope (for whatever "top-level" is in context). I hope that made sense...

ben-alkov (approved):

PR #355, opened 2023-10-13
[Do not merge] Test RHTAP pipeline copy

lkolacek:


.tekton/test-rhtap.yaml
```diff
@@ -0,0 +1,68 @@
+---
```
eskultety, line 1: **Commenting on the commit message:**

Can you be more descriptive in the commit message please? It might be obvious to all current project members at the moment what exactly this Tekton template is for or at least what's the reason behind it being added, but for new members or just irregular contributors this change is completely opaque without any context. Also rember that this commit message goes to public git history, so if you don't explain anything in the commit message, in future there's going to be literally 0 trace on why some things were done in a specific way in the template, like e.g. the 'script' YAML section.

I would also like to again suggest adopting some kind of structure in the commit subjects: e.g. `.tekton: <file>: <what>` or similar, because that immediately will help potential reviewers identify whether that's an area they're expert on/interested in and what the gist of suggested changes is without having to go through every commit in each (or most) PR, dive into the code to understand the changes and then figure out whether they feel qualified enough to review the changes. Please take ^this just as a mere suggestion from an observer :).

.tekton/test-rhtap.yaml
```diff
@@ -0,0 +1,70 @@
...
+            value: "$(params.revision)"
+        timeout: "15m"
+        taskSpec:
+          params:
+            - name: revision
+          steps:
+            - name: tag-with-latest
+              image: registry.access.redhat.com/ubi9/skopeo:latest@sha256:23c9ed4af1f42614bdc56309452568b2110a67f23102f42f6a6582a63b63dcdb
+              script: |
+                #!/usr/bin/env bash
```
eskultety, line 34: Can you give a bit of trivia for this script? IOW what is it trying to test?

.tekton/test-rhtap.yaml
lkolacek: OK I apologize I could have added some description to commit message. This PR is not meant to be merged and it's purpose is only provide some test info. Review is not needed here and it was not requested.

PR #365, opened 2023-10-18
feat(cli): add 'dev-mode' flag

ben-alkov:

- [ ] ~Docs updated (if applicable)~ (will be done in a follow up PR)

Add a flag which allows for creating requests using a package manager
which is still under development

Any such package manager(s) will be "hidden" behind the new flag, which
should prevent their use by end-users (at least those using Cachi2 in
pipelines)

This story is part of the "Yarn for Node.js" epic

- STONEBLD-1750

Signed-off-by: Ben Alkov <ben.alkov@redhat.com>

### DIFF
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -99,7 +99,7 @@ class PipPackageInput(_PackageInputBase):
 
 
 class YarnPackageInput(_PackageInputBase):
-    """Accepted input for a npm package."""
+    """Accepted input for a yarn package."""
 
     type: Literal["yarn"]
 
@@ -168,7 +168,7 @@ class Request(pydantic.BaseModel):
 
     @property
     def yarn_packages(self) -> list[YarnPackageInput]:
-        """Get the gomod packages specified for this request."""
+        """Get the yarn packages specified for this request."""
         return self._packages_by_type(YarnPackageInput)
 
     def _packages_by_type(self, pkgtype: type[T]) -> list[T]:
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -49,7 +49,9 @@ def _present_user_input_error(validation_error: pydantic.ValidationError) -> str
 # Supported package managers
 PackageManagerType = Literal["gomod", "npm", "pip", "yarn"]
 
-Flag = Literal["cgo-disable", "force-gomod-tidy", "gomod-vendor", "gomod-vendor-check"]
+Flag = Literal[
+    "cgo-disable", "dev-package-managers", "force-gomod-tidy", "gomod-vendor", "gomod-vendor-check"
+]
 
 
 class _PackageInputBase(pydantic.BaseModel, extra="forbid"):
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -4,7 +4,7 @@ from typing import Callable
 from cachi2.core.errors import UnsupportedFeature
 from cachi2.core.models.input import PackageManagerType, Request
 from cachi2.core.models.output import RequestOutput
-from cachi2.core.package_managers import gomod, npm, pip
+from cachi2.core.package_managers import gomod, npm, pip, yarn
 
 Handler = Callable[[Request], RequestOutput]
 
@@ -15,7 +15,7 @@ _package_managers: dict[PackageManagerType, Handler] = {
 }
 
 _dev_package_managers: dict[PackageManagerType, Handler] = {
-
+    "yarn": yarn.fetch_yarn_source,
 }
 
 # This is *only* used to provide a list for `cachi2 --version`
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -14,21 +14,28 @@ _package_managers: dict[PackageManagerType, Handler] = {
     "pip": pip.fetch_pip_source,
 }
 
+_dev_package_managers: dict[PackageManagerType, Handler] = {
 
+}
+
+# This is *only* used to provide a list for `cachi2 --version`
 supported_package_managers = list(_package_managers)
 
 
 def resolve_packages(request: Request) -> RequestOutput:
     """Run all requested package managers, return their combined output."""
+    _supported_package_managers = _package_managers
     requested_types = set(pkg.type for pkg in request.packages)
-    unsupported_types = requested_types - _package_managers.keys()
+    if "dev-package-managers" in request.flags:
+        _supported_package_managers = _package_managers | _dev_package_managers
+    unsupported_types = requested_types - _supported_package_managers.keys()
     if unsupported_types:
         raise UnsupportedFeature(
             f"Package manager(s) not yet supported: {', '.join(sorted(unsupported_types))}",
             # unknown package managers shouldn't get past input validation
             solution="But the good news is that we're already working on it!",
         )
-    pkg_managers = [_package_managers[type_] for type_ in sorted(requested_types)]
+    pkg_managers = [_supported_package_managers[type_] for type_ in sorted(requested_types)]
     return _merge_outputs(pkg_manager(request) for pkg_manager in pkg_managers)
 
 
--- a/tests/unit/models/test_input.py
+++ b/tests/unit/models/test_input.py
@@ -211,7 +211,7 @@ class TestRequest:
             )
 
     def test_invalid_flags(self) -> None:
-        expect_error = r"Input should be 'cgo-disable', 'force-gomod-tidy', 'gomod-vendor' or 'gomod-vendor-check'"
+        expect_error = r"Input should be 'cgo-disable', 'dev-package-managers', 'force-gomod-tidy', 'gomod-vendor' or 'gomod-vendor-check'"
         with pytest.raises(pydantic.ValidationError, match=expect_error):
             Request(
                 source_dir="/source",
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -532,6 +532,13 @@ class TestFetchDeps:
                     "force-gomod-tidy",
                 },
             ),
+            (
+                [
+                    '{"packages": [{"type":"gomod"}]}',
+                    "--dev-package-managers",
+                ],
+                {"dev-package-managers"},
+            ),
         ],
     )
     def test_specify_flags(
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -560,7 +560,7 @@ class TestFetchDeps:
             ),
             (
                 ['{"packages": [{"type": "gomod"}], "flags": ["no-such-flag"]}'],
-                "Input should be 'cgo-disable', 'force-gomod-tidy', 'gomod-vendor' or 'gomod-vendor-check'",
+                "Input should be 'cgo-disable', 'dev-package-managers', 'force-gomod-tidy', 'gomod-vendor' or 'gomod-vendor-check'",
             ),
         ],
     )
--- a/tests/unit/test_resolver.py
+++ b/tests/unit/test_resolver.py
@@ -1,10 +1,14 @@
+import re
 from pathlib import Path
 from typing import Callable
 from unittest import mock
 
+import pytest
+
 from cachi2.core import resolver
+from cachi2.core.errors import UnsupportedFeature
 from cachi2.core.models.input import Request
-from cachi2.core.models.output import EnvironmentVariable, ProjectFile, RequestOutput
+from cachi2.core.models.output import BuildConfig, EnvironmentVariable, ProjectFile, RequestOutput
 from cachi2.core.models.sbom import Component
 
 GOMOD_OUTPUT = RequestOutput.from_obj_list(
@@ -93,3 +97,74 @@ def test_resolve_packages(tmp_path: Path) -> None:
         assert resolver.resolve_packages(request) == COMBINED_OUTPUT
 
     assert calls_by_pkgtype == ["gomod", "npm", "pip"]
+
+
+@pytest.mark.parametrize(
+    "flags",
+    [
+        pytest.param(["dev-package-managers"], id="dev-package-managers-true"),
+        pytest.param([], id="dev-package-managers-false"),
+    ],
+)
+def test_dev_mode(flags: list[str]) -> None:
+    mock_resolver = mock.Mock()
+    mock_resolver.return_value = RequestOutput.empty()
+    with (
+        mock.patch.dict(
+            resolver._package_managers,
+            values={"gomod": mock_resolver},
+            clear=True,
+        ),
+        mock.patch.dict(
+            resolver._dev_package_managers,
+            values={"shrubbery": mock_resolver},
+            clear=True,
+        ),
+    ):
+        dev_package_input = mock.Mock()
+        dev_package_input.type = "shrubbery"
+
+        request = mock.Mock()
+        request.flags = flags
+        request.packages = [dev_package_input]
+
+        if flags:
+            assert resolver.resolve_packages(request) == RequestOutput(
+                components=[], build_config=BuildConfig(environment_variables=[], project_files=[])
+            )
+        else:
+            expected_error = re.escape("Package manager(s) not yet supported: shrubbery")
+            with pytest.raises(UnsupportedFeature, match=expected_error):
+                resolver.resolve_packages(request)
+
+
+def test_resolve_with_released_and_dev_package_managers() -> None:
+    mock_resolve_gomod = mock.Mock(return_value=RequestOutput.empty())
+    mock_resolve_pip = mock.Mock(return_value=RequestOutput.empty())
+
+    with (
+        mock.patch.dict(
+            resolver._package_managers,
+            values={"gomod": mock_resolve_gomod},
+            clear=True,
+        ),
+        mock.patch.dict(
+            resolver._dev_package_managers,
+            values={"pip": mock_resolve_pip},
+            clear=True,
+        ),
+    ):
+        dev_package_input = mock.Mock()
+        dev_package_input.type = "pip"
+
+        released_package_input = mock.Mock()
+        released_package_input.type = "gomod"
+
+        request = mock.Mock()
+        request.flags = ["dev-package-managers"]
+        request.packages = [released_package_input, dev_package_input]
+
+        resolver.resolve_packages(request)
+
+        mock_resolve_gomod.assert_has_calls([mock.call(request)])
+        mock_resolve_pip.assert_has_calls([mock.call(request)])
--- a/cachi2/interface/cli.py
+++ b/cachi2/interface/cli.py
@@ -132,6 +132,7 @@ def fetch_deps(
         resolve_path=True,
         help="Write output files to this directory.",
     ),
+    dev_package_managers: bool = typer.Option(False, "--dev-package-managers", hidden=True),
     cgo_disable: bool = typer.Option(
         False, "--cgo-disable", help="Set CGO_ENABLED=0 while processing gomod packages."
     ),
@@ -216,8 +217,20 @@ def fetch_deps(
             return {"packages": [{"type": raw_input}]}
 
     def combine_option_and_json_flags(json_flags: list[Flag]) -> list[str]:
-        flag_names = ["cgo-disable", "force-gomod-tidy", "gomod-vendor", "gomod-vendor-check"]
-        flag_values = [cgo_disable, force_gomod_tidy, gomod_vendor, gomod_vendor_check]
+        flag_names = [
+            "cgo-disable",
+            "dev-package-managers",
+            "force-gomod-tidy",
+            "gomod-vendor",
+            "gomod-vendor-check",
+        ]
+        flag_values = [
+            cgo_disable,
+            dev_package_managers,
+            force_gomod_tidy,
+            gomod_vendor,
+            gomod_vendor_check,
+        ]
         flags = [name for name, value in zip(flag_names, flag_values) if value]
 
         if json_flags:


cachi2/core/resolver.py
```diff
@@ -14,13 +14,19 @@
...
+_dev_package_managers: dict[PackageManagerType, Handler] = {
+    "yarn": yarn.fetch_yarn_source,
+}
 
 supported_package_managers = list(_package_managers)
 
 
 def resolve_packages(request: Request) -> RequestOutput:
     """Run all requested package managers, return their combined output."""
+    global _package_managers
```
taylormadore, line 26: Unless I'm misunderstanding the PEP584 docs, it seemed to me that this _should_ work without declaring _package_managers as a global, since the "|=" operator supposedly modifies the dict in-place, but it doesn't :thinking: 

After I went down a rabbit hole on that, I wonder if we can sidestep this entirely and just use local variables instead?

ben-alkov:
Almost ready to un-draft - I have yet to resolve [the weird mypy issue](https://github.com/containerbuildsystem/cachi2/actions/runs/6618928122/job/17978408124?pr=365)

cachi2/core/resolver.py
ben-alkov: `_package_managers` has to be declared `global` in the local scope so that it can be modified *at all*. Regardless, I'm happier with just using a local variable. `global` makes the code more concise, but it's not used very often at all, which can lead to confusion for people who've never seen it.

tests/unit/test_resolver.py
```diff
@@ -93,3 +97,71 @@ def fetch(req: Request) -> RequestOutput:
...
+        dev_package_input.type = "pip"
+
+        released_package_input = mock.Mock()
+        released_package_input.type = "gomod"
+
+        request = mock.Mock()
+        request.flags = ["dev-mode"]
+        request.packages = [released_package_input, dev_package_input]
+
+        assert resolver.resolve_packages(request) == RequestOutput(
```
lkolacek, line 165: I believe we should add more specific test for resolved packages to see that we actually resolved both released and dev packages. In case that I would remove `released_package_input` from packages now I believe test would still pass.

brunoapimentel: We could probably mock the resolver functions and then assert they were called 2x:
```
mock_resolver = mock.Mock()
mock_resolver.return_value = RequestOutput.empty()

    with (
        mock.patch.dict(
            resolver._package_managers,
            values={"gomod":mock_resolver},
            clear=True,
        ),
        mock.patch.dict(
            resolver._dev_package_managers,
            values={"pip": mock_resolver},
            clear=True,
        ),
    ):
    # ...
    assert mock_resolver.call_count == 2
```

tests/unit/test_resolver.py
```diff
@@ -93,3 +97,71 @@ def fetch(req: Request) -> RequestOutput:
...
+            resolver._dev_package_managers,
+            values={"pip": lambda _: RequestOutput.empty()},
+            clear=True,
+        ),
+    ):
+
+        class foo:
+            type: str
+
+        dev_package_input = foo()
```
brunoapimentel, line 155: We could make this a `mock.Mock()` and drop the foo class, and the code would still work as expected.

tests/unit/test_resolver.py
```diff
@@ -93,3 +97,71 @@ def fetch(req: Request) -> RequestOutput:
...
+        if flags:
+            assert resolver.resolve_packages(request) == RequestOutput(
+                components=[], build_config=BuildConfig(environment_variables=[], project_files=[])
+            )
+        else:
+            expected_error = re.escape("Package manager(s) not yet supported: shrubbery")
+            try:
+                resolver.resolve_packages(request)
+            except UnsupportedFeature as exc:
+                assert re.match(expected_error, str(exc))
```
brunoapimentel, line 135: Can we rewrite this piece using `pytest.raises`? I think it makes for a more readable syntax.

ben-alkov: It fails for me here. Pytest says the exception isn't being raised.

tests/unit/test_resolver.py
ben-alkov: Must've been an in-progress change which cause the `mock.Mock()` version to fail (something about `Mock` not being iterable). Works now.

tests/unit/test_resolver.py
ben-alkov: `FAILED tests/unit/test_resolver.py::test_dev_mode[dev-mode-false] - Failed: DID NOT RAISE <class 'cachi2.core.errors.UnsupportedFeature'>`

ben-alkov: And that's with a completely clean repo (except for my code changes), with make and tox generating everything from scratch :shrug:

ejegrova (approved):

cachi2/core/resolver.py
```diff
@@ -14,21 +14,27 @@
...
 supported_package_managers = list(_package_managers)
 
 
 def resolve_packages(request: Request) -> RequestOutput:
     """Run all requested package managers, return their combined output."""
+    local_pkg_managers = _package_managers
     requested_types = set(pkg.type for pkg in request.packages)
-    unsupported_types = requested_types - _package_managers.keys()
+    if "dev-mode" in request.flags:
+        local_pkg_managers |= _dev_package_managers
```
brunoapimentel, line 29: I still think we shouldn't update the global `_package_managers` variable, and use a local copy. Although Cachi2 is very likely to be used as a "single run" process, I'd avoid this as it could become a pitfall in some alternate scenarios.

tests/unit/test_resolver.py
```diff
@@ -93,3 +97,72 @@ def fetch(req: Request) -> RequestOutput:
...
+            assert resolver.resolve_packages(request) == RequestOutput(
+                components=[], build_config=BuildConfig(environment_variables=[], project_files=[])
+            )
+        else:
+            expected_error = re.escape("Package manager(s) not yet supported: shrubbery")
+            with pytest.raises(UnsupportedFeature, match=expected_error):
+                resolver.resolve_packages(request)
+
+
+def test_dev_mode_with_released() -> None:
```
brunoapimentel, line 141: Here's a slight improvement suggestion to this test here:
```
 def test_dev_mode_with_released() -> None:
-    mock_resolver = mock.Mock()
-    mock_resolver.return_value = RequestOutput.empty()
+    mock_resolve_gomod = mock.Mock(return_value=RequestOutput.empty())
+    mock_resolve_pip = mock.Mock(return_value=RequestOutput.empty())
+
     with (
         mock.patch.dict(
             resolver._package_managers,
-            values={"gomod": mock_resolver},
+            values={"gomod": mock_resolve_gomod},
             clear=True,
         ),
         mock.patch.dict(
             resolver._dev_package_managers,
-            values={"pip": mock_resolver},
+            values={"pip": mock_resolve_pip},
             clear=True,
         ),
     ):
@@ -165,4 +166,5 @@ def test_dev_mode_with_released() -> None:
 
         resolver.resolve_packages(request)
 
-        assert mock_resolver.call_count == 2
+        mock_resolve_gomod.assert_has_calls([mock.call(request)])
+        mock_resolve_pip.assert_has_calls([mock.call(request)])
```

cachi2/core/resolver.py
ben-alkov: `local_pkg_managers` **should** be a (true) _copy_ of `_package_managers`... If it weren't, that would defeat the entire purpose of the `global` keyword.
I'll run a debug and check object addresses to be certain.

brunoapimentel: If we only assign a new variable like `local_pkg_managers = _package_managers`, we're actually just passing the reference to the same dictionary to another variable, so no actual copy is made.

tests/unit/test_resolver.py
brunoapimentel, line 141: Nitpick: Rename this test to `test_resolve_with_released_and_dev_package_managers`. More verbose, but I'll probably forget what `released` is without more context :laughing: 

(or maybe just merge this scenario with the previous test... not sure if it'll get too complex)

cachi2/core/resolver.py
ben-alkov: Correct, but in that scope, `_package_managers` does *not* point to the module-level "global" `_package_managers` - it's a copy

brunoapimentel: As weird as it seems, it does point to the module-level variable :)

You can see it if you print the module-level variable after the `|=` operation, and make a request with `--dev-mode`.
```
diff --git a/cachi2/core/resolver.py b/cachi2/core/resolver.py
index 664332b..5bc61d3 100644
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -34,6 +34,7 @@ def resolve_packages(request: Request) -> RequestOutput:
             # unknown package managers shouldn't get past input validation
             solution="But the good news is that we're already working on it!",
         )
+    print(_package_managers)
     pkg_managers = [local_pkg_managers[type_] for type_ in sorted(requested_types)]
     return _merge_outputs(pkg_manager(request) for pkg_manager in pkg_managers)
```

tests/unit/test_resolver.py
ben-alkov: > I'll probably forget what released is without more context

Thoroughly valid :+1:

cachi2/core/resolver.py
brunoapimentel: I know, right?? I usually do a lot of small tests locally to remember how these details work.. it never sticks to my head :shrug:

ben-alkov: Ah. Apparently it's in the fine print: "[free variables may refer to globals without being declared global](https://docs.python.org/3.9/reference/simple_stmts.html#the-global-statement)", although I'm sure that I don't understand all of the implications of that statement.

lkolacek:
Do we want to mention a new flag for testing package managers in development in our documentation?

ben-alkov:
> Do we want to mention a new flag for testing package managers in development in our documentation?

If there's a "developer" section in the docs, yes. I wouldn't want to put it in end-user docs - it's *specifically* not for end-users.

lkolacek:
There's a development section in the README, I believe a short section might be useful there for newcomers. https://github.com/containerbuildsystem/cachi2#development

tests/unit/models/test_input.py
```diff
@@ -211,7 +211,7 @@ def test_invalid_package_paths(self, path: str, expect_error: str, tmp_path: Pat
             )
```
eskultety, line 211: **Commit message:**

Hold on, isn't the URL actually pointing to a private resource? IOW Isn't the board private to members only?

tests/unit/test_cli.py
```diff
@@ -532,6 +532,13 @@ def test_invalid_packages(
                     "force-gomod-tidy",
```
eskultety, line 532: Same comment about the URL as in previous patch.

cachi2/core/models/input.py
```diff
@@ -99,7 +99,7 @@ def _requirements_file_path_is_relative(cls, paths: list[Path]) -> list[Path]:
```
eskultety, line 100: Feels like this should be the first commit - before doing any functional changes to the code base start by trivial fixes, but that's just my terrible bias, feel free to happily ignore this comment :) .

eskultety:
> Do we want to mention a new flag for testing package managers in development in our documentation?

Coming late to the party, sorry I was working on another PR. So, is 'dev-mode' actually something that we want to deliver as a consumable to cachi2 users or just make it more convenient for developers? The reason why I'm asking is that we'd be constantly shipping a bleeding edge beta feature (changing day-by-day) which I'm not sure we actually want. When I look at cachi2 from a user perspective, it is a tool to somehow take care of my project and give me some (blackbox) output. If I as a user use this feature, I'd be constantly (or at least very regularly) getting different experience (from the daily built container image) using that feature, so what would the inconsistent development state tell me as a user? Probably nothing, just that I'm likely not supposed to use it on my project. So, that tells me that this feature is mostly targeted at developers. From that perspective I completely agree that we need something easy for all devs to be able to make use of in-development package managers rather than enabling it individually in their local environments. However, is a publicly consumable CLI flag actually the way we want to do this? How about having a development branch named `yarn` in the repo and use that one along with `main` as the two currently active branches where all development happens, so any developer who wishes to give an in-development package manager a try clones the branch where the package manager would already be fully enabled on CLI without any additional flag? At the end of development we'd just fast forward the branch on top of `main` so that we go without a merge commit. In large projects fast forwarding the way I propose doesn't work, but this is a small project with the focus being a single package manager most of the time. What are your thoughts on ^this?

tests/unit/models/test_input.py
ben-alkov: You mean "issues.redhat.com"? If yes, why?

ben-alkov:
> What are your thoughts on ^this?

It's a discussion for the team, given that it would be a pretty major shift in the way we work.

My *personal* opinion is that the whole point of the flag was to *not* make it super-obvious to end-users (hence the discussion about dev docs vs. user docs). Also, IIUC, it won't be possible for current end-users (i.e. RHTAP users) to enable this flag, even if they want to.

brunoapimentel:
> What are your thoughts on ^this?

I think it's something we can discuss as part of the "upstream Cachi2" meeting. I'd still keep this PR as it is, though.

Even if we decide to go with the branch approach, I do think we can benefit from a slightly different "--beta" flag, which could basically be this same flag repurposed.

tests/unit/models/test_input.py
brunoapimentel: Well, I normally just add the Jira ticket, not the full URL. But I don't think we have a strict rule about this.

brunoapimentel (approved):

brunoapimentel:
> There's a development section in the README, I believe a short section might be useful there for newcomers. https://github.com/containerbuildsystem/cachi2#development

I approved the PR as it is, but I do think adding to the docs is useful. I'm fine if it is done in a follow up PR, though.

tests/unit/models/test_input.py
eskultety: > You mean "issues.redhat.com"? If yes, why?

Because unlike 'STONEBLD-1750' which is just an opaque string literal telling absolutely nothing to an outsider contributor (which is also a good reason not to have it in a public commit message at all, minus synchrozation back to the Jira instance which is the only reason we stick to using it), a URL to a private resource in a public commit message is IMO an antipattern because everyone can (and they will) click on the URL hoping they could get more information about a given commit (if the message itself isn't descriptive enough). Why do I even mention this as a problem? Let's mention bugzilla as an example. Before Red Hat moved to Jira for tracking (although bugs are to be tracked publicly) people used to put bugzilla links into commit messages as an additional source of information because bugzilla clearly stated the overall problem a given patch is trying to address along with information about reproducers, which products were impacted in what versions, etc. but ONLY when the link pointed to a public bugzilla, because links to private bugzillas would be pointless to outside contributors. So non-RH contributors are actually very much used to put and find this kind of information in commit messages and if we continue putting private URLs in commit messages, that is what creates the anti-pattern.

eskultety:
> It's a discussion for the team, given that it would be a pretty major shift in the way we work.

Absolutely.

> I approved the PR as it is, but I do think adding to the docs is useful. I'm fine if it is done in a follow up PR, though.

Please hold on with merging this, this isn't something that is totally urgent to have in, is it?
I know that backwards compatibility isn't something this project strictly follows, but we need to keep in mind that it's also not okay to keep adding CLI options to later rename them or drop completely. Instead, let's open a discussion on what other options we have and if nothing else is plausible we can get back to having a CLI option.

eskultety:
> My _personal_ opinion is that the whole point of the flag was to _not_ make it super-obvious to end-users (hence the discussion about dev docs vs. user docs). Also, IIUC, it won't be possible for current end-users (i.e. RHTAP users) to enable this flag, even if they want to.

The problem is that RHTAP might not to be the only user, since cachi2 is a standalone tool, so the fact that this flag can't be used in one environment doesn't mean it can't be used in another. To me it feels quite unusual to ship a public CLI option that is only meant for developers to test along the way. I think achieving that desired outcome should be possible by other means, but to be fair to the proposal I would be okay with a **hidden** CLI option i.e. one that is not shown via help, still questionable, but I think that kind of approach is fairly common in the programming field and unarguably having a CLI option is indeed very convenient for anyone.
I think I mentioned it earlier but the main problem I have is that users don't actually know what to do with such a flag as "development mode" can mean anything, e.g. if this were solely about package managers in development, well, which package managers are in development? Do the users know? There is no way to do something like `--dev-mode=?` which would actually list all the package managers currently enabled only as experimental. Even if the CLI option is hidden it would still be great to have a way of telling what is currently allowed as experimental (maybe be more specific with the option name like `--experimental-pkg-managers` rather than have a catch-all one?)

> Even if we decide to go with the branch approach, I do think we can benefit from a slightly different "--beta" flag, which could basically be this same flag repurposed.

So with `--beta` things are IMO a bit tricky, because if you constantly ship beta functionality, then it must be completely transparent to the user what that functionality actually is, IOW you don't expect any input from anyone, you just enable whatever it is that's currently in development and THAT certainly doesn't conform to the proposed use case of the flag, because it requires knowledge of what could/should be used as input by the user, in this case Yarn as package manager.

eskultety:
I think achieving that desired outcome should be possible by other means, but to be fair to the proposal I would be okay with a **hidden** CLI option i.e. one that is not shown via help, still questionable, but I think that kind of approach is fairly common in the programming field and unarguably having a CLI option is indeed very convenient for anyone. 

Looks like Typer indeed wraps Click's `hidden` [flag](https://github.com/tiangolo/typer/blob/92846bb994024b7383169414e0bc250cb5311025/typer/params.py#L42) although the "docs" (if one can call that way) doesn't mention it at all, so hiding this convenience development CLI flag might not be that painful after all.

brunoapimentel (approved):
Re-adding my `LGTM` after the change for a hidden flag :rocket:

cachi2/core/models/input.py
```diff
@@ -47,9 +47,11 @@ def show_error(error: "ErrorDict") -> str:
 
 
 # Supported package managers
-PackageManagerType = Literal["gomod", "npm", "pip", "yarn"]
+PackageManagerType = Literal["gomod", "npm", "pip"]
```
eskultety: We already had "yarn" in the list (prematurely?), but its presence had no effect on unit tests, so there's no reason to drop it from the list now to add it back in last commit, let's just leave the list as is.

tests/unit/models/test_input.py
```diff
@@ -211,7 +211,7 @@ def test_invalid_package_paths(self, path: str, expect_error: str, tmp_path: Pat
...
 
     def test_invalid_flags(self) -> None:
-        expect_error = r"Input should be 'cgo-disable', 'force-gomod-tidy', 'gomod-vendor' or 'gomod-vendor-check'"
+        expect_error = r"Input should be 'dev-package-managers', 'cgo-disable', 'force-gomod-tidy', 'gomod-vendor' or 'gomod-vendor-check'"
```
eskultety, line 214: This commit needs to be squashed into the previous one, otherwise the unit test suite fails with the previous commit.

tests/unit/test_resolver.py
```diff
@@ -93,3 +97,74 @@ def fetch(req: Request) -> RequestOutput:
         assert resolver.resolve_packages(request) == COMBINED_OUTPUT
 
     assert calls_by_pkgtype == ["gomod", "npm", "pip"]
+
```
eskultety, line 100: I'll need a bit of guidance here as still don't have enough insights into the guts of the project - shouldn't we actually test the CLI flag as part of CLI tests and not the resolver? Or maybe at both places?

eskultety (commented):
Overall looking good @ben-alkov, mostly just nitpicks.

EDIT: I wanted to mention one thing, not sure if we want to tweak docs right now, but the existence of hidden dev CLI flags needs to be properly documented for contributors.

cachi2/core/models/input.py
```diff
@@ -47,7 +47,7 @@ def show_error(error: "ErrorDict") -> str:
 
 
 # Supported package managers
-PackageManagerType = Literal["gomod", "npm", "pip"]
+PackageManagerType = Literal["gomod", "npm", "pip", "yarn"]
```
eskultety, line 50: See my earlier comment, this hunk should be dropped, we already had it in the list prior to this PR.

tests/unit/models/test_input.py
eskultety: > This commit needs to be squashed into the previous one, otherwise the unit test suite fails with the previous commit.

[tangentially related]
I just realized that we've already broken bisactability (i.e. making the unit test suite pass after every single commit) on a few occasions. Despite that fact, I still very strictly stick by ^my comment on making the test suite pass after every commit. I wonder whether it were of any use (or simply worth it) to come up with mechanisms to prevent this from happening again going forward now that we know for sure we've broken that "contract" (IOW making sure every commit is properly tested), e.g. declaring cachi2 v1.0.0 stable with more guarantees?

brunoapimentel: I strive to make every commit pass the test suite when I'm writing code, but I don't run the suite for every single one of them, specially when I'm doing code reviews. If there's any automation you know of we could employ, it might be worth it.

tests/unit/test_resolver.py
brunoapimentel: It should be both, IMO. It is already being tested in the [cli](https://github.com/containerbuildsystem/cachi2/blob/76a218fbdc867f27ff7858ff07d3d18662560f6c/tests/unit/test_cli.py#L537-L538). The test is very simple because we're basically validating that the CLI is able to receive the flag. The bulk of the feature is being tested in this resolver test here.

eskultety: > It is already being tested in the [cli](https://github.com/containerbuildsystem/cachi2/blob/76a218fbdc867f27ff7858ff07d3d18662560f6c/tests/unit/test_cli.py#L537-L538).

Ah, so I'm just blind.

tests/unit/models/test_input.py
eskultety: > If there's any automation you know of we could employ, it might be worth it.

I'm not familiar with GH actions yet, but my first simplistic idea would be to spin up another container and just run (just with a single Python environment, no need for all of them) `git rebase main -x 'make test-unit'`. With a bit of tweaking, i.e. without a shell, it should fail the whole GH action at first commit that doesn't pass the test suite.

cachi2/core/models/input.py
```diff
@@ -99,7 +99,7 @@ def _requirements_file_path_is_relative(cls, paths: list[Path]) -> list[Path]:
 
 
 class YarnPackageInput(_PackageInputBase):
```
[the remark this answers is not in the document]
ben-alkov: Good point, thx

cachi2/core/models/input.py
ben-alkov: Right, but really, it **did** come in prematurely, and I'd like to see it in a commit where it makes sense.

eskultety (approved):

PR #374, opened 2023-10-25
cachi2 yarn: purls

slimreaper35:

### DIFF
--- a/cachi2/core/package_managers/yarn/resolver.py
+++ b/cachi2/core/package_managers/yarn/resolver.py
@@ -197,11 +197,32 @@ class _ComponentResolver:
                 ),
             ) from e
 
+        purl = self._generate_purl_for_package(resolved_package, self._project)
+
         return Component(
             name=resolved_package.name,
             version=resolved_package.version,
-            purl=_generate_purl_for_package(resolved_package, self._project),
+            purl=purl,
         )
+
+    @staticmethod
+    def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> str:
+        """Create a purl for a package based on its protocol.
+
+        :param package: the resolved package to be used in the purl generation.
+        :param project: the project object to resolve the configured registry url and file paths
+            for file dependencies.
+        """
+        # registry url can be accessed in project.yarnrc
+        # paths for file dependencies are relative to project.source_dir
+        return PackageURL(
+            type="npm",
+            name=package.name.lower(),
+            version=package.version,
+            # TODO: used to make sure purls are unique even for an incomplete implementation
+            #   remove raw_locator when no longer needed
+            qualifiers={"raw_locator": package.raw_locator},
+        ).to_string()
 
     def _resolve_package(self, package: Package) -> _ResolvedPackage:
         """Resolve the real name and version of the package."""
@@ -335,22 +356,3 @@ class _ComponentResolver:
             return self._project_subpath(cache_path)
         else:
             return self._output_dir.join_within_root(cache_path)
-
-
-def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> str:
-    """Create a purl for a package based on its protocol.
-
-    :param package: the resolved package to be used in the purl generation.
-    :param project: the project object to resolve the configured registry url and file paths
-        for file dependencies.
-    """
-    # registry url can be accessed in project.yarnrc
-    # paths for file dependencies are relative to project.source_dir
-    return PackageURL(
-        type="npm",
-        name=package.name.lower(),
-        version=package.version,
-        # TODO: used to make sure purls are unique even for an incomplete implementation
-        #   remove raw_locator when no longer needed
-        qualifiers={"raw_locator": package.raw_locator},
-    ).to_string()
--- a/cachi2/core/package_managers/yarn/resolver.py
+++ b/cachi2/core/package_managers/yarn/resolver.py
@@ -31,6 +31,7 @@ from cachi2.core.package_managers.yarn.locators import (
 from cachi2.core.package_managers.yarn.project import Optional, Project
 from cachi2.core.package_managers.yarn.utils import run_yarn_cmd
 from cachi2.core.rooted_path import RootedPath
+from cachi2.core.scm import get_repo_id
 
 if TYPE_CHECKING:
     # Import conditionally so that we don't have to introduce a runtime dependency on
@@ -166,10 +167,7 @@ class _ResolvedPackage:
     locator: Locator
     name: str
     version: Optional[str]
-
-    # TODO: used to make sure purls are unique even for an incomplete implementation
-    #   remove raw_locator when no longer needed
-    raw_locator: str
+    checksum: Optional[str]
 
 
 class _CouldNotResolve(ValueError):
@@ -213,15 +211,54 @@ class _ComponentResolver:
         :param project: the project object to resolve the configured registry url and file paths
             for file dependencies.
         """
-        # registry url can be accessed in project.yarnrc
-        # paths for file dependencies are relative to project.source_dir
+        qualifiers = dict()
+        subpath = None
+
+        if isinstance(package.locator, NpmLocator):
+            # package with NpmLocator doesn't need any qualifiers
+            pass
+
+        elif isinstance(package.locator, HttpsLocator):
+            qualifiers["download_url"] = package.locator.url
+            if package.checksum is not None:
+                # yarn berry (hopefully still) uses sha512 for checksums
+                # https://github.com/yarnpkg/berry/blob/017b94ae4eb20dea14ac673a053a1f2974b778ff/packages/yarnpkg-core/sources/hashUtils.ts#L84
+                qualifiers["checksum"] = f"sha512:{package.checksum}"
+
+        elif isinstance(package.locator, WorkspaceLocator):
+            project_path = project.source_dir
+            workspace_path = package.locator.relpath
+
+            repo = get_repo_id(project_path)
+
+            qualifiers["vcs_url"] = repo.as_vcs_url_qualifier()
+            subpath = str(workspace_path)
+
+        elif isinstance(package.locator, (FileLocator, LinkLocator, PortalLocator)):
+            project_path = project.source_dir
+            workspace_path = package.locator.locator.relpath
+            package_path = package.locator.relpath
+
+            normalized = project_path.join_within_root(workspace_path, package_path)
+
+            repo = get_repo_id(project_path)
+            qualifiers["vcs_url"] = repo.as_vcs_url_qualifier()
+            subpath = str(normalized.subpath_from_root)
+
+        elif isinstance(package.locator, PatchLocator):
+            # ignore patch locators
+            # the actual dependency that is patched is reported separately
+            # the patch itself will be reported via SBOM pedigree patches
+            pass
+        else:
+            assert_never(package.locator)
+
         return PackageURL(
             type="npm",
             name=package.name.lower(),
             version=package.version,
-            # TODO: used to make sure purls are unique even for an incomplete implementation
-            #   remove raw_locator when no longer needed
-            qualifiers={"raw_locator": package.raw_locator},
+            qualifiers=qualifiers,
+            subpath=subpath,
         ).to_string()
 
     def _resolve_package(self, package: Package) -> _ResolvedPackage:
@@ -231,6 +268,7 @@ class _ComponentResolver:
             log.log(level, f"%s: {msg}", package.raw_locator, *args)
 
         locator = package.parsed_locator
+        checksum = package.checksum
 
         if isinstance(locator, NpmLocator):
             # npm dependencies have reliable names and versions in yarn info output
@@ -295,7 +333,7 @@ class _ComponentResolver:
             # https://typing.readthedocs.io/en/latest/source/unreachable.html#assert-never-and-exhaustiveness-checking
             assert_never(locator)
 
-        return _ResolvedPackage(locator, name, version, package.raw_locator)
+        return _ResolvedPackage(locator, name, version, checksum)
 
     def _read_name_from_cache(self, cache_path: RootedPath) -> str:
         with zipfile.ZipFile(cache_path) as zf:
--- a/tests/unit/package_managers/yarn/test_resolver.py
+++ b/tests/unit/package_managers/yarn/test_resolver.py
@@ -4,6 +4,7 @@ import zipfile
 from pathlib import Path
 from typing import Any, NamedTuple, Optional
 from unittest import mock
+from urllib.parse import quote
 
 import pytest
 
@@ -13,6 +14,7 @@ from cachi2.core.package_managers.yarn.locators import parse_locator
 from cachi2.core.package_managers.yarn.project import PackageJson, Project, YarnRc
 from cachi2.core.package_managers.yarn.resolver import Package, create_components, resolve_packages
 from cachi2.core.rooted_path import RootedPath
+from cachi2.core.scm import RepoID
 
 
 def setup_module() -> None:
@@ -21,6 +23,10 @@ def setup_module() -> None:
 
     yarn_resolver_logger.disabled = False
     yarn_resolver_logger.setLevel("DEBUG")
+
+
+MOCK_REPO_ID = RepoID("https://github.com/org/project.git", "fffffff")
+MOCK_REPO_VCS_URL = quote("git+https://github.com/org/project.git@fffffff", safe="://")
 
 
 def mock_yarn_info_output(yarn_info_outputs: list[dict[str, Any]]) -> str:
@@ -301,6 +307,7 @@ def mock_project(project_dir: RootedPath) -> Project:
     )
 
 
+@mock.patch("cachi2.core.package_managers.yarn.resolver.get_repo_id")
 @pytest.mark.parametrize(
     "mocked_package, expect_component, expect_logs",
     [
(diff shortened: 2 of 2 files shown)


slimreaper35:
@everyone PTAL, I have no idea if I am doing it correctly.

brunoapimentel:
One thing that could help the reviews is adding in the PR description a brief example on how each protocol will resolve to a PURL, since the design doc does not get in depth with that. If it's clear to us what is the expected output, then I think the implementation is simpler to code/review.

slimreaper35:
Yeah, but I don't know which attributes from locators are important for PURL generation since the `npm` [specification](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#npm) is very lightweight.

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -344,13 +342,40 @@ def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> s
     :param project: the project object to resolve the configured registry url and file paths
         for file dependencies.
     """
-    # registry url can be accessed in project.yarnrc
-    # paths for file dependencies are relative to project.source_dir
+    qualifiers = dict()
+    if package.checksum is not None:
+        qualifiers["checksum"] = package.checksum
```
brunoapimentel, line 351: I believe we can keep the checksums exclusively for HttpsLocator packages. On the other hand, we should report whichever packages didn't have checksums verified, but that's another story (and that will be done via a property, not purls). Right @taylormadore ^^'?

brunoapimentel, line 351: Another detail we need to handle here is to prefix the checksum with the algorithm used (I'm not sure what it is, we need to look into it). The only place I could find that mentions it in the spec is [here](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairs), but there might be more info there.

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -344,13 +346,45 @@ def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> s
...
+
+        qualifiers["vcs_url"] = str(workspace_abs_path)
+        subpath = str(workspace_abs_path.subpath_from_root)
+
+    elif isinstance(package.locator, (FileLocator, LinkLocator, PortalLocator)):
+        workspace_path = package.locator.locator.relpath
+        workspace_abs_path = project.source_dir.join_within_root(workspace_path)
+        package_abs_path = workspace_abs_path.join_within_root(package.locator.relpath)
+
+        qualifiers["vcs_url"] = str(package_abs_path)
```
brunoapimentel, line 370: We need to take in consideration the `origin` of the repository, otherwise we end up with a local path that won't help locating the package on the internet:

```
pkg:npm/ansi-regex@6.0.1?vcs_url=/home/bpimente/Projects/cachito/test-repos/cachi2-yarn-berry/external-packages/ansi-regex#external-packages/ansi-regex
```

It should look like this:
```
pkg:npm/ansi-regex@6.0.1?vcs_url=git+https://github.com/cachito-testing/cachi2-yarn-berry@70515793108df42547d3320c7ea4cd6b6e505c46#external-packages/ansi-regex
```

cachi2/core/package_managers/yarn/resolver.py
slimreaper35: [sha512](https://github.com/yarnpkg/berry/blob/017b94ae4eb20dea14ac673a053a1f2974b778ff/packages/yarnpkg-core/sources/hashUtils.ts#L84)

eskultety: The other piece of useful information is that they actually use that function for checksums within the lockfile: https://github.com/yarnpkg/berry/blob/017b94ae4eb20dea14ac673a053a1f2974b778ff/packages/yarnpkg-core/sources/Project.ts#L207.
IMO we need to keep references to this information in a code commentary for future reference. This may come handy at some point since the hash algorithm used for checksums hasn't been mentioned anywhere in Yarn docs and thus the algorithm used can be changed by Yarn at will without us even noticing (e.g. when sha512 becomes insecure) ...unless we want to start recomputing checksums  just to make sure the algorithm still checks.

cachi2/core/package_managers/yarn/resolver.py
brunoapimentel: Fail it =/
https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/scm.py#L49-L56

We should probably revisit this later, as I think it makes for bad user experience.

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -344,13 +343,40 @@ def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> s
...
+
+        qualifiers["vcs_url"] = str(workspace_abs_path)
+        subpath = str(workspace_abs_path.subpath_from_root)
+
+    elif isinstance(package.locator, (FileLocator, LinkLocator, PortalLocator)):
+        workspace_path = package.locator.locator.relpath
+        workspace_abs_path = project.source_dir.join_within_root(workspace_path)
+        package_abs_path = workspace_abs_path.join_within_root(package.locator.relpath)  # noqa
+
+        repo = get_repo_id(workspace_abs_path)
```
taylormadore, line 366: We'll need to pass the path to the root of the git repository here

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -344,13 +343,44 @@ def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> s
...
-    # paths for file dependencies are relative to project.source_dir
+    qualifiers = dict()
+    subpath = None
+
+    # package with NpmLocator doesn't need any qualifiers
+
+    if isinstance(package.locator, HttpsLocator):
+        qualifiers["download_url"] = package.locator.url
+        if package.checksum is not None:
+            qualifiers["checksum"] = f"sha512-{package.checksum}"
```
brunoapimentel, line 354: The separator between algorithm and checksum should be a `:`. 

https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#generic

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -344,13 +343,44 @@ def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> s
...
+        subpath = str(workspace_path)
+
+    elif isinstance(package.locator, (FileLocator, LinkLocator, PortalLocator)):
+        project_path = project.source_dir
+        workspace_path = package.locator.locator.relpath
+        package_path = package.locator.relpath
+
+        repo = get_repo_id(project_path)
+        qualifiers["vcs_url"] = repo.as_vcs_url_qualifier()
+        subpath = str(workspace_path.joinpath(package_path))
```
brunoapimentel, line 372: I did a quick test using our test repo as [source](https://github.com/cachito-testing/cachi2-yarn-berry/tree/no-zero-installs). Looking at the results, I got two distinct results for local `strip-ansi:

```sh
# correct, probably derived from the main workspace
pkg:npm/strip-ansi@4.0.0?vcs_url=git%2Bhttps://github.com/cachito-testing/cachi2-yarn-berry%4070515793108df42547d3320c7ea4cd6b6e505c46#external-packages/strip-ansi-4.0.0.tgz

# incorrect, probably derived from packages/then-answer workspace
pkg:npm/strip-ansi@4.0.0?vcs_url=git%2Bhttps://github.com/cachito-testing/cachi2-yarn-berry%4070515793108df42547d3320c7ea4cd6b6e505c46#packages/the-answer/external-packages/strip-ansi-4.0.0.tgz
```

The second one's locator is `file:../../external-packages/strip-ansi-4.0.0.tgz`, so I'm guessing the relative import is not being properly resolved.

brunoapimentel:
Minor nitpick: it feels to me that the `_generate_purl_for_package` function should be a method of `_ComponentResolver`, since it's only ever used by this object, in the context of generation a proper Component.

(I know the change was not introduced in this PR, but maybe we could add a new commit to address this)

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -804,3 +865,103 @@ def test_create_components_cache_path_reported_but_missing(rooted_tmp_path: Root
...
+                ),
+                name="patch-locator-package",
+                version="7.0.0",
+                checksum=None,
+            ),
+            "pkg:npm/patch-locator-package@7.0.0",
+        ),
+    ],
+)
+def test_generate_purl_for_package(
```
brunoapimentel, line 959: I wonder if adding this test here brings any additional value compared to the already existing `test_create_components_single_package` and `test_create_components_patched_packages`, since `_generate_purl_for_package` is called as part of the component creation. Does this bring any additional coverage?

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -436,7 +481,11 @@ def mock_project(project_dir: RootedPath) -> Project:
                 packjson_path="node_modules/strip-ansi-tarball/package.json",
                 packjson_content=json.dumps({"name": "strip-ansi"}),
             ),
-            Component(name="strip-ansi", version="4.0.0", purl=""),
+            Component(
+                name="strip-ansi",
+                version="4.0.0",
+                purl=f"pkg:npm/strip-ansi@4.0.0?vcs_url={MOCK_REPO_VCS_URL}#external-packages/strip-ansi-4.0.0.tgz",
```
brunoapimentel, line 487: Is it worth adding a scenario that covers packages that live in parent folders? Maybe we missed it the first time because there was no scenario showing it.

brunoapimentel (commented):
The implementation LGTM :rocket: 

I only left two comments about tests, but they should be simple enough.

tests/unit/package_managers/yarn/test_resolver.py
slimreaper35: No, the coverage is the same. I can leave it there or drop it, does not matter to me. WDYT @everyone

slimreaper35: I will drop it since I move the **_generate_purl_for_package** function to be a method of _**ComponentResolver**

eskultety: > test_create_components_single_package and test_create_components_patched_packages

While the coverage may be the same I wonder if we could potentially decrease the scope of those tests? I'm asking because I think **test_generate_purl_for_package** is the correct test for this implementation, the others  are transitive only which may not be obvious at first glance and one has to go digging what happens underneath.

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -344,13 +343,46 @@ def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> s
...
+        repo = get_repo_id(project_path)
+        qualifiers["vcs_url"] = repo.as_vcs_url_qualifier()
+        subpath = str(normalized.subpath_from_root)
+
+    elif isinstance(package.locator, PatchLocator):
+        # ignore patch locators
+        # the actual dependency that is patched is reported separately
+        # the patch itself will be reported via SBOM pedigree patches
+        pass
+
```
eskultety, line 381: Would it be worth it to add a fallthrough `else` branch which would raise an exception about an unknown locator? I know it's highly unlikely a new locator would emerge and we'd have risen an exception in the resolver by now, so I'm just wondering about it, we may as well dismiss that case.

slimreaper35: something like `assert_never` ?

eskultety: Yeah, in fact I think we can just shamelessly copy-paste the whole `else` clause from https://github.com/containerbuildsystem/cachi2/blob/e228b16cfd395b0839348f014b745f1d56d25ef3/cachi2/core/package_managers/yarn/resolver.py#L273

tests/unit/package_managers/yarn/test_resolver.py
brunoapimentel: I think this test's scope is given by the number of different "locator strings vs expected purls" scenarios. I think the currents tests have a good coverage on that.

I do agree that it's less clear to have this specific piece tested as part of `create_components` instead of only `_generate_purl_for_package`, but at the same time, I wonder if it's worth duplicating the scenarios to have a separate test, that's why I raised this point.

brunoapimentel (approved):
LGTM :rocket:

tests/unit/package_managers/yarn/test_resolver.py
slimreaper35: just added the scenario, PTAL

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -195,10 +195,12 @@ def get_component(self, package: Package) -> Component:
                 ),
             ) from e
 
+        purl = self._generate_purl_for_package(resolved_package, self._project)
```
eskultety, line 198: nitpick: can assign this statement directly in the call below?

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -275,6 +277,59 @@ def log_for_locator(msg: str, *args: Any, level: int = logging.DEBUG) -> None:
 
         return _ResolvedPackage(locator, name, version, checksum)
 
+    def _generate_purl_for_package(self, package: _ResolvedPackage, project: Project) -> str:
```
eskultety, line 280: shall we make this method a `staticmethod` since it doesn't interact with `self` in any way?

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -334,57 +389,3 @@ def _cache_path_as_rooted(self, cache_path: str) -> RootedPath:
             return self._project_subpath(cache_path)
         else:
             return self._output_dir.join_within_root(cache_path)
-
```
eskultety, line 338: This commit should have happened before the first one, i.e. before implementing the logic, that way we'd end up with an overall smaller change diff.

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -287,9 +287,11 @@ def _generate_purl_for_package(self, package: _ResolvedPackage, project: Project
         qualifiers = dict()
         subpath = None
 
-        # package with NpmLocator doesn't need any qualifiers
+        if isinstance(package.locator, NpmLocator):
```
eskultety, line 290: This change should have been part of the commit which implements the guts of the method: https://github.com/containerbuildsystem/cachi2/pull/374/commits/0709ff1cdb317fbcf363637028842beca45e1341

eskultety:
Regarding functionality, I don't have any comments as @brunoapimentel has done a good job at digging into the details.

cachi2/core/package_managers/yarn/resolver.py
slimreaper35: Squashed first and last two commits into one commit, which implements a static method of `_ComponentResolver` class for generating PURLs. Everyone okay with that before I force push :fearful: ?

cachi2/core/package_managers/yarn/resolver.py
eskultety: I meant the whole hunk, not just the single line ;) .

cachi2/core/package_managers/yarn/resolver.py
eskultety: just make sure unit tests are passing for each commit in the PR (like they are now).

cachi2/core/package_managers/yarn/resolver.py
eskultety: @slimreaper35 you pushed a new revision, but didn't respond to this comment. I'm okay if we ignore this, but a resolution comment would be nice :).

cachi2/core/package_managers/yarn/resolver.py
eskultety: @slimreaper35 after thoroughly examining the latest revision what I meant by

> This commit should have happened before the first one, i.e. before implementing the logic, that way we'd end up with an overall smaller change diff.

was that:

- in commit 1 you convert `generate_purl_for_package` to a static instance method
- in commit 2 you provide the actual implementation for the method
- in commit 3 you take care of the unit tests

If you squash everything in a single commit that (from logic POV) adds too many changes in a single commit whereas we want gradual well-structured changes.

I apologize for the misunderstanding that may have originated from my comment

> just make sure unit tests are passing for each commit in the PR (like they are now).

as I only put ^this comment in not to forget about running unit tests after each patch without actually paying much attention to the details you posted before me leading to the confusion that I may have approved of squashing the 2 commits together. After all, it's better to just push a new revision right away and see if there are more comments rather than waiting for approval for future changes in the next revision.

cachi2/core/package_managers/yarn/resolver.py
slimreaper35: It was just my personal visual preference of putting the assignment of purl on a separate line. There is no other reason behind it, so, I am okay with both options as well.

cachi2/core/package_managers/yarn/resolver.py
eskultety: @slimreaper35 do you plan on taking care of the commits as suggested ^above?

slimreaper35: I will try

eskultety (approved):

PR #377, opened 2023-10-31
Yarn: Misc fixes/adjustments/improvements

eskultety:
- convert YarnRc to a dict-like structure as exposing getters/setters doesn't scale with the number of possible config options
- populate YarnRc defaults from `yarn config`
- always instantiate YarnRc (because we now have defaults populated)
- add a helper that fixes Yarn's JSON output to a properly serialized JSON for the *all* Yarn callers

### DIFF
--- a/cachi2/core/package_managers/yarn/project.py
+++ b/cachi2/core/package_managers/yarn/project.py
@@ -32,7 +32,9 @@ class YarnRc(UserDict):
     configuration file.
     """
 
-    def __init__(self, data: dict[str, Any], defaults: Optional[dict[str, Any]] = None) -> None:
+    def __init__(
+        self, data: dict[str, Any] = {}, defaults: Optional[dict[str, Any]] = None
+    ) -> None:
         """Initialize a YarnRc dictionary.
 
         :param data: the raw data for the yarnrc file.
@@ -183,7 +185,7 @@ class Project(NamedTuple):
     """A directory containing yarn sources."""
 
     source_dir: RootedPath
-    yarn_rc: Optional[YarnRc]
+    yarn_rc: YarnRc
     package_json: PackageJson
 
     @property
@@ -222,7 +224,8 @@ class Project(NamedTuple):
                 source_dir.join_within_root(".yarnrc.yml"), YarnRc.load_defaults(source_dir)
             )
         else:
-            yarn_rc = None
+            # defaults only
+            yarn_rc = YarnRc(YarnRc.load_defaults(source_dir))
 
         package_json = PackageJson.from_file(source_dir.join_within_root("package.json"))
         return cls(source_dir, yarn_rc, package_json)
--- a/cachi2/core/package_managers/yarn/project.py
+++ b/cachi2/core/package_managers/yarn/project.py
@@ -79,9 +79,9 @@ class YarnRc(UserDict):
 
         See: https://v3.yarnpkg.com/configuration/yarnrc#npmScopes
         """
-        registry = self._data.get("npmScopes", {}).get(scope, {}).get("npmRegistryServer")
+        registry = self.get("npmScopes", {}).get(scope, {}).get("npmRegistryServer")
 
-        return registry or self.registry_server
+        return registry or self["npmRegistryServer"]
 
     @staticmethod
     def load_defaults(source_dir: RootedPath) -> dict[str, Any]:
@@ -195,12 +195,12 @@ class Project(NamedTuple):
         This is determined by the existence of a non-empty yarn cache folder. For more details on
         zero-installs, see: https://v3.yarnpkg.com/features/zero-installs.
         """
-        dir = self.yarn_cache
+        cache_rooted_path = self.source_dir.join_within_root(self.yarn_rc["cacheFolder"])
 
-        if not dir.path.is_dir():
+        if not cache_rooted_path.path.is_dir():
             return False
 
-        return any(file.suffix == ".zip" for file in dir.path.iterdir())
+        return any(file.suffix == ".zip" for file in cache_rooted_path.path.iterdir())
 
     @property
     def yarn_cache(self) -> RootedPath:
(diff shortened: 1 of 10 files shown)


cachi2/core/package_managers/yarn/project.py
```diff
@@ -73,8 +81,23 @@ def registry_server_for_scope(self, scope: str) -> str:
...
+    @staticmethod
+    def load_defaults(source_dir: RootedPath) -> dict[str, Any]:
+        """Load Yarn RC defaults.
+
+        :param source_dir: directory context used to determine the right yarn binary to execute,
+                           e.g. yarn v1 doesn't support the 'config' subcommand
+        :returns: dictionary containing Yarn RC defaults if form of {"<option>": "<default_value>"}
+        """
+
+        yarn_config = json.loads(run_yarn_cmd(["config", "--json"], source_dir))
```
chmeliik, line 69: You may need to pass `YARN_IGNORE_PATH=true` as an env var here. Otherwise the globally installed `yarn` will replace itself with whatever the user has configured as `yarnPath` in the config file. And that can easily be something malicious.

(I'm assuming `load_defaults()` happens before we've had a chance to set `ignorePath` in the config file)

eskultety:
since v1:
- defaults are loaded to a separate private attribute and are only merged with the user values on demand, hence not colliding with any potential `write` method that could write the file's contents to disk
- YarnRc doesn't need the `_path` attribute; through OOP's principles this class should be very lean only abstract the config values, any other knowledge, e.g. which file the values should be read from or where should the file be written to should be handled by the caller

chmeliik (commented):
Just a fly-by comment from your local arbitrary-code-execution detector :wave:

cachi2/core/package_managers/yarn/project.py
eskultety: Oh, I was blindly relying on the directory context. I did wonder whether that's a reliable way of doing it. Thanks, I will change it accordingly.

chmeliik: The directory context probably would be enough if the .yarnrc.yml file already had `ignorePath: true` set. But if loading the defaults is part of YarnRC instantiation, and we use YarnRC to overwrite the .yarnrc.yml, then that's a bit of a chicken and egg proglem (thankfully it's not, thanks to the env vars)

eskultety: @chmeliik question: In my local setup I have (mistakenly) installed Yarn v1 globally and v4 in the user repo. Do we check early and refuse Yarn <v2+? Otherwise I'd have to add more handling and checks to make sure right Yarn is executed because v1 doesn't have the `config` subcommand.

chmeliik: I think we just assume that the global yarn is the [corepack](https://nodejs.org/api/corepack.html) yarn shim. Which is true in the cachi2 container, but not necessarily elsewhere

brunoapimentel:
From a quick test I did locally here, your changes seem to work even when rebased to the current content on the main branch (rebasing is gonna be painful though).

I definitely prefer dumping all the current `@properties` and having a leaner, even if more complex, object to handle. I personally don't like the dict syntax, so I feel we could probably also solve this 
 by keeping YarnRc as a basic class and overriding `__getattr__`/`__setattr__`. This is more of a nitpick though, since I think the performance gain from dropping `UserDict` is insignificant, if there'd be gains at all.

I do feel the team should be comfortable with the design choice here, since I really think we should change `PackageJson` to match what we're doing with `YarnRc` (minus the defaults loading, since they don't apply there). These types of patterns could even be something we reuse among different package managers.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -7,58 +7,43 @@
...
-        """Initialize a YarnRc object.
+    def __init__(
+        self, data: dict[str, Any] = {}, defaults: Optional[dict[str, Any]] = None
+    ) -> None:
+        """Initialize a YarnRc dictionary.
 
-        :param path: the path to the yarnrc file, relative to the request source dir.
         :param data: the raw data for the yarnrc file.
         """
-        self._path = path
```
brunoapimentel, line 39: > why should the data be written to a static location on the disk rather than making the interface more flexible and let the caller specify where the config should be dumped to.

Because then the caller doesn't need to have knowledge of the file location in the disk :stuck_out_tongue:  

We'll be calling `write` on yarn/main.py, and I don't think it should know/care about the file locations. We could give this responsibility to the `Project` class, though, if you want to make `write` more flexible.

eskultety:
>I personally don't like the dict syntax, so I feel we could probably have a similar result by keeping YarnRc as a basic class and overriding `__getattr__`/`__setattr__`.

So, I might agree on going down the `__getattr__/__setattr__` road as I too find it more "ergonomic" if it didn't:

- result in adding more IMO unnecessary code converting dicts to instance attributes in `__init__`, since dict is already what we get from the YAML parser and `UserDict` is nicely compatible in that regard by default
- end up having issues representing nested dictionaries (which Yarn RC does have for a few selected options) where we'd have to chain-nest YarnRc instances instead of dealing with plain `dict`s (although it is fair to say that `UserDict` is nowhere near as API compatible as plain `dict`)

cachi2/core/package_managers/yarn/project.py
eskultety: > Because then the caller doesn't need to have knowledge of the file location in the disk 

That's the thing though. A config map is just an in-memory representation with an optional bunch of utility methods, but since the only use case for the "knowledge" of the destination path of the config is for the `write` utility method it feels like this particular class isn't the right place to store it. Moreover, you need to know upfront where you'd want to dump the RC file to when you instantiate the class. What if you need to dump it elsewhere, i.e. not to the usual location (e.g. during unit tests)? From that POV to me it makes sense to push the ownership of the disk location where it makes more sense - the caller.

> We could give this responsibility to the Project class, though, if you want to make write more flexible.

Agreed, this is what the term `caller` in the commit message was originally meant to refer to. :)

eskultety:
If everyone is in agreement with this approach, then I'll update this PR to contain similar changes for the package.json representation.

brunoapimentel:
> If everyone is in agreement with this approach, then I'll update this PR to contain similar changes for the package.json representation.

I think we should do this in a follow-up PR. It is not strictly needed here, it's just a direction I think we should take later on.

PR #388, opened 2023-11-13
Copy source code directory to avoid unwanted modifications

brunoapimentel:
The initial design defined that Yarn requests should perform their operations in a copy of the source code to avoid unwanted modifications.

But this pattern might also be helpful for gomod requests (since go commands can potentially modify sources) and could be helpful for future package managers. This proposal moves the source code copying to the request level, in a way that the resulting workflow is roughly:

- parse user input from the CLI
- copy the source code directory
- performing the processing for all package managers
- delete the copied source code
- write the output to disk


brunoapimentel:
@ben-alkov @taylormadore @ejegrova @eskultety PTAQL and let me know if you agree with the proposal to perform the copy at a request level. I'll improve the second commit later if you agree with it.

cachi2/interface/cli.py
```diff
@@ -132,6 +133,17 @@ def fetch_deps(
         resolve_path=True,
         help="Write output files to this directory.",
     ),
+    backup_source: bool = typer.Option(
+        True,
+        "--backup-source",
```
eskultety, line 138: Do we really need this configurable? My understanding is that copying the sources is something cachi2 deeply cares about wrt/ "immutable" dependency fetching and validation, but not necessarily what users care about, IOW the fact we copy trees should be transparent to users. I might be missing some crucial bit of information though, so let me know if ^this is an incorrect premise.

Also, keep in mind that not only direct users of cachi2 might not really care about how we handle project processing internally,but if they use kernel 5.17 or lower with a file system supporting reflinks and the point the backup to a different mount point on the very same file system as the project source, they might encounter some `ReflinkImpossibleError` mumbo jumbo error about unsupported cross device links [1] leaving them in confusion what the real problem is, so I think this whole thing should be handled by us transparently.

[1] https://btrfs.readthedocs.io/en/latest/Reflink.html (I'm sorry I couldn't find a dedicated resource on the 5.17 restriction, so BTRFS docs will have to do to make my point)

cachi2/interface/cli.py
```diff
@@ -22,6 +22,7 @@
 log = logging.getLogger(__name__)
 
 DEFAULT_SOURCE = "."
+DEFAULT_SOURCE_BACKUP = "./cachi2-source"
```
eskultety, line 25: Static paths created and owned by a regular user (which cachi2 runs as) are predictable and thus error prone (e.g. DOS attack), so we always need to default to a directory with a random name.

Moreover, while it is perfectly fine to default to `./`  for the reflink use case because only metadata is copied and not the actual blocks, we should default to a directory residing on a pseudo filesystem on the fallback path to reduce wear on the hard drive since we'd be writing lots of small files with every project request.

cachi2/core/resolver.py
```diff
@@ -22,7 +27,39 @@
...
+    """Switches the source_dir of a Request for a full copy."""
+    original_source_dir = request.source_dir
+    source_backup = copy_directory(original_source_dir.path, backup_path)
+    request.source_dir = RootedPath(source_backup)
+
+    try:
+        yield request
+    finally:
+        if backup_path.exists():
+            shutil.rmtree(backup_path)
```
eskultety, line 41: Hmm, I'm not sure it `path.exists` is a sufficient check to perform the removal afterwards safely (it probably is for our use case), well, at least not without `ignore_errors=True`, but in context of temporary directories I mentioned in my other comments, this would likely become a `dir.cleanup(ignore_cleanup_errors=True)` which is essentially the same thing wrapped by different API.

cachi2/core/utils.py
```diff
@@ -74,3 +77,28 @@ def load_json_stream(s: str) -> Iterator:
...
         yield obj
+
+
+def copy_directory(origin: Path, destination: Path, reflink_if_supported: bool = True) -> Path:
+    """
+    Recursively copy directory to another path.
+
+    Use reflinks by default if the file system supports it.
+
+    :param reflink_if_supported: if set to false, reflinks won't be used regardless of
```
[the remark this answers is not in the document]
brunoapimentel: Since I bumped into a `ReflinkImpossibleError`, I figured there could be more corner cases in which we default to reflinks, only to make the request unprocessable. So the reasoning is that the user can, if he wants to, force Cachi2 to not use reflinks, although the default is using them if available.

We could have an automatic error fallback (default to reflink -> reflink copy fails -> try to use a normal copy), it is only more complex.

cachi2/interface/cli.py
brunoapimentel: Yeah, the idea is to make it transparent (that's why we have the default set), but also to allow the user to specify a different path if a corner case arises. We never know how the user environment will look like, so I thought the flexibilty could help.

I do agree it is to cover a very minority of cases, so maybe we could make this a [config flag](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/config.py), or even ditch it for now and add it later if a user requests it.

cachi2/interface/cli.py
brunoapimentel: I think it's fine to use a random name for the dir, but note that we already use a static path for the default output dir currently.

On the second matter, I think my knowledge is lacking here. What do you mean by pseudo filesystem, and how would we go about making this copy in one?

eskultety: > I think it's fine to use a random name for the dir, but note that we already use a static path for the default output dir currently.

That's different though, because outputdir is user facing, the copied dir proppsed in this PR is not. However, we might want to look at writing the output dir atomically.

> What do you mean by pseudo filesystem
pseudo filesystem == not a physical file system, it's purely RAM-backed, e.g. tmpfs, devfs, procfs, etc.

brunoapimentel: My initial approach was to copy it to `/tmp`, but I bumped into a `ReflinkImpossibleError: EXDEV` error since it identifies it as cross device. This is one of the reasons I thought it would be wise to allow the user to force-disable reflinks.  I'm on `6.4.10-200.fc38.x86_64`, btw.

cachi2/interface/cli.py
eskultety: > but also to allow the user to specify a different path if a corner case arises. We never know how the user environment will look like

Well, if the idea is truly making it transparent to the user, I don't think we should make it configurable, because you'd be providing configuration for behaviour that is to remain an internal implementation detail not documented anywhere - otherwise it's not transparent and requires user interaction (with a set default). The fact that we get rid of the copied directory before user request processed suggests the user isn't supposed to interact with this in any way (as they shouldn't!).

What gives us the confidence users will know what to do with `ReflinkImpossibleError` errors if we currently don't know all the corner cases. From that POV it makes sense to me to just model this as:
```
if reflink_supported:
    try:
        copy_with_reflinks()
   except ReflinkImpossibleError:
        fallback to regular copy (let the exception from copy propagate all the way up)
else:
    use regular copy
```

> so maybe we could make this a [config flag](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/config.py)

In my book config options should match CLI options (simply for proper order of precedence wrt/ configuration), so once we add a config option, it should be followed by a CLI options. Nonetheless, either of the two will instantly make this behaviour a public API which we have to be very careful about not to fall into a trap of being stuck with it and not being able to change it easily at our will without breaking CLI logic and use cases (this goes in general, not just for this proposal)

cachi2/interface/cli.py
eskultety: Yeah, that is expected, so that's why I suggested to go with a random name under `./` with reflinks as that should go smooth as butter™ and only copy to something like `/tmp` (it's not a tmpfs everywhere, but I guess very few projects actually check if it is to reduce the wear anyway) on fallback.

ben-alkov: I think the question about where to write is very environment-dependent, which (in my mind) leads to the question of what environment(s) we want to fully support. Local (bare metal) mode is fine for development, but I'm not sure that I see a *useful* use case otherwise.

All of which leads me to

> reduce wear on the hard drive since we'd be writing lots of small files with every project request

which *really* feels like
1. premature optimization
2. not our problem, given that our *primary* "production" use case is in virtualized or containerized environments

cachi2/interface/cli.py
ben-alkov: Definitely +1 on this being an implementation detail, and not something end users should even know about.

cachi2/core/utils.py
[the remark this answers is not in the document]
brunoapimentel: Changed the source code to fallback to `shutil.copy2` in case a `ReflinkImpossibleError` happens.

cachi2/interface/cli.py
brunoapimentel: Removed all the flags from the CLI and changed the code to fallback to `shutil.copy2` in case a `ReflinkImpossibleError` happens.

cachi2/interface/cli.py
brunoapimentel: I kept the copy location inside the current `source_dir` for simplicity reasons. If we figure out a way to make use of `/tmp`, I can update this PR, otherwise we can leave it as an improvement later.

brunoapimentel: What is worrying me now with the copy approach is that we might have size limitations in the `workspace` volume in RHTAP, which could essentially block users that want to build larger Yarn repos. I'll try to figure out more about size constraints tomorrow.

brunoapimentel:
New force push:
- Removed the CLI configuration flags
- Randomize the name of the copy directory
- Added `shutil.copy2` fallback in case the reflink copy fails
- Make the copy only in case Yarn packages are present

cachi2/interface/cli.py
eskultety: > which really feels like
>
>   1. premature optimization
>   2. not our problem, given that our primary "production" use case is in virtualized or containerized environments

I don't think it is premature at all, the problem is very real. While the general mindset in production clouds is that the industry doesn't care about storage wear - true, but local deployments do care very much. Again, we need to differentiate what a given company's expectations wrt/ this project are vs what is important/beneficial for upstream community in general and how these two can align. I understand that from production POV we'll probably only ever see containerized environments, but development environments should not be in a disadvantage. IOW I as a developer who prefers local deployments for instant feedback loop (let's face it CI/CD is designed to get significantly slower with time when it comes to feedback loop) very much care about not wearing down my SSD prematurely, as it's my day-in-day-out job to test my code (and others' too). So, I personally am not convinced about downplaying the development use case by saying it's not "fully" supported.

That said, I won't stand in the way of always copying to `./` and let end users deal with the fact they need to pick different file system for their repo (like it would be my case going from EXT4 to e.g. BTRFS just for git repos) if the overall consensus is that we don't want to special case this.

Another alternative I could think of  would be re-visiting what @brunoapimentel proposed originally with a CLI option, but instead calling it `--backup-dir` and pitching this as a directory where we copy the whole project and then remove it before the user could even touch it, we could have a more generic CLI option `--tmpdir` and pitch this as an "alternative directory for cachi2 to use for its temporary data which depending on the project could be very large, surpassing the capabilities of /tmp" instead. We're not giving away any internal implementation detail, it's as vague as possible and nobody should expect any kind of stability in that regard.

eskultety: > What is worrying me now with the copy approach is that we might have size limitations in the workspace volume in RHTAP, which could essentially block users that want to build larger Yarn repos. I'll try to figure out more about size constraints tomorrow.

This constraint exists outside of cachi2 though, so not something we should concern ourselves with in this project and in this PR IMO, but yes, still worth investigating what size constraints we're dealing with in that specific deployment.

cachi2/core/utils.py
```diff
@@ -74,3 +77,30 @@ def load_json_stream(s: str) -> Iterator:
...
+    """
+    Recursively copy directory to another path.
+
+    Use reflinks by default if the file system supports it.
+
+    :raise FileExistsError: if the destination path already exists.
+    :raise FileNotFoundError: if the origin directory does not exist.
+    """
+
+    def copy_using(copy_function: Callable) -> None:
```
eskultety, line 92: Can we prefix the name with a _ to make it more obvious later that `copy_using` isn't a library function?

cachi2/core/resolver.py
```diff
@@ -22,7 +28,36 @@
 supported_package_managers = list(_package_managers)
 
 
+@contextmanager
+def make_source_dir_work_copy(request: Request) -> Iterator:
+    """Switches the source_dir of a Request for a full copy."""
+    original_source_dir = request.source_dir
+    copy_dir_name = "./.cachi2-source-copy-" + secrets.token_hex(8)
```
eskultety, line 35: Why do we want to handle creation of a uniquely-named directory on our own when there's `tempfile.TemporaryDirectory`?

cachi2/interface/cli.py
eskultety: In hindsight I think we can add a global (i.e. package manager agnostic) CLI option `--tmpdir` in a follow up PR and not block this one. However, I still think we should keep in mind local environments and frequent integration tests where we shouldn't copy/write unnecessary data.

cachi2/core/resolver.py
ben-alkov: E.g. `tempfile.mkdtemp(prefix="cachi2-source-copy-", dir=./)` (the `dir` arg is optional and defaults to system-specifc temp dir(s), but it can be anything)

eskultety: > E.g. tempfile.mkdtemp(prefix="cachi2-source-copy-", dir=./) (the dir arg is optional and defaults to system-specifc temp dir(s), but it can be anything)

one doesn't even need the low-level C-wrapping API, even the high-level Python API is totally fine for our case :)

ben-alkov: That *is* [the Python API ](https://docs.python.org/3.9/library/tempfile.html#tempfile.mkdtemp)... or do you mean [`tempfile.TemporaryDirectory`](https://docs.python.org/3.9/library/tempfile.html#tempfile.TemporaryDirectory)?

eskultety: My bad, I was under the false impression that `tempfile.mkdtemp` specifically is just one of the Python API bindings over some C implementation counterpart, in this case https://linux.die.net/man/3/mkdtemp or the corresponding series of syscalls. Turns out, I was wrong and [tempile](https://github.com/python/cpython/blob/main/Lib/tempfile.py#L344) is pure Python code after all, sorry for the noise.

cachi2/core/resolver.py
```diff
@@ -30,16 +30,24 @@ def resolve_packages(request: Request) -> RequestOutput:
...
+    a package manager that can make unwanted modifications will be used.
     """
+    if request.yarn_packages:
+        return _call_on_source_dir_work_copy(request, _resolve_packages)
+    else:
+        return _resolve_packages(request)
+
+
+def _call_on_source_dir_work_copy(request: Request, handler: Handler) -> RequestOutput:
+    """Call the handler function on work copy of a request's source dir."""
```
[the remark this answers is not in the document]
brunoapimentel: I went with the latter. I kind of liked the previous version as it was compact and clear, but probably because I wrote it, so it was not confusing to me :grimacing:

tests/unit/test_resolver.py
```diff
@@ -100,18 +100,34 @@ def fetch(req: Request) -> RequestOutput:
...
-        packages=[{"type": "yarn"}],
+        packages=packages,
     )
 
     def _resolve_packages(request: Request) -> None:
-        # assert a temporary directory is being used
-        tmp_dir_name = request.source_dir.path.name
-        assert re.fullmatch("tmp[a-z0-9_]{8}.cachi2-source-copy", tmp_dir_name) is not None
+        if copy_exists:
+            # assert a temporary directory is being used
```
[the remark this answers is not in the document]
brunoapimentel: Yeah... getting the TemporaryDirectory instance would be painful, so I relied on the naming convention for it.

Containerfile
```diff
@@ -10,6 +10,7 @@ RUN microdnf -y install \
     nodejs \
     nodejs-npm \
     python3 \
+    python3-devel \
```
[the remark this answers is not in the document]
brunoapimentel: I added `gcc` explicitly, but we have quite a long list of implicit dependencies (e.g. `make`, `kernel-headers`). I'm not sure what's the correct approach here, though. Making everything explicit will make this list quite long.

brunoapimentel:
New push: addressed outstanding comments

Containerfile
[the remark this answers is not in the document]
eskultety: I know we do. IMO we need to take one for the team and maintain the exploded list, although long, but at least we'd avoid the issue of breaking the container build when any of the packages currently listed updates their SPEC file in such a way where it no longer will transitively pull dependencies we directly depend on.
Let me give you an example from the previous project I worked on: https://gitlab.com/libvirt/libvirt/-/blob/master/ci/containers/debian-11.Dockerfile?ref_type=heads#L13

The list had to be expanded explicitly because the project hit the issue I just described on multiple occasions. Worth noting though that the Dockerfile content was autogenerated completely, including the package list.

eskultety (approved):

cachi2/interface/cli.py
```diff
@@ -171,10 +172,7 @@ def fetch_deps(
     \b
     # gomod package in a subpath of the source directory (./my-repo/subpath)
     cachi2 fetch-deps --source ./my-repo '{
-        "type": "gomod",
-        "path": "subpath"
-    }'
-
+        "type": "gomod",backup_sources_and_resolve_packages
```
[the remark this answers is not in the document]
eskultety: Ah, know that feeling, it's been a while since I had that kind of mouse, but mine never had the option to paste as the rapid scrolling feature was triggered by clicking the wheel....anyway....

cachi2/core/resolver.py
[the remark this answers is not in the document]
eskultety: Every review is subjective. The only reason I commented on this hunk was that whenever I see a callable being passed around I get triggered, because IMO callbacks have a very specific use case based on architectural design, e.g. plugin implementation, drivers, etc. I simply didn't think we needed that approach in your scenario.

PR #389, opened 2023-11-14
Yarn: check version of Yarn and fail requests for non-v3 binaries

eskultety:
Reasoning:
----------------

- v1 is legacy, doesn't support PnP and hence primitives we rely on for secure dep fetching/checking
- v2 doesn't support the `config` subcommand and `--mode=skip-build` CLI option
- v4 was released only recently with some breaking changes since v3 (support for v4 will be added separately)

- [ ] ~Docs updated (if applicable)~ (will be covered in a future PR)

### DIFF
--- a/cachi2/core/package_managers/yarn/main.py
+++ b/cachi2/core/package_managers/yarn/main.py
@@ -79,6 +79,14 @@ def _configure_yarn_version(project: Project) -> None:
             ),
         )
 
+    # Note (mypy): version cannot be None anymore after the next statement
+    version = yarn_path_version if yarn_path_version else package_manager_version
+    if version.compare("3.0.0") < 0 or version.major == 4:  # type: ignore
+        raise PackageRejected(
+            f"Unsupported Yarn version '{version}' detected",
+            solution="Please pick a different version of Yarn (3.0.0<= Yarn version <4.0.0)",
+        )
+
     if (
         yarn_path_version
         and package_manager_version
--- a/tests/unit/package_managers/yarn/test_main.py
+++ b/tests/unit/package_managers/yarn/test_main.py
@@ -1,5 +1,7 @@
 import re
-from typing import Optional, Union
+from enum import Enum
+from itertools import zip_longest
+from typing import List, Optional, Union
 from unittest import mock
 
 import pytest
@@ -16,14 +18,35 @@ from cachi2.core.package_managers.yarn.project import YarnRc
 from cachi2.core.rooted_path import RootedPath
 
 
+class YarnVersions(Enum):
+    YARN_V1 = semver.VersionInfo(1, 0, 0)
+    YARN_V2 = semver.VersionInfo(2, 0, 0)
+
+    YARN_V3_RC1 = semver.VersionInfo(3, 0, 0, prerelease="rc1")
+    YARN_V3 = semver.VersionInfo(3, 0, 0)
+    YARN_V36_RC1 = semver.VersionInfo(3, 6, 0, prerelease="rc1")
+
+    YARN_V4_RC1 = semver.VersionInfo(4, 0, 0, prerelease="rc1")
+    YARN_V4 = semver.VersionInfo(4, 0, 0)
+
+    @classmethod
+    def supported(cls) -> List["YarnVersions"]:
+        return [cls.YARN_V3, cls.YARN_V36_RC1]
+
+    @classmethod
+    def unsupported(cls) -> List["YarnVersions"]:
+        return list(set(cls.__members__.values()).difference(set(cls.supported())))
+
+
 @pytest.mark.parametrize(
     "yarn_path_version, package_manager_version",
     [
-        pytest.param(semver.VersionInfo(3, 0, 0), None, id="valid-yarnpath-no-packagemanager"),
-        pytest.param(None, semver.VersionInfo(3, 0, 0), id="no-yarnpath-valid-packagemanager"),
+        pytest.param(YarnVersions.YARN_V3.value, None, id="valid-yarnpath-no-packagemanager"),
+        pytest.param(YarnVersions.YARN_V36_RC1.value, None, id="minor-version-with-prerelease"),
+        pytest.param(None, YarnVersions.YARN_V3.value, id="no-yarnpath-valid-packagemanager"),
         pytest.param(
-            semver.VersionInfo(3, 0, 0),
-            semver.VersionInfo(3, 0, 0),
+            YarnVersions.YARN_V3.value,
+            YarnVersions.YARN_V3.value,
             id="matching-yarnpath-and-packagemanager",
         ),
         pytest.param(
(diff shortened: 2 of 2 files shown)


cachi2/core/package_managers/yarn/main.py
```diff
@@ -73,21 +75,25 @@ def _configure_yarn_version(project: Project) -> None:
...
+                    f"and packageManager (yarn@{package_manager_version})"
+                ),
+                solution=(
+                    "Ensure that the versions of yarn specified by yarnPath in .yarnrc.yml and "
+                    "packageManager in package.json agree"
+                ),
+            )
+    else:
+        version = yarn_path_version if yarn_path_version else package_manager_version
+        if version < VersionInfo(3, 0, 0) or version >= VersionInfo(4, 0, 0):  # type: ignore # version cannot be None because we already checked both operands above
```
brunoapimentel, line 92: A way to avoid completely ignoring the type here is to add a dummy conditional such as:
```
if not version:
   raise RuntimeError("")
```

In case you prefer to keep the `type: ignore` solution, can you move the comment explaining why it is needed to a different line? This line currently is much larger than our usual 100 characters limit (not sure if there's a way to configure our linters to catch that though).

brunoapimentel: One last suggestion is to narrow down what we're ignoring to only `# type: ignore[operator]`

cachi2/core/package_managers/yarn/main.py
```diff
@@ -73,21 +75,25 @@ def _configure_yarn_version(project: Project) -> None:
...
+                (
+                    f"Mismatch between the yarn versions specified by yarnPath (yarn@{yarn_path_version}) "
+                    f"and packageManager (yarn@{package_manager_version})"
+                ),
+                solution=(
+                    "Ensure that the versions of yarn specified by yarnPath in .yarnrc.yml and "
+                    "packageManager in package.json agree"
+                ),
+            )
+    else:
```
brunoapimentel, line 90: This else here introduces a flaw: the version will not be verified if both `yarn_path_version` and `package_manager_version` are set, and it should.

You can test by getting a sample yarn repo and do `yarn set version 2.0.0`, then try to process it with Cachi2.

cachi2/core/package_managers/yarn/main.py
eskultety: >```
> if not version:
>   raise RuntimeError("")
>```

We shouldn't add dead code just because mypy isn't clever enough to figure out version cannot be None.

cachi2/core/package_managers/yarn/main.py
eskultety: Doh! Very true.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -73,20 +76,28 @@ def _configure_yarn_version(project: Project) -> None:
             ),
         )
 
-    if (
-        yarn_path_version
-        and package_manager_version
-        and yarn_path_version != package_manager_version
-    ):
+    if yarn_path_version and package_manager_version:
```
taylormadore, line 79: Do we need to nest an additional if/else here? Once the code gets passed the exception checks, either one of yarn_path_version or package_manager_version is defined or if they both are, they are equal.

Couldn't we just have `version = yarn_path_version if yarn_path_version else package_manager_version` right before we check whether the version is within the valid range?

eskultety: Right, moved the check to the top and dropped the extra nesting.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -73,6 +75,14 @@ def _configure_yarn_version(project: Project) -> None:
             ),
         )
 
+    # Note (mypy): version cannot be None anymore after the next statement
+    version = yarn_path_version if yarn_path_version else package_manager_version
+    if version < VersionInfo(3, 0, 0) or version >= VersionInfo(4, 0, 0):  # type: ignore
+        raise PackageRejected(
+            f"Unsupported Yarn version '{version}' detected",
+            solution="Please pick a valid release version of Yarn from the <3.0.0, 4.0.0) range",
+        )
```
slimreaper35, line 84: just a nitpick, I would stick to the mathematical conventions for intervals

e.g. "Please pick a supported Yarn release version within the interval **[**3.0.0, 4.0.0)"

brunoapimentel: I might be the stupid one, but I'd have no clue what `[ )` means without the context, and would only assume it's a typo :laughing:

eskultety: > I think it's easy to assume 4.0.0 is included, but again, I suck bad at math 🤷‍♂️

fair enough, changed - how about now? :)

brunoapimentel:
I just want to point out we're forbidding the 3.0.0-rc.1 release with this approach. I don't think it's a big deal, just want to make sure we are aware of that. Maybe adding a rc version to the unit test is worth it? https://github.com/containerbuildsystem/cachi2/pull/389/files#diff-ff0654690805ebbaf30b927f3cd0adfbb6f93fa7b25fccc63f4c59813116f63cR24

brunoapimentel (approved):
LGTM. Just left a small comment about rc versions.

slimreaper35 (approved):

eskultety:
> I just want to point out we're forbidding the 3.0.0-rc.1 release with this approach. I don't think it's a big deal, just want to make sure we are aware of that. Maybe adding a rc version to the unit test is worth it? https://github.com/containerbuildsystem/cachi2/pull/389/files#diff-ff0654690805ebbaf30b927f3cd0adfbb6f93fa7b25fccc63f4c59813116f63cR24

darn, excellent point @brunoapimentel, rc releases should not be supported which is okay wrt/ 3.0.0 check, but it's wrong for 4.0.0, because 4.0.0-rcX would actually pass the check, so I need to perform a more thorough check there.

brunoapimentel:
> > I just want to point out we're forbidding the 3.0.0-rc.1 release with this approach. I don't think it's a big deal, just want to make sure we are aware of that. Maybe adding a rc version to the unit test is worth it? https://github.com/containerbuildsystem/cachi2/pull/389/files#diff-ff0654690805ebbaf30b927f3cd0adfbb6f93fa7b25fccc63f4c59813116f63cR24
> 
> darn, excellent point @brunoapimentel, rc releases should not be supported which is okay wrt/ 3.0.0 check, but it's wrong for 4.0.0, because 4.0.0-rcX would actually pass the check, so I need to perform a more thorough check there.

Luckily, there is no rc release for the 4.0.0 version :)
https://github.com/yarnpkg/berry/releases?page=1

There's no hurt in making the check more thorough, anyways.

eskultety:
> > > I just want to point out we're forbidding the 3.0.0-rc.1 release with this approach. I don't think it's a big deal, just want to make sure we are aware of that. Maybe adding a rc version to the unit test is worth it? https://github.com/containerbuildsystem/cachi2/pull/389/files#diff-ff0654690805ebbaf30b927f3cd0adfbb6f93fa7b25fccc63f4c59813116f63cR24
> > 
> > 
> > darn, excellent point @brunoapimentel, rc releases should not be supported which is okay wrt/ 3.0.0 check, but it's wrong for 4.0.0, because 4.0.0-rcX would actually pass the check, so I need to perform a more thorough check there.
> 
> Luckily, there is no rc release for the 4.0.0 version :) https://github.com/yarnpkg/berry/releases?page=1
> 
> There's no hurt in making the check more thorough, anyways.

@brunoapimentel please have a look.

tests/unit/package_managers/yarn/test_main.py
```diff
@@ -16,20 +18,41 @@
 from cachi2.core.rooted_path import RootedPath
 
 
+class YarnVersions(Enum):
+    YARN_V1 = semver.VersionInfo(1, 0, 0)
+    YARN_V2 = semver.VersionInfo(2, 0, 0)
+
+    YARN_V3_RC1 = semver.VersionInfo(3, 0, 0, prerelease="rc1")
```
brunoapimentel, line 25: Are we including this in any unit test?

eskultety: Yes,
 `test_main.py::test_yarn_unsupported_version_fail[package_manager,yarn_path-(3.0.0-rc1, None)]`

it'll get into `parametrize` as a set difference of supported and unsupported yarn versions (it was easier to have an enum just to have list comprehensions for parametrize)

brunoapimentel: Right, I overlooked that. LGTM then!

brunoapimentel:
> @brunoapimentel please have a look.

LGTM :rocket:

ejegrova (approved):

PR #391, opened 2023-11-15
Add environment variables to yarn request output

taylormadore:
Several yarn settings must be supplied by the cachi2
request output in order for the user to make use of
the prefetched dependencies. These will be supplied by
way of environment variables, which when sourced, will
override any preexisting user configuration in .yarnrc.yml

YARN_GLOBAL_FOLDER must be configured to the deps/yarn
subdirectory of the cachi2 request output directory. This
is the path that cachi2 will populate with the prefetched
dependencies.

YARN_ENABLE_MIRROR must be set to true, otherwise the user
build will not make use of the global cache located at
the globalFolder path that cachi2 prepopulates.

YARN_ENABLE_GLOBAL_CACHE must be set to false, otherwise the
user build will exclusively rely on the cachi2 populated
globalFolder cache rather than copying the dependencies to
cacheFolder. If the volume containing globalFolder is only mounted
at build-time and the dependencies are not copied to cacheFolder,
the dependencies will be missing at run-time.

YARN_ENABLE_IMMUTABLE_CACHE must be set to false, otherwise
the `yarn install` will fail to populate cacheFolder from the
cachi2-supplied globalFolder cache.

### DIFF
--- a/cachi2/core/package_managers/yarn/main.py
+++ b/cachi2/core/package_managers/yarn/main.py
@@ -160,4 +160,11 @@ def _undo_changes(project: Project) -> None:
 
 def _generate_environment_variables() -> list[EnvironmentVariable]:
     """Generate environment variables that will be used for building the project."""
-    return []
+    env_vars = {
+        "YARN_ENABLE_GLOBAL_CACHE": {"value": "false", "kind": "literal"},
+        "YARN_ENABLE_IMMUTABLE_CACHE": {"value": "false", "kind": "literal"},
+        "YARN_ENABLE_MIRROR": {"value": "true", "kind": "literal"},
+        "YARN_GLOBAL_FOLDER": {"value": "deps/yarn", "kind": "path"},
+    }
+
+    return [EnvironmentVariable(name=name, **obj) for name, obj in env_vars.items()]
--- a/tests/unit/package_managers/yarn/test_main.py
+++ b/tests/unit/package_managers/yarn/test_main.py
@@ -1,6 +1,8 @@
+import itertools
 import re
 from enum import Enum
 from itertools import zip_longest
+from pathlib import Path
 from typing import List, Optional, Union
 from unittest import mock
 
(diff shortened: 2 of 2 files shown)


brunoapimentel:
On a first review, LGTM. It would be nice to test a build with these env variables to see if it succeeds, but I'll only look at this next week :tropical_drink:

cachi2/core/package_managers/yarn/main.py
```diff
@@ -154,4 +154,11 @@ def _undo_changes(project: Project) -> None:
```
eskultety, line 154: **Commit message**:
It actually took me a while to figure out why we're setting the global cache options instead of just overriding the local cache for the build. Because of that I'd like to see some bits from the https://github.com/chmeliik/berryscary/commit/1275d17d761090fa6999000eb1991ad4e074eacc commit message here so that even after a few months we have this crucial information tracked centrally.

slimreaper35 (approved):

brunoapimentel (approved):
LGTM!

Thanks for adding that unit test.

tests/unit/package_managers/yarn/test_main.py
```diff
@@ -238,3 +261,88 @@ def test_set_yarnrc_configuration(mock_write: mock.Mock) -> None:
 def test_generate_environment_variables(yarn_env_variables: list[EnvironmentVariable]) -> None:
     result = _generate_environment_variables()
     assert result == yarn_env_variables
+
+
+@pytest.mark.parametrize(
+    "yarn_input_packages, package_components",
+    (
+        (
+            [{"type": "yarn", "path": "."}],
```
eskultety, line 270: > "yarn_input_packages,

You lost me here to be fair. It took me a while to realize this is how you perform a fixture override (I had to scroll through the docs). The most confusing part to me was that you actually return a dummy value from the `yarn_input_packages` which is actually never used because it would be always overriden, right?

I'm also missing some test IDs here so I can at least identify what the intended purpose of each test case is.

Do you think [indirect parametrization](https://docs.pytest.org/en/7.1.x/example/parametrize.html#apply-indirect-on-particular-arguments) would help here a tiny bit [1]? IOW

```
diff --git a/tests/unit/package_managers/yarn/test_main.py b/tests/unit/package_managers/yarn/test_main.py
index ca31b86c..3bba2705 100644
--- a/tests/unit/package_managers/yarn/test_main.py
+++ b/tests/unit/package_managers/yarn/test_main.py
@@ -25,8 +25,8 @@ from cachi2.core.rooted_path import RootedPath
 
 
 @pytest.fixture
-def yarn_input_packages() -> list[dict[str, str]]:
-    return [{"type": "yarn"}]
+def yarn_input_packages(request: pytest.FixtureRequest) -> list[dict[str, str]]:
+    return request.param
 
 
 @pytest.fixture
@@ -308,6 +308,7 @@ def test_generate_environment_variables(yarn_env_variables: list[EnvironmentVari
             ],
         ),
     ),
+    indirect=["yarn_input_packages"]
 )
 @mock.patch("cachi2.core.package_managers.yarn.main._resolve_yarn_project")
 @mock.patch("cachi2.core.package_managers.yarn.project.Project.from_source_dir")
```

I'd be okay if you decide to disagree with the above, since I'd propose some test cleanup at some point anyway, see my reasoning below.

[1] Our parametrize lists are massive in general, so I think long term I think we should extract these lists to constants and keep the parametrize decorator init compact (I know we have a ton of precedent for the way you're proposing). I think would this would ultimately help overall readability of the test cases because you can still know which test a given parametrize relates to which is IMO a problem currently as parametrize tends to span multiple pages.

taylormadore: The above works for me :+1: 

I had never used or seen the indirect parametrization before, so TIL :smile:

eskultety (approved):

PR #392, opened 2023-11-20
Yarn: Set 'globalFolder' in override YarnRc && drop _fetch_dependencies helper

eskultety:

- ~[ ] Docs updated (if applicable)~
- ~[ ] Docs links in the code are still valid (if docs were updated)~

### DIFF
--- a/cachi2/core/package_managers/yarn/main.py
+++ b/cachi2/core/package_managers/yarn/main.py
@@ -35,6 +35,7 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Comp
 
     :param project: the directory to be processed.
     :param output_dir: the directory where the prefetched dependencies will be placed.
+    :raises YarnCommandError: if fetching dependencies fails
     """
     log.info(f"Fetching the yarn dependencies at the subpath {output_dir.subpath_from_root}")
 
@@ -53,7 +54,7 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Comp
     try:
         _set_yarnrc_configuration(project, output_dir)
         packages = resolve_packages(project.source_dir)
-        _fetch_dependencies(project.source_dir, output_dir)
+        _fetch_dependencies(project.source_dir)
     finally:
         _undo_changes(project)
 
@@ -142,17 +143,13 @@ def _check_yarn_cache(source_dir: RootedPath) -> None:
     pass
 
 
-def _fetch_dependencies(source_dir: RootedPath, output_dir: RootedPath) -> None:
+def _fetch_dependencies(source_dir: RootedPath) -> None:
     """Fetch dependencies using 'yarn install'.
 
     :param source_dir: the directory in which the yarn command will be called.
-    :param output_dir: the directory where the yarn dependencies will be downloaded to.
     :raises YarnCommandError: if the 'yarn install' command fails.
     """
-    cachi2_output = output_dir.join_within_root("deps", "yarn")
-
-    args = ["install", "--mode", "skip-build"]
-    run_yarn_cmd(args, source_dir, {"YARN_GLOBAL_FOLDER": str(cachi2_output)})
+    run_yarn_cmd(["install", "--mode", "skip-build"], source_dir)
 
 
 def _undo_changes(project: Project) -> None:
--- a/cachi2/core/package_managers/yarn/main.py
+++ b/cachi2/core/package_managers/yarn/main.py
@@ -127,7 +127,7 @@ def _set_yarnrc_configuration(project: Project, output_dir: RootedPath) -> None:
     yarn_rc.unsafe_http_whitelist = []
     yarn_rc.enable_mirror = True
     yarn_rc.enable_scripts = False
-    yarn_rc.global_folder = str(output_dir)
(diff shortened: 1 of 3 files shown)


slimreaper35 (approved):
:+1:

brunoapimentel:
Good catch on the `YARN_GLOBAL_FOLDER`, I did not notice it there before. We should rely on `.yarnrc.yml` to set the config we need.

tests/unit/package_managers/yarn/test_main.py
```diff
@@ -103,25 +102,6 @@ def test_configure_yarn_version_fail(
...
-    source_dir = rooted_tmp_path
-    output_dir = rooted_tmp_path.join_within_root("cachi2-output")
-
-    mock_yarn_cmd.side_effect = YarnCommandError("berryscary")
-
-    with pytest.raises(YarnCommandError) as exc_info:
-        _fetch_dependencies(source_dir, output_dir)
-
-    mock_yarn_cmd.assert_called_once_with(
-        ["install", "--mode", "skip-build"],
```
brunoapimentel, line 117: One of the things this test was checking was that `run_yarn_cmd` was being called with the right parameters. I think this could be moved to `test_resolve_yarn_project`, but this test does not exist currently.

We should do a follow up PR to increase the coverage on the yarn/main module: https://coveralls.io/builds/64062886/source?filename=cachi2%2Fcore%2Fpackage_managers%2Fyarn%2Fmain.py#L51

cachi2/core/package_managers/yarn/main.py
```diff
@@ -133,19 +134,6 @@ def _check_yarn_cache(source_dir: RootedPath) -> None:
     pass
 
 
-def _fetch_dependencies(source_dir: RootedPath, output_dir: RootedPath) -> None:
-    """Fetch dependencies using 'yarn install'.
-
-    :param source_dir: the directory in which the yarn command will be called.
-    :param output_dir: the directory where the yarn dependencies will be downloaded to.
-    :raises YarnCommandError: if the 'yarn install' command fails.
```
brunoapimentel, line 141: We should probably move this "raises" docstring to the `_resolve_yarn_project` function.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -47,7 +47,8 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Comp
         if project.is_zero_installs:
             _check_yarn_cache(project.source_dir)
         else:
-            _fetch_dependencies(project.source_dir, output_dir)
+            # fetch the dependencies
```
brunoapimentel, line 50: I like that we simplified (and fixed) the code with these changes, but I don't like that we lost the context that the function `_fetch_dependencies` brought, and now we rely on a comment to convey it.

I don't have a good suggestion to fix it though, so this is just me nitpicking :shrug:

brunoapimentel:
It would be good to have this rebased, as the new changes in main affect the content of `_resolve_yarn_project`.

cachi2/core/package_managers/yarn/main.py
eskultety: I can put it back, I don't have a preference really; since it's quite common practice to have one-liner instance methods just to convey semantic meaning, despite me not liking the same for regular functions, sure, why not, I don't really have a good argument against such practice (I'm afraid unnecessary calls and stack operations wouldn't hold and would also be inconsistent with the practice for methods anyway) :)

cachi2/core/package_managers/yarn/main.py
eskultety: Since I kept `_fetch_dependencies` around what I'll do is that I'll also copy this raises docs note to `_resolve_yarn_project` as well.

tests/unit/package_managers/yarn/test_main.py
eskultety: There's a fine line where the code coverage is a nice thing and where it leads to duplicate test cases IMO. At first I didn't see your point, but now I see you're specifically interested in testing a given set of call parameters which wouldn't be tested anywhere anymore. I'll simplify the current test and keep it around instead of removing it - we can schedule it for a conversion to the `test_resolve_yarn_project` as you noted in a follow up PR. However, as I said, we need to be careful with the test cases we test, since in this particular case we're talking about higher level helpers which transitively exercise functionality of the elementary helper primitives for which we have like 99.8% coverage of various inputs, so we'd only be creating redundancy with an extensive parametrization of say `_fetch_dependencies` or future `test_resolve_yarn_project`. Anyways, for now, I'll keep things as they are :).

eskultety:
> It would be good to have this rebased, as the new changes in main affect the content of `_resolve_yarn_project`.

Yeah, this is always an issue when creating new PRs - do you base it and depend on another PR or you start fresh from main? Does one need to rebase the series frequently without a review or not? Either way you'll end up with merge conflicts one will have to solve at some point :). Unless the code becomes completely irrelevant wrt/ to the latest main I usually go with the approach to gather ideas and comments on v1 and proceed with a rebase in next revisions.

eskultety:
since v1:
- rebased on top of latest main HEAD
- simplified and kept `_fetch_dependencies` instead of removing it
- simplified and kept `test_fetch_dependencies` instead of removing it

brunoapimentel (approved):

PR #417, opened 2023-12-05
yarn: complete main integration tests

slimreaper35:


tests/integration/test_yarn.py
```diff
@@ -45,7 +45,7 @@ def test_yarn_packages(
     request: pytest.FixtureRequest,
 ) -> None:
     """
-    Test fetched dependencies for yarn 2+.
```
eskultety, line 48: nitpick: I would change this to 'v3' explicitly as who knows how many changes will be necessary for v4. If the test invocation can remain unchained, then I say we convert the string to 'berry'.

ben-alkov (dismissed):

taylormadore:
I think we'll want to extend this into becoming an end-to-end (e2e) test, where we also install the berryscary package from the cachi2 output in a separate container and run something to ensure it installed correctly and "works". A good example of this is the e2e test for gomod: https://github.com/containerbuildsystem/cachi2/blob/a3718737d65fa25cd55757d6a6f5436359f812d5/tests/integration/test_gomod.py#L239

tests/integration/test_yarn.py
brunoapimentel: I think the test invocation will remain unchanged, but we'll need to change the ref and the test output data. Another option is to leave it simply at `yarn` :man_shrugging:

brunoapimentel:
> I think we'll want to extend this into becoming an end-to-end (e2e) test, where we also install the berryscary package from the cachi2 output in a separate container and run something to ensure it installed correctly and "works". A good example of this is the e2e test for gomod:
> 
> https://github.com/containerbuildsystem/cachi2/blob/a3718737d65fa25cd55757d6a6f5436359f812d5/tests/integration/test_gomod.py#L239

Yeah, we want to do this as part of the story, but if you want to handle the e2e part in a separate PR, I'm ok with it.

brunoapimentel:
> Should I create another repo or branch or just make the regular workflow test as e2e test?

The regular workflow should be enough :+1: 

After you run `yarn build`, you can:
```
$ yarn node dist/app.js 
Hello, World!
Today is Wednesday December 6th, 2023

What is your name?
What is your quest?
What is the capital of Assyria?

First shalt thou take out the Holy Pin.
Then shalt thou count to three, no more, no less.
Three shall be the number thou shalt count, and the number of the counting shall be three.
Four shalt thou not count, neither count thou two, excepting that thou then proceed to three.
Five is right out.
Once the number three, being the third number, be reached, then lobbest thou thy Holy Hand Grenade of Antioch towards thy foe, who, being naughty in My sight, shall snuff it.
```

slimreaper35:
moved _regular installs_ as e2e test, PTAL

tests/integration/test_yarn.py
```diff
@@ -14,48 +14,88 @@
...
                 packages=({"path": ".", "type": "yarn"},),
                 check_vendor_checksums=False,
+                flags=["--dev-package-managers"],
                 expected_exit_code=0,
                 expected_output="All dependencies fetched successfully",
             ),
-            marks=pytest.mark.xfail,  # temporary
-            id="yarn_no_zero_installs",
+            [],
+            "",
```
brunoapimentel, line 68: We should make sure that the `yarn berryscary` command is executed and that the proper output is generated.

brunoapimentel: From the tests I'm running here, adding `['yarn', 'berryscarry']` as the check_cmd makes the test fail with:
```           
AssertionError: ['yarn', 'berryscary'] command failed, Output: yarn run v1.22.19
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Couldn't find a package.json file in "/tmp/yarn_e2e_test-source"
```
I didn't dig deep into it, but maybe the image is falling back to Yarn 1?

taylormadore: The default yarn version in the node container images is usually still yarn v1. We'll need to get the correct yarn version installed in the container via another method that doesn't require network access...

corepack... :harold:

chmeliik: I thought even yarn 1 would replace itself with https://github.com/cachito-testing/cachi2-yarn-berry/blob/no-zero-installs/.yarn/releases/yarn-3.6.1.cjs, so it would work :thinking:

taylormadore: It looks like it's caused by a mismatch in the mounted directory paths at different stages of the test. When the test image is built, the source directory within the container is /tmp/{test_case_name}-source. Later, when you execute a command on the built image, the source directory path has changed to whatever the tmp_path fixture is joined with {test_case_name}-source.

To see what I'm talking about, if you copy the source directory in the Containerfile after running the yarn install commands to somewhere like /src and change the WORKDIR to /src, it should magically start working

slimreaper35: Well, it works /o\

brunoapimentel: Maybe we shouldn't to mount the tmp_dir into the image when we're running it?

https://github.com/containerbuildsystem/cachi2/blob/06f77cb4b656ae0c22142e392e1d65bb41ccf33c/tests/integration/utils.py#L67

I don't see why the image should rely on the source code mounted from an external dir just to run.

tests/integration/test_data/yarn_e2e_test/container/Containerfile
```diff
@@ -0,0 +1,19 @@
...
+# Check yarn deps
+RUN ls /tmp/yarn_e2e_test-output/deps/yarn
+
+# Check content of source repository folder
+RUN ls /tmp/yarn_e2e_test-source/
+
+WORKDIR /tmp/yarn_e2e_test-source
+RUN . /tmp/cachi2.env && yarn install --mode skip-build && yarn build
+
+RUN cp -r /tmp/yarn_e2e_test-source /src
```
brunoapimentel, line 18: A small improvement suggestion here (considering I don't want to touch the integration tests helper for now).
```
FROM docker.io/node:18.19.0

# Test disabled network access
RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi

# Print cachi2 env vars file
RUN cat /tmp/cachi2.env

# Check yarn deps
RUN ls /tmp/yarn_e2e_test-output/deps/yarn

# Check content of source repository folder
RUN ls /tmp/yarn_e2e_test-source/

# This should be a COPY, but the sources and the Containerfile are in
# different directories
RUN cp -r /tmp/yarn_e2e_test-source /src

WORKDIR /src
RUN . /tmp/cachi2.env && yarn install --mode skip-build && yarn build

CMD ["yarn", "berryscary"]
```

brunoapimentel (approved):
LGTM

ejegrova (approved):

slimreaper35:
rebased

PR #424, opened 2023-12-12
package_managers: yarn: Drop the 'undo_changes' handler

eskultety:
Originally introduced by commit a18fda989 to handle erroneous paths during the dependency pre-fetch. However, the handler has always been a NOOP and since we merged b83b621b which always creates a working copy of the input source repository, we don't need any rollback handler at the moment - we can always add it when needed, but until then, let's drop this.

### DIFF
--- a/cachi2/core/package_managers/yarn/main.py
+++ b/cachi2/core/package_managers/yarn/main.py
@@ -82,12 +82,9 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Comp
             ),
         )
 
-    try:
-        _set_yarnrc_configuration(project, output_dir)
-        packages = resolve_packages(project.source_dir)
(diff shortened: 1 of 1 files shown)


taylormadore:
Not a comment on this PR, but looking at this change reminded me: do we want to add a unit test for _resolve_yarn_project?

eskultety:
> Not a comment on this PR, but looking at this change reminded me: do we want to add a unit test for _resolve_yarn_project?

I remember having had a discussion about this in the not so distant past, could it be this one:
https://github.com/containerbuildsystem/cachi2/pull/392#discussion_r1403833573

@taylormadore would you agree with what was proposed in ^this brief discussion? If so, I think you can create and issue for this.

brunoapimentel (approved):

PR #427, opened 2023-12-14
Add preliminary Go 1.21 support

eskultety:
This PR just deals with the necessary bits to get projects building with 1.21. Proper 1.21 support including support for the `toolchain` will be proposed in the future.


cachi2/core/package_managers/gomod.py
```diff
@@ -549,6 +550,7 @@ def _resolve_gomod(
         "PATH": os.environ.get("PATH", ""),
         "GOMODCACHE": "{}/pkg/mod".format(tmp_dir),
         "GOSUMDB": "sum.golang.org",
+        "GOTOOLCHAIN": "local",
```
brunoapimentel, line 553: We also need to add this variable to the expected data in the integration tests. This is what caused a lot of failures in the last CI run.

brunoapimentel: I can do this by running `make generate-test-data`. This will also bring in the new stdlib packages from go 1.21.

brunoapimentel:
We need to update the go version in the docs: https://github.com/containerbuildsystem/cachi2/blob/main/README.md?plain=1#L311

eskultety:
Converting to draft as due to https://github.com/containerbuildsystem/cachi2/discussions/419#discussioncomment-8004802 this should be proposed as POC first.

cachi2/core/package_managers/gomod.py
```diff
@@ -258,6 +264,23 @@ def release(self) -> str:  # type: ignore
         """Release name of the Go Toolchain, e.g. go1.20 ."""
         pass
 
+    @staticmethod
+    def _locate_toolchain(release: str) -> Optional[str]:
+        """Given a release locate an alternative Go toolchain.
+
+        Locate an alternative Go toolchain under the one of the following locations:
+            - /usr/local/go/                    for container environments (pre-installed)
+            - $XDG_CACHE_HOME/cachi2/go         for local environments (download & cache)
```
chmeliik, line 299: Is `/usr/local/go/` purely our custom in-container location? Or is it the default system-wide Go location?

If it's just ours, maybe the container could pre-install to the ~/.cache location instead to simplify this

chmeliik (commented):
Approach LGTM :+1: 

Didn't go into the details

cachi2/core/package_managers/gomod.py
eskultety: > Is /usr/local/go/ purely our custom in-container location? Or is it the default system-wide Go location?

The path used by package managers on most distros will default to `/usr/lib/go`. I used `/usr/local` because the purpose of that path are host-local installations which is exactly what we're doing. We must not write to the default packaging system location.

> If it's just ours, maybe the container could pre-install to the $XDG_CACHE_HOME location instead to simplify this

We can't because:

1. `$XDG_CACHE_HOME` is usually not set, well, at least not on server-like distro spins
2. `~/.cache` points to the current container user's home, which in case of our build is root, but will most likely be something else during container runtime

and so I resorted to a widely known system location instead.

cachi2/core/utils.py
```diff
@@ -111,3 +112,12 @@ def _copy_using(copy_function: Callable) -> None:
...
 
     return destination
+
+
+def get_cache_dir() -> Path:
+    """Return cachi2's global cache directory, useful for storing reusable data."""
+    try:
+        cache_dir = Path(os.environ["XDG_CACHE_HOME"])
+    except KeyError:
+        cache_dir = Path.home().joinpath(".cache")
```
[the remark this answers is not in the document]
chmeliik: Depends :grin: 

It looks like `HOME` gets set according to the user id in the container. By default, the containers run as root, in which case `HOME=/root`. But the end user can override that: https://tekton.dev/docs/pipelines/taskruns/#example-of-running-step-containers-as-a-non-root-user. In which case `HOME=/` and isn't writable for the current user

cachi2/core/package_managers/gomod.py
chmeliik: You could set `USER` and `ENV HOME` or `ENV XDG_CACHE_HOME` in the Dockerfile, but the system location is probably easier

cachi2/core/utils.py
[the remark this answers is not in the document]
chmeliik: But as long as ~/.cache/cachi2 isn't intended for anything but local usage, this will be fine

If you wanted to future-proof it, https://github.com/containerbuildsystem/cachi2/pull/427#discussion_r1457404703 might work

cachi2/core/package_managers/gomod.py
eskultety: Hmm, but doing that would mean that konflux build pipeline needs be adjusted accordingly so that one of those is explicitly set, wouldn't it? Unless we need that sort of hardening right now I'd like to keep it simple until we need to write that cache inside a container which I don't expect to be needed (hopefully ever).

chmeliik: You can set these things in the Dockerfile, I think Tekton will respect them.

A separate problem could be that the git-clone task explicitly runs as root, so if prefetch runs as non-root it might lose the ability to write to the cloned source dir. So let's keep it as it is for now :smile:

cachi2/core/package_managers/gomod.py
```diff
@@ -213,6 +211,188 @@ def to_component(self) -> Component:
...
+            # move the newly downloaded SDK from $HOME/sdk to $HOME/.cache/cachi2/go
+            sdk_download_dir = Path.home() / f"sdk/{release}"
+            cachi2_go_dest_dir = get_cache_dir() / "go" / release
+            shutil.move(sdk_download_dir, cachi2_go_dest_dir)
+
+        log.debug(f"Go {release} toolchain installed at: {cachi2_go_dest_dir}")
+        return str(cachi2_go_dest_dir / "bin/go")
+
+    def _retry(self, cmd: list[str], **kwargs: Any) -> str:
+        """Run gomod command that downloads dependencies.
```
ben-alkov, line 355: Nitpick: Is this docstring accurate? i.e. `_retry()` is used for more than just downloading dependencies....

eskultety: ACK, will adjust the docstring accordingly.

tests/integration/test_data/gomod_e2e_multiple_modules/bom.json
```diff
@@ -196,17 +196,6 @@
       "purl": "pkg:golang/internal/goarch?type=package",
       "type": "library"
     },
-    {
```
eskultety, line 199: I wonder if these changes aren't better suited to be part of https://github.com/containerbuildsystem/cachi2/pull/427/commits/6ec74a0243c56d0fee175acf1a49c6a59f10c66b rather than a standalone commit, I think that would make more sense.

brunoapimentel:
So, I'm having a small issue running this locally when I target a Go 1.18 repo:
```
2024-01-19 11:24:45,216 INFO go.mod recommends/requires Go version: 1.18
2024-01-19 11:24:45,216 INFO Downloading the gomod dependencies
2024-01-19 11:24:49,141 ERROR The command "go install golang.org/dl/go1.20@latest" failed
2024-01-19 11:24:49,141 ERROR STDERR:
go: downloading golang.org/dl v0.0.0-20240109181448-9c0c6b2b4a66
build cache is required, but could not be located: GOCACHE is not defined and neither $XDG_CACHE_HOME nor $HOME are defined
```
I could work around this by passing my local $HOME to the `run_cmd` process, so I'm guessing this is needed in local envs?

eskultety:
> So, I'm having a small issue running this locally when I target a Go 1.18 repo:
> 
> ```
> 2024-01-19 11:24:45,216 INFO go.mod recommends/requires Go version: 1.18
> 2024-01-19 11:24:45,216 INFO Downloading the gomod dependencies
> 2024-01-19 11:24:49,141 ERROR The command "go install golang.org/dl/go1.20@latest" failed
> 2024-01-19 11:24:49,141 ERROR STDERR:
> go: downloading golang.org/dl v0.0.0-20240109181448-9c0c6b2b4a66
> build cache is required, but could not be located: GOCACHE is not defined and neither $XDG_CACHE_HOME nor $HOME are defined
> ```
> 
> I could work around this by passing my local $HOME to the `run_cmd` process, so I'm guessing this is needed in local envs?

Doh! I knew about this, but forgot to address it amongst other things that had to be done here. So Go's SDK download and install requires `git` (which, duh, everyone should have), potentially `gcc` (but this should be Go's direct dependency). Lastly, Go builds some artifacts (as can be seen in the Containerfile commit which deletes them), so yeah, I'll point Go to some `GOCACHE`

brunoapimentel:
One more small detail: apparently, very old versions of `go.mod` files lack the `go {version}` line: https://github.com/release-engineering/retrodep/blob/master/go.mod

I think we can easily leave this as unsupported... but maybe it's just as easy to assume this should be handled by go 1.20.

eskultety:
> One more small detail: apparently, very old versions of `go.mod` files lack the `go {version}` line: https://github.com/release-engineering/retrodep/blob/master/go.mod
> 
> I think we can easily leave this as unsupported... but maybe it's just as easy to assume this should be handled by go 1.20.

LOL, that must be veeery old. Well, my original unposted version of the code handling this situation looked roughly like this:
```
try:
  run go list -f {{.GoVersion}} -m
except Exception:
 # if go failed to parse go.mod, falback to regex
re.match(...)
```

But then I figured versions of Go that wouldn't pass parsing with the regex have already been long EOL'd.
I can add the code back if you want, but the error reporting was funky in that case, because we'd have reported "Failed to download dependencies" by the time we get to the fallback code in  the `except` clause, so more work would be needed. 

> but maybe it's just as easy to assume this should be handled by go 1.20.

Yeah, good point, I can set the parsed version using the regex to a fixed value of the bundled Go if the regex doesn't find a match (so that we get past the version comparisons) and let the bundled Go installation (since it claims to be backwards compatible) handle it, that's a trivial fix.

cachi2/core/package_managers/gomod.py
eskultety: Changed. Also, the error messages were made more generic.

tests/integration/test_data/gomod_e2e_multiple_modules/bom.json
eskultety: Yes, they should, moved...

eskultety:
since v1:

- `Go._retry` has now a better description along with more generic error messages since we're running more than just dependency downloading commands
- `_get_gomod_version` now returns a fixed `1.20` string if a go.mod doesn't specify the `go` directive (Note however, that this way **any** newer Go toolchain will modify go.mod with its own version in such case; not a problem of cachi2)
- explicitly set `GOCACHE` with `go install` and `<gover>/bin/go download` to prevent `build cache is required, but could not be located: GOCACHE is not defined and neither $XDG_CACHE_HOME nor $HOME are defined` error

cachi2/core/package_managers/gomod.py
```diff
@@ -559,30 +758,62 @@ def _resolve_gomod(
...
+        go_mod_version_msg += " " + "(cachi2 enforced)"
+
+    go_mod_version = version.Version(_)
+
+    log.info(go_mod_version_msg.format(go_mod_version))
+
+    if go_mod_version >= go1_21 and go_base_version < go1_21:
+        # our base Go installation is too old and we need a newer one to support new keywords
+        go = Go(release="go1.21.0")
+    elif go_mod_version < go1_21 and go_base_version >= go1_21:
```
brunoapimentel, line 781: And in case `gomod_version >= go1_21 and go_base_version >= go_121`, we can just leave Go handle any specific declared toolchain, right?

I mean, imagine go_base_version is `go1.21.1` and gomod_version is `go1.21.6`. Would Go download the specified toolchain for commands such as `download`?

brunoapimentel:
Maybe it'd be good to add some unit tests to the new `Go` class to improve coverage: https://coveralls.io/builds/65216054/source?filename=cachi2%2Fcore%2Fpackage_managers%2Fgomod.py.

I think we also should bump some of our integration test-repos to Go 1.21 (or add new branches/scenarios). But we can do this in a follow up story.

eskultety:
> Maybe it'd be good to add some unit tests to the new `Go` class to improve coverage: https://coveralls.io/builds/65216054/source?filename=cachi2%2Fcore%2Fpackage_managers%2Fgomod.py.
> 
> I think we also should bump some of our integration test-repos to Go 1.21 (or add new branches/scenarios). But we can do this in a follow up story.

Sure, I will add them provided I have a green light on this approach, see https://github.com/containerbuildsystem/cachi2/pull/427#issuecomment-1898111060

cachi2/core/package_managers/gomod.py
eskultety: > And in case `gomod_version >= go1_21 and go_base_version >= go_121`, we can just leave Go handle any specific declared toolchain, right?
> 

Yes, cachi2 is just a middle man.  We can't be responsible for someone specifying `go 1.21.6` in their mod file and expecting the build will always pass as that's awfully specific and would require essentially everyone to download that specific toolchain before proceeding. My take on this is that it was never meant to work this way, it's just starting 1.21 Go now recognizes "micro" releases in version strings due to the `toolchain` directive. The `go` directive should only ever contain language version (looking at you Syft --> `1.21.0`) which guarantees that any toolchain build conforming to the 1.21 spec will build the project.

> I mean, imagine go_base_version is `go1.21.1` and gomod_version is `go1.21.6`. Would Go download the specified toolchain for commands such as `download`?

Not with `GOTOOLCHAIN=local`, in this case it'll fail with `go: go.mod requires go >= 1.21.6 (running go 1.21.4; GOTOOLCHAIN=local)`. With `GOTOOLCHAIN=auto` (future full support) will happily pass and download both the toolchain and dependencies.

eskultety: Note that I could make the clause more generic with `if go_mod_version > go_base_version` but that would essentially always download a toolchain during runtime which we likely don't want. Like I said, if anyone puts a specific toolchain version to the `go` directive, they should suffer the consequences because literally any CI would choke on that.

brunoapimentel: Humm, whereas I agree with the approach, I'm just worried because some users are asking for Go 1.21 support specifically because some projects do have the `toolchain` directive. So this would make these projects fail if the specified toolchain does not match the one being used by Cachi2?

eskultety: I'm not sure I  follow your point, `toolchain` is only informative, not required. Adn in fact, bundled versions of Go 1.21 (e.g. Fedora) ship with `GOTOOLCHAIN=local` which means that `toolchain` is going to be ignored anyway. The whole point of this preliminary support is to make it consistent by enforcing GOTOOLCHAIN so that people don't use environments, where that bundled toolchain ships with `GOTOOLCHAIN=auto`. Can you elaborate more on your point?

brunoapimentel: Sorry, I don't have all the details on how this works in my mind, so I might have gotten something wrong.

> Not with GOTOOLCHAIN=local, in this case it'll fail with go: go.mod requires go >= 1.21.6 (running go 1.21.4; GOTOOLCHAIN=local). 

I understood that, considering the above, the prefetch would fail if the `go.mod` file specifies a `toolchain` that is higher from the one being used. If that is not the case, then we're safe.

taylormadore:
This approach LGTM :+1: 

I've tried testing a few different scenarios and it appears to work well :rocket:

cachi2/core/package_managers/gomod.py
chmeliik: But if the project specifies `go 1.22` and your base go is 1.21.x, it fails immediately no? (If you set `GOTOOLCHAIN=local` that is)

Previously it would have worked to a limited extent, the older Go just wouldn't necessarily support the features that the user might be trying to use.

With full toolchain suppor, cachi2 will download the right version of Go? But won't necessarily represent everything properly in the SBOM (unknown new features)

chmeliik:
```bash
$ go mod download
go: go.mod requires go >= 1.22 (running go 1.21.6; GOTOOLCHAIN=local)
```

eskultety: > But if the project specifies `go 1.22` and your base go is 1.21.x, it fails immediately no? (If you set `GOTOOLCHAIN=local` that is)
Previously it would have worked to a limited extent, the older Go just wouldn't necessarily support the features that the user might be trying to use.

That's right, but there's no 1.22 yet and by then, we should have full support merged.

So what exactly are we discussing here really? Let's summarize:

- The base (default) 1.21 enforces usage of 1.21 as the least common denominator if Go finds any of the dependencies to specify 1.21 while your project's version < 1.21 
    - --> dirty git tree --> we fall back to 1.20
- The base version is < 1.21 (on local environments) while the project specifies >= 1.21
   - 1.20 may or may not be able to build the project, so compared to the standard behaviour where it could fail, we go the extra mile to download and install 1.21 to eliminate the potential failure
 
So, is this whole discussion about cachi2 not going the extra mile of downloading literally ANY toolchain when `base>=1.21 and gomod>base` compared to the other cases?
The question then is, is it worth adding that functionality if we already know we'll have revert it soon when we start setting `GOTOOLCHAIN=auto` and hence render that case irrelevant?

chmeliik: I think it's only a question of priority for the full toolchain support. These issues make it a higher priority than originally expected IMO

brunoapimentel:
> This approach LGTM 👍
> 
> I've tried testing a few different scenarios and it appears to work well 🚀

I agree! I have also tested a few different scenarios, and I think any further comment on this PR will be minor.

eskultety:
since v2:

- rearranged some of the commits slightly
- added unit tests for the Go instance methods
- cosmetic code changes due to the unit tests

ben-alkov (approved):

cachi2/core/package_managers/gomod.py
```diff
@@ -664,6 +664,27 @@ def _get_repository_name(source_dir: RootedPath) -> str:
...
+
+    We need to extract the desired version of Go ourselves as older versions of Go might fail
+    due to e.g. unknown keywords or unexpected format of the version (yes, Go always performs
+    validation of go.mod).
+
+    If we cannot extract a version from the 'go' line, we return None, leaving it up to the caller
+    to decide what to do next.
+    """
+    # Spawning Go as a process could be expensive, so instead of using this manual extraction only
+    # as a fallback we use it right away.
```
eskultety, line 664: This commentary is a leftover from earlier revisions that no longer applies, I'll drop it in the next revision when I'll likely have to respin.

taylormadore (approved):
LGTM :+1:

eskultety:
Since v3:
- minor cosmetic changes

eskultety:
since v4:
- silence Hadolint

eskultety:
@taylormadore @ben-alkov please give this one more quick look and if it looks good I'll proceed with the merge.

ben-alkov (approved):

taylormadore (approved):
Still LGTM :+1:

PR #426, opened 2023-12-14
cachi2 yarn: bug fixes

slimreaper35:
While implementing integration test with multiple yarn packages, I came across some bugs.

To reproduce, run integration tests on my local branch: [here](https://github.com/slimreaper35/cachi2/tree/STONEBLD-1899)

# Maintainers will complete the following section

- [x] Commit messages are descriptive enough
- [x] Code coverage from testing does not decrease and new code is covered
- [n/a] Docs updated (if applicable)
- [n/a] Docs links in the code are still valid (if docs were updated)

**Note:** if the contribution is external (not from an organization member), the CI
pipeline will not run automatically. After verifying that the CI is safe to run:

- [approve GitHub Actions workflows][approve-gh-actions] by clicking a button
- approve the Red Hat Trusted App Pipeline container build by commenting `/ok-to-test`
  (as is the [standard for Pipelines as Code][pac-running-pipeline])

[approve-gh-actions]: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
[pac-running-pipeline]: https://pipelinesascode.com/docs/guide/running/#running-the-pipelinerun

PR #428, opened 2023-12-15
cachi2 yarn: multiple packages e2e integration test

slimreaper35:
- STONEBLD-1899


brunoapimentel:
This needs to be rebased so it can get this [fix](https://github.com/containerbuildsystem/cachi2/pull/426).

slimreaper35:
testing repo is [here](https://github.com/cachito-testing/cachi2-yarn-berry/tree/multiple-packages)

ejegrova:
please add jira issue to commit,description

tests/integration/test_data/yarn_e2e_test_multiple_packages/container/Containerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN . /tmp/cachi2.env && yarn install
+
+WORKDIR /tmp/yarn_e2e_test_multiple_packages-source/second-pkg
+RUN . /tmp/cachi2.env && yarn install
+
+WORKDIR /tmp/yarn_e2e_test_multiple_packages-source/first-pkg
+RUN . /tmp/cachi2.env && yarn install
+
+RUN cp -r /tmp/yarn_e2e_test_multiple_packages-source/first-pkg /src
+WORKDIR /src
```
slimreaper35, line 25: If I comment out these two lines, the test will fail. We encounter the same problem in main (regular installs) e2e test [here](https://github.com/containerbuildsystem/cachi2/blob/c2e5f5968df6a57f59fa3e383ff876251c3e46e2/tests/integration/test_data/yarn_e2e_test/container/Containerfile#L16)

slimreaper35:
Copying here a comment from @taylormadore (from Slack)

> FYI for anyone looking at this. The story for this test suggests that the moment dependency should have an entry for both dev and non-dev in the SBOM, but that won't be the case right now due to https://issues.redhat.com/browse/STONEBLD-1900

brunoapimentel:
Maybe we could also test different versions of the same dependency and see if Cachi2 resolves it properly? It is simply changing the version for lodash or moment in one of the packages.

taylormadore:
> Maybe we could also test different versions of the same dependency and see if Cachi2 resolves it properly? It is simply changing the version for lodash or moment in one of the packages.

Do you mean add lodash to all three packages, but use a different version of lodash in one of them? Yeah, that might be worth adding.

ejegrova (approved):

brunoapimentel (approved):

PR #448, opened 2024-01-10
Update the SBOM merge script to handle the newer metadata.tools format

brunoapimentel:
There's a new version of the CycloneDX ([1.5](https://cyclonedx.org/docs/1.5/json/)) which introduced changes in the way to specify `.metadata.tools`, and marks the current format as deprecated.

**CycloneDX 1.4:**
```
{
   "metadata": {
      "tools": [
          {
             "vendor": "red hat",
             "name": "cachi2"
          }
      ]
   }
}
```

**CycloneDX 1.5:**
```
{
   "metadata": {
      "tools": {
        "components": [
            {
               "type": "application",
               "author": "red hat",
               "name": "cachi2"
            }
        ]
    }
}
```

Syft has also adopted the newer format starting from version 0.99.0. This makes the 'merge_syft_sbom.py' script to fail in case a newer Syft SBOM is used. This PR also updates the script so both formats can be handled in Syft SBOMs.

### DIFF
--- a/utils/merge_syft_sbom.py
+++ b/utils/merge_syft_sbom.py
@@ -118,6 +118,40 @@ def _get_syft_component_filter(cachi_sbom_components: list[dict[str, Any]]) -> C
     return component_is_duplicated
 
 
+def _merge_tools_metadata(syft_sbom: dict[Any, Any], cachi2_sbom: dict[Any, Any]) -> None:
+    """Merge the content of tools in the metadata section of the SBOM.
+
+    With CycloneDX 1.5, a new format for specifying tools was introduced, and the format from 1.4
+    was marked as deprecated.
+
+    This function aims to support both formats in the Syft SBOM. We're assuming the Cachi2 SBOM
+    was generated with the same version as this script, and it will be in the older format.
+    """
+    syft_tools = syft_sbom["metadata"]["tools"]
+    cachi2_tools = cachi2_sbom["metadata"]["tools"]
+
+    if isinstance(syft_tools, dict):
+        components = []
+
+        for t in cachi2_tools:
+            components.append(
+                {
+                    "author": t["vendor"],
+                    "name": t["name"],
+                    "type": "application",
+                }
+            )
+
+        syft_tools["components"].extend(components)
+    elif isinstance(syft_tools, list):
+        syft_tools.extend(cachi2_tools)
+    else:
+        raise RuntimeError(
+            "The .metadata.tools JSON key is in an unexpected format. "
+            f"Expected dict or list, got {type(syft_tools)}."
+        )
+
+
 def merge_sboms(cachi2_sbom_path: str, syft_sbom_path: str) -> str:
     """Merge Cachi2 components into the Syft SBOM while removing duplicates."""
     with open(cachi2_sbom_path) as file:
@@ -134,9 +168,7 @@ def merge_sboms(cachi2_sbom_path: str, syft_sbom_path: str) -> str:
 
     syft_sbom["components"] = filtered_syft_components + cachi2_sbom["components"]
 
-    syft_sbom.get("metadata", {}).get("tools", []).extend(
-        cachi2_sbom.get("metadata", {}).get("tools", [])
-    )
+    _merge_tools_metadata(syft_sbom, cachi2_sbom)
 
     return json.dumps(syft_sbom, indent=2)
 
--- a/tests/unit/data/sboms/merged.bom.json
+++ b/tests/unit/data/sboms/merged.bom.json
@@ -1,21 +1,26 @@
 {
+    "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
     "bomFormat": "CycloneDX",
-    "specVersion": "1.4",
-    "serialNumber": "urn:uuid:43840228-3fae-4d88-95aa-96512d2510dd",
+    "specVersion": "1.5",
+    "serialNumber": "urn:uuid:4370d1ba-7643-4579-8313-bc715da2fa90",
     "version": 1,
     "metadata": {
         "timestamp": "2023-05-03T18:19:41Z",
-        "tools": [
-            {
-                "vendor": "anchore",
-                "name": "syft",
-                "version": "0.47.0"
-            },
-            {
-                "vendor": "red hat",
-                "name": "cachi2"
-            }
-        ],
+        "tools": {
+            "components": [
+                {
+                    "type": "application",
+                    "author": "anchore",
+                    "name": "syft",
+                    "version": "0.100.0"
+                },
+                {
+                    "type": "application",
+                    "author": "red hat",
+                    "name": "cachi2"
+                }
+            ]
+        },
         "component": {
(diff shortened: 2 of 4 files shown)


brunoapimentel:
We also should create a mechanism to add the current Cachi2 version to the metadata.tools output, since that is now supported:
```
  "metadata": {
    "timestamp": "2024-01-10T09:29:06-03:00",
    "tools": {
      "components": [
        {
          "type": "application",
          "author": "anchore",
          "name": "syft",
          "version": "0.100.0"
        }
      ]
    }
  }
```

We should do it as a follow up story, though, so we don't block the release of yarn support.

utils/merge_syft_sbom.py
```diff
@@ -134,8 +134,8 @@ def merge_sboms(cachi2_sbom_path: str, syft_sbom_path: str) -> str:
 
     syft_sbom["components"] = filtered_syft_components + cachi2_sbom["components"]
 
-    syft_sbom.get("metadata", {}).get("tools", []).extend(
-        cachi2_sbom.get("metadata", {}).get("tools", [])
+    syft_sbom["metadata"]["tools"]["components"].extend(
+        cachi2_sbom["metadata"]["tools"]["components"]
```
chmeliik, line 138: Shouldn't the script be able to handle any CycloneDX version? Besides, the 1.4 style is still valid in 1.5: https://cyclonedx.org/docs/1.5/json/#metadata_tools

brunoapimentel: That's a good point. I still think Cachi2 output the 1.5 tools format since 1.4 is valid, but already deprecated. 

I can still improve the script to handle both versions if you think it's worth the effort (it will take more effort to test it).

taylormadore:
Do we need to bump the schema version we're using in the integration tests? https://github.com/containerbuildsystem/cachi2/blob/eb4cb6c9bd2dedbd8cd766c086fa16db290ffbfb/tests/integration/utils.py#L33-L35

utils/merge_syft_sbom.py
chmeliik: A simple unit test would probably be enough no?

chmeliik: Depending on how  you release the changes, you may have to bump the versions of the prefetch-dependencies task and every task which uses the merge script.

Outputting v1.5 tools metadata breaks previous versions of the buildah task (and others that use the merge script). Changing the merge script to support only v1.5 breaks previous versions of the prefetch task.

But if you release both at the same time, maybe nobody will notice and you can get away with not versioning the breaking changes

ben-alkov (approved):

brunoapimentel:
New pushes: added a commit that updates the `merge_syft_sbom.py` script.

cachi2/core/models/sbom.py
```diff
@@ -65,14 +65,17 @@ def from_package_dict(cls, package: dict[str, Any]) -> "Component":
...
     name: str
+    author: str
 
 
 class Metadata(pydantic.BaseModel):
     """Metadata field in a SBOM."""
 
-    tools: list[Tool] = [Tool(vendor="red hat", name="cachi2")]
+    tools: dict[str, list[Tool]] = {
+        "components": [Tool(type="application", author="red hat", name="cachi2")]
```
chmeliik, line 77: components is a list of `Component` per the CycloneDX spec, should we get rid of the `Tool` object?

https://cyclonedx.org/docs/1.5/json/#metadata_tools_oneOf_i0_components

utils/merge_syft_sbom.py
```diff
@@ -118,6 +118,29 @@ def component_is_duplicated(component: dict[str, Any]) -> bool:
...
 
 
+def _merge_tools_metadata(syft_sbom: dict[Any, Any], cachi2_sbom: dict[Any, Any]) -> None:
+    """Merge the content of tools in the metadata section of the SBOM.
+
+    With CycloneDX 1.5, a new format for specifying tools was introduced, and the format from 1.4
+    was marked as deprecated.
+
+    This function aims to support both formats in the Syft SBOM. We're assuming the Cachi2 SBOM
+    was generated with the same version as this script, and it will be in the newer format.
```
chmeliik, line 128: I don't think you should assume that. The SBOM is generated in the prefetch task, the merging happens in the build task. It's up to the user whether the task versions will be in sync or not.

chmeliik:
Blocks https://github.com/redhat-appstudio/build-definitions/pull/722, which blocks the latest rh-syft build

(Syft <= 0.98.0 panics when processing the source directory of Syft >= 0.99.0)

brunoapimentel:
New push completely reworks the PR:
- To avoid breaking changes, we're keeping Cachi2's SBOM in CycloneDX 1.4
- The merge script is updated so either .metadata.tools format can be handled (assuming Cachi2's SBOM will be kept in 1.4)

utils/merge_syft_sbom.py
brunoapimentel: New approach explained [here](https://github.com/containerbuildsystem/cachi2/pull/448#issuecomment-1946179822).

cachi2/core/models/sbom.py
brunoapimentel: Not needed anymore since we'll not update Cachi2's SBOM version.

utils/merge_syft_sbom.py
```diff
@@ -118,6 +118,35 @@ def component_is_duplicated(component: dict[str, Any]) -> bool:
...
+            components.append(
+                {
+                    "author": t["vendor"],
+                    "name": t["name"],
+                    "type": "application",
+                }
+            )
+
+        syft_tools["components"].extend(components)
+    elif type(syft_tools) is list:
```
chmeliik, line 146: nitpick: unnecessary condition

or missing `else: explode`

utils/merge_syft_sbom.py
```diff
@@ -118,6 +118,35 @@ def component_is_duplicated(component: dict[str, Any]) -> bool:
...
+    With CycloneDX 1.5, a new format for specifying tools was introduced, and the format from 1.4
+    was marked as deprecated.
+
+    This function aims to support both formats in the Syft SBOM. We're assuming the Cachi2 SBOM
+    was generated with the same version as this script, and it will be in the older format.
+    """
+    syft_tools = syft_sbom["metadata"]["tools"]
+    cachi2_tools = cachi2_sbom["metadata"]["tools"]
+
+    if type(syft_tools) is dict:
```
chmeliik, line 133: nitpick:
```suggestion
    if isinstance(syft_tools, dict):
```

chmeliik (approved):
LGTM with minor nitpicks

utils/merge_syft_sbom.py
brunoapimentel: Added a `else: explode`.

utils/merge_syft_sbom.py
brunoapimentel: Fixed.

ben-alkov (approved):

taylormadore (approved):

PR #452, opened 2024-01-10
Support new tool metadata format in the merge syft sbom script

brunoapimentel:
CycloneDX 1.5 changes the way to define tools in the metadata section, and marks the 1.4 way as deprecated.

Syft has adopted the newer format starting from version 0.99.0. This makes the 'merge_syft_sbom.py' script to fail in case a newer Syft SBOM is used.

This patch updates the script so both formats can be handled.

**CycloneDX 1.4:**
```
{
   "metadata": {
      "tools": [
          {
             "vendor": "red hat",
             "name": "cachi2"
          }
      ]
   }
}
```

**CycloneDX 1.5:**
```
{
   "metadata": {
      "tools": {
        "components": [
            {
               "type": "application",
               "author": "red hat",
               "name": "cachi2"
            }
        ]
    }
}
```

### DIFF
--- a/utils/merge_syft_sbom.py
+++ b/utils/merge_syft_sbom.py
@@ -118,6 +118,27 @@ def _get_syft_component_filter(cachi_sbom_components: list[dict[str, Any]]) -> C
     return component_is_duplicated
 
 
+def _add_tools_metadata(syft_sbom: dict[Any, Any]) -> None:
+    """Add Cachi2 as a tool in the metadata section of the SBOM.
+
+    With CycloneDX 1.5, a new format for specifying tools was introduced, and the format from 1.4
+    was marked as deprecated.
+
+    This function aims to support both formats. Notice that, for simplicity reasons, we're ignoring
+    the actual 'tools' key present in the Cachi2 generated SBOM, and the data following the
+    appropriate format is hard-coded in this script.
+    """
+    metadata = syft_sbom.get("metadata", {})
+    tools = metadata.get("tools", [])
+
+    if type(tools) is dict:
+        tools.get("components", []).append(
+            {"type": "application", "author": "red hat", "name": "cachi2"}
+        )
+    elif type(tools) is list:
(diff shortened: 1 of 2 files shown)


utils/merge_syft_sbom.py
```diff
@@ -118,6 +118,27 @@ def component_is_duplicated(component: dict[str, Any]) -> bool:
...
 
 
+def _add_tools_metadata(syft_sbom: dict[Any, Any]) -> None:
+    """Add Cachi2 as a tool in the metadata section of the SBOM.
+
+    With CycloneDX 1.5, a new format for specifying tools was introduced, and the format from 1.4
+    was marked as deprecated.
+
+    This function aims to support both formats. Notice that, for simplicity reasons, we're ignoring
+    the actual 'tools' key present in the Cachi2 generated SBOM, and the data following the
```
ben-alkov, line 128: Clever. Will this end up being tech debt?

brunoapimentel: Yeah... the kind that might never be paid out. We won't likely change anything about the 'tools' content until (if) we get to the point we're actually adding Cachi2's version number to it. Then this needs to be changed.

On the other hand, the reason I took the simple approach was because I was hoping we could have this fixed ASAP to unblock users from using Yarn in Konflux, but we found a simpler [way](https://github.com/redhat-appstudio/build-definitions/pull/738) to fix it. So I can actually go back and implement this the right way.

mkosiarc (approved):

brunoapimentel:
I introduced a slightly improved version of this patch in #448. It does not hard-code the data the merge script inserts into `.metadata.tools` of the Syft SBOM, but instead relies on the fact that Cachi2 will be producing CycloneDX 1.5 with the updated `.metadata.tools` format.

I think we can close this PR and use only #448 instead.

brunoapimentel:
Closing this in favor of #448

PR #457, opened 2024-01-24
Add support for go workspaces

ejegrova:
JIRA: STONEBLD-2043


cachi2/core/package_managers/gomod.py
```diff
@@ -582,7 +582,37 @@ def _resolve_gomod(
         # Make Go ignore the vendor dir even if there is one
         go_list.extend(["-mod", "readonly"])
 
-    main_module_name = _run_gomod_cmd([*go_list, "-m"], run_params).rstrip()
+    go_list_modules = _run_gomod_cmd([*go_list, "-m", "-json"], run_params).rstrip()
+    module_list = list(load_json_stream(go_list_modules))
+    raw_main_module = module_list.pop(0)
```
eskultety, line 587: I'm curious,  why is this named a raw module? How is it different from any other module in the list?

cachi2/core/package_managers/gomod.py
```diff
@@ -582,7 +582,37 @@ def _resolve_gomod(
...
+    if len(module_list) > 0:
+        for module in module_list:
+            workspace_modules.append(
+                ParsedModule(
+                    path=module["Path"],
+                    version=version_resolver.get_golang_version(main_module_name, app_dir),
+                    replace=ParsedModule(
+                        path=get_relative_path_workspace_to_main_module(
+                            Path(raw_main_module["Dir"]), Path(module["Dir"])
+                        )
```
eskultety, line 611: I'd instantiate the replacement ParsedModule separately in a variable to enhance readability a bit.

cachi2/core/package_managers/gomod.py
```diff
@@ -582,7 +582,37 @@ def _resolve_gomod(
...
 
-    main_module_name = _run_gomod_cmd([*go_list, "-m"], run_params).rstrip()
+    go_list_modules = _run_gomod_cmd([*go_list, "-m", "-json"], run_params).rstrip()
+    module_list = list(load_json_stream(go_list_modules))
+    raw_main_module = module_list.pop(0)
+    main_module_name = raw_main_module["Path"]
+
+    def get_relative_path_workspace_to_main_module(path1: Path, path2: Path) -> str:
+        """
+        Return relative path from workspace module to main module.
```
eskultety, line 592: Are relative paths fine wrt/ SBOMs? Don't we need to canonicalize the resulting path?

eskultety (commented):
Since this is still a draft I only had some immediate comments on the proposal but didn't do a detailed in-depth review of the logic fitting the overall design.

cachi2/core/package_managers/gomod.py
ejegrova: There could probably be some better name. Later on there is `main_module` which is `ParsedModule()`, but these data that are `dict` type correspond to main module data.

cachi2/core/package_managers/gomod.py
ejegrova: Proper path is determined [here](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/gomod.py#L468), that's why we create workspace module as `ParsedModule` with `replace`.

cachi2/core/package_managers/gomod.py
```diff
@@ -582,7 +582,37 @@ def _resolve_gomod(
...
+        """
+        if not path2.is_relative_to(app_dir.root):
+            raise Exception(
+                f"Workspace module path {path2} is not relative to root path {app_dir.root}."
+            )
+        relative_path = os.path.relpath(path2, path1)
```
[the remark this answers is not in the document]
ejegrova: The corresponding function that could be used is [relative_to](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) but I'd need to use the walk_up parameter for it to function properly and it's available in python >=3.12. So I would rather use os.path

cachi2/core/package_managers/gomod.py
```diff
@@ -811,7 +816,49 @@ def _resolve_gomod(
...
+    modules_in_go_sum = _parse_go_sum_files(app_dir, module_list_paths)
+
+    def get_relative_path_workspace_to_main_module(path1: Path, path2: Path) -> str:
+        """Return relative path from workspace module to main module."""
+        if not path2.is_relative_to(app_dir.root):
+            raise PackageManagerError(
+                f"Workspace module path {path2} is not relative to root path {app_dir.root}."
+            )
+        relative_path = os.path.relpath(path2, path1)
+        return relative_path
```
brunoapimentel, line 841: I believe we need to return `f"./{relative_path}"`. The reason for that is because the function that will convert this to a `Module` later on uses the dot to [identify](https://github.com/containerbuildsystem/cachi2/blob/d654bf3b763d56783d9bee6d65eaea9fbae58b07/cachi2/core/package_managers/gomod.py#L447) it's a local path.

You can try to reproduce the error by trying to prefetch this:

```
repo: https://github.com/kubernetes/ingress-nginx
ref: 3c4e78e6b755eb33821c4477106f424bd8792d8f
path: .
```

And then check the purl on the `k8s.io/ingress-nginx/magefiles` module. You'll see the version is `@magefiles`.

cachi2/core/package_managers/gomod.py
```diff
@@ -734,7 +740,6 @@ def _resolve_gomod(
     :raises PackageManagerError: if fetching dependencies fails
     """
     _protect_against_symlinks(app_dir)
-    modules_in_go_sum = _parse_go_sum(app_dir)
```
brunoapimentel, line 737: Moving the parsing of the go.sum(s) to after downloading the dependencis is causing us a problem. It seems that Go has a tendency of adding missing hashes to `go.work.sum` when we invoke `go mod download`.

I did a few tests with our test [repo](https://github.com/ejegrova/cachi2-go-workspaces), and removing the hashes from a single dependency from both the `api/go.sum` and `gin/go.sum` folder resulted in that hash being instead added to `go.work.sum` :shrug:.

But keeping the parsing of the go.sum files before downloading seems to solve the issue (we might want to do further tests after we fix it).

ejegrova: Moving the parsing of the go.sum files before downloading fixes the issue but creates different issues because we need to have workspace modules first to parse the sum files. The current approach of using cmd `go list -e -m -json` doesn't work well when vendoring flags are used. 

@brunoapimentel and @taylormadore  came with list of solutions:
- parse the `go.work` file manually to figure out where the workspaces are
- keep track of the `go.work.sum` file and, if it changed, fail the request (it [seems](https://go.dev/ref/mod#go-work-file) that Go adds missing checksums to the go.work.sum file instead of changing the workpaces go.sums)

Solutions that might not work:
- go list only if there are workspaces
    - vendoring for individual modules work even if there are workspaces
- handle vendoring first
    - calling go mod vendor might change the go.sum files
- handle the go list error individually (works, but I think it's a bad solution)
    - would affect all requests, regardless of workspaces being involved
    - would prevent users from fixing their vendoring with the `gomod-vendor` flag (as it would fail before that part of the code arrived) (edited) 
    
   @eskultety @ben-alkov @slimreaper35 any thoughts?

ejegrova: Current preference is the first solution - parse `go.work` file, since the second one limits the user more.

chmeliik: Just FYI, you can use `go work edit -json` to parse the file for you

eskultety: > Just FYI, you can use go work edit -json to parse the file for you

+1, but just beware of the toolchain version you use to parse the file, because Go < 1.21 won't parse the new toolchain version format from either `go.mod` or `go.work` (let alone the `toolchain` keyword):

```
cachi2-go-workspaces/go.work:1: invalid go version '1.21.0': must match format 1.23
cachi2-go-workspaces/go.work:3: unknown directive: toolchain
```
That's why I resorted in parsing go.mod manually. However, once we further tweak the toolchain selection (part of #500) we can use go's native mechanisms to parse either file and work solely with a dictionary representation.

cachi2/core/package_managers/gomod.py
```diff
@@ -892,13 +901,68 @@ def go_list_deps(pattern: Literal["./...", "all"]) -> Iterator[ParsedPackage]:
...
+    case a go.work file is present in the repository.
+
+    :param app_dir: the path to the module directory
+    :param modules_json_stream: the json stream returned by "go list -m -json"
+    :return: A tuple containing the main module and a list of workspaces
+    """
+    module_list = []
+    main_module = None
+
+    for i, module in enumerate(load_json_stream(modules_json_stream)):
```
ben-alkov, line 937: Why `enumerate` if `i` is unused?

cachi2/core/package_managers/gomod.py
```diff
@@ -860,13 +860,22 @@ def _resolve_gomod(
...
+    modules_json_stream = go([*go_list, "-m", "-json"], run_params).rstrip()
+    main_module_dict, workspace_dict_list = _process_modules_json_stream(
+        app_dir, modules_json_stream
+    )
+
     main_module = ParsedModule(
-        path=main_module_name,
-        version=version_resolver.get_golang_version(main_module_name, app_dir),
+        path=main_module_dict["Path"],
+        version=version_resolver.get_golang_version(main_module_dict["Path"], app_dir),
```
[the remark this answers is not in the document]
ejegrova: fixed

cachi2/core/package_managers/gomod.py
ejegrova: removed, that was a leftover after refactoring

ejegrova:
changed integration test repo to include checks for missing_hash property

tests/unit/package_managers/test_gomod.py
```diff
@@ -665,8 +798,107 @@ def test_parse_workspace_modules(
...
+            ./app
+            ./workspace
+        )
+        """
+    )
+    tmp_path.joinpath("go.work").write_text(go_work_file)
+
+    repo_root = RootedPath(tmp_path)
+
+    go_work_root = _get_go_work_root(repo_root, Go(), {"cwd": repo_root})
```
taylormadore, line 860: Is this calling go directly? Should we be mocking that?

taylormadore (commented):
Functionally, it looks good to me :rocket:  Hopefully my knowledge of go workspaces is sufficient for that statement to mean anything :harold:

Left a couple of minor nitpicks. Also, can the packages with missing checksums be noted in the README of the integration test repo+branch?

cachi2/core/package_managers/gomod.py
```diff
@@ -452,7 +455,11 @@ def _create_module(module: ParsedModule) -> Module:
             real_path = name
 
             if mod_id not in modules_in_go_sum:
-                missing_hash_in_file = main_module_dir.subpath_from_root / "go.sum"
+                if go_work_root:
+                    missing_hash_in_file = go_work_root.subpath_from_root / "go.work.sum"
+                else:
+                    missing_hash_in_file = main_module_dir.subpath_from_root / "go.sum"
```
eskultety, line 461: So we merge all sum files together and only detect if a checksum is missing but we don't track where exactly. Not ideal, I imagine we'd improve upon this in the future, but maybe from SBOM's POV it's going to be just fine this way, no action needed here, just thinking out loud.

tests/integration/test_gomod.py
```diff
@@ -170,6 +170,18 @@
             ),
             id="gomod_multiple_modules_missing_checksums",
         ),
+        # Test case checks if cachi2 can process go workspaces properly.
+        pytest.param(
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-gomod.git",
```
eskultety, line 176: looking at the repo the workspace cases don't seem to include a nuance of a main module i.e. having a `go.mod` file checked in - shouldn't we change that and make sure the integration test repos are proper Go projects?

eskultety (commented):
Looking good, I learnt something new during the review, e.g. that any repo-local submodule still needs to be declared uniquely within all Go modules and so it is imperative one still uses the `replace` keyword to denote that the local implementation should be used even though that is the only one!
Anyway, I think we need some documentation update on workspaces, don't we?

cachi2/core/package_managers/gomod.py
brunoapimentel: I think keeping track of which checksums belong to each file is not too hard, but the problem is to know to which sum file a dependency should belong to in the first place. 

We get the list of resolved modules from a union of the output of `go mod download` and `go list -deps`, so we don't have an easy way to tell which module imported that dependency directly  (I'm guessing that should be the way to correlate it to a specific `go.sum` file, but even Go seems to place whatever it doesn't find in the individual `go.sum` files into the more generic `go.work.sum`, so :shrug:)

cachi2/core/package_managers/gomod.py
```diff
@@ -878,27 +903,125 @@ def go_list_deps(pattern: Literal["./...", "all"]) -> Iterator[ParsedPackage]:
...
-    return ResolvedGoModule(main_module, all_modules, all_packages, modules_in_go_sum)
+    return ResolvedGoModule(main_module, all_modules, all_packages, modules_in_go_sum, go_work_root)
+
+
+ModuleDict = dict[str, Any]
+
+
+def _process_modules_json_stream(
+    app_dir: RootedPath, modules_json_stream: str
+) -> tuple[ModuleDict, list[ModuleDict]]:
```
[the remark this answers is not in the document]
brunoapimentel: IMO, that will make the output a little less clear, since there's no way to tell that the first module is always the main module unless you look at this function's docstring or implementation. By returning a tuple, it is at least implicit that the first tuple element is unique somehow when compared to the elements in the list.

tests/unit/package_managers/test_gomod.py
```diff
@@ -373,6 +385,18 @@ def test_resolve_gomod_no_deps(
...
+        """
+        {
+            "Path": "github.com/release-engineering/retrodep/v2",
+            "Main": true,
+            "Dir": "{repo_dir}",
+            "GoMod": "{repo_dir}/go.mod",
+            "GoVersion": "1.19"
+        }
+        """
+    ).replace("{repo_dir}", str(module_path))
```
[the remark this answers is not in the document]
brunoapimentel: I tried the f-string approach, and the result was very ugly. I think string.Template gives us a slightly better result than what is currently there.

cachi2/core/package_managers/gomod.py
```diff
@@ -860,13 +860,25 @@ def _resolve_gomod(
         # Make Go ignore the vendor dir even if there is one
         go_list.extend(["-mod", "readonly"])
 
-    main_module_name = go([*go_list, "-m"], run_params).rstrip()
+    modules_json_stream = go([*go_list, "-m", "-json"], run_params).rstrip()
+    main_module_dict, workspace_dict_list = _process_modules_json_stream(
+        app_dir, modules_json_stream
+    )
+
+    path = main_module_dict["Path"]
```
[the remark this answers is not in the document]
brunoapimentel: I completely agree this is a better approach, but since this is refactoring work, can we do it as a follow up?

tests/integration/test_gomod.py
brunoapimentel: This case does have a `go.mod` file, it's just buried very deep within: https://github.com/cachito-testing/cachi2-gomod/tree/go_workspaces/workspace_modules/hello

cachi2/core/package_managers/gomod.py
```diff
@@ -878,27 +903,125 @@ def go_list_deps(pattern: Literal["./...", "all"]) -> Iterator[ParsedPackage]:
...
+    module_list = []
+    main_module = None
+
+    for module in load_json_stream(modules_json_stream):
+        if module["Dir"] == str(app_dir):
+            main_module = module
+        else:
+            module_list.append(module)
+
+    # should never happen, since the main module will always be a part of the json stream
```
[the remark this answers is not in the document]
brunoapimentel: I can currently see 3 branches that are related to workspaces: `nested_main_module`, `worspace_separate_modules` and `go_workspaces`. I believe the first two ones are leftovers that should be deleted. All of these were created as we evolved the code to try to cover more corner cases. 

The `go_workspaces` branch now covers the most weird scenario: the root repo is not a go project, and the workspaces root is the first nested directory (`worspace_modules`).  I am not sure we need such a complicated case in the integration test, but the bright side is that we're covering the worst we could come up with.

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
ejegrova: added a comment

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
eskultety: That is IMO an inferior approach from design POV. If an element is special it needs to be reflected in the docs instead of trying to figure out a creative way how such a semantics needs to be represented in a data type.

tests/unit/package_managers/test_gomod.py
```diff
@@ -517,6 +540,131 @@ def test_parse_broken_go_sum(rooted_tmp_path: RootedPath, caplog: pytest.LogCapt
...
 
+@pytest.mark.parametrize(
+    "project_path, stream, expected_modules",
+    (
+        (
+            "/home/cachito/simple-project",
+            dedent(
+                """
+                {
+                    "Path": "github.com/cachito-testing/simple-project",
```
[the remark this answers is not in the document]
ejegrova: changed references to `github.com/my-org/project-with-workspaces`, etc.

tests/unit/package_managers/test_gomod.py
```diff
@@ -665,8 +798,107 @@ def test_parse_workspace_modules(
...
+    tmp_path.joinpath("go.work").write_text(go_work_file)
+
+    repo_root = RootedPath(tmp_path)
+
+    go_work_root = _get_go_work_root(repo_root, Go(), {"cwd": repo_root})
+
+    assert go_work_root == repo_root
+
+
+def test_get_go_work_root_when_go_work_does_not_exist(tmp_path: Path) -> None:
```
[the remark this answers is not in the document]
ejegrova: done

cachi2/core/package_managers/gomod.py
eskultety: I suppose we'd have to do the heavy lifting ourselves and parse all `go.mod` files, correct? Anyway it seems lie a premature optimization and overcomplicating the matter given it's not necessary atm.

tests/integration/test_gomod.py
eskultety: I meant the main go.mod module, should a project, even if using workspaces, have a `go.mod` file right next to the `go.work` file?

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
eskultety: I'd be afraid it would remain TODO forever :), but okay, UNLESS it turns out more substantial changes are needed within this PR.

eskultety:
My question is this: what is the main problem of letting users point cachi2 to the `go.work` file directly the same way as we expect them to do with `go.mod` for the main module to be processed? That would lead to a consistent behaviour from UX perspective where we'd have to process all dependent modules. What are the drawbacks of that? @brunoapimentel do you see a use case where ^this wouldn't work and simply fail with cachi2 logic?

tests/unit/package_managers/test_gomod.py
```diff
@@ -244,6 +248,136 @@ def test_resolve_gomod(
...
+    # Mock the "subprocess.run" calls
+    run_side_effects = []
+
+    run_side_effects.append(
+        proc_mock(
+            "go work edit -json",
+            returncode=0,
+            stdout=get_mocked_data(data_dir, "workspaces/go_work.json"),
+        )
+    )
```
eskultety, line 288: Apart from this particular hunk, this test function is pretty much bit-for-bit identical to the existing `test_resolve_gomod` function. Given that that's the case, we should better mock `_get_go_sum_files` and add this as another parametrize case to the existing test case, this is simply a beast of a test testing a beast of a function which leads to enormous code redundancy.

tests/unit/package_managers/test_gomod.py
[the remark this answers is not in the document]
brunoapimentel: I have changed the previous occurrences, let me know if you're ok with the result.

tests/unit/package_managers/test_gomod.py
```diff
@@ -32,6 +32,8 @@
     _create_modules_from_parsed_data,
```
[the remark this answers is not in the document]
brunoapimentel: After the last changes, this is no longer happening, and I double checked each commit to be sure the unit tests pass.

tests/unit/package_managers/test_gomod.py
```diff
@@ -262,7 +271,9 @@ def test_resolve_gomod_vendor_dependencies(
         proc_mock(
             "go list -e -m",
             returncode=0,
-            stdout="github.com/cachito-testing/gomod-pandemonium",
+            stdout=get_mocked_data(data_dir, "non-vendored/go_list_modules.json").replace(
```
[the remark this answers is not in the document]
brunoapimentel: To use `Template` here, we'd need to either subclass it so it would understand curly brackets as placeholders, or change the script that generates all the data for unit tests to use `$` as a template delimiter. I think replace is a easier solution here. Are you ok if we just leave it this way?

tests/unit/package_managers/test_gomod.py
```diff
@@ -517,6 +540,131 @@ def test_parse_broken_go_sum(rooted_tmp_path: RootedPath, caplog: pytest.LogCapt
...
+    app_dir = RootedPath(project_path)
+    result = _process_modules_json_stream(app_dir, stream)
+
+    assert result == expected_modules
+
+
+@pytest.mark.parametrize(
+    "relative_app_dir, module, expected_module",
+    (
+        # main module is also the workspace root:
```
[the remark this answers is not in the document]
brunoapimentel: I changed all the newly introduced tests that use params so that they'd have proper ids.

brunoapimentel:
> My question is this: what is the main problem of letting users point cachi2 to the `go.work` file directly the same way as we expect them to do with `go.mod` for the main module to be processed? That would lead to a consistent behaviour from UX perspective where we'd have to process all dependent modules. What are the drawbacks of that? @brunoapimentel do you see a use case where ^this wouldn't work and simply fail with cachi2 logic?

Erik and I had a private chat about this, and we figured out that `go` commands are not affected by which directory they're called in, as long as they're under the directory listed by `go env GOWORK`.

@eskultety I played with the idea a little bit. The only problem we have is if the directory that contains the `go.work` file is not a go module (i.e. does not have a `go.mod` file). Initially, I was trying to work around the need to have a "main module" (the one passed as the input package path in the CLI), but the whole gomod code is based around that idea, so this led to a lot of changes. I think this could work, and maybe the code could even be improved by that, but to do it right, it would be a major change to the gomod code.

So I tried our second idea, that was electing the first workspace listed in `go.work` as the main module, which is a much simpler approach. You can see the result in the new commit I added, and with it, I could fully process a [repository](https://github.com/ejegrova/cachi2-go-workspaces) just by pointing to its root folder. I still need to double check if the fact of picking a main module can have any impact on local replacements/workspaces, since Cachi2 uses the main module's version and path to fill data on those. But it is doable in case we want to allow the user to just point to `go.work`.

tests/integration/test_gomod.py
eskultety: > I meant the main go.mod module, should a project, even if using workspaces, have a go.mod file right next to the go.work file?

The answer is no.

brunoapimentel:
> My question is this: what is the main problem of letting users point cachi2 to the `go.work` file directly the same way as we expect them to do with `go.mod` for the main module to be processed?

The conclusion we reached about this suggestion is that, although we want to provide users more flexibility by allowing them to point to a directory that contains a `go.work` file but is not a go module, such change would require a major refactoring in the code that we don't think is worth pursuing as part of this PR.

The main problem that we reached is how to work around the need of having what Cachi2 defines as a "main module", since for each of the input packages fed to the CLI, a call to `_resolve_gomod` is made where the package is defined as the "main module" for purposes of the resolve algorithm. Here's a small (and likely incomplete) list of impacted functions that would need to be changed:

- relevant files (such as go.mod and go.sum) are checked in the main module directory for symlinks that point to outside of the cloned repository
- all go commands are executed in the main module directory
- vendored deps are checked in the `vendor` folder in the main module directory
- the main module name and its path are used to identify it's version (as a version tag can be applied to a repo subpath, and also the module can be nested under a "vN" folder, where N corresponds to the major version of that module)
- the main module path is used to identify the relative path of each workspace and local replacements, which will end up composing the purls in the SBOM
- the main module path is used to verify if a local replacement resolves to a folder inside the repo

I personally believe that all of those can be worked around, and the code will likely end up cleaner after the refactoring. So we can tackle this as a follow up task. @eskultety Please add any details I might've missed.

tests/unit/package_managers/test_gomod.py
brunoapimentel: I lost track of where exactly this change was, but I believe we have changed it to use a mocked `Go._run`.

tests/unit/package_managers/test_gomod.py
brunoapimentel: I merged the code from both tests. I needed to add some conditionals, so the test code is less readable, but probably better than the duplication we had before.

Btw, we have three boolean parameters for this test, but I don't think we need to test all combinations possible, as they shouldn't affect each other, specially considering this is a unit test and the go calls are mocked. So I added a new scenario that we have only the workspaces parameter as true, and avoided increasing the number of scenarios to 9.

cachi2/core/package_managers/gomod.py
```diff
@@ -103,6 +104,7 @@ class ResolvedGoModule(NamedTuple):
     parsed_modules: Iterable[ParsedModule]
     parsed_packages: Iterable[ParsedPackage]
     modules_in_go_sum: frozenset["ModuleID"]
+    go_work_root: Optional[RootedPath] = None
```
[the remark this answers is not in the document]
brunoapimentel: Done.

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
brunoapimentel: So, I went back and implemented this, and the code looks better now. 

I'm now wondering if the workspaces scenario that we added to `test_resolve_gomod` is worth it at all. It brings more complexity to the test and a buch of new files committed to the repo, and I think the only coverage it brings us is this line [here](https://github.com/ejegrova/cachi2/blob/workspaces/cachi2/core/package_managers/gomod.py#L460). All the rest is tested individually, AFAICT.

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
brunoapimentel: My reasoning is that we seldom read a function's docstring, and it is even less likely to be the first thing we do. By separating the results, we at least stablish that the objects returned are not the same from a business logic perspective, even if they belong from the same class.

But this is subjective, so unless you feel very strongly we should not use this design, I'd rather leave it as is. Btw, this is also how the newly introduced `_parse_local_modules` function is designed :angel:

tests/unit/package_managers/test_gomod.py
```diff
@@ -509,18 +582,274 @@ def test_parse_broken_go_sum(rooted_tmp_path: RootedPath, caplog: pytest.LogCapt
...
+    }
+    {
+        "Path": "myorg.com/my-project/workspace",
+        "Main": true,
+        "Dir": "/path/to/project/workspace"
+    }
+    """
+
+    app_dir = RootedPath("/path/to/project")
+    version_resolver.get_golang_version = lambda _, __: "1.0.0"
```
brunoapimentel, line 611: Any better way to mock the result of this function?

eskultety:
```
diff --git a/tests/unit/package_managers/test_gomod.py b/tests/unit/package_managers/test_gomod.py
index 5f617d33..7362f851 100644
--- a/tests/unit/package_managers/test_gomod.py
+++ b/tests/unit/package_managers/test_gomod.py
@@ -608,7 +608,7 @@ def test_parse_local_modules(go: mock.Mock, version_resolver: mock.Mock) -> None
     """
 
     app_dir = RootedPath("/path/to/project")
-    version_resolver.get_golang_version = lambda _, __: "1.0.0"
+    version_resolver.get_golang_version.return_value = "1.0.0"
 
     main_module, workspace_modules = _parse_local_modules(go, [], {}, app_dir, version_resolver)
```

eskultety (approved):
@brunoapimentel I think we're good. A couple of trivial things, but I trust you on fixing them :).

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
eskultety: > I'm now wondering if the workspaces scenario that we added to test_resolve_gomod is worth it at all. It brings more complexity to the test and a buch of new files committed to the repo, and I think the only coverage it brings us is this line [here](https://github.com/ejegrova/cachi2/blob/workspaces/cachi2/core/package_managers/gomod.py#L460)

You're sure about the coverage? `test_resolve_gomod` has nothing to do with `_create_modules_from_parsed_data` - that one is only ever called from `fetch_gomod_source` which is out of context for `test_resolve_gomod`.

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
eskultety: > Btw, this is also how the newly introduced _parse_local_modules function is designed 👼

The fact that both are introduced as part of this PR doesn't really make it any more correct :).
However, based on our recent discussion and more specifically based on this paragraph:

> The main problem that we reached is how to work around the need of having what Cachi2 defines as a "main module", since for each of the input packages fed to the CLI, a call to _resolve_gomod is made where the package is defined as the "main module" for purposes of the resolve algorithm. If the input package points to a folder that is not a valid go module, than we can't use it as the "main module".

I think we can keep it as is, since there's already the possibility that the main module concept may need to be revised in the future in which case many functions will have to be reworked, including this one. Despite still being in disagreement with the function signature, it should not be a show-stopper for this work.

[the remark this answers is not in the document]
chmeliik: Mostly agree with Erik here, calling it a main module is false advertising. You're picking one arbitrary workspace module and calling it "main" (which probably has consequences for detecting the versions of the other workspace modules and other locally replaced modules)

But proper version detection would probably need bigger changes, this is a good enough start

cachi2/core/package_managers/gomod.py
```diff
@@ -892,13 +891,101 @@ def go_list_deps(pattern: Literal["./...", "all"]) -> Iterator[ParsedPackage]:
...
+    # the same way as a locally replaced module when converting it into a Module.
+    if not relative_dir.startswith("."):
+        relative_dir = f"./{relative_dir}"
+
+    replaced_module = ParsedModule(path=relative_dir)
+
+    return ParsedModule(
+        path=module["Path"],
+        version=main_module_version,
+        replace=replaced_module,
```
chmeliik, line 994: By setting `replace` and giving the `replaced_module` a relative path, the version eventually ends up getting detected by `version_resolver.get_golang_version(workspace_module_name, workspace_module_path)`?

(It's a little confusing, but should be correct. Maybe just don't set `version=main_module_version` to reduce confusion)

tests/integration/test_gomod.py
```diff
@@ -170,6 +170,18 @@
...
+                ref="449f904ecf268856ef335b2033cb64cc417b7c26",
+                packages=({"path": "./workspace_modules/hello", "type": "gomod"},),
```
chmeliik, line 178: I think we've already talked about this being the wrong interface
* the user shouldn't have to point to one of their workspace modules, they should be allowed to point to the workspace root
* cachi2 shouldn't treat one of the workspaces as special just because the user pointed to it

But if the version detection is not broken because of it, this is a good enough start

chmeliik (commented):
Assuming that version detection works when the workspace modules are tagged with different version tags, this LGTM

tests/integration/test_gomod.py
eskultety: > I think we've already talked about this being the wrong interface
> 
>     * the user shouldn't have to point to one of their workspace modules, they should be allowed to point to the workspace root
> 
>     * cachi2 shouldn't treat one of the workspaces as special just because the user pointed to it
> 
> 

Yeah, it was determined that from the CLI perspective nothing changes and the user is expected to point to the module they want to build (let's call it the "main" one) and so what you describe here requires a significant overhaul of the module, so based on that I accepted this as the working solution.

cachi2/core/package_managers/gomod.py
chmeliik: Tested locally, it does work. Would be worth adding to the integration test (can be a follow-up)

```bash
git clone https://github.com/cachito-testing/cachi2-gomod
cd cachi2-gomod

git tag workspace_modules/hello/v1.1.0
git tag workspace_modules/hi/hiii/v1.2.0
git tag workspace_modules/echo/v4.2.0

cachi2 --log-level debug fetch-deps '{"type": "gomod", "path": "workspace_modules/hello"}'
jq < cachi2-output/bom.json '.components[].purl | select(test("workspace_modules.*type=module"))' -r
```

```text
pkg:golang/github.com/cachito-testing/cachi2-gomod/workspace_modules/echo@v4.2.0?type=module
pkg:golang/github.com/cachito-testing/cachi2-gomod/workspace_modules/hello@v1.1.0?type=module
pkg:golang/github.com/cachito-testing/cachi2-gomod/workspace_modules/hi/hiii@v1.2.0?type=module
```

chmeliik (approved):

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
brunoapimentel: Sorry, I'm stupid, I tagged the wrong line. [This](https://github.com/ejegrova/cachi2/blob/workspaces/cachi2/core/package_managers/gomod.py#L856) is the right one.

cachi2/core/package_managers/gomod.py
brunoapimentel: I'll implement the integration test as a follow up, then.

PR #467, opened 2024-02-15
toil(pip): remove dead code

ben-alkov:
Refactor and unify pip sdist & wheel handling (STONEBLD-1705), to wit:

- wheels are enumerated, saved, and downloaded all at once asynchronously; sdists are downloaded one by one
- wheels are ignored if they already exist; sdists are re-downloaded
- wheels respect the hashes reported by PyPI; sdists do not
- sdists are recorded in the SBOM; wheels are not

## Eliminate the differences as much as possible


ben-alkov:
I've identified an issue where existing artifacts (i.e. things which have already been downloaded) are not being hashed (if required). I haven't resolved this yet.

slimreaper35 (commented):
Currently, we use artifacts and distribution packages / distribution package info. Maybe we should stick to just one name.

cachi2/core/package_managers/pip.py
```diff
@@ -1724,32 +1772,47 @@ def download_info(self) -> dict[str, Any]:
...
+
+    :param requirement: which pip package to process
+    :param str pip_deps_dir:
+    :param bool allow_binary: process wheels?
+    :return: a single DistributionPackageInfo, or a list of DPI
+    :rtype: DistributionPackageInfo
+    """
+    allowed_distros = ["sdist", "wheel"] if allow_binary else ["sdist"]
+    artifacts: list[DistributionPackageInfo] = []
+    best_sdist: list[DistributionPackageInfo] = []
```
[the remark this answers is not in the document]
eskultety: due to the return statement, however, it can be tweaked.

cachi2/core/package_managers/pip.py
```diff
@@ -1841,14 +1868,13 @@ def _process_package_distributions(
...
                 docs = PIP_NO_SDIST_DOC
-
             raise PackageRejected(
                 f"No distributions found for package {name}=={version}",
                 solution=solution,
                 docs=docs,
             )
 
-    return best_sdist, wheels
+    return best_sdist + wheels
```
eskultety, line 1877: I think `[best_sdist] + wheels` is perfectly acceptable here.

eskultety (commented):
@ben-alkov allegedly, I'm missing some deeper understanding of pip deps handling , but overall this looks good, some polishing is required. That said I do like the end result which significantly increased readability of `_download_dependencies`.

cachi2/core/package_managers/pip.py
```diff
@@ -1403,71 +1421,102 @@ def _download_dependencies(
...
+                    continue
+                else:
+                    to_download.append(artifact)
+        return to_download, to_check
+
+    def _hash_verify(
+        download_info: dict[str, Any], dpi: Optional[DistributionPackageInfo] = None
+    ) -> dict[str, Any]:
+        if dpi:
+            download_info = dpi.download_info
```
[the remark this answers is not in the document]
slimreaper35: In addition, I would keep all three helper functions out of the `_download_dependencies` and make it more concise. It has about 150 lines of code. Why should be something like hash verification so coupled with one function ?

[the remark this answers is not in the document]
brunoapimentel: The one advantage I see in keeping them in `_download_dependencies` is to make it clear they're only inteded to be used within the same function's scope. For instance, I think the generic/reusable part of `_hash_verify` is `must_match_any_checksum`, the rest is only logic local to the download function.

But I agree it makes the function look way too long.  Maybe it'd be a case to split this into a separate module? (although it's not the approach we're taking with pip, and I don't suggest we do it in this PR)

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
ben-alkov: @eskultety;

> it can be tweaked

Explain?

cachi2/core/package_managers/pip.py
```diff
@@ -1724,32 +1772,47 @@ def download_info(self) -> dict[str, Any]:
...
+    def _package_filter(pkg: pypi_simple.DistributionPackage) -> bool:
+        return (
+            pkg.version is not None
+            and canonicalize_version(pkg.version) == normalized_version
+            and pkg.package_type is not None
+            and pkg.package_type in allowed_distros
+        )
 
-    user_checksums = set(map(_to_checksum_info, requirement.hashes))
+    filtered_packages = (package for package in packages if _package_filter(package))
```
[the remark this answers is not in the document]
ben-alkov: Comprehensions basically *are* FP :stuck_out_tongue_closed_eyes:

cachi2/core/package_managers/pip.py
```diff
@@ -8,13 +8,24 @@
 import os.path
```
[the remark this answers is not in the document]
ben-alkov: > Would ^this be an acceptable explanation

Yes, thanks.

> dropped the related bits

WDYM?

cachi2/core/package_managers/pip.py
```diff
@@ -155,16 +167,16 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
     name = package["name"]
     dependency_kind = package.get("kind", None)
     version = None
-    qualifiers = None
+    qualifiers: Union[None, dict[str, str]] = None
```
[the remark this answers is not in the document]
ben-alkov: Still learning typing best practices, so...

cachi2/core/package_managers/pip.py
```diff
@@ -987,7 +999,7 @@ def _parsed(self) -> dict[str, Any]:
 
             if requirement_line:
                 parsed["requirements"].append(
-                    PipRequirement.from_line(requirement_line, requirement_options)
+                    PipRequirement.from_line(requirement_line, requirement_options)  # type: ignore
```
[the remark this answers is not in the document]
ben-alkov: It depends on how loose the typer is, but if mypy doesn't care, I'll remove it.

cachi2/core/package_managers/pip.py
```diff
@@ -1250,17 +1262,17 @@ def from_line(cls, line: str, options: list[str]) -> "PipRequirement":
...
-        requirement.package = parsed.project_name
-        requirement.raw_package = parsed.name
-        requirement.version_specs = parsed.specs
-        requirement.extras = parsed.extras
-        requirement.environment_marker = str(parsed.marker) if parsed.marker else None
+        requirement.package = req.project_name
+        requirement.raw_package = req.name
+        requirement.version_specs = req.specs
+        requirement.extras = req.extras
+        requirement.environment_marker = str(req.marker) if req.marker else ""
```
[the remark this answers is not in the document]
ben-alkov: > TL;DR pkg_resources as an API has been deprecated

I'm well aware, and I'd just as soon remove it...

cachi2/core/package_managers/pip.py
```diff
@@ -1419,72 +1421,102 @@ def _download_dependencies(
     pip_deps_dir = output_dir.join_within_root("deps", "pip")
     pip_deps_dir.path.mkdir(parents=True, exist_ok=True)
 
-    downloaded: list[dict[str, Any]] = []
-    to_download: list[DistributionPackageInfo] = []
+    def _check_local(
```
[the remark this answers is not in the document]
ben-alkov: > Do we need those lists at all?

Now that I'm looking at it, I'm guessing it's an artifact from the earliest phase of refactoring, and you're absolutely right (we don't need them).

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
ben-alkov: I want to return a single list from `_process_package_distributions`, no matter what.
The pypi type of the artifact is embedded in the DPI, so returning two lists is over-complicated.
Even treating `best_sdist` differently from everything else complicates the code unnecessarily (admittedly, it's a little weird to have it essentially be a single-element list, but not so mind-blowing as to confuse anyone).

cachi2/core/package_managers/pip.py
```diff
@@ -1419,72 +1421,102 @@ def _download_dependencies(
...
+
+    def _finalize(req: PipRequirement, download_info: dict[str, Any]) -> dict[str, Any]:
+        log.info(
+            "Successfully downloaded/checked '%s' in path '%s'",
+            req.download_line,
+            download_info["path"].relative_to(output_dir),
+        )
+        download_info["kind"] = req.kind
+        download_info["requirement_file"] = str(requirements_file.file_path.subpath_from_root)
+        return download_info
```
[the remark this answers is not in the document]
ben-alkov: > it would be great to have such a helper defined outside of _download_dependencies

The refactor would explode in complexity because of the need to pass a huge amount of context *out of* `_download_dependencies`

> download_info ... would not need to be passed around

The pattern for all artifacts looks like

```python
download_info = get_download_info_from_external_source
download_info = munge_download_info  # hashes, kind, requirements file
downloaded.append(download_info)  # hopefully self-evident
```

so, *some* passing around is required

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: >>    it can be tweaked

> Explain?

That was a forward reference to my later comment here: https://github.com/containerbuildsystem/cachi2/pull/467#discussion_r1497943711

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: I meant dropping the following chunks from the commit message itself:

> fix dodgy urllib import - this is technically incorrect: in fact, I don't understand how it works at all (note the complete lack of parse):
>
```
>>>import urllib
>>> print(dir(urllib))
['__builtins__', '__cached__', '__doc__', '__file__', `'__loader__',` '__name__', '__package__', '__path__', '__spec__']
```

cachi2/core/package_managers/pip.py
```diff
@@ -1119,7 +1133,7 @@ def __init__(self) -> None:
         self.hashes: list[str] = []
         self.qualifiers: dict[str, Any] = {}
 
-        self.kind: str = ""
+        self.kind: str = ""  # ("pypi", "url", "vcs")
```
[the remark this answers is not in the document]
ben-alkov: Either or both of "mypy" and "me" is too stupid to understand how this works, so I'm leaving it as a comment, as I think it's a valuable clarification.

cachi2/core/package_managers/pip.py
```diff
@@ -1403,71 +1420,92 @@ def _download_dependencies(
...
-            try:
-                if pkg.should_verify_checksums():
-                    must_match_any_checksum(pkg.path, pkg.checksums_to_verify)
-            except PackageRejected:
-                pkg.path.unlink()
-                log.warning("The %s is removed from the output directory", pkg.path.name)
+            for dpi in artifacts_to_download:
+                log.info("Processing pypi download '%s'", dpi.path.name)
+                asyncio.run(
+                    async_download_files({dpi.url: dpi.path}, get_config().concurrency_limit)
```
taylormadore, line 1486: Are we now downloading all of the artifacts one-by-one? Previously we were processing a list of wheels asynchronously

ben-alkov: I need to take another look at this.

cachi2/core/package_managers/pip.py
```diff
@@ -1800,17 +1827,16 @@ def _process_package_distributions(
...
-                wheels.append(dpi)
-            else:
-                log.info("Filtering out %s due to checksum mismatch", package.filename)
+            log.info("Filtering out %s due to checksum mismatch", package.filename)
 
-    if len(sdists) != 0:
-        best_sdist = max(sdists, key=_sdist_preference)
-        if best_sdist.is_yanked:
+    sdists = [dpi for dpi in artifacts if dpi.package_type == "sdist"]
+    wheels = [dpi for dpi in artifacts if dpi.package_type == "wheel"]
```
slimreaper35, line 1836: This split of artifacts could be done in the previous for loop. Something like:

```
if dpi.should_download():
    if dpi.package == "sdist":
        sdists.append(dpi)
   elif ...
```
As a result, we will not need the `artifacts` variable anymore.

cachi2/core/package_managers/pip.py
slimreaper35: How about this one?

slimreaper35 (commented):
you can fix one failing test by:

```
@mock.patch("cachi2.core.package_managers.pip._process_package_distributions")
**@mock.patch("cachi2.core.package_managers.pip.async_download_files")**
@mock.patch("cachi2.core.package_managers.pip._check_metadata_in_sdist")
def test_download_from_requirement_files(
    self,
    check_metadata_in_sdist: mock.Mock,
    mock_download_files: mock.Mock,
    mock_distributions: mock.Mock,
    rooted_tmp_path: RootedPath,
) -> None:

...
mock_distributions.side_effect = [[pypi_package1], [pypi_package2]]
...
```

ben-alkov:
> you can fix one failing test by:
> mock_distributions.side_effect = [[pypi_package1], [pypi_package2]]

I've got that locally, but then my comment in Slack becomes the problem:
> `test_download_from_requirement_files` is failing in `async_download_files()` with 'cachi2.core.errors.FetchError: exception_name: InvalidURL, details: ""'
It was failing in '.pip.download_binary_file' (same error), so I changed the mock to '.general.async_download_files'... there isn't anything "lower level" than that to mock

eskultety:
Uhm, what's wrong with this that bit of code CodeQL complains about it? I don't see a problem, it's proper instantiation, can we dismiss the error as false positive?

cachi2/core/package_managers/pip.py
```diff
@@ -1370,20 +1453,131 @@ def _split_hashes_from_options(cls, options: list[str]) -> tuple[list[str], list
...
+        return None
+
+    artifacts: list[DistributionPackageInfo] = _process_package_distributions(
+        req, pip_deps_dir, allow_binary
+    )
+    artifacts_to_download: list[DistributionPackageInfo] = [a for a in artifacts if a.path.exists()]
+    files: dict[str, Union[str, PathLike[str]]] = {
+        dpi.url: dpi.path for dpi in artifacts_to_download
+    }
+    asyncio.run(async_download_files(files, get_config().concurrency_limit))
```
slimreaper35, line 1525: Shouldn't we download all wheels from requirement.txt at once?

cachi2/core/package_managers/pip.py
```diff
@@ -1370,20 +1453,131 @@ def _split_hashes_from_options(cls, options: list[str]) -> tuple[list[str], list
...
+
+    if retval := _process_req_wrapper(artifacts_to_download, "-- Processing pypi download '%s'"):
+        download_infos.append(retval)
+    if retval := _process_req_wrapper(
```
slimreaper35, line 1529: What does this symbol `:=`  stands for ?

cachi2/core/package_managers/pip.py
```diff
@@ -1788,27 +1915,31 @@ def _process_package_distributions(
...
                 docs = PIP_NO_SDIST_DOC
-
             raise PackageRejected(
                 f"No distributions found for package {name}=={version}",
                 solution=solution,
                 docs=docs,
             )
 
-    return best_sdist, wheels
+    return best_sdist + wheels if best_sdist else wheels
```
slimreaper35, line 1942: I think what me and Erik were suggesting, is that `best_sdist` should not be a list variable. But you can return list from this function.

cachi2/core/package_managers/pip.py
```diff
@@ -1648,136 +1814,97 @@ def _validate_provided_hashes(requirements: list[PipRequirement], require_hashes
...
+            and pkg.package_type in allowed_distros
+        )
 
-    user_checksums = set(map(_to_checksum_info, requirement.hashes))
+    valid_packages = (package for package in packages if _is_valid(package))
 
-    for package in filtered_packages:
-        pypi_checksums = {
+    for package in valid_packages:
+        pypi_checksums: set[ChecksumInfo] = {
```
slimreaper35, line 1882: maybe `valid_packages` variable is not needed here as well, I can imagine something like this instead:

```
for package in packages:
    if not _is_valid(...):
        continue
```

cachi2/core/package_managers/pip.py
ben-alkov: I want to return a single list.

cachi2/core/package_managers/pip.py
ben-alkov: https://peps.python.org/pep-0572 , the "walrus" operator for assignment expressions, from Python 3.8 https://docs.python.org/3.8/reference/expressions.html?highlight=walrus#assignment-expressions

cachi2/core/package_managers/pip.py
eskultety: @slimreaper35 I think the problem here is that `best_sdist` can be None in which case you'd have an extra element in the list unless you keep the conditional @ben-alkov  has in there. Can't we have an `artifacts` (`dpis` or `packages` or `ret`) variable and do:
```
ret = []
if sdists: 
    best_sdist = max()
    if best_sdist.is_yanked:
        ...
    ret.append(best_sdist)

...

ret.extend(wheels)
return artifacts
```

That way you wouldn't need the conditional in the return statement. Note that I know you were asked to drop the `artifacts` variable in comment https://github.com/containerbuildsystem/cachi2/pull/467#discussion_r1514082878, but I think we can simplify here easily.

cachi2/core/package_managers/pip.py
```diff
@@ -1484,77 +1594,48 @@ def _download_dependencies(
...
 
+        if req.kind == "pypi":
+            download_infos = _process_pypi_req(
+                req,
+                requirements_file=requirements_file,
+                require_hashes=require_hashes,
+                pip_deps_dir=pip_deps_dir,
+                allow_binary=allow_binary,
+            )
+            processed = processed + download_infos
```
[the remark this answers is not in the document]
ben-alkov: mypy says no

cachi2/core/package_managers/pip.py
```diff
@@ -1751,41 +1832,54 @@ def _process_package_distributions(
...
+    A note on nomenclature (to address a common misconception)
+    It really is not "source" vs. "wheel" (i.e. "wheel" is somehow inherently not "source")
+    It is very much "sdist" vs. "wheel".
+    To a first approximation, sdists and wheels are BOTH "source".
+    - sdists are usually "raw" repo dumps, including docs, build scripts, etc.
+    - wheels are *only* Python source (but see below) - so, there's the irony: wheels are _usually_
+      more "source" than sdists
+
+    The confusion probably arises because wheels allow and encourage binary artifacts; in reality,
+    it's not that common.
```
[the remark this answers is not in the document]
ben-alkov: Are we considering compiled c modules "application data"?

cachi2/core/package_managers/pip.py
ben-alkov: `ret.append(best_sdist)` can still be None, so we still need to filter it out

cachi2/core/package_managers/pip.py
```diff
@@ -1247,17 +1262,17 @@ def from_line(cls, line: str, options: list[str]) -> "PipRequirement":
...
-        requirement.package = parsed.project_name
-        requirement.raw_package = parsed.name
-        requirement.version_specs = parsed.specs
-        requirement.extras = parsed.extras
-        requirement.environment_marker = str(parsed.marker) if parsed.marker else None
+        requirement.package = req.project_name
+        requirement.raw_package = req.name
+        requirement.version_specs = req.specs
+        requirement.extras = req.extras
+        requirement.environment_marker = str(req.marker) if req.marker else None
```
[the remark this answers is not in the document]
ben-alkov: It's outside the mandate of "combine sdist and wheel processing"

cachi2/core/package_managers/pip.py
```diff
@@ -1453,8 +1453,117 @@ def _split_hashes_from_options(cls, options: list[str]) -> tuple[list[str], list
...
+    artifacts_to_download: list[DistributionPackageInfo] = [a for a in artifacts if a.path.exists()]
+    files: dict[str, Union[str, PathLike[str]]] = {
+        dpi.url: dpi.path for dpi in artifacts_to_download
+    }
+    asyncio.run(async_download_files(files, get_config().concurrency_limit))
+
+    if retval := _process_req_wrapper(artifacts_to_download, "-- Processing pypi download '%s'"):
+        download_infos.append(retval)
+    if retval := _process_req_wrapper(
+        [a for a in artifacts if not a.path.exists()], "-- Checking existing pypi artifact '%s'"
```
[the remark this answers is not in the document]
ben-alkov: But then you end up in the situation where you have **no** idea as to what's _actually_ been processed. You just have a big blob of what you *attempted* to process. Yes, the logging is more verbose, but it's also **more informative**.

Digging through the logs of failed OSBS builds (for years) definitely taught me that there's such a thing as too much logging :roll_eyes:, but the work I did on forcing the OSBS integration test logs to "tell a story" about what was being **attempted** vs. what was being **completed** convinced me that the volume of logging isn't a problem - it's **what** and **how** you tell an engineer diagnosing a failure that's actually important (obviously there's a balance to be kept here)

cachi2/core/package_managers/pip.py
slimreaper35: how ?

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: Right, good point, I didn't consider those, but I guess my point still remains - does that count as encouraging someone to use it? My point being as I said earlier your app either needs it (in this case being a low level Python app) or it doesn't in which case it wouldn't ship those.

cachi2/core/package_managers/pip.py
eskultety: > `ret.append(best_sdist)` can still be None, so we still need to filter it out

Once you have a non-empty `sdists` list (we just checked it) then you can't end up with `best_sdist` being None coming from `max` or can you?

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: Sure, and the answer to that is that that particular method has `@no_type_check` on it :/

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: Note that I'm not after an incomprehensible log dump where you don't know what is being processed, I'm rather contemplating about the importance to report 2 distinct log messages:

- `"-- Processing pypi download  ... "`
- `"-- Checking existing pypi artifact ... "`

where a simple `"Processing <artifact> name"` might perfectly suffice. If both lists to be processed are dumped somewhere early in the processing then you only need to know which one is being processed at a given time, you don't need to know whether we're downloading it or just checking it, right?

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: Does it?
I just tried with the following diff and mypy didn't find any issues:
```
iff --git a/cachi2/core/package_managers/pip.py b/cachi2/core/package_managers/pip.py
index 3e52b1d7..17612c5c 100644
--- a/cachi2/core/package_managers/pip.py
+++ b/cachi2/core/package_managers/pip.py
@@ -1611,7 +1611,7 @@ def _download_dependencies(
                 pip_deps_dir=pip_deps_dir,
                 allow_binary=allow_binary,
             )
-            processed = processed + download_infos
+            processed.extend(download_infos)
         elif req.kind == "vcs":
             download_info = _process_vcs_req(
                 req,
```

tests/unit/package_managers/test_pip.py
```diff
@@ -3415,7 +3414,24 @@ def test_download_dependencies(
...
+        ] + [
+            wheel.download_info
+            | {
+                "kind": "pypi",
+                "hash_verified": use_hashes,
+                "requirement_file": str(req_file.file_path.subpath_from_root),
+            }
+            for wheel in wheels
+        ]
+
```
eskultety, line 3433: You can't generalize all wheels like this, especially not the `hash_verified` part because each wheel is supposed to test a different behaviour (matching hash, mismatching hash, no hash), so this is another issue causing `test_download_dependencies` to fail.
Moreover, there's too much happening in this statement, please unwind/open code the logic ending with a simple `pypi_downloads = [sdist] + wheels`

cachi2/core/package_managers/pip.py
```diff
@@ -1370,20 +1453,140 @@ def _split_hashes_from_options(cls, options: list[str]) -> tuple[list[str], list
...
+        _process_req_wrapper(artifacts_to_download, "-- Processing pypi artifact to download '%s'")
+    )
+
+    return download_infos
+
+
+def _process_vcs_req(
+    req: PipRequirement, pip_deps_dir: RootedPath, **kwargs: Any
+) -> dict[str, Any]:
+    log.info("-- Processing vcs repo '%s'", req.url)
```
slimreaper35, line 1549: Do these symbols have any special meaning? ("--" or "*")
I find that kind of ugly when all other messages start immediately after the log level.

```
2024-03-18 20:58:13,087 INFO Extracting metadata from pyproject.toml
2024-03-18 20:58:13,088 INFO Resolved package name: 'wheels-of-fortune'
2024-03-18 20:58:13,089 INFO Resolved package version: '0.1.0'
2024-03-18 20:58:13,090 INFO At least one dependency uses the --hash option, will require hashes
2024-03-18 20:58:13,090 INFO * Processing requirement line 'pyicu==2.12'
2024-03-18 20:58:13,713 INFO -- Processing pypi artifact to download 'PyICU-2.12.tar.gz'
2024-03-18 20:58:13,713 INFO Verifying checksums of PyICU-2.12.tar.gz
2024-03-18 20:58:13,718 INFO -- Successfully processed 'pyicu==2.12' in path 'deps/pip/PyICU-2.12.tar.gz'
2024-03-18 20:58:13,718 INFO * Finished processing requirement line 'pyicu==2.12
```

cachi2/core/package_managers/pip.py
```diff
@@ -1370,20 +1453,140 @@ def _split_hashes_from_options(cls, options: list[str]) -> tuple[list[str], list
...
+
+    return _process_req(
+        req,
+        pip_deps_dir=pip_deps_dir,
+        download_info=_download_vcs_package(req, pip_deps_dir),
+        **kwargs,
```
slimreaper35, line 1555: Why only `_process_pypi_req` does not have `**kwargs` and `_process_vcs_req`, `_process_url_req` do?

cachi2/core/package_managers/pip.py
```diff
@@ -1370,20 +1453,140 @@ def _split_hashes_from_options(cls, options: list[str]) -> tuple[list[str], list
...
+            )
+        return infos
+
+    artifacts: list[DistributionPackageInfo] = _process_package_distributions(
+        req, pip_deps_dir, allow_binary
+    )
+    artifacts_to_check: list[DistributionPackageInfo] = [a for a in artifacts if a.path.exists()]
+    artifacts_to_download: list[DistributionPackageInfo] = [
+        a for a in artifacts if not a.path.exists()
+    ]
```
slimreaper35, line 1529: This chunk of code could be executed in one for loop.

cachi2/core/package_managers/pip.py
```diff
@@ -1370,20 +1453,140 @@ def _split_hashes_from_options(cls, options: list[str]) -> tuple[list[str], list
...
+        dpi.url: dpi.path for dpi in artifacts_to_download
+    }
+    asyncio.run(async_download_files(files, get_config().concurrency_limit))
+
+    download_infos.extend(
+        _process_req_wrapper(artifacts_to_check, "-- Checking existing pypi artifact '%s'")
+    )
+    download_infos.extend(
+        _process_req_wrapper(artifacts_to_download, "-- Processing pypi artifact to download '%s'")
+    )
```
slimreaper35, line 1541: Do we need a separate log message for each artifact? There are more log messages in other functions, and some are only set in DEBUG mode.
I could imagine completely deleting the `_process_req_wrapper`.

Function with the name `_process_pypi_req` calls nested `_process_req_wrapper` which calls `_process_req` outside of the `_process_pypi_req`. 🙉

cachi2/core/package_managers/pip.py
```diff
@@ -1370,20 +1453,140 @@ def _split_hashes_from_options(cls, options: list[str]) -> tuple[list[str], list
...
+    req: PipRequirement,
+    requirements_file: PipRequirementsFile,
+    require_hashes: bool,
+    pip_deps_dir: RootedPath,
+    download_info: dict[str, Any],
+    dpi: Optional[DistributionPackageInfo] = None,
+) -> dict[str, Any]:
+    download_info["kind"] = req.kind
+    download_info["requirement_file"] = str(requirements_file.file_path.subpath_from_root)
+    download_info["hash_verified"] = False
```
slimreaper35, line 1466: I do not like the fact this function immediately rewrites one of its arguments.

cachi2/core/package_managers/pip.py
ben-alkov: IIRC, `_process_pypi_req` doesn't need it, but I'll double-check

eskultety: because unlike `_process_pypi_req` both `_process_vcs_req` and `_process_url_req` don't make use of most of the arguments `_process_req` takes, so here `**kwargs` represents just that - **passthrough arguments**.

cachi2/core/package_managers/pip.py
eskultety: Previously that was the case, yes, so in that regard this is change introduces a regression in the behaviour. Downloading everything in a single async call can be a double-edged sword though. Unless we're seeing a noticeable performance penalty I'd argue that downloading from pypi on a req by req basis gives us the possibility to either recover from errors or fail fatally early without having to potentially download hundreds of MB of data first.

cachi2/core/package_managers/pip.py
```diff
@@ -1787,28 +1917,37 @@ def _process_package_distributions(
...
-
             raise PackageRejected(
                 f"No distributions found for package {name}=={version}",
                 solution=solution,
                 docs=docs,
             )
 
-    return best_sdist, wheels
+    if wheels:
+        processed_dpis.extend(wheels)
```
slimreaper35, line 1948: `if wheels:`  can be omitted

tests/unit/package_managers/test_pip.py
```diff
@@ -3384,20 +3387,38 @@ def test_download_dependencies(
...
+        sdist_d_i = sdist_DPI.download_info | {
+            "kind": "pypi",
+            "requirement_file": str(req_file.file_path.subpath_from_root),
+            "hash_verified": True,
+        }
+        wheels_DPI = []
+        wheels_d_i = []
         if allow_binary:
-            wheels = [
+            wheels_DPI = [
```
eskultety, line 3407: This whole hunk is too verbose, would it be possible to shorten it e.g. like this (haven't tried the tests though)?
```
-        wheels_d_i = []
         if allow_binary:
-            wheels_DPI = [
-                pip.DistributionPackageInfo(
+            for wheel_path, checksums, hash_verified in zip(
+                [wheel_0_download, wheel_1_download, wheel_2_download],
+                [{ChecksumInfo("sha256", "fedcba")}, {ChecksumInfo("sha256", "abcdef")}, set()],
+                [False, True, False],
+            ):
+                dpi = pip.DistributionPackageInfo(
                     "foo",
                     "1.0",
                     "wheel",
-                    wheel_0_download,
+                    wheel_path,
                     "",
                     False,
-                    pypi_checksums={ChecksumInfo("sha256", "fedcba")},
-                ),
-                pip.DistributionPackageInfo(
-                    "foo",
-                    "1.0",
-                    "wheel",
-                    wheel_1_download,
-                    "",
-                    False,
-                    pypi_checksums={ChecksumInfo("sha256", "abcdef")},
-                ),
-                pip.DistributionPackageInfo(
-                    "foo",
-                    "1.0",
-                    "wheel",
-                    wheel_2_download,
-                    "",
-                    False,
-                    pypi_checksums=set(),
-                ),
-            ]
-            wheels_d_i = [
-                wheels_DPI[0].download_info
-                | {
-                    "kind": "pypi",
-                    "requirement_file": str(req_file.file_path.subpath_from_root),
-                    "hash_verified": False,
-                },
-                wheels_DPI[1].download_info
-                | {
-                    "kind": "pypi",
-                    "requirement_file": str(req_file.file_path.subpath_from_root),
-                    "hash_verified": True,
-                },
-                wheels_DPI[2].download_info
-                | {
-                    "kind": "pypi",
-                    "requirement_file": str(req_file.file_path.subpath_from_root),
-                    "hash_verified": False,
-                },
-            ]
-
-        pypi_downloads = [sdist_d_i] + wheels_d_i
+                    pypi_checksums=checksums,  # type: ignore
+                )
+                wheels_DPI.append(dpi)
+                pypi_downloads.append(
+                    dpi.download_info
+                    | {
+                        "kind": "pypi",
+                        "requirement_file": str(req_file.file_path.subpath_from_root),
+                        "hash_verified": False,
+                    }
+                )
```
IOW instead of explicitly creating each DPI and download_info we'd generate it since 98% of data is identical.

tests/unit/package_managers/test_pip.py
```diff
@@ -3406,101 +3427,131 @@ def test_download_dependencies(
...
+                None,  # url_download
+            ]
+        elif use_hashes and allow_binary:
+            mock_must_match_any_checksum.side_effect = [
+                None,  # sdist_download
+                PackageRejected("", solution=None),  # wheel_0_download - checksums NOK
+                None,  # wheel_1_download - checksums OK
+                None,  # vcs_download
+                None,  # url_download
+                # wheel_2_download - no checksums to verify
```
eskultety, line 3476: Shouldn't here also be a value of None rather than just a commentary?

cachi2/core/package_managers/pip.py
```diff
@@ -1648,7 +1666,7 @@ def _validate_provided_hashes(requirements: list[PipRequirement], require_hashes
             )
 
         for hash_spec in hashes:
-            algorithm, _, digest = hash_spec.partition(":")
+            _, _, digest = hash_spec.partition(":")
```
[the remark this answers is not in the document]
ben-alkov: > algorithm is used in the message that follows
No, it isn't?

cachi2/core/package_managers/pip.py
```diff
@@ -1788,17 +1822,23 @@ def _process_package_distributions(
...
                 docs = None
             else:
                 solution = (
-                    "It seems that this version does not exist or isn't published as a source distribution.\n"
-                    "Try to specify the dependency directly via a URL instead, for example, the tarball for a GitHub release."
+                    "It seems that this version does not exist or isn't published as an"
+                    " sdist.\n"
+                    "Try to specify the dependency directly via a URL instead, for example,"
+                    " the tarball for a GitHub release.\n"
+                    "Alternatively, allow the use of wheels."
```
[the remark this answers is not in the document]
ben-alkov: You must be looking at old commits...

tests/unit/package_managers/test_pip.py
ben-alkov: I'll try it out

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: my bad, sorry, wrong type of brackets :/.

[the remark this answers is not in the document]
ben-alkov: @eskultety; On second thought, is this a bug in the original code? Maybe it **should** have been `f"Not a valid hash specifier: {hash_spec!r} (expected {algorithm}:digest)"`...

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
ben-alkov: Fixed

cachi2/core/package_managers/pip.py
```diff
@@ -1788,17 +1822,23 @@ def _process_package_distributions(
                 ),
             )
     else:
-        log.warning("No source distributions found for package %s==%s", name, version)
+        log.warning("No sdist found for package %s==%s", name, version)
```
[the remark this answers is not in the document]
ben-alkov: Fixed

cachi2/core/package_managers/pip.py
```diff
@@ -28,7 +28,7 @@
 import pkg_resources
```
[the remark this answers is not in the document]
ben-alkov: Fixed

cachi2/core/package_managers/pip.py
```diff
@@ -83,6 +83,73 @@
...
+        elif self.user_checksums:
+            checksums = self.user_checksums
+            msg = "using user specified checksums"
+        else:
+            msg = "no checksums reported by PyPI or specified by the user"
+
+        log.debug("%s: %s", self.path.name, msg)
+        return checksums
+
+    def should_download(self) -> bool:
```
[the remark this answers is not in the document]
ben-alkov: Fixed

cachi2/core/package_managers/pip.py
```diff
@@ -2008,25 +2014,24 @@ def _resolve_pip(
     """
     pkg_name, pkg_version = _get_pip_metadata(app_path)
 
-    # This could be an empty list
-    if requirement_files is None:
-        resolved_req_files = _default_requirement_file_list(app_path)
-    else:
-        resolved_req_files = [app_path.join_within_root(r) for r in requirement_files]
+    def resolve_req_files(req_files: Optional[list[Path]], devel: bool) -> list[RootedPath]:
```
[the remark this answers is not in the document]
ben-alkov: Fixed

cachi2/core/package_managers/pip.py
ben-alkov: Fixed

cachi2/core/package_managers/pip.py
ben-alkov: I don't know **exactly** what you mean by "special", but, *textually*, the '*' marks the start/end of processing a single requirement, and '--' marks the start/end of processing a single artifact of an enclosing requirement.

This is all in order to demarcate different actions, to make it **easier** to read the logs. It is *extremely* difficult to figure out what's happening when one is looking at an undifferentiated wall of text.

ben-alkov: TL;DR - [This](https://privatebin.corp.redhat.com/?6d4ecc07257cc34d#9Gzthnx96JsbR5xV7dSVM1jLT8Tgbw6g83jG18VVfrxB) is the cachi2 log, with `allow_binary` enabled, from a *real-world* project ([StableCascade](https://github.com/Stability-AI/StableCascade)) which has many complex dependencies.

I ran it once and interrupted it, then ran it again and captured the log.

If you search for 'click', you can clearly see that above that point in that log, cachi2 is *checking*; below that, it's *downloading*

Yes, it's a lot of log, but the line breaks and "line headers" make it easy to see what's going on.

cachi2/core/package_managers/pip.py
ben-alkov: > There are more log messages in other functions

There are no other log messages which differentiate between CHECKING a single artifact (already on disk), and DOWNLOADING a single artifact (NOT already on disk) - these are completely different operations, and if we don't log them as such, it's harder to pinpoint failures.

cachi2/core/package_managers/pip.py
ben-alkov: I don't understand *exactly* what you mean here - for one thing, the dict arg isn't being rewritten, it's having entries added to it.

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: Probably, I wouldn't mind if you stuffed such a change in that commit :)

tests/unit/package_managers/test_pip.py
eskultety: I think we reached the point where you can squash the marked commits to their intended places in the next revision, I think we'll deal with it wrt review just fine.

cachi2/core/package_managers/pip.py
ben-alkov: Oh, I didn't understand that that's what you meant - I'll look at changing some to DEBUG.

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
ben-alkov: Fixed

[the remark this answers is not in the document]
ben-alkov: Rats. This is actually incorrect (viz. https://github.com/ben-alkov/cachi2/blob/c9f447b50322e01ffd4a4091e02d21910f54c93d/tests/unit/package_managers/test_pip.py#L3290).

TL;DR - If we pass 'algorithm' through to the log message, we end up with the nonsensical (but somewhat comical) `Not a valid hash specifier: 'malformed' (expected malformed:digest)` 🥴

tests/unit/package_managers/test_pip.py
ben-alkov: Tests fixed

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
ben-alkov: OK, I've completely refactored this.

However, be prepared for logs with **110 lines** consisting solely of e.g. "2024-03-20 12:26:04,602 INFO Verifying checksums of tokenizers- ... whl" (or 88 for "setproctitle-", 110 also for "safetensors-", etc.) because *that* log message comes from cachi2/core/checksum.py:65, and I don't plan on touching it.

cachi2/core/package_managers/pip.py
ben-alkov: Refactored

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: Needs to fixed/reverted in a different commit in this series.

tests/unit/package_managers/test_pip.py
```diff
@@ -2749,10 +2749,8 @@ def test_process_existing_package_without_any_distributions(
             )
         else:
             assert str(exc_info.value.solution) == (
-                "It seems that this version does not exist or isn't published as an"
-                " sdist.\n"
-                "Try to specify the dependency directly via a URL instead, for example,"
-                " the tarball for a GitHub release.\n"
+                "It seems that this version does not exist or isn't published as an sdist.\n"
+                "Try to specify the dependency directly via a URL instead, for example, the tarball for a GitHub release.\n"
```
eskultety, line 2753: This adjustment also needs to be in a different commit in this series.

eskultety (commented):
If there's anything else besides the last details related to bisect I don't see it so this is ready to be merged once you address those bits :).

cachi2/core/package_managers/pip.py
```diff
@@ -1650,7 +1668,7 @@ def _validate_provided_hashes(requirements: list[PipRequirement], require_hashes
         for hash_spec in hashes:
             algorithm, _, digest = hash_spec.partition(":")
             if not digest:
-                msg = f"Not a valid hash specifier: {hash_spec!r} (expected algorithm:digest)"
+                msg = f"Not a valid hash specifier: {hash_spec!r} (expected {algorithm}:digest)"
```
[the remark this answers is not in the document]
ben-alkov: That should've been fixed...

cachi2/core/package_managers/pip.py
```diff
@@ -1648,9 +1666,9 @@ def _validate_provided_hashes(requirements: list[PipRequirement], require_hashes
             )
 
         for hash_spec in hashes:
-            algorithm, _, digest = hash_spec.partition(":")
+            _, _, digest = hash_spec.partition(":")
             if not digest:
-                msg = f"Not a valid hash specifier: {hash_spec!r} (expected algorithm:digest)"
+                msg = f"Not a valid hash specifier: {hash_spec!r} (expected 'algorithm:digest')"
```
eskultety, line 1808: You added the extra quotes, so now some of the tests fail, either update the tests or drop this hunk.

cachi2/core/package_managers/pip.py
```diff
@@ -1917,8 +1957,8 @@ def _add_cachito_hash_to_url(parsed_url: urllib.parse.ParseResult, hash_spec: st
 
 
 def _to_checksum_info(hash_: str) -> ChecksumInfo:
-    algorithm, _, digest = hash_.partition(":")
-    return ChecksumInfo(algorithm, digest)
+    _, _, digest = hash_.partition(":")
```
eskultety, line 2042: You can't do this, because then it's up to the runtime what it assigns to `_` and then use in the return line below, so this hunk needs to be dropped. This change wasn't previously here.

cachi2/core/package_managers/pip.py
```diff
@@ -1467,7 +1485,7 @@ def _download_dependencies(
                     must_match_any_checksum(pkg.path, pkg.checksums_to_verify)
             except PackageRejected:
                 pkg.path.unlink()
-                log.warning("The %s is removed from the output directory", pkg.path.name)
+                log.warning("Download '%s' was removed from the output directory", pkg.path.name)
```
eskultety: This line change needs to be reflected in the unit tests otherwise they fail.

cachi2/core/package_managers/pip.py
ben-alkov: Argh

cachi2/core/package_managers/pip.py
ben-alkov: It is? https://github.com/ben-alkov/cachi2/blob/df11af3d29e70f4d1dd3c002cfd8f93e4c29cb78/cachi2/core/package_managers/pip.py#L1475

eskultety: Wrong commit though: https://github.com/containerbuildsystem/cachi2/commit/df11af3d29e70f4d1dd3c002cfd8f93e4c29cb78#diff-4c1241d88628d8032e8836ab6fa96853ce996ad356da838eca9a58daece89c4aR3557

You need the change in this commit too.

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
ben-alkov: Fixed for real :)

cachi2/core/package_managers/pip.py
ben-alkov: Fixed

cachi2/core/package_managers/pip.py
ben-alkov: Fixed

tests/unit/package_managers/test_pip.py
ben-alkov: This diff chunk should not be in this commit (it is actually **correctly** applied in "toil(package managers:pip): cosmetics and types", regardless of hash - that's where it's *always* been 🤷), and I can't make it go away.

It does **NOT** appear for me locally, nor does it show up for me locally if I do completely new fresh clone of my fork.

ben-alkov: In fact, if I open the file list diff for that specific commit (where the diff shows that incorrect chunk), use the "..." menu to "View" 'pip.py' as a file, then click the "Edit in place" button - the content is correct as expected i.e. **NOT** what appears in that diff.

ben-alkov: I've been looking at the wrong FILE this whole time - the chunk is in 'test_pip.py', not 'pip.py' :skull:

cachi2/core/package_managers/pip.py
ben-alkov: Either way, it seems to be in the correct state now.

tests/unit/package_managers/test_pip.py
ben-alkov: Fixed

eskultety (approved):

slimreaper35 (approved):

PR #500, opened 2024-03-18
 Add full Go 1.21 support

eskultety:

This PR enables full Go 1.21 support by the means of instructing Go to download the desired toolchain via`GOTOOLCHAIN=auto`. This is limited to 1.21 only, so 1.22 will be explicitly refused (1.22 will be added by a standalone PR).

Notes:
- first 3 patches are some generic tiny fixes
- `EnvironmentVariable` was adjusted to act as a more generic resolvable template (e.g. to allow `GOPROXY=file://${GOMODCACHE}/foo`)
- Go 1.21 full support is enabled using `GOTOOLCHAIN` and `GOPROXY`

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -36,7 +36,7 @@ RUN npm install && \
 #   - clean any build artifacts Go creates as part of the process.
 RUN go install 'golang.org/dl/go1.20@latest' && \
     "$HOME/go/bin/go1.20" download && \
-    mkdir /usr/local/go && \
+    mkdir -p /usr/local/go && \
     mv "$HOME/sdk/go1.20" /usr/local/go && \
     rm -rf "$HOME/go" "$HOME/.cache/go-build/"
 
--- a/Containerfile
+++ b/Containerfile
@@ -34,10 +34,12 @@ RUN npm install && \
 #   - move the SDK to a host local install system-wide location
 #   - remove the shim as it forces and expects the SDK to be used from $HOME
 #   - clean any build artifacts Go creates as part of the process.
-RUN go install 'golang.org/dl/go1.20@latest' && \
-    "$HOME/go/bin/go1.20" download && \
-    mkdir -p /usr/local/go && \
-    mv "$HOME/sdk/go1.20" /usr/local/go && \
-    rm -rf "$HOME/go" "$HOME/.cache/go-build/"
+RUN for go_ver in "go1.20" "go1.21.0"; do \
+        go install "golang.org/dl/${go_ver}@latest" && \
+        "$HOME/go/bin/$go_ver" download && \
+        mkdir -p /usr/local/go && \
+        mv "$HOME/sdk/$go_ver" /usr/local/go && \
+        rm -rf "$HOME/go" "$HOME/.cache/go-build/"; \
+    done
 
 ENTRYPOINT ["cachi2"]
--- a/README.md
+++ b/README.md
@@ -331,7 +331,9 @@ See [docs/gomod.md](docs/gomod.md) for more details.
   the intended go version, Cachi2 should handle it appropriately. If your go version is *higher* than what Cachi2 uses,
   there is a good chance it will be compatible regardless, as long as the dependency resolution did not change between
   the two versions. For example, dependency resolution did change in [go 1.18][go118-changelog] but not in
-  [go 1.19][go119-changelog].
+  [go 1.19][go119-changelog]. Things are a bit more complicated with [Go 1.21][go121-changelog], if
+  you are or have been experiencing issues with cachi2 related to Go 1.21+, please refer to
+  [docs/gomod.md](docs/gomod.md#go-121-since-cachi2-v050).
 
 ### pip
 
--- a/README.md
+++ b/README.md
@@ -333,17 +333,6 @@ See [docs/gomod.md](docs/gomod.md) for more details.
   the two versions. For example, dependency resolution did change in [go 1.18][go118-changelog] but not in
   [go 1.19][go119-changelog].
 
-#### Go 1.21+ *(since cachi2-v0.5.0)*
-  Starting with [Go 1.21][go121-changelog], Go changed the meaning of the `go 1.X` directive in
-  that it now specifies the [minimum required version](https://go.dev/ref/mod#go-mod-file-go) of Go
-  rather than a suggested version as it originally did. The format of the version string in the
-  `go` directive now also includes the micro release and if you don't include the micro release in
-  your `go.mod` file yourself (i.e. you only specify the language release) Go will try to correct
-  it automatically inside the file. Last but not least, Go 1.21 also introduced a new keyword
-  [`toolchain`](https://go.dev/ref/mod#go-mod-file-toolchain) to the `go.mod` file. What this all
-  means in practice for end users is that you may not be able to process your `go.mod` file with an older version
-  of Go (and hence older cachi2) as you could in the past for various reasons.
-
 ### pip
 
 <https://pip.pypa.io/en/stable/>
--- a/cachi2/core/models/output.py
+++ b/cachi2/core/models/output.py
@@ -1,10 +1,11 @@
 import logging
 import string
 from pathlib import Path
-from typing import Dict, Literal, Optional
+from typing import Dict, Literal, Optional, Set
 
 import pydantic
 
+from cachi2.core.errors import Cachi2Error
 from cachi2.core.models.property_semantics import merge_component_properties
 from cachi2.core.models.sbom import Component, Sbom
 from cachi2.core.models.validators import unique_sorted
@@ -43,12 +44,54 @@ class EnvironmentVariable(pydantic.BaseModel):
         substitute all placeholders defined; if no placeholders are contained within the value
         string, substitution is a NOOP (e.g. legacy "literal" variables)
         """
+
+        def get_placeholders(t: string.Template) -> Set[str]:
+            """Return a set of placeholders in a template.
+
+            Implementation is based on [1] which appeared in 3.11 [2] without additional error
+            handling which we don't need for our mostly internal use case.
+            [1] https://github.com/python/cpython/blob/3b4cd48d2988e74405838accde5edcc3b71bec48/Lib/string.py#L157
+            [2] https://docs.python.org/3.11/library/string.html#string.Template.get_identifiers
+
+            TODO: Drop this after bumping minimum required version to 3.11
+            """
+            placeholders = set()
+            matches = t.pattern.finditer(t.template)
+            for m in matches:
+                if placeholder := m.group("named") or m.group("braced"):
+                    placeholders.add(placeholder)
+
+            return placeholders
+
         # legacy path variable handling, need to prepend the base path placeholder
         if self.kind == "path" and "output_dir" in mappings:
             log.debug(f"Adjusting a legacy path variable value '{self.name}={self.value}'")
             self.value = "${output_dir}/" + self.value
 
-        return string.Template(self.value).safe_substitute(mappings)
+        # "Recursively" resolve potentially nested variables up to len(mappings) tries
+        log.debug(f"Resolving environment variable '{self.name}={self.value}'")
+        ret = self.value
+        substituted: Set[str] = set()
+        for i, m in enumerate(mappings):
+            log.debug(f"EnvironmentVariable resolution iteration {i + 1}.: {ret}")
+
+            t_old = string.Template(ret)
+            ret = t_old.safe_substitute(mappings)
+            t_new = string.Template(ret)
+            p_new = get_placeholders(t_new)
+
+            substituted |= get_placeholders(t_old) & mappings.keys()
+            if substituted & p_new:
+                raise Cachi2Error(
+                    f"Detected a cycle in environment variable expansion of '{self.name}'",
+                    solution=(
+                        "Inspect all relevant environment variables and make sure their "
+                        "expansion doesn't lead to a cycle. Ideally, avoiding nesting of "
+                        "variables altogether."
+                    ),
+                )
+
+        return ret
 
 
 class ProjectFile(pydantic.BaseModel):
--- a/cachi2/core/models/output.py
+++ b/cachi2/core/models/output.py
@@ -1,7 +1,7 @@
 import logging
 import string
 from pathlib import Path
-from typing import Literal, Optional
+from typing import Dict, Literal, Optional
 
 import pydantic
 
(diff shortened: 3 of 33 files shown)


cachi2/core/models/output.py
```diff
@@ -48,7 +49,29 @@ def resolve_value(self, mappings: Dict[str, str]) -> str:
...
             self.value = "${output_dir}/" + self.value
 
-        return string.Template(self.value).safe_substitute(mappings)
+        # "Recursively" resolve potentially nested variables up to len(mappings) tries
+        ret = self.value
+        placeholders_seen = []
+        for i, m in enumerate(mappings):
+            log.debug(f"EnvironmentVariable resolution iteration {i}.: {ret}")
+
+            placeholders_seen.append(m)
```
chmeliik, line 58: The nested variable resolution is very cool :+1: 

But in the current state, cycle detection doesn't work correctly (depends on the order of the `mapping`)
```python
>>> e = EnvironmentVariable(name="A", value="$C")
>>> e.resolve_value({"C": "$B", "B": "1"})
'1'
>>> e.resolve_value({"B": "1", "C": "$B"})
  File "<stdin>", line 1, in <module>
  File "/home/acmiel/RedHat/Cachito/cachi2/cachi2/core/models/output.py", line 66, in resolve_value
    raise Cachi2Error(
cachi2.core.errors.Cachi2Error: Detected a cycle in environment variable expansion of 'A'
```

eskultety: Doh! Nasty bug, shameful...@chmeliik thanks for pointing this out!

brunoapimentel: I know you'll try to solve this meticulously first thing in the morning, but I wonder if we need nested variable resolution in the first place. I does look nice, but it adds quite some complexity, and maybe we don't need it right now.

Our current use case can be solved if we declare GOPROXY as:
```
file://${output_dir}/deps/gomod/pkg/mod/cache/download
```
... and we can hold on adding nested variable resolution for when a use case that can't be solved this way comes up.

eskultety: @chmeliik updated with IMO a more robust solution, see if you can spot another resolution problem. I also tweaked and rearranged the test data so that your finding is covered.

eskultety: @brunoapimentel I understand the motivation and what you suggest was my initial approach, but then I figured that I hate maintaining duplicate static strings which leads to the need to fix them eventually at all places. It all comes down to immediate effect vs hardening. IMO it is the immediate impact fixes that usually lead to and add up to the technical debt that will likely never be addressed, but there surely is a line to be drawn. If you feel strongly about the added complexity (which won't even be exercised 99% of the cases) being too high, then I'll drop the change no problem.

cachi2/core/models/output.py
```diff
@@ -43,12 +44,53 @@ def resolve_value(self, mappings: Dict[str, str]) -> str:
...
+            if substituted & p_new:
+                raise Cachi2Error(
+                    f"Detected a cycle in environment variable expansion of '{self.name}'",
+                    solution=(
+                        "Inspect all relevant environment variables and make sure their "
+                        "expansion doesn't lead to a cycle. Ideally, avoiding nesting of "
+                        "variables altogether."
+                    ),
+                )
+            substituted |= get_placeholders(t_old) - p_new
```
chmeliik, line 91: If I understand correctly, this is needed for cases such as:

```python
before_substitution = "${known_variable} and ${unknown_variable}"
after_substitution = "some_value and ${unknown_variable}"
```
We don't want the `unknown_variable` to cause false-positive cycle detection errors, so we avoid marking it as substituted

I wonder if this can lead to false negatives as well though
```python
before_substitution = "${foo} and ${bar}"
after_substitution = "${bar} and ${foo}"
# get_placeholders(before_substitution) - get_placeholders(after_substitution) == empty set ?
```

eskultety: > I wonder if this can lead to false negatives as well though
>```
>before_substitution = "${foo} and ${bar}"
>after_substitution = "${bar} and ${foo}">
># get_placeholders(before_substitution) - get_placeholders(after_substitution) == empty set
>```

@chmeliik if we add an additional `before_substitution != after_substitution` that would take care of the problem wouldn't it?

cachi2/core/models/output.py
brunoapimentel: I don't feel strongly about it, my reasoning was mostly to avoid extra work, since we will need to reach a reasonable solution with this nested variable resolution algorithm.

cachi2/core/models/output.py
chmeliik: Hmm, not sure. The cycles can be more complicated
```python
mappings = {
    "cycle_part1": "$cycle_part3",
    "cycle_part2": "$cycle_part1",
    "cycle_part3": "$cycle_part2"
}
```
```python
"$cycle_part1 $cycle_part2"
# ->
"$cycle_part3 $cycle_part1"  # substituted = {"cycle_part2"}
# ->
"$cycle_part2 $cycle_part3"  # substituted = {"cycle_part2", "cycle_part1"}
# cycle detected
```
In this case I believe the cycle will be caught after the second iteration (even though it could be caught after the first one), but I'm not sure if every cycle will be caught

chmeliik: Maybe, rather than subtracting the new placeholders from the `substituted` set, a more robust solution would be to intersect `substituted` with the known `mappings`?

```python
substituted |= get_placeholders(t_old) & mappings.keys()
```

That should still cover the "don't mark unknown variables as substituted" case and shouldn't interfere with cycle detection for known variables

eskultety: > Hmm, not sure. The cycles can be more complicated
> 
> ```python
> mappings = {
>     "cycle_part1": "$cycle_part3",
>     "cycle_part2": "$cycle_part1",
>     "cycle_part3": "$cycle_part2"
> }
> ```
> 
> ```python
> "$cycle_part1 $cycle_part2"
> # ->
> "$cycle_part3 $cycle_part1"  # substituted = {"cycle_part2"}
> # ->
> "$cycle_part2 $cycle_part3"  # substituted = {"cycle_part2", "cycle_part1"}
> # cycle detected
> ```
> 
> In this case I believe the cycle will be caught after the second iteration (even though it could be caught after the first one), but I'm not sure if every cycle will be caught

I guess there is an algorithm that could catch it after the first iteration, but as long as it's detected it's not an issue.

eskultety: > Maybe, rather than subtracting the new placeholders from the `substituted` set, a more robust solution would be to intersect `substituted` with the known `mappings`?
> 
> ```python
> substituted |= get_placeholders(t_old) & mappings.keys()
> ```
> 
> That should still cover the "don't mark unknown variables as substituted" case and shouldn't interfere with cycle detection for known variables

It felt a bit counter intuitive, but yeah, this seems to work, maybe I just rename it to `seen` rather than substituted that actually might describe it better.

chmeliik: IMO `substituted` is still more accurate. In each iteration, every placeholder in the template string really does get substituted - unless it has no mapping in `mappings`

eskultety:
since v2:
- changed detection of variable resolution cycles per the review comments
- forced setting `GOTOOLCHAIN=local` for cases where `go base version > go.mod version > 1.21` (@taylormadore this relates to https://github.com/containerbuildsystem/cachi2/issues/501#issuecomment-2008970964, do you think it will be enough for now?)

cachi2/core/package_managers/gomod.py
```diff
@@ -718,6 +719,57 @@ def _find_missing_gomod_files(source_path: RootedPath, subpaths: list[str]) -> l
...
+
+    log.info(go_mod_version_msg.format(go_mod_version))
+
+    if go_mod_version >= go_max_version:
+        if go_base_version < go_max_version:
+            # our base Go installation is too old and we need a newer one to support new keywords
+            go = Go(release="go1.21.0")
+        elif go_base_version > go_mod_version:
+            # we have a newer toolchain than the requirement, so we don't want GOTOOLCHAIN=auto
+            env["GOTOOLCHAIN"] = "local"
```
taylormadore, line 755: Is this set just for the prefetch? I think we'll need to provide it in the request output

eskultety: darn, yep, good eyes!

eskultety: In fact, this shouldn't even happen here, but more changes will be needed for `_fetch_gomod_source` to be able to make this kind of decision.

eskultety: Based on the discussion in https://github.com/containerbuildsystem/cachi2/issues/501, let's keep `GOTOOLCHAIN=auto` only for prefetch.

eskultety:
since v3:

- `GOTOOLCHAIN=auto` is only used for dependency fetching, but it's not enforced upon users by the means of `generate-env`
- `toolchain` keyword in `go.mod` file is considered as well when checking the maximum supported version violation (due to the use of GOTOOLCHAIN=auto)
- container image is preinstalled with a static 1.21.0 Go SDK along with 1.20
- some minor debug log adjustments
- rebase on main

eskultety:
since v4:
- add documentation for Go 1.21 full support

eskultety:
@brunoapimentel @taylormadore sorry for the delay, please have another look at the changes. I'll work on some integration tests in the meantime.

eskultety:
Since v5:
- fix integration tests by generating fresh mock data

Containerfile
```diff
@@ -34,10 +34,12 @@ RUN npm install && \
 #   - move the SDK to a host local install system-wide location
 #   - remove the shim as it forces and expects the SDK to be used from $HOME
 #   - clean any build artifacts Go creates as part of the process.
-RUN go install 'golang.org/dl/go1.20@latest' && \
-    "$HOME/go/bin/go1.20" download && \
-    mkdir /usr/local/go && \
-    mv "$HOME/sdk/go1.20" /usr/local/go && \
-    rm -rf "$HOME/go" "$HOME/.cache/go-build/"
+RUN for go_ver in "go1.20" "go1.21.0"; do \
```
brunoapimentel, line 37: Not a problem for this PR, but I wonder how we'll enable hermetic Cachi2 builds of Cachi2 since we need to pull these toolchains from the internet. My first idea was to try to create some sort of `go-deps` directory, but then, how can we ask Go to fetch toolchains <= 1.21.0? For toolchains > 1.21.0, the download is easy to setup. 

In any case, we don't need to solve this now, I'm just thinking out loud.

brunoapimentel (commented):
The only major comment is about the GOTOOLCHAIN environment variable that we're setting up for `.build-config.json`. Besides that, it LGTM!

taylormadore (commented):
Overall, this LGTM aside from the output GOTOOLCHAIN=local :rocket: 

All of the local testing I've done looks good :+1:

taylormadore (approved):
Actually, I'm going to approve it anyway given that I won't be here tomorrow and that is a quick fix

cachi2/core/package_managers/gomod.py
```diff
@@ -548,10 +551,11 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
...
     env_vars = {
-        "GOCACHE": {"value": "deps/gomod", "kind": "path"},
-        "GOPATH": {"value": "deps/gomod", "kind": "path"},
-        "GOMODCACHE": {"value": "deps/gomod/pkg/mod", "kind": "path"},
-        "GOTOOLCHAIN": {"value": "local", "kind": "literal"},
+        "GOCACHE": "${output_dir}/deps/gomod",
+        "GOPATH": "${output_dir}/deps/gomod",
+        "GOMODCACHE": "${output_dir}/deps/gomod/pkg/mod",
+        "GOPROXY": "file://${GOMODCACHE}/cache/download",
+        "GOTOOLCHAIN": "local",
```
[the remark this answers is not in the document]
eskultety: Thank you for spotting this, you're right. When I did this I focused solely on enforcing a consistent "base" behaviour of the target image which can default either to `GOTOOLCHAIN=local` or `GOTOOLCHAIN=auto` depending on how it was packaged on that distro and I didn't like that because it can lead to negatively surprising results. However, I then neglected the technical aspect of how the hermetic builds work and falsely assumed the user would override our value to their liking just fine, but it is the other way around actually, whatever users configure will be overriden once they do a "source" on our env file. I'll drop this.

Containerfile
eskultety: > how can we ask Go to fetch toolchains <= 1.21.0? For toolchains > 1.21.0, the download is easy to setup.

Last time I looked at the cache, it seemed like we'd have to reverse engineer how Go creates the directory structure in GOMODCACHE and copy files over there by ourselves from the SDKs downloaded the same way as in this code block. I think a separate discussion is in place and things should become a bit easier if we conclude we don't want to support EOL'd versions of Go in upstream cachi2.

README.md
```diff
@@ -341,8 +341,17 @@ See [docs/gomod.md](docs/gomod.md) for more details.
...
+  older version of Go (and hence older cachi2) as you could in the past for various reasons.
+  Many projects bump their required Go toolchain's micro release as soon as it becomes available
+  upstream (i.e. not waiting for distributions to bundle them properly). This caused problems for
+  *cachi2-v0.5.0* because the container image's version simply may not have been high enough to
+  process a given project's `go.mod` file. Therefore, *cachi2-v0.6.0* implemented a mechanism to
+  always rely on the origin 0th release of a toolchain (e.g. 1.21.0) and use the `GOTOOLCHAIN=auto`
+  setting to instruct Go to fetch any toolchain as specified by the `go.mod` file automatically,
+  hence allowing us to keep up with frequent micro version bumps. **Note that such a language
+  version would still need to be officially marked as supported by cachi2, i.e. we'd not allow Go
+  to fetch e.g. a 1.22 toolchain if the maximum supported Go version by cachi2 were 1.21!**
```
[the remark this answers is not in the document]
eskultety: > which means [this](https://github.com/containerbuildsystem/cachi2/blob/278c0b241ad2aae69bd18be57409de80e7a7ea03/README.md#user-content-fn-go-compat-8d98370aad2c1c2101383579142caa1f) footnote is now incorrect

Technically, it was incorrect all along because any new keywords from releases newer than cachi2's base image would have still failed, I believe that's why it uses the 'there's a good chance' wording. I think it's appropriate to re-word this footnote to accurately reflect the reality of 1.21+.

> If you feel like this is important to be documented, why don't we move this to the docs/gomod.md specific docs?

I believe it is because it has a noticeable user impact, or at least it may have had based on the failures they'd get from requiring latest toolchain micro releases. Okay, I'll try to move it to the gomod docs and add a reference to that in the main README file.

cachi2/core/package_managers/gomod.py
```diff
@@ -718,6 +735,69 @@ def _find_missing_gomod_files(source_path: RootedPath, subpaths: list[str]) -> l
     return invalid_gomod_files
 
 
+def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
```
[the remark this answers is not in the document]
eskultety: I think it's worth noting it was me who cluttered it in the first place...

README.md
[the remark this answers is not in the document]
eskultety: Changed, let me know if you agree with the latest docs changes.

eskultety:
Since v6:
- docs changes as requested in review
- dropped explicit `GOTOOLCHAIN=local` setting in the generated env file to really let users to set the variable however they like
- moved the `GOTOOLCHAIN=auto` enablement to the correct commit

eskultety:
Since v7:
- drop the go backwards compatibility footnote as it dropping was incorrect
- convert `since cachi2-v0.6.0` from bold italic to just italic (typo)

PR #502, opened 2024-03-19
Support for prefetching RPMs [PoC]

onosek:
This is the proof of concept of RPM prefetching functionality.
Design concept: https://github.com/containerbuildsystem/cachi2/discussions/465

A new rpm package manager was added as a development version so `--dev-package-managers` flag needs to be activated.
* `fetch-deps` command will read the rpms lockfile and download listed packages and sources into the output directory structure. Also SBOM file will be generated (bom.json).
* `inject-files` command will create repofile for each arch and create repository metadata out of its RPMs for every repoid dir.

### DIFF
--- a/cachi2/core/package_managers/rpm/main.py
+++ b/cachi2/core/package_managers/rpm/main.py
@@ -1,12 +1,19 @@
+import asyncio
+import hashlib
 import logging
+from os import PathLike
+from pathlib import Path
+from typing import Any, Union
 
 import yaml
 from pydantic import ValidationError
 
+from cachi2.core.config import get_config
 from cachi2.core.errors import PackageManagerError, PackageRejected
 from cachi2.core.models.input import Request
 from cachi2.core.models.output import RequestOutput
 from cachi2.core.models.sbom import Component
+from cachi2.core.package_managers.general import async_download_files
 from cachi2.core.package_managers.rpm.redhat import RedhatRpmsLock
 from cachi2.core.rooted_path import RootedPath
 
@@ -64,7 +71,7 @@ def _resolve_rpm_project(source_dir: RootedPath, output_dir: RootedPath) -> list
 
         log.debug("Validating lockfile.")
         try:
-            _ = RedhatRpmsLock.model_validate(yaml_content)
+            redhat_rpms_lock = RedhatRpmsLock.model_validate(yaml_content)
         except ValidationError as e:
             loc = e.errors()[0]["loc"]
             msg = e.errors()[0]["msg"]
@@ -75,6 +82,84 @@ def _resolve_rpm_project(source_dir: RootedPath, output_dir: RootedPath) -> list
                 ),
             )
 
-        _ = output_dir.join_within_root(DEFAULT_PACKAGE_DIR)
+        package_dir = output_dir.join_within_root(DEFAULT_PACKAGE_DIR)
+        metadata = _download(redhat_rpms_lock, package_dir.path)
+        _verify_downloaded(metadata)
         components: list[Component] = []
         return components
+
+
+def _download(lockfile: RedhatRpmsLock, output_dir: Path) -> dict[Path, Any]:
+    """
+    Download packages mentioned in the lockfile.
+
+    Go through the parsed lockfile structure and find all RPM and SRPM files.
+    Create a metadata structure indexed by destination path used
+    for later verification (size, checksum) after download.
+    Prepare a list of files to be downloaded, and then download files.
+    """
+    metadata = {}
+    for arch in lockfile.arches:
+        log.info(f"Downloading files for '{arch.arch}' architecture.")
+        # files per URL for downloading packages & sources
+        files: dict[str, Union[str, PathLike[str]]] = {}
+        for pkg in arch.packages:
+            repoid = lockfile.internal_repoid if pkg.repoid is None else pkg.repoid
+            dest = output_dir.joinpath(arch.arch, repoid, Path(pkg.url).name)
+            files[pkg.url] = str(dest)
+            metadata[dest] = {
+                "url": pkg.url,
+                "size": pkg.size,
+                "checksum": pkg.checksum,
+            }
+            Path.mkdir(dest.parent, parents=True, exist_ok=True)
+
+        for pkg in arch.source:
+            repoid = lockfile.internal_source_repoid if pkg.repoid is None else pkg.repoid
+            dest = output_dir.joinpath(arch.arch, repoid, Path(pkg.url).name)
+            files[pkg.url] = str(dest)
+            metadata[dest] = {
+                "url": pkg.url,
+                "size": pkg.size,
+                "checksum": pkg.checksum,
+            }
+            Path.mkdir(dest.parent, parents=True, exist_ok=True)
+
+        asyncio.run(async_download_files(files, get_config().concurrency_limit))
+    return metadata
+
+
+def _verify_downloaded(metadata: dict[Path, Any]) -> None:
+    """Use metadata structure with file sizes and checksums for verification \
+    of downloaded packages and sources."""
+    log.debug("Verification of downloaded files has started.")
+
+    def raise_exception(message: str) -> None:
+        raise PackageRejected(
+            f"Some RPM packages or sources weren't verified after being downloaded: '{message}'",
+            solution=(
+                "Check the source of the data or check the corresponding metadata "
+                "in the lockfile (size, checksum)."
+            ),
+        )
+
+    # check file size and checksum of downloaded files
+    for file_path, file_metadata in metadata.items():
+        # size is optional
+        if file_metadata["size"] is not None:
+            if file_path.stat().st_size != file_metadata["size"]:
+                raise_exception(f"Unexpected file size of '{file_path}' != {file_metadata['size']}")
+
+        # checksum is optional
+        if file_metadata["checksum"] is not None:
+            alg, digest = file_metadata["checksum"].split(":")
+            method = getattr(hashlib, alg.lower(), None)
+            if method is not None:
+                h = method(usedforsecurity=False)
+            else:
+                raise_exception(f"Unsupported hashing algorithm '{alg}' for '{file_path}'")
+            with open(file_path, "rb") as f:
+                for chunk in iter(lambda: f.read(READ_CHUNK), b""):
+                    h.update(chunk)
+            if digest != h.hexdigest():
+                raise_exception(f"Unmatched checksum of '{file_path}' != '{digest}'")
--- a/cachi2/core/package_managers/rpm/main.py
+++ b/cachi2/core/package_managers/rpm/main.py
@@ -1,20 +1,80 @@
 import logging
 
+import yaml
+from pydantic import ValidationError
+
+from cachi2.core.errors import PackageManagerError, PackageRejected
 from cachi2.core.models.input import Request
 from cachi2.core.models.output import RequestOutput
 from cachi2.core.models.sbom import Component
+from cachi2.core.package_managers.rpm.redhat import RedhatRpmsLock
+from cachi2.core.rooted_path import RootedPath
 
 log = logging.getLogger(__name__)
+
+
+DEFAULT_LOCKFILE_NAME = "rpms.lock.yaml"
+DEFAULT_PACKAGE_DIR = "deps/rpm"
+
+# during the computing of file checksum read chunk of size 1 MB
+READ_CHUNK = 1048576
 
 
 def fetch_rpm_source(request: Request) -> RequestOutput:
     """Process all the rpm source directories in a request."""
     components: list[Component] = []
     for package in request.rpm_packages:
-        _ = request.source_dir.join_within_root(package.path)
+        path = request.source_dir.join_within_root(package.path)
+        components.extend(_resolve_rpm_project(path, request.output_dir))
 
     return RequestOutput.from_obj_list(
         components=components,
         environment_variables=[],
         project_files=[],
     )
+
+
+def _resolve_rpm_project(source_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+    """
+    Process a request for a single RPM source directory.
+
+    Process the input lockfile, fetch packages and generate SBOM.
+    """
+    # Check the availability of the input lockfile.
+    if not source_dir.join_within_root(DEFAULT_LOCKFILE_NAME).path.exists():
+        raise PackageRejected(
+            f"RPM lockfile '{DEFAULT_LOCKFILE_NAME}' missing, refusing to continue.",
+            solution=(
+                "Make sure your repository has RPM lockfile '{DEFAULT_LOCKFILE_NAME}' checked in "
+                "to the repository."
+            ),
+        )
+
+    lockfile_name = source_dir.join_within_root(DEFAULT_LOCKFILE_NAME)
+    log.info(f"Reading RPM lockfile: {lockfile_name}")
+    with open(lockfile_name) as f:
+        try:
+            yaml_content = yaml.safe_load(f)
+        except yaml.YAMLError as e:
+            log.error(str(e))
+            raise PackageRejected(
+                f"RPM lockfile '{DEFAULT_LOCKFILE_NAME}' yaml format is not correct.",
+                solution=("Check correct 'yaml' syntax in the lockfile."),
+            )
+
+        log.debug("Validating lockfile.")
+        try:
+            _ = RedhatRpmsLock.model_validate(yaml_content)
+        except ValidationError as e:
+            loc = e.errors()[0]["loc"]
+            msg = e.errors()[0]["msg"]
+            raise PackageManagerError(
+                f"RPM lockfile '{DEFAULT_LOCKFILE_NAME}' format is not valid: '{loc}: {msg}'",
+                solution=(
+                    "Check the correct format and whether any keys are missing in the lockfile."
+                ),
+            )
+
+        _ = output_dir.join_within_root(DEFAULT_PACKAGE_DIR)
+        components: list[Component] = []
+        return components
--- a/cachi2/core/package_managers/rpm/main.py
+++ b/cachi2/core/package_managers/rpm/main.py
@@ -1,0 +1,20 @@
+import logging
+
+from cachi2.core.models.input import Request
+from cachi2.core.models.output import RequestOutput
+from cachi2.core.models.sbom import Component
+
+log = logging.getLogger(__name__)
+
+
+def fetch_rpm_source(request: Request) -> RequestOutput:
+    """Process all the rpm source directories in a request."""
+    components: list[Component] = []
+    for package in request.rpm_packages:
+        _ = request.source_dir.join_within_root(package.path)
+
+    return RequestOutput.from_obj_list(
+        components=components,
+        environment_variables=[],
+        project_files=[],
+    )
--- a/cachi2/core/package_managers/rpm/main.py
+++ b/cachi2/core/package_managers/rpm/main.py
@@ -4,6 +4,7 @@ import logging
 from os import PathLike
 from pathlib import Path
 from typing import Any, Union
+from urllib.parse import quote
 
 import yaml
 from pydantic import ValidationError
@@ -12,10 +13,11 @@ from cachi2.core.config import get_config
 from cachi2.core.errors import PackageManagerError, PackageRejected
 from cachi2.core.models.input import Request
 from cachi2.core.models.output import RequestOutput
-from cachi2.core.models.sbom import Component
+from cachi2.core.models.sbom import Component, Property
 from cachi2.core.package_managers.general import async_download_files
 from cachi2.core.package_managers.rpm.redhat import RedhatRpmsLock
 from cachi2.core.rooted_path import RootedPath
+from cachi2.core.utils import run_cmd
 
 log = logging.getLogger(__name__)
 
@@ -85,8 +87,7 @@ def _resolve_rpm_project(source_dir: RootedPath, output_dir: RootedPath) -> list
         package_dir = output_dir.join_within_root(DEFAULT_PACKAGE_DIR)
         metadata = _download(redhat_rpms_lock, package_dir.path)
         _verify_downloaded(metadata)
-        components: list[Component] = []
-        return components
+        return _generate_sbom_components(metadata)
 
 
 def _download(lockfile: RedhatRpmsLock, output_dir: Path) -> dict[Path, Any]:
@@ -163,3 +164,60 @@ def _verify_downloaded(metadata: dict[Path, Any]) -> None:
                     h.update(chunk)
             if digest != h.hexdigest():
                 raise_exception(f"Unmatched checksum of '{file_path}' != '{digest}'")
+
+
+def _generate_sbom_components(files_metadata: dict[Path, Any]) -> list[Component]:
+    """Fill the component list with the package records."""
+    components: list[Component] = []
+    for file_path, file_metadata in files_metadata.items():
+        query_format = (
+            # all nvra macros should be present/mandatory in RPM
+            "%{NAME}\n"
+            "%{VERSION}\n"
+            "%{RELEASE}\n"
+            "%{ARCH}\n"
+            # vendor and epoch are optional in RPM file and in PURL as well
+            # return "" when vendor is not set instead of "(None)"
+            "%|VENDOR?{%{VENDOR}}:{}|\n"
+            # return "" when epoch is not set instead of "(None)"
+            "%|EPOCH?{%{EPOCH}}:{}|\n"
+        )
+        rpm_args = [
+            "-q",
+            "--queryformat",
+            query_format.strip(),
+            str(file_path),
+        ]
+        rpm_fields = run_cmd(cmd=["rpm", *rpm_args], params={})
+        name, version, release, arch, vendor, epoch = rpm_fields.split("\n")
+        log.debug(
+            f"RPM attributes for '{file_path}': name='{name}', version='{version}', "
+            f"release='{release}', arch='{arch}', vendor='{vendor}', epoch='{epoch}'"
+        )
+
+        # sanitize RPM attributes (including replacing whitespaces)
+        vendor = quote(vendor.lower())
+        download_url = quote(file_metadata["url"])
+
+        # https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm
+        # https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairsa
+        purl = (
+            f"pkg:rpm{'/' if vendor else ''}{vendor}/{name}@{version}-{release}"
+            f"?arch={arch}{'&epoch=' if epoch else ''}{epoch}&download_url={download_url}"
+        )
+
+        if file_metadata["checksum"] is None:
+            missing_hash_in_file = file_path.name
+            properties = [Property(name="cachi2:missing_hash:in_file", value=missing_hash_in_file)]
+        else:
+            properties = []
+
+        components.append(
+            Component(
+                name=name,
+                version=version,
+                purl=purl,
+                properties=properties,
+            )
+        )
+    return components
--- a/cachi2/core/package_managers/rpm/main.py
+++ b/cachi2/core/package_managers/rpm/main.py
@@ -1,6 +1,7 @@
 import asyncio
 import hashlib
 import logging
+import shlex
 from os import PathLike
 from pathlib import Path
 from typing import Any, Union
@@ -221,3 +222,79 @@ def _generate_sbom_components(files_metadata: dict[Path, Any]) -> list[Component
             )
         )
     return components
+
+
+def inject_files_post(*args: Any, **kwargs: Any) -> None:
+    """Run extra tasks for the RPM package manager (callback method) within `inject-files` cmd."""
+    if "from_output_dir" in kwargs and "for_output_dir" in kwargs:
+        from_output_dir = kwargs["from_output_dir"]
+        for_output_dir = kwargs["for_output_dir"]
+
+        if Path.exists(from_output_dir.joinpath(DEFAULT_PACKAGE_DIR)):
+            _generate_repos(from_output_dir)
+            _generate_repofiles(from_output_dir, for_output_dir)
+
+
+def _generate_repos(from_output_dir: Path) -> None:
+    """Search structure for all repoid dirs and create repository metadata \
+    out of its RPMs (and SRPMs)."""
+    package_dir = from_output_dir.joinpath(DEFAULT_PACKAGE_DIR)
+    for arch in package_dir.iterdir():
+        if not arch.is_dir():
+            continue
+        for entry in arch.iterdir():
+            if not entry.is_dir() or entry.name == "repos.d":
+                continue
+            repoid = entry.name
+            _createrepo(repoid, entry)
+
+
+def _createrepo(reponame: str, repodir: Path) -> None:
+    """Execute the createrepo utility."""
+    log.info(f"Creating repository metadata for repoid '{reponame}': {repodir}")
+    cmd = ["createrepo_c", str(repodir)]
+    log.debug("$ " + shlex.join(cmd))
+    stdout = run_cmd(cmd, params={})
+    log.debug(stdout)
+
+
+def _generate_repofiles(from_output_dir: Path, for_output_dir: Path) -> None:
+    """
+    Generate templates of repofiles for all arches.
+
+    Search structure at 'path' and generate one repofile content for each arch.
+    Each repofile contains all arch's repoids (including repoids with source RPMs).
+    Repofile (cachi2.repo) for particular arch will be stored in arch's dir in 'repos.d' subdir.
+    Repofiles are not directly created from the templates in this method - templates are stored
+    in the project file.
+    """
+    package_dir = from_output_dir.joinpath(DEFAULT_PACKAGE_DIR)
+    for arch in package_dir.iterdir():
+        if not arch.is_dir():
+            continue
+        log.debug(f"Preparing repofile content for arch '{arch.name}'")
+        content = ""
+        for entry in sorted(arch.iterdir()):
+            if not entry.is_dir() or entry.name == "repos.d":
+                continue
+            repoid = entry.name
+            localpath = for_output_dir.joinpath(DEFAULT_PACKAGE_DIR, arch.name, repoid)
+            content += f"[{repoid}]\n"
+            content += f"baseurl=file://{localpath}\n"
+            content += "gpgcheck=1\n"
+            # repoid directory matches the internal repoid
+            if repoid.startswith("cachi2-"):
+                content += (
+                    "name=Generated repository containing all packages unaffiliated "
+                    "with any official repository\n"
+                )
+        if content:
+            repo_file_path = arch.joinpath("repos.d", "cachi2.repo")
+            if repo_file_path.exists():
+                log.warning(f"Overwriting {repo_file_path}")
+            else:
+                Path.mkdir(arch.joinpath("repos.d"), parents=True, exist_ok=True)
+                log.info(f"Creating {repo_file_path}")
+
+            with open(repo_file_path, "w") as repo_file:
+                repo_file.write(content)
--- a/cachi2/core/package_managers/rpm/redhat.py
+++ b/cachi2/core/package_managers/rpm/redhat.py
@@ -1,0 +1,79 @@
+import logging
+import uuid
+from functools import cached_property
+from typing import Optional
+
+from pydantic import BaseModel, PositiveInt, field_validator, model_validator
+
+log = logging.getLogger(__name__)
+
+
+class Package(BaseModel):
+    """Package item; represents RPM or SRPM file."""
+
+    repoid: Optional[str] = None
+    url: str
+    checksum: Optional[str] = None
+    size: Optional[int] = None
+
+
+class Arch(BaseModel):
+    """Architecture structure."""
+
+    arch: str
+    packages: list[Package] = []
+    source: list[Package] = []
+
+    @model_validator(mode="after")
+    def _arch_empty(self) -> "Arch":
+        """Validate arch."""
+        if self.packages == [] and self.source == []:
+            raise ValueError("At least one field ('packages', 'source') must be set in every arch.")
+        return self
+
+
+class RedhatRpmsLock(BaseModel):
+    """
+    The class implements basic operations with specific lockfile format.
+
+    Input data comes from raw yaml stored as a dictionary.
+    """
+
+    # Top of the structure of the lockfile. Model is used for parsing the lockfile.
+    lockfileVersion: PositiveInt
+    lockfileVendor: str
+    arches: list[Arch]
+
+    @cached_property
+    def _uuid(self) -> str:
+        """
+        Generate short random string for internal repoid.
+
+        'repoid' key is not mandatory in the lockfile format. When not present,
+        fallback is set as (partly) random string containing _uuid.
+        """
+        return uuid.uuid4().hex[:6]
+
+    @property
+    def internal_repoid(self) -> str:
+        """Internal_repoid getter."""
+        return f"cachi2-{self._uuid}"
+
+    @property
+    def internal_source_repoid(self) -> str:
+        """Internal_source_repoid getter."""
+        return f"cachi2-{self._uuid}-source"
+
+    @field_validator("lockfileVersion")
+    def _version_redhat(cls, version: PositiveInt) -> PositiveInt:
+        """Evaluate whether the lockfile header matches the format specification."""
+        if version != 1:
+            raise ValueError(f"Unexpected value for 'lockfileVersion': '{version}'.")
+        return version
+
+    @field_validator("lockfileVendor")
+    def _vendor_redhat(cls, vendor: str) -> str:
+        """Evaluate whether the lockfile header matches the format specification."""
+        if vendor != "redhat":
+            raise ValueError(f"Unexpected value for 'lockfileVendor': '{vendor}'.")
+        return vendor
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -1,7 +1,7 @@
 from collections.abc import Iterable
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import Callable
+from typing import Any, Callable
 
 from cachi2.core.errors import UnsupportedFeature
 from cachi2.core.models.input import PackageManagerType, Request
@@ -84,3 +84,11 @@ def _merge_outputs(outputs: Iterable[RequestOutput]) -> RequestOutput:
         environment_variables=env_vars,
         project_files=project_files,
     )
+
+
+def inject_files_post(*args: Any, **kwargs: Any) -> None:
+    """Do extra steps for package manager."""
+    # if there is a callback method defined within the particular package manager, run it
+    if hasattr(rpm, "inject_files_post"):
+        callback_method = getattr(rpm, "inject_files_post")
+        callback_method(*args, **kwargs)
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -6,7 +6,7 @@ from typing import Callable
 from cachi2.core.errors import UnsupportedFeature
 from cachi2.core.models.input import PackageManagerType, Request
 from cachi2.core.models.output import RequestOutput
-from cachi2.core.package_managers import gomod, npm, pip, yarn
+from cachi2.core.package_managers import gomod, npm, pip, rpm, yarn
 from cachi2.core.rooted_path import RootedPath
 from cachi2.core.utils import copy_directory
 
@@ -21,7 +21,9 @@ _package_managers: dict[PackageManagerType, Handler] = {
 
 # This is where we put package managers currently under development in order to
 # invoke them via CLI
-_dev_package_managers: dict[PackageManagerType, Handler] = {}
+_dev_package_managers: dict[PackageManagerType, Handler] = {
+    "rpm": rpm.fetch_rpm_source,
+}
 
 # This is *only* used to provide a list for `cachi2 --version`
 supported_package_managers = list(_package_managers)
--- a/tests/unit/package_managers/test_rpm.py
+++ b/tests/unit/package_managers/test_rpm.py
@@ -1,0 +1,30 @@
+from unittest import mock
+
+import pytest
+
+from cachi2.core.package_managers.rpm.redhat import RedhatRpmsLock
+
+
+class TestRedhatRpmsLock:
+    @pytest.fixture
+    def raw_content(self) -> dict:
+        return {"lockfileVendor": "redhat", "lockfileVersion": 1, "arches": []}
+
+    @mock.patch("cachi2.core.package_managers.rpm.redhat.uuid")
+    def test_internal_repoid(self, mock_uuid: mock.Mock, raw_content: dict) -> None:
+        mock_uuid.uuid4.return_value.hex = "abcdefghijklmn"
+        lock = RedhatRpmsLock.model_validate(raw_content)
+        assert lock._uuid == "abcdef"
+        assert lock.internal_repoid == "cachi2-abcdef"
+
+    @mock.patch("cachi2.core.package_managers.rpm.redhat.uuid")
+    def test_internal_source_repoid(self, mock_uuid: mock.Mock, raw_content: dict) -> None:
+        mock_uuid.uuid4.return_value.hex = "abcdefghijklmn"
+        lock = RedhatRpmsLock.model_validate(raw_content)
+        assert lock._uuid == "abcdef"
+        assert lock.internal_source_repoid == "cachi2-abcdef-source"
+
+    def test_uuid(self, raw_content: dict) -> None:
+        lock = RedhatRpmsLock.model_validate(raw_content)
+        uuid = lock._uuid
+        assert len(uuid) == 6
--- a/tests/unit/package_managers/test_rpm.py
+++ b/tests/unit/package_managers/test_rpm.py
@@ -1,14 +1,17 @@
 from pathlib import Path
 from unittest import mock
+from urllib.parse import quote
 
 import pytest
 import yaml
 
 from cachi2.core.errors import PackageManagerError, PackageRejected
+from cachi2.core.models.sbom import Component, Property
 from cachi2.core.package_managers.rpm import fetch_rpm_source
 from cachi2.core.package_managers.rpm.main import (
     DEFAULT_LOCKFILE_NAME,
     _download,
+    _generate_sbom_components,
     _resolve_rpm_project,
     _verify_downloaded,
 )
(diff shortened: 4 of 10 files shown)


onosek:
* Missing unittests
* Contains some # NOTE comments - they will be removed after clarification.
* Input file _source/rpms.lock.yaml_ won't be part of the merge. It serves just for your testing:
  `cachi2 fetch-deps --log-level DEBUG --dev-package-managers rpm -source source --output output`
  `cachi2 inject-files --log-level DEBUG output`

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,111 @@
...
+                    "package": False,
+                    "url": pkg.url,
+                    "size": pkg.size,
+                    "checksum": pkg.checksum,
+                }
+                os.makedirs(os.path.dirname(dest.path), exist_ok=True)
+
+            asyncio.run(async_download_files(files, get_config().concurrency_limit))
+
+    def verify_downloaded(self) -> None:
```
onosek, line 91: Should I rather raise the exception and quit instead of displaying warnings in this method?

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,112 @@
...
+from cachi2.core.config import get_config
+from cachi2.core.package_managers.general import async_download_files
+from cachi2.core.package_managers.rpm.base import RpmsLock
+from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+
+class Package(BaseModel):
+    repoid: str
```
eskultety, line 18: `repoid` must not be mandatory, one could have packages coming from 3rd party sites which don't have a repository, just the package. Naturally, this implementation fails on such premise:
```
arches.1.packages.0.repoid
  Field required [type=missing, input_value={'url': 'https://dl.fedor...18f5d1b', 'size': 97720}, input_type=dict]
```

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,204 @@
...
+    Repofile contains all arch's repoids. Skip sources dir that contains only SRPMs.
+    """
+    project_files: list[ProjectFile] = []
+    package_dir = path.join_within_root(DEFAULT_PACKAGE_DIR)
+    for subdir in os.listdir(package_dir):
+        if subdir == "sources":
+            continue
+        arch = subdir
+        log.debug(f"Preparing repofile for arch '{arch}'")
+        abspath = os.path.join(package_dir, arch, "cachi2.repo")
```
eskultety, line 178: Repo files should live separately from actual packages. IIUC the idea is to do something like `podman build -v <repos>:/etc/yum.repos.d ...` so I can imagine another subdirectory for the repos like `cachi2-output/deps/rpm/repos.d/cachi2.repo`

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,204 @@
...
+
+def _generate_repofiles(path: RootedPath) -> list[ProjectFile]:
+    """
+    Search structure and generate repofile for each arch in its dir.
+    Repofile contains all arch's repoids. Skip sources dir that contains only SRPMs.
+    """
+    project_files: list[ProjectFile] = []
+    package_dir = path.join_within_root(DEFAULT_PACKAGE_DIR)
+    for subdir in os.listdir(package_dir):
+        if subdir == "sources":
```
eskultety, line 174: Why are we calling this directory 'sources' when the common name among RPM distros is just 'source'?

eskultety (commented):
Some high-level comments that I could spot right away, will a deeper review once this is broken into more commits.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,204 @@
...
+            continue
+        arch = subdir
+        log.debug(f"Preparing repofile for arch '{arch}'")
+        abspath = os.path.join(package_dir, arch, "cachi2.repo")
+        template = ""
+        for repoid in os.listdir(os.path.join(package_dir, arch)):
+            localpath = os.path.join(package_dir, arch, repoid)
```
eskultety, line 181: This won't work because you set this to a fixed path that won't exist in the container. Since you can't have relative paths in the `file` URI protocol, you'll need to make use of our `--for-output-dir` CLI option, have a look at other places where this is used.

cachi2/core/package_managers/rpm/redhat.py
eskultety: @onosek can you please update https://github.com/containerbuildsystem/cachi2/discussions/465 with a note on `repoid` being optional? I don't see it in there though having been requested in the comments.

source/rpms.lock.yaml
```diff
@@ -0,0 +1,32 @@
+lockfileVersion: 1
```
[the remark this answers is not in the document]
brunoapimentel: I don't think this file belongs here. I think @onosek was using it only to test some builds. In any case, I agree, we should remove it from this PR.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,204 @@
...
+                f"pkg:rpm{'/' if vendor else ''}{vendor}/{name}@{version}{release}"
+                f"?arch={arch}{'&epoch=' if epoch else ''}{epoch}&download_url={download_url}"
+            )
+
+            components.append(
+                Component(
+                    name=name,
+                    version=version,
+                    purl=purl,
+                    # NOTE: add checksum into SBOM?
```
brunoapimentel, line 135: For now, we're not adding checksums in the SBOM in any of the supporte package managers, but we are reporting cases in which the checksum was missing from the lockfile. I believe we should do the same for RPMs.

Here's an example of how it looks like in the SBOM: https://github.com/containerbuildsystem/cachi2/blob/main/tests/integration/test_data/gomod_multiple_modules_missing_checksums/bom.json#L87-L101

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,200 @@
...
+from cachi2.core.utils import run_cmd
+
+log = logging.getLogger(__name__)
+
+
+DEFAULT_LOCKFILE_NAME = "rpms.lock.yaml"
+DEFAULT_PACKAGE_DIR = "deps/rpm"
+
+
+class LockfileFormat:
```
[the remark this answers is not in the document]
onosek: OK, I dropped the _LockfileFormat_ class that processed multiple formats and made a code able to process only one lockfile format. It lead to dropping _RpmsLock_ class too and I moved the common functionality to format specific _RedhatRpmsLock_.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,200 @@
...
+            repoid = entry
+            repodir = os.path.join(package_dir, arch, repoid)
+            createrepo(repoid, repodir)
+
+
+def createrepo(repoid: str, repodir: str) -> None:
+    log.info(f"Creating repository metadata for repoid '{repoid}': {repodir}")
+    cmd = ["/usr/bin/createrepo", repodir]
+    log.debug("$ " + shlex.join(cmd))
+    subprocess.run(cmd, check=True)
```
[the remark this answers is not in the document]
onosek: No problem with it. I just forgot to unify it with other calls. I modified it even before reading the comment.

source/rpms.lock.yaml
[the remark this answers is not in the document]
onosek: Yes, I stated it here:
https://github.com/containerbuildsystem/cachi2/pull/502#issuecomment-2008205468
that the file won't be in the final code. I only tried to find a way how to share an example input.

cachi2/core/package_managers/rpm/main.py
onosek: I am making _cachi2.repo_ file for every architecture so I can't store it in a single directory. Do you think creating corresponding arch subdirectories only with a single .repo file would be OK?

cachi2/core/package_managers/rpm/main.py
onosek: _sources_ vs. _source_. Actually, I don't know. _sources_ was in the design from the beginning. It would probably look better or would be more unified with the current public repository structures. But I am afraid to change it now. I think it would require changing not just the directory name, but also lockfile section name which would mean spreading this information and changing the existing testing data across the pipeline. The section name in lockfile and output directory are two things but having them different would look weird to me and possibly confusing to others. Does the _sources_ break anything?

onosek: Yes, in the thread you link sources and packages are stored together within common architecture.
I discussed it further and the result was published here:
https://github.com/containerbuildsystem/cachi2/discussions/465#discussioncomment-8747459
including the reasons.

cachi2/core/package_managers/rpm/main.py
brunoapimentel: I just noticed that in the tests was doing with hermetic builds, I accidentally was mounting the output dir in a way that it matched what I had in my local environment, that's why the builds succeeded. 

That is not the case in my early draft of the integration tests, and I'm getting failures such as:
```
STEP 2/3: RUN microdnf -y install 	vim
Updating and loading repositories:
updates-released                       ???% |   0.0   B/s |  -1.0   B |  ?     
 >>> Curl error (37): Couldn't read a file:// file for file:///tmp/pytest-of-bpim
```

source/rpms.lock.yaml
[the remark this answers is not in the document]
eskultety: >  that the file won't be in the final code. I only tried to find a way how to share an example input.

For future reference - if the motivation is to just share some test data without the intention of merging them to the repo, one can attach files to a PR :) which means that it'll always be tracked withing the context of the PR and we hence wouldn't "lose" the record of the example input data that was used for smoke testing the PR which we kinda will to some extent in this particular case (no big deal though).

cachi2/core/package_managers/rpm/redhat.py
eskultety: `unclassified_b328f2` sounds very random with 0 connection to the code responsible for creating it. I would suggest `cachi2<randoms_tring>` instead. That way the connection to our code base is clear at first glance.

cachi2/core/package_managers/rpm/main.py
eskultety: Citing from #465:
> - Splitting sources per arches and repoids. Despite we will probably deliver single source image (SRPMs might eventually be in one common directory) I prefer not to lose the information about the structure and it is more general for other eventual use-cases.

Unrelated to the matter at hand.

> - separate all sources to their own structure so createrepo doesn't have to process them when together with rpms. And when traversing the sources structure it is not so large.

I honestly don't understand ^this comment at all. Source RPMs are still RPMs belonging to a some repository. I don't understand why we're treating them differently. Again, hitting the issue of design being tailored to a specific use case. The fact that some project/product doesn't need to create a repository out of SRPMs and only needs them in the form of plain files to stuff them to some source container is irrelevant to cachi2 as a tool. We should treat all types of RPMs consistently, i.e. if they belong to a repository denoted by a `repoid` we need to put them along other repositories. Note that official repositories will clearly mark such repoids, e.g. `fedora-source`. Also note that `createrepo` and generation of the `.repo` file isn't something that the use case you're hinting at (i.e. creating source containers) will do, so the argument about not having to process SRPMs together with RPMs IMO doesn't stand. Quite the contrary, once you get to the use case of actually wanting a `.repo` file created by cachi2 you want a repository that contains everything the lockfile specifies which currently isn't the case since you keep the sources separate and omit them from the repo file generation - that is an inconsistent and non-intuitive approach.

onosek: I think changing the output structure is quite on us while changing the lockfile format (section sources --> source) has a wider impact on teams who could already generate lockfiles.
Technically, we can have the directory name and the lockfile section different, but I would preferably keep them both unchanged.

cachi2/core/package_managers/rpm/main.py
eskultety: > I am making cachi2.repo file for every architecture so I can't store it in a single directory. Do you think creating corresponding arch subdirectories only with a single .repo file would be OK?

No, we already have the `<arch>` directory, so better nested it there like this: `cachi2-output/deps/rpm/<arch>/repos.d/cachi2.repo`

cachi2/core/package_managers/rpm/main.py
eskultety: > Technically, we can have the directory name and the lockfile section different, but I would preferably keep them both unchanged.

Regardless of the lockfile format and contents, why is directory name being consistent with the lockfile so important? Cachi2 decides what the output looks like and I think consistency is key here which the placement of the source directory as proposed doesn't comply with. Changing other parts of the consuming pipeline to match this should be pretty straightforward while keeping the lockfile format unchanged.

onosek: Ok then. I will rename both.

cachi2/core/package_managers/rpm/redhat.py
onosek: OK, I will include _cachi2_ in the repoid and will add such comment to the .repo file.

cachi2/core/package_managers/rpm/main.py
onosek:
```
<output>/deps/rpm/<arch>/<repoid>/*.rpm
<output>/deps/rpm/<arch>/cachi2.repo OR <output>/deps/rpm/<arch>/repos.d/cachi2.repo
<output>/deps/rpm/<arch>/<repoid-source>/*.src.rpm
```
Let's have the structure above. OK. I will face the situation <repoid-source> will evaluate like "fedora-source-source". I think technically it is not an issue, it just looks weird.

cachi2/core/package_managers/rpm/redhat.py
onosek: I have realized, that <repoid> key in the lockfile can be missing both for packages and sources.
So I need two random strings for both categories, right?
- cachi2<random_string>
- cachi2<random_string>-source

onosek: Can I be sure that the <output> dir is empty or should I clear it completely?
Second/third run of the `fetch-deps` in my current implementation would generate multiple random <repoid> dirs with duplicate content. And during `inject-files` I am not able to distinguish which is fresh and `createrepo` is run over all of them.
I am probably able to implement some mechanism for reusing these random dirs before the download starts, but it unnecessarily complicates the code and it might not worth it.

brunoapimentel: > Can I be sure that the dir is empty or should I clear it completely?

Are you talking about the output dir (default: `cachi2-output`)? If so, it is a known issue, and I think it should be treated globally, not for a specific package manager. I'd say you can ignore this for know, and we can file an issue to deal with this.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,200 @@
...
+    for subdir in os.listdir(package_dir):
+        if subdir == "sources":
+            continue
+        arch = subdir
+        log.debug(f"Preparing repofile for arch '{arch}'")
+        abspath = os.path.join(package_dir, arch, "cachi2.repo")
+        template = ""
+        for repoid in os.listdir(os.path.join(package_dir, arch)):
+            localpath = os.path.join(package_dir, arch, repoid)
+            template += f"[{repoid}]\n"
```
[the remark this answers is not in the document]
onosek: I searched for some resources, but I haven't found a repofile with indentation.
I tried to write another version of the code as per your suggestion and it appeared to me more complicated than the current version (since then it has grown - it has to occasionally generate name="comment about random <repoid>").

cachi2/core/package_managers/rpm/redhat.py
onosek: Yes, the output dir.

eskultety: > If so, it is a known issue, and I think it should be treated globally, not for a specific package manager. I'd say you can ignore this for know, and we can file an issue to deal with this.

This will lead to very confusing results even during testing. This is a PoC so no harm will be done if @onosek implements a very trivial cleanup. My expectation would be that when creating either the `rpm` subdir or `<arch>` dir (depending on which is the most top-level directory created as part of the rpm module) and delete it if present which would indicate a repeated run. It is true we'll need to treat this on the global level, but with the addition of random directory names we need some band aid right now. @onosek looking at how the directory structure is created, e.g. [here](https://github.com/onosek/cachi2/blob/5165784787cca7ed9da4f66671d5fb3997615fe7/cachi2/core/package_managers/rpm/redhat.py#L67) you could add a very simple `Path.rmdir()` call to get rid of any existing directory before proceeding.

cachi2/core/package_managers/rpm/main.py
eskultety: > I will face the situation will evaluate like "fedora-source-source".

Why? I didn't mean to suggest to use "fedora-source-source" with `<repoid-source>` I only needed to visually separate the placeholder from plain `<repoid>`. So what I really expect in real world would be the exact repoid that the lockfile specifies for source repos, i.e. `fedora-source` being the name of the actual directory in this case.

onosek: Ok, I understood that you meant the "-source" suffix. Then I will have to generate repository metadata for every repoid including repoids for sources via `createrepo`. Because I don't know how to distinguish package repoid and source repoid. I think it is not a big issue - metadata for sources repoids just won't have any usage.

On the other hand, I will be able not to include source repoids into repofiles.

cachi2/core/package_managers/rpm/redhat.py
onosek: Sure, will remove the existing dir before downloading.

eskultety: > Sure, will remove the existing dir before downloading.

By the time this is merged you may not even need it: #509  we'll see :).

onosek: OK, I will remove the code that deletes dirs, then. Thx.

cachi2/core/package_managers/rpm/main.py
onosek: OK, then. Let's have source repoids included in the repofile.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,204 @@
...
+            solution=("Check 'lockfileVendor' and 'lockfileVersion' keys in the lockfile."),
+        )
+
+
+def fetch_rpm_source(request: Request) -> RequestOutput:
+    """Process the input lockfile, fetch packages and generate SBOM."""
+    components: list[Component] = []
+
+    _check_lockfile(request)
+    lockfile_name = request.source_dir.join_within_root(DEFAULT_LOCKFILE_NAME)
```
brunoapimentel, line 63: We're assuming that the lockfile will be present in the root of the repository, and also that there'll be a single one of them. But Cachi2's design allows the processing of multiple [input packages](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/yarn/main.py#L26-L30) (think of them as individual projects, such as a folder that is a Python project and contains a `requirements.txt` file), and also allows the user the define the relative path to such package.

See how this is done in [yarn](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/yarn/main.py#L26-L30). I assume we want to follow the same idea for RPMs.

brunoapimentel: Here's a draft of an integration test to cover this scenario: https://github.com/containerbuildsystem/cachi2/commit/cfe3813b9b0f885d02fab5adfa685d049e2b56b2

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,204 @@
...
+            name = rpm.headers.get("name", b"").decode()
+            version = rpm.headers.get("version", b"").decode()
+            release = rpm.headers.get("release", b"").decode()
+            arch = rpm.headers.get("arch", b"").decode()
+            download_url = quote(file_metadata["url"])
+
+            # https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm
+            # https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairsa
+            purl = (
+                f"pkg:rpm{'/' if vendor else ''}{vendor}/{name}@{version}{release}"
```
hlin, line 126: '-' is missing between {version} and {release}

hlin: Here's some unit tests https://github.com/containerbuildsystem/cachi2/compare/main...hlin:cachi2:prefecting-rpms-test

cachi2/core/package_managers/rpm/main.py
onosek: OK, what should be the Property value in this case? Is file name (rpm name) enough?

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,194 @@
...
+    package_dir = path.path.joinpath(DEFAULT_PACKAGE_DIR)
+    for arch in package_dir.iterdir():
+        if not arch.is_dir():
+            continue
+        log.debug(f"Preparing repofile for arch '{arch.name}'")
+        template = ""
+        for repoid in arch.iterdir():
+            if not repoid.is_dir():
+                continue
+            localpath = Path("$output_dir", DEFAULT_PACKAGE_DIR, arch.name, repoid.name)
```
hlin, line 168: What's $output_dir?

onosek: This relates to the comment:
https://github.com/containerbuildsystem/cachi2/pull/502#discussion_r1537251809
During `inject-files` it is evaluated to `--for-output-dir` argument value.

eskultety: More precisely, it's related to the `ProjectFile` template wrapping class.

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,168 @@
...
+        return f"cachi2-{self._uuid}"
+
+    @property
+    def internal_source_repoid(self) -> str:
+        """Internal_source_repoid getter."""
+        return f"cachi2-{self._uuid}-source"
+
+    def is_internal_repoid(self, repoid: str) -> bool:
+        """Repoid directory matches the internal repoid."""
+        return repoid in (self.internal_repoid, self.internal_source_repoid)
```
eskultety, line 81: None of these properties ought to be public accessors IMO. The first 2 should be converted to "private" methods (possibly static if you  won't keep uuid as an attribute) and the last one can be dropped as it only should be needed in `inject_files` and we can handle the check there.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,197 @@
...
+            "-q",
+            "--queryformat",
+            "%|EPOCH?{%{EPOCH}}:{}|",  # return "" when epoch is not set instead of "(None)"
+            file_path,
+        ]
+        epoch = run_cmd(cmd=["rpm", *rpm_args], params={})
+
+        with rpmfile.open(file_path) as rpm:
+            vendor = rpm.headers.get("vendor", b"").decode()
+            vendor = quote(vendor.lower())
```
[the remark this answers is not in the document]
onosek: Yes, `quote` currently does a substitution.
`"Fedora Project" --> "fedora%20project"`
Did you mean a different transformation?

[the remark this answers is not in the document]
eskultety: Nope, missed that fact, thank you, all good.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -15,6 +19,8 @@
 
 
 def fetch_rpm_source(request: Request) -> RequestOutput:
+    """Process the input lockfile, fetch packages and generate SBOM."""
```
eskultety, line 25: This doc string should have been part of commit https://github.com/containerbuildsystem/cachi2/pull/502/commits/58bd8eb54b289c98951f7284b57b051fbb17ff00

cachi2/core/package_managers/rpm/redhat.py
onosek: > and we can handle the check there.
Do you mean not to pass the lockfile object anymore and not to use `is_intetnal_repoid`? And recognize an internal repoid from its specific format?

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,168 @@
...
+
+class Root(BaseModel):
+    """Top of the structure of the lockfile. Used for parsing the lockfile."""
+
+    lockfileVersion: PositiveInt
+    lockfileVendor: str
+    arches: list[Arch]
+
+
+class Header(BaseModel):
```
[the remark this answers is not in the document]
onosek: Originally the `Header` was part of the format detection mechanism (code was prepared for multiple formats).
Since then the mechanism has been removed, but I kept `Header` because I was able to load only two (sufficient) items to verify I have the right format before loading the whole lockfile.

cachi2/core/package_managers/rpm/redhat.py
eskultety: Yes.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -58,14 +64,65 @@ def fetch_rpm_source(request: Request) -> RequestOutput:
...
+        components=components,
         environment_variables=[],
         project_files=[],
     )
 
 
+def _generate_sbom_components(components: list[Component], files_metadata: dict) -> None:
+    """Fill the component list with the package records."""
+    for file_path, file_metadata in files_metadata.items():
+        # iterate only through package metadata (skip sources metadata)
```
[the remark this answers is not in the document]
onosek: I was convinced we were not going to include SRPMs. I might be wrong.
So should we include every package into SBOM?
I won't need the https://github.com/containerbuildsystem/cachi2/pull/502#discussion_r1551494775 in this case.

cachi2/core/package_managers/rpm/redhat.py
onosek: OK, you answered it elsewhere:
https://github.com/containerbuildsystem/cachi2/pull/502#discussion_r1551770780

cachi2/core/package_managers/rpm/main.py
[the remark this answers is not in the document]
eskultety: Well, just like we originally didn't create repo metadata for SRPMs, this is IMO another example of not treating both consistently. The reason for that is that we should not try predicting how people use the either of them and hence report whatever we get in SBOM. I think what I said earlier to filter out duplicates certainly applies, but that can be done as a follow up, so you don't need to bother with the deduplication right now actually.

cachi2/core/package_managers/rpm/main.py
onosek: Yes, I did it intentionally. I didn't want to modify this doc string in more commits. I put it in when the doc string was final.
I wasn't prepared for the strict requirement of having clean linters and unittests after every commit despite the commits will be pushed together. I agree original unittests had to be fixed overall, that was my fault.

eskultety: > Yes, I did it intentionally. I didn't want to modify this doc string in more commits. I put it in when the doc string was final.

No need, since we know upfront what the purpose of the method is we can put the doc string in there right away.

> ...despite the commits will be pushed together

What do you mean by pushed together? We don't do squash-merging, so the commits would appear in the history exactly as proposed, so whenever one would run a bisect they'd get false positives on unittests and whatnot stemming from the fact that not every commit passed the unit test suite (let me also state that we have already mistakenly broken this, but in general we're trying to avoid doing that again).

cachi2/core/package_managers/rpm/redhat.py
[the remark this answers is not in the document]
eskultety: > Since then the mechanism has been removed, but I kept Header because I was able to load only two (sufficient) items to verify I have the right format (and thus print out relevant error message) before loading the whole lockfile.

Well, we already parsed the whole YAML, so validation done only on `Header` as you propose elsewhere in the code is kinda pointless then, you can validate the whole lockfile for structure and then validate individual fields, it's what we do in the rest of the code base.

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    The class implements basic operations with specific lockfile format.
+
+    Input data comes from raw yaml stored as a dictionary.
+    """
+
+    # parsed structure of the lockfile
+    _lockfile: Root
+    # metadata of file records found in the lockfile (for verification and SBOM generation)
+    _files_metadata: dict[Path, Any] = {}
```
eskultety, line 64: This attribute seems to be just duplication of data we already hold via (currently via the `_lockfile` attribute, which will go away).

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,168 @@
...
+        """Parse the whole lockfile with pydantic and return whether \
+        the parsing was successful and format of the lockfile is correct."""
+        try:
+            self._lockfile = Root(**self.data)
+        except ValidationError as e:
+            log.error(str(e))  # log defail of the validation error
+            return False
+        return bool(self._lockfile)
+
+    def download(self, output_dir: Path) -> None:
```
eskultety, line 107: This method is not supposed to be part of a lockfile, that is incorrect code design, this would be much better suited in either `_fetch_rpm_source` or maybe even something like `_resolve_rpm_project`  to keep the fetch code clean and short (have a look at https://github.com/containerbuildsystem/cachi2/blob/8110f954b76fa2da11845fd7a36bc2b12d16ec0f/cachi2/core/package_managers/yarn/main.py#L92).

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        Download packages mentioned in the lockfile.
+
+        Go through the parsed lockfile structure and find all RPM and SRPM files.
+        Create a metadata structure indexed by destination path used
+        for later verification (size, checksum) after download.
+        Prepare a list of files to be downloaded, and then download files.
+        """
+        for arch in self._lockfile.arches:
```
eskultety, line 116: Looping over the model kills the whole purpose of declaring this class as a `UserDict`, so this loop needs be over the RedhatRpmsLock dict `arches` subdict.

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,168 @@
...
+            log.debug(f"Downloading files for '{arch.arch}' architecture.")
+            # files per URL for downloading packages & sources
+            files: dict[str, Union[str, PathLike[str]]] = {}
+            for pkg in arch.packages:
+                repoid = self.internal_repoid if pkg.repoid is None else pkg.repoid
+                dest = output_dir.joinpath(arch.arch, repoid, Path(pkg.url).name)
+                files[pkg.url] = str(dest)
+                self._files_metadata[dest] = {
+                    "package": True,  # RPM
```
eskultety, line 125: with this boolean gone as discussed, there's now nothing unique about this `_files_metadata` class attribute that wouldn't already be contained within the dictionary already, we just need to use it.

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,168 @@
...
+                    "package": False,  # not package ~ SRPM
+                    "url": pkg.url,
+                    "size": pkg.size,
+                    "checksum": pkg.checksum,
+                }
+                Path.mkdir(dest.parent, parents=True, exist_ok=True)
+
+            asyncio.run(async_download_files(files, get_config().concurrency_limit))
+
+    def verify_downloaded(self) -> None:
```
eskultety, line 146: same as with `download`, this method is not supposed to live within the lockfile, but elsewhere, e.g. a new `_resolve_rpm_project` function.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -28,6 +29,36 @@ def fetch_rpm_source(request: Request) -> RequestOutput:
...
 
+        redhat_rpms_lock = RedhatRpmsLock(raw_content)
+        log.debug("Checking format '{}'".format(redhat_rpms_lock.__class__.__name__))
+        if not redhat_rpms_lock.match_format():
+            raise PackageRejected(
+                f"Rpm lockfile '{DEFAULT_LOCKFILE_NAME}' format is not supported'.",
+                solution=("Check 'lockfileVendor' and 'lockfileVersion' keys in the lockfile."),
+            )
+
+        log.debug("Lockfile content validation")
```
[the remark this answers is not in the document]
eskultety: Disregard this, with my latest comments this will be completely superseded by my suggestion in: https://github.com/containerbuildsystem/cachi2/pull/502#discussion_r1554154284

cachi2/core/package_managers/rpm/main.py
```diff
@@ -69,7 +71,7 @@ def fetch_rpm_source(request: Request) -> RequestOutput:
     return RequestOutput.from_obj_list(
         components=components,
         environment_variables=[],
-        project_files=[],
+        project_files=_generate_repofiles(request.output_dir, redhat_rpms_lock),
```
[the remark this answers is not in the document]
onosek: Yes, in the original solution, I used lockfile data for generating the .repo file as well. For example for identification of internal repoids. Now the need was broken with another comment and I can generate the .repo files during `inject-files`.

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,168 @@
...
+        """Use metadata structure with file sizes and checksums for verification \
+        of downloaded packages and sources."""
+        log.debug("Verification of downloaded files has started.")
+        # check file size and checksum of downloaded files
+        for file_path, file_metadata in self._files_metadata.items():
+            if file_metadata["size"] is not None:  # size is optional
+                if file_path.stat().st_size != file_metadata["size"]:
+                    raise RuntimeError(
```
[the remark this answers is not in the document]
onosek: I realized, that `PackageRejected` exception is shown to the user at the end. `RuntimeError` is caught in _main.py_ and converted to _cachi2_ exception. Should I replace such exceptions anyway? I think the current solution saves a few lines because I have a common exception and I just extend the specific message.

cachi2/core/resolver.py
```diff
@@ -84,3 +84,9 @@ def _merge_outputs(outputs: Iterable[RequestOutput]) -> RequestOutput:
         environment_variables=env_vars,
         project_files=project_files,
     )
+
+
+def process_packages(path: Path) -> None:
```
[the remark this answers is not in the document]
onosek: Did you mean `typer.Typer()` solution with decorated callbacks or some manual approach? I tried the first, but the mechanism didn't work in my code - callbacks were not executed.

cachi2/core/package_managers/rpm/redhat.py
onosek: _internal_repoid_ / _internal_source_repoid_ are currently used in the `download` method. You recommended to move `download` out from `RedhatRpmsLock`. And then, I would have to call these (private) methods outside of object. I would speak for keeping them public.

cachi2/core/package_managers/rpm/redhat.py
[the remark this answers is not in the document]
brunoapimentel: Yeah, I noticed these exceptions were being treated. I think it would be slightly clearer if you just raised the PackageRejected from here, and that would also allow you to customize each error message individually.

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,168 @@
...
+                    )
+
+            if file_metadata["checksum"] is not None:  # checksum is optional
+                alg, digest = file_metadata["checksum"].split(":")
+                method = getattr(hashlib, alg.lower(), None)
+                if not method:
+                    raise RuntimeError(f"Unsupported hashing algorithm '{alg}' for '{file_path}'")
+                h = method(usedforsecurity=False)
+                with open(file_path, "rb") as f:
+                    for chunk in iter(lambda: f.read(4096), b""):
```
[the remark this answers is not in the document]
eskultety: > Maybe we could add a new function to the [checksum](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/checksum.py) module that verifies a file based on a single hash

@brunoapimentel agreed, can this be left for a potential follow up?

cachi2/core/resolver.py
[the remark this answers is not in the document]
eskultety: `typer.Typer` interface won't likely be an option here, even if it were, that's future optimization. For now, I really meant some simple interfacing from CLI through the resolver module to individual package managers to execute an inject files post callback if defined. Truly, what I meant was the simplest solution that would be both straightforward and generic enough to be potentially extended in the future.

cachi2/core/package_managers/rpm/redhat.py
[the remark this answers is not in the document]
brunoapimentel: Most definitely!

cachi2/core/package_managers/rpm/redhat.py
onosek: Not sure what the comment is related to:

* _files_metadata
We have the data in the loaded/parsed structure from lockfile. But `_files_metadata` (now `metadata`) is organized differently (the main key is the package path). It makes iterating through items much easier (downloading, verification after download, SBOM generation). Otherwise, for example, I would have to construct paths multiple times (including testing whether the path contains internal repoid). I prefer to keep metadata.

* _lockfile
Yes, it will disappear.

eskultety (commented):
Really mostly minor nitpicks related to wording of log messages, but other than that, this is working, doesn't crash in between commits; the commits are structured well. Anything else can be done in follow up patches as optimization/refactoring.Overall, this is looking really good, just waiting for the unit tests.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,290 @@
...
+            if digest != h.hexdigest():
+                raise_exception(f"Unmatched checksum of '{file_path}' != '{digest}'")
+
+
+def _generate_sbom_components(files_metadata: dict[Path, Any]) -> list[Component]:
+    """Fill the component list with the package records."""
+    components: list[Component] = []
+    for file_path, file_metadata in files_metadata.items():
+        query_format = (
+            "%{NAME}\n"  # all nvra macros should be present/mandatory in rpm
```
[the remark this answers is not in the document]
eskultety: Scratch this, we don't need to dump it the same way as we get it from rpm, we can customize the message.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -214,3 +215,76 @@ def _generate_sbom_components(files_metadata: dict[Path, Any]) -> list[Component
...
+            content += "gpgcheck=1\n"
+            if repoid.name.startswith("cachi2-"):  # repoid directory matches the internal repoid
+                content += (
+                    "name=Generated repository containing all packages unaffiliated "
+                    "with any official repository\n"
+                )
+        if content:
+            repo_file_path = arch.joinpath("repos.d", "cachi2.repo")
+            if repo_file_path.exists():
+                log.info(f"Overwriting {repo_file_path}")
```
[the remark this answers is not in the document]
onosek: The idea comes from:
cachi2/interface/cli.py:334
where it is an info message.
I thought the deleting was related to `fetch-deps`. This is related to `inject-files` and here, I think, we could have this situation.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,290 @@
...
+    components: list[Component] = []
+    for file_path, file_metadata in files_metadata.items():
+        query_format = (
+            "%{NAME}\n"  # all nvra macros should be present/mandatory in rpm
+            "%{VERSION}\n"
+            "%{RELEASE}\n"
+            "%{ARCH}\n"
+            # vendor and epoch are optional in rpm file and in PURL as well
+            "%|VENDOR?{%{VENDOR}}:{}|\n"  # return "" when vendor is not set instead of "(None)"
+            "%|EPOCH?{%{EPOCH}}:{}|\n"  # return "" when epoch is not set instead of "(None)"
```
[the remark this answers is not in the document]
onosek: I was thinking about it. I tried to anticipate the situation when a developer who will come in future will just do a copy&paste :-).

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,290 @@
...
+                    "Check the correct format and whether any keys are missing in the lockfile."
+                ),
+            )
+
+        log.debug("Checking format '{}'".format(redhat_rpms_lock.__class__.__name__))
+        if not redhat_rpms_lock.match_format():
+            raise PackageManagerError(
+                f"Rpm lockfile '{DEFAULT_LOCKFILE_NAME}' format is not recognized'.",
+                solution=("Check 'lockfileVendor' and 'lockfileVersion' keys in the lockfile."),
+            )
```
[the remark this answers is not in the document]
onosek: OK, I had such validators in some early version of this code. They were removed since then.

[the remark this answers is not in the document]
eskultety: I can only see something related that was dropped here: https://github.com/containerbuildsystem/cachi2/compare/5165784787cca7ed9da4f66671d5fb3997615fe7..8722b59f9aff86bcab9b8c921b6914b49749492b but I don't know why the concept was dropped completely. Also, AFAICT it was only related to the Arch model, but the model describing the lockfile, am I mistaken?

cachi2/core/package_managers/rpm/main.py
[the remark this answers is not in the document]
eskultety: > I thought the deleting was related to fetch-deps. This is related to inject-files and here, I think, we could have this situation.

True, my bad then.

>  cachi2/interface/cli.py:334

Well, IMO incorrect on its own, but since there's a precedent, you can leave this as is.

cachi2/core/package_managers/rpm/main.py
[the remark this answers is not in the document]
eskultety: Fair enough.

cachi2/core/package_managers/rpm/main.py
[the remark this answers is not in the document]
onosek: No, you are right.

brunoapimentel (commented):
Leaving my round of reviews, they're mostly nitpicks and improvement ideas (all properly identified). Please, don't bother with them too much, let's focus on the few comments that are not nitpicks.

Congrats on the work here @onosek @eskultety. The code is looking much more organized and concise than the last time I reviewed it :muscle:.

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,79 @@
...
+    source: list[Package] = []
+
+
+class Lockfile(BaseModel):
+    """Top of the structure of the lockfile. Model is used for parsing the lockfile."""
+
+    lockfileVersion: PositiveInt
+    lockfileVendor: str
+    arches: list[Arch]
+
```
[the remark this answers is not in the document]
onosek: It is, I already noticed it and removed it.

cachi2/core/resolver.py
```diff
@@ -84,3 +84,11 @@ def _merge_outputs(outputs: Iterable[RequestOutput]) -> RequestOutput:
         environment_variables=env_vars,
         project_files=project_files,
     )
+
+
+def inject_files_post(*args: Any, **kwargs: Any) -> None:
+    """Do extra steps for package manager."""
+    # if there is a callback method defined within the particular package manager, run it
+    if hasattr(rpm, "inject_files_post"):
```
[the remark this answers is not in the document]
onosek: Iterate sounds easy but I didn't find an easy method. For example, I tried to iterate through `sys.modules`. It looked quite ugly :-).
In the end, I end up with the hardcoded `hasattr(rpm, ...)`.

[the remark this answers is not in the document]
eskultety: Let's leave this for future improvement, this is simple enough and I like it that way - as for extending this for other package managers I already have an idea that should be generic enough, but that is most certainly out of scope for this.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,301 @@
...
+    """Fill the component list with the package records."""
+    components: list[Component] = []
+    for file_path, file_metadata in files_metadata.items():
+        query_format = (
+            # all nvra macros should be present/mandatory in RPM
+            "%{NAME}\n"
+            "%{VERSION}\n"
+            "%{RELEASE}\n"
+            "%{ARCH}\n"
+            # vendor and epoch are optional in RPM file and in PURL as well
```
eskultety, line 179: nitpick: can we optically split these lines that have a commentary attached?

cachi2/core/package_managers/rpm/redhat.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _arch_empty(arch: Arch) -> Arch:
+    """Validate arch."""
+    if arch.packages == [] and arch.source == []:
+        raise ValueError("At least one field ('packages', 'source') must be set in every arch.")
+    return arch
+
+
+MyArch = Annotated[Arch, AfterValidator(_arch_empty)]
```
eskultety, line 36: Is there a reason why we need this as a type validator instead of a model validator? IOW, this seems to work as expected and doesn't lead to this Annotated wildness:
```
    @model_validator(mode="after")
    def _arch_empty(self) -> "Arch":
        """Validate arch."""
        if self.packages == [] and self.source == []:
            raise ValueError("At least one field ('packages', 'source') must be set in every arch.")
        return self
```

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,301 @@
...
+        ]
+        rpm_fields = run_cmd(cmd=["rpm", *rpm_args], params={})
+        rpm_fields_list = rpm_fields.split("\n")
+        rpm_fields_debug = (
+            "name='{}', version='{}', release='{}', arch='{}', vendor='{}', epoch='{}'".format(
+                *rpm_fields_list
+            )
+        )
+        log.debug(f"RPM attributes for '{file_path}': {rpm_fields_debug}")
+        name, version, release, arch, vendor, epoch = rpm_fields_list
```
eskultety, line 199: Can we unpack the fields right away and do it this way instead?
```
f"name='{name}', version='{version}', release='{release}', ...
```

tests/unit/package_managers/test_rpm.py
```diff
@@ -0,0 +1,279 @@
...
+        mock_source_dir = mock.Mock()
+        mock_source_dir.join_within_root.return_value.path.exists.return_value = False
+        _resolve_rpm_project(mock_source_dir, mock.Mock())
+    assert f"RPM lockfile '{DEFAULT_LOCKFILE_NAME}' missing, refusing to continue" in str(
+        exc_info.value
+    )
+
+
+@mock.patch(
+    "cachi2.core.package_managers.rpm.main.open",
```
eskultety, line 77: Can we pass an actual file in these tests and avoid mocking open and let yaml actually parse the file? These tests test the top-level function resolve so mocking everything isn't IMO acceptable.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,301 @@
...
+        )
+
+    lockfile_name = source_dir.join_within_root(DEFAULT_LOCKFILE_NAME)
+    log.info(f"Reading RPM lockfile: {lockfile_name}")
+    with open(lockfile_name) as f:
+        try:
+            yaml_content = yaml.safe_load(f)
+        except Exception as e:
+            # log detail of the validation error
+            log.error(str(e))
```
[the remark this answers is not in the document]
onosek: 'msg' is not enough sometimes. Otherwise, we could end up with the message "Field required" without any context.
At least a 'loc' key needs to be considered.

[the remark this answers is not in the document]
eskultety: > msg' is not enough sometimes. Otherwise, we could end up with the message "Field required" without any context.
> At least a 'loc' key needs to be considered.

True. Makes sense, I was about to suggest something more readable, but that would end up with too much work, so 'loc' sounds like a good tradeoff.

tests/unit/package_managers/test_rpm.py
```diff
@@ -0,0 +1,30 @@
+from unittest import mock
+
+import pytest
+
+from cachi2.core.package_managers.rpm.redhat import RedhatRpmsLock
+
+
+class TestRedhatRpmsLock:
```
eskultety, line 257: I'm missing some negative test cases for the class' field validators. Since Arch isn't consumed directly I'd add one more test case to this test class that also checks whether validation in the Arch model works as expected.

cachi2/core/package_managers/rpm/main.py
onosek: I did it, but later I found that putting simple spaces doesn't make _black_ happy. Converting `query_format` to a `list` doesn't help either. Other approaches I am thinking about are more complicated.

tests/unit/package_managers/test_rpm.py
onosek: I removed mock open from tests that focused on parsing yaml/lockfile. 
Remained 2 other tests containing mock open. I hope it is not an issue.

cachi2/core/package_managers/rpm/main.py
eskultety: (heavy sigh - black)...

cachi2/core/package_managers/rpm/main.py
```diff
@@ -0,0 +1,302 @@
...
+            raise PackageRejected(
+                f"RPM lockfile '{DEFAULT_LOCKFILE_NAME}' yaml format is not correct.",
+                solution=("Check correct 'yaml' syntax in the lockfile."),
+            )
+
+        log.debug("Validating lockfile.")
+        try:
+            redhat_rpms_lock = RedhatRpmsLock.model_validate(yaml_content)
+        except ValidationError as e:
+            # log detail of the validation error
```
eskultety, line 80: this commentary can now be dropped IMO since this will end up in an exception

eskultety (approved):
One last nitpick typo from me, but I think we're good to go -> Approve.

tests/unit/package_managers/test_rpm.py
```diff
@@ -1,8 +1,233 @@
...
+                            },
+                        ],
+                    },
+                ],
+            },
+            f,
+        )
+    with pytest.raises(PackageManagerError) as exc_info:
+        _resolve_rpm_project(rooted_tmp_path, rooted_tmp_path)
+    assert f"RPM lockfile '{DEFAULT_LOCKFILE_NAME}' format is not valid" in str(exc_info.value)
```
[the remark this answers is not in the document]
brunoapimentel: We should also cover a scenario of missing checksums with the unit tests. I see we have coverage missing here:
image

brunoapimentel (approved):

tests/unit/package_managers/test_rpm.py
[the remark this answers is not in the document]
onosek: I put the test ^ into the code
https://github.com/containerbuildsystem/cachi2/pull/502/commits/38ede91cdfe697a8fe901ecb2b9f55184229145a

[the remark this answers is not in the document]
eskultety: @brunoapimentel @onosek seems to have such test case in the meantime (very recently), I think that covers your request, right?

PR #504, opened 2024-03-21
add integration tests for pip wheels

slimreaper35:
testing repository:
https://github.com/cachito-testing/cachi2-pip-wheels


slimreaper35:
I am not sure about e2e test. The build uses wheels from `requirements.txt` and `requirements-build.txt`. If the allow binary option is set to false, the build will still pass because of `requirements-build.txt`.

Should I remove `requirements-build.txt` or/and add a scenario where the build fails because of missing `requirements-build.txt` ? What do you think?

tests/integration/test_pip.py
```diff
@@ -136,6 +175,27 @@ def test_pip_packages(
...
+        pytest.param(
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-pip-wheels",
+                ref="b3dea71a1ad51da90e1161cf71c85e0f7c3e292b",
+                packages=(
+                    {
+                        "type": "pip",
+                        "requirements_files": ["requirements.txt"],
+                        "requirements_build_files": ["requirements-build.txt"],
+                        "allow_binary": "true",
```
eskultety, line 188: I wonder if `allow_binary` being the only major change justifies a dedicated e2e test instead of enhancing the existing one to pull down wheels as well.

eskultety (commented):
I'm okay with the change with a small nitpick and a question on the e2e test.
Also, we should probably include an update to the existing references to the now-nonexistent https://github.com/cachito-testing/cachito-pip-without-metadata.git repository.

tests/integration/test_pip.py
ben-alkov: I think there's value in making sure that `"allow_binary": "false",` works as intended... but maybe there's another way to do it :shrug:

eskultety: We did that in the isolated scenarios, this is full-blown e2e test which can be tested in a complex manner IMO. I'm not objecting, I'm just not convinced, that's all.

tests/integration/test_pip.py
```diff
@@ -43,7 +43,10 @@
                     {
                         "path": "second_pkg",
                         "type": "pip",
-                        "requirements_files": ["requirements.txt", "requirements-extra.txt"],
+                        "requirements_files": [
+                            "requirements.txt",
+                            "requirements-extra.txt",
+                        ],
```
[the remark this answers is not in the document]
slimreaper35: I have enabled formatting on save in VS code ... sometimes I accidentally format the code...

tests/integration/test_pip.py
slimreaper35: Is it related to the question I asked above that it is worth having two scenarios when the build will succeed if `allow_binary` is set to `true` otherwise, the build will fail because of missing `requirements-build.txt` ?

Just to make sure, for users that do not want to specify build dependencies it works

tests/integration/test_pip.py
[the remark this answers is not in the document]
ben-alkov: Felt, this happens to me sometimes. If you use black or Ruff, there shouldn't be any changes to existing code.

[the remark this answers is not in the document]
slimreaper35: Only Ruff

tests/integration/test_pip.py
slimreaper35: > or write an app that depends on a package which only distributes as a wheel.

This covers another test in the repo, but it is not e2e.

slimreaper35:
I updated the test data so it reflects #508

taylormadore (commented):
Have we covered all of the test cases listed in the story description? Specifically the ones related to hashes and the one about testing a build that requires wheels?

> about testing a build that requires wheels

I thought the e2e test might cover this, but it appears to pass with or without allow_binary

tests/integration/test_data/pip_no_sdists/fetch_deps_sha256sums.json
```diff
@@ -0,0 +1,7 @@
+{
+  "pip/tensorflow-2.16.1-cp310-cp310-win_amd64.whl": "sha256:21a3c6d76a39f52754c389326f6bef8aef3c26b5bc89ca365add4a69483e569e",
```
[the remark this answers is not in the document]
slimreaper35: Well, all three wheels have 2.1 kB, I am not sure if I can find any smaller projects.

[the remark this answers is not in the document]
taylormadore: You're right, sorry I mixed up torch and tensorflow when I was looking at the manylinux wheels

slimreaper35:
> Have we covered all of the test cases listed in the story description?

I think so, except _hash checking is/isn't used_. I just can't find a distribution package, that does not report hashes.
If there are no hashes reported by PyPI and no hashes in the requirements.txt => we should download the package without verifying the hash.

ben-alkov:
> > Have we covered all of the test cases listed in the story description?
> 
> I think so, except _hash checking is/isn't used_. I just can't find a distribution package, that does not report hashes.

Yep, agreed. AFAUI, it's a baked-in part of the PyPi API - the only way you'd get a request reply without a checksum would be someone spoofing you (or *maybe* if you're not talking to PyPi but rather some other index, but teven then I'm not certain).

> If there are no hashes reported by PyPI and no hashes in the requirements.txt => we should download the package without verifying the hash
This should read "we should NOT download the package" - no hashes at all = no download.

slimreaper35:
"should" shouldn't be in the sentence, but that's how it works right now

tests/integration/test_pip.py
eskultety: >> or write an app that depends on a package which only distributes as a wheel.

> This covers another test in the repo, but it is not e2e.

I'm trying to understand where this all got stuck. The fact that we have a fetch test for a case with no sdists doesn't mean we shouldn't test e2e in a complex manner, IOW include e.g. tensorflow or pycaret in requirements for the test app and try building it.

eskultety:
@slimreaper35 I don't see any update to the e2e tests really, so what is your stance on https://github.com/containerbuildsystem/cachi2/pull/504#issuecomment-2036487710 as you haven't replied there really whether you agree or disagree with that argument and this PR then got stuck. From my POV in this case it makes sense having a single e2e test that would test both wheels and sdists at the same time, why? Because fetching is what we really only care about and since we've taken care of testing fetches in different scenarios now we just need to make sure that whatever we fetched can be built with.

tests/integration/test_pip.py
slimreaper35: So, can I drop the e2e branch altogether and make no-sdists as the e2e test even if there is no _requirements_build.txt_ file?

slimreaper35:
Update (finally):

- There is only one integration test with packages that have no wheels and `allow_binary` option is set to true
- There is only one e2e test with *tensorboard* package that have no sdists and again `allow_binary` option is set to true
  _Luckily I found a package that does not generate a huge _requirements.txt_ file... The pre-fetching with wheels should take only ~ 50 seconds_ Should I add test ensuring that pre-fetching will fail on `allow_binary_ option set to false ?

eskultety:
>       Should I add test ensuring that pre-fetching will fail on `allow_binary` option set to false ?

Yes we should have such a test. In fact, you had two in earlier revisions, but for some reason you dropped it here: https://github.com/containerbuildsystem/cachi2/compare/ce01911eb42de201fd4d9507ad9d3a46114c23ba..06d216f8de883042e36bbbe47f29a0e452ccf5f8#diff-e4d828af95a3a935bef8644268aee7ff4190648a95b7a401e91a65dbebb9a206L99, any particular reason for that? IIRC I even emphasized the importance of those isolated fail cases here: https://github.com/containerbuildsystem/cachi2/pull/504#discussion_r1550234513

tests/integration/test_pip.py
```diff
@@ -96,6 +96,30 @@
...
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="All dependencies fetched successfully",
+            ),
+            id="pip_no_wheels",
+        ),
+        pytest.param(
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-pip-wheels.git",
+                ref="eee59273542ee7d412fb359d471386b645cf166e",
```
slimreaper35, line 113: This is the same branch as for e2e test. It fails on pre-fetching `tensorboard` as intended.

eskultety (approved):

taylormadore (approved):

PR #507, opened 2024-03-26
Convert PipRequirement classmethods to staticmethods

### LINKED ISSUE #488: Convert a couple of PipRequiremen class methods into static methods
Both ` PipRequirement._assess_direct_access_requirement` and `PipRequirement._split_hashes_from_options` can be made static helper methods instead of class methods as they really have no real connection to the class.

ianrochapg:
PipRequirement had three classmethods with no real connection to the class, so in this commit I'm converting them to staticmethods and passing the necessary data as arguments.

PipRequirement._assess_direct_access_requirement
PipRequirement._adjust_direct_access_requirement
PipRequirement._split_hashes_from_options

Resolves #488 

### DIFF
--- a/cachi2/core/package_managers/pip.py
+++ b/cachi2/core/package_managers/pip.py
@@ -22,6 +22,7 @@ from typing import (
     Iterator,
     Literal,
     Optional,
+    Pattern,
     Sequence,
     Union,
     cast,
@@ -1282,7 +1283,9 @@ class PipRequirement:
         if is_direct_access:
             if direct_access_kind in ["url", "vcs"]:
                 requirement.kind = direct_access_kind
-                to_be_parsed, qualifiers = cls._adjust_direct_access_requirement(to_be_parsed)
+                to_be_parsed, qualifiers = cls._adjust_direct_access_requirement(
+                    to_be_parsed, cls.HAS_NAME_IN_DIRECT_ACCESS_REQUIREMENT
+                )
             else:
                 raise UnsupportedFeature(
                     f"Direct references with {direct_access_kind!r} scheme are not supported, "
@@ -1325,8 +1328,8 @@ class PipRequirement:
 
         return requirement
 
-    @classmethod
-    def _assess_direct_access_requirement(cls, line: str) -> tuple[Optional[str], bool]:
+    @staticmethod
+    def _assess_direct_access_requirement(line: str) -> tuple[Optional[str], bool]:
         """Determine if the line contains a direct access requirement.
 
         :param str line: the requirement line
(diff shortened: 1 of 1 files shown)


cachi2/core/package_managers/pip.py
```diff
@@ -1299,11 +1300,15 @@ def from_line(cls, line: str, options: list[str]) -> Optional["PipRequirement"]:
         qualifiers = {}
         requirement = cls()
 
-        direct_access_kind, is_direct_access = cls._assess_direct_access_requirement(line)
+        direct_access_kind, is_direct_access = cls._assess_direct_access_requirement(
+            line, cls.URL_SCHEMES, cls.VCS_SCHEMES
```
eskultety, line 1304: Looking at the code in more detail there's actually no reason for us to keep ^these as class attributes given they're only ever used in the static methods, so there's nothing preventing us from moving the class attributes to the respective static methods and define them as constants there. Same for `HAS_NAME_IN_DIRECT_ACCESS_REQUIREMENT` below.
Note that such an attribute move should be in a separate commit preceding this one converting the methods from class methods to static methods.

eskultety (commented):
> Resolves https://github.com/containerbuildsystem/cachi2/issues/488

It's best if ^this appears in the respective commit message rather than the PR description. That way the connection to an issue is tracked on the git history level rather than solely in GH UI.

cachi2/core/package_managers/pip.py
ianrochapg: Moving these inside the methods sounds great, but can you specify what you mean by "define them as constants"? Just so we're on the same page.

eskultety: In Python constants  == upper case variables, that's all there's to it :) .

ben-alkov: @eskultety; Is there an advantage to (currently) only compiling HAS_NAME_IN_DIRECT_ACCESS_REQUIREMENT once (when the the module containing the class is loaded)?

IOW, won't we otherwise be recompiling it on every access of `_assess_direct_access_requirement`?

(yeah, it's a micro-optimization...)

eskultety: > IOW, won't we otherwise be recompiling it on every access of _assess_direct_access_requirement?

@ben-alkov @ianrochapg Oh, missed that one, in that particular case it makes sense to keep passing it to the method as a `pattern` argument.

ben-alkov (approved):
LGTM

cachi2/core/package_managers/pip.py
```diff
@@ -1375,11 +1378,15 @@ def _assess_direct_access_requirement(cls, line: str) -> tuple[Optional[str], bo
 
         return direct_access_kind, True
 
-    @classmethod
-    def _adjust_direct_access_requirement(cls, line: str) -> tuple[str, dict[str, str]]:
+    @staticmethod
+    def _adjust_direct_access_requirement(
+        line: str, has_name_in_access_requirement: Pattern[str]
```
eskultety, line 1383: `has_name_in_access_requirement` is really a bad name for an argument if you consider that now we're technically able to accept any regex pattern here (not that we're going to do that).

ianrochapg: Would you guys consider `has_name_in_requirement_pattern` a good option? It might be a bit long, but I think it makes it pretty clear.

ben-alkov: It's early for me, but maybe `pkg_name_regex`? :shrug:

I mean, I recognize that it's bad practice to put data types in variable names, but this is more about what purpose it serves than what it contains...

taylormadore: It appears we're stuck on the classic "naming things is hard" problem :sweat_smile: 

As far as I can tell, this pattern doesn't have much to do with whether there is a name present or not :shrug: 
It's just a pattern to check for [direct references](https://peps.python.org/pep-0440/#direct-references).

Perhaps direct_ref_pattern?

taylormadore: I believe so :man_shrugging:

PR #508, opened 2024-03-27
feat(pkg-mgr:pip): add 'pip_package_binary' property...

ben-alkov:
...to PropertySet and sbom::PropertyName so it can be processed

Signed-off-by: Ben Alkov <ben.alkov@redhat.com># Maintainers will complete the following section

### DIFF
--- a/cachi2/core/models/property_semantics.py
+++ b/cachi2/core/models/property_semantics.py
@@ -32,6 +32,7 @@ class PropertySet:
     missing_hash_in_file: frozenset[str] = field(default_factory=frozenset)
     npm_bundled: bool = False
     npm_development: bool = False
+    pip_package_binary: bool = False
 
     @classmethod
     def from_properties(cls, props: Iterable[Property]) -> "Self":
@@ -40,6 +41,7 @@ class PropertySet:
         missing_hash_in_file = []
         npm_bundled = False
         npm_development = False
+        pip_package_binary = False
 
         for prop in props:
             if prop.name == "cachi2:found_by":
@@ -50,10 +52,18 @@ class PropertySet:
                 npm_bundled = True
             elif prop.name == "cdx:npm:package:development":
                 npm_development = True
+            elif prop.name == "cachi2:pip:package:binary":
+                pip_package_binary = True
             else:
                 assert_never(prop.name)
 
-        return cls(found_by, frozenset(missing_hash_in_file), npm_bundled, npm_development)
+        return cls(
+            found_by,
+            frozenset(missing_hash_in_file),
+            npm_bundled,
+            npm_development,
+            pip_package_binary,
+        )
 
     def to_properties(self) -> list[Property]:
         """Convert a PropertySet to a list of SBOM component properties."""
@@ -68,6 +78,9 @@ class PropertySet:
             props.append(Property(name="cdx:npm:package:bundled", value="true"))
         if self.npm_development:
             props.append(Property(name="cdx:npm:package:development", value="true"))
+        if self.pip_package_binary:
+            props.append(Property(name="cachi2:pip:package:binary", value="true"))
+
         return sorted(props, key=lambda p: (p.name, p.value))
 
     def merge(self, other: "Self") -> "Self":
@@ -78,4 +91,5 @@ class PropertySet:
             missing_hash_in_file=self.missing_hash_in_file | other.missing_hash_in_file,
             npm_bundled=self.npm_bundled and other.npm_bundled,
             npm_development=self.npm_development and other.npm_development,
+            pip_package_binary=self.pip_package_binary or other.pip_package_binary,
         )
--- a/cachi2/core/package_managers/pip.py
+++ b/cachi2/core/package_managers/pip.py
@@ -201,10 +201,16 @@ def fetch_pip_source(request: Request) -> RequestOutput:
 
 
 def _generate_properties(dependency: dict[str, str]) -> list[Property]:
+    properties = []
     if not dependency["hash_verified"]:
-        return [Property(name="cachi2:missing_hash:in_file", value=dependency["requirement_file"])]
-    else:
-        return []
+        properties.append(
+            Property(name="cachi2:missing_hash:in_file", value=dependency["requirement_file"])
+        )
+
+    if dependency["package_type"] == "wheel":
+        properties.append(Property(name="cachi2:pip:package:binary", value="true"))
+
+    return properties
 
 
 def _generate_purl_main_package(package: dict[str, Any], package_path: RootedPath) -> str:
@@ -1487,12 +1493,17 @@ def _process_req(
         if dpi.package_type == "sdist":
             _check_metadata_in_sdist(dpi.path)
 
+        download_info["package_type"] = dpi.package_type
+
     else:
         if require_hashes or req.kind == "url":
             hashes = req.hashes or [req.qualifiers.get("cachito_hash", "")]
             download_info["hash_verified"] = _hash_verify(
                 download_info["path"], list(map(_to_checksum_info, hashes))
             )
+
+        # "package_type" is *only* needed for PyPI deps
+        download_info["package_type"] = ""
 
     log.debug(
         "Successfully processed '%s' in path '%s'",
@@ -2143,6 +2154,7 @@ def _resolve_pip(
         {
             "name": dep["package"],
             "version": _version(dep),
+            "package_type": dep["package_type"],
             "type": "pip",
             "dev": dep.get("dev", False),
             "kind": dep["kind"],
--- a/cachi2/core/package_managers/pip.py
+++ b/cachi2/core/package_managers/pip.py
@@ -49,7 +49,8 @@ from cachi2.core.config import get_config
 from cachi2.core.errors import FetchError, PackageRejected, UnexpectedFormat, UnsupportedFeature
 from cachi2.core.models.input import Request
 from cachi2.core.models.output import EnvironmentVariable, ProjectFile, RequestOutput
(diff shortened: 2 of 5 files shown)


cachi2/core/package_managers/pip.py
```diff
@@ -1482,12 +1488,15 @@ def _hash_verify(path: Path, checksum_info: Iterable[ChecksumInfo]) -> bool:
...
 
+        download_info["package_type"] = dpi.package_type
+
     else:
         if require_hashes or req.kind == "url":
             hashes = req.hashes or [req.qualifiers.get("cachito_hash", "")]
             download_info["hash_verified"] = _hash_verify(
                 download_info["path"], list(map(_to_checksum_info, hashes))
             )
+            download_info["package_type"] = ""  # hack
```
ben-alkov, line 1499: Not really happy with this, but given that DPIs only exist for pypi deps, it doesn't make sense (to me) to add "", or "none", or some other null to the `DistributionPackageInfo::package_type` Literal

cachi2/core/package_managers/pip.py
```diff
@@ -1528,6 +1537,14 @@ def _process_pypi_req(
             )
         )
 
+    # hack
```
ben-alkov, line 1540: I'm not fond of this, but I haven't been able to come up with a more elegant, Pythonic way of doing same...

ben-alkov:
I need to add a couple of unit tests (probably just a couple of scenarios for existing tests) to cover new lines in 'pip.py' and 'property_semantics.py'.

cachi2/core/package_managers/pip.py
slimreaper35: Shouldn't be this `download_info["package_type"] = "sdist"` ?

(on the same level as `if ...` )

cachi2/core/package_managers/pip.py
```diff
@@ -89,11 +89,11 @@ class DistributionPackageInfo:
 
     name: str
     version: str
-    package_type: Literal["sdist", "wheel"]
     path: Path
     url: str
     is_yanked: bool
 
+    package_type: Literal["sdist", "wheel"] = "sdist"
```
slimreaper35, line 96: I am not sure if this is necessary since we are setting it anyway in
`_process_package_distributions`

ben-alkov: Agreed. Probably a relic of trying to figure out how to work around VCS/URL reqs not having this property.

cachi2/core/package_managers/pip.py
ben-alkov: No, otherwise we hit the situation where VCS/URL reqs **don't** have the "package_type" property _at all_.

ben-alkov: But now that I'm looking at it again, I think that VCS reqs aren't covered...

ben-alkov: > Shouldn't this be ... on the same level as `if`

I shouldn't review these comments first thing in the morning :roll_eyes: - you're absolutely correct

cachi2/core/package_managers/pip.py
ben-alkov: This section is better now, but maybe could be improved?

cachi2/core/package_managers/pip.py
ben-alkov: It shouldn't be "sdist" for VCS/URL reqs, because **they aren't sdists** - "sdist"/"wheel" only apply to downloads from PyPi.

The empty string is the "hack", but I don't know of a better workaround that wouldn't require a substantial refactor.

slimreaper35: ACK

cachi2/core/package_managers/pip.py
ben-alkov: I don't actually know the correct way to enable wheels, but you can set `allow_binary: bool = True` in  `input.py::PipPackageInput` (it's very close to the top of the class).

Then run a test scenario whose deps actually have wheels...

You could try cloning [StableCascade](https://github.com/Stability-AI/StableCascade) and using the [requirements.txt](https://gist.github.com/ben-alkov/e8540fe13a36b6fb9f195ffc99aedbd8) I generated for it - it has reqs with sdist-only, sdists+wheels, wheels-only, URL, **and** VCS, so it's a pretty good test. It does take a while to run :grin:

Then you can check the BOM...

cachi2/core/package_managers/pip.py
```diff
@@ -201,10 +201,16 @@ def fetch_pip_source(request: Request) -> RequestOutput:
 
 
 def _generate_properties(dependency: dict[str, str]) -> list[Property]:
+    properties = []
     if not dependency["hash_verified"]:
-        return [Property(name="cachi2:missing_hash:in_file", value=dependency["requirement_file"])]
-    else:
-        return []
+        properties.append(
+            Property(name="cachi2:missing_hash:in_file", value=dependency["requirement_file"])
```
taylormadore, line 207: This comment is not specific to your changes, but just about consistency between handling properties across different package managers: Is there a reason why we don't create a PropertySet here like we do in both gomod and npm?
https://github.com/containerbuildsystem/cachi2/blob/0564776e552e286d7f2346d70ed079b83a5756fe/cachi2/core/package_managers/gomod.py#L149
https://github.com/containerbuildsystem/cachi2/blob/0564776e552e286d7f2346d70ed079b83a5756fe/cachi2/core/package_managers/npm.py#L655

ben-alkov: I wasn't aware that that was even a thing... I looked at 'property_semantics.py' and 'sbom.py'.

I'll change both the pip properties to use `PropertySet`.

cachi2/core/package_managers/pip.py
ben-alkov: Thanks to Bruno for enlightening me about cachi2's CLI

```bash
cachi2 fetch-deps  '{"type": "pip", "allow_binary": true}' --output ~/temp/out/StableCascade --source ~/Source/StableCascade
```

cachi2/core/package_managers/pip.py
ben-alkov: Fixed

ben-alkov:
Ready for review. Still needs unit tests to increase coverage, but I believe the implementation is complete and correct.

cachi2/core/package_managers/pip.py
taylormadore: It's late and I'm not as familiar with this feature, so perhaps I'm missing something here, but is the de-duplication of the Component property for binaries the issue you are trying to solve?

If the property should be true if there is a wheel for any component being de-duplicated, can't we just update PropertySet.merge to do a logical OR instead of AND for the binary property?

ben-alkov: > can't we just update PropertySet.merge to do a logical OR instead of AND for the binary property

Now that I know what a PropertySet is, I can look into it :wink:

cachi2/core/package_managers/pip.py
slimreaper35: Btw. why is the `package_type` attribute moved at the end of the class ?

ben-alkov: I know I had a reason, but honestly, I can't remember at this point.

ben-alkov:
Unit tests are done, new line coverage is %100 :tada:.

cachi2/core/package_managers/pip.py
slimreaper35: still no idea about this one ?

cachi2/core/models/property_semantics.py
```diff
@@ -78,4 +91,5 @@ def merge(self, other: "Self") -> "Self":
             missing_hash_in_file=self.missing_hash_in_file | other.missing_hash_in_file,
             npm_bundled=self.npm_bundled and other.npm_bundled,
             npm_development=self.npm_development and other.npm_development,
+            pip_package_binary=self.pip_package_binary or other.pip_package_binary,
```
taylormadore, line 94: Should we enhance the test_property_semantics.py::test_merge unit test with this new case?

ben-alkov: Probably.

ben-alkov: ...as well as `test_merge_component_properties`

ben-alkov: Fixed

tests/integration/test_data/pip_e2e_test/bom.json
```diff
@@ -1,134 +1,134 @@
 {
-  "bomFormat": "CycloneDX",
```
taylormadore, line 2: Any idea why this file was changed? Seems like all whitespace changes to me

cachi2/core/package_managers/pip.py
taylormadore: I'm curious about this as well. The change of order in the inputs to the DPI class resulted in quite a few other changes to the tests, etc. If it isn't necessary for some functional reason, I think we should keep it as it was to reduce the noise

tests/integration/test_data/pip_e2e_test/bom.json
ben-alkov: Very weird. I kept getting integration test failures, so I regenerated the file, but obviously messed it up...

ben-alkov: Fixed this, and the one below.

slimreaper35 (approved):

taylormadore (approved):

PR #509, opened 2024-03-28
interface: cli: Remove the top-level output directory before fetching

eskultety:
We keep reusing the old output directory hierarchy with repeated fetch runs, overwriting all relevant files. Right now it's all fine because all the files have predictable (pretty much fixed) names, however if we create files/directories with any randomness in their names, we'll be just creating more and more noise with repeated runs. Since we're rewriting all files for the processed package managers anyway (for now), i.e. not reusing data in between runs, it would be better if we first cleaned up the destination if it exists and only then proceed with the fetch.

Note that failing to process a request would be the preferred approach if the output directory existed, because we can't be sure that the existing directory we're being pointed to was indeed created by us so we may end up deleting other legitimate data, however, given that such logic would lead to regressive behaviour since we already happily overwrite anything the user points us at, so the new behaviour isn't any worse than it already is.

### DIFF
--- a/cachi2/interface/cli.py
+++ b/cachi2/interface/cli.py
@@ -2,6 +2,7 @@ import functools
 import importlib.metadata
 import json
 import logging
+import shutil
 import sys
 from pathlib import Path
 from typing import Any, Callable, Optional
@@ -249,6 +250,11 @@ def fetch_deps(
             "flags": combine_option_and_json_flags(input.flags),
         },
     )
+
+    deps_dir = output / "deps"
+    if deps_dir.exists():
+        log.debug(f"Removing existing deps directory '{deps_dir}'")
+        shutil.rmtree(deps_dir, ignore_errors=True)
 
     request_output = resolve_packages(request)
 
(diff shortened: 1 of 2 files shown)


ben-alkov:
\*faints\* all that work I did on checksumming existing pypi artifacts instead of re-downloading... :smiling_face_with_tear:

ben-alkov:
LGTM, but I don't want to ACK when we still have concerns about possibly deleting important user directories (in the case where a user points `--output` at something they might regret).

taylormadore (approved):

ben-alkov (approved):
LGTM

brunoapimentel:
LGTM, but I think we can add unit tests just to get coverage of the new functionality. I went ahead and wrote a [basic unit test](https://github.com/brunoapimentel/cachi2/commit/149fa64de65e72c41c296afc7834bd7e5f37584f) for this, you can use it if you think it's good enough.

In any case, I will go ahead and leave my LGTM so you can merge this as soon as you're back.

brunoapimentel (approved):

eskultety:
> LGTM, but I think we can add unit tests just to get coverage of the new functionality. I went ahead and wrote a [basic unit test](https://github.com/brunoapimentel/cachi2/commit/149fa64de65e72c41c296afc7834bd7e5f37584f) for this, you can use it if you think it's good enough.
> 
> In any case, I will go ahead and leave my LGTM so you can merge this as soon as you're back.

Applied, thanks!

PR #522, opened 2024-04-18
rpm PoC: Allow specifying DNF options via CLI input JSON 

eskultety:
Introduce a new input JSON attribute 'options' for the PoC rpm package
manager allowing consumers to pass arbitrary DNF repo configuration
options that would be applied later through the inject-files command.

The general schema for the input JSON is as follows:

```
{
    type: <package_manager>
    path: <relative_path>
    options: {
        <namespace>: {
            <key1>..<keyN>: Union[str, Any]<val1>
        }
    }
}
```

which translated to DNF repo options for RPM might then look like:

```
{
    type: rpm
    path: .
    options: {
        dnf: {
             repoid1: {gpgcheck: 0, deltarpm: 1},
             repoid2: {fastestmirro: 1, sslverify: 0}
        }
    }
}
```

which in turn would end up being dumped to the `.build-config.json` similarly to this:
```
{
  "environment_variables": [],
  "project_files": [],
  "options": {
    "rpm": {
      "dnf": {
        "releases": {
          "gpgcheck": 0,
          "enabled": 0,
          "type": "yum"
        }
      }
    }
  }
```
and finally the resulting generated `.repo` file might look like:
```
[releases]
gpgcheck = 0
enabled = 0
type = yum
baseurl = file:///etc/yum.repos.d/deps/rpm/x86_64/releases
```

_Notes:_
- worth mentioning is that the way this is currently designed is that `options` are only ever formatted in the build config if some were actually set on the CLI (rather than being formatted as `options: {}`) so that this would be transparent to other package manager for which we don't intend to support options at the moment; this was all achieved by setting `exclude_none=True` globally for `buildconfig.model_dump_json`
- the reason why there's an additional level of nesting in the `.build-config.json` namespaced by `"rpm"` is due to the fact that it's possible to specify more than one package with the input JSON on CLI, thus theoretically possible to provide various options for various package managers (if ever turned on) so we'd need to know which options relate to which package manager type which wouldn't otherwise be possible from current build config contents.

### DIFF
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -1,5 +1,17 @@
 from pathlib import Path
-from typing import TYPE_CHECKING, Annotated, Callable, ClassVar, Literal, Optional, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Annotated,
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    List,
+    Literal,
+    Optional,
+    TypeVar,
+    Union,
+)
 
 import pydantic
 
@@ -65,6 +77,60 @@ class _PackageInputBase(pydantic.BaseModel, extra="forbid"):
         return check_sane_relpath(path)
 
 
+class _DNFOptions(pydantic.BaseModel, extra="forbid"):
+    """DNF options model.
+
+    DNF options can be provided via 2 'streams':
+        1) global /etc/dnf/dnf.conf OR
+        2) /etc/yum.repos.d/.repo files
+
+    Config options are specified via INI format based on sections. There are 2 types of sections:
+        1) global 'main' - either global repo options or DNF control-only options
+            - NOTE: there must always ever be a single "main" section
+
+        2) <repoid> sections - options tied specifically to a given defined repo
+
+    [1] https://man7.org/linux/man-pages/man5/dnf.conf.5.html
+    """
+
+    # Don't model all known DNF options for validation purposes - it's user's responsibility!
+    dnf: Dict[Union[Literal["main"], str], Dict[str, Any]]
+
+    @pydantic.model_validator(mode="before")
+    def _validate_dnf_options(cls, data: Any, info: pydantic.ValidationInfo) -> Optional[Dict]:
+        """Fail if the user passes unexpected configuration options namespace."""
+
+        def _raise_unexpected_type(repr_: str, *prefixes: str) -> None:
+            loc = ".".join(prefixes + (repr_,))
+            raise ValueError(f"Unexpected data type for '{loc}' in input JSON: expected 'dict'")
+
+        prefixes: List[str] = ["options"]
+
+        if not data:
+            return None
+
+        if not isinstance(data, dict):
+            _raise_unexpected_type(data, *prefixes)
+
+        if "dnf" not in data:
+            raise ValueError(f"Missing required namespace attribute in '{data}': 'dnf'")
+
+        if diff := set(cls.model_fields) - set(data.keys()):
+            raise ValueError(f"Extra attributes passed in '{data}': {diff}")
+
+        prefixes.append("dnf")
+        options_scope = data["dnf"]
+        if not isinstance(options_scope, dict):
+            _raise_unexpected_type(options_scope, *prefixes)
+
+        for repo, repo_options in options_scope.items():
+            prefixes.append(repo)
+            if not isinstance(repo_options, dict):
+                _raise_unexpected_type(repo_options, *prefixes)
+
+        return data
+
+
 class GomodPackageInput(_PackageInputBase):
     """Accepted input for a gomod package."""
 
@@ -104,6 +170,7 @@ class RpmPackageInput(_PackageInputBase):
     """Accepted input for a rpm package."""
 
     type: Literal["rpm"]
+    options: Optional[Union[_DNFOptions]] = None
 
 
 class YarnPackageInput(_PackageInputBase):
--- a/cachi2/core/package_managers/rpm/main.py
+++ b/cachi2/core/package_managers/rpm/main.py
@@ -314,7 +314,8 @@ def _generate_repofiles(from_output_dir: Path, for_output_dir: Path) -> None:
             continue
         log.debug(f"Preparing repofile content for arch '{arch.name}'")
         repofile = _Repofile(defaults={"gpgcheck": "1"})
-        for entry in sorted(arch.iterdir()):
+
+        for entry in arch.iterdir():
             if not entry.is_dir() or entry.name == "repos.d":
                 continue
             repoid = entry.name
--- a/cachi2/core/package_managers/rpm/main.py
+++ b/cachi2/core/package_managers/rpm/main.py
@@ -5,7 +5,7 @@ import shlex
 from configparser import ConfigParser
 from os import PathLike
 from pathlib import Path
-from typing import Any, Union, no_type_check
+from typing import Any, Dict, Optional, Union, no_type_check
 from urllib.parse import quote
 
 import yaml
@@ -74,14 +74,39 @@ class _Repofile(ConfigParser):
 def fetch_rpm_source(request: Request) -> RequestOutput:
     """Process all the rpm source directories in a request."""
     components: list[Component] = []
+    options: Dict[str, Any] = {}
+    noptions = 0
+
     for package in request.rpm_packages:
         path = request.source_dir.join_within_root(package.path)
         components.extend(_resolve_rpm_project(path, request.output_dir))
+
+        # FIXME: this is only ever good enough for a PoC, but needs to be handled properly in the
+        # future.
+        # It's unlikely that a project would be split into multiple packages, i.e. supplying
+        # multiple rpms.lock.yaml files. We'd end up generating a single .repo file anyway,
+        # however, although trying to pass conflicting options to DNF for identical repoids via the
+        # input JSON doesn't make much sense from practical perspective (i.e. there's going to be a
+        # single .repo file) the CLI technically allows it in the input JSON.
+        # We're deliberately taking the easy route here by only assuming the "last" set of options
+        # we found in the input JSON instead of doing a deep merge of all the nested dicts.
+        # Nevertheless, we'll at least emit a warning at the end so that the user is informed
+        if package.options and package.options.dnf:
+            options = package.options.model_dump()
+            noptions += 1
+
+    if noptions > 1:
+        log.warning(
+            "Multiple sets of DNF options detected on the input: "
+            "Only one input RPM project package can specify extra DNF options, "
+            "the last one seen will take effect"
+        )
 
     return RequestOutput.from_obj_list(
         components=components,
         environment_variables=[],
         project_files=[],
+        options={"rpm": options} if options else None,
     )
 
 
@@ -272,7 +297,7 @@ def inject_files_post(from_output_dir: Path, for_output_dir: Path, **kwargs: Any
     """Run extra tasks for the RPM package manager (callback method) within `inject-files` cmd."""
     if Path.exists(from_output_dir.joinpath(DEFAULT_PACKAGE_DIR)):
         _generate_repos(from_output_dir)
-        _generate_repofiles(from_output_dir, for_output_dir)
+        _generate_repofiles(from_output_dir, for_output_dir, kwargs.get("options"))
 
 
 def _generate_repos(from_output_dir: Path) -> None:
@@ -298,7 +323,9 @@ def _createrepo(reponame: str, repodir: Path) -> None:
     log.debug(stdout)
 
 
-def _generate_repofiles(from_output_dir: Path, for_output_dir: Path) -> None:
+def _generate_repofiles(
+    from_output_dir: Path, for_output_dir: Path, options: Optional[Dict] = None
+) -> None:
     """
     Generate templates of repofiles for all arches.
 
@@ -308,6 +335,13 @@ def _generate_repofiles(from_output_dir: Path, for_output_dir: Path) -> None:
     Repofiles are not directly created from the templates in this method - templates are stored
     in the project file.
     """
+    dnf_options = None
+    dnf_options_repos = None
+
+    if options:
+        dnf_options = options.get("rpm", {}).get("dnf", {})
+        dnf_options_repos = dnf_options.keys()
+
     package_dir = from_output_dir.joinpath(DEFAULT_PACKAGE_DIR)
     for arch in package_dir.iterdir():
         if not arch.is_dir():
@@ -320,6 +354,12 @@ def _generate_repofiles(from_output_dir: Path, for_output_dir: Path) -> None:
                 continue
             repoid = entry.name
             repofile[repoid] = {}
+
+            # TODO: purposefully ignoring the fact that options might be passed within the "main"
+            # context of DNF options which would mean we'd have to generate a dnf.conf since such
+            # options are global, skipping that for now
+            if dnf_options and dnf_options_repos and repoid in dnf_options_repos:
+                repofile[repoid] = dnf_options[repoid]
 
             localpath = for_output_dir.joinpath(DEFAULT_PACKAGE_DIR, arch.name, repoid)
             repofile[repoid]["baseurl"] = f"file://{localpath}"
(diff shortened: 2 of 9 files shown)


cachi2/core/package_managers/rpm/main.py
```diff
@@ -30,17 +31,72 @@
...
+        # FIXME: this is only ever good enough for a PoC, but needs to be handled properly in the
+        # future.
+        # It's unlikely that a project would be split into multiple packages, i.e. not supplying
+        # multiple rpms.lock.yaml files. We'd end up generating a single .repo file anyway,
+        # however, although trying to pass conflicting options to DNF for identical repoids via the
+        # input JSON doesn't make much sense from that perspective (i.e. single .repo file), the
+        # CLI technically allows it in the input JSON.
+        # We're deliberately taking the easy route here by only assuming the "last" set of options
+        # we found in the input JSON instead of doing a deep merge of all the nested dicts.
+        if package.options and package.options.dnf:
```
brunoapimentel, line 92: I'm wondering if we should change this to be a global Typer option right away, since it will always affect all repofiles created in a single request.

We already have precedent for this in the project: some gomod [options](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/interface/cli.py#L140-L160) now target all gomod packages that exist in a single request, and are defined globally.

brunoapimentel: (you could argue these options are not a good example, as they are kept this way simply because they were ported from Cachito, and they probably should be defined per-package... but in the rpms case, I think it probably makes sense  to have a request-global option)

cachi2/core/models/input.py
```diff
@@ -104,6 +170,7 @@ class RpmPackageInput(_PackageInputBase):
     """Accepted input for a rpm package."""
 
     type: Literal["rpm"]
+    options: Optional[Union[_DNFOptions]] = None
```
brunoapimentel, line 173: I'm wondering if it makes sense to simplify how we define the `dnf` options dictionary. 

So my first question is: do we foresee any case in which we would pass a different set of options that is not `dnf` for rpms? Because if we're only doing this to keep a consistent interface in case the need arises someday, then I'm not sure the extra verbosity is worth it.

I'm also wondering if we need to nest it under `options` at all, since the other package managers simply add their extra options as a direct key in the PackageInput class (like, for example, `requirements_build_files` for pip).

cachi2/core/package_managers/rpm/main.py
eskultety: > (you could argue these options are not a good example, as they are kept this way simply because they were ported from Cachito, and they probably should be defined per-package... but in the rpms case, I think it probably makes sense to have a request-global option)

This I can comment on right away - I actually never liked these because while convenient I'm not convinced that this was the best design as CLI options are IMO meant to further adjust either the behaviour of the tool in general or its scoped behaviour, i.e. a subcommand (in this case `fetch-deps`). The gomod options don't fit either use case because although being used to modify a subcommand, they're tied to a specific package manager only, so I guess (without having tried putting down the effort of writing a PoC prototype to see if it works) it may have been better to introduce another level of command nesting by having a per-package manager CLI subparser for which these gomod CLI options we have would make the most sense, i.e. something along the lines `cachi2 fetch-deps gomod --<gomod_opt> <input_JSON>`

cachi2/core/models/input.py
eskultety: > I'm wondering if it makes sense to simplify how we define the `dnf` options dictionary.
> 
> So my first question is: do we foresee any case in which we would pass a different set of options that is not `dnf` for rpms? Because if we're only doing this to keep a consistent interface in case the need arises someday, then I'm not sure the extra verbosity is worth it.
> 

Yeah, normally I'd say you're right and we don't need the extra verbosity and nesting, i.e. without the usage of "namespaces" for options, however, given that these models are all designated for validation user input, they're therefore part of the public API and so we need to be extra careful about designing things so that we don't regret it later (nobody likes breaking public API) so this was my take on thinking ahead for the price of extra verbosity and hence more JSON (I know, JSON sucks for human operators...).

> I'm also wondering if we need to nest it under `options` at all, since the other package managers simply add their extra options as a direct key in the PackageInput class (like, for example, `requirements_build_files` for pip).

^This approach would not scale because:
1) the options would not be logically structured in any way, IOW you could mix options for RPM itself with DNF options on the same level in the input model
2) if you look at the manual page 5 of DNF config you'll see how many options there are, following the same principle for the input model and enumerate all those options in our input models would quickly get out of hand as it doesn't scale - IOW we should not try validaitng the correctness of the options used for the underlying toolkit (dnf in this case), we ought to only enumerate options that directly modify our behaviour and leave the rest completely opaque (like I did with the nested dictionaries and having used the type `Any`) to shield from user errors we simply cannot control.

brunoapimentel: > the options would not be logically structured in any way

Sorry, I was not clear with my second question/suggestion. I wast thinking on something like:
```
cachi2 fetch-deps --source /my/sources '{
   "type": "rpm",
   "path": "a/path",
   "dnf-options": { //options go here },
   "another-tool-we-might-add": { //extra config go here },
   "flag-we-may-need": true
}'
```

We would keep the logical nesting of the dnf options, we would just go away with the options key itself.

eskultety: > cachi2 fetch-deps --source /my/sources '{
   "type": "rpm",
   "path": "a/path",
   "dnf-options": { //options go here },
   "another-tool-we-might-add": { //extra config go here },
   "flag-we-may-need": true
}'

Right, that would of course work and I can be swayed this way although one of my main motivations for designing this was a consistent unified interface --> 'options'. That way, the keyword would ultimately be the same for all package_managers and predictible, IOW one would not need to open documentation to know how to supply further options since we could document it such that 'options' are always followed by another namespaced dictionary denoted by the tool one is trying to configure with the options, it would be a bit harder with your proposal I think.

brunoapimentel: So, if we applied this to gomod, we would end up with something like this:
```
cachi2 fetch-deps '{
   "type": "gomod",
   "path": "a/path",
   "options": {
        "vendor-check": true,
        "cgo_disable": true
   }
}'
```

Or we would nest these options under some key like we do with `dnf`?

eskultety: > Or we would nest these options under some key like we do with dnf?

In context of this proposal we'd have to nest them under `go` or `go-bin` or some other keyword for consistency reasons - far from ideal - but consistency is key if one is working with multiple different tools underneath which users may want to tweak eventually (I'm not saying it's going to happen, just looking for a generic way to handle this). I think the existence of an additional nesting is IMO not an issue pressing as much as the fact one may want to apply options globally or per-package manager, i.e. multiple `RpmPackageInputs`.
Overall I think since this closely relates to the RPMs PoC which isn't officially marked as supported could serve as a guinea pig to see whether that interface really made sense in practical terms with the option to take it back later and try to find a better way of incorporating the concept as proposed to other package managers; unless @brunoapimentel  you of course see other technical hurdles and pitfalls with this proposal. What I'm aiming at is that maybe this PR isn't the right space to have an extensive discussion on how to further built upon this concept if even the concept itself hasn't been accepted/approved yet IIUC :).

brunoapimentel: You're absolutely right. I was just trying to understand how you were imagining this unified `options` interface would work across package managers.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -4,7 +4,7 @@
 import shlex
 from os import PathLike
 from pathlib import Path
-from typing import Any, Union
+from typing import Any, List, Union
```
brunoapimentel: Whereas I don't oppose this change per se, I think we've been consistently using `list` for typing.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -224,15 +224,11 @@ def _generate_sbom_components(files_metadata: dict[Path, Any]) -> list[Component
     return components
 
 
-def inject_files_post(*args: Any, **kwargs: Any) -> None:
+def inject_files_post(from_output_dir: Path, for_output_dir: Path, **kwargs: Any) -> None:
```
brunoapimentel, line 283: Should we also change the corresponding function in `resolver.py` to match these types?

cachi2/core/models/input.py
brunoapimentel: To summarize and close this thread, we ended up with two viable options:

With a `options` key:
```
{
   "type": "rpm",
   "path": ".",
   "options": {
       "dnf": {
           "gpg-check": "1"
        }
       //additional configuration goes here
    }
}
```

Without a `options` key:
```
{
   "type": "rpm",
   "path": ".",
   "dnf-options": { 
       "gpg-check": "1"
    }
   //additional configuration goes here
}
```

After the arguments you made, I'm fine with the initial approach you proposed, Erik (with a `options` key). I'm leaving this comment here for other reviewers.

cachi2/core/package_managers/rpm/main.py
eskultety: > I am now wondering if we should have an explict error in case the user tries to define this in more than one RpmPackageInput (probably better than to cause some type of error during the build because one of the dnf options set was ignored).

I'll try to cook something up and stuff it into a new model validator for `RpmPackageInput`.

cachi2/core/package_managers/rpm/main.py
eskultety: My bad then for not checking the precedent then. While I think we should be consistent in using the types from typing that would result in an unrelated patch series, so I'll drop this hunk.

cachi2/core/package_managers/rpm/main.py
eskultety: > I'll try to cook something up and stuff it into a new model validator for `RpmPackageInput`.

Doh! A validator won't cut it, so I'll place the check elsewhere.

eskultety: In the end I opted for a simple warning since the fact that we can't handle multiple sets of DNF options is our problem but not a user error per se, they just abide what the model for the CLI interface already permits, so a fatal error didn't feel appropriate to me, but I can change that if needed.

brunoapimentel: I guess that's probably good enough. I'll leave this thread open for a little longer to see what others think about it.

brunoapimentel:
> @brunoapimentel I wonder if I should extract the `ConfigParser` bits on .repo file generation and propose that as a separate PR as I didn't get any comments on that and I'm not sure whether we're in agreement or it got lost in context of the main change - options in `RpmPackageInput`.

I did some tests with the implementation based on `ConfigParser`, and it LGTM. So I'd probably keep it the way it is since the implementation is already done.

The only thing I was in middle of reviewing before getting carried away with something else was the [custom validation](https://github.com/containerbuildsystem/cachi2/pull/522/files#diff-852b8cbcbd397f70fcea9f4b4ebc24252ea1e3eea14a56cab80b6a88b5fa7486R100-R131) you added in the last commit. From the few manual tests I did, it seemed that Pydantic handled these same use cases pretty well, and I thought that the error messages were reasonable. But I've likely missed the issues you had with unit tests.

eskultety:
> The only thing I was in middle of reviewing before getting carried away with something else was the [custom validation](https://github.com/containerbuildsystem/cachi2/pull/522/files#diff-852b8cbcbd397f70fcea9f4b4ebc24252ea1e3eea14a56cab80b6a88b5fa7486R100-R131) you added in the last commit. From the few manual tests I did, it seemed that Pydantic handled these same use cases pretty well, and I thought that the error messages were reasonable. But I've likely missed the issues you had with unit tests.

Soo, let me illustrate the issue on 1 example:

```
x = _DNFOptions.model_validate({'type': 'rpm', 'options': 'bad_type'})
*** pydantic_core._pydantic_core.ValidationError: 3 validation errors for _DNFOptions
dnf
  Field required [type=missing, input_value={'type': 'rpm', 'options': 'bad_type'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.6/v/missing
type
  Extra inputs are not permitted [type=extra_forbidden, input_value='rpm', input_type=str]
    For further information visit https://errors.pydantic.dev/2.6/v/extra_forbidden
options
  Extra inputs are not permitted [type=extra_forbidden, input_value='bad_type', input_type=str]
    For further information visit https://errors.pydantic.dev/2.6/v/extra_forbidden
```

^This is what I get without the validator, IOW the only bit of information I get is 'Field required' but it doesn't tell me anything else like what field is missing, so despite `input_value` the error message is IMO very vague.

vs

This is after introducing the validator

```
x = _DNFOptions.model_validate({'type': 'rpm', 'options': 'bad_type'})
*** pydantic_core._pydantic_core.ValidationError: 1 validation error for _DNFOptions
  Value error, Missing required namespace attribute in '{'type': 'rpm', 'options': 'bad_type'}': 'dnf' [type=value_error, input_value={'type': 'rpm', 'options': 'bad_type'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.6/v/value_error
```

It's at least IMO clear that a keyword 'dnf' is missing, so from that perspective I think that message is clearer.
Besides being vague, based on the fact that you get multiple error messages with the default behaviour I had some troubles figuring out the error msg regex for unit tests properly and so I created a custom validator with more deterministic messages. @brunoapimentel do you think there's a better way? I'm fine with dropping the validator if there is :) .

brunoapimentel:
> @brunoapimentel do you think there's a better way? I'm fine with dropping the validator if there is :) .

Oh, I see it now. When you invoke the actual CLI, the output you get is different (although I'm not sure which piece is making the conversion).

I started with you example here, and extrapolated a little:
```
$ cachi2 fetch-deps --source ../test-repos/cachi2-rpm --dev-package-managers '{"type": "rpm", "options": "bad_type"}'
2024-04-29 22:18:48,960 ERROR InvalidInput: 1 validation error for user input\npackages -> 0 -> rpm -> options\n  Value error, Unexpected data type for 'options.bad_type' in input JSON: expected 'dict'
packages -> 0 -> rpm -> options
  Value error, Unexpected data type for 'options.bad_type' in input JSON: expected 'dict'

$ cachi2 fetch-deps --source ../test-repos/cachi2-rpm --dev-package-managers '{"type": "rpm", "options": {"bad_type": "this"}}'
2024-04-29 22:19:39,650 ERROR InvalidInput: 1 validation error for user input\npackages -> 0 -> rpm -> options\n  Value error, Missing required namespace attribute in '{'bad_type': 'this'}': 'dnf'
packages -> 0 -> rpm -> options
  Value error, Missing required namespace attribute in '{'bad_type': 'this'}': 'dnf'

$ cachi2 fetch-deps --source ../test-repos/cachi2-rpm --dev-package-managers '{"type": "rpm", "options": {"dnf": "this"}}'
2024-04-29 22:19:47,235 ERROR InvalidInput: 1 validation error for user input\npackages -> 0 -> rpm -> options\n  Value error, Unexpected data type for 'options.dnf.this' in input JSON: expected 'dict'
packages -> 0 -> rpm -> options
  Value error, Unexpected data type for 'options.dnf.this' in input JSON: expected 'dict'
```

I think these messages are a lot clearer than the ones you posted before, so maybe we could just go with them. Wdyt?

eskultety:
> I think these messages are a lot clearer than the ones you posted before, so maybe we could just go with them. Wdyt?

Uhm, I'm sorry, I'm afraid you lost me a bit in your examples, so do you want me to modify the messages I emit in the validator in any way or are we okay with what is proposed?

brunoapimentel (approved):
My only nitpick here is about implementing the custom validation, but I don't see it as blocker, so I'm leaving my ack.

taylormadore (approved):

eskultety:
@brunoapimentel I needed to update this PR to reflect your recent integration test changes by adjusting them to compare on the parsed structure contents rather than verbatim file contents when it comes to the Configparser instance I introduced. Let me know if you still agree with that adjustment so that I can finally merge this. (this time for real if CI passes)

brunoapimentel:
> @brunoapimentel I needed to update this PR to reflect your recent integration test changes by adjusting them to compare on the parsed structure contents rather than verbatim file contents when it comes to the Configparser instance I introduced. Let me know if you still agree with that adjustment so that I can finally merge this. (this time for real if CI passes)

The changes on the integration test LGTM. Let's get this merged :)

PR #528, opened 2024-04-29
Switch the container image from Fedora 39 -> RH UBI 9

eskultety:
This PR switches the existing Fedora 39 container image to the RH UBI 9. A couple of tweaks were needed along the way to be able to perform the switch. This can be (doesn't necessarily need to be) perceived as a prerequisite to ultimately building the image in a multi-stage fashion.

Discussion: https://github.com/containerbuildsystem/cachi2/discussions/401

**EDIT: based on testing and the discussion the latest proposal supersedes RH UBI 9 with RockyLinux 9**


eskultety:
since v1:
- silence Hadolint DL3007: `Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag` as we actually always want to test against the latest image and get early feedback when anything changes rather than rely on an old image tag

eskultety:
Since v2:
- rebase on top of main
- reword commit 1

eskultety:
ping. I still need to have a look at the integration test failure but I'd like to get first round of reviews on the general idea.

eskultety:
ping^2

taylormadore:
I believe that yarn v4 will require node >= 18. How do we want to handle that?

eskultety:
> I believe that yarn v4 will require node >= 18. How do we want to handle that?

With multi-stage, which would come next, that would give us all the flexibility we need, both in terms of node an Go.

eskultety:
I found one fairly annoying downside to using UBI - for some weird reason it doesn't ship with `createrepo_c` which the base RHEL does ship with. We'd get it installed via `pip` to `/usr/local/bin/` but with #544 that would change because what is proposed there is to install cachi2 inside a virtual env and copy that one to the final image. I'm mentioning it here rather than in #544 is that although we can symlink `createrepo_c` from the virtual env, but I'm open to a broader discussion here if it turns out that this might turn into a maintenance nightmare with respect to UBI's limited set of packages and instead we might consider something like AlmaLinux or RockyLinux instead (afterall, both are forks of RHEL).

eskultety:
> I found one fairly annoying downside to using UBI - for some weird reason it doesn't ship with `createrepo_c` which the base RHEL does ship with. We'd get it installed via `pip` to `/usr/local/bin/` but with #544 that would change because what is proposed there is to install cachi2 inside a virtual env and copy that one to the final image. I'm mentioning it here rather than in #544 is that although we can symlink `createrepo_c` from the virtual env, but I'm open to a broader discussion here if it turns out that this might turn into a maintenance nightmare with respect to UBI's limited set of packages and instead we might consider something like AlmaLinux or RockyLinux instead (afterall, both are forks of RHEL).

For example if I try Rocky Linux's take on [UBI](https://hub.docker.com/layers/rockylinux/rockylinux/9.4-ubi/images/sha256-fd0e3e8d16ba7202fa4f0c732d5b72607ad89c3e434d57bd1fadeb4a6a6d2e4b?context=explore) then `createrepo` is present and unlike AlmaLinux Rocky still claims to be 100% bug-for-bug compatible with RHEL, i.e. a RHEL fork.
```
[root@aea9a6132ee5 /]# dnf search createrepo
Rocky Linux 9 - BaseOS                                                    1.6 MB/s | 2.2 MB     00:01    
Rocky Linux 9 - AppStream                                                 3.1 MB/s | 7.9 MB     00:02    
Rocky Linux 9 - Extras                                                     38 kB/s |  15 kB     00:00    
=================================== Name & Summary Matched: createrepo ===================================
python3-createrepo_c.x86_64 : Python 3 bindings for the createrepo_c library
======================================== Name Matched: createrepo ========================================
createrepo_c.x86_64 : Creates a common metadata repository
createrepo_c-libs.i686 : Library for repodata manipulation
createrepo_c-libs.x86_64 : Library for repodata manipulation
[root@aea9a6132ee5 /]# 
```

Any thoughts/preferences?

ben-alkov:
> for some weird reason it doesn't ship with `createrepo_c`

You've hit the main issue with using UBIs as anything*other than* a base runtime - they're very stripped down.

It sounds like Rocky is the way to go, except at that point, are we really better off than just sticking with Fed?

eskultety:
> > for some weird reason it doesn't ship with `createrepo_c`
> 
> You've hit the main issue with using UBIs as anything _other than_ a base runtime - they're very stripped down.
> 

Uhm, not entirely true I guess. See this has nothing to do with runtime, but repo setup. The Rocky flavour I tested with was also marked as UBI, except without carefully constructed limitations of the base repositories. Without disclosing too much detail, let's just say that the same repository on base RHEL contains the package while the RH UBI counterpart doesn't. And it makes sense to keep RH's UBI licensing and intended usage intact I guess, but what determines the package selection that ends up in those repositories is a mystery to me. I'll just close my thought with a simple: my bad for actually not checking the package catalog on the proposed RH UBI which would tell me immediately what I could get inside and what I couldn't.

> It sounds like Rocky is the way to go, except at that point, are we really better off than just sticking with Fed?

Well, pretty much anything is really better than Fedora for a production use (i.e. other than CI to uncover potential problems early) since it won't suffer from the 13 months EOL period.

eskultety:
> > > for some weird reason it doesn't ship with `createrepo_c`
> > 
> > 
> > You've hit the main issue with using UBIs as anything _other than_ a base runtime - they're very stripped down.
> 
> Uhm, not entirely true I guess. See this has nothing to do with runtime, but repo setup. The Rocky flavour I tested with was also marked as UBI, except without carefully constructed limitations of the base repositories. Without disclosing too much detail, let's just say that the same repository on base RHEL contains the package while the RH UBI counterpart doesn't. And it makes sense to keep RH's UBI licensing and intended usage intact I guess, but what determines the package selection that ends up in those repositories is a mystery to me. I'll just close my thought with a simple: my bad for actually not checking the package catalog on the proposed RH UBI which would tell me immediately what I could get inside and what I couldn't.
> 
> > It sounds like Rocky is the way to go, except at that point, are we really better off than just sticking with Fed?
> 
> Well, pretty much anything is really better than Fedora for a production use (i.e. other than CI to uncover potential problems early) since it won't suffer from the 13 months EOL period. EDIT: though I wonder if there are any surprising deps coming that wouldn't be available even on distros like Alma or Rocky, I guess only time will tell.

To give this a bit more perspective, here's a comparison on the number of packages available without EPEL:
```
RH UBI-9
$ cat /etc/os-release
NAME="Red Hat Enterprise Linux"
VERSION="9.3 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.3"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.3 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.3
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.3"

$ dnf list --all | wc -l
2261
====================================================
Rocky UBI-9
$ cat /etc/os-release 
NAME="Rocky Linux"
VERSION="9.4 (Blue Onyx)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="9.4"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Rocky Linux 9.4 (Blue Onyx)"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:rocky:rocky:9::baseos"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
SUPPORT_END="2032-05-31"
ROCKY_SUPPORT_PRODUCT="Rocky-Linux-9"
ROCKY_SUPPORT_PRODUCT_VERSION="9.4"
REDHAT_SUPPORT_PRODUCT="Rocky Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.4"

$ dnf list --all | wc -l
6825

File: Containerfile [hunk 1]
```diff
@@ -1,18 +1,20 @@
-FROM registry.fedoraproject.org/fedora-minimal:39
+# hadolint global ignore=DL3007
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
```
MartinBasti, line 2: To have reproducible and tested builds, we should avoid to have `latest` tag, the best option would be to have floating tag with pinned digest and make dependabot to update it regularly. Also latest tag in ubi-minimal will cause to bump minor versions, they sometimes contain not compatible changes.

MartinBasti:
Is it really so bad to use createrepo_c form pypi? I'd prefer UBI base image

Containerfile
ben-alkov: Finally, someone else who sees the evil of `:latest` :grin:

eskultety: The way you're presenting the "evilness" of latest in context of solving it with digests and dependabot is a bit misleading and maybe even false advertisement. Now, version locks monitored by dependabot make sense with individual packages because those are enclosed logical units where you always get **that** particular version. It doesn't work that way with container images, unless you never run an install on the package manager inside. The reason for that is that now we're talking about a set of packages that may get updated without you realizing and still thinking that you're using your safe pinned container image. The problem arises the moment you run e.g. `dnf install` because it'll likely update a subset of the base packages without you explicitly asking for it. Here's an example of it:

This is snippet of RockyLinux cachi2 container install with a newer (sha256:4b9c7b0285aa68f45ef42308fb7a364ffb63b461645cfb22c53212d85d16bf93) digest
```
================================================================================
 Package                 Arch    Version                       Repository  Size
================================================================================
Installing:
 gcc                     x86_64  11.4.1-3.el9                  appstream   32 M
 git-core                x86_64  2.43.0-1.el9                  appstream  4.4 M
 golang-bin              x86_64  1.21.10-1.el9_4               appstream   56 M
 nodejs                  x86_64  1:16.20.2-8.el9_4             appstream  111 k
 npm                     x86_64  1:8.19.4-1.16.20.2.8.el9_4    appstream  1.7 M
 python3-devel           x86_64  3.9.18-3.el9                  appstream  204 k
 python3-pip             noarch  21.2.3-8.el9                  appstream  1.7 M
 python3-setuptools      noarch  53.0.0-12.el9                 baseos     839 k
Upgrading:
 glibc                   x86_64  2.34-100.el9_4.2              baseos     2.0 M
 glibc-common            x86_64  2.34-100.el9_4.2              baseos     301 k
 glibc-minimal-langpack  x86_64  2.34-100.el9_4.2              baseos      27 k
Installing dependencies:
...
```

And this is a snippet of the same build using an older digest (sha256:b2ce50f292ea21d095dce94668a3ca403b8b589923d83fc78def0241a89a2c86):
```
================================================================================
 Package                 Arch    Version                       Repository  Size
================================================================================
Installing:
 gcc                     x86_64  11.4.1-3.el9                  appstream   32 M
 git-core                x86_64  2.43.0-1.el9                  appstream  4.4 M
 golang-bin              x86_64  1.21.10-1.el9_4               appstream   56 M
 nodejs                  x86_64  1:16.20.2-8.el9_4             appstream  111 k
 npm                     x86_64  1:8.19.4-1.16.20.2.8.el9_4    appstream  1.7 M
 python3-devel           x86_64  3.9.18-3.el9                  appstream  204 k
 python3-pip             noarch  21.2.3-8.el9                  appstream  1.7 M
 python3-setuptools      noarch  53.0.0-12.el9                 baseos     839 k
Upgrading:
 glibc                   x86_64  2.34-100.el9_4.2              baseos     2.0 M
 glibc-common            x86_64  2.34-100.el9_4.2              baseos     301 k
 glibc-minimal-langpack  x86_64  2.34-100.el9_4.2              baseos      27 k
 libgcc                  x86_64  11.4.1-3.el9                  baseos      91 k
 libgomp                 x86_64  11.4.1-3.el9                  baseos     267 k
 openssl                 x86_64  1:3.0.7-27.el9                baseos     1.2 M
 openssl-libs            x86_64  1:3.0.7-27.el9                baseos     2.2 M
 python3                 x86_64  3.9.18-3.el9                  baseos      25 k
 python3-libs            x86_64  3.9.18-3.el9                  baseos     7.3 M
Installing dependencies:
...
```

Notice how many more packages are upgraded during the process even though we didn't ask for it in the Containerfile. That is because the metadata database is refreshed every time to the latest one anyway and it's the direct result of package dependencies which we don't have any control over and pinning to a digest clearly isn't going to solve. To top it all off, the ^above is just for our project which installs just a few packages, for bigger projects the difference is going to be much more noticable given that the base set of packages is somewhere around 100+.

Inter-package dependencies/requirements are just one piece of the puzzle. The other is reporting the actual changes inside the container images. With individual packages, you automatically get a changelog that dependabot pulls from the given project which is nice, unless your project depends on some random garbage on github. However, that's not the case with docker images, unless the image vendor doesn't set `org.opencontainers.image.source` label on the image as the [dependabot docs states](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#docker). Now, unlike with individual projects/packages, it is apparently not that common with big reputable vendors out there and platform builder images to set that:
```
$ podman inspect registry.access.redhat.com/ubi9 | grep org.opencontainers.image.source
$ podman inspect docker.io/rockylinux/rockylinux:9-ubi | grep org.opencontainers.image.source
$ podman inspect fedora:39 | grep org.opencontainers.image.source
$ podman inspect docker.io/almalinux/9-base | grep org.opencontainers.image.source
```

And so when you look at the PR dependabot creates for such an image, you're going in blind completely as the only thing you see is a change in an opaque string, but you have literally no idea what has changed - how is that safe (or safer) when the only information you have is whether an image (as the subject of dependabot's PR) passed CI or not? What guarantees do we have about other potential breakages the CI hasn't uncovered yet?

Lastly, @mbasti you mentioned breakages in images and introducing dev versions of packages. I'd say "quite uncommon" and unlikely with platforms like RHEL where Red Hat takes huge pride in specifically not doing that and not introducing any incompatible breakages for the lifetime of the distro and instead keeping it stable for a very long time which is also the reason I even opened this PR and proposed ditching Fedora in the first place, because with Fedora you never know and I've seen broken repo metadata in CentOS Stream preventing package installs as well package upgrades multiple times (which is also something digest pinning will not shield you from at all).

Having said all of ^this and clearly not being a fan of digest pinning with container images despite that being praised as a good practice all over the internet (admittedly though in some cases, e.g. the language toolkit images it does make sense I agree) and I'm not convinced it's really solving anything in our particular case strictly in context of this project as I explained above. However, if you feel so strong about pinning the image for the "perceived" safety and it should be the only blocker for this PR to be accepted, then so be it, but I wanted to express my feelings about it backed by some real life experiment results. Let me know your thoughts so that I can start working on the necessary adjustments.

ben-alkov: `dnf update` is orthogonal to `:latest`.

The issue here is the (absolutely unfounded) assumption that `:latest` literally means "latest", i.e. the latest *content* (this is perhaps Docker's biggest design/architecture cock-up of *many*).

The *only* meaning of the auto-tag `"latest` is literally "last pushed". You can rely on the HITL to make sure that they're not mistakenly pushing e.g. an F39 image to `F40:latest`, but then you always have to worry about human error. Automation helps, but at that point, why not use a hash, or even better some non-default tag like `:latest-content`?

However: None of that addresses the uncertainty around changes made by `dnf`, but as I said, that's orthogonal, and not isolated to containers - the issue exists with *any* instantiation - VM, bare-metal, container...

eskultety: > `dnf update` is orthogonal to `:latest`.
> 
> The issue here is the (absolutely unfounded) assumption that `:latest` literally means "latest", i.e. the latest _content_ (this is perhaps Docker's biggest design/architecture cock-up of _many_).
> 
> The _only_ meaning of the auto-tag `"latest` is literally "last pushed". You can rely on the HITL to make sure that they're not mistakenly pushing e.g. an F39 image to `F40:latest`, but then you always have to worry about human error. Automation helps, but at that point, why not use a hash, or even better some non-default tag like `:latest-content`?
> 
> However: None of that addresses the uncertainty around changes made by `dnf`, but as I said, that's orthogonal, and not isolated to containers - the issue exists with _any_ instantiation - VM, bare-metal, container...

While I can accept the argument about the "latest push", that would still only apply IMO in random images which one picks as a dependency (just like any other package without a reputable history behind it, typically a package with a single maintainer and contributor). However, we're talking about the official maintained image here, which, I'd like to remind, is just some rebuild of RHEL, so if we don't have trust in maintainers of the image then what can we trust? In that case we shouldn't trust that maintainers don't push errors to RPM repo metadatada which DNF pulls, leading to broken package dependencies. Does it happen? On some distros it sometimes does. Same thing with containers, IMO the important thing here is to have some trust in the well-established community to provide adequate support in case of a problem occurs, like in this case, rebuild the image in a timely manner so that we might not even notice. Again, when we talk about our direct consumables with a massive community and ecosystem rather than "hobby abandonware" we depend on (e.g. reflink which sadly should not have happened) I'm still not seeing how hash pinning is doing us any good.
Since people seem to be in a general consensus of the value hash pinning of docker images brings in our particular case (not in general) despite using reputable sources I guess I can at least sell the idea to myself as a move towards consistency, given that we let dependabot handle the rest of our dependencies.

MartinBasti: and we shouldn't use latest if possible but stick to minor release version floating tag

lkolacek: > [@eskultety](https://github.com/eskultety) I don't get your point
> Why having dependabot gating updates in parent images is worse that blindly consuming anything from the latest? Still limits potential for unexpected breaking change.


Eskultet, I don’t get it either.

I agree with Martin and Ben that we should avoid using :latest at all costs. Dependabot will add an extra CI check for unknown changes in the next UBI version. I am not saying it will prevent us from hitting any issues after moving to the next version but it might catch an issue and IMO it does not make sense not to use it.

Plus removing the linter check is generally a bad practice. 

Maybe I just don’t get the benefit of using the latest here.

MartinBasti: > CI to be green to merge
this is the reason, CI may catch some issues

:latest tag may cause to bump to next major version and that's usually causes breakages.

Why should I have broken repo when this happens and the rush reactivelly address the issue when I can just prevent it. Saving one cycle of CI per week is nothing.

eskultety:
> Is it really so bad to use createrepo_c form pypi? I'd prefer UBI base image

UBI sadly only has 30% of the repository contents and RockyLinux is essentially RHEL, i.e. the new CentOS.

eskultety:
since v3:
- propose RockyLinux instead of RH UBI for the base image
- pin the release (no floating tag, we still want the latest 9.minor) to a particular build digest
- rename Containerfile to Dockerfile because dependabot only supports instruction files named "Dockerfile"
- add a dependabot github action to monitor the Dockerfile

Dockerfile
```diff
@@ -0,0 +1,46 @@
+FROM docker.io/library/rockylinux:9@sha256:d7be1c094cc5845ee815d4632fe377514ee6ebcf8efaed6892889657e5ddaaa6
+LABEL maintainer="Red Hat"
+
+WORKDIR /src
+RUN dnf -y install \
```
MartinBasti, line 5: Where do we actually install `createrepo`? I'm little bit confused here

eskultety: This PR is only about switching the image, i.e. `createrepo_c` is still being installed via pip. The relevant change would come in the #544 where the lack of the package would actually become apparent due to multi-stage.
If needed I can drop the mention of the package as the reason from commit message https://github.com/containerbuildsystem/cachi2/pull/528/commits/a2b4db7778f0ef3d4a4d2b6ad734290ab06a7a0e

MartinBasti: ak, thanks

MartinBasti (approved):

ben-alkov (approved):

Dockerfile
eskultety: just an FYI, I updated the commit message in question to drop any mention of the createrepo_c dependency as advertised in my previous comment.

eskultety:
Thanks for the review, merged.

PR #531, opened 2024-05-02
toil(pkg-mgr:pip): report SBOM "cachi2:missing_hash:in_file" property...

ben-alkov:
...when *any* 'requirements*.txt file is unhashed

- fix test data


ben-alkov:
This one-line change produces correct SBOMs, but I'm not completely comfortable with it as I don't fully understand why it works...

cachi2/core/package_managers/pip.py
```diff
@@ -134,14 +134,18 @@ def should_download(self) -> bool:
...
-            len(self.checksums_to_verify) > 0
+            len(self.checksums_to_match) > 0
             or len(self.pypi_checksums) == 0
             or len(self.user_checksums) == 0
         )
 
-    def should_verify_checksums(self) -> bool:
-        """Check if checksum verification is required."""
-        return len(self.checksums_to_verify) > 0
+    def has_checksums_to_match(self) -> bool:
```
a-ovchinnikov, line 142: A very minor nitpick: this method does not accept any arguments other than `self`, it might make sense to mark it as a property.

cachi2/core/package_managers/pip.py
```diff
@@ -2018,12 +2029,12 @@ def _download_url_package(
...
 
-    if url.hostname in trusted_hosts:
-        log.debug("Disabling SSL verification, %s is a --trusted-host", url.hostname)
-        insecure = True
-    elif url.port is not None and f"{url.hostname}:{url.port}" in trusted_hosts:
+    if url.port is not None and f"{url.hostname}:{url.port}" in trusted_hosts:
         log.debug("Disabling SSL verification, %s:%s is a --trusted-host", url.hostname, url.port)
         insecure = True
+    elif url.hostname in trusted_hosts:
+        log.debug("Disabling SSL verification, %s is a --trusted-host", url.hostname)
```
a-ovchinnikov, line 2036: Is there any reason for not using f-strings in log messages?

cachi2/core/package_managers/pip.py
```diff
@@ -1477,38 +1481,41 @@ def _process_req(
...
     download_info["requirement_file"] = str(requirements_file.file_path.subpath_from_root)
-    download_info["hash_verified"] = False
+    download_info["missing_checksums"] = False
+    download_info["checksum_matched"] = False
+    # "package_type" is *only* needed for PyPI deps
+    download_info["package_type"] = ""
 
-    def _hash_verify(path: Path, checksum_info: Iterable[ChecksumInfo]) -> bool:
-        verified: bool = False
+    def _checksum_match(path: Path, checksum_info: Iterable[ChecksumInfo]) -> bool:
```
a-ovchinnikov, line 1489: Another super minor nitpicky suggestion. What do you think about something like the code below:
```suggestion
try:
           must_match_any_checksum(path, cheksum_info)
except PackageRejected:
            path.unlink()
            log.warning("Download '%s' was removed from the output directory", path.name)
else:
        return True
return False
```
`matched` is not really needed here, so could be removed completely.

a-ovchinnikov (commented):
Looks good to me. Also posting two very minor non-blocking suggestions and a question.

cachi2/core/package_managers/pip.py
ben-alkov: IIRC, it's because %s strings are lazy-eval; in any case, that decision was made before my time.

cachi2/core/package_managers/pip.py
ben-alkov: You'll find I'm *very* partial to the "single exit" philosophy. Locals hardly cost anything, and a single exit point is easier to debug and reason about.

a-ovchinnikov: Fair enough. I would point out that in this specific case multiple exit points don't look like they would add extra complexity (they are right next to each other and highly unlikely to move further apart), but the published approach is totally fine. Thank you!

cachi2/core/package_managers/pip.py
a-ovchinnikov: Got it, thank you!

cachi2/core/package_managers/pip.py
```diff
@@ -1477,37 +1481,32 @@ def _process_req(
...
     download_info["kind"] = req.kind
     download_info["requirement_file"] = str(requirements_file.file_path.subpath_from_root)
-    download_info["hash_verified"] = False
+    download_info["missing_checksums"] = True
+    # "package_type" is *only* needed for PyPI deps
+    download_info["package_type"] = ""
 
-    def _hash_verify(path: Path, checksum_info: Iterable[ChecksumInfo]) -> bool:
-        verified: bool = False
+    def _checksum_match(path: Path, checksum_info: Iterable[ChecksumInfo]) -> None:
```
a-ovchinnikov, line 1488: It looks like this function does not only verify a checksum, but also produces a side-effect: it unlinks a Path. Would something like `_unlink_path_if_no_match_found`  be a more descriptive name?

cachi2/core/package_managers/pip.py
eskultety: I'm not a fan either and I like consistency, so to just add some minor context to this, I myself was recently pointed out the old string formatting style for logging is preferable and even recommended (though the official docs weren't that clear on this to me), but the truth is that this is not the regular old-style Python string formatting since the substitution happens in the logger, i.e. not immediately, so if you have a lot of debug logs, the potential penalty here is the added formatting burden of very large (e.g. JSON) strings that would otherwise not be used at all in production where debugging is disabled, which to me was really the only acceptable argument for this - we haven't formalized a guideline on this though yet.

cachi2/core/package_managers/pip.py
eskultety: Agreed on doing `return True` and `return False` here in this case, it's more direct and less confusing than using a variable.

cachi2/core/package_managers/pip.py
ben-alkov: Good point

a-ovchinnikov (approved):
LGTM!

cachi2/core/package_managers/pip.py
```diff
@@ -97,51 +97,57 @@ class DistributionPackageInfo:
...
-        if self.pypi_checksums and self.user_checksums:
-            checksums = matching
-            msg = "using intersection of user specified and PyPI reported checksums"
-        elif self.pypi_checksums:
-            checksums = self.pypi_checksums
-            msg = "using PyPI reported checksums"
-        elif self.user_checksums:
-            checksums = self.user_checksums
-            msg = "using user specified checksums"
+        intersection: set[ChecksumInfo] = self.pypi_checksum.intersection(self.req_file_checksums)
```
brunoapimentel, line 113: Nit: the intersection could be calculated in the if block (L116). I don't think having it here improves the readability, but that's just me.

cachi2/core/package_managers/pip.py
```diff
@@ -97,51 +97,57 @@ class DistributionPackageInfo:
     index_url: str
     is_yanked: bool
 
-    pypi_checksums: set[ChecksumInfo] = field(default_factory=set)
-    user_checksums: set[ChecksumInfo] = field(default_factory=set)
+    # PyPi only returns a single checksum for a given download artifact
+    pypi_checksum: set[ChecksumInfo] = field(default_factory=set)
```
brunoapimentel, line 101: Nit: even though PyPI seems to be consistent in providing only the `sha256` hash, I think that since we're keeping this variable as a set, we could keep its name in the plural form.

cachi2/core/package_managers/pip.py
```diff
@@ -1476,45 +1477,46 @@ def _split_hashes_from_options(options: list[str]) -> tuple[list[str], list[str]
...
-        return matched
 
     if dpi:
-        if dpi.has_checksums_to_match():
-            download_info["checksum_matched"] = _checksum_match(dpi.path, dpi.checksums_to_match)
-
+        if dpi.req_file_checksums:
+            download_info["missing_req_file_checksum"] = False
+        if dpi.has_checksums_to_match:
+            _checksum_must_match_or_path_unlink(dpi.path, dpi.checksums_to_match)
```
brunoapimentel, line 1504: Just checking my understanding: if the package has any `req_file_checksum`, we're marking it as having checksums in `requirements.txt` . Then, it will always pass the `if dpi.has_checksums_to_match` (since it has at least one `req_file_checksum`), and in case the checksum does not match, the package is discarded.

Could there be a scenario in which the `req_file_checksum` did not match, but a `pypi_checksum` matched, and we allowed the package to exist in the prefetched content? I think if this happened, the SBOM would lead us to believe that there was nothing wrong with the package, and that its checksums were properly matched.

ben-alkov: > if the package has any req_file_checksum, we're marking it as having checksums in requirements.txt

If this specific *download*, that is to say **this** artifact, has a checksum in the *current* requirements file.

> Could there be a scenario in which the req_file_checksum did not match

I'll have to think about it for a bit and get back to you

cachi2/core/package_managers/pip.py
```diff
@@ -2032,9 +2032,7 @@ def _download_url_package(
     if "cachito_hash" in requirement.qualifiers:
         url_with_hash = requirement.url
     else:
-        hashes = requirement.hashes
-        hash_spec = hashes[0] if hashes else requirement.qualifiers["cachito_hash"]
```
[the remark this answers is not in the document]
ben-alkov: It's a duplication of the opening `if`: `if "cachito_hash" in requirement.qualifiers:`: it's redundant. At this point, if there's a `cachito_hash` (quite literally), we're already using it.

cachi2/core/package_managers/pip.py
ben-alkov: If we:
- download an artifact
- calculate its checksum
- the checksum is correct
- we *have* `req_file_checksums` AND there's no match with `pypi_checksum`

then `checksums_to_match` *should be* empty, because `self.req_file_checksums.intersection(self.pypi_checksum)` will be empty. If `checksums_to_match` is empty, the check for that download fails at some point.

So, I think the answer to your question is that we *might* download something we "shouldn't", but the checksum matching will fail later on, so we should be OK

cachi2/core/package_managers/pip.py
ben-alkov: Actually, ignore that. I remember now, from endlessly jacking around with the unit tests, that changing any of the `checksum` variables to anything other than an iterable of some kind causes huuuge problems. I'll revert the name change.

cachi2/core/package_managers/pip.py
brunoapimentel: You're right. We actually don't even download it: https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/pip.py#L1912-L1918

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
slimreaper35: Could you add more context to the commit message ?

brunoapimentel (approved):
LGTM :rocket: :tada:

PR #532, opened 2024-05-02
RPM PoC: Integration tests

brunoapimentel:
This PR implements the final scenarios:
- multiple packages
- multiple archs
- check the contents of the .repo file


ben-alkov (commented):
LGTM, but IANAE

tests/integration/test_rpm.py
```diff
@@ -105,6 +106,71 @@ def test_rpm_packages(
...
+    utils.fetch_deps_and_check_output(
+        tmp_path, test_case, test_params, source_folder, test_data_dir, cachi2_image
+    )
+
+    # call inject-files to create the .repo file
+    cmd = [
+        "inject-files",
+        output_folder,
+        "--for-output-dir",
+        Path("/tmp").joinpath(f"{test_case}-output"),
```
eskultety, line 151: can't `tmp_path` be used here as well?

tests/integration/test_data/rpm_test_repo_file/cachi2.repo
```diff
@@ -0,0 +1,3 @@
+[releases]
```
eskultety, line 1: Can we make this slightly more complex by testing the following additionally here?
- multiple repo sections generated
- deps which are unaffiliated with any official repo (i.e. one is generated for them)?

The latter will require some mocking of the random repo name string.

eskultety (commented):
@brunoapimentel looks good, nothing substantial, but see my comments whether you think it's worth pursuing.

tests/integration/test_rpm.py
brunoapimentel: It can, but then we'd need to deal with its randomness. FWIW, we are not using this path at all, it is just a means of having consistent data for the test (also, I just copied this code from the [utils](https://github.com/containerbuildsystem/cachi2/blob/main/tests/integration/utils.py#L394-L399) module).

tests/integration/test_data/rpm_test_repo_file/cachi2.repo
brunoapimentel: Done. We can't really mock any Cachi2 method in the integration test, but I am replacing the random string with regex so that we can actually cover this case in the test.

I don't quite like the way the code is, but maybe we can improve it a bit after your `ConfigParser` change goes in.

tests/integration/test_rpm.py
eskultety: Doh! I missed the obvious connection to the expected .repo file you're creating.

Still, a nitpick - we could just write this as `Path("/tmp", f"{test_case}-output")` as `joinpath` makes everything visually longer and with `Path` and `/` I guess it only exists for completeness :), you don't need to change anything though, just my 2c.

tests/integration/test_data/rpm_test_repo_file/cachi2.repo
eskultety: You're right, we can't! One note that I'm going to make although it won't change anything right now is the fact that we're assuming the file is going to be generated always the same way because right now our code sorts the output before dumping which is technically totally unnecessary and is also proposed to be dropped with #522 where as you mention `ConfigParser` is suggested to come into play which means we'll have to compare the file on the object level rather than 1:1 content level, so this is just a NB for me to update the PR in question when this is merged. 

> I don't quite like the way the code is, but maybe we can improve it a bit after your ConfigParser change goes in.

we could drop the need for regex handling and tackle this in the `ConfigParser` instance but I doubt it will make the code that much better, it would definitely make it longer, we'll see.

eskultety (approved):

ben-alkov (approved):

Issue #482, opened 2024-03-07
Move off of the pkg_resources to importlib and packaging APIs 


eskultety:
The `pkg_resources` module has been [officially marked deprecated ](https://setuptools.pypa.io/en/latest/pkg_resources.html) by setuptools for a while although no official date for removal has been set. That said, despite the fact that much of the backend implementation of the `pkg_resources` module has already been integrated with the suggested replacements:

- [`importlib.resources`](https://docs.python.org/3/library/importlib.resources.html#module-importlib.resources)
- [`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html#module-importlib.metadata)
- [`packaging`](https://pypi.org/project/packaging)

There was a version of [pip](https://pip.pypa.io/en/stable/news/#v23-1) which due to an upgrade of setuptools which had already marked `pkg_resources` as deprecated started complaining about a deprecated vendored module:
https://github.com/pypa/pip/issues/11975. The issue has been fixed since in pip, but the reality is that once a package has been marked as deprecated we should follow the official migration guides to drop such a dependency. 

Other useful references:
- https://importlib-resources.readthedocs.io/en/latest/migration.html
- https://importlib-metadata.readthedocs.io/en/latest/migration.html

Issue #539, opened 2024-05-06
default pip-compile environment (3.9) fails to process our pyproject.toml


eskultety:
Under some circumstances running the `pip-compile` make target (Python 3.9) `pip-compile` fails to process the requirements  because `setuptools-scm` mysteriously chokes on [figuring out the project version in git](https://github.com/pypa/setuptools_scm/blob/d75201a03ea01bf2d99a2521b87cdea30320f887/src/setuptools_scm/git.py#L122) based on commit timestamps in ISO format which old (Python < 3.11) [1] wasn't able to process. The word "mysteriously" was used as this issue isn't 100% reproducible, so it actually passes at times and sometimes fails.

There isn't much help here other than running pip-compile on >Python 3.9 or moving off of setuptools.

```
$ pip-compile -v --allow-unsafe --generate-hashes --output-file=requirements.txt pyproject.toml
Using pip-tools configuration defaults found in 'pyproject.toml'.
Creating isolated environment: venv+pip...
Installing packages in isolated environment:
- setuptools
- setuptools-scm
Getting build dependencies for wheel...
Traceback (most recent call last):
  File "cachi2.git/venv39/lib64/python3.9/site-packages/pyproject_hooks/_in_process/_in_process.py", line 373, in <module>
    main()
  File "cachi2.git/venv39/lib64/python3.9/site-packages/pyproject_hooks/_in_process/_in_process.py", line 357, in main
    json_out["return_val"] = hook(**hook_input["kwargs"])
  File "cachi2.git/venv39/lib64/python3.9/site-packages/pyproject_hooks/_in_process/_in_process.py", line 134, in get_requires_for_build_wheel
    return hook(config_settings)
  File "/tmp/build-env-lf7col77/lib64/python3.9/site-packages/setuptools/build_meta.py", line 325, in get_requires_for_build_wheel
    return self._get_build_requires(config_settings, requirements=['wheel'])
  File "/tmp/build-env-lf7col77/lib64/python3.9/site-packages/setuptools/build_meta.py", line 295, in _get_build_requires
    self.run_setup()
  File "/tmp/build-env-lf7col77/lib64/python3.9/site-packages/setuptools/build_meta.py", line 311, in run_setup
    exec(code, locals())
  File "<string>", line 1, in <module>
  File "/tmp/build-env-lf7col77/lib64/python3.9/site-packages/setuptools/__init__.py", line 104, in setup
    return distutils.core.setup(**attrs)
[...]

PR #542, opened 2024-05-07
Remove deprecated dependency pkg_resources and refactor PipRequirement

### LINKED ISSUE #486: Make PipRequirement a subclass of packaging.Requirement
We created our own big class which we then populate with contents from `pkg_resources.Requirement`. Looking at the backend implementation coming from `packaging.Requirement` it looks like it's very lean and hence we should be able to properly subclass our own class from that one.

ianrochapg:
PipRequirement implemented pkg_resources which is very lean and also deprecated. Removing pkg_resources and using packaging.Requirement directly.

setuptools has been removed because it was only used by pkg_resources, which has also been removed.


slimreaper35:
You forgot to run `make pip-compile` to regenerate requirements files (it takes _pyproject.toml_ as an input)

Also, we can probably delete `--allow-unsafe` option.

cachi2/core/package_managers/pip.py
```diff
@@ -30,6 +30,7 @@
 )
 
 import tomli
```
eskultety, line 32: **Commit message:**

>  Removing pkg_resources and subclassing
directly from packaging.Requirement.

You left out the part where you actually subclass `packaging.Requirement` in your implementation ;)

> setuptools has been removed because it was only used by pkg_resources

Commits should be contained to a single logical change, so from that POV ^this removal should be placed in a separate commit. I'm surprised this worked and passed the tests since we depend on setuptools as the built backend, but given it's specified as `required` at the top of `pyproject.toml` it'll get installed anyway, so if the CI's green I'm okay with dropping it from the deps list like you do.

> Resolves #486

Please put the actual URL in the commit message rather than a numeric ID so that when browsing git history locally one can access the URL directly.

cachi2/core/package_managers/pip.py
```diff
@@ -1300,11 +1301,8 @@ def from_line(cls, line: str, options: list[str]) -> Optional["PipRequirement"]:
             requirement.kind = "pypi"
 
         try:
-            parsed: Sequence[Requirement] = list(pkg_resources.parse_requirements(to_be_parsed))
-        except (
-            RequirementParseError,
-            pkg_resources.extern.packaging.requirements.InvalidRequirement,
-        ) as exc:
+            parsed: Sequence[Requirement] = list(_parse_requirements(to_be_parsed))
```
eskultety, line 1304: Doing this effectively just vendors the old pkg_resources code same way as pkg_resources vendors packaging from which it builds upon. IOW you can drop the whole `_parse_requirements` parsing into a list and instead do plain `packaging.Requirement(to_be_parsed)`. Note that tests will have to be adjusted since we're checking some corner cases that relate to `_parse_requirements` behaviour which no longer apply to the usage of `packaging.Requirement`.

cachi2/core/package_managers/pip.py
```diff
@@ -1463,6 +1461,22 @@ def _split_hashes_from_options(options: list[str]) -> tuple[list[str], list[str]
...
+def _standardize_name(name: str) -> str:
+    return re.sub("[^A-Za-z0-9.]+", "-", name)
+
+
+def _standardize_version(version: str) -> str:
+    try:
+        return str(packaging.version.Version(version))
+    except packaging.version.InvalidVersion:
+        version = version.replace(" ", ".")
+        return re.sub("[^A-Za-z0-9.]+", "-", version)
```
eskultety, line 1473: Same thing about vendoring with ^these 2 - part of the issue is investigating the proper replacements for these and by quickly going through packaging's docs [canonicalize_name](https://packaging.pypa.io/en/stable/utils.html#packaging.utils.canonicalize_name) and [version](https://packaging.pypa.io/en/stable/utils.html#packaging.utils.canonicalize_version) seem like candidates worth exploring whether they accomplish what we need.

pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
   "beautifulsoup4",
   "createrepo_c",
   "gitpython",
+  "jaraco.text",
```
eskultety, line 29: With the helper functions potentially gone this import likely won't be needed and we also shouldn't depend on a library that pkg_resources's maintainers/creators created just to be able to keep vendoring it within setuptools.

cachi2/core/package_managers/pip.py
```diff
@@ -1323,9 +1321,9 @@ def from_line(cls, line: str, options: list[str]) -> Optional["PipRequirement"]:
 
         requirement.download_line = to_be_parsed
         requirement.options = options
-        requirement.package = req.project_name
+        requirement.package = _standardize_name(req.name)
```
eskultety, line 1324: Not objecting to this line but I'm unclear whether packaging.Requirement already doesn't have the name attribute normalized so we could potentially just do req.name, but I admit I'm unsure on this one particularly.

cachi2/core/package_managers/pip.py
ben-alkov: > but given it's specified as required at the top of pyproject.toml it'll get installed anyway
Yes, it's this. That's why I recommended removing it from direct dependencies - it was *only* there for `pkg_resources`, we're removing `pkg_resources`, so 'setuptools' should be removed from direct dependencies *in the same commit*

pyproject.toml
ben-alkov: I advised ianrochapg to directly use 'jaraco.text' - I didn't realize that it was so closely related to 'pkg_resources', and both of us were quite puzzled as to why it was being vendored there in the first place :D

cachi2/core/package_managers/pip.py
eskultety: > we're removing pkg_resources, so 'setuptools' should be removed from direct dependencies in the same commit

Not necessarily in the same one, if you mix everything that looks like it's somewhat related in a single commit it just creates noise, you only ever absolutely need to pair changes which would otherwise lead to broken builds or failing CI (on that particular commit) within a single commit which this is not the case, it'll work just fine and so the dependency can be dropped later and make the changes more linear, gradual and cleaner to review.

cachi2/core/package_managers/pip.py
eskultety: @ianrochapg I see you pushed a new revision, have you looked at the canonicalize functions I pointed to in my earlier comment to see if it worked and integration tests passed? If so, that would be obviously a win for us because we wouldn't need the 2 helpers you're introducing.

eskultety:
@ianrochapg if you don't plan on not actually working on subclassing `packaging.Requirement` inside our code, that's fine, but in that case the subject of both the PR and the commit need to drop any mention of "subclassing". Additionally, the 'resolves' reference should be dropped as well from the PR description since this PR is only one of the steps towards resolving the issue, but not quite, so on its own the issue should not get closed automatically via this PR.

eskultety:
Any update on this one?

cachi2/core/package_managers/pip.py
ianrochapg: I ran a few tests and after changing from `pkg_resources.parse_requirements` to `Requirement` directly, we do need to ensure it's normalized.

cachi2/core/package_managers/pip.py
ianrochapg: After checking it out, we can use `version`, but `name` doesn't cover every scenario required.

slimreaper35: I wonder if we can use `canonicalize_name` alongside `canonicalize_version` from packaging...

ianrochapg: @slimreaper35 It does work, I just thought I didn't need to import something else since I do need `_standardize_name` for line 316 and it covers even more scenarios than `canonicalize_name`.

cachi2/core/package_managers/pip.py
brunoapimentel: So, I dug a little deeper into this. To keep the previous behavior, we need to normalize here. 

Instantiating a req will allow the name to be anything that matches this [regex](https://github.com/pypa/packaging/blob/main/src/packaging/_tokenizer.py#L82). This means that `_` is allowed in the requirement line that is being parsed.

Previously, `project_name` was normalized using [safe_name](https://github.com/pypa/setuptools/blob/main/pkg_resources/__init__.py#L1550) from pkg_resources, which handles the `_` in the very least.

cachi2/core/package_managers/pip.py
brunoapimentel: I noticed there's actually a difference in behavior between `canonicalize_name` and `_standardize_name`.

```
>>> canonicalize_name("aio.http")
'aio-http'
>>> _standardize_name("aio.http")
'aio.http'
```

So, _standardize does allow for `.`, whereas canonicalize does not. 

In the end, for [requirement.package](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/pip.py#L1260), I don't know if it causes any issue, since `canonicalize_name` will be called regardless for PyPI deps (see [1](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/pip.py#L1879), [2](https://github.com/jwodder/pypi-simple/blob/master/src/pypi_simple/client.py#L278), [3](https://github.com/jwodder/pypi-simple/blob/master/src/pypi_simple/client.py#L10)). 

I did not dig deeper to see if it impacts file or git dependencies, but if it does not, we could just keep using `req.name`.  Otherwise, to keep the previous behavior, we want `_standardize_name`.

cachi2/core/package_managers/pip.py
brunoapimentel: This other [comment](https://github.com/containerbuildsystem/cachi2/pull/542#discussion_r1688793036) is relevant for here as well.

brunoapimentel (commented):
This LGTM, unless we want to dig deeper and see if we can drop the use of `_standardize_name`. The only nitpick I have is about the commit message, I think it is a little confusing.

cachi2/core/package_managers/pip.py
slimreaper35: The only problem that we have currently is this line (otherwise, we wouldn't need `_standardize` at all):
```
@pytest.mark.parametrize("repo_name_with_subpath", ["bar-subpath", None])
```
`/` are replaced with `-`

Is it the same for all package managers?

brunoapimentel: It is definitely not the same for gomod, as the module's name always contains a `/`. Not sure if npm/yarn allow for `/` in the package's name. Maybe we should just keep the normalize for this specific [instance](https://github.com/ianrochapg/cachi2/blob/pip-requirement-subclass/cachi2/core/package_managers/pip.py#L316), and keep the current behavior.

I'm still not sure if we can completely remove the _standardize from the other [instance](https://github.com/ianrochapg/cachi2/blob/pip-requirement-subclass/cachi2/core/package_managers/pip.py#L1314). Keeping it will maintain the same behavior prior to this PR.

slimreaper35: > I'm still not sure if we can completely remove the _standardize from the other [instance](https://github.com/ianrochapg/cachi2/blob/pip-requirement-subclass/cachi2/core/package_managers/pip.py#L1314). Keeping it will maintain the same behavior as the current code has.

As I mentioned above, we can replace it with `canonicalize_name` from the packaging package.

Well, I don't know then, I can see `/` in the rpm SBOM too.

brunoapimentel: > As I mentioned above, we can replace it with canonicalize_name from the packaging package.
Yes, that would work, but we wouldn't be able to completely get rid of `_standardize_name` since it is used to normalize package names that are resolved from a URL.

cachi2/core/package_managers/pip.py
eskultety: > Not objecting to this line but I'm unclear whether packaging.Requirement already doesn't have the name attribute normalized so we could potentially just do req.name, but I admit I'm unsure on this one particularly.

The answer is NO, `packaging.Requirement` doesn't have the name normalized yet (but there is a TODO comment, so maybe one day :) ): https://github.com/pypa/packaging/blob/bb61b7bcbc429b49e6659a8b6c2966b6ebd046b6/src/packaging/requirements.py#L32

cachi2/core/package_managers/pip.py
eskultety: I admit I got lost in this thread quite a bit already (I have switched the context a long time ago). So I'll need some refreshment and more explanation I guess in layman terms on why do we want to stick to `_standardize_name` when that regex isn't even [PEP 503](https://peps.python.org/pep-0503/#normalized-names) compatible? Conversely, `canonicalize_name` is [1].

So what am I missing here in terms of normalization?  While the proposed behaviour is in check with the old `pkg_resources` 's `safe_name` [2] behaviour it doesn't adhere to PEP. Shouldn't we strive all Python code we write to conform to PEP standards?

[1]  https://github.com/pypa/packaging/blob/bb61b7bcbc429b49e6659a8b6c2966b6ebd046b6/src/packaging/utils.py#L48C26-L48C33
[2] https://github.com/pypa/setuptools/blob/main/pkg_resources/__init__.py#L1550

brunoapimentel: There are two problems that got tangled here, and both relate to the usage of `_standardize_name`:

1- we [use](https://github.com/ianrochapg/cachi2/blob/pip-requirement-subclass/cachi2/core/package_managers/pip.py#L316) it to normalize the name of the main package (the one that is being processed). We need to keep it here since `canonicalize_name` allows `/` in the name, and the current behavior is to replace `/` to `-`. `/` is a valid use case here because the name can be inferred from the repository's URL.

2- we also [use](https://github.com/ianrochapg/cachi2/blob/pip-requirement-subclass/cachi2/core/package_managers/pip.py#L1314) it to normalize the other package's names (dependencies). We can definitely use `canonicalize_name` here, since it covers all use cases.

Ian opted to keep both instances of `_standardize_name` since we wouldn't be able to get rid of it, but I agree we can change the second instance to `canonicalize_name`.

We could probably change the current behavior and allow for `/` in the main package name, but I personally wouldn't want to go this far, at least not in this PR.

brunoapimentel: Also, PEP 503 targets Simple Repository API, and in case "1" above, we're talking about how the name will be presented in the resulting SBOM.

eskultety: > 1- we [use](https://github.com/ianrochapg/cachi2/blob/pip-requirement-subclass/cachi2/core/package_managers/pip.py#L316) it to normalize the name of the main package (the one that is being processed). We need to keep it here since canonicalize_name allows / in the name, and the current behavior is to replace / to -. / is a valid use case here because the name can be inferred from the repository's URL.


> Also, PEP 503 targets Simple Repository API, and in case "1" above, we're talking about how the main component name will be presented in the resulting SBOM.

Now I see @brunoapimentel . So, how about instead of inventing our own function for a single use case, we additionally filter out `/` from the name we get out of canonicalize to retain the current behavior wrt/ SBOM?

brunoapimentel: I'm ok with that solution :+1: 

I'm sure that doing `canonicalize + replace '/'` will make all tests pass, but I think it's best to double check if we're not uncovering any other corner case with this change. 

(sorry, the detailed info I had about this was already garbage collected :brain: :rofl:)

eskultety: So, in reality, there's a bunch of characters that would get the '-' treatment from the old normalization and not canonicalize, like e.g. '@' , then the question is whether that's a valid character and PEP says it is, so I guess like with anything we can go with canonicalize provided existing tests pass and fix other occurrences on the fly if needed, it's an easy fix.

ianrochapg: @brunoapimentel @eskultety Per your suggestion, I replaced `standardize_name` with `canonicalize_name` while also replacing the `/` with a `-` .

a-ovchinnikov (approved):
LGTM

brunoapimentel (approved):

slimreaper35:
:champagne: I assume we can close https://github.com/containerbuildsystem/cachi2/issues/482 and https://github.com/containerbuildsystem/cachi2/issues/539.

PR #544, opened 2024-05-29
Convert the cachi2 image to a multi stage docker build

eskultety:
The motivation behind this PR stems from the following:

1. ever since we started pre-installing Go SDKs into the images the image size grew out of hand (>2x)
2. the injected Go SDK installs make it impossible (among a couple of other blockers) for cachi2 itself be hermetically built because we'd have to come up with an ugly mechanism to prefetch the Go SDKs to be injected into the final image
3. slow image builds due to the size of the Go SDKs that had to be downloaded each time; this is particularly annoying with running integration tests locally since the image is rebuilt with every commit when run in an interactive rebase
     _**EDIT: Tests have shown that the RPM layers in a multi-stage scenario are rebuilt every time, so ^this issue will only be alleviated only slightly**_
5. we have an artificial dependency on nodejs via the `js-deps` directory which was only ever meant to be a build aid to make sure that:
    a) we're working with new enough Node that ships with corepack
    b) we're enabling and hence installing the correct Yarn version (v3) in the final image (Note: Yarn installation needs a namespace not to create artifacts on the global level for which we used the `js-deps` directory) 
6. Installing cachi2 and all its dependencies via pip to the global pip module store as root (i.e. /usr/lib)

~~**Depends on: https://github.com/containerbuildsystem/cachi2/pull/528**~~


Containerfile
```diff
@@ -1,47 +1,55 @@
...
+FROM base
+LABEL maintainer="Red Hat"
 
-WORKDIR /src/js-deps
-RUN npm install && \
-    ln -s "${PWD}/node_modules/.bin/corepack" /usr/local/bin/corepack && \
-    corepack enable yarn && \
-    microdnf -y remove npm
+# copy Go SDKs from official Debian images, corepack from official Node.js Alpine
+COPY --from=docker.io/library/golang:1.20.0-bullseye /usr/local/go /usr/local/go/go1.20
```
ben-alkov, line 44: Do we build this often enough to need to worry about DockerHub rate limits?

eskultety: That's a very good question but because I don't have admin rights on the org and not seeing all the settings I wonder if caching could be used, i.e. cache the intermediary images indefinitely and just copy from them which would actually be awesome because it'd speed up even GH builds by a huge margin.

Containerfile
```diff
@@ -1,45 +1,55 @@
-FROM registry.fedoraproject.org/fedora-minimal:39
-LABEL maintainer="Red Hat"
+# hadolint global ignore=DL3007
```
MartinBasti, line 1: Why do you want to use latest? I'd prefer dependabot to update versions and have reproducible builds in case we need to revert

Containerfile
```diff
@@ -1,45 +1,55 @@
...
 
-WORKDIR /src
+########################
+# PREPARE OUR BASE IMAGE
+########################
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest as base
```
MartinBasti, line 6: Using `latest` is not considered safe, it should be treated as antipattern

Containerfile
MartinBasti: I'd not be concerned about dockerlimits, we can provide docker credentials if needed

Containerfile
MartinBasti: and even better if floating tags are pinned to digests too

Containerfile
eskultety: Relying on docker credentials only exhibits the strong ties to a particular organization behind the project which IMO doesn't foster the OSS community principles, IIUC docker should have a way to cache the intermediary layers during a multi-stage build, although admittedly, everything I could find looked really hackish. Agreed with @MartinBasti that we should take care of the problem when we face it.

Containerfile
eskultety: Because this runs on the latest upstream head. With a static image digest we won't be quick enough to react to potentially breaking changes (e.g. testing early with a Fedora Rawhide images to prepare for coming changes in GCC or whatnot, not for gating though!), especially with purposefully decreasing dependabot's cadence of updates to "monthly". Reproducible builds are IMO only part of the testing strategy and part of the whole picture, on one hand you want stable tests in a predictable environment to limit all your potential errors coming from the project itself; on the other hand though, you also want early failures coming from the outside environment (and IMO you want platform ones rather quickly). Not having an image digest is not that much different from actually having one and let dependabot spam updates about it. I'm ambivalent on this, though I'm not fond of dependabot taking care of this since the digest is completely opaque. Does dependabot pair container image digest updates with something like changelog changes, IOW package updates? Because I don't see changelogs being available on docker hub itself and if that's the case then having dependabot opening a PR for a digest update doesn't really improve the situation with using 'latest' straight away.

MartinBasti: With latest you will have breaking changes more often noticing them at very last phase.

With pinning and dependabot you can see what changed and have running CI to avoid merging breaking change and notice them just before relase.

Frankly monthly dependabot updares has low cadence to address CVE in modules. But that's unrelated.

> Not having an image digest is not that much different from actually having one and let dependabot spam updates about it.
It is quite a difference, to have image broken without knowing it and have stable image and see when a update is breaking things in the PR opened by dependabot.

> you also want early failures coming from the outside environment (and IMO you want platform ones rather quickly)
You will have early failures, just in dependabot PR without affecting work in progress work and other PRs randomly.

> digest update doesn't really improve the situation with using 'latest' straight away
I completely disagree, you have no control what's in latest, sometime latest images are even dev versions

Containerfile
MartinBasti: you don't want to mess with layer caching, mainly when you are just installing RPMs and the particular line doesn't change, docker can reuse layer with old RPMs. It's safer to build image without relaying on layers.

And layers won't address the initial problem that you have to pull parent image first and also you want "latest" so you cannot just cache it, as docker would reuse that local tag forever.

We should be fine building this in GH, limits are quite generous.

eskultety: > you don't want to mess with layer caching, mainly when you are just installing RPMs and the particular line doesn't change, docker can reuse layer with old RPMs. It's safer to build image without relaying on layers.

FYI this only works for regular builds, with multi-stage there's more going on and the RPM layers tend to be rebuilt every time (not sure what the reason is I haven't investigated, maybe just ordering).

eskultety:
Since v1:
- pinned the language base image versions to exact version strings (no digests though because we don't care about the underlying platform changes, we only care about the language installations and their data which are bound to be static based on the release they're tracking)
- bumped the NodeJS engine version to 22.3 (latest) because NodeJS 16 (default on Rocky/RHEL) has been EOL for a long time and `corepack` tends to drop support for old engines leading to various crashes (either in corepack or its shims, e.g. Yarn), IOW if we copy the corepack installation from a language base image we need to make sure we also copy the NodeJS installation it is being shipped with in the official image to ensure smooth compatibility (see commit 1)

Dockerfile
```diff
@@ -1,46 +1,56 @@
-FROM docker.io/library/rockylinux:9@sha256:d7be1c094cc5845ee815d4632fe377514ee6ebcf8efaed6892889657e5ddaaa6
-LABEL maintainer="Red Hat"
+FROM docker.io/library/rockylinux:9@sha256:d7be1c094cc5845ee815d4632fe377514ee6ebcf8efaed6892889657e5ddaaa6 as rockylinux9
+FROM docker.io/library/golang:1.20.0-bullseye as golang_120
```
MartinBasti, line 2: Why not `golang:1.20.14-bullseye` 1.20.0 is quite old

eskultety: To be fair, 1.21.0 is old too but there's a reason for the X.Y.0th release choices - admittedly though, it only makes sense with 1.21+ where that allows us to fetch literally **any** newer toolchain user may require in their project and cache it as a module properly too (it's a consistent behaviour where a correct toolchain is always downloaded and we don't let Go figure out on its own whether it's bundled toolchain is new enough to process the project) .

The choice for 1.20.0 was mostly for consistency reasons with the 1.21.0 reasoning above and that's pretty much it; it also should be the toolchain used in minimum cases as 1.20 has been EOL'd already and so I plan on ditching it soon in follow-up PRs, so up to you, if you prefer pre-installing a newer 1.20 release (this would happen in a standalone commit though, just like with NodeJS to remain true to the status quo in the multi-stage conversion commit) before ditching it altogether, I can do it.

MartinBasti: Ok, I see now. Adam also told me, that we need older version of go to dynamically download newer toolchains

MartinBasti (approved):

slimreaper35 (approved):

brunoapimentel (approved):
The new

eskultety:
Rebased on top of main.

PR #430, opened 2023-12-15
docs(test): writing integration tests

ben-alkov:
Creating as a draft cuz it's a little rough, but I wanted to make sure I captured the essence and the specific commands while they were fresh in my head. Everything's on the table! Let me know if there's *anything* you'd like to do differently.

also describes the local development process:

- processing test data with cachi2
- running pytest

Closes #836

Signed-off-by: Ben Alkov <ben.alkov@redhat.com>

# Maintainers will complete the following section

- [ ] Commit messages are descriptive enough
- [ ] Code coverage from testing does not decrease and new code is covered
- [ ] Docs updated (if applicable)
- [ ] Docs links in the code are still valid (if docs were updated)

**Note:** if the contribution is external (not from an organization member), the CI
pipeline will not run automatically. After verifying that the CI is safe to run:

- [approve GitHub Actions workflows][approve-gh-actions] by clicking a button
- approve the Red Hat Trusted App Pipeline container build by commenting `/ok-to-test`
  (as is the [standard for Pipelines as Code][pac-running-pipeline])

[approve-gh-actions]: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
[pac-running-pipeline]: https://pipelinesascode.com/docs/guide/running/#running-the-pipelinerun

PR #545, opened 2024-05-30
e2e tests: Containerfile presence

slimreaper35:


slimreaper35:
@everyone
Are you okay with this change ? WDYT
In the future, when creating new e2e tests, part of the repo _can / should_ be Containerfile, too.

I think it will make life easier when testing locally. Also, I suppose users have their Containerfile in the same repo when using cachi2.

Relates to #430

tests/integration/utils.py
```diff
@@ -401,11 +401,13 @@ def build_image_and_check_cmd(
...
 
-    with build_image_for_test_case(
-        tmp_path, str(container_folder.joinpath("Containerfile")), test_case
-    ) as test_image:
+    containerfile = tmp_path.joinpath("Containerfile")
+    if not containerfile.exists():
+        container_folder = test_data_dir.joinpath(test_case, "container")
+        containerfile = container_folder.joinpath("Containerfile")
+
+    with build_image_for_test_case(tmp_path, str(containerfile), test_case) as test_image:
```
eskultety, line 410: While I appreciate you're thinking of doing this in a compatible manner, it's actually going to lead to a very messy state where part of the tests will have the containerfile inside the e2e repo and the rest will have it in cachi2 integration test data. I'd prefer if this was proposed as a complex change == move the existing container files out of cachi2 and mandate that new e2e repos follow the proposed practice. Which brings me to #430 which this change would be nicely paired with that docs where this practice would be explained.

PR #551, opened 2024-06-06
Support configurable index urls for pip and private registry auth (not just) for pip

chmeliik:
TLDR:

* `--index-url` in requirements.txt is no longer ignored
* If the index url is anything other than https://pypi.org/simple/, report it in the SBOM
* Tell aiohttp not to ignore `.netrc`
* Add way to run local PyPI server that requires auth, add integration test that uses it

Run new test with:

```shell
make test-integration TEST_LOCAL_PYPISERVER=true TOX_ARGS='-k custom_index'
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 PYTHON_VERSION_VENV ?= python3
 TOX_ENVLIST ?= py39
 TOX_ARGS ?=
+GENERATE_TEST_DATA = false
 
 .PHONY: clean
 all: venv
@@ -29,13 +30,13 @@ test-unit: venv
 	venv/bin/tox -e $(TOX_ENVLIST) -- $(TOX_ARGS)
 
 test-integration: venv
-	venv/bin/tox -e integration
+	CACHI2_GENERATE_TEST_DATA=$(GENERATE_TEST_DATA) venv/bin/tox -e integration
 
 mock-unittest-data:
 	hack/mock-unittest-data/gomod.sh
 
-generate-test-data: venv
-	CACHI2_GENERATE_TEST_DATA=true venv/bin/tox -e integration
+generate-test-data: GENERATE_TEST_DATA = true
+generate-test-data: test-integration
 
 build-image:
 	podman build -t localhost/cachi2:latest .
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ test-unit: venv
 test-integration: venv
 	CACHI2_GENERATE_TEST_DATA=$(GENERATE_TEST_DATA) \
 	CACHI2_TEST_LOCAL_PYPISERVER=$(TEST_LOCAL_PYPISERVER) \
-		venv/bin/tox -e integration
+		venv/bin/tox -e integration -- $(TOX_ARGS)
 
 mock-unittest-data:
 	hack/mock-unittest-data/gomod.sh
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ PYTHON_VERSION_VENV ?= python3
 TOX_ENVLIST ?= py39
 TOX_ARGS ?=
 GENERATE_TEST_DATA = false
+TEST_LOCAL_PYPISERVER = false
 
 .PHONY: clean
 all: venv
@@ -30,7 +31,9 @@ test-unit: venv
 	venv/bin/tox -e $(TOX_ENVLIST) -- $(TOX_ARGS)
 
 test-integration: venv
-	CACHI2_GENERATE_TEST_DATA=$(GENERATE_TEST_DATA) venv/bin/tox -e integration
+	CACHI2_GENERATE_TEST_DATA=$(GENERATE_TEST_DATA) \
+	CACHI2_TEST_LOCAL_PYPISERVER=$(TEST_LOCAL_PYPISERVER) \
+		venv/bin/tox -e integration
 
 mock-unittest-data:
 	hack/mock-unittest-data/gomod.sh
--- a/cachi2/core/package_managers/pip.py
+++ b/cachi2/core/package_managers/pip.py
@@ -94,6 +94,7 @@ class DistributionPackageInfo:
     package_type: Literal["sdist", "wheel"]
     path: Path
     url: str
+    index_url: str
     is_yanked: bool
 
     pypi_checksums: set[ChecksumInfo] = field(default_factory=set)
@@ -245,6 +246,9 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
 
     if dependency_kind == "pypi":
         version = package["version"]
+        index_url = package["index_url"]
+        if index_url.rstrip("/") != pypi_simple.PYPI_SIMPLE_ENDPOINT.rstrip("/"):
+            qualifiers = {"repository_url": index_url}
     elif dependency_kind == "vcs":
         qualifiers = {"vcs_url": package["version"]}
     elif dependency_kind == "url":
@@ -1493,6 +1497,7 @@ def _process_req(
             _check_metadata_in_sdist(dpi.path)
 
         download_info["package_type"] = dpi.package_type
+        download_info["index_url"] = dpi.index_url
 
     else:
         if require_hashes or req.kind == "url":
@@ -1898,6 +1903,7 @@ def _process_package_distributions(
             cast(Literal["sdist", "wheel"], package.package_type),
             pip_deps_dir.join_within_root(package.filename).path,
             package.url,
+            index_url,
             package.is_yanked,
             pypi_checksums,
             user_checksums,
@@ -2161,6 +2167,7 @@ def _resolve_pip(
             "name": dep["package"],
             "version": _version(dep),
             "package_type": dep["package_type"],
+            "index_url": dep.get("index_url"),
             "type": "pip",
             "dev": dep.get("dev", False),
             "kind": dep["kind"],
--- a/cachi2/core/package_managers/pip.py
+++ b/cachi2/core/package_managers/pip.py
@@ -1517,13 +1517,14 @@ def _process_pypi_req(
     req: PipRequirement,
     requirements_file: PipRequirementsFile,
     require_hashes: bool,
+    index_url: str,
     pip_deps_dir: RootedPath,
     allow_binary: bool,
 ) -> list[dict[str, Any]]:
     download_infos: list[dict[str, Any]] = []
 
     artifacts: list[DistributionPackageInfo] = _process_package_distributions(
-        req, pip_deps_dir, allow_binary
+        req, pip_deps_dir, allow_binary, index_url
     )
 
     files: dict[str, Union[str, PathLike[str]]] = {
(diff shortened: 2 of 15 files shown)


chmeliik:
This should be everything that's needed for cachi2 to support custom PyPI servers, including auth.

Auth is done via .netrc, same as pip: https://pip.pypa.io/en/stable/topics/authentication/#netrc-support.
Pip also supports keyring auth but the docs make it sound rather strange. Doesn't seem worth supporting.

Let me know if you would like an ADR for any of this :saluting_face: 

TODO: make the integration test run in CI, ideally without copy-pasting most of the Makefile target

taylormadore:
Wow, this did indeed end up being quite straightforward. LGTM so far :+1:

chmeliik:
I've made the CI work with the new int. test, should be ready for review

Makefile
```diff
@@ -32,7 +33,14 @@ test-unit: venv
 	venv/bin/tox -e $(TOX_ENVLIST) -- $(TOX_ARGS)
 
 test-integration: venv
-	CACHI2_GENERATE_TEST_DATA=$(GENERATE_TEST_DATA) venv/bin/tox -e integration
+	if [ $(TEST_LOCAL_PYPISERVER) = true ]; then \
+		PYPISERVER_PORT=$(PYPISERVER_PORT) PYPISERVER_IMAGE=$(PYPISERVER_IMAGE) tests/pypiserver/start.sh & \
+		trap "kill $$!" EXIT && \
+		while true; do sleep 1; curl -fsS http://localhost:$(PYPISERVER_PORT) && break; done; \
+	fi && \
```
eskultety, line 40: I am very much not a fan of abusing the Makefile for this purpose (particularly to use Bash as a crutch instead of modeling this properly in the Python test suite). It really only should be a recipe for tedious build/test/release processes. Now that said, one might argue that this indeed is related to the "test" part I mentioned in the previous sentence, but not quite - configuring tox is a good example of that. If we optionally need a test server deployed for a particular test's needs then I believe we should do it part of the test suite and not part of the Makefile.
Related to your `integration/conftest.py` changes, is there anything preventing us to model this server as a test fixture with a cleanup section based on Python's subprocess? After all, you guard the test with a pytest  `skip` mark, so such a fixture should not run in that case and it would be nicely contained within the test suite where it belongs.

Bonus points if we could reuse the `utils.ContainerImage` class to do this which I think it should be possible looking at the command line the container gets.

tests/integration/conftest.py
```diff
@@ -16,19 +16,20 @@ def test_data_dir() -> Path:
...
-        cachi2 = utils.build_image(cachi2_repo_root, tag="localhost/cachi2:latest")
-    else:
-        cachi2 = utils.ContainerImage(cachi2_image_ref)
-        if not cachi2_image_ref.startswith("localhost/"):
-            cachi2.pull_image()
+        utils.build_image(cachi2_repo_root, tag=cachi2_image_ref)
+
+    cachi2 = utils.Cachi2Image(cachi2_image_ref)
+    if not cachi2_image_ref.startswith("localhost/"):
+        cachi2.pull_image()
```
eskultety, line 33: nitpick: This whole hunk seems a little bit out of place in this commit, there's already quite a bit happening and this is easily missed, so I think this adjustment along with the `utils.ContainerImage` (not `utils.Cachi2Image` though) deserve their own spotlight in a separate commit.

tox.ini
```diff
@@ -62,6 +62,9 @@ commands =
 passenv =
     CACHI2_GENERATE_TEST_DATA
     CACHI2_IMAGE
+    CACHI2_TEST_LOCAL_PYPISERVER
+setenv =
+    CACHI2_TEST_NETRC_CONTENT={env:CACHI2_TEST_NETRC_CONTENT:machine localhost login cachi2-user password cachi2-pass}
```
eskultety, line 67: Why do we want to make `CACHI2_TEST_NETRC_CONTENT` variable configurable from the outside for test purposes? Isn't a fixed value enough? When would we want to reconfigure the test differently?

eskultety (commented):
Conceptually, no objections. I left a few suggestions and nitpick comments.

cachi2/core/package_managers/pip.py
```diff
@@ -1853,7 +1860,9 @@ def _process_package_distributions(
     :rtype: list[DistributionPackageInfo]
     """
     allowed_distros = ["sdist", "wheel"] if allow_binary else ["sdist"]
-    client = pypi_simple.PyPISimple()
+    client = pypi_simple.PyPISimple(
+        endpoint=index_url if index_url else pypi_simple.PYPI_SIMPLE_ENDPOINT
```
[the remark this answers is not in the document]
chmeliik: Good call, changed

Makefile
chmeliik: I can try that, but I'd like to keep the `pypiserver-run` makefile target as well

The process will need to run in the background, not sure how pretty that will be

eskultety: > The process will need to run in the background, not sure how pretty that will be

I'm not sure I follow, `subprocess.Popen` doesn't block the main process, IIUC if you put the Popen call to a fixture followed by a `yield` and then for the "cleanup" kill and wait for the server process that should theoretically work, shouldn't it?

chmeliik: > pytest eats the stdout and stderr from the subprocess

Although with `--capture=sys` or `--capture=tee-sys` passed to pytest, this looks better. Still need to find out what those do though

eskultety: Sigh, true. I guess if only

>  if the server fails to start, you'll never find out why (the output is logged when the tests finish

needs to be solved, I think redirecting podman's stdout to a file instead of the parent's process stdout (which pytest consumes) then that should also work, but then

> but it's already a bit worse than bash

will still remain true I suppose. Can we at least extract the Makefile bits, stuff it in a bash script and bury it somewhere under the integration test directory hierarchy? That shouldn't have an effect on the github action you needed the dedicated makefile target for.

eskultety: > Although with --capture=sys or --capture=tee-sys passed to pytest, this looks better. Still need to find out what those do though

Looking at the docs `tee-sys` reads exactly like what we need.

chmeliik: Added a WIP commit for what it would look like, see if you like it better

tests/integration/test_pip.py
```diff
@@ -1,15 +1,48 @@
...
+        except requests.RequestException as e:
+            log.debug(e)
+    else:
+        proc.terminate()
+        raise RuntimeError("pypiserver didn't start fast enough")
+
+    yield
+    proc.terminate()
+
+
```
eskultety, line 45: nitpick: not a big deal but strictly from correctness POV we should clean up the zombie process explicitly despite either Python runtime or whatever the parent is will take care of it.

Makefile
eskultety: > Added a WIP commit for what it would look like, see if you like it better

Unsurprisingly, it led to more lines in Python, but the code is still perfectly readable and contained within the test suite, nicely done!

tests/integration/test_pip.py
chmeliik: fixed

tests/integration/conftest.py
chmeliik: Ack, split the changes

tox.ini
chmeliik: I used it for example to see what happens when the password is wrong. It's not strictly necessary, but I would expect the tests to respect the env var if I try to override it

chmeliik:
One more small improvement: used contextlib.ExitStack to make absolutely sure the process get `.terminate()`-ed before python starts waiting for it to clean it up

tests/integration/conftest.py
```diff
@@ -28,8 +33,39 @@ def cachi2_image() -> utils.ContainerImage:
...
+
+
+# autouse=True: It's nicer to see the pypiserver setup logs at the beginning of the test suite.
+# Otherwise, pypiserver would start once the pip tests need it and the logs would be buried between
+# test output.
+@pytest.fixture(autouse=True, scope="session")
+def local_pypiserver() -> Iterator[None]:
+    if os.getenv("CACHI2_TEST_LOCAL_PYPISERVER") != "true":
+        yield
+        return
```
eskultety, line 50: You can use `mark.skipif` on fixtures as well which will enhance readability by just that tiny bit here :).

Makefile
```diff
@@ -32,6 +32,9 @@ test-unit: venv
 test-integration: venv
 	CACHI2_GENERATE_TEST_DATA=$(GENERATE_TEST_DATA) venv/bin/tox -e integration
 
+pypiserver-run:
+	tests/pypiserver/start.sh
+
```
eskultety: I'm still not convinced by this target, even more so if with all the latest changes this was reduced to a mere invocation of a script. I'd encourage it if the pypi server were to serve as a development deployment of a crucial component to the project, but it's not, we only need it for a specific test which I expect will be the only use case for a long time. If anything, we could document this as part of the integration test suite section mentioning that one can use the script to run a server locally.

tests/pypiserver/start.sh
```diff
@@ -0,0 +1,62 @@
...
+}
+
+WORKDIR=$(mktemp -d --tmpdir "cachi2-pypiserver.XXXXXX")
+trap cleanup EXIT
+
+setup() {
+    echo -e "\n--- Downloading $DIR/package-urls.txt ---\n"
+
+    mkdir "$WORKDIR/packages"
+    sed '/^#/d' "$DIR/package-urls.txt" |
```
eskultety, line 22: wouldn't `tail -n +2` to skip the first line be actually more straightforward and slightly more readable?

tests/integration/conftest.py
```diff
@@ -19,16 +19,17 @@ def test_data_dir() -> Path:
 def cachi2_image() -> utils.ContainerImage:
     cachi2_image_ref = os.environ.get("CACHI2_IMAGE")
```
eskultety, line 25: **Commit message:**
`s/once/one`     perhaps?

eskultety (commented):
Looking good, a couple more nitpics and I think we're ready to go.

chmeliik:
Just a rebase, will address comments in next force-push

Makefile
chmeliik: Ack, removed and added to the README instead

tests/integration/conftest.py
chmeliik: Tried this, but it doesn't seem to work https://github.com/pytest-dev/pytest/issues/964

Also tried `pytest.skip` instead of `yield; return`, but it looks like that also skips the tests that rely on this fixture (and together with `autouse=True`, that means skip all tests :smile: )

tests/pypiserver/start.sh
chmeliik: I went with a "comment" syntax, in case I wanted to add more packages to download (and more comments explaning how I got the urls)

tests/integration/conftest.py
eskultety: > Tried this, but it doesn't seem to work

sigh, I guess I forgot to remove the existing code when I tried it which falsely led me to thinking that it works, sorry.

eskultety (approved):

brunoapimentel (approved):
I did not test any external private registry, but the approach, code and tests LGTM.

Issue #550, opened 2024-06-03
[Go >=1.21+] Log the actual toolchain to be used for a given module instead of the pre-installed toolchain


eskultety:
Currently, we report the toolchain we selected to process a project, e.g. 1.21.0 (pre-installed) but that toolchain might not end being the one to be used for the given project as the project may specify any micro release for the toolchain it desires (e.g. 1.21.8) which we're currently handling by passing GOTOOLCHAIN=auto to the underlying Go process which will first download the desired toolchain and then process the requested command (and project) with it.
We should ideally report the target toolchain that we're invoking instead of the intermediary toolchain that will just download the target one.

eskultety:
I worried for a moment that we'd fall into the trap of reporting the wrong toolchain due to automatic toolchain switching Go's capable of (e.g. `dummy-lib@v0.0.1 requires go >= 1.21.7; switching to go1.21.10`), but that only happens during `go build` and only after `go mod tidy` (e.g. without running `tidy`: `dummy-lib@v0.0.1 requires go >= 1.21.7 (running go 1.21.4)`). In either case, for a project dependency prefetch, all of the mentioned above is irrelevant as the toolchain declared  by the main project has always has to be >= to the highest toolchain version of all of its dependencies, IOW the `toolchain` keyword specified by a dependency only gains meaning when building the dependency itself as the main module, but not when building it indirectly; as confirmed by [docs](https://go.dev/doc/toolchain):  

> Modules that are dependencies of other modules may need to set a minimum Go version requirement lower than the preferred toolchain to use when working in that module directly. In this case, the toolchain line in go.mod or go.work sets a preferred toolchain that takes precedence over the go line when the go command is deciding which toolchain to use.

PR #553, opened 2024-06-10
Go 1.22 vendored workspaces support

eskultety:
Go 1.22 introduced another toolkit feature that affects cachi2 directly - [vendored workspaces](https://go.dev/doc/go1.22#tools) meaning that projects can now combine the vendoring feature (local module replacements) and workspaces (controlling local modules centrally wrt/ dependencies and language version requirements).

Go enabled this via a new subcommand `go work vendor` which needs to be run instead of the existing `go mod vendor`. ~~This looks like the only difference and so we should be able to address it very simply by running that modified vendoring command when we detect the project uses workspaces.~~

~~Given the low complexity of the code no design doc accompanies this PoC as the changes should be straightforward, provided we finish the workspaces support in #457.~~

~~**Note this PoC is based on work proposed in #457 so as such we need to wait for that PR before we consider this to be merged**.~~
 
EDIT: Our existing workspaces code didn't account for the fact that `go list` isn't workspace aware and so we've been missing some module packages in the output SBOM (as noted by comment https://github.com/containerbuildsystem/cachi2/pull/553#discussion_r1654667805). This PR addresses that.

References: https://github.com/containerbuildsystem/cachi2/discussions/398

Notes:
- I successfully built a modified (re-enabled workspaces which upstream disables by the means of `GOWORK=off`) downstream version of [hypershift](https://github.com/openshift/hypershift/tree/release-4.17)
- I successfully prefetched upstream kubernetes and grafana repos


cachi2/core/package_managers/gomod.py
```diff
@@ -1506,6 +1508,7 @@ def parse_module_line(line: str) -> ParsedModule:
 def _vendor_deps(
     go: Go,
     app_dir: RootedPath,
+    has_workspaces: bool,
```
chmeliik, line 1511: Is `app_dir` always the one that contains `go.work`?

It seems reasonable to require that, but the current [`_find_missing_gomod_files`](https://github.com/containerbuildsystem/cachi2/blob/cd9dc05a9e3f2bb384fd750a881df22d023b86e8/cachi2/core/package_managers/gomod.py#L715) check won't always allow it (AFAIK the workspace root doesn't have to also be a module)

eskultety: > Is app_dir always the one that contains go.work

Yes, go will check the existence of a `go.work` file from current directory upwards (from [docs](https://go.dev/ref/mod#workspaces)):

> A command determines whether it is in a workspace context by first examining the GOWORK environment variable. If GOWORK is set to off, the command will be in a single-module context. If it is empty or not provided, the command will search the current working directory, and then successive parent directories, for a file go.work. If a file is found, the command will operate in the workspace it defines; otherwise, the workspace will include only the module containing the working directory. If GOWORK names a path to an existing file that ends in .work, workspace mode will be enabled. Any other value is an error. You can use the go env GOWORK command to determine which go.work file the go command is using. go env GOWORK will be empty if the go command is not in workspace mode.

As for the second part of your question
> (IIUC that's where the vendor/ directory gets created)?

Yes, because this is initialized as `go work vendor` so the vendor directory will be created in the appropriate location (from [docs](https://pkg.go.dev/cmd/go#hdr-Make_vendored_copy_of_dependencies)):

> The -o flag causes vendor to create the vendor directory at the given path instead of "vendor". The go command can only use a vendor directory named "vendor" within the module root directory, so this flag is primarily useful for other tools. 



> It seems reasonable to require that, but the current [_find_missing_gomod_files](https://github.com/containerbuildsystem/cachi2/blob/cd9dc05a9e3f2bb384fd750a881df22d023b86e8/cachi2/core/package_managers/gomod.py#L715) check won't always allow it (AFAIK the workspace root doesn't have to also be a module)

Good point. IIUC your concern we need to update the function to work with workspaces as well, i.e. update the `subpaths` argument so that it corresponds to the local paths parsed from `go.work` JSON representation. That bears the question though - if a workspace module (or any module for that matter) is missing a `go.mod` file, such a project is unbuildable by `go`, isn't it? If that's the case, isn't that a user repo configuration error that we don't need to concern ourselves with too much, or am I missing something @chmeliik ?

chmeliik: > Yes, go will check the existence of a go.work file from current directory upwards (from [docs](https://go.dev/ref/mod#workspaces)):

I was more concerned about the code that reads `app_dir / "vendor" / "modules.txt"` - if go.work isn't directly in app_dir, we fail

> if a workspace module (or any module for that matter) is missing a go.mod file, such a project is unbuildable by go

Wouldn't each of the modules referenced by go.work be buildable by themselves? I think workspaces don't have to be modules, IIUC one use case is just managing a collection of submodules

chmeliik: > Wouldn't each of the modules referenced by go.work be buildable by themselves? I think workspaces don't have to be modules, IIUC one use case is just managing a collection of submodules

i.e. I think we should stop requiring the presence of `go.mod`, `go.work` on its own should also be allowed

eskultety: > I was more concerned about the code that reads app_dir / "vendor" / "modules.txt" - if go.work isn't directly in app_dir, we fail

Agreed that `app_dir` needs to be dynamically tweaked depending on whether we're dealing with workspaces or not. I'll keep this in mind and once #457 is close to merging I'll adjust the code - there might potentially be more  places needed fixing, although maybe not as most Go commands are workspaces aware, so no matter where you execute them from, the context will always be the whole workspace.


> I think workspaces don't have to be modules, IIUC one use case is just managing a collection of submodules

Based on [docs](https://go.dev/ref/mod#workspaces):
_A workspace is a collection of modules on disk that are used as the main modules_ 

IIUC the ^above, everything that a workspace manages is a regular module, i.e. should have a `go.mod` file, otherwise it's just a package isn't it?

chmeliik: > A workspace is a collection of modules on disk that are used as the main modules
>
> IIUC the ^above, everything that a workspace manages is a regular module, i.e. should have a go.mod file, otherwise it's just a package isn't it?

I'm probably using the terminology wrong, what I meant is that the workspace _root_ - the directory that contains go.work - may not have to be a module. All the modules in the `use` section of go.work do have to be modules

eskultety: > i.e. I think we should stop requiring the presence of go.mod, go.work on its own should also be allowed

What do you mean? Not requiring `go.mod` on the "application root" level next to a potentially existing `go.work` file or not requiring its presence in the workspace managed modules? I agree with the former, while I disagree with the latter provided my understanding of https://go.dev/ref/mod#workspaces is correct specifically about the collection of modules.

In any case, I think this consideration is probably better suited at #457 rather than here. We'll likely need to adjust the original design of the feature anyway to let the user to point us to the work file instead of the mod file since with workspaces there are now multiple "main" modules. I think being pointed to a work file instead is going to be beneficial especially in those cases where no top-level `go.mod` file exists and it would otherwise be confusing, since one can extract everything necessary out of the `go.work` file.

chmeliik: But if the intention is that the user will need to point to one of the modules and the `app_dir` will be tweaked accordingly, then this is fine. The ability to just point to the workspace root can be added later

eskultety: with #457 merged and the decision to address the architecture for Go project processing in a follow up series, marking this as resolved.

eskultety:
Since v1:
- rebased on top of #457 
- added an e2e integration test (to verify that a project can actually be built with a vendored workspace)

@chmeliik @brunoapimentel I'm ambivalent on the decision to test this in an e2e test (I did make sense to me to verify building with everything vendored, i.e. we didn't need to fetch anything, also there's a lack of real projects to test the feature with), so let me know if you want me to model it as a basic integration test or e2e is fine.

cachi2/core/package_managers/gomod.py
```diff
@@ -1525,6 +1527,7 @@ def parse_module_line(line: str) -> ParsedModule:
 def _vendor_deps(
     go: Go,
     app_dir: RootedPath,
```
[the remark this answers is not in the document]
chmeliik: Tested with https://github.com/cachito-testing/gomod-pandemonium/tree/go-1.22-workspace-vendoring

```bash
cachi2 --log-level debug fetch-deps '{"type": "gomod", "path": "."}' --gomod-vendor
cp cachi2-output/bom.json root-bom.json

cachi2 --log-level debug fetch-deps '{"type": "gomod", "path": "terminaltor"}' --gomod-vendor
cp cachi2-output/bom.json submodule-bom.json

get_module_purls() {               
    jq < "$1" '.components[].purl | select(test("type=module"))' -r | sort -u
}

comm -23 <(get_module_purls root-bom.json) <(get_module_purls submodule-bom.json)
```

```text
pkg:golang/github.com/Azure/go-ansiterm@v0.0.0-20210617225240-d185dfc1b5a1?type=module
pkg:golang/github.com/google/pprof@v0.0.0-20210407192527-94a9f03dee38?type=module
pkg:golang/github.com/go-task/slim-sprig@v0.0.0-20230315185526-52ccab3ef572?type=module
pkg:golang/github.com/Microsoft/go-winio@v0.6.0?type=module
pkg:golang/golang.org/x/mod@v0.9.0?type=module
```

chmeliik:
> Since v1:
> 
>     * rebased on top of [Add support for go workspaces #457](https://github.com/containerbuildsystem/cachi2/pull/457)
> 
>     * added an e2e integration test (to verify that a project can actually be built with a vendored workspace)
> 
> 
> @chmeliik @brunoapimentel I'm ambivalent on the decision to test this in an e2e test (I did make sense to me to verify building with everything vendored, i.e. we didn't need to fetch anything, also there's a lack of real projects to test the feature with), so let me know if you want me to model it as a basic integration test or e2e is fine.
> 
> **Note: Due to the ^above the integration test currently points to my own fork and based on the decision and the overall acceptance of the integration test I will push the change to the test origin repo.**

Basic would have been enough IMO, but there's no harm in keeping it e2e when it's already e2e

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
chmeliik: Even after fixing the modules.txt parsing, there's still a problem with reporting packages:
```bash
get_nonstd_purls() {                    
    jq < "$1" '.components[].purl | select(test("^pkg:golang/[^/]*\\.[^/]*/"))' -r | sort -u
}

comm -23 <(get_nonstd_purls root-bom.json) <(get_nonstd_purls submodule-bom-fixed.json)
```

```text
pkg:golang/github.com/cachito-testing/gomod-pandemonium@v0.1.1-0.20240626113213-4d50652b5958?type=package
pkg:golang/github.com/cachito-testing/gomod-pandemonium/where-was-i-built@v0.1.1-0.20240626113213-4d50652b5958?type=package
pkg:golang/github.com/cachito-testing/retrodep/v2/retrodep/glide@v2.1.1?type=package
pkg:golang/github.com/cachito-testing/retrodep/v2/retrodep@v2.1.1?type=package
pkg:golang/github.com/Masterminds/semver@v1.4.2?type=package
pkg:golang/github.com/op/go-logging@v0.0.0-20160315200505-970db520ece7?type=package
pkg:golang/github.com/pkg/errors@v0.8.1?type=package
pkg:golang/golang.org/x/sys/execabs@v0.6.0?type=package
pkg:golang/golang.org/x/tools/go/vcs@v0.7.0?type=package
pkg:golang/gopkg.in/yaml.v2@v2.2.2?type=package
```
It looks like `go list ./...` isn't "workspace-aware", so we will need to run that in the workspace root as well. We can probably make that part of the bigger rework

cachi2/core/package_managers/gomod.py
```diff
@@ -787,13 +788,13 @@ def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
...
 
-    if target_version >= go_max_version:
+    if target_version >= go_121_version:
         # Project makes use of Go >=1.21:
         # - always use the 'X.Y.0' toolchain to make sure GOTOOLCHAIN=auto fetches anything newer
         # - container environments need to have it pre-installed
         # - local environments will always install 1.21.0 SDK and then pull any newer toolchain
         go = Go(release="go1.21.0")
-    elif go_base_version >= go_max_version:
+    elif go_base_version >= go_121_version:
```
chmeliik, line 797: I also noticed something weird in the logs when running with `--log-level debug`:

```
2024-06-26 13:59:42,508 INFO go.mod reported versions: '1.22.0'[go], '-'[toolchain]
2024-06-26 13:59:42,508 DEBUG Trying to locate Go toolchain at '/usr/local/go/go1.21.0/bin/go': FAIL
2024-06-26 13:59:42,508 DEBUG Trying to locate Go toolchain at '/home/acmiel/.cache/cachi2/go/go1.21.0/bin/go': SUCCESS
```
cachi2 says it detected 1.22.0 but runs with go1.21.0

probably not related to this diff, just bringing it up here

eskultety: https://github.com/containerbuildsystem/cachi2/issues/550

pierDipi:
@eskultety do you know when we can start using Go 1.22 ?

I'm getting
```
2024-08-08 09:19:07,145 INFO go.mod reported versions: '1.22'[go], '1.22.6'[toolchain]
2024-08-08 09:19:07,145 ERROR Failed to fetch gomod dependencies
2024-08-08 09:19:07,149 ERROR PackageManagerError: Required/recommended Go toolchain version '1.22.6' is not supported yet.
Error: PackageManagerError: Required/recommended Go toolchain version '1.22.6' is not supported yet.
  Please lower your required/recommended Go version and retry the request. You may also want to open a feature request on adding support for this version.
```

and I'm assuming the feature request has been already requested somewhere

eskultety:
> @eskultety do you know when we can start using Go 1.22 in Konflux ?
> 
> I'm getting
> 
> ```
> 2024-08-08 09:19:07,145 INFO go.mod reported versions: '1.22'[go], '1.22.6'[toolchain]
> 2024-08-08 09:19:07,145 ERROR Failed to fetch gomod dependencies
> 2024-08-08 09:19:07,149 ERROR PackageManagerError: Required/recommended Go toolchain version '1.22.6' is not supported yet.
> Error: PackageManagerError: Required/recommended Go toolchain version '1.22.6' is not supported yet.
>   Please lower your required/recommended Go version and retry the request. You may also want to open a feature request on adding support for this version.
> ```
> 
> and I'm assuming the feature request has been already requested somewhere given this PR

@pierDipi do you use vendored workspaces? If so, this work is still in progress, if not, this PR is not the right place to discuss - https://github.com/containerbuildsystem/cachi2/pull/591

eskultety:
Since v2:
- complete rework
- address the fact that `go list` isn't workspace aware and run it under every workspaced module's context and aggregate the package lists
- refactor `_resolve_gomod` such that the related logic is separated into tiny helpers, improving the test isolation in return
- refactor  `_test_resolve_gomod` to rely more on mocked output rather than carefully constructed ordered list of low level `go` commands

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
eskultety: Addressed in latest revision. Should be resolved I think.

cachi2/core/package_managers/gomod.py
```diff
@@ -414,6 +428,88 @@ def _run(self, cmd: list[str], **kwargs: Any) -> str:
...
+        if not self._parsed:
+            self._parse(go, run_params)
+
+        if self._path is None or self.get("use", []) == []:
+            return []
+
+        # This re-root is going to be useful when constructing workspace ParsedModule.
+        # mypy doesn't see that self.dir is directly connected to self._path which we checked
+        go_work_dir_reroot = RootedPath(self.dir.path)  # type: ignore
+        return iter(go_work_dir_reroot.join_within_root(p["disk_path"]) for p in self["use"])
```
a-ovchinnikov, line 510: Nit: this looks like an implied generator expression is wrapped in a generator builder. I believe generator expression should suffice (i.e. you can safely drop `iter`)

cachi2/core/package_managers/gomod.py
```diff
@@ -414,6 +428,88 @@ def _run(self, cmd: list[str], **kwargs: Any) -> str:
...
+
+        go_work_json = self._get_go_work(go, run_params)
+        self.data = ParsedGoWork.model_validate_json(go_work_json).model_dump() | self.data
+        self._parsed = True
+        return self
+
+    def workspace_paths(self, go: Go, run_params: dict[str, Any] = {}) -> Iterable[RootedPath]:
+        """Get a list of paths to all workspace modules.
+
+        Note that this call must be preceded by a call to self.parse().
```
a-ovchinnikov, line 497: L501:502 make me think that this is outdated.

cachi2/core/package_managers/gomod.py
```diff
@@ -414,6 +428,88 @@ def _run(self, cmd: list[str], **kwargs: Any) -> str:
...
+        self._app_dir = app_dir
+
+        # workspaces may not be enabled -> empty instance
+        if (rooted_path := self._get_go_work_path(app_dir)) is None:
+            return
+
+        self._path = rooted_path
+
+    def __bool__(self) -> bool:
+        return self._path is not None and self.dir is not None
```
a-ovchinnikov, line 476: `dir` looks to be `None` iff `_path` is `None`, is the second half of the expression needed?

tests/unit/package_managers/test_gomod.py
```diff
@@ -188,33 +203,52 @@ def test_resolve_gomod(
...
+    ]
 
     mock_version_resolver.get_golang_version.return_value = "v0.1.0"
     mock_go_release.return_value = "go0.1.0"
     mock_get_gomod_version.return_value = ("0.1.1", "0.1.2")
 
+    parse_packages_mocked_data: list[ParsedPackage] = []
+    mock_parse_packages.return_value = parse_packages_mocked_data
+
+    if has_workspaces:
```
a-ovchinnikov, line 226: This makes an already complex test even more complex. What do you think about moving it to a separate test-case? Also judging from the amount of mocking that is necessary this should probably be better tested with an integration test.

a-ovchinnikov (commented):
Some first-pass observations

cachi2/core/package_managers/gomod.py
```diff
@@ -105,6 +97,28 @@ class ParsedPackage(_ParsedModel):
...
+
+class ParsedGoWork(_ParsedModel):
+    """Repr of the go.work file returned by 'go work edit -json' (relevant fields only).
+
+    See: go work help edit
+    """
+
+    go: Optional[str] = None
+    toolchain: Optional[str] = None
+    use: Optional[list[_GoWorkUseStruct]] = []
```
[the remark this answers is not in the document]
eskultety: I guess Optional implies None, doesn't it? I must have gotten tangled up in pydantic here.

cachi2/core/package_managers/gomod.py
eskultety: There were many occurrences of mypy complaining about `self.dir` being None. I think this tamed some of the complaints at some point during the development, so it's there on purpose. You're right it's absolutely not needed, but mypy doesn't know. Let me double check if mypy complains about something if I drop it again (might not be needed anymore, I had a ton of changes on top since I added this line).

tests/unit/package_managers/test_gomod.py
eskultety: > This makes an already complex test even more complex. What do you think about moving it to a separate test-case?

This will prolong the feature delivery even more IMO. That test is a beast. I already have another branch with more refactoring that wasn't directly connected to this work making this function rely even more on mocking and instead test the individual pieces of logic in isolation. That is not possible though yet, so my plan was:

- migrate everything this test does to invidividual isolated tests
- get rid of the `test_resolve_gomod` test to leave the rest to the integration test suite as you correctly point out

We're just not ^there yet, so if possible, I'd like to keep it this way for the time being with the promise that I'll get rid of the test eventually when I'm done with the module.

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
a-ovchinnikov: Yes, you are right, but in the case of `use` you default to `[]` and then on L116 you explicitly convert `None` to `[]`, so I'd say there is no way for `use` to ever become `None`.

cachi2/core/package_managers/gomod.py
a-ovchinnikov: It feels like every other PR also has to wrangle mypy. If it is needed then it is needed, if it stays could you please add a short mypy-bashing comment?

tests/unit/package_managers/test_gomod.py
a-ovchinnikov: Fair enough, splitting this one is an effort of its own, I won't block on it.

cachi2/core/package_managers/gomod.py
```diff
@@ -427,6 +428,86 @@ def _run(self, cmd: list[str], **kwargs: Any) -> str:
...
+        # can't do that just yet because this is being called from fetch_gomod_source which is
+        # before we set up the correct Go toolchains. We don't need toolchains to query the GOWORK
+        # env variable, but we need correct toolchain for everything else, otherwise go might
+        # complain about not meeting the required versions, so make this effectively a "lazy"
+        # evaluation driven by the caller.
+        if self._parsed or self._path is None:
+            return self
+
+        go_work_json = self._get_go_work(go, run_params)
+        self.data = ParsedGoWork.model_validate_json(go_work_json).model_dump() | self.data
```
a-ovchinnikov, line 483: This seems to be the one and only place where you explicitly access `self.data`. If I understand your intent correctly it should be an empty dictionary on initialization, and it should remain empty unless you do something like this:
```python
foo = GoWork(".")
foo["bar"] = "baz"
```
If you are not planning on using the setter interface directly then the `| self.data` part seems redundant, further you don't really need `self._parsed`: unless you expect that someone patches `self._path` on an instance or a `go.work` file itself then you can just do a `if self.data` check. Also if you don't expect users to be adding data to the object explicitly I think it is worth it to prevent this with something like following:
```python
def __setitem__(self, *a, **k):
    raise Exception("Please don't!")
```

cachi2/core/package_managers/gomod.py
```diff
@@ -1012,26 +1029,17 @@ def _resolve_gomod(
...
-        complete module list (roughly matching the list of downloaded modules).
-        """
-        cmd = [*go_list, "-deps", "-json=ImportPath,Module,Standard,Deps", pattern]
-        return map(ParsedPackage.model_validate, load_json_stream(go(cmd, run_params)))
-
     package_modules = [
-        module for pkg in go_list_deps("all") if (module := pkg.module) and not module.main
+        module
+        for pkg in _go_list_deps(go, "all", run_params)
+        if (module := pkg.module) and not module.main
```
a-ovchinnikov, line 1050: Nitpick for the sake readability. What do you think of something like:
```python
not_main = lambda pkg: (module := pkg.module) and not module.main
package_modules = [pkg.module for pkg in _go_list_deps(go, "all", run_params) if not_main(pkg)]
```
It is a multiline one-liner otherwise.

cachi2/core/package_managers/gomod.py
```diff
@@ -941,16 +941,33 @@ def _go_list_deps(
...
+        log.debug("Querying for list of packages")
+        all_packages = _go_list_deps(go, "./...", run_params)
+    else:
+        # If there are workspace modules we need to run 'list -e ./...' under every local module
+        # path because 'go list' command isn't fully properly workspace context aware
+        for wsp in go_work.workspace_paths(go, run_params):
+            log.debug(f"Querying workspace module '{wsp.path}' for list of packages")
+
+            packages = list(_go_list_deps(go, "./...", run_params | {"cwd": wsp.path}))
+            log.debug(packages)
```
a-ovchinnikov, line 962: What do you think about adding some context to this message? It will be a list of packages which could directly follow a message on L965 (or not, that will depend on `_go_list_deps` logging behavior) otherwise. Something like f"Got {packages=} for {wsp.path=}".

a-ovchinnikov (approved):
LGTM with several non-blocking nitpicks, suggestions and a couple of small questions. The change is about refactoring existing code, the bulk of it is test data update.

hack/mock-unittest-data/gomod.sh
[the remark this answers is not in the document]
eskultety: Yes, whenever we bump it, but it's buried under hack (should probably change the location or at least document it as a developer guideline) and nobody (myself included) really remembers to run it, hence the updates at the beginning of the series.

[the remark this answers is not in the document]
a-ovchinnikov: Maybe then a one-line hint in README would suffice for now? (Until https://github.com/containerbuildsystem/cachi2/issues/776 is resolved)

cachi2/core/package_managers/gomod.py
eskultety: I don't really have a preference here, so sure, if you believe it'll make the logs clearer, it's a miniscule change.

cachi2/core/package_managers/gomod.py
eskultety: > If you are not planning on using the setter interface directly then the | self.data part seems redundant

This must be some left-over from previous iterations, seems like it (I'm re-familiarizing myself with the code after the X-mas break :) ).

> further you don't really need self._parsed: unless you expect that someone patches self._path on an instance or a go.work file itself then you can just do a if self.data check

Yep, good catch, unless I touch that in the tests in which case I'll fix that.

> Also if you don't expect users to be adding data to the object explicitly I think it is worth it to prevent this with something like following:
> ```python
> def __setitem__(self, *a, **k):
>    raise Exception("Please don't!")
>```

Hold on, without trying this locally, won't this actually prevent our own method (`self._parse`) to assign to `self.data`?

cachi2/core/package_managers/gomod.py
```diff
@@ -427,6 +428,86 @@ def _run(self, cmd: list[str], **kwargs: Any) -> str:
             ) from e
 
 
+class GoWork(UserDict):
+    """Representation of Go's go.work file."""
+
+    @staticmethod
+    def _get_go_work(go: Go, run_params: dict[str, Any]) -> str:
```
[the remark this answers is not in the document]
eskultety: Sadly, yes, but only at the moment (further changes to the go backend specifically targeted at toolchains are coming). The reason here is that at the point in time when this helper is invoked (unlike `_get_go_work_path`) we need to have set up the correct toolchain elsewhere in the module already, otherwise what might happen is that we'd execute the default toolchain on the platform which may be pointing to a Go 1.20 which doesn't have the `work` subcommand and we'd get an error. The reason why we don't care about a toolchain version in `_get_go_work_path` is because we're just checking an env variable, so the version doesn't matter since `go` understands `env` command since almost forever.

cachi2/core/package_managers/gomod.py
eskultety: lambda - why didn't I think of it?! I hate multiline comprehensions. Thanks!

cachi2/core/package_managers/gomod.py
a-ovchinnikov: >  won't this actually prevent our own method (self._parse) to assign to self.data

Unlikely -- you do a direct assignment to an instance variable, magic setitem is used under the hood when someone does `foo['bar'] = 'baz'` on an instance.

I did the following test yesterday:
```python
>>> class Foo(UserDict):                                                                                                                          
...     def add(self, k, v):                                                                                                                      
...         self.data[k] = v
>>> class Foo(Foo):                                                       
...     def __setitem__(self, *a, **k):                                   
...         raise Exception("Don't!") 
>>> a = Foo()
>>> a.add(3, 4)                                                           
>>> a
{3: 4}
>>> a[4] = 5                                                              
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in __setitem__
Exception: Don't!
```
A different one, but it shows that you can even work with `self.data` internally as usual.

eskultety: Thanks!

brunoapimentel:
Wow, great (and brave) job on doing all this refactoring. The overall solution LGTM. 

I fetched the dependencies on [K8s](https://github.com/kubernetes/kubernetes/commit/8f8c94a04d00e59d286fe4387197bc62c6a4f374) using both your branch here and a slightly modified version of main (just removing the workspace vendoring restriction and using `go work vendor` instead of `go mod vendor`), and with your branch, the resulting SBOM contained 290 extra unique dependencies.

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
eskultety: The situation is also described on L475.

cachi2/core/package_managers/gomod.py
eskultety: > if you don't expect users to be adding data to the object explicitly I think it is worth it to prevent this with something like following:

So, I've played with the suggestion for a bit and we'd need to override a few more methods to emulate immutability on the dict properly:
```
clear
update
pop
popitem
setdefault
__setitem__
__delitem__
```
This led to a mypy method signatures hell when I tried to simply (and naively) assign (or override, didn't matter) with a simple helper that would just raise an exception. The dictionary is an internal impl, so it would only ever be modified by us, is the immutability a hard requirement? I mean I'm for the idea it's just the implementation doesn't seem to be worth it and we haven't really done that for other data structures we have, so is this a hard requirement?

cachi2/core/package_managers/gomod.py
eskultety: lambda would not fly -> https://www.flake8rules.com/rules/E731.html so I solved it in a different way and still making this a one-liner.

cachi2/core/package_managers/gomod.py
```diff
@@ -427,6 +428,86 @@ def _run(self, cmd: list[str], **kwargs: Any) -> str:
...
+
+    @cached_property
+    def dir(self) -> Optional[RootedPath]:
+        """Return the base directory for the go.work file."""
+        if self._path is None:
+            return None
+
+        return RootedPath(self._app_dir.root).join_within_root(self._path.subpath_from_root.parent)
+
+    def __init__(self, app_dir: RootedPath) -> None:
```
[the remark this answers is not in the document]
brunoapimentel: Not sure if there's a widely common practice, but I also personally expect __init__ to be the first method in the class (then the other magic methods following).

cachi2/core/package_managers/gomod.py
brunoapimentel: I think this log adds way too many of lines with very little actual value in it (try it with a project with many worskpaces such as [Kubernetes](https://github.com/kubernetes/kubernetes/commit/b358cf3a791bce0df23a52bb32b0e47e23746a8d)).

eskultety: The list is potentially very large, so with processing overhead in mind I tried to only pre-format it nicely when using the DEBUG level, the code isn't the prettiest though.
I also tried the snippet below which yielded slightly worse results wrt formatting. Let me know if you think it's worth trying to add a pretty format for easier log analysis.

```diff
--- a/cachi2/interface/logging.py
+++ b/cachi2/interface/logging.py
@@ -1,6 +1,7 @@
 import enum
 import logging
 from typing import Iterable
+from pprint import pformat
 
 LOG_FORMAT = "%(asctime)s %(levelname)s %(message)s"
 
@@ -15,10 +16,21 @@ class LogLevel(str, enum.Enum):
     CRITICAL = "CRITICAL"
 
 
+class PrettyFormatter(logging.Formatter):
+    """Pretty debug formatter."""
+
+    def format(self, this: logging.LogRecord) -> str:
+        """Pretty format debug logs."""
+        if this.levelno == logging.DEBUG:
+            # if isinstance(this.msg, (dict, list, tuple, set)):
+            this.msg = pformat(this.msg, width=100, indent=2, sort_dicts=False)
+        return super().format(this)
+
+
 def setup_logging(level: LogLevel, additional_modules: Iterable[str] = ()) -> None:
     """Set up logging. By default, enables only the cachi2 root logger."""
     handler = logging.StreamHandler()
-    handler.setFormatter(logging.Formatter(LOG_FORMAT))
+    handler.setFormatter(PrettyFormatter(LOG_FORMAT))
```

eskultety: Note: I purposefully made the commit a fixup so that it's easily dropped if there's general disagreement with that approach.

eskultety: > I think this log adds way too many of lines with very little actual value in it (try it with a project with many worskpaces such as [Kubernetes](https://github.com/kubernetes/kubernetes/commit/b358cf3a791bce0df23a52bb32b0e47e23746a8d)).

@brunoapimentel I have (I only test with k8s) and the list of packages is just a large blob impossible to find anything in, so I've looked for a way to make it more appealing, I'm not saying I have found a way.

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
a-ovchinnikov: Thank you!

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
a-ovchinnikov: This is a minor suggestion, I am fine with leaving the code as is. The reason it might make sense to have init() first is to tell a reader what to expect from an instance right away (readers are most likely interested in instances of classes after all). Resolving to reduce visual noise.

a-ovchinnikov (approved):

cachi2/core/package_managers/gomod.py
a-ovchinnikov: You are right, a proper type would need all these methods. My intent was more about adding an obvious visual clue for a reader to highlight that the dictionary is not supposed to be directly modified. From practical standpoint it is much more likely to run in a situation when someone just assigns to a key rather than deletes a key or pops from a dictionary and given that code's user base won't be too big I'd say blocking just setitem() would suffice. This is not a hard requirement, so resolving now.

hack/mock-unittest-data/gomod.sh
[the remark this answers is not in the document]
eskultety: Handled separately in https://github.com/containerbuildsystem/cachi2/pull/777 -> resolving this thread.

cachi2/core/package_managers/gomod.py
brunoapimentel: I does read better, but I still wonder if adding this potentially large amount of lines to the output brings us any practical benefit at all. I think this is too verbose even for our current DEBUG level. The final list of found (and deduped) packages will be in the SBOM, in any case.

brunoapimentel (approved):

cachi2/core/package_managers/gomod.py
eskultety: > The final list of found (and deduped) packages will be in the SBOM, in any case.

The question is whether reading it from an SBOM is a win for the reader though. Let's stick to the original proposal and we can assess our verbosity on debug logs further on in the future, I'll drop the latest experimental change.

tests/unit/package_managers/test_gomod.py
```diff
@@ -1106,49 +1196,61 @@ def test_package_to_component(package: Package, expected_component: Component) -
...
-    module_path = gomod_request.source_dir.join_within_root("path/to/module")
-    version_resolver = mock.Mock()
-
-    mock_config.return_value.gomod_download_max_tries = 1
-    mock_go_release.return_value = "go0.1.0"
-    mock_get_gomod_version.return_value = ("0.1.1", "0.1.2")
-    mock_disable_telemetry.return_value = None
+@pytest.mark.parametrize("pattern", ["./...", "all"])
+@mock.patch("cachi2.core.package_managers.gomod.run_cmd")
+def test__go_list_deps(mock_run_cmd: mock.Mock, pattern: Literal["all", "./..."]) -> None:
```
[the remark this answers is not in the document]
eskultety: Fixed.

cachi2/core/package_managers/gomod.py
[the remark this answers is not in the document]
eskultety: Fixed.

tests/unit/package_managers/test_gomod.py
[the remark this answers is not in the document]
brunoapimentel: Damn, you're right, we do have some bundler and some yarn_classic tests that follow this pattern. I don't mind either way, I was just thinking we should be consistent.

cachi2/core/package_managers/gomod.py
brunoapimentel: For example, I just processed K8s with your latest patches, and wrote the logs to `output.txt`:
```
$ cat output.txt| wc -l
22815
$ cat output.txt | grep ParsedPackage | wc -l
22660
```

I think we'd be better of without this, specially when you consider that Cachi2 can be used as part of a CI system, and having very large log files can make things more cumbersome.

But I won't block the PR on this, you have my ack.

eskultety: I'll add this to my notes on topics to discuss offline.

PR #555, opened 2024-06-19
stop exposing git credentials in SBOM

### LINKED ISSUE #521: Don't expose credentials from git origin url
Putting git credentials directly in the git remote `origin` url is bad way to authenticate to git repos. Nevertheless, if someone does use that approach, cachi2 shouldn't expose their credentials in the SBOM.

```bash
# in the root of the cachi2 repo
$ git remote set-url origin https://chmeliik:foo@github.com/containerbuildsystem/cachi2
$ cachi2 fetch-deps pip
$ jq < cachi2-output/bom.json '.components[] | select(.purl | test("chmeliik"))'
{
  "name": "cachi2",
  "purl": "pkg:pypi/cachi2?vcs_url=git%2Bhttps://chmeliik:foo%40github.com/containerbuildsystem/cachi2%40fdaebb589f60d2427ecd65ef301e011705aaf19d",
  "properties": [
    {
      "name": "cachi2:found_by",
      "value": "cachi2"
    }
  ],
  "type": "library"
}
```

Drop the credentials from the vcs_url qualifier to make sure they won't appear in the SBOM https://github.com/containerbuildsystem/cachi2/blob/e6fb059d2218483b1bc2e8327cb7336bc9db7b1e/cachi2/core/scm.py#L29-L34

Or perhaps drop them before ever creating a `RepoID` in the first place, just to be sure.

slimreaper35:
if git origin url contains credentials for
auth, remove them from url so they will not
appear in PURL in SBOM

closes [#521](https://github.com/containerbuildsystem/cachi2/issues/521)

### DIFF
--- a/cachi2/core/scm.py
+++ b/cachi2/core/scm.py
@@ -3,10 +3,10 @@ import logging
 import re
 import tarfile
 import tempfile
-import urllib.parse
 from os import PathLike
 from pathlib import Path
 from typing import NamedTuple, Union
+from urllib.parse import ParseResult, SplitResult, urlparse, urlsplit
 
 from git.repo import Repo
 
@@ -22,9 +22,9 @@ class RepoID(NamedTuple):
     commit_id: str
 
     @property
-    def parsed_origin_url(self) -> urllib.parse.SplitResult:
+    def parsed_origin_url(self) -> SplitResult:
         """Get the url as a urllib.parse.SplitResult."""
-        return urllib.parse.urlsplit(self.origin_url)
+        return urlsplit(self.origin_url)
 
     def as_vcs_url_qualifier(self) -> str:
         """Turn this RepoID into a 'vcs_url' qualifier as defined by the purl spec.
@@ -62,7 +62,12 @@ def get_repo_id(repo: Union[str, PathLike[str], Repo]) -> RepoID:
 
 def _canonicalize_origin_url(url: str) -> str:
     if "://" in url:
-        return url
+        parsed: ParseResult = urlparse(url)
+        cleaned_netloc = parsed.netloc.replace(
+            f"{parsed.username}:{parsed.password}@",
+            "",
+        )
+        return parsed._replace(netloc=cleaned_netloc).geturl()
     # scp-style is "only recognized if there are no slashes before the first colon"
     elif re.match("^[^/]*:", url):
         parts = url.split("@", 1)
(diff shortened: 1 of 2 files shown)


MartinBasti (approved):

tests/unit/test_scm.py
```diff
@@ -24,6 +24,10 @@ class TestRepoID:
             # no-op
             ("ssh://user@git.host.com/some/path", "ssh://user@git.host.com/some/path"),
             ("https://git.host.com/some/path", "https://git.host.com/some/path"),
+            (
```
brunoapimentel, line 27: Just a tiny nitpick: the new test scenario is under a comment saying `#no-op`, but in this case, we're actually changing the provided url. So I'd add a new comment to separate it into a new category.

brunoapimentel (approved):

cachi2/core/scm.py
```diff
@@ -62,7 +62,21 @@ def get_repo_id(repo: Union[str, PathLike[str], Repo]) -> RepoID:
...
+        return urlunparse(
+            (
+                parsed_url.scheme,
+                cleaned_netloc,
+                parsed_url.path,
+                parsed_url.params,
+                parsed_url.query,
+                parsed_url.fragment,
+            )
+        )
```
eskultety, line 79: Can we simplify this to something like the following?

```
return urlparse(url)._replace(netloc="").geturl()
```
(you can even split this to multiple lines if you think ^this lacks readability)
`urlparse` returns a [named tuple](https://docs.python.org/3.12/library/urllib.parse.html#urllib.parse.urlparse) and named tuples can be modified via the [`_replace`](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._replace) method. Seems much cleaner to me.

eskultety: Actually, there are multiple parts in netloc, so more like this
```
parsed = urlparse(url)
return parsed._replace(netloc=parsed.netloc.split("@")[1]).geturl()
```

slimreaper35: A few were failing with this approach...
How about current solution ?
```
>           assert repo_id.origin_url == expect_result
E           AssertionError: assert 'ssh://git.host.com/some/path' == 'ssh://user@git.host.com/some/path'
E
E             - ssh://user@git.host.com/some/path
E             ?       -----
E             + ssh://git.host.com/some/path

tests/unit/test_scm.py:51: AssertionError
```

eskultety: Sorry, missed that detail, your adjustment certainly works.

eskultety (approved):

PR #565, opened 2024-07-02
Bundler design document

slimreaper35:

### DIFF
--- a/docs/design/bundler.md
+++ b/docs/design/bundler.md
@@ -1,0 +1,516 @@
+# Design document for Bundler package manager
+
+Contents:
+
+1. [Ruby ecosystem overview](#i-ruby-ecosystem-overview)
+2. [Design for the Cachi2 implementation](#ii-design-for-the-implementation-in-cachi2)
+
+## I. Ruby ecosystem overview
+
+### Development prerequisites
+In order to execute the commands in the examples below, make sure you have the following packages installed in your
+environment:
+
+```bash
+sudo dnf install rubygems rubygems-bundler
+```
+
+Or use the official Ruby image from Docker hub:
+```bash
+podman run --rm -it docker.io/library/ruby:3.3.3 bash
+```
+
+### Project structure
+```bash
+bundle init # creates Gemfile in the current directory
+bundle lock # creates Gemfile.lock in the current directory
+```
+
+```bash
+├── .bundle
+│   └── config
+├── Gemfile
+├── Gemfile.lock
+├── vendor/cache
+```
+
+### Glossary
+- **Gemfile**: A file that specifies the gems that your project depends on and their versions. Bundler uses this file
+to install the correct versions of gems for your project.
+
+  ```ruby
+  source "https://rubygems.org"
+
+  gem "rails", "= 6.1.7"
+  ```
+
+- **Gemfile.lock**: A file that locks the versions of gems that are installed for your project. Bundler uses this file
+to ensure that the correct versions of gems are installed consistently across different environments. The native
+parsing of this file is done by the
+[LockfileParser class](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb).
+
+- **RubyGems**: General package manager for Ruby. Manages installation, updating, and removal of gems globally on your
+system.
+
+  ```bash
+  gem --help
+  ```
+
+- **Bundler**: Dependency management tool for Ruby projects.
+Ensures that the correct versions of gems are installed for your project and maintains consistency with `Gemfile.lock`.
+
+  ```bash
+  bundler --help
+  ```
+
+- **Gem**: A package that can be installed and managed by Rubygems. A gem is a self-contained format that includes Ruby
+code, documentation, and a gemspec file that describes the gem's metadata.
+
+- **{gem}.gemspec**: A file that contains metadata about a gem, such as its name, version, description, authors, etc.
+RubyGems uses it to install, update, and uninstall gems.
+
+  ```ruby
+  Gem::Specification.new do |spec|
+    spec.name        = "example"
+    spec.version     = "0.1.0"
+    spec.authors     = ["Nobody"]
+    spec.email       = ["ruby@example.com"]
+    spec.summary     = "Write a short summary, because RubyGems requires one."
+  end
+  ```
+
+### Dependency types
+There are four types of
+[sources](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb#L48) for dependencies
+in the `Gemfile.lock` file:
+
+#### Gem dependencies
+Regular gem dependencies are located at the source URL, in our case, always <https://rubygems.org>. Each gem can be
+accessed by its name and version - rubygems.org/gems/`<name>`-`<version>`.gem
+
+Example of a gem dependency in the `Gemfile.lock` file:
+
+```bash
+GEM
+ remote: https://rubygems.org/
+ specs:
+  rails (6.1.4)
+    # transitive dependencies
+    actioncable (= 6.1.4)
+    actionmailbox (= 6.1.4)
+    actionmailer (= 6.1.4)
+    actionpack (= 6.1.4)
+    actiontext (= 6.1.4)
+    actionview (= 6.1.4)
+    activejob (= 6.1.4)
+    activemodel (= 6.1.4)
+    activerecord (= 6.1.4)
+    activestorage (= 6.1.4)
+    activesupport (= 6.1.4)
+    bundler (>= 1.15.0)
+    railties (= 6.1.4)
+    sprockets-rails (>= 2.0.0)
+```
+
+#### Git dependencies
+Example of a Git dependency in the `Gemfile.lock` file:
+
+```
+GIT
+  remote: https://github.com/porta.git
+  revision: 779beabd653afcd03c4468e0a69dc043f3bbb748
+  branch: main
+  specs:
+    porta (2.14.1)
+```
+
+#### Path dependencies
+Example of a path dependency in the `Gemfile.lock` file:
+
+```
+PATH
+  remote: some/pathgem
+  specs:
+    pathgem (0.1.0)
+```
+
+All path dependencies must be in the project directory. Bundler
+[does not copy](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/source/path.rb#L83) those
+dependencies that are already within the root directory of the project.
+
+#### Plugins
+Installing a plugin, even when on a folder that is a Bundler project, doesn't seem to affect the `Gemfile.lock`. The
+plugin seems to be installed by default in the `$PWD/.bundle/`. The `Gemfile.lock` does have a section for plugins,
+though, so further investigation would be needed. This initial investigation was done with the plugins listed under
+[Known Plugins](https://bundler.io/guides/plugins.html).
+
+*Don't confuse Bundler plugins with [RubyGems plugins](https://guides.rubygems.org/plugins/). The latter are meant to
+extend the functionality of `gem` itself, and don't seem to have any impact on Bundler directly.*
+
+
+### Platforms
+Some gems may contain pre-compiled binaries that provide native extensions to the Ruby package. Any gem declared in the
+`Gemfile` can be limited to specific
+[platforms](https://bundler.io/v2.5/man/gemfile.5.html#PLATFORMS), making Bundler ignore it in case the project is
+being built on a non-matching platform:
+
+```ruby
+gem "nokogiri",   platforms: [:windows_31, :jruby]
+```
+
+Here's an example of how a the `PLATFORM` section looks like in the `Gemfile.lock`:
+
+```
+PLATFORMS
+  arm64-darwin-20
+  arm64-darwin-21
+  arm64-darwin-22
+  ruby
+  x86_64-darwin-18
+  x86_64-darwin-20
+  x86_64-darwin-21
+  x86_64-darwin-22
+  x86_64-linux
+```
+
+In case a user wants to force all the binaries to be compiled from source, the `BUNDLE_FORCE_RUBY_PLATFORM` environment
+variable can be used.
+
+### Dev dependencies
+When adding a Gem into a Gemfile, the user might opt to nest them under a specific
+[group](https://bundler.io/guides/groups.html). The name of the group can be any string, but the usual groups tend to
+be common labels such as `:test`, `:development` or `:production`.
+
+Here's how it looks like in a `Gemfile`:
+
+```ruby
+# :default group
+gem 'nokogiri'
+
+group :test do
+  gem 'faker'
+  gem 'rspec'
+end
+```
+
+Another way to declare a dependency in the `:development` group is to
+[add it](https://guides.rubygems.org/specification-reference/#add_development_dependency) to the `Gem::Specification`,
+which is usually declared in the `.gemspec` file. This means we can safely assume that all dependencies under
+`:development` are dev dependencies.
+
+### Dependency checksums
+The support for checksums in the `Gemfile.lock` is still in development, and currently is an
+[opt-in feature](https://github.com/rubygems/rubygems/pull/7217). To enable it, we need to manually add a `CHECKSUMS`
+section in the `Gemfile.lock`:
+
+```shell
+# manually add `CHECKSUMS` section somewhere in the Gemfile.lock
+vim Gemfile.lock
+# install any gem
+bundle add rails --version "6.1.7"
+# check the Gemfile.lock /o\
+cat Gemfile.lock
+```
+
+Example of a checksum section in the `Gemfile.lock`:
+
+```
+CHECKSUMS
+  actioncable (6.1.7) sha256=ee5345e1ac0a9ec24af8d21d46d6e8d85dd76b28b14ab60929c2da3e7d5bfe64
+  actionmailbox (6.1.7) sha256=c4364381e724b39eee3381e6eb3fdc80f121ac9a53dea3fd9ef687a9040b8a08
+  actionmailer (6.1.7) sha256=5561c298a13e6d43eb71098be366f59be51470358e6e6e49ebaaf43502906fa4
+  actionpack (6.1.7) sha256=3a8580e3721757371328906f953b332d5c95bd56a1e4f344b3fee5d55dc1cf37
+  actiontext (6.1.7) sha256=c5d3af4168619923d0ff661207215face3e03f7a04c083b5d347f190f639798e
+  actionview (6.1.7) sha256=c166e890d2933ffbb6eb2a2eac1b54f03890e33b8b7269503af848db88afc8d4
+```
+
+This feature is available since Bundler [v2.5.0](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb#L55),
+from this [PR](https://github.com/rubygems/rubygems/pull/6374) being merged on Oct 21, 2023.
+
+## II. Design for the implementation in Cachi2
+
+### Prefetching
+
+Running a bundler command to fetch the dependencies always executes the `Gemfile`, which is arbitrary Ruby code.
+Executing arbitrary code is a security risk and makes it impossible to assert that the resulting SBOM is accurate
+(since any random package can be fetched from the Internet during the prefetch). This means that we need to implement
+custom code to fetch the dependencies.
+
+In the `Gemfile.lock`, all Gems that come from the same remote URL are grouped under the same block:
+```
+GEM
+ remote: https://rubygems.org/
+ specs:
+  rails (6.1.4)
+  json-schema (1.2.1)
+```
+
+A Gem can be fetched from its original location by using the following template:
+```ruby
+"https://#{remote}/gems/#{name}-#{version}.gem"
+```
+
+We should also leverage the existing code used to perform parallel downloads based on `asyncio` to download the necessary
+Gems from the internet.
+
+#### Output folder structure
+
+Bundler has a built-in feature to cache all dependencies locally. This is done with the `bundle cache --all` command or
+`bundle package --all` alias. In order to make bundler use the prefetched dependencies during the build, Cachi2 needs
+to recreate the exact same folder structure as bundler does.
+
+Here's an example of how the output folder should look like:
+
+```bash
+$ ls vendor/cache
+
+actioncable-6.1.7.gem
+date-3.3.4.gem
+json-schema-26487618a684
+nokogiri-1.16.6.gem
+tzinfo-2.0.6.gem
+```
+
+Notice that all the `.gem` dependencies are kept in their original format, and Git dependencies are just plain clones
+of the repository placed in a folder. For Git dependencies, the folder name must match this specific
+[format](https://github.com/rubygems/rubygems/blob/3da9b1dda0824d1d770780352bb1d3f287cb2df5/bundler/lib/bundler/source/git.rb#L130):
+
+```ruby
+"#{base_name}-#{shortref_for_path(revision)}"
+```
+
+The name of the directory **must come from the Git URL**, not the actual name of the gem, and the cloned folder must
+contain unpacked source code. Any other format will cause bundler to try to re-download the repository, causing the
+build to fail.
+
+##### Multiple Gems in a single repository
+
+A single repository can hold multiple Gems, and those can be imported as dependencies. When this happens, Bundler still
+expects a single clone to be made. Here's an example of how multiple gems imported from a single repository+revision
+looks like in the `Gemfile.lock`:
+
+```
+GIT
+  remote: https://github.com/chatwoot/azure-storage-ruby
+  revision: 9957cf899d33a285b5dfe15bdb875292398e392b
+  branch: chatwoot
+  specs:
+    azure-storage-blob (2.0.3)
+      azure-storage-common (~> 2.0)
+      nokogiri (~> 1, >= 1.10.8)
+    azure-storage-common (2.0.4)
+      faraday (~> 2.0)
+      faraday-follow_redirects (~> 0.3.0)
+      faraday-net_http_persistent (~> 2.0)
+      net-http-persistent (~> 4.0)
+      nokogiri (~> 1, >= 1.10.8)
+```
+
+### Out of scope
+
+#### Plugins
+Bundler has support for using [plugins](https://bundler.io/guides/bundler_plugins.html), which allows users to extend
+Bundler's functionality in any way that they seem fit. Since this can open the possibility for security issues, plugins
+will not be supported by Cachi2.
+
+Since we're not proposing the direct usage of Bundler to fetch the dependencies, no other actions are needed in the
+prefetch phase, existing plugin definitions will be ignored.
+
+#### Pre-compiled binaries
+For the initial implementation, we're aiming to provide support only for plain Ruby gems (which are idenfied as `ruby`
+in the `PLATFORMS` section of the `Gemfile.lock`). Platforms that relate to specific architectures will contain
+binaries that were pre-compiled for that architecture (see [Platforms](#platforms)).
+
+The URL schema in the default rubygems registry seems to follow this format:
+
+```ruby
+# Plain Ruby Gem
+"https://rubygems.org/gems/#{name}-#{version}.gem"
+
+# Platform-specific Gem
+"https://rubygems.org/gems/#{name}-#{version}-#{platform}.gem"
+```
+
+This means that we can easily just download the plain Ruby Gems, and skip all platform-specific Gems altogether. From
+the testing I did, this did not seem to affect the ability to perform a hermetic build in any way, Bundler just
+compiled everything from source.
+
+To avoid confusing users if a corner case shows up, we can add a `WARNING` log in case extra platforms are detected,
+and also document it properly.
+
+Proper support for pre-compiled binaries should be probably left as a follow-up feature, similarly to what was done
+with [pip wheels](https://github.com/containerbuildsystem/cachi2/blob/main/docs/pip.md#distribution-formats).
+
+#### Checksum verification
+Since checksums in the `Gemfile.lock` is still a feature in development (see [checksums](#dependency-checksums)), we
+can postpone implementing support for it until the feature is delivered.
+
+We need to decide if we will report all dependencies as having missing checksums in the SBOM, or not.
+
+#### Dev dependencies
+Bundler declares all dev dependencies under the `:development`
+[group](#dependency-groups-or-how-bundler-deals-with-dev-dependencies). Unfortunately, groups declared in the `Gemfile`
+are not reflected in the `Gemfile.lock`.
+
+To implement proper reporting of dev dependencies, we'll very likely need to also parse the `Gemfile`. It can be done
+as a follow-up if the need arises.
+
+#### Prefetching Bundler
+When running `bundle install`, Bundler will always try to fetch the exact version that is pinned in the `Gemfile.lock`
+to perform the install. When doing an offline install from cache, a warning message is instead printed, but Bundler
+will usually perform the install as expected.
+
+To allow users to use the pinned version instead of only relying on the Bundler version present in the base image,
+Cachi2 could also prefetch the specific Bundler version needed for that project. This is easy to achieve, since Bundler
+is treated as an ordinary Gem: https://rubygems.org/gems/bundler.
+
+This feature, however, is out of scope for the initial implementation, and could be added if there's user demand for
+it.
+
+### Providing the content for the hermetic build
+
+#### Setting the Bundler configuration
+
+The order of precedence for Bundler configuration options is as follows:
+
+1. Local config (`<project_root>/.bundle/config or $BUNDLE_APP_CONFIG/config`)
+2. Environment variables (ENV)
+3. Global config (`~/.bundle/config`)
+4. Bundler default config
+
+Since the local configuration takes higher precedence than the environment variables (except `BUNDLE_APP_CONFIG`), we
+need to set the Bundler configuration options to make the build work.
+
+In order to do this, we can either use `inject-files` to overwrite the `.bundle/config` directory in the source folder,
+or use `BUNDLE_APP_CONFIG` to point Bundler to a config directory within the Cachi2 output directory. The latter has
+the benefit of not needing to dirty the cloned sources, but it wouldn't be able to support a multiple Ruby project per
+repository scenario (since we would need to keep multiple configuration files).
+
+#### Relevant configuration for the build
+
+```
+BUNDLE_CACHE_PATH=${output_dir}/deps/rubygems
+BUNDLE_DEPLOYMENT=true
+BUNDLE_NO_PRUNE=true
+```
+
+- **BUNDLE_CACHE_PATH**: The directory that Bundler will place cached gems in when running bundle package, and that
+Bundler will look in when installing gems. Defaults to `vendor/cache`.
+
+- **BUNDLE_DEPLOYMENT**: Disallow changes to the Gemfile. This also has the side effect of forcing Bundler to use the
+local cache instead of trying to reach out for the Internet. This allows the hermetic build to work without forcing the
+users to add the `--local` flag to the `bundler install` command.
+
+- **BUNDLE_NO_PRUNE**: Whether Bundler should leave outdated gems unpruned when caching. Since we're potentially using
+a single cache folder for multiple Gems ("input packages" in Cachi2's terms), we need to make sure that the first
+install won't prune any cached dependencies that are unrelated to it.
+
+For more information, see Bundler's [documentation](https://bundler.io/v2.5/man/bundle-config.1.html).
+
+##### Other configuration that was considered
+
+- `BUNDLE_ALLOW_OFFLINE_INSTALL` is not working either with `bundle install` for some reason, which could be probably
+the most logical solution in this case.
+
+### Generating the SBOM
+
+#### Main package metadata
+
+Ruby uses [Gem::Specification](https://guides.rubygems.org/specification-reference/) as a means of defining a Gem's
+metadata, and it is usually defined in a `{gem-name}.gemspec` file. This file is not mandatory, though, and when it
+exists, it needs to be explicitly imported in the `Gemfile`:
+
+```
+source "https://rubygems.org"
+
+gemspec
+```
+
+When the `.gemspec` file exists and is properly imported, it will be listed in the `Gemfile.lock` as a `PATH`
+dependency:
+
+```
+PATH
+  remote: .
+  specs:
+    tmp (0.1.2)
+```
+
+Since the `remote` will always point to `.` in case of the main package [^main-package], we can safely use it to get
+the `name` and `version` for the SBOM component. In case this block is absent, we will need to fallback to the
+repository's remote `origin` to retrieve the main package's name, and leave the version empty, since it is not a
+mandatory field.
+
+[^main-package]: In Cachi2's terms, the **main package** is the path in the repository that is currently being
+  processed.
+
+#### PURLs
+
+Also check the Ruby PURL [specification](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#gem).
+
+##### Standard Gem
+```txt
+pkg:gem/my-gem-name@0.1.1
+```
+
+##### Git dependency
+
+```txt
+pkg:gem/my-git-dependency?vcs_url=git%2Bhttps://github.com/my-org/mygem.git%26487618a68443e94d623bb585cb464b07d36702
+```
+
+The metadata for a Git dependency can be read from the `Gemfile.lock`:
+
+```
+GIT
+  remote: https://github.com/my-org/mygem.git
+  revision: 26487618a68443e94d623bb585cb464b07d36702
+  specs:
+    json-schema (3.0.0)
+      addressable (>= 2.4)
+```
+
+##### Path dependency
+
+```txt
+pkg:gem/my-path-dependency?vcs_url=git%2Bhttps://github.com/my-org/mygem.git%40b6f47bd07e669c8d2eced8015c4bfb06db49949#subpath
+```
+Source code for "official" Bundler lockfile parsing in Ruby:
+<https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb>
+```ruby
+PATH
+  remote: subpath
+  specs:
+    my-path-dependency (1.0.0)
+```
+
+### Summary
+
+#### Implementation scoping
+- design high-level code structure into multiple modules
+- create a test repository that contains all the relevant use cases
+- define models for Bundler as a new package manager
+- parse all gems from `Gemfile.lock`
+- implement metadata parsing for the "main package"
+- download all gems from rubygems.org
+- download all gems from Git repositories
+- validate path dependencies are relative to the project root
+- inject the Bundler configuration needed for the offline install
+- generate PURLs for all dependencies
+- add integration and e2e tests
+- add documentation
+
+#### Out of scope
+- implement checksum parsing and validation when prefetching from the registry
+- downloading the Bundler version specified in the `Gemfile.lock`
+- support for pre-compiled binaries (platforms other than `ruby`)
+- Gemfile.lock checksum validation (blocked by pending official support)
+- reporting dev dependencies
+- proper support for plugins
+
+### References
+This design doc was partially based on the original implementation done in [Cachito](https://github.com/containerbuildsystem/cachito/blob/master/cachito/workers/pkg_managers/rubygems.py). Since Cachi2 has different design goals from Cachito, the implementation here will deviate from the original one, with a key difference being that Cachi2 needs to provide a way to perform an [offline install](#providing-the-content-for-the-hermetic-build) from the local prefetched content.
+
+Cachito parses the `Gemfile.lock` via [gemlock-parser](https://github.com/containerbuildsystem/gemlock-parser),
+which is vendored from
+[scancode-toolkit](https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/gemfile_lock.py), which can also be leveraged by Cachi2.


cachi2/core/package_managers/rubygems.py
```diff
@@ -0,0 +1,336 @@
+import logging
```
eskultety, line 1: Please don't put everything into the same module - those are the old ways we did implement package managers leading to a single big messy module. We should split the logic into a few self-contained modules wrapped in a package.

cachi2/core/models/input.py
```diff
@@ -179,8 +179,14 @@ class YarnPackageInput(_PackageInputBase):
     type: Literal["yarn"]
 
 
+class RubygemsPackageInput(_PackageInputBase):
```
eskultety, line 182: @slimreaper35 do you plan on proposing a design doc for this work?

docs/design/rubygems.md
```diff
@@ -0,0 +1,160 @@
...
+
+gem 'rails', '6.1.4'
+
+system('echo "Hello, world!"')
+system('sudo rm -rf /')
+```
+
+### Missing features
+
+Cachito does not cache bundler because it is not pinned as a dependency with version in the `Gemfile.lock` (even if it is pinned in the `Gemfile`).
```
eskultety, line 106: mind the 'cachito' references.

cachi2/core/models/input.py
eskultety: I'm blind, I only saw `docs` and then `rubygems.md`.

docs/design/rubygems.md
slimreaper35: removed all _cachito_  mentions

docs/design/rubygems.md
```diff
@@ -0,0 +1,284 @@
...
+
+Bundler uses this format to cache git repositories:
+
+[https://github.com/rubygems/.../bundler/source/git.rb#L139](https://github.com/rubygems/rubygems/blob/3da9b1dda0824d1d770780352bb1d3f287cb2df5/bundler/lib/bundler/source/git.rb#L139>)
+
+Renaming this directory, or changing the commit hash, will cause bundler to re-download the repository -> cache invalidation -> the build will fail.
+
+### Path dependencies
+
+Recursive copy of the gem directory to the `vendor/cache` directory - but it does not work, if the path is within the project root.
```
chmeliik, line 165: "Does not work" in what sense? Bundler doesn't copy it to `vendor/cache` because it considers it unnecessary? The PATH dependency can be installed just fine even if it's not in the cache, right?

(this would also make more sense in the Vendoring Solution section)

docs/design/rubygems.md
```diff
@@ -0,0 +1,284 @@
...
+...
+```
+
+Bundler uses this format to cache git repositories:
+
+[https://github.com/rubygems/.../bundler/source/git.rb#L139](https://github.com/rubygems/rubygems/blob/3da9b1dda0824d1d770780352bb1d3f287cb2df5/bundler/lib/bundler/source/git.rb#L139>)
+
+Renaming this directory, or changing the commit hash, will cause bundler to re-download the repository -> cache invalidation -> the build will fail.
+
+### Path dependencies
```
chmeliik, line 163: Could you add an example of a PATH dependency like you did for GEM and GIT?

chmeliik (commented):
First round of review (made it to the Checksums section) :slightly_smiling_face: 

My general observation is that it's kind of hard to derive implementation guidance from the doc because it's not always clear which parts are just general commentary on RubyGems/Bundler and which parts describe the implementation decisions

docs/design/rubygems.md
```diff
@@ -0,0 +1,284 @@
...
+  specs:
+    porta (2.14.1)
+...
+```
+
+Bundler uses this format to cache git repositories:
+
+[https://github.com/rubygems/.../bundler/source/git.rb#L139](https://github.com/rubygems/rubygems/blob/3da9b1dda0824d1d770780352bb1d3f287cb2df5/bundler/lib/bundler/source/git.rb#L139>)
+
+Renaming this directory, or changing the commit hash, will cause bundler to re-download the repository -> cache invalidation -> the build will fail.
```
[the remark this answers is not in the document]
slimreaper35: yes, exactly
I will try to rearrange the document later on

docs/design/rubygems.md
```diff
@@ -0,0 +1,357 @@
...
+PATH
+  remote: ../foobar
+  specs:
+    foobar (0.1.0)
+...
+```
+
+### Plugins
+
+TODO
```
chmeliik, line 207: IMO we just need to call out that `Plugin`-type dependencies are not supported. 

I _would_ be curious to know what they are though :grin:

docs/design/rubygems.md
```diff
@@ -0,0 +1,357 @@
...
+This will create another directory with the following structure:
+
+```bash
+├── {name}.gemspec
+├── Gemfile
+├── lib
+│   ├── {name}.rb
+├── Rakefile
+...
+```
```
chmeliik, line 34: This is a bit confusing, are we talking about two different directories here? It's still just one directory right?

IMO you can just merge `{name}.gemspec` into the first file tree, I think that's all the relevant files. If you want, you can mention that one would initialize the directory with `bundle gem {name}` and the rest of the files get created with other bundler commands

docs/design/rubygems.md
```diff
@@ -0,0 +1,284 @@
...
+
+gem 'rails', '= 6.1.7'
+
+system('echo "Hello, world!"')
+system('sudo rm -rf /')
+```
+
+### Missing features
+
+Bundler is not pinned as a dependency with version in the `Gemfile.lock` (even if it is pinned in the `Gemfile`).
```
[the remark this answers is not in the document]
slimreaper35: I installed an old version of gemlock-parser :face_in_clouds:

cachi2/core/package_managers/rubygems.py
slimreaper35: The file will be deleted anyway, so I did not bother with splitting it.

docs/design/rubygems.md
```diff
@@ -0,0 +1,344 @@
...
+  actiontext (6.1.7) sha256=c5d3af4168619923d0ff661207215face3e03f7a04c083b5d347f190f639798e
+  actionview (6.1.7) sha256=c166e890d2933ffbb6eb2a2eac1b54f03890e33b8b7269503af848db88afc8d4
+  ...
+
+BUNDLED WITH
+   2.5.14
+```
+
+I believe this feature is available since bundler [v2.5.0](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb#L55)
+from this [PR](https://github.com/rubygems/rubygems/pull/6374) being merged on Oct 21, 2023.
```
chmeliik, line 257: Hmm, so what does the checksums situation mean for us? I think gemlock-parser doesn't support them anyway, so we would first need to add support there (first to the project we vendored, then re-vendor the changes).

If we implement the parsing, we can verify the checksums after downloading (if present) or report the `cachi2:missing_hash:in_file` property (if not present). Otherwise, we probably shouldn't report anything because we just don't know. IMO implementing the parsing and validation/reporting can be a follow-up to the initial impl.

docs/design/rubygems.md
```diff
@@ -0,0 +1,344 @@
...
+those dependencies that are already within the root directory of the project.
+
+#### Plugins
+
+Not supported by cachi2.
+
+### Checksums
+
+Checksum validation is enabled by default.
+It can be disabled with the `BUNDLE_DISABLE_CHECKSUM_VALIDATION` environment variable.
```
chmeliik, line 222: So if I understand it correctly, bundler does do checksum validation even if there are no CHECKSUMS in the lockfile, but it just validates that the downloaded archive matches the checksum reported by the server (similar to how PyPI reports the expected checksums for archives and we use that to filter out non-matching wheels)

I realize now that we lose even _that_ small validation in Cachito, where we just don't validate any checksums at all. Which is not ideal, but we don't have to fix it right away in cachi2. A follow-up story to implement this checksum validation would be nice though.

docs/design/rubygems.md
```diff
@@ -0,0 +1,344 @@
...
+
+I believe this feature is available since bundler [v2.5.0](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb#L55)
+from this [PR](https://github.com/rubygems/rubygems/pull/6374) being merged on Oct 21, 2023.
+
+### Environment variables
+
+The order of precedence for bundler configuration options is as follows:
+
+1. Local config (`<project_root>/.bundle/config or $BUNDLE_APP_CONFIG/config`)
+2. Environmental variables (ENV)
```
chmeliik, line 264: Damn, local config takes precedence over ENV vars? That means if we just set the env vars, we may fail to configure the build for offline installation (if the user has some conflicting settings in their config). We might want overwrite their config instead (via `inject-files`) or parse their config and raise an error if we detect conflicting settings

docs/design/rubygems.md
```diff
@@ -0,0 +1,344 @@
...
+3. Global config (`~/.bundle/config`)
+4. Bundler default config
+
+#### Relevant environment variables
+
+```txt
+BUNDLE_FORCE_RUBY_PLATFORM=true
+BUNDLE_CACHE_ALL=true
+BUNDLE_CACHE_PATH=${output_dir}/deps/rubygems
+BUNDLE_DISABLE_CHECKSUM_VALIDATION=false
```
chmeliik, line 274: Why is `BUNDLE_DISABLE_CHECKSUM_VALIDATION=false` relevant?

docs/design/rubygems.md
```diff
@@ -0,0 +1,344 @@
...
+```
+
+**BUNDLE_CACHE_ALL**: Cache all gems, including path and git gems.
+This needs to be explicitly configured on bundler 1 and bundler 2, but will be the default on bundler 3.
+
+**BUNDLE_CACHE_PATH**: The directory that bundler will place cached gems in when running bundle package,
+and that bundler will look in when installing gems. Defaults to `vendor/cache`.
+
+**BUNDLE_FORCE_RUBY_PLATFORM**: Ignore the current machine's platform and install only ruby platform gems.
+As a result, gems with native extensions will be compiled from source.
```
chmeliik, line 284: Speaking of platforms, that would be worth describing in more detail. Cachito only supports the pure Ruby platform, I think it's mainly because gems for other platforms contain precompiled binaries and we didn't want to allow those

docs/design/rubygems.md
```diff
@@ -0,0 +1,344 @@
...
+
+### Metadata
+
+### git repository URL
+
+- git repository URL is used in other package managers as well
+- no version information available
+- gems in the repository are basically path dependencies in the `Gemfile.lock` ?!
+
+### `{gem}.gemspec` file
```
chmeliik, line 298: Cachito just takes the package name from the git url and uses the commit hash as the version. In cachi2, I'd say we can do the same as the fallback case for pip - name from git url, no version.

As an improvement, we can add some very simple regex-based parsing of the `name` and `version` fields from `{gem}.gemspec`

docs/design/rubygems.md
```diff
@@ -0,0 +1,344 @@
...
+Gemfile must contain a _gemspec_ line + the `{gem}.gemspec` file must be present in the repository.
+Bundle will add the gem as path dependency to the `Gemfile.lock` file.
+
+```Gemfile.lock
+...
+PATH
+  remote: .
+  specs:
+    tmp (0.1.2)
+...
```
chmeliik, line 312: I'm not sure what you mean by this part. Are you saying that metadata about the "main" gem can be found in Gemfile.lock?

docs/design/rubygems.md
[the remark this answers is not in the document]
chmeliik: Nice, that is quite a bit clearer. Some sections are still missing guidance on what we want to do, but we can resolve that in those sections

docs/design/rubygems.md
slimreaper35: Yeah, if the "remote" is `.` it means that there is a gemspec file in the root directory.

I found something similar in the gemlock-parser. Not sure what exactly was meant by [primary gem](https://github.com/containerbuildsystem/gemlock-parser/blob/master/gemlock_parser/gemfile_lock.py#L421)

chmeliik: Oh interesting. It sounds like we can use that to solve the metadata parsing https://github.com/containerbuildsystem/cachi2/pull/565#discussion_r1672452499
(and still fall back to git data if there's no gemspec in the root)

docs/design/rubygems.md
slimreaper35: Only BUNDLE_APP_CONFIG takes precedence

slimreaper35: There is also BUNDLE_IGNORE_CONFIG, but this will ignore _all_ environment variables

chmeliik: > There is also BUNDLE_IGNORE_CONFIG, but this will ignore all config variables

Yeah that doesn't sound like a good option, we don't want to mess with the users' settings that much

> Only BUNDLE_APP_CONFIG takes precedence

so we could potentially do something like
* copy the .bundle/config from the repo into the cachi2 output dir
* set the variables we need in the copied config
* set `BUNDLE_APP_CONFIG=<configfile in output dir>`

Although simply overwriting their config via `inject-files` would be fine too, copying it elsewhere is just an optional nice-to-have improvement IMO

cachi2/core/package_managers/rubygems.py
```diff
@@ -0,0 +1,264 @@
...
+log = logging.getLogger(__name__)
+
+
+def fetch_rubygems_source(request: Request) -> RequestOutput:
+    """Resolve and fetch RubyGems dependencies."""
+    components = []
+    environment_variables = [
+        EnvironmentVariable(name="BUNDLE_CACHE_ALL", value="true"),
+        EnvironmentVariable(name="BUNDLE_CACHE_PATH", value="${output_dir}/deps/rubygems"),
+        EnvironmentVariable(name="BUNDLE_FORCE_RUBY_PLATFORM", value="true"),
```
brunoapimentel, line 31: I'm a little confused if these variables are really relevant for the build phase. I was able to successfully build the [test repo](https://github.com/slimreaper35/cachi2-rubygems) you set up and the only thing I really needed was to properly set `BUNDLE_CACHE_PATH`.

By teading the bundler docs, I got the impression that those variables would be relevant in case we'd use bundler for doing the prefetch, which is not the case.

brunoapimentel:
I noticed that it is necessary to [add](https://github.com/slimreaper35/cachi2-rubygems/blob/main/Containerfile#L10) `--local` to the bundle install in order to successfully perform a offline install. The problem is, this is something we can't really solve just by injecting project files or source env variables.

This would force the user to modify his Dockerfile, which is still something we don't require with Cachi2. I wonder if there's a way to work around that.

brunoapimentel:
I did a small test trying to build [Porta](https://github.com/3scale/porta) with this PoC here (while disabling the Yarn deps :see_no_evil:).

The prefetch was successful, but the build failed because it was trying to reach out to grab the git dependencies. Were you able to successfully prefetch and build any project that has git deps? I think we should enhance our cachi2-rubygems repo to include at least one git dep.

cachi2/core/package_managers/rubygems.py
slimreaper35: Yeah, I will remove `BUNDLE_CACHE_ALL`. Checking the ruby platform is currently done by gemlock-parser. Do we want to set that during the build as well ?

brunoapimentel:
> I tried to set export BUNDLE_ALLOW_OFFLINE_INSTALL=true and remove --local, but it ain't work.
I found this workaround => commenting out source "https://rubygems.org" in Gemfilem, but I see a deprecation warning.

The deprecation warning worries me a little. Also, the official [docs](https://bundler.io/v2.5/man/gemfile.5.html#GLOBAL-SOURCE) seem to indicate that this is a required setting. In any case, I think we can add it as an option in the design doc and leave it up for discussion.

I wonder if turning all dependencies into [path](https://bundler.io/v2.5/man/gemfile.5.html#PATH) dependencies would work? This is a much more invasive change, and I would not like to turn to it, though.

brunoapimentel:
> I know; that was copied from cachito. Git dependencies must have this format (I did not implement it in the PoC):
https://github.com/containerbuildsystem/cachi2/pull/565/files#diff-cf42ff83f7bd882bd043479db1ad05961c5ef7218a816f2640347dd50c08fec8R193

Oh, I also did a small test with a git dependency and I was able to build the image. The thing that was not clear to me immediatelly was that the repository files need to be unpacked in a folder following the naming format you mentinoned.

Since what the prefetch does is to download a tarball of the repo+ref, it took me some time to figure it out. Maybe we can add a note in the design doc about this?

chmeliik (commented):
Could you add steps for how to locally try out the proposed solution?

There have been some fairly significant changes since the last state and I would like to test them

slimreaper35:
> Could you add steps for how to locally try out the proposed solution?

You can play around with my repo:
https://github.com/slimreaper35/cachi2-rubygems

chmeliik:
> > Could you add steps for how to locally try out the proposed solution?
> 
> You can play around with my repo: https://github.com/slimreaper35/cachi2-rubygems

I meant add the steps to the design doc. It would be a nice summary of more or less what cachi2 will do to enable offline installation (IMO it's not super easy to piece it together from the doc currently). And I would (I'm sure others too) appreciate a step-by-step guide, even I can probably figure it out by myself

slimreaper35:
I can do it, but it won't be 100% functional since offline installation of git dependencies does not work

chmeliik:
> I can do it, but it won't be 100% functional since offline installation of git dependencies does not work

IMO the steps can say to use `bundler cache` to create the cache dir (but very clearly call out that cachi2 won't be calling bundler cache but re-implementing it)

slimreaper35:
How could we set `BUNDLE_CACHE_PATH` in our copied config file before we know `${output_dir}` ?

docs/design/rubygems.md
brunoapimentel: If what this user [says](https://github.com/rubygems/rubygems/issues/3379#issuecomment-1974761996) is correct, the checksum generation for the lockfile is finished, but it is still an internal feature, not exposed in the bundler API.

I totally agree this should be dealt with in a follow-up story.

docs/design/rubygems.md
```diff
@@ -0,0 +1,450 @@
...
+When the Gemfile is changed, and the lockfile has not been updated, running Bundler commands will be blocked.
+
+**BUNDLE_FORCE_RUBY_PLATFORM**: Ignore the current machine's platform and install only ruby platform gems.
+As a result, gems with native extensions will be compiled from source.
+
+See bundle config [documentation](https://bundler.io/v2.5/man/bundle-config.1.html).
+
+---
+
+Note: _BUNDLE_FORCE_RUBY_PLATFORM_ check is done via gemlock-parser in cachito.
```
brunoapimentel, line 310: I was trying to double check this, but I think I got lost in the process. Can you point me to where we're setting this in the gemlock-parser?

docs/design/rubygems.md
```diff
@@ -0,0 +1,450 @@
...
+
+```bash
+podman build . \
+  --volume "$(realpath ./cachi2-output)":/tmp/cachi2-output:Z \
+  --volume "$(realpath ./cachi2.env)":/tmp/cachi2.env:Z \
+  --network none \
+  --tag ruby
+```
+
+### Summary
```
brunoapimentel, line 424: We could also add a task to create a custom test repo, [similar](https://github.com/cachito-testing/cachi2-gomod) to what we do with other package managers, trying to explore all the corner cases we can thing of. This should be probably one of the first tasks we tackle.

docs/design/rubygems.md
```diff
@@ -0,0 +1,450 @@
...
+
+Copy the local configuration file from the user repository to {output_dir} and set BUNDLE_APP_CONFIG to the new location.
+Then, just append all the environment variables needed to the "new" copy of the user configuration file.
+Bundler will rewrite previous values with the new ones when installing gems.
+
+#### Inject
+
+The other solution would be to inject the config file directly and rewrite the values.
+
+### Metadata
```
brunoapimentel, line 342: I think we should clarify here how each field of the SBOM component will be resolved from the metadata. My initial ideas:

- name: from .gemspec, or if missing, from the repo url
- version: from .gemspec, or if missing, generate a pseud-version (or leave it empty?)
- purl: main url based on the repo origin (ignore "source_code_uri" from .gemspec?)

Path dependencies should have their purls based on the source repo origin URL, and with the relative path appended.

docs/design/rubygems.md
```diff
@@ -0,0 +1,450 @@
...
+PATH
+ remote: .
+ specs:
+ tmp (0.1.2)
+...
+```
+
+### PURL
+
+Examples from [github.com/purl-spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#gem).
```
brunoapimentel, line 377: Do think we can add PURL examples for each of the existing dependency types (or more if there are corner cases that are interesting to clarify)?

docs/design/rubygems.md
slimreaper35: How could we set BUNDLE_CACHE_PATH in our copied config file before we know ${output_dir} ?

docs/design/rubygems.md
slimreaper35: https://github.com/containerbuildsystem/cachito/blob/master/cachito/workers/pkg_managers/rubygems.py#L101

docs/design/rubygems.md
slimreaper35: There is 
`add integration and e2e tests` below

docs/design/rubygems.md
slimreaper35: Yeah, but we don't really need to do anything with the `.gemspec` file. _Gemfile.lock_ and repo URL should give us enough information.

slimreaper35:
@brunoapimentel pointed out correctly that git dependencies can have a different name specified in the `Gemfile` than the repository name. However, the cached directory must have the name from the git URL. I fixed that. Plus, I removed all mentions of my custom repository from the document. In the meantime, I will start working on high-level design code.

brunoapimentel:
I found another bug with the git dependencies implementation. I set up this test repo [here](https://github.com/brunoapimentel/chatwoot/tree/hermetic-build-test). This one has a funky git dependency:

```
GIT
  remote: https://github.com/chatwoot/azure-storage-ruby
  revision: 9957cf899d33a285b5dfe15bdb875292398e392b
  branch: chatwoot
  specs:
    azure-storage-blob (2.0.3)
      azure-storage-common (~> 2.0)
      nokogiri (~> 1, >= 1.10.8)
    azure-storage-common (2.0.4)
      faraday (~> 2.0)
      faraday-follow_redirects (~> 0.3.0)
      faraday-net_http_persistent (~> 2.0)
      net-http-persistent (~> 4.0)
      nokogiri (~> 1, >= 1.10.8)

```

Cachi2 downloads the folders as:
```
azure-storage-blob-9957cf899d33/                             
azure-storage-common-9957cf899d33/   
```

But bundler expects it as:
```
azure-storage-ruby-9957cf899d33/
```

slimreaper35:
Yes, so:

1. The name of the directory must come from the git URL, not the name of the actual gem
2. I only copied the default branch in my PoC, so that's another thing that needs to be fixed

taylormadore (commented):
I see some mention of [specifying development dependencies](https://guides.rubygems.org/specification-reference/#add_development_dependency). Is that something we may want to (at some point) account for in the SBOM?

docs/design/rubygems.md
```diff
@@ -0,0 +1,425 @@
...
+1. Local config (`<project_root>/.bundle/config or $BUNDLE_APP_CONFIG/config`)
+2. Environment variables (ENV)
+3. Global config (`~/.bundle/config`)
+4. Bundler default config
+
+Since the local configuration takes higher precedence than the environment variables (except BUNLDE_APP_CONFIG),
+we need to set the Bundler configuration options to make the build work.
+We can easily set the environment variables if the local configuration file does not exist.
+
+#### Relevant environment variables
```
taylormadore, line 295: Do we also need to set BUNDLE_NO_PRUNE=true during the build for the multiple packages scenario to prevent one ```bundle install``` command from pruning the cache?

brunoapimentel:
I took over the design while @slimreaper35 is away. I initially just wanted to shuffle some info around to address @chmeliik's comment

> My general observation is that it's kind of hard to derive implementation guidance from the doc because it's not always clear which parts are just general commentary on RubyGems/Bundler and which parts describe the implementation decisions

But it grew bigger than I had antecipated :sweat_smile:.

Overall changes:
- Tried to bring a clearer separation between the multiple sections, specially separating the ecossystem overview from the design itself. This made me shuffle quite a few paragraphs around.
- Rewrote some paragraphs to make them clearer.
- Added examples of PURLs per dependency type.

Design points that got changed:
- Support for pre-compiled binaries is now listed as a follow-up story, instead of something we will never support. From my understanding, Cachi2's goal is not to block pre-compiled binaries, but to report them.
- Build configuration will be set via `inject-files`. It had two options up for discussion previously, but I don't think we have a choice on this. 
- Main package metadata will be read from the .gemspec file. Previously, it was supposed to be read from the `Gemfile.lock` and the repo URL, but I think that method is faulty.

For more info, refer to the specific sections of this doc and the [backup](https://github.com/slimreaper35/cachi2/tree/ruby-design-backup) of the previous design before the rewrite.

docs/design/rubygems.md
brunoapimentel: I found out that it is not mandatory to have the "main" gem entry in the `Gemfile.lock`. Bundler will process it without any issues, and the gemlock-parser will just leave the `primary_gem` attribute as empty.

So I don't think we can rely on this 100%. I'd rather just implement some basic regex parsing for the `.gemspec` file, since that one seems to be required from Bundler's perspective.

brunoapimentel: Here's a [sample file](https://github.com/chatwoot/chatwoot/blob/develop/Gemfile.lock) that does not contain the "main" gem as a PATH dependency.

brunoapimentel:
New push:
- Added a section about alternate registries and specified that support for them will be a follow-up feature
- Added a section explaining dependency groups and specified that support for dev dependencies is a potential follow-up feature
- Added  `BUNDLE_NO_PRUNE` as a necessary configuration parameter for the build phase

docs/design/rubygems.md
chmeliik: Not mandatory as in which one of these
* the user has to do something special to make sure the main gem appears in Gemfile.lock
* if the user manually removes the main gem from Gemfile.lock, bundler will process the project just fine and put it back
* if the user manually removes the main gem from Gemfile.lock, bundler will not put it back

docs/design/rubygems.md
brunoapimentel: I added this variable as part of the necessary variables for build in the design.

docs/design/rubygems.md
brunoapimentel: So, for us, this means we can reliably read the `.gemspec` metadata from the `Gemfile`. And if it does not exist, we can fall back to using the Git URL for name, and leave the version empty.

docs/design/rubygems.md
```diff
@@ -0,0 +1,522 @@
...
+Bundler's functionality in any way that they seem fit. Since this can open the possibility for security issues, plugins
+will not be supported by Cachi2.
+
+Since we're not proposing the direct usage of Bundler to fetch the dependencies, no other actions are needed in the
+prefetch phase, existing plugin definitions will be ignored.
+
+#### Pre-compiled binaries
+For the initial implementation, we're aiming to provide support only for plain Ruby gems (which are idenfied as `RUBY`
+in the `PLATFORMS` section of the `Gemfile.lock`). This means that we need to explicitly forbid the processing of any
+repository that has references to any platform besides `RUBY`.
```
chmeliik, line 346:
```suggestion
For the initial implementation, we're aiming to provide support only for plain Ruby gems (which are idenfied as `ruby`
in the `PLATFORMS` section of the `Gemfile.lock`). This means that we need to explicitly forbid the processing of any
repository that has references to any platform besides `ruby`.
```

docs/design/rubygems.md
```diff
@@ -0,0 +1,522 @@
...
+specified in the `Gemfile.lock`. This way, we can avoid downloading any pre-compiled content.
+
+We should probably implement support for pre-compiled binaries as a follow-up feature, similarly to what was done with
+[pip wheels](https://github.com/containerbuildsystem/cachi2/blob/main/docs/pip.md#distribution-formats).
+
+#### Checksum verification
+Since checksums in the `Gemlock.file` is still a feature in development (see [checksums](#dependency-checksums)), we
+can postpone implementing support for it until the feature is delivered.
+
+We need to decide if we will report all dependencies as having missing checksums in the SBOM, or not.
```
chmeliik, line 358: I would say yes, but only if `BUNDLED WITH >= $the_version_that_will_officially_add_support`

docs/design/rubygems.md
```diff
@@ -0,0 +1,522 @@
...
+
+#### Alternate registries
+Users can specify different registries by leveraging the `source` keyword on the `Gemfile`. This setting can be global
+or target to a specific group of Gems. See the [official docs](https://bundler.io/guides/gemfile.html#gemfiles) for
+more info.
+
+We can likely just support the use case of the global `source` being set to `'https://rubygems.org'` (the default)
+value. Cachi2 will need to check for ny changes on the global `source` or additional `source` blocks added to the
+`Gemfile`, and fail the request in case they're found.
```
chmeliik, line 367: It's also reflected in Gemfile.lock, right? Technically there's no reason not to support it

```Gemfile.lock
GEM
 remote: ...
```

docs/design/rubygems.md
```diff
@@ -0,0 +1,522 @@
...
+2. Environment variables (ENV)
+3. Global config (`~/.bundle/config`)
+4. Bundler default config
+
+Since the local configuration takes higher precedence than the environment variables (except `BUNDLE_APP_CONFIG`), we
+need to set the Bundler configuration options to make the build work.
+
+This can be achieved by using Cachi2's `inject-files` command to add the necessary keys to the local configuration,
+while respecting the other values that might exist there. In case the local config file does not exist, it needs to be
+created by this command.
```
chmeliik, line 398: Could we `inject-files` the config into `${output_dir}/deps/rubygems/config` and set `BUNDLE_APP_CONFIG=${output_dir}/deps/rubygems` to avoid writing directly into the repo?

docs/design/rubygems.md
```diff
@@ -0,0 +1,522 @@
...
+
+#### Relevant configuration for the build
+
+```txt
+BUNDLE_CACHE_PATH=${output_dir}/deps/rubygems
+BUNDLE_DEPLOYMENT=true
+BUNDLE_FORCE_RUBY_PLATFORM=true
+BUNDLE_NO_PRUNE=true
```
chmeliik, line 406: Does it work for git deps even without BUNDLE_CACHE_ALL?

docs/design/rubygems.md
```diff
@@ -0,0 +1,522 @@
...
+
+### Out of scope
+
+#### Plugins
+Bundler has support for using [plugins](https://bundler.io/guides/bundler_plugins.html), which allows users to extend
+Bundler's functionality in any way that they seem fit. Since this can open the possibility for security issues, plugins
+will not be supported by Cachi2.
+
+Since we're not proposing the direct usage of Bundler to fetch the dependencies, no other actions are needed in the
+prefetch phase, existing plugin definitions will be ignored.
```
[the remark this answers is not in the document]
chmeliik: > If we prefetched the plugins

Btw, do they appear in the lockfile or not?

chmeliik:
The rest LGTM :taco:

docs/design/rubygems.md
brunoapimentel: I changed the design to reflect the conclusion here.

docs/design/rubygems.md
brunoapimentel: We can probably file a story to keep an eye on this feature, and update our implementation when it is released.

docs/design/rubygems.md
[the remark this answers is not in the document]
brunoapimentel: So, I was messing around with a very bare bones project, trying to install the plugins [listed](https://bundler.io/guides/plugins.html) in the official docs. 

A lot of them simply failed to install, but some of them did. Apparently they get installed to `.bundle/plugin/gems`. I did not see any entry in the `Gemfile.lock`, and doing a `bundle plugin install` does not automatically add a `plugin` entry to the `Gemfile`.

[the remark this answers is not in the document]
brunoapimentel: Since there's no easy way to identify plugins from Cachi2's perspective, I think we can probably ignore them for now, and maybe file an issue to investigate their usage (I've yet to see a project that uses them).

docs/design/rubygems.md
brunoapimentel: Fixed

docs/design/rubygems.md
chmeliik: (actually, maybe not - to support two or more bundler projects in one repo, I think we'd have to overwrite the individual files)

docs/design/rubygems.md
slimreaper35: What is exactly wrong with taking it from `Gemfile.lock` and checking if `PATH` == `.` ?
(Even if it is not mandatory)

brunoapimentel: There's nothing wrong with it.

I had the false impression that a `.gemspec` could exist while the `PATH == .` entry in `Gemfile` was absent. That is not what happens, and the only corner case where it could happen is if `gemspec` was not included in the `Gemfile`, which I don't think is Cachi2's problem anyway.

docs/design/rubygems.md
```diff
@@ -0,0 +1,524 @@
...
+- parse all gems from `Gemfile.lock`
+- implement metadata parsing either from Git origin url or `Gemfile.lock`
+- download all gems from rubygems.org, including Bundler
+- download all gems from Git repositories
+- validate path dependencies are relative to the project root
+- handle Bundler configuration options and environment variables
+- generate PURLs for all dependencies
+- add integration and e2e tests
+- add documentation
+- implement checksum parsing and validation when prefetching (follow-up)
```
slimreaper35, line 517: Can we add all other follow-up tasks here? Ideally, separately for better readability.

taylormadore:
Overall, looking good to me :+1:

docs/design/rubygems.md
brunoapimentel: You're right, I think I went on the lazy side here. We can simply do:

```
url = f"{gem.source}/gems/{gem.name}-{gem.version}.gem"
```

If we assume that the alternate registry will behave exactly like "rubygems.org".

docs/design/rubygems.md
brunoapimentel: Yes, it does. I believe this setting only affects the `bundle cache` command, so it has no effect on the build.

brunoapimentel:
New push:
- Changed the check for pre-compiled binaires (platforms other than `ruby`)  to print a warning instead of failing the request
- Added more info in the section about plugins
- Moved the feature to download Bundler itself to a follow up
- Added the follow-up stories to the scoping in the end of the design doc

docs/design/rubygems.md
brunoapimentel: Done

docs/design/rubygems.md
```diff
@@ -0,0 +1,522 @@
...
+  spec.version     = "0.1.0"
+  spec.authors     = ["Nobody"]
+  spec.email       = ["ruby@example.com"]
+  spec.summary     = "Write a short summary, because RubyGems requires one."
+  end
+  ```
+
+### Dependency types
+There are four types of
+[sources](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb#L48) for dependencies
```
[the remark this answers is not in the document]
brunoapimentel: Fixed it, and added more info on plugins

docs/design/rubygems.md
```diff
@@ -0,0 +1,533 @@
...
+or use `BUNDLE_APP_CONFIG` to point Bundler to a config directory within the Cachi2 output directory. The latter has
+the benefit of not needing to dirty the cloned sources, but it wouldn't be able to support a multiple Ruby project per
+repository scenario (since we would need to keep multiple configuration files).
+
+#### Relevant configuration for the build
+
+```
+BUNDLE_CACHE_PATH=${output_dir}/deps/rubygems
+BUNDLE_DEPLOYMENT=true
+BUNDLE_FORCE_RUBY_PLATFORM=true
```
brunoapimentel, line 411: From the tests I did, I didn't need to set this in order for the hermetic build to work properly. Can you confirm, @slimreaper35?

brunoapimentel:
I left two test scenarios ready to be tried out:
https://github.com/brunoapimentel/cachi2-playground/tree/porta
https://github.com/brunoapimentel/cachi2-playground/tree/chatwoot

There's also a "pademonium" style test repo here: https://github.com/brunoapimentel/cachi2-rubygems/

docs/design/rubygems.md
slimreaper35: I believe we don't need `BUNDLE_FORCE_RUBY_PLATFROM` anymore. It just forces bundler to download gems without native extensions. But we download everything manually... I tried to remove the env variable and the `ruby` from `PLATFORMS`, and the build works.

brunoapimentel (approved):

brunoapimentel (changes requested):
Content-wise, this PR is approved. But we need to decide if we want to use [ADRs](https://github.com/konflux-ci/architecture/blob/main/ADR/0000-adr-template.md) or some other standard for these design docs. So let's hold on merging this until it is done.

.gitignore
```diff
@@ -117,3 +117,7 @@ settings.py
 
 # idea files
 .idea/
+
+# cachi2 default files
+cachi2-output/
+cachi2.env
```
slimreaper35, line 123: Should I drop this commit since I saw https://github.com/containerbuildsystem/cachi2/issues/597 ?

docs/design/rubygems.md
```diff
@@ -0,0 +1,528 @@
...
+3. [Design for the Cachi2 implementation](#iii-design-for-the-implementation-in-cachi2)
+
+## I. Ruby ecosystem overview
+
+### Development prerequisites
+In order to execute the commands in the examples below, make sure you have the following packages installed in your
+environment:
+
+```bash
+sudo dnf install rubygems rubygems-bundler
```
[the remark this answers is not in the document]
slimreaper35: You are right, it should be `rubygem`

docs/design/rubygems.md
```diff
@@ -0,0 +1,528 @@
...
+`:development` are dev dependencies.
+
+### Dependency checksums
+The support to checksums in the `Gemfile.lock` is still in development, and currently is an
+[opt-in feature](https://github.com/rubygems/rubygems/pull/7217). To enable it, we need to manually add a `CHECKSUMS`
+section in the `Gemfile.lock`:
+
+```shell
+# manually add `CHECKSUMS` section somewhere in the Gemfile.lock
+vim Gemfile.lock
```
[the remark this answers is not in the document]
slimreaper35: Probably the command doesn't even need to be there..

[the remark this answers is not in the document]
a-ovchinnikov: No, no, please keep a command, the command does help those without much context! My suggestion is to replace it with `echo` -- `echo` makes it obvious what is about to happen to `Gemfile.lock`.

.gitignore
ben-alkov: No, I think those are closely related to cachi2 and it makes sense to keep them in the project scope.

eskultety: Nope, these are our files so while the particular use case is uncommon and kinda a corner case, it's still ours.

eskultety: I would reword the commit message a bit to avoid 1st person narration and replace it with 3rd person narration as is common with technical writings. Also, sentences begin with capital letters, but that's just me nitpicking :).

eskultety: Uhm, looking at the PR, this commit should really be proposed standalone, outside the context of this particular PR.

slimreaper35: Will this PR be merged actually? Bruno already created a new one https://github.com/containerbuildsystem/cachi2/pull/601

brunoapimentel: Yes. #601 should only merge the basic template. This will be refactored after #601 comes in.

docs/design/rubygems.md
ben-alkov: I feel like noting missing checksums should be an **SBOM** feature which is independent of specific package managers.

If nothing else, we're going to have a situation at some point in the future where existing (old) ruby SBOMS **without** checksums look exactly like current ruby SBOMs **with** checksums (because SBOMs simply ignore checksums if they exist)...

docs/design/rubygems.md
```diff
@@ -0,0 +1,528 @@
...
+  ```ruby
+  source "https://rubygems.org"
+
+  gem "rails", "= 6.1.7"
+  ```
+
+- **Gemfile.lock**: A file that locks the versions of gems that are installed for your project. Bundler uses this file
+to ensure that the correct versions of gems are installed consistently across different environments.
+
+- **RubyGems**: General package manager for Ruby. Manages installation, updating, and removal of gems globally on your
```
ben-alkov, line 51: Or is it "RubyGem" (IDK)?

```suggestion
- **RubyGem**: General package manager for Ruby. Manages installation, updating, and removal of gems globally on your
```

docs/design/rubygems.md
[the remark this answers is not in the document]
ben-alkov: Are we OK with being RPM-centric here? Do we want to at least mention valid OS package names for e.g. Deb/Ubuntu and Arch (2 of the Big 3 :D)

docs/design/rubygems.md
```diff
@@ -0,0 +1,528 @@
...
+[group](https://bundler.io/guides/groups.html). The name of the group can be any string, but the usual groups tend to
+be common labels such as `:test`, `:development` or `:production`.
+
+Here's how it looks like in a `Gemfile`:
+
+```ruby
+# :default group
+gem 'nokogiri'
+
+group :test do
```
ben-alkov, line 189: This (and the 'gemspec' file above), seems to be executable Ruby code. If that is the case, are there any restrictions on what is allowed? I don't think that cachi2 wants to allow just anything at all (c.f. Yarnberry :skull:)

docs/design/rubygems.md
```diff
@@ -0,0 +1,528 @@
...
+end
+```
+
+Another way to declare a dependency in the `:development` group is to
+[add it](https://guides.rubygems.org/specification-reference/#add_development_dependency) to the `Gem::Specification`,
+which is usually declared in the `.gemspec` file. This means we can safely assume that all dependencies under
+`:development` are dev dependencies.
+
+### Dependency checksums
+The support to checksums in the `Gemfile.lock` is still in development, and currently is an
```
ben-alkov, line 201:
```suggestion
The support for checksums in the `Gemfile.lock` is still in development, and currently is an
```

docs/design/rubygems.md
ben-alkov: P.S. Trust me that it **can be** quite difficult to implement tracking checksum presence and responsible file(s) *after the initial implementation* (ask me how I know :grimacing:)

docs/design/rubygems.md
ben-alkov: Ignore, I see that this is addressed later in the doc

docs/design/rubygems.md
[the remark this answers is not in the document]
slimreaper35: I used `dnf` because it is used in the docs for example [README.md](https://github.com/containerbuildsystem/cachi2?tab=readme-ov-file#virtual-environment)

.gitignore
eskultety: > What is currently there is only an example to give us ground for discussion.

Correct, this PR has all the context, so the work should continue here.

docs/design/rubygems.md
brunoapimentel: Resolving this as a conclusion was reached.

docs/design/rubygems.md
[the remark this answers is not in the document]
brunoapimentel: I also wonder how much value these bring. Can we just add links to the official docs?

docs/design/rubygems.md
brunoapimentel: So you're saying we should mark all Ruby components as `missing_checksum_in_file`?

docs/design/rubygems.md
slimreaper35: RubyGems
https://github.com/rubygems/rubygems

docs/design/rubygems.md
ben-alkov: Yes

eskultety: Remember this is ultimately a user's decision, so IIUC this whole thing, unless users opt-in explicitly or use a new enough bundler, then in order to **always** report checksums, we'd not only have to opt-in that behind users's backs we'd also dirty their repo by modifying the lockfile, am getting it right? If so, then I'm also against such practice and instead report that in the SBOM, after all, this should be user's responsibility to better harden what their source code depends on, we're only a messenger here really.

docs/design/rubygems.md
```diff
@@ -0,0 +1,528 @@
...
+
+1. Local config (`<project_root>/.bundle/config or $BUNDLE_APP_CONFIG/config`)
+2. Environment variables (ENV)
+3. Global config (`~/.bundle/config`)
+4. Bundler default config
+
+Since the local configuration takes higher precedence than the environment variables (except `BUNDLE_APP_CONFIG`), we
+need to set the Bundler configuration options to make the build work.
+
+In order to do this, we can either use `inject-files` to overwrite the `.bundle/config` directory in the source folder,
```
eskultety, line 401: how would `inject-files` know how to overwrite each config based on our build config JSON? We'd have to create mappings of config paths -> env vars (or directly config keys) and re-format those as key=value pairs in the bundler config files in question. I guess there's no way around it, is it? Although we don't seem to handle this in Yarn, a bug probably?

docs/design/rubygems.md
eskultety, line 51: nitpick: is it correct to call this RubyGem support if only intend to talk to `bundler` at best (with good chunk of logic done in our code directly)?

docs/design/rubygems.md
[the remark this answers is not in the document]
eskultety: in context of the other checksum  thread this section may end up being dropped from the design doc and instead be mentioned in the user-facing pkg manager doc as a sort of guideline for a good repo configuration setup.

docs/design/rubygems.md
```diff
@@ -0,0 +1,528 @@
...
+
+To avoid arbitrary code execution, Bundler **is not used** to download dependencies. Instead, as stated above, Cachito
+parses `Gemfile.lock` file directly and download the gems from <https://rubygems.org>.
+
+**Note**: parsing `Gemfile.lock` is done via [gemlock-parser](https://github.com/containerbuildsystem/gemlock-parser),
+which is vendored from
+[scancode-toolkit](https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/gemfile_lock.py).
+
+Source code for "official" Bundler lockfile parsing in Ruby:
+<https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb>
```
eskultety, line 245: We're adding this to cachi2 and yes, much of the logic will be ported, but we don't need to mention cachito here, we're trying hard to drop all references to it ultimately to just mentioning in the main README that "origin of this project was cachito, link here..." but I don't think this should be in the design doc since we probably do end up making different decisions and the implementation will also be different in many cases.

docs/design/rubygems.md
eskultety: > Although we don't seem to handle this in Yarn, a bug probably?

Correction, we do! Because config opts have the correct order of precedence, so 1 set of env vars works like a charm.

docs/design/rubygems.md
ben-alkov: No.

> in order to always report checksums

We'll **only** report that they are *missing*, which requires nothing from the user. If the user *starts* using checksums, we'll no longer add the "missing" annotation to the SBOMs. That's all.

chmeliik: It would be to rude report that checksums are missing if the package manager doesn't even allow the user to specify checksums. And bundler currently doesn't (not officially, and I don't think our documentation should recommend that users do something their tool treats as tech preview)

ben-alkov: OK :shrug:.

From my POV, it's more important to have consistency across SBOMs regardless of package manager, particularly given that we **only** report checksums if they are *missing* (it would be completely different if we reported presence AND absence).

Now an SBOM which doesn't mention checksums will be "either *there are checksums* (which we don't report), or *there aren't checksums, but we're not reporting it*", which is so inconsistent as to be meaningless.

I'm happy to give way to the majority here, though.

docs/design/rubygems.md
```diff
@@ -0,0 +1,528 @@
...
+
+Since the local configuration takes higher precedence than the environment variables (except `BUNDLE_APP_CONFIG`), we
+need to set the Bundler configuration options to make the build work.
+
+In order to do this, we can either use `inject-files` to overwrite the `.bundle/config` directory in the source folder,
+or use `BUNDLE_APP_CONFIG` to point Bundler to a config directory within the Cachi2 output directory. The latter has
+the benefit of not needing to dirty the cloned sources, but it wouldn't be able to support a multiple Ruby project per
+repository scenario (since we would need to keep multiple configuration files).
+
+#### Relevant configuration for the build
```
a-ovchinnikov, line 406: Could it happen that there is some user configuration that clashes with our desired configuration? Or is this practically impossible?

docs/design/rubygems.md
brunoapimentel: I believe we can simply change this to Bundler. What we are proposing does not work with `gem` (the RubyGems CLI) standalone.

@slimreaper35 @a-ovchinnikov If you both agree, I can make the changes right away.

docs/design/rubygems.md
brunoapimentel: I don't think this specific section brings too much value besides linking the main reference we had, the implementation done in Cachito. So I'd be fine with droping it, but I still think there's value to link previous work, specially considering that there was much research done back then. 

Why do you think it's bad to reference Cachito? I think it's ok to say we are taking a different path with Cachi2, and the work done before is merely a reference and a starting point.

docs/design/rubygems.md
slimreaper35: Yes, I will rename `rubygems` to `bundler`.

docs/design/rubygems.md
brunoapimentel: Resolving this since a conclusion was reached.

docs/design/rubygems.md
brunoapimentel: From our initial investigation, we didn't find anything. But I believe this topic should be revisited when the actual implementation story comes up.

docs/design/rubygems.md
brunoapimentel: Ok, then I'll update the design doc as well.

docs/design/rubygems.md
slimreaper35: Technically, it is possible, but those configuration settings have no impact on the project.

a-ovchinnikov: Yes, let's postpone this until implementation time.

brunoapimentel:
New push:
- fixes typos
- renames the ending section from "potential follow-ups" to "out of scope"
- renames the package manager from "rubygems" to "bundler"

brunoapimentel (approved):

docs/design/rubygems.md
taylormadore: @MartinBasti FYI for the cachi2 OSBS compatibility layer

slimreaper35:
Drop the gitignore commit since it is already in a separate PR.

docs/design/rubygems.md
ben-alkov: This,

> there was much research done back then

and the fact that we often borrow starting code from Cachito, make it a good idea to at least provide links.

eskultety: The original project is to be sunset in the future with this one being its replacement. I don't see value in putting many references to the old project we don't really want to maintain live connections to, so while I'm okay stating that parts of the code were borrowed, linking at too many places doesn't IMO achieve what we're ultimately trying to do, because the code, yet borrowed, has to be often significantly adapted to this very code base, so providing links to code which will be 75% reworked is IMO misleading and confusing. It should only ever be nothing more than a kudos/credit to where we got inspired and that's it.

brunoapimentel: So I guess we disagree on this point. I do agree that this section as it was did not bring too much value, so I moved the content to a "References" section in the end of the document, and made it shorter to just convey the credits and links to the original project.

ben-alkov (approved):

docs/design/bundler.md
```diff
@@ -0,0 +1,516 @@
...
+compiled everything from source.
+
+To avoid confusing users if a corner case shows up, we can add a `WARNING` log in case extra platforms are detected,
+and also document it properly.
+
+Proper support for pre-compiled binaries should be probably left as a follow-up feature, similarly to what was done
+with [pip wheels](https://github.com/containerbuildsystem/cachi2/blob/main/docs/pip.md#distribution-formats).
+
+#### Checksum verification
+Since checksums in the `Gemlock.file` is still a feature in development (see [checksums](#dependency-checksums)), we
```
eskultety, line 345: typo: s/`Gemlock.file`/`Gemfile.lock` file

brunoapimentel: I'll blame that one on Javascript (and it's package-lock file) :smile:

PR #569, opened 2024-07-12
docs cleanup

slimreaper35:


docs/dependency_confusion.md
```diff
@@ -61,80 +61,78 @@ to verify that the content of all direct and indirect dependencies is what you e
...
+
+Yarn packages are identical to npm packages except that they use
+[yarn.lock](https://classic.yarnpkg.com/en/docs/yarn-lock/) as the lock file. Everything that
+applies to npm applies to yarn.
+
+The one difference is the handling of unofficial registries. If you point yarn to an unofficial
+registry via .npmrc or [.yarnrc](https://classic.yarnpkg.com/en/docs/yarnrc), this will be reflected
+in the resolved url in the lock file. Cachito will see that the url does not point to the official
+registry and will treat the dependency as a plain https dependency. If the url is accessible to
+Cachito, it will download the dependency directly without relying on npm/yarn dependency resolution.
```
eskultety: While this is probably true for yarn v1 which cachi2 can't do yet, it is definitely not true for v3 where we solely rely on yarn's native mechanism to download deps.

docs/dependency_confusion.md
```diff
@@ -61,80 +61,78 @@ to verify that the content of all direct and indirect dependencies is what you e
...
+Python packages can have various formats, which are specified across multiple PEPs. Pip supports
+most of them. Cachito support for pip, on the other hand, is quite simple. You must define all the
+direct and indirect dependencies of your package in
+[requirements.txt](https://pip.pypa.io/en/stable/user_guide/#requirements-files) style files and
+tell Cachito to process them.
+
+Cachito further restricts what you can put in your requirements.txt files. All of the dependencies
+must be
+[pinned](https://github.com/release-engineering/cachito/blob/master/docs/pip.md#pinning-versions)
+to an exact version. Cachito will refuse to process requirements files that use the --index-url or
```
eskultety: cachi2 does support `--index-url` since quite recently, not sure about `--extra-index-url` from the top of my head, so it would be worth double checking. It may or may not have an impact on the rest of the paragraph content regarding the security implications.

Needless to say that ^these fixups should go to a standalone commit to keep the changes straightforward and easy to review.

eskultety (commented):
Since modifications are being done to the file, the PR feels a bit incomplete in that it's worth checking what information still applies, otherwise we'd be still publishing inaccurate and misleading data, which in essence, yes, was still completely wrong with the original file but was also kind of tolerable with the old cachito mentions since the file was inherently inaccurate as a whole; we should avoid that with its update/conversion to cachi2.
To support my argument a bit more, we also have this sentence in the doc:

"Cachi2 never executes your code, or the code of any of your dependencies. What about your build?"

Which in hindsight is a very bold statement in general, because hey, it's only ever going to be best effort given the inherently ever-present bugs :) and is also not true with Yarn v3 where we enabled the 'exec' plugin just to get some more essential plugins working with builds with the hope that it wouldn't ever lead to arbitrary code execution (which at the time of adding it wasn't the case, but who knows, maybe future versions of Yarn v3 will change that underneath without us even knowing...)

slimreaper35:
Maybe I would throw away the whole section of package managers, and then we would need to edit the docs in 2 places, which it seems no one did, and that's why it stayed like that.

eskultety:
> Maybe I would throw away the whole section of package managers, and then we would need to edit the docs in 2 places, which it seems no one did, and that's why it stayed like that.

Sorry for not getting back. Yeah, agreed on the inconsistency. I still think there's value in explaining the background story, but we should maintain this information in the respective package manager's doc and only link to that piece of information from this document, do you agree?

@slimreaper35 can we reopen this? I don't think the discussion here should be abandoned.

slimreaper35:
I also got an idea if it is not better to select the most relevant parts and put them in the main README.md. Contributors or users would get the background story straight away.

eskultety:
> I also got an idea if it is not better to select the most relevant parts and put them in the main README.md. Contributors or users would get the background story straight away.

That is also an acceptable option. Still, package manager related bits would still need to be moved to the respective docs.

slimreaper35:
There is also another outdated [document](https://github.com/containerbuildsystem/cachi2/blob/main/docs/metadata.md) that was copied from Cachito.
This one can be either deleted or replaced by something like an SBOM explanation.

eskultety:
> There is also another outdated [document](https://github.com/containerbuildsystem/cachi2/blob/main/docs/metadata.md) that was copied from Cachito. I think this one can be either deleted or replaced with something like an SBOM explanation.

Agreed, an SBOM section should be added to the main README file for now unless we have an extensive contents for this to be put in a standalone doc and this one should be dropped - I don't think we'll concern ourselves with content manifests other than what SBOM provides.

docs/dependency_confusion.md
```diff
@@ -1,48 +1,30 @@
...
 
-To paraphrase the meaning of
-*[dependency confusion](https://medium.com/@alex.birsan/dependency-confusion-4a5d60fec610)*:
-a package manager is tricked into installing a dependency from an official, public, repository when
-it should’ve installed it from a custom, often private, repository. For example, a company may host
-its own internal npm registry which hosts internal-only packages. Then, someone on the Internet
-uploads a package of the same name to the official npm repository. When the application is
-installed, the package manager chooses the dependency package from the public repository instead
-of the internal one.
+Dependency confusion is a software supply chain attack that substitutes malicious third-party code for a legitimate internal software dependency.
```
eskultety, line 3: > that substitutes 

It doesn't substitute it, it just exploits the package manager's integral preference for default official registries.

I'd suggest rewording it slightly along the following lines:
_Dependency confusion is a spoofing type of software supply chain attack where a malicious third-party code is packaged and uploaded to common default public package repositories under an identical name as a legitimate internally hosted software dependency and thus tricking the package manager into trusting the default origin repository rather than the intended alternative source during software builds._

docs/dependency_confusion.md
```diff
@@ -1,48 +1,30 @@
...
-less. However, all the supported package managers provide means to protect yourself by following
-best practices. For example, you can remove all known attack vectors by simply verifying the
-checksums of your dependencies.
+Cachi2 is as vulnerable to dependency confusion as your package manager, not more, not less.
 
-Cachi2 enforces *some* of these best practices. Typically, each package manager provides a lockfile
-or other mechanism for pinning dependency versions. Cachi2 requires this. Verifying checksums is
-usually up to you.
+Cachi2 follows some of the best practices, such as verifying the checksums of your dependencies to ensure their integrity and prevent the inclusion of malicious code.
+Also, it requires you to pin all your dependencies to exact versions to avoid unexpected updates.
```
eskultety, line 25: These two paragraphs should come before the paragraph above:

> Cachi2 is designed to manage software dependencies without executing any code from those dependencies.

docs/dependency_confusion.md
```diff
@@ -1,48 +1,30 @@
...
+Source: <https://www.activestate.com/resources/quick-reads/dependency-confusion/>
 
-Hopefully, you are building your application in an isolated environment (such as a container build
-with no network access) to protect your own system. In the best case scenario, the malicious behaviour
-will manifest at build time and the build will fail. In the worst case scenario, the build will succeed
-but the built product will contain malicious code, which may then affect whoever tries to run it.
+## Dependency Confusion and Cachi2
 
-## Dependency Confusion vs Cachi2
+Cachi2 is designed to manage software dependencies without executing any code from those dependencies.
```
eskultety, line 14: cachi2 doesn't manage dependencies, that's a package manager's job, we just download the deps.
Again, stating that:
> without executing any code from those dependencies

there's no guarantee from our side this won't happen, we need to be clear on that it's just best effort since we can miss some internal code paths the package manager takes. So, let's be more safe with:

_cachi2 downloads software dependencies doing its best to prevent any arbitrary code execution that may be part of the dependency itself and that could lead to potentially infecting the resulting product, e.g. running a pre/post install script. This requires cachi2 to effectively reverse engineer a given package manager to understand whether the package manager itself can be fully trusted and relied upon to download a dependency using its native mechanisms or we have to work around it by implementing an abstraction ourselves. The only guarantee we can give is that it always needs to be approached with a best effort mindset end employ further attestation mechanisms to the supply chain pipeline._

docs/dependency_confusion.md
```diff
@@ -61,80 +43,64 @@ to verify that the content of all direct and indirect dependencies is what you e
...
-> a repo URL and pinned to a commit hash.
->
-> Bundler doesn't verify checksums of dependencies yet, however, there's an effort to bring
-> [internal GitLab implementation](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/92633)
-> to [the upstream](https://github.com/rubygems/rubygems/pull/5808).
-
-[cachito-1]: https://github.com/containerbuildsystem/cachito
+## pip
+
+*TL;DR: if your package is Cachito compliant, it is most likely safe. If you want to be sure, verify
```
eskultety, line 48: So you left the package managers paragraphs in, but we still mention cachito here, that's confusing.

eskultety:
> So this is version 2. I deleted _metadata.md_ and refactored _dependency_confusion.md_ except the package managers section.

so do you plan on extracting the package manager bits and  moving them to the respective docs because right now it still looks confusing especially since you left the 'cachito' references in.

docs/dependency_confusion.md
slimreaper35: I am still figuring out what to put there because I don't want to include general statements that should be part of the regular documentation. Anyway, I dropped those paragraphs so it is less confusing.

docs/dependency_confusion.md
```diff
@@ -1,140 +1,55 @@
...
+what security features it provides to protect your software supply chain. cachi2 supports the
+following package managers:
 
-## gomod
+### gomod
 
-*TL;DR: commit your go.sum.*
+- package manager for Go
+- lock file: `go.sum`
+- [docs/gomod.md](https://github.com/containerbuildsystem/cachi2/tree/main/docs/gomod.md)
```
eskultety, line 37: If you remove the whole dependency-related content from the respective package manager sections, what is the actual value of having these like you propose at all?

docs/dependency_confusion.md
```diff
@@ -1,140 +1,55 @@
...
-> most of them. Cachito support for pip, on the other hand, is quite simple. You must define all the
-> direct and indirect dependencies of your package in
-> [requirements.txt](https://pip.pypa.io/en/stable/user_guide/#requirements-files) style files and
-> tell Cachito to process them.
->
-> Cachito further restricts what you can put in your requirements.txt files. All of the dependencies
-> must be
-> [pinned](https://github.com/release-engineering/cachito/blob/master/docs/pip.md#pinning-versions)
-> to an exact version. Cachito will refuse to process requirements files that use the --index-url or
-> --extra-index-url options, which means private registries are out of the question. These two
```
eskultety, line 81: So for pip, apart from `index-url` everything else should be more or less true even now for cachi2.

docs/dependency_confusion.md
```diff
@@ -1,140 +1,55 @@
...
-> requires the lock file.
->
-> The lock file pins all dependencies to exact versions. For https dependencies, which are impossible
-> to pin, Cachito requires the integrity value (a checksum). For dependencies from the npm registry,
-> Cachito does not require integrity values, but newer versions of npm will always include them. Check
-> if all your dependencies have an integrity value, update your lock file if not.
->
-> Do not try to use private registries. If you point npm to a private registry (or any registry other
-> than the official one) via [.npmrc](https://docs.npmjs.com/cli/v6/configuring-npm/npmrc), Cachito
-> will ignore it and look in the official registry anyway.
```
eskultety, line 109: I don't think we've made any distinctive changes to NPM after porting the code from cachito, so I guess it's still valid.

docs/dependency_confusion.md
```diff
@@ -1,140 +1,55 @@
...
->
-> *TL;DR: allowing `https://rubygems.org` as the only source should mitigate the issue for GEM dependencies*
->
-> Cachito parses `Gemfile.lock` which pins all dependencies to exact versions. Cachito allows GEM dependencies
-> to be fetched only from `https://rubygems.org`, otherwise raises an error. GIT dependencies are specified using
-> a repo URL and pinned to a commit hash.
->
-> Bundler doesn't verify checksums of dependencies yet, however, there's an effort to bring
-> [internal GitLab implementation](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/92633)
-> to [the upstream](https://github.com/rubygems/rubygems/pull/5808).
```
eskultety, line 138: We don't currently have these implemented (yet), so dropping them makes sense for the time being.

docs/dependency_confusion.md
```diff
@@ -1,140 +1,55 @@
...
+### yarn
 
-[cachito-1]: https://github.com/containerbuildsystem/cachito
+- package manager for JavaScript
+- lock file: `yarn.lock`
+- [docs/yarn.md](https://github.com/containerbuildsystem/cachi2/tree/main/docs/npm.md)
```
eskultety, line 55: This is not an adequate replacement because you just replaced Yarn v1 content (which we don't support yet) with a reference to our Yarn v2+ docs (indended, **in reality you linked NPM instead**).

docs/dependency_confusion.md
eskultety: Note you can't just remove content for your convenience if it doesn't fit the immediate intention like in this PR, because we'd lose legitimate content too (and we'd have to revert the change in the future). IOW part of dropping references to the original cachito code base is also finding new home to the information that still applies instead of just outright deleting it, leaving it for a potential future follow up, that's not how things should be done. IMO it's better to have more-or-less updated content with some inaccuracies which anyone willing to do so can identify and fix than no content at all and so I didn't go too deep into our code to verify every single sentence in those paragraphs, but as I outlined in my comments, much of the information is still valid and so I'd imagine these bits to be moved to the respective package manager docs to a section clearly denoting the dependency confusion matter.

slimreaper35: I turned it back to its original state. It's not a high priority now, so I'll leave it like that.

I still don't know what should be in the package manager section; my intention is to avoid general notes that are part of the regular documentation. So, later on, we don't have to adjust docs at multiple places.

PR #578, opened 2024-07-19
Drop the reflink dependency

eskultety:
The reason for this effort is that the reflink library [1] was created
as an attempt to make use of the reflink optimization before python
gained support for the os.copy_file_range syscall. The library was
never really anything more than a band-aid and now that it's possible
to use a syscall the library even mentions on its GitHub page that
Python now implements the functionality natively.

The implementation was taken from (with some 3.9+ tweaks applied) from
an existing code proposal [2] (marked as "awaiting merge") to add the
same functionality to the 'shutil' library copying primitives and make
it completely transparent to end users. We'd have to wait a long time
to be able to use it though. Compared to the reflink library, which
used a dirty trick of copying a small file first
(in kinda error-prone way) to see if the operation raised an exception,
os.copy_file_range based solution succeeds in vast majority of
cases because if reflinks are not supported within the underlying
file system (which is nothing more than inode sharing) a copy without
the overhead of userspace <-> kernel can still continue normally, hence
reserving the 'shutil.copy2' fallback to really obscure cases
(like cross-device copying - EXDEV OR on old systems without the
 syscall - ENOSYS) or simply cases where the copying failed for some
reason which we may not even encounter ever.

[1] https://gitlab.com/rubdos/pyreflink
[2] https://github.com/python/cpython/pull/93152/files


cachi2/core/utils.py
```diff
@@ -80,11 +84,81 @@ def load_json_stream(s: str) -> Iterator:
...
+        """
+        try:
+            blocksize = max(os.fstat(fd).st_size, 2**23)  # min 8 MiB
+        except Exception:
+            # fall back to 128MiB block
+            blocksize = 2**27
+
+        # On 32-bit architectures truncate to 1 GiB to avoid OverflowError
+        if sys.maxsize < 2**32:
+            blocksize = min(blocksize, 2**30)
```
slimreaper35, line 117: How about using constants instead of comments? For example:
```
MIB_8 = 2**23
MIB_128 = 2**27
```

cachi2/core/utils.py
```diff
@@ -100,15 +174,11 @@ def _copy_using(copy_function: Callable) -> None:
...
-        try:
-            log.debug(f"Copying {origin} to {destination} using reflinks.")
-            _copy_using(reflink.reflink)
-        except reflink.ReflinkImpossibleError:
-            log.debug("Reflink copy failed, falling back to standard copy.")
-            _copy_using(shutil.copy2)
-    else:
-        log.debug(f"Copying {origin} to {destination} using a standard copy.")
+    try:
+        log.debug(f"Copying {origin} to {destination} using fast in-kernel copy.")
```
slimreaper35, line 178: Do we have an issue regarding converting the log format to a lazy variant?
I suppose that was the final decision in https://github.com/containerbuildsystem/cachi2/pull/575

cachi2/core/utils.py
```diff
@@ -1,20 +1,24 @@
...
 from cachi2.core.config import get_config
 from cachi2.core.errors import Cachi2Error
 
 log = logging.getLogger(__name__)
 
 
+class _FastCopyFailedFallback(Exception):
+    """Fast in-kernel copying failed fallback to regular copy."""
+
+
```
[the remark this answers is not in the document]
eskultety: That's fair, do we want to add it in there despite knowing it's only ever going to be used in this module and hence marked "private"?

cachi2/core/utils.py
```diff
@@ -80,11 +84,81 @@ def load_json_stream(s: str) -> Iterator:
...
+        try:
+            srcfd = fsrc.fileno()
+            destfd = fdest.fileno()
+        except OSError:
+            # invalid stream or not a regular file (doesn't use a file descriptor)
+            raise _FastCopyFailedFallback()
+
+        try:
+            while nbytes := os.copy_file_range(srcfd, destfd, count=_blocksize(srcfd)):
+                total += total + nbytes
```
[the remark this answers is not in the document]
eskultety: derp, yes.

cachi2/core/utils.py
eskultety: I guess I can do that, there's no harm, although I wanted to keep the code as authentic as possible since I've taken it from elsewhere (referenced and credited) and wanted to minimize the differences for practical reasons.

cachi2/core/utils.py
eskultety: We don't have one, I'll create one and fix this instance.

cachi2/core/utils.py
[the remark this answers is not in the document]
slimreaper35: I see, no other errors are marked as private, so it is probably okay if it is not there.

cachi2/core/utils.py
slimreaper35: I don't know what is the correct way of doing it...
Personally, I wouldn't have a problem with more significant changes compared to the implementation in CPython.

Should it be referenced in the actual function itself, or is the commit message enough?

eskultety: > Should it be referenced in the actual function itself, or is the commit message enough?

That's what commit messages are for, anyone should be capable of running blame in git and that's where they'd look for more information.

eskultety: > Personally, I wouldn't have a problem with more significant changes compared to the implementation in CPython.

The commentaries are harmless and don't require defining constants that are only going to be used here OR think of it this way - if we ever need to change the value, we need to change it at 2 places, first the value itself and then the name of the constant and all places that use it, so creating constants whose name depend on the value isn't IMO the greatest approach.

cachi2/core/utils.py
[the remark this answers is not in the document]
eskultety: For the record I extracted the blocksize getter to a standalone function to be able to mock it easily and modified the _fast_copy test case to cover this oversight of mine.

cachi2/core/utils.py
slimreaper35: I get that, but on the other hand, by using comments, we would need to change the comment and then the actual math expression. So, I don't see this as a better approach.

slimreaper35 (approved):

cachi2/core/utils.py
```diff
@@ -80,11 +84,83 @@ def load_json_stream(s: str) -> Iterator:
         yield obj
 
 
+def _get_blocksize(fd: int) -> int:
```
a-ovchinnikov, line 87: A super-minor nitpick: this is unlikely to change for a single file, so could be cached.

tests/unit/test_utils.py
```diff
@@ -84,41 +90,83 @@ def test_run_cmd_executable_not_found(
...
 
-    if reflink_supported:
-        copy_function = reflink.reflink
-    else:
-        copy_function = shutil.copy2
 
-    assert mock_shutil_copytree.call_args.kwargs["copy_function"] == copy_function
+@mock.patch("cachi2.core.utils.open")
+def test_fast_copy_fail_io_fileno(mock_open: mock.MagicMock, tmp_path: Path) -> None:
+    """Test that we correctly fall back to regular copy when encountering a non-regular file."""
```
a-ovchinnikov, line 137: This docstring is a bit misleading (the one on L152 too): here we do not test that a fallback is happening, but rather that the need to fallback was signaled.

tests/unit/test_utils.py
```diff
@@ -84,41 +90,83 @@ def test_run_cmd_executable_not_found(
...
-    origin = Path("/fake")
-    destination = Path("/phony")
+    mock_copy_range.side_effect = OSError(errno_)
 
-    copy_directory(origin, destination)
+    with pytest.raises(Exception) as ex:
+        _fast_copy(src, dest)
+
+        if errno_ == errno.EAGAIN:
+            isinstance(ex, OSError)
```
a-ovchinnikov, line 130: This line (and L132) do not seem to be doing anything. Did you want to add an assertion here?

a-ovchinnikov (commented):
Mostly LGTM and a couple of questions about tests.

tests/unit/test_utils.py
eskultety: Yes, I did.

cachi2/core/utils.py
eskultety: True.

cachi2/core/utils.py
```diff
@@ -80,11 +85,84 @@ def load_json_stream(s: str) -> Iterator:
...
+    if sys.maxsize < 2**32:
+        blocksize = min(blocksize, BLK_1GiB)
+
+    return blocksize
+
+
+def _fast_copy(src: Path, dest: Path, *, follow_symlinks: bool = True) -> int:
+    """Perform a fast in-kernel copy using os.copy_file_range syscall.
+
+    Copy data from source path to destination path using a a high-performance copy_file_range(2)
```
a-ovchinnikov, line 116: s/ a a/ a/

cachi2/core/utils.py
```diff
@@ -80,11 +85,84 @@ def load_json_stream(s: str) -> Iterator:
...
+
+            if ex.errno == errno.ENOSYS or ex.errno == errno.EXDEV:
+                raise _FastCopyFailedFallback
+
+            raise ex from None
+
+        # no data copied, copying from a pseudofilesystem? Not supported [1]
+        # [1] https://docs.python.org/3/library/os.html#os.copy_file_range
+        #
+        # this should be very seldom:
```
a-ovchinnikov, line 151: s/seldom/rare/?

cachi2/core/utils.py
```diff
@@ -1,20 +1,25 @@
...
-import reflink  # type: ignore
-
 from cachi2.core.config import get_config
 from cachi2.core.errors import Cachi2Error
 
 log = logging.getLogger(__name__)
 
 
+class _FastCopyFailedFallback(Exception):
+    """Fast in-kernel copying failed fallback to regular copy."""
```
a-ovchinnikov, line 20: It could be just me, but the sentence reads a bit hard. If you will be making another revision I'd suggest either a semicolon after 'failed' ("copying failed: fallback to regular copy") or maybe just "Signals fall back from fast-in kernel copying to regular copy"?

a-ovchinnikov (approved):
LGTM with a few non-functional nitpicks.

PR #580, opened 2024-07-22
Authenticated access for RPMs

brianwcook:
This is not ready to merge, but at this point I need a review on the approach to know whether to go ay further or not.

I was able to build cachi2 on a UBI image that can be registered with subscription-manager with just one hack - fetching an RPM for createrepo_c from the Rocky repo. 

The issue is that the rpm support is using asyncio to download RPMs instead of using dnf. This circumvents the certificate handshake the is normally handled by the dnf subscription-manager plugin. I believe that this image and the prefetch task updates I have would enable authenticated RPM prefetch..._**if** we can switch to using 'dnf download' for fetching RPMs.

Regarding the hack for createrepo_C, the only alternative I could find is to get it from a repo that requires an entitlement which would make this container non-redistributable. I tried to build a createrepo_c rpm from source but the dependencies aren't available either.

**Depends on**: https://github.com/containerbuildsystem/cachi2/pull/682


eskultety:
> I was able to build cachi2 on a UBI image that can be registered with subscription-manager with just one hack - fetching an RPM for createrepo_c from the Rocky repo.

And ^this was exactly the reason why we chose Rocky over UBI, because we'd ultimately end up with a complete Frankenstein of an image where besides fetching Golang/NodeJS from their official images we'd install individual RPMs from other RPM distros due to UBI being inadequate in terms of repo contents for cachi2 purposes (many package managers support => many potential direct dependencies) which,as a use case, ultimately may not fall exactly into UBI's intended purpose. Speaking in numbers, here's a rough comparison of [the number of packages available in both](https://github.com/containerbuildsystem/cachi2/pull/528#issuecomment-2172644004)

```
# UBI9
$ dnf list --all | wc -l
2261
================
# Rocky
$ dnf list --all | wc -l
6825
```

> The issue is that the rpm support is using asyncio to download RPMs instead of using dnf. This circumvents the certificate handshake the is normally handled by the dnf subscription-manager plugin. I believe that this image and the prefetch task updates I have would enable authenticated RPM prefetch..._if we can switch to using 'dnf download' for fetching RPMs.

^This is a direct result of the original [design](https://github.com/containerbuildsystem/cachi2/discussions/465) which heavily revolved around the idea that a package name wouldn't be enough to identify a package uniquely using PURL which is still a problem to this day and apart from https://redhatproductsecurity.github.io/security-data-guidelines/purl/ we don't anything in the official PURL upstream guideline (not a blocker, the project looks like it needs more active maintenance anyway).
TL;DR
We could start recognizing a new field `name` in the RPM lockfile and use DNF to download a package.

> Regarding the hack for createrepo_C, the only alternative I could find is to get it from a repo that requires an entitlement which would make this container non-redistributable. I tried to build a createrepo_c rpm from source but the dependencies aren't available either.

See my response at the top.
...

AFAICT subscription manager scans the underlying platform before enabling repositories, so in general there isn't a single platform that would satisfy all use cases and all builds depending on the intended product against which the build is supposed to run, is there? In other words, not all RH product will always support all its active RHEL platforms, will they?
For example, Satellite can be deployed on RHEL-7 and RHEL-8 (but not RHEL-9) depending on the exact Satellite version one is installing, so on UBI-9 the subscription manager would never enable the repositories, see the example below. This all means we'd have to maintain multiple concurrent cachi2 container images to satisfy just Red Hat pipeline build needs - sounds like a downstream candidate IMO rather than upstream. Then again, upstream aside, someone will still have to solve which UBI-based cachi2 image would be used in such a pipeline based on what the intended build is for IIUC.

Another minor issue is multi-arch. With RPM URLs, cachi2 doesn't need to care about non-x86 arches. With subscription manager the consumer has to make sure that **all** arch repos are enabled using the activation key so that `dnf download --arch` would work, but admittedly this problem relies outside of this project.

```
# UBI8
[root@64b60c84e7ef /]# dnf repolist
Updating Subscription Management repositories.
repo id                                        repo name
rhel-8-for-x86_64-appstream-rpms               Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)
rhel-8-for-x86_64-baseos-rpms                  Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)
satellite-6.15-for-rhel-8-x86_64-rpms          Red Hat Satellite 6.15 for RHEL 8 x86_64 (RPMs)
satellite-6.15-for-rhel-8-x86_64-source-rpms   Red Hat Satellite 6.15 for RHEL 8 x86_64 (Source RPMs)
ubi-8-appstream-rpms                           Red Hat Universal Base Image 8 (RPMs) - AppStream
ubi-8-baseos-rpms                              Red Hat Universal Base Image 8 (RPMs) - BaseOS
ubi-8-codeready-builder-rpms                   Red Hat Universal Base Image 8 (RPMs) - CodeReady Builder
===========================================================
# UBI9
[root@668790fa564c /]# dnf repolist
Updating Subscription Management repositories.
repo id                                  repo name
rhel-9-for-x86_64-appstream-rpms         Red Hat Enterprise Linux 9 for x86_64 - AppStream (RPMs)
rhel-9-for-x86_64-baseos-rpms            Red Hat Enterprise Linux 9 for x86_64 - BaseOS (RPMs)
ubi-9-appstream-rpms                     Red Hat Universal Base Image 9 (RPMs) - AppStream
ubi-9-baseos-rpms                        Red Hat Universal Base Image 9 (RPMs) - BaseOS
ubi-9-codeready-builder                  Red Hat Universal Base Image 9 (RPMs) - CodeReady Builder
```
Goes without saying that both were registered with the same activation key.

brianwcook:
> UBI-9 the subscription manager would never enable the repositories

I think you can enable any repository you want on any platform that has subscription manager installed and is registered, even if it is for a different RHEL version or architecture.  The only requirement is that you have rights to the repo. With activation keys you can select any repos you want and then register the system with that activation key. You can also add all the repos you care about across N builds to a a single activation key and control which repos are enabled at runtime with --enablerepo or with enabled= in cachi2 repo files refrenced by the rpms.in when resolving deps and creating the lock file. Subscription-manager does try to intelligently manage the redhat.repo file though. The repolist command only shows enabled repos by default. I believe you can pass --enablerepo and download dnfs from any repo that exists and that the activation key is enabled for.

```
dnf [options] repolist [--enabled|--disabled|--all]
Lists all enabled repositories by default.
```

brianwcook:
i don't think that is correct. You just need to pass --enablerepo when you run dnf.

eskultety:
> > UBI-9 the subscription manager would never enable the repositories
> 
> I think you can enable any repository you want on any platform that has subscription manager installed and is registered, even if it is for a different RHEL version or architecture. The only requirement is that you have rights to the repo. With activation keys you can select any repos you want and then register the system with that activation key. You can also add all the repos you care about across N builds to a a single activation key and control which repos are enabled at runtime with --enablerepo or with enabled= in cachi2 repo files refrenced by the rpms.in when resolving deps and creating the lock file. Subscription-manager does try to intelligently manage the redhat.repo file though. The repolist command only shows enabled repos by default. I believe you can pass --enablerepo and download dnfs from any repo that exists and that the activation key is enabled for.
> 
> ```
> dnf [options] repolist [--enabled|--disabled|--all]
> Lists all enabled repositories by default.
> ```

This doesn't seem to be correct:
```
# UBI9
$ grep "Red Hat Satellite" /etc/yum.repos.d/redhat.repo
name = Red Hat Satellite Client 6 for RHEL 9 x86_64 - Extended Update Support (RPMs)


# UBI8
$ grep "Red Hat Satellite" /etc/yum.repos.d/redhat.repo
name = Red Hat Satellite Utils 6.14 for RHEL 8 x86_64
name = Red Hat Satellite Capsule 6.13 for RHEL 8 x86_64
name = Red Hat Satellite Maintenance 6.12 for RHEL 8 x86_64
... // the list is really long
```

> i don't think that is correct. You just need to pass --enablerepo when you run dnf.

Have you tried with arbitrary repo? Because what I'm seeing is that the `.repo` file that RSM enables doesn't even list those repos, neither for products that don't support the base plaftorm nor for alternative arches.

brianwcook:
```[root@55dc9eeeb8f8 /]# uname -r
6.8.8-300.fc40.aarch64
[root@55dc9eeeb8f8 /]# 
[root@55dc9eeeb8f8 /]# dnf download --forcearch=x86_64  --enablerepo=rhel-9-for-x86_64-baseos-rpms  kernel-devel
Updating Subscription Management repositories.
Red Hat Universal Base Image 9 (RPMs) - BaseOS                                                                                                                                                  1.2 MB/s | 516 kB     00:00    
Red Hat Universal Base Image 9 (RPMs) - AppStream                                                                                                                                               5.9 MB/s | 2.1 MB     00:00    
Red Hat Universal Base Image 9 (RPMs) - CodeReady Builder                                                                                                                                       935 kB/s | 275 kB     00:00    
        kernel-devel-5.14.0-427.26.1.el9_4.aarch64.rpm                                                0% [                                                                                            ] ---  B/s |   0  B     --kernel-devel-5.14.0-427.26.1.el9_4.aarch64.rpm                                                                                                                                                  5.3 MB/s |  24 MB     00:04    
[root@55dc9eeeb8f8 /]#

**Comment by brianwcook** [comment]
```

[root@c86479bb8ccf yum.repos.d]# cat /etc/rhsm/facts/multiarch.facts 
{"supported_architectures": "x86_64,aarch64"}

[root@c86479bb8ccf facts]# subscription-manager register --org [org] --activationkey [key]
The system has been registered with ID: [id]
The registered system name is: [name]

[root@c86479bb8ccf yum.repos.d]# dnf info kernel-devel
Updating Subscription Management repositories.
Red Hat Enterprise Linux 9 for AMD 64 - AppStream (RPMs)                                                                                                                                     21 MB/s |  38 MB     00:01    
Last metadata expiration check: 0:00:07 ago on Mon Jul 22 17:00:55 2024.
Available Packages
Name         : kernel-devel
Version      : 5.14.0
Release      : 427.26.1.el9_4
Architecture : aarch64
Size         : 24 M
Source       : kernel-5.14.0-427.26.1.el9_4.src.rpm
Repository   : rhel-9-for-aarch64-appstream-rpms
Summary      : Development package for building kernel modules to match the kernel
URL          : https://www.kernel.org/
License      : ((GPL-2.0-only WITH Linux-syscall-note) OR BSD-2-Clause) AND ((GPL-2.0-only WITH Linux-syscall-note) OR BSD-3-Clause) AND ((GPL-2.0-only WITH Linux-syscall-note) OR CDDL-1.0) AND ((GPL-2.0-only WITH
             : Linux-syscall-note) OR Linux-OpenIB) AND ((GPL-2.0-only WITH Linux-syscall-note) OR MIT) AND ((GPL-2.0-or-later WITH Linux-syscall-note) OR BSD-3-Clause) AND ((GPL-2.0-or-later WITH Linux-syscall-note) OR
             : MIT) AND Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND BSD-3-Clause-Clear AND GFDL-1.1-no-invariants-or-later AND GPL-1.0-or-later AND (GPL-1.0-or-later OR BSD-3-Clause) AND (GPL-1.0-or-later WITH
             : Linux-syscall-note) AND GPL-2.0-only AND (GPL-2.0-only OR Apache-2.0) AND (GPL-2.0-only OR BSD-2-Clause) AND (GPL-2.0-only OR BSD-3-Clause) AND (GPL-2.0-only OR CDDL-1.0) AND (GPL-2.0-only OR
             : GFDL-1.1-no-invariants-or-later) AND (GPL-2.0-only OR GFDL-1.2-no-invariants-only) AND (GPL-2.0-only WITH Linux-syscall-note) AND GPL-2.0-or-later AND (GPL-2.0-or-later OR BSD-2-Clause) AND
             : (GPL-2.0-or-later OR BSD-3-Clause) AND (GPL-2.0-or-later OR CC-BY-4.0) AND (GPL-2.0-or-later WITH GCC-exception-2.0) AND (GPL-2.0-or-later WITH Linux-syscall-note) AND ISC AND LGPL-2.0-or-later AND
             : (LGPL-2.0-or-later OR BSD-2-Clause) AND (LGPL-2.0-or-later WITH Linux-syscall-note) AND LGPL-2.1-only AND (LGPL-2.1-only OR BSD-2-Clause) AND (LGPL-2.1-only WITH Linux-syscall-note) AND LGPL-2.1-or-later
             : AND (LGPL-2.1-or-later WITH Linux-syscall-note) AND (Linux-OpenIB OR GPL-2.0-only) AND (Linux-OpenIB OR GPL-2.0-only OR BSD-2-Clause) AND Linux-man-pages-copyleft AND MIT AND (MIT OR GPL-2.0-only) AND (MIT
             : OR GPL-2.0-or-later) AND (MIT OR LGPL-2.1-only) AND (MPL-1.1 OR GPL-2.0-only) AND (X11 OR GPL-2.0-only) AND (X11 OR GPL-2.0-or-later) AND Zlib
Description  : This package provides kernel headers and makefiles sufficient to build modules
             : against the kernel package.

Name         : kernel-devel
Version      : 5.14.0
Release      : 427.26.1.el9_4
Architecture : x86_64
Size         : 25 M
Source       : kernel-5.14.0-427.26.1.el9_4.src.rpm
Repository   : rhel-9-for-x86_64-appstream-rpms
Summary      : Development package for building kernel modules to match the kernel
URL          : https://www.kernel.org/
License      : ((GPL-2.0-only WITH Linux-syscall-note) OR BSD-2-Clause) AND ((GPL-2.0-only WITH Linux-syscall-note) OR BSD-3-Clause) AND ((GPL-2.0-only WITH Linux-syscall-note) OR CDDL-1.0) AND ((GPL-2.0-only WITH
             : Linux-syscall-note) OR Linux-OpenIB) AND ((GPL-2.0-only WITH Linux-syscall-note) OR MIT) AND ((GPL-2.0-or-later WITH Linux-syscall-note) OR BSD-3-Clause) AND ((GPL-2.0-or-later WITH Linux-syscall-note) OR
             : MIT) AND Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND BSD-3-Clause-Clear AND GFDL-1.1-no-invariants-or-later AND GPL-1.0-or-later AND (GPL-1.0-or-later OR BSD-3-Clause) AND (GPL-1.0-or-later WITH
             : Linux-syscall-note) AND GPL-2.0-only AND (GPL-2.0-only OR Apache-2.0) AND (GPL-2.0-only OR BSD-2-Clause) AND (GPL-2.0-only OR BSD-3-Clause) AND (GPL-2.0-only OR CDDL-1.0) AND (GPL-2.0-only OR
             : GFDL-1.1-no-invariants-or-later) AND (GPL-2.0-only OR GFDL-1.2-no-invariants-only) AND (GPL-2.0-only WITH Linux-syscall-note) AND GPL-2.0-or-later AND (GPL-2.0-or-later OR BSD-2-Clause) AND
             : (GPL-2.0-or-later OR BSD-3-Clause) AND (GPL-2.0-or-later OR CC-BY-4.0) AND (GPL-2.0-or-later WITH GCC-exception-2.0) AND (GPL-2.0-or-later WITH Linux-syscall-note) AND ISC AND LGPL-2.0-or-later AND
             : (LGPL-2.0-or-later OR BSD-2-Clause) AND (LGPL-2.0-or-later WITH Linux-syscall-note) AND LGPL-2.1-only AND (LGPL-2.1-only OR BSD-2-Clause) AND (LGPL-2.1-only WITH Linux-syscall-note) AND LGPL-2.1-or-later
             : AND (LGPL-2.1-or-later WITH Linux-syscall-note) AND (Linux-OpenIB OR GPL-2.0-only) AND (Linux-OpenIB OR GPL-2.0-only OR BSD-2-Clause) AND Linux-man-pages-copyleft AND MIT AND (MIT OR GPL-2.0-only) AND (MIT
             : OR GPL-2.0-or-later) AND (MIT OR LGPL-2.1-only) AND (MPL-1.1 OR GPL-2.0-only) AND (X11 OR GPL-2.0-only) AND (X11 OR GPL-2.0-or-later) AND Zlib
Description  : This package provides kernel headers and makefiles sufficient to build modules
             : against the kernel package.

```
then add repo


```
[rhel-9-for-x86_64-appstream-rpms]
name = Red Hat Enterprise Linux 9 for AMD 64 - AppStream (RPMs)
baseurl = https://cdn.redhat.com/content/dist/rhel9/$releasever/x86_64/appstream/os
enabled = 1
gpgcheck = 1
gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
sslverify = 1
sslcacert = /etc/rhsm/ca/redhat-uep.pem
sslclientkey = /etc/pki/entitlement/1217013518381510732-key.pem
sslclientcert = /etc/pki/entitlement/1217013518381510732.pem
sslverifystatus = 1
metadata_expire = 86400
enabled_metadata = 1
```

and then:
```

[root@c86479bb8ccf yum.repos.d]# dnf info kernel-devel
Updating Subscription Management repositories.
Last metadata expiration check: 0:22:48 ago on Mon Jul 22 17:00:55 2024.
Available Packages
Name         : kernel-devel
Version      : 5.14.0
Release      : 427.26.1.el9_4
Architecture : aarch64
Size         : 24 M
Source       : kernel-5.14.0-427.26.1.el9_4.src.rpm
Repository   : rhel-9-for-aarch64-appstream-rpms
Summary      : Development package for building kernel modules to match the kernel
URL          : https://www.kernel.org/
License      : ((GPL-2.0-only WITH Linux-syscall-note) OR BSD-2-Clause) AND ((GPL-2.0-only WITH Linux-syscall-note) OR BSD-3-Clause) AND ((GPL-2.0-only WITH Linux-syscall-note) OR CDDL-1.0) AND ((GPL-2.0-only WITH
             : Linux-syscall-note) OR Linux-OpenIB) AND ((GPL-2.0-only WITH Linux-syscall-note) OR MIT) AND ((GPL-2.0-or-later WITH Linux-syscall-note) OR BSD-3-Clause) AND ((GPL-2.0-or-later WITH Linux-syscall-note) OR
             : MIT) AND Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND BSD-3-Clause-Clear AND GFDL-1.1-no-invariants-or-later AND GPL-1.0-or-later AND (GPL-1.0-or-later OR BSD-3-Clause) AND (GPL-1.0-or-later WITH
             : Linux-syscall-note) AND GPL-2.0-only AND (GPL-2.0-only OR Apache-2.0) AND (GPL-2.0-only OR BSD-2-Clause) AND (GPL-2.0-only OR BSD-3-Clause) AND (GPL-2.0-only OR CDDL-1.0) AND (GPL-2.0-only OR
             : GFDL-1.1-no-invariants-or-later) AND (GPL-2.0-only OR GFDL-1.2-no-invariants-only) AND (GPL-2.0-only WITH Linux-syscall-note) AND GPL-2.0-or-later AND (GPL-2.0-or-later OR BSD-2-Clause) AND
             : (GPL-2.0-or-later OR BSD-3-Clause) AND (GPL-2.0-or-later OR CC-BY-4.0) AND (GPL-2.0-or-later WITH GCC-exception-2.0) AND (GPL-2.0-or-later WITH Linux-syscall-note) AND ISC AND LGPL-2.0-or-later AND
             : (LGPL-2.0-or-later OR BSD-2-Clause) AND (LGPL-2.0-or-later WITH Linux-syscall-note) AND LGPL-2.1-only AND (LGPL-2.1-only OR BSD-2-Clause) AND (LGPL-2.1-only WITH Linux-syscall-note) AND LGPL-2.1-or-later
             : AND (LGPL-2.1-or-later WITH Linux-syscall-note) AND (Linux-OpenIB OR GPL-2.0-only) AND (Linux-OpenIB OR GPL-2.0-only OR BSD-2-Clause) AND Linux-man-pages-copyleft AND MIT AND (MIT OR GPL-2.0-only) AND (MIT
             : OR GPL-2.0-or-later) AND (MIT OR LGPL-2.1-only) AND (MPL-1.1 OR GPL-2.0-only) AND (X11 OR GPL-2.0-only) AND (X11 OR GPL-2.0-or-later) AND Zlib
Description  : This package provides kernel headers and makefiles sufficient to build modules
             : against the kernel package.

Name         : kernel-devel
Version      : 5.14.0
Release      : 427.26.1.el9_4
Architecture : x86_64
Size         : 25 M
Source       : kernel-5.14.0-427.26.1.el9_4.src.rpm
Repository   : rhel-9-for-x86_64-appstream-rpms
Summary      : Development package for building kernel modules to match the kernel
URL          : https://www.kernel.org/
License      : ((GPL-2.0-only WITH Linux-syscall-note) OR BSD-2-Clause) AND ((GPL-2.0-only WITH Linux-syscall-note) OR BSD-3-Clause) AND ((GPL-2.0-only WITH Linux-syscall-note) OR CDDL-1.0) AND ((GPL-2.0-only WITH
             : Linux-syscall-note) OR Linux-OpenIB) AND ((GPL-2.0-only WITH Linux-syscall-note) OR MIT) AND ((GPL-2.0-or-later WITH Linux-syscall-note) OR BSD-3-Clause) AND ((GPL-2.0-or-later WITH Linux-syscall-note) OR
             : MIT) AND Apache-2.0 AND BSD-2-Clause AND BSD-3-Clause AND BSD-3-Clause-Clear AND GFDL-1.1-no-invariants-or-later AND GPL-1.0-or-later AND (GPL-1.0-or-later OR BSD-3-Clause) AND (GPL-1.0-or-later WITH
             : Linux-syscall-note) AND GPL-2.0-only AND (GPL-2.0-only OR Apache-2.0) AND (GPL-2.0-only OR BSD-2-Clause) AND (GPL-2.0-only OR BSD-3-Clause) AND (GPL-2.0-only OR CDDL-1.0) AND (GPL-2.0-only OR
             : GFDL-1.1-no-invariants-or-later) AND (GPL-2.0-only OR GFDL-1.2-no-invariants-only) AND (GPL-2.0-only WITH Linux-syscall-note) AND GPL-2.0-or-later AND (GPL-2.0-or-later OR BSD-2-Clause) AND
             : (GPL-2.0-or-later OR BSD-3-Clause) AND (GPL-2.0-or-later OR CC-BY-4.0) AND (GPL-2.0-or-later WITH GCC-exception-2.0) AND (GPL-2.0-or-later WITH Linux-syscall-note) AND ISC AND LGPL-2.0-or-later AND
             : (LGPL-2.0-or-later OR BSD-2-Clause) AND (LGPL-2.0-or-later WITH Linux-syscall-note) AND LGPL-2.1-only AND (LGPL-2.1-only OR BSD-2-Clause) AND (LGPL-2.1-only WITH Linux-syscall-note) AND LGPL-2.1-or-later
             : AND (LGPL-2.1-or-later WITH Linux-syscall-note) AND (Linux-OpenIB OR GPL-2.0-only) AND (Linux-OpenIB OR GPL-2.0-only OR BSD-2-Clause) AND Linux-man-pages-copyleft AND MIT AND (MIT OR GPL-2.0-only) AND (MIT
             : OR GPL-2.0-or-later) AND (MIT OR LGPL-2.1-only) AND (MPL-1.1 OR GPL-2.0-only) AND (X11 OR GPL-2.0-only) AND (X11 OR GPL-2.0-or-later) AND Zlib
Description  : This package provides kernel headers and makefiles sufficient to build modules
             : against the kernel package.
```

**Review by ben-alkov — CHANGES_REQUESTED** [review]
Commit message?

**Review by ben-alkov — CHANGES_REQUESTED** [review]
Commit message.

**Review by ben-alkov — CHANGES_REQUESTED** [review]
Please rebase properly.

File: requirements.txt
Comment (ben-alkov) [inline requirements.txt:1]: Why?

**Comment by ben-alkov** [comment]
@brianwcook; Please mark this as a draft PR

**Comment by ben-alkov** [comment]
@brianwcook; Your Dockerfile doesn't seem to build a usable image https://github.com/containerbuildsystem/cachi2/actions/runs/10040853225/job/27804698612?pr=580

**Comment by brianwcook** [comment]
Latest commit enables prefetch for content from dnf repositories that require client certificate authorization. Together with [this task](https://github.com/brianwcook/build-definitions/blob/STONEBLD-2618/task/prefetch-dependencies/0.1/prefetch-dependencies.yaml) the authenticated content can be fetched with cachi2 in a Konflux pipeline.

**Comment by chmeliik** [comment]
> @brianwcook would you mind splitting this into two PRs?

Or, before that, we should get confirmation from the DNF folk that their lockfile support will be able to support subscription-manager for multi-arch lockfiles. We _are_ treating the RPM support as tech-preview, but still I wouldn't like to pull this feature from under users' feet when the DNF lockfile support lands and we switch to it.

**Comment by eskultety** [comment]
> > @brianwcook would you mind splitting this into two PRs?
> 
> Or, before that, we should get confirmation from the DNF folk that their lockfile support will be able to support subscription-manager for multi-arch lockfiles. 

Firstly, I don't think ^this work has been officially announced anywhere, has it?
Secondly, how are subman and DNF related in particular? The repos subscription-manager enables are (in simple terms) based on the platform product tags (e.g. `rhel-8-server`) and the base architecture which can be tweaked with the means of facts files under `/etc/rhsm/facts` where one can "enable" for more arches for which corresponding entries in the resulting `.repo` file are generated. That should be pretty much the connection between the subscription manager and DNF as far as lockfile goes, since lockfiles can already host multiple architectures.

The problems I see are:

- this only works based on the assumption we keep using direct URLs for download and not DNF
- the repos that subman enables are matched according to product tags and so while multi-arch is okay, multi-product is likely not supported, IOW cachi2 would have to ship multiple images, since content specifically targeting RHEL-8 won't be enabled on UBI-9 (I haven't gotten a reply on that particularly) and so switching to direct DNF usage will choke on this
- messing with `/etc/rhsm` (as done in the related quoted PR) requires a privileged user which we ideally should not enforce in general; even `subscription-manager register` seems to require a privileged user

**Comment by brianwcook** [comment]
`since content specifically targeting RHEL-8 won't be enabled on UBI-9` I think this is possible as well, but honestly the direct download method seems much more straightforward to me and is likely more performant as it appears dnf serializes content downloads.

`messing with /etc/rhsm` does require a privileged user. For UBI9 we can `export SMDEV_CONTAINER_OFF=1` instead, but it does not work on UBI8.

`subscription-manager register and even `dnf install` also require user 0.  if you want to register and install RPMs you just need user 0, but the way I separated it you don't need to run cachi2 as root.

File: cachi2/core/package_managers/rpm/main.py [hunk 1]
```diff
@@ -378,3 +380,37 @@ def _generate_repofiles(
...
+    client_cert = getenv("C2_CLIENT_CERT")
+    client_key = getenv("C2_CLIENT_KEY")
+    ca_bundle = getenv("C2_CA_BUNDLE")
+    
+    ssl_ctx = ssl.create_default_context()
+
+    if client_cert is None or client_key is None:
+        log.info(f"No client certificates will be used.")
+    elif not path.isfile(path=client_cert) or not path.isfile(path=client_key) :
+        raise(FileNotFoundError)
```
eskultety: This is not a good idea because in case of an error this crashes with a traceback. You need a `Cachi2Error` subclass error here for the code to end with a user friendly error, you'll most likely be interested in the `FetchError`.

eskultety (commented):
After having a deeper look at this PR I'm not convinced the PR itself should advertise this as authenticated access to RPM entitlement repos. The way it's handled is generic SSL stuff that is IMO worth having regardless. Sure, authenticated access to RPMs is the main driver here, but the solution is by no means tied to it - see my comments on ultimately exposing this functionality on the CLI rather than env variables. I've had quite a few comments so I left the tests out of this round of reviews.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -378,3 +380,37 @@ def _generate_repofiles(
 
             with open(repo_file_path, "w") as f:
                 repofile.write(f)
+
+def _get_ssl_context():
```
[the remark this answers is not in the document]
brianwcook: I added it to the RPM package manager because in the case that someone adds certs, this way they only get sent to RPM repo endpoints. You _could_ add it at a higher level but I cannot predict what would happen if you send a client cert to other endpoints. Some endpoint will ignore it, some will fail. This could break prefetch when someone is pulling content of multiple types at the same time.

cachi2/core/package_managers/general.py
```diff
@@ -45,7 +46,8 @@ def download_binary_file(
     timeout = get_config().requests_timeout
     try:
         resp = pkg_requests_session.get(
-            url, stream=True, verify=not insecure, auth=auth, timeout=timeout
+            # todo verify needs to be : not insecure
```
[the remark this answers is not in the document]
brianwcook: note to myself, it will go away

cachi2/core/package_managers/rpm/main.py
```diff
@@ -378,3 +380,37 @@ def _generate_repofiles(
...
+    client_cert = getenv("C2_CLIENT_CERT")
+    client_key = getenv("C2_CLIENT_KEY")
+    ca_bundle = getenv("C2_CA_BUNDLE")
```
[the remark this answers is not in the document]
brianwcook: I'd say the variable names are 'made up with inspiration from the SSL module' (for example it uses the CA 'bundle' terminology'. I support moving the env variables to CLI, I just hadn't figured out how yet but I have now.

For encrypted private keys are you expecting a user to enter that password interactively?

cachi2/core/package_managers/general.py
```diff
@@ -45,7 +46,8 @@ def download_binary_file(
...
+            url, stream=True, verify=not insecure, auth=auth, timeout=timeout, cert=get_client_certs()
```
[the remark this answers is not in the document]
brianwcook: It isn't implemented yet but it was from thinking about implementing cli options (and needing to get the client cert names from the caller instead of from the env.  I also don't know what 'download_binary_file' is for - it seems to be unused when I fetch RPMs (at least nothing fails when authentication is not configured there).

cachi2/core/package_managers/rpm/main.py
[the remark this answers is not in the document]
eskultety: First of all, sorry for a late reply.

>  I support moving the env variables to CLI, I just hadn't figured out how yet but I have now.

This would ideally be done via extra options in the input JSON: https://github.com/containerbuildsystem/cachi2/blob/1be1ab417de0a1bc43db386889e5023ebd05983f/cachi2/core/models/input.py#L173 because this way it's only ever viable for a given package manager, not globally like it is for Golang. It'll need to be extended appropriately.
Even though it's generic enough for other package manager's use, that feature hasn't been enabled globally, so I'd be willing to accept tweaks only to the RPM related model and expand it to other ones in the future if needed.

> For encrypted private keys are you expecting a user to enter that password interactively?

No, the standard way to go about that is password files.

cachi2/core/package_managers/rpm/main.py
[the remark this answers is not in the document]
eskultety: If you add it as RPM only options in the JSON input as hinted in another reply then you don't need to worry about that because for other package managers those would not be used as the environment would be pristine again.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -378,3 +380,37 @@ def _generate_repofiles(
...
+    # verify_mode is for client verifying the servers cert
+    # options:
+    # ssl_ctx.verify_mode = ssl.CERT_REQUIRED  - default
+    # ssl_ctx.verify_mode = ssl.CERT_NONE - allow self signed or expired certs
+    
+    ssl_verify = getenv("C2_SSL_VERIFY", "CERT_REQUIRED")
+    if ssl_verify.lower() == "false":
+        log.info(f"Disabling SSL certificate verification. This is insecure and should not be used except for testing.")
+        ssl_ctx.check_hostname = False # required for verify_mode = ssl.CERT_NONE
+        ssl_ctx.verify_mode = ssl.CERT_NONE
```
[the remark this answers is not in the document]
brianwcook: having a choice is standard for nearly all tools; curl --insecure for example. I think it is really valuable for testing.

cachi2/core/package_managers/rpm/main.py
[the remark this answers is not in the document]
brianwcook: the cli options I've added are pretty standard / generic to SSL and cllient TLS. I think anyone with some experience will recognize them.

tests/integration/test_rpm.py
```diff
@@ -108,6 +111,56 @@
...
+def test_cert_auth(cachi2_image: utils.ContainerImage,
+                   tmp_path: Path,
+                   test_data_dir: Path,
+                   request: pytest.FixtureRequest) -> None:
+
+    client_cert = "/certificates/testuser.crt"
+    client_key = "/certificates/testuser.key"
+    ca_bundle = "/certificates/myCA.crt"
+
+    print(client_cert)
```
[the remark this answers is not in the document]
brianwcook: @eskultety this will need an override; it is necessary for the integration test and these are not sensitive certs, but ones generated specifically for this test.

cachi2/core/models/input.py
```diff
@@ -77,7 +77,42 @@ def _path_is_relative(cls, path: Path) -> Path:
...
+
+        return data
+
+
+class _SSLOptions(pydantic.BaseModel, extra="allow"):
+    """SSL options model."""
+
+    ssl: _SSLFields
+
+
```
eskultety, line 114: I just posted #651 which should make your life easier when it comes to adding a new options model, you'll only need `_SSLFields` (renamed to `_SSLOptions`), but not both as it's confusing.

eskultety (commented):
First round of comments. I still need to dissect the tests, so more comments should be expected, but for now I think there's enough to address.

brianwcook:
@eskultety Thanks for your review. I believe I have addressed all of your comments, however somehow your interim commit dropping _DNFOptions has become linked to my PR in a way that I can't seem to undo. I've tried to rebase as many ways as I know how to no avail... advice welcome.

brianwcook:
I ended up having to clone repo over again and apply a patch to get it to look right. The integration test is being blocked by Dockerhub rate limiting now though.

tests/dnfserver/container/generate-certs.sh
```diff
@@ -0,0 +1,28 @@
...
+mkdir -p certificates
+cd certificates
+rm *
+
+
+# todo - extend lifespan of all certs to 10 years
+
+# generate CA cert
+openssl genrsa   -out myCA.key 4096
+openssl req -new -x509  -key myCA.key -days 3650 -out myCA.crt \
```
eskultety, line 11: Can we use something like 36500 just to be safe we won't ever need to regenerate just because of validity :) ? Do we need to track this file though? You've clearly used it to generate certificates, but is it something we're going to be doing on a regular basis?

tests/dnfserver/container/Dockerfile
```diff
@@ -0,0 +1,9 @@
+
+FROM registry.fedoraproject.org/fedora:40
+RUN dnf -y install openssl nginx nano procps-ng python3-pip
+COPY certificates /certificates
+COPY setup.sh .
+RUN ./setup.sh
+COPY nginx.conf /etc/nginx/conf.d/
+EXPOSE 8443 8080 8081
+ENTRYPOINT /usr/sbin/nginx -g 'daemon off;'
```
eskultety, line 9: I'm not convinced we need this at all.
Instead, I think you should use the upstream `nginx:alpine-slim` image (~12MB) and just mount a bunch of files into it read-only.
You're proposing an nginx config and certificates, so the only bit missing is creating a DNF repo out of some RPM files. I don't think we want to build a new image based on Fedora (not even minimal) just to do this based on the fact that `createrepo_c` is already cachi2's direct dependency, so it'll be always present in the test environment. So I imagine the RPM will be downloaded in the `start.sh` script in the `setup()` phase just like the pypiserver does it and then all the cert files, config and the dnf repo would be mounted to the contianer.

tests/integration/test_rpm.py
```diff
@@ -265,3 +303,18 @@ def test_e2e_rpm(
...
+
+
+@pytest.mark.skipif(
+    os.getenv("CACHI2_TEST_LOCAL_DNF_SERVER") != "true", reason="CACHI2_TEST_LOCAL_DNF_SERVER!=true"
+)
+def test_nginx_config() -> None:
+    """
+    Ensures the local nginx server is properly configured to require client certificate authentication.
+
+    This test is important. If nginx is not properly configured rpm_with_cert_auth test might return a false positive.
```
eskultety, line 315: Is it important? We should have a SSL health check right as the container starts to make sure it is properly configured. This test isn't justified in my mind.

tests/dnfserver/container/nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+    # if you need to debug
+    error_log /var/log/nginx/error.log debug;
+
+
+    error_page 404 /404.html;
+        location = /40x.html {
+    }
+
+    error_page 500 502 503 504 /50x.html;
+        location = /50x.html {
```
eskultety, line 68: Is this one just for debugging without SSL?

tests/dnfserver/container/README.md
```diff
@@ -0,0 +1,19 @@
+Running the nginx container
+podman run -p 8443:8443 -p 8080:8080 quay.io/bcook/cachi2-nginx
```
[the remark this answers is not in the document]
eskultety: I don't think your own image should be used here, instead we should instruct users to go via the `start.sh` script to test this interactively.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -462,3 +480,50 @@ def _generate_repofiles(
...
+    else:
+        raise SSLContextError(
+            f"Specified client credentials files:\n \
+              ['{client_cert}']\n \
+              ['{client_key}']\n \
+              could not be opened.",
+            solution=("Ensure the files exist and that you have permission to the files."),
+        )
+
+    if ca_bundle is not None:
```
eskultety, line 512: Seems like we're checking multiple files if the path exists repeatedly, how about:
```
for ssl_file in [ca_bundle, client_cert, client_key]:
  if ssl_file is not None  and not Path(ssl_file).isfile():
    raise SSLContextError(...)
```

Then later in the code you'd only need
```
if ca_bundle is not None:
  ssl_ctx.load_verify_locations(ca_bundle)
```

cachi2/core/package_managers/rpm/main.py
```diff
@@ -320,7 +332,13 @@ def _download(lockfile: RedhatRpmsLock, output_dir: Path) -> dict[Path, Any]:
             }
             Path.mkdir(dest.parent, parents=True, exist_ok=True)
 
-        asyncio.run(async_download_files(files, get_config().concurrency_limit))
+        asyncio.run(
+            async_download_files(
+                files,
+                get_config().concurrency_limit,
+                ssl_context=_get_ssl_context(options=options.model_dump() if options else None),
```
eskultety, line 339: This should be `_get_ssl_context(options=options.model_dump()) if options else None` (mind the parenthesis after `dump`)

cachi2/core/package_managers/rpm/main.py
```diff
@@ -462,3 +480,50 @@ def _generate_repofiles(
...
                 repofile.write(f)
+
+
+def _get_ssl_context(options: Optional[dict] = None) -> ssl.SSLContext:
+
+    ssl_ctx = ssl.create_default_context()
+
+    if options is None:
+        log.debug("returning default ssl context.")
+        return ssl_ctx
```
eskultety, line 491: We always want options set in this function.

cachi2/core/errors.py
```diff
@@ -130,3 +130,23 @@ def __init__(
...
+    def __init__(self, reason: str, *, solution: Optional[str], docs: Optional[str] = None) -> None:
+        """Initialize a Package Rejected error.
+
+        Compared to the parent class, the solution param is required (but can be explicitly None).
+
+        :param reason: explain why we rejected the package
+        :param solution: politely suggest a potential solution to the user
+        :param docs: include a link to relevant documentation (if there is any)
+        """
+        super().__init__(reason, solution=solution, docs=docs)
```
eskultety, line 152: Do we really need a new exception for this? `InputError` is already in place and you can provide a `solution` there, nothing is stopping you from doing that :).

Makefile
```diff
@@ -26,6 +27,7 @@ test-integration: venv
 	CACHI2_GENERATE_TEST_DATA=$(GENERATE_TEST_DATA) \
 	CACHI2_TEST_LOCAL_PYPISERVER=$(TEST_LOCAL_PYPISERVER) \
 		venv/bin/tox -e integration -- $(TOX_ARGS)
+	CACHI2_TEST_LOCAL_DNF_SERVER=$(TEST_LOCAL_DNF_SERVER)
```
eskultety, line 30: You'll need to move this variable one line above, this won't work as expected otherwise ;)

cachi2/core/package_managers/rpm/main.py
```diff
@@ -216,7 +218,8 @@ def fetch_rpm_source(request: Request) -> RequestOutput:
         # We're deliberately taking the easy route here by only assuming the "last" set of options
         # we found in the input JSON instead of doing a deep merge of all the nested dicts.
         # Nevertheless, we'll at least emit a warning at the end so that the user is informed
-        if package.options and package.options.dnf:
+
+        if package.options and package.options.dnf:  # type: ignore
```
eskultety, line 222: Does mypy still complain? I'm not seeing any errors, so the type ignore should likely be dropped.

tests/dnfserver/start.sh
```diff
@@ -0,0 +1,27 @@
...
+
+setup() {
+    mkdir "$WORKDIR/certs"
+    echo -e "\n--- Starting dnfserver on http://localhost:${DNFSERVER_PORT:-8443} ---\n"
+}
+
+setup >&2
+
+podman run --rm --replace --name cachi2-dnfserver \
+    -p 8081:8081 \
```
eskultety, line 25: We don't need base HTTP for testing SSL right?

eskultety (commented):
Please split the changes into multiple commits in your next revision, it's extremely exhausting to review this in a single commit, we wouldn't accept this as a sigle patch anyway, I just didn't want to push that hard since you've struggled with rebasing, but I'm finding it difficult to keep context of all files changed in my head as I'm scrolling through the changes. The fact you're proposing scripts which aren't used anywhere doesn't help either. I'll expect that in the next round of reviews.

tests/dnfserver/start.sh
brianwcook: it is used for aliveness check (without certs).

cachi2/core/errors.py
brianwcook: happy to remove it, i thought it would be viewed as not specific enough

tests/dnfserver/container/nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+    }
```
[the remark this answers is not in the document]
brianwcook: I can drop it if we are ok committing the certs to the repo. Otherwise it is serving those certs so they can be fetched from the container.

eskultety:
> Please split the changes into multiple commits in your next revision, it's extremely exhausting to review this in a single commit, we wouldn't accept this as a sigle patch anyway, I just didn't want to push that hard since you've struggled with rebasing, but I'm finding it difficult to keep context of all files changed in my head as I'm scrolling through the changes. The fact you're proposing scripts which aren't used anywhere doesn't help either. I'll expect that in the next round of reviews.

I forgot to be more specific on how this should be split:

1. SSL options model introduction + unit test
2. package_managers/rpm changes + unit test
3. integration test changes

tests/dnfserver/start.sh
eskultety: > it is used for aliveness check (without certs).

If you want continuous check like this then that's what `--health-cmd` in podman is for, from outside it's not necessary to be exposed like this.

tests/dnfserver/container/nginx.conf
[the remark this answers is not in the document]
eskultety: I don't mind checking certs in the repo, provided:
- it's all base64 encoded data
- the header is generic enough not to be affiliated to any entity
- have essentially unlimited validity

I would even be okay with us generating the set of certs on demand with each CI run and using less entropy to decrease the performance penalty, it'd we fine in local executions because this integration test would not run very often, but it might potentially have an impact on the GH CI runners - these VMs are provisioned on demand and we don't know anything about them, e.g. what entropy source they use (i.e. a virtualized PRNG, or paravirtualized source from the host) and it might slow the execution down, I don't know, so I tracking test certs in the repo is fine with me.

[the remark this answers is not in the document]
eskultety: That said, we can leave out CSR, SRL and CA private key, the certs won't be maintained in any way and if a change is needed, they'll be regenerated from scratch.

tests/dnfserver/container/setup.sh
```diff
@@ -0,0 +1,28 @@
...
+mkdir "/etc/pki/nginx"
+chown -R nginx:nginx "/etc/pki/nginx"
+chmod 700 "/etc/pki/nginx"
+
+cp /certificates/localhost.crt "/etc/pki/nginx/server.crt"
+cp /certificates/localhost.key "/etc/pki/nginx/server.key"
+cp /certificates/myCA.crt "/etc/pki/myCA.crt"
+
+# create yum repo
+pip3 install createrepo_c
```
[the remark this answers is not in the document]
brianwcook: it's there so that the lock file can be regenerated if needed from the dnfserver.

tests/dnfserver/container/setup.sh
```diff
@@ -0,0 +1,28 @@
...
+mkdir -p /dnfrepo/pkg
+mkdir /dnfrepo/source
+echo "I'm alive!" > /dnfrepo/index.html
+chmod 700 "/dnfrepo"
+cd /dnfrepo/pkg
+curl -O https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/x86_64/baseos/os/Packages/r/redhat-release-9.4-0.5.el9.x86_64.rpm
+createrepo_c .
+cd /dnfrepo/source
+curl -O https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/x86_64/baseos/source/SRPMS/Packages/r/redhat-release-9.4-0.5.el9.src.rpm
+createrepo_c .
```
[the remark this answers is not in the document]
brianwcook: cachi2 complains if the source rpms are not present.

tests/dnfserver/container/test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# verify server and certs are configured correctly.
+
+curl \
+  --cacert /CertificateAuthCA/myCA.crt \
+  --key /CertificateAuthCA/testuser.key \
+  --cert /CertificateAuthCA/testuser.crt \
+  https://localhost
```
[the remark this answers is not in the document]
brianwcook: it's just helpful to know how to test the container if you need to make a change for some reason (and to not have to type that over and over). That is what I used it for.

tests/dnfserver/container/nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+    ssl_certificate_key "/etc/pki/nginx/server.key";
+
+    # enable client cert verification
+    ssl_client_certificate /etc/pki/myCA.crt; ## Use your own trusted certificate from CA/SSLTrust
+    ssl_verify_client on;
+
+    ssl_session_cache shared:SSL:1m;
+    ssl_session_timeout  10m;
+    ssl_ciphers PROFILE=SYSTEM;
+    ssl_prefer_server_ciphers on;
```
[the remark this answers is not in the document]
brianwcook: its the default config.

tests/dnfserver/container/nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+    ssl_certificate "/etc/pki/nginx/server.crt";
+    ssl_certificate_key "/etc/pki/nginx/server.key";
+
+    # enable client cert verification
+    ssl_client_certificate /etc/pki/myCA.crt; ## Use your own trusted certificate from CA/SSLTrust
+    ssl_verify_client on;
+
+    ssl_session_cache shared:SSL:1m;
+    ssl_session_timeout  10m;
+    ssl_ciphers PROFILE=SYSTEM;
```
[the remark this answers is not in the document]
brianwcook: I don't know, changing these or investigating them at all was / is irrelevant to me, I used the default config provided, it works.

tests/dnfserver/container/nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+    root         /dnfrepo;
+
+    ssl_certificate "/etc/pki/nginx/server.crt";
+    ssl_certificate_key "/etc/pki/nginx/server.key";
+
+    # enable client cert verification
+    ssl_client_certificate /etc/pki/myCA.crt; ## Use your own trusted certificate from CA/SSLTrust
+    ssl_verify_client on;
+
+    ssl_session_cache shared:SSL:1m;
```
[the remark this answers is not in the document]
brianwcook: also default config.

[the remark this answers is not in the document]
brianwcook: removed

tests/dnfserver/container/nginx.conf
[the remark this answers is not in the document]
brianwcook: removed

tests/dnfserver/container/nginx.conf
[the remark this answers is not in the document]
brianwcook: removed

tests/dnfserver/container/generate-certs.sh
brianwcook: no idea how often we would use it, IMO tracking files is free and if anyone ever needed to regenerate the certs this makes sure it is fast.

brianwcook: changed to 36500 though

tests/dnfserver/container/Dockerfile
brianwcook: I didn't choose Fedora because of createrepo_c (although it is convenient). I don't recall the exact detail but it was about needing the newer version of nginx which is in Fedora. I think it could be moved to ubi10 when that is out and get same version.

Personally I still think being able to run rpm_lockfile_prototype or whatever replaces it against the container itself is ideal.  If you worry about things like the size of container images (I don't very much) then I guess we could use alpine. I prefer Fedora because I more or less know how it works, I know who can fix it when it doesn't work and it is easy to install other packages or tools when the need arises.

Example: in 2025 we probably want to test this with quantum safe certificates - I know openssl and nginx will be updated appropriately here because I am on the project. I don't know anything about Alpine and I'm not eager to touch something again which is working.

tests/integration/test_rpm.py
brianwcook: It says it is important right in the docstring so it must be important, no?

This tests that the result of a request to the protected endpoint _without_ certificates returns a 400 as it should. If we don't test this and somehow the nginx certificates auth is misconfigured, you might get a false 'pass' from the test that _does_ send certificates.

tests/dnfserver/container/nginx.conf
brianwcook: yes, exactly.

cachi2/core/package_managers/rpm/main.py
brianwcook: definitelye should not be using an image from my quay repo, but I don't have the necessary permissions to put it anywhere else and I assume we need to get these reviews cleared before someone from the cachi2 team is going to push it for me.  And, I had to put it somewhere...

There is also test data in git (rpm lockfile) that needs to be moved to the repo where the other rpm test data is.

cachi2/core/errors.py
brianwcook: I'm assuming you are referring to `InvalidInput` and making the change

tests/dnfserver/container/nginx.conf
```diff
@@ -0,0 +1,70 @@
...
+    ssl_session_timeout  10m;
```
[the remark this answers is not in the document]
brianwcook: removed

tests/dnfserver/start.sh
brianwcook: I'm doing it [exactly the same way that the Pypi server is checked](https://github.com/brianwcook/cachi2/blob/3c562a5410ddd5f1043e7613b240bb5811682f7f/tests/integration/conftest.py#L69) in the equivalent test.

cachi2/core/errors.py
eskultety: Yes.

tests/dnfserver/container/setup.sh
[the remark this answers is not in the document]
eskultety: > it's there so that the lock file can be regenerated if needed from the dnfserver.

What use case would that be useful for in context of integration tests?

tests/dnfserver/container/test.sh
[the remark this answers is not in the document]
eskultety: You didn't document this script anywhere, so arguably nobody would even know this exists until you point this out, so right now, it were only useful to you :) . Anyway, this is a documentation material, so this docs information should go probably to the Bash script that starts the container, but I stand by my earlier comment that the code logic is better suited in the `local_dnfserver`  pytest fixture than a stanlone script nobody knows about.

tests/dnfserver/container/nginx.conf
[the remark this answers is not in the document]
eskultety: > I don't know, changing these or investigating them at all was / is irrelevant to me, I used the default config provided, it works.

Yes, but that was likely the default config from Fedora which will contain a lot of distro-only tweaks, what I'm after is the tiniest set of relevant options so the next person who looks at the config if needed doesn't need to study why the options are there and which ones matter, so it is very relevant, maybe not to you, but the person who looks at it next.

tests/dnfserver/container/Dockerfile
eskultety: >  I don't recall the exact detail but it was about needing the newer version of nginx which is in Fedora

Newer than the official nginx docker image (1.26/1.27)?

> Personally I still think being able to run rpm_lockfile_prototype or whatever replaces it against the container itself is ideal

How is ^this relevant to the integration test suite?

> I prefer Fedora because I more or less know how it works, I know who can fix it when it doesn't work and it is easy to install other packages or tools when the need arises.
...
I don't know anything about Alpine and I'm not eager to touch something again which is working.

The thing about application containers is that you shouldn't need to, what platform nginx is running is irrelevant to the cause, because as a consumer of the image you only care about the application itself unless you need to install extra packages (which is where you might start caring) which you likely don't need here. As long as you have an nginx config and a volume for data, I don't see how quantum safe certificates are an issue here, because nginx doesn't care as long as it can find them in some filesystem location it can read. Therefore I'm not convinced any of what you mentioned is a real problem that would justify our integration test suite to build a dedicated container image for its own use instead of pulling official image builds suited for the workload.

The downside of stuffing all the logic revolving around `createrepo_c` to the container build time as you currently have is that you have limited visibility into it and capabilities to debug the process of creation of the data during the build. Conversely, if you mount all of the data merely as a data volume to an nginx container, you can both issue breakpoints() to the relevant Python code that generates/handles them or inspect it from the host easily (i.e. you don't need to spawn another interactive shell session in the container).

tests/dnfserver/start.sh
eskultety: Which is a good starting point. With the exception that the nginx service you're adding requires more configuration so I see value in some (pretty much any) automatic kind of connection checking based on the SSL certs.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -203,8 +204,9 @@ def fetch_rpm_source(request: Request) -> RequestOutput:
     noptions = 0
 
     for package in request.rpm_packages:
+
```
eskultety, line 207: Extra new line.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -216,6 +218,7 @@ def fetch_rpm_source(request: Request) -> RequestOutput:
         # We're deliberately taking the easy route here by only assuming the "last" set of options
         # we found in the input JSON instead of doing a deep merge of all the nested dicts.
         # Nevertheless, we'll at least emit a warning at the end so that the user is informed
+
```
eskultety, line 221: Extra new line.

eskultety (commented):
@brianwcook please split the commits as suggested in my earlier comment as well as addressing outstanding comments (which don't have an ongoing discussion) so the review can continue and bring us closer to merging.

tests/dnfserver/start.sh
brianwcook: I've made the aliveness check use the ssl port with client certificates. If you want to use the podman health check, someone else can alter if for both dnfserver and pypi server later.

cachi2/core/package_managers/rpm/main.py
brianwcook: removed it

cachi2/core/package_managers/rpm/main.py
brianwcook: I had _SSLOptions set as default to _SSLOptions() before and you asked me to remove it and have it default to None.  If options.ssl can be None, it makes sense that the function can accept None.  If you don't want it to accept None, I think the most reasonable way to accomplish it is to have _ExtraOptions = _ExtraOptions() and _SSLOptions = _SSLOptions(). Then the optional goes away because the variable is always set, everywhere.

tests/dnfserver/start.sh
eskultety: >  If you want to use the podman health check, someone else can alter if for both dnfserver and pypi server later.

I'm okay with that, that was mostly just an idea I pitched, as long as we have some kind of an automated **pre-test** SSL check of the server, I'm good.

cachi2/core/package_managers/rpm/main.py
eskultety: > I had _SSLOptions set as default to _SSLOptions() before and you asked me to remove it and have it default to None. If options.ssl can be None, it makes sense that the function can accept None. If you don't want it to accept None, I think the most reasonable way to accomplish it is to have _ExtraOptions = _ExtraOptions() and _SSLOptions = _SSLOptions(). Then the optional goes away because the variable is always set, everywhere.

You updated the caller of this helper function exactly as I told you which 

> We always want options set in this function.

related to. With the recent change you made options is no longer optional in this helper and is always set, so the signature needs to be updated and the check dropped.

cachi2/core/package_managers/rpm/main.py
brunoapimentel: > When you use the official nginx docker image you don't need to worry about ^this.

But there's quite a lot of setup that needs to be done on top of the official image, right? It might be easier to just move the image setup to a repo in the [cachito-testing](https://github.com/cachito-testing/) org, we could try using Github's registry to store it, and a small Github Actions pipeline to build it.

cachi2/core/models/input.py
```diff
@@ -76,6 +76,30 @@ def _path_is_relative(cls, path: Path) -> Path:
...
+    """SSL options model.
+
+    Provides the 'ssl' key under 'options' for rpm package manager.
+
+    """
+
+    client_cert: Optional[str] = None
+    client_key: Optional[str] = None
+    ca_bundle: Optional[str] = None
+    ssl_verify: Optional[bool] = None
```
brunoapimentel, line 89: I wonder what the implications are to allow users to easily disable the SSL cert/host verification. I know it emits a warning, but it can still be easily used in production envs. I do see its usefulness for testing, though.

If we decide that this is a big security issue, my suggestion is to scrap this option altogether, and think of a way to introduce it as a follow up (maybe by adding a warning in the resulting SBOM?).

cachi2/core/package_managers/rpm/main.py
eskultety: > But there's quite a lot of setup that needs to be done on top of the official image, right?

A lot of setup justifying a custom build image hosted in cachito-testing org's registry making it harder to actually understand what's happening? I don't think so. If I enumerate what's needed:

- curl a single RPM package
    ^this is trivial
- call `createrepo_c` on the temporary directory we downloaded the RPM from previous step to
   ^also straight forward, we already have a hard dependency on `createrepo_c` so we can safely assume it's installed
- mount volumes with certs, nginx, and DNF repo to the container
  ^the only difference to a custom built image is the last volume, the rest we have to do anyway
- check the aliveness of the container
  ^we have to do this either way
- run the integration test

In summary, unless I have overlooked some important bit (could happen easily) then I don't think my proposal adds any overhead over what's being proposed at all. Quite the contrary, the setup is fully transparent, relying on a small, official docker image designated to one thing and one thing only with any modifications being achieved with a set of data volumes mounted on startup which we'd unarguably do either way.

cachi2/core/models/input.py
eskultety: > If we decide that this is a big security issue, my suggestion is to scrap this option altogether, and think of a way to introduce it as a follow up (maybe by adding a warning in the resulting SBOM?).

This is not our problem to solve, those are user passthrough options that we can't take full responsibility for. Documentation and potentially a warning is all that we can really do about this. With these extra options the likelihood users asking for more options like this in the future to cover their exotic needs is IMO fairly decent, so they need to be prepared to carry the consequences (as long as we ourselves never disable it by default).

brunoapimentel: I'd argue that we do take responsibility for some of these problems. For instance, we always require and validate checksums, this is not an user option.

I don't have strong feelings towards this, I don't think this is as grave as a checksum mismatch. If we think it is ok for this to be set in production environments, let's just keep the option here.

cachi2/core/package_managers/rpm/main.py
brunoapimentel: If it is as simple as you say, and if we can avoid rebuilding the image on every CI run, then I think it's a good way to go. 

Otherwise, offloading it to another repo would avoid the need for rebuilds, and to have all these test files in the Cachi2 repo. The sources for the test would still be one click away, so I don't think it would make it hard to understand (we can leave a reference to the test repo in a comment). The part I'm not sure about is the Github registry, as I have never used it before.

eskultety: > If it is as simple as you say, and if we can avoid rebuilding the image on every CI run, then I think it's a good way to go.

For a comparison, have a look at this commit on my branch where I did just that without the need for a custom container image: https://github.com/containerbuildsystem/cachi2/commit/a276ab62db81025dd579feba96953e0526543c58

eskultety:
@brianwcook sorry, getting back to this one after PTO. I don't see any update on the comments I made during the last round of review. In any case, have a look at https://github.com/containerbuildsystem/cachi2/compare/main...eskultety:cachi2:rpm-ssl-opts where I applied (hopefully) all things I have raised including splitting the work to more commits to give you an idea what I had in mind.

brianwcook:
> @brianwcook sorry, getting back to this one after PTO. 

No problem, I have been occupied with other work too. Thanks for the branch! I just pushed it.

eskultety:
Fixed the merge conflict, rebased on top of main and force pushed to @brianwcook's branch.

tests/dnfserver/README.md
```diff
@@ -0,0 +1,34 @@
+To start the testing nginx DNF repo container use our test helper script:
+Note: make sure you have `createrepo_c` installed (cachi2 has a hard dependency on this)
+```
+$ start.sh
```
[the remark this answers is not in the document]
eskultety: This is standard notation and in fact many terminal emulators have the capability to ignore the prompt sign character when copy-pasting. It also serves as a visual differentiation between commands that have to be run with elevated privileges `#` so the only thing that needs fixing here actually are the couple of lines in this file which are inconsistent with the pattern.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -462,3 +479,48 @@ def _generate_repofiles(
...
+        ("ca_bundle", ca_bundle),
+    ]:
+        if path is None:
+            continue
+
+        if not Path(path).is_file():
+            raise SSLContextError(
+                f"Specified client credential file '{id}':'{path}' is not a regular file",
+                solution=("Ensure the file exist and that is has correct permissions."),
+            )
```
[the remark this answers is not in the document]
eskultety: No, there isn't, can be moved.

Makefile
```diff
@@ -3,6 +3,7 @@ TOX_ENVLIST ?= py3
 TOX_ARGS ?=
 GENERATE_TEST_DATA = false
 TEST_LOCAL_PYPISERVER = false
+TEST_LOCAL_DNF_SERVER = false
```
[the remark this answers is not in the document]
eskultety: This isn't the right PR to do the cleanup, it's consistent with the current setup and a potential cleanup should come as a follow up.

tests/dnfserver/README.md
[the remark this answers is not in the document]
slimreaper35: Yeah, but GitHub does not support it yet. In the terminal, it does not make much of a difference. The majority of our commands in docs,  don't have `$` or `#`.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -277,14 +284,18 @@ def _resolve_rpm_project(source_dir: RootedPath, output_dir: RootedPath) -> list
...
 
         lockfile_relative_path = source_dir.subpath_from_root / DEFAULT_LOCKFILE_NAME
         return _generate_sbom_components(metadata, lockfile_relative_path)
 
 
-def _download(lockfile: RedhatRpmsLock, output_dir: Path) -> dict[Path, Any]:
+def _download(
+    lockfile: RedhatRpmsLock,
+    output_dir: Path,
+    ssl_options: Optional[Dict[str, Optional[Union[str, bool]]]] = None,
```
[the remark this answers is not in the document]
eskultety: There was a mypy type mismatch with one of the callers of this function where the ssl options where propagated from as a dictionary created with `BaseModel.model_dump` and so `_SSLOptions` could not be used. Anyway, reworked heavily to satisfy this request -> resolving.

eskultety:
Since last revision:
- Move any SSL cred file validation to the pydantic BaseModel
- reworked the unit tests completely (tests both the internal _SSLOptions base model as well as primitives working on top of it)
- fix the GH CI integration test suite issue by spawning the test DNF server as part of the job step rather than from pytest
- rebased on top of #682 

Note some commit references for the CI changes will have to tweaked once #682 is merged and the target test RPM repo has to be created (merged with an existing one), but other than that this is ready to review, these final touches will be done in the final revision.

cachi2/core/package_managers/rpm/main.py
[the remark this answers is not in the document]
eskultety: This required substantial changes to the unit test suite, but nevertheless thank you for the suggestion, at least I got to refactor the proposed unit tests as I didn't really like them much, but didn't want to invest time in refactoring every bit of the original proposal.

tests/unit/package_managers/test_rpm.py
```diff
@@ -683,3 +700,82 @@ def test_write(
...
+        assert ssl_context.verify_mode is ssl.CERT_NONE
+
+    @mock.patch.object(ssl.SSLContext, "load_cert_chain", return_value=ssl.create_default_context())
+    @mock.patch.object(ssl.SSLContext, "load_verify_locations", return_value=None)
+    def file_not_found_conditions(self) -> None:
+        with mock.patch.object(path, "isfile", side_effect=self.patched_isfile):
+
+            # Test for proper response when a file is specified which cannot be accessed.
+            options = {
+                "ssl": {
```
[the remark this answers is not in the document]
eskultety: Should be taken care of in the latest revision.

tests/unit/package_managers/test_rpm.py
```diff
@@ -683,3 +700,82 @@ def test_write(
...
+            "client_cert": None,
+            "client_key": None,
+            "ca_bundle": None,
+        }
+        ssl_context = _get_ssl_context(ssl_options)
+        assert ssl_context.verify_mode is ssl.CERT_NONE
+
+    @mock.patch.object(ssl.SSLContext, "load_cert_chain", return_value=ssl.create_default_context())
+    @mock.patch.object(ssl.SSLContext, "load_verify_locations", return_value=None)
+    def file_not_found_conditions(self) -> None:
```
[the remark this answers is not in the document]
eskultety: Unit test  suite reworked.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -15,7 +16,7 @@
 
 from cachi2.core.config import get_config
 from cachi2.core.errors import PackageManagerError, PackageRejected
-from cachi2.core.models.input import Request
+from cachi2.core.models.input import Request, _ExtraOptions, _SSLOptions
```
a-ovchinnikov, line 19: Naming nitpick: if you import them, then they probably should not start with an underscore.

a-ovchinnikov (approved):
LGTM

cachi2/core/package_managers/rpm/main.py
brunoapimentel: I would argue this is more than a nitpick, since the underscore is what denotes that a member of a module is private.

tests/integration/conftest.py
```diff
@@ -73,6 +88,56 @@ def local_pypiserver() -> Iterator[None]:
...
+
+    dnfserver_dir = top_level_test_dir / "dnfserver"
+
+    with contextlib.ExitStack() as context:
+        proc = context.enter_context(subprocess.Popen([dnfserver_dir / "start.sh"]))
+        context.callback(proc.terminate)
+
+        ssl_port = os.getenv("DNFSERVER_SSL_PORT", "8443")
+        for _ in range(60):
+            time.sleep(1)
```
brunoapimentel, line 129: This always fails in my local env with
```
---
Starting DNF server at:
	http://127.0.0.1:8081
	https://127.0.0.1:8443
---

ERROR             
```

If I increase the timeout to 3s, it passes :shrug:

brunoapimentel (approved):
Just a couple of minor comments, I'll go ahead and give you my ack.

tests/integration/conftest.py
eskultety: How exactly did you run it? Anything else running on your system?

> If I increase the timeout to 3s, it passes 🤷

What timeout? The timeout for the liveness check is 60s

cachi2/core/errors.py
```diff
@@ -130,3 +130,23 @@ def __init__(
         The output of the failing command should provide more details, please check the logs.
         """
     ).strip()
+
+
+class SSLContextError(UsageError):
```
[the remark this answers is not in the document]
eskultety: I would swear I had deleted this error type, it's a rebase artifact apparently.

cachi2/core/package_managers/rpm/main.py
eskultety: Added a new commit https://github.com/containerbuildsystem/cachi2/pull/580/commits/3bf9a96a9edd0c9b79ca20bed6ed8b62b51599cb and fixed `_SSLOptions`

eskultety:
> Note some commit references for the CI changes will have to tweaked once https://github.com/containerbuildsystem/cachi2/pull/682 is merged and the target test RPM repo has to be created (merged with an existing one), but other than that this is ready to review, these final touches will be done in the final revision.

- commit msg references tweaked after #682 
- new integration test branch created here: https://github.com/cachito-testing/cachi2-rpm/tree/dnf_tls_client_auth
- review comment tweaks

Merging.

PR #581, opened 2024-07-23
Code QL fixes reported under the 'Security' tab

eskultety:
This PR dabs at fixing the Code QL issues reported under the 'Security' tab. Most (if not all) are false positives anyway, but let's make Code QL happy for a while.
The reported problem with `tar.extractall` will remain even after this PR because Code QL doesn't see past the immediate context at the end of its nose and so it doesn't see that we kinda validated the TAR members before extraction. It probably wants to see us applying the new `filter=` argument, but that's Python 3.12 only material [1] and so we'll have to dismiss that alert manually, besides it's reported against the unit tests so the naive solution proposed here should be good enough to catch test issues early.

[1] https://docs.python.org/3.12/library/tarfile.html#tarfile.TarFile.extractall

### DIFF
--- a/cachi2/core/package_managers/yarn/resolver.py
+++ b/cachi2/core/package_managers/yarn/resolver.py
@@ -285,6 +285,8 @@ class _ComponentResolver:
 
         locator = package.parsed_locator
         checksum = package.checksum
+        name = None
+        version = None
 
         if isinstance(locator, NpmLocator):
             # npm dependencies have reliable names and versions in yarn info output
--- a/tests/unit/package_managers/conftest.py
+++ b/tests/unit/package_managers/conftest.py
@@ -1,5 +1,4 @@
 import copy
-import os
 import tempfile
 from pathlib import Path
 from typing import Any, Generator, Optional, Union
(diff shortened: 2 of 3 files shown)


tests/unit/utils.py
```diff
@@ -0,0 +1,25 @@
+import os
+import tarfile
+from pathlib import Path
+
+
+def _check_tar_members(tarpath: Path, tarmembers: list[tarfile.TarInfo], dest_path: Path) -> None:
```
brunoapimentel, line 6: I think a short docstring explaining what is being checked would be helpful here.

brunoapimentel (commented):
Whereas I don't mind adding the code for the tarball check, I wonder if we shouldn't just dismiss these specific alerts using the `not used in production code` CodeQL option.

In any case, CodeQL is dumb enough to still flag those lines even with the check added :facepalm:

eskultety:
> Whereas I don't mind adding the code for the tarball check, I wonder if we shouldn't just dismiss these specific alerts using the `not used in production code` CodeQL option.
> 
> In any case, CodeQL is dumb enough to still flag those lines even with the check added 🤦

Yes, as noted in
> The reported problem with tar.extractall will remain even after this PR because Code QL doesn't see past the immediate context at the end of its nose and so it doesn't see that we kinda validated the TAR members before extraction. 

Anyway, I spent time on adding the check just for the sake of argument, if we're okay with dismissing it, I can drop that patch and be done with it. :)

brunoapimentel:
> Anyway, I spent time on adding the check just for the sake of argument, if we're okay with dismissing it, I can drop that patch and be done with it. :)
I'm fine either way :)

eskultety:
> > Anyway, I spent time on adding the check just for the sake of argument, if we're okay with dismissing it, I can drop that patch and be done with it. :)
> 
> I'm fine either way :)

I dropped the commit in question and I'll dismiss that particular alert.

brunoapimentel (approved):

taylormadore (approved):

PR #584, opened 2024-07-29
Add bundler to dev package managers

slimreaper35:

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,7 +322,7 @@ Planned:
 
 * dnf
 * cargo
-* rubygems
+* bundler
 
 *Based on the [supported package managers](https://github.com/containerbuildsystem/cachito#package-managers) in the
 original Cachito.*
--- a/cachi2/core/package_managers/bundler/main.py
+++ b/cachi2/core/package_managers/bundler/main.py
@@ -1,0 +1,16 @@
+from cachi2.core.models.input import Request
+from cachi2.core.models.output import EnvironmentVariable, ProjectFile, RequestOutput
+from cachi2.core.models.sbom import Component
+
+
+def fetch_bundler_source(request: Request) -> RequestOutput:
+    """Resolve and process all bundler packages."""
+    components: list[Component] = []
+    environment_variables: list[EnvironmentVariable] = []
+    project_files: list[ProjectFile] = []
+
+    return RequestOutput.from_obj_list(
+        components=components,
+        environment_variables=environment_variables,
+        project_files=project_files,
+    )
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -59,7 +59,7 @@ def _present_user_input_error(validation_error: pydantic.ValidationError) -> str
 
 
 # Supported package managers
-PackageManagerType = Literal["gomod", "npm", "pip", "rpm", "yarn"]
+PackageManagerType = Literal["bundler", "gomod", "npm", "pip", "rpm", "yarn"]
 
 Flag = Literal[
     "cgo-disable", "dev-package-managers", "force-gomod-tidy", "gomod-vendor", "gomod-vendor-check"
@@ -131,6 +131,12 @@ class _DNFOptions(pydantic.BaseModel, extra="forbid"):
         return data
 
 
+class BundlerPackageInput(_PackageInputBase):
+    """Accepted input for a bundler package."""
+
+    type: Literal["bundler"]
+
+
 class GomodPackageInput(_PackageInputBase):
     """Accepted input for a gomod package."""
 
@@ -180,7 +186,14 @@ class YarnPackageInput(_PackageInputBase):
 
 
 PackageInput = Annotated[
-    Union[GomodPackageInput, NpmPackageInput, PipPackageInput, RpmPackageInput, YarnPackageInput],
+    Union[
+        BundlerPackageInput,
+        GomodPackageInput,
+        NpmPackageInput,
+        PipPackageInput,
+        RpmPackageInput,
+        YarnPackageInput,
+    ],
     # https://pydantic-docs.helpmanual.io/usage/types/#discriminated-unions-aka-tagged-unions
     pydantic.Field(discriminator="type"),
 ]
@@ -225,6 +238,11 @@ class Request(pydantic.BaseModel):
         if len(packages) == 0:
             raise ValueError("at least one package must be defined, got an empty list")
         return packages
+
+    @property
+    def bundler_packages(self) -> list[BundlerPackageInput]:
+        """Get the rubygems packages specified for this request."""
+        return self._packages_by_type(BundlerPackageInput)
 
     @property
     def gomod_packages(self) -> list[GomodPackageInput]:
--- a/cachi2/core/package_managers/bundler/__init__.py
+++ b/cachi2/core/package_managers/bundler/__init__.py
@@ -1,0 +1,3 @@
+from cachi2.core.package_managers.bundler.main import fetch_bundler_source
+
+__all__ = ["fetch_bundler_source"]
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -6,7 +6,7 @@ from typing import Any, Callable
 from cachi2.core.errors import UnsupportedFeature
 from cachi2.core.models.input import PackageManagerType, Request
 from cachi2.core.models.output import RequestOutput
-from cachi2.core.package_managers import gomod, npm, pip, rpm, yarn
+from cachi2.core.package_managers import bundler, gomod, npm, pip, rpm, yarn
 from cachi2.core.rooted_path import RootedPath
 from cachi2.core.utils import copy_directory
 
@@ -22,6 +22,7 @@ _package_managers: dict[PackageManagerType, Handler] = {
 # This is where we put package managers currently under development in order to
 # invoke them via CLI
 _dev_package_managers: dict[PackageManagerType, Handler] = {
+    "bundler": bundler.fetch_bundler_source,
     "rpm": rpm.fetch_rpm_source,
 }
 
--- a/tests/unit/models/test_input.py
+++ b/tests/unit/models/test_input.py
@@ -105,7 +105,7 @@ class TestPackageInput:
             ),
             pytest.param(
                 {"type": "go-package"},
-                r"Input tag 'go-package' found using 'type' does not match any of the expected tags: 'gomod', 'npm', 'pip', 'rpm', 'yarn'",
+                r"Input tag 'go-package' found using 'type' does not match any of the expected tags: 'bundler', 'gomod', 'npm', 'pip', 'rpm', 'yarn'",
                 id="incorrect_type_tag",
             ),
             pytest.param(
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -358,7 +358,7 @@ class TestFetchDeps:
                 [
                     "Error: InvalidInput: 1 validation error for user input",
                     "packages -> 0",
-                    "Input tag 'idk' found using 'type' does not match any of the expected tags: 'gomod', 'npm', 'pip', 'rpm', 'yarn'",
+                    "Input tag 'idk' found using 'type' does not match any of the expected tags: 'bundler', 'gomod', 'npm', 'pip', 'rpm', 'yarn'",
                 ],
             ),
             (
@@ -366,7 +366,7 @@ class TestFetchDeps:
                 [
                     "Error: InvalidInput: 1 validation error for user input",
                     "packages -> 0",
-                    "Input tag 'idk' found using 'type' does not match any of the expected tags: 'gomod', 'npm', 'pip', 'rpm', 'yarn'",
+                    "Input tag 'idk' found using 'type' does not match any of the expected tags: 'bundler', 'gomod', 'npm', 'pip', 'rpm', 'yarn'",
                 ],
             ),
             (
@@ -374,7 +374,7 @@ class TestFetchDeps:
                 [
                     "Error: InvalidInput: 1 validation error for user input",
                     "packages -> 0",
-                    "Input tag 'idk' found using 'type' does not match any of the expected tags: 'gomod', 'npm', 'pip', 'rpm', 'yarn'",
+                    "Input tag 'idk' found using 'type' does not match any of the expected tags: 'bundler', 'gomod', 'npm', 'pip', 'rpm', 'yarn'",
                 ],
             ),
             # Missing package type


cachi2/core/package_managers/rubygems/main.py
```diff
@@ -0,0 +1,87 @@
...
+
+    return RequestOutput.from_obj_list(
+        components=components,
+        environment_variables=environment_variables,
+        project_files=project_files,
+    )
+
+
+def _resolve_rubygems(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+    """TODO."""
```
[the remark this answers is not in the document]
brunoapimentel: If I understood your design correctly, this function will also call `_get_package_metadata` and  `_download_dependencies`.

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    return NotImplemented
+
+
+def _bundle_config_exists(package_dir: RootedPath) -> bool:
+    """TODO."""
+    return NotImplemented
+
+
+def _generate_environment_variables() -> list[EnvironmentVariable]:
```
brunoapimentel, line 80: I believe we won't need to set any env vars for RubyGems (unless we go down the path to set a single `.bundle/config` and point to it with `BUNDLE_APP_CONFIG`). On the other hand, I think there's no away around `_generate_project_files`. Should we replace this function here then?

slimreaper35: I thought that we could check if `.bundle/config` exists in the repo. If it does not, we can simplify by just generating environment variables instead of creating a new file.

cachi2/core/package_managers/rubygems/parser.py
```diff
@@ -0,0 +1,82 @@
...
+
+BundlerDependency = Union[GemDependency, GitDependency, PathDependency]
+
+
+def parse_gemlock(package_dir: RootedPath) -> list[BundlerDependency]:
+    """TODO."""
+    return NotImplemented
+
+
+def _validate_gem_dependency(gem: GemDependency) -> None:
```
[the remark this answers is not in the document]
slimreaper35: Added the docs.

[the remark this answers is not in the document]
brunoapimentel: One of the [threads](https://github.com/containerbuildsystem/cachi2/pull/565#discussion_r1700321119) in the design PR pointed out that there's no real reason for us to avoid downloading deps from alternate registries. So I guess we won't be needing this check.

cachi2/core/package_managers/rubygems/main.py
brunoapimentel: Yeah, it's definitely a possiblity. I think we can postpone this decision to later. I'll file the story as something more generic, such as "Set the build configuration". So we can either leave this stub here or remove it.

cachi2/core/package_managers/rubygems/parser.py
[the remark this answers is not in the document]
slimreaper35: Ah, sorry, I completely forgot about that.

cachi2/core/package_managers/rubygems/parser.py
```diff
@@ -0,0 +1,78 @@
...
+import pydantic
+
+from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+GEMFILE_LOCK = "Gemfile.lock"
+
+
+class GemMetadata(pydantic.BaseModel, extra="forbid"):
```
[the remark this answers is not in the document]
slimreaper35: I thought we could distinguish between supported classes for better readability and scalability.
Compare to one `dataclass` in [cachito](https://github.com/containerbuildsystem/cachito/blob/master/cachito/workers/pkg_managers/rubygems.py#L42).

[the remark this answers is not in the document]
brunoapimentel: I like having these classes separate, but I think Taylor's point is that they could simply extend `NamedTuple` or be a `dataclass`, instead of relying on Pydantic.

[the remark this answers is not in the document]
slimreaper35: Okay, I can change it...I assume we won't add any validators or any extra features to those classes, right ?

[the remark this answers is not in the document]
brunoapimentel: I don't think we'll need validators. Their content will always come from the gemlock-parser, so it is pretty much predictable data at this point.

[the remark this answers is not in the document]
slimreaper35: Are you okay with `dataclass` ? `NamedTuple` does not feel to me like something that should be extendable or have a base class.

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -0,0 +1,95 @@
...
+    # https://github.com/rubygems/rubygems/blob/3da9b1dda0824d1d770780352bb1d3f287cb2df5/bundler/lib/bundler/source/git.rb#L327
+    """
+
+
+def _bundle_config_exists(package_dir: RootedPath) -> bool:
+    """TODO."""
+    return NotImplemented
+
+
+def _inject_bundle_config(package_dir: RootedPath) -> None:
```
a-ovchinnikov, line 84: Would this one need an argument for bundle config?

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -0,0 +1,95 @@
...
+    """TODO."""
+
+
+def _generate_environment_variables() -> list[EnvironmentVariable]:
+    """TODO."""
+    return NotImplemented
+
+
+def _generate_purl_for(gem: BundlerDependency) -> str:
```
a-ovchinnikov, line 93: A nitpicky question: I noticed that purls are often typed as strings when in reality they are obviously purls. Is there a reason for not having `Purl` class subclassing `str`? This question holds for other fields in types defined in https://github.com/containerbuildsystem/cachi2/pull/584/files#diff-7c68c39848a484d1335d5268342184ecf184fc0a740e1203e376230d7a35cd62

a-ovchinnikov (commented):
Mostly questions to make me better informed and some minor suggestions.

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -0,0 +1,95 @@
...
+from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+
+def fetch_rubygems_source(request: Request) -> RequestOutput:
+    """Resolve and process all rubygems packages."""
+    components: list[Component] = []
+    environment_variables: list[EnvironmentVariable] = []
+    project_files: list[ProjectFile] = []
```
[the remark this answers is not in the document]
slimreaper35: Yes

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -0,0 +1,95 @@
...
+
+def _download_rubygems_package(output_dir: RootedPath, gem: GemDependency) -> None:
+    """TODO."""
+
+
+def _download_git_package(output_dir: RootedPath, gem: GitDependency) -> None:
+    """
+    TODO.
+
+    # directory has a specific format: {repository_name}-{revision}
```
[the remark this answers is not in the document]
slimreaper35: I don't think we need a subclass for that. It's just the name that we need to put as a parameter to [download_binary_file](https://github.com/slimreaper35/cachi2/blob/ruby-poc/cachi2/core/package_managers/general.py#L28) because of bundler.

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+
+def _resolve_rubygems(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+    """
+    Process a request for a single rubygems package.
+
+    TODO:
+    1. Parse Gemfile.lock
+    2. Get project metadata
```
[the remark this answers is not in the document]
slimreaper35: I renamed it to main package metadata - https://github.com/containerbuildsystem/cachito/blob/master/cachito/workers/pkg_managers/rubygems.py#L523

cachi2/core/package_managers/rubygems/main.py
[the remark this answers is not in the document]
a-ovchinnikov: Yes, however judging from the comment it is not just a name, it is a name of a very specific shape. I.e. not every DirName is a valid RepositoryDirName, and types exist to reflect such nuances among other things. With a proper class one should not be able to instantiate it from incorrect data, so once something receives an instance of such class it can be sure that received object is of right format. MOreover, a consumer would not need to know what makes this format right. Subclassing is cheap and makes intent clearer. I won't block on this, but would strongly recommend considering this option.

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -0,0 +1,95 @@
...
+    """TODO."""
+    return NotImplemented
+
+
+def _get_repository_name(package_dir: RootedPath) -> str:
+    """TODO."""
+    return NotImplemented
+
+
+def _download_dependencies(output_dir: RootedPath, dependencies: list[BundlerDependency]) -> None:
```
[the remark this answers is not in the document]
chmeliik: Are you by any chance a fan of Haskell? :grin: 

Might as well take it a step further:

```python
type IO[T] = T

def _download_dependencies(...) -> IO[None]:
    ...
```

In all seriousness, I think introducing effect types into Python is a step too far. It would make this codebase feel alien to Python devs.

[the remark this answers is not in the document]
a-ovchinnikov: That's a good one! Given that Python drifts towards becoming a poor man's ML dialect among other things (I am looking at you `match`) this might be the right decision!

TBH all the fuss with type annotations does feel a little alien to me as a Python dev: the same result could be achieved with other means without introducing that much visual noise. However annotations do give some helpful hints sometimes, and I think they give higher-quality hints when types are well-refined. That is why you'll see me commenting on types and names quite often. I would rarely if ever block on such comments though.

The main benefit here would be to immediately tell the reader what to expect. This is definitely a slippery slope which may end up with checked exceptions, so please consider this an out-loud thought experiment.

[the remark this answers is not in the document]
chmeliik: > the same result could be achieved with other means without introducing that much visual noise.

Like function and parameter naming?

In addition to being hints for the reader, the value of type hints is static type-checking (which I find immensely helpful).

A type hint like `SideEffect` would only serve the former, it has no benefits for type-checking. IMO the names `_download_dependencies, output_dir` make it clear enough what to expect from this function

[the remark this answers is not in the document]
a-ovchinnikov: Fair enough, my original concern was with effective overloading of `None` symbol to also hint on side-effects. The combination of names indeed implies that dependencies will be not only downloaded, but also probably saved to some directory.

cachi2/core/package_managers/rubygems/parser.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import Optional, Union
+
+from cachi2.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+GEMFILE_LOCK = "Gemfile.lock"
+
+
+@dataclass(init=False)
```
brunoapimentel, line 12: Do we want to keep `init=False`? If we do, we wouldn't be able to do:

```
GemDependency(name="abc", version="1.0")
```

cachi2/core/package_managers/rubygems/main.py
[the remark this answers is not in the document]
brunoapimentel: I believe we can go for a subclass when we're implementing the feature. I see the high-level code just as an initial draft, and what is currently here can (and normally is) changed during the real development.

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -0,0 +1,98 @@
...
+    # https://github.com/rubygems/rubygems/blob/3da9b1dda0824d1d770780352bb1d3f287cb2df5/bundler/lib/bundler/source/git.rb#L327
+    """
+
+
+def _bundle_config_exists(package_dir: RootedPath) -> bool:
+    """TODO."""
+    return NotImplemented
+
+
+def _inject_bundle_config(path: RootedPath) -> None:
```
brunoapimentel, line 92: During this step of the processing, I think what this function will actually do is `_generate_project_files`, since the injection will be done by another CLI command.

brunoapimentel (approved):
Two small comments, but otherwise, this LGTM.

cachi2/core/package_managers/rubygems/main.py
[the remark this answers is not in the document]
a-ovchinnikov: Sure, also I think that having a subclass in HL code is very cheap and requires approximately as much typing as a comment about expected format. Since this is just a minor observation I'll resolve the thread.

cachi2/core/package_managers/rubygems/parser.py
```diff
@@ -0,0 +1,78 @@
...
+        path:       Subpath from package root.
+    """
+
+    path: str
+
+
+BundlerDependency = Union[GemDependency, GitDependency, PathDependency]
+
+
+def parse_gemlock(package_dir: RootedPath) -> list[BundlerDependency]:
```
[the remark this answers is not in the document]
a-ovchinnikov: Suggestion applied, resolving.

cachi2/core/package_managers/rubygems/parser.py
slimreaper35: That's because `source` attribute is not optional.

cachi2/core/package_managers/rubygems/main.py
slimreaper35: I know we had a similar conversation earlier,  but what will we do if there is no `.bundle/config` in the repo?
It should be enough to just set environment variables, right ?

slimreaper35: (but I removed this function)

cachi2/core/package_managers/rubygems/main.py
a-ovchinnikov: Will move this discussion to some other place, resolving here to not hold the PR.

cachi2/core/package_managers/rubygems/parser.py
brunoapimentel: Dumb me, this is just the base class, I see the `@dataclass` (without init=False) keywords on the derived classes now.

cachi2/core/package_managers/rubygems/main.py
brunoapimentel: Yeah, but then, we will need to have both behaviors: injecting project files in case there is a `.bundle/config`, and setting env vars in case there isn't. Whereas only injecting project files works either way.

As we discussed earlier, we can decide this when the implementation story comes up, but either way, this part of the code will either `generate_env_vars` or `generate_project_files`. Injection comes at a later point.

cachi2/core/package_managers/rubygems/main.py
brunoapimentel: I think it might be a good idea to have a centralized PURL class, but then, I think we should have a separate PR that applies it project-wide, not just on the Ruby part.

a-ovchinnikov: Yes, this is definitely out of scope for this PR.

cachi2/core/package_managers/rubygems/main.py
slimreaper35: I kept both functions for the record.

slimreaper35:
rebased

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -1,7 +1,99 @@
...
+
+def _create_components(
+    package_dir: RootedPath,
+    dependencies: list[BundlerDependency],
+) -> list[Component]:
+    """TODO."""
+    return NotImplemented
+
+
+def _generate_purl_for(gem: BundlerDependency) -> str:
```
eskultety, line 51: this one should likely go to a `@dataclass/NamedTuple` instance representing a package, kinda like it's done in gomod and npm and now also proposed for rpm (#600)

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -1,7 +1,99 @@
...
+    dependencies = parse_gemlock(package_dir)
+    _download_dependencies(output_dir, dependencies)
+    return _create_components(package_dir, dependencies)
+
+
+def _create_components(
+    package_dir: RootedPath,
+    dependencies: list[BundlerDependency],
+) -> list[Component]:
+    """TODO."""
```
eskultety, line 47: Why are we adding all these empty implementations where most of them are private anyway and the only relevant entry point is `fetch_rubygems_source`? It happened to us in the past with yarn that we ended up deleting some of the functions because they never got implemented, so I wonder what's the benefit of adding a bunch of TODO functions with interfaces which we likely won't adhere to, change anyway or even revert and drop as the actual bundler implementation progresses instead of **adding the functions as they're needed by some actual code**? Since the structuring more-or-less copies what we already do with other pkg managers we can literally copy-paste on the go as PRs are opened, so I'm really not sure what value this brings if we have so much precedent when it comes to module structuring, e.g. if one is to implement lockfile parsing then such a PR will add primitives for parsing and validating the file as well as appropriate data representations as needed for the actual implementation. Trying to force these right now seems largely inefficient to me. Note that we're not implementing a public library as the situation would be very much different there - i.e. we'd start by introducing and clearly defining interfaces and class/method signatures to be future proof and backwards compatible at the same time - we don't need to do that in this project as our goal is not to provide a library but CLI tool, so this effort feels to me personally a bit misplaced.

brunoapimentel: I see these as a draft: they will very likely change as the implementation progresses. But their purpose now is to give an overall idea of the code flow, and to allow a developer to just pick a piece of the implementation such as "_inject_project_files" and limit his/her work to that specific piece, while other parallel development is made.

I believe we could also simply drop these stubs and try adding everything as it is needed, but then... what would be the criteria for things we would keep in this PR? Or should we drop this PR altogether? 

All in all, I still think this draft is bringing at least a small initial value, so I'd rather just keep it, but also not invest more time on this, since as you said it yourself, we will change it a lot.

eskultety: > I see these as a draft: they will very likely change as the implementation progresses. But their purpose now is to give an overall idea of the code flow, and to allow a developer to just pick a piece of the implementation such as "_inject_project_files" and limit his/her work to that specific piece, while other parallel development is made.

I think we're at the point where we've set enough precedent for people to copy-paste from existing pkg managers and adjust what's needed, so whoever is able to read code they should get the gist. With empty stubs, people will only gain names and sort of unfamiliar interfaces and they'll copy-paste from existing code the same way as if they didn't have anything, so the only value in empty stubs I can see is to define stable interfaces, e.g. via abstract classes and such, but that's important for libraries where definition of an interface is a careful process, not here, we're free to change any of this at will.

> All in all, I still think this draft is bringing at least a small initial value, so I'd rather just keep it, but also not invest more time on this, since as you said it yourself, we will change it a lot.

Just like with Yarn v2+ back in the day I am not a fan of empty stubs (never was) for the reasons I mentioned as I don't see any value added there. To answer the question of whether this PR should be dropped, possibly, but we could also consider this PR to be the starting line for the RubyGems/Bundler development, so the `--dev-package-manager` bits are still valid to merge IMO, the PR will just need to be renamed accordingly.

eskultety: That said, it's fairly common in other projects for contributors to post an RFC (an equivalent of our design doc + a POC implementation) where they outline everything and it's also fairly common to contain empty stubs for presentation purposes...the difference however is that those RFC POCs are never merged, they simply just act as v0, once a conclusion is reached, a proper v1 implementation (either whole, or in parts) is proposed.

slimreaper35: Hopefully, we will not change much :smile: I mean, we will at least something, but that's why this PR goes through reviews, to avoid too many changes, no ?

eskultety: Not really, because the empty stubs are (majority of them) are not even connected, so naturally people will fallback to creating their own helpers and if during review one doesn't raise that there is an empty stub supposed to take care of it, we'll have to remove it later, so to me these really are pointless apart from `_fetch_XZY` because that's the entry point.

slimreaper35:
So I made adjustment based on our discussion in https://github.com/containerbuildsystem/cachi2/pull/599
PTAL

cachi2/core/package_managers/rubygems/parser.py
```diff
@@ -0,0 +1,96 @@
...
+        checksum:   The checksum of the gem.
+    """
+
+    source: str
+    checksum: Optional[str] = None
+
+    @property
+    def purl(self) -> str:
+        """Get purl for this gem dependency."""
+        return NotImplemented
```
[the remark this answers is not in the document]
slimreaper35: Let's just drop that commit altogether from this PR.

eskultety (commented):
I think the commit message needs a bit of tweaking - we're adding bundler, not rubygems. A bit of context, e.g. a link pointing to the merged design doc (which we haven't yet) would be nice. Once you tweak those nits and provided we've merged the design, this is ready to go in.

README.md
```diff
@@ -322,7 +322,7 @@ Planned:
 
 * dnf
 * cargo
-* rubygems
+* bundler
```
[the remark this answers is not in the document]
slimreaper35: I don't know. One can say it is not clear that cargo is for Rust. So I would probably keep it consistent with all package managers.

cachi2/core/package_managers/rubygems/main.py
brunoapimentel: Well, with everything stripped down, this no longer serves the purpose of being a high level draft of the code. During this code review, we parted from having only a few high level functions with comments, to switching the comments to the stubs, to completely removing everything. I don't think the final state is better than the first one though, since we lost all the information we wanted to convey.

I don't think starting with a high-level draft is a must, the main reason that led us to this path was that this was successfully done during the Yarn implementation, and the team had a positive feedback on it. Since I believe we're dragging this way to a point where the time spent here outweights the limited usefulness of this PR, I will just agree with the direction we took here.

brunoapimentel (approved):
The newer changes LGTM

cachi2/core/package_managers/rubygems/main.py
slimreaper35: We should avoid this scenario in the future. I opened the PR 3 weeks ago. The current, relatively trivial change could be merged in a couple of days (at maximum). Others and I could already start working on different stories.

README.md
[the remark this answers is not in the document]
eskultety: > I agree that we should change both cargo and bundler if we want to clarify it here.

That was the intention when I said:

> but we might want to improve anyway in the future.

cachi2/core/package_managers/rubygems/main.py
eskultety: > I don't think the final state is better than the first one though, since we lost all the information we wanted to convey.

What was the information we wanted to convey with empty stubs that will very likely end up unused dead code?

brunoapimentel: How the code was envisioned when the design was done, and also a practical aspect of limiting what a contributor would need to touch if trying to implement a standalone feature not yet connected to the whole.

eskultety: > Others and I could already start working on different stories.

Nothing stops you or anyone else from doing so, IMO any attempt to merge some high level code was the actual blocker here. The first blocker here is the design doc which should precede any work, naturally. Once that is done there shouldn't be any further blockers for individual parts of the implementation to commence, IOW as we have the pkg manager enabled on the CLI then I don't see anyone having to wait for anything really. Those empty stubs wouldn't have prevented merge conflicts resulting from concurrent work on the feature and those are a natural part of any distributed collaboration.

brunoapimentel: >  Others and I could already start working on different stories.

> > Nothing stops you or anyone else from doing so

It is a distraction, to say the least, since he needs to both address the comments on this and also to plan his following work on the assumption this will be merged.

> Once that is done there shouldn't be any further blockers for individual parts of the implementation to commence

I agree this is a valid approach, but we had defined to go for a different route, based on the team's previous experience.

eskultety: > How the code was envisioned when the design was done, and also a practical aspect of limiting what a contributor would need to touch if trying to implement a standalone feature not yet connected to the whole.

@brunoapimentel once a design doc has been approved across the team then any attempt to lay out these implementation with these stubs for collaborators is just unnecessarily tying their hands to comply with some internal interface stubs. It should be up to the collaborator's creativity how to implement the feature based on the spec from the design doc regardless of some intended guard rails in the form of the stubs.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -0,0 +1,16 @@
+from cachi2.core.models.input import Request
+from cachi2.core.models.output import EnvironmentVariable, ProjectFile, RequestOutput
+from cachi2.core.models.sbom import Component
+
+
+def fetch_bundler_source(request: Request) -> RequestOutput:
+    """Resolve and process all bundler packages."""
+    components: list[Component] = []
```
a-ovchinnikov, line 8: Do you plan to eventually use these lists (L8:L10) in this function?

slimreaper35: yes

slimreaper35: But they will be populated with some data :smile:  I should mention

a-ovchinnikov: Thank you, they do look a bit suspicious, so I had to ask.

PR #585, opened 2024-07-29
add design doc for Yarn 1.x support

taylormadore:
You can see a barebones PoC with just prefetch (No SBOM generation) using the offline mirror here: https://github.com/taylormadore/cachi2/tree/yarn-v1-limited-poc

- [-] Code coverage from testing does not decrease and new code is covered

### DIFF
--- a/docs/design/yarn-v1.md
+++ b/docs/design/yarn-v1.md
@@ -1,0 +1,182 @@
+# Cachi2 Yarn v1 (Yarn Classic) Prefetch Design
+
+## Yarn Overview
+[Yarn](https://github.com/yarnpkg/yarn) is a package manager for JavaScript. While Yarn v1.x is no longer actively developed and has been
+succeeded by [Yarn v2+ (Berry)](https://github.com/yarnpkg/berry), it [remains widely used](https://npmtrends.com/yarn).
+
+The [official documentation](https://classic.yarnpkg.com/en/docs) contains information about the various features of Yarn, configuration settings, and CLI commands.
+
+## Proposed Cachi2 Implementation Overview
+
+### Specifying a Yarn Package in a Cachi2 Request
+Cachi2 users will specify `yarn` as the package manager or package type for all versions of Yarn. In yarn.lock, the presence of the [`__metadata`](https://github.com/yarnpkg/berry/blob/13d5b3041794c33171808fdce635461ff4ab5c4e/packages/yarnpkg-core/sources/Project.ts#L374) field can be used to presume a Yarn v2+ lockfile as can the presence of the string [`yarn lockfile v1`](https://github.com/yarnpkg/berry/blob/13d5b3041794c33171808fdce635461ff4ab5c4e/packages/yarnpkg-core/sources/Project.ts#L434) for a Yarn v1.x lockfile. The request can then be routed to the correct Yarn module in Cachi2.
+
+### Offline Mirror
+Prefetching dependencies for Yarn in Cachi2 will be done using Yarn's [offline mirror](https://classic.yarnpkg.com/blog/2016/11/24/offline-mirror/)
+feature. When a project is configured to use the offline mirror, Yarn will store compressed archives in the mirror directory and can install
+cached project dependencies from there later without network access.
+
+The offline mirror also has a setting that controls [whether it removes package archives](https://classic.yarnpkg.com/en/docs/prune-offline-mirror/)
+from the mirror directory when they are no longer needed. Cachi2 will want to disable automatic pruning of the offline mirror since we support multiple
+yarn projects in a single request.
+
+To enable the offline mirror, the following settings need to be applied either via .yarnrc or from environment variables:
+
+    yarn-offline-mirror <absolute path to the request output directory>
+    yarn-offline-mirror-pruning false
+
+    YARN_YARN_OFFLINE_MIRROR=<absolute path to the request output directory>
+    YARN_YARN_OFFLINE_MIRROR_PRUNING=false
+
+### Installing Yarn (via Corepack)
+Yarn commands will be executed in order to populate the offline mirror, so Yarn will need to be installed in the Cachi2 image. Corepack is
+already being used to install Yarn in the Cachi2 Yarn v2+ implementation, so a similar approach is suggested.
+
+Corepack installs the latest known-good Yarn v1.x release by default if no other version is specified explicitly. It seems reasonable to always
+use the latest 1.x release for the prefetch, since 1.x is no longer under active development and we never previously allowed users to specify a specific
+Yarn version to process their request in Cachito. We could add more flexibility with a later enhancement if needed.
+
+Corepack could be configured to ignore user project configuration and use the global Yarn 1.x default by specifying the following
+[environment variable](https://github.com/nodejs/corepack#environment-variables):
+
+    COREPACK_ENABLE_PROJECT_SPEC=0
+
+Alternatively we could configure corepack explicitly using the packageManager field in package.json like is done in Cachi2's Yarn v2+ implementation.
+
+We will also need to set the following to prevent corepack from prompting for interactive user input when downloading Yarn:
+
+    COREPACK_ENABLE_DOWNLOAD_PROMPT=0
+
+### Configuring Yarn (Using Environment Variables)
+The behavior of Yarn can be configured through the use of environment variables, configuration files, and CLI options for specific commands.
+
+Environment variables will be preferred by Cachi2 when available.
+
+Unfortunately the official Yarn documentation is quite limited or non-existent concerning environment variables.
+
+The environment variables specific to the Yarn package manager are prefixed by `YARN_`. When specified, they are [loaded into the registry object](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/registries/base-registry.js#L125) and then made available to the rest of the Yarn application [via the Config object](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/config.js#L231). An [example](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/config.js#L422) of this can be seen for the `YARN_YARN_OFFLINE_MIRROR_PRUNING` (undocumented) environment variable.
+
+Yarn also has support for including [NPM environment variables](https://docs.npmjs.com/cli/v10/using-npm/config#environment-variables) prefixed with `NPM_CONFIG_`.
+
+### Project Configuration
+The [package.json](https://classic.yarnpkg.com/en/docs/package-json) and [yarn.lock](https://classic.yarnpkg.com/en/docs/yarn-lock)
+configuration files must be present in order to process the request.
+
+The node_modules directory should be absent. It may not be possible to guard against a user-specified modulesFolder at build-time.
+
+Yarn projects using the [Plug'n'Play (PnP)](https://classic.yarnpkg.com/lang/en/docs/pnp/) feature will not be supported. PnP enables the use of unplugged dependencies and we are unable to verify them.
+
+Any additional configuration settings in the repository via [.yarnrc](https://classic.yarnpkg.com/en/docs/yarnrc) or .npmrc
+files should be ignored (at least initially) for the prefetch by specifying [`--no-default-rc`](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L73) for the yarn install command.
+
+We never previously honored user-defined settings with Cachito, but this can be investigated and implemented in Cachi2 in a follow-up. At minimum, the
+limitation should be documented.
+
+### SBOM Generation
+The Yarn 1.x CLI doesn't provide enough information to generate an SBOM, so the yarn.lock and package.json files will need to
+be parsed in order to gather the necessary data. The code used to process these files will need to be
+[imported from or inspired by Cachito](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py)
+and refactored/improved.
+
+#### Key Points
+ - The yarn.lock file can be parsed with [pyarn](https://github.com/containerbuildsystem/pyarn) like in cachito
+ - Workspaces will need to be [processed separately](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L95) from package.json since they do not appear in the yarn.lock file
+ - Dev dependencies can be [determined similiarly](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L140) to cachito
+ - We will need to ensure that there are no collisions in the names of the archive files being added to the offline mirror
+
+#### PURL Examples
+[NPM PURL specification](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#npm)
+
+##### Registry Package
+```txt
+pkg:npm/%40optional-scope/package-name@1.0.0
+```
+
+##### URL Package
+```txt
+pkg:npm/package-name@1.0.0?checksum=sha512:checksum&download_url=https://example.com/package-name/0101010101010101010101010101010101010101.tar.gz
+```
+
+##### Git Package
+
+```txt
+pkg:npm/package-name@1.0.0?vcs_url=git%2Bhttps://example.com/namespace/package-name.git%400101010101010101010101010101010101010101
+```
+
+##### Workspace/File Package
+
+```txt
+pkg:npm/package-name@1.0.0?vcs_url=git%2Bhttps://example.com/namespace/package-name.git%400101010101010101010101010101010101010101#subpath
+```
+
+### Avoiding Arbitrary Code Execution
+Yarn 1.x appears to honor the `--ignore-scripts` option for the install command. It can also be configured via environment variable.
+Used https://github.com/chmeliik/js-lifecycle-scripts/tree/yarn to verify.
+
+### Prefetch
+The offline mirror in the request output directory will be populated by running the following Yarn command in the project root:
+
+    yarn install --non-interactive --frozen-lockfile --disable-pnp --no-default-rc --ignore-engines
+
+Command: [yarn install](https://classic.yarnpkg.com/en/docs/cli/install)
+
+Options:
+ - [non-interactive](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-non-interactive): Disable interactive prompts
+ - [frozen-lockfile](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-frozen-lockfile): Fail if yarn.lock updates are needed
+ - [disable-pnp](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L84): Disable Plug'n'Play installation
+ - [no-default-rc](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L73): Prevent Yarn from automatically detecting .yarnrc and .npmrc files
+ - [ignore-engines](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-ignore-engines): Ignore the engine node version during the prefetch
+
+ The behavior of Yarn can also be configured using select environment variables. The following ones should be set for the prefetch:
+
+ - COREPACK_ENABLE_DOWNLOAD_PROMPT=0
+ - COREPACK_ENABLE_PROJECT_SPEC=0 (If we decide to use the global default)
+ - YARN_YARN_OFFLINE_MIRROR=\<absolute path to the request output directory\>
+ - YARN_YARN_OFFLINE_MIRROR_PRUNING=false
+ - [YARN_IGNORE_PATH](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L627)=true (Use the corepack yarn version, not any Yarn version specified by the user)
+ - [YARN_IGNORE_SCRIPTS](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/config.js#L415)=true
+
+### Build Configuration
+The user build can be configured to use the offline mirror by configuring the following environment variables:
+
+    YARN_YARN_OFFLINE_MIRROR=<request output directory>
+    YARN_YARN_OFFLINE_MIRROR_PRUNING=false
+
+## Implementation Scoping
+ - Since we're executing Yarn commands directly, process the request in an [isolated temporary directory](https://github.com/containerbuildsystem/cachi2/blob/6953607b6ef52fd3f0bef7059d2c926767b1022b/cachi2/core/resolver.py#L41) to avoid any potential changes to the repository
+ - Determine whether to process the request with yarn v1 or yarn v2+
+ - Read the project files
+   - Read/Require package.json
+   - Read/Require yarn.lock
+     - Ensure yarn.lock is formatted for Yarn 1.x
+ - Verify the repository
+   - Ensure node_modules directory is not present
+   - Ensure that Plug'n'Play (PnP) is not being used
+ - Gather the SBOM Components
+   - Get the main package component data from package.json
+   - Get the workspaces from package.json and add a component for each workspace according to the workspace package.json data
+   - Parse yarn.lock with pyarn and add components for each
+     - Report missing integrity keys from yarn.lock as missing hashes
+     - Ensure that there are no collisions between expected archive names in the offline mirror
+   - Ensure devDependencies are reported with the `cdx:npm:package:development` property
+ - Configure environment for the prefetch
+   - Ensure offline mirror is used
+   - Ensure offline mirror is not automatically pruned
+   - Ensure package scripts are ignored
+   - Ensure Yarn 1.x is used via corepack
+ - Do the prefetch
+   - Run yarn install
+ - Configure environment for the user build
+   - Ensure offline mirror is used from the request output
+   - Ensure offline mirror is not pruned automatically
+ - Add integration (and e2e) tests
+ - Add documentation
+   - Document that user-specified settings via .yarnrc/.npmrc are ignored during the prefetch
+   - Document how to install yarn itself in the user repository
+
+## Out of Scope
+ - Allowing versions of yarn other than the latest 1.x release to be specified by the user
+ - Allowing user-specified settings during the prefetch via .yarnrc/.npmrc
+
+## References
+Some of the implementation details in this design are inspired by and/or will be ported from the [Yarn 1.x implementation in Cachito](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py).


docs/design/yarn-v1.md
```diff
@@ -0,0 +1,135 @@
...
+#### Key Points
+ - The yarn.lock file can be parsed with [pyarn](https://github.com/containerbuildsystem/pyarn) like in cachito
+ - Workspaces will need to be [processed separately](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L95) from package.json since they do not appear in the yarn.lock file
+ - Dev dependencies can be [determined similiarly](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L140) to cachito
+ - We will need to ensure that there are no collisions in the names of the archive files being added to the offline mirror
+
+
+### Avoiding Arbitrary Code Execution
+Yarn 1.x appears to honor the ```--ignore-scripts``` option for the install command. It can also be configured via environment variable.
+Used https://github.com/chmeliik/js-lifecycle-scripts/tree/yarn to verify.
```
chmeliik, line 81: Does `--ignore-scripts` also apply to the scripts of your git dependencies?

(i.e. something similar to https://github.com/chmeliik/js-lifecycle-scripts/commit/f4269945d491c71bd19000883569a4b92f2785aa, but change the dep from a registry dep to a git dep)

taylormadore: It does appear to. I had added that project (specifically the yarn branch) as a git dependency for another project during testing and didn't observe the scripts execute :man_shrugging:

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,135 @@
...
+
+    YARN_YARN_OFFLINE_MIRROR=<request output directory>
+    YARN_YARN_OFFLINE_MIRROR_PRUNING=false
+
+## High-Level Procedure
+ - Since we're executing Yarn commands directly, process the request in an [isolated temporary directory](https://github.com/containerbuildsystem/cachi2/blob/6953607b6ef52fd3f0bef7059d2c926767b1022b/cachi2/core/resolver.py#L41) to avoid any potential changes to the repository
+ - Read the project files
+   - Read/Require package.json
+   - Read/Require yarn.lock
+     - Ensure yarn.lock is formatted for Yarn 1.x
```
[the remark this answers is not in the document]
taylormadore: I've successfully convinced myself of both options so far, so I'm curious what you all think :sweat_smile:

docs/design/yarn-v1.md
chmeliik: \o/

docs/design/yarn-v1.md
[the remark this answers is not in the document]
chmeliik: I'd lean towards `yarn`, feels more user-friendly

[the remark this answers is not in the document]
slimreaper35: +1 on `yarn` (or possibly `yarn-classic` , `yarn-berry`)
IIRC we plan to support yarn v4 in the future, so we would need three yarn types, right ?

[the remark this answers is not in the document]
chmeliik: Not really, we support multiple versions of each package manager without asking the user to specify the exact version.

Yarn 1 vs. Yarn 2+ is an exception - they are so different that they have less in common than NPM and Yarn.

But from the users' point of view, it's still just `yarn`

taylormadore:
Since V1:
- Non-substantive edits to wording
- Added a section describing how users will specify the Yarn (v1 and v2+) package manager: https://github.com/containerbuildsystem/cachi2/pull/585#discussion_r1700431511
- Added PURL examples
- Added a section for follow-up actions

docs/design/yarn-v1.md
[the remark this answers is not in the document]
eskultety: > But from the users' point of view, it's still just yarn

it is, but for us it's very different, so in this case I'd adhere to sticking to the official names --> `yarn-classic` because otherwise we'll have figure out ourselves which implementation to use, is that the intention here, do it dynamically or do we require the user to specify it? Because if it's supposed to be dynamic then I have to retract on the name preference as we'll have to cope with just `yarn`.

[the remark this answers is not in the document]
eskultety: Right, I noticed that bit somewhere in the document in the end.

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
+# Cachi2 Yarn v1 (Yarn Classic) Prefetch Design
+
+## Yarn Overview
+[Yarn](https://github.com/yarnpkg/yarn) is a package manager for JavaScript. While Yarn v1.x is no longer actively developed and has been
+succeeded by [Yarn v2+ (Berry)](https://github.com/yarnpkg/berry), it [remains widely used](https://npmtrends.com/yarn).
+
+## Cachito Implementation Overview
+Cachito supports the Yarn 1.x package manager. It directly parses package.json files and uses our [pyarn](https://github.com/containerbuildsystem/pyarn)
```
eskultety, line 8: Similarly to my recent ruby comment I wonder if it's a good idea for us to pull in another vendored library we created ourselves to deal with Yarn and handle parsing from grounds up instead of spinning up a Yarn environment, do: https://github.com/yarnpkg/yarn/tree/master/packages/lockfile and consume the JSON. I think it might be worth investigating.

eskultety: That said, we'd need to check if we haven't added some logic the native parser can't handle, e.g. one of the "recent" changes to our lib: https://github.com/containerbuildsystem/pyarn/commit/9ddf67ea02977c346c21a9ba29d0c98af1351eae 

Just food for thought.

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+### Offline Mirror
+Prefetching dependencies for Yarn in Cachi2 will be done using Yarn's [offline mirror](https://classic.yarnpkg.com/blog/2016/11/24/offline-mirror/)
+feature. When a project is configured to use the offline mirror, Yarn will store compressed archives in the mirror directory and can install
+cached project dependencies from there later without network access.
+
+The offline mirror also has a setting that controls [whether it removes package archives](https://classic.yarnpkg.com/en/docs/prune-offline-mirror/)
+from the mirror directory when they are no longer needed. Cachi2 will want to disable automatic pruning of the offline mirror since we support multiple
+yarn projects in a single request.
+
+To enable the offline mirror, the following settings need to be applied either via .yarnrc or from environment variables:
```
ben-alkov, line 32: > either via .yarnrc or from environment variables:

Are we going to pick one?

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,135 @@
...
+Corepack installs the latest known-good Yarn v1.x release by default if no other version is specified explicitly. It seems reasonable to always
+use the latest 1.x release for the prefetch, since 1.x is no longer under active development and we never previously allowed users to specify a specific
+Yarn version to process their request in Cachito. We could add more flexibility with a later enhancement if needed.
+
+Corepack could be configured to ignore user project configuration and use the global Yarn 1.x default by specifying the following
+[environment variable](https://github.com/nodejs/corepack#environment-variables):
+
+    COREPACK_ENABLE_PROJECT_SPEC=0
+
+Alternatively we could configure corepack explicitly using the packageManager field in package.json like is done in Cachi2's Yarn v2+ implementation.
```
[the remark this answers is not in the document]
ben-alkov: I feel like consistency with existing Yarn implementation(s) (also well-defined behavior) would be best...

but yeah

> We can always allow more flexibility *after the initial implementation* is complete if we see value in it

if someone asks for it, or we see a good reason to

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+
+We will also need to set the following to prevent corepack from prompting for interactive user input when downloading Yarn:
+
+    COREPACK_ENABLE_DOWNLOAD_PROMPT=0
+
+### Project Configuration
+The [package.json](https://classic.yarnpkg.com/en/docs/package-json) and [yarn.lock](https://classic.yarnpkg.com/en/docs/yarn-lock)
+configuration files must be present in order to process the request.
+
+The node_modules directory should be absent. I'm not certain at the moment how to guard against a user-specified modulesFolder at build-time.
```
ben-alkov, line 63: > user-specified modulesFolder at build-time

Is this a config-file option? Can we check the parsed config for it?

docs/design/yarn-v1.md
taylormadore: Yeah, environment variables are preferred for sure :+1:

I think I was just mentioning it for completeness here since the section about environment variables is further down, but I see how that could be confusing

docs/design/yarn-v1.md
ben-alkov: TL;DR - it can be a configfile/env option, but it's also a CLI option to `yarn` itself, e.g.: `yarn install --modules-folder <path>`, so we might not be able to do much about it.

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+Corepack installs the latest known-good Yarn v1.x release by default if no other version is specified explicitly. It seems reasonable to always
+use the latest 1.x release for the prefetch, since 1.x is no longer under active development and we never previously allowed users to specify a specific
+Yarn version to process their request in Cachito. We could add more flexibility with a later enhancement if needed.
+
+Corepack could be configured to ignore user project configuration and use the global Yarn 1.x default by specifying the following
+[environment variable](https://github.com/nodejs/corepack#environment-variables):
+
+    COREPACK_ENABLE_PROJECT_SPEC=0
+
+Alternatively we could configure corepack explicitly using the packageManager field in package.json like is done in Cachi2's Yarn v2+ implementation.
```
eskultety, line 53: Side note: ^this is an example of where we'd need a decision in the future design doc template to explicitly state which option we went with over a known alternative. It may help in the future if we encounter a problem requiring us to reconsider.

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+   - Read/Require package.json
+   - Read/Require yarn.lock
+     - Ensure yarn.lock is formatted for Yarn 1.x
+ - Verify the repository
+   - Ensure node_modules directory is not present
+ - Gather the SBOM Components
+   - Get the main package component data from package.json
+   - Get the workspaces from package.json and add a component for each workspace according to the workspace package.json data
+   - Parse yarn.lock with pyarn and add components for each
+     - Report missing integrity keys from yarn.lock as missing hashes
```
eskultety, line 154: Can someone more knowledgeable of Yarn v1 explain what the following is supposed to achieve and whether it's related to this problem: https://classic.yarnpkg.com/en/docs/yarnrc#toc-unsafe-disable-integrity-migration

Would it help or it's completely irrelevant?

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+ - ignore-engines: Ignore the engine node version during the prefetch
+
+ The behavior of Yarn can also be configured using select environment variables. The following ones should be set for the prefetch:
+
+    COREPACK_ENABLE_DOWNLOAD_PROMPT=0
+    COREPACK_ENABLE_PROJECT_SPEC=0 (If we decide to use the global default)
+    YARN_YARN_OFFLINE_MIRROR=<absolute path to the request output directory>
+    YARN_YARN_OFFLINE_MIRROR_PRUNING=false
+    YARN_IGNORE_PATH=true (Use the corepack yarn version, not any Yarn version specified by the user)
+    YARN_IGNORE_SCRIPTS=true
```
eskultety, line 133: Where are these env vars documented? Is the official [page](https://classic.yarnpkg.com/en/docs/envvars) a joke? 
Unless I missed something this GH issue: https://github.com/yarnpkg/yarn/issues/8784 confirms what the page lists, so are we going to fall back to NPM options? I am confused, I don't dispute the variables work, it's just we need to be able to reference them to their respective docs.

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+
+Registry dependencies are fetched through a proxy repository in Nexus and the user's build is configured to use that proxy as the registry.
+
+Non-registry dependencies are packed and uploaded to a hosted Nexus repository. The versions of these non-registry dependencies are
+updated in package.json and yarn.lock to match those in the hosted repository.
+
+## Proposed Cachi2 Implementation Overview
+
+### Specifying a Yarn Package in a Cachi2 Request
+Cachi2 users will specify ```yarn``` as the package manager or package type for all versions of Yarn. The presence of the ```__metadata``` field in yarn.lock
```
eskultety, line 19: Are we intending to parse out the __metadata field from the lockfile with a regex or are we taking a more conservative route: try parsing with V2+ and if it fails then try V1? I'm asking just out of curiousity, it's an implementation detail.

eskultety: Correction, `__metadata` is already the parsed out representation, sorry for the confusion, looking at https://github.com/yarnpkg/berry/blob/13d5b3041794c33171808fdce635461ff4ab5c4e/packages/yarnpkg-core/sources/Project.ts#L434 they just search for a static string sequence, shall we do the same to determine?

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+    COREPACK_ENABLE_PROJECT_SPEC=0
+
+Alternatively we could configure corepack explicitly using the packageManager field in package.json like is done in Cachi2's Yarn v2+ implementation.
+
+We will also need to set the following to prevent corepack from prompting for interactive user input when downloading Yarn:
+
+    COREPACK_ENABLE_DOWNLOAD_PROMPT=0
+
+### Project Configuration
+The [package.json](https://classic.yarnpkg.com/en/docs/package-json) and [yarn.lock](https://classic.yarnpkg.com/en/docs/yarn-lock)
```
eskultety, line 60: According to https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install the lockfile is not mandatory if package.json can satisfy the need, is this a deliberate decision and we want to intentionally limit our users? (Not that I object, less work for us, just double checking)

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+configuration files must be present in order to process the request.
+
+The node_modules directory should be absent. I'm not certain at the moment how to guard against a user-specified modulesFolder at build-time.
+
+Any additional configuration settings in the repository via [.yarnrc](https://classic.yarnpkg.com/en/docs/yarnrc) or .npmrc
+files should be ignored (at least initially) for the prefetch by specifying ```--no-default-rc``` for the yarn install command.
```
eskultety, line 66: where is the `--no-default-rc` install command option documented?

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+### Avoiding Arbitrary Code Execution
+Yarn 1.x appears to honor the ```--ignore-scripts``` option for the install command. It can also be configured via environment variable.
+Used https://github.com/chmeliik/js-lifecycle-scripts/tree/yarn to verify.
+
+### Prefetch
+The offline mirror in the request output directory will be populated by running the following Yarn command in the project root:
+
+    yarn install --non-interactive --frozen-lockfile --disable-pnp --no-default-rc --ignore-engines
```
eskultety, line 115: Are we potentially interested also in `--update-checksums` [1]? Since we're processing the repo in a temp environment we could see if the `yarn.lock` reports correct info that would not get updated during the actual install during hermetic build and report that fact (based on unstaged git changes) to the end user?
[1] https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-update-checksums

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+### SBOM Generation
+The Yarn 1.x CLI doesn't provide enough information to generate an SBOM, so the yarn.lock and package.json files will need to
+be parsed in order to gather the necessary data. The code used to process these files will need to be
+[imported from or inspired by Cachito](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py)
+and refactored/improved.
+
+#### Key Points
+ - The yarn.lock file can be parsed with [pyarn](https://github.com/containerbuildsystem/pyarn) like in cachito
+ - Workspaces will need to be [processed separately](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L95) from package.json since they do not appear in the yarn.lock file
+ - Dev dependencies can be [determined similiarly](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L140) to cachito
```
eskultety, line 80: So the conclusion here is that we're doing dev dependencies OR just that we want to filter them out reliably?
Would https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-production-true-false help in any way? Probably not right, because we don't control the build environment, or more specifically, the build recipes (i.e. Dockerfiles).

docs/design/yarn-v1.md
eskultety: Also, again, where does the `disable-pnp` option come from? Am I looking at the wrong docs?

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,169 @@
...
+   - Ensure Yarn 1.x is used via corepack
+ - Do the prefetch
+   - Run yarn install
+ - Configure environment for the user build
+   - Ensure offline mirror is used from the request output
+   - Ensure offline mirror is not pruned automatically
+
+## Potential Follow-Ups
+ - Allow versions of yarn other than the latest 1.x release to be specified by the user
+ - Investigate allowing more user-specified settings during the prefetch via .yarnrc/.npmrc
```
eskultety, line 169: Side note, I don't think we'll want to record this in the final design doc when the template materializes as I think the point of ADRs/RFCs/Blueprints is to capture a snapshot in a given time and provide updates in a form of new docs when needed.

docs/design/yarn-v1.md
eskultety, line 8: I'd probably refrain from mentioning Cachito too much. It is okay to mention that the core logic will be ported from it at the beginning of the doc, but then approach the doc like a completely new thing to cachi2, i.e. doing an overview of the functionality, mention the pitfalls and how we intend to approach it and treat it basically as if cachito never existed.

eskultety (commented):
I do have a few, mostly clarifying, comments, most importantly about CLI options I haven't found in the official docs, otherwise I don't see many issues.

docs/design/yarn-v1.md
taylormadore: We report missing **_user-provided_** hashes in the SBOM with the `cachi2:missing_hash:in_file` property, so if we encounter missing hashes for any dependencies in yarn.lock we should report them as such.

I think the `unsafe-disable-integrity-migration` option was created when the integrity key was first introduced to yarn.lock several years ago. It provided a mechanism for preventing yarn from automatically updating all of the existing dependencies in a lockfile with the new integrity values (which didn't become required yarn v2 IIRC).

docs/design/yarn-v1.md
ben-alkov: See also discussion in "cachi2 rubygems / bundler design" https://github.com/containerbuildsystem/cachi2/pull/565#pullrequestreview-2248406464

docs/design/yarn-v1.md
taylormadore: Isn't that page basically just saying that if yarn.lock is missing or incomplete, `yarn install` will update/create it? I think the expectation from the yarn side is that yarn.lock should always be checked in to source control: https://classic.yarnpkg.com/en/docs/yarn-lock#toc-check-into-source-control

docs/design/yarn-v1.md
taylormadore: I had thought to do pretty much exactly what they do:

1. Check for the `__metadata` field, if present assume v2+ lockfile
2. Check for `yarn lockfile v1`, if present assume v1 lockfile ([we also do this in pyarn](https://github.com/containerbuildsystem/pyarn/blob/c947823027b421c7ec42e247246d9986fbc2843d/pyarn/lockfile.py#L32))
3. Error if neither condition satisfied, since we cannot determine the lockfile version

docs/design/yarn-v1.md
taylormadore: It's another CLI option that is not mentioned in their documentation: https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L84

The PnP feature itself is though: https://classic.yarnpkg.com/en/docs/pnp

docs/design/yarn-v1.md
taylormadore: We want to prefetch all dependencies present in the lockfile (including dev dependencies) and ensure that dev dependencies are reported as such in the SBOM

docs/design/yarn-v1.md
eskultety: So, if an official docs doesn't exist, then each of these options, either CLI or env vars should be properly referenced by a link pointing to their code base.

docs/design/yarn-v1.md
eskultety: it's just a recommendation, but not mandatory IIUC, simply due to the fact it can be re-created. If we check for presence of both, we're deliberately limiting an unlikely, yet valid use case. I'm just mentioning it for correctness, I don't think there's going to be any demand for that and relaxing a rule is always the easier support path than making it stricter.

docs/design/yarn-v1.md
eskultety: Same as above, if we're dealing with internal-only option (at this point it doesn't make any sense to request a change in the docs from upstream) we should link them in the respective code base to clear up any potential confusion.

docs/design/yarn-v1.md
eskultety: @taylormadore where do we stand with this one? Is it viable or not?

eskultety: Still disagree, that thread IMO still hasn't provided any legitimate argument on why it should be a good practice over just giving credit/mentioning it once.

taylormadore: I did a very simple test last week and it worked, but I'd say that my confidence isn't high (at the moment) without digging a little deeper

docs/design/yarn-v1.md
taylormadore: I agree, but I do see value in clearly noting these as out-of-scope in the initial implementation. I'd probably move these to the proposed "decison/outcome" section of the new design doc format as limitations rather than follow-ups. How does that sound?

brunoapimentel: FWIW, I renamed the similar section in the Bundler design to "out of scope".

taylormadore:
Since V2:
- Added some links to the Yarn codebase for undocumented CLI options and environment variables
- Added a small section describing how undocumented Yarn environment variables work

docs/design/yarn-v1.md
eskultety: > Just to make sure I understand: are you proposing we not fail the prefetch due to an incorrect checksum and only report it?

Disregard please, double checked and based on the description the option seems to be effectively mutually exclusive with `frozen-lockfile`

docs/design/yarn-v1.md
eskultety: Seems to be generally accepted, works for me.

docs/design/yarn-v1.md
eskultety: It would be nice to have a PoC on this before we merge this design, since it's implementation approach decision and if we leave it for a follow-up, not that big of a deal, but we'll have to post an update design document for this one in that case.

taylormadore: @eskultety, @brunoapimentel, and I had a conversation about this and agreed to proceed with pyarn in the initial design/implementation.

I will file an issue to design and implement lockfile parsing using Yarn itself, but it will be handled as a follow-up after the initial implementation is complete. Swapping out pyarn for Yarn to handle the parsing should not impact the overall design of this feature and should be more-or-less a drop-in replacement for pyarn.

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,135 @@
...
+
+To enable the offline mirror, the following settings need to be applied either via .yarnrc or from environment variables:
+
+    yarn-offline-mirror <absolute path to the request output directory>
+    yarn-offline-mirror-pruning false
+
+    YARN_YARN_OFFLINE_MIRROR=<absolute path to the request output directory>
+    YARN_YARN_OFFLINE_MIRROR_PRUNING=false
+
+### Installing Yarn (via Corepack)
```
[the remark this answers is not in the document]
taylormadore: We'll definitely add this to the documentation story :+1:

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Project Configuration
+The [package.json](https://classic.yarnpkg.com/en/docs/package-json) and [yarn.lock](https://classic.yarnpkg.com/en/docs/yarn-lock)
+configuration files must be present in order to process the request.
+
+The node_modules directory should be absent. I'm not certain at the moment how to guard against a user-specified modulesFolder at build-time.
+
+Any additional configuration settings in the repository via [.yarnrc](https://classic.yarnpkg.com/en/docs/yarnrc) or .npmrc
+files should be ignored (at least initially) for the prefetch by specifying ```--no-default-rc``` for the yarn install command.
+We never previously honored user-defined settings with Cachito.
```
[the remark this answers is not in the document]
taylormadore: I added a note for this under the scoping for documentation and will file a follow-up issue to fix it later

docs/design/yarn-v1.md
taylormadore: References added :+1:

docs/design/yarn-v1.md
taylormadore: I dropped this section and just added a quick statement to the "References" section at the bottom, like with the Ruby design doc

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,178 @@
...
+The Yarn 1.x CLI doesn't provide enough information to generate an SBOM, so the yarn.lock and package.json files will need to
+be parsed in order to gather the necessary data. The code used to process these files will need to be
+[imported from or inspired by Cachito](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py)
+and refactored/improved.
+
+#### Key Points
+ - The yarn.lock file can be parsed with [pyarn](https://github.com/containerbuildsystem/pyarn) like in cachito
+ - Workspaces will need to be [processed separately](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L95) from package.json since they do not appear in the yarn.lock file
+ - Dev dependencies can be [determined similiarly](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L140) to cachito
+ - We will need to ensure that there are no collisions in the names of the archive files being added to the offline mirror
```
brunoapimentel, line 83: IIUC, collisions could happen when:
- there are two file dependencies with the exact same name
- the unlikely case where there are two git repos with the same name, and the exact same commit hash exists in both of them

I believe we only need to have an explicit check for the former?

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,178 @@
...
+The offline mirror in the request output directory will be populated by running the following Yarn command in the project root:
+
+    yarn install --non-interactive --frozen-lockfile --disable-pnp --no-default-rc --ignore-engines
+
+Command: [yarn install](https://classic.yarnpkg.com/en/docs/cli/install)
+
+Options:
+ - [non-interactive](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-non-interactive): Disable interactive prompts
+ - [frozen-lockfile](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-frozen-lockfile): Fail if yarn.lock updates are needed
+ - [disable-pnp](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L84): Disable Plug'n'Play installation
```
brunoapimentel, line 124: I noticed that when we set this flag, Yarn removes the `installConfig: pnp: true` key from the package.json file. Since the idea is to execute the commands in a copy of the source files, we're safe against unwanted modifications.

But I still wonder how we should treat cases where this flag is set. The prefetched content would not be in the right format for a pnp offline install, right? Should we just fail the request with an unsupported error?

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,178 @@
...
+
+### SBOM Generation
+The Yarn 1.x CLI doesn't provide enough information to generate an SBOM, so the yarn.lock and package.json files will need to
+be parsed in order to gather the necessary data. The code used to process these files will need to be
+[imported from or inspired by Cachito](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py)
+and refactored/improved.
+
+#### Key Points
+ - The yarn.lock file can be parsed with [pyarn](https://github.com/containerbuildsystem/pyarn) like in cachito
+ - Workspaces will need to be [processed separately](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L95) from package.json since they do not appear in the yarn.lock file
```
brunoapimentel, line 81: Does the offline install still work in this case? Did you test it in a repo you can link here?

docs/design/yarn-v1.md
brunoapimentel: Taylor correctly pointed out that a registry dependency could be spoofed if the user declared a file dependency that would result in the same filename when prefetched:

```
{
  "dependencies": {
    "chai": "4.2.0",
    "notchai": "https://some.host.com/archive/chai-4.2.0.tgz",
  }
}
```

So the attack surface is bigger than I expected. Still, we would report all dependencies properly in the SBOM, and the build would very likely fail, so I wonder if it is enough to simply document this limitation for now.

docs/design/yarn-v1.md
taylormadore: With workspaces? I think this is more of an SBOM reporting issue rather than a build issue since the workspace deps will be in the repository and not in the offline cache.

You can use this testing repo to see an example: https://github.com/cachito-testing/cachito-yarn-workspaces

brunoapimentel: I didn't realize that all workspace dependencies would end up in the main yarn.lock file. So I think we're good here.

docs/design/yarn-v1.md
taylormadore: > is pnp the type of installation in Yarn v1 where you check in the whole cache (or even node_modules to your repo?

It appears to be a way to avoid generating the node_modules directory for performance purposes. A pnp project will still try to fetch all of the dependencies from the cache/offline-mirror/internet

taylormadore: > Should we just fail the request with an unsupported error?

I'm leaning this way now because I tested with a slightly different project this time and saw an appearance of our old (in this case undocumented) friend, unplugged packages. Apparently this is only possible with pnp enabled.

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,181 @@
...
+ - Dev dependencies can be [determined similiarly](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L140) to cachito
+ - We will need to ensure that there are no collisions in the names of the archive files being added to the offline mirror
+
+#### PURL Examples
+[NPM PURL specification](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#npm)
+
+##### Registry Package
+```txt
+pkg:npm/@optional-scope/package-name@1.0.0
```
brunoapimentel, line 92: Does the `@` belonging to the scope need to be encoded?
```
pkg:npm/%40optional-scope/package-name@1.0.0
```

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,181 @@
...
+
+Yarn also has support for including [NPM environment variables](https://docs.npmjs.com/cli/v10/using-npm/config#environment-variables) prefixed with `NPM_CONFIG_`.
+
+### Project Configuration
+The [package.json](https://classic.yarnpkg.com/en/docs/package-json) and [yarn.lock](https://classic.yarnpkg.com/en/docs/yarn-lock)
+configuration files must be present in order to process the request.
+
+The node_modules directory should be absent. It may not be possible to guard against a user-specified modulesFolder at build-time.
+
+Yarn projects using the [Plug'n'Play (PnP)](https://classic.yarnpkg.com/lang/en/docs/pnp/) feature will not be supported.
```
brunoapimentel, line 67: Nitpick: I'd add a very brief explanation stating we won't support it because there's no way to validate unplugged deps.

docs/design/yarn-v1.md
brunoapimentel: The scoping is assuming we'll do a full check to verify collisions, closing this.

docs/design/yarn-v1.md
brunoapimentel: We have decied to not support PnP, closing the thread.

docs/design/yarn-v1.md
```diff
@@ -0,0 +1,181 @@
...
+ - Since we're executing Yarn commands directly, process the request in an [isolated temporary directory](https://github.com/containerbuildsystem/cachi2/blob/6953607b6ef52fd3f0bef7059d2c926767b1022b/cachi2/core/resolver.py#L41) to avoid any potential changes to the repository
+ - Determine whether to process the request with yarn v1 or yarn v2+
+ - Read the project files
+   - Read/Require package.json
+   - Read/Require yarn.lock
+     - Ensure yarn.lock is formatted for Yarn 1.x
+ - Verify the repository
+   - Ensure node_modules directory is not present
+   - Ensure that Plug'n'Play (PnP) is not being used
+ - Gather the SBOM Components
```
brunoapimentel, line 155: Are we reporting dev dependencies as part of the initial implementation? If so, do we want a bullet about it here?

taylormadore: Yes, we will do this as part of the initial implementation. Added :+1:

docs/design/yarn-v1.md
taylormadore: Added :+1:

docs/design/yarn-v1.md
taylormadore: Yes, it does. Updated :+1:

brunoapimentel (approved):
:taco: :rocket:

ben-alkov (approved):

PR #586, opened 2024-07-31
Use UBI9 base image for container build

brianwcook:
Change the container build to use UBI9 to support user access to Red Hat entitled content from the subscription service using the subscription manager tool (which wouldn't work with Rocky Linux). Since `createrepo_c ` isn't available from the default RPM repos on UBI, install it from pypi instead. This is tied to the WIP/tech-preview RPM prefetching work.

- [-] Code coverage from testing does not decrease and new code is covered


ben-alkov:
@brianwcook; For your awareness - please address CI failures.

brianwcook:
I don't see the same build failure locally. I re-ran pip-compile with a clean venv to see if it solves the problem.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.9
+# This file is autogenerated by pip-compile with Python 3.12
```
brunoapimentel, line 2: Our base supported version for Cachi2 is still `3.9`. You can regenerate this file easily by doing `make pip-compile`, which will keep the intended Python version

Dockerfile
```diff
@@ -1,19 +1,20 @@
...
-FROM rockylinux9 as base
+FROM ubi as base
 RUN dnf -y install \
     --setopt install_weak_deps=0 \
     --nodocs \
-    createrepo_c \
     git-core \
     python3 \
-    && dnf clean all
+    libffi-devel \
```
eskultety, line 15: Do we need this one as a dependency? The image builds just fine and createrepo_c can be used as well.

Dockerfile
```diff
@@ -1,19 +1,20 @@
-FROM docker.io/library/rockylinux:9@sha256:d7be1c094cc5845ee815d4632fe377514ee6ebcf8efaed6892889657e5ddaaa6 as rockylinux9
+FROM registry.access.redhat.com/ubi9/ubi@sha256:763f30167f92ec2af02bf7f09e75529de66e98f05373b88bef3c631cdcc39ad8 as ubi
```
eskultety, line 1: **commit message:**
Can we improve the commit message so that it explicitly states that the only reason we're switching to it is to enable access to entitled content via subscription manager that would not otherwise be possible via the current Rocky Linux image?

Dockerfile
brianwcook: at some point I saw a build failure complaining about missing libffi, but I cannot reproduce it either so I have removed that package.

requirements.txt
brianwcook: done

Dockerfile
```diff
@@ -1,19 +1,19 @@
...
-FROM rockylinux9 as base
+FROM ubi as base
 RUN dnf -y install \
     --setopt install_weak_deps=0 \
     --nodocs \
-    createrepo_c \
     git-core \
     python3 \
-    && dnf clean all
+    subscription-manager && \
```
eskultety, line 15: You have some trailing whitespaces on this line.

eskultety (commented):
@brianwcook sorry for the delay, had other duties last week, now I'm back on this. Please rebase onto the latest main HEAD to get rid of the merge commit you have, we don't do merge commits here.
I adjusted the PR description ever so slightly to make it completely clear that this change is driven solely by the need to support access to entitled subscription-based content.

Dockerfile
brianwcook: fixed

eskultety (commented):
There was a hiccup in your rebase - you've got some unrelated (already merged) changes (e.g. commits 36b011c5, 1b3bd800) in your latest revision.

brianwcook:
alright how 'bout now?

brunoapimentel (approved):
This LGTM, I only wonder if we should add a note to the commit message that the need to install `subscription-manager` comes from https://github.com/konflux-ci/build-definitions/pull/1205.

I don't think it will be very clear to people even after #580 comes in.

brianwcook:
> This LGTM, I only wonder if we should add a note to the commit message that the need to install `subscription-manager` comes from [konflux-ci/build-definitions#1205](https://github.com/konflux-ci/build-definitions/pull/1205).
> 
> I don't think it will be very clear to people even after #580 comes in.

added your comment and rebased again.

PR #590, opened 2024-08-01
[ISV-5094] cachi2 PNC package manager design

kosciCZ:

### DIFF
--- a/docs/design/pnc.md
+++ b/docs/design/pnc.md
@@ -9,7 +9,81 @@ Contents:
 ## PNC
 Also known as Project NewCastle - [open-source project](https://github.com/project-ncl/pnc) for managing, executing, 
 and tracking cross-platform builds. The part that is important for cachi2 integration is the exposed API that provides
-information about builds and their artifacts, as well as the means to download those artifacts.
+information about builds and their artifacts, as well as the means to download those artifacts. The API does not require
+authentication.
+
+### Relevant endpoints and their payloads
+#### /artifacts/{id}
+This endpoint takes an `id` of an artifact from the `fetch-artifacts-pnc.yaml` file. It will respond with
+various information about the specified artifact. The API response is [documented here](https://github.com/project-ncl/pnc-api/blob/5b35ad6fe22769510f60b0025752dc6c954c9734/src/main/java/org/jboss/pnc/api/repositorydriver/dto/RepositoryArtifact.java).
+Below is an example response, some irrelevant fields left out:
+```json
+{
+  "id": "1234",
+  "identifier": "org.example:package:zip:1.2.3.org-1",
+  "purl": "pkg:maven/org.example/package@1.2.3.org-1?type=zip",
+  "artifactQuality": "NEW",
+  "buildCategory": "STANDARD",
+  "md5": "<md5 hash>",
+  "sha1": "<sha1 hash>",
+  "sha256": "<sha256 hash>",
+  "filename": "package-1.2.3.org-1.zip",
+  "deployPath": "<path>",
+  "importDate": null,
+  "originUrl": null,
+  "size": 12345,
+  "deployUrl": "<local url>",
+  "publicUrl": "<public url>",
+  "creationTime": null,
+  "modificationTime": "2023-08-14T15:04:44.388Z",
+  "qualityLevelReason": null,
+  "targetRepository": {
+    "id": "456",
+    "temporaryRepo": false,
+    "identifier": "indy-maven",
+    "repositoryType": "MAVEN",
+    "repositoryPath": "/api/content/maven/hosted/pnc-builds/"
+  },
+  "build": {
+    "id": "ABCD123",
+    "submitTime": "2023-08-14T14:43:16.097Z",
+    "startTime": "2023-08-14T14:43:16.179Z",
+    "endTime": "2023-08-14T15:04:44.494Z",
+    "progress": "FINISHED",
+    "status": "SUCCESS",
+    "buildContentId": "build-ABCD123",
+    "temporaryBuild": false,
+    "alignmentPreference": null,
+    "scmUrl": "<source repo>",
+    "scmRevision": "<hash>",
+    "scmTag": "1.2.3.org-1-ABCD123",
+    "buildOutputChecksum": "<hash>",
+    "lastUpdateTime": "2023-08-14T15:05:02.382Z",
+    "scmBuildConfigRevision": null,
+    "scmBuildConfigRevisionInternal": null,
+    "project": {
+      "id": "5",
+      "name": "package-parent",
+      "description": null,
+      "issueTrackerUrl": null,
+      "projectUrl": null,
+      "engineeringTeam": null,
+      "technicalLeader": null
+    },
+    "attributes": {
+      "BREW_BUILD_VERSION": "1.2.3.org-1",
+      "BUILD_OUTPUT_OK": "false",
+      "BREW_BUILD_NAME": "org.example:package-parent"
+    },
+    "noRebuildCause": null
+  },
+  "creationUser": null
+}
+```
+
+#### /builds/{id}/scm-archive
+This endpoint will return the source build archive as a tarball (.tar.gz). Build `id` can be located in
+the  `fetch-artifacts-pnc.yaml` file.
 
 
 ## Overview of the current implementation in OSBS
@@ -50,7 +124,7 @@ Currently, the url of the PNC instance is configured as a part of OSBS config.
 
 [atomic_reactor/plugins/fetch_maven_artifacts.py](https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/plugins/fetch_maven_artifacts.py)
 1) OSBS loads the `fetch-artifacts-pnc.yaml`
-2) Information about individual artifacts is fetched from PNC api at `/artifacts/{id}` endpoint
+2) Information about individual artifacts is fetched from PNC API at `/artifacts/{id}` endpoint ([docs](https://github.com/project-ncl/pnc-api/blob/5b35ad6fe22769510f60b0025752dc6c954c9734/src/main/java/org/jboss/pnc/api/repositorydriver/dto/RepositoryArtifact.java))
    - `publicUrl` key is used to get download url
    - supported checksums are saved
    - build ids are saved
@@ -92,11 +166,18 @@ As mentioned in [missing features](#missing-features), the purl needs to specify
   from PNC `/artifacts/{id}` endpoint response.
 - `type` & `classifier` - solution pending
 
+Here's an example purl that should be generated:
+```
+pkg:maven/org.example/package@1.2.3.org-1?type=zip&classifier=dist&repository_url=https://repo.maven.apache.org/maven2
+```
+
 #### Relevant configuration for the build
 No environment variables need to be set.
 
-#### Injecting files to an expected location
-The actual artifacts should be injected into `artifacts/` path in the repository. 
+#### Expected location of artifacts
+The current implementation expects artifacts in the `artifacts/` path in the root of repository.
+Due to cachi2's usual design patterns, the artifacts will remain in cachi2's default output directory
+(`./cachi2-output`), and users will be required to act accordingly.
 
 
 ### Summary
--- a/docs/design/pnc.md
+++ b/docs/design/pnc.md
@@ -1,0 +1,113 @@
+**# Design document for PNC package manager
+
+Contents:
+
+1. [PNC](#PNC)
+2. [Current implementation overview (OSBS)](#overview-of-the-current-implementation-in-osbs)
+3. [Design for the Cachi2 implementation](#design-for-the-implementation-in-cachi2)
+
+## PNC
+
+### Glossary
+- **PNC**: Project NewCastle - [open-source project](https://github.com/project-ncl/pnc) for managing, executing, 
+  and tracking cross-platform builds.
+
+
+### Repository configuration 
+Projects that currently use OSBS to fetch sources from PNC do so through a file named `fetch-artifacts-pnc.yaml`
+([docs](https://osbs.readthedocs.io/en/osbs_ocp3/users.html?highlight=fetch#fetch-artifacts-pnc-yaml)).
+Example of a `fetch-artifacts-pnc.yaml` file:
+
+```yaml
+metadata:
+  # this object allows additional parameters, you can put any metadata here
+  author: shadowman
+builds:
+  # all artifacts are grouped by builds to keep track of their sources
+  - build_id: '1234' # build id must be string
+    artifacts:
+      # list of artifacts to fetch, artifacts are fetched from PNC using their IDs
+      - id : '12345' # artifact id must be string
+        # the target can just be a filename or path+filename
+        target: test/rhba-common-7.10.0.redhat-00004.pom
+      - id: '12346'
+        target: prod/rhba-common-7.10.0.redhat-00004-dist.zip
+  - build_id: '1235'
+    artifacts:
+      - id: '12354'
+        target: test/client-patcher-7.10.0.redhat-00004.jar
+      - id: '12355'
+        target: prod/rhdm-7.10.0.redhat-00004-update.zip
+```
+
+These artifacts are fetched to `artifacts/<target>` path at the root of the repository.
+
+### PNC instance configuration
+Currently, the url of the PNC instance is configured as a part of OSBS config. 
+
+## Overview of the current implementation in OSBS
+
+[atomic_reactor/plugins/fetch_maven_artifacts.py](https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/plugins/fetch_maven_artifacts.py)
+1) OSBS loads the `fetch-artifacts-pnc.yaml`
+2) Information about individual artifacts is fetched from PNC api at `/artifacts/{id}` endpoint
+   - `publicUrl` key is used to get download url
+   - supported checksums are saved
+   - build ids are saved
+3) Download queue is formed containing above info for each artifact
+4) SBOM component info for each artifact is fetched from PNC @ `/artifacts/{id}`, `purl` key
+5) Queue is executed and each artifact is downloaded, checksum verified
+6) SBOM is compiled, including PNC components from step 4)
+
+
+### Missing features
+
+PNC SBOM components don't have the most complete purl, e.g.:
+`pkg:maven/org.kie.trustyai/explainability-service@999.0.0.managedsvc-redhat-00001?type=zip`
+
+According to [purl spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#maven), `classfier` and
+`type` parameters should be also defined, method is currently pending depending on if the artifacts are single components,
+or products that contain other components. Purl should also specify a `repository_url` parameter.
+
+
+
+## Design for the implementation in Cachi2
+
+
+### Providing the content for the hermetic build
+
+#### PNC config
+Existing `fetch-artifacts-pnc.yaml` file ([docs](https://osbs.readthedocs.io/en/osbs_ocp3/users.html?highlight=fetch#fetch-artifacts-pnc-yaml))
+will be used as configuration. It needs to be extended to also include PNC instance url.
+
+#### Pre-fetch PNC dependencies
+Cachi2 will use the `artifacts` key specified in the `fetch-artifacts-pnc.yaml` file to fetch info about given artifacts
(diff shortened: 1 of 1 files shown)


kosciCZ:
@brunoapimentel @chmeliik @eskultety Here's the initial draft of the PNC package manager from KONFLUX-3403, that you've asked for in our last meeting. Please review it and I'll be happy to work in your feedback. Thanks!

cc @allda

brunoapimentel:
Overall comment: you don't need to follow exactly the structure of the other designs. I think we are mainly separating the design into two distinct parts:
- A quick overview of the package manager/tool/ecosystem to make folks that are not familiar with it up to speed. It can be brief, contain a lot of links to official docs, and only touch the parts that are relevant for Cachi2
- The implementation design by itself.

(we are still considering an "official" format for designs, so this might change a little, but don't worry about it now)

docs/design/pnc.md
```diff
@@ -0,0 +1,113 @@
...
+## Design for the implementation in Cachi2
+
+
+### Providing the content for the hermetic build
+
+#### PNC config
+Existing `fetch-artifacts-pnc.yaml` file ([docs](https://osbs.readthedocs.io/en/osbs_ocp3/users.html?highlight=fetch#fetch-artifacts-pnc-yaml))
+will be used as configuration. It needs to be extended to also include PNC instance url.
+
+#### Pre-fetch PNC dependencies
```
brunoapimentel, line 82: I feel like this section should be standalone and separate from "Providing content for the hermetic build". This is because, during the build, the prefetch has already happened. Those are two distinct phases: prefetch -> build.

(and remember that we consider that the `buildah bud`/`podman build` command is executed with the `--network none` flag, so there's absolutely no network access).

docs/design/pnc.md
```diff
@@ -0,0 +1,113 @@
...
+3. [Design for the Cachi2 implementation](#design-for-the-implementation-in-cachi2)
+
+## PNC
+
+### Glossary
+- **PNC**: Project NewCastle - [open-source project](https://github.com/project-ncl/pnc) for managing, executing, 
+  and tracking cross-platform builds.
+
+
+### Repository configuration
```
brunoapimentel, line 16: This part seems to be OSBS specific, so it is probably better fit into the `Overview of the current implementation in OSBS` section.

docs/design/pnc.md
```diff
@@ -0,0 +1,113 @@
...
+      - id: '12354'
+        target: test/client-patcher-7.10.0.redhat-00004.jar
+      - id: '12355'
+        target: prod/rhdm-7.10.0.redhat-00004-update.zip
+```
+
+These artifacts are fetched to `artifacts/<target>` path at the root of the repository.
+
+### PNC instance configuration
+Currently, the url of the PNC instance is configured as a part of OSBS config.
```
brunoapimentel, line 46: This makes me wonder how are we going to implement integration tests for this. Is it possible to spin a basic PNC instance using something like `docker-compose`? All of our integration tests currently run on Github Actions, so we can't access a private instance.

docs/design/pnc.md
```diff
@@ -0,0 +1,113 @@
...
+As mentioned in [missing features](#missing-features), the purl needs to specify more parameters
+- `repository_url` - obtainable by using host part of `publicUrl` and the path in `targetRepository.repositoryPath`
+  from PNC `/artifacts/{id}` endpoint response.
+- `type` & `classifier` - solution pending
+
+#### Relevant configuration for the build
+No environment variables need to be set.
+
+#### Injecting files to an expected location
+The actual artifacts should be injected into `artifacts/` path in the repository.
```
brunoapimentel, line 101: It is Cachi2's philosophy to avoid modifying the source files at all costs. That's why we have a separate step after [fetch-deps](https://github.com/containerbuildsystem/cachi2/blob/main/docs/usage.md#pre-fetch-dependencies) called [inject-files](https://github.com/containerbuildsystem/cachi2/blob/main/docs/usage.md#inject-project-files). The latter is used to modify source files as needed to make the build work hermetically.

But if I understood this correctly, this is simply placing the prefetched dependencies in a specific folder, right? It doesn't quite seem to be the use case for `inject-files`.

Will maven/gradle/build tool necessarily look for the content in `./artifacts`? Is this configurable somehow?

docs/design/pnc.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+
+## Design for the implementation in Cachi2
+
+
+### Providing the content for the hermetic build
+
+#### PNC config
+Existing `fetch-artifacts-pnc.yaml` file ([docs](https://osbs.readthedocs.io/en/osbs_ocp3/users.html?highlight=fetch#fetch-artifacts-pnc-yaml))
```
brunoapimentel, line 79: Will this file be commited to the source repo? A Cachi2 requests points to a single folder in the file system that contains a Git repository:

```bash
$ cachi2 fetch-deps --source local/path/to/my/repo pnc
```

So usually the needed configution is placed in the repo itself (for example, `package.lock` for npm, `rpms.lock.yaml` for dnf, `go.mod` for golang, etc).

brunoapimentel:
A suggestion that could aid the design is to implement a PoC. Note that this is not required for the design to be approved, so please only do it if you feel like it would be helpful. I do think that it usually helps to clarify a lot of the design details, and even antecipate some of the issues that can happen.

Here's the [PoC](https://github.com/slimreaper35/cachi2/tree/ruby-poc) created for the Ruby design, and the corresponding [test repo](https://github.com/brunoapimentel/cachi2-rubygems/) we are using alongside with it.

docs/design/pnc.md
kosciCZ: Yes, that is currently the case to the best of my knowledge, the `fetch-artifacts-pnc.yaml` is in the root of the source repository, so I don't think that's any different than the current way cachi2 fetches configuration

docs/design/pnc.md
kosciCZ: Looking at the PNC repositories, it does not look like it doesn't have dockerfiles/compose out of the box, and setting it all up could prove difficult. But since it's just an API, we could probably have a small mock flask application or something that would serve mock responses.

brunoapimentel: You're right, but I wonder how much value an integration test with a mocked PNC would bring. Maybe we can consider some internal testing for this, then? I'm just afraid there could be cases we would not cover with a mocked API.

docs/design/pnc.md
chmeliik: > Will maven/gradle/build tool necessarily look for the content in ./artifacts? Is this configurable somehow?

Do they look there at all? I kind of assumed ./artifacts was just a made up place where OSBS used to place the downloaded archives, and OSBS users then had to explicitly use the stuff from ./artifacts in their Dockerfile

kosciCZ: I think it's pretty arbitrary. I don't think it is 100% necessary, but on the other hand, it will ease the migration for those middleware source repos, which are the targeted customer of this change, as their Dockerfiles are already pointing at that directory. 

Let me know if this is an issue, or something that would go against how cachi2 is designed to work.

docs/design/pnc.md
kosciCZ: To the best of my knowledge, we just need a couple of API responses, and a working url to download the files referenced in those responses. In my opinion, setting up an arbitrary PNC instance just for this is an overkill, but I could see adding some test artifacts to an existing internal PNC instance.

As for internal testing, could you please expand on how that would work? I understand that it would be e.g. a repo in gitlab with a  `fetch-artifacts-pnc.yaml` file poiting to some expected artifacts, but how would we tie that back in with the existing integration testing in this repository via github actions?

docs/design/pnc.md
```diff
@@ -0,0 +1,113 @@
...
+Contents:
+
+1. [PNC](#PNC)
+2. [Current implementation overview (OSBS)](#overview-of-the-current-implementation-in-osbs)
+3. [Design for the Cachi2 implementation](#design-for-the-implementation-in-cachi2)
+
+## PNC
+
+### Glossary
+- **PNC**: Project NewCastle - [open-source project](https://github.com/project-ncl/pnc) for managing, executing,
```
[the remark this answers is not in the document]
kosciCZ: Expanded a bit, but I only have limited knowledge of PNC, I mainly familiarized myself with the parts important for cachi2 integration

docs/design/pnc.md
kosciCZ: Moved

docs/design/pnc.md
kosciCZ: Moved

kosciCZ:
@brunoapimentel thanks for the suggestion. In my opinion, implementing a PoC is an overkill in this case, as the PoC is basically already the work we want to do, so we might as well do it properly the first time around. However, using a test repo is a good idea, even if it should be manual testing during the development (we can figure out the automated integration testing in that thread above)

docs/design/pnc.md
eskultety: > but how would we tie that back in with the existing integration testing in this repository via github actions?

One of the options would IMO be self-hosted runners: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners
I don't have experience with github, but if it's anything like gitlab-runners then one just needs to set up an agent in the private environment and add it to the list of CI runners, in this case - actions. The beauty of at least gitlab runner is that it's not a push workflow, but a pull workflow, so even runners behind a firewall are able to report back provided the correct ports for egress are opened on the client side.
In practical terms, I can imagine using your already deployed instance for testing this way.

kosciCZ: @eskultety That is a really good idea! Who would own and maintain the self-hosted github runner? To the best of my knowledge and some quick looking around, we don't currently have such github runner instance available internally.

eskultety: > Who would own and maintain the self-hosted github runner?

From community POV that is usually the responsibility of people adding the support for a particular backend/driver as those are the people who actually need it :) ; but since the contributor may not always have direct involvement with the backend they're trying to add support for then the next obvious choice would IMO be whoever maintains the current internal deployment as they clearly understand the architecture the best and so it makes sense if a small infra adjustment is made such that a self hosted runner agent is created and wired so that it can relay jobs to the PNC deployment.

docs/design/pnc.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Design for the implementation in Cachi2
+
+### PNC config
+Existing `fetch-artifacts-pnc.yaml` file ([docs](https://osbs.readthedocs.io/en/osbs_ocp3/users.html?highlight=fetch#fetch-artifacts-pnc-yaml))
+will be used as configuration. It needs to be extended to also include PNC instance url.
+
+### Pre-fetch PNC dependencies
+Cachi2 will use the `artifacts` key specified in the `fetch-artifacts-pnc.yaml` file to fetch info about given artifacts
+from PNC (`/artifacts/{id}` endpoint), including url to download the artifact, checksums to verify integrity, and the purl.
```
brunoapimentel, line 82: I think it would make it much clearer if we added either docs to the PNC API, or a brief section describing the endpoints and their payloads.

docs/design/pnc.md
```diff
@@ -0,0 +1,111 @@
...
+The artifacts should be fetched into a separate subdirectory in cachi2 output folder, and into the target path.
+
+#### Fetching sources
+Cachi2 will use the `build_id` keys to fetch build sources from PNC (`/builds/{id}/scm-archive` endpoint). These sources
+should be fetched into a separate subdirectory in cachi2 output folder, different from the artifacts.
+
+### Generating the SBOM
+As mentioned in [missing features](#missing-features), the purl needs to specify more parameters
```
brunoapimentel, line 90: Can we add an example of how the final PURL should look like?

docs/design/pnc.md
```diff
@@ -0,0 +1,111 @@
...
+
+According to [purl spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#maven), `classfier` and
+`type` parameters should be also defined, method is currently pending depending on if the artifacts are single components,
+or products that contain other components. Purl should also specify a `repository_url` parameter.
+
+
+
+## Design for the implementation in Cachi2
+
+### PNC config
```
brunoapimentel, line 76: Will we need any type of authentication to consume the PNC endpoints?

docs/design/pnc.md
brunoapimentel: Cachi2 always places the downloaded files under `{output-folder}/deps/{package_manager}`, so we should probably respect that. I don't think `inject-files` is meant to simply move these files to a new location either.

It feels to me this is out of scope for Cachi2, but since this is a specific problem for the Konflux use-case, we could potentially work around this in the Konflux pipeline.

kosciCZ: Agreed. Will edit so it is clear those files will **not** be in the `/artifacts` location, but instead will use the default cachi2 location.

docs/design/pnc.md
kosciCZ: Added an example response and description of relevant endpoints into the PNC section.

docs/design/pnc.md
kosciCZ: Added

docs/design/pnc.md
kosciCZ: We've had a meeting with @brunoapimentel today where we discussed this, and arrived at the conclusion that the best available option might be extensive unit testing. The self hosted github runner seem not viable from a maintenance perspective (I really doubt your team, my team or the PNC folks will have any interest in doing that) and also possibly a security perspective. 

We're still exploring if e2e konflux tests might be also viable for this, Bruno took a follow-up item on that. But, the plan at this moment is to proceed with extensive unit tests (as in testing the manager as a whole and mocking the API responses).

docs/design/pnc.md
```diff
@@ -0,0 +1,192 @@
...
+pkg:maven/org.example/package@1.2.3.org-1?type=zip&classifier=dist&repository_url=https://repo.maven.apache.org/maven2
+```
+
+#### Relevant configuration for the build
+No environment variables need to be set.
+
+#### Expected location of artifacts
+The current implementation expects artifacts in the `artifacts/` path in the root of repository.
+Due to cachi2's usual design patterns, the artifacts will remain in cachi2's default output directory
+(`./cachi2-output`), and users will be required to act accordingly.
```
twaugh, line 180: Could we at least create a symlink `artifacts -> .cachi2-output`?

brunoapimentel: Since this is a Konflux-specific issue, I believe we could treat it on the pipeline level, so possibly a suggestion for [build-definitions](https://github.com/konflux-ci/build-definitions).

docs/design/pnc.md
```diff
@@ -0,0 +1,192 @@
...
+### Pre-fetch PNC dependencies
+Cachi2 will use the `artifacts` key specified in the `fetch-artifacts-pnc.yaml` file to fetch info about given artifacts
+from PNC (`/artifacts/{id}` endpoint), including url to download the artifact, checksums to verify integrity, and the purl.
+The artifacts should be fetched into a separate subdirectory in cachi2 output folder, and into the target path.
+
+#### Fetching sources
+Cachi2 will use the `build_id` keys to fetch build sources from PNC (`/builds/{id}/scm-archive` endpoint). These sources
+should be fetched into a separate subdirectory in cachi2 output folder, different from the artifacts.
+
+### Generating the SBOM
```
brunoapimentel, line 163: I believe we can also expand a little on how we're getting the data from the `/artifacts/{id}` endpoint to generate the SBOM component for each artifact. A SBOM component usually looks like this:

```
{
  "name": "urllib3",
  "purl": "pkg:pypi/urllib3@2.2.2",
  "version": "2.2.2",
  "properties": [
    {
      "name": "cachi2:found_by",
      "value": "cachi2"
    }
  ],
  "type": "library"
}
```

So where are we getting the name and version from? Is it from the PNC's PURL? Or some other field?

docs/design/pnc.md
```diff
@@ -0,0 +1,192 @@
...
+As mentioned in [missing features](#missing-features), the purl needs to specify more parameters
+- `repository_url` - obtainable by using host part of `publicUrl` and the path in `targetRepository.repositoryPath`
+  from PNC `/artifacts/{id}` endpoint response.
+- `type` & `classifier` - solution pending
```
brunoapimentel, line 167: What is blocking us from figuring this point out? Can the solution be released without these two qualifiers?

docs/design/pnc.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Design for the implementation in Cachi2
+
+### PNC config
+Existing `fetch-artifacts-pnc.yaml` file ([docs](https://osbs.readthedocs.io/en/osbs_ocp3/users.html?highlight=fetch#fetch-artifacts-pnc-yaml))
+will be used as configuration. It needs to be extended to also include PNC instance url.
+
+### Pre-fetch PNC dependencies
+Cachi2 will use the `artifacts` key specified in the `fetch-artifacts-pnc.yaml` file to fetch info about given artifacts
+from PNC (`/artifacts/{id}` endpoint), including url to download the artifact, checksums to verify integrity, and the purl.
```
brunoapimentel, line 156: > checksums to verify integrity

Maybe we could add a H4 heading under this one to expand on how the checksum validation will work. I am assuming that:

- the checksums are always present in the API response, and if they're not, we'll report them as missing in the SBOM for that specific artifact
- if there's a checksum mismatch, we will fail the request straight away
- we can rely on a single algorithm type (since the API provides three different ones), probably sha-256?

(the first two assumptions derive from what Cachi2 is doing for other package managers)

docs/design/pnc.md
taylormadore: Yeah, I would prefer this not be handled in cachi2 if possible. As a general principle, I think we should strive to avoid modifying the user repository unless absolutely necessary for our core functionality

docs/design/pnc.md
taylormadore: Since there are no user-defined checksums in fetch-artifacts-pnc.yaml, shouldn't all PNC SBOM components have the property `cachi2:missing_hash:in_file` regardless of whether the API response has them? Also, if the API response doesn't have them we should probably just fail since we cannot verify what was downloaded

kosciCZ: @brunoapimentel That is a correct assumption. However, based on what I saw in cachi2 code, you seem to be verifying all supported checksum algos, so I think for consistency, we should be doing all of them (even though sha-256 is clearly the strongest one)

@taylormadore I will contact prodsec to check, but I assume that the API response can be trusted.

docs/design/pnc.md
kosciCZ: Yes, the purl will need to be parsed to extract the relevant fields.

docs/design/pnc.md
brunoapimentel: Taylor has a good point, that I overlooked in my first comment. We differentiate between server-provided and user-provided checksums, and the `cachi2:missing_hash:in_file` is only used for the latter.

I was wondering if we should expand the `fetch-artifacts-pnc.yaml` definition to allow the inclusion of checksums there, but then we would push it towards being more of a lockfile, which I'm not sure if it makes sense, considering the dependency we have on the PNC API to get more data about the artifact.

Another way to go is to simply consider that the PNC package manager will not verify user checksums, and properly document it.

brunoapimentel: >  I will contact prodsec to check, but I assume that the API response can be trusted.
I think we should look at this from the perspective that it is a generic PNC support, in a sense that it extrapolates our immediate internal use-case. This means that any PNC instance could be used during the prefetching.

kosciCZ: >I think we should look at this from the perspective that it is a generic PNC support, in a sense that it extrapolates our immediate internal use-case. This means that any PNC instance could be used during the prefetching.

Okay, makes sense, will incorporate into the design.

@twaugh can you weigh in on the question of expanding the file with checksums for each artifact, from the perspective of users that will be migrating to konflux?

docs/design/pnc.md
kosciCZ: We're currently discussing what's the best way to determine those. The easy way is to look at the file extension, or a deeper inspection. It does not block the solution I think, since we still have time before we have to implement this part.

docs/design/pnc.md
kosciCZ: >I think we should look at this from the perspective that it is a generic PNC support, in a sense that it extrapolates our immediate internal use-case. This means that any PNC instance could be used during the prefetching.

Prodsec view on this is that `cachi2:missing_hash:in_file` is reasonable to do. They did ask about if it would be possible to configure cachi2 to trust PNC as a source of checksums, off by default?

PR #592, opened 2024-08-09
cachi2 coverage

slimreaper35:

### DIFF
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -25,37 +25,21 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
+
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install tox tox-gh-actions
+        pip install tox
+        pip install tox-gh-actions
+
     - name: Test with tox
       run: tox
-    - name: Run coveralls-python
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        COVERALLS_FLAG_NAME: python-${{ matrix.python-version }}
-        COVERALLS_PARALLEL: true
-      run: |
-        pip3 install --upgrade pip
-        pip3 install --upgrade setuptools
-        pip3 install --upgrade coverage[toml]
-        pip3 install --upgrade coveralls
-        coveralls --service=github
 
-  coveralls-finish:
-    name: Finish coveralls-python
-    needs: tests
-    runs-on: ubuntu-latest
-    steps:
-      - name: Finished
-        run: |
-          pip3 install --upgrade pip
-          pip3 install --upgrade setuptools
-          pip3 install --upgrade coveralls
-          coveralls --finish --service=github
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Upload coverage reports to Codecov
+      if: matrix.python-version == '3.12'
+      uses: codecov/codecov-action@v4
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
 
   linters:
     name: Linters
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,0 +1,12 @@
+codecov:
+  notify:
+    wait_for_ci: no
+
+coverage:
+  status:
+    patch: off
+    project:
+      default:
+        target: 90%
+
+comment: off
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Cachi2
 
-[![coverage][cachi2-coveralls-badge]][cachi2-coveralls]
+[![coverage][cachi2-coverage-badge]][cachi2-coverage-status]
 [![container][cachi2-container-status]][cachi2-container]
 
 Cachi2 is a CLI tool that pre-fetches your project's dependencies to aid in making your build process
@@ -406,8 +406,8 @@ See [docs/yarn.md](docs/yarn.md) for more details.
 Cachi2 was derived (but is not a direct fork) from [Cachito](https://github.com/containerbuildsystem/cachito) and is
 still in early development phase.
 
-[cachi2-coveralls]: https://coveralls.io/github/containerbuildsystem/cachi2?branch=main
-[cachi2-coveralls-badge]: https://coveralls.io/repos/github/containerbuildsystem/cachi2/badge.svg?branch=main
+[cachi2-coverage-badge]: https://codecov.io/github/containerbuildsystem/cachi2/graph/badge.svg?token=VJKRTZQBMY
+[cachi2-coverage-status]: https://codecov.io/github/containerbuildsystem/cachi2
 [cachi2-container]: https://quay.io/repository/redhat-appstudio/cachi2
 [cachi2-container-status]: https://quay.io/repository/redhat-appstudio/cachi2/status
 [cachi2-releases]: https://github.com/containerbuildsystem/cachi2/releases


ben-alkov:
LGTM so far. Nice work :taco:

slimreaper35:
Third commit:

Please look at the [top-level-section](https://docs.codecov.com/docs/codecovyml-reference) in the docs for configuring _codecov_. Also, there is a page about sending comments in pull requests. I would say the most relevant are:
- [behaviour](https://docs.codecov.com/docs/pull-request-comments#behavior)
- [requiring-changes](https://docs.codecov.com/docs/pull-request-comments#requiring-changes)

codecov.yaml
```diff
@@ -0,0 +1,13 @@
+codecov:
+  notify:
+    wait_for_ci: no
+
+coverage:
+  status:
+    project:
+      default:
+        # TODO
+        threshold: 1%
```
eskultety, line 10: We probably want a bit more slack in the threshold, i.e. IIRC we intend to maintain a 90% coverage ratio and based on the current ~97-98, I'd say 5% is reasonable.

codecov.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+comment:
+  layout: "diff, files"
```
eskultety, line 13: I think we want `condensed_` versions of ^these, we definitely don't want much verbosity from codecov.

codecov.yaml
```diff
@@ -0,0 +1,13 @@
...
+    wait_for_ci: no
+
+coverage:
+  status:
+    project:
+      default:
+        # TODO
+        threshold: 1%
+
+comment:
```
eskultety, line 12: I think we want codecov to post a comment only if necessary, so some handling of `require_changes` is going to be needed, I'd suggest: `require_changes: "coverage_drop AND uncovered_patch" `

slimreaper35:
> Looks reasonable, @slimreaper35 do you have a preview of the intended settings you have in mind showcased somewhere in practice?

Yes, I was mostly playing around with those layout options, but didn't notice big differences.
https://github.com/slimreaper35/cachi2/pulls?q=is%3Apr+is%3Aclosed

codecov.yaml
ben-alkov: They have a website for checking diffs, so having a comment with statistics and a link is fine.

eskultety: @slimreaper35 looking at your https://github.com/slimreaper35/cachi2/pulls?q=is%3Apr+is%3Aclosed experiments, I don't think `diff` layout provides any value, so having just `files` is IMO sufficient.
One thing that I didn't like is what @ben-alkov mentioned:

> so having a comment with statistics and a link is fine

I don't want to go to some other web page to have a look at which lines are not covered. It would be awesome if there were a layout switch which could render whatever the URL takes you to and have all the necessary information here in our space.

codecov.yaml
slimreaper35: I would prefer OR instead of AND so we will not miss anything.

ben-alkov: Agreed! That's actually how I read it in my pre-coffee haze this morning 😁

eskultety: > I would prefer OR instead of AND so we will not miss anything.

Please don't, it's only going to add to the noise. The reason why I suggested AND is based on [docs](https://docs.codecov.com/docs/pull-request-comments#requiring-changes):

>- "coverage_drop"[project coverage] - Only post comment if coverage drops when comparing HEAD to BASE. This > option respects [project status threshold](https://docs.codecov.com/docs/commit-status#threshold) , if >configured (i.e. drop needs to be bigger than the configured threshold for the comment to be posted).
>- "uncovered_patch" [patch coverage] - Only post comment if the patch has uncovered lines

IIUC if you put OR in there codecov is not only going to complain when overall coverage drops from today's ~97% with a given PR but also if it thinks the PR has uncovered lines in code which is sometimes hard to do and even if it is it may have diminishing result and detrimental impact on overall test readability, we need to be sensible when it comes to testing and judge what needs/doesn't need testing and putting OR in that means we'll potentially enslave ourselves at codecov's mercy, so please reconsider AND since the overall coverage is IMO a much more important factor.

eskultety: Or maybe just `coverage_drop` in case there's a refactor that doesn't add any new lines, but it simply removes a bit too much without an adjustment causing the drop.

slimreaper35: > This option respects [project status threshold](https://docs.codecov.com/docs/commit-status#threshold) , if configured (i.e. drop needs to be bigger than the configured threshold for the comment to be posted).

I am concerned about setting the threshold to 5% and having set the option. We would not see any comments until somebody accidentally removes a few tests :smile:

eskultety: Isn't that the desired behaviour though? IOW only get notified by codecov when a PR would cause the coverage to drop significantly. That said, I can see where you're going, i.e. the threshold is a relative deviation not an absolute number, so theoretically IIUC we could see 2-3 PRs deviating by 4.5% and just like that you drop below our desired 90% coverage. Setting it too conservatively, e.g. 1% is not going to cut it, because that's too low, that will be IMO triggered easily. I wonder if codecov doesn't have an absolute threshold, which we'd set to 90% and we'd never fall below.
Setting also `uncovered_patch` IMO doesn't solve this because IIUC the docs correctly, it only scans lines added, not the existing ones, so if a PR removes a few tests, coverage drops, but since no lines have been added only the existing ones were affected - I don't know what the option would do in that case.

slimreaper35: > I wonder if codecov doesn't have an absolute threshold, which we'd set to 90% and we'd never fall below.

I think it is set in [pyproject.toml](https://github.com/containerbuildsystem/cachi2/blob/main/pyproject.toml#L91)

slimreaper35:
https://github.com/containerbuildsystem/cachi2/pull/602 :fireworks: 

So the problem is that this PR was created from my repository.
I updated the badge in README.md.

codecov.yaml
eskultety: > I think it is set in [pyproject.toml](https://github.com/containerbuildsystem/cachi2/blob/main/pyproject.toml#L91)

Does codecov respect those settings? I'm now confused what the behaviour should be provided there are potentially 2 ways to configure coverage.

slimreaper35: If you run tox and the coverage is lower than 90%, it will fail.

ben-alkov: Really, we should pick something and go with it. Reconfiguring annoying behavior in a follow-up PR will be trivial.

eskultety: >  I would not have a problem with posting a comment always when the coverage has changed

IIUC, that would mean any tiny changes, both up and down, right? I mean the comment simply loses its meaning and people will start to ignore it, so we should only ever report coverage dropping.

eskultety: > If you run tox and the coverage is lower than 90%, it will fail.

@slimreaper35 can you elaborate? Does this mean we don't need any particular coverage settings and the tox settings in pyproject.toml is enough? Probably not for posting the comment, right? Essentially anything we select in that setting is just going to keep creating comments. 

> Reconfiguring annoying behavior in a follow-up PR will be trivial.

I'd prefer if we at least had some base understanding of what exactly each of those options do rather than start shooting in the dark, I never said we need to get it perfect at once, I just don't want to see another default dependabot settings :).

slimreaper35: > @slimreaper35 can you elaborate? Does this mean we don't need any particular coverage settings and the tox settings in pyproject.toml is enough? Probably not for posting the comment, right? Essentially anything we select in that setting is just going to keep creating comments.

Yes, settings in pyproject.toml are enough.
```
commands =
    pytest \
      --ignore=tests/integration \
      --cov=cachi2 \
      --cov-config=pyproject.toml \
...
```
We configured in our _pyproject.toml_ that if coverage drops below 90%, pytest will fail.  It means the tox will fail, and the GH check will fail. IMO, we don't need to set the same thing for Codecov.

slimreaper35: UPDATE: I tried to test it deeper, and I found out that there is:
- [project-status](https://docs.codecov.com/docs/commit-status#project-status)
- [patch-status](https://docs.codecov.com/docs/commit-status#patch-status)

So, we will need to configure both.

And yes, we will need to set the **project** target to 90% or an even bigger number+ threshold because the default target is current coverage, and we would not allow anyone to go below it. Same thing for **patch**.

My pull requests - https://github.com/slimreaper35/cachi2/pulls

eskultety: To share some of my recent experience based on experiments:

> [patch-status](https://docs.codecov.com/docs/commit-status#patch-status)
> 
> So, we will need to configure both (or disable one or both).

Determining the right target and threshold for a patch might be tricky, so I'd accept disabling that one completely, simply because based on [1] `uncovered_patch` doesn't respect the `target|threshold` values like `coverage_drop` does for the project status, so IOW a single uncovered line in the patch will cause the codecov action to comment on the PR. The setting in question only affects the actual gating check (unlike `coverage_drop`) whether it'll pass or fail. I don't oppose having the patch check in, but since it's gating one, I'm not sure we want _>90%_ coverage in there as target, we can definitely adjust empirically. 

> And yes, we will need to set the project target to 90% or an even bigger number+ threshold because the default target is base commit coverage, and we would not allow anyone to go below it.

Yes, I have to say I initially perceived the threshold relative to the current coverage, not the target, so IOW if one were to drop a bunch of unit tests that would cause the coverage to drop from 97% to 91%, it's more than the 5% threshold, but still over the 90% target and the check would succeed, so after understanding the settings more I have retract on one of my earlier comments and agree we probably don't even want a threshold combined with the 90% target as that one should be a hard baseline.

I also found out that the `condensed_` form of the `layout` setting for comments doesn't seem to include an important bit of information - **indirect effects**, e.g. if one deletes a bunch of unit tests, the overall coverage drops, but if it's all within the target coverage range, we'd never get otherwise notified that such a change suddenly indirectly caused many lines in code to be uncovered.

Lastly, since this is part of the gating test suite which runs in a Python env matrix, the codecov action likely uploads the same data for each Python env run. It somehow seems to work, but I think we can improve upon that and only send the coverage data once only.

[1] https://docs.codecov.com/docs/pull-request-comments#requiring-changes
[2] https://github.com/slimreaper35/cachi2/pull/12#issuecomment-2310651723

.github/workflows/gating.yaml
```diff
@@ -25,37 +25,21 @@ jobs:
...
-          pip3 install --upgrade setuptools
-          pip3 install --upgrade coveralls
-          coveralls --finish --service=github
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Upload coverage reports to Codecov
+      if: matrix.python-version == '3.12'
+      uses: codecov/codecov-action@v4
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
```
eskultety, line 42: Cool, this ended up being cleaner than anticipated :) .

codecov.yaml
slimreaper35: Have already agreed on `require_changes: "coverage_drop OR uncovered_patch"` vs `require_changes: "coverage_drop AND uncovered_patch"` ?

MartinBasti (approved):

codecov.yaml
a-ovchinnikov: +1 to having a comment opened only when there is a critical drop, these comments do add some visual noise. If it were opened only when there is a big issue it would be harder to ignore.

slimreaper35: So, I turned off comments completely because the comment would only appear if coverage drops below 90%, which is probably unrealistic. Right now, the coverage is ~ 98%. _Files changed_ section should provide enough information anyway.

eskultety (approved):

PR #593, opened 2024-08-09
Adding a new merge-sboms sub-command

a-ovchinnikov:
Functionality for merging SBOMs was already present in the codebase, this commit exposes it to a user.

### DIFF
--- a/cachi2/interface/cli.py
+++ b/cachi2/interface/cli.py
@@ -4,6 +4,7 @@ import json
 import logging
 import shutil
 import sys
+from itertools import chain
 from pathlib import Path
 from typing import Any, Callable, Optional
 
@@ -11,10 +12,12 @@ import pydantic
 import typer
 
 import cachi2.core.config as config
-from cachi2.core.errors import Cachi2Error, InvalidInput
+from cachi2.core.errors import Cachi2Error, InvalidInput, UnexpectedFormat
 from cachi2.core.extras.envfile import EnvFormat, generate_envfile
 from cachi2.core.models.input import Flag, PackageInput, Request, parse_user_input
 from cachi2.core.models.output import BuildConfig
+from cachi2.core.models.property_semantics import merge_component_properties
+from cachi2.core.models.sbom import Sbom
 from cachi2.core.resolver import inject_files_post, resolve_packages, supported_package_managers
 from cachi2.core.rooted_path import RootedPath
 from cachi2.interface.logging import LogLevel, setup_logging
@@ -24,6 +27,25 @@ log = logging.getLogger(__name__)
 
 DEFAULT_SOURCE = "."
 DEFAULT_OUTPUT = "./cachi2-output"
+
+
+OUTFILE_OPTION = typer.Option(
+    None,
+    "-o",
+    "--output",
+    dir_okay=False,
+    help="Write to this file instead of standard output.",
+)
+
+
+Paths = list[Path]
+
+
+def _bail_out_with_error(e: Cachi2Error) -> None:
+    """Report and error and set correct exit code."""
+    log.error("%s: %s", type(e).__name__, str(e).replace("\n", r"\n"))
+    print(f"Error: {type(e).__name__}: {e.friendly_msg()}", file=sys.stderr)
+    raise typer.Exit(2 if e.is_invalid_usage else 1)
 
 
 def handle_errors(cmd: Callable[..., None]) -> Callable[..., None]:
@@ -42,9 +64,7 @@ def handle_errors(cmd: Callable[..., None]) -> Callable[..., None]:
         try:
             cmd(*args, **kwargs)
         except Cachi2Error as e:
-            log_error(e)
-            print(f"Error: {type(e).__name__}: {e.friendly_msg()}", file=sys.stderr)
-            raise typer.Exit(2 if e.is_invalid_usage else 1)
+            _bail_out_with_error(e)
         except Exception as e:
             log_error(e)
             raise
@@ -291,13 +311,7 @@ FOR_OUTPUT_DIR_OPTION = typer.Option(
 def generate_env(
     from_output_dir: Path = FROM_OUTPUT_DIR_ARG,
     for_output_dir: Optional[Path] = FOR_OUTPUT_DIR_OPTION,
-    output: Optional[Path] = typer.Option(
-        None,
-        "-o",
-        "--output",
-        dir_okay=False,
-        help="Write to this file instead of standard output.",
-    ),
+    output: Optional[Path] = OUTFILE_OPTION,
     fmt: Optional[EnvFormat] = typer.Option(
         None,
         "-f",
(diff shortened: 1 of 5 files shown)


cachi2/interface/cli.py
```diff
@@ -346,6 +355,66 @@ def inject_files(
...
+    cachi2 fetch-deps <args...>
+
+    first to produce SBOMs to merge.
+    """
+    sboms_to_merge = []
+    for sbom_file in sbom_files_to_merge:
+        try:
+            sboms_to_merge.append(Sbom.model_validate_json(sbom_file.read_text()))
+        except pydantic.ValidationError:
+            raise typer.BadParameter(f"{sbom_file} does not appear to be a valid Cachi2 SBOM.")
```
brunoapimentel, line 403: I believe we're avoiding raising other errors besides the ones defined in [errors.py](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/errors.py). The downside is that we lose the nice default Typer formatting in the terminal (although our main real use cases would only see these results in pipeline logs, so the formatting would not matter much).

This is because we wanted to set up a small framework to, besides providing the error message, also provide some potential [solution](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/errors.py#L17) so the user could find his way around it.

cachi2/interface/cli.py
```diff
@@ -346,6 +355,66 @@ def inject_files(
...
+            try:
+                json.loads(sbom_file.read_text())
+            except ValueError:
+                raise typer.BadParameter(f"{sbom_file} does not look like a SBOM file")
+        return sbom_files_to_merge
+
+    return all_files_are_jsons(enough_files_for_merge(list(set(sbom_files_to_merge))))
+
+
+@app.command()
```
brunoapimentel, line 375: I wonder if we want to have this command as `hidden`. It strikes me that it will only serve a specific use case for Konflux, and probably is not something other users would care about. Although there is no hurt in exposing it, but in that case, we probably should have some small section about it in the repository docs.

brunoapimentel: @MartinBasti @eskultety Thoughts?

brunoapimentel:
Overall approach LGTM :+1:

cachi2/interface/cli.py
MartinBasti: I don't like hidden commands :D

.gitignore
```diff
@@ -117,3 +117,8 @@ settings.py
 
 # idea files
 .idea/
+
+# Vim temporary files and ctags tags
+*.swp
+*.swo
+tags
```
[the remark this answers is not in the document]
a-ovchinnikov: Sure, I'll split this out in the next revision.

cachi2/interface/cli.py
a-ovchinnikov: I'll add the docs section and will wait for more votes on hidden vs public. I have no strong opinion at the moment, but am leaning towards public command: it looks like it would be less confusing to have it exposed.

brunoapimentel: I don't have strong preferences either, but it seems folks prefer it public. So let's just add a small docs section about it and leave it public.

cachi2/interface/cli.py
a-ovchinnikov: Fair enough, I'll change exceptions to cachi2 exceptions. Raising them in CLI does produce some long and colorful tracebacks in terminal. I usually tend to hide tracebacks from users, but this is a special case.

cachi2/interface/cli.py
a-ovchinnikov: Done.

cachi2/interface/cli.py
a-ovchinnikov: Changed to proper exceptions in the follow-up revision, resolving.

docs/usage.md
```diff
@@ -106,6 +107,23 @@ it was previously clean. If any scripting depends on the cleanliness of a git re
...
 
+### Merge SBOMs
+
+Sometimes it might be necessary to merge two or more SBOMs. This could be done with `cachi2 merge-sboms`:
+
+```shell
+cachi2 merge-sboms <cachi2_sbom_1.json> ... <cachi2_sbom_n.json>
+```
+
+The subcommand expects at least two SBOMs, all in Cachi2 format and would exit with error if any of the SBOMs is in
```
brunoapimentel, line 118: I wonder if calling it `Cachi2 format` would be misleading. We're pretty much following the official CycloneDX specification, with some custom properties (the ones prefixed with `cachi2:`).

Maybe we could phrase this as:

> The subcommand expects at least two SBOMs, all produced by Cachi2 and will exit with error otherwise. The reason for this is that Cachi2 supports a [limited set](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/models/sbom.py#L7-L13) of component [properties](https://cyclonedx.org/docs/1.4/json/#components_items_properties), and it validates that no other properties exist in the SBOM.

a-ovchinnikov: Done

docs/usage.md
```diff
@@ -106,6 +107,26 @@ it was previously clean. If any scripting depends on the cleanliness of a git re
...
 
+### Merge SBOMs
+
+Sometimes it might be necessary to merge two or more SBOMs. This could be done with `cachi2 merge-sboms`:
+
+```shell
+cachi2 merge-sboms <cachi2_sbom_1.json> ... <cachi2_sbom_n.json>
+```
+
+The subcommand expects at least two SBOMs, all produced by Cachi2 and will exit with error
```
ben-alkov, line 118: Grammar pedant: 
```suggestion
The subcommand expects at least two SBOMs, all produced by Cachi2, and will exit with error
```

ben-alkov (approved):
LGTM, grammar is optional 😂

docs/usage.md
a-ovchinnikov: Done!

tests/unit/data/sboms/cachito_gomod_nodeps.bom.json
```diff
@@ -0,0 +1,39 @@
...
+    "tools": [
+      {
+        "vendor": "red hat",
+        "name": "cachi2"
+      }
+    ]
+  },
+  "specVersion": "1.4",
+  "version": 1
+}
```
brunoapimentel, line 39: Nitpick: I think that even for these test files, we are adding a trailing empty line.

a-ovchinnikov: That's how the files were generated by Cachi2, should I add a newline?

brunoapimentel: That is a good point. I looked at the other files in the same dir, and they all have a newline. Nevermind it then, as I said, it's a nitpick.

cachi2/interface/cli.py
```diff
@@ -346,6 +360,72 @@ def inject_files(
...
 
 
+def _prevalidate_sbom_files_args(sbom_files_to_merge: Paths) -> Paths:
+    def enough_files_for_merge(sbom_files_to_merge: Paths) -> Paths:
+        if len(sbom_files_to_merge) < 2:
+            # NOTE: an exception here happens during argument evaluation phase
+            # i.e. outside of handle_errors() decorator. Simply raising
+            # an exception here will not produce correct exit code, thus
+            # the explicit call to exception wrapper.
+            _bail_out_with_error(InvalidInput("Need at least two different SBOM files"))
```
brunoapimentel, line 370: Raising `InvalidInput` would have the same effect as wrapping it with `_bail_out_with_error` here, right? (considering the command is annotated with `@handle_errors`). This would avoid the need to extracting out `_bail_out_with_error`.

In case you prefer to keep it extracted, I'd say doing it in a separate commit would be better.

a-ovchinnikov: No, it won't: with a `raise` exit code is 1, with the wrapper it is set to 2 for `Cachi2Error`s. This exception will happen _before_ arguments are passed to the decorated function, thus outside of `handle_error` scope.

brunoapimentel (approved):

PR #599, opened 2024-08-14
bundler: parse Gemfile.lock

slimreaper35:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,8 @@ RUN dnf -y install \
     --nodocs \
     git-core \
     python3 \
+    rubygem-bundler \
+    rubygem-json \
     subscription-manager && \
     dnf clean all
 
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -241,7 +241,7 @@ class Request(pydantic.BaseModel):
 
     @property
     def bundler_packages(self) -> list[BundlerPackageInput]:
-        """Get the rubygems packages specified for this request."""
+        """Get the bundler packages specified for this request."""
         return self._packages_by_type(BundlerPackageInput)
 
     @property
--- a/cachi2/core/package_managers/bundler/parser.py
+++ b/cachi2/core/package_managers/bundler/parser.py
@@ -1,0 +1,127 @@
+import json
+import logging
+import subprocess
+from pathlib import Path
+from typing import Annotated, Optional, Union
+
+import pydantic
+
+from cachi2.core.errors import PackageManagerError, PackageRejected, UnexpectedFormat
+from cachi2.core.rooted_path import PathOutsideRoot, RootedPath
+from cachi2.core.utils import run_cmd
+
+log = logging.getLogger(__name__)
+
+GEMFILE = "Gemfile"
+GEMFILE_LOCK = "Gemfile.lock"
+
+AcceptedUrl = Annotated[
+    pydantic.HttpUrl,
+    pydantic.UrlConstraints(allowed_schemes=["https"]),
+]
+
+AcceptedGitRef = Annotated[
+    pydantic.StrictStr,
+    pydantic.StringConstraints(pattern=r"^[a-fA-F0-9]{40}$"),
+]
+
+
+class _GemMetadata(pydantic.BaseModel):
+    """
+    Base class for gem metadata.
+
+    Attributes:
+        name:       The name of the gem.
+        version:    The version of the gem.
+    """
+
+    name: str
+    version: str
+
+
+class GemDependency(_GemMetadata):
+    """
+    Represents a gem dependency.
+
+    Attributes:
+        source:     The source URL of the gem.
+        checksum:   The checksum of the gem.
+    """
+
+    source: str
+    checksum: Optional[str] = None
+
+
+class GitDependency(_GemMetadata):
+    """
+    Represents a git dependency.
+
+    Attributes:
+        url:        The URL of the git repository.
+        ref:        Commit hash.
+    """
+
+    url: AcceptedUrl
+    ref: AcceptedGitRef
+
+
+class PathDependency(_GemMetadata):
+    """
+    Represents a path dependency.
+
+    Attributes:
+        path:       Subpath from package root.
+    """
+
+    path: str
+
+
+BundlerDependency = Union[GemDependency, GitDependency, PathDependency]
+ParseResult = list[BundlerDependency]
+
+
+def parse_lockfile(package_dir: RootedPath) -> ParseResult:
+    """Parse a Gemfile.lock file and return a list of dependencies."""
+    lockfile_path = package_dir.join_within_root(GEMFILE_LOCK)
+    gemfile_path = package_dir.join_within_root(GEMFILE)
+    if not lockfile_path.path.exists() or not gemfile_path.path.exists():
+        reason = "Gemfile and Gemfile.lock must be present in the package directory"
+        solution = (
+            "Run `bundle init` to generate the Gemfile.\n"
+            "Run `bundle lock` to generate the Gemfile.lock."
+        )
+        raise PackageRejected(reason=reason, solution=solution)
+
+    scripts_dir = Path(__file__).parent / "scripts"
+    lockfile_parser = scripts_dir / "lockfile_parser.rb"
+    try:
+        output = run_cmd(cmd=[str(lockfile_parser)], params={"cwd": package_dir.path})
+    except subprocess.CalledProcessError:
+        raise PackageManagerError(f"Failed to parse {lockfile_path}")
+
+    json_output = json.loads(output)
+
+    bundler_version: str = json_output["bundler_version"]
+    log.info("Package %s is bundled with version %s", package_dir.path.name, bundler_version)
+    dependencies: list[dict[str, str]] = json_output["dependencies"]
+
+    result: ParseResult = []
+    for dep in dependencies:
+        if dep["type"] == "rubygems":
+            result.append(GemDependency(**dep))
+        elif dep["type"] == "git":
+            result.append(GitDependency(**dep))
+        elif dep["type"] == "path":
+            _validate_path_dependency_subpath(package_dir, dep["path"])
+            result.append(PathDependency(**dep))
+
+    return result
+
+
+def _validate_path_dependency_subpath(package_dir: RootedPath, path: str) -> None:
+    """Validate that the path dependency is within the package root."""
+    try:
+        package_dir.join_within_root(path)
+    except PathOutsideRoot:
+        reason = "PATH dependencies should be within the package root"
+        raise UnexpectedFormat(reason=reason)
--- a/cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb
+++ b/cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb
@@ -1,0 +1,36 @@
+#!/usr/bin/env ruby
+
+require 'bundler'
+require 'json'
+
+lockfile_content = File.read("Gemfile.lock")
+lockfile_parser = Bundler::LockfileParser.new(lockfile_content)
+
+parsed_specs = []
+
+lockfile_parser.specs.each do |spec|
+    parsed_spec = {
+      name: spec.name,
+      version: spec.version.to_s
+    }
+
+    case spec.source
+    when Bundler::Source::Rubygems
+      parsed_spec[:type] = 'rubygems'
+      parsed_spec[:source] = spec.source.remotes.map(&:to_s).first
+    when Bundler::Source::Git
+      parsed_spec[:type] = 'git'
+      parsed_spec[:url] = spec.source.uri
+      parsed_spec[:ref] = spec.source.revision
+    when Bundler::Source::Path
+      parsed_spec[:type] = 'path'
+      parsed_spec[:path] = spec.source.path.to_s
+    end
+
+    parsed_specs << parsed_spec
+  end
+
+puts JSON.pretty_generate({ bundler_version: lockfile_parser.bundler_version, dependencies: parsed_specs })
+
+# References:
+# https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb
--- a/tests/unit/package_managers/bundler/test_parser.py
+++ b/tests/unit/package_managers/bundler/test_parser.py
@@ -1,0 +1,182 @@
+import json
+import subprocess
+from copy import deepcopy
+from typing import Any
+from unittest import mock
+
+import pydantic
+import pytest
+
+from cachi2.core.errors import PackageManagerError, PackageRejected, UnexpectedFormat
+from cachi2.core.package_managers.bundler.parser import (
+    GEMFILE,
+    GEMFILE_LOCK,
+    GemDependency,
+    GitDependency,
+    PathDependency,
+    parse_lockfile,
+)
+from cachi2.core.rooted_path import RootedPath
+
+
+@pytest.fixture
+def empty_bundler_files(rooted_tmp_path: RootedPath) -> tuple[RootedPath, RootedPath]:
+    gemfile_path = rooted_tmp_path.join_within_root(GEMFILE)
+    gemfile_path.path.touch()
+
+    lockfile_path = rooted_tmp_path.join_within_root(GEMFILE_LOCK)
+    lockfile_path.path.touch()
+
+    return gemfile_path, lockfile_path
+
+
+SAMPLE_PARSER_OUTPUT = {
+    "bundler_version": "2.5.10",
+    "dependencies": [{"name": "example", "version": "0.1.0"}],
+}
+
+
+@pytest.fixture
+def sample_parser_output() -> dict[str, Any]:
+    return deepcopy(SAMPLE_PARSER_OUTPUT)
+
+
+def test_parse_lockfile_without_bundler_files(rooted_tmp_path: RootedPath) -> None:
+    with pytest.raises(PackageRejected) as exc_info:
+        parse_lockfile(rooted_tmp_path)
+
+    assert (
+        "Gemfile and Gemfile.lock must be present in the package directory"
+        in exc_info.value.friendly_msg()
+    )
+
+
+@mock.patch("cachi2.core.package_managers.bundler.parser.run_cmd")
+def test_parse_lockfile_os_error(
+    mock_run_cmd: mock.MagicMock,
+    empty_bundler_files: tuple[RootedPath, RootedPath],
+    rooted_tmp_path: RootedPath,
+) -> None:
+    mock_run_cmd.side_effect = subprocess.CalledProcessError(returncode=1, cmd="cmd")
+
+    with pytest.raises(PackageManagerError) as exc_info:
+        parse_lockfile(rooted_tmp_path)
+
+    assert f"Failed to parse {empty_bundler_files[1]}" in exc_info.value.friendly_msg()
+
+
+@mock.patch("cachi2.core.package_managers.bundler.parser.run_cmd")
+@pytest.mark.parametrize(
+    "error, expected_error_msg",
+    [
+        ("LOCKFILE_INVALID_URL", "Input should be a valid URL"),
+        ("LOCKFILE_INVALID_URL_SCHEME", "URL scheme should be 'https'"),
+        ("LOCKFILE_INVALID_REVISION", "String should match pattern '^[a-fA-F0-9]{40}$'"),
+        ("LOCKFILE_INVALID_PATH", "PATH dependencies should be within the package root"),
+    ],
+)
+def test_parse_lockfile_invalid_format(
+    mock_run_cmd: mock.MagicMock,
+    error: str,
+    expected_error_msg: str,
+    empty_bundler_files: tuple[RootedPath, RootedPath],
+    sample_parser_output: dict[str, Any],
+    rooted_tmp_path: RootedPath,
+) -> None:
+    if error == "LOCKFILE_INVALID_URL":
+        sample_parser_output["dependencies"][0].update(
+            {
+                "type": "git",
+                "url": "github",
+                "ref": "f" * 40,
+            }
+        )
+    elif error == "LOCKFILE_INVALID_URL_SCHEME":
+        sample_parser_output["dependencies"][0].update(
+            {
+                "type": "git",
+                "url": "http://github.com/3scale/json-schema.git",
+                "ref": "f" * 40,
+            }
+        )
+    elif error == "LOCKFILE_INVALID_REVISION":
+        sample_parser_output["dependencies"][0].update(
+            {
+                "type": "git",
+                "url": "https://github.com/3scale/json-schema.git",
+                "ref": "abcd",
+            }
+        )
+    elif error == "LOCKFILE_INVALID_PATH":
+        sample_parser_output["dependencies"][0].update(
+            {
+                "type": "path",
+                "path": "/root/pathgem",
+            }
+        )
+
+    mock_run_cmd.return_value = json.dumps(sample_parser_output)
+    with pytest.raises((pydantic.ValidationError, UnexpectedFormat)) as exc_info:
+        parse_lockfile(rooted_tmp_path)
+
+    assert expected_error_msg in str(exc_info.value)
+
+
+@mock.patch("cachi2.core.package_managers.bundler.parser.run_cmd")
+def test_parse_gemlock(
+    mock_run_cmd: mock.MagicMock,
+    empty_bundler_files: tuple[RootedPath, RootedPath],
+    sample_parser_output: dict[str, Any],
+    rooted_tmp_path: RootedPath,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    base_dep: dict[str, str] = sample_parser_output["dependencies"][0]
+    sample_parser_output["dependencies"] = [
+        {
+            "type": "git",
+            "url": "https://github.com/3scale/json-schema.git",
+            "ref": "f" * 40,
+            **base_dep,
+        },
+        {
+            "type": "path",
+            "path": "subpath/pathgem",
+            **base_dep,
+        },
+        {
+            "type": "rubygems",
+            "source": "https://rubygems.org/",
+            **base_dep,
+        },
+    ]
+
+    mock_run_cmd.return_value = json.dumps(sample_parser_output)
+    result = parse_lockfile(rooted_tmp_path)
+
+    expected_deps = [
+        GitDependency(
+            name="example",
+            version="0.1.0",
+            url="https://github.com/3scale/json-schema.git",
+            ref="f" * 40,
+        ),
+        PathDependency(name="example", version="0.1.0", path="subpath/pathgem"),
+        GemDependency(name="example", version="0.1.0", source="https://rubygems.org/"),
+    ]
+
+    assert f"Package {rooted_tmp_path.path.name} is bundled with version 2.5.10" in caplog.messages
+    assert result == expected_deps
+
+
+@mock.patch("cachi2.core.package_managers.bundler.parser.run_cmd")
+def test_parse_gemlock_empty(
+    mock_run_cmd: mock.MagicMock,
+    empty_bundler_files: tuple[RootedPath, RootedPath],
+    rooted_tmp_path: RootedPath,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    mock_run_cmd.return_value = '{"bundler_version": "2.5.10", "dependencies": []}'
+    result = parse_lockfile(rooted_tmp_path)
+
+    assert f"Package {rooted_tmp_path.path.name} is bundled with version 2.5.10" in caplog.messages
+    assert result == []
--- a/cachi2/core/package_managers/bundler/main.py
+++ b/cachi2/core/package_managers/bundler/main.py
@@ -1,6 +1,8 @@
 from cachi2.core.models.input import Request
 from cachi2.core.models.output import EnvironmentVariable, ProjectFile, RequestOutput
 from cachi2.core.models.sbom import Component
+from cachi2.core.package_managers.bundler.parser import parse_lockfile
+from cachi2.core.rooted_path import RootedPath
 
 
 def fetch_bundler_source(request: Request) -> RequestOutput:
@@ -9,8 +11,18 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
     environment_variables: list[EnvironmentVariable] = []
     project_files: list[ProjectFile] = []
 
+    for package in request.packages:
+        path_within_root = request.source_dir.join_within_root(package.path)
+        _resolve_bundler_package(package_dir=path_within_root, output_dir=request.output_dir)
+
     return RequestOutput.from_obj_list(
         components=components,
         environment_variables=environment_variables,
         project_files=project_files,
     )
+
+
+def _resolve_bundler_package(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+    """Process a request for a single bundler package."""
+    parse_lockfile(package_dir)
+    return []


cachi2/core/package_managers/rubygems/parser.py
```diff
@@ -1,9 +1,12 @@
 import logging
-from typing import Optional, Union
+import re
+from typing import Iterable, Optional, Union
 
 import pydantic
+from gemlock_parser.gemfile_lock import GEM, GIT, PATH, Gem, GemfileLockParser  # type: ignore
```
eskultety, line 6: This is a vendored version of https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/packagedcode/gemfile_lock.py which was updated in Feb 2024 whereas our code was updated in 2022.
While it's okay to copy-paste parts of the code instead of reinventing the wheel, when adding new code to this code base we should take the opportunity and evaluate if things can be improved upon given current restrictions, limitations, updates, technologies, etc.

In this particular case, when I look at both the vendored parser and its origin, that's a few hundred lines of code that need to be maintained (in this case, us, and it's not happening). So, I wonder if a we'd not be better off with a significantly shorter snippet of Ruby code checked into this code base, simply because Ruby's `bundler` library contains primitives to parse the lockfile, so we should leverage its capability to do the heavy lifting for us if possible. For example, this snippet of code is all it's supposedly needed to parse the lockfile and dump it as JSON we would then parse in our Python code (diclaimer: I did generate this prototype with an AI as I can't write any Ruby, but it's not rocket science and it seems it should be easily adjusted to our needs):

```
require 'bundler'
require 'json'

# Load and parse the Gemfile.lock using Bundler::LockfileParser
lockfile_contents = File.read('Gemfile.lock')
parser = Bundler::LockfileParser.new(lockfile_contents)

# Prepare a hash structure to hold the parsed data
parsed_data = {}

parser.specs.each do |spec|
  parsed_data[spec.name] = {
    version: spec.version.to_s,
    dependencies: spec.dependencies.map { |dep| { name: dep.name, requirement: dep.requirement.to_s } }
  }
end

# Convert the parsed data to JSON
json_output = JSON.pretty_generate(parsed_data)

# Output the JSON to a file or stdout
File.write('parsed_gemfile_lock.json', json_output)
puts json_output
```

Note this example doesn't fill in the `remote` string, but I think it could be adapted if the parser class is further inspected. The point of this comment is whether we want to rely on a few hundred lines of unmaintained code vs using a few lines of Ruby dealing with it natively. Also note that we're spawning so many subprocess calls already, so an extra call to ruby to execute such a snippet is negligible.

slimreaper35: interesting...
I slightly updated the script:
```
#!/usr/bin/env ruby

require 'bundler'
require 'json'

if ARGV.empty?
    puts "Usage: #{__FILE__} path/to/Gemfile.lock"
    exit 1
  end
  
  gemfile_lock_path = ARGV[0]
  
  begin
    lockfile_content = File.read(gemfile_lock_path)
  rescue Errno::ENOENT
    puts "Error: File not found - #{gemfile_lock_path}"
    exit 1
  end

parser = Bundler::LockfileParser.new(lockfile_content)

parsed_data = {}

parser.specs.each do |spec|
    source_details = {}
  
    case spec.source
    when Bundler::Source::Rubygems
      source_details[:type] = 'rubygems'
      source_details[:url] = spec.source.remotes.map(&:to_s).first
      source_details[:checksum] = spec.respond_to?(:checksum) ? spec.checksum : nil
    when Bundler::Source::Git
      source_details[:type] = 'git'
      source_details[:url] = spec.source.uri
      source_details[:revision] = spec.source.revision
      source_details[:branch] = spec.source.branch
    when Bundler::Source::Path
      source_details[:type] = 'path'
      source_details[:subpath] = spec.source.path.to_s
    else
      source_details[:type] = 'unknown'
    end
  
    gem_data = {
      version: spec.version.to_s,
      dependencies: spec.dependencies.map { |dep| { name: dep.name, requirement: dep.requirement.to_s } },
      type: source_details[:type]
    }
  
    gem_data[:checksum] = source_details[:checksum] if source_details[:checksum]
    gem_data[:url] = source_details[:url] if source_details[:url]
    gem_data[:revision] = source_details[:revision] if source_details[:revision]
    gem_data[:branch] = source_details[:branch] if source_details[:branch]
    gem_data[:subpath] = source_details[:subpath] if source_details[:subpath]
  
    parsed_data[spec.name] = gem_data
  end

json_output = JSON.pretty_generate(parsed_data)
File.write('parsed-gems.json', json_output)
```

There must be (at least empty) `Gemfile` in the same directory as `Gemfile.lock`.
Usage:
`ruby <script> <path-to-Gemfile.lock>`

eskultety: > There must be (at least empty) Gemfile in the same directory as you run the script.

If so, then cachi2 needs to make sure the repo is correctly structured before proceeding with processing it, for example with such a script.

chmeliik: > There must be (at least empty) Gemfile in the same directory as you run the script.

Sounds dangerous, are we sure the ruby library doesn't execute the Gemfile?

slimreaper35: Here is the problem - https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/shared_helpers.rb#L17

But we can set `BUNDLE_GEMFILE` to something non-empty as a hack :smile: 
e.g. `BUNDLE_GEMFILE="..." ruby <script> <path-to-Gemfile.lock>`

https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/shared_helpers.rb#L248

cachi2/core/package_managers/rubygems/main.py
```diff
@@ -0,0 +1,95 @@
...
+def _resolve_rubygems(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+    """
+    Process a request for a single rubygems package.
+
+    TODO:
+    1. Parse Gemfile.lock
+    2. Get project metadata
+    3. Download dependencies
+    4. Generate SBOM components
+    """
```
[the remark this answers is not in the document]
slimreaper35: I rebased this branch on high-level design. This should only cover parsing Gemfil.lock. I will update it once high-level code is merged. It is marked as a draft because of that.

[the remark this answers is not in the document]
a-ovchinnikov: Got it, thank you!

cachi2/core/package_managers/rubygems/parser.py
eskultety: > But we can set BUNDLE_GEMFILE to something non-empty as a hack 😄

@slimreaper35 I'm confused, isn't the presence of both a `Gemfile` and `Gemfile.lock` mandatory otherwise it's a repo misconfiguration? If so, then we need to the usual sequence of checking mandatory files, path validations, symlink checks, etc. before proceeding with parsing, am I missing something?

slimreaper35: In the user's repo yes, but I have this script locally in cachi2 repo and pointing to some Gemfile.lock.  When I run the script, ruby is complaining that _Could not locate Gemfile_, which apparently can be silenced by setting env variable BUNDLE_GEMFILE to non-empty value.

eskultety: > In the user's repo yes, but I have this script locally in cachi2 repo and pointing to some Gemfile.lock. When I run the script, ruby is complaining that Could not locate Gemfile (in cachi2 repo), which apparently can be silenced by setting env variable BUNDLE_GEMFILE to non-empty value.

This IMO irrelevant to the intended use, note that such a script shouldn't ever be used outside of our own Python code, so one can assume it's only ever going to be executed on a correctly configured repo. Any other use case can be considered invalid and so we don't need any variable settings even if possible in this particular case you mention.

slimreaper35: > note that such a script shouldn't ever be used outside of our own Python code

I am confused. Where should the script be used ?

brunoapimentel: > Sounds dangerous, are we sure the ruby library doesn't execute the Gemfile?

This is the main thing we should be investigating. Until now, we considered that all the heavy lifting was done by custom Python code. Adding Bundler to the equation changes the fundamentals of our initial design (which is not a problem per se). So we need to make sure that no arbitrary code execution is ever possible. 

Then, give some thought about how to maintain the Ruby/Bundler binaries in our image: is installing a single version enough (should be, right)? Or will we rely on the Bundler "toolchain" mechanism. 

Otherwise, this is a good improvement from our vendored code.

eskultety: > Sounds dangerous, are we sure the ruby library doesn't execute the Gemfile?

@chmeliik unlikely, I went ahead and spent a few hours debugging Ruby and looking at the code. It doesn't interact with the Gemfile in any way. Let me expand.

> Here is the problem - https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/shared_helpers.rb#L17

While this is technically true, it doesn't tell anyone what's happening underneath, here's the full stacktrace for the script @slimreaper35 provided:
```
/usr/local/bundle/gems/bundler-2.5.17/lib/bundler/shared_helpers.rb:17:in `root': Could not locate Gemfile (Bundler::GemfileNotFound)
        from /usr/local/bundle/gems/bundler-2.5.17/lib/bundler.rb:325:in `root'
        from /usr/local/bundle/gems/bundler-2.5.17/lib/bundler.rb:101:in `bundle_path'
        from /usr/local/bundle/gems/bundler-2.5.17/lib/bundler/source/git.rb:244:in `cache_path'
        from /usr/local/bundle/gems/bundler-2.5.17/lib/bundler/source/git.rb:375:in `git_proxy'
        from /usr/local/bundle/gems/bundler-2.5.17/lib/bundler/source/git.rb:253:in `revision'
        from /test.rb:37:in `block in <main>'
        from /test.rb:26:in `each'
        from /test.rb:26:in `<main>'
```
More specifically, bundler is just trying to set the caching dir path where the cloned repos should reside. In order to calculate the default path it's trying to locate `Gemfile` and then just use the parent stem of the path. Here is an ordered list of relevant calls made:

- https://github.com/rubygems/rubygems/blob/3fdc2f627e6bb476b722c7d37b61074430dea82e/bundler/lib/bundler/source/git.rb#L240
- https://github.com/rubygems/rubygems/blob/3fdc2f627e6bb476b722c7d37b61074430dea82e/bundler/lib/bundler.rb#L100
- https://github.com/rubygems/rubygems/blob/3fdc2f627e6bb476b722c7d37b61074430dea82e/bundler/lib/bundler.rb#L324

And that's it, provided explicit revisions are using in the lockfile, not even `git rev-parse` is run which otherwise would, but for handling of the deps there are different modules like `fetcher` and `installer`.

eskultety: > Then, give some thought about how to maintain the Ruby/Bundler binaries in our image: is installing a single version enough (should be, right)? Or will we rely on the Bundler "toolchain" mechanism.

@brunoapimentel in general, we may need to do something like with Yarn and corepack, but it's irrelevant to this particular case we're discussing IMO, because no binary apart from the Ruby interpreter is involved because everything are just standard library calls.

eskultety: > I am confused. Where should the script be used ?

@slimreaper35 you'd use it from the `rubygems` or what exact name ends up being Python module as a `subprocess.call` where you specify the exact PATH context for the call and you should completely control the execution env. Such a subprocess call would only be made once cachi2 verifies the repo configuration.

tests/unit/package_managers/test_pip.py
```diff
@@ -42,11 +42,6 @@ def setup_module() -> None:
     pip.log.setLevel(logging.DEBUG)
 
 
-@pytest.fixture
```
eskultety, line 45: If this were changes to the pip module I'd be fine with it, but this commit feels misplaced in the overall changes and theme set by this PR, so please propose it as a standalone PR. Also, for future reference, when posting a PR, please include fixes/refactors like this one at the very beginning, not in the middle of the patch series, the sequence of patches should follow some logical order, it helps the reviewer not to lose the train of thoughts, just a recommendation ;).

tests/unit/package_managers/test_gomod.py
```diff
@@ -17,7 +17,6 @@
 from cachi2.core.models.input import Flag, Request
 from cachi2.core.models.output import BuildConfig, EnvironmentVariable, RequestOutput
 from cachi2.core.models.sbom import Component, Property
-from cachi2.core.package_managers import gomod
```
eskultety, line 20: Again, as with previous patch, this one feels slightly misplaced, although it may be somewhat related to the parser test module in the next one, so either propose this one as standalone or at least include it at the beginning of the series among other refactors.

tests/unit/package_managers/test_pip.py
slimreaper35: I know. As I already responded to Alexey, I rebased on the high level code branch. Otherwise both commits would be at the beginning.

cachi2/core/package_managers/rubygems/parser.py
```diff
@@ -67,12 +70,62 @@ class PathDependency(GemMetadata):
 
 def parse_gemlock(package_dir: RootedPath) -> list[BundlerDependency]:
     """Parse a Gemfile.lock file and return a list of dependencies."""
-    return NotImplemented
+    gemlock_path = package_dir.join_within_root(GEMFILE_LOCK)
+    if not gemlock_path.path.exists():
+        reason = f"Gemfile.lock not found at {gemlock_path.path}"
+        solution = "Please ensure the Gemfile.lock file is present in the package directory"
```
eskultety, line 76: FWIW you'd call the Ruby lockfile parsing script somewhere around here, but that would change as more repo validations would be done first elsewhere.

cachi2/core/package_managers/rubygems/parser.py
```diff
@@ -0,0 +1,131 @@
...
+        elif gem.type == GIT:
+            git_dep = GitDependency(
+                name=gem.name,
+                # https://github.com/containerbuildsystem/gemlock-parser/blob/master/gemlock_parser/gemfile_lock.py#L185
+                version=gem.spec_version,
+                url=gem.remote,
+                revision=gem.revision,
+                branch=gem.branch,
+            )
+            _validate_git_dependency(git_dep)
```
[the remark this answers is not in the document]
eskultety: Good question, again, most likely consistency with other pkg managers, but my take on that is that since we're adding a new one, we might as well define an abstract interface for these dep types and let inheritance take care of this like you say.

eskultety (commented):
For a draft this looks promising, left a few comments, but for now, LGTM. Once we have a definitive conclusion on the inclusion of a ruby script handling the lockfile parsing I think this may then be converted to a normal PR.

cachi2/core/package_managers/rubygems/parser.py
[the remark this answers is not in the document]
slimreaper35: Previously, I used Pydantic base models instead of  `@dataclass`. Would pydantic validators solve this, or am I missing something?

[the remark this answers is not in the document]
a-ovchinnikov: Yes, `@field_validator` should do the trick.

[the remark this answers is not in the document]
eskultety: Agreed, this is a good example of post-init validation (the easier use case vs compared to the RPM input JSON model) so field validator is enough.

cachi2/core/package_managers/bundler/lockfile-parser.rb
```diff
@@ -0,0 +1,40 @@
...
+    parsed_spec = {
+      name: spec.name,
+      version: spec.version.to_s
+    }
+
+    case spec.source
+    when Bundler::Source::Rubygems
+      parsed_spec[:type] = 'rubygems'
+      parsed_spec[:source] = spec.source.remotes.map(&:to_s).first
+      # parsed_spec[:checksum] = spec.respond_to?(:checksum) ? spec.checksum : nil
```
eskultety, line 21: Is this some left-over that you intended to drop? :)

cachi2/core/package_managers/bundler/lockfile-parser.rb
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env ruby
```
eskultety, line 1: nitpick: I'd probably separate Python non-native resources to another subdirectory, e.g. `scripts`. Maybe I'd go as far as moving it out of the bundler package and centralize it somewhere under `cachi2/core` but that's just a tiny detail that can be tweaked later on, for now I'd just stuff this script into a subdirectory relative to the bundler package.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,160 @@
...
+
+def parse_gemlock(package_dir: RootedPath) -> ParseResult:
+    """Parse a Gemfile.lock file and return a list of dependencies."""
+    gemfile_lock_path = package_dir.join_within_root(GEMFILE_LOCK)
+    if not gemfile_lock_path.path.exists():
+        reason = f"Gemfile.lock not found at {gemfile_lock_path.path}"
+        solution = "Please ensure the Gemfile.lock file is present in the package directory"
+        raise PackageRejected(reason=reason, solution=solution)
+
+    # TODO: Merge both conditions ?
```
eskultety, line 115: Yes, please merge them.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,160 @@
...
+        raise PackageRejected(reason=reason, solution=solution)
+
+    bundler_package = files("cachi2.core.package_managers.bundler")
+    lockfile_parser_script = bundler_package.joinpath("lockfile-parser.rb")
+
+    output = subprocess.run(
+        args=[str(lockfile_parser_script)],
+        cwd=package_dir.path,
+        capture_output=True,
+        shell=True,
```
eskultety, line 130: Why do you need a shell if you're only passing path to the script itself in args and the script is marked executable? We should avoid the need for a shell whenever possible.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,160 @@
...
+    for dep in dependencies:
+        if dep["type"] == "rubygems":
+            result.append(GemDependency.model_validate(dep))
+        elif dep["type"] == "git":
+            result.append(GitDependency.model_validate(dep))
+        elif dep["type"] == "path":
+            _validate_path_dependency(package_dir, dep["path"])
+            result.append(PathDependency.model_validate(dep))
+
+        # assert_never
```
eskultety, line 149: Would you care to actually add the `assert_never` in?

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,160 @@
...
+    @property
+    def purl(self) -> str:
+        """Get purl for this git dependency."""
+        return NotImplemented
+
+    @pydantic.field_validator("url")
+    def validate_git_url(cls, url: str) -> str:
+        """Validate that the git URL is a valid git URL."""
+        if not url.startswith("https://"):
+            raise ValueError("GIT dependencies must use HTTPS protocol")
```
eskultety, line 73: Is this all we want to validate? I guess the Ruby lockfile parser would flag incorrect URLs, but if we really perform validation, it might be worth using our `extract_git_info` which parses the whole URL into bits and which we use elsewhere.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,160 @@
...
+
+        return url
+
+    @pydantic.field_validator("revision")
+    def validate_git_revision(cls, revision: str) -> str:
+        """Validate that the git revision is a valid git ref."""
+        git_ref_format = re.compile(r"^[a-fA-F0-9]{40}$")
+        if not git_ref_format.match(revision):
```
eskultety, line 81: No need to do it this way, you can just use `re.match` directly as it'll compile the pattern implicitly. If you want to leverage the benefit of compiling patterns upfront, then you need to cache it somewhere so that it'll be used for every single git dependency model instantiation, but admittedly that only becomes useful in decreasing overhead once you need to process a good number of strings (several hundreds, maybe even thousands), which I don't expect to be the case here, so just plain `re.match` is going to be sufficient for us.

eskultety (commented):
From a quick look this looks solid, I'll wait for the tests whether they uncover more potential issues.

cachi2/core/package_managers/bundler/parser.py
slimreaper35: Bandit is complaining about it.  We should probably disable this rule globally since I saw it [here](https://github.com/containerbuildsystem/cachi2/blob/b737d90806a72977d1aa664ea470d963c5248375/cachi2/core/utils.py#L51) as well.

eskultety: Yes, we need to disable it, because the rule is completely dumb - not only will it not accept a string literal passed directly (what more can you do?), but it also suggests usage of a known attack vector. To support my argument against usage of `shell=true`, have a look at this example: https://security.openstack.org/guidelines/dg_use-subprocess-securely.html

Last but not least this also goes along another of my recent comments in a different PR where I said we ought not blindly follow all rules coming from the linters and instead use common sense; this is a prime example of where things can go utterly wrong.

eskultety: Aand even Ruff documents it as bad (if I understood correctly the whole page): https://docs.astral.sh/ruff/rules/subprocess-without-shell-equals-true/

cachi2/core/package_managers/bundler/parser.py
slimreaper35: I believe `urlparse` is enough here. `extract_git_info` needs VCS URL.
```
>>> from urllib.parse import urlparse
>>> parsed = urlparse("https://github.com/3scale/json-schema.git")
>>> parsed
ParseResult(scheme='https', netloc='github.com', path='/3scale/json-schema.git', params='', query='', fragment='')
```

eskultety: Right, it was because of `urlparse` I suggested it in the first place so works for me.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,155 @@
...
+        solution = (
+            "Run `bundle init` to generate the Gemfile.\n"
+            "Run `bundle lock` to generate the Gemfile.lock."
+        )
+        raise PackageRejected(reason=reason, solution=solution)
+
+    bundler_package = files("cachi2.core.package_managers.bundler")
+    lockfile_parser_script = str(bundler_package.joinpath("scripts", "lockfile-parser.rb"))
+
+    output = subprocess.check_output(
```
brunoapimentel, line 123: Will this explode with an ugly error in case the user does not have `bundler` installed locally?

eskultety: It will, good catch.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,155 @@
...
+    for dep in dependencies:
+        if dep["type"] == "rubygems":
+            result.append(GemDependency.model_validate(dep))
+        elif dep["type"] == "git":
+            result.append(GitDependency.model_validate(dep))
+        elif dep["type"] == "path":
+            _validate_path_dependency_subpath(package_dir, dep["path"])
+            result.append(PathDependency.model_validate(dep))
+        else:
+            assert_never
```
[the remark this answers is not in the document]
slimreaper35: Do we really need _assert_never_ ? :thinking:

cachi2/core/package_managers/bundler/parser.py
slimreaper35: Fixed

cachi2/core/models/input.py
```diff
@@ -241,7 +241,7 @@ def _packages_not_empty(cls, packages: list[PackageInput]) -> list[PackageInput]
 
     @property
     def bundler_packages(self) -> list[BundlerPackageInput]:
-        """Get the rubygems packages specified for this request."""
+        """Get the bundler packages specified for this request."""
```
eskultety, line 244: This typo fix should be in a separate commit.

Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM registry.access.redhat.com/ubi9/ubi@sha256:9e6a89ab2a9224712391c77fab2ab010
 FROM docker.io/library/golang:1.20.0-bullseye as golang_120
 FROM docker.io/library/golang:1.21.0-bullseye as golang_121
 FROM docker.io/library/node:22.7.0-bullseye as node
+FROM docker.io/library/ruby:3.1.6-bullseye as ruby
```
eskultety, line 5: Dockerfile changes should be separate from core logic, so should go to a different commit (in general, see my other comment below).

Dockerfile
```diff
@@ -45,6 +46,7 @@ COPY --from=golang_120 /usr/local/go /usr/local/go/go1.20
 COPY --from=golang_121 /usr/local/go /usr/local/go/go1.21
 COPY --from=node /usr/local/lib/node_modules/corepack /usr/local/lib/corepack
 COPY --from=node /usr/local/bin/node /usr/local/bin/node
+COPY --from=ruby /usr/local/bin/ruby /usr/local/bin/ruby
```
eskultety, line 49: Copying the executable isn't by far enough, you need shared (.so) libraries for the runtime as well and bundler installation on top of it too. Having said that - do we have a minimum version requirement for ruby/bundler? I don't think so. We have legitimate reasons we copy over stuff for NodeJS and Golang:

- the version of NodeJS shipped by UBI is ~16 and we needed a higher version of corepack, most notably for future yarn v4 support
- we need the latest Go releases due to minimum version requirements in `go.mod` files which many upstream projects that use cachi2 pin to the very latest upstream Go release.

One could argue that Ruby 3.0 (UBI 9 has 3.0.7) has been eol'd in upstream, but we don't really need Ruby (for the core logic, we only need it to run the helper script), we need bundler, so as long as we don't need specific bundler features I'd stay away from copying stuff over from docker images by default and instead install whatever we need the usual way - using the platform package manager -> DNF and only fall back to docker layer copying when that is insufficient.

cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env ruby
```
eskultety, line 1: Not a deal breaker for me, but I'd also introduce this helper script in a separate commit preceding the parser changes. That is to keep git history nice and clean and not to make the script go unnoticed, IOW you can see a commit introducing a Ruby script (that doesn't happen very often, so it's an unusual change) and another one introducing the parser.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,157 @@
...
+        return NotImplemented
+
+    @pydantic.field_validator("url")
+    def validate_git_url(cls, url: str) -> str:
+        """Validate that the git URL is a valid git URL."""
+        parsed_url = urlparse(url)
+        if parsed_url.scheme != "https":
+            raise ValueError("GIT dependencies must use HTTPS protocol")
+
+        # TODO: more validation
```
eskultety, line 78: I'd drop this comment, we'll do more validation when we know we need to do it. At this point in time we think this is sufficient. The reason why I asked about validation in one of my earlier comments is that this only previously checked the protocol of the URL.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,157 @@
...
+    @property
+    def purl(self) -> str:
+        """Get purl for this git dependency."""
+        return NotImplemented
+
+    @pydantic.field_validator("url")
+    def validate_git_url(cls, url: str) -> str:
+        """Validate that the git URL is a valid git URL."""
+        parsed_url = urlparse(url)
+        if parsed_url.scheme != "https":
```
eskultety, line 75: We should probably get a bit more defensive and raise an exception if `urlparse` fails to parse the URL even though bundler seems to fail correctly on an invalid URL:
```
/usr/share/gems/gems/bundler-2.2.33/lib/bundler/vendor/uri/lib/uri/rfc3986_parser.rb:67:in `split': bad Bundler::URI(is not Bundler::URI?): "https://gi#t%hub.com/3scale/json-schema.git" (Bundler::URI::InvalidURIError)
	from /usr/share/gems/gems/bundler-2.2.33/lib/bundler/vendor/uri/lib/uri/rfc3986_parser.rb:73:in `parse'
	from /usr/share/gems/gems/bundler-2.2.33/lib/bundler/vendor/uri/lib/uri/common.rb:234:in `parse'
	from /usr/share/gems/gems/bundler-2.2.33/lib/bundler/source/git.rb:305:in `uri_hash'
	from /usr/share/gems/gems/bundler-2.2.33/lib/bundler/source/git.rb:344:in `git_scope'
	from /usr/share/gems/gems/bundler-2.2.33/lib/bundler/source/git.rb:227:in `cache_path'
	from /usr/share/gems/gems/bundler-2.2.33/lib/bundler/source/git.rb:324:in `git_proxy'
	from /usr/share/gems/gems/bundler-2.2.33/lib/bundler/source/git.rb:236:in `revision'
	from /venv/lib64/python3.9/site-packages/cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb:24:in `block in <main>'
	from /venv/lib64/python3.9/site-packages/cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb:11:in `each'
	from /venv/lib64/python3.9/site-packages/cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb:11:in `<main>'
```

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,157 @@
...
+        )
+        raise PackageRejected(reason=reason, solution=solution)
+
+    bundler_package = files("cachi2.core.package_managers.bundler")
+    lockfile_parser_script = str(bundler_package.joinpath("scripts").joinpath("lockfile_parser.rb"))
+    try:
+        output = run_cmd(cmd=[lockfile_parser_script], params={"cwd": package_dir.path})
+    except subprocess.CalledProcessError:
+        reason = "Execution of Ruby script failed"
+        raise PackageManagerError(reason=reason)
```
eskultety, line 130: Unnecessary `reason` variable, the str literal can easily fit on the same line. Also, `Execution of Ruby script failed` is too generic and unhelpful, maybe `"Failed to parse <path to Gemfile.lock>"` ?

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,157 @@
...
+        solution = (
+            "Run `bundle init` to generate the Gemfile.\n"
+            "Run `bundle lock` to generate the Gemfile.lock."
+        )
+        raise PackageRejected(reason=reason, solution=solution)
+
+    bundler_package = files("cachi2.core.package_managers.bundler")
+    lockfile_parser_script = str(bundler_package.joinpath("scripts").joinpath("lockfile_parser.rb"))
+    try:
+        output = run_cmd(cmd=[lockfile_parser_script], params={"cwd": package_dir.path})
```
eskultety, line 127: I like how bundler correctly dumps warnings to stderr which in the default configuration of subprocess.run gets ignored and hence doesn't crash the json parser: 
```
Warning: the running version of Bundler (2.2.33) is older than the version that created the lockfile (2.5.10). We suggest you to upgrade to the version that created the lockfile by running `gem install bundler:2.5.10`.
```

cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -0,0 +1,37 @@
...
+      parsed_spec[:url] = spec.source.uri
+      parsed_spec[:revision] = spec.source.revision
+      parsed_spec[:branch] = spec.source.branch
+    when Bundler::Source::Path
+      parsed_spec[:type] = 'path'
+      parsed_spec[:path] = spec.source.path.to_s
+    end
+
+    parsed_specs << parsed_spec
+  end
```
eskultety, line 32: I trust you on this one that it does everything we need from it :) .

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -0,0 +1,215 @@
+import subprocess
```
eskultety, line 1: Unit tests should be introduced along with the code that needs to be tested unlike integration tests which can come at the end because it tests functionality as a completely integrated unit.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,157 @@
...
+    for dep in dependencies:
+        if dep["type"] == "rubygems":
+            result.append(GemDependency.model_validate(dep))
+        elif dep["type"] == "git":
+            result.append(GitDependency.model_validate(dep))
+        elif dep["type"] == "path":
+            _validate_path_dependency_subpath(package_dir, dep["path"])
+            result.append(PathDependency.model_validate(dep))
+
+    return result
```
brunoapimentel, line 148: I was doing some checks trying to break the parsing of the `Gemfile.lock` by adding random strings. Apparently, our solution is robust enough to just ignore the random bits :grin:

But I noticed that an empty or pure gibberish `Gemfile.lock` will have a `[]` result, which essentially means that the prefetching for that specific Bundler project will be a noop. 

Do you think we should fail the request in such cases? Or are there any legitimate cases in which the `Gemfile.lock` will have zero dependencies?

slimreaper35: Zero dependencies is valid (I added a unit test for that as well). But I don't know about completely empty `Gemfile.lock` .

brunoapimentel: From our current parser script output the , there's no way to find out if we have an empty file vs a correctly formed file that has no deps:
```
GEM
  remote: https://rubygems.org/
  specs:

PLATFORMS
  x86_64-linux

DEPENDENCIES

BUNDLED WITH
   2.5.16
```

In the end, this feels like a very cornerish case, so maybe we can just assume this falls under the "outdated lockfile" case.

Dockerfile
slimreaper35: My bad, I should have tested it.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,153 @@
...
+    if not lockfile_path.path.exists() or not gemfile_path.path.exists():
+        reason = "Gemfile and Gemfile.lock must be present in the package directory"
+        solution = (
+            "Run `bundle init` to generate the Gemfile.\n"
+            "Run `bundle lock` to generate the Gemfile.lock."
+        )
+        raise PackageRejected(reason=reason, solution=solution)
+
+    bundler_package = files("cachi2.core.package_managers.bundler")
+    lockfile_parser_script = str(bundler_package.joinpath("scripts").joinpath("lockfile_parser.rb"))
```
eskultety, line 122: This line is unnecessarily long. I am biased to actually question both the existence and usage of  `joinpath` since it returns a new `Path` instance anyway. There are other ways to write it shorter, e.g.:

- `bundler_package / "scripts/lockfile_parser.rb"`
- `Path(bundler_package, "scripts/lockfile_parser.rb")` or (if you want to stick to it) even
- `bundler_package.joinpath("scripts/lockfile_parser.rb")`

Since pathlib allows this flexibility for path concatenation we should agree on a particular style and stick to it.

slimreaper35: How about this ?
```
  bundler_package_path = Path(__file__).parent
  lockfile_parser_script = bundler_package_path / "scripts" / "lockfile_parser.rb"
```

eskultety: > How about this ?
> 
> ```
>     bundler_package_dir = Path(__file__).parent

Not opposing, so you can definitely do that. I like the current approach though because if we move the scripts to a different location and group them all centrally together, one only needs to tweak portion of the package location identifier with this we'd have to ditch the path. That said, unit tests would fail in such case either way, so while I do have a slight preference, ^this proposal makes perfect sense too.

>     lockfile_parser_script = bundler_package_dir / "scripts" / "lockfile_parser.rb"

why do we need to split those 2 string literals instead of just writing "scripts/lockfile_parser.rb"?

slimreaper35: I changed it to `scripts_dir,` so the name of the variable won't need to be changed.

cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env ruby
+
+require 'bundler'
+require 'json'
```
brunoapimentel, line 4: I'm getting this error by using an image I build with the current version of the Dockerfile in this PR:
```
2024-09-03 17:59:53,529 ERROR The command "/venv/lib64/python3.9/site-packages/cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb" failed
2024-09-03 17:59:53,529 ERROR STDERR:
<internal:/usr/share/rubygems/rubygems/core_ext/kernel_require.rb>:85:in `require': cannot load such file -- json (LoadError)
	from <internal:/usr/share/rubygems/rubygems/core_ext/kernel_require.rb>:85:in `require'
	from /venv/lib64/python3.9/site-packages/cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb:4:in `<main>'
2024-09-03 17:59:53,529 ERROR PackageManagerError: Failed to parse /target/Gemfile.lock

```

Do we also need to install a `json` dependency?

cachi2/core/package_managers/bundler/parser.py
brunoapimentel: Yeah, apparently this is the case. I solved it by doing a `gem install json`. The problem was the amount of dependencies I had to install to get this done :smiling_face_with_tear::

```
# dnf install -y make gcc redhat-rpm-config
```

cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb
slimreaper35: Ah, I think this is the problem:
` --setopt install_weak_deps=0 `

eskultety: `rubygems-json` is enough to pull in the rest of the strong dependencies AND to parse a lockfile.

cachi2/core/package_managers/bundler/parser.py
slimreaper35: Do you mean instead of `ValueError` ?

eskultety: Right, pretty much.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,153 @@
...
+    json_output = json.loads(output)
+
+    bundler_version: str = json_output["bundler_version"]
+    log.info("Package %s is bundled with version %s", package_dir.path.name, bundler_version)
+    dependencies: list[dict[str, str]] = json_output["dependencies"]
+
+    result: ParseResult = []
+    for dep in dependencies:
+        if dep["type"] == "rubygems":
+            result.append(GemDependency.model_validate(dep))
```
a-ovchinnikov, line 137: Minor observation, non-blocking, non-actionable unless you have bandwidth and planning on another set of tweaks:  `dep` seems to be a flat dictionary, I would have just gone straight to instantiating a model from it: `GemDependency(**dep)`.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,153 @@
...
+        elif dep["type"] == "git":
+            result.append(GitDependency.model_validate(dep))
+        elif dep["type"] == "path":
+            _validate_path_dependency_subpath(package_dir, dep["path"])
+            result.append(PathDependency.model_validate(dep))
```
a-ovchinnikov, line 142: Another observation (non-blocking): I think the validator could be tucked into `PathDependency`. This would require a factory to produce `PathDependency` subclasses with a bound `package_dir`. Then all three branches will be neatly uniform.

a-ovchinnikov (approved):

brunoapimentel (approved):

eskultety:
>     * I dropped `urlparse` since there is no validation of URL at all, `urlparse` accepts empty strings btw
>       I wouldn't rely on the bundler library here, neither

Good to know, anything in particular to share?

slimreaper35:
> Good to know, anything in particular to share?

Yeah, forgot to add examples.
I was testing our `lockfile_parser.rb` script what kind of malformed URLs it can handle, and:
- `http` is supported by bundler, well okay
- but `remote:  //github.com/3scale/json-schema.git` is valid, also `remote: hello` is apparently valid
-  however, the example you provided `https://gi#t%hub.com/3scale/json-schema.git` does not pass

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -0,0 +1,133 @@
...
+    Represents a git dependency.
+
+    Attributes:
+        url:        The URL of the git repository.
+        revision:   Commit hash.
+        branch:     The branch of the git repository.
+    """
+
+    url: AcceptedUrl
+    revision: str
```
a-ovchinnikov, line 62: Obligatory "Not a blocker" marker. 

It has just struck me, why a field validator and not just `revision: str = Field(pattern=r"^[a-fA-F0-9]{40}$")`?

a-ovchinnikov (approved):

eskultety:
FYI the unit tests fail in local environments if Ruby isn't installed which is currently assumed as a hard dependency and it should be. Therefore, there's something going on with the gating CI, because it clearly has Ruby installed which is wrong, it should be a completely pristine environment.

Sigh. I know see the shortcoming of the solution based on the Ruby script. However, I'm still in favour of it due to its simplicity and clarity over the original solution based on some random 3rd party library (or even worse our own vendored version of it) which most likely will end up as an abandonware . I'd solve the unit test issue with the following:
```
@pytest.mark.skip("Gemfile.lock parser is integration tested due to its dependency on the Ruby runtime")
def test_gemfile_parse() -> None:
    pass
```
in order to:

1. not lose coverage otherwise codecov will complain in its reports that we have an uncovered function which would not be true
2. we need to skip it so that we don't report success on something that doesn't do anything.

Any objections or alternative proposals?

brunoapimentel:
> there's something going on with the gating CI

Ohh, good catch!

> I'd solve the unit test issue with the following:
> 
> ```
> @pytest.mark.skip("Gemfile.lock parser is integration tested due to its dependency on the Ruby runtime")
> def test_gemfile_parse() -> None:
>     pass
> ```
> Any objections or alternative proposals?

We don't we simply mock the `run_cmd` call, and test the rest of the function with mocked output? I assume we're fine with not unit testing the ruby code per se, so this would allow us to test all the rest of the Python code in `bundler/parser.py`.

eskultety:
> ```
> @pytest.mark.skip("Gemfile.lock parser is integration tested due to its dependency on the Ruby runtime")
> def test_gemfile_parse() -> None:
>     pass
> ```

Taking this idea back - integration testing this is not a way. We'd still have to clone a repo with a particular ref just to test these, which doesn't feel right.

> We don't we simply mock the run_cmd call, and test the rest of the function with mocked output? I assume we're fine with not unit testing the lockfile_parser.rb per se, so this would allow us to test all the rest of the Python code in bundler/parser.py.

We could do that, yes. Then again, it's our script that we're going to maintain, so we should test its correctness as well. Continuing with your thought we could have this particular test that mocks `run_cmd` and tests the rest of the python code as you suggest and then we could have an additional test for the Ruby parser which would require Ruby installed in some form which is an idea I discussed privately with @a-ovchinnikov:

- we ought not to rely on a particular interpreter to be installed in our unit test suite and so for that proposed test
- we'd introduce a new test env variable `CACHI2_TEST_EXPENSIVE` which would trigger this test (and potentially others which might take longer to execute and so by default it would be disabled)
- in such a test (we're still talking unit tests) we'd add a check whether we're running in a GH action environment, which has Ruby installed, and execute the script directly; in local dev envs we'd pull a Ruby (Alpine) image and execute the test in a container environment

That said, if everyone is in favour of including expensive tests as well, we can track that idea as a future improvement, doesn't necessarily need to be part of this particular PR.

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -0,0 +1,278 @@
...
+            "URL scheme should be 'https'",
+        ),
+        (
+            LOCKFILE_INVALID_REVISION,
+            "String should match pattern '^[a-fA-F0-9]{40}$'",
+        ),
+        (
+            LOCKFILE_INVALID_PATH,
+            "PATH dependencies should be within the package root",
+        ),
```
eskultety, line 144: Is this black's work? We should be able to fit individual params on a single line, right?

Dockerfile
```diff
@@ -12,6 +12,8 @@ RUN dnf -y install \
     --nodocs \
     git-core \
     python3 \
+    rubygem-bundler \
+	rubygem-json \
```
eskultety, line 16: Tab used for indentation --> should use spaces.

eskultety (commented):
We just need to reach an agreement on how to proceed with the unit tests that currently depend on Ruby, otherwise LGTM.

tests/unit/package_managers/bundler/test_parser.py
slimreaper35: That was caused by a comma at the end of the string.

slimreaper35:
I just added a commit that skips the tests that depend on Ruby.
I did not want to mock the `run_cmd` output with huge JSON output because it is, in the end, a temporary change.

eskultety:
> I just added a commit that skips the tests that depend on Ruby. I did not want to mock the `run_cmd` output with huge JSON output because it is, in the end, a temporary change.

Doesn't matter. You can define how big you want the JSON (I think you can take a very simple example like 2 dependencies and it won't be that big, will it?), we just want to make sure we cover ideally all python code we introduce. I don't care that much about the output being mocked for now, but I do care about testing our behaviour and agree with @brunoapimentel we should do it.

slimreaper35:
Should I add the change as a separate commit or part of the one that implements parsing?
I thought that adding a separate commit would make it easy to revert it in the future.

eskultety:
> Should I add the change as a separate commit or part of the one that implements parsing? I thought that adding a separate commit would make it easy to revert it in the future.

Same commit, with a justification. We will not revert it, we'll just update the test in the future such that it:
- doesn't mock output of the parser anymore
- only runs by default in the gating CI and 
- is guarded by an environment variable that'd be disabled by default in local development envs

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -0,0 +1,327 @@
...
+    empty_gemfile: RootedPath,
+    rooted_tmp_path: RootedPath,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    lockfile_path = rooted_tmp_path.join_within_root(GEMFILE_LOCK)
+    lockfile_path.path.write_text(LOCKFILE_VALID)
+
+    empty_json = json.loads(EMPTY_JSON)
+
+    empty_json["dependencies"] = [
```
a-ovchinnikov, line 253: An absolutely non-blocking nitpick: technically it is not empty starting from this line on.

a-ovchinnikov (approved):

eskultety (commented):
I think we can simplify the unit tests quite a bit since we're mocking the `run_cmd` results, have a look at the attached diff (passes the unit test just fine) which you can apply if you think I simplified the tests to an acceptable level.

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+    elif error == "LOCKFILE_INVALID_PATH":
+        mocked_json["dependencies"][0].update(
+            {
+                "type": "path",
+                "path": "/root/pathgem",
+            }
+        )
+
+    mock_run_cmd.return_value = json.dumps(mocked_json)
```
eskultety, line 122: I think this is just a marginal improvement over the last revision (e.g. doing json.load only to do json.dumps later instead of using a dictionary in the first place, setting fields that are not being tested in a given parametrize run, etc.), but since this is supposed to be just a temporary solution anyway, I'll let it fly and not block this further.

eskultety (approved):

tests/unit/package_managers/bundler/test_parser.py
slimreaper35: fixed

PR #601, opened 2024-08-15
Template for design docs

brunoapimentel:
This is a very rough draft of a template for merging design docs into the repository. I also added the Rubygems design adapted into the template so we can have a real use case to analyse.

### DIFF
--- a/docs/designs/rubygems-001.md
+++ b/docs/designs/rubygems-001.md
@@ -1,0 +1,547 @@
+# Initial Design for the Rubygems Package Manager
+
+- [Overview](#overview)
+- [Context](#context)
+- [Description](#description)
+    - [Ruby ecosystem overview](#i-ruby-ecosystem-overview)
+    - [Current implementation overview (Cachito)](#ii-overview-of-the-current-implementation-in-cachito)
+    - [Design for the Cachi2 implementation](#iii-design-for-the-implementation-in-cachi2)
+- [Decision](#decision)
+
+## Overview
+
+Design that covers the initial implementation of the Rubygem/Bundler package manager in Cachi2. It takes inspiration from the original implementation done in [Cachito](https://github.com/containerbuildsystem/cachito).
+
+| | |
+|---------|---------|
+|Author |[Michal Šoltis](msoltis@redhat.com) |
+|Co-author |[Bruno Pimentel](bpimente@redhat.com) |
+|Proposed on |August 01, 2024 |
+
+## Context
+In the effort to evolve Cachi2 to be a full solution for the prefetching feature of Cachito, we have decided to implement support to the currently missing package managers: Rubygems and Yarn v1. This design covers only the implementation of Rubygems, and the Yarn v1 design will follow.
+
+## Description
+
+### I. Ruby ecosystem overview
+
+#### Development prerequisites
+In order to execute the commands in the examples below, make sure you have the following packages installed in your
+environment:
+
+```bash
+sudo dnf install rubygems rubygems-bundler
+```
+
+Or use the official Ruby image from Docker hub:
+```bash
+podman run --rm -it docker.io/library/ruby:3.3.3 bash
+```
+
+#### Project structure
+```bash
+bundle init # creates Gemfile in the current directory
+bundle lock # creates Gemfile.lock in the current directory
+```
+
+```bash
+├── .bundle
+│   └── config
+├── Gemfile
+├── Gemfile.lock
+├── vendor/cache
+```
+
+#### Glossary
+- **Gemfile**: A file that specifies the gems that your project depends on and their versions. Bundler uses this file
+to install the correct versions of gems for your project.
+
+  ```ruby
+  source "https://rubygems.org"
+
+  gem "rails", "= 6.1.7"
+  ```
+
+- **Gemfile.lock**: A file that locks the versions of gems that are installed for your project. Bundler uses this file
+to ensure that the correct versions of gems are installed consistently across different environments.
+
+- **RubyGems**: General package manager for Ruby. Manages installation, updating, and removal of gems globally on your
+system.
+
+  ```bash
+  gem --help
+  ```
+
+- **Bundler**: Dependency management tool for Ruby projects.
+Ensures that the correct versions of gems are installed for your project and maintains consistency with `Gemfile.lock`.
+
+  ```bash
+  bundler --help
+  ```
+
+- **Gem**: A package that can be installed and managed by Rubygems. A gem is a self-contained format that includes Ruby
+code, documentation, and a gemspec file that describes the gem's metadata.
+
+- **{gem}.gemspec**: A file that contains metadata about a gem, such as its name, version, description, authors, etc.
+RubyGems uses it to install, update, and uninstall gems.
+
+  ```ruby
+  Gem::Specification.new do |spec|
+    spec.name        = "example"
+    spec.version     = "0.1.0"
+    spec.authors     = ["Nobody"]
+    spec.email       = ["ruby@example.com"]
+    spec.summary     = "Write a short summary, because RubyGems requires one."
+  end
+  ```
+
+#### Dependency types
+There are four types of
+[sources](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb#L48) for dependencies
+in the `Gemfile.lock` file:
+
+##### Gem dependencies
+Regular gem dependencies are located at the source URL, in our case, always <https://rubygems.org>. Each gem can be
+accessed by its name and version - rubygems.org/gems/`<name>`-`<version>`.gem
+
+Example of a gem dependency in the `Gemfile.lock` file:
+
+```bash
+GEM
+ remote: https://rubygems.org/
+ specs:
+  rails (6.1.4)
+    # transitive dependencies
+    actioncable (= 6.1.4)
+    actionmailbox (= 6.1.4)
+    actionmailer (= 6.1.4)
+    actionpack (= 6.1.4)
+    actiontext (= 6.1.4)
+    actionview (= 6.1.4)
+    activejob (= 6.1.4)
+    activemodel (= 6.1.4)
+    activerecord (= 6.1.4)
+    activestorage (= 6.1.4)
+    activesupport (= 6.1.4)
+    bundler (>= 1.15.0)
+    railties (= 6.1.4)
+    sprockets-rails (>= 2.0.0)
+```
+
+##### Git dependencies
+Example of a Git dependency in the `Gemfile.lock` file:
+
+```
+GIT
+  remote: https://github.com/porta.git
+  revision: 779beabd653afcd03c4468e0a69dc043f3bbb748
+  branch: main
+  specs:
+    porta (2.14.1)
+```
+
+##### Path dependencies
+Example of a path dependency in the `Gemfile.lock` file:
+
+```
+PATH
+  remote: some/pathgem
+  specs:
+    pathgem (0.1.0)
+```
+
+All path dependencies must be in the project directory. Bundler
+[does not copy](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/source/path.rb#L83) those
+dependencies that are already within the root directory of the project.
+
+##### Plugins
+Installing a plugin, even when on a folder that is a Bundler project, doesn't seem to affect the `Gemfile.lock`. The
+plugin seems to be installed by default in the `$PWD/.bundle/`. The `Gemfile.lock` does have a section for plugins,
+though, so further investigation would be needed. This initial investigation was done with the plugins listed under
+[Known Plugins](https://bundler.io/guides/plugins.html).
+
+*Don't confuse Bundler plugins with [RubyGems plugins](https://guides.rubygems.org/plugins/). The latter are meant to
+extend the functionality of `gem` itself, and don't seem to have any impact on Bundler directly.*
+
+
+#### Platforms
+Some gems may contain pre-compiled binaries that provide native extensions to the Ruby package. Any gem declared in the
+`Gemfile` can be limited to specific
+[platforms](https://bundler.io/v2.5/man/gemfile.5.html#PLATFORMS), making Bundler ignore it in case the project is
+being built on a non-matching platform:
+
+```ruby
+gem "nokogiri",   platforms: [:windows_31, :jruby]
+```
+
+Here's an example of how a the `PLATFORM` section looks like in the `Gemfile.lock`:
+
+```
+PLATFORMS
+  arm64-darwin-20
+  arm64-darwin-21
+  arm64-darwin-22
+  ruby
+  x86_64-darwin-18
+  x86_64-darwin-20
+  x86_64-darwin-21
+  x86_64-darwin-22
+  x86_64-linux
+```
+
+In case a user wants to force all the binaries to be compiled from source, the `BUNDLE_FORCE_RUBY_PLATFORM` environment
+variable can be used.
+
+#### Dev dependencies
+When adding a Gem into a Gemfile, the user might opt to nest them under a specific
+[group](https://bundler.io/guides/groups.html). The name of the group can be any string, but the usual groups tend to
+be common labels such as `:test`, `:development` or `:production`.
+
+Here's how it looks like in a `Gemfile`:
+
+```ruby
+# :default group
+gem 'nokogiri'
+
+group :test do
+  gem 'faker'
+  gem 'rspec'
+end
+```
+
+Another way to declare a dependency in the `:development` group is to
+[add it](https://guides.rubygems.org/specification-reference/#add_development_dependency) to the `Gem::Specification`,
+which is usually declared in the `.gemspec` file. This means we can safely assume that all dependencies under
+`:development` are dev dependencies.
+
+#### Dependency checksums
+The support to checksums in the `Gemfile.lock` is still in development, and currently is an
+[opt-in feature](https://github.com/rubygems/rubygems/pull/7217). To enable it, we need to manually add a `CHECKSUMS`
+section in the `Gemfile.lock`:
+
+```shell
+# manually add `CHECKSUMS` section somewhere in the Gemfile.lock
+vim Gemfile.lock
+# install any gem
+bundle add rails --version "6.1.7"
+# check the Gemfile.lock /o\
+cat Gemfile.lock
+```
+
+Example of a checksum section in the `Gemfile.lock`:
+
+```
+CHECKSUMS
+  actioncable (6.1.7) sha256=ee5345e1ac0a9ec24af8d21d46d6e8d85dd76b28b14ab60929c2da3e7d5bfe64
+  actionmailbox (6.1.7) sha256=c4364381e724b39eee3381e6eb3fdc80f121ac9a53dea3fd9ef687a9040b8a08
+  actionmailer (6.1.7) sha256=5561c298a13e6d43eb71098be366f59be51470358e6e6e49ebaaf43502906fa4
+  actionpack (6.1.7) sha256=3a8580e3721757371328906f953b332d5c95bd56a1e4f344b3fee5d55dc1cf37
+  actiontext (6.1.7) sha256=c5d3af4168619923d0ff661207215face3e03f7a04c083b5d347f190f639798e
+  actionview (6.1.7) sha256=c166e890d2933ffbb6eb2a2eac1b54f03890e33b8b7269503af848db88afc8d4
+```
+
+This feature is available since Bundler [v2.5.0](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb#L55),
+from this [PR](https://github.com/rubygems/rubygems/pull/6374) being merged on Oct 21, 2023.
+
+### II. Overview of the current implementation in Cachito
+
+[cachito/workers/pkg_mangers/rubygems.py](https://github.com/containerbuildsystem/cachito/blob/master/cachito/workers/pkg_managers/rubygems.py)
+
+Most work is already done by parsing the `Gemfile.lock` file, which pins all dependencies to exact versions. The only
+supported source for gem dependencies to be fetched from is <https://rubygems.org>. Git dependencies are specified
+using a repo URL and pinned to a commit hash. Path dependencies are specified using a local path.
+
+To avoid arbitrary code execution, Bundler **is not used** to download dependencies. Instead, as stated above, Cachito
+parses `Gemfile.lock` file directly and download the gems from <https://rubygems.org>.
+
+**Note**: parsing `Gemfile.lock` is done via [gemlock-parser](https://github.com/containerbuildsystem/gemlock-parser),
+which is vendored from
+[scancode-toolkit](https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/gemfile_lock.py).
+
+Source code for "official" Bundler lockfile parsing in Ruby:
+<https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb>
+
+### III. Design for the implementation in Cachi2
+
+#### Prefetching
+
+Running a bundler command to fetch the dependencies always executes the `Gemfile`, which is arbitrary Ruby code.
+Executing arbitrary code is a security risk and makes it impossible to assert that the resulting SBOM is accurate
+(since any random package can be fetched from the Internet during the prefetch). This means that we need to implement
+custom code to fetch the dependencies.
+
+In the `Gemfile.lock`, all Gems that come from the same remote URL are grouped under the same block:
+```
+GEM
+ remote: https://rubygems.org/
+ specs:
+  rails (6.1.4)
+  json-schema (1.2.1)
+```
+
+A Gem can be fetched from its original location by using the following template:
+```ruby
+"https://#{remote}/gems/#{name}-#{version}.gem"
+```
+
+We should also leverage the existing code used to perform parallel downloads based on `asyncio` to download the necessary
+Gems from the internet.
+
+##### Output folder structure
+
+Bundler has a built-in feature to cache all dependencies locally. This is done with the `bundle cache --all` command or
+`bundle package --all` alias. In order to make bundler use the prefetched dependencies during the build, Cachi2 needs
+to recreate the exact same folder structure as bundler does.
+
+Here's an example of how the output folder should look like:
+
+```bash
+$ ls vendor/cache
+
+actioncable-6.1.7.gem
+date-3.3.4.gem
+json-schema-26487618a684
+nokogiri-1.16.6.gem
+tzinfo-2.0.6.gem
+```
+
+Notice that all the `.gem` dependencies are kept in their original format, and Git dependencies are just plain clones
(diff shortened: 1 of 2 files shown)


docs/designs/template-000.md
```diff
@@ -0,0 +1,9 @@
+# Title
+
+## Overview
+
+## Context
+
+## [Free-form sections]
+
+## Decision
```
a-ovchinnikov, line 9: What do you think about moving "Decision" up, right after "Overview"? If the doc is to be used as a quick reference it would be nice to have it closer to the problem statement. Also it looks like not every design requires a decision, "Implement as described here" seems to be implied by the type of the document, maybe it should be marked as [Optional]? Or maybe even made a part of free-form sections when there are several possible strategies? Overall it looks like "Context" will ultimately dictate the structure of everything that follows.

docs/designs/template-000.md
```diff
@@ -0,0 +1,9 @@
+# Title
+
+## Overview
```
a-ovchinnikov, line 3: Would hints like "A concise, 5 sentences max, statement of the problem" be useful?

docs/designs/template-000.md
a-ovchinnikov, line 1: Filenames in the draft imply naming scheme with shared counter. This should probably be documented somewhere.

a-ovchinnikov (commented):
I like the mostly free-form approach proposed here.

docs/designs/template-000.md
eskultety: > Also it looks like not every design requires a decision, "Implement as described here" seems to be implied by the type of the document, maybe it should be marked as [Optional]?

Agreed. I would probably go as far as omitting the `Decision` section completely for now. Remember, we're dealing with design docs for package managers, so the decision inherently is to "implement support for one". We can add it optional in a follow-up patch, we may get clearer picture and overall better feeling of this once we start actively using the template.

docs/designs/template-000.md
eskultety: Why do we want to limit this to only 5 sentences for a pkg manager support design doc? A given pkg manager ecosystem will typically be extremely complex, so a solid background feels very appropriate for this type of document.

docs/designs/rubygems-001.md
```diff
@@ -0,0 +1,547 @@
...
+- [Decision](#decision)
+
+## Overview
+
+Design that covers the initial implementation of the Rubygem/Bundler package manager in Cachi2. It takes inspiration from the original implementation done in [Cachito](https://github.com/containerbuildsystem/cachito).
+
+| | |
+|---------|---------|
+|Author |[Michal Šoltis](msoltis@redhat.com) |
+|Co-author |[Bruno Pimentel](bpimente@redhat.com) |
```
eskultety, line 17: I think a single `Authors` field is enough and then list all participants in an order respecting their amount of contribution, credit, etc.

docs/designs/rubygems-001.md
```diff
@@ -0,0 +1,547 @@
...
+- add integration and e2e tests
+- add documentation
+
+### Potential follow-up features
+- implement checksum parsing and validation when prefetching from the registry
+- downloading the Bundler version specified in the `Gemfile.lock`
+- support for pre-compiled binaries (platforms other than `ruby`)
+- Gemfile.lock checksum validation (blocked by pending official support)
+- reporting dev dependencies
+- proper support for plugins
```
eskultety, line 544: This is just a summary of the implementation specification and as such is IMO redundant as it provides very little information compared to the specification section.

docs/designs/rubygems-001.md
```diff
@@ -0,0 +1,547 @@
...
+The URL schema in the default rubygems registry seems to follow this format:
+
+```ruby
+# Plain Ruby Gem
+"https://rubygems.org/gems/#{name}-#{version}.gem"
+
+# Platform-specific Gem
+"https://rubygems.org/gems/#{name}-#{version}-#{platform}.gem"
+```
+
```
eskultety: I think this scoping you have here is **actually** the decision, because we're purposefully limiting the implementation, so either it stays part of the specification or we move it to the decision (I'm ambivalent on this at the moment based on a single example).

docs/designs/rubygems-001.md
```diff
@@ -0,0 +1,547 @@
...
+environment:
+
+```bash
+sudo dnf install rubygems rubygems-bundler
+```
+
+Or use the official Ruby image from Docker hub:
+```bash
+podman run --rm -it docker.io/library/ruby:3.3.3 bash
+```
```
eskultety: This whole Ruby ecosystem overview (up to cachi2 impl.) is IMO the **actual** overview you want to have in this design doc instead of what it currently reads.

docs/designs/template-000.md
```diff
@@ -0,0 +1,9 @@
+# Title
+
```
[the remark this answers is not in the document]
eskultety: for a reference, we can get inspired with probably the origins of ADR = RFCs  --> https://www.rfc-editor.org/rfc/rfc5143, https://www.rfc-editor.org/rfc/rfc7322 just to name a few tiny examples dealing with format of the RFCs. 

Other notable mention: https://www.rfc-editor.org/rfc/rfc7841#section-3

docs/designs/template-000.md
brunoapimentel: It also feels to me that "Overview" is something more akin to TL;DR. So I wouldn't want it to be a long thing, I want it to be something I can read in under 2min.

docs/designs/template-000.md
brunoapimentel: We can mark Decision as optional. I agree it feels very out of place for initial package manage designs.

docs/designs/template-000.md
```diff
@@ -0,0 +1,9 @@
+# Title
```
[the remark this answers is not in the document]
brunoapimentel: Yes, that was in the plan. This was just a rough draft as I tried to wrap my mind around the template.

docs/designs/rubygems-001.md
```diff
@@ -0,0 +1,547 @@
...
+
+## Overview
+
+Design that covers the initial implementation of the Rubygem/Bundler package manager in Cachi2. It takes inspiration from the original implementation done in [Cachito](https://github.com/containerbuildsystem/cachito).
+
+| | |
+|---------|---------|
+|Author |[Michal Šoltis](msoltis@redhat.com) |
+|Co-author |[Bruno Pimentel](bpimente@redhat.com) |
+|Proposed on |August 01, 2024 |
```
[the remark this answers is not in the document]
brunoapimentel: This was a continuation of the Authors field, which I also struggled to decide if it was worth it or not. What I had in mind was that, although Git tracks info such as commit author and date, and Github tracks PRs author and merge date, if we happen to move this doc around, all of this gets buried under a layer of changes. My reasoning was that this info here would make it easy to reference.

I think "Last updated" is a different info that adds to "proposed date". But in the end, I really don't have a strong feeling for the date fields, so I'd agree to drop it altogether.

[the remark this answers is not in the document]
brunoapimentel: (I also feel the same for authors... I'd be ok with dropping it)

docs/designs/rubygems-001.md
brunoapimentel: I think the summary is really... just a summary. It's a way to help the reader bringing to mind all the info that was more thoroughly explained before. It can be optional as well, as I don't see it being useful for smaller docs.

There is also another benefit for this specific section here: clear separation of what's in scope and what's not. But maybe scoping doesn't belong to the design doc.

docs/designs/template-000.md
a-ovchinnikov: I'd say "Overview" is for a brief introduction to a problem, "Context" is for a solid background.

docs/designs/template-000.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Overview
+
+## Context
```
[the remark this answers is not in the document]
a-ovchinnikov: "Overview" should serve the same function as an abstract in a paper: to give a brief summary of the main body of text. It is necessary because it provides means to skip or postpone reading of the entire doc. I think naming could be discussed, but having a TL;DR section is important.

docs/designs/template-000.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Description
```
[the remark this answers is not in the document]
a-ovchinnikov: What about "Design"? This is a design doc after all.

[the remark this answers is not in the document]
brunoapimentel: I dropped this section in the following commit. The reason was that I felt it didn't add any value, but added one more level of nesting, which looks horrible IMO.

docs/designs/template-000.md
eskultety: But for pkg manager design the overview in that case is we need support for one and it uses CLI commands X,Y,Z. Specifically with pkg managers which is a complex topic on its own I don't think a short overview will give you what you need, to understand it, one needs to read further. Also note that I'm not sure if there's a formal definitiion of overview, IOW overview can be arbitrarily long which the context complies with as well in that we're not covering the whole pkg manager ecosystem, we're only covering what makes sense for us which is still technically an overview == a subset of information necessary to back the implementation.

docs/designs/template-000.md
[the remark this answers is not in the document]
eskultety: Then it should be called abstract, because I think the definition of abstract is much stricter than that of an overview, what we're doing with these design docs wrt/ pkg manager background is also an overview.
The question still remains, should an abstract be mandatory for the original design docs? I can imagine them being mandatory for the obsolete/update docs, but not for the main document, the title kinda already hints at what it's about --> implemeting new backend.

docs/designs/rubygems-001.md
eskultety: > There is also another benefit for this specific section here: clear separation of what's in scope and what's not. But maybe scoping doesn't belong to the design doc.

arguably it does, because a given pkg manager is a massive ecosystem with loads of variables and corner cases, so the design doc needs to outline what is being done and what's not, it just needs an appropriate format. I'd refrain from mentioning follow-ups as those IMO don't belong to a formal docs, one's imagination can be very wild when it comes to follow-ups so let the future decide what the follows are going to be. For the rest, an optional summary section would IMO be acceptable, worst case scenario, we'll decide it was wrong and drop it and propose a refined version of the template....RFC did the same...not once

docs/designs/rubygems-001.md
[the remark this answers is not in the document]
eskultety: Let's drop them both then, after all this is supposed to be community effort, so whatever design we merge, the whole team/community stands by it. Naturally, when it comes to giving credit, git can help if needed. I think the whole purpose of Authors/Dates etc. fields is for cases where these docs are not tracked in a VCS (is there a git for RFCs, I've never seen one) and so it needs to bear that information otherwise it's automatically public domain :) .

docs/designs/template-000.md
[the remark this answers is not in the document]
a-ovchinnikov: Mandatory fields mean that these fields are applicable to all designs, determining a set of universally applicable ones would require significant effort to analyze all existing designs and distill out all common parts. There is also a big chance of introducing too many unnecessary fields. I've seen mandatory templates with more than half fields ending up being marked as "Non-applicable" which created lots of visual noise and made it easy to miss something important. In some cases it was more productive not to follow the template at all. I don't think it is a good idea to set rules which will not be followed (for whatever reason). I am also concerned with extra maintenance and cognitive load which mandatory fields introduce so would suggest keeping them to a reasonable minimum.

docs/designs/rubygems-001.md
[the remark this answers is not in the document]
brunoapimentel: I've already had to ask clarification for an author of an old design doc, so that's the main point in having an easy-to-reference author field here, IMO. But we can always dig that up from the commits, so let's just drop these.

docs/designs/rubygems-001.md
brunoapimentel: The follow-ups mentioned here are either things that were mentioned at some point in the design, but were marked as out-of-scope since we want to keep the implementation very basic. They are all features that have been similarly implemented in other package managers.

This is part of the summary, as I see it: "the design touches all these features, this is the subset we want to implement, and this is the subset we don't want to for the first version". Maybe we can just call the section as "out of the initial scope".

docs/designs/template-000.md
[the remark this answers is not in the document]
brunoapimentel: Yes, this was another stab I took to see how it would look like. I will definitely improve and write guidelines for it.

docs/designs/template-000.md
taylormadore: IMO the "Decision" section here should be where we describe what we're going to implement given the "Context" described above. For a design, I would expect it to at a minimum include what is in-scope and out-of-scope.

We might just need a different word to describe it.

docs/designs/template-000.md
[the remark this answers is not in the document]
eskultety: @a-ovchinnikov I agree with that reasoning so as long as we don't end up with all fields optional and their intent is clearly defined, we're good.

docs/designs/rubygems-001.md
brunoapimentel: I feel like there's a benefit to it... it clearly states, in a summarized form, what the initial deliverable is.

docs/designs/0000-template.md
```diff
@@ -0,0 +1,23 @@
+# Title
+
+Updated by: [0000-this-doc-does-not-exist]()
```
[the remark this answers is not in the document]
eskultety: Yes, we do.

docs/designs/0001-golang-1-21.md
```diff
@@ -0,0 +1,206 @@
+# Go 1.21 toolchains
+
+## Overview
```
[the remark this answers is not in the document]
brunoapimentel: I proposed the change of this section to `Introducion` based on what I saw from the RFC examples Erik linked. But looking at it again, I think it is even worse to convey the meaning of TL;DR we are expecting from this section.

But point taken, this is currently inconsistent.

docs/designs/template-000.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Decision
```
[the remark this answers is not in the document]
brunoapimentel: I was wondering the same, I will try to include something in another patch today.

docs/designs/0001-golang-1-21.md
```diff
@@ -0,0 +1,206 @@
...
+
+Go 1.21 introduced a new feature called _toolchains_ which allow usage of different Go toolchains to be used for different modules simultaneously instead of the default bundled Go toolchain (i.e. the bundled `go` binary). The toolchain feature is enabled by the `toolchain` keyword which can either be in the `go.mod` file or the `go.work` (workspaces) file [[1]](#references).
+
+This document describes how this feature works and how it impacts the current implementation in Cachi2, and proposes a few ways the implementation can be approached.
+
+## Context
+
+[What was the context for this doc? I can't think of anything important to mention here, and that's why I marked this section as optional in the template].
+
+## How it works
```
eskultety, line 13: This is too free-form. I believe we need to standardize on this H2 heading. Also, I'm still fairly convinced that the paragraphs that follow this heading are **the overview** of the 1.21 changes that are relevant to the cause, it doesn't say anything about how we're going to solve it - that's good, because that will come later in `Proposed solutions`!

docs/designs/0001-golang-1-21.md
```diff
@@ -0,0 +1,206 @@
...
+	"GoMod": "/tmp/gomod/cache/download/golang.org/x/text/@v/v0.0.0-20170915032832-14c0d48ead0c.mod",
+	"Zip": "/tmp/gomod/cache/download/golang.org/x/text/@v/v0.0.0-20170915032832-14c0d48ead0c.zip",
+	"Dir": "/tmp/gomod/golang.org/x/text@v0.0.0-20170915032832-14c0d48ead0c",
+	"Sum": "h1:qgOY6WgZOaTkIIMiVjBQcw93ERBE4m30iBm00nkL0i8=",
+	"GoModSum": "h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ="
+}
+...
+```
+
+## Proposed solutions
```
eskultety, line 102: I think some standardization is needed also for ^this H2 heading, but I agree the rest of H3+ headings can be left free form and not limit the writer.

docs/designs/0001-golang-1-21.md
```diff
@@ -0,0 +1,206 @@
...
+`GOPROXY=file://${GOMODCACHE}/cache/download`
+
+#### Advantages:
+- we'll be properly supporting Go 1.21 features going in the future
+- in the very unprobable theory we might not need to keep bumping Go versions in our container image **that often** if users make use of toolchains properly, because if they set `go 1.21` and then `toolchain 1.24.0` (provided 1.24 is out) then the 1.21 binary will be able to download a newer toolchain and use it for compiling; however, this whole idea crumbles like a house of cards when it comes to features that include new keywords
+
+#### Disadvantages:
+- potentially failing builds due to transitively incompatible (i.e. with indirect dependencies) `go` and `toolchain` versions set - but this is a user problem in general though, not cachi2's (mentioning it just in case)
+
+## Decision
```
eskultety, line 143: Good use of `Decision`!

docs/designs/0001-golang-1-21.md
```diff
@@ -0,0 +1,206 @@
...
+
+[5] Snippet from the docs:
+>  toolchain downloads fail for lack of verification if GOSUMDB=off. GOPRIVATE and GONOSUMDB patterns do not apply to the toolchain downloads.
+
+[6] Snippet from the docs:
+> A module cache may be used directly as a file proxy: `GOPROXY=file://$(go env GOMODCACHE)/cache/download`
+
+## Resources
+- https://go.dev/doc/toolchain
+- https://go.dev/ref/mod#environment-variables
```
eskultety, line 195: I know I wrote the content, in retrospec ^this should have somehow be part of `References`, no need for another section, admittedly I was trying to be unnecessarily formal, just an FYI.

docs/designs/0001-golang-1-21.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Apendix
```
eskultety, line 197: Similarly here, this should have been part of the broader context sections, or overview or whatever we end up calling it.

docs/designs/0000-template.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Introduction
+
+This section should bring a short summary of what this document is covering, the main points it is going to describe, and why it is important. Ideally should be kept under two paragraphs.
+
+## Context [optional]
+
+Explains the context outside the mere technical aspects of the design that existed when it was written. It must make clear what other influences where at play that affected the decisions that were taken. In case there is nothing relevant to mention about the context, this section can be skipped.
+
+## [Free-form sections]
```
eskultety, line 17: I responded to the Go sample, I don't think this main H2 section should be free-form, its subsections sure can be.

docs/designs/0000-template.md
```diff
@@ -0,0 +1,23 @@
...
+
+The actual content of the design should be placed under one or more sections that can be named freely. Note that they should be level 2 headings (H2), but each section can also be subdivided by nested headings (H3, H4, H5 or H6) if needed. Only a single level 1 heading (H1) should exist, and it is strictly reserved for the title.
+
+## Decision/Outcome [pick one]
+
+Document here what decision was made, or what outcome is expected out of the design that was created. In case several options were presented in the design, this section must clearly state which one was chosen and the reasons for it. In case of a new package manager design, or a completely new feature to an existing package manager, this section can only include a brief summary of what the implementation will cover, since it is implied that the decision.
```
eskultety, line 23: nitpick: Is the description of the intended content for these sections being purposefully done so that it can be rendered online? I mean, the only thing I expect from a template is to be copy-pasted, renamed, filled out and proposed. So, if we make these description commentaries, people can even keep those comments in their proposed doc and that information won't be rendered, just a  bit of thinking out loud.

docs/designs/0000-template.md
```diff
@@ -0,0 +1,23 @@
...
+
+This section should bring a short summary of what this document is covering, the main points it is going to describe, and why it is important. Ideally should be kept under two paragraphs.
+
+## Context [optional]
+
+Explains the context outside the mere technical aspects of the design that existed when it was written. It must make clear what other influences where at play that affected the decisions that were taken. In case there is nothing relevant to mention about the context, this section can be skipped.
+
+## [Free-form sections]
+
+The actual content of the design should be placed under one or more sections that can be named freely. Note that they should be level 2 headings (H2), but each section can also be subdivided by nested headings (H3, H4, H5 or H6) if needed. Only a single level 1 heading (H1) should exist, and it is strictly reserved for the title.
```
eskultety, line 19: I'm not convinced there should be `1+` H2 headings for this, the design should be covered by a single H2 section and N subsections, otherwise we'll break the structure which we're trying to standardize somewhat I think.

eskultety (commented):
I think we're converging with this, maybe one more iteration :). Kudos for the inclusion of the Go sample, it's always better to have 2 different docs to compare.

docs/designs/0001-golang-1-21.md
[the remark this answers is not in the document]
brunoapimentel: `Abstract`, although when I look the word up in online dictionaries, they describe it as a `Summary`, so that could be another option.

docs/designs/template-000.md
brunoapimentel: Resolving this since we moved this discussion [here](https://github.com/containerbuildsystem/cachi2/pull/601#discussion_r1723220861).

docs/designs/template-000.md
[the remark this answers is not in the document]
brunoapimentel: Added the verbose info, so I'm resolving this.

docs/designs/rubygems-001.md
```diff
@@ -0,0 +1,547 @@
...
+    - [Ruby ecosystem overview](#i-ruby-ecosystem-overview)
+    - [Current implementation overview (Cachito)](#ii-overview-of-the-current-implementation-in-cachito)
+    - [Design for the Cachi2 implementation](#iii-design-for-the-implementation-in-cachi2)
+- [Decision](#decision)
+
+## Overview
+
+Design that covers the initial implementation of the Rubygem/Bundler package manager in Cachi2. It takes inspiration from the original implementation done in [Cachito](https://github.com/containerbuildsystem/cachito).
+
+| | |
```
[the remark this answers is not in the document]
brunoapimentel: Table dropped, resolving.

docs/designs/rubygems-001.md
brunoapimentel: We dropped the authors altogether, resolving.

docs/designs/rubygems-001.md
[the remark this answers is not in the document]
brunoapimentel: Dropped both authors and date, resolving this.

docs/designs/0000-template.md
brunoapimentel: I agree this looks good on the template, as it is more standardized, but having a large document with many subsections forced into a H2 brings a horrible result, at least IMO. As an example, check the Ruby design in this [commit](https://github.com/containerbuildsystem/cachi2/pull/601/commits/46e89a81d206ee3221e7a6a1bf0926408a9dae4e?short_path=d9b9f40#diff-d9b9f40563b29b3b51c53bc4ec6ea3b9a63fc3fadd81c8f00bb91a535561ad39), and how I changed it in the following ones. For smaller docs, I believe a single H2 works perfectly.

I am not sure what's the best option here.  If nobody minds the excessive nesting (and the hard limit on the H5), then let's just make it a single H2. 

(I even considered a separate template for new package manager designs, although I think I still prefer to keep things simple).

docs/designs/0000-template.md
brunoapimentel: Let's concentrate this discussion in this [thread](https://github.com/containerbuildsystem/cachi2/pull/601#discussion_r1724627700).

docs/designs/0000-template.md
eskultety: > with many subsections forced into a H2 brings a horrible result

We can agree that it's a matter of taste. On the other hand...

>  If nobody minds the excessive nesting (and the hard limit on the H5)

... if we ever need to go past H5 then I guess we need to rethink whether a given design really needs it, I don't expect that to be a regular thing, but I might be wrong here.

eskultety:
@brunoapimentel so there hasn't been much movement here since when this was proposed, where do we stand? We've already merged a few designs following no template (or better said, very loosely by copy-pasting from existing material), do you think we'd still benefit from a pretty generic and permissive template on a design doc? Maybe we could abandon that effort for now and only contribute the 2nd patch which actually retroactively adds a design doc for Go which we'd definitely benefit for, WDYT, would you update this PR by dropping the first commit and only focus on the other one?

Consider the context here being https://github.com/hermetoproject/hermeto/pull/1032 which actually touches the important bits (rather than formatting) on what does it mean to investigate a package manager and what a design should incorporate :) .

brunoapimentel:
> @brunoapimentel so there hasn't been much movement here since when this was proposed, where do we stand? We've already merged a few designs following no template (or better said, very loosely by copy-pasting from existing material), do you think we'd still benefit from a pretty generic and permissive template on a design doc? Maybe we could abandon that effort for now and only contribute the 2nd patch which actually retroactively adds a design doc for Go which we'd definitely benefit for, WDYT, would you update this PR by dropping the first commit and only focus on the other one?
> 
> Consider the context here being #1032 which actually touches the important bits (rather than formatting) on what does it mean to investigate a package manager and what a design should incorporate :) .

I was never fond of the approach proposed here, to be honest. If I recall correctly, I just wrote this because it was the direction our discussions were leading at the time. I still think our repo works best with a more loose design doc format compared to more formalized ADRs. I would gladly abandon this PR altogether.

eskultety:
> I was never fond of the approach proposed here, to be honest. If I recall correctly, I just wrote this because it was the direction our discussions were leading at the time. I still think our repo works best with a more loose design doc format compared to more formalized ADRs.

Well, I didn't mention #1032 by accident because one of the commits that introduces a template actually feels a bit stricter (at least to me) than what you have here, so I wanted to clarify where we stand on with this PR first. Feel free to comment on the linked PR as well because it looks like we'll end up with a template one way or the other.

>  I would gladly abandon this PR altogether.

Anyway, please don't, at the very least I'd expect us to retroactively add the Go doc, so if I could politely ask you to refine that part of the PR and propose this PR with a new title that'd be great!

brunoapimentel:
> Anyway, please don't, at the very least I'd expect us to retroactively add the Go doc, so if I could politely ask you to refine that part of the PR and propose this PR with a new title that'd be great!

Should I re-add the Go doc based on #1032 after that gets merged then?

eskultety:
> > Anyway, please don't, at the very least I'd expect us to retroactively add the Go doc, so if I could politely ask you to refine that part of the PR and propose this PR with a new title that'd be great!
> 
> Should I re-add the Go doc based on #1032 after that gets merged then?

Yes please.

PR #606, opened 2024-08-16
[Do Not Merge] Ruby test plan

a-ovchinnikov:
A place to discuss test plan for Rubygems/Bundler package manager.

### DIFF
--- a/docs/designs/ruby_test_plan_002.md
+++ b/docs/designs/ruby_test_plan_002.md
@@ -1,0 +1,139 @@
+# Test plan for RubyGems/Bundler package manager
+
+This document is intended to describe RubyGems/Bundler package manager
+usage patterns to be tested. The document follows package manager design
+proposed in [PR-565](https://github.com/containerbuildsystem/cachi2/pull/565).
+The document captures behaviors under test in a DSL for test scenario
+description. The format is straightforward and self-documenting, for a formal
+specification and tooling please refer to
+[the specification](https://cucumber.io/docs/gherkin/reference/).
+
+## General overview
+
+Cachi2 is intended to prefetch project's dependencies and to prepare a local
+environment to make it possible to build projects without network connectivity.
+Cachi2 can handle a number of package formats, RubyGems/Bundler (RG/B) is among
+the supported ones. Cachi2 does not support all features of RG/B, opting to
+support the most widespread and safe ones instead of supporting everything.
+
+The tests are to check that conforming packages could be preprocessed by Cachi2
+and that non-conforming packages are rejected by Cachi2.
+
+The checks are expected to be run against actual packages published either on
+rubygems.org, on github.com or locally.
+
+Each example in every test scenario requires a separate repository/package
+demonstrating an aspect under test. A list of necessary repositories could be
+found [below](#necessary-repositories).
+
+## Test cases
+
+@e2e_test
+Feature: Projects prepared with Cachi2 could be hermetically built
+    Scenario: a user can successfully build a project prepared with Cachi2
+        Given a clean repository of a well-defined project
+         When a user request to fetch dependencies for a well-defined project
+          And a user generates environment for well-defined project
+          And a user triggers a hermetic build
+         Then the build system exits successfully
+          And a build artifact is produced.
+
+
+@integration_test
+Feature: Cachi2 can prefetch dependencies for Ruby projects
+
+    Scenario outline: a user wants to fetch dependencies for a package containing well-defined Gemfile.lock
+       Given  clean repository of a "well-defined project"
+        When  a user request to fetch dependencies for a "well-defined project"
+        Then  direct and transitive dependencies are fetched
+         And  dependencies could be found within the "well-defined project"
+    Examples:
+        |    well-defined project                                       |
+        | Has a mixture of gems, Git and path dependencies and .gemspec |
+        | Has a mixture of gems, Git and path dependencies              |
+
+    Scenario outline: a user wants to fetch dependencies for a poorly-defined project
+       Given  clean repository of a "poorly-defined project"
+        When  a user request to fetch dependencies for a "poorly-defined project"
+        Then  direct and transitive dependencies are fetched
+         And  dependencies could be found within the "poorly-defined project"
+         But  cachi2 warns about problems with "poorly-defined project"
+    Examples:
+        |    poorly-defined project                                                     |
+        | Has a combination of dependency types in Gemfile.lock, but misses Gemfile     |
+
+    Scenario outline: a user wants to fetch dependencies for a package containing ill-defined Gemfile.lock
+       Given  clean repository of a "ill-defined project"
+        When  a user request to fetch dependencies for a "ill-defined project"
+        Then  Cachi2 reports an error
+         And  repository for "ill-defined project" remains unchanged
+    Examples:
+        |    ill-defined project                           |
+        | Gemfile.lock is missing                          |
+        | Git dependency is not pinned to a revision       |
+        | Transitive dependency is ill-defined             |
+
+
+@integration_test
+Feature: Cachi2 can preset configuration for hermetic builds
+
+    Scenario outline: a user prepares a repository with a Ruby project
+        When  a user generates environment for "a Ruby project"
+        Then  the environment contains a pointer to local Bundler config
+         And  Bundler cache directory structure is recreated for every project
+    Examples:
+        |    a Ruby project                                       |
+        | Single-project repository with mixed depenendcies types |
+        | Multi-project repository with mixed depenendcies types  |
+
+    Scenario: a user tries to prepare a project with malformed path dependency
+        When  a user generates environment for "malformed path dependency project"
+        Then  Cachi2 reports an error
+         And  repository for "malformed path dependency project" remains unchanged
+
+    Rule: cachi2-generated project configuration must coexist with user-generated configuration
+        Example: a user pre-configures a Ruby project and then runs cachi2
+            When a user modifies project configuraion
+             And generates an environment for "this project"
+            Then "cachi2-generated" configuration is stored within the project
+             And "user-generated" configuration is stored within the project
+
+    Rule: fetched dependencies directory structure must replicate Bundler cache structure
+        Example: a user prepares simple Ruby project with just gem dependencies
+            When a user generates an environment for "purely gem-dependent project"
+            Then dependencies are stored in vendor/cache
+             And .gem dependencies are kept as is
+
+        Example: a user prepares simple Ruby project with git dependencies
+            When a user generates an environment for "git and gem-dependent project"
+            Then dependencies are stored in vendor/cache
+             And Git dependnecies directories  names contain revisions
+
+        Example: a user prepares complex Ruby project
+            When a user generates an environment for "complex Ruby project"
+            Then dependencies are stored in vendor/cache
+             And dependencies are not duplicated
+
+*Note*, that actual implementations of scenarios described above might need to combine several
+steps or to share some initialization via a fixture or class-level set up for efficiency reasons.
+For example fetching from GitHub could be done once and then repository could be copied
+locally.
+
+### Necessary repositories
+
+Following the existing practice just one repository is needed. This repository
+must contain branches representing test scenarios. I.e. each Ruby project
+described below is a branch in the test repository.
+
+Each line below describes a properly defined Ruby project:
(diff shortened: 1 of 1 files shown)


docs/designs/ruby_test_plan_002.md
```diff
@@ -0,0 +1,150 @@
...
+       Given  clean repository of a "ill-defined project"
+        When  a user request to fetch dependencies for a "ill-defined project"
+        Then  Cachi2 reports an error
+         And  repository for "ill-defined project" remains unchanged
+    Examples:
+        |    ill-defined project                     |
+        | Gemfile.lock is missing                    |
+        | Gemfile.lock contains critical code        |
+        | Git dependency is not pinned to hash       |
+        | Gem dependency is sourced from unknown URL |
```
brunoapimentel, line 69: After some discussion, we decided to provide support for alternative registries: https://github.com/containerbuildsystem/cachi2/blob/a335c10d69885ba1e88e16f767cefce070a222f3/docs/design/rubygems.md?plain=1#L267

docs/designs/ruby_test_plan_002.md
```diff
@@ -0,0 +1,150 @@
...
+
+    Scenario outline: a user wants to fetch dependencies for a package containing poorly-defined Gemfile.lock
+        When  a user requests to fetch dependnecies for a "poorly-defined project"
+        Then  direct and transitive dependencies are fetched
+         And  could be found within the "poorly-defined project"
+         But  Cachi2 warns about detected problem
+    Examples:
+        |    poorly-defined project             |
+        | Gemfile contains executable code      |
+TODO: Should we detect, warn and block packages with any code in Gemfile?
```
brunoapimentel, line 81: We are not touching the `Gemfile`, so no problem there. Now, if the question is regarding the `Gemfile.lock`... can arbitrary code be added to it? How does it behave with Bundler and with the gemlock-parser?

My initial guess is that the `Gemfile.lock` is just meant to have structured data, and no code is ever executed there (and the existence of code would make the file invalid).

docs/designs/ruby_test_plan_002.md
```diff
@@ -0,0 +1,150 @@
...
+    
+    Scenario outline: a user wants to fetch dependencies for a package containing well-defined Gemfile.lock
+       Given  clean repository of a "well-defined project"
+        When  a user request to fetch dependencies for a "well-defined project"
+        Then  direct and transitive dependencies are fetched
+         And  dependencies could be found within the "well-defined project"
+    Examples:
+        |    well-defined project                          |
+        | Depends on gems from rubygems.org only           |
+        | Has correct .gemspec                             |
```
brunoapimentel, line 52: Not having a .gemspec is also valid, so we need to account for that.

docs/designs/ruby_test_plan_002.md
```diff
@@ -0,0 +1,150 @@
...
+        | Has correct Git dependency                       |
+        | Has several correct Git dependencies             |
+        | Has path dependency                              |
+        | Has several path dependencies                    |
+        | Has a mixture of gems, Git and path dependencies |
```
brunoapimentel, line 57: I think the "well-defined" scenarios can be simplified to:
- Has a mixture of gems, gems from alternate registries, Git and path dependencies. Has a gemspec file.
- Does not have a gemspec file

docs/designs/ruby_test_plan_002.md
```diff
@@ -0,0 +1,150 @@
...
+Feature: Cachi2 can preset configuration for hermetic builds
+
+    Scenario outline: a user prepares a repository with single Ruby project
+        When  a user generates environment for "single Ruby project"
+        Then  the environment contains a pointer to local Bundler config
+         And  Bundler cache directory structure is recreated for every project
+    Examples:
+        |    single Ruby project                            |
+        | Ruby project with just gem dependencies           |
+        | Ruby project with gem and git dependencies        |
```
brunoapimentel, line 95: Providing the build configuration is essentially setting some configuration keys in `.bundle/config` (or setting env vars), so I don't think the type of dependencies affect the ouput here.

brunoapimentel: We can probably have two scenarios for preset configuration: one for a single project, and one for a multi-project (or maybe even only having the multi-project scenario by itself is good enough, considering we also have the e2e that covers this)

docs/designs/ruby_test_plan_002.md
a-ovchinnikov: Makes sense, will update the doc in the next revision.

docs/designs/ruby_test_plan_002.md
a-ovchinnikov: Thank you, will remove the example.

docs/designs/ruby_test_plan_002.md
a-ovchinnikov: Ok, then this is not a valid example. Removing it. Thank you!

docs/designs/ruby_test_plan_002.md
a-ovchinnikov: The first approach sounds fine-grained, I'll use it.

a-ovchinnikov:
Trimmed down the test cases and necessary test repositories.

docs/designs/ruby_test_plan_002.md
```diff
@@ -0,0 +1,112 @@
...
+
+Each line below describes a properly defined Ruby project:
+ 1.  with a mixture of gem, git and path dependencies and a .gemspec file;
+ 2.  with a mixture of gem, git and path dependencies without a .gemspec file;
+
+These repositories must transitively depend on other gems, their depednencies must
+contain common dependencies.
+
+Each line below describes a malformed Ruby project:
+ 1.  containing a Gemfile.lock with git dependency not pinned to a hash;
```
ben-alkov, line 110: Do you mean "**commit** hash"[1]?

[1] "hash" and "checksum", especially standing alone, are woefully overloaded in our business :grin:

docs/designs/ruby_test_plan_002.md
```diff
@@ -0,0 +1,112 @@
...
+             And Git dependnecies directories  names contain revisions
+
+        Example: a user prepares complex Ruby project
+            When a user generates an environment for "complex Ruby project"
+            Then dependencies are stored in vendor/cache
+             And dependencies are not duplicated
+
+
+### Necessary repositories
+
```
slimreaper35, line 101: Ideally, we want to create one test repository with multiple branches. Each branch should have a descriptive name. That's our current approach to e2e / integration tests.

docs/designs/ruby_test_plan_002.md
slimreaper35: or revision ?

docs/designs/ruby_test_plan_002.md
```diff
@@ -0,0 +1,112 @@
...
+and that non-conforming packages are rejected by Cachi2.
+
+The checks are expected to be run against actual packages published either on
+rubygems.org, on github.com or locally.
+
+Each example in every test scenario requires a separate repository/package
+demonstrating an aspect under test. A list of necessary repositories could be
+found [below](#necessary-repositories).
+
+## Test cases
```
slimreaper35, line 28: Will we add a scenario with multiple packages in a single repo ?

docs/designs/ruby_test_plan_002.md
a-ovchinnikov: Good point, thank you, will update the document to reflect that this is a single repository with multiple branches.

docs/designs/ruby_test_plan_002.md
a-ovchinnikov: Do you mean e2e or integration test? For integration test in `Cachi2 can preset configuration for hermetic builds` feature `a user prepares a repository with a Ruby project` scenario is defined for both single project and multi-project repositories.

docs/designs/ruby_test_plan_002.md
```diff
@@ -0,0 +1,116 @@
...
+described below is a branch in the test repository.
+
+Each line below describes a properly defined Ruby project:
+ 1.  with a mixture of gem, git and path dependencies and a .gemspec file;
+ 2.  with a mixture of gem, git and path dependencies without a .gemspec file;
+
+These repositories must transitively depend on other gems, their depednencies must
+contain common dependencies.
+
+Each line below describes a malformed Ruby project:
```
eskultety, line 113: Missing Gemfile is a problem also, if for nothing else then bundler parser complains about it whenever git dependencies are used as it's trying to figure out the cache path where the git repo should be cloned to, but I think it would complain even harder during install.

docs/designs/ruby_test_plan_002.md
```diff
@@ -0,0 +1,116 @@
...
+@integration_test
+Feature: Cachi2 can preset configuration for hermetic builds
+
+    Scenario outline: a user prepares a repository with a Ruby project
+        When  a user generates environment for "a Ruby project"
+        Then  the environment contains a pointer to local Bundler config
+         And  Bundler cache directory structure is recreated for every project
+    Examples:
+        |    a Ruby project                                       |
+        | Single-project repository with mixed depenendcies types |
```
eskultety, line 75: I wonder if we could combine the single-project repo case + bundler configuration with the previous positive test case since based on the configuration opts order of precedence we have to dump a config file either way, the only question is whether it'll be combined with user options or just ours, so I'd create the previous positive case such that user defines some options in a custom config and cover  only the more complex scenario where a user config option might potentially collide with ours. Note that in order to test this we'll likely have to use the `inject-files` command which also means we need to run a prefetch first which feels redundant to me when the previous scenario has to already do that, so I'd prefer if we ran as few test cases as possible and combine whatever is related together (in general).

docs/designs/ruby_test_plan_002.md
a-ovchinnikov: This is a good call-out, thank you, and looks like a candidate for poorly-formed repository scenario. I'll add that scenario back.

docs/designs/ruby_test_plan_002.md
a-ovchinnikov: I don't think combining multiple test scenario together is a good idea. Each defined scenario should test one distinct aspect. The need to clone a repo or the need to run as few tests as possible is an implementation detail and combining multiple tests together is essentially leaking implementation details into higher-level. From what I have seen on the gate the tests tend to run for just few seconds, or tens of seconds when podman has to build something. I'd say optimizations should be handled by tests implementation: all expensive operations could be lumped together into a fixture or any other form of shared context. On the implementation level things could and likely should be combined for efficiency.

To make sure that this is not forgotten by an implementer I'll add a couple of sentences about a fixture or joining scenarios to the end of section.

ben-alkov: Agreed. I spent days trying to understand/fix some pip unit tests because they were *testing everything*.

Eventually I was forced to simply refactor them out based on type under test, which actually tested the same stuff in fewer SLOC because conditionals based on behavior/data differences in type were no longer needed.

eskultety: > Agreed. I spent days trying to understand/fix some pip unit tests because they were testing everything.

Wasn't this due to the matrix of parametrize? That's different though, because you have a matrix that explodes in complexity.

> I don't think combining multiple test scenario together is a good idea. Each defined scenario should test one distinct aspect.

I don't dispute this, but what we're really talking about here is to add essentially one more assert after running `inject-files` which is IMO still a constant complexity since we're not introducing any matrix for the prefetch test to handle. I wouldn't suggest it if we didn't have the data for a successful `inject-files` check already at hand.

It's not a blocker for me though if we keep it a separate scenario, I'm just always wary of the way we structure integration tests simply because of how much time they take to execute, e.g. the e2e's are painfully slow and based on the fact that this is inherently only going to get worse we can only control or tame the pace of it.

a-ovchinnikov: I think we are talking about pretty much the same thing: high-level test cases should remain cleanly separated, but at the implementation level could and should be optimized. I added a note at L117 in the latest revision to stress that.

ben-alkov: > Wasn't this due to the matrix of parametrize

That was one aspect, yes. The sheer amount of stuff being tested was the main issue.

a-ovchinnikov:
Feedback collected. Closing.

PR #612, opened 2024-08-27
testing: Making integration tests run conditional

a-ovchinnikov:
Before this change integration tests and e2e tests would run unconditionally on every change. This commit introduces some shortcuts to this process: changes that do not affect code (e.g. changes to text files or to anything outside of main source tree and integration tests tree) will not trigger integration and e2e tests, while changes to individual package managers or package managers tests will trigger just the affected tests. This will reduce the average time to run the tests.


tests/integration/utils.py
```diff
@@ -463,3 +463,133 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
         dep_file = output_dir / "deps" / path
         assert dep_file.exists()
         assert dep_file.read_text() == expected_content
+
+
+TEST_EVERYTHING = object()
+TEST_NOTHING = object()
+# Individual files could be added to the set as well.
+paths_to_code = set((Path("cachi2"), Path("tests/integration")))
+supported_pms = set(("rpm", "yarn", "gomod", "npm", "pip"))
```
eskultety, line 472: Can we pull this list from the resolver module from `_package_managers` and `_dev_package_managers` instead of having to update the sequence manually at two places?

tests/integration/utils.py
```diff
@@ -463,3 +463,133 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+    """Check which tests to run basing on which files were changed in a commit."""
+    if is_a_release_candidate(ref):
+        return TEST_EVERYTHING
+    changes = retrieve_changed_files_from_git(ref)
+    if no_code_was_affected_by(changes):
+        return TEST_NOTHING
+    elif just_package_managers_were_affected_by(changes):
+        return affected_package_managers(changes)
+    else:
+        return TEST_EVERYTHING
```
eskultety, line 595: How does this fare with regular releases? What ensures all tests are run when a release is tagged? If I'm not missing anything then we have to list tags for the current ref and if a tag is set then we need run everything; which goes back to my earlier comment on detecting RCs...I don't think we need to do that specifically, we need to run everything on any tag.

eskultety (commented):
This looks very promising once we iron out the details.

tests/integration/utils.py
eskultety: Oh, will if just fall through to the last `else` condition in that case? But then, when we tag a commit, that ref will only trigger a subset of tests based on files changed anyway, so I really do think that on tagging a release in git (which is currently admittedly not being done by the book IMO, but we know that) we should make sure we run **everything**.

tests/integration/conftest.py
```diff
@@ -69,3 +69,15 @@ def local_pypiserver() -> Iterator[None]:
...
+def pytest_runtest_setup(item: Any) -> None:
+    marks = set(mark.name for mark in item.iter_markers())
+    to_test = utils.determine_integration_tests_to_run_for("HEAD")
+    if to_test is utils.TEST_EVERYTHING:
+        return
+    elif to_test is utils.TEST_NOTHING:
+        pytest.skip()
+    else:
+        if not marks.intersection(to_test):
+            pytest.skip()
```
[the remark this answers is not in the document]
ben-alkov: Definitely agree about marking test cases in this way. I feel like there's a way to programmatically mark them before pytest collection runs (but don't ask me for details :grin: )...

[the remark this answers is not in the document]
a-ovchinnikov: Yes, `pytest_collection_modifyitems` is worth a try. My original implementation was caching the results of `determine_integration_tests_to_run_for("HEAD")`, but then I removed it for testing and did not add back. I am not sure that marks make things much more error-prone: there are not so many tests there ATM, every new test will be run locally first with the same command, so if a mark gets forgotten then the test would not be executed. I still would prefer not to have marks, so I'll try your suggestion.

tests/integration/utils.py
a-ovchinnikov: Oh, absolutely! Thank you!

tests/integration/utils.py
```diff
@@ -463,3 +463,133 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+TEST_NOTHING = object()
+# Individual files could be added to the set as well.
+paths_to_code = set((Path("cachi2"), Path("tests/integration")))
+supported_pms = set(("rpm", "yarn", "gomod", "npm", "pip"))
+
+
+def retrieve_changed_files_from_git(ref: str = "HEAD") -> tuple[Path, ...]:
+    cmd = f"git show --name-only --pretty=format: {ref}".split()
+    res = Popen(cmd, stdout=PIPE, stderr=PIPE, text=True)
+    files, _ = res.communicate()
```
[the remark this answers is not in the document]
a-ovchinnikov: Good point, thank you!

tests/integration/utils.py
```diff
@@ -463,3 +463,133 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+        head, *tail = [el.strip() for el in note.split(":")]
+        return head == "RC" and len(tail) > 0 and tail[0].lower() == "yes"
+
+    cmd = f"git notes show {ref}".split()
+    res = Popen(cmd, stdout=PIPE, stderr=PIPE, text=True)
+    notes, _ = res.communicate()
+
+    if res.returncode != 0:
+        return False
+    return any(contains_rc_marker(note) for note in notes.split("\n"))
```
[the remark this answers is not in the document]
a-ovchinnikov: TBH I am not sure whether this is needed at all. My idea was to have some mechanism to unconditionally trigger all tests, I chose to call it "release candidate marker" for this PoC and added a draft implementation. The thing is that as long as tests are mutually independent and there is no coupling in the code success of a modified subset means that all other tests would pass too. Thus as long as we do not merge PRs with failing ITs or e2es we should be good. So this is a back-up option of a very questionable necessity. I favor its complete removal, but would definitely want to hear more opinions.

tests/integration/utils.py
a-ovchinnikov: Basing on the current draft a release candidate note will trigger all tests (which we _maybe_ don't really need after all, see my comment above). L587 handles that. Once we are convinced that everything is OK we tag the RC and thus have it released.

tests/integration/utils.py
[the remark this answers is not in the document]
ben-alkov: I do think that it would be nice to have a simple way to 'run all', but I don't think that this particular mechanism is a good choice.

tests/integration/utils.py
```diff
@@ -463,3 +465,109 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+# Individual files could be added to the set as well.
+paths_to_code = set((Path("cachi2"), Path("tests/integration")))
+supported_pms = set(list(resolver._package_managers) + list(resolver._dev_package_managers))
+
+
+def retrieve_changed_files_from_git(ref: str = "HEAD") -> tuple[Path, ...]:
+    repo = Repo(".", search_parent_directories=True)
+    commit = repo.commit(ref)
+    modified_files = [Path(item.a_path) for item in commit.diff(commit.parents)]  # type: ignore
+    return tuple(modified_files)
```
slimreaper35, line 481: Why tuple ?

tests/integration/conftest.py
```diff
@@ -69,3 +69,23 @@ def local_pypiserver() -> Iterator[None]:
             raise RuntimeError("pypiserver didn't start fast enough")
 
         yield
+
+
+def does_not_test_any_affected_pms(test: Any, affected_pms_to_test: set[str]) -> bool:
+    pm_under_test = Path(test.fspath).stem.lstrip("test_")
```
slimreaper35, line 75: Here, you could reuse `extract_pm_name` from _utils.py_

slimreaper35:
> Full test for a commit could be forced by adding an RC note to it:
git add note -m "RC: yes"

Okay, I just found out, that git commits can have notes :smile: 
For me only works `git notes add -m ".."` Am I missing something?

tests/integration/conftest.py
```diff
@@ -69,3 +69,23 @@ def local_pypiserver() -> Iterator[None]:
...
+    return pm_under_test not in affected_pms_to_test
+
+
+def pytest_collection_modifyitems(session: Any, config: Any, items: Any) -> None:
+    skip_mark = pytest.mark.skip(reason="No changes to tested code")
+    to_test = utils.determine_integration_tests_to_run_for("HEAD")
```
slimreaper35, line 81: Suggestion:
https://docs.pytest.org/en/7.1.x/_modules/_pytest/hookspec.html#pytest_collection_modifyitems
I found type annotations and docstring for this function in pytest documentation. We can borrow that or get some inspiration.

tests/integration/conftest.py
```diff
@@ -69,3 +69,23 @@ def local_pypiserver() -> Iterator[None]:
...
+    if to_test is utils.TEST_EVERYTHING:
+        return
+    elif to_test is utils.TEST_NOTHING:
```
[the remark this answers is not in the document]
eskultety: This is a compound if-elif-else block, it visually makes more sense the way it's written despite a return being used.

tests/integration/conftest.py
eskultety: > We can borrow that or get some inspiration.

We should use it **verbatim** as is in the docs.

tests/integration/conftest.py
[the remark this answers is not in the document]
slimreaper35: I would disagree it is visually any better. By the way it is reported by [ruff](https://docs.astral.sh/ruff/rules/superfluous-else-return/) and [pylint](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html) as well.

I imagine something like that to be the best, but I don't know if it would be possible (just an idea).
```
def pytest_collection_modifyitems(session: Any, config: Any, items: Any) -> None:
    skip_mark = pytest.mark.skip(reason="No changes to tested code")
    to_be_skipped = utils.determine_integration_tests_to_be_skip("HEAD")

    for test in to_be_skipped:
        test.add_marker(skip_mark)
```

tests/integration/conftest.py
```diff
@@ -69,3 +69,23 @@ def local_pypiserver() -> Iterator[None]:
...
+        for item in items:
+            item.add_marker(skip_mark)
+    else:
+        for item in items:
+            if does_not_test_any_affected_pms(item, to_test):
+                item.add_marker(skip_mark)
+    return
```
eskultety, line 91: Apparently, you can use marks (as you clearly demonstrate), but that means that after this function is run, pytest still evaluates **every single test** and skips it on demand based on the mark. I'm not saying that's wrong by any means, I'll just share what I envisioned and let others to comment on that.

```
...
if to_test is utils.TEST_NOTHING:
  items[:] = []  # items are passed by reference, so need to be modified in-place
else:
  deselected = {i for i in items if does_not_test_any_affected_pms(i, to_test)}
  items[:] = set(items) - deselected
  
 # inform pytest of the deselected tests
 config.hook.pytest_deselected(items=deselected)


def pytest_deselected(items):
# override the built-in method to print deselected tests (mostly for verification purposes that this machinery works)
print("Deselected tests:")
for i in items:
  print(f"  - {i.nodeid}")
```

Here's the output comparison:

**Yours**:
```
collected 51 items    <<< NOTICE THIS LINE                                                             

tests/integration/test_gomod.py ssssssssssssssss
tests/integration/test_npm.py ssssssss
...
```

**Mine**:
```
collecting ... 
Deselected tests:
  - tests/integration/test_gomod.py::test_e2e_gomod[gomod_e2e_multiple_modules]
  - tests/integration/test_yarn.py::test_yarn_packages[yarn_incorrect_checksum]
  - tests/integration/test_rpm.py::test_repo_files[rpm_test_repo_file]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_without_deps]
  - tests/integration/test_yarn.py::test_yarn_packages[yarn_disallowed_protocols]
  - tests/integration/test_gomod.py::test_e2e_gomod[gomod_1.21_e2e_test]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_vendored_with_flag]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_vendor_check_wrong_vendor]
  - tests/integration/test_rpm.py::test_rpm_packages[rpm_multiple_archs]
  - tests/integration/test_yarn.py::test_yarn_packages[yarn_immutable_installs]
  - tests/integration/test_npm.py::test_npm_smoketest[npm_lockfile3_yarn_registry]
  - tests/integration/test_yarn.py::test_yarn_packages[yarn_no_lockfile]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_vendor_check_empty_vendor]
  - tests/integration/test_npm.py::test_e2e_npm[npm_smoketest_lockfile3]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_workspaces]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_go_generate_imported]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_vendored_without_flag]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_vendor_check_no_vendor]
  - tests/integration/test_npm.py::test_e2e_npm[npm_smoketest_lockfile2]
  - tests/integration/test_rpm.py::test_rpm_packages[rpm_unmatched_checksum]
  - tests/integration/test_gomod.py::test_e2e_gomod[gomod_1.18_e2e_test]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_multiple_modules_missing_checksums]
  - tests/integration/test_npm.py::test_e2e_npm[npm_lockfile3_dev_optional_peer_deps]
  - tests/integration/test_npm.py::test_e2e_npm[npm_lockfile3_multiple_dep_versions]
  - tests/integration/test_rpm.py::test_rpm_packages[rpm_multiple_packages]
  - tests/integration/test_yarn.py::test_e2e_yarn[yarn_e2e_test_multiple_packages]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_vendor_check_correct_vendor]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_with_deps]
  - tests/integration/test_rpm.py::test_rpm_packages[rpm_missing_checksums]
  - tests/integration/test_yarn.py::test_yarn_packages[yarn_zero_installs]
  - tests/integration/test_npm.py::test_npm_smoketest[npm_bundled_lockfile3]
  - tests/integration/test_gomod.py::test_gomod_packages[gomod_local_deps]
  - tests/integration/test_yarn.py::test_yarn_packages[yarn_correct_version_installed_by_corepack]
  - tests/integration/test_yarn.py::test_e2e_yarn[yarn_e2e_test]
  - tests/integration/test_rpm.py::test_rpm_packages[rpm_unexpected_size]
  - tests/integration/test_npm.py::test_e2e_npm[npm_multiple_packages_lockfile3]
  - tests/integration/test_gomod.py::test_e2e_gomod[gomod_1.20_e2e_dirty_go.mod]
  - tests/integration/test_npm.py::test_e2e_npm[npm_lockfile3_aliased_deps]
  - tests/integration/test_rpm.py::test_e2e_rpm[rpm_e2e_test]
collected 51 items / 39 deselected / 12 selected    <<< NOTICE THE SAME LINE                                            

tests/integration/test_pip.py
...
```
NB: if you decide to adopt the latter, i.e. not using marks then I think it's worth flipping the boolean logic of `does_not_test_any_affected_pms`.

tests/integration/utils.py
```diff
@@ -463,3 +465,109 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+TEST_NOTHING = object()
+# Individual files could be added to the set as well.
+paths_to_code = set((Path("cachi2"), Path("tests/integration")))
+supported_pms = set(list(resolver._package_managers) + list(resolver._dev_package_managers))
+
+
+def retrieve_changed_files_from_git(ref: str = "HEAD") -> tuple[Path, ...]:
+    repo = Repo(".", search_parent_directories=True)
+    commit = repo.commit(ref)
+    modified_files = [Path(item.a_path) for item in commit.diff(commit.parents)]  # type: ignore
```
eskultety, line 480: What did mypy complain about that you have to disable it?
For future reference I think it's worth adding a commentary to all places where mypy needs to be disabled and why.

tests/integration/utils.py
```diff
@@ -463,3 +465,109 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+    >>> just_package_managers_were_affected_by(c)
+    False
+    >>> just_package_managers_were_affected_by((Path('cachi2/core/utils.py'),))
+    False
+    """
+    return all(affects_pm(c) for c in changes)
+
+
+def must_test_all() -> bool:
+    return os.getenv("CACHI2_RERUN_ALL_ITS", "no").lower() == "yes"
```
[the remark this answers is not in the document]
slimreaper35: > How about CACHI2_RUN_ALL_INTEGRATION_TESTS?
+1

tests/integration/utils.py
```diff
@@ -463,3 +465,109 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+
+
+# 'Any' was used to let sentinel objects be returned as well as a set of names without
+# introducing full-blown singleton classes for TEST_EVERYTHING and TEST_NOTHING.
+def determine_integration_tests_to_run_for(ref: str = "HEAD") -> Any:
+    """Check which tests to run basing on which files were changed in a commit."""
+    if must_test_all():
```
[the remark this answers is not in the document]
a-ovchinnikov: `must_test_all` is the condition we are checking, `os.getenv("RUN_ALL_TESTS", "false").lower() == "true"` happens to be the way to check this condition now (it was git note check a commit ago). I prefer to distinguish between what is done and how it is done when possible. A function adds some overhead, that's true, and makes the code a little longer, but at the same time spares a reader the need to interpret a piece of code in their mind. Moreover, as long as the reader believes that `must_test_all` is implemented correctly they can ignore its implementation completely. I believe that sparing a reader a second is on average more productive than sparing a few nanoseconds for dereferencing another pointer and that we are operating on high-enough level to be able to afford that. I am open to inlining it though if there is popular demand for that.

tests/integration/utils.py
```diff
@@ -463,3 +465,109 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+        return TEST_EVERYTHING
+    changes = retrieve_changed_files_from_git(ref)
+    if no_code_was_affected_by(changes):
+        return TEST_NOTHING
+    elif just_package_managers_were_affected_by(changes):
+        return affected_package_managers(changes)
+    else:
```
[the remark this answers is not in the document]
a-ovchinnikov: I structured it this way to highlight to a reader that these outcomes are closely related. The resulting code is  leaning a bit to the right on (APL:Enterprise Java) verbosity scale, but I am inclined to keep it as is unless there is popular demand to boil it down.

tests/integration/utils.py
```diff
@@ -463,3 +465,109 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+        return "package_managers" in change.parts
+
+    def affects_pm_tests(change: Path) -> bool:
+        return change.stem.lstrip("test_") in supported_pms
+
+    return affects_pm_or_support_code(change) or affects_pm_tests(change)
+
+
+def affected_package_managers(changes: tuple[Path, ...]) -> set[str]:
+    return set(extract_pm_name(c) for c in changes if affects_pm(c))
```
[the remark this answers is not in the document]
a-ovchinnikov: Yes, it is related to `extract_pm_name` issue above, will fix in the next PR.

tests/integration/utils.py
```diff
@@ -463,3 +465,166 @@
...
+
+def retrieve_changed_files_from_git(ref: str = "HEAD") -> tuple[Path, ...]:
+    try:
+        repo = Repo(".", search_parent_directories=True)
+    except ValueError:
+        # This is supposed to happen on gate only. There
+        # the following check will succeed. It will raise another
+        # ValueError locally
+        Path(".").absolute().relative_to("/__w")
+        repo.config_writer("global").set_value("safe", "directory", "*").release()
```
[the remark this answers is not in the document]
a-ovchinnikov: L477?

tests/integration/utils.py
```diff
@@ -463,3 +465,168 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+def retrieve_changed_files_from_git(ref: str = "HEAD") -> tuple[Path, ...]:
+    repo = Repo(".", search_parent_directories=True)
+    try:
+        commit = repo.commit(ref)
+    except ValueError:
+        # This is supposed to happen on gate only. There
+        # the following check will succeed. It will raise another
+        # ValueError locally
+        Path(".").absolute().relative_to("/__w")
+        repo.config_writer("global").set_value("safe", "directory", "*").release()
```
a-ovchinnikov, line 485: This is an abomination, I am very inclined to do something about it. I am out of good ideas ATM though.

tests/integration/utils.py
eskultety: > When invoked this works fine now because deep inside the library [args sequence gets flattened](https://github.com/gitpython-developers/GitPython/blob/cfadd9e5edb8571c2c15878c8eeca91f15a69040/git/cmd.py#L1477) and also because all normal commits will have exactly one parent. OTOH multiparent commits are merge commits, and IIRC they will not have any changes in them anyway (unless they are evil), so they should return nothing and it appears to be safe to accept multiple parents. It looks like the code does the right thing while annotations miss this aspect.

Yeah, another way of fixing this would be `commit.parents[0]` because in case of a merge commit parent, that one should always point to the other branch which we merged to the target branch and so would become nr.2 parent. I completely agree though that stats seems like the superior approach.

tests/integration/conftest.py
eskultety: > It looks like some additional moves are required to actually not run them, like adding a skip mark or clearing the list of tests.

You can find all that in my example, you need to modify the items list in-place as that one is passed by reference, that's the reason why you actually ran tests which you marked as deselected. 

>  I like the skip mark approach because it leaves a clear trail in logs -- all skipped tests are skipped with a reason stated right next to test's name, and the reason is "Does not test any modified code"

Well, you can easily tweak that in my example by prepending the same msg string to the "Deselected tests" output et voila achieve the same!

> Currently I am not convinced that traversing a list of ~100 objects is expensive enough to warrant more confusing tox -eintegration output, so I will keep my approach with marks. I am open to changing that though.

Fair enough, I'm not too picky about this and marks can accomplish the same outcome, so let's keep that approach instead of fiddling with test deselection.

tests/integration/utils.py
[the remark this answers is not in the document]
eskultety: Fair enough, a matter of taste then, not a deal breaker, let's resolve this.

tests/integration/utils.py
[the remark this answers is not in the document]
eskultety: I don't have an issue with this at all and to be fair I am not a fan of blindly following all default linter rules (I have a bunch of flake8 tests disabled globally for my repos), so I'd argue that if a linter complaints, well, those complain about almost anything, so we should be sane about it and disable what we don't like. Although we wouldn't have this discussion if a formatter had changed it :) (which can be tweaked too when it comes to rules), just sayin'.

tests/integration/conftest.py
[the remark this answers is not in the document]
eskultety: I do agree from reader's POV the collector function reads easier knowing that all the selection logic on which tests to run is abstracted in a helper. So to me it's just moving code around, the logic needs to live somewhere, either in the collector (partially) or the helper (completely).

tests/integration/utils.py
eskultety: This problem is tied to the CI environments and so the code to work around it needs to live in the GH action. Can you try any of the following to see if it helps (it should): https://github.com/actions/checkout/issues/766

a-ovchinnikov: Thank you, GH actions were exactly the piece I was missing!

tests/integration/conftest.py
ben-alkov: This use case is exactly what 'skip' marks are for.

eskultety: > This use case is exactly what 'skip' marks are for.

Uhm, I'm not sure, I'd argue that adding the skip marker dynamically during test collection is a clever way to solve a problem rather than a clean and expected solution, but since we're staying within the public API boundaries....Skips are IMO meant to be used directly in the decorator in cases where tests are to be conditionally skipped but you know the condition beforehand (which is not the case here obviously). 
pytest documents the collection process here: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.hookspec.pytest_collection and they mention deselection, but adding skip marks does look a little less hacky and has a clear output, so why not.

Interestingly enough usage of `pytest.skip` directly is discouraged although I admit I don't understand why it's okay during test collection when you use that particular flag: << This is out of scope though since we moved away from that approach already

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-skip
> This function should be called only during testing (setup, call or teardown) or during collection by using the allow_module_level flag.

tests/integration/conftest.py
slimreaper35: Could you add a short docstring as well please ?

tests/integration/conftest.py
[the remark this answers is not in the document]
slimreaper35: > I tried returning a single iterable with all things to switch off, but did not like how it looked.

What was the problem? I still feel like the logic around this could be simplified.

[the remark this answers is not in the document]
a-ovchinnikov: The problem is with determining which tests not to skip. In my approach I effectively return markers for types of tests not to skip, my helpers know nothing about tests themselves, just about different types of code and which marker corresponds to which changed part. This comes at a price of teaching pytest about these markers. The alternative approach would be to implement test names accumulation in helpers, which would make the helpers more complex, thus more error-prone and would also result in re-implementing parts of pytest (i.e. tests collection). This is a more complicated approach, I tried sketching it first, did not like the complexity and opted for the solution which I posted.

tests/integration/conftest.py
a-ovchinnikov: Sure, done.

tox.ini
```diff
@@ -60,6 +60,7 @@ commands =
 passenv =
     CACHI2_GENERATE_TEST_DATA
     CACHI2_IMAGE
+    CACHI2_RUN_ALL_INTEGRATION_TESTS
```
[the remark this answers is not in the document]
a-ovchinnikov: What would be the benefit of it? Does `make` clobber environment?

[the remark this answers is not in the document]
slimreaper35: I think there is no benefit to having these variables there.
You can simply run `CACHI2_SOME_ENV=true make integration-test`.

tests/integration/utils.py
```diff
@@ -463,3 +465,167 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+# Individual files could be added to the set as well.
+paths_to_code = set((Path("cachi2"), Path("tests/integration")))
+supported_pms = set(list(resolver._package_managers) + list(resolver._dev_package_managers))
+
+
+def retrieve_changed_files_from_git(ref: str = "HEAD") -> tuple[Path, ...]:
+    repo = Repo(".", search_parent_directories=True)
+    commit = repo.commit(ref)
+    modified_files = [Path(f) for f in commit.stats.files]
+    return tuple(modified_files)
```
slimreaper35, line 480: I noticed that the function returns changes only made in the last commit.
I suppose it should return changes from `<current_branch>..main` or ?

tests/integration/conftest.py
```diff
@@ -69,3 +69,24 @@ def local_pypiserver() -> Iterator[None]:
             raise RuntimeError("pypiserver didn't start fast enough")
 
         yield
+
+
+def does_not_test_any_affected_pms(test: pytest.Item, affected_pms_to_test: set[str]) -> bool:
+    pm_under_test = utils.tested_object_name(Path(test.fspath))
```
slimreaper35, line 75: `fspath` has a type of `LEGACY_PATH`, which will be deprecated at some point. You can use just `path` without type conversion
https://github.com/pytest-dev/pytest/blob/72c682ff9773ad2690711105a100423ebf7c7c15/src/_pytest/nodes.py#L139

tests/integration/utils.py
a-ovchinnikov: My intent was to run this for the latest commit only. Diffing feature branch with main would produce too many changed modules and would effectively negate all the benefits of the change. Suppose one is working on pip in one commit, then on gomod, then on npm. My change addresses the fact that it is not necessary to retest pip when gomod is affected.

tests/integration/conftest.py
a-ovchinnikov: Good point, thank you, will update in the next revision.

tests/integration/conftest.py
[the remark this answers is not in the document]
a-ovchinnikov: Thanks for the snippet @slimreaper35 , I did indeed overthink this. I will apply the change to the next revision.

tests/integration/utils.py
```diff
@@ -463,3 +465,166 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
         dep_file = output_dir / "deps" / path
         assert dep_file.exists()
         assert dep_file.read_text() == expected_content
+
+
+ALL = object()
+# Individual files could be added to the set as well.
+paths_to_code = set((Path("cachi2"), Path("tests/integration")))
+supported_pms: set[str] = set(
```
eskultety, line 473: These are global scoped RO variables -> we should declare them as constants.

eskultety: nitpick: Also, these should go to the top of the file to not break the visual reading continuation - one expects constants at the beginning of the module, then classes, then functions, etc.

tests/integration/utils.py
```diff
@@ -463,3 +465,166 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+    """Check which tests to run basing on which files were changed in a commit."""
+    if must_test_all():
+        return set()
+    changes = select_testable_changes(retrieve_changed_files_from_git(ref))
+    if len(changes) == 0:
+        return supported_pms
+    elif just_some_package_managers_were_affected_by(changes):
+        return supported_pms - affected_package_managers(changes)
+    else:
+        return set()
```
eskultety, line 630: How about just `return set()` without the empty `else` clause at the end?

tests/integration/utils.py
```diff
@@ -463,3 +465,166 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+    False
+    """
+    return all(affects_pm(c) for c in changes)
+
+
+def must_test_all() -> bool:
+    return os.getenv("CACHI2_RUN_ALL_INTEGRATION_TESTS", "false").lower() == "true"
+
+
+def determine_integration_tests_to_run_for(ref: str = "HEAD") -> Any:
```
eskultety, line 620: This is an incredibly confusing name, because the function does the exact opposite - it doesn't determine which tests to run, it returns the ones that need to be skipped.

Additionally, this should be typed as returning a `Set`, not `Any`.

tests/integration/utils.py
```diff
@@ -463,3 +465,166 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+    """Weed out changes that cannot be tested.
+
+    If a change is outside of paths to testable code it is dropped as if
+    it never happened. Any file or module outside of watched directories
+    will be rejected.
+    """
+    return tuple(c for c in changes if is_testable_code(c))
+
+
+def just_some_package_managers_were_affected_by(changes: tuple[Path, ...]) -> bool:
```
eskultety, line 577: In context of my other comments (testing against main + affecting all pkg mgrs with changes to certain modules), we definitely DO need to test all pkg mgrs if a change was introduced to `package_managers/general.py` or essentially any non-pkg_mgr named module under the package, so this function needs to be reworked and renamed.

tox.ini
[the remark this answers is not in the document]
eskultety: For now, we should remain consistent with our current behaviour/approach towards env vars. Changes to the Makefile like you suggest should be done in a different PR.

eskultety (commented):
So, tracking all the changes done in this PR was a bit difficult because each commit introduced some fixup for one of its ancestors, please only propose new changes in subsequent commits (unless absolutely necessary, i.e. it would be actually worse to review the PR normally that way) and squash in the related bits to their respective parent commits.

Overall, I only have a single major comment - We definitely need to compare all changes against the `main` branch OR against whatever the origin base branch was (i.e. we'll need an env variable so that when one creates 3 connected branches, we only compare against the last base) and not just a single commit, because that premise only works nicely in an interactive rebase scenario which, needless to say, is a very valid and also common scenario.

tests/integration/utils.py
a-ovchinnikov: Good point, this  is a frozenset indeed.

.github/workflows/gating.yaml
```diff
@@ -116,4 +116,5 @@ jobs:
           CACHI2_IMAGE: localhost/cachi2:${{ github.sha }}
           CACHI2_TEST_LOCAL_PYPISERVER: 'true'
         run: |
+          git config --global --add safe.directory "*"
```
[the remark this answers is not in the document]
a-ovchinnikov: Done.

tests/integration/utils.py
a-ovchinnikov: I am not a big fan of dangling `elif`s, but yes, absolutely.

tests/integration/conftest.py
```diff
@@ -69,3 +69,15 @@ def local_pypiserver() -> Iterator[None]:
             raise RuntimeError("pypiserver didn't start fast enough")
 
         yield
+
+
+def pytest_runtest_setup(item: Any) -> None:
+    marks = set(mark.name for mark in item.iter_markers())
+    to_test = utils.determine_integration_tests_to_run_for("HEAD")
```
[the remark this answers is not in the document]
a-ovchinnikov: It reads much better this way (since the function is named `determine_integration_tests_to_run_for`).

tests/integration/test_pip.py
```diff
@@ -10,7 +10,6 @@
 log = logging.getLogger(__name__)
 
 
-@pytest.mark.pip
```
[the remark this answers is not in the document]
a-ovchinnikov: I was planning to eventually squash all three commits together. I'll do it now.

tests/integration/conftest.py
```diff
@@ -71,22 +71,16 @@ def local_pypiserver() -> Iterator[None]:
...
-    elif tests_to_run is None:
-        for item in items:
+    tests_to_skip = utils.determine_integration_tests_to_run_for("HEAD")
+    for item in items:
+        if utils.tested_object_name(item.path) in tests_to_skip:
             item.add_marker(skip_mark)
-    else:
-        for item in items:
-            if does_not_test_any_affected_pms(item, tests_to_run):
-                item.add_marker(skip_mark)
```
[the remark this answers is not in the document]
a-ovchinnikov: Yes, that was the plan. I kept the commits separated for a bit too long.

tests/integration/utils.py
eskultety: nitpick: Can we also use capitals here? This way it still reads more like a global variable or even worse (in context of pytest), as a fixture defined somewhere.

tests/integration/conftest.py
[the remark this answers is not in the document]
eskultety: I'll ask differently, since you're not proposing ref to be something we can override through an env variable for tests, do we need to call the function `determine_integration_tests_to_run_for` instead of `determine_integration_tests_to_run` ?

tests/integration/utils.py
eskultety: > This feels a bit like overengineering to me. It is extremely easy to trigger all tests to run

But that's exactly not what we want here. To be precise, the scenario where someone would touch all pkg mgrs at the same time is extremely rare. And I specifically don't want to run tests for pkg mgrs I'm not touching and I know that the related tests take too long to execute (gomod, pip), so if I'm touching both RPM and yarn in a single PR, then I don't want the tests to run just for yarn or RPM depending on what the last commit is. And I most certainly don't want to run tests for pip and gomod when I know they take twice as long.

> If this is the desired behavior then we already have an override variable for that. 

What variable do you mean?

I think the whole point of this PR was to kind of automate the detection of which tests need to be run for a particular change that would work both for one-shot executions as well in interactive rebase scenarios. At least that was my understanding and the value I was seeing in that, i.e. not needing to use any variable (unless absolutely needed for some heavy tweaking) and it'd just do the "right thing".

.github/workflows/gating.yaml
eskultety, line 119: **Commit message:**

nitpick: Can we be more explicit in the subject on where exactly the problem is? I know you say 'gate', but it may not be obvious to everyone, so either change the context or just mention GitHub CI somewhere in both the subject AND/OR the commit message itself.

tests/integration/utils.py
slimreaper35: What if the last commit in the PR updates only the README.md, but the previous commits are related to one or any number of package managers? We don't want to skip all integration tests in that case.

a-ovchinnikov: The fact that GH tests all commits in a sequence (see a screenshot of the current sequence below)
image
is proven by the latest (at the moment of writing) [IT job].(https://github.com/containerbuildsystem/cachi2/actions/runs/10781837227/job/29900605591?pr=612)

There you can observe that all tests were run (however second commit in the sequence does not affect any code) and there was no failure due to insufficient permissions despite the fact that the necessary permissions are added in the second commit, not the first one. Further, the job does this to fetch the code:
```
/usr/bin/git -c protocol.version=2 fetch --prune --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/* +aa22d80cf78f8d981813bfda262e8ba7f2319589:refs/remotes/pull/612/merge
```

and checks out the pull request like this:
```
 /usr/bin/git checkout --progress --force refs/remotes/pull/612/merge
```
This results in HEAD pointing to a merge commit:
```HEAD is now at aa22d80 Merge 7a546e32f90ad33e8e10d95e3164b777af957867 into b3862f2c08d62a901840f33a6a05eabaf46c0365
```
`7a546e3`  is my latest commit, `b3862f2` is the tip of `main` at the moment of the last push. It results in all files modified in the feature branch to be counted together.

.github/workflows/gating.yaml
a-ovchinnikov: Done.

tests/integration/conftest.py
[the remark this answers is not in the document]
a-ovchinnikov: I dropped `_for` in the latest revision.

tests/integration/utils.py
a-ovchinnikov: Done.

tests/integration/utils.py
eskultety: > Further, the job does this to fetch the code:
>```
>/usr/bin/git -c protocol.version=2 fetch --prune --no-recurse-submodules origin +refs/heads/*:refs/remotes/>origin/* +refs/tags/*:refs/tags/* +aa22d80cf78f8d981813bfda262e8ba7f2319589:refs/remotes/pull/612/merge
>```
>and checks out the pull request like this:
 >```
>/usr/bin/git checkout --progress --force refs/remotes/pull/612/merge
>```
>This results in HEAD pointing to a merge commit:
>```
>HEAD is now at aa22d80 Merge 7a546e32f90ad33e8e10d95e3164b777af957867 into b3862f2c08d62a901840f33a6a05eabaf46c0365
>```

I never noticed ^this, TIL GitHub automatically creates a "virtual" merge commit as a preview of the changes what it'd look like if the PR were to be merged to main immediately, thank you!

That said, what you modeled according to GitHub's behaviour won't be the case locally where your branch is properly rebased onto main and you won't have this automatic merge commit which you could inspect for changes. Instead, you'll have your individual commits and so based on your current implementation I am either: 
- forced to always run the integration test suite in an interactive rebase (which should be a choice really and a good practice, not a must) OR 
- set an environment variable to always run **ALL** tests because the last commit would end up yielding no relevant changes

none of which is desirable really because it's both not that much different from the status quo and arguably actually a regression in the behaviour because this solution only really tackles the gating environment but neglects local dev environments. If you instead compare all commits against main instead of just HEAD, then it should work consistently in both local and gating env.

tests/integration/utils.py
```diff
@@ -17,6 +17,16 @@
 import yaml
 from git import Repo
 
+from cachi2.core import resolver
+
+ALL = object()
```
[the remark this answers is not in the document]
a-ovchinnikov: Yes.

tests/integration/utils.py
```diff
@@ -463,3 +472,165 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
         assert dep_file.exists()
         assert dep_file.read_text() == expected_content
+
+
+def retrieve_changed_files_from_git() -> tuple[Path, ...]:
+    repo = Repo(".", search_parent_directories=True)
+    try:
+        # HEAD would be correctly defined locally
+        files = repo.git.diff("--name-only", f"{repo.active_branch.name}..main").split("\n")
+    except TypeError:  # will be raised when in detached HEAD state, e.g. in Github CI
```
eskultety, line 482: Not opposing per se, but can we trust it in general? IOW would it only happen in GH CI? Can't this happen in local envs is what I'm asking? For some reason it's making me uneasy.
Wouldn't 
```
if not(os.getenv("CI") or os.getenv("GITHUB_WORKSPACE")):
  repo.commit("HEAD")
else:
  repo.git.diff(...)
```
be slightly more bulletproof?

Based on https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables which are always set in GH CI env?

eskultety: Uhm, lol, this just happened to me as I added commits on top of your PR locally (I swear I didn't do it on purpose :D ). Whenever I fetch PRs locally I set git to always put them in a detached HEAD state rather than a specific branch (because once I'm done with the code review I just checkout elsewhere and don't need to delete a branch I may not need again) and I was wondering what was wrong with the test results :D.

tests/integration/utils.py
```diff
@@ -463,3 +472,165 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+    """Check that just package managers were affected.
+
+    If any code outside of package managers subtree was affected or if a module
+    shared by package managers was affected will return False.
+
+    >>> just_some_package_managers_were_affected_by((Path('tests/integration/test_pip.py'),))
+    True
+    >>> c = Path('cachi2/core/package_managers/pip.py'), Path('tests/integration/test_pip.py')
+    >>> just_some_package_managers_were_affected_by(c)
+    True
```
eskultety, line 594: FWIW I'm glad you added these to the docstring, really helpful.

a-ovchinnikov: In case this is not a widely known fact one can change directory to tests/integration and then run `python3 -m doctest utils.py` and see which of these statements fail. 

I personally prefer doctests for utility functions because such functions are not busyness logic and should not be tested with UTs directly, but they do need some smoke-tests and I find short doctests to be a nearly perfect fit. Docstrings can be also easily folded away and not pollute the precious vertical space.

tests/integration/conftest.py
```diff
@@ -69,3 +69,18 @@ def local_pypiserver() -> Iterator[None]:
...
+
+def pytest_collection_modifyitems(
+    session: pytest.Session, config: pytest.Config, items: list[pytest.Item]
+) -> None:
+    """Remove redundant tests which don't have to run for the latest code change.
+
+    This function implements a standard pytest hook. Please refer to pytest
+    docs for further information.
+    """
+    skip_mark = pytest.mark.skip(reason="No changes to tested code")
```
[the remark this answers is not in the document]
a-ovchinnikov: Initially I planned for an even more verbose "No changes to code tested by this test", but then have forgotten about it. I would have rather kept my current version because `No changes to test` is ambiguous: does it mean that there is nothing to test or does it mean that a test itself was not modified? As for line-wraps well, this is not some code to actually read, so I am inclined to tolerate that, and besides these lines are but tiny specs on modern 40' 160k+ screens.

tests/integration/utils.py
eskultety: TIL, thank you. It should be possible to do via pytest as well but `pytest --doctest-modules --doctest-continue-on-failure` is failing for me for some reason. Anyway, we should adopt this practice! I would have probably done something had I known about it :) .

tests/integration/conftest.py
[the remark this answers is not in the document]
eskultety: I'd argue that `No changes to tested code` isn't any better in terms of description, but fair enough, it wasn't a deal breaker for me anyway.

tests/integration/utils.py
eskultety: I probably shouldn't have followed with my own example, because it steered the conversation in a way I didn't intend to.

My main point was that it behaves somewhat unexpectedly if it happens, solely because you make the assumption that GitHub CI is the only env which will use detached HEADs which is a standard git feature. Moreover, you also had to add a commentary to explain the exception instead of going the env variable approach where it's immediately clear which case it handles. That is not to say that someone cannot locally set all those GH variables explicitly, but who'd do that honestly voluntarily?

eskultety: > In this specific case one could trim the scope of testing further with -k test_pattern if needed (I checked that). If, however, someone wants to run unrelated tests on a commit that does not affect tested code then these tests will be skipped with correct reason. I.e. if I detach HEAD at https://github.com/containerbuildsystem/cachi2/commit/45c8c7c9b12fb0158f623f41c13f090ca218d513 and then run tox -eintegration -- -k test_gomod_packages I'll end up with 12 skipped and 40 deselected tests vs 51 skipped for just tox -eintegration. 

This is all nice and I don't disagree, but I'm more after the automatic behaviour without any CLI args.

a-ovchinnikov: I mentioned CLI args only to highlight that if someone is up to some complicated workflow they could still have it. The whole point of the change to be drop-in compatible with any regular workflow and to save time.

tests/integration/utils.py
```diff
@@ -463,3 +472,170 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+    repo = Repo(".", search_parent_directories=True)
+    # >>> type(repo.branches)
+    # <class 'git.util.IterableList'>
+    # Despite the fact stated above mypy does not believe one can use 'in' with
+    # repo.branches because branches is an alias to heads and heads are decorated
+    # with @property:
+    #  Unsupported right operand type for in ("Callable[[], IterableList[Head]]")
+    main = "main" if "main" in repo.branches else "origin/main"  # type: ignore
+    try:
+        files = repo.git.diff("--name-only", f"{main}..HEAD").split("\n")
```
slimreaper35, line 487: So, locally, this only works for committed changes. Do we want to add a check for non-committed changes?

eskultety: Yes, but that's IMO a good candidate for a follow-up, this is doing enough already for both the local environments where you can test your changes right before submitting them to the gating CI to crunch for many minutes as well as for the gating to crunch a bit less :).
I'm saying ^this because I already have another follow-up in mind - rebase detection - and those 2 could be paired in another PR.

tests/integration/utils.py
```diff
@@ -463,3 +472,170 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+
+
+def retrieve_changed_files_from_git() -> tuple[Path, ...]:
+    repo = Repo(".", search_parent_directories=True)
+    # >>> type(repo.branches)
+    # <class 'git.util.IterableList'>
+    # Despite the fact stated above mypy does not believe one can use 'in' with
+    # repo.branches because branches is an alias to heads and heads are decorated
+    # with @property:
+    #  Unsupported right operand type for in ("Callable[[], IterableList[Head]]")
```
eskultety, line 484: side note: Digging into GitPython was a nice breeze of great Python code to have a look at :) to understand what you're saying:

> \# repo.branches because branches is an alias to heads and heads are decorated
> \# with property:

Essentially, yes, but I think it's much more poetic than that and I think you're giving mypy too much credit because I don't think mypy would be able to infer the types correctly even if `@property` weren't involved :D and it'd IMO still complain about the string. The reason for that is that `IterableList` is leveraging the `_id_attribute_` class attribute which is of type `str` and eventually translates to the `name` property because of `SymbolicReference`, but that only gets to `Head` via a triple multiple inheritance of `Reference`. For anyone interested, here are sources to infer the whole inheritance chain:

- https://github.com/gitpython-developers/GitPython/blob/cfadd9e5edb8571c2c15878c8eeca91f15a69040/git/refs/head.py#L128
- https://github.com/gitpython-developers/GitPython/blob/cfadd9e5edb8571c2c15878c8eeca91f15a69040/git/refs/reference.py#L41
- https://github.com/gitpython-developers/GitPython/blob/cfadd9e5edb8571c2c15878c8eeca91f15a69040/git/util.py#L1237
- https://github.com/gitpython-developers/GitPython/blob/cfadd9e5edb8571c2c15878c8eeca91f15a69040/git/refs/symbolic.py#L100

tests/integration/utils.py
```diff
@@ -463,3 +472,170 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+    main = "main" if "main" in repo.branches else "origin/main"  # type: ignore
+    try:
+        files = repo.git.diff("--name-only", f"{main}..HEAD").split("\n")
+    # Widest net possible in order not to interfere with testing for no good reason.
+    except Exception:
+        # If there is no main and no origin/main then someone is probably doing
```
eskultety, line 490: Please log this!

eskultety (commented):
Apart from logging the case when the main branch isn't called `main` I only have one other request - please update the README integration testing section mentioning this behaviour and how it can be overridden to either get a full test or a different subset (-k), otherwise LGTM and we're good to merge.

tests/integration/utils.py
a-ovchinnikov: True, I published a much shorter version of the story and stopped digging at an earlier point. Thank you for unraveling it further, it is an interesting story, I became mostly content with my findings when visiting reference.py. To me the most important takeaway is that mypy is incapable of doing correct inference for a legit usecase. Moreover I will never believe that it is impossible to statically infer this. I am using the word 'believe'  to denote that I am not going to prove or disprove my point with code in a language that has an actual type checker, at least not now.

tests/integration/utils.py
a-ovchinnikov: Good point, done.

a-ovchinnikov:
> Apart from logging the case when the main branch isn't called `main` I only have one other request - please update the README integration testing section mentioning this behaviour and how it can be overridden to either get a full test or a different subset (-k), otherwise LGTM and we're good to merge.

Thank you, I have completely forgotten to update README!

README.md
```diff
@@ -258,6 +258,31 @@ Similarly to unit tests, for finer control over which tests get executed, e.g. t
 tox -e integration -- tests/integration/test_package_managers.py::test_packages[gomod_without_deps]
 ```
 
+Integration tests verify package managers handlers behavior and ensure that it remains consistent.
```
eskultety, line 261: For future reference, README/doc changes should ideally go to separate commits from code changes.

eskultety (approved):

tests/integration/utils.py
slimreaper35: When I log something anywhere within these functions, I see nothing unless I add `--log-cli-level=<level>` in `[testenv:integration]`. But the option is already configured below in the [file](https://github.com/containerbuildsystem/cachi2/blob/main/tox.ini#L77)

Is it just me? I don't understand.

eskultety: > Is it just me? 

It's not, turns out, we need to set `log_cli=true` for pytest as well then I can see:

`2024-09-13 17:44:38 DEBUG HELLO FROM TEST`

based on this diff:
```
diff --git a/tests/integration/utils.py b/tests/integration/utils.py
index 80db1fd1..f7c89398 100644
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -484,6 +484,7 @@ def retrieve_changed_files_from_git() -> tuple[Path, ...]:
     #  Unsupported right operand type for in ("Callable[[], IterableList[Head]]")
     main = "main" if "main" in repo.branches else "origin/main"  # type: ignore
     try:
+        log.debug("HELLO FROM TEST")
         files = repo.git.diff("--name-only", f"{main}..HEAD").split("\n")
     # Widest net possible in order not to interfere with testing for no good reason.
     except Exception as e:
```

And I've even seen @a-ovchinnikov's warning when I made a typo in there :D:
```
------------------------------------------ live log collection -------------------------------------------
2024-09-13 17:48:10 WARNING Detection of changed files unexpectedly failed. This either indicates that both 'main' and 'origin/main' branches are missing in a repo or a more fundamental failure. The tool will attempt to recover from this by not filtering any tests out. This is the exception that was encountered: 'Logger' object has no attribute 'DEBUG'
collected 52 items
```

so yeah, `log_cli=true` works

slimreaper35: Yeah that would work, but then `tox -e py3` shows countless amount of logs.

a-ovchinnikov: `log_cli_level = DEBUG` is supposed to log everything. it should be at least `INFO` for tests, or maybe even just `WARNING`.

eskultety: > Yeah that would work, but then tox -e py3 shows countless amount of logs 🤯

It does, do you have a better idea how to handle debug logs from test? I mean, it's kinda expected that once we turn it on, debug logs litter the output, that's their inherent nature, isn't it?

Neither is IMO a problem this PR inherently needs to solve (unless you favour doing it here). We identified the problem and I agree we should decrease the default log level, allowing users to override it from tox's CLI if needed.

slimreaper35: I thought that we could use `log_cli` only for integration tests, but `log-cli` is not available in CLI.
Let's drop it in this PR.

slimreaper35 (approved):

README.md
a-ovchinnikov: Done!

PR #615, opened 2024-08-30
bandit security corrections

slimreaper35:

### DIFF
--- a/cachi2/core/package_managers/gomod.py
+++ b/cachi2/core/package_managers/gomod.py
@@ -3,7 +3,7 @@ import logging
 import os
 import re
 import shutil
-import subprocess  # nosec
+import subprocess
 import tempfile
 from datetime import datetime
 from functools import cached_property
--- a/cachi2/core/scm.py
+++ b/cachi2/core/scm.py
@@ -120,8 +120,6 @@ def clone_as_tarball(url: str, ref: str, to_path: Path) -> None:
             _reset_git_head(repo, ref)
 
             with tarfile.open(to_path, mode="w:gz") as archive:
-                # GitPython wrongly annotates working_dir as Optional, it cannot be None
-                assert repo.working_dir is not None  # nosec assert_used
                 archive.add(repo.working_dir, "app")
 
             return
--- a/cachi2/core/utils.py
+++ b/cachi2/core/utils.py
@@ -4,7 +4,7 @@ import logging
 import os
 import re
 import shutil
-import subprocess  # nosec
+import subprocess
 import sys
 from functools import cache
 from pathlib import Path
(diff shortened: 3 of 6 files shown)


pyproject.toml
```diff
@@ -94,3 +94,6 @@ exclude_lines = [
   "if TYPE_CHECKING:",
   "return NotImplemented",
 ]
+
+[tool.bandit]
+skips = ["B404", "B603"]
```
eskultety, line 99: Since this is an upstream enforced disablement, we should put each on a separate line and add a commentary with a short explanation AND/OR a URL to the official docs on the rule

cachi2/core/utils.py
```diff
@@ -48,7 +48,7 @@ def run_cmd(cmd: Sequence[str], params: dict) -> str:
             ),
         )
 
-    response = subprocess.run([executable_path, *args], **params)  # nosec
```
eskultety, line 51: Lol, okay, so we always ignored literally ALL bandit rules on our own subprocess run wrapper, good to know :D.

cachi2/core/package_managers/gomod.py
```diff
@@ -3,7 +3,7 @@
 import os
 import re
 import shutil
-import subprocess  # nosec
+import subprocess
```
eskultety, line 6: **Commit message**:
Can we add a very brief decision on why we're disabling them? This is solely to have that bit of info tracked in the git history.

eskultety: > Can we add a very brief decision on why we're disabling them? This is solely to have that bit of info tracked in the git history.

Scratch this, hit send too soon. The only change I'd make is:

_s/The commit adds the options:/It's better to explicitly list all disabled rules and so this patch disables:_

cachi2/core/scm.py
```diff
@@ -120,8 +120,6 @@ def clone_as_tarball(url: str, ref: str, to_path: Path) -> None:
             _reset_git_head(repo, ref)
 
             with tarfile.open(to_path, mode="w:gz") as archive:
-                # GitPython wrongly annotates working_dir as Optional, it cannot be None
-                assert repo.working_dir is not None  # nosec assert_used
```
eskultety, line 124: Is this not needed anymore? Since you haven't disabled a rule, does bandit not report a problem anymore?

eskultety:
LGTM - nitpicks mostly.

cachi2/core/scm.py
slimreaper35: No, `working_tree_dir` attribute is Optional, not `working_dir`.
https://github.com/gitpython-developers/GitPython/blob/main/git/repo/base.py#L126

pyproject.toml
slimreaper35: I thought the git blame command would be sufficient :smile_cat:

eskultety: I mean, yes, RTFM is a universally applicable rule and yet programmers write all sorts of commentaries all around the source code to spare other poor soul's time ;) . In general, one also can't universally rely on commit messages providing any further information - "accidents happen" and commit messages end up empty (not our case), it's just a good practice IMO that costs us nothing.

taylormadore (approved):

eskultety (approved):

PR #622, opened 2024-09-02
mypy improvements

slimreaper35:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,7 @@ profile = "black"
 line_length = 100
 
 [tool.mypy]
+python_version = "3.9"
 plugins = ["pydantic.mypy"]
 disallow_untyped_defs = true
 disallow_incomplete_defs = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,10 +80,6 @@ disallow_untyped_defs = true
 disallow_incomplete_defs = true
 disallow_untyped_decorators = true
 
-[[tool.mypy.overrides]]
-module = ["semver"]
-ignore_missing_imports = true
-
 [tool.coverage.report]
 skip_covered = true
 show_missing = true


eskultety:
Can you share the reproducer? What sequence of commands/actions did you run?

slimreaper35:
> Can you share the reproducer? What sequence of commands/actions did you run?

https://github.com/containerbuildsystem/cachi2/actions/runs/10662950660/job/29551119082

Locally, I had installed venv with python3.12, and mypy check was passing.

slimreaper35:
Hopefully, I am not missing something here. But I remember we had a variable in Makefile that could be overwritten via argument from CLI, right?

eskultety:
> > Can you share the reproducer? What sequence of commands/actions did you run?
> 
> https://github.com/containerbuildsystem/cachi2/actions/runs/10662950660/job/29551119082
> 
> Locally, I had installed venv with python3.12, and mypy check was passing.

Right.

> Hopefully, I am not missing something here. But I remember we had a variable in Makefile that could be overwritten via argument from CLI, right?

We did, and I ditched it eventually (it all started with commit 25d45878a19eff4350daa379f17a40653a81876c). It defaulted to python 3.9 which is ancient and only being kept alive in LTS distros so most development machines won't have it installed by default. Before that commit in its default setting it forced everyone to have Python 3.9 installed otherwise it failed - not a pleasant UX.
I don't think what is being proposed is the correct fix, because it brings us back to the past where we limited what a development environment should look like (IOW we thought we knew better), that's IMO wrong and developers should feel free to create the venv using their preferred version of Python and not be forced to install all versions of python (hence we run pip-compile in a 3.9 container). That's what the gating CI is for, note the environment used in gating doesn't necessarily need to correspond to the local dev environment. Besides, if you fix it in the Makefile, I can still do `python3 -m venv venv && ...` and still not  use 3.9 because let's face it, Makefiles are not that common for Python related actions in the ecosystem, so it's only fixing a very specific niche use case. I agree that since we need to stay strictly compatible with 3.9 until it's EOL (unless we adopt a different rule) it's not ideal, but then forcing one's environment that way isn't IMO a solution, so should be clearly documented and devs should either accept that they may see failures in the gating CI if they used anything other than python3.9 or should be able to create an identical environment. That said, **I agree we should bring the env variable back** as otherwise the Makefile is limiting because it only respects the default Python installation.

With development freedom in mind what I believe you're looking for is to:

- revert commit 4fba58c635a43a7e2d648dd7e359900204cb64e0 and use it to create the venv again
- allow passing `--python-version` to the mypy tox environment (note one doesn't actually need to have python 3.9 installed for this which is great)
- document this behaviour and the usage of the env variable in the test docs for the sake of preventing others hit the same problem as you did (eventually we might want to have an FAQ section for this, shall we start now?)

You can start with
```
diff --git a/tox.ini b/tox.ini
index 6da3f93c..6de76448 100644
--- a/tox.ini
+++ b/tox.ini
@@ -54,7 +54,7 @@ commands =
 
 [testenv:mypy]
 commands =
-    mypy --install-types --non-interactive cachi2 tests
+    mypy --install-types --non-interactive cachi2 tests {posargs}
```
but I think it's worth allowing {posargs} to all tox test environments so that we don't have to go through this again.

eskultety:
> > Can you share the reproducer? What sequence of commands/actions did you run?
> 
> https://github.com/containerbuildsystem/cachi2/actions/runs/10662950660/job/29551119082
> 
> Locally, I had installed venv with python3.12, and mypy check was passing.

On a positive note at least we know the gating CI works reliably :) (for obvious issues at least).

slimreaper35:
I think I found a better solution. We just need to set base python in _tox.ini_. This should work fine even if venv is created with newer python versions. Can you check that ?

```
[testenv:mypy]
basepython = python3.9
mypy --install-types --non-interactive cachi2 tests
```

eskultety:
> I think I found a better solution. We just need to set base python in _tox.ini_. This should work fine even if venv is created with newer python versions. Can you check that ?
> 
> ```
> [testenv:mypy]
> basepython = python3.9
> mypy --install-types --non-interactive cachi2 tests
> ```

This just moves the your original fix from Makefile one level deeper but it still forces you to have 3.9 installed locally. If one doesn't want to install another python version locally and just wants to let the gating CI take care of the checks, that's fine. If one wishes to replicate the environment, they can install python3.9 and create venv either manually or via the Makefile (provided we re-introduce the variable) and run tests that way or like I told you, allow running mypy with `--python-version` which actually doesn't expect you to have a particular version of python installed --> complete freedom, no version locking. I'm still not convinced this is a superior solution.

eskultety:
> Is this only relevant for mypy? Can other linters have similar issues?

Mypy technically isn't a linter, it's a static analyzer tool. So, no, I don't expect the linters we use face similar issues although I can't predict that. I'm missing context for your question, so I'm just going to make a guess - if your question is about the usage of `posargs` in tox then I can imagine other options which may be of interest to people to test with (potentially proposing them as default afterwards) those linters, so I think we can support pretty much any arbitrary option each linter supports respectively that way. If I'm off in my assumption, then please provide more context for your question so that I can improve my answer.

ben-alkov:
IMO we should ditch the Makefile all together and rely exclusively on Tox, which **doesn't require any specific Python version to be installed**, and can provide any version specified in tox.ini.

eskultety:
> IMO we should ditch the Makefile all together and rely exclusively on Tox, which **doesn't require any specific Python version to be installed**, and can provide any version specified in tox.ini.

Rebellious! This will require significant changes to the docs, but yeah, there's nothing which couldn't be documented with enough examples, so yeah, why not. That said, I'd leave that for a follow-up.

slimreaper35:
So the goal that I try to achieve with this PR is that test results are the same locally and on GitHub (without any further work).

1. I thought, that putting `basepython` into _tox.ini_ would solve this, but if it requires the installation of Python 3.9, then let's skip that.
2. I like `--python-version` that you suggested, so we can put that into pyproject.toml (in case someone runs `mypy` directly without `tox`.
3. Basically, my question was about adding similar things, e.g., `target_version` for `black` and other linters, to prevent unexpected failures in GH actions.

ben-alkov:
> Rebellious!

Wait til you see my tox.ini refactor based on https://www.seanh.cc/2018/11/14/tox :laughing:

eskultety:
> So the goal that I try to achieve with this PR is that test results are the same locally and on GitHub (without any further work).
> 
>     1. I thought, that putting `basepython` into _tox.ini_ would solve this, but if it requires the installation of Python 3.9, then let's skip that.
> 
>     2. I like `--python-version` that you suggested, so we can put that into pyproject.toml (in case someone runs `mypy` directly without `tox`.

Uhm, we can't put it into pyproject.toml though because you'd then enforce the python installation the same way. The difference between 1) and 2) is that in 1) you always enforce a particular python installation whereas with 2) **if enabled properly in the tox.ini file via `{posargs}` specifically** then you are not enforcing anything because `posargs` translates to NOP if no option is passed on the CLI. If people run mypy without tox they should be able to pass that CLI flag the same way as with tox, I don't see a difference.
So there **is** a particular way to enable this otherwise we end up enforcing the installation like with any other method that was discussed.

> 
>     3. Basically, my question was about adding similar things, e.g., `target_version` for `black` and other linters, to prevent unexpected failures in GH actions.

Yes, that's why one of my earlier comments said that `{posargs}` can be enabled unconditionally with every tox environment so that people can pass arbitrary CLI opts as needed in their particular dev environment.

slimreaper35:
https://mypy.readthedocs.io/en/stable/common_issues.html#version-and-platform-checks

> To target a different Python version, use the [--python-version X.Y](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version) flag. For example, to verify your code typechecks if were run using Python 3.8, pass in [--python-version 3.8](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version) from the command line. Note that you do not need to have Python 3.8 installed to perform this check.

eskultety:
> https://mypy.readthedocs.io/en/stable/common_issues.html#version-and-platform-checks
> 
> > To target a different Python version, use the [--python-version X.Y](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version) flag. For example, to verify your code typechecks if were run using Python 3.8, pass in [--python-version 3.8](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version) from the command line. Note that you do not need to have Python 3.8 installed to perform this check.
> 
> What is the issue with having it in pyproject.toml ?

Sigh, sorry for the noise, it was clearly at the end my working day and I didn't think before I wrote that. On one hand I told you one doesn't need to have a specific python installed for that check and then I immediately contradicted myself, should have left it for the morning :). You're right, we can adopt the flag widely based on the docs which you also should reference in the commit message  ;) .

Makefile
```diff
@@ -10,7 +10,7 @@ clean:
 	rm -rf dist venv .tox *.egg-info *.log*
 
 venv:
-	/usr/bin/env python3 -m venv venv
+	/usr/bin/env python -m venv venv
```
eskultety, line 13: This change is unnecessary + we specifically DO want to enforce **some** python3 installation (i.e. respect the python3 symlink that is compatible with `update-alternatives`) and we want to be absolutely open and clear about it, so please keep the original.

eskultety (approved):
One last nitpick to resolve, but I'm already giving you my ACK.

Makefile
slimreaper35: That wasn't intended; I just hit the backspace one more time.

eskultety:
I see you renamed the PR, do you plan on working on the follow-up to revert https://github.com/containerbuildsystem/cachi2/commit/4fba58c635a43a7e2d648dd7e359900204cb64e0 as well or not? Just checking.

slimreaper35:
> I see you renamed the PR, do you plan on working on the follow-up to revert https://github.com/containerbuildsystem/cachi2/commit/4fba58c635a43a7e2d648dd7e359900204cb64e0 as well or not? Just checking.

Do we actually want to do that? Since it has been solved differently.

eskultety:
> > I see you renamed the PR, do you plan on working on the follow-up to revert [4fba58c](https://github.com/containerbuildsystem/cachi2/commit/4fba58c635a43a7e2d648dd7e359900204cb64e0) as well or not? Just checking.
> 
> Do we actually want to do that? Since it has been solved differently.

We solved one piece of the puzzle, yes and I'm not expecting that work to happen in this PR. However, that commit still took away the possibility for users to create the venv using any python installation they wish to and always uses the default installation, otherwise they have to work around it manually. Not a big deal, it's also a corner case, but it is unarguably a limitation.

a-ovchinnikov (approved):

PR #632, opened 2024-09-05
Introduce support for Go 1.23

eskultety:
Go 1.23 was released last month [1]. The only notable change on the tooling with essentially no real impact is the introduction of telemetry data gathering which is documented to be opt-in. To stay on the safe side patch 2 disables telemetry explicitly on Go releases which support it.

[1] https://tip.golang.org/doc/go1.23

- [x] Docs links in the code are still valid (if docs were updated)

### DIFF
--- a/cachi2/core/package_managers/gomod.py
+++ b/cachi2/core/package_managers/gomod.py
@@ -767,7 +767,7 @@ def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
     GO_121 = version.Version("1.21")
     go = Go()
     target_version = None
-    go_max_version = version.Version("1.22")
+    go_max_version = version.Version("1.23")
     go_base_version = go.version
     go_mod_version_msg = "go.mod reported versions: '%s'[go], '%s'[toolchain]"
 
--- a/cachi2/core/package_managers/gomod.py
+++ b/cachi2/core/package_managers/gomod.py
@@ -827,6 +827,13 @@ def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
     return go
 
 
+def _disable_telemetry(go: Go, run_params: dict[str, Any]) -> None:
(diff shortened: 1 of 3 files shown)


cachi2/core/package_managers/gomod.py
```diff
@@ -827,6 +827,13 @@ def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
     return go
 
 
+def _disable_telemetry(go: Go, run_params: dict[str, Any]) -> None:
+    telemetry = go(["env", "GOTELEMETRY"], run_params).rstrip()
+    if telemetry and telemetry != "off":
```
a-ovchinnikov, line 832: A nitpick: `telemetry` is already a string, first check is redundant. Also do you really need a check if you are disabling telemetry anyway? I.e.
```suggestion
def _disable_telemetry(go: Go, run_params: dict[str, Any]) -> None:
        log.debug("Disabling Go telemetry")
        go(["telemetry", "off"], run_params)
will apparently do the same.

a-ovchinnikov (approved):

cachi2/core/package_managers/gomod.py
eskultety: > telemetry is already a string

and if it's empty then that Go toolchain doesn't support telemetry and the command as you suggest will fail, e.g. on Go 1.20 which we can also fallback to (still).

> Also do you really need a check if you are disabling telemetry anyway? I.e.

Yes I do, for the reason above as well as not to spin up another subprocess if I don't need to (i.e. when it's already "off")

taylormadore (approved):

cachi2/core/package_managers/gomod.py
a-ovchinnikov: I see, thank you!

PR #633, opened 2024-09-05
Makefile update - Re-introduce PYTHON_VERSION_VENV variable, enforce generic 'base_python' for Tox

eskultety:
This updates the Makefile so that we relax some of our current strict limitations on what the local dev environment should look like and gives more freedom back to developer's hands in terms of deciding which versions of Python they want and need to have installed to work with/develop cachi2 regardless of our strict gating CI environments (which remain unchanged).


tox.ini
```diff
@@ -5,7 +5,6 @@ envlist =
     black
     isort
     mypy
-    py3{9,10,11,12}
```
slimreaper35, line 8: When I run tox I don't see pytest run at all. Did you want to add `py3` to the list instead, or am I missing something ?

a-ovchinnikov:
Wouldn't this result in surprises to contributors who developed a feature with Py3.10+ and then learned from the Gate that it has to be also Py3.9 compatible?

tox.ini
eskultety: Great catch! I do need `py3` indeed. Fixed in the new revision.

eskultety:
Since v1:
- changed the last commit  so that `py3` denotes the default Python environment in tox's `env_list` using the default Python in the environment `tox` runs in, see the explanation in the commit message

If the consensus is to keep tox run ALL python environments forcing everyone to have all Pythons installed, then I'll drop the last commit and will somehow (not sure how) rework commit 2.

slimreaper35:
Could you update README.md as well ? (not to mention `python3.9`)
https://github.com/containerbuildsystem/cachi2?tab=readme-ov-file#running-unit-tests

eskultety:
Since v2:
- updated README
- dropped env changes to `tox.ini` as I found out there's the `skip_missing_interpreters` config option which pretty much does what I'd initially expected and so it solves the use case that motivated me for this PR

README.md
```diff
@@ -227,7 +224,9 @@ tox -e py39 -- tests/unit/test_cli.py::TestGenerateEnv::test_invalid_format
 tox -e py39 -- tests/unit/extras/test_envfile.py::test_cannot_determine_format
 ```
 
-In short, tox passes all arguments to the right of `--` directly to pytest.
+In short, tox passes all arguments to the right of `--` directly to pytest. For more information on
```
slimreaper35, line 227: I would rephrase this slightly to mention what `--` means. It is not something specific to tox only.

README.md
```diff
@@ -140,17 +140,14 @@ This installs the Cachi2 CLI in [editable mode](https://setuptools.pypa.io/en/la
...
-```
-
-The CLI also depends on the following non-Python dependencies:
-
-```shell
-dnf install golang-bin git
-```
+Note though that we also have several non-Pythonic runtime dependencies, so depending on which part
+of cachi2 you want to work on you may need to install further binaries or toolkits. Please refer to
+our Containerfile on what those dependencies are as that always contains the latest information.
```
slimreaper35, line 150:
```suggestion
our [Dockerfile](https://github.com/containerbuildsystem/cachi2/blob/main/Dockerfile) on what those dependencies are as that always contains the latest information.
```

eskultety: We should always promote it as a Containerfile, the only reason for having a Dockerfile (and Containerfile as a symlink) is due to Dependabot's limitations, that's not something by choice. It is true that it's currently a symlink, so GH UI will redirect it automatically which may be confusing for some people, but the moment we either replace dependabot or it magically starts supporting some alternative container file names, we don't need to change anything in the docs.

README.md
eskultety: Why? Explaining shell positional arguments and how it translates to tox `{posargs}` feels largely outside of the scope of user-facing docs on a tool that doesn't directly deal with that, IOW users don't need to understand shell in order to develop/test or even use this tool, so the original wording used in this case feels adequate for the desired outcome. Moreover, I only expanded the sentence providing references to other sources explaining the necessary internals should the user be interested in exploring them; note that I didn't try to explain how people should read a Dockerfile or how tox.ini works, so I'd preferably not like to take any action here.

README.md
eskultety: I'll update this as suggested.

a-ovchinnikov (approved):
Appears benign. LGTM.

README.md
slimreaper35:
```suggestion
Double dash (--) signifies the end of command options, so any arguments after it are passed directly to pytest without further interpretation by tox.
```
...if it is not out of scope

slimreaper35 (approved):

PR #637, opened 2024-09-05
Enable typing on pip

### LINKED ISSUE #485: Enable typing for PipRequirement.from_line class method
We purposefully don't enable typing for the `PipRequirement.from_line` class method in https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/pip.py as there seem to be quite a few problems. We should be able to fix those and enable type annotations though.

ianrochapg:
Typing was disabled in PipRequirement due to several small issues left over time, so this PR addresses all of them and re-enables typing on the file .

Resolves #485 

### DIFF
--- a/cachi2/core/package_managers/pip.py
+++ b/cachi2/core/package_managers/pip.py
@@ -25,7 +25,6 @@ from typing import (
     Pattern,
     Union,
     cast,
-    no_type_check,
 )
 
 import tomli
@@ -1068,7 +1067,7 @@ class PipRequirementsFile:
 
         :return: a dict with the keys ``requirements`` and ``options``
         """
-        parsed: dict[str, list[str]] = {"requirements": [], "options": []}
+        parsed: dict[str, list[Union[str, PipRequirement]]] = {"requirements": [], "options": []}
 
         for line in self._read_lines():
             (
@@ -1189,11 +1188,11 @@ class PipRequirement:
         self.package: str = ""
         # The package name as defined in the requirement line
         self.raw_package: str = ""
-        self.extras: list[str] = []
-        self.version_specs: list[str] = []
-        self.environment_marker: str = ""
+        self.extras: set[str] = set()
+        self.version_specs: list[tuple[str, str]] = []
+        self.environment_marker: Optional[str] = None
         self.hashes: list[str] = []
-        self.qualifiers: dict[str, Any] = {}
+        self.qualifiers: dict[str, str] = {}
 
         self.kind: Literal["pypi", "url", "vcs"]
         self.download_line: str = ""
@@ -1260,12 +1259,12 @@ class PipRequirement:
                 )
 
         requirement = self.__class__()
-
+        # Extras are incorrectly treated as part of the URL itself. If we're setting
+        # the URL, they must be skipped.
+        if not url:
+            requirement.extras = set(self.extras)
         requirement.package = self.package
         requirement.raw_package = self.raw_package
-        # Extras are incorrectly treated as part of the URL itself. If we're setting
-        # the URL, clear them.
-        requirement.extras = [] if url else list(self.extras)
         # Version specs are ignored by pip when applied to a URL, let's do the same.
         requirement.version_specs = [] if url else list(self.version_specs)
         requirement.environment_marker = self.environment_marker
@@ -1277,9 +1276,8 @@ class PipRequirement:
 
         return requirement
 
-    @no_type_check
     @classmethod
-    def from_line(cls, line: str, options: list[str]) -> Optional["PipRequirement"]:
+    def from_line(cls, line: str, options: list[str]) -> "PipRequirement":
         """Create an instance of PipRequirement from the given requirement and its options.
 
         Only ``url`` and ``vcs`` direct access requirements are supported. ``file`` is not.
@@ -1289,23 +1287,16 @@ class PipRequirement:
         :return: PipRequirement instance
         """
         to_be_parsed = line
-        qualifiers = {}
+        qualifiers: dict[str, str] = {}
         requirement = cls()
 
-        direct_access_kind, is_direct_access = cls._assess_direct_access_requirement(line)
-        if is_direct_access:
-            if direct_access_kind in ["url", "vcs"]:
-                requirement.kind = direct_access_kind
-                to_be_parsed, qualifiers = cls._adjust_direct_access_requirement(
-                    to_be_parsed, cls.HAS_NAME_IN_DIRECT_ACCESS_REQUIREMENT
-                )
-            else:
-                raise UnsupportedFeature(
-                    f"Direct references with {direct_access_kind!r} scheme are not supported, "
-                    f"{to_be_parsed!r}"
-                )
-        else:
+        if not (direct_access_kind := cls._assess_direct_access_requirement(line)):
             requirement.kind = "pypi"
+        else:
+            requirement.kind = direct_access_kind
+            to_be_parsed, qualifiers = cls._adjust_direct_access_requirement(
+                to_be_parsed, cls.HAS_NAME_IN_DIRECT_ACCESS_REQUIREMENT
+            )
 
         try:
             req = Requirement(to_be_parsed)
@@ -1328,7 +1319,7 @@ class PipRequirement:
         return requirement
 
     @staticmethod
-    def _assess_direct_access_requirement(line: str) -> tuple[Optional[str], bool]:
+    def _assess_direct_access_requirement(line: str) -> Optional[Literal["url", "vcs"]]:
         """Determine if the line contains a direct access requirement.
 
         :param str line: the requirement line
@@ -1355,10 +1346,10 @@ class PipRequirement:
             "svn+http",
             "svn+https",
         }
-        direct_access_kind = None
+        direct_access_kind: Literal["url", "vcs"]
 
         if ":" not in line:
-            return None, False
+            return None
         # Extract the scheme from the line and strip off the package name if needed
         # e.g. name @ https://...
         scheme_parts = line.split(":", 1)[0].split("@")
@@ -1373,9 +1364,11 @@ class PipRequirement:
         elif scheme in VCS_SCHEMES:
             direct_access_kind = "vcs"
         else:
-            direct_access_kind = scheme
+            raise UnsupportedFeature(
+                f"Direct references with {scheme!r} scheme are not supported, " f"{line!r}"
+            )
 
-        return direct_access_kind, True
+        return direct_access_kind
 
     @staticmethod
     def _adjust_direct_access_requirement(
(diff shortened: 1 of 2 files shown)


cachi2/core/package_managers/pip.py
```diff
@@ -1265,7 +1264,7 @@ def copy(
         requirement.raw_package = self.raw_package
         # Extras are incorrectly treated as part of the URL itself. If we're setting
         # the URL, clear them.
-        requirement.extras = [] if url else list(self.extras)
+        requirement.extras = set() if url else set(self.extras)
```
a-ovchinnikov, line 1267: Non-blocking: according to L1191 `self.extras` is already referring to a set.

cachi2/core/package_managers/pip.py
```diff
@@ -1289,13 +1287,13 @@ def from_line(cls, line: str, options: list[str]) -> Optional["PipRequirement"]:
...
+        qualifiers: dict[str, str] = {}
         requirement = cls()
 
-        direct_access_kind, is_direct_access = cls._assess_direct_access_requirement(line)
-        if is_direct_access:
+        direct_access_kind = cls._assess_direct_access_requirement(line)
+        if direct_access_kind:
             if direct_access_kind in ["url", "vcs"]:
-                requirement.kind = direct_access_kind
+                requirement.kind = cast(Literal["pypi", "url", "vcs"], direct_access_kind)
```
a-ovchinnikov, line 1296: A non-blocking question: why is this cast necessary?

a-ovchinnikov (approved):

cachi2/core/package_managers/pip.py
ben-alkov: mypy refuses to accept `requirement.kind = direct_access_kind` where `direct_access_kind` is type str.

a-ovchinnikov: This feels like `_assess_direct_access_requirement` is mistyped. If it returned `Optional[Literal[...]` instead of a string then mypy would not have complained. And since it effectively works as a constructor for this type it would be fit to reject an incorrect input value there.

cachi2/core/package_managers/pip.py
ben-alkov: Really, this should be `requirement.extras = set(self.extras) if not url`, because `requirement.extras` **is already** an empty set from `__init__()`, and it's not being assigned to anywhere else (that I can find).

cachi2/core/package_managers/pip.py
ben-alkov: > If it returned Optional[Literal[...]

It did, and mypy made the same complaint. Also, returning None here is redundant and type-confusing. Using `""` as False means we can keep it string-typed (also very Pythonic).

> reject an incorrect input value there

See above re: empty string

a-ovchinnikov: > It did, and mypy made the same complaint.
Did it make it on the same line though? For me mypy started complaining about return type of `_assess_direct_access_requirement` instead of the line it was used on. Now if someone is to use a cast then I'd say that the return statement is a better place for that. The problem here is that we want to simultaneously use a wide type and also have it refined. If this is important I would suggest introducing an enum just for that. It would also cleanly handle validation (which is now pushed to L1302).

Using `Optional` makes this an `Option` or `Maybe` type (depending on once's preference for proper statically typed languages), which is arguably a cleaner way to convey intent from type design perspective. This is probably not too Pythonic, OTOH one could argue that having casts in code is even less Pythonic.

ben-alkov: > one could argue that having casts in code is even less Pythonic

Absolutely agree there - there's a lot about mypy that is not very Pythonic

ben-alkov: > Did it make it on the same line though?

Yes

a-ovchinnikov: Oh, I know what is missing: currently the helper returns unrecognized kind as is and lets its client fail. To make it cleaner we could either add an "unsupported" literal, and have a cast in return, or better yet raise right in the helper. Or have an actual type to represent kinds. Anyway, this is not a blocker, and probably belongs to a separate refactoring effort. The change looks good to me  (except the `cast`, casts will never look good to me, sorry casts) and I have +1-ed it.

ben-alkov: > currently the helper returns unrecognized kind as is and lets its client fail. To make it cleaner we could ... raise right in the helper

Nice! :taco:

tests/unit/package_managers/test_pip.py
```diff
@@ -3,7 +3,7 @@
 from copy import deepcopy
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Collection, Literal, Optional, Union
+from typing import Any, Collection, Dict, List, Literal, Optional, Union
```
slimreaper35, line 6: why importing these types and not using dict and list ?

eskultety: Agreed. We're heavily inconsistent in usage of `dict`  vs `Dict` for type hinting, we need to improve on that. 
In context of this commit, this is the only occurrence of `Dict`, so to keep it consistent for now, let's go with `dict`.

tests/unit/package_managers/test_pip.py
```diff
@@ -2585,7 +2585,7 @@ def test_pip_requirement_copy(
         self,
         requirement_line: str,
         requirement_options: list[str],
-        new_values: dict[str, str],
+        new_values: Union[Dict[str, str], Dict[str, List[str]]],
```
eskultety, line 2588: nitpick: `dict[str, Union[str, List[str]]]` is shorter to write.

cachi2/core/package_managers/pip.py
```diff
@@ -1289,7 +1289,7 @@ def from_line(cls, line: str, options: list[str]) -> Optional["PipRequirement"]:
         :return: PipRequirement instance
         """
         to_be_parsed = line
-        qualifiers = {}
+        qualifiers: dict[str, str]  = {}
```
eskultety: flake8 ` E221: multiple spaces before operator`

cachi2/core/package_managers/pip.py
```diff
@@ -1279,7 +1279,7 @@ def copy(
 
     @no_type_check
     @classmethod
-    def from_line(cls, line: str, options: list[str]) -> Optional["PipRequirement"]:
+    def from_line(cls, line: str, options: list[str]) -> "PipRequirement":
```
eskultety, line 1280: This patch should be squashed into the previous one.

cachi2/core/package_managers/pip.py
```diff
@@ -1289,7 +1289,7 @@ def from_line(cls, line: str, options: list[str]) -> "PipRequirement":
         :return: PipRequirement instance
         """
         to_be_parsed = line
-        qualifiers: dict[str, str]  = {}
+        qualifiers: dict[str, str] = {}
```
eskultety, line 1290: This one line fix belongs to the previous patch which introduced the E221 styling issue.

cachi2/core/package_managers/pip.py
eskultety: Also, the constructor for this class defines the attribute as `self.qualifiers: dict[str, Any] = {}`, so we should be consistent and change the instance attribute to a more stricter type to follow the rest of the module.

cachi2/core/package_managers/pip.py
```diff
@@ -1260,12 +1260,12 @@ def copy(
                 )
 
         requirement = self.__class__()
-
-        requirement.package = self.package
-        requirement.raw_package = self.raw_package
         # Extras are incorrectly treated as part of the URL itself. If we're setting
         # the URL, clear them.
```
eskultety, line 1263: The commentary doesn't apply anymore after this change, so please adjust it.

cachi2/core/package_managers/pip.py
```diff
@@ -1190,8 +1190,8 @@ def __init__(self) -> None:
         # The package name as defined in the requirement line
         self.raw_package: str = ""
         self.extras: set[str] = set()
-        self.version_specs: list[str] = []
-        self.environment_marker: str = ""
+        self.version_specs: list[tuple[str, str]] = []
+        self.environment_marker: Optional[str] = None
```
eskultety, line 1193: How are the 2 hunks in this patch related? Is there a direct relationship? Unless there IS a connection, this should be squashed to commit https://github.com/containerbuildsystem/cachi2/pull/637/commits/6d10d8f85f1d2e9be7c95a789471d7814d1114af and commit message should be adjusted.

cachi2/core/package_managers/pip.py
```diff
@@ -25,7 +25,6 @@
     Pattern,
     Union,
     cast,
-    no_type_check,
```
eskultety, line 28: **Commit message:**

> ensuring completion of the task.

What task? If read the commit in the git history, then without the context of the previous patches or this PR one doesn't know what particular task you're talking about. It should be something along the lines of "Enable typing for the PipRequirement class" or something similar. This commit is also the one which should contain the `resolves` keyword (I know you have it linked in you PR description, but that shouldn't be enough, because you don't have the information in your local git log) only in the web UI.

eskultety (commented):
Pretty much all commit messages carry the same pattern:
> Fixed small typing issue in ...

That doesn't tell anyone anything without looking at the actual diff with some heavy context to see where all the attributes/variables in question are used. Please rewrite the commit messages so that they contain some useful information, e.g. why we're making that change, what exact error would mypy emit if we didn't fix this, etc.

I wouldn't mind squashing all changes specifically related to the `PipRequirement` class and listing them in the commit message with reasoning because once you have the reasoning, it's easier to see which changes are related to get the task in question done. That said I don't mind splitting the commit messages (provided some meaningful text is added), just note that every single commit message needs to pass the unit tests & linters (which it currently doesn't).

cachi2/core/package_managers/pip.py
ben-alkov: walrus++

tests/unit/package_managers/test_pip.py
eskultety: Mypy is once again acting up on dictionary unpacking on `new_requirement = original_requirement.copy(**new_values)`. It does work if the values are passed by keywords, but fair enough, I think the copy function needs a refactor anyway, so I'm okay with this as is.

cachi2/core/package_managers/pip.py
eskultety, line 28: **Commit message:**

> Resolves #485

Please use direct URLs instead of numbers in the commit messages so that it's properly rendered in local git histories.

Also, thanks to your example I think I now know what I recently did wrong in one of my commit messages where I used the keyword, but GitHub didn't close the issue and I had to close it manually - you don't use a colon! You may be onto something :) .

PR #639, opened 2024-09-06
bundler: package metadata

slimreaper35:


cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +31,49 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+    """Get package metadata - name and version."""
+    path_dependencies = [dep for dep in dependencies if isinstance(dep, PathDependency)]
+    metadata = _extract_metadata_from_gemspec_file(path_dependencies)
+    if metadata is not None:
+        return metadata
+
+    # fallback to git repository name
+    try:
+        name = extract_name_from_git_repository(package_dir)
+    except UnsupportedFeature:
```
a-ovchinnikov, line 52: Since you are already refactoring this, what do you think about moving this try block into `extract_name_from_git_repository`?

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +31,49 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
 
 def _resolve_bundler_package(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
     """Process a request for a single bundler package."""
-    parse_lockfile(package_dir)
+    dependencies = parse_lockfile(package_dir)
+    _get_package_metadata(package_dir, dependencies)
     return []
+
+
+def _get_package_metadata(
```
a-ovchinnikov, line 39: Is this really metadata? It seems to be just package name and optional version.

cachi2/core/package_managers/bundler/main.py
slimreaper35: I was thinking about it. Just `extract_name_from_git_repository` and rejecting the package because of missing metadata might be two independent things. But I don't mind including it there.

cachi2/core/package_managers/bundler/main.py
slimreaper35: Yeah, the function name is ported from cachito. How about `_get_package_info` ?

slimreaper35: Well, we also use the term `metadata` in the [pip](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/pip.py#L276) module. Would something like `_get_package_info` be more appropriate?

tests/unit/package_managers/test_pip.py
```diff
@@ -120,7 +118,7 @@ def test_get_pip_metadata(
     setup_cfg.get_name.return_value = cfg_name
     setup_cfg.get_version.return_value = cfg_version
 
-    mock_get_repo_id.return_value = MOCK_REPO_ID
+    mock_get_repo_name.return_value = "bar-subpath"
```
slimreaper35, line 121: I know this is not ideal, but I made minimum changes just to pass the CI.

The function is awful. Actually, I was the one who added pyproject.toml to the mix. A good first step in refactoring the pip module would be splitting the "metadata" (pyproject.toml, setup.py, setup.cfg) into a separate file.

cachi2/core/package_managers/bundler/main.py
slimreaper35: EDIT: https://github.com/slimreaper35/cachi2/blob/main/docs/design/bundler.md#output-folder-structure

> The name of the directory must come from the Git URL, not the actual name of the gem, and the cloned folder must contain unpacked source code. Any other format will cause bundler to try to re-download the repository, causing the build to fail.

Oops, we will need something similar for git dependencies as well :face_exhaling:

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Yes, I'd say `info` sounds better. `metadata` is data about data, this looks like just regular data.

slimreaper35: Isn't it at the end just the SBOM component?

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +31,49 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+def _extract_metadata_from_gemspec_file(
+    path_dependencies: list[PathDependency],
+) -> Optional[tuple[str, str]]:
+    """
+    Extract metadata from the gemspec file.
+
+    If the gemspec file exists in the package directory and is explicitly defined in the Gemfile,
+    then the Gemfile.lock will contain a path dependency representing the gem in the package directory.
+    """
+    log.debug("Extracting metadata from gemspec file")
```
brunoapimentel, line 74: Nitpick: I think it is more useful to get a log stating that the info was not found in a gemspec file than just a generic log that traces that this function was called.

cachi2/core/package_managers/general.py
```diff
@@ -198,3 +201,14 @@ def extract_git_info(vcs_url: str) -> dict[str, Any]:
         "namespace": namespace,
         "repo": repo,
     }
+
+
+def get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
+    """Extract repositoy name from the git URL in the package directory."""
```
brunoapimentel, line 207: Nitpick: Do you think it's worth stating in the docstring that this function can raise a `UnsupportedFeature` exception? (alternatively, we could treat the exception within the function's scope an return `Optional[str]`)

cachi2/core/package_managers/bundler/main.py
brunoapimentel: I just wrote the same suggestion [elsewhere](https://github.com/containerbuildsystem/cachi2/pull/639#discussion_r1755596285).

cachi2/core/package_managers/bundler/main.py
brunoapimentel: It is, but I think where the data came from matters. I think it's the difference between a "main package/project/app" and its dependencies... in the end, all of them can be considered "packages". But I still think the distiction is useful. Some random ideas for renaming it:

- _get_root_package_info
- _get_app_info
- _get_main_package_info

The first suggestion does not have any precedent in Cachi2, the second is used in some variable names, the third is still what we mostly use.

Or just keep it as `_get_package_info` and clarify it in the docstring, as Alexey suggested.

cachi2/core/package_managers/general.py
a-ovchinnikov: Also s/repositoy/repository/

cachi2/core/package_managers/bundler/main.py
slimreaper35: gomod: `main_module`
npm: `get_main_package`
pip: `get_pip_metadata`
:exploding_head:

eskultety: > _get_app_info

Not a bad name :).

cachi2/core/package_managers/general.py
slimreaper35: I dropped that commit altogether; it would just require too much refactoring. Sorry for the noise.

cachi2/core/package_managers/bundler/main.py
slimreaper35: How about `_get_main_package_name_and_version` ?
It is more verbose, but it is clearer than _info_.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +34,61 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+    path_dependencies = [dep for dep in dependencies if isinstance(dep, PathDependency)]
+    name_and_version = _extract_name_and_version_from_gemspec_file(path_dependencies)
+    if name_and_version is not None:
+        return name_and_version
+
+    log.info("Failed to infer name and version from gemspec file")
+
+    # fallback to origin remote
+    try:
+        name = _get_repo_name_from_origin_remote(package_dir)
```
a-ovchinnikov, line 56: Could you please add a  very short comment here or in the extractor function explaining why or when it could raise `UnsupportedFeature`?

a-ovchinnikov (commented):
Mostly LGTM, a comment about needing a comment is my main request.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Awesome, thank you!

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +34,61 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+                "Please specify package name and version in a way that Cachi2 understands,\n"
+                "or make sure that the directory Cachi2 is processing is a git repository with\n"
+                "an 'origin' remote, in which case Cachi2 will infer the package name from the remote URL."
+            ),
+        )
+
+    return name, None
+
+
+def _extract_name_and_version_from_gemspec_file(
```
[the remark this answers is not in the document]
slimreaper35: I changed it to `_get_name_and_version_from_path_deps`

a-ovchinnikov (approved):

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +34,67 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+    if name_and_version is not None:
+        return name_and_version
+
+    log.info("Failed to infer name and version from gemspec file")
+
+    # fallback to origin remote
+    try:
+        name = _get_repo_name_from_origin_remote(package_dir)
+    # if the package directory is not a git repository
+    except UnsupportedFeature:
```
brunoapimentel, line 63: This is probably something that already happens regardless of the current PR here, but I was able to get a `InvalidGitRepositoryError` trying to process a folder that is not a repo. I think this is indeed a case for failure, but the error was not treated. Should we fix this as part of this PR?

slimreaper35: Yeah, good point; this is not handled in `get_repo_id.` I should have added a test for that.

slimreaper35: I added a new test for this case, that is currently being skipped. I don't know, but it might be considered an empty stub or dead code until it is fixed in `get_repo_id`.

brunoapimentel: So we'll fix it as a follow up?

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +34,67 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+    name_and_version = _get_name_and_version_from_path_deps(path_dependencies)
+    if name_and_version is not None:
+        return name_and_version
+
+    log.info("Failed to infer name and version from gemspec file")
+
+    # fallback to origin remote
+    try:
+        name = _get_repo_name_from_origin_remote(package_dir)
+    # if the package directory is not a git repository
```
brunoapimentel, line 62: I belive this scenario only happens when the directory is a git repo, but it doesn't have a `origin` remote, right?

```suggestion
    # if the git repository does not have an origin remote configured
```

cachi2/core/package_managers/bundler/main.py
slimreaper35: Done.

brunoapimentel (approved):

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +34,67 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+    dependencies: ParseResult,
+) -> tuple[str, Optional[str]]:
+    """
+    Get main package name and version.
+
+    The main package is the package that is being processed by cachi2.
+    Not any of its dependencies.
+    """
+    path_dependencies = [dep for dep in dependencies if isinstance(dep, PathDependency)]
+    name_and_version = _get_name_and_version_from_path_deps(path_dependencies)
```
eskultety, line 53: nitpick: you're looping over dependencies (yes, just a subset of it) twice without a reason. Just move the logic to the helper directly and pass it all dependencies.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +34,67 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+                "Please specify package name and version in a way that Cachi2 understands,\n"
+                "or make sure that the directory Cachi2 is processing is a git repository with\n"
+                "an 'origin' remote, in which case Cachi2 will infer the package name from the remote URL."
+            ),
+        )
+
+    return name, None
+
+
+def _get_name_and_version_from_path_deps(
```
eskultety, line 76: This is quite an unfortunate name, because one really has to dive deeper to understand where should the name come from. It should be called more straightforward, something like `_get_name_and_version_from_gemspec` because it is essentially what it does and where the information is coming from. It's not a blocker though.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +34,67 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+    path_dependencies: list[PathDependency],
+) -> Optional[tuple[str, str]]:
+    """
+    Extract name and version from path dependencies in the Gemfile.lock.
+
+    If the gemspec file exists in the package directory and is explicitly defined in the Gemfile,
+    then the Gemfile.lock will contain a path dependency representing the gem in the package directory.
```
eskultety, line 83: This is a bit incorrect. First, there is no reference to back this statement and I had to go digging what this is based on. What I've come across was this https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/ (not an official doc, but gave me some needed perspective). So the logic you have here assumes that `Gemfile.lock` is going to be present at all times and if it happens that `.gemspec` will be present as well, we'll use the info - which based on the article will be like lower units of % of cases because it is not recommended to check in one's `Gemfile.lock` when developing gems while it's equally unusual for apps to specify a gemspec when they're technically not a gem.

So, I'd expect the docstring to be improved a bit such that it mentions that what it's doing would happen in a fairly unlikely scenario. Additionally, mention our design doc as a basis for the implementation (we do have a note on gemspec path dependency)

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +34,67 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+        if dep.path == ".":
+            return dep.name, dep.version
+
+    return None
+
+
+def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
+    """Extract repository name from git origin remote in the package directory."""
+    repo_path = get_repo_id(package_dir.root).parsed_origin_url.path.removesuffix(".git")
+    repo_name = Path(repo_path).name
```
eskultety, line 95: nitpick: `Path(repo_path).stem`? And adjust the statement above accordingly.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +34,67 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+    package_subpath = package_dir.subpath_from_root
+
+    resolved_path = Path(repo_name).joinpath(package_subpath)
+    normalized_path = canonicalize_name(str(resolved_path).replace("/", "-"))
```
eskultety, line 99: Using `canonicalize_name` doesn't seem correct to me in this case. We're applying [PEP 503](https://peps.python.org/pep-0503/#normalized-names) onto something that not only doesn't conform to it, but also doesn't have a formally defined set of acceptable characters. I could only find https://guides.rubygems.org/name-your-gem/ which allows usage of both `-` and `_` and we should respect it in the SBOM as well to avoid any confusion on the user side.

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -0,0 +1,67 @@
...
+@pytest.mark.skip(reason="Not implemented in `cachi2.core.scm.get_repo_id`")
+def test_get_main_package_name_and_version_from_non_repo(
+    rooted_tmp_path: RootedPath,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    with pytest.raises(PackageRejected) as exc_info:
+        _get_main_package_name_and_version(package_dir=rooted_tmp_path, dependencies=[])
+
+    assert "Failed to infer name and version from gemspec file" in caplog.messages
+    assert "Failed to infer package name from origin remote" in exc_info.value.friendly_msg()
```
eskultety, line 67: I guess adding a skipped test like this will get easily forgotten the same way as if it weren't even there, meaning that it's unlikely anyone will remember to enable this test when the condition no longer applies, so I'd argue it's as good as dropping it for now.

eskultety (commented):
It's mostly about normalizing the package name, the rest are minor comments, but trivial to address if a new revision is going to be proposed.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Please note, that this function does not work with gemspec, its only argument is a list of path dependencies. Moreover, gemspec is not mentioned in the code at all, all the code does is some iteration over some path dependencies! A more self-documenting code with a gemspec mentioned in function name could look like this:
```python
def select_path_dep_for_gemspec(pds):
      return next(pd for pd in pds if pd.path == ".")
gemspec_dep = select_path_dep_for_gemspec(path_dependencies)
return gemspec_dep.name, gempec_dep.version if gemspec_dep is not None else None
```
If expressed like this the name could have mentioned gemspec, it would confuse future maintainers otherwise.

cachi2/core/package_managers/bundler/main.py
slimreaper35: I never seen a repo without `Gemfile.lock`. Isn't the purpose of that file to ensure consistency and that everyone has the same dependencies installed? Similarly, as it is, for example, for pipenv `Pipfile.lock`.
https://pipenv-fork.readthedocs.io/en/latest/basics.html#pipfile-lock-security-features

eskultety: For example: https://github.com/ruby-git/ruby-git

It's explained in the article I linked. We would face issues if people tried to build gems with cachi2. Then again, in that case `bundler` would most likely not be involved. That said, the fact remains that having both files is a very rare scenario.

cachi2/core/package_managers/bundler/main.py
eskultety: > Please note, that this function does not work with gemspec

I know that it doesn't, but it's pretty much the first thing it mentions in the docstring, so there either is a relation or there isn't.

> ```
> def _get_name_and_version_from_gemspec(path_dependencies):
>     def select_path_dep_for_gemspec(pds):
>          return next(pd for pd in pds if pd.path == ".")
>     gemspec_dep = select_path_dep_for_gemspec(path_dependencies)
>     return gemspec_dep.name, gempec_dep.version if gemspec_dep is not None else None
> ```
No objection to ^this. With a tiny nitpick on working over all deps, not just path deps, so that we don't loop it "twice".

tests/unit/package_managers/bundler/test_main.py
eskultety: > Would it help to link the issue I created yesterday to this code?

Probably yes. It will still be forgotten unless you do it yourself, however, there may be a chance that if you do
```
TODO: Enable this test once XYZ is fixed
```
someone might accidentally stumble upon the URL and see if anyone has already done something about that :) .

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Another option could be to slightly tweak the name: it is not 'from', it is 'for' a gemspec (which is now represented as a path dependency). Resulting name might raise some eyebrows, but reads as a succinct description of what is going on inside the function: `_get_name_and_version_for_gemspec_from(path_dependencies)`. In this case its definition matters much less.

cachi2/core/package_managers/bundler/main.py
slimreaper35: Okay, so simultaneously having `gemspec` and `Gemfile.lock` is unusual. Got it.
Now it makes sense. Docs updated.

tests/unit/package_managers/bundler/test_main.py
slimreaper35: > unless you do it yourself

That is a very likely scenario :)

cachi2/core/package_managers/bundler/main.py
slimreaper35: I renamed it to `get name_and_version_from_lockfile` for now. I think it is along the lines of the docstring and dependencies_2 as an argument. But I leave this unresolved.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +32,67 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
+        )
+
+    return name, None
+
+
+def _get_name_and_version_from_lockfile(dependencies: ParseResult) -> Optional[tuple[str, str]]:
+    """
+    Extract package name and version from dependencies in the Gemfile.lock.
+
+    Note that having a gemspec file is an edge case when the package is not an actual gem.
```
eskultety, line 77: nitpick: If we say we're extracting name and version from Gemfile.lock then it IMO deserves a small explanation:

_Gemfile.lock only contains name/version information for the package we're processing when a gemspec file is present in the repository. 
Note that ...._

eskultety (approved):

PR #640, opened 2024-09-06
gomod: Automatically detect and check the contents of a vendor dir

brunoapimentel:
Cachi2 currently has two flags to deal with gomod vendoring:
- gomod-vendor-check: verifies if the integrity of the vendored packages
- gomod-vendor: perfoms the vendoring instead of a regular prefetch

If there's a `vendor` folder checked-in the repo, and none of the above flags are set, the request fails.

These flags are confusing and generate unnecessary action from the user's side. We want to have a single consistent behavior: if a vendor folder is checked-in the repo, Cachi2 should check if its contents are integral and up-to-date. This is essentially setting `gomod-vendor-check` if a vendor folder is found, which makes this flag useless.

We also want to deprecate the `gomod-vendor` flag and leave the vendoring completely on the behalf of the user.


brunoapimentel:
This is a very rough draft to illustrate the wanted changes. Although I believe that completely making both flags a noop would be the ideal scenario, unfortunatelly I consider that this is breaking backwards compatibility with what the user would expect from `gomod-vendor` (that is, that vendoring is done as part of the prefetch).

I see two ways forward:
- Push the change forward and make both flags noop
- Keep the behavior of `gomod-vendor` and only make the `gomod-vendor-check` behavior automatic

Either way, I think we should mark both flags as deprecated and emit warnings if a user sets them.

eskultety (commented):
I only skimmed through the code quickly since it's a draft but I'm in strong favour of this PR (sure, I'm biased). Left a major comment to spark a potential discussion on the exact approach (ignore 1 vs both CLI flags).

cachi2/core/package_managers/gomod.py
```diff
@@ -1141,44 +1139,16 @@ def __exit__(
...
-            ),
-            solution=(
-                "Consider removing the vendor/ directory and letting Cachi2 download dependencies "
-                "instead.\n"
-                "If you do want to keep using vendoring, please pass one of the required flags."
-            ),
-            docs=VENDORING_DOC,
-        )
-
-    return False, False
```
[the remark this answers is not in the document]
brunoapimentel: Right, but there's the use case when a user does not have a vendor folder checked in, and used `gomod-vendor` to create one during the prefetch. If we pushed the changes as proposed, the vendor folder would no longer be created, and although it does not result in a failure, it can potentially break the build later on.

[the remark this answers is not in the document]
brunoapimentel: I'd argue this use case is very odd and should be scrapped, my question here is how we want to do it.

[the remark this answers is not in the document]
eskultety: Oh, missed that one, solely due to:

> I'd argue this use case is very odd and should be scrapped

This was always a user error. The fact we even supported it in the first place and allowed the source tree to be dirtied by us was a mistake, so this is an attempt to correct an inherently flawed behaviour <<<< I'd sell it like this in the release notes. After all it really is odd to let some other tool do the work for you, since, without our tool would not be able to vendor their project, that never was a legitimate use case or mission of this project, we only fetch dependencies, we must never step in the way how user wishes to distribute their project.

a-ovchinnikov (approved):
The change mostly removes code, related tests and test data, the part that does not remove code makes previously user-selectable behavior default and automated. LGTM.

cachi2/core/package_managers/gomod.py
```diff
@@ -1546,28 +1516,23 @@ def parse_module_line(line: str) -> ParsedModule:
     return (module for module, has_packages in zip(modules, module_has_packages) if has_packages)
 
 
-def _vendor_deps(
+def _check_and_parse_vendored_content(
```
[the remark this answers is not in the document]
brunoapimentel: I dropped the change in the end. It's probably good enough as is.

cachi2/core/package_managers/gomod.py
```diff
@@ -619,19 +619,18 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
             components.extend(module.to_component() for module in modules)
             components.extend(package.to_component() for package in packages)
 
-        if "gomod-vendor-check" not in request.flags and "gomod-vendor" not in request.flags:
```
eskultety, line 619: If this check was enough previously, why can't we just do:
```
if app_dir.join_within_root("vendor").path.is_dir():
   ...
```

to retain the same logic. That would work, wouldn't it? Instead we're pushing it down to `_resolve_gomod` and fiddling with the GOMODCACHE setting which to me is unnecessary if we simply don't copy the temporary cache to the output and just leave it be collected by the tempdir context manager.

cachi2/core/package_managers/gomod.py
```diff
@@ -860,12 +859,22 @@ def _resolve_gomod(
...
 
+    should_vendor = app_dir.join_within_root("vendor").path.is_dir()
+
+    if should_vendor:
+        # Even though we do not perform a "go mod download" when vendoring is detected, some
+        # go commands still download dependencies as a side effect. Since we don't want those
+        # copied to the output dir, we need to set the GOMODCACHE to a different directory.
+        gomod_cache = f"{tmp_dir}/vendor-cache"
+    else:
+        gomod_cache = f"{tmp_dir}/pkg/mod"
```
eskultety, line 870: If I'm not mistaken with my comment above, this won't be necessary and the diff is going to be smaller.

cachi2/core/package_managers/gomod.py
```diff
@@ -1151,46 +1149,6 @@ def __exit__(
...
-    :raise PackageRejected: if the vendor dir is present, the flags are not used and we are strict
-    """
-    vendor = app_dir.join_within_root("vendor").path
-
-    if "gomod-vendor-check" in flags:
-        return True, not vendor.exists()
-    if "gomod-vendor" in flags:
-        return True, True
-
-    if strict and vendor.is_dir():
```
eskultety, line 1177: `strict` is determined by the `gomod_strict_vendor` config option which defaults to True. Deleting this function means we're not taking any action upon setting the variable to False. Now, having the vendor directory and not wanting to vendor it is a weird use case, i.e. `gomod_strict_vendor=False` (we've talked about it), however, you'll still need to deprecate the usage of the variable with a warning in the code (the way our config currently works it's going to be interesting to detect if the user set it somehow) as well as a note in the README.

tests/unit/package_managers/test_gomod.py
```diff
@@ -339,12 +336,7 @@ def test_resolve_gomod_vendor_dependencies(
     mock_version_resolver.get_golang_version.return_value = "v0.1.0"
     mock_go_release.return_value = "go0.1.0"
     mock_get_gomod_version.return_value = ("0.1.1", "0.1.2")
-
-    flags: list[Flag] = ["gomod-vendor"]
-    if force_gomod_tidy:
-        flags.append("force-gomod-tidy")
```
eskultety, line 345: Why are we dropping the tidy flag? We're running vendor automatically, so we don't need the vendor flag, okay, but why are we dropping tidy if the CLI flag were used? The rest of the fetch tests kept the flag untouched. What am I missing?

cachi2/core/models/input.py
```diff
@@ -4,7 +4,6 @@
     Annotated,
     Any,
     Callable,
-    ClassVar,
```
eskultety, line 7: isort is unhappy about this patch.

cachi2/core/package_managers/gomod.py
brunoapimentel: The problem is that `app_dir` (or `main_module_dir` as it is called in this scope) only exists in the context of each `subpath` being processed. We can have a subpath that contains a `vendor` dir, and another one that doesn't ><

This check used to be request-global, and now we need to do it for each input package.

brunoapimentel: We still want to copy the contents of `pkg/mod/cache/download` to the global output dir, considering that it will be populated by all input packages/subpaths that did not have a `vendor` dir.

cachi2/core/package_managers/gomod.py
brunoapimentel: Good catch. I added a deprecation warning when this config option is set, and also updated the corresponding documentation.

cachi2/core/models/input.py
eskultety: > And I'm unhappy about isort

Understandable :)

cachi2/core/package_managers/gomod.py
eskultety: Thanks for the clarification.

cachi2/interface/cli.py
```diff
@@ -165,18 +165,12 @@ def fetch_deps(
     gomod_vendor: bool = typer.Option(
         False,
         "--gomod-vendor",
-        help=(
-            "Fetch go deps via 'go mod vendor' rather than 'go mod download'. If you "
-            "have a vendor/ dir, one of --gomod-vendor/--gomod-vendor-check is required."
-        ),
+        help=("Deprecated flag. Will have no effect, and its use is discouraged."),
```
eskultety, line 168: I think we want to retain the original meaning of the flag as well, but then make 'deprecated' in capitals.

cachi2/core/config.py
```diff
@@ -22,6 +25,18 @@ class Config(BaseModel, extra="forbid"):
     requests_timeout: int = 300
     concurrency_limit: int = 5
 
+    @model_validator(mode="before")
+    @classmethod
+    def _print_deprecation_warning(cls, data: Any) -> Any:
+        if "gomod_strict_vendor" in data:
+            log.warning(
+                "The `gomod_strict_vendor` config option is deprecated and will be removed in "
+                "future versions. Notice that it no longer has any effect when set, Cachi2 will "
```
eskultety, line 34: s/Notice/Note

README.md
```diff
@@ -114,9 +114,9 @@ are names of package managers. The values are dictionaries where the keys
 are default environment variables to set for that package manager and the
 values are the environment variable values.
 * `gomod_download_max_tries` - a maximum number of attempts for retrying go commands.
-* `gomod_strict_vendor` - the bool to disable/enable the strict vendor mode. For a repo that has gomod dependencies,
-if the `vendor` directory exists and this config option is set to `True`, one of the
-[vendoring flags](gomod.md#vendoring) must be used.
+* `gomod_strict_vendor` - (deprecated) was used to disable/enable the strict vendor mode. Now, the strict mode is
+enforced, which means that if the `vendor` directory exists, its contents must be up-to-date with what `go mod vendor`
```
eskultety, line 118: nitpick: strict mode defaulted to True, so there really isn't much a change, but the current wording kinda implies there is, if you just add (Default was True) somewhere in the first sentence, then it's all fine.

cachi2/core/config.py
```diff
@@ -22,6 +25,18 @@ class Config(BaseModel, extra="forbid"):
     requests_timeout: int = 300
     concurrency_limit: int = 5
 
+    @model_validator(mode="before")
```
eskultety, line 28: I wondered why this couldn't have been a field validator since that too supports `mode="before"`. Here's the answer: https://docs.pydantic.dev/latest/concepts/validators/#validation-of-default-values - the handling would be weird and even if it technically were `"before"` you'd already get the truthy value in the method, so there goes...

eskultety (approved):
Nitpicks mostly, so giving ACK.

cachi2/interface/cli.py
brunoapimentel: Ok, I changed it. LMK if you are ok with it.

cachi2/core/config.py
brunoapimentel: With the changes, the phrase no longer exists.

cachi2/core/package_managers/gomod.py
```diff
@@ -888,17 +888,15 @@ def _resolve_gomod(
 
     # Vendor dependencies if the gomod-vendor flag is set
     flags = request.flags
-    should_vendor, can_make_changes = _should_vendor_deps(
```
eskultety, line 891: **Commit message**:
`s/being kept to/kept`

cachi2/core/config.py
eskultety: I can still see it though :) .

eskultety (approved):

cachi2/core/config.py
brunoapimentel: Right, this was on the deprecation warnings. Now it is fixed.

PR #648, opened 2024-09-10
bundler: Adding Gem downloader

a-ovchinnikov:
This change extends GemDependencies with a method
that will download them from their source.

### DIFF
--- a/cachi2/core/package_managers/bundler/parser.py
+++ b/cachi2/core/package_managers/bundler/parser.py
@@ -1,12 +1,14 @@
 import json
 import logging
 import subprocess
+from functools import cached_property
 from pathlib import Path
 from typing import Annotated, Optional, Union
 
 import pydantic
 
 from cachi2.core.errors import PackageManagerError, PackageRejected, UnexpectedFormat
+from cachi2.core.package_managers.general import download_binary_file
 from cachi2.core.rooted_path import PathOutsideRoot, RootedPath
 from cachi2.core.utils import run_cmd
 
@@ -38,18 +40,31 @@ class _GemMetadata(pydantic.BaseModel):
     name: str
     version: str
 
+    def download_to(self, fs_location: RootedPath) -> None:
+        return None
+
 
 class GemDependency(_GemMetadata):
     """
     Represents a gem dependency.
 
     Attributes:
-        source:     The source URL of the gem.
+        source:     The source URL of the gem as stated in 'remote' field from Gemfile.lock.
         checksum:   The checksum of the gem.
     """
 
     source: str
     checksum: Optional[str] = None
+
+    @cached_property
+    def remote_location(self) -> str:
+        """Return remote location to download this gem from."""
+        return f"{self.source}/gems/{self.name}-{self.version}.gem"
+
+    def download_to(self, fs_location: RootedPath) -> None:
+        """Download represented gem to specified file system location."""
+        fs_location = fs_location.join_within_root(Path(f"{self.name}-{self.version}.gem"))
+        download_binary_file(self.remote_location, fs_location)
 
 
 class GitDependency(_GemMetadata):
--- a/tests/unit/package_managers/bundler/test_parser.py
+++ b/tests/unit/package_managers/bundler/test_parser.py
@@ -1,6 +1,7 @@
 import json
 import subprocess
 from copy import deepcopy
+from pathlib import Path
 from typing import Any
 from unittest import mock
 
@@ -180,3 +181,25 @@ def test_parse_gemlock_empty(
 
     assert f"Package {rooted_tmp_path.path.name} is bundled with version 2.5.10" in caplog.messages
     assert result == []
+
+
+@pytest.mark.parametrize(
+    "source",
+    [
+        "https://rubygems.org",
+        "https://dedicatedprivategemrepo.com",
+    ],
+)
+@mock.patch("cachi2.core.package_managers.bundler.parser.download_binary_file")
+def test_dependencies_could_be_downloaded(
+    mock_downloader: mock.MagicMock,
+    source: str,
+) -> None:
+    base_destination = RootedPath("/tmp/foo")
+    dependency = GemDependency(name="foo", version="0.0.2", source=source)
+    expected_source_url = f"{source}/gems/foo-0.0.2.gem"
+    expected_destination = base_destination.join_within_root(Path("foo-0.0.2.gem"))
+
+    dependency.download_to(base_destination)
+
+    mock_downloader.assert_called_once_with(expected_source_url, expected_destination)
--- a/cachi2/core/package_managers/general.py
+++ b/cachi2/core/package_managers/general.py
@@ -3,7 +3,6 @@ import asyncio
 import logging
 import types
 from os import PathLike
-from pathlib import Path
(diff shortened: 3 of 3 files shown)


tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -180,3 +181,25 @@ def test_parse_gemlock_empty(
...
+        "https://rubygems.org",
+        "https://dedicatedprivategemrepo.com",
+    ],
+)
+@mock.patch("cachi2.core.package_managers.bundler.parser.download_binary_file")
+def test_dependnecies_could_be_downloaded(
+    mock_downloader: mock.MagicMock,
+    source: str,
+) -> None:
+    base_destination = Path("/tmp/foo")
```
slimreaper35, line 198: You can use `rooted_tmp_path` fixture instead.

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -180,3 +181,25 @@ def test_parse_gemlock_empty(
...
+    dependency = GemDependency(version="0.0.2", source=source, name="foo")
```
slimreaper35, line 199: nitpick: let's reorder the params - name, version, source :smile:

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -38,19 +40,32 @@ class _GemMetadata(pydantic.BaseModel):
...
+        return None
+
 
 class GemDependency(_GemMetadata):
     """
     Represents a gem dependency.
 
     Attributes:
-        source:     The source URL of the gem.
+        source:     The source URL of the gem as stated in 'source' field from Gemfile.lock.
```
slimreaper35, line 52: Actually, it is 'remote'. The problem is that the bundler uses 'remote' also for git and path dependencies.

tests/unit/package_managers/bundler/test_parser.py
a-ovchinnikov: I could, but this test does not touch file system at all, it only verifies the flow. There is no need to use a real path, anything path-like would do.

cachi2/core/package_managers/bundler/parser.py
a-ovchinnikov: Huh, it is 'source' in Gemfile, but 'remote' in Gemfile.lock. Will update the comment.

tests/unit/package_managers/bundler/test_parser.py
a-ovchinnikov: Done.

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -180,3 +181,25 @@ def test_parse_gemlock_empty(
...
+
+@pytest.mark.parametrize(
+    "source",
+    [
+        "https://rubygems.org",
+        "https://dedicatedprivategemrepo.com",
+    ],
+)
+@mock.patch("cachi2.core.package_managers.bundler.parser.download_binary_file")
+def test_dependnecies_could_be_downloaded(
```
brunoapimentel, line 194: Typo on dependencies.

a-ovchinnikov: Thank you!

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -38,19 +40,32 @@ class _GemMetadata(pydantic.BaseModel):
     name: str
     version: str
 
+    def download_to(self, fs_location: Path) -> None:
+        return None
+
```
slimreaper35, line 45: I would add this to a separate commit or even drop it (since it currently does nothing) until we start adding all functions together.

a-ovchinnikov: This is the default implementation, I _think_ it will remain unchanged for PathDependency. I can split it out though if this helps.

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -180,3 +181,25 @@ def test_parse_gemlock_empty(
...
+@pytest.mark.parametrize(
+    "source",
+    [
+        "https://rubygems.org",
+        "https://dedicatedprivategemrepo.com",
+    ],
+)
+@mock.patch("cachi2.core.package_managers.bundler.parser.download_binary_file")
+def test_dependencies_could_be_downloaded(
+    mock_downloader: mock.MagicMock,
```
eskultety, line 195: Out of curiosity, what dunder methods do you need to type this as `MagicMock` instead of a regular `Mock`?

eskultety (approved):

cachi2/core/package_managers/bundler/parser.py
a-ovchinnikov: I prefer keeping it simple unless we really really need extra complexity (e.g. when it makes things simpler someplace else). Moreover, this is not an interface spec, this is actual default fallback that will be executed. I would have returned `NotImplemented` or raised if I intended it to be an interface. This method exists solely for side-effects, and by default there is no side effect.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -1,12 +1,14 @@
 import json
 import logging
```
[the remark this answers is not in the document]
a-ovchinnikov: Done.

tests/unit/package_managers/bundler/test_parser.py
brunoapimentel: I'd take a `MagicMock` anytime over an ordinary `Mock` :mage_man: 

In all seriousness, I think we tend to simply default to `Mock` in Cachi2, but probably not a big deal to keep it as `MagicMock`

tests/unit/package_managers/bundler/test_parser.py
brunoapimentel: Which makes me wonder if we should try to keep the unit tests leaner.  We have already set a lot of precedent by having many tests that actually reach out for the file system, but I'm not sure if the added execution time and complexity are worth it (for example, [this fixture](https://github.com/containerbuildsystem/cachi2/blob/6a483062359b0265cd208861f74a6f35b3e37f07/tests/unit/package_managers/conftest.py#L188) in another bundler test that generates a full git repo). 

I probably still tend to see unit tests as very isolated, so I might be biased here.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -38,19 +40,32 @@ class _GemMetadata(pydantic.BaseModel):
...
 
     source: str
     checksum: Optional[str] = None
 
+    @cached_property
+    def remote_location(self) -> str:
+        """Return remote location to download this gem from."""
+        return f"{self.source}/gems/{self.name}-{self.version}.gem"
+
+    def download_to(self, fs_location: Path) -> None:
```
brunoapimentel, line 64: I believe we default to using `RootedPath` to make sure that all filepaths will not resolve outside of the repository root.

a-ovchinnikov: Good point, thank you.

brunoapimentel (approved):

PR #650, opened 2024-09-11
bundler: download git dependencies

slimreaper35:
See the commit msg for details.

- [x] Docs links in the code are still valid (if docs were updated)

### DIFF
--- a/cachi2/core/package_managers/bundler/parser.py
+++ b/cachi2/core/package_managers/bundler/parser.py
@@ -4,8 +4,10 @@ import subprocess
 from functools import cached_property
 from pathlib import Path
 from typing import Annotated, Optional, Union
+from urllib.parse import urlparse
 
 import pydantic
+from git import Repo
 
 from cachi2.core.errors import PackageManagerError, PackageRejected, UnexpectedFormat
 from cachi2.core.package_managers.general import download_binary_file
@@ -78,6 +80,32 @@ class GitDependency(_GemMetadata):
 
     url: AcceptedUrl
     ref: AcceptedGitRef
+
+    @cached_property
+    def repo_name(self) -> str:
+        """Extract the repository name from the URL."""
+        parse_result = urlparse(str(self.url))
+        return Path(parse_result.path).stem
+
+    def download_to(self, output_dir: RootedPath) -> None:
+        """Download git repository to the output directory with a specific name."""
+        short_ref_length = 12
+        short_ref = self.ref[:short_ref_length]
+
+        git_repo_path = output_dir.join_within_root(f"{self.repo_name}-{short_ref}")
+        if git_repo_path.path.exists():
+            log.info("Skipping existing git repository %s", self.url)
+            return
+
+        git_repo_path.path.mkdir(parents=True)
+
+        log.info("Cloning git repository %s", self.url)
+        repo = Repo.clone_from(
+            url=str(self.url),
+            to_path=git_repo_path.path,
+            env={"GIT_TERMINAL_PROMPT": "0"},
+        )
+        repo.git.checkout(self.ref)
 
 
 class PathDependency(_GemMetadata):
--- a/tests/unit/package_managers/bundler/test_parser.py
+++ b/tests/unit/package_managers/bundler/test_parser.py
@@ -18,6 +18,7 @@ from cachi2.core.package_managers.bundler.parser import (
     parse_lockfile,
 )
 from cachi2.core.rooted_path import RootedPath
+from tests.common_utils import GIT_REF
 
 
 @pytest.fixture
@@ -89,7 +90,7 @@ def test_parse_lockfile_invalid_format(
             {
                 "type": "git",
                 "url": "github",
-                "ref": "f" * 40,
+                "ref": GIT_REF,
             }
         )
     elif error == "LOCKFILE_INVALID_URL_SCHEME":
@@ -97,7 +98,7 @@ def test_parse_lockfile_invalid_format(
             {
                 "type": "git",
                 "url": "http://github.com/3scale/json-schema.git",
-                "ref": "f" * 40,
+                "ref": GIT_REF,
             }
         )
     elif error == "LOCKFILE_INVALID_REVISION":
@@ -136,7 +137,7 @@ def test_parse_gemlock(
         {
             "type": "git",
             "url": "https://github.com/3scale/json-schema.git",
-            "ref": "f" * 40,
+            "ref": GIT_REF,
             **base_dep,
         },
         {
@@ -159,7 +160,7 @@ def test_parse_gemlock(
             name="example",
             version="0.1.0",
             url="https://github.com/3scale/json-schema.git",
-            ref="f" * 40,
+            ref=GIT_REF,
         ),
         PathDependency(name="example", version="0.1.0", path="subpath/pathgem"),
         GemDependency(name="example", version="0.1.0", source="https://rubygems.org/"),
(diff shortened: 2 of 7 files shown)


cachi2/core/package_managers/bundler/main.py
```diff
@@ -24,5 +24,12 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
 
 def _resolve_bundler_package(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
     """Process a request for a single bundler package."""
-    parse_lockfile(package_dir)
+    dependencies = parse_lockfile(package_dir)
+
+    # TODO: remove this hunk
+    deps_dir = output_dir.join_within_root("deps", "bundler")
+    for dep in dependencies:
+        if isinstance(dep, GitDependency):
```
a-ovchinnikov, line 32: If you add a do-nothing `download_to` to _GemMetadata base class like in [PR-648](https://github.com/containerbuildsystem/cachi2/pull/648) (parser.py, L45) you would not need to check for type here.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -64,6 +66,33 @@ class GitDependency(_GemMetadata):
...
+        path = Path(parse_result.path.removesuffix(".git"))
+        return path.name
+
+    def download_to(self, output_dir: RootedPath) -> None:
+        """Download git repository to the output directory with a specific name."""
+        short_ref_length = 12
+        short_ref = self.ref[:short_ref_length]
+
+        git_repo_path = output_dir.join_within_root(f"{self.repo_name}-{short_ref}")
+        if git_repo_path.path.exists():
```
a-ovchinnikov, line 82: Does this logic belong here? Won't components be eventually deduplicated elsewhere?

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -180,3 +180,32 @@ def test_parse_gemlock_empty(
...
+        ref="f" * 40,
+    )
+
+    dep_path = rooted_tmp_path.join_within_root(f"{dep.repo_name}-{dep.ref[:12]}").path
+
+    dep.download_to(output_dir=rooted_tmp_path)
+    assert f"Downloading git repository {dep.url}" in caplog.messages
+
+    dep.download_to(output_dir=rooted_tmp_path)
+    assert f"Skipping existing git repository {dep.url}" in caplog.messages
```
a-ovchinnikov, line 204: I would recommend splitting this test into two: you are testing two scenarios here, one that something could be downloaded, another that duplicates are skipped. When such compound tests fail it is harder to pinpoint underlying issue because multiple things are tested simultaneously. Compound tests make sense in integration or e2e where tests are expensive, for UTs I think clarity is more important.

a-ovchinnikov (approved):
LGTM with some minor suggestions.

cachi2/core/package_managers/bundler/main.py
slimreaper35: I will drop this...

cachi2/core/package_managers/bundler/parser.py
slimreaper35: It may happen that the Gemfile.lock contains various duplicate or transitive dependencies, and then `mkdir(parents=True)` will raise `FileExistsError`.

@brunoapimentel sent me this repo https://github.com/brunoapimentel/chatwoot/tree/hermetic-build-test
as an example while "collaborating" on the design document.

a-ovchinnikov: Thank you for confirming this!

a-ovchinnikov (approved):

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -79,6 +81,33 @@ class GitDependency(_GemMetadata):
     url: AcceptedUrl
     ref: AcceptedGitRef
 
+    @property
```
a-ovchinnikov, line 84: A super non-significant, non-blocking nitpick, please apply only if you end up doing another revision: this won't change, so could be a cached_property.

slimreaper35: no worries, it is just a matter of seconds :smile:

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -79,6 +81,33 @@ class GitDependency(_GemMetadata):
     url: AcceptedUrl
     ref: AcceptedGitRef
 
+    @cached_property
+    def repo_name(self) -> str:
+        """Extract the repository name from the URL."""
+        parse_result = urlparse(str(self.url))
+        path = Path(parse_result.path.removesuffix(".git"))
+        return path.name
```
eskultety, line 89: How about just `return Path(parse_result.path).stem`

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -4,8 +4,10 @@
 from functools import cached_property
 from pathlib import Path
 from typing import Annotated, Optional, Union
+from urllib.parse import urlparse
```
eskultety, line 7: **Commit message:**

the search may return different results in the future (e.g. they rename the function), so I think you just want to point it to: https://github.com/rubygems/rubygems/blob/0bcc770cc623d2da98b15389cf3f3aeaf5efae6f/bundler/lib/bundler/source/git.rb#L347
OR
https://github.com/rubygems/rubygems/blob/0bcc770cc623d2da98b15389cf3f3aeaf5efae6f/bundler/lib/bundler/plugin/api/source.rb#L112

(both deal with source URL handling and I didn't want to spend time digging into how they're related, so the former should be enough of a reference for our purposes)

Also,
 > The name is constructed from name of the
> git repository (not actual name of the gem)

this should also be backed by a reference, again, you probably want https://github.com/rubygems/rubygems/blob/0bcc770cc623d2da98b15389cf3f3aeaf5efae6f/bundler/lib/bundler/source/git.rb#L339 but one has to first parse the regex they use in their head when reading the commit, so going back again to https://github.com/rubygems/rubygems/blob/0bcc770cc623d2da98b15389cf3f3aeaf5efae6f/bundler/lib/bundler/plugin/api/source.rb#L110 which is aligned with your implementation. I don't care which one you go with as long as we reference both how base name and the ref are computed.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -79,6 +81,33 @@ class GitDependency(_GemMetadata):
...
+        short_ref = self.ref[:short_ref_length]
+
+        git_repo_path = output_dir.join_within_root(f"{self.repo_name}-{short_ref}")
+        if git_repo_path.path.exists():
+            log.info("Skipping existing git repository %s", self.url)
+            return
+
+        git_repo_path.path.mkdir(parents=True)
+
+        log.info("Downloading git repository %s", self.url)
```
eskultety, line 103: `s/Downloading/Cloning` maybe?

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -203,3 +203,54 @@ def test_dependencies_could_be_downloaded(
...
+def test_download_git_dependency_works(
+    mock_git_clone: mock.Mock,
+    rooted_tmp_path: RootedPath,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    dep = GitDependency(
+        name="example",
+        version="0.1.0",
+        url="https://github.com/user/repo.git",
+        ref="f" * 40,
```
eskultety, line 218: I've seen this elsewhere already, don't we want a constant for this? e.g. `TEST_REF="f"*40`

eskultety (commented):
minor comments, LGTM.

slimreaper35:
I feel that I should have swapped the commits.

PR #651, opened 2024-09-11
Fix extra options model for rpm

eskultety:
This PR is a direct result of #580, more specifically to tackle:
- https://github.com/containerbuildsystem/cachi2/blob/848c8075caa6cde72a2b9338793786b14f1f2cae/cachi2/core/models/input.py#L109
- https://github.com/containerbuildsystem/cachi2/blob/848c8075caa6cde72a2b9338793786b14f1f2cae/cachi2/core/models/input.py#L214

The goal of this PR is make additions of new options models easier than it is today compared to what is needed in #580.

### DIFF
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -6,7 +6,6 @@ from typing import (
     Callable,
     ClassVar,
     Dict,
-    List,
     Literal,
     Optional,
     TypeVar,
@@ -75,60 +74,6 @@ class _PackageInputBase(pydantic.BaseModel, extra="forbid"):
     @pydantic.field_validator("path")
     def _path_is_relative(cls, path: Path) -> Path:
         return check_sane_relpath(path)
-
-
-class _DNFOptions(pydantic.BaseModel, extra="forbid"):
-    """DNF options model.
-
-    DNF options can be provided via 2 'streams':
-        1) global /etc/dnf/dnf.conf OR
-        2) /etc/yum.repos.d/.repo files
-
-    Config options are specified via INI format based on sections. There are 2 types of sections:
-        1) global 'main' - either global repo options or DNF control-only options
-            - NOTE: there must always ever be a single "main" section
-
-        2) <repoid> sections - options tied specifically to a given defined repo
-
-    [1] https://man7.org/linux/man-pages/man5/dnf.conf.5.html
-    """
-
-    # Don't model all known DNF options for validation purposes - it's user's responsibility!
-    dnf: Dict[Union[Literal["main"], str], Dict[str, Any]]
-
-    @pydantic.model_validator(mode="before")
-    def _validate_dnf_options(cls, data: Any, info: pydantic.ValidationInfo) -> Optional[Dict]:
-        """Fail if the user passes unexpected configuration options namespace."""
-
-        def _raise_unexpected_type(repr_: str, *prefixes: str) -> None:
-            loc = ".".join(prefixes + (repr_,))
-            raise ValueError(f"Unexpected data type for '{loc}' in input JSON: expected 'dict'")
-
-        prefixes: List[str] = ["options"]
-
-        if not data:
-            return None
-
-        if not isinstance(data, dict):
-            _raise_unexpected_type(data, *prefixes)
-
-        if "dnf" not in data:
-            raise ValueError(f"Missing required namespace attribute in '{data}': 'dnf'")
-
-        if diff := set(cls.model_fields) - set(data.keys()):
-            raise ValueError(f"Extra attributes passed in '{data}': {diff}")
-
-        prefixes.append("dnf")
-        options_scope = data["dnf"]
-        if not isinstance(options_scope, dict):
-            _raise_unexpected_type(options_scope, *prefixes)
-
-        for repo, repo_options in options_scope.items():
-            prefixes.append(repo)
-            if not isinstance(repo_options, dict):
-                _raise_unexpected_type(repo_options, *prefixes)
-
-        return data
 
 
 class BundlerPackageInput(_PackageInputBase):
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -172,11 +172,62 @@ class PipPackageInput(_PackageInputBase):
         return paths
 
 
+class _ExtraOptions(pydantic.BaseModel, extra="forbid"):
+    """Global package manager extra options model.
+
+    This model takes care of carrying and parsing various kind of extra options that need to be
+    passed through to CLI commands/services we interact with underneath to tweak their behaviour
+    rather than our own. Each option set is namespaced by the corresponding tool/service it is
+    related to.
+
+    TODO: Enable this globally for all pkg managers not just the RpmPackageInput model.
+    """
+
+    dnf: Optional[Dict[Union[Literal["main"], str], Dict[str, Any]]] = None
+
+    @pydantic.model_validator(mode="before")
+    def _validate_dnf_options(cls, data: Any, info: pydantic.ValidationInfo) -> Any:
+        """DNF options model.
+
+        DNF options can be provided via 2 'streams':
+            1) global /etc/dnf/dnf.conf OR
+            2) /etc/yum.repos.d/.repo files
+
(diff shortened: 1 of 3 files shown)


eskultety:
FWIW I tested this implementation manually in 2 ways:

1. against inputs currently accepted by main:
```
$ cachi2 fetch-deps --dev-package-managers '{"type":"rpm", "path":".", "options":{"dnf": {"ubi-9-baseos-rpms":{"sslverify":"false","gpgcheck":"false"}}}}'
```
which yields the following build config:
```
$ cat cachi2-output/.build-config.json 
{
  "environment_variables": [],
  "project_files": [],
  "options": {
    "rpm": {
      "dnf": {
        "ubi-9-baseos-rpms": {
          "sslverify": "false",
          "gpgcheck": "false"
        }
      }
    }
  }
}
```
2. by adding extra future options (inspired by #580):
```
$ git diff
diff --git a/cachi2/core/models/input.py b/cachi2/core/models/input.py
index ec6dffcc..a686522c 100644
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -117,8 +117,23 @@ class PipPackageInput(_PackageInputBase):
         return paths
 
 
+class _SSLOptions(pydantic.BaseModel, extra="forbid"):
+    client_cert: Optional[str] = None
+    client_key: Optional[str] = None
+    ca: Optional[str] = None
+    foo: Optional[str] = None
+
+    def _validate_cert_and_key_exist(self) -> Self:
+        if any((self.client_key, self.client_cert)) and not all((self.client_key, self.client_cert)):
+            raise ValueError("cert and key need to be used together")
+        return self
+
+
 class _ExtraOptions(pydantic.BaseModel, extra="forbid"):
     dnf: Optional[Dict[Union[Literal["main"], str], Dict[str, Any]]] = None
+    ssl: Optional[_SSLOptions] = None
 
     @pydantic.model_validator(mode="before")
     def _validate_dnf_options(cls, data: Any, info: pydantic.ValidationInfo) -> Any:
```
and then trying the following invocations:
```
# TEST JUST SSL OPTIONS (client_cert implies client_key otherwise -> failure)
$ cachi2 --log-level=debug fetch-deps --dev-package-managers '{"type":"rpm", "path":".", "options":{"ssl": {"client_cert":"foo","client_key":""}}}'
2024-09-11 12:32:12,181 ERROR InvalidInput: 1 validation error for user input\npackages -> 0 -> rpm -> options -> ssl\n  Value error, cert and key need to be used together
packages -> 0 -> rpm -> options -> ssl
  Value error, cert and key need to be used together

# TEST COMBINATION OF EXTRA OPTIONS WITH DNF (incorrect DNF repo options type)
$ cachi2 --log-level=debug fetch-deps --dev-package-managers '{"type":"rpm", "path":".", "options":{"ssl": {"client_cert":"foo","client_key":"asfd"}, "dnf":{"asdf":[]}}}'
2024-09-11 12:33:20,720 ERROR InvalidInput: 1 validation error for user input\npackages -> 0 -> rpm -> options\n  Value error, Unexpected data type for 'options.dnf.asdf.[]' in input JSON: expected 'dict'
packages -> 0 -> rpm -> options
  Value error, Unexpected data type for 'options.dnf.asdf.[]' in input JSON: expected 'dict'

# TEST VALID COMBO
$ cachi2 --log-level=debug fetch-deps --dev-package-managers '{"type":"rpm", "path":".", "options":{"ssl": {"client_cert":"foo","client_key":"bar"}, "dnf":{"repoid":{"opt1":"true"}}}}'
```
with the last one yielding the expected build config:
```
$ cat cachi2-output/.build-config.json 
{
  "environment_variables": [],
  "project_files": [],
  "options": {
    "rpm": {
      "dnf": {
        "repoid": {
          "opt1": "true"
        }
      },
      "ssl": {
        "client_cert": "foo",
        "client_key": "bar",
        "ca": null,
        "foo": null
      }
    }
  }
```

cachi2/core/models/input.py
```diff
@@ -172,11 +117,52 @@ def _requirements_file_path_is_relative(cls, paths: list[Path]) -> list[Path]:
         return paths
 
 
+class _ExtraOptions(pydantic.BaseModel, extra="forbid"):
```
brunoapimentel, line 120: Just to clarify: is this model intended to be used only within `RpmPackageInput`, or is it the intention to have it as possibility for other package managers to adopt if needed?

If I think of the latter, I imagine one could use `{type: pip, options: {dnf: {}}}`, and it would be valid, even though it's a noop.

eskultety: > If I think of the latter, I imagine one could use {type: pip, options: {dnf: {}}}, and it would be valid, even though it's a noop.

True. Although this is IMO not the time to deal with that, because we're not enabling it globally. Once we decide to do so, we could e.g. do:

```
class _PackageInputBase(pydantic.BaseModel, extra="forbid"):
     """Common input attributes accepted for all package types."""
 
     type: PackageManagerType
     path: Path = Path(".")
+    options: Optional[_ExtraOptions] = None
+    _option_namespaces: list[str] = []
 
     @pydantic.field_validator("path")
     def _path_is_relative(cls, path: Path) -> Path:
         return check_sane_relpath(path)
 
+    def _validate_pkgmgr_type_option_pairings(self) -> Any:
+        breakpoint()
+        if self.options is None:
+            return self
+
+        if (diff := self.options.model_fields_set - set(self._option_namespaces)):
+            raise ValueError(
+                f"Unsupported options for type:'{self.type}': {diff} | supported options: '{self._option_namespaces}'")
+        return self
+
```

AND
```
 class RpmPackageInput(_PackageInputBase):
 
     type: Literal["rpm"]
     options: Optional[_ExtraOptions] = None
+    _option_namespaces: list[str] = ["dnf"]
```

and deal with this on the global validator level. For the record I tried the above with your example and got:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for tagged-union[function-after[_validate_pkgmgr_type_option_pairings(), BundlerPackageInput],function-after[_validate_pkgmgr_type_option_pairings(), GomodPackageInput],function-after[_validate_pkgmgr_type_option_pairings(), NpmPackageInput],function-after[_validate_pkgmgr_type_option_pairings(), PipPackageInput],function-after[_validate_pkgmgr_type_option_pairings(), RpmPackageInput],function-after[_validate_pkgmgr_type_option_pairings(), YarnPackageInput]]
E       pip
E         Value error, Unsupported options for type:'pip': {'dnf'} | supported options: '[]' [type=value_error, input_value={'type': 'pip', 'requirem...': {'dnf': {'foo': {}}}}, input_type=dict]
```

brunoapimentel: Fair enough. So maybe add a small docstring stating we intend this to be a generic model for all package managers?

brunoapimentel (approved):

cachi2/core/models/input.py
```diff
@@ -172,11 +117,62 @@ def _requirements_file_path_is_relative(cls, paths: list[Path]) -> list[Path]:
...
+    rather than our own. Each option set is namespaced by the corresponding tool/service it is
+    related to.
+
+    TODO: Enable this globally for all pkg managers not just the RpmPackageInput model.
+    """
+
+    dnf: Optional[Dict[Union[Literal["main"], str], Dict[str, Any]]] = None
+
+    @pydantic.model_validator(mode="before")
+    def _validate_dnf_options(cls, data: Any, info: pydantic.ValidationInfo) -> Any:
```
a-ovchinnikov, line 134: If we are to reuse this with other PMs then we should better define `_ExtraOptions` as a union type of individual PMs options. This is not a blocker and could be done when such need arises.

a-ovchinnikov (approved):
Mostly straightforward refactoring. LGTM.

cachi2/core/models/input.py
eskultety: > If we are to reuse this with other PMs then we should better define _ExtraOptions as a union type of individual PMs options. This is not a blocker and could be done when such need arises.

I don't follow - that's how it was the previously declared and it was a PITA to integrate with it. What would the Union bring over an explicitly enumerated list of keywords (e.g. `dnf`, `ssl`, `foo`) each defined as `Optional` further specified model?

PR #652, opened 2024-09-11
Design for generic artifact fetching

kosciCZ:

### DIFF
--- a/docs/design/generic.md
+++ b/docs/design/generic.md
@@ -1,0 +1,94 @@
+# Generic artifact fetching
+
+## Introduction
+
+This document will describe high-level implementation overview for supporting generic artifact fetching in cachi2.
+Up until now cachi2 has only supported package managers for various ecosystems and languages.
+However, there are a couple of use-cases where language non-specific artifacts need to be pre-fetched in order to satisfy
+requirements of a hermetic build.
+
+## Context
+
+For context, generic artifact fetching is a use-case of its own (e.g. [OVAL feeds](https://github.com/CISecurity/OVALRepo),
+AI models), it is also necessary precursor for implementing support for fetching maven artifacts, which won't be covered
+in this design, but in a followup document.
+
+## Design
+
+In this section, I will try to cover individual parts of the design.
+
+### Source repository
+
+This section will describe the structure of the source repository, that will serve as an input to cachi2. The idea is to
+define a cachi2 lockfile that will specify individual artifacts to fetch along with necessary metadata - e.g. checksums.
+The format chosen for this lockfile is yaml, and will include [purl](https://github.com/package-url/purl-spec) for each
+of the fetched artifacts. This decision was made mainly because it allows for followup implementation of maven support,
+with accurate SBOM information. Here's an example of such a lockfile.
+
+```yaml
+artifacts:
+  - purl: pkg:generic/granite-model?download_url=https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
+    target: granite-model.safetensors
+    checksums:
+      sha256: 07123e1f482356c415f684407a3b8723e10b2cbbc0b8fcd6282c49d37c9c1abc
+```
+
+#### Lockfile format and validation
+
+##### purl (required)
+
+At this point, the only purl type allowed would be `pkg:generic`. This is because cachi2 has no good way of verifying
+additional properties of the fetched artifact that could be included in the resulting SBOM. This should create a strong
+incentive to use this feature in the only truly necessary cases, because it will generate low-quality SBOM components,
+as compared to using other package managers provided by cachi2. Additionally, the only allowed qualifier should be `download_url`.
+
+#### target (optional)
+
+This is mainly for the users convenience, so the files end up in expected locations. Target here means a specific subdirectory
+inside cachi2's output directory. Special care needs to be taken to ensure there is not a conflict with other downloaded files.
+If not specified, filename of the downloaded file will be used.
+
+##### checksums (optional)
+
+I've chosen tho separate checksums from the purl, mostly for better readability of the lockfile, but this can be up for
+discussion. If no checksum is provided, cachi2 should still download the artifact, but report this fact in the output
+SBOM component.
+
+### SBOM
+
+Letting users specify artifacts as purls begs the question of authenticity of the data provided by the users and how it
+should be handled in the resulting SBOM. As described above, the purl is restricted to its basic components, so there is
+very little space for the user to provide inaccurate information. Cachi2 should verify that the file downloaded matches
+checksums and report the purl as-is, as it contains no extra information. The section below outlines how that information
+will be verified at later time.
+
+### Validation of user input
+
+As stated above, cachi2 will perform little to no verification of identity of the downloaded artifacts besides verifying
+checksums. However, it will provide enough information in the SBOM so tooling that comes after cachi2 can enforce policies.
+An example of this would be the [Enterprise Contract](https://enterprisecontract.dev/) (EC) project, that enforces policies
+based on the provided SBOM.
+
+In the context of this feature, EC policy would be supplied with the following information by cachi2:
+
+- checksums were provided and verified
+- list of checksum algorithms used
+- download urls (as part of the purl)
+  Enterprise contract policy would then be able to restrict accepting content without checksums, enforce certain algorithms
+- for checksum verification or only allow certain patterns in the download url.
+
+### Integration testing
+
+Since this feature is generic, the testing would be done with an example source repository containing the lockfile, with
+artifacts pointing to agreed upon urls.
+
+## Outcome
+
+Here's a preliminary work breakdown:
+
+- define models for the new package manager and high-level code structure into multiple modules
+- validate & parse generic artifact lockfile
+- download artifacts from the lockfile
+- add integration tests covering the new package manager
+- generate PURLs for all downloaded artifacts
+- add documentation
--- a/docs/design/generic.md
+++ b/docs/design/generic.md
@@ -33,10 +33,8 @@ field that will indicate the version of the lockfile format.
 metadata:
   version: 1.0.0
 artifacts:
-  - purl: pkg:generic/granite-model?download_url=https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
-    target: granite-model-1.safetensors
-    checksums:
-      sha256: 07123e1f482356c415f684407a3b8723e10b2cbbc0b8fcd6282c49d37c9c1abc
+  - purl: pkg:generic/granite-model?checksums=sha256:07123e1f482356c415f684407a3b8723e10b2cbbc0b8fcd6282c49d37c9c1abc&download_url=https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
+    target: granite-model-1.safetensors      
   - download_url: https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00002-of-00003.safetensors?download=true
     target: granite-model-2.safetensors
     checksums:
--- a/docs/design/generic.md
+++ b/docs/design/generic.md
@@ -69,13 +69,14 @@ checksums. However, it will provide enough information in the SBOM so tooling th
 An example of this would be the [Enterprise Contract](https://enterprisecontract.dev/) (EC) project, that enforces policies
 based on the provided SBOM.
 
-In the context of this feature, EC policy would be supplied with the following information by cachi2:
+In the context of this feature, EC would be supplied with the following information by cachi2 in the SBOM:
 
 - checksums were provided and verified
 - list of checksum algorithms used
 - download urls (as part of the purl)
-  Enterprise contract policy would then be able to restrict accepting content without checksums, enforce certain algorithms
-- for checksum verification or only allow certain patterns in the download url.
+
+Enterprise contract policy would then be able to restrict accepting content without checksums, enforce certain algorithms
+for checksum verification or only allow certain patterns in the download url.
 
 ### Integration testing
 
--- a/docs/design/generic.md
+++ b/docs/design/generic.md
@@ -9,50 +9,67 @@ requirements of a hermetic build.
 
 ## Context
 
-For context, generic artifact fetching is a use-case of its own (e.g. [OVAL feeds](https://github.com/CISecurity/OVALRepo),
-AI models), it is also necessary precursor for implementing support for fetching maven artifacts, which won't be covered
-in this design, but in a followup document.
+For context, generic artifact fetching is a use-case of its own (e.g. [OVAL feeds][oval-feeds], AI models), it is also
+necessary precursor for implementing support for fetching maven artifacts, which won't be covered in this design, but in
+a followup document.
 
 ## Design
 
 In this section, I will try to cover individual parts of the design.
 
-### Source repository
+### Source repository and cachi2 lockfile
 
 This section will describe the structure of the source repository, that will serve as an input to cachi2. The idea is to
 define a cachi2 lockfile that will specify individual artifacts to fetch along with necessary metadata - e.g. checksums.
-The format chosen for this lockfile is yaml, and will include [purl](https://github.com/package-url/purl-spec) for each
+The format chosen for this lockfile is yaml, and will include [purl] for each
 of the fetched artifacts. This decision was made mainly because it allows for followup implementation of maven support,
-with accurate SBOM information. Here's an example of such a lockfile.
+with accurate SBOM information. Alternatively, for better user experience, the download url and checksums can be specified
+separately, always resulting in a `pkg:generic` purl. Here's an example of such a lockfile with both options.
 
 ```yaml
 artifacts:
   - purl: pkg:generic/granite-model?download_url=https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
-    target: granite-model.safetensors
+    target: granite-model-1.safetensors
     checksums:
       sha256: 07123e1f482356c415f684407a3b8723e10b2cbbc0b8fcd6282c49d37c9c1abc
+  - download_url: https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00002-of-00003.safetensors?download=true
+    target: granite-model-2.safetensors
+    checksums:
+        sha256: 90bffe1884b84d5e255f12ff0ecbd70f2edfc877b68d612dc6fb50638b3ac17c
 ```
 
-#### Lockfile format and validation
+#### Specify by purl
+
+This option is provided mostly as a necessary step for the maven support, but can be used for generic artifacts as well.
+At this time, the only purl is `pkg:generic` and will result in a `pkg:generic` SBOM component.
 
 ##### purl (required)
 
 At this point, the only purl type allowed would be `pkg:generic`. This is because cachi2 has no good way of verifying
 additional properties of the fetched artifact that could be included in the resulting SBOM. This should create a strong
 incentive to use this feature in the only truly necessary cases, because it will generate low-quality SBOM components,
-as compared to using other package managers provided by cachi2. Additionally, the only allowed qualifier should be `download_url`.
+as compared to using other package managers provided by cachi2. Additionally, the only allowed qualifiers will be
+`download_url` and `checksums`.
+
+#### Specify by download_url and checksums
+
+This option is provided for better user experience by specifying the url and checksums separately. This option will always
+result in a `pkg:generic` purl.
+
+##### download_url (required)
+
+Specified as a string containing the download url of the artifact.
+
+##### checksums (required)
+
+Specified as a dictionary of checksum algorithms and their values. At least one cachi2-verifiable checksum must be provided.
 
 #### target (optional)
 
-This is mainly for the users convenience, so the files end up in expected locations. Target here means a specific subdirectory
-inside cachi2's output directory. Special care needs to be taken to ensure there is not a conflict with other downloaded files.
-If not specified, filename of the downloaded file will be used.
-
-##### checksums (optional)
-
-I've chosen tho separate checksums from the purl, mostly for better readability of the lockfile, but this can be up for
-discussion. If no checksum is provided, cachi2 should still download the artifact, but report this fact in the output
-SBOM component.
+This key is common for both options and providedmainly for the users convenience, so the files end up in expected locations.
+Target here means a specific subdirectory inside cachi2's output directory (likely `cachi2-output/deps/generic`).
+Special care needs to be taken to ensure there is not a conflict with other downloaded files. If not specified, filename
+of the downloaded file will be used.
 
 ### SBOM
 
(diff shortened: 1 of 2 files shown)


kosciCZ:
@brunoapimentel PTAL

docs/design/generic.md
```diff
@@ -0,0 +1,95 @@
...
+#### target (optional)
+
+This is mainly for the users convenience, so the files end up in expected locations. Target here means a specific subdirectory
+inside cachi2's output directory. Special care needs to be taken to ensure there is not a conflict with other downloaded files.
+If not specified, filename of the downloaded file will be used.
+
+##### checksums (optional)
+
+I've chosen tho separate checksums from the purl, mostly for better readability of the lockfile, but this can be up for
+discussion. If no checksum is provided, cachi2 should still download the artifact, but report this fact in the output
```
brunoapimentel, line 54: I'm not sure if this is the best approach here. We have precedent in allowing artifacts to be downloaded without user-defined checksums, but we'd still have the server-provided checksums. If no checksum is verified at all, I believe it brings a greater risk for the build that will consume the prefetched content.

docs/design/generic.md
```diff
@@ -0,0 +1,95 @@
...
+artifacts:
+  - purl: pkg:generic/granite-model?download_url=https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
+    target: granite-model.safetensors
+    checksums:
+      sha256: 07123e1f482356c415f684407a3b8723e10b2cbbc0b8fcd6282c49d37c9c1abc
+```
+
+#### Lockfile format and validation
+
+##### purl (required)
```
brunoapimentel, line 38: I wonder if purls provide the best user experience here. Maybe we could allow either:
- a pair of `download_url` and `checksums`
- a purl containing the info we need (which happens to be the fields above)

I believe this lockfile will be written manually in many cases, so I think keeping the user experience here is worth it.

docs/design/generic.md
```diff
@@ -0,0 +1,95 @@
...
+
+At this point, the only purl type allowed would be `pkg:generic`. This is because cachi2 has no good way of verifying
```
brunoapimentel, line 40: We don't necessarily need to restrict input purls to `pkg:generic`. As long as a purl is parseable and we can get the download_url and checksums from it, I'd say there's no harm in allowing them. The point is that the output purl (in the SBOM) will always be `pkg:generic`.

If we want to restrict purls to `pkg:generic` to keep the implementation simple, I'm fine with it.

docs/design/generic.md
brunoapimentel: @eskultety I remember you had a point against having a Cachi2-defined lockfile format?

brunoapimentel: Also, wdyt about asking the user to specify the checksum as part of a purl qualifier (since it is officially supported)?

docs/design/generic.md
```diff
@@ -0,0 +1,95 @@
...
+additional properties of the fetched artifact that could be included in the resulting SBOM. This should create a strong
+incentive to use this feature in the only truly necessary cases, because it will generate low-quality SBOM components,
+as compared to using other package managers provided by cachi2. Additionally, the only allowed qualifier should be `download_url`.
+
+#### target (optional)
```
brunoapimentel, line 45: I'm still not sure about this feature. I know the push here is to avoid breaking the workflow of some users that will migrate to Cachi2, but I think we should primarily see this from a more generic perspective. Will it benefit users of the generic fetching? Maybe it's useful to allow users to specify a filename that differs from what exists in the remote server they'll download the file from. 

In any case, to follow the precedent we have, I'd say that the `target` must be a subpath of `cachi2-output/deps/generic`

docs/design/generic.md
eskultety: > The idea here is delegate the decision to Enterprise Contract policy, to manage that risk, but be transparent about unverified files.

Please don't mention specific components where this is consumed, at this point in time this project isn't strictly tied to the Konflux project and so if you need to make an argument in a community space then try to make it true in general, e.g. `It cannot be assumed in general that each type or artifact will always have a corresponding checksum server-side and hence the missing checksum needs to be reported in the SBOM and verified in a different software supply chain pipeline stage`.

> > This is a requirement in the feature doc ([non-public link](https://docs.google.com/document/d/15Ibevfwp7FoafaBQcbtxYYJN-aoiF5jq8aNloA8UIZg/edit#bookmark=id.n7vnmxsap2sj))

Please don't paste links to internal documents in public spaces! You wouldn't do this in Bugzilla I presume without first marking the Bugzilla private. Annotating this as a private resource is as helpful to the context as not having the link at all, so either share a public version of it or proceed over a private channel, but not a public community space.

docs/design/generic.md
brunoapimentel: And do you think it's worth allowing users to just set `download_url` and `checksums`, as an alternative input format?

brunoapimentel: I even wonder if this purl input format should be exclusive to `maven-artifacts` :thinking:

docs/design/generic.md
brunoapimentel: You're right. I was trying to find a justification for this feature looking from a more generic perspective. But as long as it falls under `cachi2-output/deps/generic`, I'm ok with this feature.

brunoapimentel: In any case, we might need to do some check to avoid naming collisions for generic artifacts downloads.

docs/design/generic.md
brunoapimentel: Coming back to the original point, after a chat with other maintainers, we feel that the checksum validation is a very important step in the prefetch. At least one checksum should be validated for each downloaded artifact, and since for the generic fetcher we won't have server-provided checksums, we should just require the user to provide them.

This is extra sensitive for the generic fetcher: we are not relying on an already stablished language ecosystem, which means the user could download anything from the internet. We can't avoid that malicious content is fetched, but we should at least have strong guarantees that the content downloaded was exactly what the user expected.

kosciCZ: We'll need to hash it out (pun intended) with the Konflux people, but generally, I don't see this as a blocker to moving ahead with the overall implementation. 

I'm leaning towards agreeing with you, however, I'll need to confirm this is not a hard requirement.

docs/design/generic.md
kosciCZ: I think allowing both is actually a great idea, if they are mutually exclusive. 

**download_url + checksums:**
- more readable, easier to manually specify
- will ever only produce a generic purl
 
**purl:**
- less readable, more of a mess to manually specify
- chance of producing quality sbom for maven pkgs 

On the exclusiveness of purl to `pkg:maven`, I honestly can't decide. Maybe leaning more towards no, because if the lockfile is going to be generated automatically, you'd maybe want to have a single format for everything. Not sure how hard of an argument that is.

docs/design/generic.md
kosciCZ: > In any case, we might need to do some check to avoid naming collisions for generic artifacts downloads.
Agreed, I have included that in this section. Closing this as resolved, feel free to unresolve if you feel there's more to discuss.

docs/design/generic.md
kosciCZ: Closing this as resolved, feel free to unresolve if you feel there's more to discuss.

docs/design/generic.md
brunoapimentel: Alright, please share your findings.

docs/design/generic.md
brunoapimentel: I'm ok with having both options, or just keeping the "download_url + checksums" one for now.

kosciCZ: Let's do both. I'll update this in the design.

docs/design/generic.md
kosciCZ: I will. For now I've changed the design to require checksums.

docs/design/generic.md
```diff
@@ -0,0 +1,95 @@
...
+discussion. If no checksum is provided, cachi2 should still download the artifact, but report this fact in the output
+SBOM component.
+
+### SBOM
+
+Letting users specify artifacts as purls begs the question of authenticity of the data provided by the users and how it
+should be handled in the resulting SBOM. As described above, the purl is restricted to its basic components, so there is
+very little space for the user to provide inaccurate information. Cachi2 should verify that the file downloaded matches
+checksums and report the purl as-is, as it contains no extra information. The section below outlines how that information
+will be verified at later time.
```
eskultety, line 80: Sorry for coming late to the party...

> @eskultety I remember you had a point against having a Cachi2-defined lockfile format?

Yes I did. This project is about dependency processing and generating a record to serve as a proof of contents (or or dependency footprint if you will) for a given project. At the moment it is not about defining any public, long term supported API (besides the CLI). Especially when it comes to formats, given its current simplicity is almost always a bad idea from future perspective due to maintenance burden when it comes to extending it with further structural primitives like grouping or nesting in the future. This design merely mentioned 2 use cases we know about at the moment, but we don't know what the future will hold and with a plethora package managers out there for which the needs are hard to predict I really don't think we want to go down the rabbit hole of maintaining several versions of a generic input lockfile (the current RPM lockfile should serve as a precedent where, luckily, we didn't have to break the structure, but multiple fields had to be added on demand in a fairly short amount of time).

My main reason for not trying to define a format from scratch, however simple and readable it is, is ^this paragraph in the design which just says cachi2 will take the PURL as a given and paste it to the output SBOM. So, to me that bears the question, what exactly is the value of this particular format over an existing CycloneDX? Is it simplicity and readability? Well, the former IMO won't stay true forever and the latter is questionable because the idea is that most lockfiles/dependency lists are generated in an automation manner (i.e. a tool generates them), so it MUST be machine readable, but not necessary human readable, because it's only ever supposed to be an intermediary format for some further processing.
If you look at CycloneDX, while it sure has its flaws, it:
- is a format which is supposed to track artifacts (dependencies in our case)
- has matured over time
- has all the necessary fields we need today and many more we may need in the future
- can be already parsed by cachi2 (in a way) because we can merge our own SBOMs already

Accepting SBOMs on the input for generic artifacts with the same limiting conditions as you outline in this design can be applied to the input SBOM as well. Since we can already do merging, accepting SBOMs feels like the next natural step forward for this project. That said, we could lift the rules and limiting conditions on what we're accepting in the input SBOM gradually, without the need to change the underlying format (CycloneDX) at all.

docs/design/generic.md
```diff
@@ -0,0 +1,123 @@
...
+##### checksums (required)
+
+Specified as a dictionary of checksum algorithms and their values. At least one cachi2-verifiable checksum must be provided.
+
+#### target (optional)
+
+This key is common for both options and providedmainly for the users convenience, so the files end up in expected locations.
+Target here means a specific subdirectory inside cachi2's output directory (likely `cachi2-output/deps/generic`).
+Special care needs to be taken to ensure there is not a conflict with other downloaded files. If not specified, filename
+of the downloaded file will be used.
```
eskultety, line 72: I'd argue that just like our packages are currently categorized based on the connection to a particular pkg manager, this should almost exclusively be `generic`, otherwise (in case of Maven or whatnot) we'll have to figure it out again based on some common pattern/relevance to some pkg ecosystem, so I question the need for this field, because currently cachi2 cannot be configured like this for any pkg manager apart from pointing it to an alternative top-level directory. I feel like not being in full control of our own directory hierarchy opens the door to potential user configuration problems  we haven't faced yet.

kosciCZ: > I question the need for this field, because currently cachi2 cannot be configured like this for any pkg manager apart from pointing it to an alternative top-level directory.

In my opinion that is the case because for pkg manager, you generally don't care where the files are, as long as they are "available" in some way for the package manager during the build itself.

However this is not true for `generic` files, because they are often directly referenced in the Dockerfile. I will for now omit migration to cachi2 from OSBS (although important for the Konflux use case). However consider the following use case. 

Say you reference a versioned file (e.g. OVAL feed) in your Dockerfile. When the time comes to update to a new version, you will need to update the cachi2 lockfile (or the SBOM, whatever we decide to go with), AND update the Dockerfile, to change the reference. 

However,  with the `target` feature, you control the dependency only through the lockfile/SBOM, because updating the version of the dependency does not affect the way you use the dependency in the build, which is closer to the general experience when using a package manager. 
 
> I feel like not being in full control of our own directory hierarchy opens the door to potential user configuration problems we haven't faced yet.

I feel like this is an implementation detail that can be solved. As mentioned, cachi2 can ensure the path is always a subdirectory of `cachi2-output/deps/generic`

eskultety: > Say you reference a versioned file (e.g. OVAL feed) in your Dockerfile. When the time comes to update to a new version, you will need to update the cachi2 lockfile (or the SBOM, whatever we decide to go with), AND update the Dockerfile, to change the reference.
> 
> However, with the target feature, you control the dependency only through the lockfile/SBOM, because updating the version of the dependency does not affect the way you use the dependency in the build, which is closer to the general experience when using a package manager.

I don't get this part. Can you follow with an example? If you reference a versioned file in your Dockerfile, then changing the reference WILL require a change to the Dockerfile as well. I don't get how target rids you off the need to change the Dockerfile accordingly. If references are the issue and you really really don't want to update your Dockerfile, then you can install as `<cmd> <cachi2-output-path>/<mydependency>*.oval`

kosciCZ: I meant that in the Dockerfile, you don't care about the version, but about having the dependency there. 

An example might be that before using cachi2, you have this kind of stuff in the dockerfile.

```
RUN curl -o feed.oval "https://foo.com/bar-08-24.oval"

RUN somecommand feed.oval
```

Now say you want to start doing hermetic builds, therefore you need to pre-fetch that oval file. You will specify the url in lockfile/SBOM and cachi2 will pre-fetch it for you. Now when you want update the version of the file, you need to:
- update the lockfile/SBOM
- update the dockerfile from `RUN somecommand cachi2-output/deps/generic/bar-08-24.oval` to  `RUN somecommand cachi2-output/deps/generic/bar-09-24.oval`

Having the ability to specify target makes this easier. Given `target: feed.oval` you only need to need to change the lockfile/SBOM, because the dockerfile references the target, e.g.  `RUN somecommand cachi2-output/deps/generic/feed.oval`

docs/design/generic.md
kosciCZ: > My main reason for not trying to define a format from scratch, however simple and readable it is, is ^this paragraph in the design which just says cachi2 will take the PURL as a given and paste it to the output SBOM. 

Yes, but the design also mentions that the purl will be very restricted, only to `pkg:generic`, and to `download_url` and `checksums`. Therefore it is taking it as is, but under very restricted conditions.  

> So, to me that bears the question, what exactly is the value of this particular format over an existing CycloneDX? Is it simplicity and readability? Well, the former IMO won't stay true forever and the latter is questionable because the idea is that most lockfiles/dependency lists are generated in an automation manner (i.e. a tool generates them), so it MUST be machine readable, but not necessary human readable, because it's only ever supposed to be an intermediary format for some further processing.

Yes, in traditional package managers, these lockfiles are machine-generated, so they don't have to be user friendly. However, that is because usually there's tooling built around them, so when user needs to update a dependency they just do e.g `poetry add "package==1.1.0"` and don't have to deal with all the specifics of the format. This is **not** the case here. While the format is still machine-readable, there is no tooling to edit it for you. So I would argue that human readability matters here. 

> Accepting SBOMs on the input for generic artifacts with the same limiting conditions as you outline in this design can be applied to the input SBOM as well. Since we can already do merging, accepting SBOMs feels like the next natural step forward for this project. 

My main reason for not wanting to go with SBOM is that it feels like an overkill for this case. And I feel like it isn't a user-friendly format to specify dependencies. 

> That said, we could lift the rules and limiting conditions on what we're accepting in the input SBOM gradually, without the need to change the underlying format (CycloneDX) at all.

While CycloneDX SBOM is flexible and mature, it servers a slightly different purpose in my opinion. We'd still be bending an existing format to our needs. We only need the purl (or its broken-down components), and we'd have to restrict everything else, which in my opinion makes little sense.

Additionally, it poses the same risk as you mentioned, because it might occur that an currently unknown use case appears that CycloneDX won't be able to accommodate, and then you're locked in a specification you _cannot_ change. Custom lockfile format is better for the use case you mentioned.

docs/design/generic.md
eskultety: Okay, very well, instead of `<cmd> cachi2-output/generic/feed.oval` you can do `<cmd> cachi2-ouptut/generic/feed*.oval` so I don't see the technical need justifying this field, but if others are in favour, it's not a blocker for me having it. FWIW this doesn't go against a different input file format other than you're proposing.

docs/design/generic.md
brunoapimentel: I think I'd still rather keep the simple yaml format for now. If we don't have any expectations of needing to have a full SBOM as an input format, I don't see how the added complexity will help us right now. I also believe we want to keep this feature simple: it should download a file, verify its checksum, and make no other assumptions besides that (that's why it's reported as `pkg:generic`). Allowing this feature to grow beyond this scope might undermine Cachi2's fundamental purpose: accurate SBOMs.

Note that the purl copying from input file to output SBOM is something we're strongly against, and it is not what we're proposing here. If we ever decide to go for a route where, instead of the full source code, we'd receive only an input SBOM to prefetch, there'd be a lot of implications to think of (maybe Cachi2 will never be the project for it).

In any case, this feature will be rolled out initially as experimental, so we have room to change our decision without needing to honor backwards compatibility.

eskultety: > In any case, this feature will be rolled out initially as experimental, so we have room to change our decision without needing to honor backwards compatibility.

How long do you expect to have this as experimental if all initial requirements have been met? That's the thing, you most likely won't have this as experimental the moment you'll need to make a breaking change. The current format doesn't even count with the possibility of being versioned, that is going to be painful. Also, the  moment we start to vendor a new format, I don't think it's appropriate for the format to be version-locked with the project itself, it's possible for sure, but probably should be a different project we maintain. You also need to ensure some backwards compatibility, not expect to break it completely, IOW a particular version of cachi2 must always be able to parse older versions of a lockfile than the version that particular release of cachi2 can handle. These are also consideration I'm not seeing anywhere to be discussed in this thread, but fair enough, it seems folks are in favour of this regardless of any considerations, so I will no longer stand in the way.

eskultety: It most certainly would.

brunoapimentel: I think Erik has a good point about the versioning of the lockfile, as much as I'd like to keep this feature simple and generic, we can't be sure that no changes will be required in the future. Should we add a version field to it then?

kosciCZ: Added versioning to the lockfile.

docs/design/generic.md
kosciCZ: Thanks, closing this conversation as resolved then. I didn't include it as necessary, but as convenience for users.

docs/design/generic.md
```diff
@@ -0,0 +1,128 @@
...
+field that will indicate the version of the lockfile format.
+
+```yaml
+metadata:
+  version: 1.0.0
+artifacts:
+  - purl: pkg:generic/granite-model?download_url=https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
+    target: granite-model-1.safetensors
+    checksums:
+      sha256: 07123e1f482356c415f684407a3b8723e10b2cbbc0b8fcd6282c49d37c9c1abc
```
brunoapimentel, line 39: So, we won't require the checksums to be part of the purl (when the purl is present)?

docs/design/generic.md
```diff
@@ -0,0 +1,128 @@
...
+  - download_url: https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00002-of-00003.safetensors?download=true
```
brunoapimentel, line 40: What is the purpose of this `download=true` qualifier?

brunoapimentel (commented):
Leaving some additional comments just to clarify some points, but I believe we have a solid direction to start the implementation.

docs/design/generic.md
kosciCZ: That is part of the url on huggingface, I just copied whatever I got to download the file.

docs/design/generic.md
kosciCZ: My mistake, this is stale. The checksums should be specified as part of the purl. Updated.

docs/design/generic.md
eskultety: With https://github.com/containerbuildsystem/cachi2/pull/663 in mind (which was opened 2 days ago I lacked any insight into Maven handling) I think we should drop modeling PURL as part of this particular design doc, and especially `pkg:generic`, and keep it simple. In context of Maven `generic` IMO just looks confusing. Maven can enhance the format with a particular PURL spec type for the use case. We could adopt `pkg:generic` for input when we have a use case, otherwise we'd similarly to my original CycloneDX proposal be addressing an immediately non-existent use case.

Should be a small and straight forward change in this design, shouldn't block anything.

docs/design/generic.md
```diff
@@ -0,0 +1,126 @@
...
+of the fetched artifacts. This decision was made mainly because it allows for followup implementation of maven support,
+with accurate SBOM information. Alternatively, for better user experience, the download url and checksums can be specified
+separately, always resulting in a `pkg:generic` purl. Here's an example of such a lockfile with both options.
+
+In order to account for possible future breaking changes, the lockfile will contain a `metadata` section with a `version`
+field that will indicate the version of the lockfile format.
+
+```yaml
+metadata:
+  version: 1.0.0
```
eskultety, line 34: minor comment: I think we'd be fine even with X.Y version scheme for a file format, definitely not a blocker though, it's just version handling (though I haven't seen `.Z` elsewhere).

docs/design/generic.md
```diff
@@ -0,0 +1,126 @@
...
+
+Specified as a string containing the download url of the artifact.
+
+##### checksums (required)
+
+Specified as a dictionary of checksum algorithms and their values. At least one cachi2-verifiable checksum must be provided.
+
+#### target (optional)
+
+This key is common for both options and providedmainly for the users convenience, so the files end up in expected locations.
```
eskultety, line 72: `s/providedmainly/provided mainly`.

docs/design/generic.md
```diff
@@ -0,0 +1,126 @@
+# Generic artifact fetching
```
eskultety, line 1: nitpick: I don't think `generic.md` is a self-descriptive name and gives way too much room for imagination without reading the design - maybe `generic_fetcher.md` might sound a little better.

eskultety (approved):
Nothing major just quick nitpick fixes. In general I am in agreement. With the first ideas of contributing guidelines having materialized recently we can merge this OR keep it open until the implementation work is done, update sections as necessary should some further technical considerations be made during development, etc. I'd be fine either way. That said, if you wish to merge this right away then please squash all commits into a single one, no fixups please.

docs/design/generic.md
```diff
@@ -0,0 +1,94 @@
...
```
eskultety, line 1: **Commit message:**

Please drop the JIRA ID from the commit message, it's an internal non-transparent designation irrelevant to upstream.

docs/design/generic.md
kosciCZ: I'm not sure I completely understand, let me see if I got it right. Are you saying that you only want to accept the `download_url + checksums` for the generic fetcher part, and add an option to specify `purl`, restricted to only `pkg:maven` in #663?

eskultety: Yes.

docs/design/generic.md
brunoapimentel: Will this be a problem to us? We're planning on limiting the url qualifiers, right? Would this also apply to the `download_url`, or only for `purl`?

brunoapimentel (approved):
Giving my ack here since I'm also in agreement with the design, but I'd suggest not to merge this (as Erik commented before), and turn it in an ADR when the feature is production-ready.

lkolacek:
> Giving my ack here since I'm also in agreement with the design, but I'd suggest not to merge this (as Erik commented before), and turn it in an ADR when the feature is production-ready.

+1

lkolacek (approved):
LGTM +1, thank you guys!

docs/design/generic.md
kosciCZ: That would be only for purl, not the url itself.

docs/design/generic.md
kosciCZ: Dropped

docs/design/generic.md
kosciCZ: Renamed

docs/design/generic.md
kosciCZ: Fixed

docs/design/generic.md
kosciCZ: Fixed, good point

docs/design/generic.md
kosciCZ: Fixed, good point.

kosciCZ:
Thanks for all your reviews, I'll keep this open and will then edit it into ADR when we're done with the implementation.

kosciCZ:
Opened a new PR #718 that converts this content into ADR and user documentation

kosciCZ:
Closing in favor of #728

PR #654, opened 2024-09-17
bundler: Provide the build configuration

a-ovchinnikov:
In order for Bundler to build from the content prefetched by Cachi2, we need to set a few configuration parameters. There are several ways of doing this, but the most robust one is to place a local configuration file into processed repository (see docs/design/bundler.md for details).

This commit introduces a fetch step that presets configuraion parameters.

### DIFF
--- a/cachi2/core/package_managers/bundler/main.py
+++ b/cachi2/core/package_managers/bundler/main.py
@@ -1,5 +1,7 @@
 import logging
+import os
 from pathlib import Path
+from textwrap import dedent
 from typing import Optional
 
 from packageurl import PackageURL
@@ -14,11 +16,15 @@ from cachi2.core.scm import get_repo_id
 
 log = logging.getLogger(__name__)
 
+CONFIG_OVERRIDE = "bundler/config_override"
+
 
 def fetch_bundler_source(request: Request) -> RequestOutput:
     """Resolve and process all bundler packages."""
     components: list[Component] = []
-    environment_variables: list[EnvironmentVariable] = []
+    environment_variables: list[EnvironmentVariable] = (
+        _prepare_environment_variables_for_hermetic_build()
+    )
     project_files: list[ProjectFile] = []
 
     for package in request.packages:
@@ -26,6 +32,7 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
         components.extend(
             _resolve_bundler_package(package_dir=path_within_root, output_dir=request.output_dir)
         )
+        project_files.append(_prepare_for_hermetic_build(request.source_dir, request.output_dir))
 
     return RequestOutput.from_obj_list(
         components=components,
@@ -118,3 +125,35 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
 
     resolved_path = Path(repo_name).joinpath(package_dir.subpath_from_root)
     return str(resolved_path)
+
+
+def _prepare_environment_variables_for_hermetic_build() -> list[EnvironmentVariable]:
+    return [
+        # Contains path to a directory where a new config could be found.
+        EnvironmentVariable(name="BUNDLE_APP_CONFIG", value="${output_dir}/" + CONFIG_OVERRIDE),
+    ]
+
+
+def _prepare_for_hermetic_build(source_dir: RootedPath, output_dir: RootedPath) -> ProjectFile:
+    """Prepare a package for hermetic build by injecting necessary config."""
+    potential_bundle_config = source_dir.join_within_root(".bundle/config").path
+    hermetic_config = dedent(
+        """
+        BUNDLE_CACHE_PATH: "${output_dir}/deps/bundler"
+        BUNDLE_DEPLOYMENT: "true"
+        BUNDLE_NO_PRUNE: "true"
+    """
+    )
+    if potential_bundle_config.is_file():
+        config_data = potential_bundle_config.read_text()
+        config_data += hermetic_config
+    elif (alternative_config := os.getenv("BUNDLE_APP_CONFIG")) is not None:
+        # Corner case: a user decides to define their own alternate config.
+        # In this scenario cachi2 must try to copy over user-defined variables
+        # to its overriding alternate config.
+        config_data = Path(alternative_config, "config").read_text()
+        config_data += hermetic_config
+    else:
+        config_data = hermetic_config
+    overriding_bundler_config_path = output_dir.join_within_root(CONFIG_OVERRIDE, "config").path
+    return ProjectFile(abspath=overriding_bundler_config_path, template=config_data)
--- a/tests/integration/test_bundler.py
+++ b/tests/integration/test_bundler.py
@@ -53,6 +53,34 @@ log = logging.getLogger(__name__)
             ),
             id="bundler_malformed_lockfile",
         ),
+        pytest.param(
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-bundler.git",
+                ref="well_formed_ruby_all_features",
+                packages=({"path": ".", "type": "bundler"},),
+                flags=["--dev-package-managers"],
+                check_output=False,
+                check_deps_checksums=False,
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="",
+            ),
+            id="bundler_everything_present",
+        ),
+        pytest.param(
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-bundler.git",
+                ref="well_formed_ruby_without_gemspec",
+                packages=({"path": ".", "type": "bundler"},),
+                flags=["--dev-package-managers"],
+                check_output=False,
+                check_deps_checksums=False,
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="",
+            ),
+            id="bundler_everything_present_except_gemspec",
+        ),
     ],
 )
 def test_bundler_packages(
--- a/tests/unit/package_managers/bundler/test_main.py
+++ b/tests/unit/package_managers/bundler/test_main.py
@@ -1,3 +1,4 @@
+from textwrap import dedent
 from unittest import mock
 
 import pytest
@@ -6,6 +7,7 @@ from git.repo import Repo
 from cachi2.core.errors import PackageRejected
 from cachi2.core.package_managers.bundler.main import (
     _get_main_package_name_and_version,
+    _prepare_for_hermetic_build,
     _resolve_bundler_package,
 )
 from cachi2.core.package_managers.bundler.parser import (
@@ -114,3 +116,104 @@ def test_get_main_package_name_and_version_from_repo_without_origin(
         _get_main_package_name_and_version(package_dir=rooted_tmp_path_repo, dependencies=[])
 
     assert "Failed to extract package name from origin remote" in exc_info.value.friendly_msg()
+
+
+def test__prepare_for_hermetic_build_injects_necessary_variable_into_empty_config(
+    rooted_tmp_path: RootedPath,
+) -> None:
+    expected_config_location = rooted_tmp_path.join_within_root(".bundle/config").path
+    expected_config_contents = dedent(
+        """
+        BUNDLE_CACHE_PATH: "${output_dir}/deps/bundler"
+        BUNDLE_DEPLOYMENT: "true"
+        BUNDLE_NO_PRUNE: "true"
+        """
+    )
+
+    assert not expected_config_location.exists(), "Unexpected .bundle/config in rooted_tmp_path"
+
+    result = _prepare_for_hermetic_build(rooted_tmp_path, rooted_tmp_path)
+
+    assert result.template == expected_config_contents
+
+
+def test__prepare_for_hermetic_build_injects_necessary_variable_into_existing_config(
+    rooted_tmp_path: RootedPath,
+) -> None:
+    expected_config_location = rooted_tmp_path.join_within_root(".bundle/config").path
+    expected_config_contents = dedent(
+        """
+        BUNDLE_CACHE_PATH: "${output_dir}/deps/bundler"
+        BUNDLE_DEPLOYMENT: "true"
+        BUNDLE_NO_PRUNE: "true"
+        """
+    )
+    existing_preamble = dedent(
+        """---
+
+        BUNDLER_NONEXISTENT_VARIABLE: "true"
+        """
+    )
+
+    assert not expected_config_location.exists(), "Unexpected .bundle/config in rooted_tmp_path"
+    assert not expected_config_location.parent.exists(), "Unexpected .bundle/ in rooted_tmp_path"
+
+    expected_config_location.parent.mkdir()
+    expected_config_location.write_text(existing_preamble)
+
+    result = _prepare_for_hermetic_build(rooted_tmp_path, rooted_tmp_path)
+
+    assert result.template == existing_preamble + expected_config_contents
+
+
+def test__prepare_for_hermetic_build_injects_necessary_variable_into_existing_alternate_config(
+    rooted_tmp_path: RootedPath,
+) -> None:
+    expected_alternate_config_location = rooted_tmp_path.join_within_root("alternate/config").path
+    expected_alternate_config_contents = dedent(
+        """
+        BUNDLE_CACHE_PATH: "${output_dir}/deps/bundler"
+        BUNDLE_DEPLOYMENT: "true"
+        BUNDLE_NO_PRUNE: "true"
+        """
+    )
+    existing_preamble = dedent(
+        """---
+        BUNDLER_NONEXISTENT_VARIABLE: "true"
+        """
+    )
+
+    assert (
+        not expected_alternate_config_location.exists()
+    ), "Unexpected .bundle/config in rooted_tmp_path"
+    assert (
+        not expected_alternate_config_location.parent.exists()
+    ), "Unexpected .bundle/ in rooted_tmp_path"
+
+    expected_alternate_config_location.parent.mkdir()
+    expected_alternate_config_location.write_text(existing_preamble)
+
+    with mock.patch("cachi2.core.package_managers.bundler.main.os.getenv") as ge:
+        ge.return_value = str(expected_alternate_config_location.parent)
+        result = _prepare_for_hermetic_build(rooted_tmp_path, rooted_tmp_path)
+
+    assert result.template == existing_preamble + expected_alternate_config_contents
+
+
+def test__prepare_for_hermetic_build_ignores_a_directory_in_place_of_config(
+    rooted_tmp_path: RootedPath,
+) -> None:
+    expected_config_location = rooted_tmp_path.join_within_root(".bundle/config").path
+    expected_config_contents = dedent(
+        """
+        BUNDLE_CACHE_PATH: "${output_dir}/deps/bundler"
+        BUNDLE_DEPLOYMENT: "true"
+        BUNDLE_NO_PRUNE: "true"
+        """
+    )
+
+    assert not expected_config_location.exists(), "Unexpected .bundle/config in rooted_tmp_path"
+    assert not expected_config_location.parent.exists(), "Unexpected .bundle/ in rooted_tmp_path"
+    result = _prepare_for_hermetic_build(rooted_tmp_path, rooted_tmp_path)
+
+    assert result.template == expected_config_contents


cachi2/core/package_managers/bundler/main.py
```diff
@@ -26,3 +31,30 @@ def _resolve_bundler_package(package_dir: RootedPath, output_dir: RootedPath) ->
     """Process a request for a single bundler package."""
     parse_lockfile(package_dir)
     return []
+
+
+def _prepare_for_hermetic_build(output_dir: RootedPath) -> None:
+    """Prepare a package for hermetic build by injecting necessary config."""
+    bundle_config = output_dir.join_within_root(".bundle/config").path
```
slimreaper35, line 38: I think we don't want to have any config file in the output dir, since users can request multiple bundler packages and this might lead to collision.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -26,3 +31,30 @@ def _resolve_bundler_package(package_dir: RootedPath, output_dir: RootedPath) ->
...
+    hermetic_config = dedent(
+        f"""
+        BUNDLE_CACHE_PATH={output_dir}/deps/rubygems
+        BUNDLE_DEPLOYMENT=true
+        BUNDLE_NO_PRUNE=true
+    """
+    )
+    if bundle_config.exists():
```
slimreaper35, line 46: Also, the config file is not mandatory.
https://github.com/3scale/porta  -  There is no `.bundle/config`

a-ovchinnikov: Yes, but it could be present and if it is present fir any reason then we do not want to just overwrite it.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Good point, where it belongs then?

slimreaper35: > Good point, where it belongs then?

It belongs in the user repo, right ?

The problem within bundler is that environment variables do not have the highest precedence (unlike all other package managers). The highest precedence is the config file. I would have to check the design doc, but I think we can set the variables only if there is no config file, and if it is, we should overwrite those settings.

There is `$BUNDLE_APP_CONFIG` variable that can overwrite default config file path :exploding_head: 
https://bundler.io/v2.5/man/bundle-config.1.html#DESCRIPTION

cachi2/core/package_managers/bundler/main.py
```diff
@@ -26,3 +31,30 @@ def _resolve_bundler_package(package_dir: RootedPath, output_dir: RootedPath) ->
...
     parse_lockfile(package_dir)
     return []
+
+
+def _prepare_for_hermetic_build(output_dir: RootedPath) -> None:
+    """Prepare a package for hermetic build by injecting necessary config."""
+    bundle_config = output_dir.join_within_root(".bundle/config").path
+    hermetic_config = dedent(
+        f"""
+        BUNDLE_CACHE_PATH={output_dir}/deps/rubygems
```
eskultety, line 41: `output_dir` is the absolute path where cachi2 stores the artifacts, but that path won't exist in the build environment, that path is determined by the CLI option `--for-output-dir` which injects the path to these env variables which we know will appear in the container. The way you apply that in code is by using `${output_dir}` which is a pattern that's going to be replaced either by the `generate-env` or `inject-files` command, we use Python's `Template` for those substitutions.

cachi2/core/package_managers/bundler/main.py
eskultety: > we do not want to just overwrite it.

I believe we have no choice, we had a long discussion on that in the design discussion, and it's due to the order of precedence where config takes priority over variables. If there are multiple ruby packages in a repo we can't use `$BUNDLE_APP_CONFIG` because it would be global for all packages we process, unfortunately, but each package can define their own options, so we have to override them individually, it's weird, it's wrong, but I don't think we've come up with a different idea in the design, have we?

cachi2/core/package_managers/bundler/main.py
eskultety: > There is $BUNDLE_APP_CONFIG variable that can overwrite default config file path 🤯

We knew about that variable, but we only define a single set of variables that are global to the whole build env, so in a multi-package repo setup you'd be applying configuration from package A + our overrides to package B.

eskultety: > Good point, where does it belong then?

Have a look at https://github.com/containerbuildsystem/cachi2/blob/3398f3de3c8ce6370be06a49701645f8edffe8e3/cachi2/core/package_managers/rpm/main.py#L230, we'll have to come up with a similar (or unified?) way to dump the options to the build config. Because of multi-package repo setups I'd suggest to namespace each set of config vars by the package name we're processing, so that when issuing the `inject-files` CLI command we'll know where to dump each of the config files and write to the input user repos.

cachi2/core/package_managers/bundler/main.py
eskultety: Actually, I have an alternative proposal. Since it's inherently wrong to dirty users' repos we could document this as a known limitation of our project. IOW, we would say that while we support compound repos, in case users use a repo with more than a single Ruby package it might happen (depending on whether `.bundle/config`s were checked in the repo for the respective packages) that during the hermetic build we might use wrong bundler configuration. <<< This is based on the fact that encountering mono-language compound repos is a fairly rare case so realistically a miniscule percentage of the user base would be actually affected. We could then make use of the `$BUNDLE_APP_CONFIG` variable. That should IMO be fine for some initial implementation of bundler support.

Then, as a follow up, we should think of enhancing our core architecture by either:
- having a per-package build_config.json OR
- having a single build_config.json, but namespace the env vars and settings we need to apply **per-package**, so in practical terms in Bundler's case we'd have a set of config options for package A and a different set for package B. We'd then have to document that users have to do `source cachi2-output/<pkg>/cachi2.env` N times depending on how many packages (N) they're trying to build in their environment. That way we could define the `$BUNDLE_APP_CONFIG` multiple times without having to worry that it would override some other package's bundler config.
Like I said ^this is a follow up material and we should track it as an issue, because this is not the first time we're seeing this problem, the same is happening with extra input JSON options for the rpm backend.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: I did not realize that, thank you!

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: So in the end it looks like having the variables set would effectively cover the most frequent use-cases, at least for now. Everything else is probably unlikely enough. I'll do that in the next revision.

a-ovchinnikov: Searching github for the variables that we set resulted in less than a hundred matches for all three of them out of more than a million repositories. I think this confirms that while possible this failure scenario is not very probable.

eskultety: > I think this confirms that while possible this failure scenario is not very probable.

It's not about the variables we set, it's more about using **some** set of variables coming from package A and using it to build package B, so it's very particular mix of events that would lead to the problem. That said, we currently don't have any such limitation on compound repos with other pkg mgrs, so we need to be open about this, yet small, likelihood of incorrect variable usage.
@a-ovchinnikov in general, did the package namespaced settings idea in the build config make sense to you?

cachi2/core/package_managers/bundler/main.py
```diff
@@ -26,3 +31,14 @@ def _resolve_bundler_package(package_dir: RootedPath, output_dir: RootedPath) ->
     """Process a request for a single bundler package."""
     parse_lockfile(package_dir)
     return []
+
+
+def _prepare_environment_variables_for_hermetic_build() -> list[EnvironmentVariable]:
+    # NOTE: these variables are necessary for hermetic builds and have to be set for
+    # the entire repository. There is a small chance that their values will be
+    # overridden by values in .bundle/configs checked in some of packages' repositories.
```
slimreaper35, line 39: Should we check at first if `BUNDLE_APP_CONFIG` is set and then use default location ?

a-ovchinnikov: My understanding is that if a variable is set in environment and is not set in any config then it will be taken from the environment so there is no need to use default location. Here are results of a quick experiment I just ran:
```
$ bundle init
Writing new Gemfile to /tmp/bc/Gemfile
$ ls -a
.  ..  Gemfile
$ bundle config
Settings are listed in order of priority. The top value will be used.
$ export BUNDLE_NO_PRUNE=true
$ bundle config
Settings are listed in order of priority. The top value will be used.
no_prune
Set via BUNDLE_NO_PRUNE: true
$ bundle config local.BUNDLE_RETRY 25
$ bundle config
Settings are listed in order of priority. The top value will be used.
local.bundle_retry
Set for your local app (/tmp/bc/.bundle/config): "25"

no_prune
Set via BUNDLE_NO_PRUNE: true

$ ls -a
.  ..  .bundle  Gemfile
$ cat .bundle/config 
---
BUNDLE_LOCAL__BUNDLE_RETRY: "25"
$ sed -i s/25/14/g .bundle/config
$ bundle config
Settings are listed in order of priority. The top value will be used.
local.bundle_retry
Set for your local app (/tmp/bc/.bundle/config): "14"

no_prune
Set via BUNDLE_NO_PRUNE: true

$ sed -i s/BUNDLE_LOCAL__BUNDLE_RETRY/BUNDLE_RETRY/g .bundle/config
$ bundle config
Settings are listed in order of priority. The top value will be used.
no_prune
Set via BUNDLE_NO_PRUNE: true

retry
Set for your local app (/tmp/bc/.bundle/config): 14
```

It looks like as long as there is no clash between local and environment we are good to set whatever we need in the environment. Judging from my yesterdays search on Github the probability of a clash is very small.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Yes, namespaces make sense as a general solution to the problem. I am not sure we have this problem now or will have it in the near future because I switched to setting environment variables and those do not seem to clash much with anything found in a local config (judging from a Github search).

cachi2/core/package_managers/bundler/main.py
```diff
@@ -99,3 +101,14 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
 
     resolved_path = Path(repo_name).joinpath(package_dir.subpath_from_root)
     return str(resolved_path)
+
+
+def _prepare_environment_variables_for_hermetic_build() -> list[EnvironmentVariable]:
+    # NOTE: these variables are necessary for hermetic builds and have to be set for
+    # the entire repository. There is a small chance that their values will be
+    # overridden by values in .bundle/configs checked in some of packages' repositories.
```
slimreaper35, line 109:
```suggestion
    # NOTE: These variables are necessary for hermetic builds and have to be set for
    # the entire repository. There is a small chance that their values will be
    # overridden by values in the .bundle/config file.
```

a-ovchinnikov: Done.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Yes, this makes total sense, thank you. I'll switch to using a dedicated config and a variable to point to it in the next revision.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -99,3 +106,36 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+        EnvironmentVariable(name="BUNDLE_APP_CONFIG", value="${output_dir}/" + CONFIG_OVERRIDE),
+    ]
+
+
+def _prepare_for_hermetic_build(source_dir: RootedPath, output_dir: RootedPath) -> None:
+    """Prepare a package for hermetic build by injecting necessary config."""
+    potential_bundle_config = source_dir.join_within_root(".bundle/config").path
+    hermetic_config = dedent(
+        f"""
+        BUNDLE_CACHE_PATH: "{output_dir}/deps/rubygems"
```
eskultety, line 123: You'll need `${output_dir}` here as well.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -99,3 +106,36 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+
+def _prepare_environment_variables_for_hermetic_build() -> list[EnvironmentVariable]:
+    return [
+        # Contains path to a directory where a new config could be found.
+        EnvironmentVariable(name="BUNDLE_APP_CONFIG", value="${output_dir}/" + CONFIG_OVERRIDE),
+    ]
+
+
+def _prepare_for_hermetic_build(source_dir: RootedPath, output_dir: RootedPath) -> None:
+    """Prepare a package for hermetic build by injecting necessary config."""
```
eskultety, line 119: This prepare can't dump the file, because in order to do so, it would need to know the real target `${output_dir}` which we won't know until either `generate-env` or `inject-files` is run. Therefore, you'll need to split the logic and dump an intermediary data to the build config and only dump the file to disk with one of the other commands (most likely `inject-files`).

cachi2/core/package_managers/bundler/main.py
```diff
@@ -12,16 +14,21 @@
 
 log = logging.getLogger(__name__)
 
+CONFIG_OVERRIDE = "config_override"
```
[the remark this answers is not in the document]
a-ovchinnikov: I wanted to keep this directory explicitly separate. `.bundle` would work too, but I am unsure if we are ok with making it look like it is something bundler-created. I'll postpone changing the name for now.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Done, thank you.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -99,3 +106,36 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+    potential_bundle_config = source_dir.join_within_root(".bundle/config").path
+    hermetic_config = dedent(
+        f"""
+        BUNDLE_CACHE_PATH: "{output_dir}/deps/rubygems"
+        BUNDLE_DEPLOYMENT: "true"
+        BUNDLE_NO_PRUNE: "true"
+    """
+    )
+    if potential_bundle_config.exists():
+        print("=== WAAAA ====")
```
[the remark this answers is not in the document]
a-ovchinnikov: Spot on! :D _Edit_: waaaa, there is a whole bunch of them here.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -99,3 +106,36 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+        with suppress(IsADirectoryError):
```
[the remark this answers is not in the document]
a-ovchinnikov: Good point, done.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -99,3 +106,36 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+            print("=== WAAAA2 ====")
+            config_data = potential_bundle_config.read_text()
+            config_data += hermetic_config
+            print("=== WAAAA3 ====")
+    else:
+        config_data = "---\n" + hermetic_config
```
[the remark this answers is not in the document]
a-ovchinnikov: Indeed. I think during one of earlier experiments with bundler I changed two variables simultaneously, one of them was adding a triple-dash. It stuck since then. Removed.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Oh, it will be a different output_dir then, not the one I thought!

cachi2/core/package_managers/bundler/main.py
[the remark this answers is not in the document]
eskultety: > doesn't really hint at any relevance to the bundler backend at all which is a bit odd IMO

Similarly to my build-config namespacing suggestion in an earlier thread we should dump pkg-mgr related artifacts to namespaced directories under `output-dir` to improve artifact clarity.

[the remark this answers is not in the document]
a-ovchinnikov: > doesn't really hint at any relevance to the bundler backend at all which is a bit odd IMO

Good point, changed to `bundle/config_override`.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -99,3 +106,36 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+def _prepare_environment_variables_for_hermetic_build() -> list[EnvironmentVariable]:
+    return [
+        # Contains path to a directory where a new config could be found.
+        EnvironmentVariable(name="BUNDLE_APP_CONFIG", value="${output_dir}/" + CONFIG_OVERRIDE),
+    ]
+
+
+def _prepare_for_hermetic_build(source_dir: RootedPath, output_dir: RootedPath) -> None:
+    """Prepare a package for hermetic build by injecting necessary config."""
+    potential_bundle_config = source_dir.join_within_root(".bundle/config").path
```
slimreaper35, line 120: I have to come back again to my comment here https://github.com/containerbuildsystem/cachi2/pull/654#discussion_r1766353638.

If a user sets `ENV BUNDLE_APP_CONFIG=....` in his Dockerfile to some location other than `.bundler/config` we should check that, if he did, copy the settings and then overwrite the variable, right ?

eskultety: Are you suggesting we start parsing users' Dockerfiles? I don't think we want to get into that business given the complexity of multi-stage builds for this kind of validation. Besides, having `BUNDLE_APP_CONFIG` set in the Dockerfile makes no practical sense because you always want to be able to build your project in any environment and in a reproducible manner and allowing the build configuration to be changed like this goes against that principle (even without cachi2 in the picture), so I don't think we even need to concern ourselves with this scenario.

slimreaper35: No parsing. Just checking env variable. I know it is corner case, but everything feels like corner case here.

```
export BUNDLE_APP_CONFIG=foo
bundler config list
```

eskultety: > No parsing. Just checking env variable. I know it is corner case, but everything feels like corner case here.

How does this relate to the Dockerfile use case? What you've described is the local direct usage of cachi2 outside of a trusted pipeline containerized environment and my point

> Besides, having BUNDLE_APP_CONFIG set in the Dockerfile makes no practical sense because you always want to be able to build your project in any environment and in a reproducible manner and allowing the build configuration to be changed like this goes against that principle (even without cachi2 in the picture), so I don't think we even need to concern ourselves with this scenario.

applies to this scenario as well. AFAIK we don't do this for other pkg mgrs.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -118,3 +125,35 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+        EnvironmentVariable(name="BUNDLE_APP_CONFIG", value="${output_dir}/" + CONFIG_OVERRIDE),
+    ]
+
+
+def _prepare_for_hermetic_build(source_dir: RootedPath, output_dir: RootedPath) -> ProjectFile:
+    """Prepare a package for hermetic build by injecting necessary config."""
+    potential_bundle_config = source_dir.join_within_root(".bundle/config").path
+    hermetic_config = dedent(
+        """
+        BUNDLE_CACHE_PATH: "${output_dir}/deps/rubygems"
```
slimreaper35, line 142:
```suggestion
        BUNDLE_CACHE_PATH: "${output_dir}/deps/bundler"
```

cachi2/core/package_managers/bundler/main.py
```diff
@@ -118,3 +125,35 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+        config_data += hermetic_config
+    # Corner case: a user decides to define their own alternate config.
+    # In this scenario cachi2 must try to copy over user-defined variables
+    # to its overriding alternate config.
+    elif (alternative_config := os.getenv("BUNDLE_APP_CONFIG")) is not None:
+        config_data = (Path(alternative_config) / "config").read_text()
+        config_data += hermetic_config
+    else:
+        config_data = hermetic_config
+    overriding_bundler_config_path = output_dir.join_within_root(CONFIG_OVERRIDE + "/config").path
```
slimreaper35, line 158:
```suggestion
    overriding_bundler_config_path = output_dir.join_within_root(CONFIG_OVERRIDE, "config").path
```

cachi2/core/package_managers/bundler/main.py
```diff
@@ -118,3 +125,35 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+    """
+    )
+    if potential_bundle_config.is_file():
+        config_data = potential_bundle_config.read_text()
+        config_data += hermetic_config
+    # Corner case: a user decides to define their own alternate config.
+    # In this scenario cachi2 must try to copy over user-defined variables
+    # to its overriding alternate config.
+    elif (alternative_config := os.getenv("BUNDLE_APP_CONFIG")) is not None:
+        config_data = (Path(alternative_config) / "config").read_text()
```
slimreaper35, line 154: Here you can use `joinpath` and remove extra parentheses, but fine.

tests/integration/test_bundler.py
```diff
@@ -53,6 +53,34 @@
...
+                packages=({"path": ".", "type": "bundler"},),
+                flags=["--dev-package-managers"],
+                check_output=False,
+                check_deps_checksums=False,
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="",
+            ),
+            id="bundler_everything_present_except_gemspec",
+        ),
```
slimreaper35, line 83: Is it worth adding a scenario with multiple packages in the same repo ?

cachi2/core/package_managers/bundler/main.py
eskultety: `joinpath` should be IMO avoided, vast majority of cases can be written in a more compact way by just using `Path` directly. Note `joinpath` doesn't even work over the same object and returns a new one so there's no practical advantage. In this case `Path(alternative_config, "config") ` is all that's needed.

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -114,3 +115,83 @@ def test_get_main_package_name_and_version_from_repo_without_origin(
         _get_main_package_name_and_version(package_dir=rooted_tmp_path_repo, dependencies=[])
 
     assert "Failed to extract package name from origin remote" in exc_info.value.friendly_msg()
+
+
+def test__prepare_for_hermetic_build_injects_necessary_variable_into_empty_config(
```
eskultety, line 120: what's the double underscore for?

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -114,3 +115,83 @@ def test_get_main_package_name_and_version_from_repo_without_origin(
...
+    rooted_tmp_path: RootedPath,
+) -> None:
+    expected_config_location = rooted_tmp_path.join_within_root(".bundle/config").path
+    expected_config_contents = (
+        '\nBUNDLE_CACHE_PATH: "${output_dir}/deps/rubygems"\n'
```
eskultety, line 125: s/rubygems/bundler

cachi2/core/package_managers/bundler/main.py
```diff
@@ -14,18 +16,23 @@
 
 log = logging.getLogger(__name__)
 
+CONFIG_OVERRIDE = "bundle/config_override"
```
eskultety, line 19: s/bundle/bundler   - _bundle_ is the location Bundler places its stuff into according to the docs, _bundler_ is the namespace for our backend

Since it appears you've received more review comments and you'll be re-spinning one more time you might as well drop the additional directory level for the config override and just have the override config file in there directly.

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -114,3 +115,83 @@ def test_get_main_package_name_and_version_from_repo_without_origin(
...
+        'BUNDLE_DEPLOYMENT: "true"\nBUNDLE_NO_PRUNE: "true"\n'
```
eskultety, line 126: Would a `textwrap.dedent` + tripple quote string be actually more readable for the expected output than what is proposed?

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -114,3 +115,83 @@ def test_get_main_package_name_and_version_from_repo_without_origin(
...
+
+def test__prepare_for_hermetic_build_injects_necessary_variable_into_existing_config(
+    rooted_tmp_path: RootedPath,
+) -> None:
+    expected_config_location = rooted_tmp_path.join_within_root(".bundle/config").path
+    expected_config_contents = (
+        '\nBUNDLE_CACHE_PATH: "${output_dir}/deps/rubygems"\n'
+        'BUNDLE_DEPLOYMENT: "true"\nBUNDLE_NO_PRUNE: "true"\n'
+    )
+    existing_preamble = '---\n\nBUNDLER_NONEXISTENT_VARIABLE: "true"\n'
```
eskultety, line 144: same as above with `textwrap.dedent` and tripple quotes, it's the pattern I believe we use consistently in most places.

cachi2/core/package_managers/bundler/main.py
slimreaper35: I see it is already implemented.

a-ovchinnikov: Yes, I have added this check. Admittedly this corner case looks somewhat contrived, as it would indicate a user doing something rather unusual, but the code to cover it is trivial and unlikely to introduce bugs so let's just have it.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Done. I'd rather kept a directory -- its name contains a clue to what to expect inside.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Done.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -118,3 +125,35 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+        BUNDLE_DEPLOYMENT: "true"
+        BUNDLE_NO_PRUNE: "true"
+    """
+    )
+    if potential_bundle_config.is_file():
+        config_data = potential_bundle_config.read_text()
+        config_data += hermetic_config
+    # Corner case: a user decides to define their own alternate config.
+    # In this scenario cachi2 must try to copy over user-defined variables
+    # to its overriding alternate config.
```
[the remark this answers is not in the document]
a-ovchinnikov: Done.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Good point, done.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: Done.

tests/unit/package_managers/bundler/test_main.py
a-ovchinnikov: One underscore to separate 'test' prefix from the name of the function, another one comes from the function name.

tests/unit/package_managers/bundler/test_main.py
a-ovchinnikov: Done.

tests/unit/package_managers/bundler/test_main.py
a-ovchinnikov: Done.

tests/unit/package_managers/bundler/test_main.py
a-ovchinnikov: Yes, absolutely, thank you!

tests/integration/test_bundler.py
a-ovchinnikov: Probably not, at least not with our current ITs structure. This scenario would essentially exercise the same code that is exercised in existing tests, but in a loop.

slimreaper35: One last thing. We don't have any e2e tests and I believe we could make one of those tests above (that are almost identical) to do full pre-fetch and build an image.

a-ovchinnikov: Good point, I'll add a e2e test in a separate commit to this PR (after it passes locally).

eskultety (approved):
e2e isn't a deal breaker for me at this point given there may be more work on the backend needed, so it'll have to be added by the time we flip the support switch on it. The rest of the changes look good -> ACK.

PR #656, opened 2024-09-17
Yarn 1.x: Prefetch dependencies

### LINKED ISSUE #626: Yarn 1.x: Prefetch dependencies
Prefetching dependencies for Yarn in Cachi2 will be done using Yarn's [offline mirror](https://classic.yarnpkg.com/blog/2016/11/24/offline-mirror/)
feature. When a project is configured to use the offline mirror, Yarn will store compressed archives in the mirror directory and can install cached project dependencies from there later without network access.

The offline mirror in the request output directory will be populated by running the following Yarn command in the project root:

    yarn install --non-interactive --frozen-lockfile --disable-pnp --no-default-rc --ignore-engines

Command: [yarn install](https://classic.yarnpkg.com/en/docs/cli/install)

Options:
 - [non-interactive](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-non-interactive): Disable interactive prompts
 - [frozen-lockfile](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-frozen-lockfile): Fail if yarn.lock updates are needed
 - [disable-pnp](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L84): Disable Plug'n'Play installation
 - [no-default-rc](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L73): Prevent Yarn from automatically detecting .yarnrc and .npmrc files
 - [ignore-engines](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-ignore-engines): Ignore the engine node version during the prefetch

 The behavior of Yarn can also be configured using select environment variables. The following ones should be set for the prefetch:

 - COREPACK_ENABLE_DOWNLOAD_PROMPT=0
 - COREPACK_ENABLE_PROJECT_SPEC=0
 - YARN_YARN_OFFLINE_MIRROR=\<absolute path to the request output directory\>
 - YARN_YARN_OFFLINE_MIRROR_PRUNING=false
 - [YARN_IGNORE_PATH](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L627)=true (Use the corepack yarn version, not any Yarn version specified by the user)
 - [YARN_IGNORE_SCRIPTS](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/config.js#L415)=true

Acceptance Criteria
- Project dependencies are prefetched into the `fetch-deps` command output directory for a Yarn 1.x project
- Settings for the prefetch are implemented as described above
- Add integration test coverage
  - Ensure that Yarn 1.x is being used by corepack
  - Ensure that multiple input packages are covered
  - ~~Ensure that the request fails if yarn needs to modify yarn.lock~~ See https://github.com/containerbuildsystem/cachi2/issues/655
  - Ensure that the request fails if yarn detects a checksum mismatch
  - Ensure that scripts are not executed by Yarn (or file a separate issue for this test)

taylormadore:
Prefetch Yarn 1.x dependencies using Yarn's [offline mirror](https://classic.yarnpkg.com/blog/2016/11/24/offline-mirror/)

Resolves #626 


cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -16,3 +81,27 @@ def _generate_build_environment_variables() -> list[EnvironmentVariable]:
     }
 
     return [EnvironmentVariable(name=key, value=value) for key, value in env_vars.items()]
+
+
+def _verify_corepack_yarn_version(source_dir: RootedPath, env: dict[str, str]) -> None:
```
taylormadore, line 86: I'd like to have one shared function for both yarn modules to check the Yarn version and ensure it is within some range of acceptable values. I started experimenting with using [semantic_version](https://github.com/rbarrois/python-semanticversion) instead of semver since they have a nice [SimpleSpec](https://github.com/rbarrois/python-semanticversion#using-a-specification) class that does what I was looking for.

If others like that idea, I can continue with it

eskultety: > I started experimenting with using [semantic_version](https://github.com/rbarrois/python-semanticversion) instead of semver since they have a nice [SimpleSpec](https://github.com/rbarrois/python-semanticversion#using-a-specification) class that does what I was looking for.

The project hasn't made a release in 2 years, seems like it has a single maintainer and when I look at PRs, issues, not much has been done really in the past year. For what it seems you need to do, pulling in another depedency seems like an overkill.
IIUC you want to be able to do:
```
Version("1.2.3") in SimpleSpec(">=1.0.6", "<4")
```

I'm not a fan of depending on the semver library in general for what we actually need to check, given the python's packaging version handling seems to cover everything semver does and some more (Yes, I know, it's specific to Python packaging only, but admittedly our use case touches only the surface). So, I would propose having a look at https://packaging.pypa.io/en/stable/specifiers.html#packaging.specifiers.SpecifierSet.__contains__  which is a generic class that seems to be doing the same thing as `SimpleSpec` and supports standard Python comparison operators, have a look at the examples if you think it covers your use case.

taylormadore: I actually started off using packaging, but cut that work short due to the fact that it [didn't appear to be strictly compatible with semver](https://packaging.python.org/en/latest/specifications/version-specifiers/#semantic-versioning) and the [version field in package.json can be any valid semver](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#version). I can take another look at it and see if it can be enough for what we need though :+1:

eskultety: > I actually started off using packaging, but cut that work short due to the fact that it [didn't appear to be strictly compatible with semver](https://packaging.python.org/en/latest/specifications/version-specifiers/#semantic-versioning) and the [version field in package.json can be any valid semver](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#version).

We shouldn't depend on pre-release and local build deps,  especially for pkg mgrs that should take care of the differences, right? We're not intending to represent valid user deps with what I'm suggesting, so I think we should be fine.

taylormadore: I might not quite understand what you mean with the last comment. Do you mean that we should ignore the prerelease and build parts of the semver? They're definitely edge cases, but we currently do not have a restriction on the user specifying an rc for Yarn v2+.

I think the packaging SpecifierSet would work well for the comparison, but I think the parsing should be done with semver, since the input is a semver. It should be possible to translate one to the other fairly easily.

eskultety: > They're definitely edge cases, but we currently do not have a restriction on the user specifying an rc for Yarn v2+

I totally neglected this aspect, I'm sorry.

> I think the packaging SpecifierSet would work well for the comparison, but I think the parsing should be done with semver, since the input is a semver. It should be possible to translate one to the other fairly easily.

It's the pre-release and dev releases that need to be translated though, correct?

taylormadore: Yeah, the prerelease part. I think we might be able to ignore the build part since according to the semver spec it should not be used for determining precedence.

The tricky thing appears to be that packaging is a lot less permissive when it comes to the contents of the prerelease part. Valid semver prereleases don't equal valid packaging prereleases.

eskultety: > The tricky thing appears to be that packaging is a lot less permissive when it comes to the contents of the prerelease part. Valid semver prereleases don't equal valid packaging prereleases.

Yes, I've seen the examples. But I wonder how it impacts us. Even if users are allowed to specify Yarn prerelease versions, all we care about are the final releases isn't it?
So if we define the spec as `(">=1.6.2, <4")` then what is preventing us to just check the versions simply based on  `major.minor.micro` and iff the input prerelease semver falls into the range then check the respective range boundaries for equality with the input semver (without prerelease) and if matched, deny it. Otherwise it means that the semver string (including prereleases) is well within the range and hence we don't care. In that case we don't even need to translate anything.

tests/unit/package_managers/conftest.py
```diff
@@ -6,6 +6,7 @@
 import git
 import pytest
 
+from cachi2.core.models.input import Request
```
eskultety, line 8: We won't have this file after #657, so you need to rebase and move it to the global conftest.

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -1,4 +1,16 @@
...
+from cachi2.core.models.output import Component, EnvironmentVariable, RequestOutput
+
+
+def fetch_yarn_source(request: Request) -> RequestOutput:
+    """Process all the yarn source directories in a request."""
+    components: list[Component] = []
+    env_vars: list[EnvironmentVariable] = []
+
+    for package in request.yarn_classic_packages:
+        pass
```
eskultety: nitpick: this should be added by future patch.

eskultety (commented):
This is solid, LGTM.

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -1,18 +1,60 @@
 from cachi2.core.models.input import Request
 from cachi2.core.models.output import Component, EnvironmentVariable, RequestOutput
+from cachi2.core.package_managers.yarn.utils import run_yarn_cmd
```
eskultety, line 8: **Commit message:**

nitpick: I'd also probably reference #626 from here for completeness and have a record in the git history directly even though you have it linked in the PR.

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -16,3 +81,27 @@ def _generate_build_environment_variables() -> list[EnvironmentVariable]:
...
+
+    try:
+        installed_yarn_version = semver.version.Version.parse(yarn_version_output)
+    except ValueError as e:
+        raise PackageManagerError(
+            "Cachi2 expected corepack to install yarn >=1.22.0,<2.0.0, but "
+            "the command `yarn --version` did not return a valid semver."
+        ) from e
+
+    min_version = semver.version.Version(1, 22, 0)
```
slimreaper35, line 98:
```suggestion
min_version_inclusive = semver.version.Version(1, 22, 0)
```

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -16,3 +81,27 @@ def _generate_build_environment_variables() -> list[EnvironmentVariable]:
...
+
+def _verify_corepack_yarn_version(source_dir: RootedPath, env: dict[str, str]) -> None:
+    """Verify that corepack installed the correct version of yarn by checking `yarn --version`."""
+    yarn_version_output = run_yarn_cmd(["--version"], source_dir, env=env).strip()
+
+    try:
+        installed_yarn_version = semver.version.Version.parse(yarn_version_output)
+    except ValueError as e:
+        raise PackageManagerError(
+            "Cachi2 expected corepack to install yarn >=1.22.0,<2.0.0, but "
```
a-ovchinnikov, line 94: Does the version check actually happen here? It looks like the only thing we can be sure about at this point is that `The command 'yarn --version' did not return a valid semver.")

taylormadore: Good point, updated :+1:

eskultety (approved):

a-ovchinnikov (approved):

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -1,4 +1,69 @@
...
+
+
+def _fetch_dependencies(source_dir: RootedPath, env: dict[str, str]) -> None:
+    """Fetch dependencies using 'yarn install'.
+
+    :param source_dir: the directory in which the yarn command will be called.
+    :param env: environment variable mapping used for the prefetch.
+    :raises PackageManagerError: if the 'yarn install' command fails.
+    """
+    run_yarn_cmd(
```
brunoapimentel, line 43: I believe it would be useful to add a `log.debug` that prints the exact command that's being invoked, similar to what we do with `gomod.Go._run`.

I know this is a shared function, so if you prefer, let's do it as a follow up.

brunoapimentel (commented):
Looks great! :taco: 

I was able to prefetch and build the [argo-cd ui](https://github.com/argoproj/argo-cd/tree/master/ui) with this (only the ui part).

I'll give a more nitpickish look tomorrow, and then my final ack.

slimreaper35 (approved):

cachi2/core/package_managers/yarn_classic/main.py
taylormadore: Filed https://github.com/containerbuildsystem/cachi2/issues/660 for this

PR #659, opened 2024-09-20
bundler: SBOM components

slimreaper35:


slimreaper35:
I am not sure if this is an ideal solution, mainly because there is always a call for `get_repo_id` for each path dependency. But I wanted to have some feedback early on.

tests/integration/test_bundler.py
```diff
@@ -0,0 +1,75 @@
...
+log = logging.getLogger(__name__)
+
+
+@pytest.mark.parametrize(
+    "test_params",
+    [
+        pytest.param(
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-bundler.git",
+                ref="ad1778cf87f45838330d63743a13fc33daa4f376",
```
a-ovchinnikov, line 17: Why did you choose to directly refer to a commit hash instead of using branch name? The [relevant part](https://github.com/containerbuildsystem/cachi2/blob/main/tests/integration/utils.py#L173) of utils seems to just check out the reference. This should work with branch names too. If someone discovers an issue in test repo they would need to not just fix it, but to also update these reference.

a-ovchinnikov (commented):
Looks straightforward and mostly LGTM with the exception for my comment in tests.

tests/integration/test_bundler.py
slimreaper35: :thinking: that's a good point; we use commit hash everywhere, so that's why I did likewise.
(There are a lot of things within integration tests that should be refactored)

slimreaper35: I would replace the e2e test for **well_formed_ruby_without_gemspec** for a multi-package test, but it depends on how we resolve conversations in #654.

https://github.com/cachito-testing/cachi2-bundler/tree/main

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -155,16 +199,6 @@ def parse_lockfile(package_dir: RootedPath) -> ParseResult:
         elif dep["type"] == "git":
             result.append(GitDependency(**dep))
         elif dep["type"] == "path":
-            _validate_path_dependency_subpath(package_dir, dep["path"])
-            result.append(PathDependency(**dep))
+            result.append(PathDependency(**dep, root=package_dir))
```
a-ovchinnikov, line 202: This cleanup is great!

a-ovchinnikov (approved):

slimreaper35:
When testing locally, I noticed it would not work with nested packages with PATH dependencies. I have a `tmp/one` bundler package and `tmp/two` bundler package. It fails with: `InvalidGitRepositoryError: ~/tmp/two` where I added one PATH dependency. A possible fix would be to add `search_parent_directories=True` to the `get_repo_id` function, but I wonder if there is a better solution...

```
$ cachi2 fetch-deps --dev-package-managers --source ../tmp '[{"path": "one", "type": "bundler"}, {"path": "two", "type": "bundler"}]'
```

a-ovchinnikov (approved):

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -8,8 +8,9 @@
 
 import pydantic
 from git import Repo
+from typing_extensions import Self
```
eskultety, line 12: **Commit message**:
I'd argue that the main motivation here is to run the validation as part of the model rather than externally and so adding the attribute is just the necessary side effect to get us there. In context of the changes themselves it reads a bit weird, so please reword it accordingly.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -84,7 +84,7 @@ def _get_name_and_version_from_lockfile(dependencies: ParseResult) -> Optional[t
     https://github.com/containerbuildsystem/cachi2/blob/main/docs/design/bundler.md
     """
     for dep in dependencies:
-        if isinstance(dep, PathDependency) and dep.path == ".":
+        if isinstance(dep, PathDependency) and dep.subpath == ".":
```
eskultety, line 108: This renaming change should be squashed into commit https://github.com/containerbuildsystem/cachi2/pull/659/commits/c4763e53121bbdd5db5ce69b62928738be679e2f and just listed as one of the renamings done by the patch.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -133,6 +148,20 @@ def validate_subpath(self) -> Self:
 
         return self
 
+    @cached_property
+    def purl(self) -> str:
+        """Get PURL for this dependency."""
+        vcs_url = get_repo_id(self.root.path).as_vcs_url_qualifier()
+        qualifiers = {"vcs_url": vcs_url}
```
eskultety, line 155: nitpick: you can use the dict directly on the call below (note: you're doing exactly that in the next commit in this series ;) ).

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -1,10 +1,77 @@
...
+from cachi2.core.package_managers.bundler.parser import (
+    GemDependency,
+    GitDependency,
+    ParseResult,
+    PathDependency,
+)
 from cachi2.core.rooted_path import RootedPath
+from tests.common_utils import GIT_REF
+
+
```
eskultety, line 20: I'm missing a unit test verifying the PURL format, we have one like that (in context of creating an SBOM component) for `gomod` and `rpm` (likely other backends as well).

eskultety (commented):
The only major comment I have is absence of testing the PURL formatting in context of SBOM component creation, the rest are nitpicks that can be resolved along the way.

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -1,10 +1,92 @@
...
+        == f"pkg:gem/my-git-dep@0.1.0?vcs_url=git%2Bhttps://github.com/rubygems/example.git%40{GIT_REF}"
+    )
+    assert (
+        components[2].purl
+        == f"pkg:gem/my-path-dep@0.1.0?vcs_url=git%2Bssh://git%40github.com/user/repo.git%40{repo_commit.hexsha}#vendor"
+    )
+    assert (
+        components[3].purl
+        == f"pkg:gem/name?vcs_url=git%2Bssh://git%40github.com/user/repo.git%40{repo_commit.hexsha}"
+    )
```
eskultety, line 89: This is unnecessarily difficult to read and understand. You can define a dedicated test case just for generating components with each type of component being an element of a parametrize. That way, you can directly stuff the expected PURL in the expected results and compare against that. It's going to be much more readable than this and not many lines will be needed for that (most lines will be "wasted" on component declaration in parametrize. Like I mentioned in my previous review, `test_generate_sbom_components` and `test_package_to_component` (rpm, gomod) can serve as an example here.

PR #663, opened 2024-09-24
Add design for fetching maven artifacts

kosciCZ:
Hey folks, this is an followup to #652, which extends generic fetching with the support for maven artifacts. I've tried to roughly outline how this will work and how it will integrate with other cachi2 features. Let me know what you think. 

### DIFF
--- a/docs/design/maven.md
+++ b/docs/design/maven.md
@@ -1,0 +1,141 @@
+# Maven artifact fetching
+
+## Introduction
+
+This document will describe high-level implementation overview for supporting fetching maven artifacts in cachi2. This
+proposal depends on generic artifact fetching, which is described in this [design PR][generic-pr].
+
+## Context
+
+[Maven] is a Java ecosystem build system and functions basically as a package manager, among other things. Maven provides a
+standardized way to specify dependencies in a `pom.xml` file, which is then used to fetch dependencies from a maven repository.
+This feature in cachi2 isn't concerned with pre-fetching entire maven projects, but rather to provide a way to fetch individual
+artifacts from maven repository. This is useful for cases where the build only needs specific artifacts from big projects,
+saving time and bandwidth.
+
+### Maven artifacts
+
+[Maven artifacts][maven-artifacts] are defined as anything that resides in a maven repository. Maven has a special way of
+addressing artifacts in its repositories called coordinates, which are usually represented by a `GAV` tuple (stands for
+'Group, Artifact, Version').
+This is used to uniquely identify an artifact in a maven repository.
+
+### Maven purls
+
+Purls for `pkg:maven` ([docs][maven-purl]) fully describe the location of an artifact in a maven repository. That is because
+all parts of the purl (`namespace`, `name`, `version` and the `classifier` and `type` qualifiers) are used to construct
+the download url for the artifact and map directly to tge `GAV` coordinates.[Maven documentation][maven-repo-layout] describes
+how to turn these coordinates into an actual download url. Here's an example:
+
+**Purl:**
+
+```
+pkg:maven/ga.io.quarkus/quarkus-core@3.8.5.redhat-00004?type=jar&repository_url=https://maven.repository.redhat.com&checksums=sha1:e4ca5fadf89e62fb29d0d008046489b2305295bf
+```
+
+**Mapping to GAV:**
+
+```
+groupId: ga.io.quarkus
+artifactId: quarkus-core
+version: 3.8.5.redhat-00004
+classifier: null
+type: jar
+```
+
+Using `classifier` and `type` we can determine the `extension` is `jar` ([docs][maven-extension]).
+
+**Download url**:
+
+Template:
+
+```
+${groupId as directory}/${artifactId}/${baseVersion}/${artifactId}-${version}.${extension}
+```
+
+Resulting url:
+
+```
+https://maven.repository.redhat.com/ga/io/quarkus/quarkus-core/3.8.5.redhat-00004/quarkus-core-3.8.5.redhat-00004.jar
+```
+
+## Design
+
+In this section, I will try to cover individual parts of the design.
+
+### Source repository (cachi2 lockfile)
+
+Currently, the cachi2 lockfile only allows `pkg:generic` purl type. This design will introduce a new purl type `pkg:maven`
+to support pre-fetching individual maven artifacts. The lockfile currently supports specifying the file to be fetched either
+as a purl or as a pair of `download_url` and `checksums`. Maven artifacts should only be specified as a purl in order to
+be reported as such in the output SBOM. Here's an example of a lockfile that fetches a maven artifact:
+
+```yaml
+metadata:
+  version: 1.0.0
+artifacts:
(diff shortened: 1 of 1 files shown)


docs/design/maven.md
```diff
@@ -0,0 +1,130 @@
+# Generic artifact fetching
```
a-ovchinnikov, line 1: Should it be "Maven"?

docs/design/maven.md
```diff
@@ -0,0 +1,130 @@
...
+
+This document will describe high-level implementation overview for supporting fetching maven artifacts in cachi2. This
+proposal depends on generic artifact fetching, which is described in this [design PR][generic-pr].
+
+## Context
+
+[Maven] is a Java ecosystem build system and functions basically as a package manager, among other things. Maven provides a
+standardized way to specify dependencies in a `pom.xml` file, which is then used to fetch dependencies from a maven repository.
+This feature in cachi2 isn't concerned with pre-fetching those, but rather to provide a way to fetch individual artifacts
+from maven repository. This is useful for cases where the build only needs specific artifacts from big projects, saving
```
a-ovchinnikov, line 13: Do you mean that instead of pre-fetching entire projects the plan is to pre-fetch just some parts of these already built projects? I am asking as someone with a very limited maven experience.

docs/design/maven.md
```diff
@@ -0,0 +1,130 @@
...
+time and bandwidth.
+
+### Maven artifacts
+
+[Maven artifacts][maven-artifacts] are defined as anything that resides in a maven repository. Maven has a special way of
+addressing artifacts in its repositories called coordinates, which is usually represented by `GAV` (Group, Artifact, Version)
```
a-ovchinnikov, line 19:
```suggestion
addressing artifacts in its repositories called coordinates, which are usually represented by `GAV` tuple (`GAV` standing for 'Group, Atifact, Version').
```
(if I read the paragraph correctly). Are Group and Artifact names? From the paragraph below it might seem that this tuple must also contain 'type' and 'classifier' fields, are those missing?

docs/design/maven.md
```diff
@@ -0,0 +1,130 @@
...
+This is used to uniquely identify an artifact in a maven repository.
+
+### Maven purls
+
+Purls for `pkg:maven` ([docs][maven-purl]) fully describe the location of an artifact in a maven repository. That is because
+all parts of the purl (`namespace`, `name`, `version` and the `classifier` and `type` qualifiers) are used to construct
+the download url for the artifact. [Maven documentation][maven-repo-layout] describes how to turn these coordinates into
```
a-ovchinnikov, line 26: This is a bit confusing. Should purls be translated to GAVs or the other way around?

docs/design/maven.md
```diff
@@ -0,0 +1,130 @@
...
+
+```yaml
+metadata:
+  version: 1.0.0
+artifacts:
+  - purl: pkg:maven/ga.io.quarkus/quarkus-core@3.8.5.redhat-00004?type=jar&repository_url=https://maven.repository.redhat.com&checksums=sha1:e4ca5fadf89e62fb29d0d008046489b2305295bf
+    target: quarkus.jar
+```
+
+### Integration with existing cachi2 features
```
a-ovchinnikov, line 80: Would be nice to see a command a user would need to run to use this feature.

docs/design/maven.md
kosciCZ: Correct, that's a typo. Updated.

docs/design/maven.md
kosciCZ: Yes, that's correct. This is analogous to how OSBS works, here's how they used to [specify](https://osbs.readthedocs.io/en/osbs_ocp3/users.html?highlight=fetch#fetch-artifacts-pnc-yaml) the artifacts to fetch. This feature is similar, but much more generic, and reusable outside of one specific usecase.

docs/design/maven.md
kosciCZ: Adjusted the wording. This is a maven quirk, while there's more to it than just the three, the name stuck. [Docs ref](https://maven.apache.org/repositories/artifacts.html)

> Artifact coordinates, are most often represented as groupId:artifactId:version, or GAV in short or when informally used (please note that Artifact coordinates has more fields, but for brevity we still call the coordinates “GAV”, not “GAVCE”).

docs/design/maven.md
kosciCZ: Added.

docs/design/maven.md
a-ovchinnikov: Thank you. Then I suggest replacing "those" with "entire maven projects" for clarity.

docs/design/maven.md
a-ovchinnikov: I suggest adding this explanation to this doc, this is important for understanding the specifics of the proposed package manager.

docs/design/maven.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### 1) Completely encapsulated in generic package manager
+
+This option would just extend the lockfile with support for maven purls, meaning the user would still be calling just the
+generic package manager. The benefit of this approach would be simplicity, as it would one package manager per lockfile.
+On the other hand, maven artifacts are not so generic.
+
+Example:
+```
+cachi2 fetch-deps --source /path/to/sources '[{"type": "generic"}]'
```
a-ovchinnikov, line 97: Thank you for the example! There is no mention of a lock file here. Does this mean that a lock file will always reside in same location? Is there an agreement on where to put it and who would do that?

docs/design/maven.md
```diff
@@ -0,0 +1,141 @@
...
+    target: quarkus.jar
+```
+
+### Integration with existing cachi2 features
+
+As mentioned above, the lockfile format is identical to the one used for describing generic artifacts, therefore it makes
+sense to extend the existing lockfile with support for maven purls. The only difference would be in the output SBOM where
+maven artifacts would be reported as `pkg:maven`, instead of `pkg:generic`.
+
+There is an open question of how the UX should be. I will try to list the available options with some pros and cons.
```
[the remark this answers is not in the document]
kosciCZ: In my opinion, the part that changes is:
- how you compute the download url (meaning, the list of accepted purl types _might_ be expanded later)
- what the output SBOM component is (depends on the above) 

So I'd argue that since the underlying mechanism does not change, it should stay as one package manager.

docs/design/maven.md
kosciCZ: Added.

docs/design/maven.md
kosciCZ: Replaced.

docs/design/maven.md
kosciCZ: At this time, yes. The generic lockfile is assumed to be present in the root of the cloned input repository. Here's a [link to the code](https://github.com/containerbuildsystem/cachi2/pull/681/files#diff-e65d82e684cdd28151a5ddccb300c06882c33725bc751263d1ffb3a0f1b878eaR14)

a-ovchinnikov: Who will put it there?

kosciCZ: Not sure I understand where this question is going. It should be commited directly in the repository. An user that wants to use cachi2's maven artifact feature would do that

kosciCZ:
Closing in favor of #692

PR #665, opened 2024-09-26
cachi2 output image

slimreaper35:

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -3,9 +3,6 @@ kind: PipelineRun
 metadata:
   name: cachi2-on-pull-request
   annotations:
-    build.appstudio.redhat.com/commit_sha: "{{revision}}"
-    build.appstudio.redhat.com/pull_request_number: "{{pull_request_number}}"
-    build.appstudio.redhat.com/target_branch: "{{target_branch}}"
     pipelinesascode.tekton.dev/max-keep-runs: "3"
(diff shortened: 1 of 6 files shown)


eskultety:
The idea here is (https://github.com/containerbuildsystem/cachi2/milestone/3) to stay independent which also means the main container registry we use for releases, which most likely ends up being GH's registry due to simplicity of use, integration with the platform and integrated RBAC based on accounts/organizations without the need for secrets hosting in external vaults not accessible to all potential maintainers.
With that done this PR will have resolve the merge conflicts it'll bring.

If we push to Konflux's Quay.io namespace as part of an action is fine with me ultimately since it's a consumer and wishes to consume upstream directly in their pipeline, but it ought not be our main release stream. That said, the one issue at hand we keep seeing with the Konflux build non-gating CI job is failures due to docker.io rate limiting (which I hope GH has resolved by massive caching which seems like the case looking at the workflows logs where images part of the integration test suite are often pulled out of a cache) and so we won't be able to continue building in Konflux the same way as we do today - I was thinking about using the `schedule` trigger [1] in actions and build the project nightly in Konflux to reveal potential problems early and not the moment we're ready to tag a release and publish a tagged container image.

With all that said, up to you if you want to keep the PR around (but the PR subject should be tweaked IMO) or close it for the time being and then open a fresh one tailored to the future situation.

[1] https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

Issue #630, opened 2024-09-04
Yarn 1.x: Parse package.json to gather workspace metadata


taylormadore:
Yarn 1.x yarn.lock files usually do not contain metadata for workspaces, so they will need to be parsed from package.json. Some inspiration for this change may be found in [Cachito's yarn module](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L95).

Some examples of how workspaces are specified and under what conditions they are added to the lockfile can be found here: https://github.com/cachito-testing/cachito-yarn-workspaces

Acceptance criteria
- Workspace dependency metadata is parsed from package.json into dataclasses
- Adapt the cachito [integration test for yarn workspaces](https://github.com/cachito-testing/cachito-yarn-workspaces)
- Ensure that path dependencies leading outside the repo are detected and an exception is raised

PR #666, opened 2024-09-26
Adding Yarn v1.x workspaces metadata extractor

a-ovchinnikov:
In Yarn v1.x workspaces metadata is stored outside of yarn.lock in package.json. This commit inroduces functions for workspaces metadata extraction and a dataclass to represent workspaces.

This addresses https://github.com/containerbuildsystem/cachi2/issues/630

### DIFF
--- a/cachi2/core/package_managers/yarn_classic/main.py
+++ b/cachi2/core/package_managers/yarn_classic/main.py
@@ -6,6 +6,7 @@ from cachi2.core.errors import PackageManagerError
 from cachi2.core.models.input import Request
 from cachi2.core.models.output import Component, EnvironmentVariable, RequestOutput
 from cachi2.core.package_managers.yarn.utils import run_yarn_cmd
+from cachi2.core.package_managers.yarn_classic.workspaces import extract_workspace_metadata
 from cachi2.core.rooted_path import RootedPath
 
 log = logging.getLogger(__name__)
@@ -27,6 +28,11 @@ def fetch_yarn_source(request: Request) -> RequestOutput:
         prefetch_env = _get_prefetch_environment_variables(request.output_dir)
         _verify_corepack_yarn_version(path, prefetch_env)
         _fetch_dependencies(path, prefetch_env)
+        # Workspaces metadata is not used at the moment, but will
+        # eventualy be converted into components. Using a noop assertion
+        # to prevent linters from complaining.
+        workspaces = extract_workspace_metadata(package, request.source_dir)
+        assert workspaces is not None  # nosec -- see comment above
 
     return RequestOutput.from_obj_list(
         components, _generate_build_environment_variables(), project_files=[]
--- a/cachi2/core/package_managers/yarn_classic/workspaces.py
+++ b/cachi2/core/package_managers/yarn_classic/workspaces.py
@@ -1,0 +1,23 @@
+import json
+from itertools import chain
+from pathlib import Path
+from typing import Any, Generator, Iterable
+
+import pydantic
+
+from cachi2.core.errors import PackageRejected
+from cachi2.core.models.input import YarnClassicPackageInput
+from cachi2.core.rooted_path import PathOutsideRoot, RootedPath
+
+
+class Workspace(pydantic.BaseModel):
+    """Workspace model."""
+
+    path: Path  # path to a workspace.
+    package_contents: dict  # package data extracted from path/"package.json".
+
+    @pydantic.field_validator("package_contents")
+    def _ensure_package_is_named(cls, package_contents: dict) -> dict:
+        if "name" not in package_contents:
+            raise ValueError("Workspaces must contain 'name' field.")
+        return package_contents
--- a/cachi2/core/package_managers/yarn_classic/workspaces.py
+++ b/cachi2/core/package_managers/yarn_classic/workspaces.py
@@ -21,3 +21,86 @@ class Workspace(pydantic.BaseModel):
         if "name" not in package_contents:
             raise ValueError("Workspaces must contain 'name' field.")
         return package_contents
+
+
+def ensure_no_path_leads_out(
+    paths: Iterable[Path],
+    source_dir: RootedPath,
+) -> None:
+    """Ensure no path leads out of source directory.
+
+    Raises an exception when any path is not relative to source directory.
+    Does nothing when path does not exist in the file system.
+    """
+    for path in paths:
+        try:
+            source_dir.join_within_root(path)
+        except PathOutsideRoot:
+            raise PackageRejected(
+                f"Found a workspace path which is not relative to package: {path}",
+                solution=(
+                    "Avoid using packages which try to access your filesystem "
+                    "outside of package directory."
+                ),
+            )
+
+
+def _ensure_workspaces_are_well_formed(
+    paths: Iterable[Path],
+) -> None:
+    """Ensure that every workspace contains package.json.
+
+    Reject the package otherwise.
+    """
+    for p in paths:
+        if not Path(p, "package.json").is_file():
+            raise PackageRejected(
+                reason=f"Workspace {p} does not contain 'package.json'",
+                solution=None,
+            )
+
+
+def _get_workspace_paths(
+    workspaces_globs: list[str],
+    source_dir: RootedPath,
+) -> Iterable[Path]:
+    """Resolve globs within source directory."""
+
+    def all_paths_matching(glob: str) -> Generator[Path, None, None]:
+        return (pth.resolve() for pth in source_dir.path.glob(glob))
+
+    return chain.from_iterable(map(all_paths_matching, workspaces_globs))
+
+
+def _extract_workspaces_globs(
+    package: dict[str, Any],
+) -> list[str]:
+    """Extract globs from workspaces entry in package dict."""
+    workspaces_globs = package.get("workspaces", [])
+    return workspaces_globs
+
+
+def _read_package_from(path: RootedPath) -> dict[str, Any]:
+    """Read package.json from a path."""
+    return json.loads(path.join_within_root("package.json").path.read_text())
+
+
+def extract_workspace_metadata(
+    package: YarnClassicPackageInput,
+    source_dir: RootedPath,
+) -> list[Workspace]:
+    """Extract workspace metadata from a package."""
+    processed_package = _read_package_from(source_dir.join_within_root(package.path))
+    workspaces_globs = _extract_workspaces_globs(processed_package)
+    workspaces_paths = _get_workspace_paths(workspaces_globs, source_dir)
+    ensure_no_path_leads_out(workspaces_paths, source_dir)
+    _ensure_workspaces_are_well_formed(workspaces_paths)
+    parsed_workspaces = []
+    for wp in workspaces_paths:
+        parsed_workspaces.append(
+            Workspace(
+                path=wp,
+                package_contents=_read_package_from(source_dir.join_within_root(wp)),
+            )
+        )
+    return parsed_workspaces
--- a/cachi2/core/package_managers/yarn_classic/workspaces.py
+++ b/cachi2/core/package_managers/yarn_classic/workspaces.py
@@ -76,7 +76,14 @@ def _extract_workspaces_globs(
     package: dict[str, Any],
 ) -> list[str]:
     """Extract globs from workspaces entry in package dict."""
+    # This could be an Array or an Array nested in an Object.
+    # Official docs mentioning the former:
+    #   https://classic.yarnpkg.com/lang/en/docs/workspaces/
+    # Official blog containing a hint about the latter:
+    #   https://classic.yarnpkg.com/lang/en/docs/workspaces/
     workspaces_globs = package.get("workspaces", [])
+    if isinstance(workspaces_globs, dict):
+        workspaces_globs = workspaces_globs.get("packages", [])
     return workspaces_globs
 
 
--- a/tests/unit/package_managers/yarn_classic/test_workspaces.py
+++ b/tests/unit/package_managers/yarn_classic/test_workspaces.py
@@ -1,0 +1,74 @@
+from pathlib import Path
+from unittest import mock
+
+import pytest
+
+from cachi2.core.errors import PackageRejected
+from cachi2.core.models.input import YarnClassicPackageInput
+from cachi2.core.package_managers.yarn_classic.workspaces import (
+    Workspace,
+    _extract_workspaces_globs,
+    _get_workspace_paths,
+    extract_workspace_metadata,
+)
+from cachi2.core.rooted_path import RootedPath
+
+
+@mock.patch("cachi2.core.package_managers.yarn_classic.workspaces._read_package_from")
+@mock.patch("cachi2.core.package_managers.yarn_classic.workspaces._get_workspace_paths")
+def test_packages_with_workspaces_outside_source_dir_are_rejected(
+    mock_get_ws_paths: mock.Mock,
+    mock_read_package_from: mock.Mock,
+) -> None:
+    package = YarnClassicPackageInput(type="yarn-classic", path=".")
+    mock_read_package_from.return_value = {"workspaces": ["../../usr"]}
+    mock_get_ws_paths.return_value = [Path("/tmp/foo/bar"), Path("/usr")]
+    source_dir = RootedPath("/tmp/foo")
+
+    with pytest.raises(PackageRejected):
+        extract_workspace_metadata(package, source_dir=source_dir)
+
+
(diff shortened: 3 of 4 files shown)


cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,98 @@
...
+            raise PackageRejected(
+                f"Found a workspace path which is not relative to package: {path}",
+                solution=(
+                    "Avoid using packages which try to access your filesystem "
+                    "outside of package directory."
+                ),
+            )
+
+
+def get_workspace_paths(
```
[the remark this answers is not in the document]
a-ovchinnikov: Good point, added a check.

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,98 @@
...
+    return json.loads(path.join_within_root("package.json").path.read_text())
+
+
+def extract_workspace_metadata(
+    package: YarnClassicPackageInput,
+    source_dir: RootedPath,
+) -> list[Workspace]:
+    """Extract workspace metadata from a package.
+
+    Currently does not deal with nested workspaces, however the way the code
```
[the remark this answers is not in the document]
eskultety: Is it? Yarn v1 is maintenance only so nested workspaces aren't happening and so we'd better not confuse the reader with the unlikely possibility of an unplanned feature in the ecosystem.

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,114 @@
...
+def get_workspace_paths(
+    workspaces_globs: list[str],
+    source_dir: RootedPath,
+) -> Iterable[Path]:
+    """Resolve globs within source directory."""
+
+    def all_paths_matching(glob: str) -> list[Path]:
+        return [pth.resolve() for pth in source_dir.path.glob(glob)]
+
+    return chain.from_iterable(all_paths_matching(g) for g in workspaces_globs)
```
eskultety, line 67: stupid nitpick: Consider `map` instead of comprehension since you have the helper defined already and you have to chain the results? Might 0.01% more idiomatic :)

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -27,6 +28,11 @@ def _ensure_mirror_dir_exists(output_dir: RootedPath) -> None:
         prefetch_env = _get_prefetch_environment_variables(request.output_dir)
         _verify_corepack_yarn_version(path, prefetch_env)
         _fetch_dependencies(path, prefetch_env)
+        # Workspaces metadata is not used at the moment, but will
+        # eventualy be converted into components. Using a noop assertion
+        # to prevent linters from complaining.
```
eskultety, line 33: I'd say either drop the direct usage for the time being, or, if you want to make sure we don't forget about it `_ = workspaces` feels cleaner than an assertion (admittedly I have a strong bias against assertions).

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def extract_workspaces_globs(
```
eskultety, line 70: Shouldn't all these extraction helpers be "private"?

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    for path in paths:
+        if not path.is_relative_to(source_dir.path):
+            raise PackageRejected(
+                f"Found a workspace path which is not relative to package: {path}",
+                solution=(
+                    "Avoid using packages which try to access your filesystem "
+                    "outside of package directory."
+                ),
+            )
```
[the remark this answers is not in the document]
taylormadore: Would the helper need more than what RootedPath already provides? I might be missing some nuance here

https://github.com/containerbuildsystem/cachi2/blob/c94a1a24c1777036bbed121281c41be1b8f156f1/cachi2/core/rooted_path.py#L111

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,114 @@
...
+from typing import Any, Iterable
+
+import pydantic
+
+from cachi2.core.errors import PackageRejected
+from cachi2.core.models.input import YarnClassicPackageInput
+from cachi2.core.rooted_path import RootedPath
+
+
+class Workspace(pydantic.BaseModel):
```
taylormadore, line 14: We should ensure that the `name` field is present in package.json for each workspace and fail otherwise

cachi2/core/package_managers/yarn_classic/workspaces.py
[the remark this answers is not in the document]
eskultety: Good catch! Unless we want to traverse the hierarchy, because RootedPath can only handle the immediate descendants, but not ancestors up to the real "root".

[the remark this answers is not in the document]
a-ovchinnikov: Done.

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,114 @@
...
+    return chain.from_iterable(all_paths_matching(g) for g in workspaces_globs)
+
+
+def extract_workspaces_globs(
+    package: dict[str, Any],
+) -> list[str]:
+    """Extract globs from workspaces entry in package dict."""
+    workspaces_globs = package.get("workspaces", [])
+    # This could be a list or a list in a dictionary. If it is not a dictionary
+    # then it is already a list that we need:
```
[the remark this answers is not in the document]
a-ovchinnikov: To be honest I trusted cachito implementation on this one. I'll expand the comment here and will try splitting everything out into a separate commit.

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,114 @@
...
+    with suppress(AttributeError):
```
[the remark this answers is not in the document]
a-ovchinnikov: Fair enough, this does attract too much attention.

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,114 @@
...
+        workspaces_globs = workspaces_globs.get("packages", [])
+    return workspaces_globs
+
+
+def read_package_from(path: RootedPath) -> dict[str, Any]:
+    """Read package.json from a path."""
+    return json.loads(path.join_within_root("package.json").path.read_text())
```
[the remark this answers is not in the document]
a-ovchinnikov: I don't think a model should be validating filesystem, so will keep this one unchanged for now.

cachi2/core/package_managers/yarn_classic/workspaces.py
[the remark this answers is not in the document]
a-ovchinnikov: Dropping everything that is related to nested workspaces.

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,114 @@
...
+    """Ensure that every workspace contains package.json.
+
+    Reject the package otherwise.
+    """
+    for p in paths:
+        if not Path(p, "package.json").is_file():
+            raise PackageRejected(
+                reason=f"Workspace {p} does not contain 'package.json'",
+                solution=None,
+            )
```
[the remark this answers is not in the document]
eskultety: > Checking that some file is present somewhere is not models responsibility. Someone else must do this

I'd argue that the model is to represent a workspace which *is defined** by the presence of a `package.json` file; no `package.json` file -> no package or workspace in this case.

Fair enough though, given the not-the-most-pleasant pydantic model interface when it comes to testing the validation, it's not a blocker for me -> resolving.

cachi2/core/package_managers/yarn_classic/workspaces.py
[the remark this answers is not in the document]
eskultety: > I don't think a model should be validating filesystem, so will keep this one unchanged for now.

Generally true, see my other comment on this though; resolving based on https://github.com/containerbuildsystem/cachi2/pull/666#discussion_r1806177005

cachi2/core/package_managers/yarn_classic/workspaces.py
eskultety: > Sorry, I don't follow why map would be more idiomatic? I need to join several flat iterables into anther flat iterable, chain.from_iterable does exactly that. To the best of my knowledge map cannot do this.

Yes, I only meant the comprehension parth:
`chain.from_iterable(map(all_paths_matching), workspaces_globs)`

> there does not seem to be any good justification for using map instead of GC.

It's slightly more readable at times is all :) , why do you think I didn't even dare to call this a "nitpick" but a stupid one?

cachi2/core/package_managers/yarn_classic/main.py
eskultety: Still feels weird to have in the code like that at any point in time, but okay, not a blocker.

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,23 @@
+import json
```
eskultety, line 1: **Commit subject:**

s/yar/yarn

eskultety: **Commit message:**

To make it more useful please mention what constitutes a workspace which lead us to this representation.

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -21,3 +21,86 @@ def _ensure_package_is_named(cls, package_contents: dict) -> dict:
...
+            raise PackageRejected(
+                f"Found a workspace path which is not relative to package: {path}",
+                solution=(
+                    "Avoid using packages which try to access your filesystem "
+                    "outside of package directory."
+                ),
+            )
+
+
+def ensure_workspaces_are_well_formed(
```
eskultety, line 48: A private helper maybe?

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -0,0 +1,23 @@
...
+
+from cachi2.core.errors import PackageRejected
+from cachi2.core.models.input import YarnClassicPackageInput
+from cachi2.core.rooted_path import PathOutsideRoot, RootedPath
+
+
+class Workspace(pydantic.BaseModel):
+    """Workspace model."""
+
+    path: Path  # path to a workspace.
```
eskultety, line 16: There's missing a validator for the path if it even exists as part of the project.

cachi2/core/package_managers/yarn_classic/workspaces.py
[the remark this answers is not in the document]
eskultety: > A model represents some structured data, its responsibility is to make sure that all necessary fields were supplied.

Actually, since you're missing a validator for the `path` attribute, you might as well extend the model with a `model_validator(after)` one that checks `package.json` presence as it's closely related.
I have found 
- https://github.com/containerbuildsystem/cachi2/blob/c94a1a24c1777036bbed121281c41be1b8f156f1/cachi2/core/models/input.py#L70
- https://github.com/containerbuildsystem/cachi2/blob/c94a1a24c1777036bbed121281c41be1b8f156f1/cachi2/core/models/input.py#L154
- [most recent one] https://github.com/containerbuildsystem/cachi2/blob/c94a1a24c1777036bbed121281c41be1b8f156f1/cachi2/core/models/input.py#L88

which are similar but definitely not trying to do a deep file system checking. So, either we've already broken good data-logic separation there as well or we could do something similar and a bit more here.

tests/unit/package_managers/yarn_classic/test_workspaces.py
```diff
@@ -0,0 +1,83 @@
...
+    assert expected == result
+
+
+def test_extracting_workspace_globs_works_with_glons_defined_in_dict() -> None:
+    package = {"workspaces": {"packages": ["foo"]}}
+
+    expected = ["foo"]
+    result = _extract_workspaces_globs(package)
+
+    assert expected == result
```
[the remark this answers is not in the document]
a-ovchinnikov: Good point. Done.

cachi2/core/package_managers/yarn_classic/workspaces.py
a-ovchinnikov: Oh, I see it now! While I am not 100% agree with the "more readable" part I'll give it a try.

cachi2/core/package_managers/yarn_classic/workspaces.py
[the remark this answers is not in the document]
a-ovchinnikov: I am sorry, but mode="after" is useless: package.json is used to instantiate a Workspace, moreover, there is a validator that makes sure that package.json contains a name. By the time we get to "after" we have already proven that all paths are correct. 'path' field is validated to be a correct Path-like string though, and we also check that no path leads out of the model by checking that it is joinable within the root. Moreover, testing any filesystem layout is still not a job for a model, it becomes bound to a specific filesystem otherwise. It is already a little mixed up by having package.json contents in it, I don't think it should be bound to a FS tighter than that.

cachi2/core/package_managers/yarn_classic/workspaces.py
a-ovchinnikov: Done

cachi2/core/package_managers/yarn_classic/workspaces.py
a-ovchinnikov: Validation will be happening elsewhere (namely when we attempt to read package.json to create an instance). A validator here would introduce coupling with filesystem.

tests/unit/package_managers/yarn_classic/test_workspaces.py
```diff
@@ -0,0 +1,92 @@
...
+        ),
+    ],
+)
+def test_extracting_workspace_globs_works_for_all_types_of_workspaces(
+    package: dict,
+    expected: list,
+) -> None:
+    package = {"workspaces": ["foo"]}
+
+    expected = ["foo"]
```
eskultety, line 80: You need to actually make use of the parametrize values :).

a-ovchinnikov: What? Are you saying no sneaky overrides to a value that will always pass the test are allowed?!

eskultety (approved):

cachi2/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -21,3 +21,86 @@ def _ensure_package_is_named(cls, package_contents: dict) -> dict:
...
+    source_dir: RootedPath,
+) -> None:
+    """Ensure no path leads out of source directory.
+
+    Raises an exception when any path is not relative to source directory.
+    Does nothing when path does not exist in the file system.
+    """
+    for path in paths:
+        try:
+            source_dir.join_within_root(path)
```
taylormadore, line 37: There may be a subtle path-handling issue here. The source_dir is a RootedPath where the root is the root of the source git repository. In this case, I believe we want to be testing against a RootedPath where the root is the package_path (which I believe is the [workspace root](https://classic.yarnpkg.com/lang/en/docs/workspaces/#toc-limitations-caveats) referenced in the docs).

We may want to pass the package_root to extract_workspace_metadata rather than the input package and source_dir. We could then use the re_root method of the RootedPath class to set the appropriate root for the check.

Something like this:
```python
def extract_workspace_metadata(package_path: RootedPath) -> list[Workspace]:
    """Extract workspace metadata from a package."""
    processed_package = _read_package_from(package_path)
    workspaces_globs = _extract_workspaces_globs(processed_package)
    workspaces_paths = _get_workspace_paths(workspaces_globs, package_path)
    ensure_no_path_leads_out(workspaces_paths, package_path.re_root(package_path.subpath_from_root))
    _ensure_workspaces_are_well_formed(workspaces_paths)
    parsed_workspaces = []
    for wp in workspaces_paths:
        parsed_workspaces.append(
            Workspace(
                path=wp,
                package_contents=_read_package_from(package_path.join_within_root(wp)),
            )
        )
    return parsed_workspaces
```

taylormadore: Not a blocker for me, since this is not really a risk to cachi2 and more of a Yarn install/build-time issue

taylormadore (approved):

Issue #672, opened 2024-10-03
Bundler PM does not prefetch platfrom-specific gems


a-ovchinnikov:
When a project depends on gems which are platform-specific or depend on platform-specific gems (e.g. [nokogiri](https://rubygems.org/gems/nokogiri) or [sorbet](https://rubygems.org/gems/sorbet) the necessary gems do not get prefetched which results in hermetic builds failures.

PR #128, opened 2023-04-04
Add Cachi2 Logging Standards

fepas:
A better-defined logging format will help users to better understand the status of requests.
INFO messages should be enough to let the user know what is happening with a request.

# Maintainers will complete the following section

- [ ] Commit messages are descriptive enough
- [ ] Code coverage from testing does not decrease and new code is covered
- [ ] New code has type annotations
- [ ] Docs updated (if applicable)
- [ ] Docs links in the code are still valid (if docs were updated)

PR #673, opened 2024-10-03
bundler: Support prefetching binary dependencies

a-ovchinnikov:
Some dependencies contain architecture-specific components.  Their name and location differ from standard Ruby-only dependencies. As a result prior to this commit such dependencies would be skipped during prefetch. This, in turn, failed hermetic builds because a dependency would end up missing.

Resolves: https://github.com/containerbuildsystem/cachi2/issues/672


tests/integration/test_data/bundler_everything_present/container/Containerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN ls /tmp/bundler_everything_present-output/deps/bundler
+
+# Check content of source repository folder
+RUN ls /tmp/bundler_everything_present-source/
+
+# This should be a COPY, but the source code and Containerfile are in different directories
+RUN cp -r /tmp/bundler_everything_present-source /src
+
+WORKDIR /src
+# Bundler would try and install whichever version was used to generate
```
a-ovchinnikov, line 19: `... used to generate` what? End of sentence is missing.

tests/integration/test_data/bundler_everything_present/container/Containerfile
```diff
@@ -0,0 +1,25 @@
...
+# This will cause bundler to attempt to download an earlier version even if
+# just microversions diverged. This, int turn, would cause a build to fail.
```
a-ovchinnikov, line 21: s/int/in/

cachi2/core/package_managers/bundler/main.py
```diff
@@ -30,7 +36,11 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
     for package in request.packages:
         path_within_root = request.source_dir.join_within_root(package.path)
         components.extend(
-            _resolve_bundler_package(package_dir=path_within_root, output_dir=request.output_dir)
+            _resolve_bundler_package(
+                package_dir=path_within_root,
+                output_dir=request.output_dir,
+                allow_binary=cast(BundlerPackageInput, package).allow_binary,
```
slimreaper35, line 42: This is unnecessarily difficult to read, I forgot to use `request.bundler_packages` instead of `request.packages` when adding the function. Then `allow_binary=package.allow_binary` should be fine.

cachi2/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -11,7 +11,9 @@
 lockfile_parser.specs.each do |spec|
     parsed_spec = {
       name: spec.name,
-      version: spec.version.to_s
+      version: spec.version.to_s,
+      full_name: spec.full_name,
```
slimreaper35, line 15: I don't think we need `full_name`, it can be easily constructed from `name` and `version` as you have in the code.
That said, I wonder if we can move the platform attribute to the `GemDependency` class and make it default to ruby (for example). I see a lot of duplication in both classes.

+ When I look at ~our favorite~ PURL spec [doc](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#gem) I see a platform qualifier. The default is `ruby`.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -194,7 +222,23 @@ def parse_lockfile(package_dir: RootedPath) -> ParseResult:
...
+                    log.warning(
+                        "Downloading binary dependency %s because 'allow_binary' is set to True",
+                        dep["full_name"],
+                    )
+                    result.append(GemPlatformSpecificDependency(**dep))
+                else:
+                    # No need to force a platform if we skip the packages.
+                    log.warning(
+                        "Skipping binary dependency %s because 'allow_binary' is set to False."
+                        " This will likely result in an unbuildable package.",
```
slimreaper35, line 237: I would argue, that this is not true, it is a rare scenario when a gem does not ship a source distribution.
We only found one during the development.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -80,6 +80,32 @@ def download_to(self, deps_dir: RootedPath) -> None:
...
+        """Download represented gem to specified file system location."""
+        fs_location = deps_dir.join_within_root(
+            Path(f"{self.name}-{self.version}-{self.platform}.gem")
+        )
+        log.info(
+            "Downloading platform-specific gem %s-%s-%s", self.name, self.version, self.platform
+        )
+        download_binary_file(self.remote_location, fs_location)
+
+
```
slimreaper35, line 108: See my previous comment about duplication.

tests/integration/test_bundler.py
```diff
@@ -100,3 +100,76 @@ def test_bundler_packages(
...
+                ref="well_formed_ruby_all_features",
+                packages=({"path": ".", "type": "bundler", "allow_binary": "true"},),
+                flags=["--dev-package-managers"],
+                check_output=False,
+                check_deps_checksums=False,
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="",
+            ),
+            [],  # No additional commands are run to verify the build
```
slimreaper35, line 120: Is this marked as TODO or on purpose ?

slimreaper35 (commented):
It would be fine to keep e2e tests as separate commits. You forgot to generate test data.

cachi2/core/package_managers/bundler/parser.py
a-ovchinnikov: We could rework it, but after I am done with sqlite3 (and potentially other interesting ways of storing versions).

cachi2/core/models/input.py
```diff
@@ -73,6 +73,7 @@ class BundlerPackageInput(_PackageInputBase):
     """Accepted input for a bundler package."""
```
eskultety, line 74: **Commit message:**

If this patch is supposed to resolve the issue in question then it should be prefixed with one of the magic words (e.g. Resolves) to automatically close the issue on merge.

slimreaper35:
I am not sure if overall that's what we want. When I was testing it locally, it looked like our parser script always adds platform = "ruby" if the source gem exists. So if users set allow binary to true, we download source gems anyway.

When I had (sorbet)[https://rubygems.org/gems/sorbet-static/versions] in Gemfile.lock. JSON output had 3 gems for 3 different platforms. Then, we would not download all pre-compiled gems available.

Everything kind of depends on the parser.

a-ovchinnikov:
>I am not sure if overall that's what we want.

My understanding is that we want to download whatever is specified in a Gemfile.lock. Technically we do not even need to ask a user if they want precompiled binaries or not, but for consistency with pip we do this. I think it is also good to be explicit about binary blobs and make user acknowledge that this is indeed what they want: there is always a chance that this was never noticed before and made its way into a stricter environment by accident.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -80,6 +80,44 @@ def download_to(self, deps_dir: RootedPath) -> None:
...
+            "Downloading platform-specific gem %s-%s-%s", self.name, self.version, self.platform
+        )
+        try:
+            download_binary_file(self.remote_location, fs_location)
+        except FetchError:
+            # A combination of Ruby v.3.0.7 and some Bundler dependencies results in
+            # -gnu suffix being dropped from some platforms. This was observed on
+            # sqlite3-aarch-linux-gnu. We cannot control user's Ruby version,
+            # but we could try and guess correct path. If this fails then we should
+            # assume that package is broken.
```
eskultety, line 113: What test data did you use? How did you come to the conclusion it has something to do with Ruby v3.0.7? When I tried the following `Gemfile` with that Ruby version (bundler 2.2.33):
```
source "https://rubygems.org"

gem "sqlite3", "= 2.0.4"
```
The lockfile I got for mult-arch platforms was (note the `mini_portile` dep):
```
GEM
  remote: https://rubygems.org/
  specs:
    mini_portile2 (2.8.7)
    sqlite3 (2.0.4)
      mini_portile2 (~> 2.8.0)

PLATFORMS
  aarch64-linux
  arm-linux
  arm-linux-musl
  arm64-darwin
  x64-mingw-ucrt
  x86-linux
  x86_64-darwin
  x86_64-linux
  x86_64-linux-musl

DEPENDENCIES
  sqlite3 (= 2.0.4)
```

I was not able to reproduce otherwise with newer Ruby which apparently changed how the compile targets are handled and with 3.1.0 I got:
```
GEM
  remote: https://rubygems.org/
  specs:
    sqlite3 (2.0.4-aarch64-linux-gnu)
    sqlite3 (2.0.4-arm-linux-gnu)
    sqlite3 (2.0.4-arm-linux-musl)
    sqlite3 (2.0.4-arm64-darwin)
    sqlite3 (2.0.4-x64-mingw-ucrt)
    sqlite3 (2.0.4-x86-linux-gnu)
    sqlite3 (2.0.4-x86_64-darwin)
    sqlite3 (2.0.4-x86_64-linux-gnu)
    sqlite3 (2.0.4-x86_64-linux-musl)

PLATFORMS
  aarch64-linux
  arm-linux
  arm-linux-musl
  arm64-darwin
  x64-mingw-ucrt
  x86-linux
  x86_64-darwin
  x86_64-linux
  x86_64-linux-musl

DEPENDENCIES
  sqlite3 (= 2.0.4)
```

That said, I found this issue which reported a similar problem, but for the `musl` target: https://github.com/rubygems/rubygems/issues/7432 . I wonder if these could be related in any way.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -80,6 +80,44 @@ def download_to(self, deps_dir: RootedPath) -> None:
...
+            self.platform = self.platform + "-gnu"
+            fs_location = deps_dir.join_within_root(
+                Path(f"{self.name}-{self.version}-{self.platform}.gem")
+            )
+            download_binary_file(self.remote_location, fs_location)
```
eskultety, line 118: Based on my limited research above (missing `musl` target in the lockfile) I am not convinced that we want to assume this problem exists only for `-gnu`. The reporter of https://github.com/rubygems/rubygems/issues/7432 mentioned that he could work around the problem by explicitly doing `bundle lock --add-platform xyz-musl` and then it produced the expected results. At best I'd suggest keeping a neutral attitude here, not trying to recover from an error that likely doesn't originate at cachi2, and simply re-raise  `FetchError` (or change the type) providing a solution and in that solution mention that the lockfile should be inspected and the actual URL checked for existence.

cachi2/core/package_managers/bundler/parser.py
eskultety: > We cannot force Ruby platform too: this requires modifications to the lock file

We should not touch the lock file - ever - as that is the only source of truth for us, if we change it, it's going to be hard arguing about incorrect platform targets, bad resolution (see the issue I linked elsewhere) to the consumer base.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -194,7 +234,24 @@ def parse_lockfile(package_dir: RootedPath) -> ParseResult:
...
+                    log.warning(
+                        "Downloading binary dependency %s because 'allow_binary' is set to True",
+                        full_name,
+                    )
+                    result.append(GemPlatformSpecificDependency(**dep))
+                else:
+                    # No need to force a platform if we skip the packages.
+                    log.warning(
+                        "Skipping binary dependency %s because 'allow_binary' is set to False."
+                        " This will likely result in an unbuildable package.",
```
eskultety, line 250: nitpick: please invert the `platform != "ruby"` condition - a short-circuit evaluation reads much easier than a loong `if` and a one-liner `else`

cachi2/core/package_managers/bundler/parser.py
eskultety: What I potentially don't agree with is the warning statement as argued by @slimreaper35 for one particular reason - I don't see any further scanning on dupes in case `allow_binary == False`, IOW there could be an identical dependency defined with platform `ruby` withing the lockfile along with the platform specific ones, couldn't it (for the same reasons as Python's wheels - GPU bound builds, complex and long builds) ?
If I'm not mistaken in that reasoning, then by having at least the `ruby` one, a subsequent build would still pass and so we ought not scare the user with such statements; and if it so happens that they'd fail the build, then we can point them to the docs and say there's a flag for binary deps.

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -23,6 +24,15 @@
 from tests.common_utils import GIT_REF
 
 
+def relaxed_in(substring: str, messages: Iterable[str]) -> bool:
```
eskultety, line 27: This name doesn't really hint at what it's doing. Additionally, since you already went down the road, could you generalize it even more and accept a regex pattern and match that?

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -215,6 +226,62 @@ def test_dependencies_could_be_downloaded(
...
+    ]
+
+    mock_downloader.side_effect = [FetchError(reason="testing"), "ok"]
+
+    dependency.download_to(base_destination)
+
+    assert relaxed_in("Downloading platform-specific gem", caplog.messages)
+    mock_downloader.assert_has_calls(expected_calls)
+
+
```
eskultety, line 284: In context of this review, this test case should go away.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -194,7 +234,24 @@ def parse_lockfile(package_dir: RootedPath) -> ParseResult:
     result: ParseResult = []
     for dep in dependencies:
         if dep["type"] == "rubygems":
-            result.append(GemDependency(**dep))
+            if dep["platform"] != "ruby":
+                full_name = "-".join([dep["name"], dep["version"], dep["platform"]])
+                log.warning("Found a binary dependency %s", full_name)
```
eskultety, line 239: nitpick: This warning feels a bit redundant since there are only 2 branches and each one reports a different WARNING. Also, I think those are INFO level messages, not warning, a potentially failed build aside there isn't any immediate harm that could be the result of this code. And since there is a flag for including binary deps, it simply doesn't feel justified as a warning anyway.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -80,6 +80,44 @@ def download_to(self, deps_dir: RootedPath) -> None:
...
+    Attributes:
+        platform:     Platform for which the dependency was built.
+    """
+
+    platform: str
+
+    @property
+    def remote_location(self) -> str:
+        """Return remote location to download this gem from."""
+        return f"{self.source}/downloads/{self.name}-{self.version}-{self.platform}.gem"
```
eskultety, line 96: Out of curiosity, is it okay for us to hard-code the 'downloads' part of the remote path? Do we want to be easily prepared for custom Ruby "package indices" ?

cachi2/core/models/input.py
```diff
@@ -73,6 +73,7 @@ class BundlerPackageInput(_PackageInputBase):
...
 
     type: Literal["bundler"]
+    allow_binary: bool = False
```
eskultety, line 76: I believe the CLI change should go to a separate commit and be decoupled from the core logic.

eskultety, line 76: **Commit message:**

I don't think this is a "fix" anymore. By handling this with a CLI flag it is a genuine feature ;) and so the commit subject should be adjusted to actually match the vibe.

cachi2/core/package_managers/bundler/parser.py
a-ovchinnikov: No, it won't pass, at least not always. It started failing with nokogiri despite nokogiri having ruby version. If a version is present in a lockfile it seems to be needed by Bundler. If it were the case I would not have noticed this at all. I don't like the idea of silently falling back to ruby platform. Binary implementations exist for a reason, and usually this reason is performance. With silent ruby fallback we'll end up with users having noticeably worse performance by default. And since everything will work they will probably grumble, but accept it. And then some time in the future they will discover that someone made a suboptimal choice for them and will be very upset.

tests/unit/package_managers/bundler/test_parser.py
a-ovchinnikov: Do you mean `test_binary_gem_dependencies_could_be_downloaded_for_damaged_platforms`? If yes then I disagree: it verifies that our code handles externally produced damaged data so it must stay.

cachi2/core/package_managers/bundler/parser.py
eskultety: > Binary implementations exist for a reason, and usually this reason is performance. With silent ruby fallback we'll end up with users having noticeably worse performance by default. And since everything will work they will probably grumble, but accept it. And then some time in the future they will discover that someone made a suboptimal choice for them and will be very upset.

Not a problem IMO as there's going to be a documented CLI flag to avert this situation.

> If a version is present in a lockfile it seems to be needed by Bundler

Fair enough, that makes sense. I guess I was more focused on the parity with our `pip` module so that we don't try downloading any binary packages (unless specified in the lockfile explicitly) without the CLI flag, but admittedly the situation with Bundler is different.

a-ovchinnikov: > Not a problem IMO as there's going to be a documented CLI flag

A user could be someone other than a package builder. We have a pretty long supply chain here, there are many settings along it, I would not want to introduce another one with a surprising choice of defaults. I think this is something we can and should decide on on behalf of the users. This choice provides experience closer to one of installing dependencies on a connected machine, most users would expect that.

Another alternative I thought about is to build everything _including binary dependencies_ ourselves. This looks to be out of scope for cachi2: while we could download sources someone else would need to determine the right build sequence etc. This feels like a task for a proper package-manager. The next best thing is to try and replicate regular installation as close as possible. We can do this by downloading all binaries that Bundler deems necessary after a user agrees to that. Highlighting this is important because it is very easy to overlook a performance setting in the middle of a pipeline.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -60,7 +74,16 @@ def _resolve_bundler_package(package_dir: RootedPath, output_dir: RootedPath) ->
...
-        components.append(Component(name=dep.name, version=dep.version, purl=dep.purl))
+        if isinstance(dep, GemPlatformSpecificDependency):
+            c = Component(
+                name=dep.name,
+                version=dep.version,
+                purl=dep.purl,
+                properties=PropertySet(bundler_package_binary=True).to_properties(),
+            )
+        else:
+            c = Component(name=dep.name, version=dep.version, purl=dep.purl)
```
slimreaper35, line 85:
```suggestion
        c = Component(name=dep.name, version=dep.version, purl=dep.purl)
        if isinstance(dep, GemPlatformSpecificDependency):
            c.properties = PropertySet(bundler_package_binary=True).to_properties()
```

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -194,7 +226,24 @@ def parse_lockfile(package_dir: RootedPath) -> ParseResult:
...
-            result.append(GemDependency(**dep))
+            if dep["platform"] == "ruby":
+                result.append(GemDependency(**dep))
+            else:
+                full_name = "-".join([dep["name"], dep["version"], dep["platform"]])
+                log.info("Found a binary dependency %s", full_name)
+                if allow_binary:
+                    log.warning(
+                        "Downloading binary dependency %s because 'allow_binary' is set to True",
+                        full_name,
```
slimreaper35, line 237: It is not really downloading that is happening here. Later on, there is another INFO log message anyway.

tests/integration/test_data/bundler_everything_present/container/Containerfile
```diff
@@ -0,0 +1,25 @@
...
+# This should be a COPY, but the source code and Containerfile are in different directories
+RUN cp -r /tmp/bundler_everything_present-source /src
+
+WORKDIR /src
+# Bundler would try and install whichever version was used to generate Gemfile.lock.
+# This will cause bundler to attempt to download an earlier version even if
+# just microversions diverged. This, in turn, would cause a build to fail.
+# Running 'bundle _<version>_ install' is supposed to attempt to run installation
+# with this specific version. The extra code below ensures that any present
+# version is used:
```
eskultety, line 24: Interesting behaviour....which means that it's not just the problem of our e2e test, users will suffer from this as well in hermetic envs and we surely don't want Bundler to fail the installation due to a no network access! I don't think what follows is the correct fix, instead I think we should override the `BUNDLE_VERSION` [1] variable along with the other ones and document.

[1] https://bundler.io/v2.5/man/bundle-config.1.html#LIST-OF-AVAILABLE-KEYS
[2] https://github.com/rubygems/rubygems/issues/6613
[3] https://github.com/rubygems/rubygems/pull/6824/

eskultety (commented):
Mostly minor comments on top of @slimreaper35's . The default bundler version downgrade behaviour raised a red flag for me and so I believe we need to set `BUNDLE_VERSION`.

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -15,13 +16,26 @@
...
+def some_message_contains_substring(substring: RegexpStr, messages: Iterable[str]) -> bool:
+    """Check if substring-matching regexp could be found in any message.
+
+    This produces a bit less coupling between tests and code than
+    checking for a full message.
+    """
+    r = re.compile(substring)
+    return any(r.match(m) is not None for m in messages)
+
```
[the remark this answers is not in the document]
slimreaper35: If I replace the function with `assert "..." in caplog.text` everything passes. But it is not ideal either.

> Now imagine an error message which has been constructed at multiple places (unwinding of the exception stack), each adding some data specific bits.

This feels like a message that could be constructed at multiple places:
```
assert some_message_contains_substring("Found a binary dependency", caplog.messages)
assert some_message_contains_substring("Downloading binary dependency", caplog.messages)
```
So I don't think we have to do anything here (at least for now). 
Just use `caplog.messages` and check full log messages as we do elsewhere in the tests.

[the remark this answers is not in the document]
eskultety: > This feels like a message that could be constructed at multiple places:
> ```
> assert some_message_contains_substring("Found a binary dependency", caplog.messages)
> assert some_message_contains_substring("Downloading binary dependency", caplog.messages)
>```

What I meant was a message like this:
`FetchError: Failed to fetch dependencies: Failed to process file 'XYZ': 'No such file or directory'`

where the message itself consists of 3 parts consisting of 3 different streams of information or just context wrapping whatever you want to call it where the 2nd part of the msg is tied to file some operations and the 3rd part gives you the exact system error (based on errno) why the file operation failed. The messages can be arbitrarily complex and regex handling is useful in such cases.
That said, it's not a hard blocker for me, the function name isn't ideal though and I'd surely use the word `assert` in its name to signal it's a wrapper over `assert`.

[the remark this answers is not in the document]
eskultety: Note that our logging is far from ideal, so we may want to improve on that in the future. That is irrelevant to this thread and PR and so I won't digress.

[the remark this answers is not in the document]
slimreaper35: Oh I found this: https://github.com/containerbuildsystem/cachi2/pull/128

cachi2/core/package_managers/bundler/main.py
```diff
@@ -27,10 +33,14 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
...
-    for package in request.packages:
+    for package in request.bundler_packages:
         path_within_root = request.source_dir.join_within_root(package.path)
         components.extend(
-            _resolve_bundler_package(package_dir=path_within_root, output_dir=request.output_dir)
+            _resolve_bundler_package(
+                package_dir=path_within_root,
+                output_dir=request.output_dir,
+                allow_binary=package.allow_binary,
+            )
```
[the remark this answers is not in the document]
a-ovchinnikov: Done.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -80,6 +80,36 @@ def download_to(self, deps_dir: RootedPath) -> None:
...
+        fs_location = deps_dir.join_within_root(
+            Path(f"{self.name}-{self.version}-{self.platform}.gem")
+        )
+        log.info(
+            "Downloading platform-specific gem %s-%s-%s", self.name, self.version, self.platform
+        )
+        # A combination of Ruby v.3.0.7 and some Bundler dependencies results in
+        # -gnu suffix being dropped from some platforms. This was observed on
+        # sqlite3-aarch-linux-gnu. We discourage using outdated platforms
+        # for building dependencies and cnsider this to be a limitation of Ruby.
```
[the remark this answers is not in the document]
eskultety: Yes, which means that this hunk needs to go.

tests/integration/test_data/bundler_everything_present/container/Containerfile
eskultety: > Can we do this? We can detect a version that is used in a fetch environment, but there is no guarantee that this same version is used in build environment. 

Yes we can! The reason I suggested is the keyword `system` that the variable accepts. From the docs:

> system will use the system version of Bundler

cachi2/core/package_managers/bundler/parser.py
eskultety: The logic was eventually dropped -> resolving.

tests/unit/package_managers/bundler/test_parser.py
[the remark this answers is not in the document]
eskultety: > Checking messages is a questionable idea because it tightly couples code and tests. A typo or a new, clearer message would mean that tests must change too and this is bad because no change is logic resulted in a change to tests. Tests must check logic, exact shape of side effects is unimportant. Our logic demands that we emit some messages...

100% agreed! That's why I suggested regex pattern searching in the first place to avoid the need to change tests when we want to add a small additional check that we're emitting an error/exception of the right type and that the contents more or less makes sense instead of matching exact strings which I'm also not a fan of.
We're doing that already, we can improve and it's not a blocker for me if we don't do it here ultimately, I just had a suggestion on regexes and you did that. Generalizing the function can come later (if we don't forget).

cachi2/core/package_managers/bundler/parser.py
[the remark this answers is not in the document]
a-ovchinnikov: Done.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -226,7 +226,24 @@ def parse_lockfile(package_dir: RootedPath) -> ParseResult:
...
-            result.append(GemDependency(**dep))
+            if dep["platform"] == "ruby":
+                result.append(GemDependency(**dep))
+            else:
+                full_name = "-".join([dep["name"], dep["version"], dep["platform"]])
+                log.info("Found a binary dependency %s", full_name)
+                if allow_binary:
+                    log.warning(
+                        "Will download binary dependency %s because 'allow_binary' is set to True",
+                        full_name,
```
slimreaper35, line 237: IMO, this is an unnecessary warning. Users should know, that they set `allow_binary`.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -226,7 +226,24 @@ def parse_lockfile(package_dir: RootedPath) -> ParseResult:
...
+                    )
+                    result.append(GemPlatformSpecificDependency(**dep))
+                else:
+                    # No need to force a platform if we skip the packages.
+                    log.warning(
+                        "Skipping binary dependency %s because 'allow_binary' is set to False."
+                        " This will likely result in an unbuildable package.",
```
slimreaper35, line 244:
```suggestion
log.warning("Skipping platform-specific gem %s because 'allow_binary' option is disabled", full_name)
```

slimreaper35 (commented):
I miss generated SBOM from e2e tests.

tests/integration/test_data/bundler_everything_present/container/Containerfile
a-ovchinnikov: Missed the 'system' part yesterday, it does work after all.

cachi2/core/package_managers/bundler/parser.py
a-ovchinnikov: I would rather give users another clue. There are many ways a build could go wrong, messing with binary dependencies is just one of those. I would expect a user to start debugging a failing build with going through logs, thus putting clues right there seems reasonable. Besides that our exceptions often have "solution" field, I think having a detailed warning is in the spirit of having solutions for exceptions.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -159,6 +159,7 @@ def _prepare_for_hermetic_build(source_dir: RootedPath, output_dir: RootedPath)
         BUNDLE_CACHE_PATH: "${output_dir}/deps/bundler"
         BUNDLE_DEPLOYMENT: "true"
         BUNDLE_NO_PRUNE: "true"
+        BUNDLE_VERSION: "system"
```
eskultety, line 162: **Commit message**:
nitpick: All true except that bundler apparently attempts to do this only with Z-streams, a version mismatch in Y-stream or major leads to a warning only that bundler isn't trying to correct by itself.

cachi2/core/package_managers/bundler/parser.py
eskultety: This warning debate has IMO gone out of hand a bit, let's just agree on keeping it, it's just a string of characters after all that doesn't decrease the quality of the output and can potentially only improve it.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -60,7 +73,10 @@ def _resolve_bundler_package(package_dir: RootedPath, output_dir: RootedPath) ->
     components = [Component(name=name, version=version, purl=main_package_purl.to_string())]
     for dep in dependencies:
         dep.download_to(deps_dir)
-        components.append(Component(name=dep.name, version=dep.version, purl=dep.purl))
+        c = Component(name=dep.name, version=dep.version, purl=dep.purl)
+        if isinstance(dep, GemPlatformSpecificDependency):
+            c.properties = PropertySet(bundler_package_binary=True).to_properties()
```
eskultety, line 79: This line actually causes this commit not to pass mypy in a rebase because the attribute is added by the next one.

a-ovchinnikov: Indeed, thank you!

slimreaper35:
One more thing, that I probably mentioned somewhere, if we want to add platform qualifier in the PURLs:
https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#gem

eskultety (approved):

eskultety:
> One more thing, that I probably mentioned somewhere, if we want to add platform qualifier in the PURLs: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#gem

Good point.

a-ovchinnikov:
>> One more thing, that I probably mentioned somewhere, if we want to add platform qualifier in the PURLs: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#gem

> Good point.

If I read the link correctly this is a different platform: not a platform a gem is compiled to, but a Ruby implementation. The given example explicitly mentions JRuby.

slimreaper35:
> If I read the link correctly this is a different platform: not a platform a gem is compiled to, but a Ruby implementation. The given example explicitly mentions JRuby.

Okay. So we can probably leave it. We are not handling this for `pip` either.

a-ovchinnikov:
> And how about the SBOM from e2e tests ?

Do we need it? The logic of SBOM generation is pretty static and well-tested with UTs (97+% coverage is no joke). The tests check that bundler can build a package after it was prepared by cachi2 and that there are no failures there, they do catch quite a lot of interesting behaviours as we have learned. I think adding more data is excessive and will not bring any more confidence.

slimreaper35:
We have it for all e2e tests for all package managers, so at least for consistency, I would add it. We can discuss it when we put all integration tests into a single repo. That should be a good starting point for improving the integration tests overall.

slimreaper35 (approved):

a-ovchinnikov:
> We have it for all e2e tests for all package managers

Done!

PR #680, opened 2024-10-09
pip metadata refactoring

slimreaper35:
My local approximate results:

`(venv) ~/cachi2 (main) $ time tox -e py312`
```
real    0m24.600s
user    0m23.449s
sys     0m1.419s
```

`(venv) ~/cachi2 (pip-refactoring) $ time tox -e py312`
```
real    0m13.625s
user    0m12.713s
sys     0m0.997s
```


a-ovchinnikov (approved):
LGTM with some minor nitpicks.

cachi2/core/package_managers/pip.py
```diff
@@ -273,72 +272,71 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
...
+    """
+    Attempt to retrieve the name and version of a pip package.
+
+    The function attempts to extract metadata in the following order:
+    1. pyproject.toml
+    2. setup.py
+    3. setup.cfg
+
+    If both name and version cannot be resolved from these sources,
+    it tries to infer the name from origin remote URL of the git repository.
```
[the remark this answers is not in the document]
slimreaper35: I renamed it to `_extract_metadata_from_*config*_files`. I would say it is more accurate.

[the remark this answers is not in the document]
a-ovchinnikov: Yes, this sounds better! Thank you.

cachi2/core/package_managers/pip.py
```diff
@@ -1,4 +1,3 @@
-# SPDX-License-Identifier: GPL-3.0-or-later
```
a-ovchinnikov, line 1: Why do you remove license identifier?

cachi2/core/package_managers/pip.py
```diff
@@ -273,72 +272,63 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
...
-    if version:
-        log.info("Resolved package version: %r", version)
-    else:
-        log.warning("Could not resolve package version")
+
+def _get_pip_metadata(package_dir: RootedPath) -> tuple[str, Optional[str]]:
+    """Attempt to retrieve name and version of a pip package."""
+    name, version = _extract_metadata_from_config_files(package_dir)
+
+    if not name:
```
a-ovchinnikov, line 330: Or even simpler:
```python
return name or _infer_name..(..), version
```

cachi2/core/package_managers/pip.py
slimreaper35: That is a leftover from Cachito.

eskultety: It should stay IMO. It's always harder to add SPDX license IDs afterwards should we adopt modules which would simply not comply with GPL, but would with LGPL (e.g. contributions from entities working primarily closed source). IOW should it happen that we wouldn't have a global license, these SPDX tags very clearly denote what module is shipped under what license. They're harmless and I'd argue we should add them to all modules if we haven't done so yet while we can (i.e. while the contributions are coming from the same entity).

a-ovchinnikov: Back in the day every module in OpenStack would contain several lines of header to indicate which license it is distributed under. The reasoning was that if any of the modules start living on their own there would be no confusion about how they must be licensed. Back then the general consensus was that some hints about governing license must be present everywhere.

slimreaper35: I don't know anything about this topic so I will add it back. Btw. there are only a few files that contain these license identifiers.

eskultety: > Back in the day every module in OpenStack would contain several lines of header to indicate which license it is distributed under. The reasoning was that if any of the modules start living on their own there would be no confusion about how they must be licensed. Back then the general consensus was that some hints about governing license must be present everywhere.

Yes. A bit of trivia on a different project:

We had a bit different situation in libvirt - the project is distributed under both LGPL and GPL (the library code was LGPL and library unrelated code was GPL). Another major factor there were copyrighted contributions all of which predated SPDX license tags. Since the project has matured and has been around for a while there was only a limited maneuvering space to do something about this and so some modules had to retain the full blown LGPL preamble at the beginning mostly due to copyright mentions while new modules only use SDPX tags. In some cases the adoption of SPDX tags (in a sibling project under the same project umbrella I think) required contacting individual contributors to get their blessing for ridding of the copyright parts in the modules' preambles.

Arguably very few engineers actually understand all the nooks and crannies of OSS licensing (I'm definitely not one of them, quite the opposite!). Our situation is very different and much easier because we're operating under the global GPL license and we don't carry any copyrighted contributions. That said, removing SPDX license identifiers is rarely a good idea and we MUST be very careful with that. With all that said we have 2 options:

1. in order to be consistent, we should add the SPDX license IDs to all modules since thanks to the global license we don't need to go and ask anyone for permission to add these tags to all modules, but asking every contributor to add them to their newly created modules may sound tedious
2. Do nothing and keep it as is! Modules carried over from another project must keep whatever licensing burden they originally had and new ones are covered by the global license.

Having thought about it overnight, I have to retract on my earlier comment:

> I'd argue we should add them to all modules if we haven't done so yet while we can

and express my clear preference for 2.

Tangentially related:
As a consumer of a project one naturally cannot "downgrade" the license, only the opposite direction is possible. However, I wonder what the steps for the project itself are if they wish to re-license, downgrading to LPGL for example, it surely would require a unilateral vote of the community (yeah, good luck with that), but I don't know what the process actually looks like at all.

cachi2/core/package_managers/pip.py
```diff
@@ -273,72 +272,71 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
...
     if pyproject_toml.exists():
-        log.info("Extracting metadata from pyproject.toml")
+        log.debug("Checking pyproject.toml for metadata")
         if pyproject_toml.check_dynamic_version():
-            log.warning("Parsing dynamic metadata from pyproject.toml is not supported")
+            log.warning("Dynamic version parsing from pyproject.toml is not supported")
 
-        name = pyproject_toml.get_name()
-        version = pyproject_toml.get_version()
+        name, version = pyproject_toml.get_name(), pyproject_toml.get_version()
```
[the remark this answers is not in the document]
eskultety: @a-ovchinnikov What is `or_` aliased to?

eskultety:
There's too much going on in this single commit, so it's difficult to follow all the changes in the diff, please introduce them gradually.

cachi2/core/package_managers/pip.py
```diff
@@ -273,74 +273,61 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
...
-                    "Please specify package metadata in a way that Cachi2 understands"
-                    " (see the docs)\n"
-                    "or make sure that the directory Cachi2 is processing is a git"
-                    " repository with\n"
-                    "an 'origin' remote in which case Cachi2 will infer the package name from"
-                    " the remote url."
-                ),
-                docs=PIP_METADATA_DOC,
-            )
+    return None, None
```
eskultety, line 324: This refactor isn't a 1:1 replacement - previously if a name was found it was carried over to other source checks. Here, if you extract name but can't find version your code will resort to returning `(None,None)` which means we'll infer the name of the project from the URL instead, that doesn't sound right.

cachi2/core/package_managers/pip.py
```diff
@@ -273,74 +273,61 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
...
 
-    log.info("Resolved package name: %r", name)
-    if version:
-        log.info("Resolved package version: %r", version)
-    else:
-        log.warning("Could not resolve package version")
```
eskultety, line 341: We lost this warning/info completely.

cachi2/core/package_managers/pip.py
```diff
@@ -273,74 +273,61 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
...
+            return name, version
 
-    if None in (name, version) and setup_py.exists():
-        log.info("Filling in missing metadata from setup.py")
-        name = name or setup_py.get_name()
-        version = version or setup_py.get_version()
+    # setup.py
+    if setup_py.exists():
+        log.debug("Checking setup.py for metadata")
+        name, version = setup_py.get_name(), setup_py.get_version()
```
eskultety, line 315: Please keep these on separate lines, it's more readable, we don't gain anything with this compound line and we don't lose anything by having it nicely split (all occurrences).

cachi2/core/package_managers/pip.py
```diff
@@ -273,74 +273,61 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
...
     setup_py = SetupPY(package_dir)
     setup_cfg = SetupCFG(package_dir)
 
+    # pyproject.toml
     if pyproject_toml.exists():
-        log.info("Extracting metadata from pyproject.toml")
+        log.debug("Checking pyproject.toml for metadata")
         if pyproject_toml.check_dynamic_version():
-            log.warning("Parsing dynamic metadata from pyproject.toml is not supported")
+            log.warning("Dynamic version parsing from pyproject.toml is not supported")
```
eskultety, line 306: I believe this was a bug all the time - if we can't support dynamic version parsing from pyproject.toml we should not proceed with parsing it and instead should skip it and try other source - can we do that?

cachi2/core/package_managers/pip.py
```diff
@@ -273,74 +273,61 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
...
-    if version:
-        log.info("Resolved package version: %r", version)
-    else:
-        log.warning("Could not resolve package version")
 
-    return name, version
+def _get_pip_metadata(package_dir: RootedPath) -> tuple[str, Optional[str]]:
+    """Attempt to retrieve name and version of a pip package."""
+    name, version = _extract_metadata_from_config_files(package_dir)
+    return name or _infer_package_name_from_origin_url(package_dir), version
```
eskultety, line 330: I think having this conditionally assigned is more readable, but I believe this was @a-ovchinnikov review suggestion IIRC, so I'll refrain from demanding a conflicting approach, so let's keep it.

eskultety (commented):
Very much in favour of this work, needs some polishing though.

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
a-ovchinnikov: Sorry, that's `or` as a function as imported from operator module.

cachi2/core/package_managers/pip.py
slimreaper35: I would prefer to keep the previous approach too.

cachi2/core/package_managers/pip.py
slimreaper35: Can we at least try to parse the name ?

eskultety: Sure, that would be aligned with the previous approach.

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: Well, we could write our own lambda for this, but then again, doesn't seem much readable to me nor performant, IIUC reduce is one of the functions (like filter) that shine with big data sets; we're talking 3 values here.

cachi2/core/package_managers/pip.py
```diff
@@ -273,74 +273,61 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
...
+        )
 
-    First, try to parse the setup.py script (if present) and extract name and version
-    from keyword arguments to the setuptools.setup() call. If either name or version
-    could not be resolved and there is a setup.cfg file, try to fill in the missing
-    values from metadata.name and metadata.version in the .cfg file.
+    repo_name = Path(repo_id.parsed_origin_url.path.removesuffix(".git")).name
+    subpath = package_dir.subpath_from_root
+    resolved_name = Path(repo_name).joinpath(subpath)
+    return canonicalize_name(str(resolved_name).replace("/", "-")).strip("-.")
```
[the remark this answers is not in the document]
slimreaper35: I would like to not rearrange the commits in this PR at all + don't waste too much time here. Each commit would require modification in unit tests, and that is not trivial based on the extremely complicated logic in there. I chose a much easier path, so, delete everything and write this from scratch. 

I know, in GitHub the PR is basically unreviewable. Sorry for that.

a-ovchinnikov (dismissed):

tests/unit/package_managers/test_pip.py
```diff
@@ -58,116 +54,116 @@ def make_dpi(
...
-@mock.patch("cachi2.core.package_managers.pip.SetupPY")
+@pytest.mark.parametrize(
+    "dynamic_version",
+    [True, False],
+)
 @mock.patch("cachi2.core.package_managers.pip.PyProjectTOML")
-@mock.patch("cachi2.core.package_managers.pip.get_repo_id")
-def test_get_pip_metadata(
-    mock_get_repo_id: mock.Mock,
+def test_get_pip_metadata_from_pyproject_toml(
```
eskultety, line 62: You decreased the value of the unit test by adding per-project file test case because you never test the combinations by making the main test split into multiple ones which is surely more readable but only part of the original logic is being tested, that was the point of the complex test.

cachi2/core/package_managers/pip.py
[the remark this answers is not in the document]
eskultety: > I know, in GitHub the PR is basically unreviewable. Sorry for that.

First things first, GH has nothing to do with this, I review stuff locally with 50 lines of context usually and even with flags to git to detect moved code doesn't change the situation in any way.

> I would like to not rearrange the commits in this PR at all + don't waste too much time here. Each commit would require modification in unit tests, and that is not trivial based on the extremely complicated logic in there. I chose a much easier path, so, delete everything and write this from scratch.

The only blocker for splitting here is the unit tests here, I agree. But given that this refactor actually changes the behavioral logic such that we don't return the same results as previously from the `pip_metadata` helper function. Additionally, the tests that you introduce actually don't test the compound cases which is the whole point of the pip metadata querying logic (e.g. name comes from one project file, version from another one). Since it looks like you'll have to rework the unit tests quite a bit that gives you the opportunity to split the changes into commits - I have actually tried myself and was pretty straightforward except for the tests, but like I said, looks like you'll have to rework that.

cachi2/core/package_managers/pip.py
```diff
@@ -460,14 +460,6 @@ def get_version(self) -> Optional[str]:
             log.warning("No project.version in pyproject.toml")
             return None
 
-    def check_dynamic_version(self) -> bool:
```
eskultety, line 456: You don't explain anywhere why you're dropping this check.

eskultety:
> > After having carefully gone through the unit tests which I didn't do in my first round of reviews I think we're actually opening us up for potential issues with pyproject.toml setup.py etc. mixed metadata.
> 
> That's a good point. The fact that we were mixing metadata from multiple project configuration files was there reason why we ended up with extremely complicated and long unit tests. Splitting name and version into multiple configuration files makes no sense on its own. In the end, we only need the `name`, one string, for an SBOM component.

Well, what just did is a breaking change from the behaviour POV without any warning. There probably was a reason we did this way in the past. It's true that mixing metadata is wrong, however, we allowed it and it also wasn't against the ecosystem practices, was it? (although very unexpected without a doubt). So this can't be compared to our recent dropping of Go vendoring flags, because those actually allowed projects to use incorrect repo setups which would not be buildable using standard toolkits the way users intended to in the first place, I'm not sure that's the case here.

If we end up wanting this, then you'll have to accompany this change with a docs update (we'll also need to mention that in the release notes). That said, although I'm definitely not a fan of breaking backwards compatibility, strictly speaking SemVer [1]:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

and so I won't stop this work based on this argument, but we'll probably need more voices in favour.

[1] https://semver.org/#semantic-versioning-specification-semver

eskultety (commented):
You still stuffed everything into commit 1. The changes can be introduced gradually by adding one unit test at a time and turning off that particular test area in the more complex unit test you're trying to kill. That way, you'd keep most of the things as is until you're ready to switch and then remove everything you don't need in a single commit, it can be done and the diff will be much more readable IMO. I'm not fond of  trying to argument squashed changes by a complex unit test that isn't easily to be replaced (as I mentioned one option how to do it) as a justification - things can be made cleaner for the reader/reviewer.

slimreaper35:
> You still stuffed everything into commit 1. The changes can be introduced gradually by adding one unit test at a time and turning off that particular test area in the more complex unit test you're trying to kill. That way, you'd keep most of the things as is until you're ready to switch and then remove everything you don't need in a single commit, it can be done and the diff will be much more readable IMO. I'm not fond of trying to argument squashed changes by a complex unit test that isn't easily to be replaced (as I mentioned one option how to do it) as a justification - things can be made cleaner for the reader/reviewer.

I'll try my best

slimreaper35:
Also, it might be worth discussing [setup.py](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setuppy-discouraged) as:

> New projects are advised to avoid setup.py configurations (beyond the minimal stub) when custom scripting during the build is not necessary. Examples are kept in this document to help people interested in maintaining or contributing to existing packages that use setup.py. Note that you can still keep most of configuration declarative in [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html) or [pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) and use setup.py only for the parts not supported in those files (e.g. C extensions). See [note](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setuppy-discouraged).

We can at least add a warning when extracting metadata from setup.py

eskultety:
> We can at least add a warning when extracting metadata from setup.py

People should be aware by now, do we want to hold everyone's hand? I mean displaying a warning for users who genuinely need `setup.py` (because `pyproject.toml` simply doesn't cut it for them as they may have C deps) doesn't feel right. I wouldn't strictly argue against having a warning if you proposed it somewhere in the code, I'm just questioning the usefulness given the circumstances.

eskultety:
> > > After having carefully gone through the unit tests which I didn't do in my first round of reviews I think we're actually opening us up for potential issues with pyproject.toml setup.py etc. mixed metadata.
> > 
> > 
> > That's a good point. The fact that we were mixing metadata from multiple project configuration files was there reason why we ended up with extremely complicated and long unit tests. Splitting name and version into multiple configuration files makes no sense on its own. In the end, we only need the `name`, one string, for an SBOM component.
> 
> Well, what this PR just did is a breaking change from the behaviour POV without any warning. There probably was a reason we did this way in the past. It's true that mixing metadata is wrong, however, we allowed it and it also wasn't against the ecosystem practices, was it? (although very unexpected without a doubt). So this can't be compared to our recent dropping of Go vendoring flags, because those actually allowed projects to use incorrect repo setups which would not be buildable using standard toolkits the way users intended to in the first place, I'm not sure that's the case here.
> 
> If we end up wanting this, then you'll have to accompany this change with a docs update (we'll also need to mention that in the release notes). That said, although I'm definitely not a fan of breaking backwards compatibility, strictly speaking SemVer [1]:
> 
> > Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.
> 
> and so I won't stop this work based on this argument, but we'll probably need more voices in favour.
> 
> [1] https://semver.org/#semantic-versioning-specification-semver

@brunoapimentel @a-ovchinnikov @taylormadore @ben-alkov any opinions on simpler yet backwards incompatible behaviour?

cachi2/core/package_managers/pip.py
slimreaper35: Can I use similar reasoning?

> People should be aware of setup.py soft deprecation by now, do we want to hold everyone's hand? I mean displaying a warning for users who genuinely need setup.py (because pyproject.toml simply doesn't cut it for them as they may have C deps) doesn't feel right. I wouldn't strictly argue against having a warning if you proposed it somewhere in the code, I'm just questioning the usefulness given the circumstances.

Even though it is not deprecated. But the version is optional so I don't see much value in the warning message.

cachi2/core/package_managers/pip.py
```diff
@@ -312,9 +312,6 @@ def _get_pip_metadata(package_dir: RootedPath) -> tuple[str, Optional[str]]:
```
a-ovchinnikov: **Commit message**

s/optioanl/optional

a-ovchinnikov (approved):

cachi2/core/package_managers/pip.py
```diff
@@ -300,30 +300,37 @@ def _extract_metadata_from_config_files(
     1. pyproject.toml
     2. setup.py
     3. setup.cfg
```
eskultety, line 302: **Commit message:**

I don't see "breaking change" mentioned anywhere in the commit. This needs to be communicated clearly that this adjustment, while simplifying things, is a breaking change to some extent and it'll also have to be mentioned in the release notes.

eskultety (approved):
Needs commit msg adjustment wrt/ this being a breaking change, otherwise ACK.

PR #682, opened 2024-10-09
[GH workflows] Execute the integration test suite directly on the VM runner & run the simple PyPI test server as a service

eskultety:
This PR is effectively just a vol.1 in a series of small refactorings of our GH CI.
This one only takes care of how:
- we run the integration test suite
- we spin up helper services

This PR **does not** tweak how logs are captured from GH workflows, those remain messy and will handled as part of a different work.

### DIFF
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -91,7 +91,10 @@ jobs:
 
   build-image:
     name: Build Cachi2 image and run integration tests on it
-    runs-on: ubuntu-latest
+
+    # TODO: Replace this with ubuntu-latest once GH completes the migration of the VM runners to
+    # ubuntu 24.04 and respect the YAML tag (revert the commit that added this)
+    runs-on: ubuntu-24.04
     container:
       image: registry.fedoraproject.org/fedora:40
       options: --privileged
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -107,6 +107,20 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: PyPI test server start
+        run: |
+          tests/pypiserver/start.sh &
+
+          # Testing basic HTTP request
+          status=$(curl -sSI \
+                        --output /dev/null \
+                        --write-out %{http_code} \
+                        --retry-delay 1 \
+                        --retry 60 \
+                        --retry-all-errors \
+                        http://127.0.0.1:8080)
+          [[ ${status} == "200" ]] || exit 1
+
       - name: Build Cachi2 image
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/gating.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
 
   build-image:
     name: Build Cachi2 image and run integration tests on it
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
a-ovchinnikov, line 94: Will a TODO comment here help not to forget to revert it?

a-ovchinnikov (approved):
LGTM!

eskultety:
FWIW: here's a workflow run on my own fork with #580 changes applied as well, so looks like this works with more services in the background nicely.

.github/workflows/gating.yaml
```diff
@@ -104,6 +104,15 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: PyPI test server start
```
MartinBasti, line 107: Shouldn't we just start it right before integration tests?

tests/integration/conftest.py
```diff
@@ -47,7 +47,11 @@ def cachi2_image() -> utils.Cachi2Image:
 # test output.
 @pytest.fixture(autouse=True, scope="session")
 def local_pypiserver() -> Iterator[None]:
-    if os.getenv("CACHI2_TEST_LOCAL_PYPISERVER") != "true":
+    if (
+        os.getenv("CI")
+        and os.getenv("GITHUB_ACTIONS")
```
slimreaper35, line 52: Both variables are needed ?

eskultety: Technically no, but IMO decreases the likelyhood of a conflict, i.e. a false positive

.github/workflows/gating.yaml
eskultety: Does it really matter when we start it?

eskultety: To support my argument a bit, the server usually takes a couple of seconds to spin up, so if a container fails to execute (it's CI after all) I want to fail early and not wait until the cachi2 image is built, so I prefer spinning up the servers as soon as possible in the build & integration job.

MartinBasti: Running process that takes resources when you don't need it matters to me, but I assume it's small.

eskultety: Sure, but running a container image build takes them as well and the image is useless if a test server which is required isn't running ;) , so I'd argue the lesser evil wins -> marking as resolved.

MartinBasti (approved):

slimreaper35 (approved):

PR #692, opened 2024-10-21
Proposal: Enhanced purls for generic artifacts

brunoapimentel:
This draft introduces a more generic alternative to solve the need for specific Maven purls covered in #663.

In summary, it is a mechanism that would allow us to  support a subset of [purl types](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst) that do not fit into existing package manager ecosystems. Cachi2 would receive purls as input, resolve them into a download URL, and report them in the SBOM.

### DIFF
--- a/docs/design/generic-enhanced-purls.md
+++ b/docs/design/generic-enhanced-purls.md
@@ -1,0 +1,271 @@
+# Support for different purl types in the generic artifact fetcher
+The generic artifact package manager is being added to Hermeto as a means for users to introduce files that do not belong to traditional package manager ecosystems (e.g. pip, npm, golang) to their hermetic container builds. Since Hermeto does not have any extra information about the file that's being fetched, the purls are always reported as [pkg:generic](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#generic).
+
+There are use cases that would benefit from more accurate purls, though, such as the one described in the Maven artifacts [proposal](https://github.com/hermetoproject/hermeto/pull/663). Considering that the purl specification already identifies several types of packages that don't fit into traditional package manager (e.g. github, docker, huggingface; see the [purl types spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst) for more info), this proposal builds on top of the fundamentals of the generic fetcher to provide an extensible mechanism that would allow Hermeto to fetch files from specific sources and report them with matching purl types.
+
+## Generating purls of a specific type
+A purl is simply a way to represent a unique package and its location on the internet. When fetching a generic artifact, Hermeto will generate a purl of type `generic` by default, since there's no further info on the artifact's type. With this proposal, the lockfile will be extended to support a `type` field for each artifact, and for each supported purl type, a set of attributes that are necessary for resolving both the download URL and generating the purl for that artifact.
+
+This is a summary of how the resolution of a single package will look like:
+- Check if the purl type is supported
+- Read and validate the attributes
+- Resolve the download URL
+- Fetch the artifact
+- Verify that at least one of the provided checksums match
+- Generate a purl based on the input attributes
+- Generate a SBOM component containing the type-specific purl and the validated checksums as qualifiers
+
+### What confidence Hermeto has in the reported SBOM components
+- The attributes were resolved to a valid download URL where the artifact was fetched from
+- At least one of the provided checksums was validated
+- There is no guarantee over the reported purl type, Hermeto will trust the user input
+
+## Adding support for new purl types
+Although the proposal shows this feature as extensible to pretty much any purl type that exists, the decision on which types will be supported will be done on a case-by-case basis. The initial implementation was done for [Maven artifacts](https://github.com/hermetoproject/hermeto/pull/735), and support for new purl types will be analyzed by the maintainers when the need for them arises from the community.
+
+There are specific details regarding each purl type, such as extending the feature to fetch Git repositories or OCI artifacts instead of regular files, and also a potential overlap with existing or planned fully supported package managers, such as pip or npm. All of this needs to be taken in consideration before there's a decision of supporting a new purl type in the generic fetcher.
+
+## Examples of workflows for different purl types
+
+### Maven
+
+#### artifacts.lock.yaml
+```yaml
+metadata:
+  version: 1.0.0
+artifacts:
+  - type: maven
+    attributes:
+      group_id: io.quarkus
+      artifact_id: quarkus-core
+      type: jar
+      classifier: null # optional, can be omitted if null
+      repository_url: https://maven.repository.redhat.com/ga
+      version: 3.8.5.redhat-00004
+    checksum: sha256:e4ca5fad
+    target: quarkus.jar
+```
+
+#### processing the lockfile
+First, we need to determine the file extension by using the rules defined [here](https://maven.apache.org/repositories/artifacts.html#but-where-do-i-set-artifact-extension).
+
+```python
+extension = find_extension(type, classifier)
+
+# generate the download URL
+{repository_url}/{as_dir(group_id)}/{artifact_id}/{version}/{artifact_id}-{version}.{extension}
+
+# generate the purl
+pkg:maven/{group_id}/{artifact_id}@{version}[-{classifier}]?type={type}&repository_url={repository_url}&checksum={checksum}
+```
+
+#### SBOM component
+```json
+{
+  "name": "quarkus-core",
+  "version": "3.8.5.redhat-00004",
+  "purl": "pkg:maven/io.quarkus/quarkus-core@3.8.5.redhat-00004?type=jar&repository_url=https://maven.repository.redhat.com/ga&checksum=sha256:e4ca5fad",
+  "properties": [
+    {
+      "name": "hermeto:found_by",
+      "value": "hermeto:generic"
+    }
+  ],
+  "externalReferences": [
+    {
+      "type": "distribution",
+      "url": "https://maven.repository.redhat.com/ga/io/quarkus/quarkus-core/3.8.5.redhat-00004/quarkus-core-3.8.5.redhat-00004.jar"
+    }
+  ],
+  "type": "file"
+}
+```
+
+### Nuget
+
+#### artifacts.lock.yaml
+```yaml
+metadata:
+  version: 1.0.0
+artifacts:
+  - type: nuget
+    attributes:
+      namespace: Google
+      package: Protobuf
+      version: 3.28.3
+      repository_url: https://globalcdn.nuget.org/packages # default value, can be omitted
+    checksum: sha256:e4ca5fad
+```
+
+#### processing the lockfile
+```python
+# generate the download URL
+{repository_url}/{lowercase(namespace + '.' + package)}.{version}.nupkg?packageVersion={version}
+
+# generate the purl
+pkg:nuget/[{namespace}.]{package}@{version}[?repository_url={repository_url}]
+```
+
+#### SBOM component
+```json
+{
+  "name": "Google.Protobuf",
+  "version": "3.28.3",
+  "purl": "pkg:nuget/Google.Protobuf@3.28.3?checksum=sha256:e4ca5fad&repository_url=https://globalcdn.nuget.org/packages",
+  "properties": [
+    {
+      "name": "hermeto:found_by",
+      "value": "hermeto:generic"
+    }
+  ],
+  "externalReferences": [
+    {
+      "type": "distribution",
+      "url": "https://globalcdn.nuget.org/packages/google.protobuf.3.28.3.nupkg?packageVersion=3.28.3"
+    }
+  ],
+  "type": "file"
+}
+```
+
+### Hugging Face
+
+#### artifacts.lock.yaml
+```yaml
+  - type: huggingface
+    attributes:
+      namespace: nvidia
+      project: Llama-3.1-Nemotron-70B-Instruct-HF
+      commit_hash: 043235d6088ecd3dd5fb5ca3592b6913fd516027
+      repository_url: https://huggingface.co # default value, can be omitted
+  - type: huggingface
+    attributes:
+      namespace: nvidia
+      project: Llama-3.1-Nemotron-70B-Instruct-HF
+      commit_hash: 043235d6088ecd3dd5fb5ca3592b6913fd516027
+      file_name: model-00001-of-00030.safetensors
+      repository_url: https://huggingface.co # default value, can be omitted
+    checksum: sha256:e4ca5fad
+```
+
+#### processing the lockfile
+```python
+# download URL (in case a file_name is present)
+{repository_url}/{namespace}/{project}/blob/{commit_hash}/{file_name}
+
+# git clone command (in case file_name is absent) 
+git clone {repository_url}/{namespace}/{project}
+git checkout {commit_hash}
+
+# generating the purl
+pkg:huggingface/{namespace}/{project}@{commit_hash}[?repository_url={repository_url}][&file_name={file_name}]
+```
+
+#### SBOM component
+```json
+[
+  {
+    "name": "Llama-3.1-Nemotron-70B-Instruct-HF",
+    "purl": "pkg:huggingface/nvidia/Llama-3.1-Nemotron-70B-Instruct-HF@043235d6088ecd3dd5fb5ca3592b6913fd516027",
+    "properties": [
+      {
+        "name": "hermeto:found_by",
+        "value": "hermeto:generic"
+      }
+    ],
+    "externalReferences": [
+      {
+        "type": "distribution",
+        "url": "https://huggingface.co/nvidia/Llama-3.1-Nemotron-70B-Instruct-HF"
+      }
+    ],
+    "type": "library"
+  },
+  {
+    "name": "Llama-3.1-Nemotron-70B-Instruct-HF",
+    "purl": "pkg:huggingface/nvidia/Llama-3.1-Nemotron-70B-Instruct-HF@043235d6088ecd3dd5fb5ca3592b6913fd516027?file_name=model-00001-of-00030.safetensors&checksum=sha256%e4ca5fad",
+    "properties": [
+      {
+        "name": "hermeto:found_by",
+        "value": "hermeto:generic"
+      }
+    ],
+    "externalReferences": [
+      {
+        "type": "distribution",
+        "url": "https://huggingface.co/nvidia/Llama-3.1-Nemotron-70B-Instruct-HF/blob/043235d6088ecd3dd5fb5ca3592b6913fd516027/model-00001-of-00030.safetensors"
+      }
+    ],
+    "type": "file"
+  }
+]
+```
+
+### OCI
+
+#### artifacts.lock.yaml
+```yaml
+metadata:
+  version: 1.0.0
+artifacts:
+  - type: oci
+    attributes:
+      namespace: konflux-ci
+      name: buildah-task
+      digest: sha256:b2d6c32d1e05e91920cd4475b2761d58bb7ee11ad5dff3ecb59831c7572b4d0c
+      repository_url: quay.io
+    checksum: sha256:e4ca5fad
+```
+
+#### processing the lockfile
+```python
+# generate the download command
+podman pull docker://{repository_url}/{namespace}/{name}@{digest}
+
+# generate the purl
+pkg:oci/{name}@{encode(digest)}?repository_url={repository_url}/{namespace}
+```
+
+#### SBOM component
+```json
+{
+    "name": "buildah-task",
+    "purl": "pkg:oci/buildah-task@sha256:b2d6c32d1e05e91920cd4475b2761d58bb7ee11ad5dff3ecb59831c7572b4d0c?repository_url=quay.io/konflux-ci",
+    "properties": [
+        {
+            "name": "hermeto:found_by",
+            "value": "hermeto:generic"
+        }
+    ],
+    "externalReferences": [
+        {
+            "type": "distribution",
+            "url": "quay.io/konflux-ci/buildah-task@sha256:b2d6c32d1e05e91920cd4475b2761d58bb7ee11ad5dff3ecb59831c7572b4d0c"
+        }
+    ],
+    "type": "container"
+}
+```
+
+## Dropped alternative: consume purls as input
+
+We initially considered the idea of consuming purls as part of a package input, instead of relying on a decomposed attributes. This idea was abandoned because the following advantages were identified on the attribute-based approach:
+- More readable and predictable format, since the purl spec provides much flexibility on how the purls can look like
+- Better user experience, specially for users who are not familiar with purls
+- Attributes can be added or changed as needed without the need to follow the purl spec
+
+#### artifacts.lock.yaml
+```yaml
+metadata:
+  version: 1.0.0
+artifacts:
+  - purl: pkg:maven/io.quarkus/quarkus-core@3.8.5.redhat-00004?type=jar&repository_url=https://maven.repository.redhat.com/ga
+    checksums:
+      sha256: d16bf783cb6670f7f692ad7d6885ab957c63cfc1b9649bc4a3ba1cfbdfd5230c
+    target: quarkus.jar
+```
+
+**Resulting purl**
+```
+pkg:maven/io.quarkus/quarkus-core@3.8.5.redhat-00004?type=jar&repository_url=https://maven.repository.redhat.com/ga&checksum=sha256:e4ca5fad
+```


docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,150 @@
...
+        - maven.repository.internal.com
+        - maven.org
+    huggingface:
+        - huggingface.co
+```
+
+### Cachi2 CLI usage
+
+```
+cachi2 fetch-deps --source /path/to/repo generic
```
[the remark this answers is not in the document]
taylormadore: I think we'll just use/enhance the existing [config-file functionality](https://github.com/containerbuildsystem/cachi2?tab=readme-ov-file#configuration), right?

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,150 @@
...
+metadata:
+    version: '1.0'
+artifacts:
+  - download_url: https://github.com/containerbuildsystem/cachi2/archive/refs/tags/0.11.0.tar.gz
+    target: cachi2_0_11_0.tar.gz
+    checksums: 
+      sha256: fa0d536389db15fb3dabdb3b3d08354f47f765a653178140bfbe1b3de1a6ee76
+  - download_url: https://maven.repository.internal.com/ga/io/quarkus/quarkus-core/3.8.5.internal-00004/quarkus-core-3.8.5.internal-00004.jar
+    target: quakus.jar
+    type: maven
```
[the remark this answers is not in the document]
brunoapimentel: That is a good point. If we agree on enforcing the registry URL as a global rule for enhancing purls, we can probably drop this field.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
a-ovchinnikov: Yes, we can default to a location within a repo and use defaut configuration mechanism, but would need a way to point to some external config which will be version-controlled elsewhere. Suppose someone needs to build ScaPy (or some other kilostarred package) and suppose they _also_ need some prebuilt generic artifact. I don't think it is safe to assume that upstream ScaPy would just accept any cachi2 yaml. If it is a local clone the someone would still need to add the config and maintain it. Likely not a blocker right now, but I feel like we'll need to deal with this at some point.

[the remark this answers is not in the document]
eskultety: I agree with some of the @a-ovchinnikov's reasoning when it comes to checking-in some random file, but then again, imagine a blackbox pipeline, where would such a file come from? It might easily become a devops engineer's nightmare. I also agree with @brunoapimentel  that from reproducibility POV and ease of debugging you want users to commit the file to their repos. Now, why would any random big upstream care about cachi2? Exactly, they won't, that's why downstream exists and then such a file is perfectly fine to be checked in to a repo. However, downstream isn't very popular in K8S world and so users tend to solve it with git submodules instead to compensate for the need of downstream control and here again, checking in the cachi2 lockfile is perfectly fine. I don't even think we need a different way of providing it ATM. If the need arises, we can discuss again.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,150 @@
...
+                    "value": "cachi2:generic"
+                }
+            ],
+            "externalReferences": [
+                {
+                    "type": "distribution",
+                    "url": "https://github.com/containerbuildsystem/cachi2/archive/refs/tags/0.11.0.tar.gz"
+                }
+            ],
+            "type": "file"
```
eskultety, line 111: If this draft is to materialize into something mergeable then this needs to be removed as it IMO has no direct connection to the discussed matter.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,150 @@
...
+    - Mapping famous and public registries will prevent the use of private registries. To solve this, we need to allow users to change Cachi2's default configuration to enable additional source URLs for each purl type.
+    - More rules might be useful or even necessary when generating specific purl types. For example, Maven has the concept of groupId, artifactId and version (GAV) for each of its artifacts, and these values are fundamental in defining both the download URL and the purl.
+- Extend the generic artifacts lockfile specification to introduce a `type` attribute that allows users to hint at which purl type that artifact should have.
+    - The only types that can be selected are the ones that had their respective `enhancers` implemented. We don't want to allow users to be able to freely specify the purl type for a generic artifact, but rather restrict this use to a specific subset of purl types.
+    - Any failures to match the hinted `type` will not cause the request to fail. The file will still be fetched, but the purl will revert to `pkg:generic`.
```
eskultety, line 15: I don't think this is a good idea. Warnings are easily missed (nobody reads logs if things are passing!) and we'd essentially deceive our users when it comes to the resulting SBOM gibberish that is only supposed to be machine readable that they'd only find out in a later stage of their release process (or worse even - never!) that we replaced the PURL with a generic type. Type should be a strict requirement for the artifact spec in the lockfile and the model validator should very carefully validate all the information we got regarding the type since the idea here is that we'd gradually implement most of the data PURL types.
Please make this a hard failure and let's treat the lockfile as the ultimate source of truth to make our lives a bit easier as well, no shortcuts.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Input files
+
+**generic_artifacts.yaml**
+```yaml
+metadata:
+    version: '1.0'
+artifacts:
+  - download_url: https://github.com/containerbuildsystem/cachi2/archive/refs/tags/0.11.0.tar.gz
+    target: cachi2_0_11_0.tar.gz
```
eskultety, line 27: [off-topic]
Why did we call this 'target'? I think it's not a very descriptive name (which we can still change!) Wouldn't something like `filename` or `fs_filename` be more descriptive in the user facing docs?

docs/design/generic-enhanced-purls.md
kosciCZ: Agree that any unexpected mismatch should fail loud and clear.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
kosciCZ: Agree with Erik, it should be rather explicit, we definitely don't want to infer that information, as that could be potentially _a lot_ of code.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,150 @@
...
+**.cachi2-config.yaml**
+```yaml
+generic-artifact-sources:
+    maven:
+        # sample internal registry
+        - maven.repository.internal.com
+        - maven.org
+    huggingface:
+        - huggingface.co
+```
```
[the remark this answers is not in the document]
brunoapimentel: Right, this was one of the rough edges of the design. This idea stemmed from the notion that it is easy to validate famous public registries to have more certainty about the artifact type. This could even be hard-coded, but then we'd lose the possibility of more specific use cases, such as internal registries. 

My hope was that we could figure out a way to have configured for use cases where Cachi2 was being used in a managed environment, but I agree it does not make sense to allow the user to easily specify this. This would make sense only as a server-side configuration in a managed environement.

To clarify, the intention was never to replace purls, but instead to have one more validation to justify the resulting purl.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,150 @@
...
+**generic_artifacts.yaml**
+```yaml
+metadata:
+    version: '1.0'
+artifacts:
+  - download_url: https://github.com/containerbuildsystem/cachi2/archive/refs/tags/0.11.0.tar.gz
+    target: cachi2_0_11_0.tar.gz
+    checksums: 
+      sha256: fa0d536389db15fb3dabdb3b3d08354f47f765a653178140bfbe1b3de1a6ee76
+  - download_url: https://maven.repository.internal.com/ga/io/quarkus/quarkus-core/3.8.5.internal-00004/quarkus-core-3.8.5.internal-00004.jar
```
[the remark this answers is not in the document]
aloubyansky: One of the challenges with this approach is that in some edge cases you will not be able to accurately derive the `classifier` and `extension` qualifiers for Maven artifacts from a URL.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,150 @@
...
+
+- MavenPurlEnhancer:
+    - validates the origin URL
+    - parses the download url and converts it into the expected purl
+        ```bash
+        # sample url
+        https://maven.repository.internal.com/ga/io/quarkus/quarkus-core/3.8.5.internal-00004/quarkus-core-3.8.5.internal-00004.jar
+
+        # how the parsing will be done
+        https://{repository_url}/{as_dir(group_id)}/{artifactId}/{version}/{artifact_id}-{version}.{extension}
```
aloubyansky, line 67: This is missing a potential artifact `classifier`. With a `classifier` it would be
```
https://{repository_url}/{as_dir(group_id)}/{artifactId}/{version}/{artifact_id}-{version}[{-classifier}].{extension}
```
As mentioned in one of the comments above, if a `classifier` is present, you may struggle to figure out where the `classifier` ends and where the `extension` begins.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,150 @@
...
+
+There are use cases that would benefit from more accurate purls, though, such as the recent Maven artifacts [proposal]. Considering that the purl specification already identifies several types of packages that don't fit into traditional package manager (e.g. github, docker, huggingface; see the [purl types spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst) for more info), this proposal builds on top of the fundamentals of the generic fetcher to provide an extensible mechanism that would allow Cachi2 to fetch files from specific sources and report them with matching purl types.
+
+## Enhanced purls overview
+
+- Implement an `enhancer` for every supported purl type, which is essentially a set of rules that will be applied to a generic artifact and, in case those rules can be matched, it will replace the generic purl for a more specific type.
```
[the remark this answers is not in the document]
brunoapimentel: I think this is the core of the discussion that led us to the change. We, as Cachi2 maintainers, are not confident we should allow users to specify how the reported purls should be. And to clarify, this design proposes that only resolved `download_url`s are used as input, which means we wouldn't accept purls as input.

If we consider Cachi2 as a standalone project, its goal is to prefetch the dependencies necessary for a hermetic build, and report them the most accurate way possible, but only to the extent of the info it has at hand during the prefetch. Sometimes, the user will have a better purl than what Cachi2 can provide. That is not true in all cases, though. Allowing purls to be passed through creates a precedent that goes against our ultimate goal of accurate SBOMs.

Now, if we need better purls for our specific internal case, maybe they can be enhanced in other points of the pipeline? But we can lead this discussion somewhere else.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
brunoapimentel: I agree with what Erik is proposing here. We should take the GAV as input for maven artifacts, which enable us to generate proper SBOMs.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
aloubyansky: > this design proposes that only resolved download_urls are used as input

Thanks, I overlooked that.

> If we consider Cachi2 as a standalone project, its goal is to prefetch the dependencies necessary for a hermetic build, and report them the most accurate way possible, but only to the extent of the info it has at hand during the prefetch.

Agreed.

> Allowing purls to be passed through creates a precedent that goes against our ultimate goal of accurate SBOMs.

Could you please elaborate on that?

Whatever Cachi2 produces will be produced (ultimately) for the user who provides the input. The user will be responsible for the quality of the input and, as a consequence, the output.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
ben-alkov: If nothing else, based on `Mapping famous and public registries will prevent the use of private registries. To solve this, we need to allow users to change Cachi2's default configuration to enable additional source URLs for each purl type`, there's **no way** to _always_ infer package type.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
brunoapimentel: > Whatever Cachi2 produces will be produced (ultimately) for the user who provides the input. The user will be responsible for the quality of the input and, as a consequence, the output.

I agree the user has substantial responsibility towards the quality of the input/output, but Cachi2 has traditionally been on the restrictive side of things. For instance, for `pip`, Cachi2 simply won't process repos that don't have a fully resolved requirements file (with pinned versions and hashes). Cachi2 will never execute a `setup.py` file because this allows the execution of arbitrary code, which undermines our certainty that the only downloaded files were exactly the ones that are described in a certain ref of a repo.

Having Cachi2 receive a purl as input, resolve it to download an artifact, and simply copy paste the purl to the output SBOM seems to be opening a wide scope of things to happen. From the accuracy perspective, it could allow from simple "typos" to non-compliant purls to be passed through. From a practical standpoint, we would still need to validate the purls to an extent, and to keep up with the purl-spec.

Looking from another perspective, what we're proposing here is already allowing the user to specify the purl (with a limitation of the subset of purl types and details about the ouput). The decision here is mostly to not have purls as input, and to limit what can be done as output. So my question is, would the output purl for maven proposed be outside of what you expect? Are there any corner cases of concern?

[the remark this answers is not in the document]
brunoapimentel: I would rather Cachi2 to keep focused on package manager environments and have only the generic fetcher (with `pkg:generic` purls) as an alternative to get other type of content into a build. Anything besides that does feel a little out of scope of "prefetching to enable hermetic builds, download only what is clearly defined in the repo, and accurately report what was prefetched with the best info we have".

However, the maven artifacts use case pushes things further than that. What we're trying to achieve with this proposal is to reach a wider scope so we're not just special-casing maven artifacts. Cachi2 is still an open-source project that tries to reach a wide range of use cases.

I also think it's important to add that generic artifact fetching is a very new ground for this project, so I think we're still learning as we go.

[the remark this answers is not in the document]
ben-alkov: I think the point here is that one of cachi2's core responsibilities is exactly "being responsible for the output" _in spite of_ what users do. If users were good at setting up the proper conditions for a hermetic build (or if it were easy to start with :roll_eyes: ), cachi2 would be out of a job :grin:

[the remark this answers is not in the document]
aloubyansky: A lot of other comments on this proposal seem to suggest it'd be a major struggle to produce a non-generic PURL from a generic one, which creates a major problem if the goal is to have an accurate SBOM, which we are all after here.

In case of Maven, it's as straightforward as it can be to go from a PURL to the download URL but not always the other way around. If we have to provide a generic PURL for a Maven artifact then we also have to add extra config to provide enough instructions for Cachi2 to generate a proper Maven PURL. For the Maven use-case this adds a lot of complexity for Cachi2 to implement and user configs. Just in case, I'll have ~6500 Maven artifacts in a lockfile. It'd be a lot easier to simply express essentially the same info in a single PURL.

Now I am not sure how different this would be for other component types.

docs/design/generic-enhanced-purls.md
eskultety: I think we can mark this one as resolved as based on the https://github.com/containerbuildsystem/cachi2/pull/692#discussion_r1810861033 thread is seems like we'll need to keep adding more PURL-type related attributes.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
eskultety: > Sometimes, the user will have a better purl than what Cachi2 can provide

That's the sad reality of the PURL spec. In ~~an ideal~~ a normal world the spec (like any other spec) itself **would define** what a perfect PURL looks like for each artifact type and with cachi2 strictly following the spec this statement would hold no ground; PURL spec doesn't live in that world :( .

> Whatever Cachi2 produces will be produced (ultimately) for the user who provides the input. The user will be responsible for the quality of the input and, as a consequence, the output.

@aloubyansky this is AFAIK called tainting in our field, i.e. consuming user input with little to no validation and use it for output generation (in this case taking the PURL verbatim) and as such tainting is undesirable in general and so we'd like to avoid it by forcing users to provide the information in form of attributes leading to the desired PURL on the output. This is especially problematic with the PURL spec, which, like I mentioned earlier somewhere the spec is quite vague for some artifact types and we'd have to trust user's judgement on the quality of an input PURL.

>  I'll have ~6500 Maven artifacts in a lockfile. It'd be a lot easier to simply express essentially the same info in a single PURL.
> 
> Now I am not sure how different this would be for other component types.

@aloubyansky I guess you kinda answered your concerns with ^this - cachi2 cannot tailor a solution to a single use case knowing not all consumers would do their due diligence when it comes to providing input using PURLs straightaway.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
eskultety: > Will Cachi2 actually validate that they correspond

Since the motivation here is to understand individual PURL spec types (like I mentioned, I am not particularly fond of the maintenance burden, but it is what it ise) we should understand them properly and validate them, whatever that means in the Maven world, at least that's my understanding of it. It's the disadvantage of maintaining our own manifest format for these purposes.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
aloubyansky: I honestly don't see yet how this proposal is better from the "undesirable tainting" perspective.
Could we have an example side-by-side of a final PURL vs a generic one + extra info that would show the advantage of the latter?

Cachi2 is already hooked on PURLs either way.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
aloubyansky: I wouldn't like to put extra burden on Cachi2 maintenance. So far, it seems to me this proposal creates more complexity than the alternative.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,150 @@
...
+                    "type": "distribution",
+                    "url": "https://github.com/containerbuildsystem/cachi2/archive/refs/tags/0.11.0.tar.gz"
+                }
+            ],
+            "type": "file"
+        },
+        {
+            "name": "quakus-core",
+            "version": "3.8.5.internal-00004",
+            "purl": "pkg:maven/ga.io.quarkus/quarkus-core@3.8.5.internal-00004?type=jar&repository_url=https://maven.repository.internal.com&checksums=sha1:e4ca5fadf89e62fb29d0d008046489b2305295bf",
```
aloubyansky, line 116:
```suggestion
            "purl": "pkg:maven/io.quarkus/quarkus-core@3.8.5.internal-00004?type=jar&repository_url=https://maven.repository.internal.com/ga&checksums=sha1:e4ca5fadf89e62fb29d0d008046489b2305295bf",
```

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
kosciCZ: > i.e. consuming user input with little to no validation and use it for output generation (in this case taking the PURL verbatim) and as such tainting is undesirable in general and so we'd like to avoid it by forcing users to provide the information in form of attributes leading to the desired PURL on the output.

I don't think this is true. Cachi2 is quite literally in control of how much validation it does. It does not have to take the purl verbatim, it is free to decompose it into individual attributes and validate those. 

> I guess you kinda answered your concerns with ^this - cachi2 cannot tailor a solution to a single use case knowing not all consumers would do their due diligence when it comes to providing input using PURLs straightaway.

You're thinking about other use cases and that is good. However, those are so far all theoretical, while this is a very real one. I can do some research and give you the data, if other purl types convert to url in similar way, if that's the issue. 

Other than that, in the case of the custom lockfile for the generic fetcher, you suggested that we use SBOMs instead, which would likely involve letting users supply the purl as part of those SBOM components. I can't see why now, when there's decision between an established format for specifying packages (purl) and a custom implementation that is essentially url + some custom attributes, your reasoning is different.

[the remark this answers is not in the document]
aloubyansky: >  I can do some research and give you the data, if other purl types convert to url in similar way, if that's the issue.

I think it'd be worth it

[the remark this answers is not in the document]
eskultety: > I don't think this is true. Cachi2 is quite literally in control of how much validation it does. It does not have to take the purl verbatim, it is free to decompose it into individual attributes and validate those.

Does such decomposition involve detection of all extra non-spec qualifiers that we could flag and forbid? Because with YAML attributes this is as easy as validating the lockfile model and failing early and loudly about unrecognized attributes. Besides, I have a feeling that accepting a PURL is inviting the user to expect that, well, since we already consumed the PURL we just use it, right? Not really. Admittedly, either way this is an implementation detail.

> Other than that, in the case of the custom lockfile for the generic fetcher, you suggested that we use SBOMs instead, which would likely involve letting users supply the purl as part of those SBOM components. I can't see why now, when there's decision between an established format for specifying packages (purl) and a custom implementation that is essentially url + some custom attributes, your reasoning is different.

I don't think it is different at all, actually. The proposal of input SBOMs was shut down as too complex and we moved on to a different approach. I don't think I ever advocated for input PURLs anywhere (going back to https://github.com/containerbuildsystem/cachi2/pull/652 and more specifically this thread: https://github.com/containerbuildsystem/cachi2/pull/652#discussion_r1762653777), for me it was the future proof type of input when it comes to the plethora of standardized attributes we could consume and look at during artifact processing from an input SBOM, PURLs are just an inherent part of SBOMs which we'd get as a side effect we could not do anything about at all if we had decided to consume it, so I think we may have misunderstood each other in terms of my stance towards PURLs in general. However, this is a new proposal where we have the choice for all input data type/format.

> I can do some research and give you the data, if other purl types convert to url in similar way, if that's the issue.

I'm not sure I follow the outcome here, I guess you mean if other PURLs have a direct download URL mandatory? Is that what you're saying you'd like to check?

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Consuming input purls
+
+A purl is simply a way to represent a unique package and it's location on the internet. By allowing users to specify a purl as part of a package definition in the lockfile, Cachi2 can verify that it actually resolves to a valid download location, and that it adheres to the fundamentals of the purl spec. Note that there's a great variety of purl types, and resolving them will involve type-specific logic, which means we'll only be able to support a small subset of purl types.
+
+This is a summary of how the resolution of a single package will look like:
+
+- Instead of consuming a `download_url`, the package will specify a `purl` attribute in the lockfile
+- Parse the purl by using the [packageurl-python](https://github.com/package-url/packageurl-python) library
+- Validate that the purl is within the supported types
```
arewm, line 15: Earlier you mentioned that purls are reported as generic. Would Cachi2 also support a generic purl as input? If the supported types mentioned below are not used, would Cachi2 try to just fall back to fetch a generic artifact?

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+- Resolve the purl to a download URL by using a type-specific algorithm
+- Download the file
+- Validate the file's checksum
+- Generate a SBOM component containing the input purl
+
+### What confidence Cachi2 has in the reported components
+
+- The purl is valid
```
arewm, line 23: This is confidence that the purl is syntactically valid?

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+- The purl resolves to an actual URL that contains an artifact
+- The checksum was validated
```
arewm, line 25: These both refer to the artifact fetched, right? Would something like this be appropriate?
```suggestion
- The purl resolves to an actual URL where the artifact was fetched from
- The artifact's checksum was validated 
```

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
+# Support for different purl types in the generic artifact fetcher
+
+The generic artifact package manager is being added to Cachi2 as a means for users to introduce files that do not belong to traditional package manager ecosystems (e.g. pip, npm, golang) to their hermetic container builds. Since Cachi2 does not have any extra information about the file that's being fetched, the purls are always reported as [pkg:generic](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#generic).
```
arewm, line 3: Is this a generic purl type and/or a generic property, i.e.
```
    "properties": [
        {
            "name": "cachi2:found_by",
            "value": "cachi2:generic"
        }
```

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+            "url": "quay.io/konflux-ci/buildah-task@sha256:b2d6c32d1e05e91920cd4475b2761d58bb7ee11ad5dff3ecb59831c7572b4d0c"
+        }
+    ],
+    "type": "container"
+}
+```
+
+## Initial thoughts
+
+From a Cachi2 perspective, we can separate purls types into ones that are part of existing package manager ecosystems (such as nuget, composer, maven) and ones that are not (github, huggingface, oci):
```
arewm, line 223: What benefit do we have from this separation? Do we do anything with this classification?

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+- Types that refer to existing package managers seem to have a straightforward resolution. The package is likely a file in a predictable URL within a registry.
+- Types that don't refer to existing package managers are trickier to resolve, since they might be poiting to a container image or a full git repository. They might also need extra tools to provide a smooth resolution (podman, git, etc)
+
+
+### Decision points
```
arewm, line 228: I am not familiar with the review process. Are these open questions that we should try to resolve?

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+
+- Should the checksums be specified as part of the input purl?
```
arewm, line 230: What are the Cachi2 requirements for the checksums for other package managers? This should at least be a best practice for Cachi2 users if it isn't a requirement.

arewm (commented):
Do we have a record of what confidence Cachi2 has for the reported contents for the various supported package managers outside of this proposal?

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Maven
+
+#### generic_artifacts.yaml
+
+```yaml
+metadata:
+  version: 1.0.0
+artifacts:
+  - purl: pkg:maven/ga.io.quarkus/quarkus-core@3.8.5.redhat-00004?type=jar&repository_url=https://maven.repository.redhat.com&checksum=sha256:e4ca5fad
```
aloubyansky, line 38:
```suggestion
  - purl: pkg:maven/io.quarkus/quarkus-core@3.8.5.redhat-00004?type=jar&repository_url=https://maven.repository.redhat.com/ga&checksum=sha256:e4ca5fad
```

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+https://{repository_url}/{as_dir(group_id)}/{artifact_id}/{version}/{artifact_id}-{version}[{-classifier}].{extension}
+```
+
+#### SBOM component
+
+```json
+{
+    "name": "quarkus-core",
+    "version": "3.8.5.redhat-00004",
+    "purl": "pkg:maven/ga.io.quarkus/quarkus-core@3.8.5.redhat-00004?type=jar&repository_url=https://maven.repository.redhat.com&checksum=sha256:e4ca5fad",
```
aloubyansky, line 57:
```suggestion
    "purl": "pkg:maven/io.quarkus/quarkus-core@3.8.5.redhat-00004?type=jar&repository_url=https://maven.repository.redhat.com/ga&checksum=sha256:e4ca5fad",
```

docs/design/generic-enhanced-purls.md
brunoapimentel: We will require checksums, it is one of the few validations we have in place. The question here is about where the checksum should be in the lockfile. A generic artifact looks like this:
```yaml
- download_url: 
   checksums: https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
        sha256: 90bffe1884b84d5e255f12ff0ecbd70f2edfc877b68d612dc6fb50638b3ac17c
```

The artifact with a specific type will have a `purl` instead of the `download_url`, so should we drop the extra `checksums` attribute and make it part of the `purl`?

docs/design/generic-enhanced-purls.md
brunoapimentel: This is a draft, so I wrote this essentially as a way to express my own questions given the current state the design is. I'd say we want to resolve them and incorporate the answers in the design.

docs/design/generic-enhanced-purls.md
brunoapimentel: We don't do anything with them. I initially had a hard time wrapping my head around non-package manager types (oci, github, huggingface), because they don't simply point to a file in a URL. I initally assumed we don't want to allow this feature to cover other existing package managers, because we'd want full support for a package manager.

I think the question here is: how flexible we want this feature to become. It started as a way to fetch a files, but do we want to extend it to OCI artifacts or even git repos?

docs/design/generic-enhanced-purls.md
brunoapimentel: In this case, a generic purl type. The property is a way to inform which backend/package manager in Cachi2 produced that SBOM component.

docs/design/generic-enhanced-purls.md
brunoapimentel: Yes, I added this bit to this sentence.

docs/design/generic-enhanced-purls.md
brunoapimentel: :+1:

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+From a Cachi2 perspective, we can separate purls types into ones that are part of existing package manager ecosystems (such as nuget, composer, maven) and ones that are not (github, huggingface, oci):
+- Types that refer to existing package managers seem to have a straightforward resolution. The package is likely a file in a predictable URL within a registry.
+- Types that don't refer to existing package managers are trickier to resolve, since they might be poiting to a container image or a full git repository. They might also need extra tools to provide a smooth resolution (podman, git, etc)
+
+
+### Decision points
+
+- Should the checksums be specified as part of the input purl?
+- Should we limit the qualifiers to the types that are strictly available to that type?
+- Should we allow types that are not files (git repos, OCI artifacts)? Should they be reported as different component types?
```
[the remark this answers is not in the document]
brunoapimentel: I think this is completely up to discussion. I can see it working either way, but since it's supposed to be a short-lived cache, maybe it wouldn't be helpful to tar it.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+
+From a Cachi2 perspective, we can separate purls types into ones that are part of existing package manager ecosystems (such as nuget, composer, maven) and ones that are not (github, huggingface, oci):
+- Types that refer to existing package managers seem to have a straightforward resolution. The package is likely a file in a predictable URL within a registry.
+- Types that don't refer to existing package managers are trickier to resolve, since they might be poiting to a container image or a full git repository. They might also need extra tools to provide a smooth resolution (podman, git, etc)
+
+
+### Decision points
+
+- Should the checksums be specified as part of the input purl?
+- Should we limit the qualifiers to the types that are strictly available to that type?
```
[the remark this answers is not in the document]
brunoapimentel: Yes, the purl specification loosely determines what are the qualifiers for each type. See [oci,](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci) for example. 

There are also global qualifiers that can be applied to any type (check the qualifiers bullet [here](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component)).

docs/design/generic-enhanced-purls.md
brunoapimentel: I think we should keep the current lockfile option of provinding `donwload_url` and `checksums` for the `pkg:generic` cases, but we can also support consuming purls as input for the sake of consistency.

As for the types, I think we need to be very explicit about which ones are supported, and anything that falls outside of that means a failed request.

arewm: So the only way that users would be able to prefetch in a way that produces a generic purl would be through the current lockfile option? There is no proposal to add a purl-supported generic package fetching?

docs/design/generic-enhanced-purls.md
arewm: Would it be possible to add an example purl that Cachi2 might generate with an input that could produce it?

docs/design/generic-enhanced-purls.md
arewm: Thanks.

docs/design/generic-enhanced-purls.md
arewm: Thanks. I cannot resolve threads apparently.

docs/design/generic-enhanced-purls.md
arewm: Any of these artifact types vary with how metadata is stored and associated with the content that is downloaded. 

Being able to specify/support purl types that could match to future package manager support is one way to simplify the required work for fetching content. The downside to this (from a users' perspective) is that the native dependency files for the package managers are not supported. If, in the future, a package manager-native way to resolve dependencies is developed, it would be an option for the Cachi2 tooling to deprecate the generic purl-based fetching mechanism. This deprecation would, of course, require a breaking change.

arewm: I would consider flexibility to be driven by a case-by-case basis. If there is requested support for fetching some form of artifact, is it required to have the native dependency format (i.e. lockfiles)? Or is it possible/reasonable to implement the fetching with a generic purl-based approach? Implementation with a purl-based approach seems like it would at least be faster if that is an acceptable solution.

docs/design/generic-enhanced-purls.md
arewm: If the checksum is a simple/standard addition to a purl, it makes sense to require it. The usability of keeping the checksum separate would just be to make it simpler to see/update it.

Would it be reasonable to support both flows? If the checksum is specified separately, Cachi2 would still generate the purl with the checksum to inject into the SBOM.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
arewm: I think I would expect the qualifiers to be verified as part of the parser. Cachi2 would need to ensure that the required qualifiers are present and that the optional ones may be present.

Which of the package-specific qualifiers that are directly specified would Cachi2 use? For example, would OCI artifact support be able to handle the repository_url, namespace, version, ... or would the input purl provide those _plus_ a download_url? Or would support for these qualifiers be considered more for "full" package manager support instead of generic?

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
arewm: After further thought, I agree that tarring would likely not be helpful. If tarring is desired, that should be handled by some other tool/process.

Cachi2 should strive to fetch artifacts which can be "quickly" be used by various tools. This means that git repos would likely not be tarred and that OCI artifacts would likely be stored as on-disk OCI.

docs/design/generic-enhanced-purls.md
ben-alkov: > simplify the required work for fetching content

We want to make sure, for anything with an *existing* package manager, that we don't break builds, i.e. anything which happens *after* the fetch.

For instance: if we have to handle some arbitrary level of complexity to resolve the dependency graph we need to download, and/or some arbitrarily complex setup step is involved (e.g. a specific filesystem layout)... we're better off simply using the existing package manager ecosystem and it's lockfiles to resolve the dependency graph, get URLs, download to correct locations from those URLs, etc...

To a first approximation, cachi2 is a very smart "package manager wrapper" - our secondary brief is to make sure that e.g. `go build`, `pip install -r foo`, `yarn` (er... build? >_<) *succeed*, and our **primary** brief is that ***we*** control the downloads which the build relies on.

docs/design/generic-enhanced-purls.md
aloubyansky: FWIW, CycloneDX components have dedicated fields for checksums (hashes). Perhaps both ways are supported. The CDX maven generator includes ~8 checksums. Probably more than necessary and it doesn't add them to purls, which is probably not a bad idea in such a case.

docs/design/generic-enhanced-purls.md
brunoapimentel: There isn't, but we can add one. Do you think it's worth doing it now?

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
brunoapimentel: I believe the specification for [oci](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci) seems to indicate the use of `repository_url`.

I changed the design so that it mentions we will validate the qualifiers depending on type, it is probably better than to leave it completely open.

docs/design/generic-enhanced-purls.md
brunoapimentel: I changed the design so that checksums are only consumed as separate attributes and only a single one of them is reported in the output purl. This is mainly to keep consistency with other existing package managers: we always report a single checksum in case of file dependencies.

We can extend this to support checksums in the purl, in case the need comes up.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,233 @@
...
+- Should we allow types that are not files (git repos, OCI artifacts)? Should they be reported as different component types?
+- How should our policy be regarding extending the generic fetcher for other package managers we don't fully support? Would this impact the will of contributors to provide full support for a package manager?
```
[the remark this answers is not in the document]
brunoapimentel: I was thinking about this. If we compare it to most package managers that allow file dependencies (e.g. npm, pip), in the end, we're still using `aiohttp` to download the files, the main difference is that the dependency is defined in the expected lockfile for that package manager (pacakge.json, requirements.txt), which gives a little more confidence on the package type. I don't think this means we have any other guarantees, nor greater confidence in fidelity or integrity. Any file downloaded will have it's checksum validated, so we're sure we're getting the right content.

From this reasoning, we could probably even accept the generic fetcher to download `pip` or other supported package manager artifacts, which would only be useful in case a user needed a single artifact and did not want to initialize a project in his repo.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,286 @@
...
+        {
+            "type": "distribution",
+            "url": "quay.io/konflux-ci/buildah-task@sha256:b2d6c32d1e05e91920cd4475b2761d58bb7ee11ad5dff3ecb59831c7572b4d0c"
+        }
+    ],
+    "type": "container"
+}
+```
+
+## Alternative: consume options instead of a purl
```
brunoapimentel, line 221: @aloubyansky How do you feel about this alternate proposal? Would it be difficult to generate the lockfiles for PNC/Maven artifacts?

My initial feeling is that having the attributes broken down is clearer and avoids the need to parse the purl according to the spec.  The flexibilty to define the output purl would still exist, since we can add any needed attributes in order to generate the purl we want.

docs/design/generic-enhanced-purls.md
arewm: I don't think that it is Cachi2's job to ensure that builds are not broken if a purl-driven package manager was used. This should be handled by users' specifications of what content to download ... whether it is by specifying a path to a lockfile or a set of purls. It does, however, seem reasonable that users expectations are set based on the choice of interface.

For example, you may have a lockfile that only has one tarball: https://github.com/stolostron/image-builder/blob/master/yarn-source/package-lock.json

A user should be able to use the npm package manager to resolve that if they want to maintain the package lock. If they don't want to maintain the package lock, they should be able to just fetch that tarball.

Similarly, if there is not support for a specific package manager in Cachi2 but a user wants to prefetch all content, they should be able to take the complexity to individually specify all resolved packages for their builds. Cachi2 wouldn't know that there is complete resolution of artifacts but that is irrelevant.

If Cachi2 wants to record how dependencies were defined (i.e. if they were resolved with a package manager or defined individually), it seems reasonable that the generated SBOM would have metadata to indicate how the specific dependency was resolved.

docs/design/generic-enhanced-purls.md
arewm: That makes sense. Presumably Cachi2 would verify all checksums that are provided and it would report the "most secure" one, i.e. sha256  over md5.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
arewm: I haven't read the content that you have changed yet, but that change makes sense.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
arewm: You can resolve this comment thread if you want.

docs/design/generic-enhanced-purls.md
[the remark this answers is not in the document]
arewm: My comment earlier is also relevant for this thread: https://github.com/containerbuildsystem/cachi2/pull/692#discussion_r1829983150

What you say here makes sense and I think it matches with my mental model of the interface/results.

docs/design/generic-enhanced-purls.md
arewm: I haven't read the alternative yet, but based on your comment, it is effectively the effect your comment above: https://github.com/containerbuildsystem/cachi2/pull/692#discussion_r1828416919?

It is a recognition that different purl types will have different field requirements, so you would expose _those_ fields as being required?

It seems like it would be harder to validate the validity of fields because it could be possible for a user to add content to an attribute which would end up matching some other part of the purl spec (i.e. if control characters are not properly escaped). If you default to consuming a purl itself, then you could validate the parsing of the purl and then only update those specific sub-elements which are relevant for cachi2 to control (i.e. uri/hash).

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,286 @@
...
+#### generic_artifacts.yaml
+
+```yaml
+metadata:
+  version: 1.0.0
+artifacts:
+  - purl: pkg:maven/io.quarkus/quarkus-core@3.8.5.redhat-00004?type=jar&repository_url=https://maven.repository.redhat.com/ga
+    checksums:
+      sha256: d16bf783cb6670f7f692ad7d6885ab957c63cfc1b9649bc4a3ba1cfbdfd5230c
+    target: quarkus.jar
```
arewm, line 42: Is this target the location on-disk for where the fetch artifacts would be put? Would all artifacts need to specify this so that they can be appropriately used by some later stage or would there be default/assumptions for package managers about locations and file names?

docs/design/generic-enhanced-purls.md
brunoapimentel: No, I did not write anything about consuming input `pkg:generic` purls here because this is already supported by the current lockfile. So, initally, Cachi2 would not consume `pkg:generic` purls and for this reason, users wouldn't be able to produce `pkg:generic` SBOM components by providing input purls.

The mechanism to extend support to any purl type it is already described here, though, so supporting input `pkg:generic` is trivial (and we probably should, for the sake of consistency).

docs/design/generic-enhanced-purls.md
brunoapimentel: Do you need an example different from the ones in https://github.com/containerbuildsystem/cachi2/pull/692/files#diff-aadc5e0bf6d8c1eefd2742bad409f3f2fb7a684aa7cbe0d47ebfb752e07d9fd6R31?

arewm: I guess it is sufficient. I cannot remember why I was asking specifically. This can be resolved.

docs/design/generic-enhanced-purls.md
arewm: To record some parts of a conversation that I had out-of-band:

* I don't see an issue with this approach, but you should ensure that you are making the decision for the proper reason.
* With this method, I think it would still be required to validate the entire purl after it is assembled. Therefore, I don't think that this would alleviate the need to parse a purl
* This seems to mostly affect user input. Is it a better user experience to provide values based on some set of required/optional keys? Would users be more likely to do this properly than if they were to specify the purl in its entirety? 
* Will the verbosity (i.e. the increased number of lines in the configuration file) be an issue? Or would it be a desired feature?

eskultety:
@brunoapimentel this PR has been dormant for a while, but has quite a few comments in the discussion, is this work still relevant to us or has the desire for it faded and we can potentially close it for the time being as "not planned" ?

eskultety:
@brunoapimentel do you think this PR is still worth pursuing in context of an alternative approach that IIUC is supposed to supersede the Maven support we currently have outlined here: https://github.com/hermetoproject/hermeto/pull/1046? Or could we potentially close this as not planned?

brunoapimentel:
> @brunoapimentel do you think this PR is still worth pursuing in context of an alternative approach that IIUC is supposed to supersede the Maven support we currently have outlined here: #1046? Or could we potentially close this as not planned?

This work was [implemented](https://github.com/hermetoproject/hermeto/pull/735) as the current Maven extension for the generic fetcher. So, to my mind, this PR is the design for that. I am not sure why we didn't merge it back then.

So I think we should merge it to serve as a historical document supporting #735. If you have a strong inclination to just close this, I don't mind. In any case, I believe that for as long as we support the Maven extension, this proposal is still in effect. (How would its existence be justified otherwise? Just an one-off feature?)

eskultety:
> This work was https://github.com/hermetoproject/hermeto/pull/735 as the current Maven extension for the generic fetcher. So, to my mind, this PR is the design for that.

Makes sense. @brunoapimentel would you please replace the cachi2 references with Hermeto (or drop them altogether where appropriate) and quickly refine the contents of the doc for correctness? I'll re-review.

brunoapimentel:
> > This work was #735 as the current Maven extension for the generic fetcher. So, to my mind, this PR is the design for that.
> 
> Makes sense. @brunoapimentel would you please replace the cachi2 references with Hermeto (or drop them altogether where appropriate) and quickly refine the contents of the doc for correctness? I'll re-review.

@eskultety Done. Please re-review.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,271 @@
...
+- Fetch the artifact
+- Verify that at least one of the provided checksums match
+- Generate a purl based on the input attributes
+- Generate a SBOM component containing the type-specific purl and the validated checksums as qualifiers
+
+### What confidence Hermeto has in the reported SBOM components
+- The attributes were resolved to a valid download URL where the artifact was fetched from
+- At least one of the provided checksums was validated
+- There is no guarantee over the reported purl type, Hermeto will trust the user input
+
```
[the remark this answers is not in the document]
brunoapimentel: :rabbit2: Doesn't know what it's talking about here, so I'm just dismissing this comment.

eskultety (approved):

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,271 @@
...
+#### processing the lockfile
+```python
+# generate the download URL
+{repository_url}/{lowercase(namespace + '.' + package)}.{version}.nupkg?packageVersion={version}
+
+# generate the purl
+pkg:nuget/[{namespace}.]{package}@{version}[?repository_url={repository_url}]
+```
+
```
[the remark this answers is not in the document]
eskultety: @brunoapimentel the bot is correct here.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,271 @@
...
+pkg:huggingface/{namespace}/{project}@{commit_hash}[?repository_url={repository_url}][&file_name={file_name}]
+```
+
+#### SBOM component
+```json
+[
+  {
+    "name": "Llama-3.1-Nemotron-70B-Instruct-HF",
+    "purl": "pkg:huggingface/nvidia/Llama-3.1-Nemotron-70B-Instruct-HF@043235d6088ecd3dd5fb5ca3592b6913fd516027",
+    "properties": [
```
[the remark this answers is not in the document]
eskultety: @brunoapimentel I think the bot is right with 2. Dismissing 1. as we don't encode default repository urls.

docs/design/generic-enhanced-purls.md
```diff
@@ -0,0 +1,271 @@
...
+```
+
+#### processing the lockfile
+```python
+# generate the download command
+podman pull docker://{repository_url}/{namespace}/{name}@{digest}
+
+# generate the purl
+pkg:oci/{name}@{encode(digest)}?repository_url={repository_url}/{namespace}
+```
```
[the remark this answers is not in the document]
eskultety: @brunoapimentel also correct.

brunoapimentel:
This was forgotten here for a while, so I'm closing it because I don't think we'll ever merge it. 

There is one implementation of this that is the Maven extension to the generic fetcher, but I personally think we should not keep adding any more of these (and eventually even remove the Maven one and keep the generic fetcher simply producing generic purls).

From a pure Hermeto perspective, Hermeto is writing in the purl exactly what it knows about the package: it is a package fetched from a URL, not necessarily tied to any ecosystem, and no additional processing was done on it besides checking the checksum.

PR #705, opened 2024-10-23
feat(yarnv1): read package.json  & yarnlock

ben-alkov:


cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,294 @@
...
+    def install_config(self) -> Optional[dict[str, Any]]:
+        """Get the installConfig dict."""
+        return self.data.get("installConfig", "")
+
+    @classmethod
+    def from_file(cls, path: RootedPath) -> "PackageJson":
+        """Parse the content of a package.json file."""
+        try:
+            with path.path.open("r") as f:
+                package_json_data = json.load(f)
```
a-ovchinnikov, line 94:
```suggestion
package_json_data = json.loads(path.path.read_text())
```

a-ovchinnikov (commented):
A few minor suggestions and some questions.

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,294 @@
...
+    def data(self) -> dict[str, Any]:
+        return self._data
+
+    @property
+    def path(self) -> RootedPath:
+        return self._path
+
+    @property
+    def config_type(self) -> Literal["package_json", "yarnrc", "yarnlock"]:
+        if isinstance(self, PackageJson):
```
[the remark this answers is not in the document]
ben-alkov: :thinking: That's nice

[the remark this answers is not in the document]
ben-alkov: @a-ovchinnikov; As always, the Devil is in the (typing) details...

```
cachi2/core/package_managers/yarn_classic/project.py:68: error: Signature of "_self_type" incompatible with supertype "_CommonConfigFile"  [override]
cachi2/core/package_managers/yarn_classic/project.py:68: note:      Superclass:
cachi2/core/package_managers/yarn_classic/project.py:68: note:          Never
cachi2/core/package_managers/yarn_classic/project.py:68: note:      Subclass:
cachi2/core/package_managers/yarn_classic/project.py:68: note:          str
```

Thoughts?

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Nice! I need to get more familiar with `Path`...

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,294 @@
...
+        """Parse the content of a string containing the contents of a valid yarnrc file."""
+        yarnrc_data = {}
+        try:
+            for line in string.splitlines():
+                line.strip()
+                if line:
+                    key_and_val = line.split(" ", 1)
+                    yarnrc_data[key_and_val[0]] = key_and_val[1]
+        except ValueError as e:
+            raise PackageRejected(
```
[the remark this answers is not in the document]
ben-alkov: I see what you're saying; this is the pattern I copied from Yarnberry. If we have an error parsing any config file *for the current package*, we can't continue, so `PackageRejected` is appropriate.

Isn't it? :laughing:

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,294 @@
...
+                ),
+            )
+
+        if not yarnrc_data:
+            # warn
+            pass
+
+        return yarnrc_data
+
+    def write(self) -> None:
```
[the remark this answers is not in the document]
ben-alkov: '.yarnrc' and 'package.json' both have `write()` methods in cachi2's Yarnberry implementation.

cachi2/core/package_managers/yarn_classic/project.py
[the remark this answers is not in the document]
a-ovchinnikov: From a purist's perspective we should be raising ParsingError here which must be converted to PackageRejected at caller's level. I am not a purist and won't block on this.

cachi2/core/package_managers/yarn_classic/project.py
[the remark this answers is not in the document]
a-ovchinnikov: Actually it seems like we have overthought this. The following code should do the trick:
```python
from abc import ABC, abstractmethod
ConfigTypes = Literal["package_json", "yarnrc", "yarnlock"]

@dataclass
class _CommonConfigFile(ABC):
    ...
    @property
    @abstractmethod
    def config_type(self) -> ConfigTypes:
        pass
...

class YarnRc(_CommonConfigFile):
...
    @property
    def config_type(self) -> ConfigTypes:
        return "yarnrc"
```

I think you want `_CommonComfig` to be semi-abstract, so it makes sense to subclass it from ABC too and mark `config_type` as an abstract property. All children will have to define the method and return correct literal.

cachi2/core/package_managers/yarn_classic/project.py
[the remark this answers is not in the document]
a-ovchinnikov: I have not checked yet, but I have a strong suspicion that this would break the existing pattern. Refactoring half of PMs is definitely not in scope, so we likely can keep it as is for now.

cachi2/core/package_managers/yarn_classic/project.py
eskultety: Note that this is only usable in cases where we **have to** load the whole file from disk to memory, in all other cases a context manager is preferred, that's why I prefer context managers for file ops, because the reads always the same, no surprises :) .

cachi2/core/package_managers/yarn_classic/project.py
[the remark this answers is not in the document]
eskultety: > Why would we want to write to any of these files? Should not they be read-only for us?

We actually do write them in the temp clean source repo clone when fetching & validating to disable some default Yarn functionality. I don't remember whether everything could have been set with env variables, but we do write the config actually before interacting with Yarn V2 since users are able to ship their own Yarn shim in their repo. IIRC plugins can't be disabled/set via env variables and so they have to be overridden via `.yarnrc` and so I presume we decide to write the config instead of doing both env vars AND the config.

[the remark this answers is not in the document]
eskultety: @brunoapimentel wrote that specific bit of code for Yarn v2 so his memory is probably better than mine on that matter.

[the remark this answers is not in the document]
taylormadore: In the **_initial implementation_** for yarn 1.x support, we shouldn't need to write these files, but that may change in the future with follow-ups

ben-alkov:
Implemented `_check_pnp_installs`, `is_pnp_install`; removed all `write()` methods; added basic unit tests

cachi2/core/package_managers/yarn_classic/project.py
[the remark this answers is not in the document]
ben-alkov: Removed all `writes()`

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,294 @@
...
+
+        yarnrc_data = {}
+        try:
+            with path.path.open("r") as f:
+                for line in f:
+                    line = line.replace("\n", "").strip()
+                    if line:
+                        key_and_val = line.split(" ", 1)
+                        yarnrc_data[key_and_val[0]] = key_and_val[1]
+        except ValueError as e:
```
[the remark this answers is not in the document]
ben-alkov: Removed; yarnrc is out of scope here

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,294 @@
...
+        but paths don't have to be quoted(?!)
+
+        `--cache-folder /tmp/yarn-cache/`
+        """
+
+        yarnrc_data = {}
+        try:
+            with path.path.open("r") as f:
+                for line in f:
+                    line = line.replace("\n", "").strip()
```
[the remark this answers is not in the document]
ben-alkov: Removed; yarnrc is out of scope here

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Updated

cachi2/core/package_managers/yarn_classic/project.py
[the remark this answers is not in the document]
ben-alkov: Fixed, thanks @a-ovchinnikov!

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class Project(NamedTuple):
+    """A directory containing yarn sources."""
+
+    source_dir: RootedPath
+    package_json: PackageJson
+
+    @property
+    def is_pnp_install(self) -> bool:
```
taylormadore, line 163: This conditional is a bit dense. Can it be broken down to make it more readable? Maybe some clearly-named, individual boolean variables for each of the conditions and then combine them for the return

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,185 @@
...
+    is relevant for the request processing.
+    """
+
+    @property
+    def config_kind(self) -> ConfigKind:
+        """Return kind of this ConfigFile."""
+        return "package_json"
+
+    @property
+    def package_manager(self) -> Optional[str]:
```
taylormadore, line 70: At least for the initial implementation, I don't think we'll need to get/set the packageManager field in package.json

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -91,11 +90,55 @@ def _generate_build_environment_variables() -> list[EnvironmentVariable]:
     return [EnvironmentVariable(name=key, value=value) for key, value in env_vars.items()]
 
 
+def _check_pnp_installs(project: Project) -> None:
+    if project.is_pnp_install:
+        raise PackageRejected(
+            reason=("Yarn zero install detected, PnP zero installs are unsupported by cachi2"),
```
taylormadore, line 96: I think "zero-installs" are a term/feature that only really exists in Yarn v2+, so we might want to avoid using it here for Yarn 1.x. Same with the solution text mention of .yarn/cache

ben-alkov: Gah. Copy-pasta.

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Fixed

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: removed

cachi2/core/package_managers/yarn_classic/main.py
ben-alkov: Fixed

ben-alkov:
We've got 80% coverage of new code.

I figure somewhere between 2-5 more unit tests should cover everything, and they'll all be pretty straightforward.

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,176 @@
...
+        - the presence of an expanded node_modules directory
+        For more details on PnP, see: https://classic.yarnpkg.com/en/docs/pnp
+        """
+        install_config_pnp_on = self.package_json.install_config.get("pnp", False) is True
+        pnp_cjs_exists = any(
+            PurePath(file).match("*.pnp.cjs") for file in self.source_dir.path.iterdir()
+        )
+        node_modules_exists = self.source_dir.join_within_root("node_modules").path.exists()
+        is_pnp = False
+        if install_config_pnp_on or pnp_cjs_exists or node_modules_exists:
```
a-ovchinnikov, line 168:
```suggestion
return install_config_pnp_on or pnp_cjs_exists or node_modules_exists
```

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -91,11 +90,56 @@ def _generate_build_environment_variables() -> list[EnvironmentVariable]:
...
+    # _fetch_dependencies(project.source_dir)
+
+    # return create_components(packages, project, output_dir)
+
+
 def _verify_corepack_yarn_version(source_dir: RootedPath, env: dict[str, str]) -> None:
     """Verify that corepack installed the correct version of yarn by checking `yarn --version`."""
     installed_yarn_version = extract_yarn_version_from_env(source_dir, env)
 
-    if installed_yarn_version not in VersionsRange("1.22.0", "2.0.0"):
```
a-ovchinnikov, line 98: This looks like a conflict resolution gone wrong, [a00a8320](https://github.com/containerbuildsystem/cachi2/commit/a00a8320625a093700522f9f951a5d93801de39b) introduced a reverse change (L8 in the same module too), could you please check?

ben-alkov: Rats, I must've resolved in the wrong direction...

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Derp, this is what comes from "losing track of the meta" while refactoring.

ben-alkov: Fixed

cachi2/core/package_managers/yarn_classic/main.py
ben-alkov: Fixed.

ben-alkov:
Not ready to merge, as I need to increase coverage, but I'm confident in the implementation.

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -91,6 +92,46 @@ def _generate_build_environment_variables() -> list[EnvironmentVariable]:
...
+    log.info(f"Fetching the yarn-classic dependencies at the subpath {project.source_dir}")
+
+    _verify_repository(project)
+
+    # Placeholders for implementations in other PRs
+    # _set_yarnrc_configuration(project, output_dir)
+    # packages = resolve_packages(project.source_dir)
+    # _fetch_dependencies(project.source_dir)
+
+    # return create_components(packages, project, output_dir)
```
slimreaper35, line 132: This function is not used anywhere

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,173 @@
...
+
+    This class abstracts the underlying attributes.
+    """
+
+    yarn_lockfile: lockfile.Lockfile
+
+    @property
+    def config_kind(self) -> ConfigKind:
+        """Return kind of this ConfigFile."""
+        return "yarnlock"
```
slimreaper35, line 110: IMO, this property should be dropped and replaced with `isinstance(<class-name>)` somewhere else in the code.

requirements.txt
```diff
@@ -489,6 +489,114 @@ packaging==24.1 \
...
+    --hash=sha256:f571aea50ba5623c308aa146eb650eebf7dbe0fd8c5d946e28343cb3b5aad577 \
+    --hash=sha256:f60f0ac7005b9f5a6091009b09a419ace1610e163fa5deaba5ce3484341840e7 \
+    --hash=sha256:f6475a1b2ecb310c98c28d271a30df74f9dd436ee46d09236a6b750a7599ce57 \
+    --hash=sha256:f6d5749fdd33d90e34c2efb174c7e236829147a2713334d708746e94c4bde40d \
+    --hash=sha256:f902804113e032e2cdf8c71015651c97af6418363bea8d78dc0911d56c335032 \
+    --hash=sha256:fa1076244f54bb76e65e22cb6910365779d5c3d71d1f18b275f1dfc7b0d71b4d \
+    --hash=sha256:fc2db02409338bf36590aa985a461b2c96fce91f8e7e0f14c50c5fcc4f229016 \
+    --hash=sha256:ffcad6c564fe6b9b8916c1aefbb37a362deebf9394bd2974e9d84232e3e08504
+    # via yarl
+pyarn==0.2.0 \
```
taylormadore, line 596: pyarn is pinned to different versions in requirements.txt and requirements-extras.txt

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,173 @@
...
+                solution=(
+                    "The package.json file must contain valid JSON. "
+                    "Refer to the parser error and fix the contents of the file."
+                ),
+            )
+
+        return cls(path, package_json_data)
+
+
+class YarnLock(_CommonConfigFile):
```
slimreaper35, line 99: I don't understand the inheritance here, `data` and `path` attributes are not used anywhere inside `YarnLock` class.

I would say package.json and yarn.lock are quite different from each other.

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,173 @@
...
+
+        This is determined by
+        - `installConfig.pnp: true` in 'package.json'
+        - the existence of file(s) with glob name '*.pnp.cjs'
+        - the presence of an expanded node_modules directory
+        For more details on PnP, see: https://classic.yarnpkg.com/en/docs/pnp
+        """
+        install_config_pnp_on = self.package_json.install_config.get("pnp", False) is True
+        pnp_cjs_exists = any(
+            PurePath(file).match("*.pnp.cjs") for file in self.source_dir.path.iterdir()
```
slimreaper35, line 164: What is `PurePath` ?  The regular `Path` class is a subclass of `PurePath` btw. (I don't see `PurePath` class anywhere in cachi2.)

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,173 @@
...
+    def is_pnp_install(self) -> bool:
+        """Is the project is using Plug'n'Play (PnP) workflow or not.
+
+        This is determined by
+        - `installConfig.pnp: true` in 'package.json'
+        - the existence of file(s) with glob name '*.pnp.cjs'
+        - the presence of an expanded node_modules directory
+        For more details on PnP, see: https://classic.yarnpkg.com/en/docs/pnp
+        """
+        install_config_pnp_on = self.package_json.install_config.get("pnp", False) is True
```
slimreaper35, line 162: `is True` is redundant here

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,173 @@
...
+                solution=(
+                    "The yarn.lock file must be valid. "
+                    "Refer to the parser error and fix the contents of the file."
+                ),
+            )
+
+        if not yarn_lockfile:
+            raise PackageRejected(
+                reason="The yarn.lock file must not be empty",
+                solution=("Please verify the content of the file."),
```
slimreaper35, line 137: `(` and `)` are not necessary here

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,173 @@
...
+            raise PackageRejected(
+                reason="The yarn.lock file must be present for the yarn package manager",
+                solution=(
+                    "Please double-check that you have specified the correct path "
+                    "to the package directory containing this file"
+                ),
+            )
+        except ValueError as e:
+            raise PackageRejected(
+                reason=(f"Can't parse the {path.subpath_from_root} file.\n" f"{e}"),
```
slimreaper35, line 127: I think giving the user a full path to the file is more valuable.

taylormadore, line 127: The output here could potentially be lengthy, so I think we definitely want the exceptions chained but without repeating the output in the new PackageRejected exception we are raising

slimreaper35:
As I often get these recommendations, you could split the changes into more commits. For example
- implement yarn.lock
- implement package.json
(package.json has nothing to do with `pyarn`  being added to requirements.txt)

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: THey will be eventually. Also, they are siblings of the yarnc class which will be implemented later

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: The use case here is being able to use the pattern `what_are_you = foo.config_kind` rather than being forced to use

```
if isinstance(foo):
    what_are_you = i_am_foo
elif isinstance(bar):
    what_are_you = i_am_bar
elif isinstance(baz):
    what_are_you = i_am_baz
```

every time.

If we're truly switching a different code block for *each* and *all* classes, then yeah, the if-tree is necessary.

slimreaper35: What would be the real use case here ? `__CommonConfigFile` is the abstract class so you can't instantiate it anyway.
`kind = PackageJson.config_kind == "package_json"` this will awkward to me.

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -91,6 +92,46 @@ def _generate_build_environment_variables() -> list[EnvironmentVariable]:
...
+                " and any 'node_modules' directories."
+            ),
+        )
+
+
+def _verify_repository(project: Project) -> None:
+    _check_pnp_installs(project)
+    # _check_lockfile(project)
+
+    return None
```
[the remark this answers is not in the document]
ben-alkov: Copypasta :harold:

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -91,6 +92,46 @@ def _generate_build_environment_variables() -> list[EnvironmentVariable]:
     return [EnvironmentVariable(name=key, value=value) for key, value in env_vars.items()]
 
 
+def _check_pnp_installs(project: Project) -> None:
```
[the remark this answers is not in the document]
ben-alkov: Copypasta :harold:

cachi2/core/package_managers/yarn_classic/project.py
slimreaper35: EDIT: So my point about `isinstance` is not valid, `ConfigKind` is just Literal.

ben-alkov: Exactly :)

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Refactor fallout :rolleyes:

cachi2/core/package_managers/yarn_classic/main.py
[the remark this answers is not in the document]
ben-alkov: > What does an explicit return add here?

`None` 🤣

Removed

cachi2/core/package_managers/yarn_classic/project.py
slimreaper35: But still don't understand the use case of such property.

slimreaper35: EDIT 2: I understand, but using the class as type insted of literal is much clearer IMO.

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: `match` is only offered on PurePath (https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.match)

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+
+        return cls(path, yarn_lockfile.data)
+
+
+ConfigFile = Union[PackageJson, YarnLock]
+
+
+class Project(NamedTuple):
+    """A directory containing yarn sources."""
```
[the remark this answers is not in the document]
ben-alkov: Copypasta from New Yarn

cachi2/core/package_managers/yarn_classic/project.py
slimreaper35: > The regular Path class is a subclass of PurePath

ben-alkov: :headdesk:

ben-alkov: :headdesk:

cachi2/core/package_managers/yarn_classic/main.py
ben-alkov: Removed

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Fixed

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Fixed

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Fixed

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Fixed

ben-alkov:
> As I often get these recommendations, you could split the changes into more commits. For example
> 
>     * implement yarn.lock
> 
>     * implement package.json
>       (package.json has nothing to do with `pyarn`  being added to requirements.txt)

Fixed

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -91,6 +92,46 @@ def _generate_build_environment_variables() -> list[EnvironmentVariable]:
...
+
+
+def _verify_repository(project: Project) -> None:
+    _check_pnp_installs(project)
+    # _check_lockfile(project)
+
+    return None
+
+
+def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Component]:  # type: ignore
```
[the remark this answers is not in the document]
ben-alkov: Fixed

tests/unit/package_managers/yarn_classic/test_project.py
```diff
@@ -0,0 +1,217 @@
...
+            False,
+            PackageJson,
+            "package.json",
+            VALID_PACKAGE_JSON_FILE,
+            "None",
+            id="non-pnp",
+        ),
+    ],
+)
+def test_pnp_installs_detection(
```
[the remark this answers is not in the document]
ben-alkov: Fixed

requirements.txt
ben-alkov: Fixed

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Fixed

cachi2/core/package_managers/yarn_classic/main.py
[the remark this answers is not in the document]
ben-alkov: Fixed

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -91,6 +92,46 @@ def _generate_build_environment_variables() -> list[EnvironmentVariable]:
...
+    # Temporary type ignore
+    """Process a request for a single yarn source directory.
+
+    :param project: the directory to be processed.
```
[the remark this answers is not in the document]
ben-alkov: Fixed

cachi2/core/package_managers/yarn_classic/project.py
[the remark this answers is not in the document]
ben-alkov: Fixed

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,174 @@
...
+        This is determined by
+        - `installConfig.pnp: true` in 'package.json'
+        - the existence of file(s) with glob name '*.pnp.cjs'
+        - the presence of an expanded node_modules directory
+        For more details on PnP, see: https://classic.yarnpkg.com/en/docs/pnp
+        """
+        install_config_pnp_enabled = self.package_json.install_config.get("pnp", False)
+        pnp_cjs_exists = any(
+            Path(file).match("*.pnp.cjs") for file in self.source_dir.path.iterdir()
+        )
```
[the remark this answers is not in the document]
eskultety:
```
any(self.source_dir.path.glob("*.pnp.cjs")
```

should do just fine I think.

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,171 @@
...
+
+    @property
+    def path(self) -> RootedPath:
+        return self._path
+
+    @property
+    @abstractmethod
+    def config_kind(self) -> ConfigKind:
+        """Return kind of ConfigFile instance."""
+        pass
```
a-ovchinnikov, line 47: Non-blocking: `pass` is redundant here (since the method already has one statement).

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,171 @@
...
+                    "Please double-check that you have specified the correct path "
+                    "to the package directory containing this file"
+                ),
+            )
+        except ValueError:
+            raise PackageRejected(
+                reason=(f"Can't parse the {path} file.\n"),
+                solution=(
+                    "The yarn.lock file must be valid. "
+                    "Refer to the parser error and fix the contents of the file."
```
a-ovchinnikov, line 129: I am sorry, but where is the parser error surfaced?

cachi2/core/package_managers/yarn_classic/project.py
```diff
@@ -0,0 +1,171 @@
+"""
+Parse the relevant files of a yarn project.
+
+It also provides basic utility functions. The main logic to resolve and prefetch
```
a-ovchinnikov, line 4: Does it still provide these functions?

a-ovchinnikov (commented):
99% LGTM (with a couple small and trivial fixes).

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Copypasta.

🤔 I don't exactly know *what* pyarn will say if it sees an invalid lock file. I can do a follow-up to improve the messaging here (I'll delete the inapplicable bit in this PR).

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Yes, e.g. creating ConfigFile instances from existing files, creating Project instances from source dir

cachi2/core/package_managers/yarn_classic/project.py
ben-alkov: Deleted

a-ovchinnikov (approved):

cachi2/core/package_managers/yarn_classic/project.py
a-ovchinnikov: Fair enough.

slimreaper35 (approved):

PR #710, opened 2024-10-25
generic fetcher: Generate SBOM components

kosciCZ:
Generic fetcher now generates SBOM components of purl type pkg:generic. It also produces externalReferences of type distribution for each generic component to allow for gating at later time.

### DIFF
--- a/cachi2/core/models/sbom.py
+++ b/cachi2/core/models/sbom.py
@@ -21,6 +21,13 @@ class Property(pydantic.BaseModel):
     value: str
 
 
+class ExternalReference(pydantic.BaseModel):
+    """An ExternalReference inside an SBOM component."""
+
+    url: str
+    type: Literal["distribution"] = "distribution"
+
+
 FOUND_BY_CACHI2_PROPERTY: Property = Property(name="cachi2:found_by", value="cachi2")
 
 
@@ -35,7 +42,10 @@ class Component(pydantic.BaseModel):
     purl: str
     version: Optional[str] = None
     properties: list[Property] = pydantic.Field(default_factory=list, validate_default=True)
-    type: Literal["library"] = "library"
+    type: Literal["library", "file"] = "library"
+    external_references: Optional[list[ExternalReference]] = pydantic.Field(
+        serialization_alias="externalReferences", default=None
+    )
 
     def key(self) -> str:
         """Uniquely identifies a package.
--- a/cachi2/core/package_managers/generic/main.py
+++ b/cachi2/core/package_managers/generic/main.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Union
 
 import yaml
+from packageurl import PackageURL
 from pydantic import ValidationError
 
 from cachi2.core.checksum import must_match_any_checksum
@@ -12,7 +13,7 @@ from cachi2.core.config import get_config
 from cachi2.core.errors import PackageRejected
 from cachi2.core.models.input import Request
 from cachi2.core.models.output import RequestOutput
-from cachi2.core.models.sbom import Component
+from cachi2.core.models.sbom import Component, ExternalReference
 from cachi2.core.package_managers.general import async_download_files
 from cachi2.core.package_managers.generic.models import GenericLockfileV1
 from cachi2.core.rooted_path import RootedPath
@@ -69,7 +70,7 @@ def _resolve_generic_lockfile(source_dir: RootedPath, output_dir: RootedPath) ->
     # verify checksums
     for artifact in lockfile.artifacts:
         must_match_any_checksum(artifact.target, artifact.formatted_checksums)
-    return []
+    return _generate_sbom_components(lockfile)
 
 
 def _load_lockfile(lockfile_path: RootedPath, output_dir: RootedPath) -> GenericLockfileV1:
@@ -102,3 +103,29 @@ def _load_lockfile(lockfile_path: RootedPath, output_dir: RootedPath) -> Generic
                 ),
             )
     return lockfile
+
+
+def _generate_sbom_components(lockfile: GenericLockfileV1) -> list[Component]:
+    """Generate a list of SBOM components for a given lockfile."""
+    components: list[Component] = []
+
+    for artifact in lockfile.artifacts:
+        name = Path(artifact.target).name
+        url = str(artifact.download_url)
+        checksums = ",".join([f"{algo}:{digest}" for algo, digest in artifact.checksums.items()])
+        component = Component(
+            name=name,
+            purl=PackageURL(
+                type="generic",
+                name=name,
+                qualifiers={
+                    "download_url": url,
+                    "checksums": checksums,
+                },
+            ).to_string(),
+            type="file",
+            external_references=[ExternalReference(url=url, type="distribution")],
+        )
+        components.append(component)
+
+    return components
--- a/tests/unit/package_managers/test_generic.py
+++ b/tests/unit/package_managers/test_generic.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Type
+from typing import Any, Type
 from unittest import mock
 
 import pytest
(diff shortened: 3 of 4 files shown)


cachi2/core/models/sbom.py
slimreaper35, line 1: The changes in this file would deserve a separate commit.

tests/unit/package_managers/test_generic.py
```diff
@@ -207,6 +207,61 @@ def test_resolve_generic_lockfile_invalid(
     assert expected_err in str(exc_info.value)
 
 
+@pytest.mark.parametrize(
+    ["lockfile", "expected"],
```
slimreaper35, line 211:
```suggestion
    ["lockfile_content", "expected_components"]
```
nitpick: I would choose slightly more verbose names.

cachi2/core/package_managers/generic/main.py
```diff
@@ -102,3 +103,29 @@ def _load_lockfile(lockfile_path: RootedPath, output_dir: RootedPath) -> Generic
...
+            name=name,
+            purl=PackageURL(
+                type="generic",
+                name=name,
+                qualifiers={
+                    "download_url": url,
+                    "checksums": ",".join([f"{a}:{h}" for a, h in artifact.checksums.items()]),
+                },
+            ).to_string(),
+            package_manager="generic",
```
slimreaper35, line 125: Is `package_manager` an attribute in the `Component` class ?

cachi2/core/models/sbom.py
kosciCZ: Split to a separate commit.

tests/unit/package_managers/test_generic.py
kosciCZ: Changed.

cachi2/core/package_managers/generic/main.py
kosciCZ: Good catch, that's a leftover from something I was testing. Removed.

cachi2/core/package_managers/generic/main.py
```diff
@@ -102,3 +103,28 @@ def _load_lockfile(lockfile_path: RootedPath, output_dir: RootedPath) -> Generic
...
+        name = Path(artifact.target).name
+        url = str(artifact.download_url)
+        component = Component(
+            name=name,
+            purl=PackageURL(
+                type="generic",
+                name=name,
+                qualifiers={
+                    "download_url": url,
+                    "checksums": ",".join([f"{a}:{h}" for a, h in artifact.checksums.items()]),
```
slimreaper35, line 122: Maybe one more suggestion, `a` and `h` do not tell much about what it stands for.  Btw. in the `LockfileArtifact` you have used: `ChecksumInfo(algo, digest) for algo, digest in self.checksums.items()`. For readability, you can define checksums just below the `url` variable.

But, I will approve the PR.

a-ovchinnikov (approved):
LGTM

cachi2/core/package_managers/generic/main.py
kosciCZ: I agree, it dows improve readability. Fixed.

kosciCZ:
This has acquired the necessary number of approvals. Can this be merged? It'd be great if this made it to this week's release.

cachi2/core/models/sbom.py
```diff
@@ -21,6 +21,13 @@ class Property(pydantic.BaseModel):
     value: str
 
 
+class ExternalReference(pydantic.BaseModel):
+    """An ExternalReference inside an SBOM component."""
+
+    url: str
+    type: Literal["distribution"] = "distribution"
+
+
```
eskultety, line 30: Can you justify this change? The commit message is empty with such an IMO substantial SBOM update. Why is this needed? Why have these specific fields been chosen for the external reference? Can you elaborate?

eskultety:
Please don't merge until https://github.com/containerbuildsystem/cachi2/pull/710#discussion_r1820475942 is addressed.

cachi2/core/models/sbom.py
kosciCZ: It is sort of described in the follow-up commit message, as I'd say the introduction itself bears no meaning, if not used within the context of the generic fetcher (other package managers do not utilize it).

> Generic fetcher now generates SBOM components of purl type
pkg:generic. It also produces externalReferences of type
distribution for each generic component to allow for gating
at a later time.

I chose `externalReferences` of type `distribution` to describe the `download_url` in a way that's easily processable in whatever will look at the SBOM later (thinking about EC, but could be anything, really). The [CycloneDX spec](https://github.com/CycloneDX/guides/blob/main/SBOM/en/0x51-External-References.md#external-references) describes this type of externalReferences as "Direct or repository download location" which in my opinion makes sense for the `download_url`

cachi2/core/package_managers/generic/main.py
```diff
@@ -102,3 +103,29 @@ def _load_lockfile(lockfile_path: RootedPath, output_dir: RootedPath) -> Generic
...
+
+
+def _generate_sbom_components(lockfile: GenericLockfileV1) -> list[Component]:
+    """Generate a list of SBOM components for a given lockfile."""
+    components: list[Component] = []
+
+    for artifact in lockfile.artifacts:
+        name = Path(artifact.target).name
+        url = str(artifact.download_url)
+        checksums = ",".join([f"{algo}:{digest}" for algo, digest in artifact.checksums.items()])
```
brunoapimentel, line 115: We're reporting all checksums present in the lockfile here, but this does not mean that all of them were validated. I'm wondering if this is the behavior we want. It'd be nice to double check how this is being done across different package managers, but I believe we can investigate this as a follow up, no need to block this PR.

cachi2/core/models/sbom.py
brunoapimentel: I think what you wrote here can be worked into a message for this specific commit. For now, it'll only be used for `generic`, but that might change in the future. The reasoning here is that this field should carry the URL from which a component was downloaded, as a means of having more meaningful data in the SBOM so that other tooling can build on top of it.

cachi2/core/package_managers/generic/main.py
kosciCZ: Huh, I guess that's technically right. I guess we would need to implement something like `must_match_all_checksums`? Good point though.

cachi2/core/models/sbom.py
kosciCZ: I will amend the commit message for that commit.

kosciCZ: Amended.

cachi2/core/package_managers/generic/main.py
brunoapimentel: From a quick look, it seems that the currently existing package managers are only reporting a single checksum in the SBOM. So I think it's worth to have a discussion about this before we end the experimental phase on this.

brunoapimentel (approved):

cachi2/core/models/sbom.py
eskultety: @kosciCZ  please include https://cyclonedx.org/docs/1.6/json/#components_items_externalReferences as a reference in the commit message where the following is IMO the most important bit justifying the selection of this particular component item:

> External references provide a way to document systems, sites, and information that may be relevant but are not included with the BOM. They may also establish specific relationships within or external to the BOM.

cachi2/core/package_managers/generic/main.py
eskultety: > From a quick look, it seems that the currently existing package managers are only reporting a single checksum in the SBOM. So I think it's worth to have a discussion about this (should we report checksums that were not validated) before we end the experimental phase on generic.

@brunoapimentel can you please open an issue on this so that that discussion is not forgotten?

eskultety (approved):
@kosciCZ with the CycloneDX reference added to commit nr.1 -> ACK

cachi2/core/models/sbom.py
kosciCZ: Added to the commit message.

PR #717, opened 2024-11-01
Bundler package manager: fixes / improvements

slimreaper35:


tests/integration/test_data/generic_e2e/fetch_deps_sha256sums.json
```diff
@@ -1,4 +1,4 @@
- {
-   "generic/archive.zip": "sha256:386428a82f37345fa24b74068e0e79f4c1f2ff38d4f5c106ea14de4a2926e584",
-   "generic/v1.0.0.zip": "sha256:4fbcaa2a8d17c1f8042578627c122361ab18b7973311e7e9c598696732902f87"
- }
+{
+  "generic/archive.zip": "sha256:386428a82f37345fa24b74068e0e79f4c1f2ff38d4f5c106ea14de4a2926e584",
+  "generic/v1.0.0.zip": "sha256:4fbcaa2a8d17c1f8042578627c122361ab18b7973311e7e9c598696732902f87"
+}
```
slimreaper35, line 4: It shouldn't be in this PR, but someone accidentally messed up the formatting.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -90,9 +91,17 @@ class GemPlatformSpecificDependency(GemDependency):
 
     platform: str
 
+    @property
+    def purl(self) -> str:
+        """Get PURL for this dependency."""
+        qualifiers = {"platform": self.platform}
+        purl = PackageURL(type="gem", name=self.name, version=self.version, qualifiers=qualifiers)
+        return purl.to_string()
```
slimreaper35, line 99: Do you have any other possible solution in your mind ?

eskultety: `platform` isn't a standard PURL qualifier for PURL, is it? We should avoid inventing our own qualifiers and go around the PURL spec. I'd start by opening an issue with the PURL spec. I would then look at the CycloneDX spec and see if there's some extra info data field we could stuff this information to. As for dedup in that case we'll have to skip components that would define such a cyclonedx extra field from the process.

I thought Bundler doesn't suffer from the same problem as pip does where we don't really know what "wheel" we should download and so we download all of them; with Bundler, the platform is encoded in the lock file, isn't it? So when does this bug occur?

tests/integration/test_data/generic_e2e/fetch_deps_sha256sums.json
eskultety: So, can you propose it in a separate PR then?

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -103,10 +105,6 @@ def download_to(self, deps_dir: RootedPath) -> None:
         log.info(
             "Downloading platform-specific gem %s-%s-%s", self.name, self.version, self.platform
         )
-        # A combination of Ruby v.3.0.7 and some Bundler dependencies results in
-        # -gnu suffix being dropped from some platforms. This was observed on
-        # sqlite3-aarch-linux-gnu. We discourage using outdated platforms
-        # for building dependencies and cnsider this to be a limitation of Ruby.
```
eskultety, line 109: If you want to remove this hunk, you need to do it in a separate commit and describe the motivation accordingly.

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -209,7 +209,7 @@ def test_parse_gemlock_empty(
 @pytest.mark.parametrize(
     "source",
     [
-        "https://rubygems.org",
+        "https://rubygems.org/",
```
eskultety, line 212: This will be unnecessary with `urljoin`, so you can keep it as is.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -71,7 +71,8 @@ def purl(self) -> str:
     @cached_property
     def remote_location(self) -> str:
         """Return remote location to download this gem from."""
-        return f"{self.source}/gems/{self.name}-{self.version}.gem"
```
eskultety, line 74: **Commit message:**

I'm missing any context in the commit message, so I had to look at the code to see what's going on based on some suspicious changes in this commit (adding commentaries on trailing slash etc.) so please provide some description, it helps reviewers who don't hold the necessary context in their mind anymore to get up to speed more quickly.

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -71,7 +71,8 @@ def purl(self) -> str:
...
+        # The source URL ends with a slash
+        return f"{self.source}downloads/{self.name}-{self.version}.gem"
```
eskultety, line 75: The fact that we'd need a commentary to explain kinda disqualifies the change as is IMO. Unfortunately the precedent we have for platform gems that led you to this change slipped my eye when it was added. The way this should be handled is with `urljoin` from `urllib.parse` which shields us from potentially constructing an invalid URL, e.g. by adding an extra slash (as I TIL would be invalid).

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -93,6 +94,7 @@ class GemPlatformSpecificDependency(GemDependency):
     @property
     def remote_location(self) -> str:
         """Return remote location to download this gem from."""
+        # The source URL ends with a slash
```
eskultety, line 104: Please fix the URL construction rather than adding a commentary.

cachi2/core/package_managers/bundler/parser.py
eskultety: I'm imagining something along the lines:
_Access all gems using direct download URL instead of going through the hosted package index._

cachi2/core/package_managers/bundler/parser.py
slimreaper35: I thought it was a leftover from the last bundler PR.

tests/integration/test_data/generic_e2e/fetch_deps_sha256sums.json
slimreaper35: Is it really necessary? It is just 4 spaces. I will add that as the first commit in this PR.

> It shouldn't be in this PR, but someone accidentally messed up the formatting.

I mean by that it was not intended at first :smile:

cachi2/core/package_managers/bundler/parser.py
a-ovchinnikov: Please don't. This is not a leftover, this is intentional. Figuring out that certain combinations cause interesting failures was rather time-consuming and I had to document that somewhere. This spot is close enough to a place in code that could suffer from the issue, so I chose it.

cachi2/core/package_managers/bundler/parser.py
slimreaper35: > What exactly is the bug? Is it that platform-specific gems are not reported individually?
Yes, at least I thought that, but since it is the same for `pip` I don't know. Maybe not.

slimreaper35: I built a sample ruby application in konflux this week and here is a component from the final CycloneDX SBOM:

```
        {
            "bom-ref": "pkg:gem/nokogiri@1.16.6-arm64-darwin?package-id=9388ea05e3cbfe2f",
            "type": "library",
            "name": "nokogiri",
            "version": "1.16.6-arm64-darwin",
            "cpe": "cpe:2.3:a:nokogiri:nokogiri:1.16.6-arm64-darwin:*:*:*:*:*:*:*",
            "purl": "pkg:gem/nokogiri@1.16.6-arm64-darwin",
            "properties": [
                {
                    "name": "syft:package:foundBy",
                    "value": "ruby-gemfile-cataloger"
                },
                {
                    "name": "syft:package:language",
                    "value": "ruby"
                },
                {
                    "name": "syft:package:type",
                    "value": "gem"
                },
                {
                    "name": "syft:cpe23",
                    "value": "cpe:2.3:a:nokogiri:nokogiri:1.16.6-arm64-darwin:*:*:*:*:ruby:*:*"
                },
                {
                    "name": "syft:location:0:path",
                    "value": "/Gemfile.lock"
                }
            ]
        },
```

a-ovchinnikov: > "version": "1.16.6-arm64-darwin"

Is this the only entry for nokogiri? Apart from being an interesting choice of platform it also tells us that version gets mangled with architecture identifier. Maybe we should be doing the same?

slimreaper35: There are 3 entries for nokogiri. In the `cachi2-output/deps` I see those 3 + the source gem, which is reported by cachi2 somewhere else in the SBOM. All three have this custom property:

```
  {
      "name": "syft:location:0:path",
      "value": "/Gemfile.lock"
  }
```

tests/integration/test_data/generic_e2e/fetch_deps_sha256sums.json
eskultety: 1st in the commit list is okay with me.

cachi2/core/package_managers/bundler/parser.py
eskultety: > I didn't realize that we are doing the same way for pip.

If this was a reaction to https://github.com/containerbuildsystem/cachi2/pull/717/commits/ff4f85a740a8b08d13642a9c5ef5b0538df69d5c#r1828958364 then I didn't hint at this anywhere. To be fair I don't know what we do with pip, but it is definitely worth checking to remain consistent and potentially file an issue to improve the status quo if the same happens with the pip ecosystem because I do agree that filtering out downloaded artifacts because they happen to have the same PURL is wrong and undermines our mission of accurate SBOMs. So again, we should check what happens with pip wheels, @slimreaper35 please have a look at this using, we may already have the data at hand with the integration test suite SBOMs.

slimreaper35: With pip, the PURLs are the same for each requirement, so the number of components differs from the downloaded artifacts. I fetched-deps in our https://github.com/cachito-testing/cachi2-pip-wheels/tree/no-sdists integrations test to make sure.

```
diff --git a/cachi2/core/package_managers/pip.py b/cachi2/core/package_managers/pip.py
index 08f55027..e41f135c 100644
--- a/cachi2/core/package_managers/pip.py
+++ b/cachi2/core/package_managers/pip.py
@@ -209,6 +209,7 @@ def fetch_pip_source(request: Request) -> RequestOutput:
         replaced_requirements_files = map(_replace_external_requirements, info["requirements"])
         project_files.extend(filter(None, replaced_requirements_files))

+    print(len(components))
     return RequestOutput.from_obj_list(
         components=components,
         environment_variables=environment_variables,
diff --git a/cachi2/interface/cli.py b/cachi2/interface/cli.py
index d98771df..3198617e 100644
--- a/cachi2/interface/cli.py
+++ b/cachi2/interface/cli.py
@@ -286,6 +286,7 @@ def fetch_deps(
     )

     sbom = request_output.generate_sbom()
+    print(len(sbom.components))
     request.output_dir.join_within_root("bom.json").path.write_text(
         # the Sbom model has camelCase aliases in some fields
         sbom.model_dump_json(indent=2, by_alias=True, exclude_none=True)
```

Result:

```
168
12
2024-11-06 13:02:12,650 INFO All dependencies fetched successfully \o/
```

eskultety: So it appears we have the same situation with `pip` and now we need to come to an agreement on whether this is the wrong behaviour and fix it or be consistent and leave it be. My take on this is that we report dependencies in the SBOM - does it matter how many wheels/gems we downloaded? That's a build stage detail which of the artifacts gets used, but essentially coming from the same resource page (though a different path in the URL) it should not matter whether a platform package is used over a general Ruby platform one, it's still the same dependency IMO. While I'm not disputing the fact that we might do better I'm not sure the benefit outweighs the process of addressing this.

slimreaper35: I suspect we can leave it for now and if the decision is to report packages for each platform separately, we can do it for `pip` and `bundler` at once.

eskultety: I would track this though in a low priority issue so that we don't forget about this. We can evaluate at any time. Will you open an issue @slimreaper35?
Marking this thread as resolved so that review can continue.

a-ovchinnikov (approved):

PR #726, opened 2024-11-02
Update pip usage example with CS9

jcapiitao:
CS8 is EOL since May 31 2024 [1], this patch updates the example of pip usage to use CS9.
Plus, we need to enable the CRB repository which provides gobject-introspection-devel.

[1] https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -377,13 +377,15 @@ already contains everything you need.
 Containerfile.baseimage:
 
 ```Dockerfile
-FROM quay.io/centos/centos:stream8
+FROM quay.io/centos/centos:stream9
 
-# python3.8 runtime, C build dependencies
-RUN dnf -y install \
-        python38 \
-        python38-pip \
-        python38-devel \
+# python3.9 runtime, C build dependencies
+RUN dnf -y install 'dnf-command(config-manager)' && \
+    dnf config-manager --set-enabled crb && \
+    dnf -y install \
+        python3 \
+        python3-pip \
+        python3-devel \
         gcc \
         make \
         libffi-devel \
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -377,13 +377,15 @@ already contains everything you need.
...
-# python3.8 runtime, C build dependencies
-RUN dnf -y install \
-        python38 \
-        python38-pip \
-        python38-devel \
+FROM quay.io/centos/centos:stream9
+
+# python3.9 runtime, C build dependencies
+RUN dnf -y install 'dnf-command(config-manager)' && \
+    dnf config-manager --set-enabled crb && \
```
eskultety, line 384: Disclaimer: I didn't know we had an example like this until you pointed it out.

CS is a rolling distro which breaks package updates more often than not sadly (at least it used to be the case). I mean what you have here is all fine but strictly from docs POV I just wonder if it deserves the promotion (even just as an example) since it requires the Content Ready Builder repo to enable headers. Anyway, I think this hunk would be slightly simplified if you replaced CS with an example using AlmaLinux or Rocky where you don't need either the config manager or the CRB repo.

brunoapimentel: Atomic reactor [uses](https://github.com/containerbuildsystem/atomic-reactor/blob/master/Dockerfile#L1) Fedora, so we might as well just do the same.

eskultety: Works too.

brunoapimentel:
@jcapiitao Do you still intend to work on this, or can we close this PR?

slimreaper35:
Can we use `fedora` or `ubi` instead of `centos` ?

eskultety:
> Can we use `fedora` or `ubi` instead of `centos` ?

UBI would be more stable for a docs type of change since with Fedora the docs (like the Fedora images) would quickly get outdated again.

ben-alkov:
>> Can we use fedora or ubi instead of centos ?

> UBI would be more stable for a docs type of change since with Fedora the docs (like the Fedora images) would quickly get outdated again.

Def not Cent. Bruno previously mentioned Fed, but Erik makes a very good point...

I'm going to test with UBI and see if the example works as intended.

ben-alkov:
After consulting with colleagues, it looks like atomic-reactor (mostly because of osbs client) is a very bad choice for an introductory example (it might not even be possible to build it *easily* with Hermeto without some gnarly hacks). I'm going to close this PR and work up something much simpler.
Thanks very much to @jcapiitao for opening this, and to @eskultety, @slimreaper35, and @brunoapimentel for reviews!

PR #718, opened 2024-11-01
generic fetcher: Add usage docs

kosciCZ:
Add documentation on how to use the generic fetcher and also an ADR to help move out of the experimental phase.

I've tried to follow the CONTRIBUTING.md guidelines, but since I did not see any previous ADR, I just tried to come up with something. Feel free to give more direction on how you want the docs to look like. Thanks!

# Maintainers will complete the following section

- [ ] Commit messages are descriptive enough
- [ ] Code coverage from testing does not decrease and new code is covered
- [ ] Docs updated (if applicable)
- [ ] Docs links in the code are still valid (if docs were updated)

**Note:** if the contribution is external (not from an organization member), the CI
pipeline will not run automatically. After verifying that the CI is safe to run:

- [approve GitHub Actions workflows][approve-gh-actions] by clicking a button
- approve the Red Hat Trusted App Pipeline container build by commenting `/ok-to-test`
  (as is the [standard for Pipelines as Code][pac-running-pipeline])

[approve-gh-actions]: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
[pac-running-pipeline]: https://pipelinesascode.com/docs/guide/running/#running-the-pipelinerun

PR #728, opened 2024-11-06
generic fetcher: Official support with ADR

kosciCZ:
This change includes the generic fetcher ADR which is a necessary step for the feature to be officially supported. Split from #718 

### DIFF
--- a/cachi2/core/package_managers/generic/main.py
+++ b/cachi2/core/package_managers/generic/main.py
@@ -62,14 +62,14 @@ def _resolve_generic_lockfile(source_dir: RootedPath, output_dir: RootedPath) ->
 
     for artifact in lockfile.artifacts:
         # create the parent directory for the artifact
-        Path.mkdir(Path(artifact.target).parent, parents=True, exist_ok=True)
-        to_download[str(artifact.download_url)] = artifact.target
+        Path.mkdir(Path(artifact.filename).parent, parents=True, exist_ok=True)
+        to_download[str(artifact.download_url)] = artifact.filename
 
     asyncio.run(async_download_files(to_download, get_config().concurrency_limit))
 
     # verify checksums
     for artifact in lockfile.artifacts:
-        must_match_any_checksum(artifact.target, artifact.formatted_checksums)
+        must_match_any_checksum(artifact.filename, artifact.formatted_checksums)
     return _generate_sbom_components(lockfile)
 
 
@@ -110,7 +110,7 @@ def _generate_sbom_components(lockfile: GenericLockfileV1) -> list[Component]:
     components: list[Component] = []
 
     for artifact in lockfile.artifacts:
-        name = Path(artifact.target).name
+        name = Path(artifact.filename).name
         url = str(artifact.download_url)
         checksums = ",".join([f"{algo}:{digest}" for algo, digest in artifact.checksums.items()])
         component = Component(
--- a/cachi2/core/package_managers/generic/main.py
+++ b/cachi2/core/package_managers/generic/main.py
@@ -19,7 +19,7 @@ from cachi2.core.package_managers.generic.models import GenericLockfileV1
 from cachi2.core.rooted_path import RootedPath
 
 log = logging.getLogger(__name__)
-DEFAULT_LOCKFILE_NAME = "generic_lockfile.yaml"
+DEFAULT_LOCKFILE_NAME = "artifacts.lock.yaml"
 DEFAULT_DEPS_DIR = "deps/generic"
 
 
--- a/cachi2/core/package_managers/generic/main.py
+++ b/cachi2/core/package_managers/generic/main.py
@@ -69,7 +69,7 @@ def _resolve_generic_lockfile(source_dir: RootedPath, output_dir: RootedPath) ->
 
     # verify checksums
     for artifact in lockfile.artifacts:
-        must_match_any_checksum(artifact.filename, artifact.formatted_checksums)
+        must_match_any_checksum(artifact.filename, [artifact.formatted_checksum])
     return _generate_sbom_components(lockfile)
 
 
@@ -112,7 +112,6 @@ def _generate_sbom_components(lockfile: GenericLockfileV1) -> list[Component]:
     for artifact in lockfile.artifacts:
         name = Path(artifact.filename).name
         url = str(artifact.download_url)
-        checksums = ",".join([f"{algo}:{digest}" for algo, digest in artifact.checksums.items()])
         component = Component(
             name=name,
             purl=PackageURL(
@@ -120,7 +119,7 @@ def _generate_sbom_components(lockfile: GenericLockfileV1) -> list[Component]:
                 name=name,
                 qualifiers={
                     "download_url": url,
-                    "checksums": checksums,
+                    "checksum": artifact.checksum,
                 },
             ).to_string(),
             type="file",
--- a/cachi2/core/package_managers/generic/main.py
+++ b/cachi2/core/package_managers/generic/main.py
@@ -32,24 +32,29 @@ def fetch_generic_source(request: Request) -> RequestOutput:
     components = []
     for package in request.generic_packages:
         path = request.source_dir.join_within_root(package.path)
-        components.extend(_resolve_generic_lockfile(path, request.output_dir))
+        lockfile = package.lockfile or path.join_within_root(DEFAULT_LOCKFILE_NAME).path
+        if not lockfile.is_absolute():
+            raise PackageRejected(
+                f"Supplied generic lockfile path '{lockfile}' is not absolute, refusing to continue.",
+                solution="Make sure the supplied path to the generic lockfile is absolute.",
+            )
+        components.extend(_resolve_generic_lockfile(lockfile, request.output_dir))
     return RequestOutput.from_obj_list(components=components)
 
 
-def _resolve_generic_lockfile(source_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+def _resolve_generic_lockfile(lockfile_path: Path, output_dir: RootedPath) -> list[Component]:
     """
     Resolve the generic lockfile and pre-fetch the dependencies.
 
-    :param source_dir: the source directory to resolve the lockfile from
+    :param lockfile_path: absolute path to the lockfile
     :param output_dir: the output directory to store the dependencies
     """
-    lockfile_path = source_dir.join_within_root(DEFAULT_LOCKFILE_NAME)
-    if not lockfile_path.path.exists():
+    if not lockfile_path.exists():
         raise PackageRejected(
-            f"Cachi2 generic lockfile '{DEFAULT_LOCKFILE_NAME}' missing, refusing to continue.",
+            f"Cachi2 generic lockfile '{lockfile_path}' does not exist, refusing to continue.",
             solution=(
-                f"Make sure your repository has cachi2 generic lockfile '{DEFAULT_LOCKFILE_NAME}' checked in "
-                "to the repository."
+                f"Make sure your repository has cachi2 generic lockfile '{DEFAULT_LOCKFILE_NAME}' "
+                f"checked in to the repository, or the supplied lockfile path is correct."
             ),
         )
 
@@ -73,7 +78,7 @@ def _resolve_generic_lockfile(source_dir: RootedPath, output_dir: RootedPath) ->
     return _generate_sbom_components(lockfile)
 
 
-def _load_lockfile(lockfile_path: RootedPath, output_dir: RootedPath) -> GenericLockfileV1:
+def _load_lockfile(lockfile_path: Path, output_dir: RootedPath) -> GenericLockfileV1:
     """
     Load the cachi2 generic lockfile from the given path.
 
--- a/cachi2/core/package_managers/generic/models.py
+++ b/cachi2/core/package_managers/generic/models.py
@@ -1,6 +1,7 @@
+import re
 from collections import Counter
 from pathlib import Path
-from typing import List, Literal
+from typing import Literal
 from urllib.parse import urlparse
 
 from pydantic import AnyUrl, BaseModel, field_validator, model_validator
@@ -9,6 +10,8 @@ from pydantic_core.core_schema import ValidationInfo
 from cachi2.core.checksum import ChecksumInfo
 from cachi2.core.errors import PackageManagerError
 from cachi2.core.rooted_path import RootedPath
+
+CHECKSUM_FORMAT = re.compile(r"^[a-zA-Z0-9]+:[a-zA-Z0-9]+$")
 
 
 class LockfileMetadata(BaseModel):
@@ -23,24 +26,24 @@ class LockfileArtifact(BaseModel):
 
     :param download_url: The URL to download the artifact from.
     :param filename: The target path to save the artifact to. Subpath of the deps/generic folder.
-    :param checksums: Map that represents checksums for the artifact where keys are hashing algs and values are hashes.
+    :param checksum: Checksum of the artifact in the format "algorithm:hash"
     """
 
     download_url: AnyUrl
     filename: str = ""
-    checksums: dict[str, str]
+    checksum: str
 
-    @field_validator("checksums")
+    @field_validator("checksum")
     @classmethod
-    def no_empty_checksums(cls, value: dict[str, str]) -> dict[str, str]:
+    def checksum_format(cls, value: str) -> str:
         """
-        Validate that at least one checksum is present for an artifact.
+        Validate that the provided checksum string is in the format "algorithm:hash".
 
         :param value: the checksums dict to validate
         :return: the validated checksum dict
         """
-        if len(value) == 0:
-            raise ValueError("At least one checksum must be provided.")
+        if not CHECKSUM_FORMAT.match(value):
+            raise ValueError("Checksum must be in the format 'algorithm:hash'")
         return value
 
     @model_validator(mode="after")
@@ -62,9 +65,10 @@ class LockfileArtifact(BaseModel):
         return self
 
     @property
-    def formatted_checksums(self) -> List[ChecksumInfo]:
-        """Return the checksums as a list of ChecksumInfo objects."""
-        return [ChecksumInfo(algo, digest) for algo, digest in self.checksums.items()]
+    def formatted_checksum(self) -> ChecksumInfo:
+        """Return the checksum as a ChecksumInfo object."""
+        algorithm, digest = self.checksum.split(":", 1)
+        return ChecksumInfo(algorithm, digest)
 
 
 class GenericLockfileV1(BaseModel):
--- a/cachi2/core/package_managers/generic/models.py
+++ b/cachi2/core/package_managers/generic/models.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from typing import Literal
 from urllib.parse import urlparse
 
-from pydantic import AnyUrl, BaseModel, field_validator, model_validator
+from pydantic import AnyUrl, BaseModel, ConfigDict, field_validator, model_validator
 from pydantic_core.core_schema import ValidationInfo
 
 from cachi2.core.checksum import ChecksumInfo
@@ -18,6 +18,7 @@ class LockfileMetadata(BaseModel):
     """Defines format of the metadata section in the lockfile."""
 
     version: Literal["1.0"]
+    model_config = ConfigDict(extra="forbid")
 
 
 class LockfileArtifact(BaseModel):
@@ -32,6 +33,7 @@ class LockfileArtifact(BaseModel):
     download_url: AnyUrl
     filename: str = ""
     checksum: str
+    model_config = ConfigDict(extra="forbid")
 
     @field_validator("checksum")
     @classmethod
@@ -76,6 +78,7 @@ class GenericLockfileV1(BaseModel):
 
     metadata: LockfileMetadata
     artifacts: list[LockfileArtifact]
+    model_config = ConfigDict(extra="forbid")
 
     @model_validator(mode="after")
     def no_artifact_conflicts(self) -> "GenericLockfileV1":
--- a/cachi2/core/package_managers/generic/models.py
+++ b/cachi2/core/package_managers/generic/models.py
@@ -22,12 +22,12 @@ class LockfileArtifact(BaseModel):
     Defines format of a single artifact in the lockfile.
 
     :param download_url: The URL to download the artifact from.
-    :param target: The target path to save the artifact to. Subpath of the deps/generic folder.
+    :param filename: The target path to save the artifact to. Subpath of the deps/generic folder.
     :param checksums: Map that represents checksums for the artifact where keys are hashing algs and values are hashes.
     """
 
     download_url: AnyUrl
-    target: str = ""
+    filename: str = ""
     checksums: dict[str, str]
 
     @field_validator("checksums")
@@ -44,11 +44,11 @@ class LockfileArtifact(BaseModel):
         return value
 
     @model_validator(mode="after")
-    def set_target(self, info: ValidationInfo) -> "LockfileArtifact":
+    def set_filename(self, info: ValidationInfo) -> "LockfileArtifact":
         """Set the target path if not provided and resolve it into an absolute path."""
-        if not self.target:
+        if not self.filename:
             url_path = urlparse(str(self.download_url)).path
-            self.target = Path(url_path).name
+            self.filename = Path(url_path).name
 
         # needs to have output_dir context in order to be able to resolve the target path
         # and so that it can be used to check for conflicts with other artifacts
@@ -57,7 +57,7 @@ class LockfileArtifact(BaseModel):
                 "The `LockfileArtifact` class needs to be called with `output_dir` in the context"
             )
         output_dir: RootedPath = info.context["output_dir"]
-        self.target = str(output_dir.join_within_root(self.target).path.resolve())
+        self.filename = str(output_dir.join_within_root(self.filename).path.resolve())
 
         return self
 
@@ -75,15 +75,15 @@ class GenericLockfileV1(BaseModel):
 
     @model_validator(mode="after")
     def no_artifact_conflicts(self) -> "GenericLockfileV1":
-        """Validate that all artifacts have unique targets and download_urls."""
+        """Validate that all artifacts have unique filenames and download_urls."""
         urls = Counter(a.download_url for a in self.artifacts)
-        targets = Counter(a.target for a in self.artifacts)
+        filenames = Counter(a.filename for a in self.artifacts)
         duplicate_urls = [str(u) for u, count in urls.most_common() if count > 1]
-        duplicate_targets = [t for t, count in targets.most_common() if count > 1]
-        if duplicate_urls or duplicate_targets:
+        duplicate_filenames = [t for t, count in filenames.most_common() if count > 1]
+        if duplicate_urls or duplicate_filenames:
             raise ValueError(
                 (f"Duplicate download_urls: {duplicate_urls}\n" if duplicate_urls else "")
-                + (f"Duplicate targets: {duplicate_targets}" if duplicate_targets else "")
+                + (f"Duplicate filenames: {duplicate_filenames}" if duplicate_filenames else "")
             )
 
         return self
--- a/docs/adr/0001-add-generic-fetcher.md
+++ b/docs/adr/0001-add-generic-fetcher.md
@@ -1,0 +1,102 @@
+# Add generic fetcher
+
+## Context
+
+Some users need to download arbitrary files that don't fit within an established package ecosystem cachi2 could
+potentially otherwise support. The target audience is users that want to use cachi2 to achieve hermetic builds
+and also want an easy way to include these arbitrary files, that cachi2 will account for in the SBOM it produces.
+
+## Decision
+
+A new package manager for generic artifacts must be introduced. This package manager utilizes a custom
+lockfile based on which it will download files, save them into a requested location, and verify checksums.
+Below is a more detailed overview of the implementation.
+
+### Lockfile format
+
+Cachi2 expects the lockfile to be named `artifacts.lock.yaml`.
+In order to account for possible future breaking changes, the lockfile will contain a `metadata` section with a `version`
+field that will indicate the version of the lockfile format. It will also contain a list of artifacts (files) to download,
+each of the artifacts to having a URL, a checksum, and optionally output filename specified.
+
+```yaml
+metadata:
+  # uses X.Y semantic versioning
+  version: "1.0"
+artifacts:
+  - download_url: https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
+    filename: granite-model-1.safetensors
+    checksum: sha256:d16bf783cb6670f7f692ad7d6885ab957c63cfc1b9649bc4a3ba1cfbdfd5230c
+```
+
+#### Lockfile properties
+
+Below is an explanation of individual properties of the lockfile.
+
+##### download_url (required)
+
+Specified as a string containing the download url of the artifact.
+
+##### checksum (required)
+
+Specified as string in the format of "algorithm:hash". Must be provided to ensure the identity of the artifact.
+
+#### filename (optional)
+
+This key is provided mainly for the users convenience, so the files end up in expected locations. It is optional and if
+not specified, it will be derived from the download_url. Filename here is a path inside cachi2's output directory for
+the generic fetcher (`{cachi2-output-dir}/deps/generic`). Cachi2 will verify that the resulting filenames, including those
+derived from download urls do not overlap.
+
+### SBOM components
+
+Artifacts fetched with the generic fetcher will all be recorded in the SBOM cachi2 produces. Given the inability to derive
+any extra information about these files beyond a download location and a filename, these files will always be recorded
+as SBOM components with purl of type generic.
+
+Additionally, the SBOM component will contain [externalReferences] of type `distribution` to indicate the url used to download
+the file to allow for easier handling for tools that might process the SBOM.
+
+Here's an example SBOM generated for above file.
+
+```json
+{
+  "bomFormat": "CycloneDX",
+  "components": [
+    {
+      "name": "granite-model-1.safetensors",
+      "purl": "pkg:generic/granite-model-1.safetensors?checksum=sha256:d16bf783cb6670f7f692ad7d6885ab957c63cfc1b9649bc4a3ba1cfbdfd5230c&download_url=https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors",
+      "properties": [
+        {
+          "name": "cachi2:found_by",
+          "value": "cachi2"
+        }
+      ],
+      "type": "file",
+      "externalReferences": [
+        {
+          "url": "https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors",
+          "type": "distribution"
+        }
+      ]
+    }
+  ],
+  "metadata": {
+    "tools": [
+      {
+        "vendor": "red hat",
+        "name": "cachi2"
+      }
+    ]
+  },
+  "specVersion": "1.4",
+  "version": 1
+}
+```
+
+## Consequences
+
+As mentioned before, this package manager enables users to fetch arbitrary files with cachi2 and have them accounted for
+in the SBOM.
+
+[externalReferences]: https://cyclonedx.org/docs/1.6/json/#components_items_externalReferences
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -612,8 +612,9 @@ git clone -b sample-app https://github.com/cachito-testing/cachi2-generic.git
 ```
 
 #### Pre-fetch dependencies (generic fetcher)
-In order to retrieve the archive with the tool, a `artifacts.lock.yaml` needs to be in the repository. You can find a
-sample lockfile below. It is identical to one found in the [sample repository](https://github.com/cachito-testing/cachi2-generic/tree/sample-app).
+In order to retrieve the archive with the tool, either a `artifacts.lock.yaml` needs to be in the repository, or an absolute
+path needs to be supplied in the JSON input, pointing to a lockfile. You can find a sample lockfile below. It is identical
+to the one found in the [sample repository](https://github.com/cachito-testing/cachi2-generic/tree/sample-app).
 A lockfile for the generic fetcher must contain a `metadata` header and a list of artifacts, where each artifact is
 represented as a pair of URL and a checksum string in the format of `"algorithm:checksum"`:
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -634,7 +634,7 @@ is assumed to be `.`. Since generic fetcher is still an experimental feature, it
 enabled with the `--dev-package-managers` flag.
 
 ```
-cachi2 fetch-deps --source ./cachi2-generic --output ./cachi2-output generic --dev-package-managers
+cachi2 fetch-deps --source ./cachi2-generic --output ./cachi2-output generic
 ```
 
 #### Build the application image (generic fetcher)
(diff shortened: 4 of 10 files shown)


docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Decision
+
+This change introduces a generic fetcher, an additional cachi2 package manager. This package manager utilizes a custom
+lockfile that is located in the input repository. Based on that lockfile, it will download files, save them into a requested
+location, and verify checksums. Below is a more detailed overview of the implementation.
+
+### Lockfile format
+
+Cachi2 expects the lockfile to be named `generic_lockfile.yaml`.
```
brunoapimentel, line 21: Nitpick: I don't quite like using underscore on filenames. I also think the name could be slightly clearer, such as `generic-artifacts.lock.yaml`.

Another suggestion would be to namespace it with `cachi2`, so that we make a clear statement that it is a Cachi2-defined file (something like `cachi2-artifacts.lock.yaml`). I know we had opinions against this in the original design, but even with the imminent Cachi2 name change, it might still make sense to namespace the lockfile name.

In the end, this is still a nitpick, I see no big problems with keeping the current name.

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,109 @@
...
+## Context
+
+The main motivation for this change is to cover use cases of users that need to download arbitrary files that don't fit
+within an established package ecosystem cachi2 could potentially otherwise support. The target audience is users that
+want to use cachi2 to achieve hermetic builds and want an easy way to also include these arbitrary files, that cachi2
+will account for in the SBOM it produces.
+
+## Decision
+
+This change introduces a generic fetcher, an additional cachi2 package manager. This package manager utilizes a custom
```
a-ovchinnikov, line 15:
```diff
- This change introduces a generic fetcher, an additional cachi2 package manager.
+ A new package manager for generic artifacts must be introduced.
```

(since this is a design doc, not a change that introduces anything).

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
kosciCZ: It is possible, but I wonder what would be the benefit, if the lockfile has to be generated specifically for cachi2's use. But going off RPM package manager, it would probably be good to keep consistency and allow users to specify that name.

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,109 @@
...
+
+##### checksums (required)
+
+Specified as a dictionary of checksum algorithms and their values. At least one cachi2-verifiable checksum must be provided
+to ensure at least some degree of confidence in the identity of the artifact.
+
+#### target (optional)
+
+This key is provided mainly for the users convenience, so the files end up in expected locations. It is optional and if
+not specified, it will be derived from the download_url. Target here means a specific subdirectory inside cachi2's output
```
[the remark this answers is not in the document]
kosciCZ: I think user would be more likely to expect a filename, same as with e.g. when you use curl to download a file. 

I might have not worded it correctly, it is actually a filepath. I don't think it needs to be mandatory, because there might be files, where you just don't care.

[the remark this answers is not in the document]
kosciCZ: (Changed the wording to mention a filepath instead of subdir)

docs/adr/0001-add-generic-fetcher.md
kosciCZ: My thought was that we would discuss here what needs to be changed, and those changes would be merged together with the ADR

docs/adr/0001-add-generic-fetcher.md
kosciCZ: I think it's not a good idea to namespace it with cachi2 for two reasons: 1) it's gonna change soon 2) it's inconsistent with the RPM package manager naming. 

Thefore I think it should be something similar as the [RPM lockfile name](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/rpm/main.py#L30-L30), which is `rpms.lock.yaml`, like you mentioned `artifacts.lock.yaml`, or perhaps even more verbose `build-artifacts.lock.yaml`

eskultety: `artifacts.lock.yaml` is something I would have gone with. I was also thinking about namespacing it later on, because it does make sense to indicate that the format is custom to a particular tool, but we'd then have to support multiple names for the files which I don't want. If anything, we should allow users to pass these files via CLI and then it can be named literally anything, so what @kosciCZ proposes works for me.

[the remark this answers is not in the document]
eskultety: One wouldn't need to check a random cachi2 only file to their upstream repo, but instead sideload it which makes IMO sense for formats which aren't ecosystem native or widely common.

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,109 @@
...
+field that will indicate the version of the lockfile format. It will also contain a list of artifacts (files) to download,
+each of the artifacts to have a URL, list of checksums, and optionally target location specified.
+
+```yaml
+metadata:
+  # uses X.Y semantic versioning
+  version: "1.0"
+artifacts:
+  - download_url: https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
+    target: granite-model-1.safetensors
```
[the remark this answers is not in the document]
eskultety: > Will `output_filename` be too verbose? L53 also tells us that this is, in fact, a directory name.

I think this is a matter of documentation, so a shorter `filename` name paired with a meaning description is all we need IMO.

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,109 @@
+# Add generic fetcher
+
+- Status: proposed
+- Date: 2024-10-30
+
```
eskultety, line 5: @brunoapimentel did we want ^this to be captured? I think in your ADR template attempt we said this bit of information isn't really useful to anyone, or do I remember this wrong?

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,109 @@
...
+Here's an example SBOM generated for above file.
+
+```json
+{
+  "bomFormat": "CycloneDX",
+  "components": [
+    {
+      "name": "granite-model-1.safetensors",
+      "purl": "pkg:generic/granite-model-1.safetensors?checksums=sha256:d16bf783cb6670f7f692ad7d6885ab957c63cfc1b9649bc4a3ba1cfbdfd5230c&download_url=https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors",
```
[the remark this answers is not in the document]
eskultety: >  the strongest algorithm among the ones that were valid. This might be a good resolution for the matter here.

Do we want to get into the business of comparing and ordering hash algs based on how strong we think they are? Shouldn't this be in the user's own interest to provide strong algorithms which it wouldn't matter which one would be picked?

That said, remind me, do we really want N checksums per-artifact and validate all of them? It's going to slow down the whole process and I wonder whether the benefits outweigh the slight performance penalty paid.

@brunoapimentel  What is the inherent problem with reporting N checksums in a purl? The whole SBOM is a write only machine readable artifact anyway, so why does it matter how many checksums appear in there?

eskultety (commented):
@kosciCZ minor suggestions I'd say but content wise not much really, I see the important stuff I would expect from an ADR mentioned there.

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
kosciCZ: So, to clarify the format of the supplied path, should the input be an absolute path of the lockfile? I've looked over other package managers, and it seems like the pip package manager only allows relative paths, but that would defeat the purpose of sideloading the lockfile from outside the upstream repo. 

Just making sure if using an absolute path is okay.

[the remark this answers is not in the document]
eskultety: > but that would defeat the purpose of sideloading the lockfile from outside the upstream repo.

Yes, correct. That go me thinking though - we don't allow resources outside the repo in general, this would be a first IIRC. This would allow the operator to override a checked in version of the lockfile (because CLI should always take precedence) making us download something else. I don't think that's a problem though because everything would be recorded in the SBOM which might then not pass additional verification and we're not able to tell good from bad content pointed to by a URL today anyway, so from that POV there's no difference between a checked in version of the manifest vs an external one, so I think it's the right way forward.

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,109 @@
...
+  "specVersion": "1.4",
+  "version": 1
+}
+```
+
+## Consequences
+
+As mentioned before, this package manager enables users to fetch arbitrary files with cachi2 and have them accounted for
+in the SBOM. Possible downside could be maintaining the lockfile format, as it is specific to cachi2 (which should be
+partially mitigated by versioning it).
```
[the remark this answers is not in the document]
kosciCZ: Removed.

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
eskultety: >  I did not bump up the lockfile version, because the feature is still in the experimental phase and I don't think supporting the previous name is of any benefit.

**Not bumping any version** whilst still within the experimental stage is okay and expected I'd argue, period :) .

docs/adr/0001-add-generic-fetcher.md
brunoapimentel: +1 for `artifacts.lock.yaml`

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
brunoapimentel: > Do we want to get into the business of comparing and ordering hash algs based on how strong we think they are

Fair point, we could only pick the first valid one in the lockfile.

> That said, remind me, do we really want N checksums per-artifact and validate all of them? It's going to slow down the whole process and I wonder whether the benefits outweigh the slight performance penalty paid.

I believe the use cases we already have identified based on Maven have several checksums per artifact. We can try to confirm this. In any case, if we believe a single checksum is enough, I'd say the lockfile should support only a single one.

>  The whole SBOM is a write only machine readable artifact anyway, so why does it matter how many checksums appear in there?

I was looking at cases where the SBOM was being read by humans, but I agree with you it is the minority of cases. I'd be ok with having many checksums in the purls, but I still think only the valid ones should be there.

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
brunoapimentel: I still think this hurts our ability to enforce reproducibility. As you said, this is a behavior without precedents we'd be introducing. 

We could get into a debate about how much should Cachi2 be in the business of enforcing reproducibility, given that for some package managers (e.g. `go`), that claim is already somewhat flawed (because `go` can potentially resolve things differently based on many factors). I'd love to have a strong claim on reproducibility, in a way that we could cache a prefetch output based on repo + commit hash (and potentially Cachi2's version).

In any case, I think we could discuss this sideloading as a follow up, as it won't be a breaking change to the interface we're establishing here.

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,109 @@
...
+
+```yaml
+metadata:
+  # uses X.Y semantic versioning
+  version: "1.0"
+artifacts:
+  - download_url: https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
+    target: granite-model-1.safetensors
+    checksums:
+        sha256: d16bf783cb6670f7f692ad7d6885ab957c63cfc1b9649bc4a3ba1cfbdfd5230c
```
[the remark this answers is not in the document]
eskultety: Yes.

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
eskultety: > We could get into a debate about how much should Cachi2 be in the business of enforcing reproducibility, given that for some package managers (e.g. go), that claim is already somewhat flawed (because go can potentially resolve things differently based on many factors). I'd love to have a strong claim on reproducibility, in a way that we could cache a prefetch output based on repo + commit hash (and potentially Cachi2's version).

I think arguments passed via the CLI doesn't inherently doesn't contradict the claim, does it? Once you inject the same file, the outcome should not be different, otherwise it's a bug. If you inject a different one, then it's a different fetch/build, so we can't really talk about reproducibility, can we?

> In any case, I think we could discuss this sideloading as a follow up, as it won't be a breaking change to the interface we're establishing here.

I was wondering about this one. If we make it a follow up, then we inherently force (once we move this out of experimental before introducing the CLI option to pass an alternative manifest) early adopters to check in their manifest files to their upstream projects. I don't think they'll then be bothered much to revert that change since it'll work, you know, renovate/dependabot stuff would be set up... So I'm tempted to say that the CLI extra arg work should happen before the experimental status is lifted.

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
kosciCZ: > > Do we want to get into the business of comparing and ordering hash algs based on how strong we think they are

> Fair point, we could only pick the first valid one in the lockfile.

This is a potentially bad approach. I know folks in EC plan on gating content based on a list of allowed algos, so the reporting one at random, as you suggest, might potentially end up creating an unnecessary problem.  

> I believe the use cases we already have identified based on Maven have several checksums per artifact. We can try to confirm this. In any case, if we believe a single checksum is enough, I'd say the lockfile should support only a single one.

I agree with this. This will let user select what they think is the most appropriate algo, so any potential gating failure will be directly caused by them, not cachi2's algo selection heurestic.  I will go with this approach, and implement it so it's a string of "alg:hash", like this:

```
artifacts:
  - download_url: https://huggingface.co/instructlab/granite-7b-lab/resolve/main/model-00001-of-00003.safetensors?download=true
    filename: granite-model-1.safetensors
    checksum: "sha256:d16bf783cb6670f7f692ad7d6885ab957c63cfc1b9649bc4a3ba1cfbdfd5230c"
```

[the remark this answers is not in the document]
eskultety: > if we believe a single checksum is enough, I'd say the lockfile should support only a single one.

I understand that specifying a list of checksums and letting a different tool further down a pipeline to choose from but if we filter out SHA1 then having N hashes doesn't really bring any more confidence than a single hash out of the family of recommended strong algorithms, so why bother supporting that, it doesn't improve either the quality of the SBOM or provide any more confidence in it than going with a single one.

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
eskultety: I think we can resolve this based on https://github.com/containerbuildsystem/cachi2/pull/728#discussion_r1836733350

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
eskultety: > See [this PR](https://github.com/red-hat-data-services/trustyai-explainability/pull/44/files) that does sideloading that way

Thank you for pointing this out - I think for the sake of higher SLSA levels and build artifact confidence we should treat this as a bug and force a clean repo usage.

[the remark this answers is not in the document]
eskultety: > I was unable to find a clear definition for what a "dirty" repository is

Simple, git knows about untracked files - if not empty OR there are staged/unstaged changes -> dirty repo.

> but couldn't that be avoided by just committing the file locally?

I'm not sure I follow.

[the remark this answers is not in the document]
kosciCZ: > I'm not sure I follow.
Instead of just copying that file into the repo, once could also actually commit it locally in the task, without actually pushing it anywhere. 

Something like this:
```
cp /tmp/generic_lockfile.yaml .
git config --global user.name "Foo Bar"
git config --global user.email "foo.bar@example.com"
git add generic_lockfile.yaml
git commit -m "Add file"
```
This would leave a clean repo.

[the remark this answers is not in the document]
eskultety: Oh, not a cachi2 problem. If one is determined enough to go and use such a dirty hack in their pipeline they're should be ready to reap the potential consequences, there's nothing we can do about this in cachi2 and - on a side note - I'd argue noone should even try providing support for pipeline configurations like that.

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
kosciCZ: Updated to only support a single checksum.

docs/adr/0001-add-generic-fetcher.md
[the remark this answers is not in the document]
kosciCZ: Updated to support specifying an absolute lockfile path as part of the JSON input.

docs/adr/0001-add-generic-fetcher.md
kosciCZ: Changed.

docs/adr/0001-add-generic-fetcher.md
kosciCZ: Default  lockfile name is now `artifacts.lock.yaml`.

docs/adr/0001-add-generic-fetcher.md
kosciCZ: Removed.

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,104 @@
...
+## Context
+
+The main motivation for this change is to cover use cases of users that need to download arbitrary files that don't fit
+within an established package ecosystem cachi2 could potentially otherwise support. The target audience is users that
+want to use cachi2 to achieve hermetic builds and want an easy way to also include these arbitrary files, that cachi2
+will account for in the SBOM it produces.
+
+## Decision
+
+A new package manager for generic artifacts must be introduced.. This package manager utilizes a custom
```
a-ovchinnikov, line 12: s/.././

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,104 @@
...
+lockfile that is located in the input repository. Based on that lockfile, it will download files, save them into a requested
```
a-ovchinnikov, line 13:
```diff
- lockfile that is located in the input repository. Based on that lockfile, it will download files,
+ lockfile basing on which it will download files,
```

(since the file can be outside a repo now)

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,104 @@
+# Add generic fetcher
+
+## Context
+
+The main motivation for this change is to cover use cases of users that need to download arbitrary files that don't fit
```
a-ovchinnikov, line 5: Non-blocking suggestion:
```diff
- The main motivation for this change is to cover use cases of users that need to download arbitrary files that don't fit
+ Some users need to download arbitrary files that don't fit
```

(since the document is not describing an individual change, but a new pseudo-package manager; also a bit easier to read)

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,104 @@
...
+within an established package ecosystem cachi2 could potentially otherwise support. The target audience is users that
+want to use cachi2 to achieve hermetic builds and want an easy way to also include these arbitrary files, that cachi2
```
a-ovchinnikov, line 7:
```diff
- and want an easy way to also include these arbitrary files
+ and also want an easy way to include these arbitrary files

File: docs/adr/0001-add-generic-fetcher.md [hunk 13]
```diff
@@ -0,0 +1,104 @@
...
+A new package manager for generic artifacts must be introduced.. This package manager utilizes a custom
+lockfile that is located in the input repository. Based on that lockfile, it will download files, save them into a requested
+location, and verify checksums. Below is a more detailed overview of the implementation.
+
+### Lockfile format
+
+Cachi2 expects the lockfile to be named `artifacts.lock.yaml`.
+In order to account for possible future breaking changes, the lockfile will contain a `metadata` section with a `version`
+field that will indicate the version of the lockfile format. It will also contain a list of artifacts (files) to download,
+each of the artifacts to have a URL, list of checksums, and optionally resulting filename specified.
```
a-ovchinnikov, line 21:
```diff
- each of the artifacts to have a URL, list of checksums, and optionally resulting filename specified
+ each of the artifacts having a URL, a checksums, and optionally output filename specified
```

docs/adr/0001-add-generic-fetcher.md
```diff
@@ -0,0 +1,104 @@
...
+Below is an explanation of individual properties of the lockfile.
+
+##### download_url (required)
+
+Specified as a string containing the download url of the artifact.
+
+##### checksum (required)
+
+Specified as string in the format of "algorithm:hash". Must be provided to ensure at least some
+degree of confidence in the identity of the artifact.
```
a-ovchinnikov, line 44: Non-blocking, opinion based: I suggest dropping either "at least some degree of confidence" or the entire second sentence altogether. A checksum is a mean to validate identity of an artifact, if someone insists on arbitrary files in their build and provide a checksum then we must assume that this is exactly what they want as long as the checksum matches.

a-ovchinnikov (commented):
A few trivial modifications to ADR to reflect the most recent changes, also some wording nitpicks, LGTM otherwise.

cachi2/core/package_managers/generic/models.py
```diff
@@ -10,6 +11,8 @@
 from cachi2.core.errors import PackageManagerError
 from cachi2.core.rooted_path import RootedPath
 
+CHECKSUM_FORMAT = re.compile(r"^[a-zA-Z0-9]+:[a-zA-Z0-9]+$")
+
 
 class LockfileMetadata(BaseModel):
```
brunoapimentel, line 17: Just a small, non-blocking suggestion:

It might be worth to add a `extra="forbid"` option here. I was doing a test with an outdated lockfile, and the presence of the `target` attribute did not trigger any error.

docs/usage.md
```diff
@@ -612,20 +612,19 @@ git clone -b sample-app https://github.com/cachito-testing/cachi2-generic.git
...
 A lockfile for the generic fetcher must contain a `metadata` header and a list of artifacts, where each artifact is
-represented as a pair of URL and a map containing one or more items in the shape of `"hashing algorithm": "checksum"`:
+represented as a pair of URL and a checksum string in the format of `"algorithm:checksum"`:
 
 ```
 ---
 metadata:
   version: "1.0"
 artifacts:
   - download_url: "https://github.com/jeremylong/DependencyCheck/releases/download/v11.1.0/dependency-check-11.1.0-release.zip"
```
brunoapimentel, line 626: Should we also add the `filename` attribute here? I noticed this attribute isn't mentioned anywhere in the documentation.

cachi2/core/package_managers/generic/main.py
```diff
@@ -32,24 +32,29 @@ def fetch_generic_source(request: Request) -> RequestOutput:
     components = []
     for package in request.generic_packages:
         path = request.source_dir.join_within_root(package.path)
-        components.extend(_resolve_generic_lockfile(path, request.output_dir))
+        lockfile = package.lockfile or path.join_within_root(DEFAULT_LOCKFILE_NAME).path
```
brunoapimentel, line 35: I'm not sure if requiring users to provide absolute paths for the lockfile via the CLI  is the best approach here. It might be tricky in the context of a CI system where the user does not know exactly what the path to the lockfile will be. We could potentially allow for relative paths if we resolved them within `request.source_dir`. 

Did we verify if this approach would solve any current use case? I'm getting the impression that fetching the lockfile from a remote location could be needed in some cases.

brunoapimentel: Another concern is security implications of allowing users to specify a filepath that resolves outside of the request's root dir, as it could be used to exploit any larger CI system that Cachi2 is part of. For the current use cases, Cachi2 always verifies that all the files it is reading are within the request's root dir, and that's the reason we have the `RootedPath` class.

This is very important in context of using package managers built-in commands that are outside of our control. I reckon that in this case, we'd be safer since we have control over what we're doing with the file. Still, I think some extra consideration is needed since this sets a new precedent for the project.

Ideally, from a security perspective, we'd allow only relative paths based on `request.source_dir`. But on the other hand, we're trying to solve the use case of upstream projects that would not commit an `artifacts.lock.yaml` file to it, so this would not be sufficient.

brunoapimentel: As I suggested before, we could postpone this to another moment in case we're not certain of the decision now. I don't think we should block this ADR because of this.

brunoapimentel (commented):
This LGTM, the only thing I wanted is more feedback on this concern [here](https://github.com/containerbuildsystem/cachi2/pull/728#discussion_r1838535149).

My last comment is that I think we should use this PR to move `generic` out of `dev-package-managers`, effectively making it stable.

cachi2/core/package_managers/generic/main.py
eskultety: FYI We as reviewers are contradicting each other in our suggestions: https://github.com/containerbuildsystem/cachi2/pull/728#discussion_r1836490114 which is what @kosciCZ followed so we need to get things straight rather quickly otherwise it's an awfully frustrating experience for the contributor.

> Another concern is security implications of allowing users to specify a filepath that resolves outside of the request's root dir, as it could be used to exploit any larger CI system that Cachi2 is part of. For the current use cases, Cachi2 always verifies that all the files it is reading are within the request's root dir, and that's the reason we have the `RootedPath` class.
> 

We do and that makes total sense for most scenarios, manifest files IMO not being one of them, but in general we should keep honoring this I agree.

> This is very important in context of using package managers built-in commands that are outside of our control. I reckon that in this case, we'd be safer since we have control over what we're doing with the file. Still, I think some extra consideration is needed since this sets a new precedent for the project.
> 

The manifest file is a declarative yaml file, so in order for this to be exploitable the absolute path would have to be a symlink to a valid YAML file containing some hidden parser instructions to exploit the YAML parser itself which would then need to gain full control over our own process; other than that I don't see how we are safer with a potentially malicious checked-in manifest file, IOW this manifest file only contains data pointing to online resources which we are not able to evaluate on our own anyway, so how do you see this manifest file exploiting the containing CI systems cachi2 is part of? I think any pipeline subsystem/service should be hardened on its own for this very reason that any component can essentially be exploited and so nothing can be implicitly trusted and attack vectors can only be mitigated, not avoided.

> Ideally, from a security perspective, we'd allow only relative paths based on `request.source_dir`. But on the other hand, we're trying to solve the use case of upstream projects that would not commit an `artifacts.lock.yaml` file to it, so this would not be sufficient.

Relative paths solve nothing in general because the same file a user would provide via an absolute path they'd now purposfully copy over to the repository to pass our check, so from security perspective we didn't achieve anything, the file is either checked-in to the repo in which case we're putting our confidence into the community collective responsibility for the project or it isn't checked into the repository in which case we have absolutely no guarantee on anything regardless of where the file came from and since users can customize their pipeline automation (e.g. tekton) you can bet they'll use whatever power they have to workaround our trivial path checks (also demonstrated by an example posted by @kosciCZ here: https://github.com/red-hat-data-services/trustyai-explainability/pull/44/files)



> As I suggested before, we could postpone this to another moment in case we're not certain of the decision now. I don't think we should block this ADR because of this.

I disagree , I outlined why in: https://github.com/containerbuildsystem/cachi2/pull/728#discussion_r1836731971 . If we make this a follow up work, then:

1. it'll likely never happen and 
2. we'd be forcing our user base to check the manifest file to their repos and once they do so, they won't be willing to go back easily which is bad UX IMO solely because we're forcing people to commit random manifest files to their repos which don't really relate to their project in any way it's just that someone told the project to be built in a particular manner which requires them to use a non-standard configuration file which has no widely-accepted ecosystem around it - perfectly fine for downstream solutions (do whatever you want), not okay with upstream projects IMO

eskultety:
> This LGTM, the only thing I wanted is more feedback on this concern [here](https://github.com/containerbuildsystem/cachi2/pull/728#discussion_r1838535149).
> 
> My last comment is that I think we should use this PR to move `generic` out of `dev-package-managers`, effectively making it stable.

If you want to lift the experimental status, then I believe the ability to consume manifest files for the generic fetcher via a secondary channel needs to addressed, but most importantly all maintainers need to be in agreement with the approach.

kosciCZ:
> If you want to lift the experimental status, then I believe the ability to consume manifest files for the generic fetcher via a secondary channel needs to addressed, but most importantly all maintainers need to be in agreement with the approach.

@eskultety So do I understand correctly that despite everything else being resolved and ready, this PR is now blocked for two weeks since Bruno is on vacation, despite the feature of being able to specify an absolute lockfile path being approved by two out of three reviewers?

docs/adr/0001-add-generic-fetcher.md
kosciCZ: Fixed.

docs/adr/0001-add-generic-fetcher.md
kosciCZ: Fixed.

docs/adr/0001-add-generic-fetcher.md
kosciCZ: Fixed.

docs/adr/0001-add-generic-fetcher.md
kosciCZ: Fixed.

docs/adr/0001-add-generic-fetcher.md
kosciCZ: Fixed.

docs/adr/0001-add-generic-fetcher.md
kosciCZ: Fixed.

docs/usage.md
kosciCZ: Added.

cachi2/core/package_managers/generic/models.py
kosciCZ: Added.

a-ovchinnikov (approved):

cachi2/core/package_managers/generic/main.py
chmeliik: The potential exploit would be this: `{"type": "generic", "lockfile": "/root/.docker/config.json"}` - an attempt to make cachi2 raise a validation error which would print the content of a file that the CI task wasn't intended to access

kosciCZ: That should not happen. Here's an attempt to supply cachi2 with an arbitrary json file:
```
$ cachi2 fetch-deps --source ../cachi2-generic  --output ../cachi2-generic/cachi2-output '{"type": "generic", "lockfile": "/home/jakoscie/redhat/cachi2-generic/cachi2-output/bom.json"}' --dev-package-managers
2024-11-14 13:52:42,074 INFO Reading generic lockfile: /home/jakoscie/redhat/cachi2-generic/cachi2-output/bom.json
2024-11-14 13:52:42,077 ERROR PackageRejected: Cachi2 lockfile '/home/jakoscie/redhat/cachi2-generic/cachi2-output/bom.json' format is not valid: '('metadata', 'version'): Field required'
Error: PackageRejected: Cachi2 lockfile '/home/jakoscie/redhat/cachi2-generic/cachi2-output/bom.json' format is not valid: '('metadata', 'version'): Field required'
  Check the correct format and whether any keys are missing in the lockfile.


```

eskultety: > The potential exploit would be this: `{"type": "generic", "lockfile": "/root/.docker/config.json"}` - an attempt to make cachi2 raise a validation error which would print the content of a file that the CI task wasn't intended to access

This is already a problem with our own config which can be supplied on the CLI from an arbitrary location and fail validation the same way already, so this issue exists regardless of this PR. What you're describing is only a problem in environments where cachi2 runs with elevated privileges and cannot be prevented as long as you accept *any* paths in your project, that's why it's fundamentally wrong running cachi2 with elevated privileges as is running pretty much any tool out there. Furthermore, what you're going to get at best is data out of cachi2's own environment which is already a public image; that said, the attacker would have to know what data is mounted into what location to cachi2's own environment where they're certain we run with elevated privileges and at the same time they would have to make sure what they're interested in is either a JSON or a YAML file that the parser can consume, but that would fail validation.

eskultety: > That should not happen. Here's an attempt to supply cachi2 with an arbitrary json file:

@kosciCZ that is a result of the fact that the global lockfile model isn't defined with `extra=forbid` which I guess we want as it's consistent with other models we have which will then get you exactly to what @chmeliik is trying to point out - by complaining about each extra element (here's an example of our own YAML config spoofed):
```
cachi2 --config-file /tmp/pytest-of-eskultety/pytest-current/test_e2e_bundler_bundler_everycurrent/bundler_everything_present-output/bom.json fetch-deps
2024-11-14 14:22:25,164 ERROR InvalidInput: 5 validation errors for user input\nbomFormat\n  Extra inputs are not permitted\ncomponents\n  Extra inputs are not permitted\nmetadata\n  Extra inputs are not permitted\nspecVersion\n  Extra inputs are not permitted\nversion\n  Extra inputs are not permitted
Error: InvalidInput: 5 validation errors for user input
bomFormat
  Extra inputs are not permitted
components
  Extra inputs are not permitted
metadata
  Extra inputs are not permitted
specVersion
  Extra inputs are not permitted
version
  Extra inputs are not permitted
```
which gives you an idea about the top-level keys in a file.

That still doesn't make this a feasible attack vector because it's truly relying on the fact that the attacker knows exactly what data is going to be mount where inside the cachi2 container to leverage the elevated privileges. If done properly, everything is mounted to random locations out of which cachi2 reads data, meaning that any attempt to spoof data out of the environment is only limited to cachi2's own container image which, like I said, is public already.

eskultety:
> > If you want to lift the experimental status, then I believe the ability to consume manifest files for the generic fetcher via a secondary channel needs to addressed, but most importantly all maintainers need to be in agreement with the approach.
> 
> @eskultety So do I understand correctly that despite everything else being resolved and ready, this PR is now blocked for two weeks since Bruno is on vacation, despite the feature of being able to specify an absolute lockfile path being approved by two out of three reviewers?

First of all, I don't think this personal plan/situation was publicly announced anywhere and so I don't consider it appropriate doing it on his behalf this way in a public space, so please respect others' privacy and refrain from doing so in the future.

Now, to your question, I will summarize what I said in the relevant threads regarding this concern:

- allowing external YAML config files is only as safe as the YAML parser itself we rely on
- we already allow our own cachi2 config to be passed from an arbitrary location with an absolute path and it is a YAML, so if a manifest could be used to exploit the YAML parser so could be via our own cachi2 config
- we don't check whether files are truly checked into a repo and so if we don't allow arbitrary paths for the artifact manifest people will deliberately circumvent that by injecting the file from a different location just before running cachi2
- forcing the manifest to be read from the repo and repo only is IMO only a marginal improvement from security POV (that said, additional hardening should be employed wherever cachi2 is to be executed with elevated privileges), but a bigger convenience penalty for the user base
- makes IMO no sense to force people checking a random YAML manifest file to their repos to satisfy a particular 3rd party build toolchain

With all that, no, I am not convinced this work needs to be blocked for any longer with the approvals in place.

cachi2/core/package_managers/generic/main.py
kosciCZ: > @kosciCZ that is a result of the fact that the global lockfile model isn't defined with extra=forbid which I guess we want as it's consistent with other models we have which will then get you exactly to what 

But it is defined with that option: https://github.com/containerbuildsystem/cachi2/pull/728/files#diff-a870dba66695ba9805db0cc2ebd3799b87de8d5f58544d8a2482d7ce17240f2dR81

Anyway, do I understand correctly that this is out of scope for this PR?

eskultety: > But it is defined with that option: https://github.com/containerbuildsystem/cachi2/pull/728/files#diff-a870dba66695ba9805db0cc2ebd3799b87de8d5f58544d8a2482d7ce17240f2dR81
>

Nope, this line is the important one: https://github.com/containerbuildsystem/cachi2/pull/728/files#diff-a870dba66695ba9805db0cc2ebd3799b87de8d5f58544d8a2482d7ce17240f2dL70
 
> Anyway, do I understand correctly that this is out of scope for this PR?

From my personal POV the problem already exists so your PR makes it only marginally worse and best we can do, if anything, is to overload the default validator for each model so that it doesn't disclose the top-level keys which also feels wrong and would be rather ugly. Since we're talking JSON which is a dict-like format, the chance of disclosing actual sensitive values is also IMO rather small.

MartinBasti: > The potential exploit would be this: {"type": "generic", "lockfile": "/root/.docker/config.json"} - an attempt to make cachi2 raise a validation error which would print the content of a file that the CI task wasn't intended to access

but this configuration must be done by user right? In CI tools user can add an extra task to print env and whatever, so this is not a security issues. It's expected and reason why CI tools don't automatically start for non maintainers, as code must be reviewed first

kosciCZ:
I appreciate the approval. I will add one more commit that lifts the experimental status.

cachi2/core/package_managers/generic/main.py
chmeliik: But if the validation errors only print the keys and not values, it should be fine in 99% of cases.

I also tried a non-yaml file, but that's a safe error message as well

```
2024-11-14 15:22:10,403 INFO Reading generic lockfile: /home/acmiel/RedHat/Stonesoup/tssc-dev-multi-ci/.env
2024-11-14 15:22:10,403 ERROR PackageRejected: Cachi2 lockfile '/home/acmiel/RedHat/Stonesoup/tssc-dev-multi-ci/.env' yaml format is not correct: expected '<document start>', but found '<scalar>'\n  in "/home/acmiel/RedHat/Stonesoup/tssc-dev-multi-ci/.env", line 10, column 1
Error: PackageRejected: Cachi2 lockfile '/home/acmiel/RedHat/Stonesoup/tssc-dev-multi-ci/.env' yaml format is not correct: expected '<document start>', but found '<scalar>'
  in "/home/acmiel/RedHat/Stonesoup/tssc-dev-multi-ci/.env", line 10, column 1
  Check correct 'yaml' syntax in the lockfile.
```

chmeliik: The point is more that until now, cachi2 promised not to read/write files outside the `--source`, `--output` and `--config-file` locations. This change breaks that promise, but if the error messages are safe it's not a huge deal

eskultety: > The point is more that until now, cachi2 promised not to read/write files outside the `--source`, `--output` and `--config-file` locations. This change breaks that promise, but if the error messages are safe it's not a huge deal

Do we have a promise like that anywhere? Also, that promise isn't broken with a default CLI, cachi2 would only ever look beyond if instructed with additional JSON input arguments on the CLI, so I don't think we're breaking any promises.

chmeliik: TLDR: the error message could be a tiny bit better (avoid exposing the keys), but no further concerns from my side

MartinBasti: I had discussion with Adam, so I see there 2 POW:

1) INPUT is just considered to be regular CLI params, just in json form, then wrapper around CLI (user, OSBS) must ensure, that provide paths are right (any paths from nested files must not get out of CWD)

2) INPUT is just inline config file and wrapper cannot do much about it, then cachito should ensure that it's not reading outside of CWD, as wrapper may not have the power to ensure paths are safe.

Given the complexity of JSON file, I'm probably more eager to agree with 2


*CWD or dir explicitly set by separate option

chmeliik: idealistically my view is option 2, but practically making an exception for the generic fetcher is fine IMO

MartinBasti: let's go with 1) ; I think users may want to provide this file from a separate task, not via source

kosciCZ:
This is now ready to be merged, per my understanding.

PR #731, opened 2024-11-11
Replace tox.ini and Makefile for noxfile.py

slimreaper35:
https://nox.thea.codes/en/stable

### Notes

- Each session is defined as a Python function with the `@nox.session` decorator
- `reuse_venv=True` is used to speed up the execution of sessions by reusing the existing venv
- Docstrings below each session function are used as session descriptions in the help output
- The module docstring is used in the help output when running `nox -l, --list` too
- The `*session.posargs` is used to pass additional arguments to the command
- The `silent=True` argument is used to suppress the output of the command *only* if the command was successful
- To use external commands, set the `external=True` argument in the `session.run()` method
  (it should work without the `external=True` argument, but it's better to be explicit and silence the warning)
- The only command from Makefile that is not included here is `make venv`
  (it probably doesn't make sense to have a session for that, it would require contributors to install `nox` first)
- I made some minor adjustments to `pytest` commands to make them more readable and clean
---


noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+    )
+    session.run(
+        "pip-compile",
+        "--all-extras",
+        "--generate-hashes",
+        "--output-file=requirements-extras.txt",
+        "--rebuild",
+        "pyproject.toml",
+        silent=True,
+    )
```
eskultety, line 228: This one becomes confusing with the other target (podman) in context. Since we have a hard dependency on pip-compile to always run with 3.9, risking that a developer unaware of this rule will try submitting requirements updates which are easy to slip in during the review process, so I'd say this one should be dropped and the target should always run in a container as it's been the case for a while via the Makefile.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+To list all available sessions, run the following command:
+$ nox -l
+"""
+
+import nox
+from nox.sessions import Session
+
+# default sessions (sorted alphabetically)
+nox.options.sessions = ["bandit", "black", "flake8", "isort", "mypy", "unit-tests"]
+
```
eskultety, line 23: you should be able to set `nox.options.reuse_venv` globally as well instead of setting it with each target. I also think we'd want to set it to 'always' if possible (you can do that over CLI, not sure if it requires a boolean this way) as that is what we want I guess based on the proposal.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+    session.run(
+        "podman",
+        "build",
+        "--pull-always",
+        "--no-cache",
+        "--tag",
+        "localhost/cachi2:latest",
+        ".",
+        external=True,
+    )
```
eskultety, line 205: I wonder if anyone's ever used this feature. If one suspects a problem originating in an old base image then `--pull-always` and `--no-cache` should be in their docker arsenal. I think we can safely drop this, it was just a weird kludge that doesn't really relate to the project itself in any way.
That said, since this existed as a Makefile target, it should be dropped either prior this conversion or after, not in the same commit, the migration should be a 1:1 migration of Makefile + tox, maybe even split this particular commit in 2 - first migrate tox.ini to nox, then Makefile (or vice versa), then remove the corresponding origins.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+def build_image(session: Session) -> None:
+    """Build the cachi2 container image"""
+    session.run(
+        "podman",
+        "build",
+        "--tag",
+        "localhost/cachi2:latest",
+        ".",
+        external=True,
+    )
```
eskultety, line 190: Similarly to the below I don't think this should exist as it pretty much just wraps standard podman CLI interface without any exotic CLI options so I question its usefulness - the leaner we keep this file the better.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+        "--install-types",
+        "--non-interactive",
+        "cachi2",
+        "tests",
+        *session.posargs,
+        silent=True,
+    )
+
+
+@nox.session(name="unit-tests", reuse_venv=True, python=["3.9", "3.10", "3.11", "3.12"])
```
eskultety, line 107: nitpick:
`["3.9", "3.10", "3.11", "3.12"]`

In order to avoid changing multiple files at the same time, since noxfile is a Python script we could introduce a simple helper that would parse the `requires-python` field from the `pyproject.toml` file and enumerate the version set from the specifier (based on `packaging.specifiers.SpecifierSet`), i.e. if we adjust the specifier to `>=3.9, <=3.12` such helper would yield this exact list. That way, any time we add a new supported version of Python (e.g. 3.13) or deprecate the lower boundary (i.e. 3.9 in this case), `noxfile` will likely not need an adjustment because of such update.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+        *session.posargs,
+        silent=True,
+    )
+
+
+@nox.session(reuse_venv=True)
+def flake8(session: Session) -> None:
+    """Run flake8 on cachi2 and tests directories"""
+    _install_requirements(session)
+    # TODO: Move flake8-pyproject to requirements-extras.txt
```
eskultety, line 64: What's the blocker for this TODO work? Seems simple enough which means I'm missing something.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+    )
+
+
+@nox.session(name="all-integration-tests", reuse_venv=True)
+def all_integration_tests(session: Session) -> None:
+    """Run all integration tests that are available"""
+    _run_integration_tests(
+        session,
+        {
+            "CACHI2_RUN_ALL_INTEGRATION_TESTS": "true",
```
eskultety, line 163: Aren't environment variables automatically passed through? I would expect us to only define the `integration_tests` target and let it be tweaked with an env variable, that's how I used to it anyway prior this work. I think documenting all the env vars we use anywhere in the project is paramount for that. Unless it doesn't work the way I imagine it does/should in which case my argument becomes invalid.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+
+@nox.session(name="generate-test-data", reuse_venv=True)
+def generate_test_data(session: Session) -> None:
+    """Run all integration tests that are available and update SBOMs"""
+    _run_integration_tests(
+        session,
+        {
+            "CACHI2_RUN_ALL_INTEGRATION_TESTS": "true",
+            "CACHI2_GENERATE_TEST_DATA": "true",
+        },
```
eskultety, line 176: Similarly here, noxfile is IMO opaque, meaning that nobody will bother looking into it unless they need to add or fix a target, so I think documenting the usage of these env vars to adjust the behaviour of the test target is IMO more transparent than essentially defining "sugar coated" targets which just hinder transparency; I think less is more in this case.

pyproject.toml
```diff
@@ -100,3 +100,8 @@ skips = [
 [tool.pytest.ini_options]
 log_format = "%(asctime)s %(levelname)s %(message)s"
 log_date_format = "%Y-%m-%d %H:%M:%S"
+
```
eskultety, line 103: Empty commit message, both here and in the previous commit - some reasoning would be nice.

eskultety (commented):
@slimreaper35 this is a very shallow top-level review I've done without trying it locally myself, but from the first look it looks fine, definitely doesn't hinder readability when compared to the declarative tox.ini config, so from that POV seems like a solid alternative to what we have today.

noxfile.py
slimreaper35: I added this one as _optional_ or alternative if someone does not want to run a container.
> @nox.session(name="pip-compile", reuse_venv=True, python="3.9")

python="3.9" ensures that the session will not run if python3.9 is not installed locally

noxfile.py
slimreaper35: I will do that in the next revision.

noxfile.py
slimreaper35: With the growing amount of our custom env variables, this process becomes much more complicated. Recently, I wanted to generate data for integration tests and I had to define two or more mouthful env variables for that. The next day I would forget the exact names and I could start again looking for them somewhere in the code base.

I would rather have an abstraction layer as it is in this file. As I mentioned in the PR template each session's docstring is displayed in the help message. That seems enough IMO. Just keep it simple.

noxfile.py
eskultety: Seems pointless to me doing it this way when you already have a solution in place that enforces the correct Python version regardless of the one installed locally AND works unconditionally all the time.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+def unit_tests(session: Session) -> None:
+    """Run unit tests and generate coverage report"""
+    _install_requirements(session)
+    session.run(
+        "pytest",
+        "--log-level=DEBUG",
+        "--cov=cachi2",
+        "--cov-config=pyproject.toml",
+        "--cov-report=term",
+        "--cov-report=html",
```
a-ovchinnikov, line 117: While we are here could you please add `--no-cov-on-fail`? Coverage report is a bit useless when there are failures, but consumes screen space.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+        "tests/unit",
+        *session.posargs,
```
a-ovchinnikov, line 119: What do you think about making UTs silent as well? (This would require to somehow exfiltrate coverage which will get clobbered along with tests output otherwise)

a-ovchinnikov (commented):
This looks so neat!

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+    )
+
+
+@nox.session(reuse_venv=True)
+def black(session: Session) -> None:
+    """Run black on cachi2 and tests directories and noxfile.py"""
+    _install_requirements(session)
+    session.run(
+        "black",
+        "--check",
```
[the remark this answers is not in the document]
slimreaper35: why ?

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+
+To list all available sessions, run the following command:
+$ nox -l
+"""
+
+import nox
+from nox.sessions import Session
+
+# default sessions (sorted alphabetically)
+nox.options.sessions = ["bandit", "black", "flake8", "isort", "mypy", "unit-tests"]
```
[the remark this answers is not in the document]
slimreaper35: I don't know. That would mean different behavior compared to GH actions.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+    session.run("coverage", "html")
+    session.run("xdg-open", "htmlcov/index.html", external=True)
+
+
+def _run_integration_tests(session: Session, env: dict[str, str]) -> None:
+    _install_requirements(session)
+    default_env = {
+        "CACHI2_TEST_NETRC_CONTENT": "machine 127.0.0.1 login cachi2-user password cachi2-pass",
+        "CACHI2_TEST_LOCAL_PYPISERVER": "true",
+        "CACHI2_TEST_LOCAL_DNF_SERVER": "true",
```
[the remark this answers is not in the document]
slimreaper35: Well, it depends on how important these special tests are and how we define the sessions. I would assume if I want to run all integration tests, that _all_ integration tests would run.

noxfile.py
slimreaper35: Do you mean something like that ?
```
def _get_supported_python_versions() -> list[str]:
    """
    Read pyproject.toml and return a list of all supported Python versions
    based on the requires-python specification.
    """
    pyproject_path = Path("pyproject.toml")
    pyproject = tomli.loads(pyproject_path.read_text())
    requires_python = pyproject["project"]["requires-python"]

    specifiers = SpecifierSet(requires_python)
    min_minor = None
    max_minor = None

    for spec in specifiers:
        minor = int(re.match(r"3\.(\d+)", spec.version).group(1))

        if spec.operator in (">=", ">"):
            if min_minor is None or minor > min_minor:
                min_minor = minor
        elif spec.operator in ("<=", "<"):
            if max_minor is None or minor < max_minor:
                max_minor = minor

    return [f"3.{minor}" for minor in range(min_minor, max_minor + 1)]
```

noxfile.py
eskultety: > Such targets add clues to how cachi2 could be tested. I suggest keeping them in place until there is a superb troubleshooting document.

I don't think this is useful in any way, we're pretty much the only project I've encountered that tried to wrap *some* of podman's flags to do this, anyone even remotely familiar with podman would not really use this. I mean, I think with this @slimreaper35's effort it's time that we drop some of these kludges altogether and only define what's necessary for the project to decrease clutter, improve transparency, promote flexibility in a way where every user is motivated to use standard invocation mechanisms and be in full control over the configuration of such an invocation if what I just wrote makes sense at all. In this very case I don't think any extensive documentation is needed at all, having a pristine image is just one of the troubleshooting and debugging approaches, we're not wrapping other common invocations.

noxfile.py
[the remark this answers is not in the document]
eskultety: > To automatically apply `black` suggestions. What is the normal workflow btw? Does everyone generate a separate diff, inspect it and then apply?

Yes. The reason is that it would only make sense during a rebase, if you run it over a patch series you still have to do your due diligence and apply individual changes and still do a rebase to squash those changes to the right commit. Until we run UTs+linters in a rebase in our GH CI (we'd still have to tweak the action so that commit amends are allowed, do we want that in general?) I don't think we want automatic handling of these.

noxfile.py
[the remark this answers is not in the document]
eskultety: > My reasoning is that it is nice to fail as fast as possible when working on a change locally.

I understand the sentiment and also see some value in this, but then

>  if there is a flake8 issue I know that I need to fix it or it will cascade to other linters, it does not make sense to wait until all targets are done

This isn't true for black/isort errors, so e.g. what happens for me often is that I have flake8 + formatter errors reported and currently all it takes is one execution of tox to tell me about all of the errors immediately and I don't need to run tox more than twice per reported error (1 initial and 1 to verify). Linters are pretty fast IMO, most time is consumed by mypy and UTs which this setting would not solve. Also, I think the overall team sentiment is to move off of these individual linters onto Ruff which will just render all this irrelevant given that it's written in a compiled language and so, again, the only real bottleneck in nox execution will be the UTs.

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+        "isort",
+        "--check",
+        "--diff",
+        "--color",
+        "cachi2",
+        "tests",
+        "noxfile.py",
+        *session.posargs,
+        silent=True,
+    )
```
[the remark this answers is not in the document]
eskultety: Matter of taste - how about we format invocations that fit within the 100 columns on a single line and only split those which would not?

noxfile.py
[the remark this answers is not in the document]
eskultety: > Good point, but you have defined this env not for all_tests target, but on a higher level

Fully agreed, I don't want those servers to be spawned most of the time on my local machine, so this should only ever be defaulting to true for the `all_tests` target.

noxfile.py
eskultety: Fair enough, I guess I need to think about this (after trying) harder and potentially propose a follow-up PR if anything comes out of that thought process, so you can disregard my earlier comment for the purposes of getting the PR accepted.

noxfile.py
slimreaper35: It is doable, but if one test fails we lose the colorful output from pytest.

noxfile.py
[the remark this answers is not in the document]
slimreaper35: Done.

noxfile.py
eskultety: Using `int(re.match(r"3\.(\d+)", spec.version).group(1))` regexes for this purpose seems like a big overkill to me when packaging already provides a `Version` object that can be compared.

Also,
```
if spec.operator in (">=", ">"):
            if min_minor is None or minor > min_minor:
                min_minor = minor
```
is incorrect for e.g. `>=3.6, >3.7` because your `min_minor` would be 7 but due to the `>` operator it has to be 8, so you'd have to further special case that. We're already making a compromise here assuming only Python3 as the major, my idea suggestion crumbles like a house of cards with a hypothetical Python4 release in general, because then you cannot enumerate the versions without scraping Python's release page for the minor releases for each major to produce a correct version range, i.e. not ending up with something like Python 3.40 if hypothetically Python 4 is released and compatibility isn't such a problem like it were with Python2-3 (unlikely).
Therefore, my suggestion to this operator handling is to make a further compromise and **only assume** the `>=, <=` operators because we're in full control over our `pyproject.toml` and we only ever add/remove Python versions when we know our project would work with them, IOW we wouldn't skip a version of Python, that makes no sense from our POV. This alone will fix the problem I mentioned in the first sentence.

```
...
min_version = None
max_version = None
...
version = Version(spec.version)
...
return [f"3.{minor}" for minor in range(min_version.minor, max_version.minor + 1)]
```
should do the trick

noxfile.py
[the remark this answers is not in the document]
a-ovchinnikov: The same holds true for UTs: if py3.9 fails then there is no need to run other targets.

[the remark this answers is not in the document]
eskultety: UTs take most of the execution time, so if you change the order in which targets are executed by stopping there at failing UTs you're deliberately robbing yourself of the opportunity to fix everything in one go without having to re-run the sessions more twice, so it's IMO actually worse that way (I mean run UTs first and skip linters on failure)

noxfile.py
```diff
@@ -0,0 +1,261 @@
...
+        "pyproject.toml",
+        silent=True,
+    )
+
+
+@nox.session(name="pip-compile-podman", reuse_venv=True)
+def pip_compile_podman(session: Session) -> None:
+    """Generate requirements.txt and requirements-extras.txt files inside a container"""
+    PWD = session.env["PWD"]
+    session.run(
```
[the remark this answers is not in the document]
slimreaper35: I have a feeling that we lean towards keeping both `Makefile` and `noxfile.py`, or not ?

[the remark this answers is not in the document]
a-ovchinnikov: I'd say we should definitely get rid of Makefile. Having two tools to do one task is worse than having one tool for one task.

noxfile.py
a-ovchinnikov: Ok, I am not insisting on this, and there is still the need to extract just the coverage.

pyproject.toml
```diff
@@ -102,3 +98,17 @@ skips = [
 log_cli_level = "DEBUG"
 log_format = "%(asctime)s %(levelname)s %(message)s"
 log_date_format = "%Y-%m-%d %H:%M:%S"
+
+[tool.flake8]
+show_source = true
+ignore = [
+  "D100", # missing docstring in public module
```
a-ovchinnikov, line 104: These comments are a great addition! The config becomes much clearer this way.

tox.ini
```diff
@@ -70,10 +70,3 @@ commands =
       {posargs}
 allowlist_externals = rm
 skipsdist = true
-
-[flake8]
-show-source = True
-exclude = venv,.git,.tox,dist,*egg,.env,hack
```
a-ovchinnikov, line 76: `exclude` session is omitted in flake8 section of the toml file above, is it defined elsewhere?

noxfile.py
```diff
@@ -0,0 +1,133 @@
...
+from nox.sessions import Session
+
+# default sessions (sorted alphabetically)
+nox.options.sessions = ["bandit", "black", "flake8", "isort", "mypy", "unit-tests"]
+
+# reuse virtual environment across all sessions
+nox.options.reuse_venv = "always"
+
+# stop other sessions if one session fails
+nox.options.stop_on_first_error = True
```
a-ovchinnikov, line 30: Have we reached a consensus on this one? (It seemed to me that I was the only one in favor of this option).

a-ovchinnikov (commented):
A couple small questions, but LGTM overall.

noxfile.py
slimreaper35: After finishing [this](https://github.com/containerbuildsystem/cachi2/pull/680) I think the execution of all sessions is quite fast now.

tox.ini
slimreaper35: I stated that in the commit message, we run flake8 on specific targets so we don't have to exclude venv etc.

noxfile.py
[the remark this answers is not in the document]
slimreaper35: I think @eskultety has a different opinion.

tox.ini
a-ovchinnikov: Missed that one, sorry!

noxfile.py
eskultety: Like I expressed elsewhere in a thread, I'm not so much in favour of an early failure here given that the bottleneck are always UTs no matter what order this is executed in.

noxfile.py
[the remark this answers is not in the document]
eskultety: > I think @eskultety has a different opinion.

Different opinion in what? I'm all for ditching both the Makefile and tox in favour of this.
Why was the pip-compile target dropped from the latest revision?

noxfile.py
slimreaper35: Could we parse the versions from `classifiers` ?

noxfile.py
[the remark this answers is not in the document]
slimreaper35: OK, maybe I misunderstood something, added `pip-compile` and `build-image` sessions (targets).

noxfile.py
eskultety: I did consider using classifiers in my original thought process, but ruled them out for being more error prone. This made me think again and I think my original worries weren't justified since we're in full control of the file, so I think we could give it a try.

noxfile.py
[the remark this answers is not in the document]
eskultety: You added `build_pristine_image` not `build_image` if you look at the command line:
`podman build --tag localhost/cachi2:latest .`
That said, I argued in a different thread that those build image targets should be IMO dropped completely (in a separate commit though).

noxfile.py
```diff
@@ -0,0 +1,170 @@
...
+    cmd = "isort --check --diff --color cachi2 tests noxfile.py"
+    session.run(*cmd.split(), *session.posargs, silent=True)
+
+
+@nox.session()
+def mypy(session: Session) -> None:
+    """Run mypy on cachi2 and tests directories"""
+    install_requirements(session)
+    cmd = "mypy --install-types --non-interactive cachi2 tests"
+    session.run(*cmd.split(), *session.posargs, silent=True)
```
a-ovchinnikov, line 87: Obligatory [non-blocking] tag.

Would this be too much of generalization?

```python
@nox.session()
def mypy(session: Session) -> None:
    """Run mypy on cachi2 and tests directories"""
    run_cmd_for(session, "mypy --install-types --non-interactive cachi2 tests")

def run_cmd_for(session, cmd, loud=False):
    install_requirements(session)
    session.run(*cmd.split(), *session.posargs, silent=not loud)

**Review by a-ovchinnikov — APPROVED** [review]

File: noxfile.py (hunk 20 above)
Comment (slimreaper35) [reply noxfile.py:87]: I don't know, to me it feels like an unnecessary wrapper around each command.

File: noxfile.py (hunk 11 above)
Comment (slimreaper35) [reply noxfile.py]: > We really need to implement HTML reporting for pytest results (but in a follow-up PR, I guess)

I added a session (target)  `coverage-html`. Is that what you mean ?

**Comment by ben-alkov** [comment]
```
clean:

mock-unittest-data:
	hack/mock-unittest-data/gomod.sh
```

I'm not advocating for keeping these, but they *must* be documented somewhere. I can't even remember the complete list of what "clean" takes care of, much less how to refresh the mock data 😆 ...

**Comment by ben-alkov** [comment]
> > I'm not advocating for keeping these, but they must be documented somewhere. I can't even remember the complete list of what "clean" takes care of, much less how to refresh the mock data 😆 ...
> 
> I've never used `hack/mock-unittest-data/gomod.sh` or `hack/mock-unittest-data/yarn.py`

All the more reason to document them, before the last person who actually did run them gets hit by a bus 😁

File: noxfile.py (hunk 20 above)
Comment (ben-alkov) [reply noxfile.py:86]: Given that all of the commands are exactly the same, maintainability would be increased by

```
def run_cmd(cmd, session):
    session.run(*cmd.split(), *session.posargs, silent=True)
```

called from the individual sessions as `run_cmd(cmd, session)`.

Why? I realize that it seems excessive, but:

- It actually reduces the amount of repetition. If we need to change the  options to `session.run`, we can do it in ONE place instead of ten
- If we do have one specific session which needs to have a slightly different `session.run`, we can call `session.run` directly *from that specific session*, i.e. ignoring our "wrapper". As it stands, if one out the many sessions has different options, it's really hard to see.

File: noxfile.py [hunk 21]
```diff
@@ -0,0 +1,166 @@
...
+        --generate-hashes \
+        --output-file=requirements.txt \
+        --rebuild \
+        pyproject.toml && \
+    pip-compile \
+        --all-extras \
+        --generate-hashes \
+        --output-file=requirements-extras.txt \
+        --rebuild \
+        pyproject.toml
```
eskultety, line 163: Has this actually worked for you? It didn't when I executed it locally, some '&&' stringifying happening that prevented the shell inside the container to actually run the commands properly.

What eventually worked for me though was dumping the commands to a temporary script and mounting that to the container:
```    tempfile = None
    with NamedTemporaryFile(mode="w", prefix="cachi2_pipcompile", suffix=".sh", delete=False) as tf:
        tempfile = tf.name
        deps_cmds = """
        #!/bin/sh
        # git must be installed in the image due to setuptools-scm that has it as a direct dependency
        apk add --no-cache git
        pip3 install pip-tools
        """

        pip_compile_cmds = """
        pip-compile \
            --generate-hashes \
            --output-file=requirements.txt \
            --rebuild \
            pyproject.toml
        pip-compile \
            --all-extras \
            --generate-hashes \
            --output-file=requirements-extras.txt \
            --rebuild \
            pyproject.toml
        """
        tf.write(deps_cmds + pip_compile_cmds)

    target = Path(tempfile).name
    cmd = f"podman run --rm --volume {PWD}:/cachi2:rw,Z --volume {tf.name}:/{target}:ro,Z --workdir /cachi2 docker.io/library/python:3.9-alpine"
    session.run(*cmd.split(), *["sh", f"/{target}"], external=True)
    os.unlink(tempfile)
```

I didn't bother with `dedent`, but it could be incorporated to make the temp script nicer for debugging purposes. The point being, something along ^those lines should be done instead.

noxfile.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    cmd = f"podman run --rm --volume {PWD}:/cachi2:rw,Z --workdir /cachi2 docker.io/library/python:3.9-alpine sh -c {pip_compile_cmd}"
+    session.run(*cmd.split(), pip_compile_cmd, silent=True, external=True)
```
eskultety, line 166: If not for anything else, running podman from nox using a session runner is exotic enough to justify not silencing the output. The other reason being that you actually want to see what pip-compile is doing.

eskultety (commented):
What I'm missing in this PR:
- adding `.nox` to `.gitignore`
- making the final switch to nox from tox/Makefile, i.e. dropping latter
- docs changes

pyproject.toml
```diff
@@ -82,6 +82,14 @@ disallow_untyped_decorators = true
...
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+module = "nox.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "noxfile"
+disallow_untyped_decorators = false
+
```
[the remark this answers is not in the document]
slimreaper35: I believe `nox.*` notation is a wildcard pattern that matches all modules and submodules under the nox namespace and `noxfile` matches only `noxfile.py`. I haven't found a way how to merge both yet :shrug:

noxfile.py
```diff
@@ -0,0 +1,166 @@
...
+
+import os
+from pathlib import Path
+
+import nox
+import tomli
+from nox.sessions import Session
+
+# default sessions (sorted alphabetically)
+nox.options.sessions = ["bandit", "black", "flake8", "isort", "mypy", "unit-tests-python"]
```
slimreaper35, line 25: Do you agree with `unit-tests-python-3.x` or do you prefer a shorter name ?

Makefile
```diff
@@ -5,12 +5,9 @@ GENERATE_TEST_DATA = false
 TEST_LOCAL_PYPISERVER = false
 TEST_LOCAL_DNF_SERVER = false
 
-.PHONY: clean pip-compile
```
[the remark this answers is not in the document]
slimreaper35: Should be `venv` target be part of that commit as well?

noxfile.py
a-ovchinnikov: I would have shortened this to `utests` for the sake of those who would want to invoke just this section (less typing to do), but that's not binding.

ben-alkov: I'd say `.*python.*` is redundant in the context of unit tests

.github/workflows/gating.yaml
```diff
@@ -25,18 +25,20 @@ jobs:
         # We need to install git inside the container otherwise the checkout action will use Git
         # REST API and the .git directory won't be present which fails due to setuptools-scm
         apt-get update && apt-get install --no-install-recommends --no-install-suggests -y git
-        python3 -m pip install --upgrade pip
-        pip install tox
+        python3 -m venv /var/tmp/venv
+        /var/tmp/venv/bin/pip3 install --upgrade pip
+        /var/tmp/venv/bin/pip3 install nox
+        /var/tmp/venv/bin/pip3 install tomli
```
eskultety, line 31: why are you installing it in a venv in a Container, is it because Debian complains about installing to system-wide locations?

eskultety:
> This LGTM, it only needs the changes that Erik [pointed out](https://github.com/containerbuildsystem/cachi2/pull/731#pullrequestreview-2469309757).
> 
> With these changes, the only reason to create a venv outside of `.nox` is to run the `cachi2` command itself? I'm wondering if we need to keep the `make venv` command at all. It seems to me that the best approach is to install `nox` as a system-wide package, so we can probably move the creation of the cachi2 venv to nox.

How about just documenting it? ;) Having a shortcut command for `python3 -m venv venv; venv/bin/pip install -r requirements.txt -r requirements-extras.txt` seems suspicious, I was never a fan of that, but went with the flow since we had the Makefile and some Makefile targets depended on having a venv, otherwise I'd have deleted it already a while ago, but now we have a chance to get rid of any cruft we don't need and given nox/tox default to creating venvs for all jobs I say it's time to say goodbye to holding hands when it comes to venv creation, it's a standard mechanism anyway :) .

.github/workflows/gating.yaml
eskultety: Have you considered installing to `/usr/local` instead? I'm asking because installing to a venv in a throwaway CI container feels like rubbing the wrong way, but more importantly, I don't know how these containers are set up, IOW `/tmp` is more often than not a `tmpfs` with a fairly low size, you have `/var/tmp` which I think should be fine, but I think with these CI infras using a system-wide location for package installs is a safer bet. Not a blocker though, it's just that this caught my eye.

eskultety (approved):

.github/workflows/gating.yaml
slimreaper35: fixed

noxfile.py
slimreaper35: this is pretty much the last thread that needs to be resolved

brunoapimentel: I think this is nitpick territory. Let's resolve it and merge the PR.

ben-alkov (approved):

PR #733, opened 2024-11-14
Making bundler respect hermetic requirement

a-ovchinnikov:
It turned out that due to a yet unknown change somewhere
in the ecosystem bundler started ignoring instructions
to not access the web during a build when a git dependency
was present.

This change adds a workaround that explicitly forbids
consulting remotes for branches and references and also
pins a path to local repository in a config.

### DIFF
--- a/cachi2/core/package_managers/bundler/main.py
+++ b/cachi2/core/package_managers/bundler/main.py
@@ -13,6 +13,7 @@ from cachi2.core.models.property_semantics import PropertySet
 from cachi2.core.models.sbom import Component
 from cachi2.core.package_managers.bundler.parser import (
     GemPlatformSpecificDependency,
+    GitDependency,
     ParseResult,
     PathDependency,
     parse_lockfile,
@@ -32,17 +33,20 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
         _prepare_environment_variables_for_hermetic_build()
     )
     project_files: list[ProjectFile] = []
+    git_paths = []
 
     for package in request.bundler_packages:
         path_within_root = request.source_dir.join_within_root(package.path)
-        components.extend(
-            _resolve_bundler_package(
-                package_dir=path_within_root,
-                output_dir=request.output_dir,
-                allow_binary=package.allow_binary,
-            )
+        _comps, _git_paths = _resolve_bundler_package(
+            package_dir=path_within_root,
+            output_dir=request.output_dir,
+            allow_binary=package.allow_binary,
         )
-        project_files.append(_prepare_for_hermetic_build(request.source_dir, request.output_dir))
+        components.extend(_comps)
+        git_paths.extend(_git_paths)
+    project_files.append(
+        _prepare_for_hermetic_build(request.source_dir, request.output_dir, git_paths)
+    )
 
     return RequestOutput.from_obj_list(
         components=components,
@@ -51,11 +55,17 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
     )
 
 
+# Aliases for git dependency name and git dependency name as
+# it is written to file system:
+DepName = str
+FSDepName = str
+
+
 def _resolve_bundler_package(
     package_dir: RootedPath,
     output_dir: RootedPath,
     allow_binary: bool = False,
-) -> list[Component]:
+) -> tuple[list[Component], list[tuple[DepName, FSDepName]]]:
     """Process a request for a single bundler package."""
     deps_dir = output_dir.join_within_root("deps", "bundler")
     deps_dir.path.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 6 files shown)


cachi2/core/package_managers/bundler/main.py
```diff
@@ -162,6 +167,27 @@ def _prepare_for_hermetic_build(source_dir: RootedPath, output_dir: RootedPath)
         BUNDLE_VERSION: "system"
     """
     )
+    # Note: if a package depends on a git revision then
+    # BUNDLE_DISABLE_LOCAL_BRANCH_CHECK is necessary for a hermetic build
```
eskultety, line 171: I don't think this is sufficient, I think we need `BUNDLE_DISABLE_LOCAL_REVISION_CHECK` as well. IIUC these checks are tied to whether the git dependency is going to be referenced with `ref:` AND/OR `branch:` in the Gemfile (which then somehow translates to Gemfile.lock) .

tests/integration/test_data/bundler_everything_present_except_gemspec/.build-config.yaml
```diff
@@ -9,3 +9,5 @@ project_files:
     BUNDLE_DEPLOYMENT: "true"
     BUNDLE_NO_PRUNE: "true"
     BUNDLE_VERSION: "system"
+    BUNDLE_DISABLE_LOCAL_BRANCH_CHECK: "true"
+    BUNDLE_LOCAL.JSON___SCHEMA: "${output_dir}/deps/bundler/json-schema-26487618a684"
```
eskultety, line 13: Uhm, would it be possible for us to make the ref segment of the file name a placeholder too? I guess not, anyway, not a blocker it's just that whenever the test repo is updated it'll break the ITs again requiring and explicit update of the ref. Also, the convenience test data generator would not be able to cope with it. It is what it is....

eskultety (commented):
This approach seems to be the only one that makes bundler stop accessing the internet with git dependencies, regardless of any extra env variables used sadly.

While at it, I think this effort should also contain the following changes:

- enforce `BUNDLE_ALLOW_OFFLINE_INSTALL` - didn't do anything for git variables, but seems very relevant to our use case in general
- enforce `BUNDLE_DISABLE_VERSION_CHECK` - I've encountered different bundler behaviour with different versions and older versions were particularly stubborn trying to check the latest version of bundler online, sometimes it was a hard failure, sometimes bundler ignored it and proceeded further where it failed on the build
- enforce `BUNDLE_DISABLE_LOCAL_REVISION_CHECK` - it has a similar function to your `BUNDLE_DISABLE_LOCAL_BRANCH_CHECK`
- stop enforcing `BUNDLE_DEPLOYMENT` - this is a user's choice for how they wish to distributed their built app which we don't control and should not try doing; it's essentially a way of vendoring one's app which is not our business. Moreover, it doesn't seem to have any effect on the offline install per se

eskultety:
FWIW I just opened https://github.com/rubygems/rubygems/issues/8265 to request the `--local` installation flag to be exposed as a configuration setting, mentioning it here just for completeness, I will follow up with the docs update.

cachi2/core/package_managers/bundler/main.py
```diff
@@ -13,6 +13,7 @@
 from cachi2.core.models.sbom import Component
 from cachi2.core.package_managers.bundler.parser import (
     GemPlatformSpecificDependency,
+    GitDependency,
```
eskultety, line 16: nitpick: Do we really want to start the commit subject with "bug" :) ? It looks like you're introducing one, not fixing it.

eskultety (approved):
One commit msg nitpick and I still believe commits 2 and 3 should be swapped. With that addressed -> ACK.

cachi2/core/package_managers/bundler/main.py
a-ovchinnikov: I might be. Any piece of code could contain a bug or enable a previously hidden bug. Starting every commit message with `bug:` would make us explicitly acknowledge this.

eskultety: > I might be. Any piece of code could contain a bug or enable a previously hidden bug. Starting every commit message with `bug:` would make us explicitly acknowledge this.

Amen! :D

cachi2/core/package_managers/bundler/main.py
```diff
@@ -55,7 +59,7 @@ def _resolve_bundler_package(
     package_dir: RootedPath,
     output_dir: RootedPath,
     allow_binary: bool = False,
-) -> list[Component]:
+) -> tuple[list[Component], list[tuple]]:
```
slimreaper35, line 62:
```suggestion
) -> tuple[list[Component], list[tuple[str, str]]]:
```

slimreaper35 (approved):

docs/bundler.md
```diff
@@ -71,6 +71,9 @@ works correctly:
 - BUNDLE_DEPLOYMENT: "true"
 - BUNDLE_NO_PRUNE: "true"
 - BUNDLE_VERSION: "system"
+- BUNDLE_ALLOW_OFFLINE_INSTALL: "true"
```
eskultety, line 74: I just noticed this needs to be updated in the design as well: https://github.com/containerbuildsystem/cachi2/blob/main/docs/design/bundler.md#other-configuration-that-was-considered

> BUNDLE_ALLOW_OFFLINE_INSTALL is not working either with bundle install for some reason, which could be probably the most logical solution in this case.

I don't know what the version of Bundler was back then, but this variable doesn't seem to be causing any issues with default Bundler on either Ruby 3.3 or 3.2, so I'd just drop that hunk and add the variable to the list.

eskultety (changes requested):
@a-ovchinnikov Sorry for a late notice, I know you have 2 ACKs, but I just noticed something in the design doc which we should IMO fix here, it's a trivial change, see my comment.

eskultety:
> @a-ovchinnikov Sorry for a late notice, I know you have 2 ACKs, but I just noticed something in the design doc which we should IMO fix here, it's a trivial change, see my comment.

Disregard, it's better I'll do that in my docs patches which I'll rebase onto this work, so re-approving, sorry for the noise.

eskultety (approved):

PR #739, opened 2024-11-19
yarn: SBOM components

slimreaper35:

### DIFF
--- a/cachi2/core/package_managers/yarn_classic/resolver.py
+++ b/cachi2/core/package_managers/yarn_classic/resolver.py
@@ -1,9 +1,8 @@
 import re
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from itertools import chain
 from pathlib import Path
-from typing import Iterable, Optional, Union
+from typing import Optional, Union
 from urllib.parse import urlparse
 
 from packageurl import PackageURL
@@ -338,14 +337,15 @@ def _get_workspace_packages(
     ]
 
 
-def resolve_packages(project: Project) -> Iterable[YarnClassicPackage]:
+def resolve_packages(project: Project) -> list[YarnClassicPackage]:
     """Return a list of Packages corresponding to all project dependencies."""
     workspaces = extract_workspace_metadata(project.source_dir)
     yarn_lock = YarnLock.from_file(project.source_dir.join_within_root("yarn.lock"))
     runtime_deps = find_runtime_deps(project.package_json, yarn_lock, workspaces)
 
-    return chain(
-        [_get_main_package(project.source_dir, project.package_json)],
-        _get_workspace_packages(project.source_dir, workspaces),
-        _get_packages_from_lockfile(project.source_dir, yarn_lock, runtime_deps),
-    )
+    result: list[YarnClassicPackage] = []
+
+    result.append(_get_main_package(project.source_dir, project.package_json))
+    result.extend(_get_workspace_packages(project.source_dir, workspaces))
+    result.extend(_get_packages_from_lockfile(project.source_dir, yarn_lock, runtime_deps))
+    return result
--- a/cachi2/core/package_managers/yarn_classic/resolver.py
+++ b/cachi2/core/package_managers/yarn_classic/resolver.py
@@ -1,8 +1,10 @@
+import json
 import re
+import tarfile
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Optional, Union
+from typing import Any, Iterable, Optional, Union
 from urllib.parse import urlparse
 
 from packageurl import PackageURL
@@ -12,7 +14,11 @@ from cachi2.core.checksum import ChecksumInfo
 from cachi2.core.errors import PackageRejected, UnexpectedFormat
 from cachi2.core.package_managers.npm import NPM_REGISTRY_CNAMES
 from cachi2.core.package_managers.yarn_classic.project import PackageJson, Project, YarnLock
-from cachi2.core.package_managers.yarn_classic.utils import find_runtime_deps
+from cachi2.core.package_managers.yarn_classic.utils import (
+    find_runtime_deps,
+    get_git_tarball_mirror_name,
+    get_tarball_mirror_name,
+)
 from cachi2.core.package_managers.yarn_classic.workspaces import (
     Workspace,
     extract_workspace_metadata,
@@ -181,8 +187,11 @@ YarnClassicPackage = Union[
 
 
 class _YarnClassicPackageFactory:
-    def __init__(self, source_dir: RootedPath, runtime_deps: set[str]) -> None:
+    def __init__(
+        self, source_dir: RootedPath, mirror_dir: RootedPath, runtime_deps: set[str]
+    ) -> None:
         self._source_dir = source_dir
+        self._mirror_dir = mirror_dir
         self._runtime_deps = runtime_deps
 
     def create_package_from_pyarn_package(self, package: PYarnPackage) -> YarnClassicPackage:
@@ -200,6 +209,7 @@ class _YarnClassicPackageFactory:
         dev = package_id not in self._runtime_deps
 
         if _is_from_npm_registry(package.url):
+            # registry packages should already have the correct name
             return RegistryPackage(
                 name=package.name,
                 version=package.version,
@@ -214,29 +224,47 @@ class _YarnClassicPackageFactory:
             path = self._source_dir.join_within_root(package.path)
             # File packages have a url, whereas link packages do not
             if package.url:
+                real_name = _read_name_from_tarball(path)
+
                 return FilePackage(
-                    name=package.name,
+                    name=real_name,
                     version=package.version,
                     path=path,
                     integrity=package.checksum,
                     dev=dev,
                 )
+
+            package_json_path = path.join_within_root("package.json")
+            # package.json is not required for link packages
+            if package_json_path.path.exists():
+                real_name = _read_name_from_package_json(package_json_path)
+            else:
+                real_name = package.name
+
             return LinkPackage(
-                name=package.name,
+                name=real_name,
                 version=package.version,
                 path=path,
                 dev=dev,
             )
         elif _is_git_url(package.url):
+            tarball_name = get_git_tarball_mirror_name(package.url)
+            tarball_path = self._mirror_dir.join_within_root(tarball_name)
+            real_name = _read_name_from_tarball(tarball_path)
+
             return GitPackage(
-                name=package.name,
+                name=real_name,
                 version=package.version,
                 url=package.url,
                 dev=dev,
             )
         elif _is_tarball_url(package.url):
+            tarball_name = get_tarball_mirror_name(package.url)
+            tarball_path = self._mirror_dir.join_within_root(tarball_name)
+            real_name = _read_name_from_tarball(tarball_path)
+
             return UrlPackage(
-                name=package.name,
+                name=real_name,
                 version=package.version,
                 url=package.url,
                 integrity=package.checksum,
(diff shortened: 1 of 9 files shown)


a-ovchinnikov (commented):
LGTM with a couple small questions/nitpicks. Good to go once those are resolved.

cachi2/core/package_managers/yarn_classic/resolver.py
```diff
@@ -47,26 +53,99 @@ class _RelpathMixin(BaseModel):
...
 class UrlPackage(_BasePackage, _UrlMixin):
     """A Yarn 1.x package from a http/https URL."""
 
+    @property
+    def purl(self) -> str:
+        """Return package URL."""
+        qualifiers = {"download_url": self.url}
+        return PackageURL(
+            type="npm",
+            name=self.name,
```
taylormadore, line 102: For some of the non-registry packages, we may want to read their respective package.json files from the cached tarballs to get their true names. As a user, I can call these non-registry dependencies whatever I want in my project's package.json and it could be completely different from what it really is:
```
"dependencies": {
  "not-fecha": "https://github.com/taylorhakes/fecha.git"
}
```

cachi2/core/package_managers/yarn_classic/resolver.py
```diff
@@ -47,26 +53,99 @@ class _RelpathMixin(BaseModel):
 class RegistryPackage(_BasePackage, _UrlMixin):
     """A Yarn 1.x package from the registry."""
 
+    @property
+    def purl(self) -> str:
+        """Return package URL."""
+        qualifiers = {}
+
+        if self.url != DEFAULT_NPM_REGISTRY:
```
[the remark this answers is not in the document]
slimreaper35: We do have a check for that:
https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/yarn_classic/resolver.py#L184

IOW, if a package is `RegistryPackage`  it is one of those npm registries.

cachi2/core/package_managers/yarn_classic/resolver.py
slimreaper35: Does the filename in the cache come from package.json or the URL ?

cachi2/core/package_managers/yarn_classic/resolver.py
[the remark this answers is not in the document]
a-ovchinnikov: Thank you for pointing me to this. Could you please add this explicitly? Essentially like this:
```python
class RegistryPackage(_BasePackage, _UrlMixin):
    """A Yarn 1.x package from the registry."""
    url: Literal[*NPM_REGISTRY_CNAMES]
```
It will significantly reduce the number of such questions in the future.

cachi2/core/package_managers/yarn_classic/resolver.py
taylormadore: I believe it is the filename from the resolved URL

cachi2/core/package_managers/yarn_classic/resolver.py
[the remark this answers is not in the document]
slimreaper35: Should the URL be like `"https://registry.yarnpkg.com/foo/-/foo-1.0.0.tgz#fffffff"` or just `https://registry.yarnpkg.com` ? :thinking: cc @taylormadore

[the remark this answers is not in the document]
slimreaper35: Currently, we store `https://registry.yarnpkg.com/foo/-/foo-1.0.0.tgz#fffffff` as the `url` attribute so I would need to change the implementation slightly.

cachi2/core/package_managers/yarn_classic/resolver.py
slimreaper35: That's a tricky situation. Theoritacily, we could have three potential names:

- the name declared in the project's package.json
- the actual package name from the dependency's package.json
- the repository name from the Git URL

eskultety (commented):
Straight forwards, but some details need a bit more polishing.

cachi2/core/package_managers/yarn_classic/resolver.py
slimreaper35: Is this only relevant for `GitPackage` and `UrlPackage` ?

brunoapimentel: We actually have [precedent](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/yarn/resolver.py#L356) for using the name present in the `package.json` file in the yarn v3 implementation. 

We don't support git deps in the Yarn v3 backend, but I think we should use the same approach for both file and git deps.

cachi2/core/package_managers/yarn_classic/resolver.py
```diff
@@ -30,14 +34,20 @@
...
     name: str
     version: Optional[str] = None
     integrity: Optional[str] = None
     dev: bool = False
 
+    @property
+    @abstractmethod
+    def purl(self) -> str:
+        """Return the package URL."""
+        pass
```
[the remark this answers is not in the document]
slimreaper35: This removal increased our coverage :smile:

cachi2/core/package_managers/yarn_classic/resolver.py
```diff
@@ -329,6 +353,36 @@ def resolve_packages(project: Project) -> Iterable[YarnClassicPackage]:
...
+            _get_packages_from_lockfile(project.source_dir, mirror_dir, yarn_lock, runtime_deps),
         )
     )
+
+
+def _read_name_from_tarball(tarball_path: RootedPath) -> str:
+    """Read the package name from the package.json file in the cached tarball."""
+    with tarfile.open(tarball_path) as tar:
+        names = tar.getnames()
+        package_json_subpath = next((n for n in names if n.endswith("package.json")), None)
```
a-ovchinnikov, line 365: A quick experiment tells me that picking the first name that ends with `packages.json` will most likely work. I wonder if this will always be the case? I.e. is it guaranteed that top-level `packages.json` will always come first even if there are nested packages? (This is not a blocker, more of a general question about the tool and I believe it is reasonably safe to just pick the first match)

cachi2/core/package_managers/yarn_classic/resolver.py
```diff
@@ -329,6 +353,36 @@ def resolve_packages(project: Project) -> Iterable[YarnClassicPackage]:
...
+        if package_json_subpath is None:
+            raise ValueError(f"No package.json found in the tarball {tarball_path}")
+
+        package_json = tar.extractfile(package_json_subpath)
+        if package_json is None:
+            raise ValueError(f"Failed to extract package.json from {tarball_path}")
+
+        package_json_content: dict[str, Any] = json.loads(package_json.read())
```
a-ovchinnikov, line 373: Non-blocking observation: L373:L378 look suspiciously like an inlined call to `_read_name_from_package_json`.

cachi2/core/package_managers/yarn_classic/resolver.py
```diff
@@ -329,6 +353,36 @@ def resolve_packages(project: Project) -> Iterable[YarnClassicPackage]:
...
+            raise ValueError(f"No 'name' field found in package.json in {tarball_path}")
+
+        return name
+
+
+def _read_name_from_package_json(path: RootedPath) -> str:
+    """Read the package name from a package.json file."""
+    package_json = PackageJson.from_file(path)
+
+    if (name := package_json.data.get("name")) is None:
```
a-ovchinnikov, line 385: I wonder if this has to just move to `PackageJson` after- validator. Out of scope for this change though.

a-ovchinnikov (approved):
LGTM with a couple of small non-blocking questions and observations.

cachi2/core/package_managers/yarn_classic/resolver.py
```diff
@@ -325,8 +325,10 @@ def resolve_packages(project: Project) -> Iterable[YarnClassicPackage]:
     yarn_lock = YarnLock.from_file(project.source_dir.join_within_root("yarn.lock"))
     runtime_deps = find_runtime_deps(project.package_json, yarn_lock, workspaces)
 
-    return chain(
-        [_get_main_package(project.package_json)],
-        _get_workspace_packages(project.source_dir, workspaces),
-        _get_packages_from_lockfile(project.source_dir, yarn_lock, runtime_deps),
+    return list(
+        chain(
```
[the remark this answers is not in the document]
slimreaper35: mypy seems to complain about this, I found this could work:
```
    result: list[YarnClassicPackage] = []

    result.append(_get_main_package(project.package_json))
    result.extend(_get_workspace_packages(project.source_dir, workspaces))
    result.extend(
        _get_packages_from_lockfile(project.source_dir, mirror_dir, yarn_lock, runtime_deps)
    )
    return result
```

cachi2/core/package_managers/yarn_classic/resolver.py
slimreaper35: `package_json` in this function has the type `IO[bytes]`. Is it possible to extract the `Path` object after extracting the file from the tarball?

cachi2/core/package_managers/yarn_classic/resolver.py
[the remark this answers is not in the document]
a-ovchinnikov: `mypy` is at it again. Let's see. In the proposed amendment the first call returns `WorkspacePackage` which is then gets wrapped into a `list`. The second call returns `list[WorkspacePackage]`. The third call returns `list[YarnClassicPackage]`. The function is supposed to return an `Iterable` over `YarnClassicPackage`s. By definition `YarnClassicPackage` is a union containing `WorkspacePackage`, i.e. a wider type which could be constructed from some combination of narrower ones. A list is an interable. Concating a list of narrower types to a list of wider type should result in a list of wider type.

More importantly `list(chain(...))` does exactly the same. I wonder why `mypy` decides that one way of conversion is fine from typing perspective while another one is not. I am personally not a fan of adding extra hints just to pacify mypy.

cachi2/core/package_managers/yarn_classic/resolver.py
brunoapimentel: I'm pretty sure that any tarball/git repo that is used as a dependency needs to have its package.json defined in the root folder.

cachi2/core/package_managers/yarn_classic/resolver.py
```diff
@@ -48,26 +59,98 @@ class _RelpathMixin(BaseModel):
...
+            version=self.version,
+            qualifiers=qualifiers,
+        ).to_string()
+
 
 class FilePackage(_BasePackage, _RelpathMixin):
     """A Yarn 1.x package from a local file path."""
 
+    @property
+    def purl(self) -> str:
```
taylormadore, line 123: Shouldn't this have a  `vcs_url` qualifier, since it lives within the source repo? Same comment for Link and Workspace packages

slimreaper35: I think so, we use `vcs_url` for PATH dependencies for [bundler](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/bundler/parser.py#L184).

slimreaper35: should be fixed

cachi2/core/package_managers/yarn_classic/resolver.py
```diff
@@ -27,47 +37,137 @@
...
+        """Return the package URL."""
+
 
 class _UrlMixin(BaseModel):
     url: str
 
 
 class _RelpathMixin(BaseModel):
+    root: RootedPath
     relpath: Path
```
slimreaper35, line 63: 1. Do you think the mixin's name should be slightly reworded ?
2. And what is the more appropriate type for `root` ? `RootedPath` or just `Path`

taylormadore: Maybe just `_PathMixin`?

I'm wondering if we could just have one RootedPath and then just use the root and subpath_from_root properties of that one RootedPath to get everything we need

taylormadore (approved):

PR #741, opened 2024-11-21
[bundler] Adjust the design docs/user facing docs on the build config settings we need for hermetic builds

eskultety:
This PR tries to capture the essence of https://github.com/containerbuildsystem/cachi2/pull/733, mainly the discussion that happened around it, e.g. https://github.com/containerbuildsystem/cachi2/pull/733#issuecomment-2482819831


slimreaper35:
Do we want to add a section to the docs about using `--local` flag when installing dependencies and the current limitation that it is not available as environment variable ?

docs/design/bundler.md
```diff
@@ -449,6 +453,58 @@ because then it's trying to enforce further checks as outlined in the
...
+
+is that this creates a local `vendor/bundle` directory local to the application repository instead
+of using a system-wide location during the install. What that means in practice is that regardless
+of the Rubygems ecosystem recommendations we're enforcing a particular way of installing
+applications onto our whole user base. If they wish to override it, they need to make changes to
+their build recipes and explicitly by setting the `BUNDLE_PATH` configuration option and point it
+to a system location. That is official the only way to tell bundler to skip creating the local
+`vendor/bundle` directory using the deployment mode.
+
+Trying not to use the deployment mode (without making use of `--local` on the user side) leads to
```
a-ovchinnikov, line 480: Structural nitpick: the rest of this section belongs to a separate section, `Typical problems which arise when trying to enforce offline installation`. I would have also swapped this subsection with `Offline installation using deployment mode`

docs/bundler.md
```diff
@@ -78,8 +78,19 @@ The directory that Bundler will look into when installing gems.
 
 ### BUNDLE_DEPLOYMENT
 
-Disallow changes to the **Gemfile**. When the **Gemfile** is changed
-and the lockfile has not been updated, running Bundler commands will be blocked.
+Disallow changes to the **Gemfile**.
```
a-ovchinnikov, line 81:
```diff
- Disallow changes to the **Gemfile**.
+ Disallow changes to the **Gemfile**. When the **Gemfile** is changed
```
`When the **Gemfile** is changed` went missing.

a-ovchinnikov (commented):
One structural suggestion, one minor correction, LGTM otherwise.

docs/design/bundler.md
eskultety: > I would have also swapped this subsection with Offline installation using deployment mode

What do you mean? Like only use that bit of content and drop the rest?

eskultety:
> Do we want to add a section to the docs about using `--local` flag when installing dependencies

Depends, I'd mention it only as a recommendation to potentially consider at best and only if they experience issues, remember, we don't want our users to need to make adjustments to their standard build recipes just because of cachi2. Sometimes it's inevitable, but in this case it's our problem to take care of. We'll need to improve upon that by handling this also on the CLI, but ideally we don't want anyone to need to update their Dockerfile.
I can add the recommendation to the footnote, how about that @slimreaper35?

> and the current limitation that it is not available as environment variable ?

No, this bit of information should be completely transparent to the user.

docs/bundler.md
a-ovchinnikov: Ok, because with the change as it is now the next sentence starts from the middle: `and the lockfile has not been updated, running Bundler commands will be blocked.`

docs/design/bundler.md
a-ovchinnikov: 1. Add a new header `Typical problems which arise when trying to enforce offline installation` on L479;
2. Cut L479-L506 and paste after L455 (i.e. right before `Offline installation...`);
3. Keep everything else.

The contents itself is fine, it is just about ordering of subsections. Also the suggested subsection name is verbose and clunky, something more succinct would be much better, I just could not come up with a better one yet. This is not a fundamental blocker, but I think it will help the reader a bit.

slimreaper35:
> Depends, I'd mention it only as a recommendation to potentially consider at best and only if they experience issues, remember, we don't want our users to need to make adjustments to their standard build recipes just because of cachi2. Sometimes it's inevitable, but in this case it's our problem to take care of. We'll need to improve upon that by handling this also on the CLI, but ideally we don't want anyone to need to update their Dockerfile.
I can add the recommendation to the footnote, how about that @slimreaper35?

Ok

eskultety:
> > Depends, I'd mention it only as a recommendation to potentially consider at best and only if they experience issues, remember, we don't want our users to need to make adjustments to their standard build recipes just because of cachi2. Sometimes it's inevitable, but in this case it's our problem to take care of. We'll need to improve upon that by handling this also on the CLI, but ideally we don't want anyone to need to update their Dockerfile.
> > I can add the recommendation to the footnote, how about that @slimreaper35?
> 
> Ok

@slimreaper35 turns out I can't mention this. I forgot one thing - deployment doesn't mix with `--local`, IOW deployment takes precedence, so there really isn't a way for users to enforce local install with deployment, they can only use the `BUNDLE_PATH` configuration key to force the installation directory, I will mention that instead in the footnote.

docs/bundler.md
eskultety: Adjusted, although a bit differently than suggested @a-ovchinnikov. Merely cutting and pasting the suggested content to a location earlier in the document didn't feel (contextually) right to me and so I reworded the paragraph slighly and added a `Offline installation pitfalls` section to host it, let me know if you agree.

docs/design/bundler.md
eskultety: Adjusted, although a bit differently than suggested @a-ovchinnikov. Merely cutting and pasting the suggested content to a location earlier in the document didn't feel (contextually) right to me and so I reworded the paragraph slighly and added a `Offline installation pitfalls` section to host it, let me know if you agree.

docs/bundler.md
eskultety: Derp, sorry, I thought you were referring to the other occurrence in  the design doc where there's only the first sentence (pre-existing) and because I was editing mainly that one I thought I omitted something on purpose. Sorry, for the confusion, this is now fixed!

slimreaper35 (approved):

PR #743, opened 2024-11-21
yarn: Automated Yarn version detector

a-ovchinnikov:
In order to simplify user interface some logic is added to automatically distinguish between supported Yarn versions.

### DIFF
--- a/cachi2/core/config.py
+++ b/cachi2/core/config.py
@@ -25,6 +25,8 @@ class Config(BaseModel, extra="forbid"):
     requests_timeout: int = 300
     concurrency_limit: int = 5
 
+    allow_yarnberry_processing: bool = True
+
     @model_validator(mode="before")
     @classmethod
     def _print_deprecation_warning(cls, data: Any) -> Any:
--- a/cachi2/core/package_managers/metayarn.py
+++ b/cachi2/core/package_managers/metayarn.py
@@ -1,0 +1,28 @@
+from cachi2.core.config import get_config
+from cachi2.core.models.input import Request
+from cachi2.core.models.output import RequestOutput
+from cachi2.core.package_managers.yarn.main import fetch_yarn_source as fetch_yarnberry_source
+from cachi2.core.package_managers.yarn_classic.main import MissingLockfile, NotV1Lockfile
+from cachi2.core.package_managers.yarn_classic.main import (
+    fetch_yarn_source as fetch_yarn_classic_source,
+)
+from cachi2.core.utils import merge_outputs
+
+
+def fetch_yarn_source(request: Request) -> RequestOutput:
+    """Fetch yarn source."""
+    # Packages could be a mixture of yarn v1 and v2 (at least this is how it
+    # looks now). To preserve this behavior each request is split into individual
+    # packages which are assessed one by one.
+    fetched_packages = []
+    for package in request.packages:
+        new_request = request.model_copy(update={"packages": [package]})
+        try:
+            fetched_packages.append(fetch_yarn_classic_source(new_request))
+        except (MissingLockfile, NotV1Lockfile) as e:
+            # It is assumed that if a package is not v1 then it is probably v2.
+            if get_config().allow_yarnberry_processing:
+                fetched_packages.append(fetch_yarnberry_source(new_request))
+            else:
+                raise e
+    return merge_outputs(fetched_packages)
--- a/cachi2/core/package_managers/yarn_classic/main.py
+++ b/cachi2/core/package_managers/yarn_classic/main.py
@@ -2,7 +2,7 @@ import logging
 import re
 from collections import Counter
 from pathlib import Path
-from typing import Iterable
+from typing import Any, Iterable
 from urllib.parse import urlparse
 
 from cachi2.core.errors import PackageManagerError, PackageRejected
@@ -27,6 +27,26 @@ log = logging.getLogger(__name__)
 
 
 MIRROR_DIR = "deps/yarn-classic"
+_yarn_classic_pattern = "yarn lockfile v1"  # See [yarn_classic_trait].
+
+
+class MissingLockfile(PackageRejected):
+    """Indicate that a lock file is missing."""
+
+    def __init__(self) -> None:
+        """Initialize a Missing Lockfile error."""
+        reason = "Yarn lockfile 'yarn.lock' missing, refusing to continue"
+        solution = "Make sure your repository has a Yarn lockfile (i.e. yarn.lock) checked in"
+        super().__init__(reason, solution=solution)
+
+
+class NotV1Lockfile(PackageRejected):
+    """Indicate that a lockfile is of wrong tyrpoe."""
+
+    def __init__(self, package_path: Any) -> None:
+        """Initialize a Missing Lockfile error."""
+        reason = f"{package_path} not a Yarn v1"
+        super().__init__(reason, solution=None)
 
 
 def fetch_yarn_source(request: Request) -> RequestOutput:
@@ -36,7 +56,7 @@ def fetch_yarn_source(request: Request) -> RequestOutput:
     def _ensure_mirror_dir_exists(output_dir: RootedPath) -> None:
         output_dir.join_within_root(MIRROR_DIR).path.mkdir(parents=True, exist_ok=True)
 
-    for package in request.yarn_classic_packages:
+    for package in request.yarn_packages:
         package_path = request.source_dir.join_within_root(package.path)
         _ensure_mirror_dir_exists(request.output_dir)
         _resolve_yarn_project(Project.from_source_dir(package_path), request.output_dir)
@@ -121,9 +141,27 @@ def _reject_if_pnp_install(project: Project) -> None:
         )
 
 
+def _get_path_to_yarn_lock(project: Project) -> Path:
+    return project.source_dir.join_within_root("yarn.lock").path
+
+
+def _reject_if_wrong_lockfile_version(project: Project) -> None:
+    yarnlock_path = _get_path_to_yarn_lock(project)
+    text = yarnlock_path.read_text()
+    if _yarn_classic_pattern not in text:
+        raise NotV1Lockfile(project.source_dir)
+
+
+def _reject_if_lockfile_is_missing(project: Project) -> None:
+    yarnlock_path = _get_path_to_yarn_lock(project)
+    if not yarnlock_path.exists():
+        raise MissingLockfile()
+
+
 def _verify_repository(project: Project) -> None:
+    _reject_if_lockfile_is_missing(project)
+    _reject_if_wrong_lockfile_version(project)
     _reject_if_pnp_install(project)
-    # _check_lockfile(project)
 
 
 def _verify_corepack_yarn_version(source_dir: RootedPath, env: dict[str, str]) -> None:
@@ -194,3 +232,7 @@ def _get_git_tarball_mirror_name(url: str) -> str:
         package_filename = package_filename[1:]
 
     return package_filename
+
+
+# References
+# [yarn_classic_trait]:  https://github.com/yarnpkg/berry/blob/13d5b3041794c33171808fdce635461ff4ab5c4e/packages/yarnpkg-core/sources/Project.ts#L434
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -1,4 +1,3 @@
-from collections.abc import Iterable
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any, Callable
@@ -6,9 +5,9 @@ from typing import Any, Callable
 from cachi2.core.errors import UnsupportedFeature
 from cachi2.core.models.input import PackageManagerType, Request
 from cachi2.core.models.output import RequestOutput
-from cachi2.core.package_managers import bundler, generic, gomod, npm, pip, rpm, yarn, yarn_classic
+from cachi2.core.package_managers import bundler, generic, gomod, metayarn, npm, pip, rpm
 from cachi2.core.rooted_path import RootedPath
-from cachi2.core.utils import copy_directory
+from cachi2.core.utils import copy_directory, merge_outputs
 
 Handler = Callable[[Request], RequestOutput]
 
@@ -17,7 +16,7 @@ _package_managers: dict[PackageManagerType, Handler] = {
     "gomod": gomod.fetch_gomod_source,
     "npm": npm.fetch_npm_source,
     "pip": pip.fetch_pip_source,
-    "yarn": yarn.fetch_yarn_source,
+    "yarn": metayarn.fetch_yarn_source,
     "generic": generic.fetch_generic_source,
 }
 
@@ -25,7 +24,6 @@ _package_managers: dict[PackageManagerType, Handler] = {
 # invoke them via CLI
 _dev_package_managers: dict[PackageManagerType, Handler] = {
     "rpm": rpm.fetch_rpm_source,
-    "yarn-classic": yarn_classic.fetch_yarn_source,
 }
 
 # This is *only* used to provide a list for `cachi2 --version`
@@ -68,26 +66,7 @@ def _resolve_packages(request: Request) -> RequestOutput:
             solution="But the good news is that we're already working on it!",
         )
     pkg_managers = [_supported_package_managers[type_] for type_ in sorted(requested_types)]
-    return _merge_outputs(pkg_manager(request) for pkg_manager in pkg_managers)
-
-
-def _merge_outputs(outputs: Iterable[RequestOutput]) -> RequestOutput:
-    """Merge RequestOutput instances."""
-    components = []
-    env_vars = []
-    project_files = []
-
-    for output in outputs:
-        components.extend(output.components)
-        env_vars.extend(output.build_config.environment_variables)
-        project_files.extend(output.build_config.project_files)
-
-    return RequestOutput.from_obj_list(
-        components=components,
-        environment_variables=env_vars,
-        project_files=project_files,
-        options=output.build_config.options if output.build_config.options else None,
-    )
+    return merge_outputs(pkg_manager(request) for pkg_manager in pkg_managers)
 
 
 def inject_files_post(from_output_dir: Path, for_output_dir: Path, **kwargs: Any) -> None:
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -37,7 +37,7 @@ def resolve_packages(request: Request) -> RequestOutput:
     This function performs the operations in a working copy of the source directory in case
     a package manager that can make unwanted modifications will be used.
     """
-    if request.yarn_packages or request.yarn_classic_packages:
+    if request.yarn_packages:
         original_source_dir = request.source_dir
 
         with TemporaryDirectory(".cachi2-source-copy", dir=".") as temp_dir:
--- a/tests/unit/package_managers/yarn_classic/test_project.py
+++ b/tests/unit/package_managers/yarn_classic/test_project.py
@@ -230,6 +230,15 @@ def test_pnp_installs_false(
         config_file_name,
         config_file_content,
     )
+    # A yarn v1 project needs a valid yarn lock to be accepted.
+    _prepare_config_file(
+        rooted_tmp_path,
+        # Ignoring type checking because of
+        #   incompatible type "type[YarnLock]"; expected "Union[PackageJson, YarnLock]"
+        YarnLock,  # type: ignore
+        "yarn.lock",
+        VALID_YARN_LOCK_FILE,
+    )
 
     project = Project.from_source_dir(rooted_tmp_path)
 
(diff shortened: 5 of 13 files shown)


cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,123 @@
...
+def _separate_packages_by_yarn_version(
+    packages: Iterable[PackageInput],
+    source_dir: RootedPath,
+) -> dict[str, Union[list[PackageInput], tuple[PackageInput, Exception]]]:
+    """Sorts packages to bins depending on which Yarn version was used.
+
+    The output dictionary contains "uncategorized" entry to capture anything
+    that could not be categorized (likely yet-unsupported versions of Yarn).
+
+    The output dictionary also contains categroy "exceptions" to accumulate
```
ben-alkov, line 52:
```diff
-    categroy
+    category
```

ben-alkov:
@a-ovchinnikov; Definitely a "it's not you, it's me" thing, and something which I know there are many valid opinions about; if it were me, I'd order it like this, to make it easier to review:

```python
def _get_path_to_yarn_lock()
def _contains_yarn_version_trait()
_yarnv1_pattern
_yarnv2_pattern
contains_yarn_classic
_yarn_versions
_yarn_processors
def _yarn_selector()
def _separate_packages_by_yarn_version()
def dispatch_to_correct_fetcher()
```

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,123 @@
...
+
+_yarnv1_pattern = re.compile("yarn lockfile v1")  # See [yarn_v1_trait].
+_yarnv2_pattern = re.compile("__metadata:")  # See [yarn_v2_trait] and [yarn_v2_test_repo].
+contains_yarn_classic = partial(_contains_yarn_version_trait, trait_pattern=_yarnv1_pattern)
+
+
+_yarn_versions = {
+    "yarn_classic": contains_yarn_classic,
+    "yarn": partial(_contains_yarn_version_trait, trait_pattern=_yarnv2_pattern),
+}
```
ben-alkov, line 35: Firstly, :taco: for `partial()` - loving it!

Then:
Zen of Python: "Explicit is better than implicit"
English copyediting class (😆): paraphrasing - "Parallelism is desirable" (i.e. things which serve similar purposes should look similar) 

```python
_yarnv1_pattern = re.compile("yarn lockfile v1")  # See [yarn_v1_trait].
_yarnv2_pattern = re.compile("__metadata:")  # See [yarn_v2_trait] and [yarn_v2_test_repo].
contains_yarn_classic = partial(_contains_yarn_version_trait, trait_pattern=_yarnv1_pattern)
contains_yarnberry = partial(_contains_yarn_version_trait, trait_pattern=_yarnv2_pattern)

_yarn_versions = {
    "yarn_classic": contains_yarn_classic,
    "yarn": contains_yarnberry,
}
```

N.B "yarnberry" is to make it painfully obvious, but the names aren't as important as the structure.

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,123 @@
...
+    exceptions that occured during pre-processing of packages.
+    """
+    output = defaultdict(list)
+    for p in packages:
+        category, exception = _yarn_selector(p, source_dir)
+        if exception is None:
+            output[category].append(p)
+        else:  # This is an exceptional result.
+            # Categories would never clash, but mypy does not know that and
+            # does not want to learn.
```
ben-alkov, line 62: Good old mypy, butt of so many jokes 🤣

ben-alkov (approved):
Overall, LGTM with small nitpicks, except for the refactor around `_yarn_versions`: IMO it could be clarified.

cachi2/core/package_managers/metayarn.py
a-ovchinnikov: Done.

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,123 @@
...
+def _yarn_selector(
+    package: PackageInput,
+    source_dir: RootedPath,
+) -> tuple[str, Optional[Exception]]:
+    try:
+        for yarn_version, version_matches_for in _yarn_versions.items():
+            if version_matches_for(package, source_dir):
+                return yarn_version, None
+        else:
+            return "uncategorized", None
```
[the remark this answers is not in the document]
a-ovchinnikov: No, I have intentionally put the final return into `for`'s `else` clause, primarily to avoid confusion about intended indentation level. I guess it dis not work as intended.

cachi2/core/package_managers/metayarn.py
a-ovchinnikov: Good point, thank you, done. I added `-berry` suffix here, but maybe we should add it everywhere then? (I am also squirreling away "parallelism is desirable", it is a great way of referring to the concept)

a-ovchinnikov:
> @a-ovchinnikov; Definitely a "it's not you, it's me" thing, and something which I know there are many valid opinions about; if it were me, I'd order it like this, to make it easier to review:
> 
> ...

Fair enough, done.

cachi2/core/package_managers/metayarn.py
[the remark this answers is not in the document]
ben-alkov: Ohh, it's just me then >_<. I think this is literally my first time seeing a `for...else` construct "in the wild"

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,122 @@
...
+
+    fetched_packages = []
+    for pm, packages in sorted_packages.items():
+        new_request = request.model_copy(update={"packages": packages})
+        fetched_packages.append(_yarn_processors[pm](new_request))
+    # Judging from the resolver code it is safe to merge multiple packages.
+    # Moreover, it does not look like there is any mechanism in place now to
+    # prevent users from requesting both PMs simultaneously.
+    # The code below preserves this behavior.
+    return _merge_outputs(fetched_packages)
```
eskultety, line 116: How do plan on plugging this to the existing `_resolve_packages` helper in the `resolver` module?

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,122 @@
...
+) -> Path:
+    """Construct a path to package's lockfile.
+
+    Raise an exception when there is no lockfile.
+    """
+    yarnlock_path = source_dir.join_within_root("yarn.lock")
+    if yarnlock_path.path.exists():
+        return yarnlock_path.path
+
+    raise PackageRejected(f"Yarn lockfile is missing in {package.path}", solution=None)
```
eskultety, line 30: I understand and you're right. I'd just like to avoid having the same check at multiple places, it just makes debugging a bit harder (definitely not impossible). Like I mentioned in another comment I raise, this meta module should be as simple as possible, and so if "yarn lockfile v1" isn't matched, I'd default to processing everything with yarn v2+ since that has this check and is raising this exception already.

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,122 @@
+import logging
```
eskultety, line 1: Oh, and I think we should stuff this under a new package rather than a top-level module, IOW, make this a `meta/yarn.py` or `_meta/yarn.py`.

cachi2/core/package_managers/metayarn.py
a-ovchinnikov: `_merge_outputs` takes `Iterable[RequestOutput]` and returns `RequestOutput`. Each `fetch_foo_deps` returns a `RequestOutput` so merging any two or more of those in any order should be fine (if I understand your question correctly).

eskultety: Stupid me didn't look at the return type, just the input type. In that case it should be simple.

cachi2/core/package_managers/metayarn.py
eskultety: > if we could be sure that no one would ever mix different Yarn versions. Since mixing them is possible now I opted to preserving the behavior. This resulted in the need to separate potentially mixed dependencies. If this is not what we want then yes, most of this code could be scraped and replaced with the one method above.

True. I still think though that some of the proposed logic would be better hosted in the respective yarn backend modules we have, maybe even paired with more thourough checks we already have in place in those modules when it comes to project files and just let the given parser handle the lockfile - this would actually be a good field test of our yarn backends.

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,122 @@
...
+    trait_pattern: re.Pattern[str],
+) -> Optional[re.Match[str]]:
+    text = _get_path_to_yarn_lock(package, source_dir).read_text()
+    return trait_pattern.search(text)
+
+
+_yarn_classic_pattern = re.compile("yarn lockfile v1")  # See [yarn_classic_trait].
+_yarnberry_pattern = re.compile("__metadata:")  # See [yarnberry_trait] and [yarn_v2_test_repo].
+contains_yarn_classic = partial(_contains_yarn_version_trait, trait_pattern=_yarn_classic_pattern)
+contains_yarnberry = partial(_contains_yarn_version_trait, trait_pattern=_yarnberry_pattern)
```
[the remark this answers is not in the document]
eskultety: Okay, definitely not wrong, but I think it's the parser's job to take care of this, so if we're already not doing that in the respective yarn backend, we need to fix that and not bother with it in the meta handler.

cachi2/core/package_managers/metayarn.py
a-ovchinnikov: Yes, concrete Yarn backends are definitely a better place for that. I have just realized that I can achieve the same results by lifting the check to corresponding PM and splitting packages list into individual packages here before feeding them to PMs instead of separating them by Yarn version and then sending to handlers. I'll push the next change with it (as a separate commit while this is a draft, will squash later).

a-ovchinnikov:
Marking as ready for review to encourage reviews. Will squash the commits once we agree upon the approach.

taylormadore:
The updated approach here seems straightforward :+1: 

IIRC the current v1 parser does check for a v1 lockfile, but that parsing occurs later in the request processing, so I don't have an objection to doing it separately and earlier as proposed here

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,21 @@
...
+    """Fetch yarn source."""
+    # Packages could be a mixture of yarn v1 and v2 (at least this is how it
+    # looks now). To preserve this behavior each reqiest is split into individual
+    # packages which are assessed one by one.
+    fetched_packages = []
+    for package in request.packages:
+        new_request = request.model_copy(update={"packages": [package]})
+        try:
+            fetched_packages.append(yarn_classic.fetch_yarn_source(new_request))
+        except PackageRejected:
```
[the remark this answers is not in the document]
a-ovchinnikov: Good point, thank you. This PR adds two checks: one for lock file presence and another for lock file version. Both raise PackageRejected error to fail early and to signal that a package cannot be processed. This could be problematic if something else raises PackageRejected when processing actual v1 package (for some other reason). I'll subclass PackageRejected with a couple more specific exceptions (one for missing lock file and another for not a v1 lock file) and will catch them here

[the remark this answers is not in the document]
brunoapimentel: Maybe it'd better to have these new errors as part of either the `metayarn.py` or the `yarn_classic/main.py` module. I feel like errors under `core/errors.py` are meant to be globally used in the application, and we should limit the number of variants there.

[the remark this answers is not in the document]
a-ovchinnikov: Good point, thank you. I had doubts about `core/errors.py` being the right home for these exceptions, but placed them there anyway. Will move them in the next push.

taylormadore:
I think we might need to add a config option that would disable the processing of yarn v2+ requests altogether. I'm thinking specifically of cachi2-in-OSBS here, where we probably want to restrict input to yarn v1.

I like the idea of a config option over a flag, since I don't believe we will allow cachi2 config access to users in OSBS.

CC: @MartinBasti

a-ovchinnikov:
> I think we might need to add a config option that would disable the processing of yarn v2+ requests altogether. 

Done!

cachi2/core/config.py
```diff
@@ -25,6 +25,8 @@ class Config(BaseModel, extra="forbid"):
     requests_timeout: int = 300
     concurrency_limit: int = 5
 
+    allow_yarnberry_processing: bool = True
```
slimreaper35, line 28: Why do we need this option ?

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -27,6 +27,26 @@
...
+        super().__init__(reason, solution=solution)
+
+
+class NotV1Lockfile(PackageRejected):
+    """Indicate that a lockfile is of wrong tyrpoe."""
+
+    def __init__(self, package_path: Any) -> None:
+        """Initialize a Missing Lockfile error."""
+        reason = f"{package_path} not a Yarn v1"
+        super().__init__(reason, solution=None)
```
slimreaper35, line 49: Recently, I got a comment from Erik, that all custom cachi2 exceptions should be inside `cachi2.core.errors.py`. But that's just a nitpick.

tests/unit/package_managers/yarn_classic/test_project.py
```diff
@@ -230,6 +230,16 @@ def test_pnp_installs_false(
         config_file_name,
         config_file_content,
     )
+    # A yarn v1 project needs a valid yarn lock to be accepted.
+    _prepare_config_file(
+        rooted_tmp_path,
+        # I will deal with
+        #   incompatible type "type[YarnLock]"; expected "Union[PackageJson, YarnLock]"
+        # later, will ignore mypy for now:
```
slimreaper35: I would suggest to drop personalized comments / notes and create issues instead for example.

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -27,6 +27,26 @@
 
 
 MIRROR_DIR = "deps/yarn-classic"
+_yarn_classic_pattern = re.compile("yarn lockfile v1")  # See [yarn_classic_trait].
```
slimreaper35, line 30: Could we place the link from down below here ?

cachi2/core/resolver.py
```diff
@@ -68,26 +66,7 @@ def _resolve_packages(request: Request) -> RequestOutput:
...
-        env_vars.extend(output.build_config.environment_variables)
-        project_files.extend(output.build_config.project_files)
-
-    return RequestOutput.from_obj_list(
-        components=components,
-        environment_variables=env_vars,
-        project_files=project_files,
-        options=output.build_config.options if output.build_config.options else None,
-    )
+    return merge_outputs(pkg_manager(request) for pkg_manager in pkg_managers)
```
slimreaper35, line 69: Why are we moving this function elsewhere ?

cachi2/core/package_managers/yarn_classic/main.py
eskultety: Correction - I meant any publicly exposed top-level exceptions like `PathOutsideRoot`, `PackageRejected`, etc. ...custom exceptions which are module-scoped are fine to keep in the given module where it makes the most sense for them. That said, these **are** module-scoped exceptions and so we should start their name with an underscore. Additionally, these should be re-raised in the `fetch-yarn-...` meta entrypoint as a `PackageRejected` exception IMO as that's the public-facing one.

cachi2/core/package_managers/yarn_classic/main.py
eskultety: Why do we need regular expressions if the regex pattern is a plain literal? Isn't `"yarn lockfile v1" in ...` more straight forward in this case based on the fact that we read the input file as text in one go?

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,21 @@
...
+from cachi2.core.models.output import RequestOutput
+from cachi2.core.package_managers import yarn, yarn_classic
+from cachi2.core.package_managers.yarn_classic.main import MissingLockfile, NotV1Lockfile
+from cachi2.core.resolver import _merge_outputs
+
+
+def fetch_yarn_source(request: Request) -> RequestOutput:
+    """Fetch yarn source."""
+    # Packages could be a mixture of yarn v1 and v2 (at least this is how it
+    # looks now). To preserve this behavior each reqiest is split into individual
```
eskultety, line 12: s/reqiest/request

cachi2/core/config.py
eskultety: Our config file needs some love, we kinda managed to get rid of some Go stuff with the hope of getting rid of some more. If we want to add a backend-related option, then we should scope this by a settings group (an INI section if you will) before we roll this out, because then it's going to be harder to get rid of the option. The config change doesn't necessarily need come with this PR and can be introduced in a follow-up PR with more config changes/cleanups.

eskultety: Moreover, some/many (definitely not all, that's a rare exhibit out there) of the config options (in general) could have a CLI counterpart to make behaviour configuration of cachi2 more convenient.

cachi2/core/package_managers/yarn_classic/main.py
a-ovchinnikov: These are not module-scoped, but they are also not cachi2-scoped: they are raised here to send a message to meta-PM. They do not need to be re-wrapped because as we have agreed earlier everything that is not v1 is v2+ (and exceptions from v2+ propagate freely). Furthermore, any other exception will propagate to the user, so if something happens to conform to v1 and not trigger any of these, and also is a broken v1 then a user would know.

cachi2/core/package_managers/yarn_classic/main.py
a-ovchinnikov: > Could we place the link from down below here ?

I would rather had it down at the end of file unless there is strong opposition to that. I think having a short reference does not pollute the code as much as a long URL: the line will be read more often than the link will be followed, so I chose to make the line easier to read at the expense of making the link slightly harder to follow (now it is just one select-move/chord/click-and-something-unknown away).

> Why do we need regular expressions if the regex pattern is a plain literal?

No good reason. Will amend in the next revision.

cachi2/core/config.py
a-ovchinnikov: > Why do we need this option ?

I added it basing on https://github.com/containerbuildsystem/cachi2/pull/743#issuecomment-2532884713 ,
the comment also explicitly mentions a config option and not a flag.

> Our config file needs some love
Absolutely! However on L17 I see `goproxy_url`, on L19 -- `gomod_download_max_tries`. The config already looks like a flat space for random variables as it is, do you think it is in scope of this change to start reworking it?

brunoapimentel: > Moreover, some/many (definitely not all, that's a rare exhibit out there) of the config options (in general) could have a CLI counterpart to make behaviour configuration of cachi2 more convenient.

This use case makes me think that we might want to have a set of config options that are only configurable by modifying the config file itself, so that indirect users of Cachi2 as part of a backend system couldn't change them.

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,25 @@
+from cachi2.core.config import get_config
+from cachi2.core.models.input import Request
+from cachi2.core.models.output import RequestOutput
+from cachi2.core.package_managers import yarn, yarn_classic
+from cachi2.core.package_managers.yarn_classic.main import MissingLockfile, NotV1Lockfile
+from cachi2.core.utils import merge_outputs
+
+
+def fetch_yarn_source(request: Request) -> RequestOutput:
```
brunoapimentel, line 9: We need a unit test for this function, as the coverage is very low on this module.

cachi2/core/package_managers/metayarn.py
```diff
@@ -0,0 +1,25 @@
...
+    """Fetch yarn source."""
+    # Packages could be a mixture of yarn v1 and v2 (at least this is how it
+    # looks now). To preserve this behavior each request is split into individual
+    # packages which are assessed one by one.
+    fetched_packages = []
+    for package in request.packages:
```
brunoapimentel, line 15: By implementing a loop to process the input packages here, we are making the same loop that is done in the `fetch_yarn_source` functions for both v1 and v2 redundant.

From one perspective, I'd simplify the `fetch_yarn_source` functions and get rid of the redundancy, since yarn requests will always be initially handled by `metayarn`. On the other hand, keeping them as they are will keep the `yarn` and `yarn_classic` packages independent in case we decide to get rid of `metayarn`, or create another workflow.

Wdyt?

brunoapimentel (approved):

taylormadore:
One detail I missed before is that part of https://github.com/containerbuildsystem/cachi2/commit/78d0eebf18ce151b9e1a9e97a2524d6cba51372d (specifically the part allowing yarn-classic as an input to the CLI) still needs to be reverted

Otherwise LGTM :100:

a-ovchinnikov:
> One detail I missed before is that part of 78d0eeb (specifically the part allowing yarn-classic as an input to the CLI) still needs to be reverted

Done.

taylormadore (approved):

Issue #634, opened 2024-09-05
Yarn 1.x: Ensure there are no collisions in the offline mirror


taylormadore:
When Yarn 1.x fetches a tarball into the offline mirror, the name of the tarball is used as the name of the tarball in the mirror. This potentially opens up the possibility of collisions in the offline mirror, where two different tarballs have the same name and Yarn gets confused about which is being installed.

See: https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/fetchers/tarball-fetcher.js#L65

Practically, it is likely that if integrity keys are being used that Yarn will detect the mismatch and fail, so it is possible this could be remediated by documenting the issue.

Acceptance Criteria
- Offline mirror collisions are detected and an exception is raised
- OR this issue is satisfied via documentation
- Add an integration test for collisions

PR #753, opened 2024-11-28
yarn: Detect offline mirror collisions

slimreaper35:
When yarn fetches a URL dependency into the offline mirror, the name of the tarball is used as the name of the tarball in the mirror.

After running `yarn install` command, in a project where are two different URL dependencies have the same name of the tarball, there will be only one tarball in the offline mirror with such name, because the first one gets replaced by the second one.

closes https://github.com/containerbuildsystem/cachi2/issues/634

### DIFF
--- a/cachi2/core/package_managers/yarn_classic/main.py
+++ b/cachi2/core/package_managers/yarn_classic/main.py
@@ -1,4 +1,9 @@
 import logging
+import re
+from collections import Counter
+from pathlib import Path
+from typing import Iterable
+from urllib.parse import urlparse
 
 from cachi2.core.errors import PackageManagerError, PackageRejected
 from cachi2.core.models.input import Request
@@ -9,7 +14,13 @@ from cachi2.core.package_managers.yarn.utils import (
     run_yarn_cmd,
 )
 from cachi2.core.package_managers.yarn_classic.project import Project
-from cachi2.core.package_managers.yarn_classic.resolver import resolve_packages
+from cachi2.core.package_managers.yarn_classic.resolver import (
+    GitPackage,
+    RegistryPackage,
+    UrlPackage,
+    YarnClassicPackage,
+    resolve_packages,
+)
 from cachi2.core.rooted_path import RootedPath
 
 log = logging.getLogger(__name__)
@@ -43,7 +54,8 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> None:
     prefetch_env = _get_prefetch_environment_variables(output_dir)
     _verify_corepack_yarn_version(project.source_dir, prefetch_env)
     _fetch_dependencies(project.source_dir, prefetch_env)
-    resolve_packages(project)
+    packages = resolve_packages(project)
+    _verify_no_offline_mirror_collisions(packages)
 
 
 def _fetch_dependencies(source_dir: RootedPath, env: dict[str, str]) -> None:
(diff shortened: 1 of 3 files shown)


cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -125,3 +134,18 @@ def _verify_corepack_yarn_version(source_dir: RootedPath, env: dict[str, str]) -
         )
 
     log.info("Processing the request using yarn@%s", installed_yarn_version)
+
+
+def _verify_offline_mirror_collisions(packages: Iterable[YarnClassicPackage]) -> None:
```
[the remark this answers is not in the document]
slimreaper35: I used the word "verify" since we have other similar function `_verify_repository`, `verify_corepack_version` ...

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -125,3 +134,18 @@ def _verify_corepack_yarn_version(source_dir: RootedPath, env: dict[str, str]) -
...
+    """Verify that there are no duplicate tarballs in the offline mirror."""
+    tarballs = []
+    for p in packages:
+        if isinstance(p, UrlPackage):
+            parsed_url = urlparse(p.url)
+            path = Path(parsed_url.path)
+            tarballs.append(path.name)
+
+    c = Counter(tarballs)
+    duplicate_tarballs = [name for name, value in c.most_common() if value > 1]
```
a-ovchinnikov, line 149: Super minor naming nitpick that I missed the last time. In the case of a `Counter` it is not just a `value` it is `count`. Not a blocker.

a-ovchinnikov (approved):

cachi2/core/package_managers/yarn_classic/main.py
slimreaper35: fixed

a-ovchinnikov: Thank you!

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -125,3 +134,18 @@ def _verify_corepack_yarn_version(source_dir: RootedPath, env: dict[str, str]) -
         )
 
     log.info("Processing the request using yarn@%s", installed_yarn_version)
+
+
+def _verify_no_offline_mirror_collisions(packages: Iterable[YarnClassicPackage]) -> None:
+    """Verify that there are no duplicate tarballs in the offline mirror."""
+    tarballs = []
+    for p in packages:
+        if isinstance(p, UrlPackage):
```
taylormadore, line 143: Do we want to handle collisions between other types of Packages as well? For example, I could have a dependency from the registry that results in cached tarball named foo-1.0.0.tgz and have a url dependency with the same name in the path

slimreaper35: I added registry packages to check as well. It should be all I believe or?

taylormadore: IIRC git packages are the only additional package type that is copied into the offline mirror. The others (file, workspace, link) are just present in the source repo.

Does this check also work for scoped packages? For example, in the main integration test, `@colors/colors` is one of the dependencies with a resolved url `https://registry.yarnpkg.com/@colors/colors/-/colors-1.6.0.tgz#ec6cd237440700bc23ca23087f513c75508958b0` and a tarball in the offline mirror `@colors-colors-1.6.0.tgz`

We may need to incorporate some of the logic found here: https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/fetchers/tarball-fetcher.js#L77

slimreaper35: I tried to rewrite the implementation in yarnpkg to Python.

> IIRC git packages are the only additional package type that is copied into the offline mirror. The others (file, workspace, link) are just present in the source repo.

Added git packages too, even though it seems an extremely rare scenario because of hash being added to the name of the repo.

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -43,7 +54,8 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> None:
     prefetch_env = _get_prefetch_environment_variables(output_dir)
     _verify_corepack_yarn_version(project.source_dir, prefetch_env)
     _fetch_dependencies(project.source_dir, prefetch_env)
-    resolve_packages(project)
+    packages = resolve_packages(project)
+    _verify_no_offline_mirror_collisions(packages)
```
a-ovchinnikov, line 58: Non-blocker, non-actionable, just a minor discussion point. I wonder if it would be even easier to follow if we adapted an approach where verifiers are transparent and return the verified object when no issues are detected. I.e. here it could be
```python
packages = _verify_no_offline_mirror_collisions(resolve_packages(project))
```
(pretty much like pydantic does it)
What I don't like about it is the way it reads (alternating direction with start in the middle of statement) and the way it will read for a longer chain of verifiers. Maybe it won't be that bad if we could have had some chaining like this:
```python
packages = resolve_packages_for(project) |> _verify_no_offline_mirror_collisions
```
?

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -125,3 +137,58 @@ def _verify_corepack_yarn_version(source_dir: RootedPath, env: dict[str, str]) -
...
+    """Verify that there are no duplicate tarballs in the offline mirror."""
+    tarballs = []
+    for p in packages:
+        if isinstance(p, (RegistryPackage, UrlPackage)):
+            tarball_name = _get_tarball_mirror_name(p.url)
+            tarballs.append(tarball_name)
+
+        elif isinstance(p, GitPackage):
+            tarball_name = _get_git_tarball_mirror_name(p.url)
+            tarballs.append(tarball_name)
```
a-ovchinnikov, line 152: Non-blocking, optional suggestion
```python
for p in packages:
        if isinstance(p, (RegistryPackage, UrlPackage)):
            tarball_name = _get_tarball_mirror_name(p.url)
        elif isinstance(p, GitPackage):
            tarball_name = _get_git_tarball_mirror_name(p.url)
        else:  # we explicitly don't care about other types; this ensures no one would ever ask again.
            continue
       tarballs.append(tarball_name)
```

cachi2/core/package_managers/yarn_classic/main.py
```diff
@@ -125,3 +137,58 @@ def _verify_corepack_yarn_version(source_dir: RootedPath, env: dict[str, str]) -
...
+
+
+# https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/fetchers/tarball-fetcher.js#L65
+def _get_tarball_mirror_name(url: str) -> str:
+    parsed_url = urlparse(url)
+    path = Path(parsed_url.path)
+
+    match = re.search(RE_URL_NAME_MATCH, str(path))
+
+    if match:
```
a-ovchinnikov, line 171: Non-blocking PEP8 zeal: match could be `None` so it is better to be explicit by checking if it `is` th` `None`.

a-ovchinnikov (approved):
A couple of minor suggestions, but they don't block an approval. LGTM.

cachi2/core/package_managers/yarn_classic/main.py
slimreaper35: The `packages` will be used for generating SBOM components, so maybe I would continue on this point after this PR is merged and the focus moves to SBOM PR.

a-ovchinnikov: Yes, the comment above was more to start a discussion about something that definitely out of scope for this PR.

taylormadore (approved):

cachi2/core/package_managers/yarn_classic/main.py
taylormadore: Nice job, thanks for taking care of this :100:

PR #762, opened 2024-12-04
Adding optional RPM summary to SBOMs

a-ovchinnikov:
This change adds an option to include RPM summary in a SBOM.

### DIFF
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -214,6 +214,7 @@ class RpmPackageInput(_PackageInputBase):
     """Accepted input for a rpm package."""
 
     type: Literal["rpm"]
+    include_summary_in_sbom: bool = False
     options: Optional[ExtraOptions] = None
 
 
--- a/cachi2/core/models/property_semantics.py
+++ b/cachi2/core/models/property_semantics.py
@@ -34,6 +34,7 @@ class PropertySet:
     npm_development: bool = False
     pip_package_binary: bool = False
     bundler_package_binary: bool = False
+    rpm_summary: str = ""
 
     @classmethod
     def from_properties(cls, props: Iterable[Property]) -> "Self":
@@ -44,6 +45,7 @@ class PropertySet:
         npm_development = False
         pip_package_binary = False
         bundler_package_binary = False
+        rpm_summary = ""
 
         for prop in props:
             if prop.name == "cachi2:found_by":
@@ -58,6 +60,8 @@ class PropertySet:
                 pip_package_binary = True
             elif prop.name == "cachi2:bundler:package:binary":
                 bundler_package_binary = True
+            elif prop.name == "cachi2:rpm_summary":
+                rpm_summary = prop.value
             else:
                 assert_never(prop.name)
 
@@ -68,6 +72,7 @@ class PropertySet:
             npm_development,
             pip_package_binary,
             bundler_package_binary,
+            rpm_summary,
         )
 
     def to_properties(self) -> list[Property]:
@@ -87,6 +92,8 @@ class PropertySet:
             props.append(Property(name="cachi2:pip:package:binary", value="true"))
         if self.bundler_package_binary:
             props.append(Property(name="cachi2:bundler:package:binary", value="true"))
+        if self.rpm_summary:
+            props.append(Property(name="cachi2:rpm_summary", value=self.rpm_summary))
 
         return sorted(props, key=lambda p: (p.name, p.value))
 
@@ -100,4 +107,5 @@ class PropertySet:
             npm_development=self.npm_development and other.npm_development,
             pip_package_binary=self.pip_package_binary or other.pip_package_binary,
             bundler_package_binary=self.bundler_package_binary or other.bundler_package_binary,
+            rpm_summary=self.rpm_summary or other.rpm_summary,
         )
--- a/cachi2/core/models/sbom.py
+++ b/cachi2/core/models/sbom.py
@@ -7,6 +7,7 @@ from cachi2.core.models.validators import unique_sorted
 PropertyName = Literal[
     "cachi2:bundler:package:binary",
     "cachi2:found_by",
+    "cachi2:rpm_summary",
     "cachi2:missing_hash:in_file",
     "cachi2:pip:package:binary",
     "cdx:npm:package:bundled",
--- a/cachi2/core/package_managers/rpm/main.py
+++ b/cachi2/core/package_managers/rpm/main.py
@@ -47,6 +47,7 @@ class Package:
     vendor: Optional[str] = None
     checksum: Optional[str] = None
     repository_id: Optional[str] = None
+    summary: Optional[str] = None
 
     @classmethod
     def from_filepath(cls, rpm_filepath: Path, rpm_download_metadata: dict[str, Any]) -> "Package":
@@ -85,6 +86,7 @@ class Package:
             "version=%{VERSION}\n"
             "release=%{RELEASE}\n"
             "arch=%{ARCH}\n"
+            "summary=%{SUMMARY}\n"
             # vendor and epoch are optional RPM tags; return "" if not set instead of "(None)"
             "vendor=%|VENDOR?{%{VENDOR}}:{}|\n"
             "epoch=%|EPOCH?{%{EPOCH}}:{}|\n"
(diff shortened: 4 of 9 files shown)


cachi2/core/models/property_semantics.py
```diff
@@ -87,6 +92,8 @@ def to_properties(self) -> list[Property]:
             props.append(Property(name="cachi2:pip:package:binary", value="true"))
         if self.bundler_package_binary:
             props.append(Property(name="cachi2:bundler:package:binary", value="true"))
+        if self.add_rpm_summary:
+            props.append(Property(name="cachi2:bundler:package:binary", value="true"))
```
taylormadore, line 96: Won't this property be a bool in the output after the component properties are merged? Also, shouldn't the name of this Property be `cachi2:rpm_summary`?

cachi2/core/models/sbom.py
```diff
@@ -7,6 +7,7 @@
 PropertyName = Literal[
     "cachi2:bundler:package:binary",
     "cachi2:found_by",
+    "cachi2:rpm_summary",
```
taylormadore, line 10: Should we use a custom cachi2 property or perhaps the description field for a CycloneDX Component instead?

a-ovchinnikov: Maybe. I'll need to think about it more. A Property looks the least invasive way of doing this.

a-ovchinnikov: FWIW properties can be straightforwardly converted to annotations.

cachi2/core/models/input.py
```diff
@@ -216,6 +216,7 @@ class RpmPackageInput(_PackageInputBase):
     """Accepted input for a rpm package."""
 
     type: Literal["rpm"]
+    add_rpm_summary: bool = False
```
eskultety, line 219: what is the reasoning behind making this configurable? On its own the option doesn't IMO make much sense and I'd like to avoid littering the code with semi-random single-purpose tailored SBOM options in the future by setting this precedent. It's a straightforward change, so why don't we make it the default?

cachi2/core/models/sbom.py
eskultety: > Should we use a custom cachi2 property or perhaps the description field for a CycloneDX Component instead?

Description feels more natural, but then we need to sit down and define what description is expected to represent a particular PM backend's component to remain consistent across all backends. In that case, also for consistency reasons, we should also introduce the same functionality to the rest of the backends in a short timeline span in order not to leave the rest of PMs behind; note that a custom property IMO feels like less of a problem for consistency than making use of an existing CycloneDX field suited directly for this purpose.

cachi2/core/package_managers/rpm/main.py
```diff
@@ -85,6 +86,7 @@ def _query_rpm_fields(file_path: Path) -> dict[str, str]:
             "version=%{VERSION}\n"
             "release=%{RELEASE}\n"
             "arch=%{ARCH}\n"
+            "summary=%|SUMMARY?{%{SUMMARY}}:{}|\n"
```
eskultety, line 89: Sometimes the summary is incredibly useless. Thinking ahead, I think we're more likely to find descriptions rather than summaries provided by other PM backends/services in case we want to be consistent, so considering the pros/cons of using 'DESCRIPTION' here instead could be food for thought.

cachi2/core/models/input.py
a-ovchinnikov: It is either a bloated SBOM or a bloated config. The config might need refactoring and confining individual PM options to dedicated sections. I would say that keeping it flat as long as all options together can still fit on one screen should be fine. Having a flag that someone must explicitly toggle to switch from de facto standard behavior looks like the path of least disturbance to me (i.e. if you are a user of the tool, and have no idea that some flag was added then you should not notice the addition because it did not silently alter any of your existing workflows), so I'd say it is ok for users to ask for more flags for their specific use cases. Summary is one of such use cases. As for granularity -- as long as it is transparent to those who do not want it (i.e. default value of a flag does not alter the tool's behavior) anything could be subject to toggling. I do not know why some might want toggling features off, but can imagine an overly tightly coupled standalone validators as one reason. An alternative to this is postprocessing, JSON is machine-friendly and can be pruned after it is generated, however that could require action from more users than toggling a switch from default position. Essentially parsing out just whatever they want means that users must define a model too which they might want to avoid doing.

cachi2/core/package_managers/rpm/main.py
a-ovchinnikov: Having a summary is a resolution for a specific use-case where a user needs this field in a SBOM for feeding some downstream tooling. It does seem to be rather meaningless at times, but that's was the request. I don't think it needs to be present in a SBOM at all, at least not as long as SBOMs are treated as evidences of artifact contents, but also don't think it is a problem as long as it can be turned off.

cachi2/core/models/input.py
eskultety: > Essentially parsing out just whatever they want means that users must define a model too which they might want to avoid doing.

They need to do it regardless, but fair enough.

cachi2/core/package_managers/rpm/main.py
eskultety: > Having a summary is a resolution for a specific use-case where a user needs this field in a SBOM for feeding some downstream tooling

Exactly and so I've been wondering (tied somewhat to my earlier comment on the flag existence) how we imagine things to continue working going in the future, more specifically, if this is a downstream-only request then we really should keep the upstream code base lean and maintain a downstream fork exactly for this purpose which used to be the common practice everywhere else apart from the Openshift/k8s world. IOW we should evaluate every such request and decide not only what implications it has, but most importantly whether it makes sense for the project in general, because I don't think a flag for just about anything is a way to a sustainable and maintainable future given how many (in general) combinations you'll end up with and toggling particular SBOM output segments isn't really behaviour configuration per se.
That said, this PR isn't the right place to start having these discussions and so I won't stand in the way.

cachi2/core/models/sbom.py
eskultety: @taylormadore would you care enough to create an issue on investigating the possibility of adopting a more general approach covering all backends via the built-in description fields? I like the idea in general and I feel we should invest some time into exploring the possibility before ruling it out completely.

cachi2/core/models/property_semantics.py
```diff
@@ -58,6 +60,8 @@ def from_properties(cls, props: Iterable[Property]) -> "Self":
                 pip_package_binary = True
             elif prop.name == "cachi2:bundler:package:binary":
                 bundler_package_binary = True
+            elif prop.name == "cachi2:rpm_summary":
+                rpm_summary = ""
```
eskultety, line 64: This line here will cause this feature to never actually format the property to the output. I'd suggest adding an integration test case for this just to be sure we're covered.

eskultety: Tangential nitpick - since this is a custom property only relevant to RPMs and only ever used with RPM components, do we need the additional slight redundancy in the the property name - "rpm_summary" instead of just "summary"?

cachi2/core/package_managers/rpm/main.py
eskultety: @a-ovchinnikov IIUC https://rpm-software-management.github.io/rpm/manual/tags.html#base-package-tags SUMMARY is a mandatory field so you should be able to safely assume it's always present and hence treat it the same way as VERSION or RELEASE.

cachi2/core/models/input.py
eskultety, line 219: nitpick - it's part of `RpmPackageInput` already so the 'rpm' part of the option name is redundant, but at the same time the name itself doesn't indicate this is an SBOM only flag - what about `sbom_inlude_summary` OR `include_summary_in_sbom`? What I'm after is some kind of naming pattern that could be reasonably reusable for other such flags we might be required to add, IOW `include_XYZ_in_sbom`.

cachi2/core/models/property_semantics.py
a-ovchinnikov: Oof, good catch, thank you!

> do we need the additional slight redundancy?

I believe yes, it slightly pollutes SBOMs, but at the same time makes it easier to comprehend where it belongs when reading Properties code.

a-ovchinnikov: Added an integration test to cover this case.

brunoapimentel (approved):

eskultety (approved):
Approve, but please try to address the input JSON option naming nitpick I raised I think it's justified for users sake.

tests/integration/test_data/rpm_summary_is_reported/bom.json
```diff
@@ -0,0 +1,91 @@
...
+  "components": [
+    {
+      "name": "glibc-common",
+      "properties": [
+        {
+          "name": "cachi2:found_by",
+          "value": "cachi2"
+        },
+        {
+            'name': 'cachi2:rpm_summary',
```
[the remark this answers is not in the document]
a-ovchinnikov: Fixed.

cachi2/core/models/input.py
a-ovchinnikov: Good point, done.

PR #765, opened 2024-12-09
Introduce a new dependabot-related pip-compile GH actions workflow [pydantic saga]

eskultety:
This should finally resolve the problem of pydantic async release cycles for `pydantic` and `pydantic-core` causing CI issues preventing merges of pydantic-related changes (see patch 2 for detailed explanation).
The effect of the workflow is illustrated here: https://github.com/eskultety/cachi2/pull/33#issuecomment-2509808483

Note, after the workflow, manual intervention is still going to be needed, because the only thing the workflow does is issue a comment, but then it's just a matter of applying the patch from the comment and force pushing to dependabot's branch and that should be it.

**Depends on #764**

### DIFF
--- a/.github/workflows/dependabot-pipcompile.yml
+++ b/.github/workflows/dependabot-pipcompile.yml
@@ -1,0 +1,76 @@
+name: Pip-compile
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    paths:
+      - requirements.txt
+      - requirements-extras.txt
+  workflow_dispatch:
+    inputs: {}
+
+# Need these permissions for the GITHUB_TOKEN to be able to post a comment to a PR
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  versions-check:
+    runs-on: ubuntu-24.04
+    container:
+      image: python:3.9-alpine
+
+    steps:
+      # Need to install git before running the checkout action in a container
+      - name: Install dependencies
+        run: apk update && apk add --no-cache git
+
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Install pip-tools
+        run: |
+          pip install --upgrade pip
+          pip install --no-cache-dir pip-tools
+
+        # This step uses multi-line string injection to GitHub environment [1]
+        # [1] https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#multiline-strings
+      - name: Run pip-compile to update requirements.txt
+        run: |
+          git config --global --add safe.directory "*"
+          pip-compile --generate-hashes --output-file=requirements.txt pyproject.toml
+          pip-compile \
+              --all-extras \
+              --allow-unsafe \
+              --generate-hashes \
+              --output-file=requirements-extras.txt \
+              pyproject.toml
+          {
+            echo 'GIT_DIFF<<EOF'
+            git diff -p
+            echo EOF
+          } >> "$GITHUB_ENV"
+
+      # Only comment on PRs when changes to requirements files are needed, based on:
+      #   - https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution
+      #   - https://github.com/actions/github-script?tab=readme-ov-file#comment-on-an-issue
+      #   - https://github.com/actions/github-script/issues/247#issuecomment-1079839739
+      #   - https://github.com/actions/github-script/issues/220#issuecomment-1007633429
+      - name: Comment on pull request
+        uses: actions/github-script@v7
+        if: env.GIT_DIFF != ''
+        env:
+          DIFF: "Changes to requirements files are needed. If you're experiencing CI test failures, please apply the following patch and update the pull request:\n```diff\n${{ env.GIT_DIFF }}\n```"
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: process.env.DIFF
+            })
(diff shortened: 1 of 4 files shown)


eskultety:
Note that I also considered force pushing directly from the workflow to the same branch, but I originally faced some permissions issue. Admittedly, that could have been due to missing GH token permissions (addressed), but then I'd be most likely spawning a new concurrent gating job out of a different workflow just because we'd be force pushing a pip-compile change. Anyway, I reckoned such approach would have been obfuscating and opaque so I went with with a clearer approach requiring a bit of manual intervention though.

eskultety:
ping.

cachi2/core/package_managers/generic/models.py
a-ovchinnikov, line 1: Does this commit belong here? https://github.com/containerbuildsystem/cachi2/pull/764 looks suspiciously similar.

.github/workflows/dependabot-pipcompile.yml
a-ovchinnikov, line 1: Sorry, why cannot we just merge the PR then? Why is this part needed:
>If changes are needed, the GitHub actions bot will ... provide a patch to the
reviewer to apply and update the pull request

cachi2/core/package_managers/generic/models.py
eskultety: It is mentioned in the description:

_Depends on https://github.com/containerbuildsystem/cachi2/pull/764_

and so this PR was rebased onto #764 because that commit in question needs to go in first for this PR to start making sense. I mean we could do it in the inverse order as well, but you would not see the the behaviour described in this PR in the dependabot PR in question right away because unit tests would start failing which #764 is taking care of, so that's the context :).

.github/workflows/dependabot-pipcompile.yml
eskultety: I'm sorry I don't follow, can you be more verbose?

cachi2/core/package_managers/generic/models.py
a-ovchinnikov: Oh, got it, thank you!

.github/workflows/dependabot-pipcompile.yml
a-ovchinnikov: After reading the change once again I realized I misunderstood the flow on the first pass. Now I see that you propose to check if pydantic version was bumped and if yes then you also update pydantic-core, but this does not happen in the same patch (for some reason at the moment of writing of my earlier comment I believed that pydantic-core update would also be contained in the same dependabot PR).

I suggest the following change to the commit message:

```diff
- run a dedicated workflow on every dependabot pull request that would
- check whether changes to our requirements files are needed. If changes
- are needed, the GitHub actions bot will comment on the pull request
+ run a dedicated workflow on every dependabot pull request that would
+ check whether pydantic version was changed. If yes then the Github actions bot
+ will generate necessary change to requirements.txt to upgrade to matching 
+ pydantic-core version and publish it as a comment on dependabot's pull request
```

a-ovchinnikov: I wonder if it is possible to automatically add a commit to dependabot's PR? (I found [this discussion](https://github.com/orgs/community/discussions/24945) which looks promising)

eskultety: See my earlier comment https://github.com/containerbuildsystem/cachi2/pull/765#issuecomment-2527112587.
In essence, force pushing within a workflow would inherently trigger another workflow with the original most likely still failing. I have tried force pushing earlier, but faced some issues (probably just token permissions which may have been solved with the permission changes I have here. In any case, that approach seemed opaque and non-transparent to me, potentially source of hidden issues. Let me know if you insist on me pursuing that approach instead of this one.

eskultety: > I suggest the following change to the commit message:
> 
> ```diff
> - run a dedicated workflow on every dependabot pull request that would
> - check whether changes to our requirements files are needed. If changes
> - are needed, the GitHub actions bot will comment on the pull request
> + run a dedicated workflow on every dependabot pull request that would
> + check whether pydantic version was changed. If yes then the Github actions bot
> + will generate necessary change to requirements.txt to upgrade to matching 
> + pydantic-core version and publish it as a comment on dependabot's pull request
> ```

 Okay, but this isn't only about pydantic, this results in *any* pip-compile changes, IOW, this workflow looks whether a change to any of our requirements files was presented in which case it'll run pip-compile to make sure *all* deps can be successfully re-pinned. I'll try to capture it better in the commit msg.

a-ovchinnikov: > Let me know if you insist 

Absolutely not! Having a rare and well-documented manual intervention has its benefits in this specific case.

> this isn't only about pydantic

Fair enough, then what about this:
```diff
- whether changes to our requirements files are needed. If changes are
+ whether any additional changes to our requirements files are needed. If yes
```
I have completely missed the 'additional' part, having it explicitly highlighted would be my only request, the change LGTM otherwise.

a-ovchinnikov (approved):
LGTM with a suggestion to clarify wording in one of commits.

.github/workflows/dependabot-pipcompile.yml
eskultety: > > Let me know if you insist
> 
> Absolutely not! Having a rare and well-documented manual intervention has its benefits in this specific case.

I would drop "rare" from that sentence :), this is going to happen pretty much with every grouped monthly dependabot update, because we will no longer separate pydantic from the rest of the deps, so the next time deps are updated, only `pydantic` is going to be proposed by dependabot with this workflow suggesting the correct version of `pydantic-core` in a comment-supplied patch via the GH script action.

eskultety: Commit msg adjustment applied.

a-ovchinnikov: Thank you!

slimreaper35:
I wonder if we could use this workflow to completely replace dependabot and regenerate `requirements` each month from scratch.

eskultety:
> I wonder if we could use this workflow to completely replace dependabot and regenerate `requirements` each month from scratch.

As appealing as that might sound (especially to me since I'm not a fan of dependabot), you'd lose the one useful bit of information it provides during these updates - changelog! I can only wish the industry would use the dedicated OCI labels for this purpose in container images  as well so that the hashes wouldn't be completely opaque leaving us with no other choice but to blindly just hit merge if the CI passes. Well, it is what it is for container images...

slimreaper35 (approved):

PR #767, opened 2024-12-12
Support for Yarn v4

brunoapimentel:
This PR brings the investigation regarding introducing support for Yarn v4 in the form of an ADR. Since the changes seem to be trivial, both the implementation and the ADR are included.

The ADR includes the full investigation made in the `Context` section, since I belive it is important to keep this information for historical reasons.


slimreaper35:
Is this supposed to be ADR or a design doc? We have both right now, so I am a bit confused what is the use case for both.

docs/adr/0002-yarn-v4.md
```diff
@@ -0,0 +1,175 @@
...
+
+Yarn v4 introduced the subdirectory (`#external-packages/strip-ansi-4.0.0.tgz`) and the hash (`::hash=e17689`), which are already [handled](https://github.com/containerbuildsystem/cachi2/blob/a5f19c6f9be90be4289beee35ecccd2827bbb328/cachi2/core/package_managers/yarn/locators.py#L77-L79) by the v3 implementation in Cachi2.
+
+**Changes to some instances of the patch locator:**
+```
+-  resolution: "typescript@patch:typescript@npm%3A5.1.6#~builtin<compat/typescript>::version=5.1.6&hash=5da071"
++  resolution: "typescript@patch:typescript@npm%3A5.1.6#optional!builtin<compat/typescript>::version=5.1.6&hash=5da071"
+```
+
+Instances of the patch locator are currently [being ignored](https://github.com/containerbuildsystem/cachi2/blob/a5f19c6f9be90be4289beee35ecccd2827bbb328/cachi2/core/package_managers/yarn/resolver.py#L264-L268) in the v3 implementation in Cachi2.
```
chmeliik, line 152: Ooh, that seems like a bug. They're not being ignored, they're being reported as if they were regular registry dependencies. Which is not quite right, the patched dependency doesn't have to be a registry dependency.

Fixing the parsing of patch locators would be out of scope then? It would have to be fixed before cachi2 can start reporting the patches

cachi2/core/package_managers/yarn/main.py
```diff
@@ -174,9 +175,12 @@ def _get_plugin_allowlist(yarn_rc: YarnRc) -> list[Plugin]:
     This list should only have official plugins that add new protocols and that also do not
     implement the 'fetchPackageInfo' hook, since it would allow arbitrary code execution.
 
+    Note that starting from v4, the official plugins are enabled by default and can't be disabled.
```
eskultety, line 178: Do we need to be concerned by any of the official plugins (disclaimer: I haven't checked the full set of plugins)?

docs/adr/0002-yarn-v4.md
```diff
@@ -0,0 +1,175 @@
...
+### All official plugins are enabled by default
+
+And can no longer be disabled by `.yarnrc.yml` configuration. The current official plugins don't seem to introduce any behavior that would taint the accuracy of the prefetched dependencies, though. Most of them only add support to protocols (which will still be filtered using the same rules of Yarn v4) or CLI commands.
+
+The list of all official plugins can be found under the "Default Plugins" section in the [API](https://yarnpkg.com/api) page on the official Yarn documentation. Here's a short summary of every official plugin (as of 4.5.3):
+
+<details>
+    <summary>Plugins that add support for a protocol</summary>
+
+- plugin-exec
```
eskultety, line 31: Can you refresh my mind on how the `exec` protocol cannot lead to arbitrary code execution in a transitive package? My memory has already faded.

brunoapimentel: It can lead to arbitrary code execution, that's why we disallow it [here](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/yarn/locators.py#L172-L176).

The reason for not disabling the plugin straight away is that it will make `yarn info` fail with an ugly error.

cachi2/core/package_managers/yarn/main.py
brunoapimentel: AFAICT, no, we don't. I wrote a small section about the official plugins and their functions in the ADR to try to TL;DR  the reason why I believe this.

docs/adr/0002-yarn-v4.md
brunoapimentel: IIRC, we wanted to report patched dependencies as a regular dependency type and add the [pedigree.patches ](https://cyclonedx.org/docs/1.6/json/#components_items_pedigree_patches) information to the SBOM component.

I think the latter part is not being done, so you're right, we need to fix it.

chmeliik: > IIRC, we wanted to report patched dependencies as a regular dependency type and add the [pedigree.patches ](https://cyclonedx.org/docs/1.6/json/#components_items_pedigree_patches) information to the SBOM component.

That makes sense, but the current approach has a bug: https://github.com/containerbuildsystem/cachi2/blob/d89b5674c55d48d33eeb9cf3acc6d77f8692f980/tests/integration/test_data/yarn_e2e_test/bom.json#L323

That's a purl for a regular registry dependency, generated for https://github.com/cachito-testing/cachi2-yarn-berry/blob/70515793108df42547d3320c7ea4cd6b6e505c46/yarn.lock#L272 (definitely not a registry dependency)

The correct purl also happens to be present, because yarn.lock also lists the "un-patched" dependency separately: https://github.com/containerbuildsystem/cachi2/blob/d89b5674c55d48d33eeb9cf3acc6d77f8692f980/tests/integration/test_data/yarn_e2e_test/bom.json#L335

So in practice, the bug is that cachi2 reports bogus registry dependencies in these situations

docs/adr/0002-yarn-v4.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Yarn now caches npm version metadata
+
+Although Yarn only seems to create the cache folder (`{globalFolder}/metadata/npm`) when the [hardened mode](https://yarnpkg.com/configuration/yarnrc#enableHardenedMode) is enabled. Simply running `yarn install` on a local project or even prefetching with Cachi2 doesn't seem to generate this extra folder.
+
+The metadata, in itself, is a collection of json files with a few kilobytes each. In case we decide to enable the hardened mode, it'd be ideal to delete this npm metadata this from the output folder, although there's no harm in keeping it.
+
+### Changes in yarn.lock
+
+When updating v3 projects to v4, some differences in the lockfile appear. These are some picks from updating the [disallowed-protocols]() branch on the cachi2-yarn-berry test project:
```
a-ovchinnikov, line 117: `disallowed-protocols` link appears to be broken.

docs/adr/0002-yarn-v4.md
```diff
@@ -0,0 +1,243 @@
...
++  checksum: ...
+```
+
+**Changes to some instances of the file locator:**
+```
+-  resolution: "strip-ansi-tarball@file:external-packages/strip-ansi-4.0.0.tgz::locator=berryscary%40workspace%3A."
++  resolution: "strip-ansi-tarball@file:external-packages/strip-ansi-4.0.0.tgz#external-packages/strip-ansi-4.0.0.tgz::hash=e17689&locator=berryscary%40workspace%3A."
+```
+
+Yarn v4 introduced the subdirectory (`#external-packages/strip-ansi-4.0.0.tgz`) and the hash (`::hash=e17689`), which are already [handled](https://github.com/containerbuildsystem/cachi2/blob/a5f19c6f9be90be4289beee35ecccd2827bbb328/cachi2/core/package_managers/yarn/locators.py#L77-L79) by the v3 implementation in Cachi2.
```
a-ovchinnikov, line 144: So, does this mean that no change was introduced after all?

docs/adr/0002-yarn-v4.md
```diff
@@ -0,0 +1,243 @@
...
+173.75s user 92.09s system 336% cpu 78.99 total<br>
+111.99s user 69.36s system 315% cpu 57.568 total<br>
+*average time 67.155*
+
+hardened:<br>
+245.61s user 103.60s system 351% cpu 99.30 total<br>
+274.81s user 109.91s system 361% cpu 106.42 total<br>
+244.61s user 107.89s system 349% cpu 100.81 total<br>
+*average time 102.176*<br>
+*~52 increase*
```
a-ovchinnikov, line 231: s/52/52%

docs/adr/0002-yarn-v4.md
```diff
@@ -0,0 +1,243 @@
...
+
+</details>
+
+## Decision
+
+The changes needed to support Yarn v4 boil down to raising the maximum supported version, adding a few integration tests to cover v4 scenarios and updating the documentation. The Corepack shim is compatible with v4, so the project will still be processed with the exact Yarn version that is defined in its configuration files.
+
+Enabling the newly introduced hardened mode during the prefetch might prove useful to further increase the security of the process, but is by no means necessary to introduce basic support for v4. The impact on the prefetching speed is something that needs to be investigated in depth before we decide to enable it, but the decision can be deferred to a later moment.
```
a-ovchinnikov, line 239: Could we just add a user-accessible toggle switch for that?

a-ovchinnikov (commented):
A few nitpicks/suggestions and minor questions, LGTM overall.

docs/adr/0002-yarn-v4.md
```diff
@@ -0,0 +1,243 @@
...
+The [enableGlobalCache](https://yarnpkg.com/configuration/yarnrc#enableGlobalCache) option allows the user to set a shared location for the cache folder. Cachi2 already treats this option the following way:
+
+- Set to `true` during the prefetch, and point it to a specific location in order to keep the prefetched dependencies.
+- Set to `false` during the build, since the build needs to read from the cache, but the dependencies need to be installed to a local folder since they'll be used during the runtime.
+
+The only action that might be necessary is to document this behavior, since now during build-time, Cachi2 is setting the opposite option that would be expected as default.
+
+### All official plugins are enabled by default
+
+And can no longer be disabled by `.yarnrc.yml` configuration. The current official plugins don't seem to introduce any behavior that would taint the accuracy of the prefetched dependencies, though. Most of them only add support to protocols (which will still be filtered using the same rules of Yarn v4) or CLI commands.
```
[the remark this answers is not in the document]
brunoapimentel: I tried to bend the parser rules here :sweat_smile:

docs/adr/0002-yarn-v4.md
```diff
@@ -0,0 +1,243 @@
...
+
+Yarn v4 was released in October 23, 2023, which coincided with the time that the support for Yarn v3 was being introduced in Cachi2. To limit the scope of implementation, we decided to leave support for v4 to a later time. Now, as more projects migrate to v4, introducing proper support has become a priority.
+
+This document analyzes the the most impactful changes introduced with v4, and how they relate to the current v3 implementation.
+
+Main references:
+- [v4 blog post](https://yarnpkg.com/blog/release/4.0)
+- [Breaking changes for Yarn 4](https://github.com/yarnpkg/berry/issues/3591)
+
+### Global cache is enabled by default
```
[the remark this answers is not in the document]
brunoapimentel: I like it, but then we're adding a new header that is not present in the standard [ADR template](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions).

I wonder if we should move this section to another document, something like a `docs/designs` folder, that can be supplemental to the ADRs.

Or... we can simply take liberties with the original proposed ADR template, not sure what's better.

docs/adr/0002-yarn-v4.md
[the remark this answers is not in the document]
brunoapimentel: Fixed.

docs/adr/0002-yarn-v4.md
```diff
@@ -0,0 +1,243 @@
+# Extend Yarn Berry support to v4
+
+## Context
+
+Yarn v4 was released in October 23, 2023, which coincided with the time that the support for Yarn v3 was being introduced in Cachi2. To limit the scope of implementation, we decided to leave support for v4 to a later time. Now, as more projects migrate to v4, introducing proper support has become a priority.
+
+This document analyzes the the most impactful changes introduced with v4, and how they relate to the current v3 implementation.
```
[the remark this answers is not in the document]
brunoapimentel: Fixed.

docs/adr/0002-yarn-v4.md
brunoapimentel: Fixed.

docs/adr/0002-yarn-v4.md
brunoapimentel: Fixed.

docs/adr/0002-yarn-v4.md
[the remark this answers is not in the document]
a-ovchinnikov: > I tried to bend the parser rules here

I actually like the idea, it is just it does not seem to work that well in ADR setting -- jumping to the paragraph from table of content will result in some confusion (at least for me).

docs/adr/0002-yarn-v4.md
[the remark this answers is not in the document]
a-ovchinnikov: This document is a fine place to keep the section, having it here does not introduce discontinuity. I'd say having an extra section that improves the flow of the doc is better than obedience to the template (but this is a debatable point).

What about going one level deeper then?

```diff
+ ### Changes in Yarn behavior
+ #### Global cache is enabled by default
```

cachi2/core/package_managers/yarn/main.py
```diff
@@ -210,6 +210,15 @@ def _set_yarnrc_configuration(project: Project, output_dir: RootedPath) -> None:
...
 
+    # version can be read from `package.json` since we have already executed
+    # `_configure_yarn_version` at this point
+    version = get_semver_from_package_manager(project.package_json.package_manager)
+
+    # In Yarn v4, constraints can be automatically executed as part of `yarn install`, so they
+    # need to be explicitly disabled
+    if version in VersionsRange("4.0.0-rc1", "5.0.0"): #type: ignore
+        yarn_rc.enable_constraints_checks = False
+
```
brunoapimentel, line 221: I really don't like this solution, but I think to make this better, we'll need to move `_configure_yarn_version` to `project.py`, and expose a `Project.version` property to avoid needing to parse the version from `package.json`/`.yarnrc.yml` twice.

It is a big refactoring, though, so I prefer not to mix it with this PR.

brunoapimentel (commented):
New changes: force `enableConstraintsChecks` to be `false` during the prefetch, to avoid arbitrary code execution.

tests/integration/test_yarn.py
```diff
@@ -48,6 +48,17 @@
...
+            utils.TestParameters(
+                repo="https://github.com/cachito-testing/cachi2-yarn-berry.git",
+                ref="53a2bfe8d5ee7ed9c2f752fe75831a881d54895f",
+                packages=({"path": ".", "type": "yarn"},),
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="Processing the request using yarn@4.5.2",
+            ),
+            id="yarn_v4",
+        ),
```
slimreaper35, line 61: Shouldn't this be among e2e tests with a Dockerfile?

tests/unit/package_managers/yarn/test_main.py
```diff
@@ -313,7 +318,28 @@ def test_set_yarnrc_configuration(
...
+) -> None:
+    yarn_rc = YarnRc(mock.Mock(), {})
+
+    package_json = mock.Mock()
+    package_json.package_manager = "yarn@4.0.0"
+
+    project = mock.Mock()
+    project.yarn_rc = yarn_rc
+    project.package_json = package_json
+    output_dir = RootedPath("/tmp/output")
```
slimreaper35, line 333: nitpick: this could be replaced by `rooted_path` fixture

tests/unit/package_managers/yarn/test_main.py
```diff
@@ -48,9 +48,12 @@ class YarnVersions(Enum):
     YARN_V4_RC1 = semver.VersionInfo(4, 0, 0, prerelease="rc1")
     YARN_V4 = semver.VersionInfo(4, 0, 0)
 
+    YARN_V5_RC1 = semver.VersionInfo(5, 0, 0, prerelease="rc1")
+    YARN_V5 = semver.VersionInfo(5, 0, 0)
```
slimreaper35, line 52: Are these constants used anywhere ?

tests/integration/test_yarn.py
brunoapimentel: My reasoning that a more simple integration test would cover the new use case. Do you think we need to have a full e2e test to it? Any special reason for it?

tests/unit/package_managers/yarn/test_main.py
brunoapimentel: Actually, yes. If you look at the `unsupported` method of this class, you'll see that all class attributes are used to calculated the list of unsupported versions.

tests/unit/package_managers/yarn/test_main.py
brunoapimentel: Fixed

docs/adr/0002-yarn-v4.md
brunoapimentel: We most definitely could, I just wonder if it is justifiable to have the added complexity, and if we would want a user to disable it at all.

If we went that route, we would probably need to write in the SBOM that hardened mode was off, so that a pipeline could gate builds made with hardened mode off.

docs/adr/0002-yarn-v4.md
brunoapimentel: The lockfile was modified, but Cachi2 is already able to parse the changes. In the end, there was no effective change from Cachi2's perspective, since the prefetch will still be performed the same way.

docs/adr/0002-yarn-v4.md
```diff
@@ -0,0 +1,175 @@
...
+
+<details>
+    <summary>Plugins that add support for a protocol</summary>
+
+- plugin-exec
+- plugin-file 
+- plugin-git
+- plugin-http
+- plugin-link
+- plugin-npm
```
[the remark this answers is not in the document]
brunoapimentel: What disables the execution of lifecyle scripts is the [--mode=skip-build](https://yarnpkg.com/cli/install#options-mode) option, that we use in v3 and has the same behavior in v4.

docs/adr/0002-yarn-v4.md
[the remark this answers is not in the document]
brunoapimentel: I made it one level of nesting deeper, but I don't really like the result. It is fine since this is a small investigation, but would probably be troublesome for larger docs. We can probably add extra sections in those cases, I see no reason why we need to be extra strict with the original template.

docs/adr/0002-yarn-v4.md
a-ovchinnikov: Fair enough, thank you.

docs/adr/0002-yarn-v4.md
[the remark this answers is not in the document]
a-ovchinnikov: Thank you. I think extra sections are fine, and the nesting is mostly fine too.

a-ovchinnikov (approved):

tests/integration/test_yarn.py
slimreaper35: I thought we used `e2e` for hermetic builds with generating SBOMs. That's why I applied [this](https://github.com/containerbuildsystem/cachi2/blob/main/noxfile.py#L140) env variable when generating a new SBOM via noxfile. Currently, the yarn v4 test is ignored for that.

> Any special reason for it?

Maybe just for consistency with other pkg managers. We can improve it somehow later after moving all tests to a single repo.

brunoapimentel: > I thought we used e2e for hermetic builds with generating SBOMs

Hummm, I don't think that is the case. I think that e2e tests are the ones that we actually go the extra  step of building the image with the prefetched content. 

I just looked at `test_gomod.py` and `test_pip.py`, and all the cases under `test_gomod_packages` or `test_pip_packages` have their output SBOM commited to the repo. I might still be misunderstanding your point, but maybe the env variable you pointed is not correct.

slimreaper35: Oh okay, you are right, so I need to fix that :disappointed:

slimreaper35 (approved):

PR #770, opened 2024-12-18
bundler git deps fixes

slimreaper35:

### DIFF
--- a/cachi2/core/package_managers/bundler/parser.py
+++ b/cachi2/core/package_managers/bundler/parser.py
@@ -116,10 +116,12 @@ class GitDependency(_GemMetadata):
 
     Attributes:
         url:        The URL of the git repository.
+        branch:     The branch to checkout.
         ref:        Commit hash.
     """
 
     url: AcceptedUrl
+    branch: Optional[str] = None
     ref: AcceptedGitRef
 
     @cached_property
@@ -153,7 +155,11 @@ class GitDependency(_GemMetadata):
             to_path=git_repo_path.path,
             env={"GIT_TERMINAL_PROMPT": "0"},
         )
-        repo.git.checkout(self.ref)
+
+        if self.branch is not None:
+            repo.git.checkout(self.branch)
+
+        repo.git.reset("--hard", self.ref)
 
 
 class PathDependency(_GemMetadata):
--- a/cachi2/core/package_managers/bundler/main.py
+++ b/cachi2/core/package_managers/bundler/main.py
@@ -90,7 +90,7 @@ def _resolve_bundler_package(
         else:
             properties = []
         if isinstance(dep, GitDependency):
-            git_paths.append((dep.name, dep.name + "-" + dep.ref[:12]))
+            git_paths.append((dep.name, dep.repo_name + "-" + dep.ref[:12]))
 
         c = Component(name=dep.name, version=dep.version, purl=dep.purl, properties=properties)
         components.append(c)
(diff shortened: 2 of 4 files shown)


cachi2/core/package_managers/bundler/parser.py
```diff
@@ -153,7 +154,11 @@ def download_to(self, deps_dir: RootedPath) -> None:
             to_path=git_repo_path.path,
             env={"GIT_TERMINAL_PROMPT": "0"},
         )
-        repo.git.checkout(self.ref)
```
a-ovchinnikov, line 156: If there is no branch specified we might actually want to checkout the ref, don't we? It is either us or bundler during install phase (and that could cascade in a very wrong direction).

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -47,6 +47,7 @@ def test_resolve_bundler_package(
     git_dep = GitDependency(
         name="my-git-dep",
         version="0.1.0",
+        branch=None,
```
a-ovchinnikov, line 50: Can this be skipped? (since it is optional in the model anyway).

cachi2/core/package_managers/bundler/parser.py
```diff
@@ -120,6 +120,7 @@ class GitDependency(_GemMetadata):
     """
 
     url: AcceptedUrl
+    branch: Optional[str]
```
a-ovchinnikov, line 123:
```diff
-    branch: Optional[str]
+    branch: Optional[str] = None
```
It is (hopefully) a corner-case, I suggest not polluting everything just because of it.

a-ovchinnikov (commented):
Short and mostly LGTM with a few super small questions.

cachi2/core/package_managers/bundler/parser.py
slimreaper35: Yes, but I changed it to `reset --hard <ref>` so we don't end up in the detached state

slimreaper35: What do you mean by
> it is unlikely that these repos will be later used for development

---
> Since it does not make sense to specify a ref when someone has specified a branch, but they still decided to specify the ref, let's do what they have asked and reset to it anyway.

What if the ref is from a different branch than the specified branch? :smile: Maybe this is an unrealistic scenario, but who knows with bundler?

a-ovchinnikov: > it is unlikely that these repos will be later used for development

Being in headless state is a problem if someone is to try and commit something to this repository. If they are to build it then it is probably fine (assuming that this ref does not point to a state in which it is unbuildable -- which we cannot protect against and also should not).

> What if the ref is from a different branch than the specified branch? 

Possible, but I'd argue that here we are in the area of deliberate and methodical attempts to shoot oneself in the foot. We cannot shield a user from all possible issues with their repo. We might issue a warning though.

slimreaper35: But that's not their repo, it is a dependency they specified in the `Gemfile`.

a-ovchinnikov: > But that's not their repo

Exactly! That's why it is not a problem. I think no change to your current code is needed.

a-ovchinnikov (approved):

brunoapimentel (approved):

PR #775, opened 2025-01-02
Yarn Berry refactoring and minor improvements

brunoapimentel:
- Turns the YarnRc and PackageJson project classes into UserDict subclasses (to avoid mapping all necessary attributes)
- Avoids parsing the version from PackageJson twice when setting the `.yarnrc.yml` configuration


cachi2/core/package_managers/yarn/project.py
```diff
@@ -49,220 +45,33 @@ def __init__(self, path: RootedPath, data: dict[str, Any]) -> None:
...
-        """Get the globally configured registry server.
+    def __getitem__(self, key: str) -> Any:
+        try:
+            return self.data[key]
+        except KeyError:
+            try:
+                # try the defaults potentially raising KeyError
+                return self._defaults[key]
+            except KeyError:
+                return None
```
a-ovchinnikov, line 58: I wonder if adding something along these lines would be a good idea:
```python
def __getattr__(self, attr):
    if attr in self.__dict__:
         return self.__dict__[attr]
    return self[attr]

def __setattr__(self, attr, value):
    self.data[attr] = value
```
This would let keeping most client code intact.

a-ovchinnikov (approved):
LGTM with a minor suggestion.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -35,12 +33,10 @@ class Plugin(TypedDict):
...
 
-class YarnRc:
-    """A yarnrc file.
+class ConfigurationFile(UserDict):
+    """This is it."""
 
-    This class abstracts the underlying attributes and only exposes what
-    is relevant for the request processing.
-    """
+    _defaults: dict[str, Any] = {}
```
slimreaper35, line 39: Have you considered using `defaultdict` instead ? It should cover our use case I believe.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -49,220 +45,33 @@ def __init__(self, path: RootedPath, data: dict[str, Any]) -> None:
...
-    @enable_global_cache.setter
-    def enable_global_cache(self, enable_global_cache: Optional[bool]) -> None:
-        self._data["enableGlobalCache"] = enable_global_cache
+    _defaults = {
+        "cacheFolder": "./.yarn/cache",
+        "lockfileFilename": "yarn.lock",
+        "unsafeHttpWhitelist": [],
+        "plugins": [],
+        "npmRegistryServer": "https://registry.yarnpkg.com",
+    }
```
eskultety, line 74: When I saw this PR I was wondering how you solved this, because handling defaults properly create a (fixable in our case) chicken-egg problem. By properly I mean calling into a given `yarn` version and query the defaults from it directly. That however, would happen from within the `YarnRC` instantiation (via a `Project` instance) before we could have set the yarn version for corepack which is what held me up when I was working on this in the past: https://github.com/eskultety/cachi2/commit/39a89cad330f18a225bc2682da4c1143b23945e4#diff-cbc68c73831a63c3d995893ed401dc07bf92cd8e8654b498177c465e8584ef18R85 and sadly I didn't find the time to finish it.
That all said, based on the fact that we're intending to support multiple versions of modern Yarn with ideally the same backend and data structures, I don't think it's wise to hard code (only) a couple of options; moreover, a particular Yarn release might change the actual value of any of these options getting us into potential problems, making this approach opaque. Instead, the solution should be robust enough to easily start querying any supported YarnRC option. Therefore, I'd like to see this being filled in properly from the given yarn version's output.
In order to be able to do ^this, you'll have to do a few things to avert the chicken-egg problem:

1. defer the instantiation of `project` to `_resolve_yarn_project` (it's too soon in https://github.com/containerbuildsystem/cachi2/blob/10144e70972af69fb20d9f682d576e0c799a5cfa/cachi2/core/package_managers/yarn/main.py#L26 anyway)
2. parse the `.yarnrc` outside of `Project` just to be able to query `yarnPath` as that's the only thing blocking you at this point to execute the correct Yarn binary initializing the defaults; honestly, as bad as it sounds, regex might be worth a shot here to avoid passing a raw dictionary with user `.yarnrc` values around until they find their way to a `YarnRC` instance.
3. instantiate `YarnRC` and `Project` (passing one to the other)

If I haven't missed anything ^this should be all we need to handle defaults properly.

cachi2/core/package_managers/yarn/project.py
```diff
@@ -35,12 +33,10 @@ class Plugin(TypedDict):
     spec: str
 
 
-class YarnRc:
-    """A yarnrc file.
+class ConfigurationFile(UserDict):
```
eskultety, line 36: Hmm, we modeled this as a module-scoped (private) ABC parent class in the yarn classic backend: https://github.com/containerbuildsystem/cachi2/blob/10144e70972af69fb20d9f682d576e0c799a5cfa/cachi2/core/package_managers/yarn_classic/project.py#L25, so I think ideally we'd want to be consistent with that, however, I suspect (haven't tried) you might face problems if you tried to inherit from both ABC and UserDict, especially if you tried to use ABC as a metaclass.
As much as I'd like to see us being consistent, it may not be the nicest path getting us there in this case. If anything, the class description should be changed or dropped.

eskultety (commented):
First of all, thank you for this effort which I started in the past but failed to finish (https://github.com/eskultety/cachi2/commits/yarn-misc/). I left a couple of design comments, one major, the rest is minor.

cachi2/core/package_managers/yarn/project.py
brunoapimentel: I'm not sure how we could do it. [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) allows for a `default_factory` function that accepts no args, so we could no stablish a defaut dict to be used. I might be missing something, though.

cachi2/core/package_managers/yarn/project.py
a-ovchinnikov: This is a terrible idea, at least the way I posted it, it takes more than the snippet to make it work properly.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -166,6 +166,8 @@ def _configure_yarn_version(project: Project) -> None:
     # Note (mypy): version cannot be None anymore
     _verify_corepack_yarn_version(version, project.source_dir)  # type: ignore
 
+    return version  # type: ignore
```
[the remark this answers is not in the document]
brunoapimentel: I really don't like forcefully silencing mypy, but the typecast here also did not please me. I tried an alternative solution, which did the job, LMK what you think about it.

cachi2/core/package_managers/yarn/project.py
brunoapimentel: > Yarn release might change the actual value of any of these options getting us into potential problems

I agree that this is a risk, but I don't think we need to prepare for it ahead of time. The only defaults that really matter to us are: "cacheFolder" and "lockfileFilename", which are very unlikely to change. Even the assumption of plugins being an empty list is something we are assuming by ourselves (it does not show up in the [docs](https://yarnpkg.com/configuration/yarnrc), nor in `yarn config --json`).

So I'd say we can postpone adding this complexity to the time a problem does appear (if it appears at all). In any case, we're not doing any regressions, as the values are hard-coded now. The main thing I wanted to solve here is the manual mapping of every single attribute (which was a shortsighted solution I wrote myself :sweat_smile: )

brunoapimentel: Btw, this newly added [patch](https://github.com/containerbuildsystem/cachi2/pull/775/commits/15cac3c05974287f3946c62e75aecd79b7706efb) removes even more defaults.

cachi2/core/package_managers/yarn/project.py
brunoapimentel: Oh damn, I completely missed that. I think there's no point in trying to use this as a base class (or mirror the approach), since the premise there is to have a foundation that allows manually mapping each necessary config option to an attribute, which is the approach we're trying to drop here for Yarn Berry.

I think we can either:
1- not be consistent, keep Yarn Berry's implemetation different and based on UserDict
2- try to use _CommonconfigFile as a base class and figure out a way to magically handle the attributes/options (Alexey's suggestion [here](https://github.com/containerbuildsystem/cachi2/pull/775#discussion_r1901368985))
3- Drop this PR altogether (we probably won't need to add many more config options in the near future)

cachi2/core/package_managers/yarn/main.py
```diff
@@ -104,17 +104,17 @@ def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Comp
     """
     log.info(f"Fetching the yarn dependencies at the subpath {project.source_dir}")
 
-    _configure_yarn_version(project)
+    version = _configure_yarn_version(project)
     _verify_repository(project)
 
-    _set_yarnrc_configuration(project, output_dir)
+    _set_yarnrc_configuration(project, output_dir, version)
```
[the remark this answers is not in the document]
brunoapimentel: Fixed

cachi2/core/package_managers/yarn/project.py
slimreaper35: Something like that maybe :thinking: 

```
class ConfigurationFile(UserDict):
    """This is it."""

    def __init__(self, path: RootedPath, data: dict[str, Any]) -> None:
        """Initialize a YarnRc object.

        :param path: the path to the yarnrc file, relative to the request source dir.
        :param data: the raw data for the yarnrc file.
        """
        self._path = path
        super().__init__(data)


class YarnRc(ConfigurationFile):
    """A yarnrc file.

    This class abstracts the underlying attributes and only exposes what
    is relevant for the request processing.
    """

    def write(self) -> None:
        """Write the data to the yarnrc file."""
        with self._path.path.open("w") as f:
            yaml.safe_dump(self.data, f)

    @classmethod
    def from_file(cls, file_path: RootedPath) -> "YarnRc":
        """Parse the content of a yarnrc file."""
        try:
            with file_path.path.open("r") as f:
                yarnrc_data = defaultdict(lambda: None, yaml.safe_load(f))
```
If you feel like it does not improve the readability, just leave it as it is.

eskultety: > ```
>    @classmethod
>    def from_file(cls, file_path: RootedPath) -> "YarnRc":
>       """Parse the content of a yarnrc file."""
>       try:
>            with file_path.path.open("r") as f:
>                yarnrc_data = defaultdict(lambda: None, yaml.safe_load(f))
>```

Just make sure that we merge the hardcoded defaults with the user values before passing it to defaultdict otherwise we won't consult the defaults at all.

cachi2/core/package_managers/yarn/project.py
eskultety: > Btw, this newly added [patch](https://github.com/containerbuildsystem/cachi2/pull/775/commits/15cac3c05974287f3946c62e75aecd79b7706efb) removes even more defaults.

The URL is dead because you force pushed in the meantime :).

brunoapimentel: :facepalm: 

It should be the last commit in this PR, in any case :)

cachi2/core/package_managers/yarn/project.py
brunoapimentel: Yeah, I think we could even converge to using a simple UserDict in both Yarn Berry and Classic. I agree that the current PR already brings us benefits the way it is.

tests/unit/package_managers/yarn/test_project.py
```diff
@@ -86,56 +86,54 @@ def _prepare_yarnrc_file(rooted_tmp_path: RootedPath, data: str) -> YarnRc:
...
-    assert yarn_rc.patch_folder is None
-    assert yarn_rc.pnp_data_path is None
-    assert yarn_rc.pnp_mode is None
-    assert yarn_rc.pnp_unplugged_folder is None
-    assert yarn_rc.registry_server == "https://registry.yarnpkg.com"
-    assert yarn_rc.registry_server_for_scope("foobar") == "https://registry.yarnpkg.com"
-    assert yarn_rc.unsafe_http_whitelist == []
-    assert yarn_rc.yarn_path is None
-    assert yarn_rc.virtual_folder is None
+    assert yarn_rc["cacheFolder"] == "./.yarn/cache"
```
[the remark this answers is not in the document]
brunoapimentel: So, `yarn_rc.values()` is actually an empty dict, and popping does not work if the key is not set. Your suggestion did inspire me to clean up the test, LMK if you liked the result.

tests/unit/package_managers/yarn/test_project.py
```diff
@@ -86,56 +86,54 @@ def _prepare_yarnrc_file(rooted_tmp_path: RootedPath, data: str) -> YarnRc:
 def test_parse_yarnrc(rooted_tmp_path: RootedPath) -> None:
     yarn_rc = _prepare_yarnrc_file(rooted_tmp_path, VALID_YARNRC_FILE)
```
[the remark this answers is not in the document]
brunoapimentel: I'm not sure I follow, can you clarify? :sweat_smile:

tests/unit/package_managers/yarn/test_project.py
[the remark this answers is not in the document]
a-ovchinnikov: ~According to the previous state of the code "cacheFolder" is present (`assert yarn_rc["cacheFolder"] == "./.yarn/cache"` would always fail otherwise). Once it is popped the dictionary is either empty, then the assertion holds, or all its keys are equal to None and it still holds.~ Ok, this is a bit more magical object than I thought when starting the comment. Yes, the new version does the trick, thank you.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -41,11 +41,11 @@ def fetch_yarn_source(request: Request) -> RequestOutput:
 def _verify_yarnrc_paths(project: Project) -> None:
     paths_conf_opts = {
         # pnpDataPath is only configurable in Yarn v3
-        project.yarn_rc.pnp_data_path: "pnpDataPath",
```
eskultety, line 44: **Commit msg:**

> yarnrc¹

Is ^this missing a footer link to the reference or was this supposed to be 'v1' instead?

cachi2/core/package_managers/yarn/project.py
eskultety: We're not going to use defaultdict in this iteration - resolving.

cachi2/core/package_managers/yarn/project.py
eskultety, line 58: @brunoapimentel why do you need to keep the defaults in a dedicated `_defaults` attribute? That on its own isn't a bad thing if you want to have better bookkeeping although I don't think it's really necessary, but why do you need to overload `__getitem__` like this instead of populating `self.data` first with defaults and then loading values from `yarnrc`? I mean this shadows the usual practice of using the dedicated exception mechanism to handle this case and so strictly speaking about returning None instead of raising a KeyError @slimreaper35 had a point about the `defaultdict`. So I'd say, let's just either use a try-except clause where needed OR always query dict items with `get` to get the same value, but overloading the magic method for this very purpose seems a bit like obfuscation to me.

tests/unit/package_managers/yarn/test_project.py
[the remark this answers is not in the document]
eskultety: Very good catch! Why not compare dict objects instead? Comparing by assertions was necessary back then, but now it feels inferior. Comparing by strings is prone to ordering even though dicts are now by default ordered.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -126,7 +126,10 @@ def _configure_yarn_version(project: Project) -> None:
         project.package_json["packageManager"]
     )
 
-    if yarn_path_version is None and package_manager_version is None:
```
eskultety, line 126: **Commit subj: [multiple occurrences]**

_Yarn Berry_: I know it used to be the code name back in the v2 day? Is it still accurate? We don't refer to it as berry in the package name in the code base, instead, we just refer to it as "yarn". I think we should be consistent in that otherwise it might a be a little confusing to people who are not that familiar with Yarn development.

cachi2/core/package_managers/yarn/main.py
```diff
@@ -126,7 +126,10 @@ def _configure_yarn_version(project: Project) -> None:
...
+    version = yarn_path_version if yarn_path_version else package_manager_version
+
+    # this check is done here to make mypy understand that version can't be Optional anymore
+    if version is None:
```
eskultety, line 132: Hold on, so NOW mypy understood? I tried a couple of compound if clause variations and it still complained, so this is the way? Stupid mypy.

cachi2/core/package_managers/yarn/project.py
eskultety: Agreed with 1. For the record I wasn't trying to advocate for porting the ABC stuff from classic here as I find that approach inferior to this one.

cachi2/core/package_managers/yarn/project.py
eskultety: I guess we disagree on the approach here, but it's a non-blocker for me, worst case scenario - it breaks (like everything else). This PR undeniably does improve things a lot.

cachi2/core/package_managers/yarn/main.py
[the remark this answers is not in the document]
eskultety: Sure. Looking at your attempt to make mypy finally shut up is definitely superior. Nevertheless, mypy should up its game when it comes to compound conditions.

cachi2/core/package_managers/yarn/main.py
brunoapimentel: Ack, I can change all instances to simply `Yarn`. The reason I wrote it as `Yarn Berry` is because it's the official name for Yarn versions >=2, and now that we have Yarn classic, I thought it was the best way to describe it;

a-ovchinnikov: It is referred to as "Berry", at least in metayarn.py. This was done on request to make it even more clear which yarn is  dealt with where.

cachi2/core/package_managers/yarn/main.py
brunoapimentel: I think it was supposed to be a footer link? Can't really remember. I dropped it, anyways.

cachi2/core/package_managers/yarn/main.py
eskultety: Not a fan of codenames in general, because you can never really assume everyone has the same context and understanding, e.g. Android versions (sure, ours is not the same case), then again, this is an easy enough change to pretty much just the commit messages which is what people are supposed to read.

cachi2/core/package_managers/yarn/project.py
eskultety: @brunoapimentel the last revision is still not exactly what I meant by my last comment, IOW now you have some kind of hybrid approach where on one hand you let `KeyError` to be raised (that's fine) but on the other you still overloaded `__getitem__`, I think if you merge the dicts in `__init__` it's going to be cleaner.

brunoapimentel: They definitely will be cleaner, but the downside is that we are making the defaults actual attributes, which means that they will be written to file when we call `write`. That is a change in the current behavior, unless we create another mechanism to avoid it happening.

eskultety: Please correct me here if I'm wrong, but we're processing Yarn in a temporary working copy of the source directory so the `yarnrc` changes aren't going to be carried over to the actual repo (good), so why is it a problem that we're going to write the default options to the output? Does it matter?

brunoapimentel: I would expect that yarn would behave the same way if defaults are written to file, I do find it a little unecessary though. But if you still feel like the magic method is worse, we can go that way.

tests/unit/package_managers/yarn/test_project.py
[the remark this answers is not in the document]
eskultety: This is pretty much the last open thread in the PR and here (commit https://github.com/containerbuildsystem/cachi2/pull/775/commits/9c324effd82cb80602c4572c87e068e6e259f37b) you replaced the asserts with a single comparison which is good. Both @a-ovchinnikov and I expressed the potential issue of reordering when it comes to strings, so wouldn't `assert yarn_rc.data == yaml.safe_load(VALID_YARNRC_FILE)` take care of it for good and still have it as a oneliner?

eskultety:
So the integration test is failing because `lockfileFilename` doesn't exist as a setting in v4 [1][2] which is one of the things I hinted on during my review. This issue only got visible when the change to merge the defaults inside the `YarnRc` constructor was made and `.yarnrc` got written to the disk. While we could revert that change and go back to `__getitem__` I think this shows us that we need to be more diligent about new version support for the backends and in order to make our lives easier when it comes to maintenance and make all of this clear in the code we should guard any extra options or those which were dropped guarded by a version conditional wherever needed, in this case - the `YarnRc` constructor.

[1] https://yarnpkg.com/configuration/yarnrc
[2] https://v3.yarnpkg.com/configuration/yarnrc/#lockfileFilename

tests/unit/package_managers/yarn/test_project.py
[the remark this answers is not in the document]
brunoapimentel: `VALID_YARNRC_FILE`  is already ordered in the way that safe_dump produces it, but fair enough, your suggestion seems a little bit less volatile.

brunoapimentel:
> So the integration test is failing because `lockfileFilename` doesn't exist as a setting in v4 [1][2] which is one of the things I hinted on during my review.

:shakes_fist: yarn

Well, this is tricky, because we'd still need to have the "default" `yarn.lock` string regardless of it being dropped from the yarnrc settings so that we can actually find the file.

Since we only rely on three config keys, I changed the approach to completely drop the defaults, and treat the cases where we need them during the code (each key is only accessed once, anyways).

eskultety:
> Since we only rely on three config keys, I changed the approach to completely drop the defaults, and treat the cases where we need them during the code (each key is only accessed once, anyways).

That's fair and should suffice for now. We can improve upon that in the future.

eskultety (approved):

slimreaper35 (approved):

PR #780, opened 2025-01-07
pip: Recommend pybuild-deps package for generating requirements-build.txt

slimreaper35:


docs/pip.md
```diff
@@ -434,13 +434,13 @@ but contains build dependencies rather than runtime dependencies.
 
 There's no great way to generate such a file. As far as we know, the best solution is pip-compile combined with this
 standalone script that lives in the old Cachito repo:
-[pip\_find\_builddeps.py](https://github.com/containerbuildsystem/cachito/blob/master/bin/pip_find_builddeps.py).
+[pip\_find\_build_deps.py](https://github.com/containerbuildsystem/cachi2/blob/main/hack/pip_find_build_deps.py).
```
eskultety, line 437: You'll also have to tweak the sentence above it, most importantly this part:

> script that lives in the old Cachito repo

docs/pip.md
```diff
@@ -434,13 +434,13 @@ but contains build dependencies rather than runtime dependencies.
...
 
 #### Prerequisites
 
 Generate a [fully resolved requirements.txt](#requirementstxt)
 
-Get the script ([download](https://raw.githubusercontent.com/containerbuildsystem/cachito/master/bin/pip_find_builddeps.py)
+Get the script ([download](https://raw.githubusercontent.com/containerbuildsystem/cachi2/main/hack/pip_find_build_deps.py)
```
eskultety, line 443: This section would probably benefit from some more rewording, i.e. previously people weren't expected to have cachito repo cloned, well, they may already have cachi2 cloned from some other unrelated work, so a notice that the script can be used from within the repo itself would be nice IMO. Something like
```
You can use the 'hack/pip_find_build_deps.py' script if you already have our repository cloned. If not then you can download the script directly from...
```

hack/pip_find_build_deps.py
```diff
@@ -0,0 +1,236 @@
+#!/usr/bin/env python3
```
eskultety, line 1: **commit subject nitpick:**

s/cachito/the cachito project

**commit msg:**
Commit message lacks some more context for this change.

eskultety (commented):
The PR lacks context for this change, e.g. so we're moving some script from other repo to our repo, why? What's the motivation behind the change, what is it improving/addressing? Otherwise LGTM.

chmeliik:
Is it time to start recommending https://github.com/bruno-fs/pybuild-deps instead?

brunoapimentel:
> Is it time to start recommending https://github.com/bruno-fs/pybuild-deps instead?

Yes, I pushed for this at the time you filed those [issues](https://github.com/bruno-fs/pybuild-deps/issues?q=is%3Aissue+is%3Aclosed). We can revisit this now.

docs/pip.md
slimreaper35: fixed

docs/pip.md
slimreaper35: fixed

hack/pip_find_build_deps.py
a-ovchinnikov, line 1: _Commit message_:
s/genearte/generate

a-ovchinnikov (approved):

brunoapimentel:
> I thought, that when we break the relationship with Cachito completely and update all the references, we should move the script for generating build dependencies here. I don't know in what shape is https://github.com/bruno-fs/pybuild-deps but we can for example have this as an alternative option mentioned in the docs.

I would rather just stop referecing the script at all and simply point to pybuild-deps, as it is a more robust and efficient solution.

slimreaper35:
> I would rather just stop referecing the script at all and simply point to pybuild-deps, as it is a more robust and efficient solution.
(btw, the developer has already asked us to move pybuild-deps under our organization, so we in case we decide to go that way, we would become the maintainers of this project, making it more "official")

Do you want me to do it in this PR?  Or wait until we move our repo under a new organization and decide on `pybuild-deps`  location?

brunoapimentel:
> Do you want me to do it in this PR? Or wait until we move our repo under a new organization and decide on `pybuild-deps` location?

I have a slightly preferer doing it now rather than moving the script to this repo and replacing it with pybuild-deps later. Wdyt @eskultety @a-ovchinnikov? Otherwise, let's follow up this PR as is, I'd be fine with it too.

slimreaper35:
So let's do it now. It is a simple change in the docs. Does not make sense to "temporarily" merge those two commits.

eskultety:
> > Do you want me to do it in this PR? Or wait until we move our repo under a new organization and decide on `pybuild-deps` location?
> 
> I slightly preferer doing it now rather than moving the script to this repo and replacing it with pybuild-deps later. Wdyt @eskultety @a-ovchinnikov? Otherwise, let's follow up this PR as is, I'd be fine with it too.

Can you be more specific on what "prefer doing it now" encompasses? Do you mean ONLY referencing that external project/script in our docs **OR** move their whole repo under this org and become maintainers straight away? Because I am totally on board with the former while I may need a little while to think about the latter as I've never seen the project in question.

brunoapimentel:
> Can you be more specific on what "prefer doing it now" encompasses? Do you mean ONLY referencing that external project/script in our docs **OR** move their whole repo under this org and become maintainers straight away? Because I am totally on board with the former while I may need a little while to think about the latter as I've never seen the project in question.

Only referencing the external project (pybuild-deps) and stop referencing the old script altogether. The reasoning is that pybuild-deps is a more robust solution, with tests and more readable code, it has a much faster resolution time and it solves problems such as conflicting versions for build dependencies that the current script does not.

The latter (adopting the project) is something we'll need to carefully consider, but it is out of scope for this PR (I only mentioned it because I wanted to show that this is not just a random external project)

brunoapimentel (approved):

PR #783, opened 2025-01-10
integration tests minor improvements

slimreaper35:

### DIFF
--- a/tests/integration/test_data/generic_maven_e2e/bom.json
+++ b/tests/integration/test_data/generic_maven_e2e/bom.json
@@ -2,47 +2,47 @@
   "bomFormat": "CycloneDX",
   "components": [
     {
+      "externalReferences": [
+        {
+          "type": "distribution",
+          "url": "https://repo1.maven.org/maven2/org/apache/ant/ant-launcher/1.10.14/ant-launcher-1.10.14.jar"
+        }
+      ],
       "name": "ant-launcher",
-      "purl": "pkg:maven/org.apache.ant/ant-launcher@1.10.14?checksum=sha256:f0909725a7a24e393888f3fbb558347abf506ce2f7ebc581ff26331b94d951a5&repository_url=https://repo1.maven.org/maven2&type=jar",
-      "version": "1.10.14",
       "properties": [
         {
           "name": "cachi2:found_by",
           "value": "cachi2"
         }
       ],
+      "purl": "pkg:maven/org.apache.ant/ant-launcher@1.10.14?checksum=sha256:f0909725a7a24e393888f3fbb558347abf506ce2f7ebc581ff26331b94d951a5&repository_url=https://repo1.maven.org/maven2&type=jar",
       "type": "library",
-      "externalReferences": [
-        {
-          "url": "https://repo1.maven.org/maven2/org/apache/ant/ant-launcher/1.10.14/ant-launcher-1.10.14.jar",
-          "type": "distribution"
-        }
-      ]
+      "version": "1.10.14"
     },
     {
+      "externalReferences": [
+        {
+          "type": "distribution",
+          "url": "https://repo1.maven.org/maven2/org/apache/ant/ant/1.10.14/ant-1.10.14.jar"
+        }
+      ],
       "name": "ant",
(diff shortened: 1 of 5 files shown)


tests/integration/utils.py
```diff
@@ -35,7 +35,7 @@
     )
 )
 SUPPORTED_PMS: frozenset[str] = frozenset(
-    list(resolver._package_managers) + list(resolver._dev_package_managers)
+    list(resolver._package_managers) + list(resolver._dev_package_managers) + ["yarn_classic"]
```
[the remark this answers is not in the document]
slimreaper35: Do you mean something like this ?
```
SUPPORTED_PMS = list(resolver._package_managers)
DEV_PMS = list(resolver._dev_package_managers)
EXTRA_PMS = ["yarn_classic"]

ALL_PMS: frozenset[str] = frozenset(SUPPORTED_PMS + DEV_PMS + EXTRA_PMS)
```

tests/integration/test_data/generic_maven_e2e/bom.json
a-ovchinnikov, line 1: This data looks merely shuffled, is this intentional? Is it done to make it compliant with the new data generator?

a-ovchinnikov (approved):

tests/integration/test_data/generic_maven_e2e/bom.json
slimreaper35: I don't know, honestly. This data was generated by other contributors. Usually, when you add a new package manager, you generate data, and that's it. There is no need to touch it anymore unless there is a significant change in the implementation.

a-ovchinnikov: Thank you. The commit looks like it could be dropped, OTOH this change might pop up unexpectedly the next time generator is run, so let's keep it and see how that pans out.

tests/integration/utils.py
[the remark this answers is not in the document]
eskultety: Pretty much, except that ALL_PMS would be named something else, e.g. `TRACKED_DIRS` or something more descriptive once you leave the realm of only wanting to filter tests solely based on actual pkg mgr backends. Like I said, we don't need to do that part now since we only have a single representative for the "extra" part so that would suffice :) .

tests/integration/utils.py
```diff
@@ -34,8 +34,11 @@
         Path("pyproject.toml"),
     )
 )
+
+# package managers that are not part of the resolver
```
a-ovchinnikov, line 38: They are, they are just hidden behind an alias.

slimreaper35: renamed

tests/integration/test_generic.py
```diff
@@ -88,7 +87,7 @@ def test_generic_fetcher(
 )
 def test_e2e_generic(
     test_params: utils.TestParameters,
-    check_cmd: List[str],
+    check_cmd: list[str],
```
eskultety, line 90: Is this commit (nr. 5) a new one? I don't remember seeing it. In any case, if you want to do this then I suggest doing it globally in a single commit for the whole code base rather than doing it in parts to reduce the noise of this change - i.e. not introducing any functional change.
TIL since Python 3.9 the builtin types (list, dict, tuple) can be used for type hinting directly, so I would imagine if we make this kind of change then the conversion would happen in a single commit and be done with it, otherwise it doesn't make much sense to me (I'm not advocating against the change in general to be clear, just the execution of the effort).

eskultety (approved):
ACK with the last commit dropped (unless you want to perform the change globally :) ).

tests/integration/test_generic.py
slimreaper35: Yes, the last two commits are new. (I moved them from the other branch here as it seems a more appropriate place and the PR hasn't got 2 approvals.)

> In any case, if you want to do this then I suggest doing it globally in a single commit for the whole code base rather than doing it in parts to reduce the noise of this change - i.e. not introducing any functional change.

I think I did it globally in one go. At least I don't see `List` except for:
```
tests/integration/utils.py:    # <class 'git.util.IterableList'>
tests/integration/utils.py:    #  Unsupported right operand type for in ("Callable[[], IterableList[Head]]")
```
I should remove the words "in tests" from the commit title.

eskultety: I meant for the other types as well (Dict & Tuple) ;)

slimreaper35: dropped the last commit

PR #784, opened 2025-01-13
Report Yarn v3/v4 patches as pedigree rather than components

taylormadore:
Yarn has a `patch` protocol for the [package patching feature](https://yarnpkg.com/features/patching). Patches should not be reported as independent Components in the SBOM, but should instead be reported as [pedigree](https://cyclonedx.org/docs/1.4/json/#components_items_pedigree_patches) for the patched "regular" package Component.

### DIFF
--- a/cachi2/core/models/sbom.py
+++ b/cachi2/core/models/sbom.py
@@ -27,6 +27,25 @@ class ExternalReference(pydantic.BaseModel):
     type: Literal["distribution"] = "distribution"
 
 
+class PatchDiff(pydantic.BaseModel):
+    """A Diff inside a Patch."""
+
+    url: str
+
+
+class Patch(pydantic.BaseModel):
+    """A Patch inside a SBOM Component Pedigree."""
+
+    type: Literal["backport", "cherry-pick", "monkey", "unofficial"] = "unofficial"
+    diff: PatchDiff
+
+
+class Pedigree(pydantic.BaseModel):
+    """A Pedigree inside a SBOM component."""
+
+    patches: list[Patch]
+
+
 FOUND_BY_CACHI2_PROPERTY: Property = Property(name="cachi2:found_by", value="cachi2")
 
 
@@ -45,6 +64,7 @@ class Component(pydantic.BaseModel):
     external_references: Optional[list[ExternalReference]] = pydantic.Field(
         serialization_alias="externalReferences", default=None
     )
+    pedigree: Optional[Pedigree] = None
 
     def key(self) -> str:
         """Uniquely identifies a package.
--- a/cachi2/core/package_managers/yarn/locators.py
+++ b/cachi2/core/package_managers/yarn/locators.py
@@ -205,10 +205,14 @@ def _parse_patch_locator(locator: "_ParsedLocator") -> PatchLocator:
 
     original_package = parse_locator(reference.source)
 
-    # https://github.com/yarnpkg/berry/blob/b6026842dfec4b012571b5982bb74420c7682a73/packages/plugin-patch/sources/patchUtils.ts#L92
     def process_patch_path(patch: str) -> Union[str, Path]:
-        # '~' denotes an optional patch (failing to apply the patch is not fatal, only a warning)
+        # Yarn patches can be optional, where failing to apply the patch is not fatal, only a warning
+        # '~' denotes an optional patch in Yarn v3
+        # https://github.com/yarnpkg/berry/blob/b6026842dfec4b012571b5982bb74420c7682a73/packages/plugin-patch/sources/patchUtils.ts#L92
         patch = patch.removeprefix("~")
+        # `optional!' denotes an optional patch in Yarn v4
+        # https://github.com/yarnpkg/berry/blob/93a56643ba3c813a87920dcf75c644eaf3b38e6f/packages/plugin-patch/sources/patchUtils.ts#L147
+        patch = patch.removeprefix("optional!")
         if re.match(r"^builtin<([^>]+)>$", patch):
             return patch
         else:
--- a/cachi2/core/package_managers/yarn/resolver.py
+++ b/cachi2/core/package_managers/yarn/resolver.py
@@ -7,18 +7,22 @@ processed.
 
 import json
 import logging
+import re
 import zipfile
+from collections import defaultdict
 from dataclasses import dataclass
 from functools import cached_property
 from pathlib import Path
 from textwrap import dedent
 from typing import TYPE_CHECKING, Any, Mapping, Union
+from urllib.parse import quote
 
 import pydantic
 from packageurl import PackageURL
+from semver import Version
 
 from cachi2.core.errors import PackageManagerError, PackageRejected, UnsupportedFeature
-from cachi2.core.models.sbom import Component
+from cachi2.core.models.sbom import Component, Patch, PatchDiff, Pedigree
 from cachi2.core.package_managers.yarn.locators import (
     FileLocator,
     HttpsLocator,
@@ -31,7 +35,7 @@ from cachi2.core.package_managers.yarn.locators import (
     parse_locator,
 )
 from cachi2.core.package_managers.yarn.project import Optional, Project
-from cachi2.core.package_managers.yarn.utils import run_yarn_cmd
+from cachi2.core.package_managers.yarn.utils import extract_yarn_version_from_env, run_yarn_cmd
 from cachi2.core.rooted_path import RootedPath
 from cachi2.core.scm import get_repo_id
 
@@ -42,6 +46,10 @@ if TYPE_CHECKING:
     from typing_extensions import assert_never
 
 log = logging.getLogger(__name__)
+
+COMPAT_PATCHES_SUBPATH = "packages/plugin-compat/sources/patches"
+COMPAT_PATCHES_REGEX = re.compile(r"builtin<compat/([^>]+)>")
+YARN_REPO_URL = "https://github.com/yarnpkg/berry"
 
 
 @dataclass(frozen=True)
@@ -165,8 +173,18 @@ def create_components(
     packages: list[Package], project: Project, output_dir: RootedPath
 ) -> list[Component]:
     """Create SBOM components for all the packages parsed from the 'yarn info' output."""
-    package_mapping = {package.parsed_locator: package for package in packages}
-    component_resolver = _ComponentResolver(package_mapping, project, output_dir)
+    package_mapping: dict[Locator, Package] = {}
+    patch_locators: list[PatchLocator] = []
+
+    # Patches are not components themselves, but they are necessary to
+    # resolve pedigree for their non-patch parent package components
+    for package in packages:
+        if isinstance(package.parsed_locator, PatchLocator):
+            patch_locators.append(package.parsed_locator)
+        else:
+            package_mapping[package.parsed_locator] = package
+
+    component_resolver = _ComponentResolver(package_mapping, patch_locators, project, output_dir)
     return [component_resolver.get_component(package) for package in package_mapping.values()]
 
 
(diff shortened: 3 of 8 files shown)


cachi2/core/models/sbom.py
```diff
@@ -28,6 +28,25 @@ class ExternalReference(pydantic.BaseModel):
     type: Literal["distribution"] = "distribution"
 
 
+class PatchDiff(pydantic.BaseModel):
```
taylormadore, line 31: The CycloneDX schema offers the option of a text diff for the patch, but it is not required. I opted not to include so as not to clutter the SBOM. Let me know if you disagree.

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -192,11 +202,41 @@ class _CouldNotResolve(ValueError):
...
         self._project = project
         self._output_dir = output_dir
         self._package_mapping = package_mapping
+        self._pedigree_mapping = self._get_pedigree_mapping(patch_locators)
+
+    def _get_pedigree_mapping(self, patch_locators: list[PatchLocator]) -> dict[Locator, Pedigree]:
+        """Map locators for dependencies that get patched to their Pedigree."""
+        pedigree_mapping: dict[Locator, Pedigree] = {}
+        for patch_locator in patch_locators:
+            # Filter out builtin patches
```
taylormadore, line 220: Builtin patches are applied by Yarn itself and tied to a specific version of Yarn. They are used to make certain features of the package manager work correctly: [example](https://yarnpkg.com/getting-started/qa/#why-is-typescript-patched-even-if-i-dont-use-plugnplay).

Do we want to report these?

cachi2/core/package_managers/yarn/locators.py
```diff
@@ -207,8 +207,8 @@ def _parse_patch_locator(locator: "_ParsedLocator") -> PatchLocator:
 
     # https://github.com/yarnpkg/berry/blob/b6026842dfec4b012571b5982bb74420c7682a73/packages/plugin-patch/sources/patchUtils.ts#L92
     def process_patch_path(patch: str) -> Union[str, Path]:
-        # '~' denotes an optional patch (failing to apply the patch is not fatal, only a warning)
-        patch = patch.removeprefix("~")
+        # '~' or `optional!' denotes an optional patch (failing to apply the patch is not fatal, only a warning)
```
eskultety, line 210: **Commit message:**
Can you please also include a reference to the docs or yarn code?

eskultety: nitpick: As for the commentary, maybe splitting the line in 2 and making it clearer that `~` is specifically for v3 and `optional!` additionally for v4 would IMO make it easier to read.

cachi2/core/models/sbom.py
eskultety: We now have a precedent :) in https://github.com/containerbuildsystem/cachi2/blob/38f8cd9d79669c60edc134d96ef0c69822d371db/cachi2/core/models/input.py#L217, so we could enable this the same way eventually.

cachi2/core/models/sbom.py
```diff
@@ -28,6 +28,25 @@ class ExternalReference(pydantic.BaseModel):
...
+    """A Diff inside a Patch."""
+
+    url: str
+
+
+class Patch(pydantic.BaseModel):
+    """A Patch inside a SBOM Component Pedigree."""
+
+    type: Literal["backport", "cherry-pick", "monkey", "unofficial"] = "unofficial"
```
eskultety, line 40: How can we determine which of the types: `unofficial`,`cherry-pick`, `backport` is the right one in order to improve or rather retain and not hinder the SBOM's accuracy?
I'm not questioning the use of Pedigree to record patch information, I'm just wondering whether by always reporting "unofficial" we're not going to cause more troubles for ourselves since I don't see a way for us to be able to detect it, it's a shame that the schema requires the type.
Can the choice of unofficial be recorded in an ADR at least? Because I feel like this is something worth recording from architectural POV.

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -18,7 +18,7 @@
 from packageurl import PackageURL
 
 from cachi2.core.errors import PackageManagerError, PackageRejected, UnsupportedFeature
-from cachi2.core.models.sbom import Component
```
eskultety, line 21: **Commit msg nitpick:**
I know you have it in the previous one, but it doesn't hurt to put the same cyclonedx URL here for reference as well. The reason for that being that one doesn't miss it in context of this PR, but the situation would be different when looking at the git history where you don't have this logical envelope.

cachi2/core/package_managers/yarn/resolver.py
eskultety: It would make our code appear "more" functional from SBOM perspective because we could report these patches as "cherry-pick" and therefore claim that we have "some" way of detecting the right type for the patch :D. In all seriousness, I think it would be more factually correct if we did, but based on the way they document it I'd say it all depends on how transparent these are to the user, IOW can we see the patch in the lockfile? Seems like we can, so we should report it as such IMO.

brunoapimentel: My first impression is that there's no harm in reporting them, but then, `unnoficial` seems an incorrect way to report them in the sbom, an apparently there's no better term for it in CycloneDx.

eskultety (commented):
Since this is a WIP, I approached this as a first round of reviews, so not very thorough. That said, it is pretty straight forward and makes sense, so you should flip the Draft switch :) .

cachi2/core/package_managers/yarn/resolver.py
brunoapimentel: So, here's the explanation: https://cyclonedx.org/docs/1.5/json/#components_items_pedigree_patches_items_type

I think we can keep `unnoficial` for user-added patches, and maybe `monkey` for official Yarn patches? I can't find any better way to express this, the choices in CycloneDX are very limiting.

cachi2/core/models/sbom.py
brunoapimentel: As I commented in the other [thread](https://github.com/containerbuildsystem/cachi2/pull/784#discussion_r1920169529), I think we can do a best-effort to assume that all patches added by the user (that is, that are present in the `package.json` file) are unnoficial. I think there'll be rare cases in which the maintainer of the library releases a patch to be used, if any.

We could see Yarn patches as unnoficial (since they are not written by the maintainers of the library being patched), but I'd like to differentiate them from user-added patches, since they are automatically added by Yarn, and are "official" if considering Yarn as the owner of the patch. Or, let's just omit them altogether, although I think the info could be somewhat useful (they do show up in `yarn.lock`).

brunoapimentel: I don't think we need to go as formal as an ADR, let's just record the decision in a comment and the commit message, should be enough.

brunoapimentel (commented):
Overall, this LGTM. I think we can mark it as ready to review, I just need to take a better look at tests before giving the final ack.

cachi2/core/package_managers/yarn/resolver.py
taylormadore: Just an FYI, for builtin patches we're probably going to need to clone the yarn (berry) repo and figure out a [.ts file](https://github.com/yarnpkg/berry/tree/master/packages/plugin-compat/sources/patches) there to point to rather than a standard diff, since the patches don't exist in the user repo and the URL for them does not appear in yarn info/lock.

eskultety: > So, here's the explanation: https://cyclonedx.org/docs/1.5/json/#components_items_pedigree_patches_items_type
> 
> I think we can keep `unnoficial` for user-added patches, and maybe `monkey` for official Yarn patches? I can't find any better way to express this, the choices in CycloneDX are very limiting.

`monkey` doesn't IMO describe them well - these are official patches that are applied on top of Yarn by Yarn itself and they carry those patches over to newer releases ---> this is a definition of cherry picking.

eskultety: > Just an FYI, for builtin patches we're probably going to need to clone the yarn (berry) repo and figure out a [.ts file](https://github.com/yarnpkg/berry/tree/master/packages/plugin-compat/sources/patches) there to point to rather than a standard diff, since the patches don't exist in the user repo and the URL for them does not appear in yarn info/lock.

That's unfortunate, but if our goal is to accurately record this in the SBOM, I guess we need to something like that. What ref/branch would we be cloning in that case? Always "main"? Or the latest `vX` tag?

I'm also wondering if fetching the official Yarn `vX` tarball, unpacking it and walking the hierarchy to identify the path wouldn't be slightly cheaper AND/OR faster than git operations on blobless shallow git clones.

cachi2/core/models/sbom.py
eskultety: What is the purpose of ADRs then? There clearly isn't one standard way on how one goes about recording patches in an SBOM, especially since there's a mandatory field we're not quite sure how to properly fill in based on the information we have.

cachi2/core/package_managers/yarn/resolver.py
brunoapimentel: > these are official patches that are applied on top of Yarn by Yarn itself and they carry those patches over to newer releases

They applied by Yarn on top of a few libraries, like Typescript ([e.g.](https://github.com/cachito-testing/cachi2-yarn-berry/blob/regular-installs/yarn.lock#L1331))

taylormadore: Why wouldn't builtin patches be "unofficial"? If yarn patches Typescript, it may be an official patch in the context of Yarn, but it almost certainly isn't in the context of Typescript, which is the Component being reported.

brunoapimentel: Well, my reasoning came from wanting to set them apart, but on the other hand, the purls will still carry the `~builtin` qualifier, right? I'm fine to just make them `unofficial`.

cachi2/core/package_managers/yarn/resolver.py
taylormadore: Done

cachi2/core/package_managers/yarn/locators.py
taylormadore: Done

cachi2/core/models/sbom.py
taylormadore: An ADR feels like overkill for a relatively minor decision like this, but :man_shrugging: 

There are two threads with discussion about the patch type, so I'm not sure which to use. My opinion is here: https://github.com/containerbuildsystem/cachi2/pull/784#discussion_r1922571758

brunoapimentel: My final opinion is in the same thread. Let's make it simply "unofficial".

cachi2/core/package_managers/yarn/resolver.py
eskultety: > Why wouldn't builtin patches be "unofficial"? If yarn patches Typescript, it may be an official patch in the context of Yarn, but it almost certainly isn't in the context of Typescript, which is the Component being reported.

But Typescript is Yarn's own dependency, not user's one, correct? IOW It is Yarn who needs it not user's app, doesn't it? If I'm wrong in this base assumption you can just stop reading here :), no need to continue the discussion.

Otherwise, it is Yarn's own patch applied on top of Yarn code base which doesn't sound unofficial to me.

In any case, I don't dwell too much on whether this will be unofficial or not in the SBOM, but there must be a clear record on why we think it is the case, it cannot sound just like an arbitrary decision.

cachi2/core/models/sbom.py
eskultety: > An ADR feels like overkill for a relatively minor decision like this, but 🤷‍♂️

Fair enough, although this isn't an internal implementation detail because it has an undocumented visible impact on users in form of our SBOM. Please just make sure that it's described accordingly at least in the commit msg, ideally with a tiny commentary in the code accompanying the change.

cachi2/core/package_managers/yarn/resolver.py
taylormadore: > But Typescript is Yarn's own dependency, not user's one, correct?

Typescript would be a dependency of the user project. The builtin patches are only applied if the dependency they are targeted to patch is included

eskultety: In which case my argument is invalid -> "unofficial" it is, sorry for the noise and thank you for the clarification.

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -192,11 +209,50 @@ class _CouldNotResolve(ValueError):
...
+        output_dir: RootedPath,
     ) -> None:
         self._project = project
         self._output_dir = output_dir
         self._package_mapping = package_mapping
+        self._pedigree_mapping = self._get_pedigree_mapping(patch_locators)
+
+    def _get_pedigree_mapping(self, patch_locators: list[PatchLocator]) -> dict[Locator, Pedigree]:
+        """Map locators for dependencies that get patched to their Pedigree."""
+        pedigree_mapping: dict[Locator, Pedigree] = {}
```
eskultety, line 225: I was recently correctly pointed out to consider `defaultdict` in a different scenario. Would `defaultdict(Pedigree(patches=[]))` be applicable here in this case as well?

brunoapimentel (approved):

cachi2/core/package_managers/yarn/resolver.py
```diff
@@ -412,3 +450,33 @@ def _cache_path_as_rooted(self, cache_path: str) -> RootedPath:
...
+
+    def _get_builtin_patch_url(self, patch: str, yarn_version: Version) -> str:
+        """Return a PURL-style VCS URL qualifier with subpath for a builtin Patch."""
+        match = re.match(COMPAT_PATCHES_REGEX, patch)
+        if not match:
+            raise UnsupportedFeature(f"{patch} is not a builtin patch from plugin-compat")
+
+        patch_filename = f"{match.group(1)}.patch.ts"
+        patch_subpath = Path(COMPAT_PATCHES_SUBPATH, patch_filename)
+        yarn_cli_tag = quote(f"@yarnpkg/cli/{yarn_version}")
```
eskultety, line 480: Sure Yarn, why wouldn't you put a special character into a git tag name /o\. I stumbled upon this and spent a few minutes verifying the VCS URL as it crashed my brain momentarily. Nitpick - maybe naming the variable `yarn_git_tag` would have alleviated the problem by just a tiny bit.

PR #789, opened 2025-01-20
Consolidate the integration test repos to a single one

slimreaper35:


tests/integration/utils.py
```diff
@@ -328,37 +323,50 @@ def fetch_deps_and_check_output(
...
-    :return: Path to output folder with fetched dependencies and output.json
+    :return: None
     """
-    output_folder = tmp_path.joinpath(f"{test_case}-output")
+    repo = Repo(test_repo_dir)
+    repo.git.reset("--hard")
+    repo.git.clean("-fdx")
+
+    node_modules = test_repo_dir.joinpath("node_modules")
+    shutil.rmtree(node_modules, ignore_errors=True)
```
slimreaper35, line 348: Why is `git reset` and `git clean` not sufficient here ?

tests/integration/utils.py
```diff
@@ -118,8 +118,8 @@ def __exit__(self, exc_type: Any, exc_value: Any, exc_traceback: Any) -> None:
 class Cachi2Image(ContainerImage):
     def run_cmd_on_image(
         self,
-        cmd: List,
-        tmpdir: StrPath,
+        cmd: list[Any],
```
a-ovchinnikov, line 123: [non-blocker][nitpick]
I don't think a `cmd` is a list of Any, usually commands are lists of strings (this does not really matter that much in test utils though).

tests/integration/utils.py
```diff
@@ -336,7 +336,7 @@ def fetch_deps_and_check_output(
     """
     Fetch dependencies for source repo and check expected output.
 
-    :param tmp_path: Temp directory for pytest
+    :param tmp_path: pytest fixture
```
a-ovchinnikov, line 334: [optional][non-blocking]
s/pytest fixture/pytest fixture pointing to Cachi2 temporary directory/

tests/integration/test_data/bundler_everything_present/container/Containerfile
```diff
@@ -1,19 +1,5 @@
 FROM docker.io/ruby:3.3
 
-# Test disabled network access
-RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
```
a-ovchinnikov, line 4: I would have kept this intact (just in case).

tests/integration/test_npm.py
```diff
@@ -73,7 +70,7 @@ def test_npm_packages(
         pytest.param(
             utils.TestParameters(
                 repo="https://github.com/cachito-testing/npm-cachi2-smoketest.git",
```
a-ovchinnikov: ~I tried searching for `npm/smoketes-lockfile3` branch in this repo and could not find it. Is this the right repository?~ _Edit_: maybe it would make sense to squash the last commit removing this parameter with this one?

tests/integration/utils.py
slimreaper35: oh thanks for that, I squashed this commit so I don't modify those lines twice

tests/integration/test_data/bundler_everything_present/container/Containerfile
slimreaper35: I can do that, but it is not something any user would have in their Dockerfile.

tests/integration/test_npm.py
slimreaper35: to squash last two commits ?

tests/integration/test_data/bundler_everything_present/container/Containerfile
a-ovchinnikov: Yes, users don't have it, however this is a test container where we verify a number of conditions. One of the conditions is hermeticity. If (for any reason) this container gains network access (maybe a typo in some other set-up script) and we don't test for it and also there is a new bug in say, Bundler, which makes it want to talk to the web again then we won't know of it until a bug report. I think it is worth it to be explicit about all assumptions that we make.

tests/integration/test_npm.py
a-ovchinnikov: No, first to rearrange commits with ff5f385 right after dab19e0 and then squash. I.e. in a rebase it would look like this:
```
pick dab19e0
squash ff5f385
```

It would be a trail of leftover and unused parameter otherwise, someone else might get confused.

slimreaper35: done

tests/integration/test_bundler.py
```diff
@@ -13,7 +13,7 @@
     [
         pytest.param(
             utils.TestParameters(
-                ref="bundler/missing-gemfile",
```
eskultety: This commit should be squashed into the one consolidating everything into a single repo.

tests/integration/test_rpm.py
```diff
@@ -183,27 +172,24 @@ def test_repo_files(
...
     request: pytest.FixtureRequest,
 ) -> None:
     """Test if the contents of the generated .repo file are correct."""
     test_case = request.node.callspec.id
-    output_folder = tmp_path.joinpath(f"{test_case}-output")
-
-    source_folder = utils.clone_repository(
-        test_params.repo, test_params.ref, f"{test_case}-source", tmp_path
-    )
+    output_dir = tmp_path.joinpath(utils.DEFAULT_OUTPUT_DIR)
```
eskultety, line 181: Okay, makes sense, but please do this in a separate commit, this change has nothing to do with the consolidation on its own, the diff is already massive and hence hard to follow even without it.

tests/integration/test_pip.py
```diff
@@ -195,14 +181,13 @@ def test_pip_packages(
                 expected_exit_code=0,
                 expected_output="All dependencies fetched successfully",
             ),
-            ["python3", "/opt/test_package_cachi2"],
+            ["python3", "/app/src/test_package_cachi2/main.py"],
```
eskultety, line 184: What's this path change related to?

tests/integration/utils.py
```diff
@@ -60,10 +60,12 @@
     "https://raw.githubusercontent.com/CycloneDX/specification/1.4/schema/bom-1.4.schema.json"
 )
 
+DEFAULT_OUTPUT_DIR = Path("cachi2-output")
```
eskultety, line 63: You should pull this one from `cachi2/interface/cli.py` so that we won't have to change it at 2 places soon ;)

tests/integration/utils.py
```diff
@@ -60,10 +60,12 @@
...
+DEFAULT_ENV_FILE = Path("cachi2.env")
```
eskultety, line 64: We're only ever going to use this in a single test primitive related to the `generate-env` command, so we can skip the constant and hardcode the path in the invocation, it's fine.

tests/integration/utils.py
```diff
@@ -94,12 +96,12 @@ def pull_image(self) -> None:
 
     def run_cmd_on_image(
         self,
-        cmd: List,
-        tmpdir: StrPath,
+        cmd: list[str],
+        tmp_path: Path,
```
eskultety, line 100: This hunk (multiple occurrences) seems like a non-related cosmetic change.

tests/integration/utils.py
eskultety: That's because `node_modules` contain a nested git repository which git by default ignores when cleaning. From git-clean man page:

> -f, --force
           If the Git configuration variable clean.requireForce is not set to false, git clean will
           refuse to delete files or directories unless given -f. **Git will refuse to modify untracked
           nested git repositories (directories with a .git subdirectory) unless a second -f is given.**

tests/integration/utils.py
```diff
@@ -431,55 +437,71 @@ def build_image_and_check_cmd(
...
+def _replace_tmp_path_with_placeholder(
+    project_files: list[dict[str, str]], test_repo_dir: Path
+) -> None:
     for item in project_files:
-        relative_path = item["abspath"].replace(str(tmp_path), "")
-        item["abspath"] = "${test_case_tmpdir}" + str(relative_path)
+        if "bundler" in item["abspath"]:
+            # special case for bundler, as it is not a real project file
+            item["abspath"] = "${test_case_tmp_path}/cachi2-output/bundler/config_override/config"
+            continue
```
eskultety, line 501: This seems like fixing a pre-existing problem not inherently tied to the consolidation, is that so? I think if you want to fix whatever it is then it should gain visibility in a separate commit, in this consolidation huge one it's easily lost, I almost overlooked the hunk completely.

tests/integration/utils.py
eskultety: Please mention the double `-f` in a commentary in the module so that some poor soul doesn't try to remove it as a typo in the future (even though it would fail the tests).

eskultety (commented):
Not only did this consolidation clean up the mess we have in integration tests (some bits still remaining for a follow up), but by doing so it sped up the execution by at least 13%! (measured solely on the NPM test suite, so extrapolating it to the whole test suite where there are many more test cases may yield even better results).

I left some comments which should be simple to address and we're good to go.

tests/integration/test_data/npm_lockfile3_aliased_deps/container/Containerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:18
+FROM docker.io/node:18
```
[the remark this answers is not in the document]
slimreaper35: Is this a candidate for a follow-up issue?

tests/integration/test_npm.py
```diff
@@ -61,252 +56,66 @@ def test_npm_smoketest(
...
-            ],
+            [],
+            [],
             id="npm_lockfile3_aliased_deps",
         ),
         pytest.param(
             utils.TestParameters(
-                repo="https://github.com/cachito-testing/npm-dev-optional-peer-deps.git",
-                ref="77018bf73295ef1248d24479da897d960576f933",
+                branch="77018bf73295ef1248d24479da897d960576f933",
```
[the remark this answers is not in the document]
slimreaper35: Ah, I forgot to change this :face_exhaling:

tests/integration/test_data/npm_lockfile3_aliased_deps/container/Containerfile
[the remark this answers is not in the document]
eskultety: No. It's up to you whether you want to fix it or not. I raised this only because it seemed like you'd be pushing at least one more revision anyway, so might as well address it if you feel like it, but I had to acknowledge that it's not a functional request since what you have works (then again, either one approach is a cosmetic change only, so there goes :) ).

tests/integration/utils.py
slimreaper35: > This seems like fixing a pre-existing problem not inherently tied to the consolidation, is that so? 

No. Previously each test was tied to its own test directory with the source code and cachi2 output. Now, the source code lives only in one repo and the cachi2 output is inside `tmp_path` fixture.

tests/integration/test_pip.py
slimreaper35: https://github.com/containerbuildsystem/cachi2/pull/789/files#diff-d39fbd1e984fcf660f7f0e15d3272e06ae3571cc48014872d6cbec95f3aee7b8L19

simplified copy of the source code to different location

slimreaper35:
I tried to resolve the majority of the comments. I did not rebase, so the GitHub diff (Compare button) should provide all modified lines only by me.

tests/integration/test_rpm.py
slimreaper35: what do you mean exactly here ?

a-ovchinnikov (approved):

tests/integration/test_data/gomod_1.18_e2e_test/bom.json
```diff
@@ -572,9 +572,9 @@
           "value": "cachi2"
         }
       ],
-      "purl": "pkg:golang/github.com/cachito-testing/retrodep/retrodep/glide@v2.1.2-0.20230109085255-c3496edd5d45?type=package",
+      "purl": "pkg:golang/github.com/cachito-testing/integration-tests/retrodep/glide@v2.0.0-20230109085255-c3496edd5d45?type=package",
       "type": "library",
-      "version": "v2.1.2-0.20230109085255-c3496edd5d45"
+      "version": "v2.0.0-20230109085255-c3496edd5d45"
```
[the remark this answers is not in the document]
slimreaper35: I don't know honestly. I just run `nox -s generate-test-data`. Isn't it related to your golang PR ?
https://github.com/containerbuildsystem/cachi2/pull/553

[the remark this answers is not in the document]
eskultety: No, it didn't touch this file. This change was added by commit 4b06bd0f6 without providing further context on where the information came from.

tests/integration/utils.py
eskultety: This hasn't been addressed. Again, we're close to merging so let's not make a deal out of this, but please make sure you perform cosmetic changes like these (if at all) in a separate commit to avoid polluting a diff with unrelated changes, it just prolongs the review itself.

eskultety:
@slimreaper35 seems like an update to `yarn_v4` integration test data got lost somehow (I don't remember the tests failing originally).

slimreaper35:
> @slimreaper35 seems like an update to yarn_v4 integration test data got lost somehow (I don't remember the tests failing originally).

I had to update test data because of the recent changes to yarn pedigree patches - https://github.com/containerbuildsystem/cachi2/commit/db6e6d90533566e763873204fdbc92feba2d7b12

Issue #712, opened 2024-10-25
Cachi2 prefetching is making changes on a source


eskultety:
We only use a working copy of the input source repository for the Yarn backends. All other package manager backends work on top of the input source directly which may result in some operations we perform dirtying the repo which is completely undesirable.
Unlike https://github.com/containerbuildsystem/cachi2/issues/707 the global fix isn't as straight forward since the main issue only pops up with integration tests which also compare the contents of our `.build-config.json` file where we fill out incorrect path data for all `project_files`, using the temporary working copy path instead of the actual input repo.

We need to make sure that while all operations are performed on top of the working copy, any generated build-config data are tied to the original input repo path as the temporary working copy is discarded by the time both `.build-config.json` and SBOM are dumped to the disk. 

Extends: https://github.com/containerbuildsystem/cachi2/issues/707

eskultety:
We'll need to pass around 2 paths in the request blob - one for the working copy for all operations and one for the original repository path for logging and formatting purposes.

PR #790, opened 2025-01-20
Fix project files absolute paths when a work copy is used

### LINKED ISSUE #785: Using yarn will cause other package managers that rely on project files to fail to build
Since we process the yarn request on a [copy](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/resolver.py#L40-L43) of the sources, any project file that gets generated (e.g. [npm](https://github.com/containerbuildsystem/cachi2/blob/main/cachi2/core/package_managers/npm.py#L634)) will have its abspath set to the copy path's, and will cause the build to fail.

*How to reproduce*:
Prefetch and build [this branch](https://github.com/brunoapimentel/cachi2-experiments/tree/work-copy-bug).
Prefetching only `npm` will cause the [npm install](https://github.com/brunoapimentel/cachi2-experiments/blob/work-copy-bug/Dockerfile#L9) clause succeed, whereas when we add both `yarn` and `npm`, it fails.

Relates to: #712

brunoapimentel:
To avoid unintended source code modifications, Cachi2 makes a copy of the source directory when some package managers are used (currently, it's being done just for yarn).

This creates a problem with project files, since the absolute path of files now is pointing to the temporary working copy. The solution introced by this patch reverts these paths back to the original source dir path.

This is intended to be a quick fix, and should only be kept until we solve #712.

Resolves #785.

### DIFF
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -48,6 +48,13 @@ def resolve_packages(request: Request) -> RequestOutput:
             output = _resolve_packages(request)
             request.source_dir = original_source_dir
 
+            # Temporary solution to project files paths that are pointing to the work copy.
+            # Should be replaced once we extend the work copy solution to other package managers.
+            # https://github.com/containerbuildsystem/cachi2/issues/712
+            for project_file in output.build_config.project_files:
+                subpath = project_file.abspath.relative_to(source_backup)
+                project_file.abspath = original_source_dir / subpath
+
             return output
     else:
         return _resolve_packages(request)
--- a/tests/unit/test_resolver.py
+++ b/tests/unit/test_resolver.py
@@ -120,7 +120,7 @@ def test_source_dir_copy(
         packages=packages,
     )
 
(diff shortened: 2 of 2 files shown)


taylormadore (approved):
This seems reasonable as a short-term fix to unblock users until the larger issue is resolved

cachi2/core/resolver.py
```diff
@@ -48,6 +48,12 @@ def resolve_packages(request: Request) -> RequestOutput:
             output = _resolve_packages(request)
             request.source_dir = original_source_dir
 
+            # Temporary solution to project files paths that are pointing to the work copy.
```
eskultety, line 51: **commit message nitpick:**

> solve #712¹.

no need for a footer, just use the URL directly and GH will render it just fine.

a-ovchinnikov (approved):
LGTM as a temporary solution (with a minor request).

eskultety (approved):

slimreaper35 (approved):

PR #793, opened 2025-01-24
chore: Applying underscores decision

a-ovchinnikov:
And recording it in contributor's guidelines.


CONTRIBUTING.md
```diff
@@ -175,6 +175,10 @@ If your code was inspired by any third-party sources, consider adding a comment
 
 When extending an existing feature, please add a new test case instead of modifying any existing ones. Large test scenarios with many branching paths are very hard to understand and to maintain. It is ok to copy and paste large parts of an existing test if needed for a new scenario. It is also fine to add a new parameter group to an existing test, as long as the test function remains unchanged.
 
+When naming tests please make sure that prefix `test` is followed by just _one_ underscore
+even when testing methods that start with an underscore (two undrescores are acceptable when
+testing methods whose names start with two underscores).
+
```
eskultety, line 181: I thought this was supposed to be just an agreement among the core members so as not to point it out again during reviews. I'm not convinced this is something we should have in the guidelines as a strict expectation from external contributors (I'd at least drop the part in the parentheses as that isn't IMO helping to keep things consistent).

Since the second part of this patch is fixing the inconsistency then I'd modify ^this suggestion to a more generic one: 

_"Please stick to naming which is consistent with our existing function/test names. There's a lot of precedent to follow in our code base. As with anything, despite the best of efforts we acknowledge that there always will be some inconsistencies across the code base which we have missed during reviews. Feel free to fix them along the way if you happen to stumble upon them and help us on improving the status quo."_

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -119,7 +119,7 @@ def test_get_main_package_name_and_version_from_repo_without_origin(
     assert "Failed to extract package name from origin remote" in exc_info.value.friendly_msg()
 
 
-def test__prepare_for_hermetic_build_injects_necessary_variable_into_empty_config(
```
eskultety, line 122: I know these changes correspond to your guideline change, but I think the easiest is to just use the same scheme everywhere, because after this patch there are still the following violators (yes, mine, from a recent golang changes):
```
tests/unit/package_managers/test_gomod.py

test__init__
test__init__fail
test__bool__
```

CONTRIBUTING.md
a-ovchinnikov: I'll just drop it then. Maybe this should be expressed with a flake8 plugin at some point.

tests/unit/package_managers/bundler/test_main.py
a-ovchinnikov: Yes, I have left those intact intentionally, because a function name starts with two underscores (in my naming scheme a hypothetical test for any these three would have started with triple underscore).

eskultety: I understand, but if this patch is to have any tangible meaning then we should apply it globally, no exceptions, pure consistency and be done with it :) .

a-ovchinnikov: Fair enough.

tests/unit/package_managers/test_gomod.py
```diff
@@ -2455,7 +2455,7 @@ class TestGoWork:
         ],
     )
     @mock.patch("cachi2.core.package_managers.gomod.run_cmd")
-    def test__init__(
+    def test_init__(
```
eskultety, line 2458: `test_init__` - I would hope that the trailing underscores would be gone too :), looks weird otherwise.

eskultety (approved):

tests/unit/package_managers/test_gomod.py
a-ovchinnikov: Done.

slimreaper35 (approved):

PR #802, opened 2025-01-29
Some changes to make local IT runs more stable

a-ovchinnikov:
The PR adds several retries to functions which tend to sporadically fail locally for no good reason.

It also addresses several warnings which pollute UTs output. It does not handle
`coroutine 'AsyncMockMixin._execute_mock_call' was never awaited` which is very annoying, but has to be chased with specially trained dogs.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -138,6 +138,11 @@ def local_dnfserver(top_level_test_dir: Path) -> Iterator[None]:
             try:
                 _check_ssl_configuration()
                 break
+            except requests.ConnectionError:
+                # ConnectionResetError is often reported locally, waiting it over
+                # helps.
+                log.info("Failed to connect to the DNF server, retrying...")
+                continue
             except requests.RequestException as e:
                 raise RuntimeError(e)
         else:
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,3 +1,4 @@
+import sys
 import tarfile
 from pathlib import Path
 from typing import Generator
@@ -20,7 +21,10 @@ def data_dir() -> Path:
 def golang_repo_path(data_dir: Path, tmp_path: Path) -> Path:
     """Return extracted Golang git repository inside a temporary directory."""
     with tarfile.open(data_dir / "golang_git_repo.tar.gz") as tar:
-        tar.extractall(tmp_path)
+        if sys.version_info >= (3, 12):
+            tar.extractall(tmp_path, filter="fully_trusted")
+        else:
+            tar.extractall(tmp_path)
 
     return tmp_path / "golang_git_repo"
 
--- a/tests/unit/models/test_sbom.py
+++ b/tests/unit/models/test_sbom.py
@@ -559,12 +559,14 @@ DEFAULT_ROOT_PACKAGE = SPDXPackage(
     }
 )
 
-DEFAULT_ROOT_RELATION = {
-    "spdxElementId": "SPDXRef-DOCUMENT",
-    "comment": "",
-    "relatedSpdxElement": "SPDXRef-DocumentRoot-File-",
-    "relationshipType": "DESCRIBES",
-}
+DEFAULT_ROOT_RELATION = SPDXRelation(
+    **{
+        "spdxElementId": "SPDXRef-DOCUMENT",
+        "comment": "",
+        "relatedSpdxElement": "SPDXRef-DocumentRoot-File-",
+        "relationshipType": "DESCRIBES",
+    }
+)
 
 
 def _gen_ref(locator: str) -> dict:
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import shutil
+import sys
 import tempfile
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -285,7 +286,12 @@ def _safe_extract(tar: TarFile, path: str = ".", *, numeric_owner: bool = False)
         if not abs_member_path.is_relative_to(abs_path):
             raise ExtractError("Attempted Path Traversal in Tar File")
 
-    tar.extractall(path, numeric_owner=numeric_owner)
+    # This 'if' block is to deal with deprectaion warning for unfiltered tar
+    # extraction in 3.12.
+    if sys.version_info >= (3, 12):
+        tar.extractall(path, numeric_owner=numeric_owner, filter="fully_trusted")
+    else:
+        tar.extractall(path, numeric_owner=numeric_owner)
 
 
 def _json_serialize(data: dict[str, Any]) -> str:
--- a/tests/unit/test_scm.py
+++ b/tests/unit/test_scm.py
@@ -1,4 +1,5 @@
 import filecmp
+import sys
 import tarfile
 from pathlib import Path
 from typing import Union
@@ -76,7 +77,10 @@ def test_clone_as_tarball(golang_repo_path: Path, tmp_path: Path) -> None:
     clone_as_tarball(f"file://{original_path}", INITIAL_COMMIT, to_path)
 
     with tarfile.open(to_path) as tar:
-        tar.extractall(tmp_path / "my-repo")
+        if sys.version_info >= (3, 12):
+            tar.extractall(tmp_path / "my-repo", filter="fully_trusted")
+        else:
+            tar.extractall(tmp_path / "my-repo")
 
     my_path = tmp_path / "my-repo" / "app"
 


cachi2/core/package_managers/gomod.py
```diff
@@ -1247,6 +1257,13 @@ def __init__(self, repo: git.Repo, commit: git.objects.commit.Commit):
         self._commit = commit
 
     @classmethod
+    @backoff.on_exception(
+        fast_constant_delay,
+        Exception,
+        jitter=None,  # use deterministic backoff, do not apply jitter
```
eskultety, line 1263: I'm wondering why we concluded that jitter is bad (originally appeared first in commit 9d6f0603c migrating the logic from cachito).

cachi2/core/package_managers/gomod.py
```diff
@@ -1247,6 +1257,13 @@ def __init__(self, repo: git.Repo, commit: git.objects.commit.Commit):
...
+        max_tries=5,
```
eskultety, line 1264: This number is repeated multiple times across the patch, so I think it deserves a generic constant.

cachi2/core/package_managers/gomod.py
```diff
@@ -392,7 +395,7 @@ def _retry(self, cmd: list[str], **kwargs: Any) -> str:
         n_tries = get_config().gomod_download_max_tries
 
         @backoff.on_exception(
-            backoff.expo,
+            fast_constant_delay,
```
eskultety, line 398: What's the reasoning behind dropping exponential? Why is it deemed bad in our case? I'm genuinely curious because expo seemed to me to be the "industry" standard when it comes to dealing with networking responses.

cachi2/core/package_managers/gomod.py
```diff
@@ -410,6 +413,13 @@ def run_go(_cmd: list[str], **kwargs: Any) -> str:
             )
             raise PackageManagerError(err_msg) from None
 
+    @backoff.on_exception(
+        fast_constant_delay,
+        Exception,
+        jitter=None,
+        max_tries=5,
+        logger=log,
+    )
```
eskultety, line 422: This hunk is incorrect, historically, we executed Go in 2 different modes: the one nowadays represented by `_run` dealing with go commands that don't make use of any networking and the other represented by `_retry` which we use when we fetch packages mainly. We most certainly don't want failing local go executions that have nothing to do with networking to be executed again.

noxfile.py
```diff
@@ -107,7 +107,7 @@ def _run_integration_tests(session: Session, env: dict[str, str]) -> None:
     netrc = "machine 127.0.0.1 login cachi2-user password cachi2-pass"
     default_env = {"CACHI2_TEST_NETRC_CONTENT": os.getenv("CACHI2_TEST_NETRC_CONTENT", netrc)}
     default_env.update(env)
-    cmd = "pytest --log-cli-level=WARNING tests/integration"
+    cmd = "pytest --log-cli-level=WARNING tests/integration --maxfail=1"
```
eskultety, line 110: Haven't we had a discussion about this in the past in a different PR that failing fast might be sometimes counterproductive? I might be misremembering things.
This approach feels to me like computer games where you need to complete a task to get to the next checkpoint only to find out there's a failure waiting for you instead of revealing all failures at once. If the test output is bothersome then the test suite can be filtered to execute only on a subset of tests, but in general having an idea on how many failures one will need to deal with seems like a benefit rather than being something inferior.

tests/unit/package_managers/test_gomod.py
```diff
@@ -2215,8 +2215,7 @@ def test_retry(
         assert mock_sleep.call_count == tries_needed - 1
 
         for n in range(tries_needed - 1):
-            wait = 2**n
-            assert f"Backing off run_go(...) for {wait:.1f}s" in caplog.text
+            assert "Backing off _run(...) for " in caplog.text
```
eskultety, line 2218: Like I mentioned in a different comment, this is very much appreciated in tests, but sadly won't make the **actual** cachi2 log output problem coming from Backoff go away.

tests/unit/package_managers/test_gomod.py
```diff
@@ -2227,21 +2226,24 @@ def test_retry_failure(
         mock_config.return_value.gomod_download_max_tries = 5
 
         failure = subprocess.CalledProcessError(returncode=1, cmd="foo")
-        mock_run.side_effect = [failure] * 5
+        # _run is wrapped with another 5 retries, so another x5
+        mock_run.side_effect = [failure] * 5 * 5
```
eskultety, line 2230: This shouldn't be necessary once you drop backoff from `_run`.

tests/unit/models/test_sbom.py
```diff
@@ -559,12 +559,14 @@ def test_cyclonedx_sbom_can_be_converted_to_spdx_and_back_without_loosing_any_da
...
-    "relationshipType": "DESCRIBES",
-}
+DEFAULT_ROOT_RELATION = SPDXRelation(
+    **{
+        "spdxElementId": "SPDXRef-DOCUMENT",
+        "comment": "",
+        "relatedSpdxElement": "SPDXRef-DocumentRoot-File-",
+        "relationshipType": "DESCRIBES",
+    }
+)
```
eskultety, line 569: nitpick: I'm wondering if `SPDXRelation.from_dict()` wouldn't be more readable, but that's just me, I missed the unpack at first glance, feel free to ignore.

eskultety (commented):
Just a heads up - I looked at the backoff project and it very much seems like an abandonware: https://github.com/litl/backoff/issues/211 . We may or may not face an issue once we bump to 3.10/3.11 unless they figure out whether they want to fork the project or not.
[tenacity](https://github.com/jd/tenacity) seemed like a viable alternative not so long ago, but looking at their issues/PRs, seems to be in the same situation.

tests/unit/conftest.py
```diff
@@ -20,7 +21,10 @@
 def golang_repo_path(data_dir: Path, tmp_path: Path) -> Path:
     """Return extracted Golang git repository inside a temporary directory."""
     with tarfile.open(data_dir / "golang_git_repo.tar.gz") as tar:
-        tar.extractall(tmp_path)
+        if sys.version_info >= (3, 12):
+            tar.extractall(tmp_path, filter="fully_trusted")
```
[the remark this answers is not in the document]
a-ovchinnikov: It won't stop then. Here I default to pre-3.12 behavior which is, as CodeQL has dutifully noted, extremely dangerous!

tests/unit/models/test_sbom.py
a-ovchinnikov: `SPDXRelation` does not have `from_dict` at the moment, I'll keep it as expansion for now.

a-ovchinnikov (commented):
The current implementation of backoffs does not play well with UTs and solves a local problem. I will drop this commit in the next revision and will try addressing it separately at some other point.

For the record, I tried using `pytest-retry` and mark all tests as flaky (which, admittedly, is a wrong way of doing it), but that, in turn, did not play well with our fixtures.

tests/integration/conftest.py
```diff
@@ -138,6 +138,11 @@ def _check_ssl_configuration() -> None:
             try:
                 _check_ssl_configuration()
                 break
+            except requests.ConnectionError:
+                # ConnectionResetError is often reported locally, waiting it over
+                # helps.
```
slimreaper35, line 143: Did you mean `requests.ConnectionError` or `ConnectionResetError` is something different? (Also, the comment could be wrapped in one line :smile: )

slimreaper35 (approved):
Approved, the changes in the PR shrank quite a lot since the last time I checked on it.

tests/unit/conftest.py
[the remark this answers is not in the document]
eskultety: I meant when we support 3.12+ only which means we could start using the `filter` argument which is where it was introduced ;) -> resolving.

tests/unit/models/test_sbom.py
eskultety: I meant we could add one :) , it's simple, out of scope though.

cachi2/core/package_managers/gomod.py
eskultety: > Since we have effectively just one process I'd say we can safely rely on jitter occurring naturally.

This isn't true though for PaC services, where there are many concurrent pipelines running our tool as a standalone process each, potentially contacting the same service concurrently.

eskultety: > Initially I have implemented my own retry decorator (a much simpler one than that provided by backoff) only to realize that we have a dependency which I could reuse.

Well, you might as well propose it :) since backoff seems to have been abandoned, so that way, we could maintain our own simple implementation that would suit our needs just fine without additional cruft.

tests/unit/package_managers/test_gomod.py
eskultety: >  I would have dropped all string verifications because this functionality should be tested on backoff side, there is no good reason for re-testing someone else's code.

Agreed. "My first issue" material? :)

tests/unit/package_managers/test_gomod.py
eskultety: Since the commit was dropped for now -> resolving.

eskultety (approved):

cachi2/core/package_managers/gomod.py
eskultety: Since the commit was dropped in the meantime -> resolving.

cachi2/core/package_managers/gomod.py
a-ovchinnikov: > there are many concurrent pipelines running our tool

They still don't do this in sync. Usually artificial jitter is a small (on the order of tens percent maximum) deviation from certain value, and that's what we'll get when jobs are triggered randomly.

tests/integration/conftest.py
a-ovchinnikov: It is a `ConnectionError` which gets a field set to `ConnectionResetError`. I think it is safe to retry on any ConnectionError that happens locally.

PR #807, opened 2025-01-31
SPDX: make SPDXIDs pass validation + cosmetic improvement

chmeliik:
See commit messages for details

### DIFF
--- a/cachi2/core/models/sbom.py
+++ b/cachi2/core/models/sbom.py
@@ -2,6 +2,7 @@ import datetime
 import hashlib
 import json
 import logging
+import re
 from collections import defaultdict
 from functools import cached_property, partial, reduce
 from itertools import chain, groupby
@@ -115,6 +116,17 @@ def spdx_now() -> str:
     return datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
 
 
+def sanitize_spdxid(spdxid: str) -> str:
+    """Sanitize an SPDXID.
+
+    See https://spdx.github.io/spdx-spec/v2.3/package-information/#7.2:
+
+    Format  "SPDXRef-"[idstring]
+            where [idstring] is a unique string containing letters, numbers, ., and/or -.
+    """
+    return re.sub(r"[^0-9a-zA-Z\.\-]", "-", spdxid)
+
+
 class Sbom(pydantic.BaseModel):
     """Software bill of materials in the CycloneDX format.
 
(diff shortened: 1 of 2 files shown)


cachi2/core/models/sbom.py
```diff
@@ -184,6 +185,12 @@ def libs_to_packages(libraries: list[Component]) -> list[SPDXPackage]:
             args = dict(annotator=annottr, annotationDate=now, annotationType="OTHER")
             pAnnotation = partial(SPDXPackageAnnotation, **args)
 
+            def get_spdxid(component: Component) -> str:
+                spdxid = f"SPDXRef-Package-{component.name}-{component.version}-{package_hash}"
+                # According to a validation error from https://tools.spdx.org/app/validate/,
+                # the allowed format for SPDXID is SPDXRef-[0-9a-zA-Z\.\-\+]+
+                return re.sub(r"[^0-9a-zA-Z\.\-\+]", "-", spdxid)
```
eskultety: nitpick: I think the interface would be more ergonomic and more re-usable (do we need that?) if it were a blackbox accepting strings directly and only performed the transformation in accordance to the spec. It would require a minor change in commit 2/2, but that's a trivial change anyway.

cachi2/core/models/sbom.py
```diff
@@ -2,6 +2,7 @@
 import hashlib
 import json
 import logging
+import re
```
eskultety, line 5: **Commit message:**

> According to https://tools.spdx.org/app/validate/, SPDXID strings must

Would this be the actual docs to link with this commit in addition to the reference to the online validator? https://spdx.github.io/spdx-spec/v2.3/snippet-information/#921-description

eskultety (approved):
I think we can do slightly better from the interface POV for the helper, but it's non-blocking.

cachi2/core/models/sbom.py
chmeliik: Hmm, the spec seems to disagree with the validator

> containing letters, numbers, `.`, and/or `-`.

I'll exclude `+` as well to be sure

cachi2/core/models/sbom.py
chmeliik: Refactored as suggested

cachi2/core/models/sbom.py
chmeliik: Done and re-worded commit message

a-ovchinnikov (approved):

PR #808, opened 2025-01-31
Initial cargo support

slimreaper35:


cachi2/core/package_managers/cargo/main.py
```diff
@@ -0,0 +1,59 @@
...
+
+    return RequestOutput.from_obj_list(components, environment_variables, project_files)
+
+
+def _resolve_cargo_package(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+    """Resolve a single cargo package."""
+    vendor_dir = output_dir.join_within_root("deps/cargo")
+    cmd = ["cargo", "vendor", "--locked", str(vendor_dir)]
+    log.info("Vendoring crates at %s", package_dir)
+    run_cmd(cmd=cmd, params={"cwd": package_dir})
```
slimreaper35, line 36: One thing, that I don't understand is (for example) a scenario when `Cargo.lock` is not in the package directory. The command logically fails and we:

- print the actual command that failed :heavy_check_mark: 
- print STDERR :heavy_check_mark: 
- and then re-raise the `CalledProcessError` and pollute the whole screen :question:

a-ovchinnikov: Missing Cargo.lock looks like an improbable situation: it means that someone is trying to build a (deliberately) broken package as part of some larger build. We could perfrm a basic sanity check ourselves (i.e. just check that all necessary files are there) or just let cargo stderr propagate.

a-ovchinnikov (approved):
Straightforward basic Rust support. LGTM.

slimreaper35:
I had to add one more commit to update test data for gomod to pass the CI.

Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN dnf -y install \
     python3 \
     rubygem-bundler \
     rubygem-json \
+    cargo \
```
eskultety, line 18: This unfortunately bloats our image with a ton of direct cargo dependencies like GCC, CPP, rust standard libraries, etc. growing the image size ~3.5 fold:
```
========================================================================================
Install  35 Packages

Total download size: 157 M
Installed size: 534 M
```
from
```
=======================================================================================
Install  17 Packages

Total download size: 56 M
Installed size: 161 M
```

Sometimes this is unavoidable, but I'm not a fan of massive images (I don't think anyone is with a slowish connection) and so we should always strive to figure out a reasonable way of getting rid of anything unnecessary (that's why people love all alpine container image build spinoffs). That's why I never was aligned with the idea of us installing 2 Go SDKs and today we're actually very close to culling one of them and hence shrinking the image again by ~100MB.

In your particular case it should be very easy for us, because the cargo binary itself doesn't have many library requirements and all of them are already in place in the base UBI image:
```
# forgive some of the misleading packages which only think they *provide* the actual file path
$ ldd bin/cargo | awk '/=> \// {print $3}' | xargs rpm -q --path
openssl-libs-3.2.2-6.el9_5.x86_64
zlib-1.2.11-40.el9.x86_64
libcurl-minimal-7.76.1-31.el9.x86_64
libgcc-11.5.0-2.el9.x86_64
glibc-2.34-125.el9_5.1.x86_64
glibc-2.34-125.el9_5.1.x86_64
libnghttp2-1.43.0-6.el9.x86_64
krb5-libs-1.21.1-4.el9_5.x86_64
krb5-libs-1.21.1-4.el9_5.x86_64
krb5-libs-1.21.1-4.el9_5.x86_64
libcom_err-1.46.5-5.el9.x86_64
krb5-libs-1.21.1-4.el9_5.x86_64
keyutils-libs-1.6.3-1.el9.x86_64
glibc-2.34-125.el9_5.1.x86_64
libselinux-3.6-1.el9.x86_64
pcre2-10.40-6.el9.x86_64
```

Assuming our logic is only going to revolve around `cargo` invocations (integration tests needed), I just did
```diff
diff --git a/Dockerfile b/Dockerfile
index 929c6a85..d4762ecc 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,7 @@ COPY . .
 RUN dnf -y install \
     --setopt install_weak_deps=0 \
     --nodocs \
+    cargo \
     gcc \
     python3-devel \
     python3-pip \
@@ -50,6 +51,7 @@ COPY --from=golang_120 /usr/local/go /usr/local/go/go1.20
 COPY --from=golang_121 /usr/local/go /usr/local/go/go1.21
 COPY --from=node /usr/local/lib/node_modules/corepack /usr/local/lib/corepack
 COPY --from=node /usr/local/bin/node /usr/local/bin/node
+COPY --from=builder /usr/bin/cargo /usr/bin/cargo
 COPY --from=builder /venv /venv
```

and then tried to vendor a random project on GH: https://github.com/librespot-org/librespot  - no problem!
If you want this to be cleaner then another build stage just to install cargo would be the right thing to start with.

Please give it a try and consider not installing cargo along with every Rust development dependency it lists as a requirement, you'll also be able to ditch the next commit because of that.

tests/integration/test_data/gomod_1.18_e2e_test/bom.json
```diff
@@ -1436,6 +1436,17 @@
       "purl": "pkg:golang/regexp?type=package",
       "type": "library"
     },
+    {
```
eskultety, line 1439: First of all, this should have been part of  the previous commit if it introduced this breakage. Secondly,

> Probably, by adding cargo to cachi2 image some test data (SBOMs) must
> be updated, otherwise CI won't pass.

^this isn't a very descriptive message, there's too much uncertainty in it and not many facts. The real reason behind this needed update are cargo's direct dependencies, more specifically the `glibc-devel` package which provides the object (i.e. .so) standard C libraries.

Lastly, if we manage to install cargo in a clever manner we shouldn't need this test data update just yet.

pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
   "requests",
   "semver",
   "tomli",
+  "tomlkit",
```
eskultety, line 34: nitpick: Care enough to add an extra commit switching our own logic from tomli to tomlkit :)? It should be really simple since there are only 2 files where it's imported: `pip.py` and `nox.py`. Note we won't be able to get rid of it completely, it'll still remain in the requirements files as it's pulled in by something else, but we could replace it in `pyproject.toml`.

Dockerfile
slimreaper35: Thanks for that !

Dockerfile
```diff
@@ -28,6 +28,7 @@ RUN dnf -y install \
     --setopt install_weak_deps=0 \
     --nodocs \
     gcc \
+    cargo \
```
eskultety, line 31: nitpick: I know it's my suggestion (I was just prototyping quickly), maybe just add a short commentary here that this is not a direct build dependency, instead we just want to copy the binary over in the next stage.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -8,4 +18,42 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+        project_files.append(_use_vendored_sources(package_dir))
+
     return RequestOutput.from_obj_list(components, environment_variables, project_files)
+
+
+def _resolve_cargo_package(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+    """Resolve a single cargo package."""
+    vendor_dir = output_dir.join_within_root("deps/cargo")
+    cmd = ["cargo", "vendor", "--locked", str(vendor_dir)]
+    log.info("Vendoring crates at %s", package_dir)
```
eskultety, line 35: nitpick: I forgot to comment on this log message during first round of reviews - please stick to something like "Fetching" or whatever wording we use across the other backends, the fact that we're vendoring is an internal implementation detail which hypothetically could change at some point in the future.

cachi2/core/package_managers/pip.py
```diff
@@ -462,11 +462,9 @@ def get_version(self) -> Optional[str]:
...
     def _parsed_toml(self) -> dict[str, Any]:
         try:
-            with open(self._setup_file, "rb") as f:
-                log.debug("Parsing pyproject.toml at %r", str(self._setup_file))
-                data = tomli.load(f)
-                return data
-        except tomli.TOMLDecodeError as e:
+            log.debug("Parsing pyproject.toml at %r", str(self._setup_file))
+            return tomlkit.parse(self._setup_file.path.read_text())
+        except Exception as e:
```
eskultety, line 467: I'm wondering why they haven't exposed their top-level [TOMLKitError](https://github.com/python-poetry/tomlkit/blob/94e193afd679ca8cf5b0f45b879ffa6af20cf528/tomlkit/exceptions.py#L6) in the public API reference, that's weird. I'm tempted to say "it's been stable since its introduction in 2018 so let's use it", but let's keep this as is until the team responds to my issue: https://github.com/python-poetry/tomlkit/issues/399.

Just add a "TODO" commentary here and mention we'd want to replace Exception with a finer grained one once a resolution is reached in https://github.com/python-poetry/tomlkit/issues/399.

eskultety (approved):
Please address the commentary nitpicks before merging :) .

cachi2/core/package_managers/pip.py
slimreaper35: Why not just use `from tomlkit.exceptions import TOMLKitError` ? I don't see any exception being exposed in the API.

eskultety: > Why not just use from tomlkit.exceptions import TOMLKitError ?

Like I said earlier

> I'm tempted to say "it's been stable since its introduction in 2018 so let's use it (i.e. TOMLKitError)"

but even that exception is not part of their public API so technically can change, though I don't think it will.

Issue #782, opened 2025-01-10
[tests] [integration] Ignore our git change-based pytest filter if an explicit pytest filter was used ('-k XYZ')


eskultety:
By default the behaviour of our integration test suite is to save execution time on local dev machines by identifying changes to specific components and only execute relevant integration tests by filtering out all backends not affected by a given change.
That however, doesn't really work in case one is trying to explicitly run a particular set of integration tests without having actual changes in the corresponding backend and so no tests are run. The way one has to go about this is to use both `CACHI2_RUN_ALL_INTEGRATION_TESTS=true` and the corresponding pytest filter `-k XYZ` to get the desired outcome.
Instead, we should modify our `pytest_collection_modifyitems` function so that pytest native filters take precedence, IOW it would identify that a pytest native filter was used and in that case ignore our default change-based test selection behaviour.

PR #811, opened 2025-01-31
Skip test filtering when using pytest -k option

slimreaper35:
When the -k option is used to select tests by keyword, the existing logic would still skip some tests. This prevents users from running specific tests that are normally skipped. The check ensures that test filtering is bypassed when a keyword is provided.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -159,6 +159,10 @@ def pytest_collection_modifyitems(
     This function implements a standard pytest hook. Please refer to pytest
     docs for further information.
     """
+    # do not try to skip tests if a keyword or marker is specified
+    if config.getoption("-k") or config.getoption("-m"):
+        return
+
     skip_mark = pytest.mark.skip(reason="No changes to tested code")
     tests_to_skip = utils.determine_integration_tests_to_skip()
     for item in items:


tests/integration/conftest.py
```diff
@@ -159,6 +159,10 @@ def pytest_collection_modifyitems(
     This function implements a standard pytest hook. Please refer to pytest
     docs for further information.
     """
+    # do not try to skip tests if a keyword is provided (-k option)
+    if config.getoption("keyword"):
```
a-ovchinnikov, line 163: Won't this be addressed by requesting to run all tests and then filtering?

slimreaper35: In that case, we can close the issue and this PR. One minor problem that I noticed is that if you run `nox -s all-integration-tests -- -k test_something`  you always spin up local pypi server and dnf server fixtures.

slimreaper35: It might not be always desired, before merging #802 the dnf server failed to start and then everything cascaded and failed too.

tests/integration/conftest.py
```diff
@@ -159,6 +159,10 @@ def pytest_collection_modifyitems(
...
+        return
+
```
eskultety, line 165: Was this simple, huh? Nice! :)

That said, `keyword` is an internal implementation detail when it comes to option representation, only visible in a debugger, but not mentioned anywhere in the docs: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.Config (you should link the docs in the commit message)

So you want to do `config.getoption("-k")` instead, proof:
```
ipdb> p config.getoption("-k")
'test_gomod.py or test_pip.py'
```

Since you're at it, you may as well expand the check to the `-m` option (as in `-m MARKEXPR`) as well. We're not currently using marks much (or at all), but it's another official way of filtering which we should consider.

eskultety (changes requested):

tests/integration/conftest.py
eskultety: > Won't this be addressed by requesting to run all tests and then filtering?

It would, but that's not very ergonomic.

eskultety:
@slimreaper35 you should also reference https://github.com/containerbuildsystem/cachi2/issues/782 as resolved in the commit message.

tests/integration/conftest.py
eskultety: For example, I almost never run the "expensive" (i.e. requiring a server) tests and very rarely execute the UT suite via nox.

a-ovchinnikov (approved):

eskultety (approved):

PR #823, opened 2025-02-05
[reorg] Replace mentions of the old containerbuildsystem

eskultety:
The project moved to a new GitHub organization "hermetoproject". Change the references accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,8 +259,8 @@ See [docs/yarn.md](docs/yarn.md) for more details.
 
 Cachi2 was derived (but is not a direct fork) from [Cachito](https://github.com/containerbuildsystem/cachito).
 
-[cachi2-coverage-badge]: https://codecov.io/github/containerbuildsystem/cachi2/graph/badge.svg?token=VJKRTZQBMY
-[cachi2-coverage-status]: https://codecov.io/github/containerbuildsystem/cachi2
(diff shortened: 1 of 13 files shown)


eskultety:
Opening as a draft before moving the repository, but we'll need this right after the migration.

I also omitted references to the project in codecov settings as I'm unclear at this point what and where we need changes in the references (if at all).

eskultety:
There's more, I'm seeing some references in integration test data which I could replace by hand but I wanted to do it properly with test data re-generation which didn't work at all for many of the test data directories so I need to investigate further.

eskultety:
Since last revision:
- changed codecov references
- added some cachito ref purging that was directly tied to other references of the old containerbuildsystem org
- omitted integration test data changes because the migrated repository has some missing tags that the original had (needs to be addressed nicely in a follow-up PR)

eskultety:
> There's more, I'm seeing some references in integration test data which I could replace by hand but I wanted to do it properly with test data re-generation which didn't work at all for many of the test data directories so I need to investigate further.

As mentioned in https://github.com/hermetoproject/cachi2/pull/823#issuecomment-2724138917 this will not be handled in this PR.

eskultety:
Create https://github.com/hermetoproject/cachi2/issues/866 to track the rest.

MartinBasti (approved):

slimreaper35 (approved):

PR #826, opened 2025-02-06
[tests] Integration test fixes

eskultety:
A couple of small fixes that either fix a regression after the IT test suite repo consolidation or that have bothered me for a while.

### DIFF
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -134,6 +134,15 @@ class Cachi2Image(ContainerImage):
                 )
         return super().run_cmd_on_image(cmd, tmp_path, mounts, net)
 
+    def __exit__(self, exc_type: Any, exc_value: Any, exc_traceback: Any) -> None:
+        """Override the context manager exit to remove any dangling layers as build side effect."""
+        remove_dangling = "podman rmi --force $(podman images -q --filter 'dangling=true')"
+
+        # Best effort: ignore the return code, because if the dangling filter returns an empty set
+        # podman exits with 125 which is an internal podman error
+        # (in this case: "image name or ID must be specified")
+        run_cmd(remove_dangling, shell=True)
+
 
 def build_image(context_dir: Path, tag: str) -> ContainerImage:
     return _build_image(["podman", "build", str(context_dir)], tag=tag)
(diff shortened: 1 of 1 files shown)


tests/integration/utils.py
```diff
@@ -185,19 +185,28 @@ def _build_image(podman_cmd: list[str], *, tag: str) -> ContainerImage:
...
     :rtype: Tuple
     """
     log.info("Run command: %s.", cmd)
 
-    process = Popen(cmd, stdout=PIPE, stderr=PIPE)
-    out, err = process.communicate()
-    return (out + err).decode("utf-8"), process.returncode
+    forced_options = {
+        "stdout": subprocess.PIPE,
+        "stderr": subprocess.STDOUT,
```
slimreaper35, line 210: Is there a special reason for this?  We use `capture_output=True` in `cachi2.core.utils.py` instead.

eskultety: `capture` will cause 2 separate streams, we want the whole output condensed, that's more useful and simpler here in tests I think. Additionally, that's what the original logic did anyway.

tests/integration/utils.py
```diff
@@ -478,19 +478,19 @@ def build_image_and_check_cmd(
     log.info("Build container image with all prerequisites retrieved in previous steps")
     container_folder = test_data_dir.joinpath(test_case, "container")
 
-    test_image = build_image_for_test_case(
+    with build_image_for_test_case(
```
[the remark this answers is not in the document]
eskultety: It's easily missed during reviews ;) , that happens.

slimreaper35 (approved):

tests/integration/utils.py
```diff
@@ -134,6 +134,15 @@ def run_cmd_on_image(
                 )
         return super().run_cmd_on_image(cmd, tmp_path, mounts, net)
 
+    def __exit__(self, exc_type: Any, exc_value: Any, exc_traceback: Any) -> None:
+        """Override the context manager exit to remove any dangling layers as build side effect."""
+        remove_dangling = "podman rmi --force $(podman images -q --filter 'dangling=true')"
```
brunoapimentel, line 139: This would remove all dangling images in the user's filesystem, not just the ones generated by this test run, right? I'm just wondering if this could have an unintended side-effect, since Cachi2 is potentially deleting things it did not generate.

eskultety: > This would remove all dangling images in the user's filesystem, not just the ones generated by this test run, right? 

Yes, it would. But because these are always marked as `<none> <none>` you can't really tell without deep inspection on what generated those, so they're practically meaningless and it's actually podman's issue that it didn't delete them despite being instructed. Worst that happens is that future unrelated builds would get a bit slower due to missing layers from the intermediate images which nobody actually cares about in multi-stage builds, so this is a safe change from that perspective.

eskultety: > you can't really tell without deep inspection on what generated those

and I might even be wrong about this, because I simply didn't feel like tracing all the hashes on my local file system, so I **only suspect** that it's even possible trace them, but can't tell for sure.

brunoapimentel: I saw that podman-images has a `after/since` filter:

```
 Filter by images created after the given IMAGE (name or tag).     
```

Which we could maybe leverage? But you're probably right it's not a big deal to delete them all, so I'm thinking we shouldn't waste time investigating.

eskultety: > I saw that podman-images has a after/since filter:

IIUC in order to execute this the way you suggest we'd have to remember the last built image as a reference for the removal. However, there could be parallel builds happening in the meantime in which case you'd still delete unrelated dangling images.

> so I'm thinking we shouldn't waste time investigating.

Agreed, I mean they're called dangling for a reason :) .

Issue #803, opened 2025-01-30
Revisit older integration tests


slimreaper35:
Before merging #789, the whole integration test suite was a big mess, mostly one repository representing a single test case.
We should revisit older integration tests to **potentially** speed up overall execution even more.

See the new integration test repo -> https://github.com/cachito-testing/integration-tests and check if:

- there is an extremely outdated branch (cachi2 does not support the feature or specific version of pkg manager)
- possible duplicates (it happened already for bundler https://github.com/containerbuildsystem/cachi2/commit/03b433f1c878fbd7922c5565acc879a6df6ee66e)
- one branch (for example e2e test) covers scenarios in other branches -> the other branches could be deleted
- etc.

The following numbers represent the amount of integration tests that we currently have:

- bundler: 5
- generic: 3
- gomod: **14**
- npm: 8
- pip: **12**
- rpm: 10
- yarn berry: 9
- yarn classic: 8

A total of 69 integration tests at the time of creating this issue.

PR #831, opened 2025-02-10
Adding SBOM components and integration tests

a-ovchinnikov:
This change adds SBOM components generation and integration tests to finalize Cargo support implementation. It also fixes a problem of missing git dependencies substitution: git dependencies have to be explicitly replaced with a local copy. This change deals with this by collecting cargo output which is one substitution away from a correct config file template.


cachi2/core/package_managers/cargo/main.py
```diff
@@ -22,20 +21,29 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+        project_files.append(_use_vendored_sources(package_dir, cfg_template))
 
     return RequestOutput.from_obj_list(components, environment_variables, project_files)
 
 
-def _resolve_cargo_package(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+def _resolve_cargo_package(
+    package_dir: RootedPath,
+    output_dir: RootedPath,
+) -> tuple[list[Component], str]:
```
slimreaper35, line 40: I would split the function and keep the components and config file handling separately.

tests/integration/test_cargo.py
```diff
@@ -0,0 +1,121 @@
...
+                branch="cargo/mixed-git-crate-dependency",
+                packages=({"path": ".", "type": "cargo"},),
+                flags=["--dev-package-managers"],
+                check_output=False,
+                check_deps_checksums=False,
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="",
+            ),
+            id="mixed_git_crate_dependency",
```
slimreaper35, line 51: This is a duplicate test already present as e2e

tests/integration/test_cargo.py
```diff
@@ -0,0 +1,121 @@
...
+    check_cmd: list[str],
+    expected_cmd_output: str,
+    cachi2_image: utils.ContainerImage,
+    tmp_path: Path,
+    test_repo_dir: Path,
+    test_data_dir: Path,
+    request: pytest.FixtureRequest,
+) -> None:
+    """
+    End to end test for bundler.
```
slimreaper35, line 102:
```suggestion
    End to end test for cargo.
```

tests/integration/test_cargo.py
```diff
@@ -0,0 +1,121 @@
...
+
+    :param test_params: Test case arguments
+    :param tmp_path: Temp directory for pytest
```
slimreaper35, line 105: Let's just drop these "params" docstring. They seem to be copied all over the integration tests, just these two, nothing else.

tests/integration/test_data/cargo_mixed_dep/container/Containerfile
```diff
@@ -0,0 +1,9 @@
+FROM docker.io/rust:1.67
+
+# Test disabled network access
+RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
+
+WORKDIR /src
+RUN . /tmp/cachi2.env && \
+	sed -i 's/version = 4/version = 3/' Cargo.lock && \
```
slimreaper35, line 8: Why is `sed` used here ?

tests/integration/test_cargo.py
```diff
@@ -0,0 +1,121 @@
...
+                branch="cargo/just-a-git-dependency",
+                packages=({"path": ".", "type": "cargo"},),
+                flags=["--dev-package-managers"],
+                check_output=False,
+                check_deps_checksums=False,
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="",
+            ),
+            id="just_a_git_dependency",
```
slimreaper35, line 38: I think we could merge these test cases IMO to a more complex one. Having a git dependency and crate dependency together is a common scenario IMO. I would maybe add a negative scenario instead, for example missing Cargo.lock etc.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -34,7 +36,17 @@ def _resolve_cargo_package(package_dir: RootedPath, output_dir: RootedPath) -> l
     cmd = ["cargo", "vendor", "--locked", str(vendor_dir)]
     log.info("Fetching cargo dependencies at %s", package_dir)
     run_cmd(cmd=cmd, params={"cwd": package_dir})
-    return []
+
+    downloaded_deps = os.listdir(vendor_dir)
+    all_configs = [vendor_dir.path / x / "Cargo.toml" for x in downloaded_deps]
+    all_tomls = [TOMLFile(x).read() for x in all_configs]
```
eskultety, line 50: This can potentially be a list of fairly sized objects, can we make it lazily evaluated with `map` instead? Especially if you use `iterdir` from above then the 3 lines above can be condensed to a single `map` call.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -34,7 +36,17 @@ def _resolve_cargo_package(package_dir: RootedPath, output_dir: RootedPath) -> l
...
+    # Ignoring mypy error since name and version will be present in all Cargo.tomls downloaded
+    # from a registry.
+    name_versions = [(x["package"]["name"], x["package"]["version"]) for x in all_tomls]  # type: ignore
+    mk_purl = lambda n, v: PackageURL(type="cargo", name=n, version=v).to_string()  # noqa: E731
+    mk_comp = lambda n, v: Component(name=n, version=v, purl=mk_purl(n, v))  # noqa: E731
```
eskultety, line 55: This doesn't seem to be aligned with how other backends produce the information. I know in case of cargo most of the job has been done by cargo itself via vendoring, but I think **consistency** is key regardless, i.e. following the precedent set by other modules (i.e. using `purl` and `to_component` instance methods) for the small tradeoff of defining tiny dataclasses and some negligible overhead I think.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -34,7 +36,17 @@ def _resolve_cargo_package(package_dir: RootedPath, output_dir: RootedPath) -> l
...
+    components = [mk_comp(n, v) for n, v in name_versions]
```
eskultety, line 56: nitpick: this gets immediately added to a different list, which means for a small period of time we'd be storing 2 lists in memory for no good reason, can this be turned into a generator instead?

cachi2/core/package_managers/cargo/main.py
```diff
@@ -22,20 +21,29 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
         package_dir = request.source_dir.join_within_root(package.path)
-        components.extend(_resolve_cargo_package(package_dir, request.output_dir))
         # cargo allows to specify configuration per-package
         # https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure
-        project_files.append(_use_vendored_sources(package_dir))
+        fetched_components, cfg_template = _resolve_cargo_package(package_dir, request.output_dir)
+        components.extend(fetched_components)
+        tail = 'directory = "${output_dir}/deps/cargo"'
+        # absolute path has to be replaced with relative path for sources relocation to work:
+        # (absolute path also happens to be the last line generated by cargo)
```
eskultety, line 30: Can this path mangling be incorporated in the resolve method and return a final template product? I mean the resolve method has all the necessary information.

eskultety: >\# absolute path has to be replaced with relative path for sources relocation to work:
>\# (absolute path also happens to be the last line generated by cargo)

This is very error prone I'm afraid, we're relying on cargo's CLI output ordering which in general isn't considered part of the public API, so we'll need to be more thorough with the check (`startswith` should be sufficient, `re` would IMO be overkill for such a simple replacement).

cachi2/core/package_managers/cargo/main.py
```diff
@@ -22,20 +21,29 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+        cfg_template = "\n".join(cfg_template.split("\n")[:-2] + [tail])
```
eskultety, line 31: `cfg_template.split("\n")[:-2] + [tail]` this is quite cryptic to be fair even if documented, but like I said above, we can't rely on this, the ordering of the config sections may change or new sections may be generated in the future.

eskultety: To be honest, the original version with a static template and a `TOMLKit.update` was more readable than this new attempt.

a-ovchinnikov:
The new revision has components generation unified with a dedicated dataclass, package qualifiers are added to purls, Containerfile is switched to a different base image.

tests/integration/test_cargo.py
slimreaper35: I guess this is the same topic as discussed in https://github.com/hermetoproject/cachi2/pull/832#discussion_r1951099181.

BTW I removed duplicate tests for bundler in [#783.](https://github.com/hermetoproject/cachi2/pull/783/commits/c71d29ecbf149393d241670d4289702d8b62eb23).

test_e2e_x is a basically superset of test_x_packages, it adds `utils.build_image_and_check_cmd`.
In total, there are only two function calls, so I don't see any big investigation here after a failed test.

tests/integration/test_cargo.py
eskultety: > When things go wrong in a complex scenario one would need to invest more time into figuring out which part caused the issue. Having separate tests helps in making sure that individual types work fine.

Depends, these tests are fairly cheap compared to e2e, but then if e2e is a superset, then I also don't see that much value in individual positive tests, because positive cases should be covered in an e2e. Testing negative scenarios individually is a different story.

tests/integration/test_data/cargo_mixed_dep/container/Containerfile
```diff
@@ -0,0 +1,9 @@
+FROM docker.io/rust:1.67
```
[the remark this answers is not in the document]
eskultety: Hold on, I'm getting confused now.

>   1.84 generates lock file v4 while 1.67 generates v3.

What does ^this mean for us? Are we entering a yarn-like territory with these versions or are we completely unaffected by different lockfile versions? Can you elaborate some more please?

cachi2/core/package_managers/cargo/main.py
eskultety: If we don't have it even for yarn which is the most recent example, then well, we should be more diligent about it :) .

tests/integration/test_cargo.py
slimreaper35: We have an older IT, which is now a public archive, created by @brunoapimentel - https://github.com/cachito-testing/cachi2-rust.

I always thought about e2e tests to be more or less real-world projects with many dependencies. Personally, I don't like the idea of having simple tests just to have less work to do when something fails.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -20,40 +39,54 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
-    run_cmd(cmd=cmd, params={"cwd": package_dir})
-    return []
+    # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
+    config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
+
+    packages = tomlkit.parse((package_dir.path / "Cargo.lock").read_text())["package"]
+    # a shorthand helper, thus noqa:
+    is_a_dep = lambda p: p["name"] != package_dir.path.stem  # noqa: E731
+    # mypy has issues with parsed tomls:
+    components = (CargoPackage(**p).to_component() for p in packages if is_a_dep(p))  # type: ignore
```
slimreaper35, line 66: Let's use a regular list instead. We used a generator for yarn classic (https://github.com/hermetoproject/cachi2/commit/5d0c8cdf0feb2b40fd1f5dae8094cd2ba168e696) and we had a hidden bug because we iterated through the components multiple times even though it may not be the case this time

tests/integration/test_data/cargo_mixed_dep/container/Containerfile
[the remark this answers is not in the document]
a-ovchinnikov: v4 could be incompatible with v3 and older versions of cargo would refuse dealing with lock files generated by newer ones.

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -1,42 +0,0 @@
-import tomlkit
```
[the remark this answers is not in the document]
slimreaper35: ok, resolving thread

cachi2/core/package_managers/cargo/main.py
```diff
@@ -40,47 +39,54 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
 
-def _use_vendored_sources(package_dir: RootedPath) -> ProjectFile:
+def _swap_sources_directory_for_subsitution_slot(template: str) -> str:
+    toml_template = tomlkit.parse(template)
+    # Absolute path has to be replaced with relative path for sources relocation to work:
+    toml_template["source"]["vendored-sources"]["directory"] = "${output_dir}/deps/cargo"  # type: ignore
+    #  side note:  mypy does not believe that we'll always be receiving templates of correct
+    #              structure and complains that in theory we might get a not deep enough
+    #              nesting of containers.
+    # A correct output_dir value will be supplied iby hermeto during a later stage.
```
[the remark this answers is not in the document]
a-ovchinnikov: Still not yet *sigh*

cachi2/core/package_managers/cargo/main.py
```diff
@@ -40,47 +39,54 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
 
 
-def _use_vendored_sources(package_dir: RootedPath) -> ProjectFile:
+def _swap_sources_directory_for_subsitution_slot(template: str) -> str:
+    toml_template = tomlkit.parse(template)
+    # Absolute path has to be replaced with relative path for sources relocation to work:
+    toml_template["source"]["vendored-sources"]["directory"] = "${output_dir}/deps/cargo"  # type: ignore
+    #  side note:  mypy does not believe that we'll always be receiving templates of correct
+    #              structure and complains that in theory we might get a not deep enough
+    #              nesting of containers.
```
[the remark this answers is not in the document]
a-ovchinnikov: I would have argued about relative cleanliness of using an existing interface vs unwrapping into an object with effectively the same interface for the only purpose of placating a linter, but I won't. I'll let mypy live with this.

tests/integration/test_cargo.py
a-ovchinnikov: The point of having simple tests is the same as for having well organized code -- to compartmentalize behavior in manageable components. I can easily buy the argument for having more tests, some of which could be quite complex, but I still think that having simpler tests increases precision.

tests/integration/test_cargo.py
a-ovchinnikov: 50% reduction of potentially problematic are at a glance is quite a feat, I'd say :)

cachi2/core/package_managers/cargo/main.py
slimreaper35: I meant something like this:

```
def _generate_sbom_components(package_dir: RootedPath) -> list[Component]:
    """Generate SBOM components for a cargo package."""
    # TODO: Here
    return []


def _resolve_cargo_package(
    package_dir: RootedPath,
    output_dir: RootedPath,
) -> ProjectFile:
    """Resolve a single cargo package."""
    vendor_dir = output_dir.join_within_root("deps/cargo")
    cmd = ["cargo", "vendor", "--locked", str(vendor_dir)]
    log.info("Fetching cargo dependencies at %s", package_dir)
    # STDOUT contains the exact values ​​that need to be added to the .cargo/config.toml file for hermetic build
    config = run_cmd(cmd=cmd, params={"cwd": package_dir})

    config_template = _swap_vendored_directory_for_placeholder(config)
    return _update_cargo_config(package_dir, config_template)


def _swap_vendored_directory_for_placeholder(config: str) -> dict[str, Any]:
    """Swap the vendored directory for a placeholder in the cargo config file."""
    toml_template = tomlkit.parse(config).value
    toml_template["source"]["vendored-sources"]["directory"] = "${output_dir}/deps/cargo"
    return toml_template


def _update_cargo_config(package_dir: RootedPath, config_template: dict[str, Any]) -> ProjectFile:
    """Make sure cargo will use the vendored sources when building the project."""
    cargo_config = package_dir.join_within_root(".cargo/config.toml")
    cargo_config.path.parent.mkdir(parents=True, exist_ok=True)
    cargo_config.path.touch(exist_ok=True)

    toml_file = TOMLFile(cargo_config)
    original_content = toml_file.read()
    original_content.update(config_template)

    return ProjectFile(abspath=cargo_config.path, template=tomlkit.dumps(config_template))
```

cachi2/core/package_managers/cargo/main.py
```diff
@@ -28,13 +48,20 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
     """Resolve a single cargo package."""
     vendor_dir = output_dir.join_within_root("deps/cargo")
     cmd = ["cargo", "vendor", "--locked", str(vendor_dir)]
     log.info("Fetching cargo dependencies at %s", package_dir)
     run_cmd(cmd=cmd, params={"cwd": package_dir})
-    return []
+
+    packages = tomlkit.parse((package_dir.path / "Cargo.lock").read_text())["package"]
+    # a shorthand helper, thus noqa:
+    is_a_dep = lambda p: p["name"] != package_dir.path.stem  # noqa: E731
```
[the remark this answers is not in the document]
eskultety: Seems like named lambdas are going to be a thing in the code base, not a blocker for me if it can be easily identifiable. One thing that my eyes don't like is the repeating `noqa: E731` hint - can you apply a global ignore to our flake8 rules and rid us of all the occurrences?

cachi2/core/package_managers/cargo/main.py
```diff
@@ -28,13 +48,20 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
-def _resolve_cargo_package(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+def _resolve_cargo_package(package_dir: RootedPath, output_dir: RootedPath) -> Generator:
     """Resolve a single cargo package."""
     vendor_dir = output_dir.join_within_root("deps/cargo")
     cmd = ["cargo", "vendor", "--locked", str(vendor_dir)]
     log.info("Fetching cargo dependencies at %s", package_dir)
     run_cmd(cmd=cmd, params={"cwd": package_dir})
-    return []
+
+    packages = tomlkit.parse((package_dir.path / "Cargo.lock").read_text())["package"]
```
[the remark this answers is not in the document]
eskultety: I mean, do they need to match necessarily? The fact that it behaves like a dict doesn't mean it's **just** a dict, so to me that isn't that big of an issue, after all, it's us who's trying to unpack the container to a class constructor ;).

cachi2/core/package_managers/cargo/main.py
a-ovchinnikov: Oh, I see, thank you. I believe this does not apply anymore since we have a dataclass now.

tests/integration/test_cargo.py
a-ovchinnikov: Do you suggest dropping all three ITs in favor of the unified e2e?

cachi2/core/package_managers/cargo/main.py
```diff
@@ -20,40 +53,60 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
     log.info("Fetching cargo dependencies at %s", package_dir)
-    run_cmd(cmd=cmd, params={"cwd": package_dir})
-    return []
+    # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
+    config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
+
+    # 'value' unwraps the underlying dict and that makes mypy happy (it complains about
+    # mismatching type otherwise despite parsed document having the necessary interface).
+    packages = tomlkit.parse((package_dir.path / "Cargo.lock").read_text()).value["package"]
+    is_a_dep = lambda p: p["name"] != package_dir.path.stem  # a shorthand, thus # noqa: E731
```
brunoapimentel, line 79: This will fail if the cloned directory name does not match the `package` name in Cargo.toml. Here's an [example](https://github.com/cachito-testing/cachi2-rust/blob/main/Cargo.toml#L13) where the repository name differs from the package name, and will be cloned by default to a mismatched folder.

slimreaper35: I think the package section in the `Cargo.toml` is not required if the repo has only workspace members.

brunoapimentel: That is a good point. I think the main problem here is if there's no package whose name matches the cloned directory.

a-ovchinnikov: Yes, good catch, thank you, cachi2 vendors `rustic` as a dependency just fine, but does not vendor `rustic` as the main package.

a-ovchinnikov: > the repo has only workspace members

@slimreaper35 do you have an example of such repo?

slimreaper35: I was referencing cargo [documentation](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html#creating-a-workspace), and I believe it is a valid case. If necessary, I may try to find one.

a-ovchinnikov: I found one in the wild after all: https://github.com/rwf2/Rocket/tree/master

a-ovchinnikov:
Latest rev: addressed main package name generation concern.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -28,13 +64,65 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+    # 'value' unwraps the underlying dict and that makes mypy happy (it complains about
+    # mismatching type otherwise despite parsed document having the necessary interface).
+    return tomlkit.parse(path_to_toml.read_text()).value["package"]
+
+
+def _resolve_main_package(package_dir: RootedPath) -> dict:
+    try:
+        return _extract_package_info(package_dir.path / "Cargo.toml")
+    # We'll get here in the case of virtual workspaces. A real-world example is
+    # https://github.com/rwf2/Rocket/tree/master
```
slimreaper35, line 77: Why do we have to reference a random project on github ?

cachi2/core/package_managers/cargo/main.py
```diff
@@ -28,13 +64,65 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+
+    packages = _extract_package_info(package_dir.path / "Cargo.lock")
+    main_package = _resolve_main_package(package_dir)
+    is_a_dep = lambda p: p["name"] != main_package["name"]  # a shorthand, thus # noqa: E731
+    deps_components = (CargoPackage(**p).to_component() for p in packages if is_a_dep(p))
+
+    vcs_url = get_repo_id(package_dir.root).as_vcs_url_qualifier()
+    main_component = CargoPackage(
+        name=main_package["name"], version=main_package["version"], vcs_url=vcs_url
+    ).to_component()
```
slimreaper35, line 122: I would split the logic of handling "main package" to a separate function.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -28,13 +64,65 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+            "name": package_dir.path.stem,
+            "version": "0.0.1",
+        }
+
+
+def _verify_repository_is_locked_or_fail(package_dir: RootedPath) -> None:
+    # Most packages will be locked, however metapackages (i.e. those, which
+    # contain just a workspace and could even lack a name) could arrive without
+    # a lock file. A user could try and fix this by explicitly locking the
+    # package first.
```
slimreaper35, line 94: I think this is a 1:1 situation as we have for yarn when we use `--frozen-lockfile` flag.
It has nothing to do with metapackages or workspaces. Better name would be `_verify_lockfile_is_synced` or `_verify_lockfile_is_up_to_date`

cachi2/core/package_managers/cargo/main.py
```diff
@@ -1,17 +1,53 @@
...
+@dataclass(frozen=True)
+class CargoPackage:
+    """CargoPackage."""
+
+    name: str
+    version: str
+    source: Optional[str] = None  # [git|registry]+https://github.com/<org>/<package>#[|<sha>]
+    checksum: Optional[str] = None
+    dependencies: Optional[list] = None
+    vcs_url: Optional[str] = None
```
slimreaper35, line 31: Does vcs_url exist in `Cargo.lock` ?

cachi2/core/package_managers/cargo/main.py
```diff
@@ -28,13 +64,65 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+    _verify_repository_is_locked_or_fail(package_dir)
     vendor_dir = output_dir.join_within_root("deps/cargo")
     cmd = ["cargo", "vendor", "--locked", str(vendor_dir)]
     log.info("Fetching cargo dependencies at %s", package_dir)
     run_cmd(cmd=cmd, params={"cwd": package_dir})
-    return []
+
+    packages = _extract_package_info(package_dir.path / "Cargo.lock")
+    main_package = _resolve_main_package(package_dir)
+    is_a_dep = lambda p: p["name"] != main_package["name"]  # a shorthand, thus # noqa: E731
```
slimreaper35, line 116: This seems to me like a very weak differentiation. I could name my directory like any dependency from crates.io.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -20,40 +55,93 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+
+    return components, _swap_sources_directory_for_subsitution_slot(config_template)
 
 
-def _use_vendored_sources(package_dir: RootedPath) -> ProjectFile:
+def _swap_sources_directory_for_subsitution_slot(template: str) -> dict:
+    toml_template = tomlkit.parse(template).value
+    # Absolute path has to be replaced with relative path for sources relocation to work:
+    toml_template["source"]["vendored-sources"]["directory"] = "${output_dir}/deps/cargo"
+    # A correct output_dir value will be supplied by cachi2 during a later stage.
```
slimreaper35, line 133: nitpick: These comments seem too obvious but ok.

slimreaper35:
I added a few comments, but I will touch the code when doing #833, so if you want to get this out of your way and merge it, you can ignore them. 

But let's make a team decision on how to approach integration tests and keep it consistent later with other package managers. This is a bit contradictory with #803 and my recent work to move all test cases to a single repo to speed up the test suite.

cachi2/core/package_managers/cargo/main.py
a-ovchinnikov: Because it is an actual example of something that has to be handled. I think providing context and reference points helps readers and future maintainers. It took me some effort to find one, no one seemed to be able to name an example off the top of their head and thus I copied the link over for posterity.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -28,13 +64,65 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+    # https://github.com/rwf2/Rocket/tree/master
+    # In this case there is no package name per se, but there still is a metapackage.
+    # We'll use directory name as a fallback in this case.
+    # Version won't make much sense here: this is a meta-package, the state of a
+    # repository will be captured in VCS_URL, and individual components will be
+    # versioned, so the version will be pinned artificially to a default value.
+    except KeyError:
+        return {
+            "name": package_dir.path.stem,
+            "version": "0.0.1",
```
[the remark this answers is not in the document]
a-ovchinnikov: Good point, thank you.

cachi2/core/package_managers/cargo/main.py
a-ovchinnikov: When there is no lock file it cannot be up to date and there is nothing to be synced. Moreover, the name matches what the function does: it checks if the repository is locked (i.e. if lock file is present) and fails otherwise. Maybe a better name would be `_verify_lockfile_is_present_or_fail`, I'll try that.

cachi2/core/package_managers/cargo/main.py
a-ovchinnikov: Technically yes, one could definitely do that. What would be the practical point though? I think we should clearly differentiate between a set of all possibilities and a set of those that could be seen in practice. Otherwise it is easy to come up with dozens of creative ways to subvert any of assumptions that were made.

cachi2/core/package_managers/cargo/main.py
a-ovchinnikov: Yes, I agree, this would increase code orthogonality. At the same time the function code is linear, there are no branches nor cycles and it fits on one standard screen in its entirety. I am strongly inclined to keeping it as is unless there is more demand for splitting.

cachi2/core/package_managers/cargo/main.py
a-ovchinnikov: To the best of my knowledge no, but IIUC we would still want it included for checked out projects, so I decided to lump it together with other fields in CargoPackage and set it only for the main one. Do you have a better place for that in mind?

eskultety: Not having all the necessary context here, @a-ovchinnikov can you please elaborate more on the VCS_URL field and its relevance to cargo as you tried to hint at in your response? I got lost, sorry.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -28,13 +64,65 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
     return RequestOutput.from_obj_list(components, environment_variables, project_files)
 
 
-def _resolve_cargo_package(package_dir: RootedPath, output_dir: RootedPath) -> list[Component]:
+def _extract_package_info(path_to_toml: Path) -> dict:
+    # 'value' unwraps the underlying dict and that makes mypy happy (it complains about
+    # mismatching type otherwise despite parsed document having the necessary interface).
+    return tomlkit.parse(path_to_toml.read_text()).value["package"]
```
eskultety, line 70: I'm wondering - this likely won't be our sole TOML parsing edeavour, I think it would make sense for us to subclass the parser and add a `to_dict` method wrapping the `value` accessor to make it easier and more predictable for everyone. I simply don't like the commentary about shutting up mypy, yeah, it's that, but at the same time, it's an official interface...
Up to you.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -28,13 +64,65 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
+    # Most packages will be locked, however metapackages (i.e. those, which
+    # contain just a workspace and could even lack a name) could arrive without
+    # a lock file. A user could try and fix this by explicitly locking the
+    # package first.
+    if not (package_dir.path / "Cargo.lock").exists():
+        raise PackageRejected(
+            f"{package_dir.path} is not locked",
+            solution="Please lock it first by running 'cargo generate-lockfile",
+        )
```
eskultety, line 99: Hold on, so what is the common practice with Cargo workspaces wrt to lockfiles? Is there supposed to be just a single lockfile on the repo root level or is there supposed to be one for each workspace package since there is one `Cargo.toml` per workspace member?

cachi2/core/package_managers/cargo/main.py
```diff
@@ -122,26 +123,25 @@ def _resolve_cargo_package(
...
+    toml_template["source"]["vendored-sources"]["directory"] = "${output_dir}/deps/cargo"
+    # A correct output_dir value will be supplied by cachi2 during a later stage.
+    return toml_template
+
+
+def _use_vendored_sources(package_dir: RootedPath, config_template: dict) -> ProjectFile:
     """Make sure cargo will use the vendored sources when building the project."""
     cargo_config = package_dir.join_within_root(".cargo/config.toml")
     cargo_config.path.parent.mkdir(parents=True, exist_ok=True)
     cargo_config.path.touch(exist_ok=True)
```
eskultety, line 141: So the intent of this patch is to only update `.cargo/config.toml` with the `inject-files` command, but this pre-existing hunk touches the file if missing, thus making premature changes to the repo, so that needs to be addressed as well so that the file is ONLY created in `inject-files`.

tests/integration/test_cargo.py
eskultety: I'm on the same boat as @slimreaper35 in that the suite exercises the same code paths twice. I don't think the debugging burden is so substantial that one would take significantly longer to pinpoint a problem from a more complex e2e test. That said though, I don't intend to dwell on tests too much to the point where it should delay acceptance of a PR as long as the tests meet some criteria:
- tests are readable
- tests are easily identifiable (i.e. naming)
- tests are meaningful
- coverage is objectively reasonable (i.e. positive, but mainly negative test cases are introduced)

What bothers me about our ITs is how much time they take to complete and would like to address it. Especially during a rebase it's a nightmare because even with the auto contextual filtering we have in place it basically snowballs as you progress through the rebase; not only that, it rebuilds the container image with every step which is insane, but that's a discussion for another time and I may have an idea how to tackle it.

That said, I'm okay with the composition, so we can resolve the thread.

cachi2/core/package_managers/cargo/main.py
slimreaper35: This class would deserve to be split and use similar approach as for other package managers, e.g. yarn classic mixins.

Dependencies attribute is never used. Checksum is not relevant for all package managers. There is no information about subpath, which is a generic and also used in purls.

tests/integration/test_cargo.py
slimreaper35: I spent non-trivial time to speed up integration tests when merging them to single repo. Having duplicate tests here does complete oposite.

cachi2/core/package_managers/cargo/main.py
a-ovchinnikov: I think this is `mypy`'s problem -- the class in question provides correct interface which is supposed to be enough and it is enough from user's perspective, however from `mypy`'s perspective it is not. There is no other necessity for a subclass than placating `mypy` and I find both approaches off-putting. I suggest keeping this as an internal function until we need it somewhere else and then either lift it to `utils` as is or subclass.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -28,13 +64,65 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+    _verify_lockfile_is_present_or_fail(package_dir)
     vendor_dir = output_dir.join_within_root("deps/cargo")
     cmd = ["cargo", "vendor", "--locked", str(vendor_dir)]
     log.info("Fetching cargo dependencies at %s", package_dir)
     run_cmd(cmd=cmd, params={"cwd": package_dir})
-    return []
+
+    packages = _extract_package_info(package_dir.path / "Cargo.lock")
+    main_package = _resolve_main_package(package_dir)
+    is_a_dep = lambda p: p["name"] != main_package["name"]  # a shorthand, thus # noqa: E731
```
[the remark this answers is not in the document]
a-ovchinnikov: Done

cachi2/core/package_managers/cargo/main.py
a-ovchinnikov: Thank you for catching this!

tests/integration/test_cargo.py
a-ovchinnikov: There are a few ways to improve ITs, for example running several instances in parallel or working through a local cache to eliminate extra network delay. These bigger changes are out of scope. While I still believe in the value of incremental positive testing I am OK to drop all the ITs in favor of one E2E. This will be a trivial change. Could everyone please react with thumbs up for dropping and thumbs down for keeping them?

cachi2/core/package_managers/cargo/main.py
```diff
@@ -1,17 +1,52 @@
...
+    checksum: Optional[str] = None
+    dependencies: Optional[list] = None
+    vcs_url: Optional[str] = None
+
+    @cached_property
+    def purl(self) -> PackageURL:
+        """Return corrsponding purl."""
+        qualifiers = {}
+        if self.source is not None:
+            qualifiers.update({"source": self.source})
```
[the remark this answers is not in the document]
brunoapimentel: I was thinking of consistency with other package managers, since we do report it only for vcs or file dependencies. But I'm not against it, so I'll just close this thread.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -1,17 +1,53 @@
...
+
+    name: str
+    version: str
+    source: Optional[str] = None  # [git|registry]+https://github.com/<org>/<package>#[|<sha>]
+    checksum: Optional[str] = None
+    dependencies: Optional[list] = None
+    vcs_url: Optional[str] = None
+
+    @cached_property
+    def purl(self) -> PackageURL:
```
taylormadore, line 34: Non-blocking question: Do we want/need to handle alternate registries? If so, I would envision that as a follow-up

a-ovchinnikov: Alternate registries are theoretically possible, but to the best of my knowledge are not widely used. I think we can safely postpone adding their support.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -20,40 +55,93 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+    config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
+
+    packages = _extract_package_info(package_dir.path / "Cargo.lock")
+    main_package = _resolve_main_package(package_dir)
+    is_a_dep = lambda p: p["name"] != main_package["name"]  # a shorthand, thus # noqa: E731
+    deps_components = (CargoPackage(**p).to_component() for p in packages if is_a_dep(p))
+
+    vcs_url = get_repo_id(package_dir.root).as_vcs_url_qualifier()
+    main_component = CargoPackage(
+        name=main_package["name"], version=main_package["version"], vcs_url=vcs_url
```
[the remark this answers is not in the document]
brunoapimentel: I think the origin of components is not clear with the current approach. For instance, look at these sample purls:

*main package*
```
pkg:cargo/serde?vcs_url=git%2Bhttps://github.com/serde-rs/serde%40a8bdd17333889cebe7af97064066f89aab1bef8e
```

*worskpace package*
```
pkg:cargo/serde_derive@1.0.218
```

The vcs url makes clear that the main package is not simply a registry dependency, so I think we can have the same treatment for workspaces. Another argument is that they're the only packages that do not have a source qualifier, so I think we could at least provide the same vcs_url as the main package (ideally with the workspace's subpath).

Now, I'm not aware of the limitations that the cargo commands impose. Will the approach I proposed be too troublesome to implement (as you mentioned, would it need a dedicated toml processor)?

cachi2/core/package_managers/cargo/main.py
eskultety: Okay, thanks for clarifying, then it's reasonable for us to ask the user to do it.

brunoapimentel (approved):

taylormadore (approved):

eskultety (approved):

PR #832, opened 2025-02-11
[tests] [unit] Fix the CLI test so it doesn't leave a bunch of temp files behind

eskultety:

### DIFF
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -949,22 +949,30 @@ class TestMergeSboms:
     @pytest.mark.parametrize(
         "sbom_files_to_merge",
         [
-            [
-                "./tests/unit/data/sboms/cachi2.bom.json",
-                "./tests/unit/data/sboms/cachito_gomod.bom.json",
-            ],
-            [
-                "./tests/unit/data/sboms/cachi2.bom.json",
-                "./tests/unit/data/sboms/cachito_gomod.bom.json",
-                "./tests/unit/data/sboms/cachito_gomod_nodeps.bom.json",
-            ],
+            pytest.param(
+                [
+                    "./tests/unit/data/sboms/cachi2.bom.json",
+                    "./tests/unit/data/sboms/cachito_gomod.bom.json",
+                ],
+                id="merge_our_own_cyclonedx",
+            ),
+            pytest.param(
+                [
+                    "./tests/unit/data/sboms/cachi2.bom.json",
+                    "./tests/unit/data/sboms/cachito_gomod.bom.json",
+                    "./tests/unit/data/sboms/cachito_gomod_nodeps.bom.json",
+                ],
+                id="merge_our_own_cyclonedx_more",
+            ),
         ],
     )
     def test_a_user_can_successfully_save_sboms_merge_results_to_a_file(
         self,
+        request: pytest.FixtureRequest,
         sbom_files_to_merge: list[str],
     ) -> None:
-        with tempfile.NamedTemporaryFile() as fp:
+        prefix = f"cachi2-{Path(__file__).stem}-{request.node.callspec.id}-"
+        with tempfile.NamedTemporaryFile(prefix=prefix) as fp:
             invoke_expecting_sucess(app, ["merge-sboms", "-o", fp.name, *sbom_files_to_merge])
             assert Path(fp.name).lstat().st_size > 0, "SBOM failed to be written to output file!"
 
@@ -979,9 +987,11 @@ class TestMergeSboms:
     )
     def test_a_user_can_successfully_save_sboms_merge_results_to_a_file_in_spdx_format(
         self,
+        request: pytest.FixtureRequest,
         sbom_files_to_merge: list[str],
     ) -> None:
-        with tempfile.NamedTemporaryFile() as fp:
+        prefix = f"cachi2-{Path(__file__).stem}-{request.node.callspec.id}-"
+        with tempfile.NamedTemporaryFile(prefix=prefix) as fp:
             invoke_expecting_sucess(
                 app,
                 ["merge-sboms", "-o", fp.name, "--sbom-output-type", "spdx", *sbom_files_to_merge],
@@ -991,17 +1001,22 @@ class TestMergeSboms:
     @pytest.mark.parametrize(
         "sbom_files_to_merge",
         [
-            [
-                "./tests/unit/data/sboms/cachi2.bom.json",
-                "./tests/unit/data/sboms/syft.bom.spdx.json",
-            ],
+            pytest.param(
+                [
+                    "./tests/unit/data/sboms/cachi2.bom.json",
+                    "./tests/unit/data/sboms/syft.bom.spdx.json",
+                ],
+                id="merge_mixed_format",
+            ),
         ],
     )
     def test_a_user_can_successfully_save_mixed_sboms_merge_results_to_a_file_in_spdx_format(
         self,
+        request: pytest.FixtureRequest,
         sbom_files_to_merge: list[str],
     ) -> None:
-        with tempfile.NamedTemporaryFile() as fp:
+        prefix = f"cachi2-{Path(__file__).stem}-{request.node.callspec.id}-"
+        with tempfile.NamedTemporaryFile(prefix=prefix) as fp:
             invoke_expecting_sucess(
                 app,
                 ["merge-sboms", "-o", fp.name, "--sbom-output-type", "spdx", *sbom_files_to_merge],


tests/unit/test_cli.py
```diff
@@ -935,79 +935,51 @@ def test_a_user_sees_error_when_they_provide_a_non_cachi2_sbom_for_a_merge(
...
         sbom_files_to_merge: list[str],
+        format_opt: list[str],
     ) -> None:
         prefix = f"cachi2-{Path(__file__).stem}-{request.node.callspec.id}-"
-        with tempfile.NamedTemporaryFile(prefix=prefix) as fp:
-            invoke_expecting_sucess(
-                app,
-                ["merge-sboms", "-o", fp.name, "--sbom-output-type", "spdx", *sbom_files_to_merge],
-            )
+        with tempfile.NamedTemporaryFile(prefix=prefix, delete=False) as fp:
```
slimreaper35, line 982: In the second commit, you removed `delete=False` argument. Did you want to add it here again ?

eskultety: Nope, that is a leftover from an interactive rebase test, good catch!

slimreaper35 (approved):

tests/unit/test_cli.py
```diff
@@ -956,8 +956,7 @@ def test_a_user_can_successfully_save_sboms_merge_results_to_a_file(
         self,
         sbom_files_to_merge: list[str],
     ) -> None:
-        with tempfile.NamedTemporaryFile(delete=False) as fp:
-            fp.close()
+        with tempfile.NamedTemporaryFile() as fp:
```
a-ovchinnikov: I introduced a close() call (accompanied by `delete=False`) back in the days of exposing `merge-sboms` as a subcommand to work around some problem with a temporary file, likely a file contention. Back then it was a conscious decision which had to make it into a comment, but apparently did not. Now I wonder what went wrong back then and is fixed now.

tests/unit/test_cli.py
```diff
@@ -928,10 +928,6 @@ def test_a_user_sees_error_when_they_provide_a_non_cachi2_sbom_for_a_merge(
     @pytest.mark.parametrize(
         "sbom_files_to_merge",
         [
```
a-ovchinnikov, line 930: I strongly suggest not skipping simpler tests. The idea behind having a progression of complexity is to help with detecting a breaking point. If some update breaks two-SBOMs case then it will likely break three-SBOMs case. If, however, something breaks three-SBOMs, but keeps two-SBOMs intact then there is quite a bit of narrowing down. It usually does not make sense to go beyond three, that's why I have never added more cases here.

tests/unit/test_cli.py
```diff
@@ -935,79 +935,51 @@ def test_a_user_sees_error_when_they_provide_a_non_cachi2_sbom_for_a_merge(
             ],
         ],
     )
-    def test_a_user_can_successfully_merge_several_cachi2_sboms(
+    def test_merge_several_cachi2_sboms(
```
a-ovchinnikov, line 938: Please don't. It is totally fine to have duplications in test cases. Test cases are not code, test cases exist to state boring facts about program's units behavior. Test cases names are to tell a summary of what is going to happen in a test, to tell someone what to expect from this test. This summary falls apart when one has to go back and forth and assemble it from ids and test names. I insisted on not modifying pre-existing tests when SPDX was added and I insist on that now. Tests must not be touched unless those are non-valuable tests which do not actually test anything. Having multiple test cases gradually adding features to be tested could greatly help during refactoring.

tests/unit/test_cli.py
```diff
@@ -945,18 +945,23 @@ def test_a_user_can_successfully_merge_several_cachi2_sboms(
...
+            )
         ],
     )
     def test_a_user_can_successfully_save_sboms_merge_results_to_a_file(
         self,
+        request: pytest.FixtureRequest,
         sbom_files_to_merge: list[str],
     ) -> None:
-        with tempfile.NamedTemporaryFile() as fp:
+        prefix = f"cachi2-{Path(__file__).stem}-{request.node.callspec.id}-"
```
a-ovchinnikov: This addition is so cool!

a-ovchinnikov (changes requested):
Two commits in the middle are awesome, two commits at the edges are questionable (please see my comments inline).

tests/unit/test_cli.py
eskultety: > I strongly suggest not skipping simpler tests. The idea behind having a progression of complexity is to help with detecting a breaking point. If some update breaks two-SBOMs case then it will likely break three-SBOMs case

May be true in general, but common sense tells me otherwise in this particular case. If you look at the actual functional code:
https://github.com/hermetoproject/cachi2/blob/2235854bfc5cb76ea168cb04d50be16efce9fa0a/cachi2/interface/cli.py#L459-L460

We are really only doing a string concatentation of JSONs and calling into the JSON formatter exactly once. I don't think testing 2 and 3 SBOM merges brings us any value because of this. Again, might make sense elsewhere but I don't think it makes it here. Additionally, if you look at our test suite, there's always only going to be more tests, so we really need to be careful and pragmatic for the number of variants we test to not go out of hand so as to make the test suite lose its meaning - provide instant local feedback (ideally) fast! This also goes in line with a recently filed issue https://github.com/hermetoproject/cachi2/issues/803 which is a very reasonable ask and a very real case, the tests are still taking ~17minutes (give or take), well, do they? Or can we get the feedback faster and be more productive in the meantime?

So, if possible, I'd like to keep this particular change in, because I don't think the parameter matrix as it currently is defined isn't IMO justified well.

tests/unit/test_cli.py
eskultety: It is unusual to request a context manager on a temporary file not to be deleted afterwards AND close it right on the next line, I'd be curious to learn what the reason was back then :).

tests/unit/test_cli.py
eskultety: Oh, one more thing - using and not using `parametrize` at the same time is also confusing and very unusual IMO which was another motivator for the change in question :) .

a-ovchinnikov: >  Tests are still code

Treating them as regular code (and not a very special kind) implies that they need tests. This is a bit paradoxical.

> any refactoring ... will result in many identical changes to several tests

I'd say this is an indication of brittle tests which are tightly coupled to an  implementation. Brittleness aside this usually indicates that the tests are, in fact, just a crude reimplementation of existing logic and exist solely to bump coverage numbers which is something to avoid. Tests which test through a contract, on the other hand, don't need to and must not to be changed during refactoring at all. Tests are supposed to verify external behavior, how this behavior is implemented is immaterial.

Pytest part reads a bit like an ad populum. While parametrization is an awesome tool for one-dimensional substitutions and should be used for that, I think multidimensional expansion is a tool that has to be used sparingly. Parametrization and overgeneralization lead to tests like those for gomod. It is a bit hard to tell what is going on in some of those because of the multitude of parameters supplied.

The other thing which I find problematic is that here we are effectively lumping together two distinct features into one test. Merging different types of SBOMs are different features and have been introduced separately in separate user stories, that's why I think it is worth it keeping their tests separate. The fact that they now share implementation is again an implementation detail and should not affect how tests are written.

>  when the time comes and we need to update the tests because of a code change

This means that our tests are tightly coupled to our implementation and that we are checking how we wrote some piece of code vs checking if it does what it is expected to do.

This all stems from the fact that we have not too many busyness rules on one hand and use a lot of external services on the other. This is somewhat similar to microservices which are usually heavily tested with integration tests and sporadically with unit and e2e. I believe we should revisit our approach to testing and more importantly to coverage calculation and interpretation.

This is definitely a bigger discussion than could (and should) fit into a cleanup PR thread.

tests/unit/test_cli.py
a-ovchinnikov: >  If you look at the actual functional code

This is exactly how tests get coupled to implementations. What is happening inside a unit is unit's busyness, a test writer should at least abstract that away.

I am sorry, I don't follow the IT argument: this is a unit test. UTs run for ~10 seconds on my machine, I'd say that is pretty fast. The full nox suit runs for about a minute which is a regression from tox where --parallel flag would make the entire test matrix finish in about 12 seconds.

I won't block on this one, but I still believe that small incremental tests (especially when each takes a tiny fraction of a second) are better than a lump of all features crammed into a single one. If tests _are_ code then lumped tests are the equivalent of enterprise-grade god-class which is demonstrably bad.

a-ovchinnikov (approved):

tests/unit/test_cli.py
eskultety: > I thought about this in the background a bit and realized that smaller, more compact test cases followed by an expanded tests that build on the previous ones act as a form of induction proof steps for me.

I would agree with this if this were tested e.g. as 1,N occurrences, but not 2,3 in particular. Anyway, let's agree to disagree, I think that's fair. I don't care about this particular change enough to continue this thread, so I'm dropping that commit.

a-ovchinnikov: Yes, sure, I have already +1-ed the change, this is more about articulating my approach (since similar discussions resurface now and then).

eskultety: Cool, merging then.

PR #838, opened 2025-02-18
Add user documentation for cargo package manager

slimreaper35:

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ when downloading Go modules. See [Go environment variables](https://go.dev/ref/m
 | Package manager                                | Ecosystem            |
 |------------------------------------------------|----------------------|
 | [bundler](#bundler)                            | Ruby                 |
+| [cargo](#cargo)                                | Rust                 |
 | [generic](#generic-fetcher)                    | N/A                  |
 | [gomod](#gomod)                                | Go                   |
 | [npm](#npm)                                    | JavaScript           |
@@ -153,6 +154,17 @@ on the dependencies specified in the [Gemfile](https://bundler.io/v2.5/man/gemfi
 Both files must be present in the source repository so you should check them into your git repository.
 
 See [docs/bundler.md](docs/bundler.md) for more details.
+
+### cargo
+
+<https://doc.rust-lang.org/cargo/>
+
+Cachi2 supports Cargo by invoking the Cargo CLI to fetch Rust dependencies as
+declared in your project, ensuring reproducible builds with explicitly defined
+versions. Make sure to have up-to-date [Cargo.lock](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html)
+present in your project.
+
+See [docs/cargo.md](docs/cargo.md) for more details.
 
 ### generic fetcher
 
--- a/docs/cargo.md
+++ b/docs/cargo.md
@@ -1,0 +1,66 @@
+# Cargo
+
+<https://doc.rust-lang.org/cargo/>
+
+## Prerequisites
+
+To use Cachi2 with Cargo locally, ensure you have Cargo binary installed on your
+system. Then, ensure that the **Cargo.toml** and **Cargo.lock** are in your
+project directory.
+
+## Usage
+
+Run the following commands in your terminal to prefetch your project's
+dependencies specified in the **Cargo.lock**. It must be synchronized with the **Cargo.toml**
+file. Otherwise, the command will fail.
+
+```bash
+cd path-to-your-rust-project
+cachi2 fetch-deps cargo
+```
+
+The default output directory is `cachi2-output`. You can change it by passing
+the `--output-dir` option for the `fetch-deps` command. See the help message
+for more information.
+
+After prefetching the dependencies, you can use the prefetched dependencies
+to build your project. Make sure to run the following command to update the
+`.cargo/config.toml` to use them. (If the file does not exist in your repository,
+it will be created).
+
+```bash
+cachi2 inject-files --for-output-dir /tmp/cachi2-output cachi2-output
+```
+
+Use `--for-output-dir` to specify the output directory you want to mount or copy
+to the container. The command will update the `.cargo/config.toml` file to use the
+prefetched dependencies from the specified directory.
+
+_There are no environment variables that need to be set for the build phase._
+
+## Hermetic build
+
+After using the `fetch-deps`, and `inject-files` commands to set up the directory,
+you can build your project hermetically. Here is an example of a Dockerfile with
+basic instructions to build a Rust project:
+
+```Dockerfile
+FROM docker.io/library/rust:latest
+
+WORKDIR /app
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -128,19 +128,18 @@ when downloading Go modules. See [Go environment variables](https://go.dev/ref/m
...
-* [pip](#pip)
-* [npm](#npm)
-* [yarn](#yarn)
-* [bundler](#bundler)
-* [generic fetcher](#generic-fetcher)
-* [cargo](#cargo)
-
-Planned:
-
-* dnf
```
eskultety, line 142: This commit could easily be proposed standalone and have no connection to the cargo document, it's a change that doesn't need to be held up by anything else or conversely shouldn't block the other change, so please open a dedicated PR for this.

README.md
```diff
@@ -128,19 +128,18 @@ when downloading Go modules. See [Go environment variables](https://go.dev/ref/m
...
+| Package manager                                | Programming language |
+|------------------------------------------------|----------------------|
+| [gomod](#gomod)                                | Go                   |
+| [pip](#pip)                                    | Python               |
+| [npm](#npm)                                    | JavaScript           |
+| [yarn](#yarn)                                  | JavaScript           |
+| *rpm                                           | N/A                  |
+| [bundler](#bundler)                            | Ruby                 |
+| [generic](#generic-fetcher)                    | N/A                  |
+| [cargo](#cargo)                                | Rust                 |
```
eskultety, line 140: Would you mind sorting the backends alphabetically since you're making this change?

README.md
```diff
@@ -128,19 +128,18 @@ when downloading Go modules. See [Go environment variables](https://go.dev/ref/m
...
-* [npm](#npm)
-* [yarn](#yarn)
-* [bundler](#bundler)
-* [generic fetcher](#generic-fetcher)
-* [cargo](#cargo)
-
-Planned:
-
-* dnf
+| Package manager                                | Programming language |
```
eskultety, line 131: I'd like to limit the number of N/A (won't apply to generic ever) in the table, does "package ecosystem" sound better than _Programming language_?

README.md
slimreaper35: Then it would also make sense to rearrange all those paragraphs below the table

README.md
eskultety: N/A is okay for generic, but not for RPMs, hence the "package ecosystem" suggestion. Once we adopt DNF as the backend the situation would repeat with N/A, whereas with ecosystem you simply put RPM in there. Similarly to e.g. DEB if we ever adopted Debian's package ecosystem, etc.

README.md
slimreaper35: Dropped

docs/cargo.md
```diff
@@ -0,0 +1,69 @@
...
+
+```bash
+cd path-to-your-rust-project
+cachi2 fetch-deps cargo
+```
+
+The default output directory is `cachi2-output`. You can change it by passing
+the `--output-dir` option for the `fetch-deps` command.
+
+In addition, the command will update the `.cargo/config.toml` to use the vendored
```
eskultety, line 31: This is the first time the word `vendored` is mentioned in this context. It may be confusing to the user to realize at this point what we're doing. I think this whole paragraph deserves a bit of a careful rewrite on how our fetching works and what it means for the user (i.e. the need to update their cargo config file).

eskultety (commented):
Pretty straight forward, I think with respect to #831 this doesn't need much adjustment, just some fine touches.

docs/cargo.md
```diff
@@ -0,0 +1,69 @@
...
+dependencies. (If the file does not exist, it will be created). At this point,
+the file contains a placeholder for the path to the vendored dependencies. Replace
```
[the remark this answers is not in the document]
slimreaper35: why not, but it should be changed in #831 at first

docs/cargo.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+The default output directory is `cachi2-output`. You can change it by passing
+the `--output-dir` option for the `fetch-deps` command. See the help message
+for more information.
+
+<!-- TOOD: Remove one of the following paragraphs. -->
+
+In addition, the command will update the `.cargo/config.toml` to use the prefetched
+dependencies. (If the file does not exist, it will be created). At this point,
```
a-ovchinnikov, line 29: This sounds a bit confusing: with my change the file is not modified during prefetch anymore, a project file template is generated instead which consists of existing data plus vendoring config. The second paragraph captures this behavior better.

eskultety: > This sounds a bit confusing: with my change the file is not modified during prefetch anymore

Correct behaviour, but the `.cargo/config.toml` is still touched somewhere in the logic, I'll mention that in the other PR.

slimreaper35: should be fixed

docs/cargo.md
```diff
@@ -0,0 +1,66 @@
...
+```Dockerfile
+FROM docker.io/library/rust:latest
+
+WORKDIR /app
+
+COPY Cargo.toml Cargo.lock .
+
+...
+
+RUN . /tmp/cachi2.env && cargo build --release
```
brunoapimentel, line 55: This will fail, because we do not generate a cachi2.env file.

slimreaper35: fixed

docs/cargo.md
```diff
@@ -0,0 +1,66 @@
...
+cachi2 fetch-deps cargo
+```
+
+The default output directory is `cachi2-output`. You can change it by passing
+the `--output-dir` option for the `fetch-deps` command. See the help message
+for more information.
+
+After prefetching the dependencies, you can use the prefetched dependencies
+to build your project. The following command will update the `.cargo/config.toml`
+to use them. (If the file does not exist, it will be created).
```
eskultety, line 28: > The following command will update the ...

I think ^this wording is too soft, meaning that it doesn't sound like something mandatory for the user to do. They won't learn about this until a few paragraphs later mentioned below in the 

> \#\# Hermetic build

section. I would personally go for something stronger, e.g. _Make sure you run the command bellow which ..._

Also, I think we can be a tiny bit more explicit about the modified TOML config file being modified in their own repo and say: _... will update the `.cargo/config.toml in your repository to use them._

docs/cargo.md
```diff
@@ -0,0 +1,66 @@
...
+
+COPY Cargo.toml Cargo.lock .
+
+...
+
+RUN cargo build --release
+```
+
+Assuming `cachi2-output` is in the same directory as the Dockerfile, build the
+container image:
```
eskultety, line 59: Please add _(don't forget mounting the cachi2 directory to the container build environment)_ just to be a bit more explicit.

eskultety (approved):
Nitpicks only -> ACK.

a-ovchinnikov (approved):

brunoapimentel (approved):

PR #841, opened 2025-02-24
Allow overriding verify_ssl in generic artifact downloader

jctanner:
Similar to rpm, this allows the user to override the generic package fetcher's asyncio.get method with `verify_ssl=False` given an input such as ...
```
cachi2 --log-level=debug --config-file=config.yaml fetch-deps --dev-package-managers \
'{"packages": [{"type": "generic", "options": {"ssl": {"ssl_verify": false}}}]}'
```

The input models file needed some re-arranging to resolve class dependencies.

### DIFF
--- a/cachi2/core/models/input.py
+++ b/cachi2/core/models/input.py
@@ -19,6 +19,97 @@ T = TypeVar("T")
 ModelT = TypeVar("ModelT", bound=pydantic.BaseModel)
 
 
+class SSLOptions(pydantic.BaseModel, extra="forbid"):
+    """SSL options model.
+
+    Defines extra options fields for client TLS authentication.
+    """
+
+    client_cert: Optional[str] = None
+    client_key: Optional[str] = None
+    ca_bundle: Optional[str] = None
+    ssl_verify: bool = True
+
+    @pydantic.field_validator("client_key", "client_cert", "ca_bundle")
+    @classmethod
+    def _validate_auth_file_paths(cls, val: str, info: pydantic.ValidationInfo) -> Optional[str]:
+        if val is None:
+            return val
+
+        if not Path(val).is_file():
+            raise ValueError(
+                (
+                    f"Specified ssl auth file '{info.field_name}':'{val}' is not a regular file.",
+                    "Make sure the file exists and that it has correct permissions.",
+                )
+            )
+
+        return val
+
+    @pydantic.model_validator(mode="after")
+    def _validate_ssl_options(self) -> Self:
+
+        cert_and_key = (self.client_cert, self.client_key)
+        if any(cert_and_key) and not all(cert_and_key):
+            raise ValueError(
+                "When using client certificates, client_key and client_cert must both be provided."
+            )
+
+        return self
+
+
+class ExtraOptions(pydantic.BaseModel, extra="forbid"):
+    """Global package manager extra options model.
+
+    This model takes care of carrying and parsing various kind of extra options that need to be
+    passed through to CLI commands/services we interact with underneath to tweak their behaviour
+    rather than our own. Each option set is namespaced by the corresponding tool/service it is
+    related to.
+
+    TODO: Enable this globally for all pkg managers not just the RpmPackageInput model.
+    """
+
+    dnf: Optional[Dict[Union[Literal["main"], str], Dict[str, Any]]] = None
+    ssl: Optional[SSLOptions] = None
+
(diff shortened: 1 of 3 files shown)


cachi2/core/models/input.py
```diff
@@ -19,6 +19,97 @@
 ModelT = TypeVar("ModelT", bound=pydantic.BaseModel)
 
 
+class SSLOptions(pydantic.BaseModel, extra="forbid"):
```
eskultety, line 22: You're redefining this class, one already exists on line 164.

cachi2/core/models/input.py
```diff
@@ -19,6 +19,97 @@
...
+            _raise_unexpected_type(str(dnf_opts), *prefixes)
+
+        for repo, repo_options in dnf_opts.items():
+            prefixes.append(repo)
+            if not isinstance(repo_options, dict):
+                _raise_unexpected_type(str(repo_options), *prefixes)
+
+        return data
+
+
```
eskultety, line 112: Seems like you need this code movement - please do so in a separate commit for better visibility and logical change separation.

cachi2/core/package_managers/general.py
```diff
@@ -81,7 +81,7 @@ async def _async_download_binary_file(
             f"aiohttp.ClientSession.get(url: {url}, timeout: {timeout}, raise_for_status: True)"
         )
         async with session.get(
-            url, timeout=timeout, auth=auth, raise_for_status=True, ssl=ssl_context
+            url, timeout=timeout, auth=auth, raise_for_status=True, verify_ssl=getattr(ssl_context, 'ssl_verify', True)
```
eskultety, line 84: This won't fly I'm afraid, with RPMs we may need to pass client TLS certificates in the context, the feature was never about disabling SSL verification, in fact, we only kept it in by accident, we originally didn't even plan on merging that setting :) .

Now, what we need to correctly handle this we'll need to expand the type that we pass to the `ssl` attribute. Looking at the [docs](https://docs.aiohttp.org/en/stable/client_reference.html#basic-api):

> ssl –
>
>SSL validation mode. True for default SSL check ([ssl.create_default_context()](https://docs.python.org/3/library/>ssl.html#ssl.create_default_context) is used), False for skip SSL certificate validation, >[aiohttp.Fingerprint]>(https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.Fingerprint) for >fingerprint validation, [ssl.SSLContext](https://docs.python.org/3/library/ssl.html#ssl.SSLContext) for custom SSL >certificate validation.
>
>Supersedes verify_ssl, ssl_context and fingerprint parameters.

IOW, we need to change our code so that we can successfully propagate the boolean value of `False` all the way to `async with session.get(..., ssl=ssl_context) # ssl_context being False` to do what you need. On the other hand, when you support the whole SSL options structure, you can't really choose to ignore the possibility of passing TLS certificates because it suddenly becomes supported in the input JSON, so we should additionally add proper SSL support to the generic fetcher backend. Would you mind updating the code to reflect that?

I guess passing certificates isn't an option for your use case? I don't remember from the top of my head, but I think we at least discussed the possibility of recording the fact that insecure download was used in the SBOM somehow, but I don't think that's the case, this is just a heads up what may come in the future - reporting some SBOM records as tainted.

cachi2/core/package_managers/generic/main.py
```diff
@@ -69,7 +74,7 @@ def _resolve_generic_lockfile(lockfile_path: Path, output_dir: RootedPath) -> li
         Path.mkdir(Path(artifact.filename).parent, parents=True, exist_ok=True)
         to_download[str(artifact.download_url)] = artifact.filename
 
-    asyncio.run(async_download_files(to_download, get_config().concurrency_limit))
+    asyncio.run(async_download_files(files_to_download=to_download, concurrency_limit=get_config().concurrency_limit, ssl_context=ssl_context))
```
eskultety, line 77: nitpick: Passing the first 2 by keyword is an unrelated change, those are declared as positional, not optional, so it should either be changed in the called function itself (in a different commit though) or just drop the changes up to `ssl_context` which you need.

eskultety (commented):
@jctanner thanks for the PR, couple of comments inline. Additionally, can you please add an integration test to verify the SSL functionality since we'd be enabling the whole thing for a new backend? We have an existing test using a DNF server in the repo which can get inspired by. Also, the generic backend is properly documented (unlike RPM), so the feature should be documented as well.

PR #842, opened 2025-02-24
Replace "pip-compile" for "uv pip compile" for requirements

slimreaper35:


noxfile.py
```diff
@@ -146,26 +146,23 @@ def generate_test_data(session: Session) -> None:
...
+    session.install("uv")
     cmd = [
-        "podman",
-        "run",
-        "--rm",
-        "--volume",
-        f"{PWD}:/cachi2:rw,Z",
-        "--workdir",
-        "/cachi2",
-        f"docker.io/library/python:{PYTHON_VERSION_MINIMAL}-alpine",
```
eskultety, line 166: ouch, I never noticed we used an env variable here, that should have never happened, because our pip-compile was supposed to run on 3.9 as that's the production python version we use, even our requirements files claim that:
```
  1 # This file is autogenerated by pip-compile with Python 3.9
```
which means that, luckily, we haven't updated these files from local dev environments for a while.

eskultety, line 166: I'm afraid you won't be able to get rid of the container just yet, because its sole purpose was to shield one from having to install python 3.9 on their devel machine just to be able to run pip-compile. Unit tests and linters are tested with different versions of python so we're covered both in upstream CI and locally.

noxfile.py
```diff
@@ -146,26 +146,23 @@ def generate_test_data(session: Session) -> None:
 @nox.session(name="pip-compile")
 def pip_compile(session: Session) -> None:
     """Update requirements.txt and requirements-extras.txt files."""
-    PWD = session.env["PWD"]
```
eskultety, line 149: **Commit message:**

> Closes https://github.com/hermetoproject/cachi2/issues/539

I don't think it does. We don't know what the underlying error was, but it was intermittent. Let me close it explicitly as WORKSFORME since we haven't been able to reproduce it since.

eskultety:
I'm missing the corresponding change in our GH actions :) . Overall, I don't mind giving it a try, I believe `uv` is the future to replace other existing linters we have, so why not. Just be wary that adopting `uv pip-compile` on its own won't solve the need for a local pip-compile container, it has a convenience meaning. If we were to adopt a faster migration policy wrt Python versions then I think we could consider dropping the container altogether since we wouldn't be forcing devs to install old Python versions just to update dependencies correctly and not break production deployments.

eskultety (commented):
I'm missing the corresponding change in our GH actions :) . Overall, I don't mind giving it a try, I believe uv is the future to replace other existing linters we have, so why not. Just be wary that adopting uv pip-compile on its own won't solve the need for a local pip-compile container, it has a convenience meaning. If we were to adopt a faster migration policy wrt Python versions then I think we could consider dropping the container altogether since we wouldn't be forcing devs to install old Python versions just to update dependencies correctly and not break production deployments.

eskultety (commented):
Seems non-intrusive, but most importantly, it's almost 5x faster \o/.

requirements.txt
```diff
@@ -1,93 +1,94 @@
...
-#    pip-compile --generate-hashes --output-file=requirements.txt pyproject.toml
-#
-aiohappyeyeballs==2.4.4 \
-    --hash=sha256:5fdd7d87889c63183afc18ce9271f9b0a7d32c2303e394468dd45d514a757745 \
-    --hash=sha256:a980909d50efcd44795c4afeca523296716d50cd756ddca6af8c65b996e27de8
+# This file was autogenerated by uv via the following command:
+#    uv pip compile --generate-hashes --output-file=requirements.txt --python=3.9 --refresh pyproject.toml
+aiohappyeyeballs==2.4.6 \
+    --hash=sha256:147ec992cf873d74f5062644332c539fcd42956dc69453fe5204195e560517e1 \
+    --hash=sha256:9b05052f9042985d32ecbe4b59a77ae19c006a78f1344d7fdad69d28ded3d0b0
```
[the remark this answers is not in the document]
eskultety: Did I just comment on on the exploded view from the merge commit that GH creates for PR instead of the actual commit? Sigh, looks like it - please ignore this, I should have done the review locally as I usually do rather than post comments on GH right away :(

eskultety (approved):

a-ovchinnikov (approved):

PR #847, opened 2025-02-26
[flake8] Add E731 to the ignored rules

eskultety:
E731 - Do not assign a lambda expression, use a def

We have a bunch of occurrences already and seems like this convenience helper pattern is to stay. Rather than putting #noqa directives everywhere let's just ignore the rule globally.

### DIFF
--- a/cachi2/core/models/sbom.py
+++ b/cachi2/core/models/sbom.py
@@ -196,11 +196,10 @@ class Sbom(pydantic.BaseModel):
             args = dict(annotator=annottr, annotationDate=now, annotationType="OTHER")
             pAnnotation = partial(SPDXPackageAnnotation, **args)
 
-            # noqa for trivial helpers.
-            mkcomm = lambda p: json.dumps(dict(name=f"{p.name}", value=f"{p.value}"))  # noqa: E731
-            hashdict = lambda c: dict(name=c.name, version=c.version, purl=c.purl)  # noqa: E731
+            mkcomm = lambda p: json.dumps(dict(name=f"{p.name}", value=f"{p.value}"))
+            hashdict = lambda c: dict(name=c.name, version=c.version, purl=c.purl)
             erefbase = dict(referenceCategory="PACKAGE-MANAGER", referenceType="purl")
-            erefdict = lambda c: dict(referenceLocator=c.purl, **erefbase)  # noqa: E731
+            erefdict = lambda c: dict(referenceLocator=c.purl, **erefbase)
 
             for component in libraries:
                 package_hash = SPDXPackage._calculate_package_hash_from_dict(hashdict(component))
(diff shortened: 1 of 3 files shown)


slimreaper35:
https://www.flake8rules.com/rules/E731.html
> The primary reason for this is debugging. Lambdas show as <lambda> in tracebacks, where functions will display the function’s name.

Are we okay with this trade-off?

eskultety:
> https://www.flake8rules.com/rules/E731.html
> 
> > The primary reason for this is debugging. Lambdas show as  in tracebacks, where functions will display the function’s name.
> 
> Are we okay with this trade-off?

Based on the fact that we already merged a bunch of these already to the main branch and I remember @a-ovchinnikov being fairly adamant about the lambda helpers then I'd say yes we are, that debugging note is negligible, not a real issue.

My motivation is just being annoyed by all those commentaries which can be avoided.

a-ovchinnikov:
> Are we okay with this trade-off?

As long as we all adhere to using named lambdas as a shorthand for cumbersome conditions in
list comprehensions we should be fine. I.e. module-level named lambda is most likely bad, a locally scoped one right above a comprehension is most likely fine (of course conditions apply). Any complex logic that has higher probability of failure should not be wrapped in named lambdas for the reasons @eskultety outlined above. I believe we can collectively do this easily and without any problem.

Initially I thought about adding a specific `noqa` on a per-module basis, just to highlight
that these lambdas are added consciously, but apparently  it is impossible to selectively block flake8 on a per-module basis, so this looks like the only way to do this. If I am wrong I suggest adding a proper switch to the top of the files.

slimreaper35 (approved):

eskultety:
> Initially I thought about adding a specific `noqa` on a per-module basis, just to highlight that these lambdas are added consciously, but apparently it is impossible to selectively block flake8 on a per-module basis, so this looks like the only way to do this. If I am wrong I suggest adding a proper switch to the top of the files.

Looks like rules can be applied only per-file or globally. Anyway, based on our review workflow where we tend to carefully review every change that goes in I'm not worried much :) . That said @a-ovchinnikov how is a module-level named lambda worse than local-scoped ones? Yeah, it's going to be weird during debugging and one will have to open the code next to the debugger (since ipdb doesn't have a TUI like gdb does) and it does annoy me (not gonna lie), but my intention here is not going and dropping all existing named lambdas, just to sort of formally adopt a rule that we've been disabling for a while already.

slimreaper35:
Let's disable it globally so it is visible to everyone. Anything else will lead to a bunch of "noqas" in the codebase.

a-ovchinnikov:
> How is a module-level named lambda worse than local-scoped ones?

They have a much wider scope and can potentially be imported by someone else.

These  lambdas are not supposed to be debugged and not supposed to raise any exceptions, all potentially raising activity should happen before their invocation. If there is a chance for such exception then a named lambda is a wrong tool.

PR #854, opened 2025-03-04
fix: gomod: fetch --tags fail

fmudry:
'git fetch --tags' fails in a Go repository with submodules in a konflux pipeline.
recurse_submodules="no" will prevent this error.

**Context**
When trying to build a Go repository containing at least one submodule with [konflux](https://konflux-ci.dev/), the pipeline would fail during the `prefetch-dependencies` stage, since cahi2 uses `git fetch --tags`. 
Additionally, the error during the konflux (Go repository with submodules) build doesn't happen, when `fetchTags` is enabled, see  this [PR](https://github.com/stolostron/observatorium-operator/pull/261), users will get this error:
```bash
If the issue seems to be on the Cachi2 side, please contact the maintainers.
Unregistering from: subscription.rhsm.redhat.com:443/subscription
System has been unregistered.
```

**Solution**
This is a workaround, since the root cause of the problem wasn't found. Using `--recurse-submodules=no`  prevents this error.

----------------------------------------------

How I tested it:
1. activated cachi2 venv 
2. created temp folder
```bash
mkdir ~/dev
````
3. moved into the ~/dev folder  
 ```bash 
cd ~/dev
```
4. created test.sh file with a konflux git-clone image and arguments based on the PR mentioned above which would clone the Go repository into a local enviroment 
```bash
#!/bin/bash
set -o errexit -o nounset -o pipefail

test_repo_path=${1?Missing argument: path where to clone the test repo}

mkdir -p "$test_repo_path"

podman run --rm -it \
    -v "$(realpath "$test_repo_path"):/test-repo:z" \
    -w "/test-repo" \
    --user root \
    quay.io/konflux-ci/git-clone@sha256:4e53ebd9242f05ca55bfc8d58b3363d8b9d9bc3ab439d9ab76cdbdf5b1fd42d9 \
        -url=https://github.com/jacobbaungard/observatorium-operator \
        -revision=konflux-213-fail-prefetch \
        -submodules=true
````  
5. cloned the repo with:
```bash
./test.sh repo
```
6. moved to repo folder
```bash
cd repo
```
7. ran cachi2 fetch deps
```bash
cachi2 fetch-deps gomod
```

For the first time it failed as expected. After adding recurse_submodules="no", deleting /repo folder and recreating steps from point 5 it worked.
(Note that you have to delete the /repo folder, since running `cachi2 fetch-deps gomod` for the second time works.)

### DIFF
--- a/cachi2/core/package_managers/gomod.py
+++ b/cachi2/core/package_managers/gomod.py
@@ -1252,7 +1252,7 @@ class ModuleVersionResolver:
         repo = git.Repo(repo_path)
         commit = repo.commit(repo.rev_parse("HEAD").hexsha)
         try:
-            repo.remote().fetch(force=True, tags=True)
+            repo.remote().fetch(force=True, tags=True, recurse_submodules="no")
         except Exception as ex:
             raise FetchError(
                 f"Failed to fetch the tags on the Git repository ({type(ex).__name__}) "


eskultety:
@fmudry just a word of caution, this is a public OSS project, the URLs you supplied in the description
are not publicly visible (though issues.redhat.com being resolvable) because those JIRA boards are not public , so please update the PR description by providing all relevant information explicitly in the description rather than posting URLs that are not accessible to everyone on the internet :) .

eskultety:
@fmudry can this be reproduced outside the environment you mentioned? i.e. a setup where plain `git clone`/ `git fetch` actions would yield the same error? If so, then please share that one as the official reproducer in the PR description. To add a bit more context to my ask, if the problem exists solely in Konflux's environment, but cannot be reproduced outside, then that would indicate that it needs to be addressed there too, i.e. in the build-definitions repo by injecting the `fetch.recurseSubmodules` setting to a git config file and making cachi2's git invocations pick it up automatically in the corresponding tekton task. Depending on the actual reproducer my suggestion may or may not be the actual correct place to address it, waiting for your input.
That said, we could investigate whether we can apply this kind of change here in this project globally, but the PR description would have to be framed differently for that, i.e. explaining why it is a good idea and okay for us to ignore submodules when fetching tags.

fmudry:
@eskultety Thank you for the clarification. I've updated the description. Let me know if more information is needed.

brunoapimentel:
> if the problem exists solely in Konflux's environment, but cannot be reproduced outside, then that would indicate that it needs to be addressed there too

It only happens when the repo is cloned by this custom git-init command [here](https://github.com/konflux-ci/build-definitions/blob/main/task/git-clone/0.1/git-clone.yaml#L235-L243) (which is defined [here](https://github.com/tektoncd-catalog/git-clone/tree/9fab698afab55734702f7a4af7b054c51b17101d/image/git-init)). It also seems to only affect some repos, and we don't know exactly what causes it.

@fmudry I though this [patch](https://github.com/konflux-ci/build-definitions/pull/1971) took care of the error, so it's not clear to me why we need to do it in Cachi2 as well :thinking:. Could you clarify?

fmudry:
This issue is being resolved on the [konflux side](https://github.com/konflux-ci/build-definitions/pull/2108).

PR #856, opened 2025-03-06
Update yarn.md - workaround when working with sources not fetched from a remore git repo

nickboldt:
Update yarn.md - workaround when working with sources not fetched from a remore git repo

### DIFF
--- a/docs/yarn.md
+++ b/docs/yarn.md
@@ -127,6 +127,11 @@ document has been completed, then it's essentially just a matter of cachi2 inter
 ``yarn install --mode=skip-build`` to fetch all dependencies (including transitive dependencies).
 
 ### Known pitfalls
+
+If your sources were not fetched from a remote git repo, Cachi2 will get confused about the name of 
+the main package. As workaround, run ``git init; git remote add origin git@url`` to give Cachi2 a
+value when it runs ``git remote show origin`` to define the name of the main package.
+
 If your repository isn't in a pristine state (i.e. you tried to run ``yarn install`` previously on
 your own without Cachi2) what may happen is that Cachi2 will assume the repository makes use of
 [Zero-Installs](#dealing-with-yarn-zero-installs). The workaround here is simple, just run ``yarn
--- a/docs/yarn.md
+++ b/docs/yarn.md
@@ -129,8 +129,9 @@ document has been completed, then it's essentially just a matter of cachi2 inter
 ### Known pitfalls
 
 If your sources were not fetched from a remote git repo, Cachi2 will get confused about the name of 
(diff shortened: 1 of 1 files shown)


docs/yarn.md
```diff
@@ -127,6 +127,12 @@ document has been completed, then it's essentially just a matter of cachi2 inter
 ``yarn install --mode=skip-build`` to fetch all dependencies (including transitive dependencies).
 
 ### Known pitfalls
+
+If your sources were not fetched from a remote git repo, Cachi2 will get confused about the name of 
+the main package, throwing an error: ``InvalidGitRepositoryError for file: URL``. As workaround, run 
+``git init; git remote add origin orgin https://github.com/somewhere/out-there/`` so that ``git remote show origin`` will return a value
```
gashcrumb, line 133: actually it turns out also an initial commit is needed as well, otherwise the command throws this error:

```text
2025-03-06 19:13:39,975 ERROR ValueError: Reference at 'refs/heads/main' does not exist
```

eskultety:
@nickboldt I don't think this is a pitfall and specifically one only related to Yarn; this is by design as git repositories are a vital source of information to make our SBOMs accurate, we might be left in the dark completely with some dependencies if we'd permit inputs which weren't a git repository. What you propose as a workaround would IMO lead to even worse UX because not only would a standard (i.e not a git) directory fail to be processed, our own documented workaround might not even work correctly, producing more errors and/or inaccurate data.
Looking at our usage docs, we **do** mention that input is supposed to be a git repo https://github.com/hermetoproject/cachi2/blob/fc04c06b23793701f43d5541d1d02281a10ed6e1/docs/usage.md?plain=1#L44, however, it may not be obvious at first glance. Therefore, I'd suggest you convert this docs PR to an update to the usage docs where you make this bit of information more visible, IOW making it a strict pre-requisite.

nickboldt:
True, it might make sense to turn this into a PR for the usage.md.

I can also try to spiff up weird statements like those in https://github.com/hermetoproject/cachi2/blob/fc04c06b23793701f43d5541d1d02281a10ed6e1/docs/usage.md?plain=1#L83-L84 about "don't worry about this flag for now" :rofl:

docs/yarn.md
nickboldt: workaround:
```
git init && git add -A && git commit -m "initial commit" && git remote add origin https://github.com/someorg/somerepo
```

nickboldt:
> @nickboldt are you going to follow up with the `usage.md` changes and change the scope of this PR?

See https://github.com/hermetoproject/cachi2/pull/862/files

PR #859, opened 2025-03-11
Replace backoff in elastic requests

slimreaper35:


a-ovchinnikov (approved):

taylormadore (approved):

tests/unit/package_managers/test_gomod.py
```diff
@@ -2238,11 +2237,8 @@ def test_retry_failure(
         assert mock_run.call_count == 5
         assert mock_sleep.call_count == 4
 
-        assert "Backing off run_go(...) for 1.0s" in caplog.text
-        assert "Backing off run_go(...) for 2.0s" in caplog.text
-        assert "Backing off run_go(...) for 4.0s" in caplog.text
-        assert "Backing off run_go(...) for 8.0s" in caplog.text
-        assert "Giving up run_go(...) after 5 tries" in caplog.text
+        expected = ["Running '['go', 'mod', 'download']'" for _ in range(5)]
+        assert expected == caplog.messages
```
eskultety, line 2241: Since we're changing this, do we want to keep asserting the output instead of asserting that we completed N backoff/retry loops?

a-ovchinnikov: Generally we should not be doing this at all, this effectively tests other project's code which should be done within the scope of the other project.

slimreaper35: Should I remove those lines?

eskultety: Yes please, but maybe in a separate commit for better visibility. Since it's unlikely that we'd do it all in a single go, instead whenever relevant tests are touched, we'd address it gradually.

eskultety (approved):
Given that the gomod backend is the only consumer of this logic I'm also more in favour of just creating a simple solution ourselves, but for the time being this is good enough and takes care of the issue at hand, we can replace it again in the future in an effort to decrease our dependency list.

PR #860, opened 2025-03-11
setuptools-scm fixes

eskultety:
This is an attempt to fix one version-related aspect of our container image builds. The last 2 patches are a bit controversial, so I'd be inclined to drop them if requested, but they fix a legitimate workflow issue.
See the patch descriptions for detailed info.

FWIW I used a dummy release on my fork to experiment with this, please give it a try: https://github.com/eskultety/cachi2/releases/tag/1.2.3-alpha

### DIFF
--- a/.containerignore
+++ b/.containerignore
@@ -1,6 +1,7 @@
 *
 
 !.git/
+!.git_archival.txt
 !cachi2/
 !pyproject.toml
 !requirements.txt
--- a/.git_archival.txt
+++ b/.git_archival.txt
@@ -1,0 +1,3 @@
+node: $Format:%H$
+node-date: $Format:%cI$
+describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+.git_archival.txt  export-subst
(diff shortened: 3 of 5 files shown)


a-ovchinnikov (approved):
LGTM, I don't find the second part of the PR controversial, it addresses a corner case and does not introduce much additional logic to maintain.

eskultety:
> LGTM, I don't find the second part of the PR controversial, it addresses a corner case and does not introduce much additional logic to maintain.

The thing is, it's not a corner case, it's actually a standard workflow, e.g. I use git worktrees a lot to be able to work on several branches in parallel due to frequent interrupts. That said, adopting a MANIFEST.in file when the official setuptools docs say it's no longer needed for most cases (unless adding non-source package data) it feels like rubbing the wrong way to add it, hence I said I could be convinced to drop it. It looks like though that there might be a bug in setuptools when combined with the SCM plugin that using a manifest seems to be the only way I could find which can create a proper sdist, in turn leading to a working cachi2 install inside a development container.

brunoapimentel (approved):
I'm fine with the approach here.

> It looks like though that there might be a bug in setuptools when combined with the SCM plugin that using a manifest seems to be the only way I could find which can create a proper sdist

Is it worth filing a bug in the setuptools-scm project?

eskultety:
> Is it worth filing a bug in the setuptools-scm project?

Absolutely, it's on my todo list but I need to create a small reproducer first. I also think that this falls into base setuptools rather than setuptools-scm, but I need to confirm with the reproducer, so yeah, I'll do that at some point (but looking at the setuptools issues that are related to MANIFEST it doesn't seem like something that would get fixed).
Thanks for the reviews.

slimreaper35 (approved):

PR #862, opened 2025-03-12
Update usage.md - workaround when working with sources not fetched from a remote git repo

nickboldt:
adds a pre-req section that calls out the need for a git repo to exist on disk already

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -30,9 +30,23 @@ In order to support this class of hermetic builds, not only does Cachi2 need to 
 build flows will need additional changes (i.e. leveraging defined [environment variables](#generate-environment-variables)
 or using Cachi2 to [inject project files](#inject-project-files)).
 
+### Pre-requirements
+
+The first step in creating hermetic builds is to have a git repo from which to get sources, as Cachi2 relies on git metadata.
+
+If your sources have not already been fetched from a remote git repo, Cachi2 will get confused about the name of 
+the main package, throwing an error: ``InvalidGitRepositoryError for file: URL``. 
+
+As workaround, run the following to create a local-only git repo:
+
+```shell
+git init && git add -A && git commit -m "initial commit" && \
+git remote add origin https://github.com/someorg/somerepo
+```
+
 ### Pre-fetch dependencies
 
-The first step in creating hermetic builds is to fetch the dependencies for one of the [supported package managers](../README.md#package-managers).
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -30,9 +30,23 @@ In order to support this class of hermetic builds, not only does Cachi2 need to
...
+The first step in creating hermetic builds is to have a git repo from which to get sources, as Cachi2 relies on git metadata.
+
+If your sources have not already been fetched from a remote git repo, Cachi2 will get confused about the name of 
+the main package, throwing an error: ``InvalidGitRepositoryError for file: URL``. 
+
+As workaround, run the following to create a local-only git repo:
+
+```shell
+git init && git add -A && git commit -m "initial commit" && \
+git remote add origin https://github.com/someorg/somerepo
```
eskultety, line 44: This is highly specific to one particular use case you've encountered and for everyone else reading this (myself included) this sounds very confusing because it lacks the context on why there's need to add a remote especially after the sentence one line above where you mention 

> a local-only git repo

I'd hence expect users to just perform a `git init` in that case, because, well, a local-only git repo is needed, why the remote?
Therefore, I'd skip these exact steps and if not for the reason I just mentioned then at least because any developer who's told by the official docs a git repo is needed on the input should be capable of configuring one following steps needed for their particular use case. 

To give you an idea on what I'd imagine being mentioned in the docs I'd reword this whole part a bit more carefully in order not to give users any sign of any other setup then the recommended one being officially supported. E.g.

_This is paramount for successful execution. If for some reason you don't have a git repository, e.g. you're trying to use cachi2 on an unpacked tarball, you may also get acceptable results by forcefully creating a git repository from it first. Make sure to set the remote as well. Note however, that this is only good for smoke testing a scenario and we can't guarantee any results without proper and correct git metadata._

I'd also make ^this only a footer in this doc rather than being an integral part of the docs text to make it 100% clear what our expectations on the input are and that this is merely a best effort type of thing.
@nickboldt how does ^that sound?

nickboldt: > why the remote?

My understanding is that the cachi2 process requires:
* a git repo
* with a remote origin
* and at least one commit so there's a history. 

See discussion in https://github.com/hermetoproject/cachi2/pull/856#pullrequestreview-2665406387

eskultety:
Closing as the docs change was incorporated in 7e2150efd

PR #867, opened 2025-03-17
Fix: exception message inside rooted_path.py and unit tests

derasdf:
Updated the exception message in the `re_root` function for the `PathOutsideRoot` error to be clearer and more user-friendly. Also, modified the unit tests to check just for the `PathOutsideRoot` error without relying on the exact message matching.

Closes https://github.com/hermetoproject/cachi2/issues/759

### DIFF
--- a/cachi2/core/rooted_path.py
+++ b/cachi2/core/rooted_path.py
@@ -89,11 +89,10 @@ class RootedPath(PathLike[str]):
         """
         subpath = self.path.joinpath(*other).resolve()
         if not subpath.is_relative_to(self.root):
-            s_other = str(Path(*other))
-            s_self = str(self)
-            s_root = str(self.root)
             raise PathOutsideRoot(
-                f"Path {s_self!r}/{s_other!r} outside {s_root!r}, refusing to proceed"
+                s_self=str(Path(*other)),
+                s_other=str(self.path),
+                s_root=str(self.root),
             )
         cls = type(self)
         return cls(subpath)
--- a/cachi2/core/rooted_path.py
+++ b/cachi2/core/rooted_path.py
@@ -93,7 +93,7 @@ class RootedPath(PathLike[str]):
             s_self = str(self)
             s_root = str(self.root)
             raise PathOutsideRoot(
-                f"Joining path {s_other!r} to {s_self!r}: target is outside {s_root!r}"
+                f"Path {s_self!r}/{s_other!r} outside {s_root!r}, refusing to proceed"
             )
         cls = type(self)
         return cls(subpath)
--- a/tests/unit/package_managers/test_generic.py
+++ b/tests/unit/package_managers/test_generic.py
@@ -185,7 +185,7 @@ def test_resolve_generic_no_lockfile(mock_load: mock.Mock, rooted_tmp_path: Root
         pytest.param(
             LOCKFILE_INVALID_FILENAME,
             PathOutsideRoot,
-            "target is outside",
+            "",
             id="invalid_filename",
         ),
         pytest.param(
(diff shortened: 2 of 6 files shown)


cachi2/core/rooted_path.py
```diff
@@ -93,7 +93,7 @@ def re_root(self: RootedPathT, *other: StrPath) -> RootedPathT:
             s_self = str(self)
             s_root = str(self.root)
             raise PathOutsideRoot(
-                f"Joining path {s_other!r} to {s_self!r}: target is outside {s_root!r}"
+                f"Path {s_self!r}/{s_other!r} outside {s_root!r}, refusing to proceed"
```
a-ovchinnikov, line 96: I wonder if it would make sense to lift message generation to exception class itself? In this case `PathOutsideRoot` would need to be called with just the paths.

tests/unit/package_managers/test_generic.py
```diff
@@ -185,7 +185,7 @@ def test_resolve_generic_no_lockfile(mock_load: mock.Mock, rooted_tmp_path: Root
         pytest.param(
             LOCKFILE_INVALID_FILENAME,
             PathOutsideRoot,
-            "target is outside",
+            "",
```
a-ovchinnikov, line 188: Out of scope for this change, but I think dropping direct string checks should be done everywhere in favor of checking error type. That would require a richer set of error types. I vaguely remember that not using a richer set of errors was a conscious design decision, maybe we should revisit it (@eskultety , @brunoapimentel)?

brunoapimentel: Yeah, one of the motivators was because we were trying to group errors in a way that it would be easier to generate metrics (for example, separate user errors from actual system errors).

I don't think we need to limit the error types, but maybe the original intention can be kept by using a smaller set of superclasses.

brunoapimentel (approved):

tests/unit/package_managers/test_generic.py
eskultety: > Yeah, one of the motivators was because we were trying to group errors in a way that it would be easier to generate metrics (for example, separate user errors from actual system errors).

@brunoapimentel do you happen to know approximately where this decision dates back to?

> I don't think we need to limit the error types, but maybe the original intention can be kept by using a smaller set of superclasses.

Yes, limiting error types is actually counter-productive when it comes to testing, because the more specialized types (within boundaries of course) we have, the easier it is to keep tests well maintained and target to expected error code paths, IOW one of the big pain points we currently have is checking verbatim error messages in our tests which, though doesn't seem like it, cause a significant problem when adding complex logic which affects most of the test cases leading to weird test failures whereas with more specialized exception types, one only needs to check the type of the expected exception.

a-ovchinnikov: Thanks. Filed https://github.com/hermetoproject/cachi2/issues/876 for exception classes.

cachi2/core/rooted_path.py
eskultety: I think this is a reasonable and simple ask, @derasdf can you please incorporate the suggestion?

tests/unit/package_managers/test_generic.py
a-ovchinnikov: Resolving to avoid confusion

cachi2/core/rooted_path.py
derasdf: Yeah, makes sense, i'll get to it.

eskultety (approved):

slimreaper35 (approved):

PR #869, opened 2025-03-18
Use temporary directory when resolving each package manager

slimreaper35:

### DIFF
--- a/cachi2/core/resolver.py
+++ b/cachi2/core/resolver.py
@@ -39,26 +39,22 @@ def resolve_packages(request: Request) -> RequestOutput:
     This function performs the operations in a working copy of the source directory in case
     a package manager that can make unwanted modifications will be used.
     """
-    if request.yarn_packages:
-        original_source_dir = request.source_dir
+    original_source_dir = request.source_dir
 
-        with TemporaryDirectory(".cachi2-source-copy", dir=".") as temp_dir:
-            source_backup = copy_directory(original_source_dir.path, Path(temp_dir).resolve())
+    with TemporaryDirectory(".cachi2-source-copy", dir=".") as temp_dir:
+        source_backup = copy_directory(original_source_dir.path, Path(temp_dir).resolve())
 
-            request.source_dir = RootedPath(source_backup)
-            output = _resolve_packages(request)
-            request.source_dir = original_source_dir
+        request.source_dir = RootedPath(source_backup)
+        output = _resolve_packages(request)
+        request.source_dir = original_source_dir
 
-            # Temporary solution to project files paths that are pointing to the work copy.
-            # Should be replaced once we extend the work copy solution to other package managers.
-            # https://github.com/hermetoproject/cachi2/issues/712
-            for project_file in output.build_config.project_files:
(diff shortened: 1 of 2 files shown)


eskultety:
This is on my radar for sure, I just want to get some more rebranding stuff out of hands so that I can switch focus on this one based on the context I have.

a-ovchinnikov (approved):

slimreaper35:
rebased

cachi2/core/resolver.py
```diff
@@ -39,26 +39,24 @@ def resolve_packages(request: Request) -> RequestOutput:
...
+        output = _resolve_packages(request)
+        request.source_dir = original_source_dir
 
-            # Temporary solution to project files paths that are pointing to the work copy.
-            # Should be replaced once we extend the work copy solution to other package managers.
-            # https://github.com/hermetoproject/cachi2/issues/712
-            for project_file in output.build_config.project_files:
+        for project_file in output.build_config.project_files:
+            # bundler config file is copied to the output directory, so we don't need to update the path
+            if "bundler" in project_file.template:
```
taylormadore, line 53: I just noticed this bug in an OSBS build, so this PR is very timely :heart:

taylormadore: I know bundler is currently our only case where a project file is in the output directory, but could we generalize this check to make it valid for all project files that are already in the output directory?

brunoapimentel:
When I wrote the solution before, I thought of it as a hotfix, because I imagined a scenario where the package manager code would be aware of the work copy path and the original path. But this is probably overcomplicating things, and with Taylor's [suggestion](https://github.com/hermetoproject/cachi2/pull/869#discussion_r2004173143), I believe this becomes more future-proof. 

LGTM

cachi2/core/resolver.py
slimreaper35: Done.

cachi2/core/resolver.py
```diff
@@ -39,26 +39,24 @@ def resolve_packages(request: Request) -> RequestOutput:
...
 
-            # Temporary solution to project files paths that are pointing to the work copy.
-            # Should be replaced once we extend the work copy solution to other package managers.
-            # https://github.com/hermetoproject/cachi2/issues/712
-            for project_file in output.build_config.project_files:
+        for project_file in output.build_config.project_files:
+            # if the project file is in the output directory, we don't need to update the path
+            if project_file.abspath.is_relative_to(request.output_dir):
+                pass
+            else:
```
eskultety, line 55: How about just
```
# we don't need to update paths if a project file already lies in the output directory
if not project_file.abspath.is_relative_to(request.output_dir):
    subpath = project_file.abspath.relative_to(source_backup)
    ...
```

eskultety (approved):
Turned out to be simpler than https://github.com/hermetoproject/cachi2/issues/712#issuecomment-2597889519 :) .
This also closes https://github.com/hermetoproject/cachi2/issues/707 so add that to your commit message

brunoapimentel (approved):

PR #871, opened 2025-03-18
[Rebranding] vol.1 : A tiny prelude

eskultety:
First round of rebranding related changes:
- small but unnoticeable fixes that otherwise impact the rebranding process
- put a rebranding notice in the main README for the time being

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,11 @@
+# 🚧 REBRANDING NOTICE 🚧
+**This project is currently undergoing rebranding
+(see https://github.com/orgs/hermetoproject/discussions/824).**
+
+**You will gradually notice changes to the code base and eventually - the repository name and
+release artifacts. We are taking steps to ensure backwards compatibility, so feel free to continue
+using the project as normal, stay tuned!**
+
 # Cachi2
 
 [![coverage][cachi2-coverage-badge]][cachi2-coverage-status]
--- a/cachi2/interface/cli.py
+++ b/cachi2/interface/cli.py
@@ -97,7 +97,7 @@ def version_callback(value: bool) -> None:
 
 @app.callback()
 @handle_errors
-def cachi2(  # noqa: D103; docstring becomes part of --help message
+def main(  # noqa: D103; docstring becomes part of --help message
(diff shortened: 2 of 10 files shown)


slimreaper35:
Third commit: Have you considered renaming [Cachi2Error](https://github.com/hermetoproject/cachi2/blob/main/cachi2/core/errors.py#L7) to a neutral name too? (for the same reason as you mentioned in the commit message)

eskultety:
> Third commit: Have you considered renaming [Cachi2Error](https://github.com/hermetoproject/cachi2/blob/main/cachi2/core/errors.py#L7) to a neutral name too? (for the same reason as you mentioned in the commit message)

I have but `Cachi2Error` is a top-level exception which is user facing. Even though we're not a library it is common for top-level exceptions to have some connotation to the project itself. I have a patch renaming this to `HermetoError` in vol.2 , so it'd be good to reach some consensus before proceeding. I'm ambivalent, I can imagine both approaches. In any case, it would happen in vol.2 anyway because commit 3/5 only replaces static string literals, not identifiers (there's going to be a commit dedicated to just that).

slimreaper35:
> I have but Cachi2Error is a top-level exception which is user facing.

Well, the docstring is misleading a bit, so I can imagine `InternalError` or `BaseError` would fit better. But if the consensus is to use `HermetoError` I am fine with that.
```
class Cachi2Error(Exception):
    """Root of the error hierarchy. Don't raise this directly, use more specific error types."""
```

a-ovchinnikov (approved):
LGTM, small and straightforward changes.

eskultety:
> > I have but Cachi2Error is a top-level exception which is user facing.
> 
> Well, the docstring is misleading a bit, so I can imagine `InternalError` or `BaseError` would fit better. But if the consensus is to use `HermetoError` I am fine with that.
> 
> ```
> class Cachi2Error(Exception):
>     """Root of the error hierarchy. Don't raise this directly, use more specific error types."""
> ```

`BaseError` is a good suggestion, like I said, it would happen in a different PR.

PR #873, opened 2025-03-18
cargo: Hide .cargo/config.toml before prefetching

slimreaper35:

### DIFF
--- a/cachi2/core/package_managers/cargo/main.py
+++ b/cachi2/core/package_managers/cargo/main.py
@@ -1,9 +1,10 @@
 import logging
+from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import cached_property
 from itertools import chain
 from pathlib import Path
-from typing import Optional
+from typing import Generator, Optional
 
 import tomlkit
 from packageurl import PackageURL
@@ -101,6 +102,24 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
         )
 
 
+@contextmanager
+def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None, None]:
+    """Hide the cargo config file if it exists.
+
+    The file may contain various settings that could result in potential attack vectors.
+    Therefore, it is better to "hide" it before running the `cargo vendor` command.
+    """
+    do_nothing = None
+
+    config = package_dir.join_within_root(".cargo/config.toml")
+    data = config.path.read_text() if config.path.exists() else None
+    config.path.unlink() if data is not None else do_nothing
+    try:
+        yield
+    finally:
+        config.path.write_text(data) if data is not None else do_nothing
+
+
 def _resolve_cargo_package(
     package_dir: RootedPath,
     output_dir: RootedPath,
@@ -110,8 +129,9 @@ def _resolve_cargo_package(
     vendor_dir = output_dir.join_within_root("deps/cargo")
     cmd = ["cargo", "vendor", "--locked", str(vendor_dir)]
     log.info("Fetching cargo dependencies at %s", package_dir)
-    # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
-    config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
+    with _hidden_cargo_config_file(package_dir):
+        # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
+        config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
 
     packages = _extract_package_info(package_dir.path / "Cargo.lock")
     main_package = _resolve_main_package(package_dir)


cachi2/core/package_managers/cargo/main.py
```diff
@@ -101,12 +101,25 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
...
+    config.path.unlink(missing_ok=True)
+
+
 def _resolve_cargo_package(
     package_dir: RootedPath,
     output_dir: RootedPath,
 ) -> tuple[chain[Component], dict]:
     """Resolve a single cargo package."""
     _verify_lockfile_is_present_or_fail(package_dir)
+    _delete_cargo_config_file(package_dir)
```
a-ovchinnikov, line 121: Is this the same `package_dir` that will be later used in `_use_verndored_sources`? If yes then we'll loose any additional configuration this way. I think this warrants a new e2e with a minimal package containing customizations in `config.toml`. It does feel like an overkill to run an entire e2e to verify a single side effect containment.

Alternatively this could be converted into a context manager, say `hidden_project_config` and the change could look like this:
```python
@contextmanager
def hidden_project_config(package_dir):
    p = Path(package_dir) / ".cargo" / "config.toml"
    data = p.read_text() if p.exists() else ""
    if data:
        p.unlink()
    try:
         yield
    finally:
        p.write_text(data) if data else None

... ... ...

    with hidden_project_config(package_dir):
        config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
```
This way while still having a nasty side effect we'll know its duration (by definition), we will be sure that it will be rolled back (again, by definition of this very side effect) and we will have a fall-back mechanism in place (platform's guarantee). (The other benefit is that we will decouple the concept of hiding a config from working on a project with a hidden config -- if someone someday decides that renaming a file would do as well the client side won't change and won't notice any difference).

slimreaper35: > Is this the same package_dir that will be later used in _use_verndored_sources? If yes then we'll loose any additional configuration this way.

I thought this was the desired state - make sure any cargo settings don't affect the cargo vendor command that we use for prefetching.

> I think this warrants a new e2e with a minimal package containing customizations in config.toml. It does feel like an overkill to run an entire e2e to verify a single side effect containment.

I planned to raise the question of whether this change deserves an integration test or at least a unit test.

> Alternatively this could be converted into a context manager, say hidden_project_config and the change could look like this:

Hmm, the approach looks good to me. :+1:

a-ovchinnikov: > I thought this was the desired state - make sure any cargo settings don't affect the cargo vendor command that we use for prefetching.

Yes, however we need the contents of the original `config.toml` later on, or we'll be overwriting it completely during inject phase otherwise: we attempt extending it with vendored sources, not replacing it, and this way we'll always replace it. (L64 generates templates for files to inject and relies on L154 for that which attempts to read the config which gets deleted by L62->L121 path, and this happens unconditionally).

a-ovchinnikov (approved):
LGTM. Inline comments are optionally actionable and exist mostly to share some ideas.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -101,17 +102,36 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
         )
 
 
+@contextmanager
+def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None, None]:
+    """Hide the cargo config file if it exists.
+
+    The file may contain various settings that could result in potential attack vectors.
+    Therefore, it is better to "hide" it before running the `cargo vendor` command.
+    Unlink the file if it exists, restore it after the command is executed.
```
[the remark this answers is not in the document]
slimreaper35: Removed the sentence. Yes, repeating the code in the docs is not a good idea.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -101,17 +102,37 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
...
 
 
+@contextmanager
+def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None, None]:
+    """Hide the cargo config file if it exists.
+
+    The file may contain various settings that could result in potential attack vectors.
+    Therefore, it is better to "hide" it before running the `cargo vendor` command.
+    """
+    do_nothing = None
```
a-ovchinnikov, line 112: Since you have added it, could you please move the alias to the outer scope or even maybe to utils? (This is not a blocker, the change still LGTM)

taylormadore (approved):

cachi2/core/package_managers/cargo/main.py
slimreaper35: Do you mean like a global constant ?

cachi2/core/package_managers/cargo/main.py
```diff
@@ -101,17 +102,37 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
...
+        config.path.write_text(data) if data is not None else do_nothing
+
+
 def _resolve_cargo_package(
     package_dir: RootedPath,
     output_dir: RootedPath,
 ) -> tuple[chain[Component], dict]:
     """Resolve a single cargo package."""
     _verify_lockfile_is_present_or_fail(package_dir)
+
```
eskultety, line 129: nitpick: extra new line

eskultety (approved):
ACK as is or even with `do_nothing` alias dropped completely, there needs to be balance between what shortcuts we define, e.g. simple named lambdas, not renaming builtins, keywords, base types, etc.

cachi2/core/package_managers/cargo/main.py
slimreaper35: I have concerns (after opening many PRs to remove dead code) that such a constant will never be used again.

a-ovchinnikov: Ok, definitely not a blocker, could be merged as is. `pass` won't work in this case since this requires a value and `pass` is not a value.

eskultety: > pass won't work in this case since this requires a value and pass is not a value.

It was meant in the broader context ;)

PR #874, opened 2025-03-18
[Rebranding] vol.2 : Straightforward string replacements

eskultety:
None of the strings replaced are user facing, so error messages are still reporting "cachi2" in them.
I also followed https://github.com/hermetoproject/cachi2/pull/871#issuecomment-2732590110 since there weren't any other comments/objections and reworked my initial approach to incorporate more "generic" words as replacements in several places where the project reference felt a bit off, make sure you review those particular replacements carefully to double check that the sentences still read like English :).

# Maintainers will complete the following section

- [ ] Commit messages are descriptive enough
- [ ] Code coverage from testing does not decrease and new code is covered
- [ ] Docs updated (if applicable)
- [ ] Docs links in the code are still valid (if docs were updated)

**Note:** if the contribution is external (not from an organization member), the CI
pipeline will not run automatically. After verifying that the CI is safe to run:

- [approve GitHub Actions workflows][approve-gh-actions] by clicking a button
- approve the Red Hat Trusted App Pipeline container build by commenting `/ok-to-test`
  (as is the [standard for Pipelines as Code][pac-running-pipeline])

[approve-gh-actions]: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
[pac-running-pipeline]: https://pipelinesascode.com/docs/guide/running/#running-the-pipelinerun

PR #875, opened 2025-03-18
[Rebranding] vol.3 - Rising action: The messy name replacements

eskultety:
Depends on: #874 


eskultety:
The CI failures are not caused by this PR, some dependency of ours broke the pip install with hashes.

https://github.com/hermetoproject/cachi2/actions/runs/14036450875/job/39295509671

eskultety:
> The CI failures are not caused by this PR, some dependency of ours broke the pip install with hashes.
> 
> https://github.com/hermetoproject/cachi2/actions/runs/14036450875/job/39295509671

If you have a working local env you can test the PR locally to verify.

slimreaper35 (approved):

tests/unit/models/test_sbom.py
```diff
@@ -533,22 +541,27 @@ def test_cyclonedx_sbom_can_be_converted_to_spdx_and_back_without_loosing_any_da
             ]
         )
         cyclonedx_sbom = sbom.to_spdx("NOASSERTION").to_cyclonedx()
+
+        # We have to re-order components because 'hermeto' now comes alphabetically later
+        # TODO: Stop storing components/properties as a list and use sets; that, however, will
```
a-ovchinnikov, line 546: We can keep them as lists, but provide a custom `__eq__`.

a-ovchinnikov (approved):

cachi2/__init__.py
```diff
@@ -0,0 +1,22 @@
+import importlib.util as imputil
+import sys
+from functools import cache
+from pathlib import Path
+
+
+@cache
```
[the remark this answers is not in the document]
eskultety: Early development remnant, this looked very differently when I started. I may have left there either by accident or as a safety if for some reason it wasn't called once only, I honestly don't remember anymore. I can delete it.

tests/unit/models/test_sbom.py
eskultety: Hmm, that would work. I guess it might also turn out to be less messy than conversions, but one must not forget to update the custom `__eq__` if a new field is added/recorded in the representation, otherwise it may end up being a bigger mess in debugging than array->dict conversions.
I'll leave the commentary as is for now, it's harmless, the TODO note on its own is important, the rest is implementation detail and we can do as you suggest :).

a-ovchinnikov: Sure, let's keep it and figure out the best way once it is addressed.

eskultety:
Since last revision:
- had to introduce a new commit to resolve the flake8 crashes in GH CI which were only reproducible in pristine container envs but not in my local venv
- cherry-picked https://github.com/hermetoproject/cachi2/pull/884/commits/1fa36a86662617382d0af5f98e6ee6fadd9718e8 because, again, pristine CI container environments kept failing, so I squashed it in patch 4

Please re-review.

a-ovchinnikov (approved):

PR #878, opened 2025-03-20
Containerfile: Maximize layer caching

eskultety:
This is mainly about commit 140881e42fd29cecc872407bf5d07328612a6305 which prioritized storage considerations and  a clean overview (due to the dangling images being opaque) over efficiency which, in hindsight, was a **detrimental** decision wrt integration test suite performance. On its own though we'd still not follow best practices when it comes to layer caching and so additional patches are added on top. All patches combined resulted in decreasing a single e2e test execution (`yarn_e2e_test_multiple_packages` was used) from 86s -> 18s (>4x speedup) on my machine using the `throughput-performance` tuned profile.

We could still squeeze some more out of it by doing a similar thing for the test application image builds (which are still removed at the end of the run) with not-as-significant returns, but this PR is supposed to be merely a simple quick hotfix with minimum "new" changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,11 +34,14 @@ RUN dnf -y install \
     python3-setuptools \
     && dnf clean all
 
-COPY . .
+# Install dependencies in a separate layer to maximize layer caching
+COPY requirements.txt .
 RUN python3 -m venv /venv && \
     /venv/bin/pip install --upgrade pip && \
-    /venv/bin/pip install -r requirements.txt --no-deps --no-cache-dir --require-hashes && \
-    /venv/bin/pip install --no-cache-dir .
+    /venv/bin/pip install -r requirements.txt --no-deps --no-cache-dir --require-hashes
+
+COPY . .
+RUN /venv/bin/pip install --no-cache-dir .
 
 ##########################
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -24,12 +26,15 @@ RUN dnf -y install \
...
 COPY . .
+
+# Keep the list of dependencies alphabetically sorted
 RUN dnf -y install \
     --setopt install_weak_deps=0 \
     --nodocs \
-    gcc \
-    # not a build dependency, but we copy the binary to the final image
+    # cargo is not a build dependency, but it pulls in the Rust compiler & library adding >200M
+    # of installed size which is excessive when we only need a single binary which we'll copy over
```
a-ovchinnikov, line 34: We don't need it _yet_. Once we get to building Python dependencies with `rustc` we'll need these. Obligatory not a blocker, will be dealt separately.

a-ovchinnikov (approved):

Dockerfile
eskultety: Oh no! ... Also, not entirely unexpected :)
I'll drop this hunk then since the effort you mention will land in the foreseeable future.

brunoapimentel (approved):

PR #884, opened 2025-03-24
[Rebranding] vol. 4 - Climax: Adopt the new name for our project

eskultety:
Depends on https://github.com/hermetoproject/cachi2/pull/875


eskultety:
Rebased on freshly merged #875 . Some changes had to be incorporated into #875, so technically it could have been squeezed to #885, but I just kept it as is to keep things simple.

noxfile.py
```diff
@@ -53,39 +53,39 @@ def parse_supported_python_versions() -> list[str]:
 
 @nox.session()
 def bandit(session: Session) -> None:
-    """Run bandit on cachi2 directory and noxfile.py."""
+    """Run bandit."""
```
brunoapimentel, line 56: Are we changing the references to the `cachi2` symlink in vol5?
```
 cmd = "bandit -c pyproject.toml -r cachi2 noxfile.py"
```
 (in line 58 right below this comment)

eskultety: Good catch! Nope, rebase issue artifact...turns out, I forgot the change in stash and forgot to re-apply it:
```
stash@{0}: WIP on (no branch): d898b3b4 Rebranding: Rename the top-level source directory cachi2 -> hermeto

diff --git a/noxfile.py b/noxfile.py
index 1a6d8f3e..44c99519 100644
--- a/noxfile.py
+++ b/noxfile.py
@@ -55,7 +55,7 @@ def parse_supported_python_versions() -> list[str]:
 def bandit(session: Session) -> None:
     """Run bandit on cachi2 directory and noxfile.py."""
     install_requirements(session)
-    cmd = "bandit -c pyproject.toml -r cachi2 noxfile.py"
+    cmd = "bandit -c pyproject.toml -r hermeto noxfile.py"
     session.run(*cmd.split(), *session.posargs, silent=True)
 
 
@@ -63,7 +63,7 @@ def bandit(session: Session) -> None:
 def black(session: Session) -> None:
     """Run black on cachi2 and tests directories and noxfile.py."""
     install_requirements(session)
-    cmd = "black --check --diff cachi2 tests noxfile.py"
+    cmd = "black --check --diff hermeto tests noxfile.py"
     session.run(*cmd.split(), *session.posargs, silent=True)
 ...
```

eskultety: fixed.

brunoapimentel (approved):

slimreaper35:
Would you change [netrc](https://github.com/hermetoproject/cachi2/pull/884/files#diff-f7a16a65f061822bcc73b8296f4dc837353d379d8d9cc5307982cb6941442835L108) content to something generic too?

slimreaper35 (approved):

eskultety:
> Would you change [netrc](https://github.com/hermetoproject/cachi2/pull/884/files#diff-f7a16a65f061822bcc73b8296f4dc837353d379d8d9cc5307982cb6941442835L108) content to something generic too? EDIT: I see it requires many changes to the integration test, so let's skip it now.

Yes, I have some work on a local branch, but yes I'll have regenerate the container server configuration.

noxfile.py
```diff
@@ -53,49 +53,49 @@ def parse_supported_python_versions() -> list[str]:
...
     install_requirements(session)
-    cmd = "black --check --diff cachi2 tests noxfile.py"
+    cmd = "black --check --diff hermeto tests noxfile.py"
     session.run(*cmd.split(), *session.posargs, silent=True)
 
 
 @nox.session()
 def flake8(session: Session) -> None:
-    """Run flake8 on cachi2 and tests directories and noxfile.py."""
+    """Run flake8."""
```
a-ovchinnikov, line 72: These docstrings are not really helpful and could be dropped entirely, should we just turn off the corresponding warning for flake8 for this module? (out of scope for this change, this non-actionable)

a-ovchinnikov (commented):
One minute too late!

noxfile.py
eskultety: I'd rather satisfy flake with unhelpful doc strings since the code is self explanatory rather than disabling rules. It's cheaper IMO and more transparent.

PR #887, opened 2025-03-24
[CI] noxfile: Use --no-deps with pip requirements installation

eskultety:
The upstream CI fails with:

    ERROR: In --require-hashes mode, all requirements must have their
    versions pinned with ==. These do not: coverage[toml]>=7.5

By default, pip tries to resolve transitive dependencies of packages when reading the requirements files. Well, that fails if the 'require-hashes' mode is ON. In case one uses a fully resolved and pinned requirements file (we do), pip shouldn't need to spend time trying to re-resolve something that's already resolved, it should use it directly.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -34,7 +34,7 @@ nox.options.default_venv_backend = "venv"
 
 def install_requirements(session: Session) -> None:
     """Install requirements for all sessions."""
-    session.install("-r", "requirements-extras.txt")
+    session.install("--no-deps", "-r", "requirements-extras.txt")
 
 
 def parse_supported_python_versions() -> list[str]:


noxfile.py
```diff
@@ -34,7 +34,7 @@
 
 def install_requirements(session: Session) -> None:
     """Install requirements for all sessions."""
-    session.install("-r", "requirements-extras.txt")
+    session.install("-r", "--no-deps", "requirements-extras.txt")
```
a-ovchinnikov, line 37: What bothers me is that this issue started appearing out of the blue, that some combo of dependencies changes uncovered it.

a-ovchinnikov (approved):

noxfile.py
slimreaper35:
```suggestion
    session.install("--no-deps", "-r", "requirements-extras.txt")
```

eskultety: Sigh, I tested it in an existing venv  so I didn't notice, thanks @slimreaper35!

> What bothers me is that this issue started appearing out of the blue, that some combo of dependencies changes uncovered it.

Exactly, that's the one thing I can't explain, i.e. what exactly led to this, but this option does the right thing according to the docs.

slimreaper35 (approved):

PR #889, opened 2025-03-25
Rust extensions for Python packages

a-ovchinnikov:
Adds support for Rust extensions for Python.

Closes https://github.com/hermetoproject/hermeto/issues/798


eskultety (commented):
Just some top-level observations.

a-ovchinnikov:
Second revision:
 - Applied suggestions;
 - Switched from build layer to UBI for testing;
Remaining items:
 - Extend documentation;
 - Move test branch from my fork to the main repository.

cachi2/core/package_managers/cargo/main.py
```diff
@@ -127,20 +130,24 @@ def _resolve_cargo_package(
...
 
     packages = _extract_package_info(package_dir.path / "Cargo.lock")
     main_package = _resolve_main_package(package_dir)
-    vcs_url = get_repo_id(package_dir.root).as_vcs_url_qualifier()
-    is_a_dep = lambda p: p["name"] != main_package["name"]
-    deps_components = (
-        CargoPackage(**p, vcs_url=vcs_url).to_component() for p in packages if is_a_dep(p)
-    )
+    is_a_dep = lambda p: p["name"] != main_package["name"]  # a shorthand, thus # noqa: E731
+    deps_components = (CargoPackage(**p).to_component() for p in packages if is_a_dep(p))
```
a-ovchinnikov, line 144: This looks like an incorrectly resolved merge conflict, will fix in the next revision.

noxfile.py
```diff
@@ -108,7 +108,7 @@ def _run_integration_tests(session: Session, env: dict[str, str]) -> None:
     netrc = "machine 127.0.0.1 login cachi2-user password cachi2-pass"
     default_env = {"HERMETO_TEST_NETRC_CONTENT": os.getenv("HERMETO_TEST_NETRC_CONTENT", netrc)}
     default_env.update(env)
-    cmd = "pytest --log-cli-level=WARNING tests/integration"
+    cmd = "pytest --log-cli-level=WARNING -W ignore::DeprecationWarning tests/integration"
```
a-ovchinnikov, line 111: This one sneaked in, but I do find it useful. If someone else finds it useful I'll re-home this change into a separate commit else I'll drop it in the next revision.

cachi2/core/package_managers/pip.py
```diff
@@ -2168,9 +2197,44 @@ def _version(dep: dict[str, Any]) -> str:
...
+        fname = p.get("path")
+        try:
+            tf = tarfile.open(fname)
+        except tarfile.ReadError:
+            log.warning(f"Failed to read '{fname}' as a tar file")
+            continue
+        for member in tf.getmembers():
+            # Top-level toml works just fine.
+            # examples is a dirty hack around setuptools-rust.
+            if member.name.endswith("Cargo.toml") and "examples" not in member.name:
```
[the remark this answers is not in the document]
eskultety: > I am not aware of any strict rules for placing Cargo.toml, it will likely be in the root of a source tree

Would this snippet from the [docs](https://doc.rust-lang.org/cargo/guide/project-layout.html) be enough of an authoritative answer?:

> Cargo.toml and Cargo.lock are stored in the root of your package (package root).

So it seems like the check I proposed should suffice then.

If you want to make it even more robust, then you could just grep for any `Cargo.toml` file in the tarball (including setuptools-rust), IOW use your current check without the examples filter, extract the tarball, see if there's a package directory (there is one for setuptools-rust) and run
`cargo locate-project` from there.

```
# https://doc.rust-lang.org/cargo/commands/cargo-locate-project.html searches upwards
/setuptools_rust-1.11.0/setuptools_rust # cargo locate-project
error: could not find `Cargo.toml` in `/setuptools_rust-1.11.0/setuptools_rust` or any parent directory
/setuptools_rust-1.11.0 /setuptools_rust # echo $?
101
```
That should give us IMO effectively as much robustness as possible, but that's just an idea, I'd be okay to start with a plain top-level Cargo.toml check and harden the solution further if we ever encounter an issue.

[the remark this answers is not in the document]
a-ovchinnikov: Ok, I have updated the code following your original suggestion. It seems to work fine.

noxfile.py
a-ovchinnikov: Moved to a separate commit. Will drop in case there is any opposition to it.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -10,7 +10,7 @@
 from packageurl import PackageURL
 from tomlkit.toml_file import TOMLFile
 
-from hermeto.core.errors import PackageRejected
+from hermeto.core.errors import NotAGitRepoPackageRejected, PackageRejected
```
eskultety, line 13: **Commit message:**
The msg is fairly humble with words on the proposed implementation. It could be IMO more verbose on providing some top-level overview on how this is being implemented, something like:
- we fetch Python package sdists like normal
- we identify which packages use Rust extensions and unpack the tarballs to the output deps directory
- call the cargo backend over the Rust extension packages from within the output deps directory
- vendor the Rust extensions packages like regular Rust packages

hermeto/core/package_managers/cargo/main.py
```diff
@@ -127,20 +129,25 @@ def _resolve_cargo_package(
...
     with _hidden_cargo_config_file(package_dir):
         # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
         config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
 
     packages = _extract_package_info(package_dir.path / "Cargo.lock")
     main_package = _resolve_main_package(package_dir)
-    vcs_url = get_repo_id(package_dir.root).as_vcs_url_qualifier()
     is_a_dep = lambda p: p["name"] != main_package["name"]
+    try:
+        vcs_url = get_repo_id(package_dir.root).as_vcs_url_qualifier()
```
eskultety, line 144: Please remind me why we always populate the vcs_url, I seem to have forgotten.

eskultety:
One more thing, I think it's time to open a PR against the integration-tests repo for this PR to consume the test case from there directly instead of your fork.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -95,6 +95,8 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
     # contain just a workspace and could even lack a name) could arrive without
     # a lock file. A user could try and fix this by explicitly locking the
     # package first.
+    if not package_dir.path.exists():
+        raise ValueError(f"Path does not exist: {package_dir.path=}")
```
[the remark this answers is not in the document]
slimreaper35: What is an extra `=` at the end ?

hermeto/core/package_managers/cargo/main.py
slimreaper35: PURL qualifier - https://github.com/package-url/purl-spec/blob/main/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairs

hermeto/core/package_managers/cargo/main.py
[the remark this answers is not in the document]
eskultety: > What is an extra `=` at the end ?

It'll make the string evaluate like this: `Path does not exist: package_dir.path=<package_dir.path value>`

Which actually makes me think, that's useful debugging, but this ends up being a user-facing error, right @a-ovchinnikov ?

tests/integration/conftest.py
```diff
@@ -19,7 +19,8 @@
 
 @pytest.fixture(scope="session")
 def test_repo_dir(tmp_path_factory: pytest.FixtureRequest) -> Path:
-    test_repo_url = "https://github.com/cachito-testing/integration-tests.git"
+    # test_repo_url = "https://github.com/cachito-testing/integration-tests.git"
```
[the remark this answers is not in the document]
a-ovchinnikov: Yes! (removed in the new revision).

hermeto/core/package_managers/cargo/main.py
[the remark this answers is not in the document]
a-ovchinnikov: > right?
Yes, that's correct, it is about to be dropped.

docs/pip.md
```diff
@@ -8,6 +8,7 @@
 * [Project metadata](#project-metadata)
 * [Distribution formats](#distribution-formats)
 * [Using fetched dependencies](#using-fetched-dependencies)
+* [Working with Rust-based extension](#working-with-rust-based-extensions)
```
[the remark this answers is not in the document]
a-ovchinnikov: Thank you for catching this!

hermeto/core/package_managers/pip.py
```diff
@@ -214,12 +220,29 @@ def fetch_pip_source(request: Request) -> RequestOutput:
...
+def _find_and_fetch_rust_dependencies(
+    request: Request, packages_containing_rust_code: list
+) -> RequestOutput:
+    if packages_containing_rust_code:
+        # Need to swap source for output since this should be happening within output_dir
+        cargo_request = request.model_copy(
+            update={"packages": [], "source_dir": request.output_dir}
+        )
+        cargo_request.packages = packages_containing_rust_code
+        return fetch_cargo_source(cargo_request)
```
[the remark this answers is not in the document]
eskultety: Like I said, food for thought, I wanted to make sure that we're aligned with future expectations when it ever comes to it.

docs/pip.md
```diff
@@ -8,6 +8,7 @@
 * [Project metadata](#project-metadata)
 * [Distribution formats](#distribution-formats)
 * [Using fetched dependencies](#using-fetched-dependencies)
+* [Working with Rust-based extension](#rust-based-deps)
```
slimreaper35, line 11:
```diff
-- * [Working with Rust-based extension](#rust-based-deps)
++ * [Working with Rust-based extensions](#working-with-rust-based-extensions)
```

docs/pip.md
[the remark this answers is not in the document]
slimreaper35:
```diff
-- * [Working with Rust-based extension](#rust-based-deps)
++ * [Working with Rust-based extensions](#working-with-rust-based-extensions)
```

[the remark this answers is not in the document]
slimreaper35: The link should match the heading title (converted to lowercase).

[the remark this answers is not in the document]
a-ovchinnikov: I added an explicit link down in the document: I assumed that the hyphen in the middle of the name was the problem. Now I see that it was, in fact, a missing 's' in the table of contents item.

slimreaper35:
So the only missing piece is to finish the integration test?

a-ovchinnikov:
> So the only missing piece is to finish the integration test?

Sorry, why is it missing? I have already updated the branch and the test is passing on the gate now. One piece that must be updated before merging is the documentation part -- it turned out more confusing than I intended.

docs/pip.md
```diff
@@ -484,6 +485,50 @@ hermeto-output/deps/pip
...
+processed. Note, that this step requires internet connection.
+
+Once requirements-build.txt is populated fetch could be done as usual. Hermeto will fetch all
+runtime and build time dependencies for both Python and Rust parts.
+
+To build such Python package one would need to supply a proper `.cargo/config.toml`. It has
+to be provided manually because the build systems in use appear to be very sensitive to its
+placement which could change between fetch and build phases. `.cargo` directory must be
+created on the same level as `hermeto-output` directory and must contain a `config.toml`
+file of the following contents:
```
eskultety, line 512: Somehow I didn't connect the dots in my head when I saw this being done in the test Dockerfile, it was only I read this paragraph that made me realize that. Can't we achieve this in the inject-files command? We can make a note on rust extensions being used by the repo in the build config file and we know where to create it.

docs/pip.md
```diff
@@ -484,6 +485,50 @@ hermeto-output/deps/pip
...
+```
+[source.crates-io]
+replace-with = "local"
+
+[source.local]
+directory = "<path_to_project>/<hermeto_output>/deps/cargo"
+```
+
+where _path_to_project_ is the path to a directory containing the project to be built and
+_hermeto_output_ is the name of an output directory (defaults to `hermeto-output`).
```
eskultety, line 523: Does this mean that the fetch dir needs to be mounted to the project dir inside the build container env? Usually that's not the case, the fetch dir ends up somewhere under `/tmp` while the input sources are elsewhere, e.g. under `/<project_dir>`.

eskultety: That said, I guess it makes sense as a general note for local environments where container builds are not being utilized.

docs/pip.md
```diff
@@ -484,6 +485,50 @@ hermeto-output/deps/pip
 └── ...
 ```
 
+## Working with Rust-based dependencies
+
+Hermeto provides a way to prepare hermetic build of a Python package which depends on
+on Rust-based packages. This process cannot be fully automated on Hermeto's side, thus it would
```
eskultety, line 491: s/on on/on

docs/pip.md
```diff
@@ -484,6 +485,50 @@ hermeto-output/deps/pip
...
+require minor intervention from users.
+
+Building such project requires that all build dependencies are present. To achieve this
+[pybuild-deps](https://pypi.org/project/pybuild-deps) could be used. The following command
+will take care of generating the list:
+
+```
+echo <package_name==package_version> | pybuild-deps compile /dev/stdin --generate-hashes -o requirements-build.txt
```
eskultety, line 499: Is this the recommended invocation? So one cannot specify the package spec as a string on the command line and instead it needs to be read from stdin? That doesn't sound very robust or convenient and we should improve that since now it falls into our maintenance domain.

eskultety:
> > So the only missing piece is to finish the integration test?
> 
> Sorry, why is it missing? I have already updated the branch and the test is passing on the gate now. One piece that must be updated before merging is the documentation part -- it turned out more confusing than I intended.

I think what @slimreaper35 meant was to contribute the test to our integration tests repo instead of using it from your fork, replacing the references and commentaries you currently have in the code.

eskultety:
@a-ovchinnikov I noticed the last change you pushed has some, apparently bugfix in it. It almost went unnoticed by me, so please when you provide further fixes for issues that have been encountered outside the regular review cycle, please mention them in comment summary with the new revision so that it makes it more visible to the reviewers on what to focus on.

eskultety:
> > > So the only missing piece is to finish the integration test?
> > 
> > 
> > Sorry, why is it missing? I have already updated the branch and the test is passing on the gate now. One piece that must be updated before merging is the documentation part -- it turned out more confusing than I intended.
> 
> I think what @slimreaper35 meant was to contribute the test to our integration tests repo instead of using it from your fork, replacing the references and commentaries you currently have in the code.

Nevermind. I noticed you bypassed the PR workflow and pushed the test branch directly.

hermeto/core/package_managers/pip.py
```diff
@@ -2171,9 +2203,40 @@ def _version(dep: dict[str, Any]) -> str:
...
+        # a Rust package (see https://doc.rust-lang.org/cargo/guide/project-layout.html
+        # for details).
+        except KeyError:
+            continue
+        tf.extractall(path=Path(str(tf.name)).parent, filter="data")
+        subpath_to_add = "deps/pip/" + str(pname)
+        path_to_use = output_dir.join_within_root(subpath_to_add)
+        packages_containing_rust_code.append(
+            CargoPackageInput(type="cargo", path=str(path_to_use.path.relative_to(output_dir.path)))
+        )
```
slimreaper35, line 2236:
```suggestion
        rust_package_source_dir = output_dir.join_within_root("deps", "pip", pkg_name)
        packages_containing_rust_code.append(
            CargoPackageInput(
                type="cargo", path=rust_package_source_dir.path.relative_to(output_dir)
            )
        )
```

hermeto/core/package_managers/pip.py
```diff
@@ -2171,9 +2203,40 @@ def _version(dep: dict[str, Any]) -> str:
         "package": {"name": pkg_name, "version": pkg_version, "type": "pip"},
         "dependencies": dependencies,
         "requirements": [*resolved_req_files, *resolved_build_req_files],
+        "packages_containing_rust_code": packages_containing_rust_code,
     }
 
 
+def _filter_packages_with_rust_code(
+    packages: list, app_path: RootedPath, output_dir: RootedPath, source_dir: RootedPath
```
slimreaper35, line 2211: `app_path` param is not used

docs/pip.md
brunoapimentel: The cargo [docs](https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure) also indicate that we could make use of $CARGO_HOME to try to bypass this? I did not test it, though.

eskultety: > but breaking the isolation of the project folder would be new ground to us

How so? We inject to the prefetch dir for bundler as well as rpm.

a-ovchinnikov:
In the new revision:
 * simplified handling package name suffixes;
 * added code to inject CARGO_HOME variable pointing to injected config.toml -- this makes it unnecessary for a user to manually set the override;
 * updated the docs by dropping everything related to config.toml;
 * updated Containerfile to drop creation of the config;
 * renamed the test to align it with test branch name.

tests/integration/test_cargo.py
```diff
@@ -68,6 +68,26 @@ def test_cargo_packages(
...
+                check_output=True,
+                check_deps_checksums=False,
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="",
+            ),
+            # Invocation will fail if there was a failure to build the dependencies.
+            ["python3", "/app/src/test_package_cachi2/main.py"],
+            [],
+            id="e2e_rust_extensions",
```
slimreaper35, line 89: Last nitpick: to be consistent with other ids, this could be:
`id="pip_e2e_rust_extensions"` which leads me to thinking if `test_pip.py` isn't the more appropriate location ... :thinking:

slimreaper35 (approved):

a-ovchinnikov:
In the latest revision I moved the ultimate test from cargo to pip.

docs/pip.md
```diff
@@ -484,6 +485,34 @@ hermeto-output/deps/pip
...
+```
+
+Note, that this step requires internet connection.
+
+Once requirements-build.txt is populated fetch could be done as usual. Hermeto will fetch all
+run time and build time dependencies for both Python and Rust parts.
+
+Note, that a system which is to be used for building these extensions must have
+both `rustc` and `cargo` installed. In some cases it might be necessary to preinstall
+C libraries which the Rust packages need to be built.
```
eskultety, line 510: I'd probably reword this to a less verbose version of
_must have `rustc`, `cargo`, and all necessary C libraries in order to build the Rust extension packages installed._

tests/integration/test_pip.py
```diff
@@ -206,6 +206,26 @@ def test_pip_packages(
...
+                check_deps_checksums=False,
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="",
+            ),
+            # Invocation will fail if there was a failure to build the dependencies.
+            ["python3", "/app/src/test_package_cachi2/main.py"],
+            [],
+            id="pip_e2e_rust_extensions",
+        ),
```
eskultety, line 228: I just noticed the test tooks a whopping 5 minutes to execute. Can be addressed in a follow up, but this is going to get frustrating with time. It'd be worth exploring whether the dependency set can be decreased. If nothing else, we should adopt a new tag "expensive" and run ITs in local dev envs without that tag active.

eskultety (approved):

tests/integration/test_pip.py
a-ovchinnikov: Decreasing dependency set will be rather hard since the only dependency the test package has now is `cryptography`. However the very nature of a hermetic build of such complexity results in a test instance downloading half the Internet on each run: to build a Rust extension for Python we need not just the Python package, but also entire dependency trees of both Python and Rust parts (regular plus build ones) and also rpms to supply toolchains and necessary C libraries. This is an inherent problem for most e2e tests, it just happens to manifest itself here more prominently. There are ways to tackle this and I am going to discuss them and their trade-offs during F2F.

eskultety: I know, it was a long shot if there wasn't just one that wasn't needed :).

PR #890, opened 2025-03-26
docs: Remove releasing section from CONTRIBUTING.md

slimreaper35:
The release process is only internal workflow for maintainers of the project. It does not belong to the CONTRIBUTING.md. The file is here for contributors as the name suggests.

Additionaly, the process may and most likely will change in the future.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@
   * [Running unit tests](#running-unit-tests)
   * [Running integration tests](#running-integration-tests)
   * [Adding new dependencies to the project](#adding-new-dependencies-to-the-project)
-* [Releasing](#releasing)
+* [Release schedule](#release-schedule)
 
 ## How to start a contribution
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -271,23 +271,6 @@ re-generate the set of dependencies, run the following in the repository and com
 nox -s pip-compile
 ```
 
-## Releasing
```
a-ovchinnikov, line 274: What do you think about moving this paragraph someplace else instead of just dropping it?

eskultety: Agreed. We can create a private repo to serve as storage for maintainer-only related stuff. Losing the information completely doesn't sound like a good idea to me.

a-ovchinnikov: Does it need to be private though? Just a MAINTAINING.md should work -- if someone who is not a maintainer is willing to read it then by all means, there is not secret information there, but they are unlikely to gain any insight from that. Having more than one repo adds to maintenance burden, having all relevant docs in one place makes it easier to locate necessary data.

eskultety: Fair enough. I was trying to foresee more information in there that we may not want to share publicly, like the infra details, and maintenance information, etc. after all, we're currently not solely relying on GH's native infra....

PR #892, opened 2025-03-26
Ensure clean destination dir when fast copy fails

slimreaper35:

### DIFF
--- a/hermeto/core/utils.py
+++ b/hermeto/core/utils.py
@@ -185,6 +185,7 @@ def copy_directory(origin: Path, destination: Path) -> Path:
         _copy_using(_fast_copy)
     except _FastCopyFailedFallback:
         log.debug("Fast copying failed, falling back to standard copy.")
+        shutil.rmtree(destination)
         _copy_using(shutil.copy2)
 
     return destination


a-ovchinnikov (approved):

taylormadore (approved):

eskultety (changes requested):
See my comment to the PR, I don't think change is correct.

slimreaper35:
Thanks! Deleting the temporary directory works too.

a-ovchinnikov (approved):

eskultety (approved):

cgwalters:
A whole lot of our builds for centos/rhel-bootc are broken, see https://gitlab.com/redhat/centos-stream/containers/bootc/-/issues/1184
I think it's either fallout from or fixed by this? Haven't dug in here but just glancing at this I am really wondering why you aren't just forking off `/bin/cp -a`...

eskultety:
> A whole lot of our builds for centos/rhel-bootc are broken, see https://gitlab.com/redhat/centos-stream/containers/bootc/-/issues/1184 I think it's either fallout from or fixed by this?

@cgwalters  with this PR those errors will go away.

> Haven't dug in here but just glancing at this I am really wondering why you aren't just forking off `/bin/cp -a`...

I don't follow. Can you elaborate please?

cgwalters:
> @cgwalters with this PR those errors will go away.

Thanks.

> I don't follow. Can you elaborate please?

AFAICS this is basically doing the same thing that forking off `/bin/cp` from coreutils is doing - modern versions use `--reflinks=auto` by default, etc. Why reimplement that with a lot of custom Python code?

eskultety:
> AFAICS this is basically doing the same thing that forking off `/bin/cp` from coreutils is doing - modern versions use `--reflinks=auto` by default, etc. Why reimplement that with a lot of custom Python code?

1. it's not a lot of code really
2. the code was taken verbatim (minus the exception fallback) directly from the official reflink support proposal PR to cpython (properly referenced in the related commit message: 7480c02cf); this is to be flipped to `shutil` in the very distant future where we'll bump the minimum required version of Python to the relevant release with native reflink primitives.
3. unnecessary subprocess overhead, not a reasonable backpropagation of errors to differentiate recoverable from unrecoverable errors, especially in this particular case we'd have zero visibility into what copy was used and hence struggle to identify and investigate any copy-related problems by not knowing what impl `cp` used. In general, subprocess should be reserved only to cases where there isn't a straightforward way of addressing a particular issue in python directly

cgwalters:
When will this appear in an updated Konflux task? Looking at https://gitlab.com/redhat/centos-stream/containers/bootc/-/merge_requests/493 we're still seeing this symptom with a recent prefetch-dependencies run.

eskultety:
@cgwalters I presume you've already found your answer, but mentioning it for completeness: https://github.com/konflux-ci/build-definitions/pull/2159#event-17092240767

PR #893, opened 2025-03-28
Add Python3.13 to our supported platforms list

eskultety:
Python 3.13 has been out for a while, but we couldn't easily integrate with it in our CI due to `createrepo_c` which we depend on is slow with releases and didn't provide a 3.13 wheel (we'd have to build it from scratch in our CI). That has recently changed and there is a 3.13 build available -> enable support for 3.13 and test with it in unit tests.

### DIFF
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
 
     container:
       image: python:${{ matrix.python-version }}-slim
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -25,9 +25,7 @@ jobs:
         # We need to install git inside the container otherwise the checkout action will use Git
         # REST API and the .git directory won't be present which fails due to setuptools-scm
         apt-get update && apt-get install --no-install-recommends --no-install-suggests -y git
-        python3 -m pip install --upgrade pip
-        pip install nox
-        pip install tomlkit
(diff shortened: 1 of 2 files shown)


eskultety:
> What exactly does support mean here? I've always wondered why we need unit tests for **four** different versions of Python. We are not deploying to PyPI, and the container image only uses 3.9 (same for ITs, same for requirements).

It means the code base doesn't break with future versions of Python we'll gradually transition to due to removal of deprecated functionality announced in earlier versions. It also means that you can use any of these Python version in your own local development environments, e.g. a recent Fedora 41 upgrade broke everyone who tried to re-create their venv with plain `python3` invocation because unlike Fedora 40, 41 now defaults to 3.13 instead of 3.12 which was the last version that `createrepo_c` could have been pulled on without having to install additional tooling to build it from sources.

> the container image only uses 3.9 

This has almost nothing to do with the upstream code base, because that is a production deliverable, unless you publish to PyPI a project will, platform-wise, only ever ship a single container anyway typically utilizing the least supported version to stay true to the claim. Deliverables are merely snapshots of the codebase at a given time. Conversely, the code base is a "living organism ", constantly moving forward, striving to remain up to date with latest trends and prepare for upcoming changes from external factors.

> same for ITs,

That's fine, it is a correct thing to do to run ITs over what ends up being the actual consumable by our user base. Note that if we ever decided to start releasing to PyPI, we'd have to change our ITs such that we'd always test over the PyPI build (yet in a container) to ensure consistent experience.

> same for requirements

I don't get this part.

slimreaper35 (approved):

eskultety:
ping. After https://github.com/hermetoproject/hermeto/pull/899 this really could be simplified to the last commit only. Sine  `createrepo_c` is no longer a blocker for this this statement is pretty much just a formality. I'll be dropping the second commit since that one is no longer needed.

a-ovchinnikov (approved):

PR #896, opened 2025-03-28
gomod: Print the Git error message a failure to fetch tags happens

brunoapimentel:
This is very useful when troubleshooting these types of errors. Currently, the only printed output is this:

 ```
Error: FetchError: Failed to fetch the tags on the Git repository [...]
```

Which hides the underlying cause.

### DIFF
--- a/hermeto/core/package_managers/gomod.py
+++ b/hermeto/core/package_managers/gomod.py
@@ -1257,7 +1257,8 @@ class ModuleVersionResolver:
         except Exception as ex:
             raise FetchError(
                 f"Failed to fetch the tags on the Git repository ({type(ex).__name__}) "
-                f"for {repo.working_tree_dir}"
+                f"for {repo.working_tree_dir}: "
+                f"{str(ex)}"
             )
 
         return cls(repo, commit)

### CODE REVIEW

**Review by a-ovchinnikov — APPROVED** [review]

**Review by taylormadore — APPROVED** [review]
We've needed this for a while :sweat_smile:

File: hermeto/core/package_managers/gomod.py [hunk 1]
```diff
@@ -1255,6 +1255,7 @@ def from_repo_path(cls, repo_path: RootedPath) -> "Self":
         try:
             repo.remote().fetch(force=True, tags=True)
         except Exception as ex:
+            log.error(ex)
```
eskultety, line 1258: nitpick: This will look weird in the user log, almost as if we had an uncaught error somewhere. I think this should be appended to the `FetchError` string, after all `Failed to fetch the tags on the Git repository` truly begs for a more detailed error message :).

eskultety (approved):
One small trivial but IMO functional nitpick, LGTM otherwise.

hermeto/core/package_managers/gomod.py
brunoapimentel: That is a good point. I don't think we need to print the stack trace at all. Updated.

PR #904, opened 2025-04-02
WIP: DNF manifest design document

slimreaper35:
Document the DNF-native RPM manifest approach to replace the existing prototype solution. This provides a supported method using libpkgmanifest for generating and parsing lockfiles.

### DIFF
--- a/docs/design/dnf.md
+++ b/docs/design/dnf.md
@@ -1,0 +1,148 @@
+# DNF/RPM Manifest/Lockfile swap
+
+DNF4 - <https://github.com/rpm-software-management/dnf>
+
+DNF5 - <https://github.com/rpm-software-management/dnf5>
+
+RPM - <https://github.com/rpm-software-management/rpm>
+
+## Context
+
+The current solution for RPM manifests is based on the so-called RPM lockfile
+[prototype](https://github.com/konflux-ci/rpm-lockfile-prototype)
+which is a proof-of-concept tool that generates a lockfile containing all RPMs required to build a
+container image. Hermeto parses this file manually and downloads all RPMs locally to enable hermetic
+builds.
+
+The `rpm` package manager is not released as a supported feature, thus, it can only be used with
+`--dev-package-manager` flag. User documentation does not exist either. Only available documentation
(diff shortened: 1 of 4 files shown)


docs/design/dnf.md
```diff
@@ -0,0 +1,387 @@
...
+
+## Implementation
+
+### Prefetching (approach 1)
+
+Prefetching the packages can be done by simply using the `dnf4 manifest` command:
+
+```bash
+dnf4 manifest download --destdir "${output_dir}/deps/rpm"
+```
```
eskultety, line 372: My expectation here was that we'd finally be able to consume the functionality via Python APIs, i.e. not having to run another subprocess command which doesn't make it any easier for error handling.

I looked at the manifest library but those are pretty much just essential APIs to extract information from the manifest file, but not to control any action over it. Then I immediately thought of how DNF's `manifest` command would consume it and whether we could plug into that and so I thought of `libdnf`. However, now I realized that `manifest` is and will be enabled via a plugin, not `libdnf` and so we need to get familiar with the plugin's implementation. I wasn't able to find (admittedly didn't spend too much time digging) the repo for the plugin itself, but that should be part of this investigation and work with the community on the integration side of things for this.
That said, subprocess should be the last resort solution if no other option exists for us.

eskultety (commented):
Since this is still in the draft state, just a couple of top-level observations.

docs/design/dnf.md
slimreaper35: @mcurlej May I ask you for a quick insight here on the usage of above mentioned Python API?

(This PR is supposed to be a design document on integrating support of DNF [manifest](https://github.com/rpm-software-management/libpkgmanifest) to the hermeto project.)

eskultety: More specifically, I'd scope the question to just: Is there a way for us to consume the `manifest` plugin's functionality through a Python interface that would be officially supported (i.e. apart from direct `dnf manifest` invocation in a dedicated process)? The question stems from the fact that IIUC `libpkgmanifest` only provides primitives to extract data from the manifest file.

PR #906, opened 2025-04-03
Link README.md to the index page on the website

slimreaper35:

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,10 @@ The primary intended use of Hermeto's outputs is for network-isolated container 
 Please note that Hermeto is rather picky, aiming to:
 
 * encourage or enforce best practices
-* never execute arbitrary code - looking at you [setup.py (discouraged)][setuppy-discouraged]
+* never execute arbitrary code [^pip-download-example]
 * keep the implementation simple
+
+[^pip-download-example]: See for example <https://discuss.python.org/t/pip-download-just-the-source-packages-no-building-no-metadata-etc/4651>
 
 To play nicely with Hermeto, the build process for your project must be
 
@@ -89,7 +91,7 @@ hermeto fetch-deps \
 
 The `fetch-deps` command fetches your project's dependencies and stores them on your disk. Hermeto also produces a detailed SBOM containing information about all the project's components and packages. You can find the SBOM in the output directory.
 
-See [docs/usage.md](docs/usage.md) for a more detailed, practical (*cough*) example of Hermeto usage.
+See [docs/usage.md](docs/usage.md) for a more detailed, practical example of Hermeto usage.
 
 You might also like to check out `hermeto --help` and the `--help` texts of the available subcommands.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -56,7 +55,7 @@ technologies - such as containers - to achieve isolation (see [usage](docs/usage
 
 We do not distribute Hermeto as a standalone package as of now.
 
-To install Hermeto for local development, see the [CONTRIBUTING.md](CONTRIBUTING.md).
+To install Hermeto for local development, see the [CONTRIBUTING.md](https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md).
```
eskultety, line 58: I think with markdown it's fine as it is. See, unlike with GH pull request/issues IDs vs full URLs outside of the webUI context which are immutable (i.e. only in read-only mode via commit messages), these references in markdown are actually mutable, i.e. if we were to either rename the project again tomorrow or switch GH for GitLab, the link would have to be replaced again whereas the built-in linking/referencing mechanism in markdown prevents this. Even when reading the README file offline it mentions the CONTRIBUTING.md file directly which resides in the repo root directory, so I'm afraid we're not gaining anything with this hunk.

README.md
```diff
@@ -102,7 +101,7 @@ You can change Hermeto's configuration by specifying a configuration file while
 ```
 
 Any parameter specified in this file will override the default values present in the
-[config.py](hermeto/core/config.py) module.
+[config.py](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/config.py) module.
```
eskultety, line 104: same here, when you read the README file offline, the path to the file still resolves perfectly whether you read it online, offline, or from a source tarball.

mkdocs.yml
```diff
@@ -27,6 +27,9 @@ theme:
     - content.action.view
     - content.code.copy
     - content.tooltips
+    - navigation.sections
```
eskultety, line 30: Commit message:

The change sounds cool, any way I can have a preview of it first, e.g. on your fork?

README.md
slimreaper35: Same as above, only related to the website.

mkdocs.yml
slimreaper35: Uh, I removed it from my fork. If you checkout my branch/PR just run `mkdocs serve`.

README.md
eskultety: Oh, I see it now. Sadly, 
https://github.com/mkdocs/mkdocs/discussions/3230
https://github.com/squidfunk/mkdocs-material/issues/6879
https://github.com/mkdocs/mkdocs/issues/608#issuecomment-109855759

none of the above solves the issue, sigh.
What you have is an acceptable trade-off, so let's keep it as is.

README.md
```diff
@@ -25,7 +25,7 @@ The primary intended use of Hermeto's outputs is for network-isolated container
 Please note that Hermeto is rather picky, aiming to:
 
 * encourage or enforce best practices
-* never execute arbitrary code - looking at you [setup.py (discouraged)][setuppy-discouraged]
+* never execute arbitrary code
```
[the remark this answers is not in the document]
eskultety: There you go - https://discuss.python.org/t/pip-download-just-the-source-packages-no-building-no-metadata-etc/4651 sorry for the delay. I think it's also worth noting that this is specifically due to `pip download` trying to extract package metadata reliably.

On a side note, the thread also mentions some future work on standardizing sdist naming so that pip would not need to execute that particular code path anymore. However, I have not tried it myself but more importantly, we'd have to verify if pypi now enforces the PEP standard mentioned in the thread even for old packages (my guess is NO) to eliminate the need to build the package first in order to get the metadata. I'm only mentioning that so that we may want to (or be tempted to) flip our implementation to rely on https://peps.python.org/pep-0643/. It would be nice to simplify the code base, but we simply may not be able to do so to cope with old package builds not conforming to that PEP.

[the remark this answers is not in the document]
slimreaper35: Is a footnote enough here?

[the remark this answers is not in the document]
eskultety: Yep, I'd argue that's even the recommended approach.

a-ovchinnikov (approved):

eskultety (approved):

PR #911, opened 2025-04-24
Add integration test covering multiple package managers

slimreaper35:
Closes https://github.com/hermetoproject/hermeto/issues/857

**Note:** If the contribution is external (not from an organization member), the CI


tests/integration/test_multiple.py
```diff
@@ -0,0 +1,56 @@
...
+@pytest.mark.parametrize(
+    "test_params, check_cmd, expected_cmd_output",
+    [
+        pytest.param(
+            utils.TestParameters(
+                branch="multiple/gomod-and-npm",
+                packages=(
+                    ({"type": "gomod", "path": "gomod-package"}),
+                    {"type": "npm", "path": "npm-package"},
+                    # using RPM to prefetch both package managers
```
a-ovchinnikov, line 18: This feels like naming clash -- Hermeto's package managers are a different entity. WDYT about `# using RPM to provide gomod and npm`?

tests/integration/utils.py
```diff
@@ -38,10 +38,11 @@
     )
 )
 
-# package managers that are not exposed to the user but are used internally
-EXTRA_PMS = ["yarn_classic"]
+# yarn classic is not exposed to users, but it is used internally
+# multiple is used for tests covering multiple package managers
```
a-ovchinnikov, line 42: Is it exposed in any way? Why do you need to add it to a list of supported PMs?

a-ovchinnikov (commented):
LGTM with two small questions

tests/integration/utils.py
slimreaper35: I changed the comment completely to make it clearer.

slimreaper35: The only integration test that always runs is `test_legacy_entrypoint`, but IIUC, we keep it only for backwards compatibility, and it will be removed in the future.

a-ovchinnikov: Thank you for updating the comment, it is still a bit confusing: what you want to share is that there is a _test scenario_ for multiple PMs which must be triggered each time a module containing this test scenario is changed. It is not a PM (unlike `yarn_classic` which is a hidden PM). Here lies a problem: you would want to trigger the same test on every change to any of PMs that are run in the scope of the test. What is actually needed is a translation rule to map `(rpm, gomod, npm)` to `test_multiple` in addition to mapping it to `test_pm`.

Come to think of it I am not so sure anymore that this short-circuiting logic is as helpful as I originally thought.

tests/integration/utils.py
```diff
@@ -626,7 +626,9 @@ def pm_name(pm_change: Path) -> str:
 
 
 def affected_package_managers(pm_changes: tuple[Path, ...]) -> set[str]:
```
a-ovchinnikov, line 628: You would still need "multiple" in `EXTRA_PMS` for it to be properly excluded though. I think adding it back with a comment is the correct thing to do ATM. Everything else seems to work fine.

tests/integration/utils.py
```diff
@@ -626,7 +626,9 @@ def pm_name(pm_change: Path) -> str:
...
-    return set(pm_name(c) for c in pm_changes)
+    result = set(pm_name(c) for c in pm_changes)
+    result.add("multiple") if result.intersection(("gomod", "npm", "rpm")) else None
```
a-ovchinnikov, line 630: I played a bit more with this code, that's what I had to add to ensure `multiple` gets skipped when `test_gomod` is modified:
```python
   result = set(pm_name(c) for c in pm_changes)
   def filter_pure_pm_names(changes):
       return [pm_name(c) for c in changes if name_of(c) in SUPPORTED_PMS and not name_of(c).startswith("test_")]
   def affects_only_pms_from_multiple():
       return set(filter_pure_pm_names(pm_changes)).intersection(("gomod", "npm", "rpm"))
   result.add("multiple") if affects_only_pms_from_multiple()  else None
```

It will trigger `mulitple` on changes to tests otherwise. The resulting code is pretty ugly and I am fine both ways with and without it, so this is not a blocker and I will +1 it.

a-ovchinnikov (approved):

a-ovchinnikov:
> ...mentioned ... he would consider dropping the logic completely
In this PR as well (see [my comment](https://github.com/hermetoproject/hermeto/pull/911#discussion_r2066676794) above). Rewriting the rules to be less arcane than they appear now would also work. I am not entirely sure this is in the scope of the change though, maybe a follow-up issue would be a better place for that?

eskultety:
> > ...mentioned ... he would consider dropping the logic completely
> 
> In this PR as well (see [my comment](https://github.com/hermetoproject/hermeto/pull/911#discussion_r2066676794) above). Rewriting the rules to be less arcane than they appear now would also work. I am not entirely sure this is in the scope of the change though, maybe a follow-up issue would be a better place for that? _Edit_: formatting

I mean, we'd be deliberately merging a misbehaving code, yeah, in tests, but that shouldn't be a justification factor really if the logic is off. I don't mind tracking and tackling the problem in a different PR, but then this PR would have to wait and rebase to merge clean changes, i.e. a test. It's not like this work is urgent or anything, effort has been invested, work has been done, we can just wait for some related polishing to happen first.

tests/integration/utils.py
```diff
@@ -602,7 +602,7 @@ def affects_pm_directly(change: Path) -> bool:
         return "package_managers" in change.parts and name_belongs_to_a_pm(change)
 
     def affects_pm_tests(change: Path) -> bool:
-        return tested_object_name(change) in SUPPORTED_PMS
+        return tested_object_name(change) in SUPPORTED_PMS or change.stem.endswith("multiple")
```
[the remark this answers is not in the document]
ben-alkov: > we're trying to produce a set of tests to skip rather than a set of tests to run and then inverting the logic in pytest_collection_modifyitems

Heartily agree that this is not The Way

tests/integration/test_data/multiple_gomod_and_npm/container/Containerfile
```diff
@@ -0,0 +1,13 @@
+FROM registry.access.redhat.com/ubi9/ubi@sha256:f4ebd46d3ba96feb016d798009e1cc2404c3a4ebdac8b2479a2ac053e59f41b4
```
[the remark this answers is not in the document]
slimreaper35: It's not for dependabot. I cannot build the image when using the latest tag.

```
E           STEP 3/7: RUN dnf -y install golang npm
E           Updating Subscription Management repositories.
E           Unable to read consumer identity
E           
E           This system is not registered with an entitlement server. You can use subscription-manager to register.
E           
E           Repository 'ubi-9-baseos-rpms' is missing name in configuration, using id.
E           Repository 'ubi-9-appstream-rpms' is missing name in configuration, using id.
E           ubi-9-baseos-rpms                               8.6 MB/s | 8.8 kB     00:00    
E           ubi-9-appstream-rpms                             90 MB/s |  92 kB     00:00    
E           Error: 
E            Problem: package golang-bin-1.23.6-2.el9_5.x86_64 from ubi-9-appstream-rpms requires /usr/bin/gcc, but none of the providers can be installed
E             - package golang-1.23.6-2.el9_5.x86_64 from ubi-9-appstream-rpms requires golang-bin = 1.23.6-2.el9_5, but none of the providers can be installed
E             - package gcc-11.5.0-5.el9_5.x86_64 from ubi-9-appstream-rpms requires glibc-devel >= 2.2.90-12, but none of the providers can be installed
E             - conflicting requests
E             - nothing provides glibc = 2.34-125.el9_5.8 needed by glibc-devel-2.34-125.el9_5.8.x86_64 from ubi-9-appstream-rpms
E           (try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
E           Error: building at STEP "RUN dnf -y install golang npm": while running runtime: exit status 1
```

tests/integration/utils.py
slimreaper35: Applied.

tests/integration/test_data/multiple_gomod_and_npm/container/Containerfile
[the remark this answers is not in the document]
eskultety: Well, I think we have a problem Houston then.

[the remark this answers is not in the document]
eskultety: I don't think using shortcuts like the one introduced to **silently** overcome the issue you mention, that is not the way! If an issue like that exists, it has to be either documented or shared upfront. At the same time, such a bug needs to be reported rather than silently workarounding it given what UBI is supposed to represent in the container ecosystem. I think if anything, this is only intermittent caused by some automation glitch on the CI/CD side (more specifically, on the repo metadata vsd content side).

tests/integration/utils.py
```diff
@@ -505,7 +505,6 @@ def build_image_and_check_cmd(
         containerfile_path=container_folder.joinpath("Containerfile"),
         test_case=test_case,
     ) as test_image:
-
```
eskultety, line 508: Unrelated change.

tests/integration/test_data/multiple_gomod_and_npm/container/Containerfile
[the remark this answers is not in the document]
slimreaper35: I don't understand why the test is passing on GitHub but not locally :exploding_head:

tests/integration/utils.py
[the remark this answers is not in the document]
slimreaper35: Dropped this part in favor of https://github.com/hermetoproject/hermeto/issues/950

eskultety (approved):

a-ovchinnikov (approved):

PR #913, opened 2025-04-25
cargo-pip: Adds support for nested Cargo.toml

### LINKED ISSUE #912: pip+cargo integration: python+rust packages with Cargo.toml located outside won't have their cargo packages fetched
As the title said. Unfortunately, the heuristic adopted [here](https://github.com/hermetoproject/hermeto/blob/4199123830bc212ec79a0eb3343a1f5f1cf67b77/hermeto/core/package_managers/pip.py#L2247-L2252) will lead to false negatives.

One relatively popular example is bcrypt, which is maintained by the same team that maintains cryptography. Its manifest file is located at [`src/_bcrypt/Cargo.toml`](https://github.com/pyca/bcrypt/blob/4.3.0/pyproject.toml#L54).

However, from a research I made a few months ago with the top downloaded pypi packages, inferring which ones that depend on rust by searching for Cargo.toml anywhere I know it would solve this problem but create another one: we would start detecting false positives.

I propose that you look to build dependencies of each package. If the package depends on either maturin or setuptools-rust, then find the topmost `Cargo.toml`. This heuristic worked just fine when I did that research, without false positives or negatives. I also made this suggestion on the design document [here](https://github.com/hermetoproject/hermeto/blob/e5fa5c0fcd0dff62cf02be5b0d219e04c1ea440c/docs/design/cargo-support.md#L808-L812)

bruno-fs:
Adds support for prefetching cargo dependencies for packages with Cargo.toml nested while avoiding false positives. This is achieved by only considering only as true python+rust packages those that contain maturin or setuptools-rust as a build dependency.

Closes #912


requirements-extras.txt
```diff
@@ -810,6 +818,14 @@ pbr==6.1.1 \
...
     --hash=sha256:93ea72ce6989eb2eed99d0f75721474f69ad88128afdef5ac377eb797c4bf76b
     # via stevedore
+pip==25.0.1 \
+    --hash=sha256:88f96547ea48b940a3a385494e181e29fb8637898f88d88737c5049780f196ea \
+    --hash=sha256:c46efd13b6aa8279f33f2864459c8ce587ea6a1a59ee20de055868d8f7688f7f
+    # via pip-tools
+pip-tools==7.4.1 \
+    --hash=sha256:4c690e5fbae2f21e87843e89c26191f0d9454f362d8acdbd695716493ec8b3a9 \
+    --hash=sha256:864826f5073864450e24dbeeb85ce3920cdfb09848a3d69ebf537b521f14bcc9
+    # via pybuild-deps
```
bruno-fs, line 828: for this particular use case of pybuild-deps, these dependencies are not really required.

maybe they should become optional if you find interesting using pybuild-deps as a lib 🤔

hermeto/core/package_managers/pip.py
```diff
@@ -2231,6 +2232,41 @@ def _version(dep: dict[str, Any]) -> str:
...
+        # the file is decoded as utf-8-sig because plain utf-8 has proven to
+        # be problematic with certain sources from pypi
+        # see https://github.com/hermetoproject/pybuild-deps/blob/4dc40ffabddb8aad1279978b8741111fb64452e6/src/pybuild_deps/finder.py#L45-L51
+        file_contents = file.read().decode("utf-8-sig") # type: ignore
+        try:
+            build_dependencies = parser(file_contents)
+        except parsers.SetupPyParsingError:
+            # unfortunately pybuild-deps parser has some known edge-cases for older packages
+            # relying on setup.py
+            log.error("Unable to parse build dependencies for %s.", pkg_name)
```
a-ovchinnikov, line 2264: You would probably want to continue to the next pair here or hide `if _has_rust_build_deps(build_dependencies):` in an `else:` clause: it looks like `build_dependencies` will be undefined when `log.error` is reached.

bruno-fs: nice catch! thanks

a-ovchinnikov (approved):

hermeto/core/package_managers/pip.py
```diff
@@ -2231,6 +2232,42 @@ def _version(dep: dict[str, Any]) -> str:
...
+    for file_name, parser in file_parser_map.items():
+        pkg_name = source_tarball.getnames()[0].split("/")[0]
+        try:
+            file = source_tarball.extractfile(f"{pkg_name}/{file_name}")
+        except KeyError:
+            continue
+        # the file is decoded as utf-8-sig because plain utf-8 has proven to
+        # be problematic with certain sources from pypi
+        # see https://github.com/hermetoproject/pybuild-deps/blob/4dc40ffabddb8aad1279978b8741111fb64452e6/src/pybuild_deps/finder.py#L45-L51
+        file_contents = file.read().decode("utf-8-sig") # type: ignore
```
a-ovchinnikov, line 2258: This line offends flake8 and black, could you please add one more space before the comment? Also what exactly does mypy complain about?

a-ovchinnikov (commented):
LGTM, but the commit has to be signed.

hermeto/core/package_managers/pip.py
```diff
@@ -2231,6 +2232,42 @@ def _version(dep: dict[str, Any]) -> str:
...
+def _depends_on_rust(source_tarball: tarfile.TarFile) -> bool:
+    file_parser_map = {
+        "pyproject.toml": parsers.parse_pyproject_toml,
+        "setup.cfg": parsers.parse_setup_cfg,
+        "setup.py": parsers.parse_setup_py,
+    }
+    for file_name, parser in file_parser_map.items():
+        pkg_name = source_tarball.getnames()[0].split("/")[0]
+        try:
+            file: tarfile.ExFileObject = source_tarball.extractfile(f"{pkg_name}/{file_name}")  # type: ignore[assignment]
```
bruno-fs, line 2252: in order to not have mypy complaining below, I added this type annotation (I got the type using a debugger). 

not sure if this is the best approach (I've zero experience with mypy)

hermeto/core/package_managers/pip.py
bruno-fs: managed to get rid ot this mypy error, but I'm not sure if the approach was the best: https://github.com/hermetoproject/hermeto/pull/913/files#r2064395166

slimreaper35:
@bruno-fs You will probably need to regenerate SBOMs using `nox -s generate-test-data -- -k test_pip`

bruno-fs:
Huh, when I ran just `nox` before everything was passing, but now that I attempted to regenerate SBOMs I got a failure

Looks like the relevant issue in the traceback is this one.

```
E         2025-04-29 11:25:05,524 INFO Fetching cargo dependencies at /tmp/pytest-of-bciconel/pytest-0/test_pip_packages_pip_full_has0/hermeto-output/deps/pip/cryptography-43.0.0/src/rust
E         2025-04-29 11:25:11,504 ERROR ValueError: Invalid purl: version argument must be a string: {'workspace': True}.
```

Full traceback [here](https://paste.centos.org/view/ca7b977d). I'm scratching my head trying to understand how I broke this.

bruno-fs:
I see where the problem begins. [cryptography version 43](https://github.com/pyca/cryptography/tree/43.0.0) do not have Cargo.toml in its root, so this wasn't being detected before.

Now that it is, we are having trouble parsing the version out of it: https://github.com/pyca/cryptography/blob/43.0.0/src/rust/Cargo.toml#L11

hermeto/core/package_managers/cargo/main.py
```diff
@@ -70,7 +70,15 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
-    return tomlkit.parse(path_to_toml.read_text()).value["package"]
+    parsed_toml = tomlkit.parse(path_to_toml.read_text())
+    package_info = parsed_toml.value["package"]
+    if isinstance(package_info, dict) and package_info.get("version") == {"workspace": True}:
+        # support for workspace package table https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
+        # real world example: https://github.com/pyca/cryptography/blob/43.0.0/src/rust/Cargo.toml
+        # we condition this to when package_info is a dict because this same function has usages where package_info is a list
+        # (Cargo.lock) and this info is only available when it is a dict (Cargo.toml)
+        package_info["version"] = parsed_toml.value["workspace"]["package"]["version"]
+    return package_info
```
slimreaper35, line 81: Good catch, we didn't know that version could be specified inside the workspace table.
One question: Is it possible to have the name specified in the workspace table, too?

For the future:
I just feel that there is too much coupling going on here. It took some time to get my head around that. We basically use one function to process two formats. IMO, 6 comments and 5 lines of code are a hint that we could do better.

hermeto/core/package_managers/pip.py
```diff
@@ -2231,6 +2232,43 @@ def _version(dep: dict[str, Any]) -> str:
     }
 
 
+def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
+    rust_build_deps = ("maturin", "setuptools-rust", "setuptools_rust")
+    for dep in raw_build_dependencies:
+        if dep.strip().lower().startswith(rust_build_deps):
```
slimreaper35, line 2238: Was this intended to use the `startswith` method with a tuple?

hermeto/core/package_managers/pip.py
```diff
@@ -2231,6 +2232,43 @@ def _version(dep: dict[str, Any]) -> str:
...
+        # be problematic with certain sources from pypi
+        # see https://github.com/hermetoproject/pybuild-deps/blob/4dc40ffabddb8aad1279978b8741111fb64452e6/src/pybuild_deps/finder.py#L45-L51
+        # mypy: it thinks file type is "IO[bytes] | None", but that's not right as .extractfile won't return None.
+        file_contents = file.read().decode("utf-8-sig")  # type: ignore
+        try:
+            build_dependencies = parser(file_contents)
+        except parsers.SetupPyParsingError:
+            # unfortunately pybuild-deps parser has some known edge-cases for older packages
+            # relying on setup.py
+            log.error("Unable to parse build dependencies for %s.", pkg_name)
```
slimreaper35, line 2265: What exactly is the problem here? Is it related to pybuild-deps ?

hermeto/core/package_managers/cargo/main.py
bruno-fs: AFAIK (from what I saw both on documentation and cryptography), name stays on "package" section. But I don't have a 100% confidence on this one, it is mostly a empirical knowledge 😅

hermeto/core/package_managers/pip.py
bruno-fs: yes!

raw_build_dependencies is a list that could have things like `maturin<2.0`. We only care about the package name, and if it specifically match `rust_build_deps` tuple.

hermeto/core/package_managers/pip.py
bruno-fs: It is! pybuild-deps has a limitation when parsing `setup.py` and it will throw this exception when dealing with such case. The setup.py parser test has some use-cases where this might happen:

https://github.com/hermetoproject/pybuild-deps/blob/main/tests/test_parsers/test_setup_py.py#L99-L132

hermeto/core/package_managers/pip.py
slimreaper35: Oh, I just saw that `startswith` accepts a string or a tuple of strings to try.

slimreaper35 (approved):

PR #924, opened 2025-05-02
chore: Make pip skip Rust sources when binaries allowed

a-ovchinnikov:
Previously pip would download and process Rust sources even when a user requested just binaries. This change introduces a check to skip Rust sources processing when a user is fine with having just binaries.

### DIFF
--- a/hermeto/core/package_managers/pip.py
+++ b/hermeto/core/package_managers/pip.py
@@ -229,6 +229,7 @@ def fetch_pip_source(request: Request) -> RequestOutput:
         environment_variables=environment_variables,
         project_files=project_files,
     )
+
     cargo_packages = _find_and_fetch_rust_dependencies(request, packages_containing_rust_code)
     return pip_packages + cargo_packages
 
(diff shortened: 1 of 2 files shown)


brunoapimentel (approved):

brunoapimentel:
Since we have plans to revisit the pip backend to enable the fetching of specific wheels (as opposed to a toggle to fetch all wheels or no wheels at all), we will need to enhance this later on. But this is a problem for our future selves, this LGTM.

a-ovchinnikov:
> What if there are no binaries (wheels) for a pip package?

It should not be fetched with `allow_binaries: True` then.

slimreaper35:
One source distribution is always fetched - https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/pip.py#L1965

a-ovchinnikov:
> One source distribution is always fetched

I am sorry, I don't follow how this is related to the problem of excessive processing of Py+Rust packages, could you please elaborate?

slimreaper35:
Regardless of `allow_binary` option, we always download one source distribution of each pip package (if it exists).

My concern is that if there are no wheels available on PyPI, which can happen, we would skip source distributions that could potentially contain Rust code. IOW, `allow_binary` does not guarantee downloading wheels.

slimreaper35:
Anyway, it should be solved when we implement https://github.com/hermetoproject/hermeto/pull/924#issuecomment-2847866364

slimreaper35 (approved):

PR #928, opened 2025-05-05
chore: Improve messaging for corrupt Cargo.locks

a-ovchinnikov:
When a Cargo.lock and Cargo.toml do not match `cargo vendor --locked` will refuse to proceed. This is a corner case which manifests itself in several Python packages extended with Rust. While making sure that locks match to configs is not a direct responsibility of Hermeto the messaging around this specific failure was confusing.

This change adds a dedicated exception class to raise when there is a mismatch which contains a more thorough explanation of the problem.

Adds visibility to https://github.com/hermetoproject/hermeto/issues/925

### DIFF
--- a/docs/pip.md
+++ b/docs/pip.md
@@ -511,6 +511,15 @@ Note, that a system which is to be used for building these extensions must have
 With these preparations running a pip installation as usual should be sufficient to
 build and install a Rust-based extension.
 
+Note, that sometimes Rust-based extensions can break a build. This could happen when
+such dependency is distributed with a Cargo.lock not matching Cargo.toml (while rare this seems
+to happen due to peculiarities of the release process for some packages). In this case
+a package will be rejected with a note about lock file mismatch and an additional report from
+Cargo about inability to load package lock file due to a mismatch. There is no
+good solution for this problem on Hermeto's side and the best course of action is to reach out
+to maintainers of this extension and notify them about the mismatch. Switching to a binary
+distribution of the package would also resolve this problem at the price of not building
+it from sources (see [Building with wheels](#building-with-wheels) for additional context).
 
 ## Troubleshooting
 
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -1,4 +1,5 @@
 import logging
+import subprocess
 from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import cached_property
@@ -18,6 +19,23 @@ from hermeto.core.scm import get_repo_id
 from hermeto.core.utils import run_cmd
 
 log = logging.getLogger(__name__)
+
+
+class PackageWithCorruptLockfileRejected(PackageRejected):
+    """Package lock file does not match package config."""
+
+    def __init__(self, package_path: str) -> None:
+        """Initialize the error."""
+        reason = (
+            f"{package_path} contains a Cargo.lock that does not match"
+            " the corresponding Cargo.toml"
+        )
+        super().__init__(reason, solution=self.default_solution, docs="")
+
+    default_solution = (
+        "Consider reaching out to maintainer of the dependency in question to address"
+        " inconsistencies between Cargo.lock and Cargo.toml"
+    )
 
 
 @dataclass(frozen=True)
@@ -141,7 +159,20 @@ def _resolve_cargo_package(
     log.info("Fetching cargo dependencies at %s", package_dir)
     with _hidden_cargo_config_file(package_dir):
         # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
-        config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
+        try:
+            config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
+        except subprocess.CalledProcessError as e:
+            # Search for a very specific failure state to better report it.
+            # This is not a robust solution in any way, however it seems to be the only one
+            # readily available: cargo returns a generic 101 code on this failure and on multiple
+            # others, thus the only way to check for this specific type of failure is to process
+            # stderr. Two parts of a string are used to decrease the likelihood of false positives.
+            lock_corruption_marker1 = "failed to sync"
+            lock_corruption_marker2 = "needs to be updated but --locked was passed"
+            if lock_corruption_marker1 in e.stderr and lock_corruption_marker2 in e.stderr:
+                raise PackageWithCorruptLockfileRejected(f"{package_dir.path}")
+            else:
+                raise
 
     packages = _extract_package_info(package_dir.path / "Cargo.lock")
     main_package = _resolve_main_package(package_dir)
--- a/tests/unit/package_managers/test_pip.py
+++ b/tests/unit/package_managers/test_pip.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import re
+import subprocess
 from copy import deepcopy
 from pathlib import Path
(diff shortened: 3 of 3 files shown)


hermeto/core/errors.py
```diff
@@ -104,6 +104,20 @@ class NotAGitRepo(PackageRejected):
     """A package turned out to be not a git repository."""
 
 
+class PackageWithCorruptLockfileRejected(PackageRejected):
+    """Package lock file does not match package config."""
+
+    def __init__(self, package_path: str) -> None:
+        """Initialize the error."""
+        reason = f"{package_path} contains a lock file that does not match corresponding Cargo.toml"
```
eskultety, line 112: nitpick: this is a backend specific error the whole purpose of which is to wrap the parent with a better msg string. Even the string itself, by the nature of containing a reference to Cargo.toml, suggests that it would likely be better placed in the cargo backend package.

hermeto/core/errors.py
```diff
@@ -104,6 +104,20 @@ class NotAGitRepo(PackageRejected):
...
+        super().__init__(reason, solution=self.default_solution, docs="")
+
+    default_solution = (
+        "Consider reaching out to package maintainer to address inconsistencies between Cargo.lock"
+        " and Cargo.toml"
+    )
```
eskultety, line 118: substitute nitpicks:
`s/corresponding/the corresponding`
`s/lock file/Cargo.lock`
`s/to package maintainer/to maintainers of the dependency in question`

hermeto/core/errors.py
slimreaper35, line 112:
```suggestion
        reason = f"{package_path} contains Cargo.lock that does not match the corresponding Cargo.toml"
```

hermeto/core/package_managers/cargo/main.py
```diff
@@ -141,7 +142,20 @@ def _resolve_cargo_package(
...
+            config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
+        except subprocess.CalledProcessError as e:
+            # Search for a very specific failure state to better report it.
+            lock_corruption_marker1 = "failed to sync"
+            lock_corruption_marker2 = "needs to be updated but --locked was passed"
+            if (
+                e.returncode == 101
+                and lock_corruption_marker1 in e.stderr
+                and lock_corruption_marker2 in e.stderr
+            ):
```
eskultety, line 155: Hmm, not a fan of this handling to be honest. I think we share the general attitude towards matching strings in error messages to be fairly fragile in the log term because CLI output isn't guaranteed to not change and we'd never notice this code path wasn't exercised anymore. So, my question would be - how many different error paths tied to `cargo vendor` do we realistically expect here? Can we make a reasonable assumption that if the vendor command fails it **has to be** related to the Cargo.toml - Cargo.lock ? Would it make sense for us to make the error more generic in case we cannot reliably make ^this assumption? Something along the lines of: _"Please check that your Cargo.lock conforms to the Cargo.toml file and see the underlying command error"_.

Also, I looked at https://github.com/rust-lang/cargo/blob/bfe41282be384e55eee6a87c2151fab00075b374/src/cargo/util/errors.rs#L322 and it seems 101 is a generic code for **all** CLI errors so realistically I don't expect getting anything else from cargo sadly (I wish they had a more sophisticated list of error codes being an established CLI tool) so I'm not sure this "magical" constant provides much value anyway.

eskultety (commented):
Can we have a really dumb unit test verifying the exception type raised?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -141,7 +142,20 @@ def _resolve_cargo_package(
     log.info("Fetching cargo dependencies at %s", package_dir)
     with _hidden_cargo_config_file(package_dir):
         # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
-        config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
+        try:
+            config_template = run_cmd(cmd=cmd, params={"cwd": package_dir})
+        except subprocess.CalledProcessError as e:
+            # Search for a very specific failure state to better report it.
+            lock_corruption_marker1 = "failed to sync"
+            lock_corruption_marker2 = "needs to be updated but --locked was passed"
```
slimreaper35, line 150: Do we want to do a verbatim check for internal cargo error messages ?

hermeto/core/package_managers/cargo/main.py
eskultety: > It does not indeed add much value to be checking it. Since it can happen due to quite a few cargo errors and I am interested in a very specific one I opted for a verbatim check. 

And hence I questioned its usefulness in context of matching verbatim strings in the CLI output, IOW we may as well just check the strings ignoring the exit code completely.

> Unfortunately this won't work -- it is not something users have much influence on. The PR stems from reports of failures in dependencies (see https://github.com/hermetoproject/hermeto/issues/925 and a linked jiter issue for details). Even if a user can check that they can't do much about it. The purpose of this PR is to surface the fact that some dependency is misbehaving

Fair enough, the commit msg is descriptive enough justifying this change, I can dislike it all I want, I couldn't think of a better way of exposing the problem, hopefully it won't be as common as I fear it would leading to false negatives and hence continuation of weird errors in our log.

eskultety: > Once https://github.com/hermetoproject/hermeto/issues/934 is implemented in some form we should amend the exception with a suggestion to use a less strict mode to try and pass the build.

I'm not sure we want to advertise the permissive mode too much otherwise users will happily use it as a general solution to all their problems hoping they'll disappear and builds will start passing instead of looking at the real problems underneath, but I see your point.

taylormadore: The only guard I could think of against a changing interface here is having a test that reproduces and checks for the expected behavior. With that said it feels like more effort than is warranted for a more specific error message.

a-ovchinnikov: > a test that reproduces and checks for the expected behavior

Yes, that would give us an early warning, however the very idea of basing anything off of a volatile and not strictly specified part of a contract is fundamentally flawed. Unfortunately I could not find a more robust foundation for the check, so had to go with what we have.

taylormadore (approved):

hermeto/core/package_managers/cargo/main.py
```diff
@@ -56,6 +74,8 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
     environment_variables: list[EnvironmentVariable] = []
     project_files: list[ProjectFile] = []
 
+    # raise Exception(f"{request=}")
```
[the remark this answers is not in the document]
a-ovchinnikov: Yes

docs/pip.md
```diff
@@ -511,6 +511,15 @@ Note, that a system which is to be used for building these extensions must have
...
 build and install a Rust-based extension.
 
+Note, that sometimes Rust-based extensions can break a build. This could happen when
+such dependency is distributed with a Cargo.lock not matching Cargo.toml (while rare this seems
+to happen due to peculiarities of the release process for some packages). In this case
+a package will be rejected with a note about lock file mismatch and an additional report from
+Cargo about inability to load package lock file due to a mismatch. There is no
+good solution for this problem on Hermeto's side and the best course of action is to reach out
+to maintainers of this extension and notify them about the mismatch. Switching to a binary
+distribution of the package would also resolve this problem at the price of not building
```
eskultety, line 521: I think the last sentence warrants a small reference back to `### Building with wheels` in this same document.

eskultety (approved):
Tiny docs nitpick, ACK.

PR #930, opened 2025-05-05
Konflux onboarding: Add a dedicated Konflux Dockerfile

brunoapimentel:
This PR will introduce a new Dockerfile that will be used in the future to enable Konflux builds. For now, this Dockerfile will have no effect in the repository, and it is worth mentioning that it can't be built locally without quite a few specific configurations. 

I tried to keep all the files that are Konflux-specific under the `.konflux` directory, so we can easily move them out in the future in case we decide to do so. Let me know if you agree with this approach.

### DIFF
--- a/.containerignore
+++ b/.containerignore
@@ -6,3 +6,4 @@
 !pyproject.toml
 !requirements.txt
 !MANIFEST.in
+!.konflux/requirements-build.txt
--- a/.konflux/Dockerfile.konflux
+++ b/.konflux/Dockerfile.konflux
@@ -1,0 +1,85 @@
+# This file is meant to be built hermetically in a Konflux instance.
+# It can only be build locally if the prefetching of dependencies is properly configured.
+
+FROM registry.access.redhat.com/ubi9/ubi@sha256:a1804302f6f53e04cc1c6b20bc2204d5c9ae6e5a664174b38fbeeb30f7983d4e as ubi
+
+###########################
+# PREPARE GENERIC ARTIFACTS
+###########################
+FROM ubi as artifacts
+
+RUN mkdir /artifacts && \
+    cd /artifacts && \
+    mkdir go1.20 go1.21 nodejs && \
+    # Assume that the Hermeto prefetched content will be mounted to /cachi2/output
+    tar -xzf /cachi2/output/deps/generic/go1.20.linux-amd64.tar.gz -C ./go1.20 && \
+    tar -xzf /cachi2/output/deps/generic/go1.21.0.linux-amd64.tar.gz -C ./go1.21 && \
+    # Use --strip-components to change the default nodejs-{version}-{os}-{arch} naming schema
+    tar -xzf /cachi2/output/deps/generic/nodejs.tar.gz -C ./nodejs --strip-components 1
+
+########################
+# PREPARE OUR BASE IMAGE
+########################
+FROM ubi as base
+RUN dnf -y install \
+    --setopt install_weak_deps=0 \
+    --nodocs \
+    git-core \
+    jq \
+    python3 \
+    rubygem-bundler \
+    rubygem-json \
+    subscription-manager && \
+    dnf clean all
+
+###############
+# BUILD/INSTALL
+###############
+FROM base as builder
+WORKDIR /src
+RUN dnf -y install \
+    --setopt install_weak_deps=0 \
+    --nodocs \
+    gcc \
+    # not a build dependency, but we copy the binary to the final image
+    cargo \
+    python3-devel \
+    python3-pip \
+    python3-setuptools \
+    && dnf clean all
+
+# Install dependencies in a separate layer to maximize layer caching
+COPY requirements.txt .
+COPY .konflux/requirements-build.txt requirements-build.txt
+
+RUN python3 -m venv /venv && \
+    /venv/bin/pip install --upgrade pip && \
+    /venv/bin/pip install -r requirements.txt --no-deps --no-cache-dir --require-hashes
+
+COPY . .
+RUN /venv/bin/pip install setuptools && \
+    /venv/bin/pip install --no-cache-dir .
+
+##########################
+# ASSEMBLE THE FINAL IMAGE
+##########################
+FROM base
+LABEL maintainer="Red Hat"
+
+# copy Go SDKs and Node.js installation from official images
+COPY --from=artifacts /artifacts/go1.20/go /usr/local/go/go1.20
+COPY --from=artifacts /artifacts/go1.21/go /usr/local/go/go1.21
+COPY --from=artifacts /artifacts/nodejs/lib/node_modules/corepack/dist/corepack.js /usr/local/lib/corepack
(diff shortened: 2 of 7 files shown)


.konflux/artifacts.lock.yaml
```diff
@@ -0,0 +1,10 @@
+metadata:
```
eskultety, line 1: **Commit subject nitpick:**
`s/Create/Define OR Introduce` <<< sounds better in my head in connection with "the generic ..." that follows :)

.konflux/rpms.in.yaml
```diff
@@ -0,0 +1,19 @@
+contentOrigin:
```
eskultety, line 1: **Commit msg**:
`rpms-lockfile-prototype¹` <<< There's some footer note reference that is missing a link. Using old-school reference markers `[1..N]` is also probably slightly more visible :).

.konflux/rpms.in.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - python3-pip
+  - python3-setuptools
+  - python3-devel
+  - rubygem-bundler
+  - rubygem-json
+  - subscription-manager
+context:
+  containerfile:
+    file: Dockerfile.konflux
+    stageName: ubi
```
eskultety, line 19: nitpick: This line might be obvious to some who have used the prototype generator several times already, but I feel like adding a brief commentary on what this setting does AND/OR why it's necessary. Just my 2c, others might feel differently about it.

eskultety (commented):
I imagine converging to this very shape of the changes must have been a painful process, BUT it looks very straightforward to me. I also much appreciate the intent to put all of this to a hidden `konflux` directory to kinda abstract upstream from the changes and avoid confusion when looking at the repository.
Basically I'm missing 2 things for an approval:

1. Can you attach a snippet of a successful build log to the PR using these changes?
2. What further changes do we need to our CI for the Konflux GH action to consume the changes in this PR?

.containerignore
```diff
@@ -6,3 +6,4 @@
 !pyproject.toml
 !requirements.txt
 !MANIFEST.in
+!.konflux/requirements-build.txt
```
slimreaper35, line 9: nitpic: This should not be needed if we copy the requirements file directly in the Dockerfile.

eskultety: I don't think this is true. If you look at the ignore file the very first line says `*` which means all files should be excluded from the build context and then we re-include files/directories explicitly. If `.konflux/requirements-build.txt` is missing from the build context, then the `COPY` command cannot work because the builder would not see it in the build context.

slimreaper35: Yes, you are right. Closing thread.

.konflux/ubi.repo
```diff
@@ -0,0 +1,62 @@
+[ubi-9-for-x86_64-baseos-rpms]
+name = Red Hat Universal Base Image 9 (RPMs) - BaseOS
+baseurl = https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/$basearch/baseos/os
+enabled = 1
+gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
+gpgcheck = 1
```
eskultety, line 6: Do we need to maintain this file at all if all we need is to fetch packages from BaseOS and Appstream? Recently, we did the following to go about this just fine: https://github.com/hermetoproject/integration-tests/blob/multiple/gomod-and-npm/rpms.in.yaml#L3-L6

.konflux/rpms.in.yaml
```diff
@@ -14,6 +15,8 @@ packages:
   - rubygem-json
   - subscription-manager
 context:
+  # When resolving the packages, use the image defined in this specific stage of the
+  # referenced Dockerfile
```
eskultety, line 19: Same here, should be in previous commit.

.konflux/rpms.in.yaml
```diff
@@ -1,3 +1,4 @@
+# This file was copied directly from the UBI image referenced in ./Dockerfile.konflux
```
eskultety, line 1: Should be in previous commit, not this one :).

eskultety (commented):
Minor adjustment, but I guess we're good to go. I'd still like to see a response to:

> Can you attach a snippet of a successful build log to the PR using these changes?

to be sure these changes are complete wrt/ hermetic builds, otherwise I can comment solely on the diff whether it makes sense or not, but without a proof it actually works as intended.

.konflux/rpms.lock.yaml
```diff
@@ -0,0 +1,326 @@
+# This file was generated via the following command:
```
taylormadore, line 1: Not something missing from this PR specifically--just an open question: What is our process going to be for keeping this up-to-date?

brunoapimentel: This is supported by Renovate/Mintmaker, so we'll be getting updates via automated PRs.

.konflux/ubi.repo
brunoapimentel: We need a small [tweak](https://github.com/brunoapimentel/hermeto/commit/f3114611956d5344a6d1e0217f0a90ba00b3d1a2) to have the repoids that will end up in the SBOM to comply with Konflux's rules. So we can either keep this file as is, or have a `sed` command to update the resulting `rpms.lock.yaml` file.

Having the file here is probably a little simpler, but I have no objections about going the other way.

.konflux/rpms.lock.yaml
brunoapimentel: So we will need to move the file out of the nicely organized `.konflux` folder =/

brunoapimentel:
> Did pybuild-deps not work? Is it missing an --allow-binary option?

It does work, but we're opting for doing the build using [prefetched wheels](https://github.com/brunoapimentel/hermeto/blob/konflux-hermeto/.tekton/hermeto-pull-request.yaml#L34-L43) for now. Even with `allow_binary` set to `true`, these build requirements ended up being necessary. Pybuild-deps would have generated the full list of build dependencies, which was overkill.

.konflux/Dockerfile.konflux
```diff
@@ -0,0 +1,86 @@
...
+COPY .konflux/requirements-build.txt requirements-build.txt
+
+RUN python3 -m venv /venv && \
+    /venv/bin/pip install --upgrade pip && \
+    /venv/bin/pip install -r requirements.txt --no-deps --no-cache-dir --require-hashes && \
+    /venv/bin/pip install -r requirements-build.txt --no-deps --no-cache-dir
+
+COPY . .
+RUN /venv/bin/pip install setuptools && \
+    /venv/bin/pip install --no-cache-dir .
```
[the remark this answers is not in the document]
brunoapimentel: Good catch. Fixed it.

taylormadore:
Just reviewing stale PRs: This looks like it has been overcome-by-events and is no longer relevant? I think we can close this after the new downstream is live

brunoapimentel:
Closing because we abandoned the intent of onboarding this repository to Konflux.

PR #941, opened 2025-05-07
cli: Add global mode option for input validation strictness

slimreaper35:
Closes: https://github.com/hermetoproject/hermeto/issues/934

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,19 @@ You might also like to check out `hermeto --help` and the `--help` texts of the 
 
 ## Configuration
 
+### Modes
+
+Hermeto can be run in two modes using the global CLI option `--mode`.
+
+* strict
+* permissive
+
+The default mode is `strict`. In this mode, _some_ input requirements that are not met are treated
+as errors. On the other hand, the `permissive` mode treats them as warnings. This means that Hermeto
+will proceed and generate an SBOM, but it may not be complete or accurate.
+
+### Available configuration parameters
+
 You can change Hermeto's configuration by specifying a configuration file while invoking any of the CLI commands:
 
 ```shell
@@ -106,8 +119,6 @@ Any parameter specified in this file will override the default values present in
 [config.py](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/config.py) module.
 
 The only supported format for the config file is YAML.
-
-### Available configuration parameters
 
 * `default_environment_variables` - a dictionary where the keys
 are names of package managers. The values are dictionaries where the keys
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -1,3 +1,4 @@
+import enum
 import logging
 from pathlib import Path
 from typing import TYPE_CHECKING, Annotated, Any, Callable, Dict, Literal, Optional, TypeVar, Union
@@ -60,6 +61,13 @@ Flag = Literal[
 ]
 
 
+class Mode(str, enum.Enum):
+    """Represents a global CLI option to relax input expectations and requirements checks."""
+
+    STRICT = "strict"
+    PERMISSIVE = "permissive"
+
+
 class _PackageInputBase(pydantic.BaseModel, extra="forbid"):
     """Common input attributes accepted for all package types."""
 
@@ -100,7 +108,6 @@ class SSLOptions(pydantic.BaseModel, extra="forbid"):
 
     @pydantic.model_validator(mode="after")
     def _validate_ssl_options(self) -> Self:
-
         cert_and_key = (self.client_cert, self.client_key)
         if any(cert_and_key) and not all(cert_and_key):
             raise ValueError(
@@ -254,6 +261,7 @@ class Request(pydantic.BaseModel):
     output_dir: RootedPath
     packages: list[PackageInput]
     flags: frozenset[Flag] = frozenset()
+    mode: Mode = Mode.STRICT
 
     @pydantic.field_validator("packages")
     def _unique_packages(cls, packages: list[PackageInput]) -> list[PackageInput]:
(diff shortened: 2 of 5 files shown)


hermeto/interface/cli.py
```diff
@@ -174,6 +174,11 @@ def main(  # noqa: D103; docstring becomes part of --help message
         case_sensitive=False,
         help="Set log level.",
     ),
+    mode: Mode = typer.Option(
+        Mode.STRICT,
+        "--mode",
+        help="Set input expectations and requirements validation mode.",
```
taylormadore, line 180: I'm trying to think how to make this a little more clear without making it too long :sweat_smile:

What I feel like we want to convey is: "Set best‑practice validation mode: strict fails fast, while permissive converts violations to warnings, continues, and potentially reduces the accuracy of the SBOM".

hermeto/core/models/input.py
```diff
@@ -60,6 +61,13 @@ def show_error(error: "ErrorDict") -> str:
 ]
 
 
+class Mode(str, enum.Enum):
+    """Represents global CLI option."""
```
taylormadore, line 65: Can we be a little more specific to this specific option with this docstring?

taylormadore (commented):
LGTM with some minor comments

a-ovchinnikov (commented):
LGTM once https://github.com/hermetoproject/hermeto/pull/941/commits/ae1a9ac7882878391b2560c94834d90a744aa065#r2077606269 is addressed.

tests/unit/test_cli.py
```diff
@@ -180,6 +180,23 @@ def test_config_file_option_invalid(
...
 
+    @pytest.mark.parametrize(
+        "mode, is_valid",
+        [
+            ("strict", True),
+            ("permissive", True),
+            ("bad", False),
+        ],
+    )
+    def test_mode_option(self, mode: str, is_valid: bool) -> None:
```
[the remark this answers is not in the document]
eskultety: Yes, most of the time (like in this case), otherwise, applying common sense is the way to go.

hermeto/interface/cli.py
eskultety: How about:
`Input validation mode, 'strict' treats all violations as fatal errors, 'permissive' just audits them`

eskultety, line 180: This is a NOP at the moment, can we mark it as such in the CLI help? Putting something like `(currently a NOP)` to the help output?

eskultety (commented):
Since this is a user-facing option, I'm lacking a documentation mention where the violation case can be explained to a more detail along with a note that this doesn't really do anything at the moment and that pieces of logic impacted by the setting will be added gradually.

hermeto/interface/cli.py
a-ovchinnikov: That would need some hook to not forget to remove it once it is no longer a stub.

eskultety: Fair enough, just docs it is then.

tests/unit/test_cli.py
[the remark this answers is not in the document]
slimreaper35: Done.

hermeto/interface/cli.py
slimreaper35: I updated the README.md with a detailed explanation.

> "Set best‑practice validation mode: strict fails fast, while permissive converts violations to warnings, continues, and potentially reduces the accuracy of the SBOM"
> Input validation mode, 'strict' treats all violations as fatal errors, 'permissive' just audits them

If we don't want to add more modes, then I guess it's fine. Otherwise, the help message will get very long.

README.md
```diff
@@ -96,6 +96,19 @@ You might also like to check out `hermeto --help` and the `--help` texts of the
...
 
+### Modes
+
+Hermeto can be run in two modes using the global CLI option `--mode`.
+
+* strict
+* permissive
+
+The default mode is `strict`. This mode treats all violations as fatal errors, and Hermeto will
+refuse to proceed. Meanwhile, the `permissive` mode treats all violations as warnings and Hermeto
```
eskultety, line 107: > mode treats all violations as warnings

This is a very bold statement and we need to soften the language - with the best of intentions I am not confident we'll be able to convert 100% of what we imagine to mere warnings, there will always be configuration setups (though we may not know about them yet at this point) that will simply prevent us from proceeding.

hermeto/interface/cli.py
eskultety: The current wording is still pretty vague to me, okay, let's try another attempt:

_treat input requirement violations as errors|warnings (the latter may affect SBOM accuracy)_

^Still too long? :)

README.md
slimreaper35: Updated the paragraph.

eskultety: Just a side note - I think as we put the flag in practice the paragraph should grow with listing examples of violation instances which the flag can be applied to.

eskultety (approved):
I think this is a good starting point. I'd prefer https://github.com/hermetoproject/hermeto/pull/941#discussion_r2077801976 to have been addressed, but at the same time, the docs paragraph is currently worded such that noone is really able to tell what input requirements it applies to which can also mean "none" and we can fine tune the docs with a further list of common requirements as we gradually enable it across all backends.

README.md
```diff
@@ -96,6 +96,19 @@ You might also like to check out `hermeto --help` and the `--help` texts of the
...
+### Modes
+
+Hermeto can be run in two modes using the global CLI option `--mode`.
+
+* strict
+* permissive
+
+The default mode is `strict`. In this mode, **some** input requirements that are not met are treated
+as errors. On the other hand, the `permissive` mode treats them as warnings. This means that Hermeto
+will proceed, but it may not be able to produce a complete and accurate SBOM.
```
a-ovchinnikov, line 108: This is a bit unclear: it makes it look like it is possible to have an error ignoring which would still result in a complete and accurate SBOM. Did you mean something like "Hermeto will not fail and generate an SBOM, but it will contain omissions and inaccuracies of varying degree."?

a-ovchinnikov (approved):
LGTM with two minor non-blocking observations.

taylormadore (approved):

PR #944, opened 2025-05-13
chore: Report correct version tag by configuring setuptools_scm

derasdf:
The weird version format was caused by missing version_scheme and local_scheme in [tool.setuptools_scm].

Closes: https://github.com/hermetoproject/hermeto/issues/822

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,8 @@ packages = ["hermeto"]
 
 [tool.setuptools_scm]
 fallback_version = "0.0.0+dev.fallback"
+version_scheme = "only-version"
+local_scheme = "no-local-version"
 
 [tool.black]
 line-length = 100


eskultety:
If you look at https://github.com/hermetoproject/hermeto/issues/822#issuecomment-2753443377 I don't think the proposed changes address any of that, because the way our current release process is set up (which is far from good), the git tag doesn't really exist at the time when the container image (which is to be released) is built, because the semver tag is pushed via the "on-release" PaC trigger on an existing image that has been built from latest HEAD at some point.

When it comes to the proposed changes themselves:
`only-version` - it's debatable whether this is useful at all because the only thing this does is suppose you have locally a bunch of commits which would otherwise report `cachi2 X.Y.post1+<digest>.<datestamp>` and so now it would report `cachi2 X.Y`. But that setting is only ever going to be exercised in local dev envs because if the release process is done correctly, the top commit will already be tagged as `X.Y` so setuptools-scm doesn't really have a choice to report something else regardless of the `only-version` setting.

`local_scheme = "no-local-version"` - I have to admit that without trying this out, the [docs](https://setuptools-scm.readthedocs.io/en/latest/extending/#setuptools_scmversion_scheme) is unclear to me on what exactly it does, IIUC this would only remove the `.<datestamp>` part of the string from my example above? So what does `local_scheme` do if `version_scheme` is already set to `only-version`?

In any case, unless I'm terribly misunderstanding something I'm not seeing much value in this proposed change and I'm hence tempted to NACK it unless it proves me otherwise. At the very least, this PR definitely doesn't close the linked issue, because the problem will still persist even after due to our current release process setup.

PR #945, opened 2025-05-14
Use gz compression for RPM repository metadata

slimreaper35:


eskultety (commented):
IIUC this was seen in the wild on some RHEL-7 systems. I think this change warrants an integration test based on a UBI7 to make sure we're properly covered. Can be an existing integration test copy, content doesn't matter as long as the base image is tested.

slimreaper35:
https://github.com/hermetoproject/integration-tests/pull/7

> IIUC this was seen in the wild on some RHEL-7 systems. I think this change warrants an integration test based on a UBI7 to make sure we're properly covered. Can be an existing integration test copy, content doesn't matter as long as the base image is tested.

:+1: While testing, I realized that `--general-compress-type` is the option to go with instead of  `--compress-type`.

a-ovchinnikov (approved):

eskultety:
> [hermetoproject/integration-tests#7](https://github.com/hermetoproject/integration-tests/pull/7)
> 
> > IIUC this was seen in the wild on some RHEL-7 systems. I think this change warrants an integration test based on a UBI7 to make sure we're properly covered. Can be an existing integration test copy, content doesn't matter as long as the base image is tested.
> 
> 👍 While testing, I realized that `--general-compress-type` is the option to go with instead of `--compress-type`.

Do share your findings :), was the original option not present on R7 and the new incantation you went with serves as an alias these days or do the options have different meaning? I'm lazy to check, sorry...

tests/integration/test_rpm.py
```diff
@@ -230,6 +230,26 @@ def read_and_normalize_repofile(path: Path) -> str:
...
+                    },
+                ),
+                flags=["--dev-package-managers"],
+                check_vendor_checksums=False,
+                expected_exit_code=0,
+                expected_output="All dependencies fetched successfully",
+            ),
+            ["vim", "--version"],
+            ["Vi IMproved 7.4"],
+            id="rpm_compress_type",
```
eskultety, line 251: nitpick: the test could be renamed to something more meaningful, see the suggestions you got in the other pr.

eskultety (commented):
1 minor non-functional comment, otherwise LGTM but you have an outstanding issue with the integration test and I believe it has something to do with the external repo file you used as I commented on in the other linked PR.

slimreaper35:
> Do share your findings :), was the original option not present on R7 and the new incantation you went with serves as an alias these days or do the options have different meaning? I'm lazy to check, sorry...

Sharing the snippet from createrepo_c help message, that is quite verbose so it's easy to overlook an option.

```
--compress-type=COMPRESSION_TYPE    Which compression type to use for additional metadata files (comps, updateinfo, etc). Supported values are: bz2, gz, zck, zstd, xz.

--general-compress-type=COMPRESSION_TYPE    Which compression type to use (even for primary, filelists and other xml). Supported values are: bz2, gz, zck, zstd, xz.
```

slimreaper35:
oh no, some rust e2e tests are failing

eskultety:
> oh no, some rust e2e tests are failing

Oh crap. The problem doesn't come from this PR, but is caused by what I advocated for e.g. in https://github.com/hermetoproject/hermeto/pull/911#discussion_r2077406378, i.e. not fixing the test image to a particular hash. In general, I stand by that argument that in order to have thorough testing detecting possible system package dependency and subpackage naming changes early (this really happens, mostly on Fedora though) before users can experience them in production, one cannot be fixed to a particular image hash in tests. 
However, distro package formats (e.g. RPM) in combination with hermetic builds were simply never made for this - in this case, the updated UBI image comes with a newer python3 preinstalled, so when we try to "install" it again in the Dockerfile, we only prefetched an old version (because we don't have a bot watching the RPM lockfile) of python which naturally won't be installed. Even if we omitted python3 because it comes pre-installed, `python3-devel` would fail the same way.

Sigh. I don't see a good way out of this other than pinning all RPM-based test container images to a particular digest, but we'll have to live with the fact that we'll always be testing against obsolete images and not with the latest ones to detect potential issues early. That said, the silver lining for us testing with old RPM images can be that we're only really testing fetching logic and not dependency resolution which we have no business with, so using old & obsolete images might not be a tragedy. However, doing that with other images could be potentially detrimental to our ability to test because e.g. we already know that future Node images will not come with yarn v1 nor corepack pre-installed which is something that would affect our testing with generating false negatives.

eskultety:
> > in order to have [thorough testing] [base image] cannot be fixed to a particular image hash in tests.
> 
> I agree with this statement in general, however would like to note that this would mix test scenarios and instead of testing just one aspect or just a group of closely related aspects we will be testing unrelated components as well -- this way there is always a risk of sporadic tests failures due to some unrelated changes somewhere. Pinning a hash, on the other hand, removes such risks. 

Since I expect all e2e tests to only ever be run with network isolation, then it does eliminate those risks, but only based on that assumption, because platform package managers behave differently than the language ones and the moment you'd consider online access in any of those tests, pinning the container image would not help you at all with reproducibility because the repository metadata is a separate entity living its own life separate from the base image contents.

eskultety (approved):

PR #952, opened 2025-05-15
cargo: Force old-style config to the new style

a-ovchinnikov:
Turns out it is still possible to occasionally find projects that keep pre-v1.39.0 style config which with any newer cargo would take precedence over any existing new style config. Since the only important difference between the two styles of configs is their name just renaming an old style one is enough to proceed (unless the project is to be built with a very old Cargo+Rust combination, but this does not look likely at the moment).

This commit adds a transparent guard to implicitly rename the config and to thus ensure that Hermeto's changes to it are respected.

Addresses: https://github.com/hermetoproject/hermeto/issues/946


taylormadore:
Will this have any effect on the user build if we are only renaming this file in an ephemeral copy of the source?

a-ovchinnikov:
> Will this have any effect on the user build if we are only renaming this file in an ephemeral copy of the source?

You are right, it won't. I think this is a rare enough case to be documented and handled by the user: they would need to remove the old config manually, the new one will be injected by Hermeto. I don't think adding logic to _remove_ files during injection is warranted at this point, and would require replacing `inject-files` with `prepare-repository-for-build` or something similar, essentially introducing a big and breaking change for a very localized problem.

a-ovchinnikov:
In the updated version I followed @taylormadore's suggestion and made the selection of a config transparent to the user. As a result there is no need for a manual intervention and a documentation change and the code simply checks which config file to inject basing on what is present in a repository.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -137,13 +137,27 @@ def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None,
...
+    # to .cargo/config.toml for projects that are built with both old and new versions
+    # of Cargo. Unlinking a symlink first is safe.
+    all_possible_config_names = (".cargo/config", ".cargo/config.toml")
+    configs_contents = []
+
+    for cfgname in all_possible_config_names:
+        config = package_dir.join_within_root(cfgname)
+        data = config.path.read_text() if config.path.exists() else None
+        configs_contents.append((config, data))
+        config.path.unlink() if data is not None else do_nothing
```
eskultety, line 154: nitpick (readability improvement): Pre-existing, but since you're touching the code this compound statement should be written as multiline instead of a one-liner. It's one thing to do this with assignments, but statements are harder to read this way and I'd argue it feels like unnecessary obfuscation. As a bonus, we'll be able to drop the `do_nothing` variable then as it's only needed to satisfy the one-liner condition syntax.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -137,13 +137,27 @@ def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None,
...
+
     try:
         yield
     finally:
-        config.path.write_text(data) if data is not None else do_nothing
+        for config, data in configs_contents:
+            config.path.write_text(data) if data is not None else do_nothing
```
eskultety, line 160: Same here with the one-liner statement...

hermeto/core/package_managers/cargo/main.py
```diff
@@ -202,9 +216,26 @@ def _swap_sources_directory_for_subsitution_slot(template: str) -> dict:
...
+    # take precedence on .cargo/config.toml if present and the latter one would be ignored.
+    # The recommended practice for dealing with a situation when an older build system
+    # has to build a more modern project is to symlink .cargo/config.toml to .cargo/config.
+    # And vice versa: renaming .cargo/config to .cargo/config.toml would have no effect on
+    # any post-2019 toolchain.
+    # Refer to https://doc.rust-lang.org/cargo/reference/config.html for further details.
+    # Since we could potentially end up building a somewhat stale Rust-based
+    # Python extension it is better to check if there is an old-style config present and
+    # process it if found.
+    cf_name = ".cargo/config" if _old_config_is_present_in(package_dir) else ".cargo/config.toml"
```
eskultety, line 237: nitpick (readability): I still don't know how defining an extra function just to be able to do ^this is better than plain
```
cf_name = ".cargo/config.toml"
if (package_dir.path / ".cargo/config").exists():
  cf_name = ".cargo/config"
```
I think we should limit the one-liner syntax to elementary checks (possibly the named lambdas we already have elsewhere), but overall we should not abuse the syntax so as to define alias functions for simple checks, it's IMO easier to decode the check than to jump into the definition and back in the editor. My 2c.

eskultety (commented):
Thumbs up for the approach taken. I think we can do slightly better on the readability side (minor).

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Ok, I find both styles quite readable with slight preference to one-liners, but won't insist on it.

slimreaper35 (approved):

taylormadore (approved):

PR #953, opened 2025-05-16
Allow URLs with .whl file extension when wheels are allowed

slimreaper35:

### DIFF
--- a/hermeto/core/package_managers/pip.py
+++ b/hermeto/core/package_managers/pip.py
@@ -9,7 +9,6 @@ import os
 import os.path
 import re
 import tarfile
-import urllib
 import zipfile
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
@@ -28,6 +27,7 @@ from typing import (
     Union,
     cast,
 )
+from urllib import parse as urlparse
 
 import tomlkit
 from packageurl import PackageURL
@@ -311,10 +311,10 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
     elif dependency_kind == "vcs":
         qualifiers = {"vcs_url": package["version"]}
     elif dependency_kind == "url":
-        parsed_url = urllib.parse.urldefrag(package["version"])
-        fragments: dict[str, list[str]] = urllib.parse.parse_qs(parsed_url.fragment)
+        defragmented_url, fragment = urlparse.urldefrag(package["version"])
+        fragments: dict[str, list[str]] = urlparse.parse_qs(fragment)
         checksum: str = fragments["cachito_hash"][0]
-        qualifiers = {"download_url": parsed_url.url, "checksum": checksum}
+        qualifiers = {"download_url": defragmented_url, "checksum": checksum}
     else:
         # Should not happen
         raise RuntimeError(f"Unexpected requirement kind: {dependency_kind}")
@@ -1449,12 +1449,12 @@ class PipRequirement:
         if "; " in url:
             url, environment_marker = url.split("; ", 1)
 
-        parsed_url = urllib.parse.urlparse(url)
+        parsed_url = urlparse.urlparse(url)
         if parsed_url.fragment:
             for section in parsed_url.fragment.split("&"):
                 if "=" in section:
                     attr, value = section.split("=", 1)
-                    value = urllib.parse.unquote(value)
+                    value = urlparse.unquote(value)
                     qualifiers[attr] = value
                     if attr == "egg":
                         # Use the egg name as the package name to avoid ambiguity when both are
@@ -1799,7 +1799,7 @@ def _validate_requirements(requirements: list[PipRequirement], allow_binary: boo
 
         # Fail if VCS requirement uses any VCS other than git or does not have a valid ref
         elif req.kind == "vcs":
-            url = urllib.parse.urlparse(req.url)
+            url = urlparse.urlparse(req.url)
 
             if not url.scheme.startswith("git"):
                 raise UnsupportedFeature(
@@ -1839,7 +1839,7 @@ def _validate_requirements(requirements: list[PipRequirement], allow_binary: boo
             else:
                 allowed_extensions = SDIST_FILE_EXTENSIONS
 
-            url = urllib.parse.urlparse(req.url)
+            url = urlparse.urlparse(req.url)
             if not any(url.path.endswith(ext) for ext in allowed_extensions):
                 msg = (
                     "URL for requirement does not contain any recognized file extension: "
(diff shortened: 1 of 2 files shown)


a-ovchinnikov (commented):
The code as it is LGTM, just a bit unclear where does the problem which the patch addresses come from. Was there an issue reported or did someone run into an unexpected behavior? If yes what exactly happened? I think adding a couple of lines of background to the commit message would greatly help future us and reviewers. It is also hard to tell whether the original problem was resolved.

hermeto/core/package_managers/pip.py
```diff
@@ -72,9 +72,7 @@
 ZIP_FILE_EXT = ".zip"
 COMPRESSED_TAR_EXT = ".tar.Z"
 SDIST_FILE_EXTENSIONS = [ZIP_FILE_EXT, ".tar.gz", ".tar.bz2", ".tar.xz", COMPRESSED_TAR_EXT, ".tar"]
-SDIST_EXT_PATTERN = r"|".join(map(re.escape, SDIST_FILE_EXTENSIONS))
-
-PYPI_URL = "https://pypi.org"
+WHEEL_FILE_EXTENSIONS = [".whl"]
```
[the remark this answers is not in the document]
slimreaper35: Maybe just a one-string constant is enough here :thinking:

hermeto/core/package_managers/pip.py
```diff
@@ -72,9 +72,7 @@
 ZIP_FILE_EXT = ".zip"
 COMPRESSED_TAR_EXT = ".tar.Z"
 SDIST_FILE_EXTENSIONS = [ZIP_FILE_EXT, ".tar.gz", ".tar.bz2", ".tar.xz", COMPRESSED_TAR_EXT, ".tar"]
```
eskultety, line 74: **Commit message:**

> for checking the type of package distribution (sdist or wheel) [1].

It's nice you're referencing some work, but note that you referenced a particular line on the main branch which is likely to change in the future, making the context reference invalid - either pick it from a particular release or just paste a tiny code snippet verbatim to the commit message, hence making it immutable. I guess I prefer the latter, but I'd be fine with either.

eskultety (commented):
Small comments, but LGTM.

eskultety (approved):

taylormadore:
I get a stack trace when trying to run this change against the repo+ref that generated the ticket: https://github.com/raptorsun/lightspeed-rag-content/tree/byok-tool-build

It looks like we need to account for this change here: https://github.com/hermetoproject/hermeto/blob/774a4119b8bdc4e038e2bf4f050cf93b9a572d4a/hermeto/core/package_managers/pip.py#L2328

slimreaper35:
Thanks @taylormadore !
It should work now. Can you double check ?

(Added one small commit)

taylormadore:
@slimreaper35  Have you tried to install from the injected requirements file? Pip seems unhappy with our external deps naming format for wheels:

> Processing /remote-source/deps/pip/external-torch/torch-external-sha256-5b6ae523bfb67088a17ca7734d131548a2e60346c622621e4248ed09dd0790cc (from -r requirements.cpu.txt (line 96))
ERROR: torch@ file:///remote-source/deps/pip/external-torch/torch-external-sha256-5b6ae523bfb67088a17ca7734d131548a2e60346c622621e4248ed09dd0790cc from file:///remote-source/deps/pip/external-torch/torch-external-sha256-5b6ae523bfb67088a17ca7734d131548a2e60346c622621e4248ed09dd0790cc (from -r requirements.cpu.txt (line 96)) does not appear to be a Python project: neither 'setup.py' nor 'pyproject.toml' found.

eskultety:
> Seems like that `pip` is a next candidate for big refactoring after `gomod`. Having some classes would really help to reduce duplication.

Yes.

hermeto/core/package_managers/pip.py
```diff
@@ -72,9 +72,7 @@
 ZIP_FILE_EXT = ".zip"
 COMPRESSED_TAR_EXT = ".tar.Z"
```
eskultety, line 73: nitpick: Pre-existing, but since you're touching the code you could just add one trivial change - move ^this one to the list of `SDIST_FILE_EXTENSIONS` and hard-code it on the single spot where it's used to skip metadata parsing. The reasoning behind this is that it's confusing because `.tar.gz, .tar.bz2, .tar.xz` are all compressed, it's just that `.tar.Z` is older and as a matter of fact `tarfile` doesn't support it, no idea why we're not attempting to read the metadata (historical reasons...)

hermeto/core/package_managers/pip.py
```diff
@@ -1834,11 +1835,15 @@ def _validate_requirements(requirements: list[PipRequirement]) -> None:
                     docs=PIP_EXTERNAL_DEPS_DOC,
                 )
 
+            allowed_extensions = list(SDIST_FILE_EXTENSIONS)
+            if allow_binary:
+                allowed_extensions.append(WHEEL_FILE_EXTENSION)
```
eskultety, line 1840: optional:
```
if allow_binary:
  allowed_extensions = ALL_FILE_EXTENSIONS
else:
  allowed_extensions = SDIST_FILE_EXTENSIONS
```

eskultety (commented):
> Seems like that pip is a next candidate for big refactoring after gomod. Having some classes would really help to reduce duplication.

Yes, this module is in a much worse state than gomod.

hermeto/core/package_managers/pip.py
```diff
@@ -28,6 +27,7 @@
     Union,
     cast,
 )
+from urllib import parse
```
eskultety, line 30: Probably better to namespace this as `from urllib import parse as urlparse`. Apologies for having forgotten to mention it in the previous round of review.

hermeto/core/package_managers/pip.py
```diff
@@ -1799,14 +1796,15 @@ def _validate_requirements(requirements: list[PipRequirement]) -> None:
...
-            url = urllib.parse.urlparse(req.url)
+            url = parse.urlparse(req.url)
 
             if not url.scheme.startswith("git"):
                 raise UnsupportedFeature(
                     f"Unsupported VCS for {req.download_line}: {url.scheme} (only git is supported)"
                 )
 
-            if not GIT_REF_IN_PATH.search(url.path):
+            git_ref_in_path = re.compile(r"@[a-fA-F0-9]{40}$")
```
eskultety, line 1806: This particular change wasn't requested though. What this does is it'll compile the regex over and over again for every VCS `req` in the requirements list instead of just applying a pre-compiled one. It's definitely true that we don't expect many VCS type of requirements, so probably not a big loss if it's used **iff** a VCS req exists, but:
- you're already stretching it with the unrelated changes stuffed into this one particular commit to the point where a standalone commit should be introduced :)
- if matching a regex is a one-off task the call to `compile` is pointless because all `search`-like `re` APIs will compile the pattern themselves, IOW either do it once on the global level like it were before or just issue `re.search` directly

eskultety (approved):
I was able to install the deps prefetched with the latest revision -> ACK.

PR #961, opened 2025-05-26
Konflux onboarding: add Renovate configuration

brunoapimentel:
In order to avoid getting a flood of PRs by Renovate, we need to configure package group for the updates. The initial `renovate.json` file introduced here tries to mostly copy what we have configured for Dependabot. The only configuration I have not copied yet is the supression of `pydating-core` updates (see this commit), since I want to test how Renovate deals with this particular dependency.

In any case, this change assumes that we want all the updates to be handled by Renovate. Alternatively, we can still keep all the updates by Dependabot, and only rely on Renovate to update the Tekton CI definitions.

For more info on Renovate configuration, check the [official docs](https://docs.renovatebot.com/configuration-options/#packagerules).

This should only be merged after #930. 

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,49 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base"
+  ],
+  "packageRules": [
+    {
+      "groupName": "Tekton CI",
+      "matchFileNames": [
+        ".tekton/**"
+      ],
+      "schedule": "weekly"
+    },
+    {
+      "groupName": "Github Actions",
+      "matchManagers": [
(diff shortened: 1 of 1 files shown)


eskultety:
> Does the renovate bot support Containerfiles too ? Currently, we have a symlink to Dockerfile.

Yes, it does: https://docs.renovatebot.com/modules/manager/dockerfile/#default-config

taylormadore:
Just reviewing stale PRs: This looks like it has been overcome-by-events and is no longer relevant? I think we can close this after the new downstream is live

brunoapimentel:
Closing because we abandoned the intent of onboarding this repository to Konflux.

PR #974, opened 2025-06-02
Break pip module into multiple ones

slimreaper35:
Closes https://github.com/hermetoproject/hermeto/issues/487

### DIFF
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -4,20 +4,18 @@ import logging
 import tarfile
 import zipfile
 from collections.abc import Iterable, Iterator
-from dataclasses import dataclass, field
 from os import PathLike
 from pathlib import Path
-from typing import Any, Literal, Optional, Union, cast
+from typing import Any, Optional, Union
 from urllib import parse as urlparse
 
 import pypi_simple
-import requests
 from packageurl import PackageURL
-from packaging.utils import canonicalize_name, canonicalize_version
+from packaging.utils import canonicalize_name
 
 from hermeto.core.checksum import ChecksumInfo, must_match_any_checksum
 from hermeto.core.config import get_config
-from hermeto.core.errors import FetchError, PackageRejected, UnsupportedFeature
+from hermeto.core.errors import PackageRejected, UnsupportedFeature
 from hermeto.core.models.input import Request
 from hermeto.core.models.output import EnvironmentVariable, ProjectFile, RequestOutput
 from hermeto.core.models.property_semantics import PropertySet
@@ -26,6 +24,11 @@ from hermeto.core.package_managers.general import (
     async_download_files,
     download_binary_file,
     extract_git_info,
+)
+from hermeto.core.package_managers.pip.package_distributions import (
+    PIP_NO_SDIST_DOC,
+    DistributionPackageInfo,
+    process_package_distributions,
 )
 from hermeto.core.package_managers.pip.project_files import PyProjectTOML, SetupCFG, SetupPY
 from hermeto.core.package_managers.pip.requirements import (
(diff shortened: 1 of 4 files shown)


taylormadore (approved):
[non-blocking] [This](https://github.com/hermetoproject/hermeto/blob/86a6418690f866c885d7ff776f291daa4f504fbe/hermeto/core/package_managers/pip/main.py#L1058) feels like it belongs with the PipRequirementsFile

a-ovchinnikov (dismissed):

slimreaper35:
> [non-blocking] [This](https://github.com/hermetoproject/hermeto/blob/86a6418690f866c885d7ff776f291daa4f504fbe/hermeto/core/package_managers/pip/main.py#L1058) feels like it belongs with the PipRequirementsFile

Done. But only as a function, apparently it would require a lot of changes in unit tests to be a method.

eskultety:
@slimreaper35 Can we go one step further and extract `DistributionPackageInfo` (and friends) out of the main module as well?

slimreaper35:
> @slimreaper35 Can we go one step further and extract DistributionPackageInfo (and friends) out of the main module as well?

In this pull request ?

eskultety:
> > @slimreaper35 Can we go one step further and extract DistributionPackageInfo (and friends) out of the main module as well?
> 
> In this pull request ?

Yes.

slimreaper35:
New version:
- Added more functions into `requirements.py` (validate_requirements, validate_requirements_hashes)
- New minor improvements commits were added (some of them were necessary)
- There is also new `package_distributions.py` module so the main module has 816 LOC

tests/unit/package_managers/pip/test_project_files.py
```diff
@@ -5,7 +5,7 @@
 import pytest
 
 from hermeto.core.errors import BaseError, PackageRejected
-from hermeto.core.package_managers.pip import main as pip
+from hermeto.core.package_managers.pip.project_files import PyProjectTOML, SetupCFG, SetupPY
```
eskultety, line 8: **Commit message:**

> Unfortunately, this should part of
> https://github.com/hermetoproject/hermeto/commit/e21e487fbb95f83d216d6f76d6dcec081fc07961.
> (Apologies)

_"This should have been part of commit e21e487fbb95f83d216d6f76d6dcec081fc07961."_

- with commit refs digest is enough (instead of a URL), GH renderer can automatically recognize those and locally you can just double click to highlight and paste to terminal, win-win :)
- it's just code, no need to apologize ;)

tests/unit/package_managers/pip/test_main.py
```diff
@@ -52,6 +52,62 @@ def make_dpi(
     )
 
 
+def mock_pypi_simple_distribution_package(
```
eskultety: Future note - do not couple code movement with renaming, it only makes changes harder to trace and prolongs that particular commit's review time. With plain code movement, `git show --color-moved` can apply visual magic in most cases which just makes the reviewer see that there's close to 100% match and they move on. If you rename along the way, that magic is often lost. I'll let this fly this time, but please make sure to not pair renames with code movement in the future  ;).

hermeto/core/package_managers/pip/main.py
```diff
@@ -6,15 +6,13 @@
 import os
 import os.path
 import re
-import shutil
```
eskultety, line 9: In context of previous patch, just for the record this one is fairly good (yet still simple) example of a commit where code movement is correctly displayed by git (if you haven't used `--color-moved` before) and one doesn't need to focus on those lines in the diff at all :).

hermeto/core/package_managers/pip/main.py
```diff
@@ -521,127 +449,6 @@ def _download_dependencies(
...
-                solution=solution,
-                docs=docs,
-            )
-
-    processed_dpis.extend(wheels)
-
-    return processed_dpis
-
-
 def _sdist_preference(sdist_pkg: DistributionPackageInfo) -> tuple[int, int]:
```
eskultety: This function is now defined in 2 modules.

hermeto/core/package_managers/pip/main.py
```diff
@@ -521,127 +449,6 @@ def _download_dependencies(
...
-                docs = None
-            else:
-                solution = (
-                    "It seems that this version does not exist or isn't published as an"
-                    " sdist.\n"
-                    "Try to specify the dependency directly via a URL instead, for example,"
-                    " the tarball for a GitHub release.\n"
-                    "Alternatively, allow the use of wheels."
-                )
-                docs = PIP_NO_SDIST_DOC
```
eskultety, line 1359: With this we now have 2 modules referring to the same doc URL, one using a hardcoded URL, one a constant, can we put the constant to the `__init__.py` file and use it consistently since we apparently want to link to particular docs? That said, I'm not such a big fan of linking docs in error messages, docs is pretty much outdated all the time and so while the idea sounds nice, the execution may end up being confusing and you need to actively maintain the error messages if the location of the doc changes. Your call on this one, but please make us consistent again.

hermeto/core/package_managers/pip/main.py
slimreaper35: oh no!

hermeto/core/package_managers/pip/main.py
slimreaper35: > With this we now have 2 modules referring to the same doc URL, one using a hardcoded URL, one a constant, can we put the constant to the __init__.py file and use it consistently since we apparently want to link to particular docs?

Hmm, having it in `__init__.py` would case circular import error, so I put the constant into `package_distributions.py` and import it from there to `main.py`.

> That said, I'm not such a big fan of linking docs in error messages, docs is pretty much outdated all the time and so while the idea sounds nice, the execution may end up being confusing and you need to actively maintain the error messages if the location of the doc changes. 

I see this pattern of linking docs is only used for gomod and pip so it might come from Cachito. If you want I can create a tracker issue for that at least.

eskultety: >  Hmm, having it in __init__.py would case circular import error, so I put the constant into package_distributions.py and import it from there to main.py.

Huh, forgot we use:
```
from hermeto.core.package_managers.pip.main import fetch_pip_source
```
in our backends.

> I see this pattern of linking docs is only used for gomod and pip so it might come from Cachito. If you want I can create a tracker issue for that at least.

Yeah, please do, let's get rid of that practice.

eskultety:
> > > @slimreaper35 Can we go one step further and extract DistributionPackageInfo (and friends) out of the main module as well?
> > 
> > 
> > In this pull request ?
> 
> Yes.

Thank you!

eskultety (approved):

PR #981, opened 2025-06-11
Use pydantic-settings to manage Hermeto app config

taylormadore:
This introduces pydantic-settings to manage our config settings. It attempts to do the following:

- Allow the use of environment variables to control Hermeto settings
- Allow the use of additional YAML configuration files at certain expected locations
- Maintain backwards compatibility with the existing CLI file input

There are a few workarounds used here for known issues with pydantic-settings that are targeted to be fixed in pydantic v3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,38 +122,26 @@ The permissive mode can currently suppress the following:
   information)
 - cargo manifest file `Cargo.toml` is out of sync with `Cargo.lock`
 
-### Available configuration parameters
+### Settings
 
-You can change Hermeto's configuration by specifying a configuration file while
-invoking any of the CLI commands
+Settings can be provided via the following sources (highest priority first):
 
-```shell
-hermeto --config-file config.yaml fetch-deps --source ./my-repo gomod
-```
+1. **Environment variables**: prefixed with `HERMETO_`, using `__` for nested
+   settings (e.g., `HERMETO_GOMOD__DOWNLOAD_MAX_TRIES=10`)
+2. **CLI option**: `--config-file path/to/config.yaml`
+3. **Config files** (automatically loaded if present): `~/.config/hermeto/config.yaml`,
+   `hermeto.yaml`, `.hermeto.yaml`
 
-Any parameter specified in this file will override the default values present in
-the [config.py][] module.
+Any settings specified will override the default values present in the
+[config.py][] module. The only supported format for config files is YAML.
 
-The only supported format for the config file is YAML.
-
-- `default_environment_variables` a dictionary where the keys are names of
-  package managers. The values are dictionaries where the keys are default
-  environment variables to set for that package manager and the values are the
-  environment variable values.
-- `gomod_download_max_tries` a maximum number of attempts for retrying go
-  commands.
-- `gomod_strict_vendor` (deprecated: *This option no longer has any effect
-  when set*) the bool to disable/enable the strict vendor mode. For a repo that
-  has gomod dependencies, if the `vendor` directory exists and this config
-  option is set to `True`, one of the vendoring flags must be used.
-- `goproxy_url` sets the value of the GOPROXY variable that Hermeto uses
-  internally when downloading Go modules. See [Go environment variables][].
-- `requests_timeout` a number (in seconds) for `requests.get()`'s 'timeout'
-  parameter, which sets an upper limit on how long `requests` can take to make a
-  connection and/or send a response. Larger numbers set longer timeouts.
-- `subprocess_timeout` a number (in seconds) to set a timeout for commands
-  executed by the `subprocess` module. Set a larger number to give the
-  subprocess execution more time.
+- `gomod.download_max_tries` max retry attempts for go commands.
+- `gomod.environment_variables` default environment variables for gomod.
+- `gomod.proxy_url` sets the GOPROXY variable that Hermeto uses internally when
+  downloading Go modules. See [Go environment variables][].
+- `http.timeout` timeout (seconds) for HTTP requests.
+- `runtime.concurrency_limit` max concurrent operations.
+- `runtime.subprocess_timeout` timeout (seconds) for subprocess commands.
 
 ## Package managers
 
--- a/hermeto/core/config.py
+++ b/hermeto/core/config.py
@@ -3,7 +3,7 @@ from pathlib import Path
 from typing import Any
 
 import yaml
-from pydantic import ValidationError, model_validator
+from pydantic import BaseModel, ValidationError, model_validator
 from pydantic_core import ErrorDetails
 from pydantic_settings import (
     BaseSettings,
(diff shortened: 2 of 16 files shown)


eskultety:
@taylormadore I think the simplest way of making the failing check pass would be to have a dedicated commit re-generating the requirements files before introducing a new dependency. The problem with that workflow is that it is run on any requirements file change which is usually done by dependabot (_Edit: IIUC GH actions don't allow you to filter based on source branches/actors, that's why this behavior_). I'm unclear on why it's failing like that at the moment. I do remember seeing resource failures back when I introduced the workflow, so I *think* it's related to the `GITHUB_TOKEN` permissions, but I'm not sure how to verify/debug this.

Luckily, we don't need to create another exception for pydantic-settings (or pydantic for that matter), since unlike pydantic/pydantic-core isn't pinned to a fixed version, for now that is...

hermeto/core/config.py
```diff
@@ -15,6 +15,11 @@
 from hermeto import APP_NAME
 from hermeto.core.errors import InvalidInput
 
+CONFIG_FILE_PATHS = [
+    f"~/.config/{APP_NAME.lower()}/config.yaml",
+    f"{APP_NAME.lower()}.yaml",
+    f".{APP_NAME.lower()}.yaml",
```
a-ovchinnikov, line 22: Do I understand this correctly and the last element in the list wins?

hermeto/core/config.py
```diff
@@ -27,7 +27,11 @@
 class Config(BaseSettings):
     """Singleton that provides default configuration for the application process."""
 
-    model_config = SettingsConfigDict(extra="forbid")
+    model_config = SettingsConfigDict(
+        case_sensitive=False,
+        env_prefix=f"{APP_NAME.upper()}_",
```
a-ovchinnikov, line 117: To verify my understanding again, does this mean that if I have FOO set to 0 in `~/.config/hermeto/config.yaml`, to 1 in `hermeto.yaml`, to 3 in `.hermeto.yaml` and to 4 via `HERMETO_FOO` environment variable I will get 4 in the code?

taylormadore: Yes, exactly :+1:

hermeto/core/config.py
taylormadore: Yes, that is correct :+1:

eskultety: Maybe also add a commentary that the order of predecence in the list is ascending? :)

hermeto/core/config.py
```diff
@@ -27,7 +27,11 @@
 class Config(BaseSettings):
     """Singleton that provides default configuration for the application process."""
 
-    model_config = SettingsConfigDict(extra="forbid")
+    model_config = SettingsConfigDict(
```
eskultety, line 112: nitpick: maybe just add a note to the commit message that the highest precedence of env variables is done by pydantic natively because it's not obvious from the code in any way :) .

README.md
```diff
@@ -122,38 +122,26 @@ The permissive mode can currently suppress the following:
...
-  parameter, which sets an upper limit on how long `requests` can take to make a
-  connection and/or send a response. Larger numbers set longer timeouts.
-- `subprocess_timeout` a number (in seconds) to set a timeout for commands
-  executed by the `subprocess` module. Set a larger number to give the
-  subprocess execution more time.
+### Settings
+
+Settings can be provided via the following sources (highest priority first):
+
+1. **Environment variables**: prefixed with `HERMETO_`, using `__` for nested
```
eskultety, line 129: why do we want the double score? I'm not sure I've encountered something like that anywhere when working on various ecosystems for which we ended up creating a backend.

hermeto/core/config.py
```diff
@@ -23,40 +23,131 @@
...
+        - Migrate legacy flat fields to new namespaced structure
+        """
+        if not isinstance(data, dict):
+            return data
+
+        _remove_gomod_strict_vendor(data)
+
+        for old_key, (namespace, new_key) in _FLAT_FIELD_MIGRATIONS:
+            if old_key in data:
+                _migrate_flat_to_namespace(data, old_key, data.pop(old_key), namespace, new_key)
```
eskultety, line 144: Maybe add a `FIXME` commentary with a note to drop  this `flat_to_namespace` logic with Hermeto 1.0

hermeto/core/config.py
taylormadore: The priority is configured here: https://github.com/taylormadore/hermeto/blob/95c96f408bacc43c05a65391a933d3c904cff12a/hermeto/core/config.py#L156-L167

README.md
taylormadore: I was just trying to avoid any collisions/ambiguity. If we had YARN and YARN_V1 namespaces, is V1 a setting under YARN or referring to the YARN_V1 namespace?

hermeto/core/config.py
taylormadore: Added

hermeto/core/config.py
taylormadore: Added

taylormadore:
Rebased on main

hermeto/core/config.py
eskultety: Didn't do my due diligence, sorry.

README.md
eskultety: Isn't this more of a documentation issue rather than an implementation thing? Also, wouldn't that specific case more likely become `YARN_CLASSIC`?

hermeto/core/config.py
```diff
@@ -73,6 +77,7 @@ def settings_customise_sources(
         """
         return (
             init_settings,
+            env_settings,
```
eskultety, line 178: Why does `init_settings` have higher priority? What use case did you have in mind where we'd not want to respect user settings?

eskultety, line 178: Why does `init_settings` have higher priority? What use case did you have in mind where we'd not want to respect user settings?

taylormadore: Potential convenience for testing. It otherwise does no harm

README.md
taylormadore: Having the delimiter in any of the nested field names can be problematic, which the double underscore will help avoid. See https://github.com/pydantic/pydantic-settings/issues/51#issuecomment-1801837619

```
from pydantic import BaseModel
from pydantic_settings import BaseSettings, SettingsConfigDict
import os

class Nested(BaseModel, extra="forbid"):
    field_name: str = "bar"


class Settings(BaseSettings):
    model_config = SettingsConfigDict(
        env_prefix="HERMETO_",
        env_nested_delimiter="_",
        extra="forbid",
    )
    nested_model: Nested = Nested()


os.environ["HERMETO_NESTED_MODEL_FIELD_NAME"] = "foo"
s = Settings()
```
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Settings
nested_model.field
  Extra inputs are not permitted [type=extra_forbidden, input_value={'name': 'foo'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/extra_forbidden
```

This conversation helped me realize I missed excluding extra fields from the nested models. Will fix. :sweat_smile:

hermeto/core/config.py
eskultety: Since the docs contains an inverted example, please add a commentary to keep this order, it wasn't obvious at first glance.

README.md
eskultety: Oh, that invalidates my argument completely. Again, please either add a code commentary linking the GH issue or even mention it in the commit message, either works for me -> resolving.

eskultety (approved):
With the reason for using double underscore for nesting as a technological limitation documented -> ACK

README.md
taylormadore: I'm glad you had asked, because I had completely forgotten the main reason why this was needed since the summer :smile:

taylormadore: I think this captures it:

The double underscore is pydantic-settings' convention for nested config structures. Single underscores can't be used as delimiters since they appear in field names (e.g., `concurrency_limit`). With `__` as the delimiter, `HERMETO_RUNTIME__CONCURRENCY_LIMIT=5` maps to `runtime.concurrency_limit = 5`.

See: https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values

PR #982, opened 2025-06-11
Allow easier installation of Hermeto on all operating systems

slimreaper35:


hermeto/core/utils.py
```diff
@@ -136,7 +136,8 @@ def _fast_copy(src: Path, dest: Path, *, follow_symlinks: bool = True) -> int:
             raise _FastCopyFailedFallback()
 
         try:
-            while nbytes := os.copy_file_range(srcfd, destfd, count=_get_blocksize(srcfd)):
+            # mypy: `os` module has no attribute "copy_file_range" on some platforms
```
eskultety, line 139: > The fast copy function relies on the Linux-specific os.copy_file_range
> syscall. It should not be used on other platforms as it would fail. Now
> we explicitly raise an exception on non-Linux systems to trigger the
> fallback to regular copy methods.

The commit msg needs an update as it currently doesn't match the  actual changes proposed.

pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = [
   "tenacity",
   "tomlkit",
   "typer",
-  "createrepo-c",
+  "createrepo-c ; sys_platform == 'linux'",
```
eskultety: `platform_system == 'Linux'` is also an alternative, but no only it results in generating `sys_platform` markers in the files, `sys_platform == 'linux'` as a string looks to be stable based on its description in the docs [1]
[1] https://packaging.pypa.io/en/stable/markers.html#packaging.markers.Environment

requirements-extras.txt
```diff
@@ -110,7 +110,7 @@ argcomplete==3.6.2 \
     --hash=sha256:65b3133a29ad53fb42c48cf5114752c7ab66c1c38544fdf6460f450c09b42591 \
     --hash=sha256:d0519b1bc867f5f4f4713c41ad0aba73a4a5f007449716b16f385f2166dc6adf
     # via nox
-async-timeout==5.0.1 \
+async-timeout==5.0.1 ; python_full_version < '3.11' \
```
eskultety, line 113: This is...interesting, I guess it's completely harmless.

requirements-extras.txt
```diff
@@ -360,7 +360,7 @@ coverage==7.8.2 \
     --hash=sha256:ef2f22795a7aca99fc3c84393a55a53dd18ab8c93fb431004e4d8f0774150f54 \
     --hash=sha256:fd51355ab8a372d89fb0e6a31719e825cf8df8b6724bee942fb5b92c3f016ba3
     # via pytest-cov
-createrepo-c==1.2.1 \
+createrepo-c==1.2.1 ; sys_platform == 'linux' \
```
eskultety, line 360: I'd like to see `createrepo-c` to be extracted from the core set of deps into its own extra in the future as it's been giving us a hard time for a while, something like `pip install .[rpm]` and keep it isolated in its own requirements file, and keep markers only in that file. This is because `nox` tries to install `requirements.txt` for all sessions, even unit tests which don't need `createrepo_c` at all, it's a runtime dependency only. That said, it's been broken since we added this so this PR doesn't make anything worse in that regard, instead, it improves **some** of the aspects.

.github/workflows/dependabot-pipcompile.yml
```diff
@@ -47,13 +47,15 @@ jobs:
             --output-file=requirements.txt \
             --python=3.9 \
             --refresh \
+            --no-strip-markers \
```
eskultety, line 50: s/no-stip-markers/no-skip-markers

eskultety (approved):
not a fan of the additional environment markers, but what can you do. We can later revert the markers back and improve upon that by tracking `createrepo_c` outside of `requirements.txt`

pyproject.toml
slimreaper35: Added this link -> https://docs.python.org/3/library/sys.html#sys.platform in the commit message.

pyproject.toml
```diff
@@ -34,30 +34,26 @@ dependencies = [
   "tenacity",
   "tomlkit",
   "typer",
-  "createrepo-c ; sys_platform == 'linux'",
-  "packageurl-python"
+  "typing-extensions",
```
eskultety: This is a new commit, right? I don't remember seeing it before. Anyway,

> typing-extensions is a direct dependency (not optional)

No it's not, `typing-extensions` is only needed for `mypy`. Additionally, direct vs optional is not a good a way of thinking about dependencies, it's runtime | build | development/test. `typing-extensions` falls in to the development category and so it should stay where it was.

slimreaper35: Interesting. But, it remains inside `requirements.txt` since other libraries use it as _runtime_ dependency (e.g. https://github.com/pydantic/pydantic/blob/main/pyproject.toml#L47)

eskultety: Well, that's not our issue to address, one can only do as much :). The point is, the way we currently consume `typing-extensions` is strictly mypy-related and hence has no place in runtime dependencies. Fancy a bug report on pydantic? :)

eskultety: One thing to keep in mind is that the absolute vast majority of projects don't lock their complete dependency list and so the issue you describe only gets visible and emphasized once you do that.

PR #987, opened 2025-06-23
Handle missing copy_file_range attribute in os module

slimreaper35:
The `_fast_copy` function uses `os.copy_file_range` for efficient copying. However, this function is not available on all platforms such as macOS.

Previously, an uncaught AttributeError would crash the process when the function was missing. By catching AttributeError explicitly and raising _FastCopyFailedFallback, we trigger the existing fallback mechanism for unsupported platforms. This maintains cross-platform compatibility while preserving fast copy operations where supported.

Should be part of 14e1491.

### DIFF
--- a/hermeto/core/utils.py
+++ b/hermeto/core/utils.py
@@ -136,9 +136,12 @@ def _fast_copy(src: Path, dest: Path, *, follow_symlinks: bool = True) -> int:
             raise _FastCopyFailedFallback()
 
         try:
-            # mypy: `os` module has no attribute "copy_file_range" on some platforms
             while nbytes := os.copy_file_range(srcfd, destfd, count=_get_blocksize(srcfd)):  # type: ignore
                 total += nbytes
+
+        # `os` module deos not have attribute `copy_file_range` on some platforms (see type ignore above)
+        except AttributeError:
+            raise _FastCopyFailedFallback()
 
         except OSError as ex:
             # ...in oder to have a more informative exception.


a-ovchinnikov (approved):

hermeto/core/utils.py
```diff
@@ -140,6 +140,9 @@ def _fast_copy(src: Path, dest: Path, *, follow_symlinks: bool = True) -> int:
             while nbytes := os.copy_file_range(srcfd, destfd, count=_get_blocksize(srcfd)):  # type: ignore
                 total += nbytes
 
+        except AttributeError:
+            raise _FastCopyFailedFallback("os.copy_file_range is not available on this platform")
```
eskultety: The error string isn't being utilized anywhere. That said, it might be worth having that context everywhere (consistency), however, in such case this PR should be paired with one additional commit that would provide some context error messages to the rest of the fallback cases and report them at the single spot where the fallback exception is caught, it's a simple change.

eskultety: Looking at the code base, one way to go about this could be defining a `__str__` method on `_FastCopyFailedFallback`:
```
def __str__(self):
        if self.__cause__:
            # provide a more informative message
            return self.__cause__
        
        return super().__str__()
```
and enforce the `raise from` pattern which we inconsistently use.

slimreaper35: Isn't one debug log enough here? It's not a user-facing functionality.

eskultety: What do you mean by debug log? There is exactly one debug log relevant here, but what you currently have isn't going to be reported anywhere and so I'm curious why the error string was added in the first place if it's not output anywhere, `_FastCopyFailedFallback` is essentially a sink. Now, that is not to say that you shouldn't report the message like you do, hence my reflection on the current code. You're right this is not a user-facing functionality, neither are debug logs and seeing the exact cause of the fallback turn out to be useful for developers in the future, that's all I'm saying.

slimreaper35: Okay, so if the error message is not captured anywhere, I can just delete it.

eskultety: Technically, you can, but let me yet again reiterate the potential usefulness of it :) as I tried to communicate.

eskultety: I may have prepared something on my local branch a few days ago, so don't bother, this is already good enough, let's merge this as is :) .

eskultety (approved):

PR #988, opened 2025-06-23
Simplify cargo purls

slimreaper35:

## Summary by CodeRabbit

* **Refactor**
  * Improved handling of Cargo packages, distinguishing between local/workspace and external dependencies.
  * Enhanced support for virtual workspaces and local path-based dependencies.
  * Refined SBOM generation and package URL formatting for better accuracy and clarity.

* **Bug Fixes**
  * Adjusted package URL (purl) generation to use updated query parameters and formats, improving metadata consistency.

* **Tests**
  * Added unit tests for main package resolution and package URL generation for Cargo packages.

### DIFF
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -7,11 +7,10 @@ from dataclasses import dataclass
 from functools import cached_property
 from itertools import chain
 from pathlib import Path
-from typing import Optional
+from typing import Any, Optional
 
 import tomlkit
 from packageurl import PackageURL
-from tomlkit.toml_file import TOMLFile
 
 from hermeto.core.errors import NotAGitRepo, PackageRejected
 from hermeto.core.models.input import Mode, Request
@@ -45,7 +44,7 @@ class CargoPackage:
     """CargoPackage."""
 
     name: str
-    version: str
+    version: Optional[str] = None
     source: Optional[str] = None  # [git|registry]+https://github.com/<org>/<package>#[|<sha>]
     checksum: Optional[str] = None
     dependencies: Optional[list] = None
@@ -87,35 +86,29 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
     return RequestOutput.from_obj_list(components, environment_variables, project_files)
 
 
-def _extract_package_info(path_to_toml: Path) -> dict:
-    # 'value' unwraps the underlying dict and that makes mypy happy (it complains about
-    # mismatching type otherwise despite parsed document having the necessary interface).
-    parsed_toml = tomlkit.parse(path_to_toml.read_text())
-    package_info = parsed_toml.value["package"]
-    if isinstance(package_info, dict) and package_info.get("version") == {"workspace": True}:
+def _parse_toml_project_file(path: Path) -> dict[str, Any]:
+    """Parse any Cargo related TOML file into a dictionary."""
+    parsed_toml = tomlkit.parse(path.read_text())
+    return parsed_toml.value
+
+
+def _resolve_main_package(package_dir: RootedPath) -> tuple[str, Optional[str]]:
+    """Resolve package name and version from Cargo.toml."""
+    parsed_toml = _parse_toml_project_file(package_dir.path / "Cargo.toml")
+    package_info = parsed_toml.get("package", {})
+
+    # use default values if the project is a virtual workspace without any package information
+    name = package_info.get("name", package_dir.path.stem)
+    version = package_info.get("version", None)
+
+    if package_info.get("version") == {"workspace": True}:
         # support for workspace package table https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
         # real world example: https://github.com/pyca/cryptography/blob/43.0.0/src/rust/Cargo.toml
         # we condition this to when package_info is a dict because this same function has usages where package_info is a list
         # (Cargo.lock) and this info is only available when it is a dict (Cargo.toml)
-        package_info["version"] = parsed_toml.value["workspace"]["package"]["version"]
-    return package_info
+        version = parsed_toml["workspace"]["package"]["version"]
 
-
-def _resolve_main_package(package_dir: RootedPath) -> dict:
-    try:
-        return _extract_package_info(package_dir.path / "Cargo.toml")
-    # We'll get here in the case of virtual workspaces. A real-world example is
-    # https://github.com/rwf2/Rocket/tree/master
-    # In this case there is no package name per se, but there still is a metapackage.
-    # We'll use directory name as a fallback in this case.
-    # Version won't make much sense here: this is a meta-package, the state of a
-    # repository will be captured in VCS_URL, and individual components will be
-    # versioned.
-    except KeyError:
-        return {
-            "name": package_dir.path.stem,
-            "version": None,
-        }
+    return name, version
 
 
 def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
@@ -222,9 +215,11 @@ def _resolve_cargo_package(
             cmd=cmd, params={"cwd": package_dir}, package_dir=package_dir, mode=request.mode
         )
 
-    packages = _extract_package_info(package_dir.path / "Cargo.lock")
-    main_package = _resolve_main_package(package_dir)
-    is_a_dep = lambda p: p["name"] != main_package["name"]
+    parsed_lockfile = _parse_toml_project_file(package_dir.path / "Cargo.lock")
+    packages = parsed_lockfile.get("package", [])
+
+    main_package_name, main_package_version = _resolve_main_package(package_dir)
+    is_a_dep = lambda p: p["name"] != main_package_name
     try:
         vcs_url = get_repo_id(package_dir.root).as_vcs_url_qualifier()
     except NotAGitRepo:
@@ -234,7 +229,7 @@ def _resolve_cargo_package(
         CargoPackage(**p, vcs_url=vcs_url).to_component() for p in packages if is_a_dep(p)
     )
     main_component = CargoPackage(
-        name=main_package["name"], version=main_package["version"], vcs_url=vcs_url
+        name=main_package_name, version=main_package_version, vcs_url=vcs_url
     ).to_component()
 
     components = chain((main_component,), deps_components)
@@ -269,10 +264,8 @@ def _use_vendored_sources(package_dir: RootedPath, config_template: dict) -> Pro
     # Python extension it is better to check if there is an old-style config present and
     # process it if found.
     cfn = ".cargo/config" if _old_style_config_is_present_in(package_dir) else ".cargo/config.toml"
-    cargo_config = package_dir.join_within_root(cfn)
+    config_path = package_dir.join_within_root(cfn).path
 
-    toml_file = TOMLFile(cargo_config)
-    original_content = toml_file.read() if cargo_config.path.exists() else {}
+    original_content = _parse_toml_project_file(config_path) if config_path.exists() else {}
     original_content.update(config_template)
-
-    return ProjectFile(abspath=cargo_config.path, template=tomlkit.dumps(config_template))
+    return ProjectFile(abspath=config_path, template=tomlkit.dumps(config_template))
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -82,7 +82,7 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
         # https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure
         config_template = _fetch_dependencies(package_dir, request)
         project_files.append(_use_vendored_sources(package_dir, config_template))
-        components.extend(_resolve_cargo_package(package_dir))
+        components.extend(_generate_sbom_components(package_dir))
 
     return RequestOutput.from_obj_list(components, environment_variables, project_files)
 
@@ -220,8 +220,8 @@ def _run_cmd_watching_out_for_lock_mismatch(
             raise
 
 
-def _resolve_cargo_package(package_dir: RootedPath) -> chain[Component]:
-    """Resolve a single cargo package."""
+def _generate_sbom_components(package_dir: RootedPath) -> chain[Component]:
+    """Generate SBOM components from Cargo.lock and for the main package."""
     parsed_lockfile = _parse_toml_project_file(package_dir.path / "Cargo.lock")
     packages = parsed_lockfile.get("package", [])
 
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -88,13 +88,19 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
 
 
 def _fetch_dependencies(package_dir: RootedPath, request: Request) -> dict[str, Any]:
+    """Fetch cargo dependencies and return a config template for hermetic build."""
     vendor_dir = request.output_dir.join_within_root("deps/cargo")
-    # --no-delete to keep everything already present. It does not matter for a fresh
-    # single package, but it does matter when there is pip interaction.
+    # --locked           Assert that `Cargo.lock` will remain unchanged.
+    # --versioned-dirs   Always include version in subdir name.
+    # --no-delete        Don't delete older crates in the vendor directory.
+    #                    It is necessary to make Cargo keep dependencies that are already
+    #                    present in the vendored directory. This flag has no effect on standalone
+    #                    cargo operations however is crucial when it is invoked from pip.
     cmd = ["cargo", "vendor", "--locked", "--versioned-dirs", "--no-delete", str(vendor_dir)]
     log.info("Fetching cargo dependencies at %s", package_dir)
     with _hidden_cargo_config_file(package_dir):
-        # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
+        # The necessary configuration to use the vendored sources will be printed to STDOUT.
+        # https://doc.rust-lang.org/cargo/commands/cargo-vendor.html#description
         config_template = _run_cmd_watching_out_for_lock_mismatch(
             cmd=cmd, params={"cwd": package_dir}, package_dir=package_dir, mode=request.mode
         )
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -77,13 +77,29 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
 
     for package in request.cargo_packages:
         package_dir = request.source_dir.join_within_root(package.path)
+        _verify_lockfile_is_present_or_fail(package_dir)
         # cargo allows to specify configuration per-package
         # https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure
-        fetched_components, cfg_template = _resolve_cargo_package(package_dir, request)
-        components.extend(fetched_components)
-        project_files.append(_use_vendored_sources(package_dir, cfg_template))
+        config_template = _fetch_dependencies(package_dir, request)
+        project_files.append(_use_vendored_sources(package_dir, config_template))
+        components.extend(_resolve_cargo_package(package_dir))
 
     return RequestOutput.from_obj_list(components, environment_variables, project_files)
+
+
+def _fetch_dependencies(package_dir: RootedPath, request: Request) -> dict[str, Any]:
+    vendor_dir = request.output_dir.join_within_root("deps/cargo")
+    # --no-delete to keep everything already present. It does not matter for a fresh
+    # single package, but it does matter when there is pip interaction.
+    cmd = ["cargo", "vendor", "--locked", "--versioned-dirs", "--no-delete", str(vendor_dir)]
+    log.info("Fetching cargo dependencies at %s", package_dir)
+    with _hidden_cargo_config_file(package_dir):
+        # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
+        config_template = _run_cmd_watching_out_for_lock_mismatch(
+            cmd=cmd, params={"cwd": package_dir}, package_dir=package_dir, mode=request.mode
+        )
+
+    return _swap_sources_directory_for_subsitution_slot(config_template)
 
 
 def _parse_toml_project_file(path: Path) -> dict[str, Any]:
@@ -198,23 +214,8 @@ def _run_cmd_watching_out_for_lock_mismatch(
             raise
 
 
-def _resolve_cargo_package(
-    package_dir: RootedPath,
-    request: Request,
-) -> tuple[chain[Component], dict]:
+def _resolve_cargo_package(package_dir: RootedPath) -> chain[Component]:
     """Resolve a single cargo package."""
-    _verify_lockfile_is_present_or_fail(package_dir)
-    vendor_dir = request.output_dir.join_within_root("deps/cargo")
-    # --no-delete to keep everything already present. It does not matter for a fresh
-    # single package, but it does matter when there is pip interaction.
-    cmd = ["cargo", "vendor", "--locked", "--versioned-dirs", "--no-delete", str(vendor_dir)]
-    log.info("Fetching cargo dependencies at %s", package_dir)
-    with _hidden_cargo_config_file(package_dir):
-        # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
-        config_template = _run_cmd_watching_out_for_lock_mismatch(
-            cmd=cmd, params={"cwd": package_dir}, package_dir=package_dir, mode=request.mode
-        )
-
     parsed_lockfile = _parse_toml_project_file(package_dir.path / "Cargo.lock")
     packages = parsed_lockfile.get("package", [])
 
@@ -234,7 +235,7 @@ def _resolve_cargo_package(
 
     components = chain((main_component,), deps_components)
 
-    return components, _swap_sources_directory_for_subsitution_slot(config_template)
+    return components
 
 
 def _swap_sources_directory_for_subsitution_slot(template: str) -> dict:
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -5,9 +5,9 @@ from collections.abc import Generator
 from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import cached_property
-from itertools import chain
 from pathlib import Path
 from typing import Any, Optional
+from urllib.parse import urlparse, urlunparse
 
 import tomlkit
 from packageurl import PackageURL
@@ -41,31 +41,64 @@ class PackageWithCorruptLockfileRejected(PackageRejected):
 
 @dataclass(frozen=True)
 class CargoPackage:
-    """CargoPackage."""
+    """Represents a package from Cargo.lock file."""
 
     name: str
-    version: Optional[str] = None
+    version: str
     source: Optional[str] = None  # [git|registry]+https://github.com/<org>/<package>#[|<sha>]
     checksum: Optional[str] = None
-    dependencies: Optional[list] = None
-    vcs_url: Optional[str] = None
 
     @cached_property
     def purl(self) -> PackageURL:
-        """Return corrsponding purl."""
+        """Return corresponding package URL."""
         qualifiers = {}
-        if self.source is not None:
-            qualifiers["source"] = self.source
-        # The condition below holds for either the main package or any packages
-        # that originate from the filesystem (for example, workspace and patched source ones).
-        if self.vcs_url is not None and self.source is None:
-            qualifiers["vcs_url"] = self.vcs_url
+        # depends on https://github.com/hermetoproject/hermeto/issues/852
         if self.checksum is not None:
             qualifiers["checksum"] = self.checksum
+
+        if self.source is not None and self.source.startswith("git+"):
+            parsed_url = urlparse(self.source)
+            commit_id = parsed_url.fragment
+            base_url = urlunparse(parsed_url._replace(query="", fragment=""))
+            qualifiers["vcs_url"] = f"{base_url}@{commit_id}"
+
         return PackageURL(type="cargo", name=self.name, version=self.version, qualifiers=qualifiers)
 
     def to_component(self) -> Component:
         """Convert CargoPackage into SBOM component."""
+        return Component(name=self.name, version=self.version, purl=self.purl.to_string())
+
+
+@dataclass
+class LocalCargoPackage:
+    """Represents a local dependency in the project or a workspace."""
+
+    name: str
+    version: Optional[str] = None
+    vcs_url: Optional[str] = None
+    subpath: Optional[str] = None
+
+    @cached_property
+    def purl(self) -> PackageURL:
+        """Return corresponding package URL."""
+        qualifiers = {}
+        if self.vcs_url is not None:
+            qualifiers["vcs_url"] = self.vcs_url
+        else:
+            # The subpath does not make sense if there is no VCS URL. This usually happens because
+            # of missing .git directory in an unpacked tarball that comes from a pip request.
+            self.subpath = None
+
+        return PackageURL(
+            type="cargo",
+            name=self.name,
+            version=self.version,
+            qualifiers=qualifiers,
+            subpath=self.subpath,
+        )
+
+    def to_component(self) -> Component:
+        """Convert LocalCargoPackage into SBOM component."""
         return Component(name=self.name, version=self.version, purl=self.purl.to_string())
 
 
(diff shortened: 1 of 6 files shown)


hermeto/core/package_managers/cargo/main.py
```diff
@@ -45,7 +44,7 @@ class CargoPackage:
     """CargoPackage."""
 
     name: str
-    version: str
+    version: Optional[str] = None
```
a-ovchinnikov: I am sorry, I believe this change is incorrect: version cannot be `None`. If this is done to silence mypy (again) then a comment must be present in the code at the very least.

Looking at the change further I have discovered why `None` could end up being supplied as a version: this could happen and is expected when the PM is dealing with a virtual workspace, this was stated in a comment which got deleted in this commit as an unnecessary. I think comments explaining motivation behind certain pieces of code should not be dropped.

Since this is wrong typing a better solution would be to provide an empty string and not `None` _and_ to keep a comment explaining why a version could become empty.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -87,35 +86,28 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
-        return {
-            "name": package_dir.path.stem,
-            "version": None,
-        }
+def _resolve_main_package(package_dir: RootedPath) -> tuple[str, Optional[str]]:
+    """Resolve package name and version from Cargo.toml."""
+    parsed_toml = _parse_toml_project_file(package_dir.path / "Cargo.toml")
+
+    package_info: dict[str, Any] = parsed_toml.get("package", {})
+    workspace_info: dict[str, Any] = parsed_toml.get("workspace", {})
```
a-ovchinnikov: How is annotation `dict[str, Any]` better than just `dict` and what new information does it tell a reader? (A dictionary is the default clause of a `get` which means that we will be getting a dictionary anyway). I would strongly recommend dropping this.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -87,35 +86,28 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
-    # We'll use directory name as a fallback in this case.
-    # Version won't make much sense here: this is a meta-package, the state of a
-    # repository will be captured in VCS_URL, and individual components will be
-    # versioned.
-    except KeyError:
-        return {
-            "name": package_dir.path.stem,
-            "version": None,
-        }
+def _resolve_main_package(package_dir: RootedPath) -> tuple[str, Optional[str]]:
```
a-ovchinnikov, line 150: Let's make it a strict `str` and fall back to `""` while we are here.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -254,7 +248,7 @@ def _old_style_config_is_present_in(package_dir: RootedPath) -> bool:
     return (package_dir.path / ".cargo/config").exists()
 
 
-def _use_vendored_sources(package_dir: RootedPath, config_template: dict) -> ProjectFile:
+def _use_vendored_sources(package_dir: RootedPath, config_template: dict[str, Any]) -> ProjectFile:
```
a-ovchinnikov: What insight does this new annotation provide into the nature of `config_template`?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -77,15 +77,35 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
 
     return RequestOutput.from_obj_list(components, environment_variables, project_files)
 
 
+def _fetch_dependencies(package_dir: RootedPath, request: Request) -> dict[str, Any]:
+    """Fetch cargo dependencies and return a config template for hermetic build."""
+    vendor_dir = request.output_dir.join_within_root("deps/cargo")
+    # --locked           Assert that `Cargo.lock` will remain unchanged
+    # --versioned-dirs   Always include version in subdir name
+    # --no-delete        Don't delete older crates in the vendor directory
```
a-ovchinnikov: Why wouldn't we want to delete old crates in the vendor directory?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -197,23 +217,8 @@ def _run_cmd_watching_out_for_lock_mismatch(
             raise
 
 
-def _resolve_cargo_package(
-    package_dir: RootedPath,
-    request: Request,
-) -> tuple[chain[Component], dict]:
+def _resolve_cargo_package(package_dir: RootedPath) -> chain[Component]:
```
a-ovchinnikov: I don't think this should be called `_resolve_cargo_package` anymore. What it does now is more like `_generate_components_for_package`.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -41,27 +42,34 @@ def __init__(self, package_path: str) -> None:
...
+    Represents a package from Cargo.lock file.
+
+    The [[package]] table in the Cargo.lock file may contain many combinations of the following
+    fields (the name and version are mandatory). Even though some fields are not used, it is
+    convenient to simply unpack the parsed dictionary into this dataclass.
+    """
 
     name: str
-    version: Optional[str] = None
-    source: Optional[str] = None  # [git|registry]+https://github.com/<org>/<package>#[|<sha>]
```
a-ovchinnikov, line 49: Why do you think that a local example of what is supposed to go into a field is excessive?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -229,12 +237,20 @@ def _resolve_cargo_package(package_dir: RootedPath) -> chain[Component]:
...
-    deps_components = (
-        CargoPackage(**p, vcs_url=vcs_url).to_component() for p in packages if is_a_dep(p)
+
+    deps_components = (CargoPackage(**p).to_component() for p in packages if is_a_dep(p))
+
+    main_purl = PackageURL(
+        type="cargo",
+        name=main_package_name,
+        version=main_package_version,
+        qualifiers={"vcs_url": vcs_url} if vcs_url else None,
```
a-ovchinnikov: PEP8 nitpick alert. The recommended way of checking for `None` is with `is`.

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -0,0 +1,42 @@
...
+                "name": "foo",
+                "version": "0.1.0",
+                "source": "git+https://github.com/rust-random/rand?rev=abc123#abc123",
+            },
+            "pkg:cargo/foo@0.1.0?vcs_url=git%2Bhttps://github.com/rust-random/rand%40abc123",
+            id="package_with_git_source",
+        ),
+    ],
+)
+def test_purls(pkg: dict[str, Any], expected_purl: str) -> None:
```
a-ovchinnikov: Could you please be more specific about what is being tested? Alternatively this looks like a good candidate for becoming a doctest in the class itself:

```python
"""
    Represents a package from Cargo.lock file
    <the rest of the docstring>

    >>> CargoPackage(**{"name": "foo", "version": "0.1.0"}).purl.to_string()
    'pkg:cargo/foo@0.1.0'
    >>> CargoPackage(
    ...     **{"name": "foo", "version": "0.1.0",
    ...        "source": "registry+https://github.com/rust-lang/crates.io-index",
    ...        "checksum": "abc123"}).purl.to_string()
    'pkg:cargo/foo@0.1.0?checksum=abc123'
"""
```

hermeto/core/package_managers/cargo/main.py
slimreaper35: > this could happen and is expected when the PM is dealing with a virtual workspace,

Yes, that's right. Later in the next patch, I changed it back to `str` and moved the logic for handling "main package" out of this class.

(I should have explained this better in the commit message.)

hermeto/core/package_managers/cargo/main.py
slimreaper35: These comments just briefly document the options we use for `cargo vendor`. Actually, they were copied from the command help message. I can drop them if desired.

hermeto/core/package_managers/cargo/main.py
slimreaper35: Do you mean just `str, Any` part or the whole type annotation?

hermeto/core/package_managers/cargo/main.py
slimreaper35: But I regenerated SBOMs, and I see a new field that came up:

```diff
       "purl": "pkg:cargo/cryptography-44.0",
-      "type": "library"
+      "type": "library",
+      "version": 
```

which does not make much sense. The version is an optional field in the SBOM, so I would vote for the current state.

hermeto/core/package_managers/cargo/main.py
slimreaper35: just a more specific type annotation

hermeto/core/package_managers/cargo/main.py
slimreaper35: So it looks like, if you replace `None` with `""` it does not affect the actual PURL but it does affect the SBOM component. The JSON file would contain a version key with an empty value. I am not sure if that is correct.

slimreaper35:
new revision with more commits and with suggested improvements

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: The entire annotation. Th reason for it is that `dict` part is obvious from `get` default argument, an annotation would have been necessary if this `get` call were supposed to return a union type (which, in turn, is a questionable design decision), and I think `str, Any` refinement is at the very least useless or even slightly harmful: `Any` is  the broadest type possible, its use in annotations should be limited to some very generic code which would probably generate some other code or would need to be able to process literally any object (strings, numbers, file objects, functions, classes etc). A synonym to `Any` is `object`, which is everything a Python interpreter deals with. By telling a reader that this dictionary maps strings to objects you a) don't tell anything specific and also b) tell them that _actually_ you are OK with this dictionary having a class as a value, or maybe a thread. The last part is plainly not true. The fact that this annotation is widely used does not counter any of my statements above.

A correct approach to providing such annotations would be to have a special type for config section, either defined as a proper class or as an alias to `dict[str, str | Number | Boolean]` (note, that suggested refinement of values types is likely incomplete and should serve as a prototype, not the final definition), however I do not think this is necessary at all: it is pretty obvious from the code context that this is a config-like dictionary, and the fact that configs usually map strings to strings, numbers or booleans is widely known and understood. I think it is safe to rely on this industry-wide knowledge and just use good names for the variables to convey your intent (which you already do).

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: This one apparently boils down to the fact that `Component` allows version to be `None` (which TBH is semantically questionable and would have benefited from a comment explaining why so so much). I think that the best place to fix that would be [right there](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/models/sbom.py#L94) with something like 
```python
version = v if (v:=package.get("version")) is not None and v else None
```

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Please see my reasoning about `str, Any` refinements above.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -95,18 +95,18 @@ def _parse_toml_project_file(path: Path) -> dict[str, Any]:
...
-    if package_info.get("version") == {"workspace": True}:
-        # support for workspace package table https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
-        # real world example: https://github.com/pyca/cryptography/blob/43.0.0/src/rust/Cargo.toml
-        # we condition this to when package_info is a dict because this same function has usages where package_info is a list
-        # (Cargo.lock) and this info is only available when it is a dict (Cargo.toml)
-        version = parsed_toml["workspace"]["package"]["version"]
+    # check for a workspace package version
+    # https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
+    if version is None:
+        version = workspace_info.get("package", {}).get("version", None)
```
a-ovchinnikov: Minor nitpick: `None` is the default return value for `get`.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Would it be safe to drop `--no-delete` flag?

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Resolving this, but shouldn't it be done in the same commit?

tests/unit/package_managers/cargo/test_main.py
a-ovchinnikov: Thank you for renaming it, it reads better now! What do you think about converting this test into a doctest?

hermeto/core/package_managers/cargo/main.py
slimreaper35: Maybe this is a good candidate for a follow-up issue/pull request.
A similar implementation is in [bundler](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/bundler/main.py#L105) module.

hermeto/core/package_managers/cargo/main.py
slimreaper35: Well, you added the flag in 4503e60. But I guess it is not harmful to keep it.

hermeto/core/package_managers/cargo/main.py
slimreaper35: reverted

hermeto/core/package_managers/cargo/main.py
slimreaper35: fixed

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Thank you!

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Yes, filed https://github.com/hermetoproject/hermeto/issues/990 for it

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Sorry, this was a trick question: I know I added it, I also know why I added it. Moreover, I deemed this specific flag so important and non-obvious that I added a comment (admittedly not a very good one) explaining why this flag is needed. The comment describing the intent is getting removed in this commit and is replaced with some trivia instead. I suggest rewording my commit instead of dropping  it. Let's try something like this:
```python
# --no-delete is necessary to make Cargo keep dependencies that are already
# present in the vendored directory. This flag have no effect on standalone 
# Cargo PM operations however is crucial when it is invoked from Pip PM.
```

WDYT?

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov, line 1: Is it worth it to have a separate renaming commit right after a refactoring commit?

slimreaper35: After which commit ?

hermeto/core/package_managers/cargo/main.py
slimreaper35: Added

hermeto/core/package_managers/cargo/main.py
```diff
@@ -41,27 +42,34 @@ def __init__(self, package_path: str) -> None:
...
-        if self.vcs_url is not None and self.source is None:
-            qualifiers["vcs_url"] = self.vcs_url
+        # depends on https://github.com/hermetoproject/hermeto/issues/852
         if self.checksum is not None:
             qualifiers["checksum"] = self.checksum
+
+        if self.source is not None and self.source.startswith("git+"):
+            parsed_url = urlparse(self.source)
+            commit_id = parsed_url.fragment
+            base_url = urlunparse(parsed_url._replace(query="", fragment=""))
```
eskultety, line 62: Interesting, a private method that is publicly documented in a usage example https://docs.python.org/3/library/urllib.parse.html:

>  As is the case with all named tuples, the subclass has a few additional methods and attributes that are particularly useful. One such method is _replace(). The _replace() method will return a new ParseResult object replacing specified fields with new values.

I'll take that, this is neat.

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -0,0 +1,42 @@
+from typing import Any
```
eskultety, line 1: Uhm, did we really have no unit tests for cargo up until now? Am I reading it right? :)

hermeto/core/package_managers/cargo/main.py
```diff
@@ -41,27 +42,34 @@ def __init__(self, package_path: str) -> None:
...
+            qualifiers["vcs_url"] = f"{base_url}@{commit_id}"
```
eskultety, line 63: Do we not care about the subpath element for the PURL, basically what I'm asking is - the linked SPEC `vcs_url` section further links to https://github.com/spdx/spdx-spec/blob/cfa1b9d08903/chapters/3-package-information.md#37-package-download-location which specifies the vcs url as:
`<vcs_tool>+<transport>://<host_name>[/<path_to_repository>][@<revision_tag_or_branch>][#<sub_path>]`

so I'm just looking for an answer, do we want/need to encode the subpath in the purl?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -41,27 +42,34 @@ def __init__(self, package_path: str) -> None:
...
+    Represents a package from Cargo.lock file.
+
+    The [[package]] table in the Cargo.lock file may contain many combinations of the following
+    fields (the name and version are mandatory). Even though some fields are not used, it is
+    convenient to simply unpack the parsed dictionary into this dataclass.
+    """
 
     name: str
-    version: Optional[str] = None
+    version: str
```
eskultety: I have to admit I got lost here a bit - IIUC the main package can be the virtual workspace itself, correct? If so, then based on previous patches' context a workspace isn't required to specify a version, it only CAN do so for workspace members to inherit it, so what am I missing? What gives the confidence that `version` can no longer be None?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -7,11 +7,10 @@
 from functools import cached_property
 from itertools import chain
```
eskultety: > Improve parsing of Cargo.toml, Cargo.lock and .cargo/config.toml files
by separating their logic into a distinct function. This enhances
code clarity without unnecessary comments and removes duplication.
>

This should be 2 commits, split right here, to clean up the diff a bit and the version part that follows should be squashed into the next commit in the series where's it's contextually suited much more.

>Change type of `version` field from `str` to `Optional[str]` to silence
>mypy error. The class is used for "main package" component, which can be
>a virtual workspace without defined package name and version in
>Cargo.toml. The version is None in that case. In the future patches,
>I will revert this line back to `str` type and move the logic for "main
>package" component away from this class.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -95,18 +95,18 @@ def _parse_toml_project_file(path: Path) -> dict[str, Any]:
...
 
-    if package_info.get("version") == {"workspace": True}:
-        # support for workspace package table https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
-        # real world example: https://github.com/pyca/cryptography/blob/43.0.0/src/rust/Cargo.toml
-        # we condition this to when package_info is a dict because this same function has usages where package_info is a list
-        # (Cargo.lock) and this info is only available when it is a dict (Cargo.toml)
-        version = parsed_toml["workspace"]["package"]["version"]
+    # check for a workspace package version
+    # https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
+    if version is None:
```
eskultety, line 163: like I'm mentioning elsewhere I'm still unclear on where's the tipping point where we can assume it to be non-None again.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -198,23 +221,8 @@ def _run_cmd_watching_out_for_lock_mismatch(
...
-    # --no-delete to keep everything already present. It does not matter for a fresh
-    # single package, but it does matter when there is pip interaction.
-    cmd = ["cargo", "vendor", "--locked", "--versioned-dirs", "--no-delete", str(vendor_dir)]
-    log.info("Fetching cargo dependencies at %s", package_dir)
-    with _hidden_cargo_config_file(package_dir):
-        # stdout contains exact values to add to .cargo/config.toml for a build to become hermetic.
-        config_template = _run_cmd_watching_out_for_lock_mismatch(
-            cmd=cmd, params={"cwd": package_dir}, package_dir=package_dir, mode=request.mode
-        )
-
```
eskultety, line 224: There's some code movement along with changes going on. That's inevitable but if you want to update commentaries AND/OR function names, whatever, please do that in a standalone commit limiting the actual changes needed to the bare functional minimum. Not sure if GH is going to attach my screenshot or not, but basically when done right, most of this big hunk could actually be gray (dimmed-zebra) which means - uninteresting, aka no review time needed.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -221,8 +221,8 @@ def _run_cmd_watching_out_for_lock_mismatch(
             raise
 
 
-def _resolve_cargo_package(package_dir: RootedPath) -> chain[Component]:
-    """Resolve a single cargo package."""
+def _generate_sbom_components(package_dir: RootedPath) -> chain[Component]:
```
eskultety: Much better, yes, this is (overall) more consistent with the rest of the backends.

tests/unit/package_managers/cargo/test_main.py
a-ovchinnikov: Yes, that was a conscious decision. UTs do not make much sense for a PM, ITs and E2Es are better positioned to verify that a PM actually works (since it is a thin wrapper around a third party tool).

hermeto/core/package_managers/cargo/main.py
slimreaper35: Right now, the CargoPackage class is only used for items from Cargo.lock, where each item must have a version. The code above is used for checking the "main" package, IOW, by parsing top-level Cargo.toml.

Example of workspace package table - [cryptography](https://github.com/pyca/cryptography/blob/main/Cargo.toml#L15) where nested packages can inherit the version.

Also, you can have a minimal top-level Cargo.toml file just listing all workspaces in the project:

```
[workspace]
members = [
  "a",
  "b",
  "c"
]
```

In that case version is None for the "main" package.

slimreaper35: I tried to find a scheme for Cargo.lock, but I was unsuccessful.

hermeto/core/package_managers/cargo/main.py
chmeliik, line 63: Does Hermeto's Cargo implementation support only git dependencies and dependencies from the default crates.io registry? What about [dependencies from other registries](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-other-registries) and [local path dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies)? And workspaces, if that's a thing?

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Is it worth it to modify a function in one commit and then rename it in a commit right after that one? I'd say squashing them is a better approach (I refer to https://github.com/hermetoproject/hermeto/pull/988/commits/697f83425d02a13a3f95e6b5ad1836a3d3ee8d78 and  https://github.com/hermetoproject/hermeto/pull/988/commits/06360ffb99596cfeb176983432dbd4818bdb5586, sorry I did not realize that GitHub would just apply the same comment to the entire chain).

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: I believe the minimal example @slimreaper35 has provided was the reason for introducing `None` as a possible version.

hermeto/core/package_managers/cargo/main.py
slimreaper35: https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field

According to the docs, if there is no version field, the default is `0.0.0`, which will be recorded then in `Cargo.lock`.

tests/unit/package_managers/cargo/test_main.py
eskultety: I'm just surprised we didn't have any other helpers worth testing in isolation.

hermeto/core/package_managers/cargo/main.py
eskultety: We currently don't have it captured in ITs (we should though), but IIUC we do! Here's a test repo that should be migrated over to this org and be used as an e2e test (and indeed is used during development): https://github.com/cachito-testing/cachi2-rust

It doesn't cover alternative registries, but given that it's just a flag in the TOML spec and we don't stand in cargo's way I dare to say it shouldn't pose an issue. That said, I don't think this is captured in the SBOM very clearly, then again, we'd have come up with some custom qualifiers I reckon (`source` being one of them).

chmeliik: Alternative registries should use `repository_url` I think

hermeto/core/package_managers/cargo/main.py
eskultety: I guess my original question wasn't 100% clear based on the limited context I had, so now having read https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field provided by @slimreaper35 my question is - in cases with such barbones Cargo.toml do we find it okay to not report a version for the "main" (aka workspace) package in the SBOM instead of forcing it to the `0.0.0` string literal and link this decision to the official docs? I tried to simulate this with https://github.com/cachito-testing/cachi2-rust# and yes, the version gets enforced as 0.0.0 in `Cargo.lock`, but 
```
    {
      "name": "rustic",
      "purl": "pkg:cargo/rustic?vcs_url=git%2Bhttps://github.com/cachito-testing/cachi2-rust.git%408ed95a9ae1b4110dd97c722e7c47522bc6cc2a5a",
      "properties": [
        {
          "name": "hermeto:found_by",
          "value": "hermeto"
        }
      ],
      "type": "library"
```
This may be acceptable, but I'm wondering if it's the best we can do, NB we don't currently document this anywhere, so users may expect to see some version since `Cargo.lock` will have it. At the very least I'd prefer this workspace behaviour to be documented somehow, but I would definitely not oppose mirroring the `cargo` default for the field in this case so that our SBOM matches the actual lock file.

eskultety: Hmm, I guess the answer to the above is - "yes, it's fine" solely based on the fact that `version` isn't a mandatory SBOM component field.

hermeto/core/package_managers/cargo/main.py
eskultety: Having this discussion in the first place means there are gaps in the design doc and we should probably add a note on https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-workspace and what it means for us or the resulting SBOM since there will be no "main" package.

tests/unit/package_managers/cargo/test_main.py
a-ovchinnikov: I would rather had this as a doctest on the class: more compact, adds usage example to the class, does not get run on every occasion (no need to since chances to changes to this function are close to zero). I think that would have been a good compromise between no tests at all and a unit test for a helper (which, _by definition_, is not a valid unit since it does not bear any busyness logic and exists solely to facilitate something). However that suggestion did not fly for some reason.

eskultety: > unit test for a helper (which, by definition, is not a valid unit since it does not bear any busyness logic and exists solely to facilitate something).

Hmm, either I'm missing something or by that same definition we'd never write any unit test at all since in the end all business logic is just a composition of small private helpers facilitating some elementary functionality which we tend to create for easier maintenance, better readability, and straightforward mocking (when it actually comes to testing) :) . I'm stating this in context of another remark you made earlier in this thread:

> UTs do not make much sense for a PM, ITs and E2Es are better positioned to verify that a PM actually works

which in code translation equals testing `fetch_xyz_sources` (or other variants of the name) AND/OR `resolve_xyz`.

a-ovchinnikov: Correct, since PMs deal with arranging external components I think the main focus should be on integration and E2E tests, with UTs playing auxiliary role. For helpers which do not represent busyness logic and just abstract repeated patterns away doctest could provide both documentation and reproducible verification steps. Doctests would effectively outline the supposed usage of an object and provide a trace of actual testing done on the object.

Having UTs dedicated to helpers is problematic -- they can artificially inflate coverage. I.e. if a helper is covered, but is never accessed from actual busyness logic (for example, because of a bug in the aforementioned logic), and that branch in busyness logic unit is not covered for some reason, then the coverage would drop by a fraction of a percent. When helpers are tested through units which actually call them then there is a higher chance of noticing that something is not happening. In my opinion having helpers tested directly gives false sense of confidence.

The other thing with helpers is them being usually linear or almost linear fragments of code. What is the value of a test in such a case? It is not like anyone doubts Python's ability to branch or to do trivial substitutions. The most important aspect of such tests is an additional layer of documentation which in the case of dedicated UTs, especially those which add fancy parametrization, becomes much less clear. Below are two tests out of four compacted into a docstring:

```python
"""
    Represents a package from Cargo.lock file
    <the rest of the docstring>

    >>> CargoPackage(**{"name": "foo", "version": "0.1.0"}).purl.to_string()
    'pkg:cargo/foo@0.1.0'
    >>> CargoPackage(
    ...     **{"name": "foo", "version": "0.1.0",
    ...        "source": "registry+https://github.com/rust-lang/crates.io-index",
    ...        "checksum": "abc123"}).purl.to_string()
    'pkg:cargo/foo@0.1.0?checksum=abc123'
"""
```

They are compact, demonstrate how a piece of code is to be used, are right where someone might be already looking at and they do not run every time an unrelated bit of code is modified.

eskultety: > Having UTs dedicated to helpers is problematic -- they can artificially inflate coverage. I.e. if a helper is covered, but is never accessed from actual busyness logic (for example, because of a bug in the aforementioned logic), and that branch in busyness logic unit is not covered for some reason, then the coverage would drop by a fraction of a percent. When helpers are tested through units which actually call them then there is a higher chance of noticing that something is not happening. In my opinion having helpers tested directly gives false sense of confidence.

Matter of perception IMO. Everyone should be clear on what UT coverage really means - the only thing coverage really tells you is how much effort is a project investing in ensuring its stability, seeing a ridiculously low coverage to me screams maintainers don't care and hence I'd have no confidence in such a project. Conversely, hight UT coverage tells me that maintainers take stability seriously, but doesn't give me actual confidence which can only be achieved with ITs. So, basically how I perceive UTs/ITs is - UTs are **only for devs**, ITs **are for everyone** but the tradeoff is execution time.
I would definitely not downplay the importance of UTs because having an already standardized and automated way of executing smoke tests is a great way of getting instant feedback on interface compatibility breakages without the need to give something a second thought.

> They are compact, demonstrate how a piece of code is to be used, are right where someone might be already looking at and they do not run every time an unrelated bit of code is modified.

I don't dispute ^this argument at all, but I'll personally never be a fan of using swiss army knife type of solutions over dedicated tooling. The fact that we as a project would prefer not running ^these docstring tests by default as you proposed means I'd have to remember running them which isn't going to happen and with a dedicated, standardized test framework already at hand I also don't see the point of mixing docs with exec logic. The moment you automate their execution they're no different to existing UT solutions IMO.
You need to also think about following common development practices that, if the desire is to gain contributor attraction in the future, are in sync with nowadays trends and I think that's where _pytest_ shines.

Let me re-iterate, while I find those docs strings + tests neat in a way, I personally will most likely never endorse them or won't pick them as my tool of preference over using a standardized, common, widely popular, time tested test framework which is already utilized within this project. I don't like spending any more time with tests than absolutely necessary despite acknowledging their utmost importance, so I'd always opt into a "set it and forget it" type of **test** solution revolving around copy-pasting existing tests to remain consistent with the overall code base trend. The thing for us to focus on really needs to be the dichotomy UTs vs ITs for a given code change.

tests/unit/package_managers/cargo/test_main.py
slimreaper35: Looks like the discussion has moved to another thread.

hermeto/core/package_managers/cargo/main.py
slimreaper35: So, what do we want to do with the version? Keep it None or "0.0.0".

In case of minimal workspace with no metadata, it should stay None. If there is a `[package]` table with no version defined, then we can fallback to "0.0.0".

(to be aligned with Cargo)

hermeto/core/package_managers/cargo/main.py
slimreaper35: Sure, I can squash them. At the same time, I see some comments from @eskultety about splitting commits.

hermeto/core/package_managers/cargo/main.py
taylormadore: Does cargo add subpaths to the source for git deps? I tried a quick example with a dependency located in a subpath in the main repo and didn't see cargo add a subpath to the source

hermeto/core/package_managers/cargo/main.py
taylormadore: At the risk of adding an _additional_ (not specific to this issue) opinion about commit splitting, I think the test-data-change commits should be squashed into the commit that causes them to change. Otherwise the commits with the functional changes have failing tests

I will go hide now :face_with_peeking_eye:

hermeto/core/package_managers/cargo/main.py
taylormadore: > minimal workspace with no metadata

Just to clarify (sorry, I am a cargo noob), do you mean virtual workspace?

If so, I agree with the above.

slimreaper35: yes

eskultety: Let's keep it as None then.

hermeto/core/package_managers/cargo/main.py
eskultety: Okay, thanks for trying.

hermeto/core/package_managers/cargo/main.py
eskultety: >  I think the test-data-change commits should be squashed into the commit that causes them to change. Otherwise the commits with the functional changes have failing tests

Oh, absolutely! I didn't even think this needs to be discussed, splitting in exchange for breaking tests due to granularity, that's wrong. Splitting should always be done in a way where **at least** unit tests are not broken in between (it's sometimes hard not to break integration tests in between and we have violated that on many occasions in the past and that's why I tend to be more lenient about it, but I sure do not dispute the argument, quite the contrary!). That said, please use common sense and let's move on :).

tests/unit/package_managers/cargo/test_main.py
a-ovchinnikov: Yes, I agree in general with one small comment:
```
$ pytest --doctest-modules
```
appears to be the stock way of running them indirectly (i.e. through [pytest](https://docs.pytest.org/en/stable/how-to/doctest.html))

hermeto/core/package_managers/cargo/main.py
slimreaper35: resolving thread -> #1001

hermeto/core/package_managers/cargo/main.py
eskultety: Resolving based on https://github.com/hermetoproject/hermeto/pull/988#discussion_r2179359534, I will open a docs issue on our SBOM (if we don't have one already that could track such an update as well)

eskultety: https://github.com/hermetoproject/hermeto/issues/1012

hermeto/core/package_managers/cargo/main.py
slimreaper35: The version part is only necessary to pass the CI. Otherwise, I wouldn't touch it.

hermeto/core/package_managers/cargo/main.py
slimreaper35: Which comment is this about?

eskultety: Commit named '_cargo: Split fetching dependencies and component resolution_' where you not only moved code cleanly, but also made other small changes along the way. In general, when moving code, first the code needs to be moved 1:1 wherever applicable, then the rest of the changes can follow that.

slimreaper35: fixed

hermeto/core/package_managers/cargo/main.py
```diff
@@ -88,13 +88,19 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
 def _fetch_dependencies(package_dir: RootedPath, request: Request) -> dict[str, Any]:
+    """Fetch cargo dependencies and return a config template for hermetic build."""
     vendor_dir = request.output_dir.join_within_root("deps/cargo")
-    # --no-delete to keep everything already present. It does not matter for a fresh
-    # single package, but it does matter when there is pip interaction.
+    # --locked           Assert that `Cargo.lock` will remain unchanged.
+    # --versioned-dirs   Always include version in subdir name.
+    # --no-delete        Don't delete older crates in the vendor directory.
+    #                    It is necessary to make Cargo keep dependencies that are already
+    #                    present in the vendored directory. This flag have no effect on standalone
```
a-ovchinnikov: s/flag have/flag has/

hermeto/core/package_managers/cargo/main.py
```diff
@@ -220,26 +253,72 @@ def _run_cmd_watching_out_for_lock_mismatch(
...
 
-def _generate_sbom_components(package_dir: RootedPath) -> chain[Component]:
+def _find_local_packages(package_dir: RootedPath) -> dict[str, str]:
+    """Find local packages in the Cargo.toml file and return their subpaths."""
+    parsed_toml = _parse_toml_project_file(package_dir.path / "Cargo.toml")
+
+    result = {}
+
+    runtime_deps = parsed_toml.get("dependencies", {})
+    # https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html
```
a-ovchinnikov, line 265: Thank you for adding the link! I would strongly recommend adding a short summary of a special case right in a comment and adding a link _after that_. The reason for that is to reduce switching for a reader.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -220,26 +253,72 @@ def _run_cmd_watching_out_for_lock_mismatch(
...
+        if pkg_name == main_package_name:
+            components.append(
+                LocalCargoPackage(
+                    name=main_package_name,
+                    version=main_package_version,
+                    vcs_url=vcs_url,
+                    subpath=str(package_dir.path.relative_to(package_dir.root)),
+                ).to_component()
+            )
+        elif pkg_name in local_packages.keys():
```
a-ovchinnikov: Non-blocking observation: having a `is_local_package()` defined somewhere would have slightly improved the flow. Returning it from `_find_local_packaged()` (properly renamed) could be done. Bending the rules a little and agreeing on the helper to return either an empty string or a path could result in something like this:
```python
elif subpath := is_local_package(pcg_name):
   ...
   subpath=subpath,
   ...
```

a-ovchinnikov (approved):

PR #991, opened 2025-06-26
Support Rust extensions for pip URL and VCS dependencies

### LINKED ISSUE #983: hermeto looking for Cargo.lock in wrong location when downloading dependency from https
tiktoken is not locked, so I created a repo with the contents and locked the file
https://github.com/jmontleon/tiktoken

When I adjust my requirements.txt following the documentation like so:
```
tiktoken @ https://github.com/jmontleon/tiktoken/archive/refs/heads/main.tar.gz \
    --hash=sha256:bd7265958ae33394646906614e23829e41f4053080c7be64979791b403ed283e
```

hermeto complains it is not locked:
```
2025-06-19 02:31:37,589 ERROR PackageRejected: /home/jason/Downloads/kai/hermeto-output/deps/pip/tiktoken-main is not locked
Error: PackageRejected: /home/jason/Downloads/kai/hermeto-output/deps/pip/tiktoken-main is not locked
```

But that path does not exist. It's at:
```
$ ls /home/jason/Downloads/kai/hermeto-output/deps/pip/external-tiktoken/tiktoken-main/
Cargo.lock  CHANGELOG.md  MANIFEST.in  PKG-INFO        README.md  setup.cfg  src    tiktoken           tiktoken_ext
Cargo.toml  LICENSE       perf.svg     pyproject.toml  scripts    setup.py   tests  tiktoken.egg-info
```

slimreaper35:
Closes #983


slimreaper35:
Seems like a much easier solution than modifying `filter_packages_with_rust_code` function, I will try to look at it.

hermeto/core/package_managers/pip/main.py
```diff
@@ -676,22 +676,15 @@ def _get_external_requirement_filepath(requirement: PipRequirement) -> Path:
             filename = Path(orig_url.path).name
             filepath = Path(urlparse.unquote(filename))
         else:
-            # e.g. external-pyarn/pyarn-external-sha256-deadbeef.tar.gz
-            filepath = Path(
-                f"external-{package}", f"{package}-external-{algorithm}-{digest}{file_ext}"
-            )
+            # e.g. pyarn-external-sha256-abc123.tar.gz
+            filepath = Path(f"{package}-external-{algorithm}-{digest}{file_ext}")
```
eskultety: Hmm, based on https://github.com/hermetoproject/hermeto/issues/983#issuecomment-2995258774 I'm still wondering whom are we helping with too much of a customization of the tarball name really. The deps directory is user facing so from UX POV if anything I as a user would expect to just find package-named sdists as listed in my requirements file. IOW the fact that we're dealing with direct dependencies instead of those coming from an index should be completely transparent to the end user unless we need to something more with them in which case that should happen via some additional logic, files, other artifacts, but not on the `deps` directory structure I think.

Now, this apparently is enough to fix the issue at hand, great, it's a dead simple change, but I'd like to take this to a higher level - let's pretend 'external' isn't really a thing we need to do (seems like it was never used anyway), so let's add a commentary somewhere around here stating that we potentially might need to start tracking external deps for whatever purposes in the future and do that with a hidden file on the FS or just a data structure.
Then, let's start naming those files appropriately as we'd get them from an index, i.e. `{package}.tar.gz`. With `algo:digest` I have the exact same reasoning as with `external`, whom are we helping with it to have it encoded in the name? IOW a fully resolved requirements file lists digests for all deps, so why don't other deps have it encoded in their names? At this point in time I don't see a reason for us to go an extra mile dealing with those particular deps more than necessary and so we should just "pretend" those are deps like any other and make no differences.

For reference, this is what the logs are going to look like after this PR:
```
...
2025-06-27 11:48:09,973 DEBUG Successfully processed 'tiktoken @ https://github.com/jmontleon/tiktoken/archive/refs/heads/main.tar.gz' in path 'deps/pip/tiktoken-external-sha256-d9ad52c4e2ff5f04873285c2b19a83ff7aad29da72ab1ab9ab16b838bf14ff5c.tar.gz'
...
2025-06-27 11:48:18,776 INFO Fetching cargo dependencies at /var/tmp/kai/kai_mcp_solution_server/hermeto-output/deps/pip/tiktoken-main
```

I mean, yeah, `tiktoken-main` is clearly a by-product of Gh's auto-generated tarball out of a magic source URL that I'm not sure can be solved in an elegant way on our Cargo backend side, but I think mentioning just `tiktoken.tar.gz` in the logs may have been less confusing. That said, ^this all is just my idea, I can be swayed in the exact same direction as this PR is going with good arguments, but I don't see any good use for the heavily customized names ATM.

hermeto/core/package_managers/pip/main.py
```diff
@@ -676,22 +676,15 @@ def _get_external_requirement_filepath(requirement: PipRequirement) -> Path:
             filename = Path(orig_url.path).name
             filepath = Path(urlparse.unquote(filename))
         else:
-            # e.g. external-pyarn/pyarn-external-sha256-deadbeef.tar.gz
-            filepath = Path(
-                f"external-{package}", f"{package}-external-{algorithm}-{digest}{file_ext}"
-            )
+            # e.g. pyarn-external-sha256-abc123.tar.gz
```
eskultety: FYI `deadbeef` (over `abc123`) was perfectly fitting in the original comment, it's a common hex radix notation for debug values and examples: https://en.wikipedia.org/wiki/Magic_number_(programming)#Debug_values

That said, if you end up applying my functional suggestion, the comment will become completely irrelevant.

hermeto/core/package_managers/pip/main.py
slimreaper35: Should I create an issue for this "historical feature"? There are more things that we should revisit, like dealing with hashes or specifically `cachito_hash`addition.

eskultety: > Should I create an issue for this "historical feature"? 

If we address it in full in this PR then I don't necessarily need to see an issue for it. You can do that if you want - your call, but I'm not sure how much more aside from the existing comment above we could put in there based on:
1. what we know
2. how much we can disclose

> There are more things that we should revisit, like dealing with hashes or specifically cachito_hashaddition.

Yes, this one has been on my radar for a while, yet we don't track it anywhere, so yes an issue would be nice for this one.

eskultety: > > Should I create an issue for this "historical feature"?

@slimreaper35 oh, did you mean start tracking this in case we may need it in the future again? I may have misunderstood, sorry. In that case, let's not add issues to the backlog we're unsure about.

slimreaper35:
Oh Lord, `appr` is causing some problems.

Why would we have the same dependency [twice](https://github.com/hermetoproject/integration-tests/blob/pip/mixed-hashes/requirements.txt) in requirements.txt ?

chmeliik:
> Oh Lord, `appr` is causing some problems.
> 
> Why would we have the same dependency [twice](https://github.com/hermetoproject/integration-tests/blob/pip/mixed-hashes/requirements.txt) in requirements.txt ?

Probably to test that Hermeto uses unique enough filenames. Remember that Hermeto can be used to process multiple python projects in the same repo, so the same dependency can very well appear multiple times with different versions (or urls). And in requirements-build.txt, it's not unusual for the same dependency to appear twice within the same file (though it would be unusual to find url dependencies there)

slimreaper35:
Hmm, so the cause is here - https://github.com/hermetoproject/hermeto/blob/main/tests/integration/utils.py#L234

eskultety:
> > Oh Lord, `appr` is causing some problems.
> > Why would we have the same dependency [twice](https://github.com/hermetoproject/integration-tests/blob/pip/mixed-hashes/requirements.txt) in requirements.txt ?
> 
> Probably to test that Hermeto uses unique enough filenames. Remember that Hermeto can be used to process multiple python projects in the same repo, so the same dependency can very well appear multiple times with different versions (or urls).

Ah, that's right, we need to ensure there's still some uniqueness, so having the digest is indeed important.

eskultety:
> > Oh Lord, `appr` is causing some problems.
> > Why would we have the same dependency [twice](https://github.com/hermetoproject/integration-tests/blob/pip/mixed-hashes/requirements.txt) in requirements.txt ?
> 
> Probably to test that Hermeto uses unique enough filenames. 

Looking at the existing test, yeah, it gets the job done, but having a dedicated test (based off of the one about mixed hashes) explicitly to test this direct dependency naming scheme would IMO improve the clarity AND hopefully cause less confusion in the future if we are to re-open this discussion again.

hermeto/core/package_managers/pip/main.py
```diff
@@ -676,20 +676,15 @@ def _get_external_requirement_filepath(requirement: PipRequirement) -> Path:
             filename = Path(orig_url.path).name
             filepath = Path(urlparse.unquote(filename))
         else:
-            # e.g. external-pyarn/pyarn-external-sha256-deadbeef.tar.gz
-            filepath = Path(
-                f"external-{package}", f"{package}-external-{algorithm}-{digest}{file_ext}"
-            )
+            # e.g. pyarn-external-sha256-deadbeef.tar.gz
+            filepath = Path(f"{package}-external-{algorithm}-{digest}{file_ext}")
```
eskultety: You put the `external-` fragment back in with the partial revert :) . Digest on its own should provide the required uniqueness (which was an oversight on my side during the initial review), and since we only need it for uniqueness, we IMO don't need to track the algorithm, but that's optional, `external` should go away regardless.

hermeto/core/package_managers/pip/main.py
```diff
@@ -676,22 +676,13 @@ def _get_external_requirement_filepath(requirement: PipRequirement) -> Path:
...
         repo = git_info["repo"]
         ref = git_info["ref"]
-        # e.g. github.com/containerbuildsystem/pyarn/pyarn-external-gitcommit-badbeef.tar.gz
-        filepath = Path(
-            git_info["host"],
-            git_info["namespace"],  # namespaces can contain '/' but pathlib can handle that
-            repo,
-            f"{repo}-external-gitcommit-{ref}.tar.gz",
-        )
+        filepath = Path(f"{repo}-external-gitcommit-{ref}.tar.gz")
```
eskultety: I think you still forgot one more `external` instance :)

slimreaper35: Updated + added this change:

```diff
- if "-external-gitcommit-" in file_name:
+ if "-gitcommit-" in file_name:
    files_checksums[rel_file] = _get_git_commit_from_tarball(
        root_dir.joinpath(rel_file)
    )
```

a-ovchinnikov (commented):
> Tested with simple requirements.txt file ...

Did you run an intergation test or an E2E test?

slimreaper35:
> Did you run an integration test or an E2E test? Edit: typo

Just made sure the prefetching works with the example in the issue description.

a-ovchinnikov:
> made sure the prefetching works with the example in the issue description

I think this warrants at least an IT, better a E2E -- this was caught as a part of a complex workflow, I think it is worth it to try and replicate it, especially if there is a reference to it in the commit message.

eskultety:
> > made sure the prefetching works with the example in the issue description
> 
> I think this warrants at least an IT, better a E2E -- this was caught as a part of a complex workflow, I think it is worth it to try and replicate it, especially if there is a reference to it in the commit message.

I don't think we need a full-fledged E2E for a structural (but not logical) change like this one since the problem was caught during prefetch (not even build). It would only lead to defining yet another multi-backend E2E test which take a very long time. That said, if an e2e is truly desired then the existing Python+Rust ext. test should be enhanced to cover this scenario.

slimreaper35:
> I don't think we need a full-fledged E2E for a structural (but not logical) change like this one since the problem was caught during prefetch (not even build). It would only lead to defining yet another multi-backend E2E test which take a very long time. That said, if an e2e is truly desired then the existing Python+Rust ext. test should be enhanced to cover this scenario.

Created PR https://github.com/hermetoproject/integration-tests/pull/11

hermeto/core/package_managers/pip/rust.py
```diff
@@ -24,70 +23,69 @@ def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
...
+        if not package_path:
+            continue
+
+        filename = Path(Path(package_path).name)
+        while filename.suffix in (".zip", ".tar", ".gz", ".tgz"):
+            filename = filename.with_suffix("")
+
+        pip_deps_dir = Path(package_path).parent
+        extract_dir = pip_deps_dir / filename
+        shutil.unpack_archive(package_path, extract_dir=extract_dir)
```
taylormadore: Do we need to worry that we're extracting archives that are untrusted from our perspective?

slimreaper35: I forgot to add the [filter](https://docs.python.org/3/library/shutil.html#shutil.unpack_archive) keyword. But that is not valid for zip files.

hermeto/core/package_managers/pip/rust.py
```diff
@@ -24,70 +23,69 @@ def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
...
 
 
 def filter_packages_with_rust_code(packages: list[dict[str, Any]]) -> list[CargoPackageInput]:
     """Filter packages that contain Rust code from a list of pip packages."""
     packages_containing_rust_code = []
-    tar_packages = [p for p in packages if tarfile.is_tarfile(p.get("path", ""))]
-    for p in tar_packages:
-        fname = p.get("path", "")
-        # File name and package name may differ e.g. when there is a hyphen in
-        # package name it might be replaced by an underscore in a file name.
```
a-ovchinnikov, line 63: The original code relied on reusing a package name to find the archive. That worked perfectly in 95% of cases until I ran into a package with a hyphen. I left the comment to guard against an overzealous refactoring. I suggest not dropping comments.

hermeto/core/package_managers/pip/rust.py
```diff
@@ -24,70 +23,69 @@ def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
...
+
+        filename = Path(Path(package_path).name)
+        while filename.suffix in (".zip", ".tar", ".gz", ".tgz"):
+            filename = filename.with_suffix("")
+
+        pip_deps_dir = Path(package_path).parent
+        extract_dir = pip_deps_dir / filename
+        shutil.unpack_archive(package_path, extract_dir=extract_dir)
+
+        cargo_files = list(extract_dir.rglob("Cargo.toml"))
```
a-ovchinnikov, line 73: Neat!

hermeto/core/package_managers/pip/rust.py
```diff
@@ -24,70 +23,74 @@ def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
...
-        pkg_name = source_tarball.getnames()[0].split("/")[0]
-        try:
-            file = source_tarball.extractfile(f"{pkg_name}/{file_name}")
-        except KeyError:
+        file_path = extracted_dir / file_name
+        if not file_path.exists():
             continue
-        # the file is decoded as utf-8-sig because plain utf-8 has proven to
-        # be problematic with certain sources from pypi
-        # see https://github.com/hermetoproject/pybuild-deps/blob/4dc40ffabddb8aad1279978b8741111fb64452e6/src/pybuild_deps/finder.py#L45-L51
```
a-ovchinnikov, line 41: Why did you drop L39-L41?

hermeto/core/package_managers/pip/rust.py
```diff
@@ -24,70 +23,74 @@ def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
...
+        cargo_files = list(extract_dir.rglob("Cargo.toml"))
+        if not cargo_files:
+            shutil.rmtree(extract_dir, ignore_errors=True)
+            continue
+
+        # The unpacked URL/VSC package may have an arbitrary directory name that we cannot control.
+        # Therefore, it is inside a predictable directory derived from the package name.
+        subdirs = [d for d in extract_dir.iterdir() if d.is_dir()]
+        if len(subdirs) == 1:
+            real_extract_dir = subdirs[0]
```
a-ovchinnikov: Minor naming question: do you mean `real_extract_dir` to be `actual_extract_dir` or `dir_with_sources` (aka `sources_root`)?

slimreaper35: Maybe `source_root` fits better. The structure is `cryptography-somehash/cryptography-41.0.2/<source-code>`.

EDIT: or `source_dir`

hermeto/core/package_managers/pip/rust.py
```diff
@@ -24,70 +23,78 @@ def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
...
+        # The unpacked URL/VSC package may have an arbitrary directory name that we cannot control.
+        # Therefore, it is inside a predictable directory derived from the package name.
+        subdirs = [d for d in extract_dir.iterdir() if d.is_dir()]
+        if len(subdirs) == 1:
+            source_dir = subdirs[0]
+        else:
+            # should not happen
+            source_dir = extract_dir
+
+        if not _depends_on_rust(source_dir):
```
a-ovchinnikov, line 87: A minor, optional, non-blocking naming suggestion: s/not _depends_on_rust/no_dependency_on_rust_detected_in/ (if accepted would require an inversion in the original logic).

a-ovchinnikov (approved):

hermeto/core/package_managers/pip/rust.py
```diff
@@ -24,70 +23,78 @@ def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
...
-            rust_root = Path(sorted(cargo_manifests, key=len)[0]).parent
-
-        tf.extractall(path=Path(str(tf.name)).parent, filter="data")
-        packages_containing_rust_code.append(CargoPackageInput(type="cargo", path=rust_root))
+        package_path = p.get("path", "")
+        if not package_path:
+            continue
+
+        filename = Path(Path(package_path).name)
+        extract_filter = None if filename.suffix == ".zip" else "data"
```
eskultety: nitpick: the convention is to default to None not lead with it, IOW
`a = foo if <predicate> else None`

hermeto/core/package_managers/pip/rust.py
```diff
@@ -24,70 +23,78 @@ def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
...
+            shutil.rmtree(extract_dir, ignore_errors=True)
+            continue
+
+        # The unpacked URL/VSC package may have an arbitrary directory name that we cannot control.
+        # Therefore, it is inside a predictable directory derived from the package name.
+        subdirs = [d for d in extract_dir.iterdir() if d.is_dir()]
+        if len(subdirs) == 1:
+            source_dir = subdirs[0]
+        else:
+            # should not happen
```
eskultety: "should not happen" only IF it's a proper packaged tarball, we're talking about a random URL tarball that pretends to be an installable python/rust package which may not have the top-level directory (yes, I would not expect this to be very common). The check above would crumble in that case because you're counting on the number of directories, but you could have files on the top level which immediately invalidate it.

hermeto/core/package_managers/pip/rust.py
```diff
@@ -24,70 +23,78 @@ def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
...
+        pip_deps_dir = Path(package_path).parent
+        extract_dir = pip_deps_dir / filename
+        shutil.unpack_archive(package_path, extract_dir=extract_dir, filter=extract_filter)  # type: ignore[arg-type]
+
+        cargo_files = list(extract_dir.rglob("Cargo.toml"))
+        if not cargo_files:
+            shutil.rmtree(extract_dir, ignore_errors=True)
+            continue
+
+        # The unpacked URL/VSC package may have an arbitrary directory name that we cannot control.
```
taylormadore: minor typo: VSC --> VCS

hermeto/core/package_managers/pip/rust.py
slimreaper35: which safeguard ? (not sure if I get it)

eskultety: essentially it's just about converting
```
subdirs = [d for d in extract_dir.iterdir() if d.is_dir()]
if len(subdirs) == 1:
```
to
```
nitems = sum(1 for i in extract_dir.iterdir())   <<< no need to construct a list to count len
if nitems != 1:
  # non-standard packaging scheme that doesn't come with a top-level <project>-<version> directory
  source_dir = extract_dir
else:
  source_dir = next(extract_dir.iterdir())      <<< running iterdir twice is cheap in this case because it either always had a single item or we only end up running it once; if this is not a directory, it'll fail in a different part of the code anyway
```

PR #992, opened 2025-06-27
Use mirror.gcr.io for pulling images

MartinBasti:
Avoiding usage docker hub directly as it causes issues with rate limiting in the build system and require also dealing with credentials to get higher rate limit, which is not suitable for developers of OSS project.

GCR mirror should be suitable replacement without strict ratelimiting


a-ovchinnikov (approved):

Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM registry.access.redhat.com/ubi9/ubi@sha256:304b50df1ea4db9706d8a30f4bbf26f582936ebc80c7e075c72ff2af99292a54 as ubi
-FROM docker.io/library/golang:1.20.0-bullseye as golang_120
```
eskultety, line 2: **Commit msg:**
Please link the following resource for more context: https://cloud.google.com/artifact-registry/docs/pull-cached-dockerhub-images

Also, while at it:
`s/usage/usage of`
`s/require/requires`

eskultety (approved):

Dockerfile
MartinBasti: added

PR #1020, opened 2025-07-10
WIP - Maven Support with Maven-Lockfile Plugin

adambkaplan:
Adding experimental support for pre-fetching Maven dependencies. This process requires developers to add the [maven-lockfile plugin](https://github.com/chains-project/maven-lockfile) to their repositories and generate the `lockfile.json` file.


hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,397 @@
...
+from hermeto.core.models.output import ProjectFile, RequestOutput
+from hermeto.core.models.property_semantics import PropertySet
+from hermeto.core.models.sbom import Component
+from hermeto.core.package_managers.general import async_download_files
+from hermeto.core.rooted_path import RootedPath
+
+log = logging.getLogger(__name__)
+
+# Maven scope types
+MAVEN_SCOPES = ("compile", "provided", "runtime", "test", "system", "import")
```
eskultety, line 25: Java noob here, what are these scopes?
Seeing `compile` in the list, does this mean it's a binary artifact? Does it come with a corresponding source artifact or is like some e.g. Python packages which only distribute as wheels? If that's the case, we don't do binary artifacts by default because one doesn't know how those were built and we have a special option in the input JSON to enable that - `allow_binary`.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,397 @@
...
+
+    @property
+    def included(self) -> bool:
+        """Return True if this dependency should be included."""
+        return self._dependency_dict.get("included", True)
+
+    @property
+    def children(self) -> list[dict[str, Any]]:
+        """Get the children dependencies."""
+        return self._dependency_dict.get("children", [])
```
eskultety: We don't do properties in dictionary-like cases anymore, so this will have to inherit from UserDict or potentially only have instance attributes, we do the former with our more recent backends. We did something different for Yarn v1, though I don't remember off the top of my head without diving into the code base (maybe a frozen class or pydantic's basemodel), point is - this generated property list isn't really that readable if you need to extend it in the future and only creates churn, so this can definitely be written in a more compact form.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,397 @@
...
+    def get_main_package(self) -> MavenComponentInfo:
+        """Get the main package as a MavenComponentInfo."""
+        main_info = self.main_package_info
+        purl = PackageURL(
+            type="maven",
+            namespace=main_info["groupId"],
+            name=main_info["artifactId"],
+            version=main_info["version"],
+        )
+
```
eskultety, line 261: Is main package a thing with Java/Maven? Can there be something like workspaces where your application doesn't really define a main package, rather a collection of packages that constitute your whole application Foo. If workspaces are a thing, we need to take them into account.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,397 @@
...
+def _get_maven_dependencies(
+    download_dir: RootedPath, deps_to_download: dict[str, dict[str, Optional[str]]]
+) -> dict[str, RootedPath]:
+    """Download Maven dependencies and return their local paths."""
+    download_paths = {}
+    files_to_download = {}
+    
+    for url, dep_info in deps_to_download.items():
+        # Create Maven directory structure: groupId/artifactId/version/
```
eskultety, line 306: Is this structure standardized and documented anywhere?

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,397 @@
...
+        asyncio.run(
+            async_download_files(
+                files_to_download,
+                get_config().concurrency_limit,
+            )
+        )
+
+    # Verify checksums
+    for url, dep_info in deps_to_download.items():
+        if dep_info["checksum"] and dep_info["checksum_algorithm"]:
```
eskultety, line 381: Are checksums a mandatory dependency attribute in Maven's world? If not then this can generate an exception easily so we'd need to account for that.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,397 @@
...
+                local_path = download_paths[resolved_url]
+                # Use template variable for output directory
+                templated_path = Path("${output_dir}", local_path.subpath_from_root)
+                dep_dict["resolved"] = f"file://{templated_path}"
+            
+            # Recursively update children
+            if dep_dict.get("children"):
+                update_dependency_paths(dep_dict["children"])
+
+    update_dependency_paths(maven_lockfile.lockfile_data.get("dependencies", []))
```
eskultety, line 466: So, what's the realistic scenario with children? Can we assume the level of `children` nesting here is going to stay within reasonable boundaries, otherwise we're just going to waste memory which already is quite limited in container workloads, so I'd personally always prefer to see an iterative approach rather than recursion unless we're absolutely sure that we're only ever going to work with a few levels of nesting and the iterative approach would not be worth the hindered readability, which is not the case here, it can be rewritten (even with iterators) easily.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,397 @@
...
+        download_paths[url] = local_path
+
+    # Download all files
+    if files_to_download:
+        asyncio.run(
+            async_download_files(
+                files_to_download,
+                get_config().concurrency_limit,
+            )
+        )
```
eskultety: I don't see any deduplication here. We just download everything based on `groupId/artifactId/version/` . What if you have:
```
Foo1 -> Baz
Foo2 -> Bar -> Baz
Foo3 -> FooBar -> Bar -> Baz
```
how many copies of `Baz` are we going to fetch? If this is anything like NPM's `node_modules` directory explosion then I think we'll need to deduplicate and only download a single copy and link it properly where needed to save bandwidth since it looks like we're in full control over this process, i.e. we're not relying on any native tooling underneath (e.g. `cargo`).

README.md
```diff
@@ -230,6 +233,13 @@ some cases drastically so.
 
 See docs/gomod.md for more details.
 
+### [maven][]
+
+Hermeto supports maven by parsing one or more `lockfile.json` files generated by the
+[maven-lockfile plugin](https://github.com/chains-project/maven-lockfile). Maintainers
+are responsible for adding this plugin to their code and invoking the required Maven
+targets to generate the lockfile and build using a "frozen" `pom.xml` file.
+
```
eskultety, line 242: This sadly won't be enough for user facing docs, we'll need a more detailed `maven.md` docs describing the usage. Also, this is the only occurrence of `pom.xml` being mentioned. How does one generate it? Is Hermeto expected to do that? What information do we need to generate it in that case?

hermeto/core/models/input.py
```diff
@@ -54,7 +54,7 @@ def show_error(error: "ErrorDict") -> str:
 
 
 # Supported package managers
-PackageManagerType = Literal["bundler", "cargo", "generic", "gomod", "npm", "pip", "rpm", "yarn"]
+PackageManagerType = Literal["bundler", "cargo", "generic", "gomod", "maven", "npm", "pip", "rpm", "yarn"]
```
eskultety, line 57: I don't believe we'll merge this in one go (I think it's been a loong time since we did, probably just at the beginning), so this will have to live in under the `dev-package-managers` (or what the structure holding these is called).

eskultety (commented):
@adambkaplan Thank you very much for the draft. Given it's mostly AI generated I am impressed by the overall structure. Based on the fact that no context was provided for this work I could only comment on a few things that immediately caught my eye without diving too much into the implementation (also because this is a draft). Can you point us to some projects in the wild making use of this plugin to see whether there's really only a single pattern this can be used with?
Now, in order for this work to be considered ready for merge, a few things to focus on before lifting the Draft status:

- we'll need a design doc that serves as the condensed source of truth not only for the implementation decisions and trade-offs, but also as a reference to all state-of-the-art work this implementation is based on (i.e. references to docs, code, issues, blogs, etc.)
- does this approach fully deprecate https://github.com/hermetoproject/hermeto/blob/main/docs/generic.md#maven-artifacts ?
- the changes will need to be broken into individual commits, otherwise it's unreviewable. Also we don't do squash merge here so any fixup commits you may have need to be squashed in the respective commits introducing a related change. CONTRIBUTING.md should give you a rough idea (although truth to be told we don't have a detailed "new package manager backend" type of guide yet) or any past pull request adding a new backend, e.g. `Cargo/Rust` OR `Yarn v1` could both serve as a good example of how we expect to review additions such as this one. We can assist further.
- I can only see `fetch` being implemented, but how does one actually build off of the directory structure created by fetch with all the file paths mangled? That's where the `inject-files` command comes in, but that isn't implemented anywhere in this PR.
- can Maven be configured with any environment variables? Do we need to set any variables for a successful hermetic build based on the modified lockfile? If so -> `generate-env` is the command to do that and should be part of this PR to be accepted
- we'll need an integration test as a proof of functionality, there's https://github.com/hermetoproject/integration-tests that hosts the integration test scenarios.

chmeliik:
> reference to all state-of-the-art work this implementation is based on (i.e. references to docs, code, issues, blogs, etc.)

+1, I'll need more info about https://github.com/chains-project/maven-lockfile. How does it solve the allegedly unsolvable problem of generating a Java lockfile? Why should we have confidence that it actually works?

tests/unit/package_managers/test_maven.py
```diff
@@ -0,0 +1,260 @@
...
+from hermeto.core.package_managers.maven.main import (
+    MavenDependency,
+    MavenLockfile,
+    fetch_maven_source,
+)
+from hermeto.core.rooted_path import RootedPath
+
+
+def test_maven_dependency_properties():
+    """Test that MavenDependency correctly parses dependency properties."""
```
a-ovchinnikov, line 20: @eskultety this is a prime example of type of tests I talked about on multiple occasions, most recently in purls simplification PR. It bumps coverage nicely, however what exactly is it testing? If one checks `MavenDependency` implementation we'll see this:

```python
    def __init__(self, dependency_dict: dict[str, Any]) -> None:
        """..."""
        self._dependency_dict = dependency_dict
```

I.e. it just binds another name to an externally controlled mutable object (and I will not delve into why this is a questionable approach). Now on top of that it defines a bunch of accessors like the one below:
```python
    @property
    def group_id(self) -> str:
        """Get the group ID."""
        return self._dependency_dict["groupId"]
```
Again, not discussing the approach AI assistant took to implement it, note that it simply provides a property wrapper for a dictionary access.

Getting back to the test: in the test it arranges a dictionary, acts by instantiating an object which does nothing more than bind another name to a pre-existing object and then asserts that properties act as properties and that dictionary elements could be accessed. It looks convincing, however all it does is exercising some code which is guaranteed to work by someone else.

README.md
ben-alkov: > we'll need a more detailed maven.md docs describing the usage

Including an example walkthrough. PTAL at the .MDs for the individual package managers, paying special attention to the Example section (e.g. https://github.com/hermetoproject/hermeto/blob/main/docs/npm.md#example)

hermeto/core/package_managers/maven/__init__.py
ben-alkov, line 1: Missing EOF newline

hermeto/core/package_managers/maven/main.py
ben-alkov: Death to Recursion

hermeto/core/package_managers/maven/main.py
```diff
@@ -331,12 +331,47 @@ def _get_maven_dependencies(
             "repository_id": repository_id,
             "url": url
         })
+        
+        # WORKAROUND: Also track .pom files for _remote.repositories
+        pom_filename = Path(filename).stem + ".pom"
```
a-ovchinnikov, line 336: Two quick observations about this function:

1. It is already almost 150 lines long and keeps growing and
2. There are multiple one-line comments explaining what the next block code of is about to do.

The standard rule for 2. is to split out a helper each time a comment explaining what a block of code is about to do appears. This would also help with 1. _Edit_: word order.

hermeto/core/package_managers/maven/main.py
```diff
@@ -361,6 +396,41 @@ def _get_maven_dependencies(
...
+                    if pom_local_path.exists():
+                        checksum_info = ChecksumInfo(
+                            algorithm=python_algorithm,
+                            hexdigest=pom_checksum,
+                        )
+                        must_match_any_checksum(pom_local_path, [checksum_info])
+                        log.debug(f"Verified checksum for {pom_local_path}")
+                    else:
+                        log.warning(f"POM file {pom_local_path} not found for checksum verification")
+                except Exception as e:
```
a-ovchinnikov, line 429: This is extremely broad, could you please either make sure it is narrowed or there is a concise explanation why the broadest possible catch is necessary?

hermeto/core/package_managers/maven/main.py
```diff
@@ -331,12 +331,47 @@ def _get_maven_dependencies(
             "repository_id": repository_id,
             "url": url
         })
+        
+        # WORKAROUND: Also track .pom files for _remote.repositories
```
a-ovchinnikov, line 335: Could you please elaborate workaround around what?

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,397 @@
...
+            version=self.version,
+        )
+
+        return MavenComponentInfo(
+            name=self.name,
+            purl=str(purl),
+            version=self.version,
+            scope=self.scope,
+            included=self.included,
+            missing_hash_in_file=None if self.checksum else Path("lockfile.json"),
```
a-ovchinnikov, line 177: `Path("lockfile.json")` will be relative to the current working directory, is this expected? _Edit_: `Path` is used as a refined `str` in this context, this probably should be mentioned somewhere. Or are there any plans to use it as `Path` in the future?

a-ovchinnikov (commented):
A few quick observations.

adambkaplan:
Added a few more commits here to preserve my progress, but this PoC needs to be halted for the time being. I encountered a few blockers with the maven-lockfile plugin that prevent Hermeto from creating a valid Maven repository layout:

1. Lockfile is missing the POM (required metadata) and its checksum: https://github.com/chains-project/maven-lockfile/issues/1064
2. Lockfile is not resolving the dependencies of plugins: https://github.com/chains-project/maven-lockfile/issues/424
3. Lockfile is not providing the identifiers of the repositories used to fetch content: https://github.com/chains-project/maven-lockfile/issues/1269

The latter item reveals a challenge with Hermeto's philosophy, which attempts to prefetch dependencies _**independent of the language ecosystem tool**_ (am I correct here?). This requires Hermeto to reverse-engineer the expected layout of a given package manager. In the case of Maven, some of the details are considered "internal" and at risk of changing with a version upgrade.

chmeliik:
> The latter item reveals a challenge with Hermeto's philosophy, which attempts to prefetch dependencies _**independent of the language ecosystem tool**_ (am I correct here?). This requires Hermeto to reverse-engineer the expected layout of a given package manager. In the case of Maven, some of the details are considered "internal" and at risk of changing with a version upgrade.

More or less correct. IMO there's two main points to the philosophy:

* It should not matter what version of the tool the user has. If Hermeto uses Maven vX during prefetching and the build fails because the user has vY, that's a problem (poor design of Maven support)
  * For some package managers, it's ok to use the tool itself to prefetch dependencies. For example:
    * `cargo`, which has a stable vendor directory format, so it shouldn't matter what version Hermeto uses
    * `yarn`, because the ecosystem is built around recording the exact version of `yarn` in package.json and fetching the correct version before processing the dependencies (so Hermeto is easily able to use the same version as the user)
* If Hermeto is going to rely on the tool itself, it needs confidence that the tool will download only the dependencies that Hermeto knows about and not do anything unexpected. For example:
  * Hermeto cannot rely on `pip` for downloading, because `pip` will often execute arbitrary code defined by the dependencies (yes, even when just _downloading_ and not installing)
  * Hermeto can rely on `yarn` but has to disable some features (plugins) and ban others (git dependencies and "exec" dependencies, because both involve arbitrary code execution)
  * Hermeto can rely on `cargo` while restricting some configuration options that could lead to arbitrary code execution

eskultety:
> The latter item reveals a challenge with Hermeto's philosophy, which attempts to prefetch dependencies _**independent of the language ecosystem tool**_ (am I correct here?).

Yeah, I wouldn't really say ^this is our philosophy, it's more about applying safeguards along the way to avoid compromising the whole pipeline run. I'd condense @chmeliik's response on what we want to essentially:

"ensure native tooling can verify its own lockfile's integrity & deterministically download the dependencies in a way where the tooling doesn't employ any pre/post install OR other user configurable hook scripts"

If there's anything that would stand in the ^way, we need to come up with a crafted solution ourselves to comply with the premise.

hermeto/core/package_managers/maven/main.py
eskultety: @a-ovchinnikov I think at this point there already are quite a few substantial comments and this looks like still very much like a live WIP, far from a regular proposal, so I think you don't need to spend your precious time digging too deep into function-level stuff right now :)

a-ovchinnikov: This was a drive-by review, not a proper one, essentially things that caught my eye the moment I opened the diff.

adambkaplan:
> Hermeto cannot rely on pip for downloading, because pip will often execute arbitrary code defined by the dependencies (yes, even when just downloading and not installing)

@chmeliik in the case of Maven (and other Java builders that leverage Maven plugins, such as [Gradle](https://gradle.org/)), the answer is an emphatic _**no**_. Maven heavily relies on a plugin ecosystem that can execute arbitrary code. Some can even ship their own package managers that will attempt to install dependencies dynamically during a typical build lifecycle. Our own [Quarkus Quinoa](https://github.com/quarkiverse/quarkus-quinoa) plugin meets this criteria - it lets developers install + manage NodeJS dependencies within a Java build chain!

As @eskultety alluded to - this is all worth codifying in a design document.

a-ovchinnikov:
@coderabbitai review

slimreaper35:
This work has been suppressed by other **two** pull requests (https://github.com/hermetoproject/hermeto/pull/1381 and https://github.com/hermetoproject/hermeto/pull/1246)

PR #1022, opened 2025-07-11
Upgrade async HTTP requests

slimreaper35:

## Summary by CodeRabbit

* **New Features**
  * Added robust file download capabilities using a new library, supporting both synchronous and asynchronous downloads with retry logic and improved error handling.
  * Introduced enhanced Git repository URL parsing, extracting detailed information such as reference, namespace, host, and repository name.

* **Dependency Changes**
  * Replaced `aiohttp` and related packages with `httpx` and its dependencies for HTTP operations, simplifying and reducing the dependency set.

* **Bug Fixes**
  * Ensured correct file permissions are set for temporary `.netrc` files during integration tests.

* **Tests**
  * Updated and expanded tests to cover new Git URL parsing logic and file download features.
  * Removed obsolete tests related to deprecated download and parsing implementations.

* **Chores**
  * Cleaned up and reorganized import statements and removed unused code and files.


MartinBasti:
@coderabbitai review

hermeto/core/package_managers/general.py
```diff
@@ -117,7 +117,6 @@ async def _async_download_binary_file(
 
 async def async_download_files(
     files_to_download: dict[str, Union[str, PathLike[str]]],
```
eskultety: **Commit message:**

> There is no such a thing like controling concurrency in
> asynchronous programming

I think this is a bold statement which I don't think is correct. I think you might be mistaking concurrency for parallelism, there's plenty of sources online defining the meaning of each, one such could be: https://docs.oracle.com/cd/E19455-01/806-5257/6je9h032b/index.html

tests/unit/package_managers/test_rpm.py
```diff
@@ -587,7 +587,6 @@ def test_download(
                 )
             ),
         },
-        5,
```
eskultety, line 590: We could actually increase this number to a more reasonable one.

tests/unit/package_managers/test_general.py
```diff
@@ -235,11 +235,9 @@ async def mock_download_binary_file(
         "file3": str(tmp_path / "path3"),
     }
 
-    concurrency_limit = 2
```
eskultety, line 238: Hmm, this was IMO too low for no reason unless we were fetching models from hugging face which isn't our case.

hermeto/core/package_managers/general.py
```diff
@@ -151,20 +150,6 @@ async def on_request_start(
...
-                try:
-                    await asyncio.gather(*done)
-                except FetchError:
-                    # Close retry_client if any request fails (other tasks can be running,
-                    # if a task is closed with the client open, an Warning is raised).
-                    await retry_client.close()
-                    for t in tasks:
-                        t.cancel()
-                    raise
-
```
eskultety, line 152: Oh, please don't. This actually has a great bit of meaning - you can't just launch (potentially) hundreds of download tasks and expect the code to be more performant, quite the contrary:

- the server side has finite resources -> you'll likely face rate limiting or connection rejections (stability issues -> regression)
- client-side bandwidth is limited -> your download tasks are going to be throttled which will most likely cause timeouts requiring more users to ultimately override the default timeout value in the config
- storage is slow -> downloads are most likely going to be buffered in memory before flushed to the non-volatile storage -> much bigger memory footprint (big problem in container environments!!!)

eskultety: At best we may need to rethink the default value for concurrency to actually increase the throughput somewhat.

hermeto/core/package_managers/general.py
```diff
@@ -119,36 +117,11 @@ async def async_download_files(
...
-        trust_env=True,
-    )
-
-    async with retry_client as session:
-        tasks: set[asyncio.Task] = set()
+    """Download multiple files asynchronously."""
+    tasks: set[asyncio.Task] = set()
 
+    # respect proxy settings and .netrc
+    async with aiohttp.ClientSession(trust_env=True) as session:
```
eskultety: IIUC you just dropped any retry for a failed async download in this commit, correct?

hermeto/core/config.py
```diff
@@ -20,9 +20,6 @@ class Config(BaseModel, extra="forbid"):
     gomod_download_max_tries: int = 5
```
eskultety, line 20: > The change replaces aiohttp with httpx as the HTTP client library for
> async downloads.

This is e pretty vague explanation. What is the motivation behind switching libraries, what is it that aiohttp doesn't do that httpx does better, if anything `aiohttp` seems to be quite fast paced, hence maintained. What makes the `httpx` implementation compelling that we'd want to switch immediately?

It's a bit nicer to look at I totally give you that, but a quick skim through Google results told me that `httpx` may actually be bigger than `aiohttp` since it has both synchronous and asynchronous implementation which isn't a deal breaker for me, but `aiohttp` is heavily optimized specifically for `async` which is our case and hence will most likely be more performant for many concurrent tasks, again, our case. So if you want to make this switch, I'm afraid I'll need to see some performance benchmarks to know we're actually not regressing in performance significantly. Again, readability is absolutely important, but only if it doesn't get too much in the way of performance (to be fair this is just a single file and 2-3 functions that are ugly).

tests/unit/package_managers/test_general.py
```diff
@@ -143,48 +142,6 @@ def test_extract_git_info(url: str, nonstandard_info: Any) -> None:
     assert general.extract_git_info(url) == info
 
 
-@pytest.mark.asyncio
-async def test_async_download_binary_file(
```
eskultety, line 148: To be fair, the fact you switched libraries doesn't on its own IMO make this test automatically disappear, in essence it should still be "applicable". The thing is, the test has always been heavily mocked and so I'm wondering what it is actually testing, so if you want to drop it, just do that in a different commit because I don't think it's much relevant here.

hermeto/core/package_managers/general.py
```diff
@@ -70,35 +71,17 @@ def download_binary_file(
...
-        ) as resp:
-            with open(download_path, "wb") as f:
-                while True:
-                    chunk = await resp.content.read(chunk_size)
-                    if not chunk:
-                        break
-                    f.write(chunk)
+        response = await client.get(url, timeout=timeout, follow_redirects=True)
+        response.raise_for_status()
+        Path(download_path).write_bytes(response.content)
```
eskultety: So you deliberately got rid of chunking, so IIUC when you try to download a hugging face model that is 8GB in size, you'd store it in memory in full and only once it's downloaded only then you'd flush it to the storage destination, is that assessment correct?

hermeto/core/config.py
slimreaper35: This switch to the httpx library was only to pass our CI. At that time it appeared to be a solution. That's why the commit message is missing "WHY" part. I will probably drop the commit altogether.

PR #1023, opened 2025-07-11
conan: Add initial overview

a-ovchinnikov:
This commit adds initial overview of Conan (C/C++ package manager) and briefly discusses how its support could be implemented.

Conan: https://conan.io/

Addresses: https://github.com/hermetoproject/hermeto/issues/1002


docs/design/conan.md
```diff
@@ -0,0 +1,373 @@
...
+as well as an option for hosting [private repositories][]. Conan is written in
+Python and could be used both as a stand-alone tool and as a library (however
+Python APIs are [considered experimental][]). Conan allows to use Python as
+a dependencies specification language as well which creates ample opportunities
+for arbitrary code execution during both collection and build time.
+
+Conan is [available via PyPI][] or [as rpm][].
+
+Technically Conan cannot download dependencies from anywhere except from a
+central repository, however it is possible to amend _conanfile.py_
```
a-ovchinnikov: Correction, from _a repository_, not the central one. I.e. it cannot just grab something from GitHub.

docs/design/conan.md
```diff
@@ -0,0 +1,373 @@
...
+in such a way that it would download a package from GitHub and would expose it
+via a local repository.
+
+Apparently it has nothing to do with a certain Cimmerian.
```
ben-alkov: :rofl:

eskultety: Probably not something that needs to be mentioned in a design doc :).

docs/design/conan.md
```diff
@@ -0,0 +1,373 @@
+# Adding Conan Support to Hermeto
+
+## Background
+
+[Conan][] is one of many package managers for C++. At the moment of writing
+it is not abandoned and supports quite a few C++ libraries. Support for this
```
eskultety, line 16: nitpick:
> Support for this PM was  inquired about by users

I only see an upstream issue for `vcpkg`, but not Conan. How many users were requesting this? In any case, the sentence itself doesn't IMO need to be part of a design doc.

docs/design/conan.md
```diff
@@ -0,0 +1,373 @@
...
+
+## Specifying dependencies
+
+Any project that uses Conan must define either _conanfile.txt_ or
+_conanfile.py_.  The former one is an ini-like list of dependencies and
+generators for config files for build systems in use, the latter one is a
+full-blown Python script allowing users to dynamically modify dependencies (and
+to execute arbitrary code).
+
+_conanfile.py_ could be defined by both package consumer and by recipes for
```
eskultety, line 43: This one sounds very problematic wrt/ arbitrary code execution and its prevention.

docs/design/conan.md
```diff
@@ -0,0 +1,373 @@
...
+related to various aspects of producing a package: `build()`, `generate()`,
+`requirements()` and others.  These methods are called when a user executes
+corresponding CLI commands, sometimes at rather unexpected moments. For
+instance, running `conan build . ` (a command which is supposed to build a
+recipe) results in `requirements()` and `build()` invocation (which is quite
+logical), and running `conan graph info .` (a command which is supposed to
+report a dependency graph) results in the same two invocations.
+
+There does not seem to be any provisions limiting what a piece of code could be
+doing, invocation of external tooling appears to be encouraged.  Thus the only
```
eskultety, line 135: > invocation of external tooling appears to be encouraged

What does this mean for us, can it mean that we'll need to start packaging these all kinds of tools in our image to get pretty much any C/C++ Conan-based application to complete prefetch successfully?

eskultety: If so, then it's concerning to say the least, because suddenly not only the image will get too big to maintain, our own hermetic build is going to get more difficult and all auxiliary CVE problems will be our problem to solve in our image to retain healthy state in various container registry image scan DBs.

docs/design/conan.md
```diff
@@ -0,0 +1,373 @@
...
+    conaninfo.txt  conanmanifest.txt  include  lib  license
+  ```
+
+There are very few environment variables which affect Conan, the one useful is
+CONAN_HOME. All other settings should be done via profile or global.conf residing
+in CONAN_HOME.
+
+Another option that is present in Conan, but which does not seem to be used in the
+wild is package lock. A package could be locked with `conan lock create <path>`.
+That would create a lock file similar to one below:
```
eskultety, line 341: This means that the whole Conan backend would be strictly dependent on the `--permissive` mode just because lockfiles are not common.

docs/design/conan.md
```diff
@@ -0,0 +1,373 @@
...
+Conan could be used as a library or as a binary. While it is lucrative to use
+it as a library I think we should not do that to ensure isolation between
+Hermeto and recipes which could potentially contain anything and would end up
+being executed effectively from within Hermeto.  While this won't eliminate the
+threat it at least would not also break Hermeto too easily.
+
+The most universal option for getting everything seems to be to let Conan
+build  everything non-hermetically first, then purge resulting binaries. This addresses
+the permitted variety of specifying dependencies and build dependencies. This seems
+to be the most feasible option since `build()` will be run on multiple occasions.
```
eskultety, line 392: The problem with this solution though is that due to the unlimited nature of `conanfile.py` recipes which can be chained, one cannot deterministically verify that the sources that got fetched as part of the `build` command have not been tampered with (i.e. patched dynamically) by any of the recipes without a deep analysis. Unless I missed something in your write-up, then just removing the binaries afterwards isn't a proof of anything and I'm afraid this is not a way.

eskultety (commented):
Just some top-level observations without diving deep into Conan's docs.

docs/design/conan.md
a-ovchinnikov: Yes, sorry, this will go away once the doc is a bit more mature.

docs/design/conan.md
a-ovchinnikov: Fair enough. Conan was inquired about in a Slack thread, it has not made it to a proper issue yet.

docs/design/conan.md
a-ovchinnikov: Oh yes, it absolutely is! This would require some sort of SBOM signalling at the very least. I am saying this with an assumption that there are users who are fully blocked on Conan and must use it no matter the cost. An alternative, as it was briefly discussed offline, is not to support Conan due to the extreme volatility of sources that it effectively enables.

docs/design/conan.md
a-ovchinnikov: If we are fine with sacrificing an unspecified (judging from what I observed low) number of corner-cases where fetch and build phases cannot be separated then we do not need to have a full-blown build image for just fetching. In my opinion the bigger problem are a few packages which rely on system-wide tools for dependencies installation. An example of those is xorg's `build()` method which contains instructions to invoke `apt` or `dnf` or some other system-level package manager and to install some libraries with it. This is something that I am willing to draw a line at. As for the image -- maybe the image could be split into several?

docs/design/conan.md
a-ovchinnikov: Yes (at least if central repository is used as the packages source).

docs/design/conan.md
a-ovchinnikov: No, unfortunately we cannot guarantee anything about conan packages. That's why I think that the only way of using it is by marking SBOM as "best effort" and communicating to users that this is the best they can get with conan.

docs/design/conan.md
eskultety: > An example of those is xorg's build() method which contains instructions to invoke apt or dnf or some other system-level package manager and to install some libraries with it. 

Yeah, this kinda breaks the logical isolation of our backends, not that it would be a technical problem, since our default container user is root, so we can do anything when it comes to system actions. That said, it's sketchy.

eskultety: > maybe the image could be split into several

I thought about this in the past, there's just one big issue which doesn't make that effort very feasible -  compound repositories! IOW you never know what type of application you're building, e.g. JS frontend with Python or Rust backend and you immediately need 3+ backends and toolchains installed in the Hermeto image. Some applications will build a single container image, those would be fine, but those that deploy in a distributed service-based environment would be impacted. The biggest impact, however, would be on layered pipeline solutions integrating with Hermeto which would have to employ logic to determine the "right" small image to use for prefetch in every scenario, but ultimately the moment a layered solution needs to consider the monolithic Hermeto image as a fallback for a complicated analysis then IMO the whole effort of breaking the image and the added maintenance burden isn't worth it.

docs/design/conan.md
eskultety: Honestly, without a having had a deep read of all the resource material there just seem to be too many "but"s with Conan for me:

- it would be an everlasting strictly permissive backend (not the biggest issue in my mind)
- unlimited python recipes coming from anywhere producing random binaries not captured in the resulting SBOM potentially tampering with downloaded sources; this is actually not even permissive, the resulting SBOM has IMO zero informative value
- potentially unlimited set of tools needed inside our image to run a smoke build to produce a fully buildable set of dependecies WILL increase the CVE surface for our container image that we'll have to take care of to comply with static analysis tools, IOW increased maintenance
- dynamic, i.e. runtime toolchain "enhancements" as a result of the Python recipes completely invalidate the premise of a static & vetted container image, IOW if you look at some container image static analyzers even if Hermeto were to pass those checks and hence be considered safe to use in a pipeline, the fact that those recipes can make literally any changes to the image (note our default user is root) renders those checks worthless because at that point it's a completely different container image then what one intended to use.

taylormadore (approved):
I agree that Conan doesn't appear to be a good fit for Hermeto right now. Let's merge this to document what we've learned and we can revisit Conan if things change in the future.

ben-alkov (approved):

docs/design/conan.md
```diff
@@ -0,0 +1,415 @@
...
+This document captures the decision to not support [Conan][] in Hermeto.  This
+decision is made basing on the fact that Conan allows and encourages arbitrary
+code execution at effectively every step in dependencies processing.  The
+document contains the necessary background that made Hermeto team strongly
+decide against supporting it.
+
+
+## Background
+
+[Conan][] is one of many package managers for C++. At the moment of writing
```
eskultety: s/moment of writing/time of writing   (multiple occurrences)

docs/design/conan.md
```diff
@@ -0,0 +1,415 @@
+# Adding Conan Support to Hermeto Should Never Happen
```
eskultety: nitpick: Do we want this strong wording - _...Should Never Happen_? I would just put _[REJECTED]_ in there and the reason being that while it may currently be the case that it's not feasible for us to spend time on this, the trend in various ecosystems is to move towards the direction we'd like so we may just need to wait a few years and I think that should be captured in the decision.

eskultety (approved):
ACK with a minor nitpick on the design title.

docs/design/conan.md
a-ovchinnikov: Fair enough, updated the heading and also added a couple of sentences at the end of the decision paragraph to keep the door open.

docs/design/conan.md
eskultety: resolving as we're not implementing this any time soon.

docs/design/conan.md
eskultety: resolving as we're not implementing this any time soon.

PR #1027, opened 2025-07-17
add binary platform filtering design

taylormadore:
This adds a small design for introducing platform filters to package manager backends that support prefetching binaries.

### DIFF
--- a/docs/design/multi-arch.md
+++ b/docs/design/multi-arch.md
@@ -1,0 +1,257 @@
+# Multi-Architecture Support in Hermeto
+
+## Overview
+
+This design introduces platform-aware dependency fetching to Hermeto. The goal is to allow users to filter pre-fetched binary dependencies (like RPMs, Python wheels or platform-specific Gems) based on platform selectors (e.g., architecture, OS). The result is a lean, accurate set of dependencies tailored for each target in a multi-architecture build.
+
+## Design Principles
+
+### Hermeto
+- **Source-First Posture**: Hermeto remains source-first. Binary prefetching must either be part of the design for a package manager backend (e.g., RPM) or specifically enabled by the user.
+- **Binary Filtering is an Optional Extension** The current behavior remains the default. All binary filtering and its associated input parameters are opt-in.
+- **Simple, User-Supplied Filters**: Filtering is performed based on explicit, user-provided platform parameters. Hermeto will not perform runtime platform detection or infer compatibility.
+- **Backend-Specific Selectors**: The input parameters for filtering may differ significantly between package managers. The design accommodates this by scoping platform selectors to each package manager backend.
+
+### Konflux
+- **Single Platform Per Hermeto Invocation**: In a Konflux multi-arch build pipeline, each Hermeto invocation will target exactly one platform, producing one prefetch artifact and one SBOM. These artifacts will then be propagated to the build for that platform, ensuring natural separation between different architectures. Konflux is outside the scope of this design and is only mentioned here to understand a targeted use-case for this feature.
+
+
+## Proposed Implementation
+
+The implementation is centered around a new, optional, and backend-specific input model for specifying binary filtering options. The output structure and SBOM format remain unchanged.
+
+### Input Models & CLI
+
+- Optional platform fields are encapsulated within a `binary` submodel for relevant package types. This approach provides a clear namespace for binary-related options and allows for future extensions without cluttering the top-level package input.
+- Platform fields are handled at the PackageInput level rather than globally because the fields needed to filter/select prefetched binaries may differ between package managers
+- The `allow_binary` option will be renamed (deprecated) and moved to the `binary` submodel, but will be retained for backwards compatibility
+- Platform fields will accept a list to allow the flexibility of multiple filters
+- Platform fields can use Literals for input validation for small, stable sets like Python implementations
+
+```python
+class BinaryModeOptions(pydantic.BaseModel):
+    enabled: bool = False  # Renamed from allow_binary
+
+class PipBinaryOptions(BinaryModeOptions):
+    arch: Optional[list[str]] = None
+    os: Optional[list[str]] = None
+    py_version: Optional[list[str]] = None  # e.g., ["39", "310", "311"] (no dots)
+    py_impl: Optional[list[str]] = None  # e.g., ["cp", "pp", "py"]
+
+class BundlerBinaryOptions(BinaryModeOptions):
+    platform: Optional[list[str]] = None
+
+class RpmBinaryOptions(pydantic.BaseModel):
+    # binary.enabled is not relevant for RPM
+    arch: Optional[list[str]] = None
+
+class PipPackageInput(_PackageInputBase):
+    type: Literal["pip"]
+    allow_binary: bool = False  # Deprecated by binary.enabled, but kept for compatibility
+    binary: Optional[PipBinaryOptions] = None
+
+    @pydantic.model_validator(mode="after")
+    def _handle_legacy_allow_binary(self):
+        # If allow_binary is True and binary options are not set,
+        # initialize binary options with enabled=True.
+        # Issue a deprecation warning.
+
+class BundlerPackageInput(_PackageInputBase):
+    type: Literal["bundler"]
+    allow_binary: bool = False  # Deprecated by binary.enabled, but kept for compatibility
+    binary: Optional[BundlerBinaryOptions] = None
+
+    @pydantic.model_validator(mode="after")
+    def _handle_legacy_allow_binary(self):
+        ...
+
+class RpmPackageInput(_PackageInputBase):
+    type: Literal["rpm"]
+    binary: Optional[RpmBinaryOptions] = None
+```
+
+#### CLI Usage
+
+Complex input is provided by an input JSON string. This is the only way to specify binary options.
+
+```bash
+# Simple invocation (no binaries)
+hermeto fetch-deps pip
+
+# With binaries enabled
+hermeto fetch-deps '{"type": "pip", "binary": {"enabled": true}}'
+
+# With platform filtering
+hermeto fetch-deps '{
+  "type": "pip",
+  "binary": {
+    "enabled": true,
+    "os": ["linux"],
+    "arch": ["x86_64"],
+    "py_version": ["39"],
+    "py_impl": ["py"]
+  }
+}'
+```
+
+### Platform Filters
+- When multiple values are provided for a single filter field, they are combined with **OR** logic
+- When multiple filter fields are provided, they are combined with **AND** logic
+
+**Example**
+```
+{
+    "type": "pip",
+    "binary": {
+      "enabled": true,
+      "os": ["linux", "macosx"],
+      "arch": ["x86_64", "arm64"],
+      "py_version": ["310", "311", "312"],
+      "py_impl": ["cp"]
+    }
+  }
+```
+  Matches wheels for:
+  - ✅ torch-2.7.1-cp310-cp310-manylinux_2_28_x86_64.whl (linux + x86_64 + py310 + cp)
+  - ✅ torch-2.7.1-cp311-none-macosx_11_0_arm64.whl (macosx + arm64 + py311 + cp)
+  - ✅ torch-2.7.1-cp312-cp312-manylinux_2_28_x86_64.whl (linux + x86_64 + py312 + cp)
+  - ❌ torch-2.7.1-cp39-none-macosx_11_0_arm64.whl (Python 3.9 not in version list)
+  - ❌ torch-2.7.1-cp313-cp313-win_amd64.whl (Windows not in OS list)
+
+
+### pip
+
+The pip backend will be updated to filter Python wheels based on user-provided platform selectors, falling back to source distributions (if available) when no matching binary is found.
+
+| Current Behavior (**When `allow_binary=true`**) | Proposed Changes (**When platform filtering**) |
+|---|---|
+| Always prefetch sdists when available | - |
+| Packages with prefetched wheels are specifically noted in the SBOM | - |
+| Prefetch all available wheels | Prefetch *wheels that match platform filters* |
+| **Fail** when a package has no matching distributions | **Warn** when a package has no matching distributions *due to platform filtering* |
+| - | Perform filtering *before* downloading from the registry |
+
+**Implementation Details**:
+- Hermeto will use the Python packaging library to [parse wheel filenames][] and do platform matching between the user-provided filters and [wheel tags][].
+- Filtering logic will match against the standard `{python tag}-{abi tag}-{platform tag}` tag format (e.g., numpy-1.24.3-cp311-cp311-linux_x86_64.whl). See [packaging platform tags][]
+  - The user-specified python implementation is a match if:
+    - The wheel python implementation from the python tag is either `py` or equal to the user-specified python implementation
+  - The user-specified python version is a match if:
+    - Equal to the wheel python version from the python tag
+    - Greater than the wheel python version from the python tag and the wheel abi from the abi tag is either `abi3` or `none`
+  - The user-specified os is a match if:
+    - The wheel platform tag is `any`
+    - Contained by the wheel platform tag
+  - The user-specified arch is a match if:
+    - The wheel platform tag is `any`
+    - Contained by the wheel platform tag
+
+**Platform Matching Examples**:
+- Example 1: Prefetching for Python 3.11 on Linux
+  ```json
+  {
+    "os": ["linux"],
+    "arch": ["x86_64", "aarch64"],
+    "py_version": ["311"],
+    "py_impl": ["cp"]
+  }
+  ```
+  Matches torch 2.7.1 wheels:
+  - ✅ `torch-2.7.1-cp311-cp311-manylinux_2_28_x86_64.whl` (linux + x86_64 + py311 + cp)
+  - ✅ `torch-2.7.1-cp311-cp311-manylinux_2_28_aarch64.whl` (linux + aarch64 + py311 + cp)
+  - ❌ `torch-2.7.1-cp310-cp310-manylinux_2_28_x86_64.whl` (Python 3.10 not in version list)
+  - ❌ `torch-2.7.1-cp311-none-macosx_11_0_arm64.whl` (macOS not in OS list)
+
+- Example 2: Prefetching for multiple Python versions on macOS ARM64
+  ```json
+  {
+    "os": ["macosx"],
+    "arch": ["arm64"],
+    "py_version": ["311", "312"],
+    "py_impl": ["cp"]
+  }
+  ```
+  Matches torch 2.7.1 wheels:
+  - ✅ `torch-2.7.1-cp311-none-macosx_11_0_arm64.whl` (macosx + arm64 + py311 + cp)
+  - ✅ `torch-2.7.1-cp312-none-macosx_11_0_arm64.whl` (macosx + arm64 + py312 + cp)
+  - ✅ `torch-2.7.1-cp310-none-macosx_11_0_arm64.whl` (macosx + arm64 + cp310 with abi=none works for py311/312)
+  - ❌ `torch-2.7.1-cp313-cp313t-macosx_14_0_arm64.whl` (Python 3.13 not in version list)
+  - ❌ `torch-2.7.1-cp312-cp312-manylinux_2_28_aarch64.whl` (Linux not in OS list)
+
+**Python Compatibility**:
+- Version format: `["311"]` for Python 3.11, `["39", "310"]` for Python 3.9 or 3.10
+- Implementation: `["cp"]` (CPython), `["pp"]` (PyPy), `["py"]` (generic)
+
+This filtering is unlikely to produce a single "best" wheel, but is likely to reduce the number of downloads significantly. See the [wheelios][] repository for the original investigation.
+
+**Related Work/Optional Extensions**:
+- Enable selection of specific packages to prefetch wheels for
+
+### bundler
+
+The bundler backend will be updated to filter platform-specific Gems based on user-provided platform selectors.
+
+| Current Behavior (**When `allow_binary=true`**) | Proposed Changes (**When platform filtering**)|
+|---|---|
+| Platform information for Gems is parsed from `Gemfile.lock` | - |
+| Always prefetch platform-agnostic (`platform="ruby"`) gems | - |
+| Platform-specific Gems are specifically noted in the SBOM | - |
+| Prefetch all platform-specific Gems in `Gemfile.lock` | Prefetch all platform-specific Gems *that match platform filters* |
+| - | Perform filtering *before* downloading from the registry |
+| - | **Warn** when a Gem is skipped entirely *due to platform filtering* |
+
+**Implementation Details**:
+- Hermeto will perform matching between the user-specified platforms and platform-specific Gems
+- The lockfile parser appears to return normalized values for platform, so to be safe, Hermeto will also likely need to use the [Gem:Platform][] class to perform normalization of the user specified platforms before any comparisons/matches can be made.
+
+**Platform Matching Examples**:
+- Example 1: Prefetching for x86_64 Linux (any libc variant)
+  ```json
+  {
+    "platform": ["x86_64-linux-gnu", "x86_64-linux-musl"]
+  }
+  ```
+  Matches nokogiri 1.18.8 gems:
+  - ✅ `nokogiri (1.18.8-x86_64-linux-gnu)`
+  - ✅ `nokogiri (1.18.8-x86_64-linux-musl)`
+  - ✅ `nokogiri (1.18.8)` platform="ruby" (pure Ruby fallback, always included)
+  - ❌ `nokogiri (1.18.8-aarch64-linux-gnu)`
+  - ❌ `nokogiri (1.18.8-java)`
+
+- Example 2: Prefetching for JRuby
+  ```json
+  {
+    "platform": ["java"]
+  }
+  ```
+  Matches nokogiri 1.18.8 gems:
+  - ✅ `nokogiri (1.18.8-java)` (exact match - special case: single-component platform)
+  - ✅ `nokogiri (1.18.8)` platform="ruby" (pure Ruby fallback, always included)
+  - ❌ `nokogiri (1.18.8-x86_64-linux-gnu)`
+  - ❌ `nokogiri (1.18.8-arm64-darwin)`
+
+### rpm
+
+The rpm backend will filter packages based on a user-provided architecture list, failing if a requested architecture is not available.
+
+| Current Behavior | Proposed Changes (**When platform filtering**) |
+|---|---|
+| `allow_binary=true` is assumed without being specified | - |
+| RPMs are grouped by arch in `rpms.lock.yaml`. Noarch is present in all groups | - |
+| RPM Architecture noted in SBOM PURLs | - |
+| Output structure is per-architecture | - |
+| Prefetch RPMs for all arches in `rpms.lock.yaml` | Prefetch RPMs for *user-requested* arches |
+| - | Perform filtering *before* downloading from the registry |
+| - | **Fail** if a requested architecture is not present in the lockfile |
+
+### JavaScript & Source-based Package Managers
+- **npm/yarn**: JS packages can publish prebuilt binaries using the [prebuildify][] tool. Separate investigation needed to assess impact to Hermeto.
+- **gomod/cargo**: These package managers are fundamentally source-based and do not have an equivalent concept of pre-compiled binary artifacts. Therefore, binary filtering options do not apply.
+
+
+[parse wheel filenames]: https://packaging.pypa.io/en/stable/utils.html#packaging.utils.parse_wheel_filename
+[wheel tags]: https://packaging.pypa.io/en/stable/tags.html
+[prebuildify]: https://github.com/prebuild/prebuildify
+[wheelios]: https://github.com/chmeliik/wheelios
+[packaging platform tags]: https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/
+[Gem:Platform]: https://github.com/rubygems/rubygems/blob/8ad4509f95c90cf9523a82ca917b6b842fd37132/lib/rubygems/platform.rb#L10


docs/design/multi-arch.md
```diff
@@ -0,0 +1,257 @@
...
+        ...
+
+class RpmPackageInput(_PackageInputBase):
+    type: Literal["rpm"]
+    binary: Optional[RpmBinaryOptions] = None
+```
+
+#### CLI Usage
+
+Complex input is provided by an input JSON string. This is the only way to specify binary options.
```
slimreaper35: Maybe we could add support for an input JSON file. It could improve our UX. Every time I need to use a more complex JSON object on the command line for `fetch-deps` command, I always make a typo somewhere.

The input becomes wild if a user wants to prefetch content, for example, for multiple pip packages in one repo.

docs/design/multi-arch.md
```diff
@@ -0,0 +1,257 @@
...
+  - ✅ torch-2.7.1-cp310-cp310-manylinux_2_28_x86_64.whl (linux + x86_64 + py310 + cp)
+  - ✅ torch-2.7.1-cp311-none-macosx_11_0_arm64.whl (macosx + arm64 + py311 + cp)
+  - ✅ torch-2.7.1-cp312-cp312-manylinux_2_28_x86_64.whl (linux + x86_64 + py312 + cp)
+  - ❌ torch-2.7.1-cp39-none-macosx_11_0_arm64.whl (Python 3.9 not in version list)
+  - ❌ torch-2.7.1-cp313-cp313-win_amd64.whl (Windows not in OS list)
+
+
+### pip
+
+The pip backend will be updated to filter Python wheels based on user-provided platform selectors, falling back to source distributions (if available) when no matching binary is found.
```
slimreaper35: Will we always download one sdist when wheels are enabled, like in the current implementation?

docs/design/multi-arch.md
```diff
@@ -0,0 +1,257 @@
+# Multi-Architecture Support in Hermeto
```
eskultety: Having gone through the content I think ^this name may be a bit misleading and "Platform filtering for binary artifacts" might be a more descriptive name for the work IMO. The thing about multi-arch has more to do with pipeline solutions potentially consuming Hermeto rather than hermeto strictly caring about individual architectures. Note that multi-arch in context of Hermeto could also mean that we're about to start providing architecture specific builds of our container image which is not the case.

docs/design/multi-arch.md
```diff
@@ -0,0 +1,257 @@
...
+- **Binary Filtering is an Optional Extension** The current behavior remains the default. All binary filtering and its associated input parameters are opt-in.
+- **Simple, User-Supplied Filters**: Filtering is performed based on explicit, user-provided platform parameters. Hermeto will not perform runtime platform detection or infer compatibility.
+- **Backend-Specific Selectors**: The input parameters for filtering may differ significantly between package managers. The design accommodates this by scoping platform selectors to each package manager backend.
+
+### Konflux
+- **Single Platform Per Hermeto Invocation**: In a Konflux multi-arch build pipeline, each Hermeto invocation will target exactly one platform, producing one prefetch artifact and one SBOM. These artifacts will then be propagated to the build for that platform, ensuring natural separation between different architectures. Konflux is outside the scope of this design and is only mentioned here to understand a targeted use-case for this feature.
```
eskultety: > Konflux is outside the scope of this design and is only mentioned here to understand a targeted use-case for this feature.

Yep, which is also something I don't think we need to record in a design doc for platform filtering. The motivation for this could be mentioned in the commit message or the PR description instead as additional context, but we should remain agnostic in every decision we take.
The Konflux envisioning briefly mentioned here should live in the dependent work once our Hermeto part is complete :) .

docs/design/multi-arch.md
```diff
@@ -0,0 +1,257 @@
...
+
+```python
+class BinaryModeOptions(pydantic.BaseModel):
+    enabled: bool = False  # Renamed from allow_binary
+
+class PipBinaryOptions(BinaryModeOptions):
+    arch: Optional[list[str]] = None
+    os: Optional[list[str]] = None
+    py_version: Optional[list[str]] = None  # e.g., ["39", "310", "311"] (no dots)
+    py_impl: Optional[list[str]] = None  # e.g., ["cp", "pp", "py"]
```
eskultety: Do you think it would be a good idea if we could provide a default for the most common values? E.g. `cp` for implementation, `linux` for OS and `x86_64` for arch to alleviate the verbosity problem of our input JSON? Leaving Python version strictly defaulting to None. Would that be too much of a maintenance burden?

docs/design/multi-arch.md
```diff
@@ -0,0 +1,257 @@
...
+class BundlerBinaryOptions(BinaryModeOptions):
+    platform: Optional[list[str]] = None
+
+class RpmBinaryOptions(pydantic.BaseModel):
+    # binary.enabled is not relevant for RPM
+    arch: Optional[list[str]] = None
+
+class PipPackageInput(_PackageInputBase):
+    type: Literal["pip"]
+    allow_binary: bool = False  # Deprecated by binary.enabled, but kept for compatibility
```
eskultety: Let me create a Hermeto 1.0 milestone targeting removal of these deprecations.

docs/design/multi-arch.md
eskultety: Agreed, that would be a nice addition, although

>  I always make a typo somewhere

A file doesn't shield you from ^this I'm afraid, the only thing that can be omitted are an extra pair of Shell quotes. It's a JSON problem...

docs/design/multi-arch.md
eskultety: > falling back to source distributions (if available) when no matching binary is found

If a package was explicitly requested as binary (in the optional extension scenario) then we must not fall back to an sdist I think because that would IMO be confusing from UX POV - the user expects a wheel and we'd give them an sdist and so I'd be careful with this fallback in general. Yes, that's the current simple `allow_binary` behaviour, but once we introduce granular filtering then I think that gives us a right to be more strict with the checks.

docs/design/multi-arch.md
```diff
@@ -0,0 +1,257 @@
...
+### pip
+
+The pip backend will be updated to filter Python wheels based on user-provided platform selectors, falling back to source distributions (if available) when no matching binary is found.
+
+| Current Behavior (**When `allow_binary=true`**) | Proposed Changes (**When platform filtering**) |
+|---|---|
+| Always prefetch sdists when available | - |
+| Packages with prefetched wheels are specifically noted in the SBOM | - |
+| Prefetch all available wheels | Prefetch *wheels that match platform filters* |
+| **Fail** when a package has no matching distributions | **Warn** when a package has no matching distributions *due to platform filtering* |
```
eskultety: Why do want to default to a warning if the current simple behaviour results in an error?

docs/design/multi-arch.md
```diff
@@ -0,0 +1,257 @@
...
+| RPMs are grouped by arch in `rpms.lock.yaml`. Noarch is present in all groups | - |
+| RPM Architecture noted in SBOM PURLs | - |
+| Output structure is per-architecture | - |
+| Prefetch RPMs for all arches in `rpms.lock.yaml` | Prefetch RPMs for *user-requested* arches |
+| - | Perform filtering *before* downloading from the registry |
+| - | **Fail** if a requested architecture is not present in the lockfile |
+
+### JavaScript & Source-based Package Managers
+- **npm/yarn**: JS packages can publish prebuilt binaries using the [prebuildify][] tool. Separate investigation needed to assess impact to Hermeto.
+- **gomod/cargo**: These package managers are fundamentally source-based and do not have an equivalent concept of pre-compiled binary artifacts. Therefore, binary filtering options do not apply.
```
eskultety: I agree in principle, but don't forget about toolchains which are pre-compiled. Yes, we don't support this right now, but maybe that will have to change at some point, so let's try to be less descriptive in this case.

docs/design/multi-arch.md
```diff
@@ -0,0 +1,257 @@
...
+| `allow_binary=true` is assumed without being specified | - |
+| RPMs are grouped by arch in `rpms.lock.yaml`. Noarch is present in all groups | - |
+| RPM Architecture noted in SBOM PURLs | - |
+| Output structure is per-architecture | - |
+| Prefetch RPMs for all arches in `rpms.lock.yaml` | Prefetch RPMs for *user-requested* arches |
+| - | Perform filtering *before* downloading from the registry |
+| - | **Fail** if a requested architecture is not present in the lockfile |
+
+### JavaScript & Source-based Package Managers
+- **npm/yarn**: JS packages can publish prebuilt binaries using the [prebuildify][] tool. Separate investigation needed to assess impact to Hermeto.
```
eskultety: If this doesn't happen as part of the same work I think ^this will deserve a user facing docs update, just an FYI, no action here needed.

docs/design/multi-arch.md
```diff
@@ -0,0 +1,257 @@
...
+  - ❌ `torch-2.7.1-cp312-cp312-manylinux_2_28_aarch64.whl` (Linux not in OS list)
+
+**Python Compatibility**:
+- Version format: `["311"]` for Python 3.11, `["39", "310"]` for Python 3.9 or 3.10
+- Implementation: `["cp"]` (CPython), `["pp"]` (PyPy), `["py"]` (generic)
+
+This filtering is unlikely to produce a single "best" wheel, but is likely to reduce the number of downloads significantly. See the [wheelios][] repository for the original investigation.
+
+**Related Work/Optional Extensions**:
+- Enable selection of specific packages to prefetch wheels for
```
eskultety: This would be a very impactful change if incorporated to this work :).

docs/design/multi-arch.md
slimreaper35: Depends on editor or vim plugins :) I cannot live without autocompletion of parentheses, quotes, or autoindentation. (But that's not the point here.)

taylormadore: This is a feature I have been wanting as well, so I agree. In terms of planning that work, I lean towards filing a standalone issue for it. WDYT?

docs/design/multi-arch.md
taylormadore: Good point--I agree and will update it. I think we've been calling this feature "Multi-arch" for so long that I wasn't thinking about it :slightly_smiling_face:

docs/design/multi-arch.md
taylormadore: Yup, that's an excellent idea

docs/design/multi-arch.md
taylormadore: Having some sensible defaults (where they make sense) sounds reasonable to me +1

taylormadore: For the python implementation specifically, I'd probably lean towards defaulting to `["cp", "py"]` rather than `cp` alone

docs/design/multi-arch.md
taylormadore: Fair enough--as for making it less descriptive, I'd probably just focus on the current state with something like "These backends currently only manage source-based packages, so binary filtering does not apply. Adding support for prefetching other binary artifacts, like Go toolchains, is outside the scope of this design."

The main point of this was to convey that there is "nothing to do here right now"

docs/design/multi-arch.md
taylormadore: You raise a good point and I'm still pondering it myself. Are the binary options we provide for a package guaranteeing a binary or permitting one? In the case of pip itself (the tool, not the Hermeto backend), if there is no available or matching wheel for a package, it falls back to an sdist (if available).

taylormadore: Yes, I had planned on keeping the same behavior, but there is a related discussion here: https://github.com/hermetoproject/hermeto/pull/1027#discussion_r2218873226

docs/design/multi-arch.md
eskultety: > This is a feature I have been wanting as well, so I agree. In terms of planning that work, I lean towards filing a standalone issue for it.

Definitely! Specifying JSON as a file on the input isn't strictly tied to this work. We should probably link that issue with https://github.com/hermetoproject/hermeto/issues/533 to have some connection.

docs/design/multi-arch.md
eskultety: > For the python implementation specifically, I'd probably lean towards defaulting to ["cp", "py"] rather than cp alone

I don't feel very strongly about this, so I'll defer to your judgement. Are `["cp", "pp", "py"]` the only options it can take?

docs/design/multi-arch.md
eskultety: > In the case of pip itself (the tool, not the Hermeto backend), if there is no available or matching wheel for a package, it falls back to an sdist (if available).

That's true, but I think only to the point where the user would start supplying CLI options to pip:
```
  --no-binary <format_control>
                              Do not use binary packages. Can be supplied
                              multiple times, and each time adds to the
                              existing value. Accepts either ":all:" to
                              disable all binary packages, ":none:" to empty
                              the set (notice the colons), or one or more
                              package names with commas between them (no
                              colons). Note that some packages are tricky to
                              compile and may fail to install when this option
                              is used on them.
  --only-binary <format_control>
                              Do not use source packages. Can be supplied
                              multiple times, and each time adds to the
                              existing value. Accepts either ":all:" to
                              disable all source packages, ":none:" to empty
                              the set, or one or more package names with
                              commas between them. Packages without binary
                              distributions will fail to install when this
                              option is used on them.
  --prefer-binary             Prefer binary packages over source packages,
                              even if the source packages are newer.
```
Seems like `prefer-binary` is the default, however, the moment you allow wheel filtering in Hermeto then my understanding is that we'd be leaning more towards `--only-binary` and a fatal failure seems to be consistent with what pip would do if it could not find a suitable wheel to satisfy the request.

eskultety: If a user requested wheels then my understanding is they really DO NOT want sdists for whatever reason:

- sdist is not available
- building a wheel from the sdist is either not easily possible in their build infrastructure OR just extremely expensive

docs/design/multi-arch.md
eskultety: That's fair.

docs/design/multi-arch.md
slimreaper35: I think there was a discussion somewhere about SBOM components - in case we download wheels for a package, we add `pip_package_binary` property, but we still download one sdist.

eskultety: >  in case we download wheels for a package, we add pip_package_binary property, but we still download one sdist.

This is how `allow_binary` works, it doesn't necessarily have to mean that the new way of filtering wheels has to be strictly compatible with that (I might be wrong). The thing here is that the user requested a wheel and so, like I said, they must have had a reason for that to request it explicitly, so downloading an sdist to me feels like wasted bandwidth. I originally wanted to change this, but first - it would have been an incompatible change and second - I don't think our project at that time was mature enough when it comes to incorporating sustainable design decisions like we're trying to do today.
If we conclude that we'd like to follow pip's built-in install options more closely then we have a unique opportunity in this design to change our original approach.

slimreaper35: > This is how allow_binary works, it doesn't necessarily have to mean that the new way of filtering wheels has to be strictly compatible with that (I might be wrong). 

I know, it was more like a question of whether there is something that we can improve.

So, if we change the design and don't download one sdist when a user requests wheels, it should be solved.

docs/design/multi-arch.md
taylormadore: Filed #1035

docs/design/multi-arch.md
taylormadore: These are the common options: https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#python-tag

Many pure python projects distribute `py` wheels, which are implementation agnostic, so it seems sound to also  include them by default

eskultety: 2/5 sounds as a reasonable default (hopefully we won't regret having a default with this particular one) :) .

docs/design/multi-arch.md
taylormadore: Extending this to bundler--if the user specifies that they want to allow binaries for specific Gem(s), would we no longer fetch the pure ruby (platform: ruby) Gems also listed in the lockfile?

docs/design/multi-arch.md
taylormadore: Added

docs/design/multi-arch.md
taylormadore: Fixed

docs/design/multi-arch.md
taylormadore: Removed

docs/design/multi-arch.md
taylormadore: Explicitly mentioned that we do not currently support binaries with JS backends and that adding that support is out of scope for this design.

docs/design/multi-arch.md
taylormadore: Added to the base design--I think it makes sense to be included

docs/design/multi-arch.md
taylormadore: I think I intended this to apply to the sdist fallback, but since that is still being discussed--I'll hold off on fixing it until a decision is finalized

docs/design/binary-platform-filtering.md
```diff
@@ -0,0 +1,257 @@
...
+class BinaryModeOptions(pydantic.BaseModel):
+    enabled: bool = False  # Renamed from allow_binary
+    packages = Optional[list[str]] = None  # List of packages to allow binaries for
+
+class PipBinaryOptions(BinaryModeOptions):
+    arch: Optional[list[str]] = ["x86_64"]
+    os: Optional[list[str]] = ["linux"]
+    py_version: Optional[list[str]] = None  # e.g., ["39", "310", "311"] (no dots)
+    py_impl: Optional[list[str]] = ["py", "cp"]  # e.g., ["cp", "pp", "py"]
+
```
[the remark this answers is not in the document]
taylormadore: Use of default_factory for mutable defaults isn't necessary with pydantic models: https://docs.pydantic.dev/latest/concepts/fields/#mutable-default-values

(also this is really pseudocode in a document) :wink:

docs/design/multi-arch.md
eskultety: > Extending this to bundler--if the user specifies that they want to allow binaries for specific Gem(s), would we no longer fetch the pure ruby (platform: ruby) Gems also listed in the lockfile?

With the new proposal, yes, that would be my expectation. The situation is similarly messy even for bundler. Pasting our code for reference:

https://github.com/hermetoproject/hermeto/blob/555b00dbd8b21688676e37c44b5e4dc3995588c5/hermeto/core/package_managers/bundler/parser.py#L235-L252

I think this positions us well for a future removal of `allow_binary` in Hermeto v1.0. That would mean that `allow_binary` isn't fully deprecated (hmm can we even mark something as deprecated if the new functionality doesn't retain 1:1 parity? I guess we can as long as it remains functional), due to the new behaviour not being backwards compatible, but I'm not sure anyone would really miss a ton of fetched packages that don't end up being used during the build.

slimreaper35: AFAIK a lot of the gems only have the source distribution (ruby platform), so we would probably use a lot of fallbacks if a user requests all gems for one specific architecture.

We fetch everything that our custom parser script returns, which somehow resolves each gem's platform.

If you try to `feth-deps` in https://github.com/hermetoproject/integration-tests/blob/bundler/e2e, you notice that the only gem that is resolved with a specific platform is `nokogiri`, even though there is no ruby platform in the `Gemfile.lock`.

docs/design/binary-platform-filtering.md
```diff
@@ -0,0 +1,257 @@
...
+## Overview
+
+This design introduces platform-aware dependency fetching to Hermeto. The goal is to allow users to filter pre-fetched binary dependencies (like RPMs, Python wheels or platform-specific Gems) based on platform selectors (e.g., architecture, OS). The result is a lean, accurate set of dependencies tailored for each target in a multi-architecture build.
+
+## Design Principles
+
+- **Source-First Posture**: Hermeto remains source-first. Binary prefetching must either be part of the design for a package manager backend (e.g., RPM) or specifically enabled by the user.
+- **Binary Filtering is an Optional Extension** The current behavior remains the default. All binary filtering and its associated input parameters are opt-in.
+- **Simple, User-Supplied Filters**: Filtering is performed based on explicit, user-provided platform parameters. Hermeto will not perform runtime platform detection or infer compatibility.
```
chmeliik: > Hermeto will not ... infer compatibility

For python, this may be a dealbreaker. To download the same wheels that `pip` would consider by default, all users would need to pass something like this:

```jsonc
{
  "binary": {
    "os": ["linux", "any"],  // pip always considers the "any" platform as compatible, so I must always add it
    "arch": ["x86_64", "any"],  // or should I put "any" here?  both os and arch are part of the platform tag
    // I'm on py 3.13, but pip will also consider all lower versions of python, so I must specify them all
    "py_version": ["313", "312", "311", "310", "39", "38", "37", "..."],
    // Except pip will only consider the lower versions of python if the abi tag is "abi3" or "none"
    // So to make the filtering effective, I must also specify the correct abi and always add "abi3" and "none"
    "abi": ["cp313", "abi3", "none"],
    // I use CPython, but the "py" implementation is always compatible, so I must always add it
    "py_impl": ["cp", "py"],
  }
}
```

Without specifying all of these, the vast majority of builds will have to fall back to building from sdists (assuming we're still unconditionally fetching sdists). Each project distributes a wildly different set of wheels. If pip didn't have the logic to fall back to compatible platforms, hardly any set of dependencies could be resolved with just wheels.

eskultety: @chmeliik would it help if we made (and documented) `any`, `abi3`, `none` and `py` as defaults that are **always** preset and cannot be disabled? With ABI it will get tricky, but I think based on the python version I think we could derive the native abi denominator.

chmeliik: +1, that would be nice.

> With ABI it will get tricky, but I think based on the python version I think we could derive the native abi denominator.

Yeah, it should be doable. And ideally also treat the py_version is an upper bound rather than an exact version?

It's not that complex to implement, I believe this code snippet does it all: https://github.com/chmeliik/wheelios/blob/c924ac6dac69c47e5d49727d4400269225efce2a/filterwheels.py#L34-L52

To me it just sounded like the design was specifically calling out that Hermeto would not have any additional logic beyond matching exactly what the user specifies

taylormadore: @chmeliik Yeah, you're absolutely right--it got complicated with pip. I'll replace that line. What I really want to emphasize is that Hermeto will remain ignorant of what platform it is executing on and will only use user-input and backend defaults to make decisions.

chmeliik: :+1: feel free to resolve this, I don't have that power

docs/design/multi-arch.md
taylormadore: Incorporated this as discussed. The understanding should be captured in the "Binary Fetching Strategies" section

docs/design/multi-arch.md
taylormadore: Fixed

docs/design/binary-platform-filtering.md
```diff
@@ -0,0 +1,279 @@
...
+- Platform fields may have sensible defaults for the most common use cases. The existing `allow_binary` behavior must be maintained until that field is removed.
+- Introduce an optional packages field within the binary model. This field accepts a list of package names, acting as an allow-list for binary prefetching. If this list is provided, only packages named in it will be considered for binary downloads; all others will be treated as source-only. This option applies only to backends like pip and bundler where binary fetching is optional.
+
+```python
+class BinaryModeOptions(pydantic.BaseModel):
+    enabled: bool = False  # Renamed from allow_binary
+    packages: Optional[list[str]] = None  # List of packages to allow binaries for
+
+class PipBinaryOptions(BinaryModeOptions):
+    arch: Optional[list[str]] = ["x86_64"]
```
a-ovchinnikov: What would `None` mean in the context of architecture?

docs/design/binary-platform-filtering.md
```diff
@@ -0,0 +1,279 @@
...
+- Platform fields are handled at the PackageInput level rather than globally because the fields needed to filter/select prefetched binaries may differ between package managers
+- The `allow_binary` option will be renamed (deprecated) and moved to the `binary` submodel, but will be retained until the next major version for backwards compatibility
+- Platform fields will accept a list to allow the flexibility of multiple filters
+- Platform fields can use Literals for input validation for small, stable sets like Python implementations
+- Platform fields may have sensible defaults for the most common use cases. The existing `allow_binary` behavior must be maintained until that field is removed.
+- Introduce an optional packages field within the binary model. This field accepts a list of package names, acting as an allow-list for binary prefetching. If this list is provided, only packages named in it will be considered for binary downloads; all others will be treated as source-only. This option applies only to backends like pip and bundler where binary fetching is optional.
+
+```python
+class BinaryModeOptions(pydantic.BaseModel):
```
a-ovchinnikov: A usage example from L83 makes me think that this is not just options, but rather filters. WDYT about reflecting this in class names? E.g. something like `BinaryFilteringOptions`.

docs/design/binary-platform-filtering.md
```diff
@@ -0,0 +1,279 @@
...
+### Binary Fetching Strategies
+The combination of the enabled and packages fields within the binary model allows users to select from three distinct fetching strategies. These strategies are analogous to pip's binary-related flags and apply only to package managers where binary fetching is optional (e.g., pip, bundler).
+
+**No Binaries (Analogous to pip install --no-binary)**
+Configuration: binary.enabled: false (or the legacy allow_binary: false).
+
+Behavior: Hermeto operates in a source-only mode. No binary artifacts will be prefetched for any packages. This is the default behavior if no binary options are specified.
+
+**Prefer Binaries (Analogous to pip install --prefer-binary)**
+Configuration: binary.enabled: true with no packages list defined.
```
a-ovchinnikov: This feels a little error-prone: if a user thinks about restricting binary-allowed packages set and then forgets to provide the list they will get all packages. I wonder if it would make sense to add an explicit glob-like entry meaning `all packages`? Something that will never intersect with any package name, like `*`. This could also remove the need for `enabled`: `*` would mean yes to all, `None` would mean none and a list of packages would mean just those packages.

docs/design/binary-platform-filtering.md
taylormadore: As currently proposed, a `None` value for any filter essentially means no filter and would not reduce the result set based on architecture

docs/design/binary-platform-filtering.md
taylormadore: I had originally wanted to keep the namespace generic enough so that it could be extended for other hypothetical non-filter binary options in the future, but they are indeed all filtering-related right now so I see your point :thinking:

docs/design/binary-platform-filtering.md
a-ovchinnikov: Thank you, this makes sense. I think it would be easier to understand if the class were named `PipBinaryFilters` or something like that: that would have left no room for doubt -- `None` stands for no filters on architecture.

taylormadore: That sounds reasonable to me. Is it enough to rename the classes or do you think it's also unclear from a user-input perspective?

a-ovchinnikov: No, I think just classes names could be updated. An additional idea is to have some glob-looking value to mean `all`, but that could prove to be rather unwieldy implementation-wise, so I won't insist on it. (I.e. something like `Union[None, Literal["all"], list[str]]`)

docs/design/binary-platform-filtering.md
eskultety: Missing `code` formatting for the options in the description.

eskultety: Not a bad idea, since we're introducing a brand new control mechanism. I'd just slightly adjust the proposal with `:all:, :none:, <comma separated list>` (we might be able to parse the value with the packaging library? Haven't actually checked) instead of globbing which would further emphasize our compatibility with https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-only-binary. It would also be compatible with poetry (should we ever consider it for a backend): https://python-poetry.org/docs/configuration/#installerno-binary

docs/design/binary-platform-filtering.md
```diff
@@ -0,0 +1,279 @@
...
+    @pydantic.model_validator(mode="after")
+    def _handle_legacy_allow_binary(self):
+        ...
+
+class RpmPackageInput(_PackageInputBase):
+    type: Literal["rpm"]
+    binary: Optional[RpmBinaryOptions] = None
+```
+
+#### CLI Usage
```
eskultety: nitpick: this section splits the mental flow a bit IMO since the next section defines the strategies. It might be better to swap the order of the two, nonblocking though, feel free to ignore.

docs/design/binary-platform-filtering.md
eskultety: > (I.e. something like Union[None, Literal["all"], list[str]])

If we strictly keep this as a string, provided there's a `packaging` primitive to infer the control.

docs/design/binary-platform-filtering.md
taylormadore: Would we apply this pattern to all of the filters--not just package names? If so:

- I think we should drop the enabled and just assume enabled if the binary field contains anything
- All filters should default to `:all:` in the binary model (still thinking about this in terms of the current "fetch everything" approach vs. our desire for some more specific defaults with pip)
- I'm not sure that `:none:` makes any sense in this context. If the user wants no binaries, they shouldn't populate the binary model

eskultety: > I think we should drop the enabled and just assume enabled if the binary field contains anything

That's fair.

> All filters should default to :all: in the binary model

Yep.

> I'm not sure that :none: makes any sense in this context. If the user wants no binaries, they shouldn't populate the binary model

Right. No, I got carried away by being overly pip-compliant, but ^this makes total sense.

> (still thinking about this in terms of the current "fetch everything" approach vs. our desire for some more specific defaults with pip)

Will reply to ^this in a different thread.

eskultety: > (still thinking about this in terms of the current "fetch everything" approach vs. our desire for some more specific defaults with pip)

Hmm, truth to be told I'm still kinda undecided on the "prefer binary" mode because this means that our current behaviour will continue being carried over even to new major releases and I'm not sure I'd want that because I don't think it makes that much sense. From a user's POV they have the following options:

- build entirely from sources if possible
- build entirely from wheels if possible
- build from sources with the exception of a few packages

We pretty much cleanly support only the first option with the second one being a hybrid between the 1st and 3rd (which we currently don't have standalone). That would of course mean that we'd need separate logic for `allow_binary` to retain compatibility but which we'd eventually ditch in a major release. I'm sorry making this argument now despite having had a private meeting on it, it was easy for me to get tangled up in the details with all the modes and filtering logic. That said, I can be swayed to the current proposal, i.e. by carrying the current "prefer binary" mode even to future major releases with a strong argument.

taylormadore: The "`prefer-binary`" mode is helpful for good user experience because it avoids forcing users into a difficult all-or-nothing choice with just the strict "source-only" mode and a strict "binary-only" modes.

Imagine a project with 50 dependencies. The user wants binaries, so they enable the strict "binary-only" mode. However, one of those dependencies doesn't distribute a wheel. The prefetch would fail. To fix this, the user would have to manually list all 49 other dependencies individually in the packages filter to tell Hermeto to fetch binaries for them, while excluding the one problematic package.

The "`prefer-binary`" mode would be a "best-effort" mode. It would try to prefetch binaries for all packages but gracefully fall back to downloading an sdist when it can't find a wheel.

I'm OK with not **_also_** prefetching sdists for packages where we can find a wheel.

eskultety: > The "prefer-binary" mode is helpful for good user experience because it avoids forcing users into a difficult all-or-nothing choice with just the strict "source-only" mode and a strict "binary-only" modes. Imagine a project with 50 dependencies....

Yep, good argument.

> The "prefer-binary" mode would be a "best-effort" mode. It would try to prefetch binaries for all packages but gracefully fall back to downloading an sdist when it can't find a wheel.
> 
> I'm OK with **not also** prefetching sdists for packages where we can find a wheel.
> 

That's very much acceptable! To be fair not only am I motivated to be more pip-compliant but more importantly I'd like to reduce the bandwidth needed in cases like you described above resulting in our current "fetch absolutely everything you can find" approach :) .

docs/design/binary-platform-filtering.md
```diff
@@ -0,0 +1,272 @@
...
+
+**No Binaries (Analogous to pip install `--no-binary`)**
+Configuration: binary field unspecified (or the legacy `allow_binary: false`).
+
+Behavior: Hermeto operates in a source-only mode. No binary artifacts will be prefetched for any packages. This is the default behavior if no binary options are specified.
+
+**Prefer Binaries (Analogous to pip install `--prefer-binary`)**
+Configuration: binary field specified with no packages list defined.
+
+Behavior: Hermeto will attempt to prefetch compatible binaries for all dependencies where possible. In this mode, it will also prefetch a corresponding source distribution if there is no binary available to serve as a fallback.
```
eskultety: > if there is no binary available

nitpick: I'd make this part in bold.

eskultety (approved):
I don't have further comments on this design, looks very solid.

a-ovchinnikov (approved):

docs/design/binary-platform-filtering.md
taylormadore: Done :+1:

PR #1028, opened 2025-07-17
Add RPM modularity label

slimreaper35:

### DIFF
--- a/hermeto/core/models/property_semantics.py
+++ b/hermeto/core/models/property_semantics.py
@@ -21,6 +21,7 @@ class PropertyEnum(str, Enum):
     PROP_MISSING_HASH_IN_FILE = f"{APP_NAME}:missing_hash:in_file"
     PROP_PIP_PACKAGE_BINARY = f"{APP_NAME}:pip:package:binary"
     PROP_PIP_PACKAGE_BUILD_DEPENDENCY = f"{APP_NAME}:pip:package:build-dependency"
+    PROP_RPM_MODULARITY_LABEL = f"{APP_NAME}:rpm_modularity_label"
     PROP_RPM_SUMMARY = f"{APP_NAME}:rpm_summary"
 
     def __str__(self) -> str:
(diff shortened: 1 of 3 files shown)


hermeto/core/models/property_semantics.py
```diff
@@ -38,61 +39,64 @@ class Property(pydantic.BaseModel):
 class PropertySet:
     """Represents the semantic meaning of the set of Properties of a single Component."""
 
+    # sorted alphabetically
```
a-ovchinnikov: This comment (and an identical one above) does not add anything to explain the intent behind the code below. It should be either "Keep this list sorted for XYZ reasons" or removed.

a-ovchinnikov (commented):
LGTM with one minor nitpick. Will approve once tests are passing.

hermeto/core/models/property_semantics.py
slimreaper35: Removed

eskultety:
**Depends on: https://github.com/hermetoproject/hermeto/pull/1031**

hermeto/core/models/property_semantics.py
```diff
@@ -21,6 +21,7 @@ class PropertyEnum(str, Enum):
     PROP_MISSING_HASH_IN_FILE = f"{APP_NAME}:missing_hash:in_file"
     PROP_PIP_PACKAGE_BINARY = f"{APP_NAME}:pip:package:binary"
     PROP_PIP_PACKAGE_BUILD_DEPENDENCY = f"{APP_NAME}:pip:package:build-dependency"
```
eskultety, line 23: **Commit message**:

> This enables proper tracking of modular RPM packages and their
> relationships
> which is important for accurate dependency resolution in
> modular repositories.

Okay, this is essentially true, but we're stating what the RPM header tag is designated for, not why we're adding this and so it is a bit out of context IMO.
From my POV there are 2 reasons we're adding this:

1.  `modularity_label` is guaranteed to be in the RPM tags for RPM modules, so providing this additional information about an RPM that in fact comes from a module is reasonable from SBOM's accuracy POV
2. there are some automated static analysis checks out there running on container images cross checking the container image contents with the SBOM we produce causing the tests to choke because they don't have anything to match the modularity data they've extracted against.

eskultety (approved):
Minor wording change in the commit message, ACK.

PR #1030, opened 2025-07-18
vcpkg: Adds initial overview

a-ovchinnikov:
This commit adds initial overview of vcpkg (C/C++ package manager) and briefly discusses how its support could be implemented.

vcpkg: https://learn.microsoft.com/en-us/vcpkg/

### DIFF
--- a/docs/design/vcpkg.md
+++ b/docs/design/vcpkg.md
@@ -1,0 +1,159 @@
+# Adding Cargo support to Hermeto
+
+## Background
+
+[vcpkg][] is a C/C++ package manager maintained by Microsoft. At the moment of
+writing it is not abandoned and supports quite a few (about 2600)  C++
+libraries. Support for this PM was inquired about by users.  vcpkg supports
+multiple build systems, provides both default registry as well as allows
+setting up custom ones, provides binaries and sources, [the documentation][]
+specifically mentions air-gapped environments which in theory maps nicely to
+hermetic builds concept. Despite being advertised the underlying feature
+necessary for hermetic builds is [considered experimental][] at the moment of
+writing.
+
+While vcpkg is available through at least DNF installing it from the system
+package manager resulted in a partially working version due to auxiliary
+scripts not being present in expected location. [The recommended way][] of
+doing this appears to be by running a bootstrap script shipped via GitHub which
+either downloads a release from GitHub or builds vcpkg locally.
+
+vcpkg relies on json and cmake files to describe a package. Compiled binaries
+reside within [vcpkg installation tree][].
+
+vcpkg does not seem to provide any lock file mechanism besides pinning a
+dependency to a specific version.
+
+
+## Specifying dependencies
+
+vcpkg operates with a concept of a port: a versioned recipe for building some
+artifacts. Ports may depend on other ports and may have optional dependencies.
+Each port specifies some metadata about its package including, but not limited
+to name, version and dependencies, location of package source and build
+instructions.
+
+Every port contains a `portfile.cmake` with fetch and build instructions for a
+package and [vcpkg.json][] containing metadata about the package and its
+dependencies. Ports can carry patches to be applied to sources of packages.
+
+<details>
+  <summary>A typical vcpkg.json</summary>
+
+  ```
+  {
+  "name": "fmt",
+  "version": "11.0.2",
+  "port-version": 1,
+  "description": "{fmt} is an open-source formatting library providing a fast and safe alternative to C stdio and C++ iostreams.",
+  "homepage": "https://github.com/fmtlib/fmt",
+  "license": "MIT",
+  "dependencies": [
+      {
+        "name": "vcpkg-cmake",
+        "host": true
+      },
+      {
+        "name": "vcpkg-cmake-config",
+        "host": true
+      }
+    ]
+  }
+  ```
+
+</details>
+
+
+<details>
+  <summary>The relevant part of corresponding portfile.cmake</summary>
+
+  ```
+  vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO fmtlib/fmt
+    REF "${VERSION}"
+    SHA512 47ff6d289dcc22681eea6da465b0348172921e7cafff8fd57a1540d3232cc6b53250a4625c954ee0944c87963b17680ecbc3ea123e43c2c822efe0dc6fa6cef3
+    HEAD_REF master
+    PATCHES
+        fix-write-batch.patch
+        fix-pass-utf-8-only-if-the-compiler-is-MSVC-at-build.patch # remove in next release
+  )
+  ...
+  ```
+
+</details>
+
+Dependencies come from registries which could be of three types:
+
+ 1. The "built-in" registry (effectively a local copy of [vcpkg][] repository);
+ 1. A git registry -- any repository which follows certain layout rules and
+    contains ports;
+ 1. A directory maintaining certain structure.
+
+
+## Downloading dependencies
+
+vcpkg would work on a project created with it. Before using the tool for a new
+project [the official guide][] recommends acquiring vcpkg repository and
+pointing an environment variable to it. This location (`$VCPKG_ROOT`) will be
+used to store sources for dependencies. There is no separate command for just
+downloading sources, however `vcpkg install` accepts `--only-downloads` flag
+which does exactly that. When sources are downloaded they modify two locations:
+vcpkg root directory (effectively a location to which [vcpkg][] repository was
+cloned) and registry cache. The latter defaults to `$HOME/.cache/vcpkg/registries/`
+but could be retargeted with `$X_VCPKG_REGISTRIES_CACHE`. Once both caches are
+populated a package could be built by running `vcpkg install --no-downloads`.
+Setting registries cache apparently always results in re-downloading a registry
+(unless there is a `--no-downloads` flag present).  Note, that if any patches
+are present in a port they will be applied to sources at this time. Transitive
+dependencies appear to be handled as well.
+
+In case of necessity dependency graph could be computed and represented in a
+variety of formats (excluding json) by running `vcpkg depend-info fmt`.
+
+
+## Proposed solution
+
+Following the precedent set in other package managers vcpkg should be used as
+an external tool. It will need to be downloaded and bootstrapped first, then
+caches locations will need to be set, after which package's dependencies
+could be resolved.
+
+The main problem with vcpkg appears to be the lack of any reporting mechanism
+other than dumping build information to stdout:
+
+<details>
+  <summary>Sample dry-run install output</summary>
+
+  ```
+  $ vcpkg install --dry-run
+  Detecting compiler hash for triplet x64-linux...
+  Compiler found: /usr/bin/c++
+  The following packages are already installed:
+    * vcpkg-cmake:x64-linux@2024-04-23 -- git+https://github.com/microsoft/vcpkg@e74aa1e8f93278a8e71372f1fa08c3df420eb840
+    * vcpkg-cmake-config:x64-linux@2024-05-23 -- git+https://github.com/microsoft/vcpkg@97a63e4bc1a17422ffe4eff71da53b4b561a7841
+  The following packages will be rebuilt:
+      fmt:x64-linux@10.1.1 -- git+https://github.com/microsoft/vcpkg@dfe9aa860f5a8317f341a21d317be1cf44e89f18
+  The following packages will be built and installed:
+      zlib:x64-linux@1.3.1 -- git+https://github.com/microsoft/vcpkg@3f05e04b9aededb96786a911a16193cdb711f0c9
+  ```
+
+</details>
+
+While it contains enough data to populate a SBOM it has to be parsed which is
+inherently error-prone. There does not seem to be any way to get around this.
+
+Once packages are downloaded two environment variables will need to be set to
+account for the caches. Note, that this would create a tighter coupling between
+vcpkg used to download artifacts and one that will be used to build them.  It
+is not immediately clear how big this problem is, if it is a big problem then
+sources will have to be injected into a new $VCPKG_ROOT on a build system.
+
+
+[vcpkg]: https://github.com/microsoft/vcpkg
+[the documentation]: https://learn.microsoft.com/en-us/vcpkg/
+[The recommended way]: https://learn.microsoft.com/en-us/vcpkg/get_started/get-started?pivots=shell-bash#1---set-up-vcpkg
+[vcpkg installation tree]: https://learn.microsoft.com/en-us/vcpkg/reference/installation-tree-layout
+[vcpkg.json]: https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json
+[the official guide]: https://learn.microsoft.com/en-us/vcpkg/get_started/get-started?pivots=shell-bash
+[considered experimental]: https://learn.microsoft.com/en-us/vcpkg/concepts/asset-caching


docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+## Background
+
+[vcpkg][] is a C/C++ package manager maintained by Microsoft. At the moment of
+writing it is not abandoned and supports quite a few (about 2600)  C++
+libraries. Support for this PM was inquired about by users.  vcpkg supports
+multiple build systems, provides both default registry as well as allows
+setting up custom ones, provides binaries and sources, [the documentation][]
+specifically mentions air-gapped environments which in theory maps nicely to
+hermetic builds concept. Despite being advertised the underlying feature
+necessary for hermetic builds is [considered experimental][] at the moment of
```
eskultety, line 12: >  The X_VCPKG_ASSET_SOURCES environment variable, or
>  The --x-asset-sources command-line option.

Yeah, which means our implementation would have to remain experimental as well until ^this is resolved upstream, because if they dropped it, we'd have a broken, yet supported, implementation.

docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+        "name": "vcpkg-cmake",
+        "host": true
+      },
+      {
+        "name": "vcpkg-cmake-config",
+        "host": true
+      }
+    ]
+  }
+  ```
```
eskultety, line 64: This design seems to assume that the only operating mode vcpkg has includes ^these manifests files. The docs [mentions also a "classic" mode](https://learn.microsoft.com/en-us/vcpkg/concepts/manifest-mode) which, yes, while discouraged, it does exist and we'd likely need a way of detecting what structure a given project follows and fail for anything else than the "manifest mode":

> vcpkg has two operation modes: [classic mode](https://learn.microsoft.com/en-us/vcpkg/concepts/classic-mode) and manifest mode. For most users, we recommend manifest mode.

docs/design/vcpkg.md
eskultety: Also, IIUC https://learn.microsoft.com/en-us/vcpkg/users/assetcaching#x-azurl, more specifically, the
`x-azurl,file://<filesystem-location>[,,<rw>]` variant along with `x-block-origin` would be the **only** feasible option for us, correct?
What about git registries, would that be a feasible alternative for us? To set up a local-only git registry with the expected hierarchy structure?

docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+ 1. A directory maintaining certain structure.
+
+
+## Downloading dependencies
+
+vcpkg would work on a project created with it. Before using the tool for a new
+project [the official guide][] recommends acquiring vcpkg repository and
+pointing an environment variable to it. This location (`$VCPKG_ROOT`) will be
+used to store sources for dependencies. There is no separate command for just
+downloading sources, however `vcpkg install` accepts `--only-downloads` flag
```
eskultety, line 144: From the docs:

> ### --only-downloads
> When passed this option, vcpkg will run each build in the plan until it makes its first non-downloading external process call. 

I guess I'm unclear on what _first non-downloading external_ means, probably because I don't understand Cmake, IOW can `vcpkg` make an internal non-downloading call? I'm asking because I'm so far not 100% confident there's no arbitrary code execution based on the above sentence in context of execution of "build" plans. Maybe I'm just paranoid.

docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+which does exactly that. When sources are downloaded they modify two locations:
+vcpkg root directory (effectively a location to which [vcpkg][] repository was
+cloned) and registry cache. The latter defaults to `$HOME/.cache/vcpkg/registries/`
```
eskultety, line 147: I don't understand this bit, primarily because I haven't actually run `vcpkg`. Does ^this mean that a local copy of the default vcpkg git repository will always exist after running some `vcpkg` command?

docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+  The following packages will be rebuilt:
+      fmt:x64-linux@10.1.1 -- git+https://github.com/microsoft/vcpkg@dfe9aa860f5a8317f341a21d317be1cf44e89f18
+  The following packages will be built and installed:
+      zlib:x64-linux@1.3.1 -- git+https://github.com/microsoft/vcpkg@3f05e04b9aededb96786a911a16193cdb711f0c9
+  ```
+
+</details>
+
+While it contains enough data to populate a SBOM it has to be parsed which is
+inherently error-prone. There does not seem to be any way to get around this.
```
eskultety, line 219: Why do we need to rely on `stdout` for SBOM populating? If there's some registry data available locally, don't we get the cmake portfile recipe with every depedency? That would be a more reliable source for SBOM contents IMO.

docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+
+Once packages are downloaded two environment variables will need to be set to
+account for the caches. Note, that this would create a tighter coupling between
+vcpkg used to download artifacts and one that will be used to build them.  It
+is not immediately clear how big this problem is, if it is a big problem then
+sources will have to be injected into a new $VCPKG_ROOT on a build system.
```
eskultety: I'm unclear on this bit, can you elaborate? Do you mean similarly to cargo and some toolchain pinning, but in this case incompatible cache structuring?

docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+<details>
+  <summary>Sample dry-run install output</summary>
+
+  ```
+  $ vcpkg install --dry-run
+  Detecting compiler hash for triplet x64-linux...
+  Compiler found: /usr/bin/c++
+  The following packages are already installed:
+    * vcpkg-cmake:x64-linux@2024-04-23 -- git+https://github.com/microsoft/vcpkg@e74aa1e8f93278a8e71372f1fa08c3df420eb840
+    * vcpkg-cmake-config:x64-linux@2024-05-23 -- git+https://github.com/microsoft/vcpkg@97a63e4bc1a17422ffe4eff71da53b4b561a7841
```
eskultety, line 209: Since you're mentioning architectures, I think in context of https://github.com/hermetoproject/hermeto/pull/1027 it might be worth considering and documenting also the `install --allow-unsupported` option IIUC its architecture relevance.

docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+vcpkg would work on a project created with it. Before using the tool for a new
+project [the official guide][] recommends acquiring vcpkg repository and
+pointing an environment variable to it. This location (`$VCPKG_ROOT`) will be
+used to store sources for dependencies. There is no separate command for just
+downloading sources, however `vcpkg install` accepts `--only-downloads` flag
+which does exactly that. When sources are downloaded they modify two locations:
+vcpkg root directory (effectively a location to which [vcpkg][] repository was
+cloned) and registry cache. The latter defaults to `$HOME/.cache/vcpkg/registries/`
+but could be retargeted with `$X_VCPKG_REGISTRIES_CACHE`. Once both caches are
+populated a package could be built by running `vcpkg install --no-downloads`.
```
eskultety, line 149: So, what does a downloaded dependency really look like? I wasn't able to derive that from the docs - the cmake recipes mostly point to Github tarballs/zips of a particular version, does `vcpkg` unpack, patch and repack?

docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+Setting registries cache apparently always results in re-downloading a registry
+(unless there is a `--no-downloads` flag present).  Note, that if any patches
+are present in a port they will be applied to sources at this time. Transitive
```
eskultety, line 152: I think this "patches" bit deserves more visibility in the doc.

docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+In case of necessity dependency graph could be computed and represented in a
+variety of formats (excluding json) by running `vcpkg depend-info fmt`.
+
+
+## Proposed solution
+
+Following the precedent set in other package managers vcpkg should be used as
+an external tool. It will need to be downloaded and bootstrapped first, then
+caches locations will need to be set, after which package's dependencies
+could be resolved.
```
eskultety: So, since `vcpkg` is OSS (kudos Microsoft!), can we have a look at their implementation for inspiration, see what it is they do with `vcpkg install --only-download` and assess if reimplementing the fetching logic is totally impractical from our POV?

eskultety (commented):
I have to say, compared to Conan, `vcpkg` definitely gives me far less shivers.

docs/design/vcpkg.md
a-ovchinnikov: To the best of my understanding "classic" mode means manually invoking vcpkg to install select libraries into vcpkg-specific location, i.e. a slightly less global set of dependencies than a global install. These installed packages are not directly associated with any package. Having a project using vcpkg in classic mode would mean that a dependency list is stored in some script which wraps an invocation of `vcpkg install`. As long as we are limiting ourselves to just running manifest mode commands we should be safe.

docs/design/vcpkg.md
a-ovchinnikov: Sorry, I do not follow. Yes, we can mimic a registry by providing correct directory/files structure on the file system and pointing vcpkg to it.

docs/design/vcpkg.md
a-ovchinnikov: From my limited understanding of cmake, vcpkg scripts and some experiments a dependency from a json would be searched in a cache first then attempted to be downloaded if not found, cmake scripts behave the same way. The scripts have some guardrails  against running arbitrary code in download only mode: there is a check in [one of the internal scripts ](https://github.com/microsoft/vcpkg/blob/master/scripts/cmake/vcpkg_execute_required_process.cmake#L22) which is hit when I add random commands to a project file. I cannot tell now how hard it would be to tamper with cmake from within another script, tampering with these scripts within vcpkg root allowed me to continue processing the port file. Now, once vcpkg makes non-downloading call (in the simplest case I had a `message()` function) it halts processing of a portfile, however apparently proceeds with everything else. I tested this with a simple package with just one dependency in json, no code and several messages in the port file.

docs/design/vcpkg.md
a-ovchinnikov: No, a copy of registry will exist. This is essentially a huge map from port names to available port versions and git tree ids which correspond to port versions (git tree ids in vcpkg ports repository). To the best of my understanding it will be populated by any relevant command unless a `--no-download` option is present and passed.

docs/design/vcpkg.md
a-ovchinnikov: From my experiments the sources end up in `buildtrees` subdirectory in $VCPKG_ROOT in unpacked form with applied patches.

docs/design/vcpkg.md
a-ovchinnikov: We would need to not only re-implement fetch logic, but also provide correct repository structure for a vcpkg instance down the road. As always -- this is possible, but not very practical since we would need to maintain it. And since vcpkg relies heavily on cmake wouldn't that also mean that we would need to re-implement cmake functionality?

docs/design/vcpkg.md
a-ovchinnikov: My main concerns are transient dependencies (admittedly they seem to be somewhat rare), and any potential deviation which the tool makes from a list of versions and cares to report. Building a list of transitive dependencies is not that big of a task, but making sure that what was downloaded matches the spec is more complicated, that is why I reluctantly recommend parsing logs (while fully understanding how brittle this is).

docs/design/vcpkg.md
a-ovchinnikov: No. By default the sources end up in $VCPKG_ROOT which also contains port definitions, versions database and cmake scripts necessary for vcpkg to operate. If, for any reason, someone would use a clean version of vcpkg from, say, our image then we would at the very least need to link `buildtrees` within it to `buildtrees` generated during fetch. If, however, one could reuse effectively entire vcpkg root directory from the fetch phase then this is not a problem.

docs/design/vcpkg.md
eskultety: Which part specifically does need more clarification?

docs/design/vcpkg.md
eskultety: > that is why I reluctantly recommend parsing logs

This is too error prone and unstable to back our "accurate SBOM" claim I'm afraid. I think we need to consider a different approach.

docs/design/vcpkg.md
```diff
@@ -0,0 +1,159 @@
...
+vcpkg relies on json and cmake files to describe a package. Compiled binaries
+reside within [vcpkg installation tree][].
+
+vcpkg does not seem to provide any lock file mechanism besides pinning a
+dependency to a specific version.
+
+
+## Specifying dependencies
+
+vcpkg operates with a concept of a port: a versioned recipe for building some
```
eskultety, line 30: FWIW I'm missing the mention of (if it's relevant at all) different port types: https://learn.microsoft.com/en-us/vcpkg/concepts/ports#port-categories

docs/design/vcpkg.md
eskultety: Just to capture what I discussed over a private channel on this topic - could we maybe leverage CMake to create our own little cmake script that would import a given portfile.cmake and see if there's a way to access the signature (e.g. https://learn.microsoft.com/en-us/vcpkg/maintainers/functions/vcpkg_from_github) attributes/variables in CMake's context and dump them in a JSON/YAML format that would allow us to easily consume and process the crucial set of data?

docs/design/vcpkg.md
eskultety: > As long as we are limiting ourselves to just running manifest mode commands we should be safe.

Just for the record - how do we enforce this?

a-ovchinnikov: Bad wording, sorry. My understanding is that if a project has manifests present then it is in manifest mode. If the manifests are missing one is free to run  `vcpkg install foo` to install `foo` for this project and every other project that is to be processed by vcpkg and which is lacking manifests. In other words, there does not seem to be any specific automation for classic mode.
_Edit_: I have to mention another option that vcpkg provides: an integration whith a build system (e.g. cmake) which would trigger an update to shared dependencies _when the build is run_. There does not seem to be a way to pre-populate the cache other than install something manually.

docs/design/vcpkg.md
a-ovchinnikov: To the best of my understanding these types are not reflected in a registry in any explicit way. E.g. it is possible to learn that a port is a script port by looking at port contents and by parsing portfile contents, there is no flag that tells us that this is a script port. A bunch of cmake scripts and a lack of any download commands, on the other hand, indicate that this is a script port.

taylormadore (approved):
LGTM. The investigation was thorough and the design appears comprehensive.

docs/design/vcpkg.md
```diff
@@ -0,0 +1,385 @@
...
+Note, that `git-tree` hash is being replaced by `path` pointing to a directory
+containing a port relative to $VCPKG_ROOT. It should be possible to use
+absolute paths here too, but that has not resulted in a success so far. Another
+amendment that appears necessary is removing port version: for some reason
+versions other than 0 (the default one, assumed when port version is missing)
+result in vcpkg being unable to process the recipe. This procedure appears to
+be safe though because port version indicates how many changes have been made
+to a recipe itself after the last package version change, thus it exists mainly
+for internal bookkeeping. Nevertheless Hermeto could preserve the original
+values and report them in SBOM.
```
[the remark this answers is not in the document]
a-ovchinnikov: This claim is based on observed behavior, not on the documentation.

slimreaper35 (approved):

docs/design/vcpkg.md
```diff
@@ -0,0 +1,385 @@
...
+in a registry.  Once all dependencies are known `ports` directory must be
+traversed to extract data about versions (the latest available unless
+explicitly specified otherwise) and upstream location.
+
+`2. ` requires some rudimentary parsing of cmake scripts. [cmake_parser][]
+appears to be a good enough readily available solution for that given that
+cmake does not publish the language grammar. An alternative would be to produce
+either an ad-hoc parser with [PyParsing][] or to invest into cmake grammar
+specification and then use some standard parser generator. Preliminary
+experiments have shown that `cmake_parser` is good enough for the task.
```
eskultety, line 261: Hmm, a plausible approach to use the python package. That said, it looks like it's one of them abandonware, so we need to keep that in mind. Out of curiosity, have you tried to write a cmake native script that would try to parse the data? I think that might be a worth AI exercise these days and see what it spits out :) .

docs/design/vcpkg.md
```diff
@@ -0,0 +1,385 @@
...
+install any of the tools by itself, electing to fail with a warning about a
+missing tool, however it must be noted that CPython has two recipes for it in
+the official registry, so in theory one could introduce a Python dependency
+which will end up being resolved by downloading CPython sources and building
+them. Another common practice is reliance on an additional set of cmake
+functions distributed as a meta-port. Such shared scripts can provide a lot of
+source-modifying functionality and can provide wrappers for collecting sources.
+The latter is the way Qt5 ports are written, that is why it is highly likely
+that Qt5 will require separate treatment after the more common cases are
+supported.
```
eskultety, line 343: This paragraph gives me shivers. It looks like implementing this won't go without documenting serious limitations and out of scope handling, possibly making this backend as "best effort only" (or out right "forever experimental"?)

eskultety (approved):
A solid starting point.

docs/design/vcpkg.md
a-ovchinnikov: Possible, but I would rather not touch cmake unless strictly necessary. Even if all parsers, parser generators and parser libraries go extinct writing one for cmake from scratch would not be an excessively complicated task, the language itself is rather primitive.

docs/design/vcpkg.md
a-ovchinnikov: Yes, making Qt5 work without cmake would definitely be a feat in itself. This must be clearly communicated to users when we get to implementing vcpkg support.

eskultety: Depends....we already have one for yarn v1 lockfiles and we have to maintain it which is always an added burden. In that particular case it's a separate repo vs a few tens of lines of native JS code to do the same thing (yes, with some caveat that hermeto repo would have to emulate a fake JS project to do that which we did in the past anyway). Look at pybuild-deps, it's another of our responsibilities and we're not doing a particularly good job at maintaining it....we also use it as a library despite it being just a CLI tool. Anyway, comparing pros & cons is a discussion out of scope of this PR :) .

docs/design/vcpkg.md
eskultety: Depends....we already have one for yarn v1 lockfiles and we have to maintain it which is always an added burden. In that particular case it's a separate repo vs a few tens of lines of native JS code to do the same thing (yes, with some caveat that hermeto repo would have to emulate a fake JS project to do that which we did in the past anyway). Look at pybuild-deps, it's another of our responsibilities and we're not doing a particularly good job at maintaining it....we also use it as a library despite it being just a CLI tool. Anyway, comparing pros & cons is a discussion out of scope of this PR :) .

Issue #1019, opened 2025-07-09
Documentation for adding package managers is missing


a-ovchinnikov:
Currently adding a package manager starts with a somewhat free-form draft document outlining the general features of the packaging system to be supported and its shortcomings. It might be worth it to have a short document describing the main aspects of a packaging system to consider and investigate when adding support for it. This will help new contributors and speed up their contributions.

The document should mention the most crucial topics a PM author must cover: the general format of a package, whether there is a risk of arbitrary code execution during fetch and build phases, how the tool could be made work in network isolated environment.

brunoapimentel:
The [contributing guide](https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md#how-to-start-a-contribution) touches a lot of important aspects of this. But I do think it is probably too long, and it does not go straight to the point, so maybe we could rework it?

eskultety:
Much of the current contents would be better suited at something like `HACKING.md` to keep the controbuting guide focused and on point.

adambkaplan:
I drafted  #1032  to propose a "people process" around adding a package manager. This came out of the feedback from my  vibe-coded PoC for Maven support (#1020). A "big bang" PR with a new package manager is too challenging for the community to review and vet, plus we would like to have technical design docs to live alongside the code.

eskultety:
closing per #1032

PR #1032, opened 2025-07-18
doc: ADR for Package Manager Promotion

adambkaplan:
Decision record that creates a lifecycle policy for package managers. This policy in many ways expands Hermeto's mission from a supply-chain security tool to a catalyst for software supply chain best practices.

The policy is designed to be flexible enough to support a wide range of package managers, while also providing a clear path for package managers to graduate to full support. Space is made for the package manager's community to iterate on their own tooling for generating a reproducible list of dependencies. This ADR also makes clear that Hermeto is not responsible for verifying the correctness of the dependency list.

The community's current stance on how we support RPM/DNF may be impacted by this ADR. Much of the process for pinning RPM dependencies is not documented as the current process relies on external prototype tools. RPM support should be considered "incubating" in this lifecycle and have appropriate documentation (technical designs and end-user procedures).

Related to #1019 - though this doesn't touch on the specifics of implementing a package manager. This is more about driving the discussion and iterative development of adding a new package manager.

### DIFF
--- a/docs/adr/0003-package-manager-guide.md
+++ b/docs/adr/0003-package-manager-guide.md
@@ -1,0 +1,146 @@
+# Process for Adding Package Managers
+
+Date: 2025-07-18
+
+## Context
+
+Hermeto is intended to support a wide range of programming languages and dependency management
+systems. Now that we have a "critical mass" of package managers, the project should codify the
+requirements Hermeto needs to support a given package manager, and the "lifecycle" for evolving
+a package manager.
+
+The phases of the lifecycle are inspired by our experiences with [DNF](https://docs.fedoraproject.org/en-US/quick-docs/dnf/)
+and [Maven](https://maven.apache.org/), which currently do not have mature mechanisms for declaring
+and pinning dependencies.
+
+## Decision
+
+### Candidates for Package Manager Support
+
+Hermeto can support _any_ package manager that is able to generate a comprehensive list of uniqely
+identified dependencies, and has a process for persisting these dependencies locally to disk. The
+dependency list does not have to include a set of expected checksums for each item, however this
+practice is **strongly encouraged**. The dependecy list must include:
+
+* Direct dependencies of the artifact to be built.
+* Indirect (transitive) dependencies of the artifact.
+* Dependencies required for compilation that are external to the package manager. For example,
+  Maven plugins. Indirect/transitive dependencies of these compilation tools/packages must also be
+  identified.
+
+Hermeto is **not** responsible for generating or validating this dependency list.
+
+### Contribution and Promotion Process
+
+The Hermeto project will use a phased approach for accepting new package managers. The "lifecycle"
+of a package manager can be broken down into the following phases:
+
+1. Proposal
+2. Experimental
+3. Incubating
+4. Graduated
+
+#### Proposal Phase
+
+In the "proposal" phase, the candidate package manager is evaluated for project fit and
+feasibility. Those who wish to add a new package manager should submit an ADR describing the
+package manager and the high-level process developers should use to generate the dependency list.
+
+Maintainers at this stage should evaluate the following:
+
+   1. Can a developer generate a dependency list? Even an incomplete one?
+   2. What is the value to the community? Is the package manager worth maintaining?
+   
+Part of Hermeto's mission is to encourage software supply chain best practices in the software
+development lifecycle. Therefore maintainers should **not** consider limitations on how the
+dependency list is generated at this phase. Procedures that rely on proof of concept tools that
+have not been adopted "upstream" are fine.
+
+Package managers graduate to "Experimental" when the ADR is approved and merged.
+
+#### Experimental Phase
+
+During the experimental phase, the new package manager is added to the
+[_dev_package_managers](https://github.com/hermetoproject/hermeto/blob/0.31.0/hermeto/core/resolver.py#L27-L29)
+list. This disables the package manager by default. Contributors begin implementing the package
+manager by creating a new module within `hermeto.core.package_managers`, then declaring a main
+`fetch_*` function that obtains the package manager's dependencies.
+
+Experimental package managers do not need to work end to end, or have complete functionality. They
+_should_ warn users that the package manager is experimental and not guaranteed to produce a
+successful hermetic build.
+
+As the experimental package manager is iterated on, key technical decisions should be captured in a
+corresponding document in the `docs/design` folder. The document should eventually include the
+following:
+
+* Background context of the package manager - how it works, the typical "day in the life" of a
+  developer.
+* Procedure for generating the dependency list and/or checksum file(s). Links to external tools
+  should be provided if the package manager does not natively support the generation of such files.
+* Links to technical requirements/specifications if Hermeto needs to download files to a specific
+  directory tree structure, or supply other metadata.
+* Environment variables that need to be set so the package manager utilizes the files downloaded by
+  Hermeto.
+* Configuration files that need to be updated or generated so that the package manager builds with
+  the dependencies pinned by Hermeto.
+
+Package managers graduate to "Incubating" when the following criteria are satisfied:
+
+* The package manager fully implements the pre-fetch process:
+  * Downloads dependencies in a structure the package manager understands.
+  * Sets environment variables so that the package manager utilizes Hermeto's pre-fetched
+    dependencies.
+  * Generates or updates configuration files so that the package manager builds with Hermeto's pre-
+    fetched dependencies.
+* Maintainers are satisfied with the design document details.
+* Demonstration of a proof of concept hermetic build for the package manager. The PoC does not need
+  to use a "real" project - a "hello world" project with minimal dependencies is sufficient.
+
+#### Incubating Phase
+
+During the incubating phase, the package manager should still be disabled by default, but does not
+need to emit any "experimental" warnings. The Hermeto community should seek feedback from the
+package manager's community regarding the process and demonstrate hermetic builds with increasing
+levels of project complexity.
+
+Incubating package managers should use this phase to draft end user documentation (under the
+`docs/` directory). Documentation should instruct end users how to generate the dependency list and
+execute a hermetic build with Hermeto. These procedures can reference external tools that are proof
+of concept or otherwise not fully supported by the package manager's community. If such a tool is
+utilized, the documentation should indicate these procedures are subject to change at any time.
+
+Incubating package managers are allowed to make breaking changes. If a breaking change is
+introduced, it should be documented in a release note.
+
+Incubating package managers are promoted to "Graduated" when the Hermeto community agrees the
+following criteria are satisfied:
+
+1. The process for generating the dependency list is supported by the underlying package manager's
+   community - either directly or through a well-maintained external party tool.
+2. A developer with reasonable package manager experience can use Hermeto to pre-fetch dependencies
+   by reading the project documentation.
+2. A sufficient number of demonstration/pilot projects have been able to use Hermeto to execute a
+   hermetic build with the package manager.
+
+#### Graduated Phase
+
+Graduated package managers are moved from the `_dev_package_managers` set to the full set of
+supported package managers. At this point the package manager is enabled by default and suitable
+for general use.
+
+Graduated package managers can only be removed with a major version update to Hermeto (ex: v1 to
+v2). User-facing details such as configuration values/flags can be deprecated and rendered non-
+functional with a minor version update, but cannot be removed without a major version update. Any
+deprecation must include a release note announcing the change.
+
+## Consequences
+
+* Contribution guidelines will need to be updated to codify the process/lifecycle in this ADR,
+  including:
+  * Iterative design documentation alongside code changes.
+  * Documentation requirements for promoting/graduating package managers.
+* Our current support for RPM/DNF should be documented (technical designs and end-user procedures).
+* The Hermeto project should create a design document template used when contributors add (and
+  iterate on) package manager support.
+* Hermeto's release notes process must be able to announce deprecations and removals.


ben-alkov:
@adambkaplan; I'm sure there's more for us humans to discuss, but in terms of CodeRabbit's review - no lies detected. I fully agree with its findings.

ben-alkov:
@adambkaplan; Please lint with markdownlint using [our config](https://github.com/hermetoproject/hermeto/blob/9bdaf9be729539e0b6bee82bb749bb253fbfd9fa/.markdownlint-cli2.yaml), or simply copypasta/cherry pick https://github.com/hermetoproject/hermeto/commit/a41b57a56641bd79c1bffcbc6a6077c6e468f4ac from my fork - no need for credit, took me about 3 minutes and markdownlint did all of the work :grin:  (except for - no `.` at the end of list items; - no `:` introducing lists. It's a style thing, but neither are really necessary).

P.S. We don't run CI on the "ADR" and "design" dirs because there's no real point to reformatting old ADRs and designs.

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
+# Process for Adding Package Managers
+
+Date: 2025-07-18
+
+## Context
+
+Hermeto is intended to support a wide range of programming languages and dependency management
+systems. Now that we have a "critical mass" of package managers, the project should codify the
+requirements Hermeto needs to support a given package manager, and the "lifecycle" for evolving
```
ben-alkov: A "given" package manager, or a "net-new" package manager?

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+v2). User-facing details such as configuration values/flags can be deprecated and rendered non-
+functional with a minor version update, but cannot be removed without a major version update. Any
+deprecation must include a release note announcing the change.
+
+## Consequences
+
+* Contribution guidelines will need to be updated to codify the process/lifecycle in this ADR,
+  including:
+  * Iterative design documentation alongside code changes.
+  * Documentation requirements for promoting/graduating package managers.
```
ben-alkov: Do we mean "*Documentation* requirements" or "*Document the* requirements"

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+* Our current support for RPM/DNF should be documented (technical designs and end-user procedures).
+* The Hermeto project should create a design document template used when contributors add (and
```
ben-alkov: Nitpick

```suggestion
* The Hermeto project should create a design document template for contributors to use when they add (and
```

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+
+The phases of the lifecycle are inspired by our experiences with [DNF](https://docs.fedoraproject.org/en-US/quick-docs/dnf/)
+and [Maven](https://maven.apache.org/), which currently do not have mature mechanisms for declaring
+and pinning dependencies.
+
+## Decision
+
+### Candidates for Package Manager Support
+
+Hermeto can support _any_ package manager that is able to generate a comprehensive list of uniqely
```
ben-alkov:
```suggestion
Hermeto can support _any_ package manager that is able to generate a comprehensive list of uniquely
```

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+identified dependencies, and has a process for persisting these dependencies locally to disk. The
+dependency list does not have to include a set of expected checksums for each item, however this
+practice is **strongly encouraged**. The dependecy list must include:
```
ben-alkov:
```suggestion
practice is **strongly encouraged**. The dependency list must include:
```

ben-alkov (commented):
Preliminary stuff. I'll go over the language of "Candidates for Package Manager Support" in toto once we resolve mardownlinting.

docs/adr/0003-package-manager-guide.md
eskultety: > _any_ package manager

nitpick: I'd add "package manager ecosystem" here because it doesn't need to be the package manager tool itself directly that takes care of the list but it can be a different tool officially maintained with the actual package manager, something like Clippy as a linter for Rust.

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+
+* Direct dependencies of the artifact to be built.
+* Indirect (transitive) dependencies of the artifact.
+* Dependencies required for compilation that are external to the package manager. For example,
+  Maven plugins. Indirect/transitive dependencies of these compilation tools/packages must also be
+  identified.
```
eskultety: We cannot mandate this under "must", because not all external toolchains need to come from the same ecosystem. What I'd emphasize here instead is the need to either make sure that all necessary toolchains are:

- listed under the same dependency list if applicable
- listed under a different dependency list where appropriate
- pre-installed in the target base image if none of the 2 ^above can be satisfied (otherwise the hermetic build would fail)

docs/adr/0003-package-manager-guide.md
eskultety: > The dependecy list must include

nitpick: I'd prepend _'For a given package manager ecosystem to be even considered for Hermeto support, ...'_ here

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+
+Hermeto is **not** responsible for generating or validating this dependency list.
```
eskultety: > or validating this dependency list

This is true, but I'm wondering how important it is to mention that IFF an ecosystem provides a way of validation of such a list (e.g. integrity check), we'll run that as part of our process. This can only ever be best effort though since like you say - we don't usually have enough data to validate the contents.

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+development lifecycle. Therefore maintainers should **not** consider limitations on how the
+dependency list is generated at this phase. Procedures that rely on proof of concept tools that
+have not been adopted "upstream" are fine.
+
+Package managers graduate to "Experimental" when the ADR is approved and merged.
+
+#### Experimental Phase
+
+During the experimental phase, the new package manager is added to the
+[_dev_package_managers](https://github.com/hermetoproject/hermeto/blob/0.31.0/hermeto/core/resolver.py#L27-L29)
```
eskultety: `_dev_package_managers` is an internal implementation detail which may change in the future which would instantly invalidate ^this paragraph and what we should be trying to achieve with this doc is to provide a guideline containing high-level information that remains true (ideally not changing at all, but let's face it...) regardless of implementation changes we incorporate along the way. Therefore, I suggest rephrasing this; something along the lines of:
_The `hermeto.core.resolver.py` module is where package managers are enabled, please follow the established structure and comments on how to enable a new backend._
would IMO suffice, but that's just an example, I'm open to other wordings.

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+
+   1. Can a developer generate a dependency list? Even an incomplete one?
+   2. What is the value to the community? Is the package manager worth maintaining?
+   
+Part of Hermeto's mission is to encourage software supply chain best practices in the software
+development lifecycle. Therefore maintainers should **not** consider limitations on how the
+dependency list is generated at this phase. Procedures that rely on proof of concept tools that
+have not been adopted "upstream" are fine.
+
+Package managers graduate to "Experimental" when the ADR is approved and merged.
```
eskultety: I don't think ADRs as a concept gained traction in our project, specifically due to its 'time record' nature. We kind of naturally converged (or rather returned back) to design docs that are hosted under the `docs/designs` directory. What is proposed in this paragraph isn't really what we follow in practice. It's true that we mandated full design docs prior for implementation to start, but we lifted that hard requirement a while ago as it was deemed impractical. Instead:
- we expect an RFE in form of a "live" design doc starting with what you mention in ^this paragraph, ideally including the "day in the life" of a developer" bits for maintainers to familiarize themselves quickly with the area more swiftly
- upon initial approval ^this very doc then continues to evolve with more technical details, documentation & other material references; PoC is proposed, etc.
- this is already where development already started and the design doc keeps being updated with new gotchas, pitfalls and findings to ultimately become "a record" of how & why this was implemented

The tricky part based on ^this is that the line between a proposal and an experimental phase isn't very clear, but the thing is - does it need to be? Do we need to be stiff about this and overprocess the development? I personally don't think so and so I think a more looser guideline here is more beneficial.

eskultety: Also, given that everything is tracked in git, we don't need to submit further ADRs, it turns out to be much more practical from dev perspective to just update the design doc along with any other code changes we adopt in the future for a particular backend.

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+  * Sets environment variables so that the package manager utilizes Hermeto's pre-fetched
+    dependencies.
+  * Generates or updates configuration files so that the package manager builds with Hermeto's pre-
+    fetched dependencies.
+* Maintainers are satisfied with the design document details.
+* Demonstration of a proof of concept hermetic build for the package manager. The PoC does not need
+  to use a "real" project - a "hello world" project with minimal dependencies is sufficient.
+
+#### Incubating Phase
+
```
eskultety: We currently don't have a notion of an "incubating phase". What would it look like in practice strictly from development perspective? Let me elaborate some more. Currently all development happens on the main branch for its inherent simplicity, ease of local testing (remember, you need to specify the backend on the CLI to do anything with it), guaranteed frequent rebases, etc. which is in contrast with what many other projects do where features are developed on a separate branch which is then merged to main and which is also where this experimental vs incubating distinction makes more sense. We strictly follow linear history in our repository. It's not impossible to achieve the same with feature branches, but it's much easier to get out of sync, making the final rebase potentially tedious and painful in the end. And so we took the easy path here.
The thing I'm struggling to envision is how ^this phase would translate to real world in terms of how a feature is consumed, how does one tell from a single beta CLI flag in what stage a given backend is and whether they even want to give it a try?
This is a matter of our CLI design where you need to specify a backend to even test some code path locally. I guess along with `x-<pkgmgr>` we could ultimately keep the `--dev-package-managers` flag specifically for this (even though we didn't really plan keeping it around), but I think the important question to answer here is - What value does the incubating phase distinction bring over experimental vs supported? Unarguably I'm seeing this in a very binary way, because to me "experimental" in general means:

- is the implementation complete?
- is the functionality on par with other similar features?
- are we confident about the approach taken? Have we covered all/most design pitfalls?
- has all dependent work that were to happen first in some other project that we need to consume for a particular piece of implementation finished and was released?

If the answer is NO to any of the above questions then a feature is not production ready whether one wants to call it experimental or incubating :), hence my position on this. Not sure how much it helps in your PR, but I think documenting the experimental status carefully is probably all we need.

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+for general use.
+
+Graduated package managers can only be removed with a major version update to Hermeto (ex: v1 to
+v2). User-facing details such as configuration values/flags can be deprecated and rendered non-
+functional with a minor version update, but cannot be removed without a major version update. Any
+deprecation must include a release note announcing the change.
+
+## Consequences
+
+* Contribution guidelines will need to be updated to codify the process/lifecycle in this ADR,
```
eskultety: One thing I should have mentioned at the very beginning but kept for last (I'm sorry) since you have an explicit note here in your proposal - I believe this whole ADR should be converted into a development guideline that I can imagine either as a standalone document OR as part of a comprehensive HACKING.md document, I'd say TBD, but for an ADR this feels slightly too much and ultimately based on the expectation and the information already provided here we'd be ultimately duplicating it.

 If I were to contribute to a project for the first time I'd not bother with any ADRs until pointed out during the review process, one can ingest only so much information before making their first contribution and so guidelines is where our focus needs to be :) .

Now, with all ^that, I'd also like to emphasize that the fact this PR proposal is in form of an ADR in the first place is very likely a consequence of us not making it clear at all in our `docs` directory  what is what  and how new documentation content should really be proposed, since we have both `docs/adr` and `docs/design` hosting similar content. So, @adambkaplan we're sorry for any confusion we may have caused.

docs/adr/0003-package-manager-guide.md
```diff
@@ -0,0 +1,146 @@
...
+  including:
+  * Iterative design documentation alongside code changes.
+  * Documentation requirements for promoting/graduating package managers.
+* Our current support for RPM/DNF should be documented (technical designs and end-user procedures).
```
eskultety: Do we still need to do anything about this with https://github.com/hermetoproject/hermeto/pull/1024 having been merged?

docs/adr/0003-package-manager-guide.md
eskultety: > should create a design document template

Yeah, https://github.com/hermetoproject/hermeto/pull/601, shame on us.

docs/adr/0003-package-manager-guide.md
ben-alkov: > It's true that we mandated full design docs prior for implementation to start, but we lifted that hard requirement a while ago as it was deemed impractical

And very much a burden for contributors, and frankly anti-agile.

docs/adr/0003-package-manager-guide.md
ben-alkov: @eskultety; DYT we should delete 'docs/adr'? It will live on in git history, of course...

eskultety: First we need to tweak the files in there a bit to be moved to design docs and then we definitely can remove the dir :) .

eskultety: On a second thought, not very related to this PR, but on the topic of ADRs, we could use them as decision records on rejecting implementation of some ecosystem backend due to serious security & trust implications, e.g. arbitrary code execution with Conan #1023, just food for thought, but let's discuss specifics of that somewhere else.

brunoapimentel: On this topic, I think there's an overlap in what's proposed here with what's currently in our [contributing guide](https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md#how-to-start-a-contribution). That text was written in the very moment we were deciding to drop the requirement of a full design doc before any implementation for a package manager began, and I think it is outdated and/or could definitely use some improvement. 

So I'd say we should either replace/enhance the contributing guide with whatever is agreed upon here, or delete that part from there altogether (I do like the Hermeto's Ethos bit though).

docs/adr/0003-package-manager-guide.md
adambkaplan: I think it is fair to add this as requirement for us to graduate/GA a package manager ecosystem. If the dependency list provides checksums, Hermeto should validate it.

adambkaplan:
Based on the feedback here - I think it makes sense to re-work the content + discussion into the contributing guide itself, with a design doc template to help contributors add package managers (the main form of feature requests we'll receive).

docs/adr/0003-package-manager-guide.md
adambkaplan: I took my own fresh stab at a design doc template, as a separate commit in this PR. It's a bit more involved than @brunoapimentel 's idea because it contains a lot of prompting questions. The design doc is also intended to be iterated on over time.

adambkaplan:
I decided to draft a new design doc template (independent of #601) based on feedback here. I think it makes sense to drop the commits that modify the ADR and just keep the design doc template (which we can improve over time), with the link/update to the contributing guide.

Should I proceed to drop/removed the ADR, since it seems that we aren't going to use those?

adambkaplan:
@eskultety @brunoapimentel updated this PR, which still has the "ADR" element. I have otherwise squashed + rebased.

Should we keep the ADR, or should I drop it?

brunoapimentel:
> @eskultety @brunoapimentel updated this PR, which still has the "ADR" element. I have otherwise squashed + rebased.
> 
> Should we keep the ADR, or should I drop it?

The design template looks good to me. I think there's value in adding it even without the 'ADR for Package Manager Promotion' part. As for the ADR itself, I prefer to leave it to other folks with more context to comment on it.

In case the ADR is moved forward, I left one very small comment on it that I think is important (sorry in advance, I really don't want to prolong this any further).

docs/design/package-manager-template.md
```diff
@@ -0,0 +1,219 @@
...
+
+#### Network Requirements
+
+Describe network-related considerations:
+
+- **Registry endpoints**: URLs and APIs Hermeto needs to access
+- **Authentication**: Any authentication requirements for package registries
+- **Rate limiting**: Considerations for API rate limits
+- **Mirror support**: Support for alternative registries or mirrors
+
```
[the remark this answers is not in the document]
eskultety: We don't need to go overboard with this at the moment.

eskultety:
> @eskultety @brunoapimentel updated this PR, which still has the "ADR" element. I have otherwise squashed + rebased.
> 
> Should we keep the ADR, or should I drop it?

This has taken us extra long. @adambkaplan based on my comment above, I don't think the ADR concept stuck with this project so please remove ADR-specific bits. Other than that the thread on _Incubating phase_ is still open where I don't think we need to go to that depth especially since this guide is supposed to be updated with time. Otherwise, this has an ACK from me.

docs/adr/0003-package-manager-guide.md
adambkaplan: Apologies for not replying on this.

I ended up dropping the "incubating" phase as it doesn't fit well with the current community process. We have coalesced around the `x-` prefix to communicate the package manager is not GA/fully supported, which seems to get the job done.

adambkaplan:
Slimmed this down to just the design template and an update to the contributing guide. Thanks everyone!

docs/design/package-manager-template.md
```diff
@@ -0,0 +1,218 @@
...
+- Complete the "Implementation Notes" sections as desired, or when the package manager is ready to
+  be enabled by default.
+
+**Maintainers**:
+
+Design documentation review is meant to facilitate conversation and future improvements. Regular
+review _and merging_ is vital to keep the discussion and implementation moving forward. Aim for
+_progress_ over _perfection_ when reviewing and approving submissions.
+
+
```
eskultety: I think we can drop this paragraph, we're kind of already piling maintainer-relevant info to host centrally elsewhere instead of having maintainer notes here and there.

docs/design/package-manager-template.md
```diff
@@ -0,0 +1,218 @@
...
+- **Ecosystem considerations**: Features and discussion in the package manager ecosystem that may
+  impact Hermeto's implementation
+
+### Testing Strategy
+
+Describe how the package manager implementation is tested:
+
+- **Unit tests**: Key areas covered by unit tests
+- **Integration tests**: End-to-end testing scenarios
+- **Test data**: Sample projects and dependencies used for testing
```
eskultety: I don't think we need an explicit paragraph on this in the design template, tests are an integral part of any technical PR, so we woudln't accept just about any technical contribution without a reasonable coverage discussed as part of the review process. If it's not the case currently, any mention of tests should go to CONTRIBUTING.md instead.

docs/design/package-manager-template.md
```diff
@@ -0,0 +1,218 @@
...
+
+- **Official documentation**: Links to package manager documentation
+- **Specifications**: Relevant technical specifications or RFCs
+- **Community resources**: Forums, mailing lists, or chat channels
+- **Related tools**: Other tools in the ecosystem
+
+## Changelog
+
+Optional - mark key changes to this document with dates.
+
```
eskultety: This can be dropped as well, ^this practice would simulate wiki kind of behaviour where every change is accompanied with a changelog entry for bookkeeping purposes, we have git for that. This is even more emphasized by the fact that these docs are not part of our user facing docs.

docs/design/package-manager-template.md
```diff
@@ -0,0 +1,218 @@
...
+Describe any files that Hermeto should generate or provide to the package manager. This will form
+the basis of the `inject-files` implementation. A tree diagram can be helpful here:
+
+```
+hermeto-deps/
+├── [package-manager-name]/
+│   ├── metadata/
+│   ├── packages/
+│   └── manager-config.json
+```
```
eskultety, line 170: This sample tree doesn't resemble anything that we do as part of that command actually. While it will differ among backends, the common pattern worth mentioning might instead be:
```
project.git/
├── <depedencies_lockfile>  # e.g. requirements.txt, yarn.lock, packages.json <<< We may overwrite this
└── <package_manager_data>/ # e.g. .bundler, .yarn, .config directory <<< We may overwrite configuration here 
    ├── cache
    ├── config
    ├── some_dir
    └── some_file
```

docs/design/package-manager-template.md
```diff
@@ -0,0 +1,218 @@
...
+- Are there related tools that should be considered in or out of scope?
+- Is there any behavior or configuration that should be considered an "edge case"?
+
+### Dependency List Generation
+
+Hermeto needs a reliable list of dependencies to pre-fetch. This section documents the tools and
+procedure needed to generate this list. Hermeto is **not** responsible for generating the
+dependency list on its own.
+
+_Note: The subsections below are not required, but serve as a useful starting point_.
```
eskultety, line 71: The document is very comprehensive and detailed and so this tiny note can easily get lost in the surrounding information noise. I'd suggest marking each subsection with an 'optional' tag just like you did on L177 so that the fact a subsection is optional screams right at you from the heading.

Edit: multiple occurrences across the doc.

docs/design/package-manager-template.md
```diff
@@ -0,0 +1,218 @@
+# [Package Manager Name] Design Document
+
+This template is intended to help contributors add support for a new package manager (ecosystem).
+Yes it is big - please try not to be intimidated by its size! The sections and probative questions
+are based on the community's experiences adding new package manager ecosystems to Hermeto.
```
eskultety, line 5: I think we should also add a note either here or to the respective CONTRIBUTING section mentioning this template to also look at past designs that were successfully merged as a cross-reference.

eskultety (commented):
Comprehensive design template, yes, intimidating...yes, it has to be for what it represents, I like it.

eskultety:
@adambkaplan we'd like to merge this as the content is solid, but there are a few outstanding comments unresolved.

docs/design/package-manager-template.md
adambkaplan: Done!

docs/design/package-manager-template.md
adambkaplan: Done!

docs/design/package-manager-template.md
adambkaplan: Done!

docs/design/package-manager-template.md
adambkaplan: Done!

docs/design/package-manager-template.md
adambkaplan: Added a link in this document to this document with a note, "You may also use past [design documents](https://github.com/hermetoproject/hermeto/tree/main/docs/design) as inspiratoin"

adambkaplan:
Updated per feedback, and rebased.

PR #1036, opened 2025-07-23
toil(docs): replace pip example

ben-alkov:
Replace the 'atomic-reactor' pip example with a simpler example, which actually builds hermetically 🫤

This PR depends on having the example code in a repo under "hermeto/doc-examples" (currently in https://github.com/ben-alkov/doc-examples; I'll create that repo after this PR is ACKed().


docs/pip.md
```diff
@@ -716,97 +718,27 @@ export PIP_FIND_LINKS=/tmp/hermeto-output/deps/pip
...
 #### Build the application image
 
-We will base the final application image on our custom base image. The base
-image build installed all the RPMs we will need, so the final phase can use
-network isolation again 🎉. In order to support the network isolated build, we
-need to remember to `source` the environment file in the step that executes `pip
-install`. Because `osbs-client` comes from GitHub, the source code in
-`/src/atomic-reactor` has also been changed so that the dependencies are
-pointing to the cached versions.
+In order to support the network isolated build, we need to remember to `source`
```
taylormadore, line 736: Do we want to include the inject-files subcommand in this example (for completeness) even if it isn't strictly needed in this specific case (because there are no non-registry dependencies)?

The build.sh script in the pip-basic example uses the inject files command

docs/pip.md
```diff
@@ -692,10 +695,9 @@ See [the pip documentation][] for more details about running Hermeto for
 pre-fetching pip dependencies.
```
a-ovchinnikov: Github won't let me nitpick on L689 (since we are here anyway): "The steps for pre-fetching the dependencies s/is/are/ similar to before" -- is it me or is something missing between "to" and "before"? Would inserting "ones outlined" make it smoother? The part about using pip _this time_ might need to be amended as well.

docs/pip.md
ben-alkov: 6 of one, 1/2 dozen the other. I think I'd rather remove the command from the shell script: it's confusing to have a noop without calling it out, but then that itself is a little silly.

I fully intend to add a more complex example to the doc-examples repo at some point, which would (hopefully) demonstrate all of these features. For that, based on what I've seen so far, I will almost certainly need @a-ovchinnikov's help with building Rust from pip.

docs/pip.md
ben-alkov: Ah, yes, I noticed that 'wtf', but let it slide. I'll try to figure out what it's referring to.

docs/pip.md
```diff
@@ -669,16 +676,12 @@ In requirements.txt, specify the dependency [via an https url](#https-urls).
...
-Atomic-reactor already builds with Cachito (Hermeto's spiritual ancestor), which
-makes it a rare example of a Python project that meets Hermeto's requirements
-out of the box (see [the pip documentation][]).
+Let's build a [basic pip project][].
 
 Get the repo if you want to try for yourself
 
 ```shell
-git clone https://github.com/containerbuildsystem/atomic-reactor --branch=4.4.0
+git clone https://github.com/hermeto/doc-examples --branch=pip-basic
```
slimreaper35: Was that supposed to be a real project ? Because the URL does not exist.

ben-alkov: Not yet, but it will. Preview at https://github.com/ben-alkov/doc-examples/tree/pip-basic . It's in the PR notes.

docs/pip.md
ben-alkov: Pretty sure this is left over from when these examples were in 'usage.md'. Will rephrase.

docs/pip.md
```diff
@@ -669,36 +676,51 @@ In requirements.txt, specify the dependency [via an https url](#https-urls).
...
+The defaults are
+
+- "path": "."
+- package manager type provided as a CLI arg (pip, gomod, npm, etc.)
+
+e.g.
+
+```shell
+hermeto fetch-deps pip
+```
```
a-ovchinnikov, line 701: WDYT?

```diff
- The defaults are
-
- - "path": "."
- - package manager type provided as a CLI arg (pip, gomod, npm, etc.)-
-
- e.g.
- 
- ```shell 
- hermeto fetch-deps pip
- ```
+ Sources could be fetced with
+ ```shell 
+ hermeto fetch-deps pip
+ ```
+
+ Here `path` is not specified explicitly and defaults to `.` and a short-hand notation
+ for project manager is used.
```

If you accept this then L717-L723 would become redundant.

docs/pip.md
a-ovchinnikov: Rust extensions should just work as long as they do not require cargo>1.84.

ben-alkov:
OK, ready for re- review.

docs/pip.md
a-ovchinnikov: s/hermeto/hermetoproject

ben-alkov: ~Tha heq... I did this yesterday 👀~

derp

PR #1038, opened 2025-07-24
bundler: Extract pydantic model definitions from the parser module

eskultety:
Noticed by accident when reviewing https://github.com/hermetoproject/hermeto/pull/1027

### DIFF
--- a/hermeto/core/package_managers/bundler/gem_models.py
+++ b/hermeto/core/package_managers/bundler/gem_models.py
@@ -1,0 +1,191 @@
+import logging
+from functools import cached_property
+from pathlib import Path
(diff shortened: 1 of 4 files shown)


eskultety:
Looking at the model definitions it's odd that the models which should mainly be just static data containers define download methods which requires importing dependencies the module would otherwise not need. I was considering creating wrapping classes inside the `main.py` of the backend that would define the download methods in order to keep the model definition isolated and the functional logic inside `main.py`. Let me know if you'd prefer it that way and I'd add another commit to the series.

eskultety:
Polite ping, this is a trivial change.

PR #1043, opened 2025-07-29
Allow the selection of the container engine to be used in integration tests

brunoapimentel:
Introduces the `HERMETO_CONTAINER_ENGINE` environment variable that can be set to choose which container engine will be used to run integration tests. Currently, only `podman` and `buildah` are supported.

### DIFF
--- a/tests/integration/container_engine.py
+++ b/tests/integration/container_engine.py
@@ -1,8 +1,11 @@
+import json
 import logging
 import os
+import secrets
 import subprocess
 from abc import ABC, abstractmethod
-from typing import Any, Optional, Union
+from contextlib import contextmanager
+from typing import Any, Generator, Optional, Union
 
 log = logging.getLogger(__name__)
 
@@ -101,11 +104,112 @@ class PodmanEngine(ContainerEngine):
         return self._run_cmd(image_cmd)
 
 
+class BuildahEngine(ContainerEngine):
+    """Buildah engine."""
+
+    @property
+    def name(self) -> str:
+        """Get the name of the container engine."""
+        return "buildah"
+
+    @contextmanager
+    def _configure_buildah_container(self, image: str) -> Generator[str, None, None]:
+        """Configure buildah container.
+
+        Buildah requires a container to be explicitly created from an image before it can be used.
+        The built container then can be referenced by the name given when creating it.
+
+        :param image: The image to create a container from.
+        :return: The container name.
+        """
+        container_name = f"hermeto-test-container-{secrets.token_hex(6)}"
+        from_cmd = ["buildah", "from", "--name", container_name, image]
+        stdout, exit_code = self._run_cmd(from_cmd)
+
+        if exit_code != 0:
+            raise RuntimeError(f"Failed to create buildah container from image {image}\n{stdout}")
+
+        try:
+            yield container_name
+        finally:
+            self._run_cmd([self.name, "rm", container_name])
+
+    def _generate_cmd(
+        self,
+        container_name: str,
+        image: str,
+        cmd: list[str],
+        entrypoint: Optional[str] = None,
+        flags: Optional[list[str]] = None,
+    ) -> list[str]:
+        """Generate container run command.
+
+        Buildah does not automatically use the image's configured CMD or ENTRYPOINT when running a command.
+        This means that the run command needs to be generated manually considering the image's configuration
+        and the flags present in the original run command.
+
+        :param container_name: The container name to run the command on.
+        :param image: The image on which the command will be run.
+        :param cmd: The original command provided.
+        :param entrypoint: The entrypoint to be used for the image.
+        :param flags: The flags present in the run command.
+        :return: The generated command.
+        """
+        if flags is None:
+            flags = []
+
+        image_cmd, image_entrypoint = self._get_image_config(image)
+
+        # fallback to the image's default cmd if a custom one is not provided
+        cmd = cmd or image_cmd
+
+        # if an entrypoint flag is provided, use it
+        if entrypoint:
+            return ["run", *flags, container_name, "--", entrypoint, *cmd]
+
+        # if the image has an entrypoint, prepend it to the command
+        if image_entrypoint:
+            return ["run", *flags, container_name, "--", *image_entrypoint, *cmd]
+
+        # if no entrypoint is provided, use only the cmd
+        return ["run", *flags, container_name, "--", *cmd]
+
+    def _get_image_config(self, image: str) -> tuple[list[str], list[str]]:
+        """Parse entrypoint and cmd from image's JSON configuration."""
+        output, exit_code = self._run_cmd(["buildah", "inspect", image])
+
+        if exit_code != 0:
+            raise RuntimeError(f"Failed to inspect image {image}.")
+
+        parsed_output = json.loads(output)
+        docker_config = parsed_output.get("Docker", {}).get("config", {})
+        cmd = docker_config.get("Cmd", [])
+        entrypoint = docker_config.get("Entrypoint", [])
+
+        return cmd, entrypoint
+
+    def run(
+        self,
+        image: str,
+        cmd: list[str],
+        entrypoint: Optional[str] = None,
+        flags: Optional[list[str]] = None,
+    ) -> tuple[str, int]:
+        """Run command using buildah."""
+        with self._configure_buildah_container(image) as container_name:
+            generated_cmd = self._generate_cmd(container_name, image, cmd, entrypoint, flags)
+
+            return self._run_cmd([self.name, *generated_cmd])
+
+
 def get_container_engine() -> ContainerEngine:
     """Get the configured container engine."""
     engine_name = os.getenv("HERMETO_TEST_CONTAINER_ENGINE", "podman").lower()
 
     if engine_name == "podman":
         return PodmanEngine()
+
+    if engine_name == "buildah":
+        return BuildahEngine()
 
     raise RuntimeError(f"Invalid container engine: {engine_name}")


tests/integration/container_runtime.py
```diff
@@ -0,0 +1,62 @@
+import logging
+import subprocess
+from abc import ABC
+from typing import Any, Union
+
+log = logging.getLogger(__name__)
+
+
+def _run_cmd(cmd: Union[list[str], str], **subprocess_kwargs: Any) -> tuple[str, int]:
```
ben-alkov: Sheer curiosity on my part: Why isn't this a Class method?

ben-alkov:
EDITED: buildah is troublesome

I don't want to bog us down in nomenclature pedantry, but IIUC, *these* are container "runtimes"

- containerd
- CRI-O
- docker daemon
- gVisor
- Kata
- LXD
- runC

*these* are container "engines"

- ctr (containerd)
- crictl (CRI-O)
- docker
- Hyper-V
- podman
- rkt

although the terminology is still settling, and in colloquial usage, esp. docker and podman are referred to as either.

A quick Google backs this up to a first approximation, and there's a pretty good summary in https://sarusso.github.io/blog/container-engines-runtimes-orchestrators.html, paragraph "Definitions."

~This is actually **even more** relevant for buildah, which is technically _neither_ (but "engine" is closer than "runtime").~ I correct myself (forgot about `buildah run`): buildah is more than a builder, but still not exactly a runtime.

TL'DR - We should probably say (and implement) "Allow the selection of the container ***engine***"

tests/integration/container_runtime.py
brunoapimentel: I just forgot about it while fighting the Cursor spaghetti :sweat_smile:

ben-alkov:
The bot brings up a good point RE: using the `@abstractmethod` decorator

tests/integration/container_runtime.py
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    subprocess_kwargs.update(forced_options)
+    process = subprocess.run(cmd, **subprocess_kwargs)
+
+    return process.stdout, process.returncode
+
+
+class ContainerRuntime(ABC):
+    """Abstract base class for container runtimes."""
```
a-ovchinnikov: I am sorry, why is this class abstract? It seems to have just concrete methods.

tests/integration/utils.py
```diff
@@ -23,6 +22,8 @@
 from hermeto.core import resolver
 from hermeto.interface.cli import DEFAULT_OUTPUT
 
+from .container_runtime import get_container_runtime
```
a-ovchinnikov: A nitpick: do we do relative imports?

tests/integration/container_runtime.py
```diff
@@ -83,11 +85,138 @@ def __init__(self) -> None:
...
+    def _configure_buildah_container(self, image: str) -> str:
+        """Configure buildah container.
+
+        Buildah requires a container to be explicitly created from an image before it can be used.
+        The built container then can be referenced by the name given when creating it.
+
+        :param image: The image to create a container from.
+        :return: The container name.
+        """
+        container_name = f"hermeto-test-container-{secrets.token_hex(6)}"
```
ben-alkov: T.I.L. - very cool

tests/integration/container_runtime.py
```diff
@@ -83,11 +85,138 @@ def __init__(self) -> None:
...
+        :param flags: The flags to treat.
+        :return: The treated flags and the entrypoint value.
+        """
+        if flags is None:
+            return [], None
+
+        # find the index of the entrypoint flag, if it exists
+        index = next((i for i, flag in enumerate(flags) if flag.startswith("--entrypoint")), None)
+
+        if index is not None:
```
ben-alkov: `if index:`?

tests/integration/container_runtime.py
a-ovchinnikov, line 1: Nitpick:

*Commit message* in 4464aaf6d2

s/previouc/previous/

tests/integration/container_runtime.py
```diff
@@ -41,9 +41,17 @@ def __init__(self) -> None:
         """Initialize container runtime."""
         self.name = ""
 
-    def build_image(self, build_cmd: list[str], tag: str) -> tuple[str, int]:
+    def build_image(
+        self, flags: Optional[list[str]] = None, context_dir: Optional[StrPath] = None
```
a-ovchinnikov: Nitpick: `StrPath` is supertype of `str`, `str` is immutable. I think it is safe to just have `context_dir` default to `"."`

tests/integration/container_runtime.py
```diff
@@ -83,11 +85,138 @@ def __init__(self) -> None:
...
+                entrypoint_value = flags[index + 1]
+                return flags[:index] + flags[index + 2 :], entrypoint_value
+            # entrypoint flag exists but no value is provided
+            else:
+                raise RuntimeError(
+                    f"Unable to parse the run command flags: {flags}."
+                    "Check if the --entrypoint flag defines a value."
+                )
+
+        return flags, None
```
ben-alkov: I can't see why we're returning `None` here.

tests/integration/container_runtime.py
a-ovchinnikov: PEP8 absolutists recommend comparing to `None` explicitly (since `None` is the third option).

Maybe it would make more sense to have a positive check for `None` instead and return fast?
```python
if index is None:
    return flags, None
entrypoint_flag = flags[index]
...
```

tests/integration/container_runtime.py
```diff
@@ -83,11 +85,138 @@ def __init__(self) -> None:
...
+    ) -> tuple[str, int]:
+        """Run command using buildah."""
+        container_name = self._configure_buildah_container(image)
+
+        try:
+            generated_cmd = self._generate_cmd(container_name, image, cmd, flags)
+
+            return _run_cmd([self.name, *generated_cmd])
+        finally:
+            _run_cmd([self.name, "rm", container_name])
```
a-ovchinnikov: The base class defines `remove_image`. Would it make sense to have this expressed via `remove_image`? Also `remove_image` is used in ContainerImage for managing context. Would `"rmi", "--force"` work with buildah?

brunoapimentel: This is not removing the image, it is removing the work container that needs to be manually created (and removed) in buildah. Notice that the comand is `rm` instead of `rmi`. And `rm` does not have the `--force` flag, whereas `rmi` does.

tests/integration/container_runtime.py
brunoapimentel: Because stupid me thought making it abstract would prevent it from being instantiated (and thus from being used directly), but after a quick test, it seems this is not the case in Python.

So, I believe the abstraction makes sense, we have a base ContainerRuntime that defines the method interface, and that will be the only thing that is known outside of this module. All other runtimes should be specializations of this base class. The base class shouldn't be used directly, only the specialized classes.

Most methods are 100% reusable, with the exception of `run_cmd_on_image`. This one could be made abstract, but I kept it the way it is since it could very likely be used as is if we decided to go for a DockerRuntime. But it probably should just be abstract, on a second though. Maybe I'll also add a dumb method just to return the name of the runtime, since that is one of the major changes between each specialized class.

I'll post an update on this soon.

tests/integration/container_runtime.py
brunoapimentel: This function is intended to return the entrypoint of a given image + run flags, and remove it in case it was part of the flags. 

This means that if the image has an entrypoint, we need to return it. If the image does not have an entrypoint, but the flags are something like `--rm --entrypoint hermeto`, we also need to return `hermeto` as the intended entrypoint.

This line is the case in which neither the image nor the flags contain an entrypoint, so we need to return the flags as they were, and `None` as the entrypoint.

tests/integration/container_runtime.py
brunoapimentel: Fixed.

tests/integration/container_runtime.py
brunoapimentel: I'm not a PEP8 absolutist, so is `if index:` okay?

tests/integration/container_runtime.py
brunoapimentel: Ok, see how it looks like now

tests/integration/container_runtime.py
brunoapimentel: Fixed.

tests/integration/utils.py
brunoapimentel: I saw one in `tests/integraion/conftest.py` and wrote this without a second thought. Fixed now.

tests/integration/container_runtime.py
brunoapimentel: done.

tests/integration/container_runtime.py
brunoapimentel: No, it's not ok. :rabbit: has rightfully pointed out that if `index = 0`, this condition would fail in case it was as simple `if index`

tests/integration/container_engine.py
```diff
@@ -0,0 +1,232 @@
...
+        :param flags: The flags to treat.
+        :return: The treated flags and the entrypoint value.
+        """
+        if flags is None:
+            return [], None
+
+        # find the index of the entrypoint flag, if it exists
+        index = next((i for i, flag in enumerate(flags) if flag.startswith("--entrypoint")), None)
+
+        if index:
```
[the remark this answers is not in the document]
brunoapimentel: Fixed

tests/integration/container_runtime.py
```diff
@@ -0,0 +1,215 @@
...
+    ) -> tuple[str, int]:
+        """Run command using buildah."""
+        try:
+            container_name = self._configure_buildah_container(image)
+            generated_cmd = self._generate_cmd(container_name, image, cmd, flags)
+
+            return _run_cmd([self.name, *generated_cmd])
+        finally:
+            _run_cmd([self.name, "rm", container_name])
+
```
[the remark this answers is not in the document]
brunoapimentel: Fixed.

tests/integration/container_runtime.py
a-ovchinnikov: Oh, I see, thank you!

tests/integration/container_engine.py
```diff
@@ -60,5 +61,10 @@ def name(self) -> str:
 
 
 def get_container_engine() -> ContainerEngine:
-    """Get the container engine."""
-    return PodmanEngine()
+    """Get the configured container engine."""
+    engine_name = os.getenv("HERMETO_CONTAINER_ENGINE", "podman").lower()
```
eskultety: Since this is only test related I think `HERMETO_TEST_` is the prefix you'd want to go with. Commit message update will be needed as well.

tests/integration/container_engine.py
```diff
@@ -40,15 +42,27 @@ def _run_cmd(self, cmd: Union[list[str], str], **subprocess_kwargs: Any) -> tupl
...
-        :rtype: Tuple
-        """
-        return self._run_cmd([self.name, *cmd])
+    def pull_image(self, image: str) -> tuple[str, int]:
+        """Pull container image."""
+        return self._run_cmd([self.name, "pull", image])
+
+    def remove_image(self, image: str) -> tuple[str, int]:
+        """Remove container image."""
+        return self._run_cmd([self.name, "rmi", "--force", image])
```
eskultety: nitpick: would it make sense to map the method names to the actual command they wrap?

Additionally, do we want to add a `**podman-kwargs` just in case we'd want to pass more options with the engine command? It wouldn't hurt IMO even if we don't use it right now.

tests/integration/utils.py
```diff
@@ -185,37 +185,12 @@ def build_image_for_test_case(
...
-        "encoding": "utf-8",
-        "text": True,
-    }
-
-    subprocess_kwargs.update(forced_options)
-    process = subprocess.run(cmd, **subprocess_kwargs)
-
-    return process.stdout, process.returncode
-
-
```
eskultety, line 218: Technically this should be dropped in a standalone commit for easier read just to emphasize "it's no longer needed because we replaced it"

tests/integration/container_engine.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @property
+    def name(self) -> str:
+        """Get the name of the container engine."""
+        return "podman"
+
+
+def get_container_engine() -> ContainerEngine:
+    """Get the container engine."""
+    return PodmanEngine()
```
eskultety: nitpick: This could be introduced in a separate commit as an implementation of the interface and basically keep this "boilerplate" commit just for declaring the interface. Not a dealbreaker though, feel free to ignore unless there are other substantial comments.

tests/integration/container_engine.py
eskultety: I might also not be a fan of these hidden (hidden in terms of being buried in a wrapper) "default" engine CLI options, but then again, this is for tests and only for our own purposes, so I don't expect heavy tweaking, so let's keep it. We can always adjust in the future.

tests/integration/container_engine.py
```diff
@@ -40,15 +42,27 @@ def _run_cmd(self, cmd: Union[list[str], str], **subprocess_kwargs: Any) -> tupl
 
         return process.stdout, process.returncode
 
-    def run_cmd(self, cmd: list[str]) -> tuple[str, int]:
-        """
-        Run command via subprocess.
```
eskultety: If `run_cmd` is no longer needed, the diff would be cleaner and more readable if you just dropped it in a subsequent commit.

tests/integration/utils.py
```diff
@@ -129,17 +125,17 @@ def run_cmd_on_image(
         tmp_path: Path,
         mounts: Sequence[tuple[StrPath, StrPath]] = (),
         net: Optional[str] = "host",
-        podman_flags: Optional[list[str]] = None,
+        flags: Optional[list[str]] = None,
```
eskultety: nitpick: these flags changes also kinda steer the focus from the main changes here which are the interfaces, but unless there's anything else to fix, you can ignore this comment.

tests/integration/container_engine.py
```diff
@@ -40,15 +42,27 @@ def _run_cmd(self, cmd: Union[list[str], str], **subprocess_kwargs: Any) -> tupl
...
+    def build_image(self, build_cmd: list[str], tag: str) -> tuple[str, int]:
```
eskultety: **Commit message**:
In context of the next commit in line, maybe just add a quick note that the build_image interface is strictly compatible with the existing usage and will be addressed in the next commit.

tests/integration/container_engine.py
```diff
@@ -92,11 +94,139 @@ def run_cmd_on_image(
...
+        # entrypoint flag is in the format --entrypoint <entrypoint>
+        elif index + 1 < len(flags):
+            entrypoint_value = flags[index + 1]
+            return flags[:index] + flags[index + 2 :], entrypoint_value
+        # entrypoint flag exists but no value is provided
+        else:
+            raise RuntimeError(
+                f"Unable to parse the run command flags: {flags}."
+                "Check if the --entrypoint flag defines a value."
+            )
```
eskultety: Uhm, IIUC this is all because of a single integration test - `test_legacy.py`, correct? This is quite some work just to make that one particular test work. I mean, yes, it's an important test for the time being, but ultimately we'd drop the entrypoint. That said, our GitHub CI will continue using podman IIUC, not buildah and buildah is more of a niche enterprise test scenario rather than something we'd be hitting anywhere in production with the `--entrypoint` CLI option. How about instead of introducing this logic we make that particular integration test conditionally enabled based on the container engine used for tests? That would be much simpler.

tests/integration/container_engine.py
```diff
@@ -92,11 +94,139 @@ def run_cmd_on_image(
...
+
+        if exit_code != 0:
+            raise RuntimeError(f"Failed to inspect image {image}.")
+
+        parsed_output = json.loads(output)
+        docker_config = parsed_output.get("Docker", {}).get("config", {})
+        cmd = docker_config.get("Cmd", [])
+        entrypoint = docker_config.get("Entrypoint", [])
+
+        return cmd, entrypoint
```
eskultety: If I'm correct in my entrypoint and `test_legacy.py` reasoning, this method should also not be needed.

tests/integration/container_runtime.py
eskultety: > Because stupid me thought making it abstract would prevent it from being instantiated

TIL that in order to achieve this in Python one HAS TO define at least one abstract method, how obvious....

tests/integration/container_runtime.py
ben-alkov: Got it, thx

tests/integration/container_engine.py
brunoapimentel: I think we'll need to keep this one around. There are a few tests that already rely on the image's configured entrypoint (e.g. test_e2e_cargo). 

Buildah is weird when used to run commands, since it completely ignores both the cmd and the entrypoint in the image. I think keeping this here will make this implementation more future-proof.

tests/integration/container_engine.py
brunoapimentel: Yeah, we can, and I agree it's a lot of work for a single test. On the other hand, it is already implemented, and can be useful for any future test that might need to use the --entrypoint flag. It is also only contained in BuildahEngine, so it's not like it'll affect podman.

I'm ok with dropping it too, but give it a second thought considering it might be needed again later.

tests/integration/container_engine.py
brunoapimentel: Fixed.

tests/integration/utils.py
brunoapimentel: Fixed.

tests/integration/container_engine.py
brunoapimentel: Fixed.

tests/integration/container_engine.py
brunoapimentel: Fixed.

tests/integration/utils.py
brunoapimentel: Fixed.

tests/integration/container_engine.py
brunoapimentel: Fixed.

tests/integration/container_engine.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class ContainerEngine(ABC):
+    """Abstract base class for container engines."""
+
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """Get the name of the container engine."""
+        pass
```
a-ovchinnikov: Technicality nitpick: this `pass` is redundant since you already have a docstring.

tests/integration/container_engine.py
```diff
@@ -42,9 +42,14 @@ def _run_cmd(self, cmd: Union[list[str], str], **subprocess_kwargs: Any) -> tupl
 
         return process.stdout, process.returncode
 
-    def build_image(self, build_cmd: list[str], tag: str) -> tuple[str, int]:
+    def build_image(
+        self, flags: Optional[list[str]] = None, context_dir: Optional[StrPath] = "."
```
a-ovchinnikov: Minor nitpick: `context_dir` should not be `Optional` anymore unless `None` is a valid argument to `_run_cmd()`

tests/integration/container_engine.py
```diff
@@ -92,11 +94,139 @@ def run_cmd_on_image(
...
+        the run command can be properly formatted later on.
+
+        :param flags: The flags to treat.
+        :return: The treated flags and the entrypoint value.
+        """
+        if flags is None:
+            return [], None
+
+        # find the index of the entrypoint flag, if it exists
+        index = next((i for i, flag in enumerate(flags) if flag.startswith("--entrypoint")), None)
```
a-ovchinnikov: Minor nitpick: `core.utils` have `first_for` which takes a predicate and an iterable:
```python
is_entrypoint = lambda x: x[1].startswith("--entrypoint")
entrypoint_index = first_for(is_entrypoint, enumerate(flags), (None,))[0]
```
Not sure if it reads any better though. In any case I would suggest replacing `index` with `entrypoint_index` to make it clear what we are dealing with. This would make comment on L168 redundant.

tests/integration/container_engine.py
```diff
@@ -92,11 +94,139 @@ def run_cmd_on_image(
...
+        return cmd, entrypoint
+
+    def run_cmd_on_image(
+        self,
+        image: str,
+        cmd: list[str],
+        flags: Optional[list[str]] = None,
+    ) -> tuple[str, int]:
+        """Run command using buildah."""
+        container_name = self._configure_buildah_container(image)
```
a-ovchinnikov: I wonder if it makes sense to lift `finally` into `_configure_buildah_container` and turn it into a context manager? It could yield a container name and then dispose of the container. Or would that mess with any other intended use?

tests/integration/container_engine.py
eskultety: > I think we'll need to keep this one around. There are a few tests that already rely on the image's configured entrypoint (e.g. test_e2e_cargo).

But that's based on the assumption that we'd want to use buildah to run e2e tests.

tests/integration/container_engine.py
eskultety: > On the other hand, it is already implemented, and can be useful for any future test that might need to use the

Well, you can make the "might come handy in the future" argument literally about any code snippet :). AFAIK the use case for which you need buildah is pretty limited and doesn't involve any e2e tests and so skipping one more test would not make a difference since we're running the whole test suite here in upstream GH.
The problem is (without diving too much into the code details), you're trying to figure out "if" and "what syntax variant" was used with `--entrypoint`. The idea doesn't end here however - it's very likely that another CLI argument would have to be used, hence new flag parsing introduced and the snippet here would just grow like spaghetti beyond maintenance and maintenance matters a lot to me. Any test failure might get particularly interesting to debug based on the exact combination of flags passed on the CLI instead of just failing a particular test because we forgot to disable it with buildah.
I'd proceed very conservatively here - if you were to use buildah as your preferred choice for all interactions instead of podman then I guess you'd have to add more bits. We could track that work in a dedicated issue/milestone once that time comes, but for now I'd really keep this simple and to the bare minimum what is needed for your use case which doesn't strictly relate to upstream matters.

tests/integration/container_engine.py
brunoapimentel: I changed the variable name, but didn't think `first_for` made it more readable that what it was.

tests/integration/container_engine.py
brunoapimentel: Fixed

tests/integration/container_engine.py
brunoapimentel: Fixed

tests/integration/container_engine.py
brunoapimentel: Right, I don't see why not. I think buildah should be able to run all the current integration tests.

tests/integration/container_engine.py
brunoapimentel: Well, but the thing is that is currently useful, and it is not likely to change. I still think that adding buildah means that we should support all test cases that we currently have, unless we have a very big limitation.

I added another commit with an option around having this large piece of code, lmk wyt.

tests/integration/container_engine.py
brunoapimentel: I did the methods renaming. I started dealing with the kwargs, but then I figured that some of the flags don't really have a value counterpart, so you'd need to pass them as `name=None`, which I found odd.

brunoapimentel: That being said, we can still add `flags: list[str]` argument to all methods.

eskultety: > That being said, we can still add `flags: list[str]` argument to all methods.

^This is what I actually meant, I was working with a different context when I suggested `kwargs` that of course can't work since we're spawning a new process, anyway, `flags` it is!

tests/integration/container_engine.py
eskultety: Fair enough, let's not block on this.

tests/integration/container_engine.py
```diff
@@ -0,0 +1,48 @@
+import logging
```
eskultety, line 2: **commit msg:**
s/was implement/was implemented

tests/integration/container_engine.py
```diff
@@ -0,0 +1,231 @@
...
+        return self._run_cmd([self.name, "rmi", "--force", image])
+
+    @abstractmethod
+    def run(
+        self,
+        image: str,
+        cmd: list[str],
+        podman_flags: Optional[list[str]] = None,
+    ) -> tuple[str, int]:
+        """Run command on the image."""
```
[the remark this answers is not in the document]
eskultety: Valid point.

tests/integration/container_engine.py
brunoapimentel: Yeah, it makes total sense, the code looks better now.

tests/integration/container_engine.py
brunoapimentel: Added.

tests/integration/container_engine.py
eskultety: Yeah, good suggestion.

tests/integration/container_engine.py
```diff
@@ -72,6 +72,7 @@ def run(
         self,
         image: str,
         cmd: list[str],
+        entrypoint: Optional[str] = None,
```
eskultety, line 75: Sure, why not, this is reasonable.

tests/integration/container_engine.py
```diff
@@ -96,11 +99,140 @@ def run(
...
+        from_cmd = ["buildah", "from", "--name", container_name, image]
+        _, exit_code = self._run_cmd(from_cmd)
+
+        if exit_code != 0:
+            raise RuntimeError(f"Failed to create buildah container from image {image}.")
+
+        try:
+            yield container_name
+        finally:
+            self._run_cmd([self.name, "rm", container_name])
```
eskultety, line 135: What if this fails for some reason? Would it be a hard failure from buildah's POV, should we caught any exception and soft report it without failing a given test?

tests/integration/container_engine.py
```diff
@@ -96,11 +99,140 @@ def run(
...
+            return flags[:entrypoint_index] + flags[entrypoint_index + 2 :], entrypoint_value
+        # entrypoint flag exists but no value is provided
+        else:
+            raise RuntimeError(
+                f"Unable to parse the run command flags: {flags}."
+                "Check if the --entrypoint flag defines a value."
+            )
+
+    def _get_image_config(self, image: str) -> tuple[list[str], list[str]]:
+        """Get the configured entrypoint and cmd of an image."""
```
eskultety: nitpick: Replace ` Get the configured entrypoint and cmd of an image.` with `Parse entrypoint and cmd from image's JSON configuration`.

tests/integration/container_engine.py
```diff
@@ -96,11 +99,140 @@ def run(
...
+        if entrypoint:
+            return ["run", *filtered_flags, container_name, "--", entrypoint, *cmd]
+
+        # if the image has an entrypoint, prepend it to the command
+        if image_entrypoint:
+            return ["run", *filtered_flags, container_name, "--", *image_entrypoint, *cmd]
+
+        return ["run", *filtered_flags, container_name, "--", *cmd]
+
+    def _filter_entrypoint_flags(
```
eskultety: This will be gone after incorporating the last patch right?

brunoapimentel: Yes

tests/integration/container_engine.py
brunoapimentel: Ok, I implemented it properly then.

tests/integration/container_engine.py
brunoapimentel: Changed it.

tests/integration/container_engine.py
brunoapimentel: If `buildah from` fails, then the container is not created and the test is not run at all. Then it will make the whole test suite fails, and I think that's the appropriate behavior.

tests/integration/container_engine.py
brunoapimentel: Fixed.

tests/integration/container_engine.py
eskultety: > If buildah from fails, then the container is not created and the test is not run at all. Then it will make the whole test suite fails, and I think that's the appropriate behavior

So no unhandled exception in such case?

eskultety (approved):
If the buildah handler doesn't cause unhandled exceptions in the test suite and it just makes it fail -> ACK

tests/integration/container_engine.py
brunoapimentel: The exception in this case is this `RuntimeError` [here](https://github.com/hermetoproject/hermeto/blob/f9f4410fe73553a8879d1dad170ab6cb76a37d47/tests/integration/container_engine.py#L130). Would you suggest another way to make the tests fail?

eskultety: Ah, that one! Can buildah tell us why it could not create a container, otherwise the error is pretty vague. :)

brunoapimentel: I think the best I could do is to also print the stdout from the `buildah from` command. Wdyt?

eskultety: We do that with other commands already, so that's the expected approach, I was just wondering if it fails silently or not :).

brunoapimentel: Ok, added.

PR #1044, opened 2025-07-29
pip: chore: Move Cargo root resolution into a separate function

a-ovchinnikov:
Heuristic for resolving Cargo root directory was reintroduced recently (https://github.com/hermetoproject/hermeto/pull/1040). Unfortunately the problem that it was solving turned out to be system-dependent and thus not directly testable with an integration test. Since this heuristic has once been replaced during refactoring there must be some test that guards against future regressions. This change pulls heuristic into a separate function and covers that function with unit tests to warn future maintainers about possible regressions.

### DIFF
--- a/hermeto/core/package_managers/pip/rust.py
+++ b/hermeto/core/package_managers/pip/rust.py
@@ -4,7 +4,7 @@ import logging
 import shutil
 from pathlib import Path
 from textwrap import dedent
-from typing import Any
+from typing import Any, Iterable
 
 from pybuild_deps import parsers
 
@@ -50,6 +50,16 @@ def _depends_on_rust(extracted_dir: Path) -> bool:
     return False
 
 
+def _get_rust_root_dir(cargo_files: Iterable[Path]) -> Path:
+    # find the top-most Cargo.toml in the package - that's not necessarily the most accurate
+    # solution, but this heuristic has proven to work on the most popular python packages
+    # that have rust dependencies; if this stops working, then we would probably need to check
+    # pyproject toml config section for maturin or setuptools-rust...
+    # More info on this issue in the design doc
+    # https://github.com/hermetoproject/hermeto/blob/e5fa5c0fcd0dff62cf02be5b0d219e04c1ea440c/docs/design/cargo-support.md#L806
+    return min(cargo_files, key=lambda x: len(x.parts)).parent
+
+
 def filter_packages_with_rust_code(packages: list[dict[str, Any]]) -> list[CargoPackageInput]:
     """Filter packages that contain Rust code from a list of pip packages."""
     packages_containing_rust_code = []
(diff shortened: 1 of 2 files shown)


hermeto/core/package_managers/pip/rust.py
```diff
@@ -87,13 +97,7 @@ def filter_packages_with_rust_code(packages: list[dict[str, Any]]) -> list[Cargo
...
             shutil.rmtree(extract_dir, ignore_errors=True)
             continue
-        # find the top-most Cargo.toml in the package - that's not necessarily the most accurate
-        # solution, but this heuristic has proven to work on the most popular python packages
-        # that have rust dependencies; if this stops working, then we would probably need to check
-        # pyproject toml config section for maturin or setuptools-rust...
-        # More info on this issue in the design doc
-        # https://github.com/hermetoproject/hermeto/blob/e5fa5c0fcd0dff62cf02be5b0d219e04c1ea440c/docs/design/cargo-support.md#L806
-        rust_root_dir = min(cargo_files, key=lambda x: len(x.parts)).parent
+        rust_root_dir = _infer_rust_root_dir_from(cargo_files)
```
eskultety: nitpick: for some reason I see e.g. `get_cargo_root_dir` as a simpler, easier to read, name.

tests/unit/package_managers/pip/test_rust.py
```diff
@@ -0,0 +1,29 @@
...
+    [
+        ((Path("/tmp/foo/Cargo.toml"), Path("/tmp/bar/baz/Cargo.toml")), Path("/tmp/foo")),
+        ((Path("/tmp/bar/baz/Cargo.toml"), Path("/tmp/foo/Cargo.toml")), Path("/tmp/foo")),
+        (
+            (
+                Path("/tmp/bar/baz/Cargo.toml"),
+                Path("/tmp/foo/Cargo.toml"),
+                Path("/tmp/foo/quux/Cargo.toml"),
+            ),
+            Path("/tmp/foo"),
```
eskultety: Please use test case IDs with parametrize.

tests/unit/package_managers/pip/test_rust.py
```diff
@@ -0,0 +1,29 @@
...
+        ),
+    ],
+)
+def test_the_shortest_path_in_cargo_package_is_inferred_as_root(
```
eskultety, line 32: What is supposed to happen if we'd encounter a package that would have `<root>/src/pkg1/Cargo.toml` and `<root>/src/pkg2/Cargo.toml`? The inference currently uses `min` , but since `glob` is unstable so is our inference in that case. Sorting would only help stabilizing the tests, but it'd do nothing in real life. I guess that's a topic for a different time though, this PR is only simplifying existing code, it was just a thought from the top of my head.

a-ovchinnikov: It would pick whichever path happens to be first in the list. Realistically that would require a package that a) does not have Cargo.toml at root and b) has several subpackages of same structure. I cannot say how frequent this structure is, basing @bruno-fs's research linked in the code probably it is not very common.

eskultety: Right, `min` is going to give you that, but it'll be an assumption. Agreed, let's deal with it when it comes to that.

bruno-fs: if it comes to that, I think we would have to check `pyproject.toml` configuration.

tests/unit/package_managers/pip/test_rust.py
```diff
@@ -0,0 +1,38 @@
...
+                Path("/tmp/foo/Cargo.toml"),
+                Path("/tmp/foo/quux/Cargo.toml"),
+            ),
+            Path("/tmp/foo"),
+            id="tricky_ordering",
+        ),
+    ],
+)
+def test_the_shortest_path_in_cargo_package_is_inferred_as_root(
+    cargo_files: list, expected_rust_root_dir: Path
```
slimreaper35: The parameter has type `list`, but in pytest.param I see tuples.

a-ovchinnikov: mypy apparently cannot reason about decorators, but it can nag about practically interchangeable types. The ludicrously correct way of fixing this "problem" is to properly annotate `_get_rust_root_dir`. I labelled it "ludicrously correct" for the sole reason that this is Python, it **is not** designed with absolute static type correctness in mind, there is the whole ML family for when that is necessary any representative of which does far better job at this than mypy.

a-ovchinnikov: And then I pushed without commiting the new annotations.

PR #1045, opened 2025-07-30
Add requirements-build.txt file

brunoapimentel:
The setuptools-scm library is needed as a dependency when building the project. The lack of this dependency will make the build fail in a hermetic environment.

### DIFF
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -1,0 +1,3 @@
+setuptools==80.9.0
+setuptools-scm==8.2.0
+


requirements-build.txt
```diff
@@ -0,0 +1,2 @@
+setuptools-scm==8.2.0
```
eskultety, line 2: I think the important bit here is also why this is being introduced in a separate file, i.e. `requirements-build.txt` instead of just adding it to the requirements.txt file :) .

Speaking of which, can you also add setuptools to this file and remove them from requirements.txt and shave a whooping 1.3M of our resulting package install?

brunoapimentel: Humm, I think it won't be possible to that, as setuptools is being included as a transitive dependency for pybuild-deps (which was reintroduced [here](https://github.com/hermetoproject/hermeto/commit/1b1e2a60fabb9ddafa1d08bc0b95be6d2d879f95#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552)).

eskultety: Oh, right. Well, we can still add it here and pretend the other reference didn't happen :). I wonder if pybuild-deps actually needs it for runtime or it's just yet another runtime/build dependency that was declared incorectly.

brunoapimentel: I did it. And dependabot will be able to keep it up-to-date: https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories#pip-and-pip-compile

PR #1046, opened 2025-07-30
doc: Maven Package Manager Design

adambkaplan:
This is an initial design document for adding Maven as a package manager, based on the proof of concept in PR #1021 [1]. It provides background context on how Maven projects are structured, and how the Maven lockfile plugin can be used to generate a dependency list for Hermeto.

The feature is constrained to Maven 3.x, as Maven 4.x is currently under development and will introduce several breaking changes. The feature will not support other Maven-compatible package managers, such as Gradle.

Assisted-by: Cursor

### DIFF
--- a/docs/design/maven-3.md
+++ b/docs/design/maven-3.md
@@ -1,0 +1,478 @@
+# Maven 3 Design Document
+
+<!-- 
+This template is intended to help contributors add support for a new package manager (ecosystem).
+Completed design documents are not required prior to contributing code - they are meant to
+facilitate conversation and technical decisionmaking.
+
+To get started:
+
+- Make a copy of this template.
+- Fill out the "Overview" section describing the package manager to the best of your ability.
+  Submit a PR to start a conversation with the community!
+- Complete the "Design" sections as code is written, or if feedback is desired prior to
+  implementation.
+- Complete the "Implementation Notes" sections as desired, or when the package manager is ready to
+  be enabled by default.
+-->
+
+## Overview
+
+[Maven](https://maven.apache.org) is a package manager (ecosystem) for Java applications. It is one
+of the first full-featured package managers for any programming language and drove the wide open
+source ecosystem for Java application development. Java programs are _compiled_ into bytecode,
+which is then executed within a Java virtual machine (JVM). Java programs are meant to be agnostic
+to host operating systems and CPU architectures, though in practice some programs need dependencies
+that have native OS/CPU extensions.
+
+Maven's current active major version - Maven 3 - was first released in 2010, with backwards
+compatibilty support for Maven 2 (released in 2005). Maven 4 is currently in a beta phase, with a
+GA release planned for later in 2025.
+
+### Developer Workflow
+
+To use Maven, developers first install the [mvn](https://maven.apache.org/install.html) command
+line tool. Next, they set up a local "Maven project" where the source code resides - see the
+["Maven in 5 Minutes"](https://maven.apache.org/guides/getting-started/maven-in-five-minutes.html)
+tutorial for a quick guide on how this works.
+
+A Maven project typically has this directory tree structure:
+
+```
+my-app
+|-- pom.xml
+`-- src
+    |-- main
+    |   `-- java
+    |       `-- io
+    |           `-- mydomain
+    |               `-- app
+    |                   `-- App.java
+    `-- test
+        `-- java
+            `-- io
+                `-- mydomain
+                    `-- app
+                        `-- AppTest.java
+```
+
+- A `pom.xml` file in the project's root directory.
+- A `src` directory where Java code resides. This is further sub-divided into `main`, `test`, and
+  other directories.
+
+Dependencies are declared in the `pom.xml` file. These include:
+
+- Dependencies of the Java code needed at compile time.
+- Dependencies used to test the Java code.
+- [Plugins](https://maven.apache.org/plugins/index.html) which extend Maven's behavior for building
+  and testing applications.
+
+Maven 3 has a [default lifecycle](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference)
+that sequentially executes several [build phases](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference).
+Most Maven 3 projects invoke the `package`, `install`, or `deploy` phases to compile and/or test
+their applications.
+
+### How the Package Manager Works
+
+#### Dependency Management
+
+Developers declare dependencies in the `pom.xml` file using the `<dependencies>`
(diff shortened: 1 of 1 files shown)


adambkaplan:
This is my first "test run" of the design doc template I proposed in #1032

docs/design/maven-3.md
```diff
@@ -0,0 +1,478 @@
...
+## Implementation Notes
+
+<!-- 
+This section helps the community evaluate the maturity of the package manager. Experimental package
+managers are hidden behind the `--dev-package-managers` flag (disabled by default). It is optional
+for experimental package managers, but should be completed before the package manager is enabled
+by default.
+-->
+
+### Deprecation of `maven` Generic Prefetcher
```
brunoapimentel: As a side note, I think we should also deprecate the whole generic fetcher extensibility proposal when we remove the Maven support from it. I'm still not convinced it was a good idea.

docs/design/maven-3.md
```diff
@@ -0,0 +1,478 @@
...
+for generated content.
+
+### Dependency List Generation
+
+Unfortunately, Maven 3 does not have a native notion of a "lockfile." A typical Maven project only
+lists a minimum set of direct dependencies and required plugins, relying on the resolver to obtain
+transitive dependencies at compile or test time. Full specification of transitive dependencies is
+[not required](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Transitive_Dependencies).
+
+This design relies on the [Maven Lockfile](https://github.com/chains-project/maven-lockfile)
```
brunoapimentel, line 143: I'm curious what approach they took to come up with the list of dependencies. I reckon it is a hard problem to solve. In any case, I think that what will give us proof of its accuracy is our ability to build projects hermetically from the output.

brunoapimentel:
> This is my first "test run" of the design doc template I proposed in #1032

I really like the approach you took with the design doc template, I think the prompts really help a person who's not too familiar with the project to ask themselves the right questions and to properly convey their proposal to the team.

I just wonder if we can somehow simplify it, it does seem to be too extensive. I will poke at it later to provide more useful feedback on it, but that's my first impression at least.

docs/design/maven-3.md
adambkaplan: Maven has internal APIs that let plugin developers invoke the dependency resolution process. These APIs are well tested by the official `dependency` plugin, and the lockfile plugin uses similar code paths.

I found it's a bit more difficult/challenging to resolve the dependencies of plugins that a project depends on. I still need to explore this a bit more.

adambkaplan:
> I just wonder if we can somehow simplify it, it does seem to be too extensive. I will poke at it later to provide more useful feedback on it, but that's my first impression at least.

I struggle with this too in the other projects I'm involved in. I may add a "preamble" section to emphasize that not everything needs to be filled out all at once. This goes both for contributors and maintainers - design docs shouldn't need to be complete in order to be merged!

eskultety:
> I struggle with this too in the other projects I'm involved in. I may add a "preamble" section to emphasize that not everything needs to be filled out all at once

In general, I don't object to ^this line of thinking, simply because we have historically explicitly stated in past design docs what the doc's scope is and what isn't, i.e. what is going to be delivered and what won't work and will be subject to further investigation and improvement, so I don't see an issue there. However, what I specifically don't want is that design docs would be accepted or rushed with severe architectural gaps, adding support for backends with open questions on the ecosystem practices and security, IOW how does an ecosystem handle depsolving, can we do integrity checking, where and how are deps stored, what does the native tooling really do and provide to determine whether arbitrary code execution can or cannot be ruled out and hence nudge the implementation in certain direction. So, I expect every design doc, while scoped, to have clear answers on the important bits.

docs/design/maven-3.md
```diff
@@ -0,0 +1,478 @@
...
+Maven build lifecycle. We recommend the following plugin configuration:
+
+```xml
+<plugin>
+  <groupId>io.github.chains-project</groupId>
+  <artifactId>maven-lockfile</artifactId>
+  <version>5.6.0</version>
+  <executions>
+    <execution>
+      <phase>validate</phase>
```
adambkaplan, line 175: I should consult with the chains-project community on which phase is appropriate. `verify` may be the better lifecycle phase to attach this goal to, since it ensures the project compiles first.

See https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html

adambkaplan:
> However, what I specifically don't want is that design docs would be accepted or rushed with severe architectural gaps, adding support for backends with open questions on the ecosystem practices and security

I think this is where the "experimental" practice/paradigm is needed, and iterative merging of designs. Why should we block merge on an idea that is permitted to make breaking changes? I would reserve this scrutiny for when a package manager wants to promote itself from experimental to supported.

Maven 3 is especially challenging because of its age - it is locked into practices that were set 20 years ago with Maven 2, and has struggled to evolve. The fixes for Maven 3 will have to come from outside its sponsor, Apache, since the project has decided to make significant breaking changes with Maven 4 (currently in a beta/release candidate phase). We need space to innovate within the boundaries of the current ecosystem, which is enormous. A majority of the world's enterprise software is written in Java and built with Maven 3.

docs/design/maven-3.md
adambkaplan: Decided not to deprecated this - if anything, there are strong arguments for expanding its support beyond maven.

eskultety:
> I think this is where the "experimental" practice/paradigm is needed, and iterative merging of designs. Why should we block merge on an idea that is permitted to make breaking changes? I would reserve this scrutiny for when a package manager wants to promote itself from experimental to supported.

This is a strong argument, you convinced me. That said, maybe we need a small update here: https://github.com/hermetoproject/hermeto/pull/1032 - every design doc introducing an experimental support for a given backend (maturing over time) should state in the design doc that the functionality is still experimental, that way, we gain the clarity in our support statement everywhere (since I really don't want us to push vague & opaque designs for our docs page to publish) and you won't be blocked by merging a solid design for some working PoC implementation, @adambkaplan is that acceptable?

adambkaplan:
@eskultety @brunoapimentel apologies it has taken so long to respond to your comment. Added a "Status" line at the top of the design doc. I have also rebased to avoid any potential merge issues.

docs/design/maven-3.md
```diff
@@ -0,0 +1,478 @@
...
+```
+
+Developers can ensure the lockfile is kept up to date by adding the `validate` goal to the
+Maven build lifecycle. We recommend the following plugin configuration:
+
+```xml
+<plugin>
+  <groupId>io.github.chains-project</groupId>
+  <artifactId>maven-lockfile</artifactId>
+  <version>5.6.0</version>
```
adambkaplan, line 172: Sorry bot, your information is stale. The latest version is actually 5.10.0

slimreaper35:
This works has been moved to https://github.com/hermetoproject/hermeto/pull/1381

PR #1058, opened 2025-08-05
GitHub page final tweaks

slimreaper35:

### DIFF
--- a/docs/generic.md
+++ b/docs/generic.md
@@ -44,7 +44,7 @@ hermeto fetch-deps \
 
 where 'JSON input' is
 
-```jsonc
+```js
 {
   "type": "generic",
   // path to the package (relative to the --source directory)
--- a/docs/yarn_classic.md
+++ b/docs/yarn_classic.md
@@ -48,32 +48,11 @@ Yarn Classic PM is capable of processing the following types of dependencies
 
 ## Fetching dependencies for Yarn Classic projects
 
-The process of fetching dependencies for Yarn Classic is similar to that for any
-other package manager. The name of package manager is `yarn_classic`, and it
-does not expect any additional arguments.
+The process of fetching dependencies for Yarn Classic is identical to the Yarn
+backend. Hermeto automatically detects Yarn Classic (i.e. v1) projects and so
+the logic is all conveniently hidden behind a single `yarn` CLI backend option.
 
-Hermeto can be run as follows
-
-```shell
-hermeto fetch-deps \
-  --source ./my-repo \
-  --output ./hermeto-output \
-  '<JSON input>'
-```
-
-where 'JSON input' is
-
-```jsonc
-{
-  // "yarn_classic" tells Hermeto to process Yarn packages
-  "type": "yarn_classic",
-  // path to the package (relative to the --source directory)
-  // defaults to "."
-  "path": ".",
-}
-```
-
-For complete example of how to pre-fetch dependencies, see [Pre-fetch dependencies][].
+For complete example of how to pre-fetch dependencies for see [Yarn documentation][].
 
 ## Prerequisites for an offline build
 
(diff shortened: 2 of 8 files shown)


mkdocs.yml
```diff
@@ -49,6 +49,8 @@ nav:
 # Markdown extensions
 markdown_extensions:
   - pymdownx.emoji
+  - pymdownx.highlight:
+  - pymdownx.superfences
```
eskultety, line 55: you should also add [pygments](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#+pymdownx.highlight.use_pygments) so that highlighting happens prior rendering the page instead of leaving it to the browser.

docs/generic.md
```diff
@@ -44,7 +44,7 @@ hermeto fetch-deps \
 
 where 'JSON input' is
 
-```jsonc
+```json
```
eskultety: Oh, no, this breaks the GH renderer making my eyes bleed with the comments. (TIL) Since comments are apparently linguistically invalid in JSON (hence GH renderer highlighting it in red), we could either do `js` or `json5` (for json5 you need the pygments plugin which added JSON5 not so long ago, but doesn't look as good as when formatted as javascript IMO).

slimreaper35: How about moving some comments outside of the code block, and maybe dropping some of them as well?

eskultety: I haven't inspected the JSON snippets we have in the docs, so while there may be some snippets where dropping the comments would make sense I hardly doubt we could do it for all of them. What do you mean by moving them outside of the code block? Like code block annotations or something else? I looked at annotations and I wasn't impressed really, it was actually more confusing that proper code comments AND I'd be curious how GH renderer would handle them.

Do we truly want to spend and invest more work on refining the JSON snippets in the docs just to enable proper color theming for `json` instead of using an, IMO viable and dead simple to do though slightly less colored, alternative? :)

slimreaper35: > Do we truly want to spend and invest more work on refining the JSON snippets in the docs just to enable proper color theming for json instead of using an, IMO viable and dead simple to do though slightly less colored, alternative? :)

definitely not

slimreaper35: I dropped that commit.

eskultety: > I dropped that commit.

Why though? You spent effort on enabling colors in the web docs only to let JSON formatting produce a lexer error (hidden of course) inside pygments -> hence no coloring. I mean if you'd be okay with formatting JSON as javascript (for the record JSON stands for Javascript Object Notation), then it's fairly simple to `sed` it:
`git grep -l '```json' docs/ | xargs sed -E -i 's/```json[c]?/```js/` 

and that should be all we need to do unless I missed something.

docs/yarn_classic.md
```diff
@@ -48,32 +48,11 @@ Yarn Classic PM is capable of processing the following types of dependencies
 
 ## Fetching dependencies for Yarn Classic projects
 
-The process of fetching dependencies for Yarn Classic is similar to that for any
-other package manager. The name of package manager is `yarn_classic`, and it
-does not expect any additional arguments.
+The process of fetching dependencies for Yarn Classic is identical as for Yarn.
```
eskultety: s/identical as for/identical to the Yarn backend.

docs/yarn_classic.md
```diff
@@ -48,32 +48,11 @@ Yarn Classic PM is capable of processing the following types of dependencies
...
+But only `yarn` is a supported CLI option. Hermeto automatically detects Yarn
```
eskultety: > But only `yarn` is a supported CLI option. Hermeto automatically detects Yarn...

I think _Hermeto automatically detects Yarn Classic (i.e. v1) projects and so the logic is all conveniently hidden behind a single `yarn` CLI backend option._ might read slightly better.

docs/generic.md
```diff
@@ -44,7 +44,7 @@ hermeto fetch-deps \
 
 where 'JSON input' is
```
eskultety, line 45: > Converting code block to pure json on the other
> hand causes issues on GitHub

I'd just add why - because comments are not supported in JSON.

ben-alkov (approved):
LGTM, with comments.

- I agree with Erik's comments, so only merge after taking care of those
- Do you have a rendered version somewhere? Like from your fork of Hermeto? f not, no worries, but please consider it in the future.

slimreaper35:
> Do you have a rendered version somewhere? Like from your fork of Hermeto? f not, no worries, but please consider it in the future.

No, I did not set up a GitHub page (or GitHub actions) in my fork. I usually, during a review, check out the PR/branch and run `mkdocs serve` and see the proposed changes locally.

PR #1060, opened 2025-08-07
Enable platform-specific binary package filtering input

taylormadore:
This PR introduces flexible binary package filtering input parameters that allow users to specify which platform-specific packages to prefetch based on architecture, OS, Python version, and other characteristics.

This is **only** the input model implementation and does not change the current behavior. New behavior for the individual package manager backends will be added in subsequent PRs.

### DIFF
--- a/hermeto/core/binary_filters.py
+++ b/hermeto/core/binary_filters.py
@@ -1,0 +1,30 @@
+"""Base classes for binary package filtering."""
+
+from abc import ABC, abstractmethod
+from typing import Any, Optional, Set
+
+from hermeto.core.models.input import BINARY_FILTER_ALL
+
+
+class BinaryPackageFilter(ABC):
+    """Abstract base class for binary package filtering."""
+
+    def _parse_filter_spec(self, spec: str) -> Optional[Set[str]]:
+        """Parse filter specification into allowed values set.
+
+        Returns None if spec is ':all:' or contains ':all:' as any item.
+        This matches pip's behavior where any occurrence of ':all:' means accept all.
+        """
+        if spec == BINARY_FILTER_ALL:
+            return None
+
+        filters = {stripped_filter for item in spec.split(",") if (stripped_filter := item.strip())}
+
+        if BINARY_FILTER_ALL in filters:
+            return None
+
+        return filters
+
+    @abstractmethod
+    def __contains__(self, item: Any) -> bool:
+        """Check if item passes the filter criteria."""
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -11,6 +11,8 @@ from hermeto.core.errors import InvalidInput
 from hermeto.core.models.validators import check_sane_relpath, unique
 from hermeto.core.rooted_path import PathOutsideRoot, RootedPath
 
+BINARY_FILTER_ALL = ":all:"
+
 if TYPE_CHECKING:
     from pydantic.error_wrappers import ErrorDict
 
@@ -138,6 +140,27 @@ class SSLOptions(pydantic.BaseModel, extra="forbid"):
             )
 
         return self
+
+
+def _validate_binary_filter_format(value: Any) -> str:
+    """Validate binary filter format as either a single value or comma-separated string.
+
+    May return a slightly modified string with non-empty, unique, trimmed values.
+    """
+    if not isinstance(value, str):
+        raise ValueError(f"Binary filter must be a string, got {type(value).__name__}")
+
+    # Maintain input order
+    unique_stripped_filters = dict.fromkeys(
+        stripped_filter for item in value.split(",") if (stripped_filter := item.strip())
+    )
+    if not unique_stripped_filters:
+        raise ValueError("Binary filter cannot contain only empty values")
+
+    return ",".join(unique_stripped_filters.keys())
+
+
+BinaryFilterStr = Annotated[str, pydantic.BeforeValidator(_validate_binary_filter_format)]
 
 
 class BundlerPackageInput(_PackageInputBase):
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -23,6 +23,40 @@ T = TypeVar("T")
 ModelT = TypeVar("ModelT", bound=pydantic.BaseModel)
 
 
+def _handle_legacy_allow_binary(
+    instance: Union["PipPackageInput", "BundlerPackageInput"],
+    binary_filter_class: Union[type["PipBinaryFilters"], type["BundlerBinaryFilters"]],
+) -> None:
+    """Handle backward compatibility for allow_binary field.
+
+    May modify instance attributes.
+    """
+    # If allow_binary is already False, nothing to process
+    if not instance.allow_binary:
+        return
+
+    # Check if user provided both fields originally
+    user_provided_both = instance.allow_binary and instance.binary is not None
+    # Determine if allow_binary should be migrated to binary field
+    should_migrate_allow_binary = instance.allow_binary and instance.binary is None
+
+    if user_provided_both:
+        log.warning(
+            "Both 'allow_binary' and 'binary' fields specified. "
+            "The 'binary' field will take precedence. "
+            "Please remove 'allow_binary' as it is deprecated."
+        )
+    elif should_migrate_allow_binary:
+        log.warning(
+            "The 'allow_binary' field is deprecated and will be removed in the next major version. "
+            "Please use 'binary': {} instead of 'allow_binary': true."
+        )
+        instance.binary = binary_filter_class.with_allow_binary_behavior()
+
+    # Set allow_binary to False to prevent duplicate processing
+    instance.allow_binary = False
+
+
 def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
     """Parse user input into a model, re-raise validation errors as InvalidInput."""
     try:
@@ -177,11 +211,25 @@ class PipBinaryFilters(BinaryModeOptions):
     py_version: BinaryFilterStr = BINARY_FILTER_ALL
     py_impl: BinaryFilterStr = "cp"
 
+    @classmethod
+    def with_allow_binary_behavior(cls) -> Self:
+        """Create filters that mimic the old allow_binary=True behavior."""
+        return cls(
+            arch=BINARY_FILTER_ALL,
+            os=BINARY_FILTER_ALL,
+            py_impl=BINARY_FILTER_ALL,
+        )
+
 
 class BundlerBinaryFilters(BinaryModeOptions):
     """Binary filters specific to bundler packages."""
 
     platform: BinaryFilterStr = BINARY_FILTER_ALL
+
+    @classmethod
+    def with_allow_binary_behavior(cls) -> Self:
+        """Create filters that mimic the old allow_binary=True behavior."""
+        return cls()
 
 
 class RpmBinaryFilters(pydantic.BaseModel, extra="forbid"):
@@ -195,6 +243,13 @@ class BundlerPackageInput(_PackageInputBase):
 
     type: Literal["bundler"]
     allow_binary: bool = False
+    binary: Optional[BundlerBinaryFilters] = None
+
+    @pydantic.model_validator(mode="after")
+    def _handle_legacy_allow_binary_field(self) -> Self:
+        """Handle backward compatibility for allow_binary field."""
+        _handle_legacy_allow_binary(self, BundlerBinaryFilters)
+        return self
 
 
 class CargoPackageInput(_PackageInputBase):
@@ -229,6 +284,7 @@ class PipPackageInput(_PackageInputBase):
     requirements_files: Optional[list[Path]] = None
     requirements_build_files: Optional[list[Path]] = None
     allow_binary: bool = False
+    binary: Optional[PipBinaryFilters] = None
 
     @pydantic.field_validator("requirements_files", "requirements_build_files")
     def _no_explicit_none(cls, paths: Optional[list[Path]]) -> list[Path]:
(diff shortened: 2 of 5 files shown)


taylormadore:
@coderabbitai review

hermeto/core/models/input.py
```diff
@@ -120,11 +153,113 @@ def _validate_ssl_options(self) -> Self:
...
+    if not isinstance(value, str):
+        raise ValueError(
+            (
+                f"Binary filter must be a string (e.g., 'x86_64,aarch64' or ':all:'). "
+                f"Got type: {type(value).__name__}"
+            )
+        )
+
+    value = value.strip()
+    if value == BINARY_FILTER_ALL:
```
slimreaper35: I see uppercase `:ALL:` in the unit tests, but here is a check only for the lowercase value.

taylormadore: Currently the magic `:all:` keyword is case-sensitive, so I think that is the expected behavior. The test case is showing that `:ALL:` is accepted from an input standpoint but does not trigger the same behavior as `:all:`.

slimreaper35:
I don't think I have any major comments 👍 Hopefully, I didn't miss something.

hermeto/core/models/input.py
```diff
@@ -120,6 +122,60 @@ def _validate_ssl_options(self) -> Self:
...
+        return BinaryFilter()
+
+    filters = {f.strip() for f in value.split(",") if f.strip()}
```
eskultety: nitpick: can we use walrus here? `{f for f in value.split(",") if (f := f.strip())}` (or with any other variable other than `f` :) ).

tests/unit/models/test_input.py
```diff
@@ -393,3 +397,160 @@ def test_empty_packages(self) -> None:
...
+    ) -> None:
+        if input_filters is None:
+            filter = BinaryFilter()
+        else:
+            filter = BinaryFilter(filters=input_filters)
+        assert filter.filters == expected_filters
+        assert filter.is_all is expected_is_all
+
+
+class TestParseBinaryFilter:
```
eskultety: nitpick: Would it make sense to combine these 2 test classes into a single `TestBinaryFilter` one?

tests/unit/models/test_input.py
```diff
@@ -393,3 +397,160 @@ def test_empty_packages(self) -> None:
...
+            pytest.param("x86-64,arm-linux", {"x86-64", "arm-linux"}, id="hyphenated_values"),
+            pytest.param(
+                "linux_gnu,darwin_x64", {"linux_gnu", "darwin_x64"}, id="underscore_values"
+            ),
+            pytest.param("3.9,3.10,3.11", {"3.9", "3.10", "3.11"}, id="dotted_values"),
+            pytest.param("x86_64-linux-gnu", {"x86_64-linux-gnu"}, id="mixed_special_chars"),
+            pytest.param("123,456,789", {"123", "456", "789"}, id="numeric_string_values"),
+            pytest.param("x86_64,123,abc", {"x86_64", "123", "abc"}, id="mixed_numeric_alpha"),
+        ],
+    )
```
eskultety: Woa, what a coverage!

hermeto/core/models/input.py
```diff
@@ -120,6 +122,60 @@ def _validate_ssl_options(self) -> Self:
...
+            f"<{type(self).__name__} "
+            f"filters={':all:' if self.is_all else ','.join(sorted(self.filters))}>"
+        )
+
+
+def _parse_binary_filter(value: Any) -> BinaryFilter:
+    """Parse either None, BinaryFilter, or comma-separated string into a BinaryFilter object."""
+    if value is None:
+        return BinaryFilter()
+    if isinstance(value, BinaryFilter):
```
eskultety: When do we expect this use case with an instance, that's not something that could be done by the user, are we going to instantiate like this ourselves?

eskultety:
> This is **only** the input model implementation and does not change the current behavior. New behavior for the individual package manager backends will be added in subsequent PRs.

Ah, so this means that

> or python implementation, the implementation-agnostic py and ...

will be done in a follow up, correct?

tests/unit/models/test_input.py
```diff
@@ -148,6 +163,66 @@ class TestPackageInput:
...
+            (
+                {
+                    "type": "pip",
+                    "binary": {
+                        "arch": "aarch64,armv7l",
+                        "os": "darwin,windows",
+                        "py_version": "3.9,3.10",
+                        "py_impl": "pp,jy",
+                        "packages": "numpy,pandas",
+                    },
```
eskultety, line 175: Preexisting, but can we start adding new parametrize cases including an ID? That shouldn't break pytest, right?

hermeto/core/models/input.py
```diff
@@ -120,6 +122,60 @@ def _validate_ssl_options(self) -> Self:
...
+        return ":all:" if self.is_all else ",".join(sorted(self.filters))
+
+    def __repr__(self) -> str:
+        return (
+            f"<{type(self).__name__} "
+            f"filters={':all:' if self.is_all else ','.join(sorted(self.filters))}>"
+        )
+
+
+def _parse_binary_filter(value: Any) -> BinaryFilter:
```
a-ovchinnikov: This feels like a classmethod for BinaryFilter, is there any reason to keep this separate from the class?

hermeto/core/models/input.py
```diff
@@ -120,6 +122,60 @@ def _validate_ssl_options(self) -> Self:
...
+        """Return True if this filter accepts all (no filtering)."""
+        return not self.filters
+
+    def __str__(self) -> str:
+        return ":all:" if self.is_all else ",".join(sorted(self.filters))
+
+    def __repr__(self) -> str:
+        return (
+            f"<{type(self).__name__} "
+            f"filters={':all:' if self.is_all else ','.join(sorted(self.filters))}>"
```
a-ovchinnikov: The [recommended way](https://docs.python.org/3/reference/datamodel.html#object.__repr__) of defining a \_\_repr\_\_() is to have it return something that
```
If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value
```
I.e. here you would want to have something like `BinaryFilter(filters=set())`. Maybe this is a sign to move `:all:` processing into the initializer?

tests/unit/models/test_input.py
a-ovchinnikov: It looks like we are testing Python stdlib here, do we need to do that?

hermeto/core/models/input.py
```diff
@@ -23,6 +23,37 @@
...
+    if user_provided_both:
+        log.warning(
+            "Both 'allow_binary' and 'binary' fields specified. "
+            "The 'binary' field will take precedence. "
+            "Please remove 'allow_binary' as it is deprecated."
+        )
+    elif should_migrate_allow_binary:
+        log.warning(
+            "The 'allow_binary' field is deprecated and will be removed in the next major version. "
+            "Please use 'binary': {} instead of 'allow_binary': true."
```
a-ovchinnikov, line 52: Should it be something like `'binary': {'arch...}` or do you mean that an empty map is fine too?

tests/unit/models/test_input.py
```diff
@@ -554,3 +655,111 @@ def test_field_accepts_none(self) -> None:
         model = self.TestModel(arch_filter=None)
         assert model.arch_filter.filters == set()
         assert model.arch_filter.is_all is True
+
+
+class TestPipBinaryFilters:
+    def test_default_values(self) -> None:
```
a-ovchinnikov: Here we test that pydantic sets default values to the values that we asked it to use. Which means that the test is for pydantic. Does it provide value to Hermeto?

a-ovchinnikov (approved):
LGTM with several nitpicks and optional suggestions.

tests/unit/models/test_input.py
eskultety: I would definitely not oppose reducing the test footprint if we can.

taylormadore:
Changes in V2:
- Kept the input validation focused on the input string format. I felt like I was trying to do too much parsing there in V1 and abusing the validator
- Cut down and simplified the unit tests
- Added the RPM arch filtering as an example (to be dropped from this before merge and opened separately)

hermeto/core/models/input.py
taylormadore: When we dropped the enabled option (`"binary": {"enabled": true}`) from the design, the empty mapthe way to specify that binaries are requested when there are no non-default options specified

taylormadore:
Rebased on main

slimreaper35:
Do you want to include RPM filtering in this PR?

taylormadore:
> Do you want to include RPM filtering in this PR?

No, I'll drop that final commit from this before merging it and open it in a separate PR. I only added it to this one so the reviewers could see where I was going with the ABC.

slimreaper35:
Sigh, I didn't expand the commit message 😞

hermeto/core/models/input.py
```diff
@@ -140,6 +142,27 @@ def _validate_ssl_options(self) -> Self:
         return self
 
 
+def _validate_binary_filter_format(value: Any) -> str:
+    """Validate binary filter format as either a single value or comma-separated string.
+
+    May mutate the input value slightly to return non-empty, unique, trimmed values
```
a-ovchinnikov: A nitpick/terminology warning: "mutate" makes this read as if it modifies an object in an enclosing scope which is not true since the function deals exclusively with strings. Not a blocker, but produces a tiny bit of confusion.

hermeto/core/models/input.py
```diff
@@ -23,6 +23,37 @@
 ModelT = TypeVar("ModelT", bound=pydantic.BaseModel)
 
 
+def _handle_legacy_allow_binary(
+    instance: Union["PipPackageInput", "BundlerPackageInput"],
+    binary_filter_class: Union[type["PipBinaryFilters"], type["BundlerBinaryFilters"]],
+) -> None:
+    """Handle backward compatibility for allow_binary field."""
```
a-ovchinnikov: Another non-blocking nitpick: now this one mutates objects in outer scope! Would have been nice is this was reflected in the docstring:
"""... .
Modifies instance attributes."""

hermeto/core/binary_filters.py
```diff
@@ -0,0 +1,31 @@
...
+
+        if BINARY_FILTER_ALL in filters:
+            return None
+
+        return filters
+
+    @abstractmethod
+    def __contains__(self, item: Any) -> bool:
+        """Check if item passes the filter criteria."""
+        pass
```
a-ovchinnikov: A negligible, non-blocking nitpick: there is no need in `pass` here, a docstring is enough to delineate the method's boundaries.

hermeto/core/package_managers/rpm/main.py
```diff
@@ -327,7 +337,8 @@ def _download(
     Prepare a list of files to be downloaded, and then download files.
     """
     metadata = {}
-    for arch in lockfile.arches:
+
+    for arch in _filter_architectures(lockfile, arch_filter):
```
a-ovchinnikov: Just wondering what do you think about making `_filter_architecture`  a method of `RpmArchitectureFilter`? I.e. to have something like this:
```diff
- for arch in _filter_architectures(lockfile, arch_filter):
+ for arch in arch_filter.filter(lockfile):
# or even
+ for arch in arch_filter(lockfile):
```

hermeto/core/models/input.py
taylormadore: I agree that should be mentioned for this one. Will fix :+1:

hermeto/core/models/input.py
taylormadore: What I wanted to convey is that this validator isn't strictly validating the input--it _**may**_ also make some small changes to normalize the input string

a-ovchinnikov: Yes, that's not a big issue, just something that I had to get back to and re-read once I realized that the input is actually immutable. Not a blocker, maybe something like "may return a slightly modified string" will be clearer?

hermeto/core/package_managers/rpm/main.py
taylormadore: Good point. I'm going to drop this WIP commit from this PR before merging, but I like this for the follow-up PR implementing the filters for RPM :+1:

tests/unit/models/test_input.py
taylormadore: Tests substantially reworked in V2

hermeto/core/models/input.py
taylormadore: Removed in V2

hermeto/core/models/input.py
taylormadore: Removed in V2

hermeto/core/models/input.py
taylormadore: Done

tests/unit/models/test_input.py
taylormadore: Tests substantially reworked and reduced in V2

tests/unit/models/test_input.py
taylormadore: Done

tests/unit/models/test_input.py
```diff
@@ -431,3 +519,68 @@ def test_rejects_invalid_formats(self, input_value: Any, error_match: str) -> No
         """Test that invalid formats are rejected."""
         with pytest.raises(ValueError, match=error_match):
             _validate_binary_filter_format(input_value)
+
+
+class TestLegacyAllowBinary:
+    """Test legacy allow_binary field migration functionality."""
+
+    @pytest.mark.parametrize(
```
slimreaper35, line 527: Some new tests are missing `pytest ids`. Not a big deal for me, but I know from the past that it might be useful to others, as I was requested to add them.

taylormadore: I thought these ones were relatively self-documenting, but I'm also not opposed to adding the ids. Added.

taylormadore:
As planned, dropped the final, WIP demonstration commit before merge

PR #1062, opened 2025-08-08
Dockerfile: ship both cargo & rustc

ptoscano:
The `cargo` executable alone may not be enough to resolve/check Rust dependencies. In some cases, it tries to run `rustc`, thus failing because it is missing in the container image.

To make sure that `cargo` can run properly all the time, install that and `rust` directly in the base image used for the final image, rather than copying the `cargo` executable from an helper image. This way not only `cargo` is available, a working `rustc` is too.

Resolves: https://github.com/hermetoproject/hermeto/issues/1061

Fixes commit 55c218dc529e99b284a66d04ef48aa56d3fcc4c0.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,13 @@ FROM ubi as base
 RUN dnf -y install \
     --setopt install_weak_deps=0 \
     --nodocs \
+    cargo \
     git-core \
     jq \
     python3 \
     rubygem-bundler \
     rubygem-json \
+    rust \
     subscription-manager && \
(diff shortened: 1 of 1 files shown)


taylormadore:
@ptoscano As mentioned in this [comment](https://github.com/hermetoproject/hermeto/issues/1061#issuecomment-3175363120), we would like to try another approach first without installing rustc in our image, so I'm going to close this for now. Thank you again for reporting this issue and we'll continue discussing it in the issue thread

PR #1076, opened 2025-08-21
Revert "build(deps): bump backrefs from 5.9 to 6.0.1"

eskultety:
This reverts commit e2ddf643dcb21cc75df0b638ffea73796dd9f212.

One can't even create a virtual environment due to:
    ERROR: Cannot install -r requirements-extras.txt (line 687) and
    backrefs==6.0.1 because these package versions have conflicting
    dependencies.

    The conflict is caused by:
        The user requested backrefs==6.0.1
        mkdocs-material 9.6.16 depends on backrefs~=5.7.post1

I don't really know how CI could *not* have caught this since we install to a virtual environment as part of it, so beats me why we were able to merge the change in the first place.


slimreaper35:
So it was because of the `--no-deps` option, which prevents installing package dependencies from the requirements file.

eskultety:
> So it was because of the `--no-deps` option, which prevents installing package dependencies from the requirements file.

Ah, that's right! Still, the version spec on `mkdocs-material` doesn't officially allow us to upgrade just yet `backrefs~=5.7.post1` and so we ought to revert.

PR #1078, opened 2025-08-21
cli: Accept JSON file as input for fetch-deps command

slimreaper35:

### DIFF
--- a/hermeto/interface/cli.py
+++ b/hermeto/interface/cli.py
@@ -66,6 +66,10 @@ FETCH_DEPS_HELP = f"""\
             "gomod-vendor"
         ]
     }}'
+
+    \b
+    # any input as a JSON file
+    {APP_NAME} fetch-deps ./input.json
     """
 
 MERGE_SBOMS_HELP = f"""\
@@ -186,16 +190,28 @@ def main(  # noqa: D103; docstring becomes part of --help message
 
 
 def _if_json_then_validate(value: str) -> str:
+    if _looks_like_input_file(value):
+        try:
+            json.loads(Path(value).read_text())
+        except json.JSONDecodeError:
+            raise typer.BadParameter(f"Looks like JSON file but is not valid JSON file: {value!r}")
+
     if _looks_like_json(value):
         try:
             json.loads(value)
         except json.JSONDecodeError:
             raise typer.BadParameter(f"Looks like JSON but is not valid JSON: {value!r}")
+
     return value
 
 
 def _looks_like_json(value: str) -> bool:
     return value.lstrip().startswith(("{", "["))
+
+
+def _looks_like_input_file(value: str) -> bool:
+    path = Path(value)
+    return path.exists() and path.is_file()
 
 
 class _Input(pydantic.BaseModel, extra="forbid"):
@@ -209,7 +225,7 @@ def fetch_deps(  # noqa: D103; docstring becomes part of --help message
     ctx: typer.Context,
     raw_input: str = typer.Argument(
         ...,
-        help="Specify package (within the source repo) to process. See usage examples.",
+        help="Specify JSON input or path to JSON input file to process. See usage examples.",
         metavar="PKG",
         callback=_if_json_then_validate,
     ),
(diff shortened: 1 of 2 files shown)


hermeto/interface/cli.py
```diff
@@ -191,13 +195,25 @@ def _if_json_then_validate(value: str) -> str:
...
     return value
 
 
 def _looks_like_json(value: str) -> bool:
     return value.lstrip().startswith(("{", "["))
 
 
+def _looks_like_json_file(value: str) -> bool:
+    path = Path(value)
+    return path.exists() and path.suffix == ".json"
```
a-ovchinnikov: This is a nice helper! WDYT about a few more conditions:
```diff
- return path.exists() and path.suffix == ".json"
+ return all((path.exists(), os.access(path, os.R_OK), path.is_file(), path.suffix == ".json"))
```
?
_Edit_: TypeError

slimreaper35: Added a condition if the path is a file. Should be enough IMO.

hermeto/interface/cli.py
```diff
@@ -191,13 +195,25 @@ def _if_json_then_validate(value: str) -> str:
...
     return value
 
 
 def _looks_like_json(value: str) -> bool:
     return value.lstrip().startswith(("{", "["))
 
 
+def _looks_like_json_file(value: str) -> bool:
+    path = Path(value)
+    return all((path.exists(), path.is_file(), path.suffix == ".json"))
```
eskultety: `.json` is no guarantee of anything, I suggest dropping that check, extension checks are a Windows thing; the only proper JSON check is feeding the parser. I can imagine doing `hermeto fetch-deps input` given that `input` is a properly formatted JSON file it should just work.  This is because we're not officially document that `input.json` is the name of the file, it can be any file really, same as with our config file.
You may want to drop & extract or at least rename the function after doing making this change.

hermeto/interface/cli.py
```diff
@@ -191,13 +195,25 @@ def _if_json_then_validate(value: str) -> str:
             json.loads(value)
         except json.JSONDecodeError:
             raise typer.BadParameter(f"Looks like JSON but is not valid JSON: {value!r}")
+
+    if _looks_like_json_file(value):
+        try:
```
eskultety: This handling isn't a fallback, i.e. if the first check fails with an exception, say with `{}.json`, it'll fail with
`Invalid value for 'PKG': Looks like JSON but is not valid JSON: '{}.json'` without even trying the JSON file code path.

Also, and this is a nitpick, the file check feels naturally more robust than the string check IMO, so it might be worth starting with the file check instead of the JSON string check. I'm okay either way as long as a fall back check is implemented.

slimreaper35: That's just wild. Why would anyone use such names as `{}.json`?

eskultety: Well, it's called defensive programming, fixing this doesn't cost us literally anything AND prevents awkward bugs that would pop up if anyone ran some fuzzy testers, e.g. with the help of AI.

slimreaper35: So I swapped the string check with the input file check. If there is a `{}` or `{}.json` file that exists but is not a valid JSON file, does it still make sense to fallback to string check?

eskultety: > file that exists but is not a valid JSON file, does it still make sense to fallback to string check?

No, since you have already validated it is a file that exists, there's no reason to continue with the string check.

hermeto/interface/cli.py
```diff
@@ -186,18 +190,29 @@ def main(  # noqa: D103; docstring becomes part of --help message
...
+
     return value
 
 
 def _looks_like_json(value: str) -> bool:
     return value.lstrip().startswith(("{", "["))
 
 
+def _looks_like_input_file(value: str) -> bool:
+    return Path(value).exists()
```
eskultety: You lost the `is_file()` check from previous iterations, this way it can still be a directory.

tests/unit/test_cli.py
```diff
@@ -372,6 +373,52 @@ def test_specify_packages(
...
+                    "packages": [
+                        {"type": "gomod", "path": "pkg_a"},
+                        {"type": "gomod", "path": "pkg_b"},
+                    ]
+                },
+                [
+                    {"type": "gomod", "path": "pkg_a"},
+                    {"type": "gomod", "path": "pkg_b"},
+                ],
+            ),
```
eskultety: when adding parametrize, please always try to pair a test case with some test ID, it makes the test output more readable and a bit easier to filter.

eskultety (approved):
ACK with minor comments addressed.

taylormadore (approved):

PR #1079, opened 2025-08-21
Add permissive mode design document

slimreaper35:

### DIFF
--- a/docs/design/permissive-mode.md
+++ b/docs/design/permissive-mode.md
@@ -1,0 +1,116 @@
+# Hermeto Permissive Mode
+
+## Overview
+
+This document describes the design for tracking permissive mode violations during
+SBOM generation in Hermeto. The goal is to provide granular tracking of
+violations that occur when running in permissive mode.
+
+## Context
+
+Hermeto, in permissive mode, bypasses certain validation checks and continues
+processing even when issues are encountered. However, this information is not
+currently captured in the generated SBOMs, making it difficult to audit what
+violations occurred during processing.
+
+Current validation checks that are relaxed in the permissive mode:
+
+- `gomod` - the output directory is changed after vendoring
+- `cargo` - the lockfile is not up to date with the project file
+
+## Goals
+
+- Record specific violation, not the whole permissive mode
+- Try to avoid adding the same property to every component
+- Compatibility with both supported SBOM formats - CycloneDX and SPDX
+
+The strict mode remains as default, and the permissive mode is reserved for
+rare and deliberately audited exceptions. By embedding detailed information into
+the SBOM, we will ensure that a more granular and precise policy can be
+implemented in the future.
+
+CycloneDX schema: [http://cyclonedx.org/docs/1.6/json](http://cyclonedx.org/docs/1.6/json)
+SPDX schema: [https://spdx.github.io/spdx-spec/v3.0.1/](https://spdx.github.io/spdx-spec/v3.0.1/)
+
+## Solution
+
+### Option 1
+
+#### CycloneDX
+
+Violations are added as properties in the metadata section:
+[https://cyclonedx.org/docs/1.6/json/#metadata_properties](https://cyclonedx.org/docs/1.6/json/#metadata_properties)
+
+Example:
+
+```json
+{
+    "bomFormat": "CycloneDX",
+    "specVersion": "1.4",
+    "version": 1,
+    "metadata": {
+      "tools": [
+        {
+          "vendor": "red hat",
+          "name": "hermeto"
+        }
+      ],
+      "properties": [
+        {
+          "name": "hermeto:permissive_mode:cargo_lockfile",
+          "value": "Cargo.lock is not synced with Cargo.toml"
+        },
+        {
+          "name": "hermeto:permissive_mode:gomod_vendor_dir",
+          "value": "Go vendor directory inconsistent with go.mod"
+        },
+      ]
+    },
+    "components": []
+  }
+```
+
+#### SPDX
+
+Violations are added as document-level annotations:
+[https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/Annotation/](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/Annotation/)
+
+Example:
+
+```json
+{
+  "spdxVersion": "SPDX-2.3",
+  "SPDXID": "SPDXRef-DOCUMENT",
+  "dataLicense": "CC0-1.0",
+  "name": "my-project",
+  "documentNamespace": "NOASSERTION",
+  "creationInfo": {
+    "created": "2024-01-15T10:30:00Z",
+    "creators": ["Tool: hermeto"]
+  },
+  "packages": [],
+  "relationships": [],
+  "annotations": [
+    {
+      "annotator": "Tool: hermeto",
+      "annotationDate": "2025-01-15T10:30:00Z",
+      "annotationType": "OTHER",
+      "comment": "Cargo.lock is not synced with Cargo.toml"
+    },
+    {
+      "annotator": "Tool: hermeto",
+      "annotationDate": "2025-01-15T10:30:00Z",
+      "annotationType": "OTHER",
+      "comment": "Go vendor directory inconsistent with go.mod"
+    }
+  ]
+}
+```
+
+## Implementation
+
+- [ ] Record permissive violations in the SBOM
+- [ ] Document the behavior for each package manager
+- [ ] Add at least one integration test with SPDX format (or convert existing one)
+- [ ] Make sure Conforma blocks releases with permissive mode
+- [ ] Add mode parameter to prefetch-dependencies task


slimreaper35:
I was wondering whether we should always report the permissive mode in the SBOM even if the prefetch would pass in the strict mode.

taylormadore:
> I was wondering whether we should always report the permissive mode in the SBOM even if the prefetch would pass in the strict mode.

My initial thought is that we probably don't need to if we can reliably report the use of certain features/overrides that are only permitted in permissive mode

a-ovchinnikov:
> I was wondering whether we should always report the permissive mode in the SBOM even if the prefetch would pass in the strict mode

I'd say we should signal violations if they happen rather than just mentioning that we have generally allowed a more lenient approach. From this perspective this would reduce noise. On the other hand mentioning that a  SBOM _could_ have had some violations but does not have them might prove useful for debugging. I wonder if having SBOM verbosity is too fine-grained? I.e. with verbose mode we mention that SBOM is verbose and report everything we could including permissive mode flag?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,116 @@
...
+CycloneDX schema: [http://cyclonedx.org/docs/1.6/json](http://cyclonedx.org/docs/1.6/json)
+SPDX schema: [https://spdx.github.io/spdx-spec/v3.0.1/](https://spdx.github.io/spdx-spec/v3.0.1/)
+
+## Solution
+
+### Option 1
+
+#### CycloneDX
+
+Violations are added as properties in the metadata section:
```
eskultety: I don't think metadata properties are a 1:1 mapping to SPDX annotations. Our SBOMs in either format should be predictable in how particular information is encoded. I think https://cyclonedx.org/docs/1.6/json/#annotations is a better fit and when you look at the fields they match the SPDX ones more closely.
The problem with that is annotations are available only since [v1.5](https://cyclonedx.org/docs/1.5/json/), but we should be able to bump the cyclonedx minor version without consequences or without the need for a major version release.

See more on the annotations topic in my comment below..

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,116 @@
...
+    "creators": ["Tool: hermeto"]
+  },
+  "packages": [],
+  "relationships": [],
+  "annotations": [
+    {
+      "annotator": "Tool: hermeto",
+      "annotationDate": "2025-01-15T10:30:00Z",
+      "annotationType": "OTHER",
+      "comment": "Cargo.lock is not synced with Cargo.toml"
```
eskultety: I think annotations would be a great fit for recording this type of information. However, the way this is proposed I'm a bit skeptical on the delivery - we just have this encoded as a generic document level annotation and the free form text that is left to the consumer to deal with. IIUC our vision when it comes to reporting these permissive mode violations is to do it in a granular fashion, i.e. report specific violations. We also want to enable other SBOM tools to do a deeper analysis on this type of information. However, ^this design is missing 2 crucial elements on how to do that:

1. the proposed annotations have no connection to the violating components and so in a multi-"package" scenarios apply to the whole deliverable which the SBOM was generated for instead of just parts of it
- use the https://spdx.github.io/spdx-spec/v2.3/annotations/#124-spdx-identifier-reference-field to reference the element for which a particular violation was hit
2. The free form text is not machine readable.
- use some machine-readable string formatting for the annotation free-form text (yes, we'll have to invent the format for the string). This might help the SBOM processing tools further filter or categorize the violations we report.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,116 @@
...
+- Record specific violation, not the whole permissive mode
+- Try to avoid adding the same property to every component
+- Compatibility with both supported SBOM formats - CycloneDX and SPDX
+
+The strict mode remains as default, and the permissive mode is reserved for
+rare and deliberately audited exceptions. By embedding detailed information into
+the SBOM, we will ensure that a more granular and precise policy can be
+implemented in the future.
+
+CycloneDX schema: [http://cyclonedx.org/docs/1.6/json](http://cyclonedx.org/docs/1.6/json)
```
eskultety: This is for v1.6 of the schema, we only guarantee 1.4 at the moment.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,116 @@
...
+SPDX schema: [https://spdx.github.io/spdx-spec/v3.0.1/](https://spdx.github.io/spdx-spec/v3.0.1/)
```
eskultety: This references v3.0.1 of the SPDX spec, but we declare that we output v2.3 only: https://github.com/hermetoproject/hermeto/blob/main/docs/usage.md#pre-fetch-dependencies

The link therefore should probably be: https://spdx.github.io/spdx-spec/v2.3/annotations/

eskultety:
> > I was wondering whether we should always report the permissive mode in the SBOM even if the prefetch would pass in the strict mode
> 
> I'd say we should signal violations if they happen rather than just mentioning that we have generally allowed a more lenient approach. From this perspective this would reduce noise. On the other hand mentioning that a SBOM _could_ have had some violations but does not have them might prove useful for debugging. I wonder if having SBOM verbosity is too fine-grained? I.e. with verbose mode we mention that SBOM is verbose and report everything we could including permissive mode flag?

+1 except for this part:

> On the other hand mentioning that a SBOM _could_ have had some violations but does not have them might prove useful for debugging.

Logging the permissive mode usage in logs (log-level=DEBUG) is okay, tainting the SBOM if it doesn't need to be is not.

docs/design/permissive-mode.md
slimreaper35: Does it make sense to investigate bumping the version of the schema for both formats?

eskultety: You need to consider compatibility though: with cyclonedx we're talking minor version bumps which, if cyclonedx conforms to semver correctly, we can achieve without a major release. With SPDX we're talking a new major spec revision and hence you'd need to investigate backwards compatibility. There may not be v3.0.1 tools common out there and if there's no strict compatbility with v2.3 our SPDX would suddenly not be easily parseable.

docs/design/permissive-mode.md
taylormadore: > the proposed annotations have no connection to the violating components

Should we state this as a requirement for the design? I feel like the properties/annotations/metadata that we'll be introducing make more sense linked to the components (even if duplicative between Components in a Hermeto "Package") rather than the SBOM document as a whole. WDYT?

I also agree 100% with the second point.

eskultety: > Should we state this as a requirement for the design?

Good suggestion.

> (even if duplicative between Components in a Hermeto "Package") 

I guess that makes sense wrt to being consistent with how we process things in SBOM today, the question is how formal we want to make this, IOW is it more likely that a generic SBOM processing tool would look for and report annotations or do we expect tailored processors that would "know" to look at particular specialized properties? I'd be okay with either, but have a small preference for the, let's say, slightly more formal approach since you would not lose the component links.

a-ovchinnikov:
>> On the other hand mentioning that a SBOM could have had some violations but does not have them might prove useful for debugging.
>
> Logging the permissive mode usage in logs (log-level=DEBUG) is okay, tainting the SBOM if it doesn't need to be is not.

My suggestion stems from the idea of a self-contained SBOM, i.e. a SBOM that carries as much data about the system and itself as possible. Logs may get scrubbed by the time someone wonders which mode was used to generate this particular SBOM, having more sources means having more points of failure. It is not for debugging of Hermeto itself, but rather whatever it is used to fetch. I realize that not everyone might want that that's why I am suggesting an additional switch for SBOM verbosity which would result in such data inclusion (as machine-readable strings embedded into annotation fields). Something like `hermeto --verbose-sbom fetch pip`. Maybe I am overthinking this one though.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,116 @@
...
+
+Current validation checks that are relaxed in the permissive mode:
+
+- `gomod` - the output directory is changed after vendoring
+- `cargo` - the lockfile is not up to date with the project file
+
+## Goals
+
+- Record specific violation, not the whole permissive mode
+- Try to avoid adding the same property to every component
```
a-ovchinnikov: Why? Is there any estimate of how much extra space that could consume?

docs/design/permissive-mode.md
a-ovchinnikov: We already handle conversion of individual component properties from CyDX to SPDX annotations. Would it not make sense to have this data encoded as a property/annotation of known format per affected component? To make UX slightly better we could add a property listing all offending components to the top of CyDX SBOM and use SPDX package comment section to store the same machine-readable string in SPDX documents. Here I assume CuDX v1.4 and SPDX v2.3. An alternative for SPDX would be to add a phony element describing certain violations and add `EXAMPLE_OF` relationship between an offending SPDXRef and the violation element, but that feels as clunky as embedding data into comments.

slimreaper35: If it is possible to add custom properties only to affected components, that would be good.

I had in mind, for example, in https://github.com/hermetoproject/hermeto/issues/938, what would be the value for users if we include the same information about the permissive flag in each component? Wasn't the permissive mode meant to be the last option if they cannot workaround certain requirements that we have?

On the other hand, I understand the arguments about consistency and linking properties with components. I am thinking about it only from a logical point of view. I don't have much context about machine processing tools, etc.

a-ovchinnikov: > If it is possible to add custom properties only to affected components, that would be good.

Setting extra properties per component looks like something you might want to re-use. You also might be interested in checking out [`to_spdx()`](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/models/sbom.py#L168) and [`to_cyclonedx()`](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/models/sbom.py#L626) methods of the two supported SBOM types.

> what would be the value for users if we include the same information about the permissive flag in each component?

This would probably be an overkill, but having something at the root of a document pointing to affected components could be handy. It could be just a json in the worst case (to make it machine-friendly).

slimreaper35: Is recording various violations in a different way an option?

eskultety:
> > > On the other hand mentioning that a SBOM could have had some violations but does not have them might prove useful for debugging.
> > 
> > 
> > Logging the permissive mode usage in logs (log-level=DEBUG) is okay, tainting the SBOM if it doesn't need to be is not.
> 
> My suggestion stems from the idea of a self-contained SBOM, i.e. a SBOM that carries as much data about the system and itself as possible.

But ^this is not the point of an SBOM, SBOM is pretty much just an attestation (a glorified checksum) of a product, IIUC it must be system agnostic. What we're trying to do is just annotate some components in the SBOM without dumping too much in there. We need to draw a clear line between logs and what SBOM is supposed to be. The fact that someone may scrub logs is a problem on a different level. My personal POV is that we should stick to the original suggestion - *only* enhance the SBOM *when* there would be a violation if it weren't for the permissive mode, do nothing otherwise.

docs/design/permissive-mode.md
eskultety: It just creates noise, if we can avoid it, let's say in an elegant way, we should do that, otherwise, if it requires too much gibberish code --> let's not duplicate.

docs/design/permissive-mode.md
eskultety: > I had in mind, for example, in https://github.com/hermetoproject/hermeto/issues/938, what would be the value for users if we include the same information about the permissive flag in each component? Wasn't the permissive mode meant to be the last option if they cannot workaround certain requirements that we have?
> 
> On the other hand, I understand the arguments about consistency and linking properties with components. I am thinking about it only from a logical point of view. I don't have much context about machine processing tools, etc.

Well, if you just list violations at the top of the file, you have no idea which component generated it, so it's *almost* good as nothing because we'd invent a custom error string for that that wouldn't tell analyzers anything, but linked components would! IOW without understanding much about the error, SBOM analyzers would say "hey, there's an annotation that looks like some kind of violation and is related to these components...."

Ideally, the linking of components should not be intrusive too much, i.e. they should only contain a reference to the annotation without any details, the details would be contained only in the top-level annotation, but take it with a grain of salt, I have almost no visibility into the specs (especially SPDX), so I'd have to dig deep to see what it would actually look like.

eskultety: And again, I'm not so much in favour of recording this as properties, because, yes, that's how do stuff today, doesn't mean it's the way how one should go about annotations in SBOM in general, if we tailor this too much to our existing approaches, then the usefulness of the feature may become questionable.

@a-ovchinnikov out of curiosity, do you think the annotation/reference duplication in components would be solved if we (at some point in the future) converted from a flat list of components to a tree-like structure? IOW have a component e.g. for a `go` package and then a list of subcomponents, then a sibling component for a `yarn` package (in a multi-package repo I mean) with its own subcomponents. That way, if my imagination isn't running too wild, we could only capture the violation/property/annotation only with the top-level package component but not with every component that is in any way related to it, IOW with all dependencies of a project we captured in the SBOM.

docs/design/permissive-mode.md
a-ovchinnikov: I have re-read the goal and if it is _every_ component then I agree, if it is _every affected_ component then I would still suggest amending every affected component.

docs/design/permissive-mode.md
a-ovchinnikov: > the annotation/reference duplication in components would be solved if we ... converted from a flat list of components to a tree-like structure?

We already have a tree in a sense, it just happens to be very wide and not tall at all. The main reason I keep suggesting storing this extra data in components is local context: this way when processing a component one could always tell at a glance whether it were created in strict mode or not. I think that some duplication is fine if it diminishes a chance of mistake (like forgetting to check a top-level SBOM entry with violations). Having said that I don't think this is a deal-breaker, we could store a list of violators together with violations at the top level. Consider something like this:

```
    {
      "annotator": "Tool: hermeto:jsonencoded",
      "annotationDate": "2025-01-15T10:30:00Z",
      "annotationType": "OTHER",
      "comment": "{
          'strict-mode-violations': True,
          'violators': [
               {
                'component': 'foobar-v1.2.3-beta',
                'violation': 'Go vendor directory inconsistent with go.mod',
                'status': 'overridden with permissive flag'*
               }
          ]
        }"
    }
* this one is very optional, I rather agree with the notion of SBOM being a lean
  snapshot than not.
```


This will work for both SPDX and CyDX, with an additional bonus of not needing to have different types of entries for each type of violation in SPDX and not needing to add any relations to individual components. The downside is json in json.

eskultety: > this way when processing a component one could always tell at a glance whether it were created in strict mode or not.

SBOM should be primarily machine readable, and when it comes to human readability, then a top-level annotation will achieve IMO the same thing, the rest is just automagic reference linking.

docs/design/permissive-mode.md
eskultety: > if it is every affected component then I would still suggest amending every affected component.

I think conceptually we all reached a consensus, the only design decision left is IMO to decide whether we want to duplicate the same violation info in all affected components (i.e. the violation string) as properties, or have it more generic with a top-level annotation and only references to the annotations in the affected components. IIUC agreeing on that particular bit will make this PR resolved and we can move onto implementation.

docs/design/permissive-mode.md
slimreaper35: > Is recording various violations in a different way an option?

Any opinion on that?

a-ovchinnikov: > Is recording various violations in a different way an option?

Do you have an example of what you mean? Basing on how it sounds I'd say this is a questionable idea, one would expect uniformity of representation because uniform representation is the least surprising one. Breaking uniformity is done when there is no other good choice available. I do not think that we are in this situation yet, but if you have a good example for why this has to be done then yes.

slimreaper35: I don't have an example. I was just curious if something like that even makes sense.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,221 @@
...
+      "tools": [
+        {
+          "vendor": "red hat",
+          "name": "hermeto"
+        }
+      ],
+      "properties": [
+        {
+          "name": "hermeto:permissive_mode:cargo_lockfile",
+          "value": "Cargo.lock is not synced with Cargo.toml"
```
a-ovchinnikov: This annotation lacks pointers to a violating component. Maybe a json-string with a list of components could be used here? And name could be expanded to explicitly mention the violation: "hermeto:hermeticity_violation:permissive_mode:cargo_lockfile_out_of_sync" or something like that.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,221 @@
...
+Example:
+
+```json
+{
+    "bomFormat": "CycloneDX",
+    "components": [],
+    "annotations": [
+      {
+        "bom-ref": "hermeto:permissive_mode:cargo_lockfile",
+        "subjects": [
```
taylormadore: I believe that subjects needs to be an array of strings, specifically referencing unique bom-refs that we will need to start generating for all Components

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,221 @@
...
+          {
+            "bom-ref": "hermeto:permissive_mode:cargo_lockfile"
+          }
+        ],
+        "annotator": "Tool: hermeto",
```
taylormadore: From the schema it looks like this should be an object instead of a string. Along with the bump to CycloneDX 1.6, we should update the Hermeto tool section to the new format We'll then be able to create a bom-ref for Hermeto to refer to here.

docs/design/permissive-mode.md
taylormadore: This example could probably use at least one Component defined so we can see the whole picture

docs/design/permissive-mode.md
slimreaper35: I think I will drop those examples, since we have decided to go with annotations.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,251 @@
...
+  "components": [
+    {
+      "name": "package",
+      "properties": [
+        {
+          "name": "hermeto:found_by",
+          "value": "hermeto"
+        },
+        {
+          "name": "hermeto:permissive-mode:gomod",
```
a-ovchinnikov: Would adding a word "violation", "warning" or something similar to the name make sense? I believe that would add an immediate clue that something is not right with the component. (As a user I am not sure if "permissive-mode" is bad for my packages, but would likely investigate a "violation").

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,251 @@
...
+
+#### Cons
+
+- **Slightly More Complex:** Requires understanding the reference system, though
+  this is minimal complexity for automated tools.
+
+#### CycloneDX
+
+Violations are added as annotations in the metadata section. Assuming we will
+use version 1.6 of the CycloneDX schema.
```
a-ovchinnikov: Why properties from v1.4 cannot be used?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,251 @@
...
+  }
+}
+```
+
+#### SPDX
+
+### Approach 2: Top-Level Annotation with Component References
+
+This approach involves adding top-level annotations that describe permissive mode
+violations, and each affected component references these annotations by ID. This
```
a-ovchinnikov: "...and each affected component references these annotation..." -- does not this result in duplication? My initial understanding was that the plan is for having a top-level field containing effectively a list of violating components. In other words, are back references really necessary?

eskultety: > My initial understanding was that the plan is for having a top-level field containing effectively a list of violating components. In other words, are back references really necessary?

No references -> no SBOM automation, just free-form strings
back-references -> enables full SBOM post-processing automation

I'd say the reasoning here being that we should IMO work on how our SBOM can be improved in terms of general processing rather than sticking with information represented by custom invented strings only.

docs/design/permissive-mode.md
taylormadore: I agree with this somewhat, but I'm wondering if violation is too strong a word. It's more of an override--but I'm not sure I like that word either... Sorry, I didn't come with any solutions here yet :innocent:

docs/design/permissive-mode.md
a-ovchinnikov: No references != no _back_ references. Consider something like this:
```
{
    'name': 'hermeto:hermeticity_warning:permissive_mode:gomod',
    'value': 'foobar-v.1.2.3;quux-v2.1.4'
}
```
These are direct links from a problem statement to components. Maybe I am missing something fundamental here, but I cannot see much difference between processing a custom entry once or processing it once and then processing an extra entry per component, at least implementation-wise.  Moreover, I do not see much difference between storing a violation type per component vs storing a link to a violation entry stored in some other location: the amount of extra data is not that different in both cases (approximately 30 bytes for a back reference vs approximately 100 bytes for direct annotation). I do not have strong feelings towards any of the implementations and would like to understand what makes one better than another. Both approaches represent an enhancement to the standard (at least for CyDX) and both could be made machine-friendly.

taylormadore: I may be missing something too. I only expected references from the annotations to the affected components--which aligns with the current state of the example. Where are the back-references coming from?

docs/design/permissive-mode.md
slimreaper35: I thought we decided to go with annotations.

a-ovchinnikov: If this is a finalized decision then I recommend replacing "Assuming" with "This requires an upgrade to CycloneDX version 1.6" and adding something along these lines to this paragraph: "an alternative of staying with v1.4 and using properties to store this data was considered and rejected for <xyz> reason".

slimreaper35:
@coderabbitai ignore

docs/design/permissive-mode.md
slimreaper35: I didn't think too much about the property name. It can be discussed later when we start implementing it.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,269 @@
...
+  }
+}
+```
+
+### SPDX
+
+JSON Schema: [https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json](https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json)
+
+Validation can done using the SPDX [online](https://tools.spdx.org/app/validate)
+tool.
```
a-ovchinnikov, line 190: Could you please add a few words on why relationships cannot be used? I think missing proper objects to relate to and the need to craft something in place of those is worth mentioning.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,269 @@
...
+    {
+      "spdxElementId": "SPDXRef-DOCUMENT",
+      "relationshipType": "DESCRIBES",
+      "relatedSpdxElement": "SPDXRef-Package-golang-github.com-gorilla-mux-1.8.0"
+    }
+  ]
+}
+```
+
+#### Approach 2: Top-Level Annotation with Package References
```
a-ovchinnikov: WDYT about mentioning externalRefs as Approach 3?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,353 @@
...
+      "versionInfo": "1.8.0",
+      "filesAnalyzed": false,
+      "externalRefs": [
+        {
+          "referenceCategory": "PACKAGE-MANAGER",
+          "referenceType": "purl",
+          "referenceLocator": "pkg:golang/github.com/gorilla/mux@v1.8.0"
+        },
+        {
+          "referenceCategory": "OTHER",
```
a-ovchinnikov, line 354: We have two alternatives here: `OTHER` and `SECURITY`. `OTHER` is a generic catch-all without any set rules for fields, while `SECURITY` is more strict, but allows to point to a specific URL describing a problem at hand. I think this is worth mentioning in the doc.

docs/design/permissive-mode.md
a-ovchinnikov: Thank you! One more suggestion here: I think it is worth it to mention phony packages (because we've seen a precedent for that). These are not exactly non-standard, non-standard objects won't fit into a SPDX document, but this is a questionable approach nevertheless.

docs/design/permissive-mode.md
slimreaper35: > while SECURITY is more strict, but allows to point to a specific URL describing a problem at hand.

How? Do you mean in the `referenceLocator` field?

a-ovchinnikov: Yes.

slimreaper35:
Added SPDX 3 as the second option.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,360 @@
...
+      "relatedSpdxElement": "SPDXRef-Package-golang-github.com-gorilla-mux-1.8.0"
+    }
+  ]
+}
+```
+
+## Implementation
+
+- Record permissive violations in the SBOM
+- CycloneDX: bump the schema version to 1.6
```
a-ovchinnikov: Wasn't this handled in https://github.com/hermetoproject/hermeto/commit/58afb1cf235911c263a9495abac0099d7fb3e3d6?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,360 @@
...
+- Document the behavior for each package manager
```
a-ovchinnikov: s/for each package manager/for each affected package manager/?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,360 @@
...
+- Add at least one integration test with SPDX format (or convert existing ones)
```
a-ovchinnikov: Is this related to the proposed change?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,360 @@
...
+JSON Schema: [https://cyclonedx.org/docs/1.4/json](https://cyclonedx.org/docs/1.4/json)
+
+Validation can be done using the CycloneDX [CLI](https://github.com/CycloneDX/cyclonedx-cli)
+tool.
+
+```bash
+cyclonedx validate --input-file bom.json
+```
+
+#### Approach 1: Linked-Component Properties ❌
```
a-ovchinnikov: I think it would have been nice to have a one or two sentences comment on why this approach was not chosen, especially given that properties already map to SPDX annotation (and vise versa). This suggestion  applies to all marked as rejected approaches lacking any commentary on why they were not selected.

Another thing which I won't push too hard for: do we need saltires and check marks right next to title elements? It could be just me, but when I see something marked with a crossout I initially tend to skip it. I would suggest avoiding excessive typography tricks like bold font and bright marks since they attract reader's attention and when done too often that dilutes it. _Generally_ when you want to _emphasize_ something it is better to do that with a slanted font: it does not interrupt the way one reads a text. Again, this is my take on the problem, so I won't be pushing for it.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,360 @@
...
+    {
+      "name": "github.com/gin-gonic/gin",
+      "properties": [
+        {
+          "name": "hermeto:found_by",
+          "value": "hermeto"
+        },
+        {
+          "name": "hermeto:permissive-mode:violation:gomod:vendor-directory-changed-after-vendoring",
+          "value": "vendor directory changed after vendoring"
```
a-ovchinnikov: Is there any good reason for duplicating this info? The same could be rewritten as
```json
{
    "name": "hermeto:permissive-mode:violation",
    "value": "gomod:vendor-directory-changed-after-vendoring"
}
```
without any loss.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,360 @@
...
+```
+
+### SPDX
+
+JSON Schema: [https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json](https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json)
+
+Validation can be done using the SPDX [online](https://tools.spdx.org/app/validate)
+tool.
+
+**Note:** Relationships cannot be used for tracking permissive mode violations.
```
a-ovchinnikov: WDYT about adding "SPDX lets one define relationships between its elements, the specification provide multiple variant of those. Unfortunately relationships cannot be used...". The way it is written now assumes that every reader is familiar with SPDX spec.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,360 @@
...
+The fundamental issue is that relationships require proper SPDX elements to
+relate to, but permissive mode violations are not first-class objects in the SPDX
+specification. While it would be possible to create artificial elements to represent
```
a-ovchinnikov: I am sorry, I lost the track a bit, which artificial elements are you referring to here?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,360 @@
...
+violations, this would require crafting non-standard objects that don't align with
+SPDX's intended data model and could confuse existing SPDX tooling.
+
+An alternative approach would be to create "phony packages" that represent violations
+as SPDX package elements, which could then be linked via relationships. While phony
+packages are not non-standard objects, this approach is questionable as it pollutes
+the package list with non-package entities and could mislead tools that expect
+actual software packages.
+
+#### Approach 1: Linked-Component Annotations ✅
```
a-ovchinnikov: What is a linked-component?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,360 @@
...
+      "relatedSpdxElement": "SPDXRef-Package-golang-github.com-gorilla-mux-1.8.0"
+    }
+  ]
+}
+```
+
+#### Approach 2: Top-Level Annotation with Package References ❌
+
+This approach involves adding top-level annotations that describe permissive mode
+violations, and referencing the affected packages by their IDs.
```
a-ovchinnikov: Are you sure this is correct? Judging from the schema linked above you can add a top-level annotation and nothing prevents you from listing all affected SPDXIDs in the comment section. What cannot be done with v2.3 is having an annotation and then referencing it from someplace else (because, as you correctly note below, it does not have SPDXID).

docs/design/permissive-mode.md
slimreaper35: I will drop the whole section.

docs/design/permissive-mode.md
slimreaper35: I will drop the whole section.

docs/design/permissive-mode.md
slimreaper35: I will drop the whole section.

docs/design/permissive-mode.md
slimreaper35: I forgot to update this example.

docs/design/permissive-mode.md
slimreaper35: phony packages :D, fixed

docs/design/permissive-mode.md
slimreaper35: updated all titles

docs/design/permissive-mode.md
slimreaper35: dropped all xmarks and green checks

docs/design/permissive-mode.md
slimreaper35: Since we bumped the version, I also deleted the section about bumping the schema for approach 1.

docs/design/permissive-mode.md
a-ovchinnikov: Thank you!

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,349 @@
...
+    }
+  ]
+}
+```
+
+#### Approach 2 (alternative): Top-Level annotations with package references
+
+This approach involves adding top-level annotations that describe permissive mode
+violations and referencing the affected components by their IDs. Similar to the
+CycloneDX approach 1.
```
a-ovchinnikov, line 293: I still don't understand what blocks us from having a top-level annotation referencing affected elements. Could you please elaborate a bit? I believe something like this is absolutely possible:
```json
{
  "spdxVersion": "SPDX-2.3",
  "dataLicense": "CC0-1.0",
  "SPDXID": "SPDXRef-DOCUMENT",
  "name": "unknown",
  "documentNamespace": "NOASSERTION",
  "creationInfo": {
    "licenseListVersion": "3.24",
    "creators": ["Organization: Red Hat, Inc", "Tool: hermeto"],
    "created": "2025-01-15T10:30:00Z"
  },
  "annotations": [
    {
         "annotator": "Tool: hermeto",
         "annotationDate": "2025-01-15T10:30:00Z",
         "annotationType": "OTHER",
         "comment": "hermeto:permissive-mode:violation:gomod:vendor-directory-changed-after-vendoring::SPDXRef-Package-golang-github.com-gin-gonic-gin-1.9.1;SPDXRef-Package-golang-github.com-gorilla-mux-1.8.0"
        }
      ]
  "packages": [
    {
      "name": "github.com/gin-gonic/gin",
      "downloadLocation": "NOASSERTION",
      "SPDXID": "SPDXRef-Package-golang-github.com-gin-gonic-gin-1.9.1",
      "versionInfo": "1.9.1",
      "filesAnalyzed": false,
      "externalRefs": [ ... ],

    },
    {
      "name": "github.com/gorilla/mux",
      "downloadLocation": "NOASSERTION",
      "SPDXID": "SPDXRef-Package-golang-github.com-gorilla-mux-1.8.0",
      "versionInfo": "1.8.0",
      "filesAnalyzed": false,
      "externalRefs": [ ... ]
    }
  ],
  "relationships": [ ... ]
}
```

Am I reading [schema](https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json) wrong?

docs/design/permissive-mode.md
a-ovchinnikov: Thank you, now they are much easier to comprehend.

docs/design/permissive-mode.md
slimreaper35: Where is the reference to the components/packages?

a-ovchinnikov: Here it is: `"comment": "hermeto:permissive-mode:violation:gomod:vendor-directory-changed-after-vendoring::SPDXRef-Package-golang-github.com-gin-gonic-gin-1.9.1;SPDXRef-Package-golang-github.com-gorilla-mux-1.8.0"`

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,353 @@
...
+Validation checks that are currently relaxed in the permissive mode:
+
+- `gomod` - vendor directory contents differ from go.mod/go.sum after vendoring
+- `cargo` - Cargo.lock is out of sync with Cargo.toml (it must exist though)
+
+## Goals
+
+- Record specific violation, not the whole permissive mode
+- Compatibility with both supported SBOM formats - CycloneDX and SPDX
+- Machine-readable format for easy parsing and processing
```
a-ovchinnikov: Non-blocking nitpick:
```diff
- Machine-readable format for easy parsing and processing
+ Present violations in machine-readable format for easy parsing and processing

```

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,353 @@
...
+
+## Solution
```
a-ovchinnikov: Non-blocking nitpick: s/Solution/Proposed Solutions/ because there are many solutions and some of them are not selected.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,353 @@
...
+```bash
+cyclonedx validate --input-file bom.json
+```
+
+#### Approach 1 (preferred): Top-level annotation with component references
+
+This approach involves adding top-level annotations that describe permissive mode
+violations and referencing the affected components by their IDs.
+
+Example:
```
a-ovchinnikov: Non-blocking suggestion: WDYT about hiding these examples in a drop-down field?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,353 @@
...
+```
+
+### SPDX
+
+JSON Schema: [https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json](https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json)
+
+Validation can be done using the SPDX [online](https://tools.spdx.org/app/validate)
+tool.
+
+#### Approach 1 (preferred): Annotations within the packagess
```
a-ovchinnikov: s/packagess/packages

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,353 @@
...
+}
+```
+
+#### Approach 2 (alternative): Top-level annotations with package references
+
+This approach involves adding top-level annotations that describe permissive mode
+violations and referencing the affected components by their IDs. Similar to the
+CycloneDX approach 1.
+
+However, only since SPDX version 3, annotations can become actual SPDX elements
```
a-ovchinnikov: This is irrelevant to approach with a list of SPDXIDs in a well-defined annotation. To keep this paragraph I think you would need to at least add something like this: "SPDX spec allows defining relationships between SPDX elements of multitude of types. Two `spdxElementId` are required for that one per element in a relationship.". Maybe this should be moved to `Approach 2.5` section?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,353 @@
...
+      {
+        "vendor": "red hat",
+        "name": "hermeto"
+      }
+    ]
+  }
+}
+```
+
+### SPDX
```
a-ovchinnikov: The very important bit of information that is missing is that we are using SPDX v2.3 and are unlikely to switch to v3.0 in the near future. I think it is necessary to add a clear statement that by SPDX in this document you mean v2.3 unless stated otherwise.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,353 @@
...
+
+JSON Schema: [https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json](https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json)
+
+Validation can be done using the SPDX [online](https://tools.spdx.org/app/validate)
+tool.
+
+#### Approach 1 (preferred): Annotations within the packagess
+
+This approach involves adding annotations directly to each package that describe
+permissive mode violations.
```
a-ovchinnikov: Why is this approach preferred?

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,353 @@
...
+JSON Schema: [https://cyclonedx.org/docs/1.6/json](https://cyclonedx.org/docs/1.6/json)
+
+Validation can be done using the CycloneDX [CLI](https://github.com/CycloneDX/cyclonedx-cli)
+tool.
+
+```bash
+cyclonedx validate --input-file bom.json
+```
+
+#### Approach 1 (preferred): Top-level annotation with component references
```
a-ovchinnikov, line 51: I would strongly suggest adding a few words about why this approach is preferred.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,353 @@
...
+However, only since SPDX version 3, annotations can become actual SPDX elements
+with a `spdxElementId` [property](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/Annotation/)
+which means, that we could use the `relationships` field to link them to the affected
+packages.
+
+#### Approach 3 (alternative): External references for violation tracking
+
+This approach uses the `externalRefs` field to add custom reference types that describe
+permissive mode violations. This method leverages SPDX's existing external reference
+system to embed violation metadata in a structured, machine-readable format.
```
a-ovchinnikov, line 304: `SECURITY` externalRef can contain a `URL`. What is the reason for not using that?

slimreaper35: Why do we need to keep expanding this document about potential solutions that we will not implement?

a-ovchinnikov: The document as it is written leaves a lot of reasoning behind the presented decisions hidden. It does a good job of answering "what will be done", but is sparse on "why". These are the questions which someone somewhat familiar with the spec would have asked. It does not require much to answer them, a few sentences in total, but that will come handy once we need to figure out why we chose an implementation.

slimreaper35: I added short paragraphs about the chosen approaches and the benefits.

I still don't understand why we need to explore and elaborate on alternative solutions that we will not implement, as we already have one for each SBOM format. If we decide to change the recording of violations in the SBOM in the future, we can continue the investigation.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,384 @@
...
+
+This document describes the design for tracking permissive mode violations during
+SBOM generation in Hermeto. The goal is to provide granular tracking of these
+violations in the SBOM.
+
+## Context
+
+Hermeto, in permissive mode, bypasses certain validation checks and continues
+processing even when issues are encountered. However, this information is not
+currently captured in the generated SBOMs, making it difficult to audit what
```
taylormadore: "audit any violations bypassed during processing"

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,384 @@
...
+</details>
+
+### SPDX
+
+- JSON Schema: [https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json](https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json)
+- Validation can be done using the SPDX [online](https://tools.spdx.org/app/validate)
+tool.
+
+**NOTE:** We are using SPDX version 2.3 right now. The latest version is however,
+3.1 at the time of writing. We are not planning to upgrade to 3.1 in the near future.
```
taylormadore: nit: 3.0.1 is the latest released version. 3.1 is still under development

docs/design/permissive-mode.md
a-ovchinnikov: Thank you. Please note, that my comments are not about the investigation, it is over now and we know enough about all approaches, I am concerned that we write down what we know and remember at the moment for the benefit of future maintainers and implementers (probably us again, but having context-switched enough to forget most of this). Without this information the document is useless and should have been just ~40 lines of code.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,384 @@
...
+
+#### Approach 1 (preferred): Annotations within the packages
+
+This approach involves adding annotations directly to each package that describe
+permissive mode violations. Annotations are currently used for converting CycloneDX
+component properties.
+
+*This is the most practical and well-supported method for attaching custom,
```
a-ovchinnikov: I am lost, which is best, per component or top-level? One part of the document says one thing, another one another. I will not block on this now, however I think it is better to refer to just one thing as the best.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,384 @@
...
+package-specific metadata in SPDX 2.3. Since the annotations field is already
+used in Hermeto to convert CycloneDX properties, using it for violation tracking
```
a-ovchinnikov: The document explicitly tells us that top-level is chosen and preferred, not per-component (L56 in this revision). There is arbitrary and underarticulated mismatch between chosen approaches.

a-ovchinnikov (approved):
I am approving the document to unblock the feature despite still being convinced that the choice is rather arbitrary and mismatching. I think this will not result in excessive maintenance burden and that we can go with it as is.

docs/design/permissive-mode.md
slimreaper35: Oh, thanks. That is some AI leftover. Fixed.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,386 @@
...
+          "referenceType": "purl",
+          "referenceLocator": "pkg:golang/github.com/gin-gonic/gin@v1.9.1"
+        }
+      ],
+      "annotations": [
+        {
+          "annotator": "Tool: hermeto",
+          "annotationDate": "2025-01-15T10:30:00Z",
+          "annotationType": "OTHER",
+          "comment": "hermeto:permissive-mode:violation:gomod:vendor-directory-changed-after-vendoring"
```
eskultety, line 245: We might need to give this string a bit more thought, but that's an implementation detail, irrelevant to the doc.

docs/design/permissive-mode.md
eskultety: > I still don't understand why we need to explore and elaborate on alternative solutions that we will not implement, as we already have one for each SBOM format. If we decide to change the recording of violations in the SBOM in the future, we can continue the investigation.

I tend to have the same feeling about these docs as it spans across other design docs we already have. That said, looking at the matter from a different angle, it's not a bad thing to have a reference in the future when nobody remembers anything anymore on what approaches we considered, so there is some value, but maybe a follow-up improvement could be that we put all the considered solutions to a different section for better optical separation, IOW make it so that the reader won't feel the urge (or even exhaustion) to read the rest unless they're really skeptical about the approach we ended up implementing :) .

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,386 @@
...
+      "relationshipType": "DESCRIBES",
+      "relatedSpdxElement": "SPDXRef-Package-golang-github.com-gorilla-mux-1.8.0"
+    }
+  ]
+}
+```
+
+</details>
+
+#### Approach 4 (alternative): Relationships with phony packages
```
eskultety, line 378: This last one may be a bit of a stretch - was this really considered or is it just a mention of "a possibility" but at no point was anyone really convinced we'd want to do that? Non-blocking, but I could imagine seeing this dropped completely.

docs/design/permissive-mode.md
```diff
@@ -0,0 +1,386 @@
...
+
+</details>
+
+#### Approach 2 (alternative): Top-level annotations with package references
+
+This approach involves adding top-level annotations that describe permissive mode
+violations and referencing the affected components by their IDs. Similar to the
+CycloneDX approach 1.
+
+Unfortunately, only since SPDX version 3, annotations can become actual SPDX elements
```
eskultety, line 295: Out of curiosity, is ^this still on the table with a future SPDX 3.0 support bump or do we intend to keep the other approach due to compatibility? 3.0 features would be nice, but compatibility is more important.

slimreaper35: > Out of curiosity, is ^this still on the table with a future SPDX 3.0 support bump or do we intend to keep the other approach due to compatibility?

That would be on the table if we have a plan to adopt SPDX 3 in the near future, which does not seem realistic.

> 3.0 features would be nice, but compatibility is more important.

I also raised a question of supporting both versions, if that even makes sense.

taylormadore: I think it should be on the table if/when we add support for SPDX 3.0 in a future major release of Hermeto

docs/design/permissive-mode.md
slimreaper35: @a-ovchinnikov has proposed it as a corner-case solution.

PR #1081, opened 2025-08-22
Add support for oci:// URLs to fetch RPMs from OCI registries

gnotlooy:
## Summary
- Implement OCI registry support for RPM downloads using oci:// URL scheme
- Parse oci:// URLs and extract registry/repository/tag components  
- Authenticate using standard ~/.docker/config.json mechanisms
- Fetch image manifests via OCI Registry API and match layers by checksum
- Download specific layer blobs directly without pulling entire image

## Test plan
- [ ] Test with valid oci:// URLs in rpms.lock.yaml
- [ ] Verify authentication works with Docker config
- [ ] Confirm checksum matching finds correct RPM layers
- [ ] Test error handling for invalid URLs and missing layers
- [ ] Run existing RPM tests to ensure no regressions

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -99,15 +99,145 @@ async def _async_download_binary_file(
     log.debug(f"Download completed - {url}")
 
 
+async def _async_download_oci_file(
+    oci_url: str,
+    download_path: Union[str, PathLike[str]],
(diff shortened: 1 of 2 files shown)


lcarva:
This is a hot mess. Let's close this out and I'll work with @gnotlooy to make this better.

PR #1082, opened 2025-08-26
noxfile: Group linters and static analyzers to a single session - lint

eskultety:
Each linter invocation installs the whole requirements-extras.txt file which results in 1.5G of disk space for just individual nox linter sessions.

    $ du -h -d1 .nox
    294M .nox/bandit
    313M .nox/mypy
    294M .nox/flake8
    294M .nox/isort
    294M .nox/black
    1.5G .nox/

I explored the possibility of not installing the full requirements-extras.txt file, but the problem is our resolved and locked requirements file would lose its meaning as we'd always be installing the latest (any) package from pypi. Additionally, the current code transcript relying on a single file is much cleaner than enumerating individual dependencies for a session, so keeping the current approach is better.

The linters take only a few seconds when reusing the virtual environment (pre-existing), so there's no loss in running all linters vs invoking individual linters on nox's CLI.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -22,7 +22,7 @@ import nox
 from nox.sessions import Session
 
 # default sessions to run (sorted alphabetically)
-nox.options.sessions = ["bandit", "black", "flake8", "isort", "mypy", "python"]
+nox.options.sessions = ["lint", "python"]
 
 # reuse virtual environment for all sessions
 nox.options.reuse_venv = "always"
@@ -45,43 +45,26 @@ def parse_supported_python_versions() -> list[str]:
 
 
 @nox.session()
-def bandit(session: Session) -> None:
-    """Run bandit."""
+def lint(session: Session) -> None:
+    """Run linters."""
+    exc = None
     install_requirements(session)
-    cmd = "bandit -c pyproject.toml -r hermeto noxfile.py"
-    session.run(*cmd.split(), *session.posargs, silent=True)
+    cmds = [
+        "bandit -c pyproject.toml -r hermeto noxfile.py",
+        "black --check --diff hermeto tests noxfile.py",
+        "flake8 hermeto tests noxfile.py",
+        "isort --check --diff --color hermeto tests noxfile.py",
+        "mypy --install-types --non-interactive hermeto tests noxfile.py",
+    ]
 
+    for cmd in cmds:
+        try:
+            session.run(*cmd.split(), *session.posargs, silent=True)
+        except Exception as e:
+            exc = e
 
-@nox.session()
-def black(session: Session) -> None:
-    """Run black."""
-    install_requirements(session)
-    cmd = "black --check --diff hermeto tests noxfile.py"
-    session.run(*cmd.split(), *session.posargs, silent=True)
-
-
-@nox.session()
-def flake8(session: Session) -> None:
-    """Run flake8."""
-    install_requirements(session)
-    cmd = "flake8 hermeto tests noxfile.py"
-    session.run(*cmd.split(), *session.posargs, silent=True)
-
-
-@nox.session()
-def isort(session: Session) -> None:
-    """Run isort."""
-    install_requirements(session)
-    cmd = "isort --check --diff --color hermeto tests noxfile.py"
-    session.run(*cmd.split(), *session.posargs, silent=True)
-
-
-@nox.session()
-def mypy(session: Session) -> None:
-    """Run mypy."""
-    install_requirements(session)
-    cmd = "mypy --install-types --non-interactive hermeto tests noxfile.py"
-    session.run(*cmd.split(), *session.posargs, silent=True)
+    if exc:
+        raise exc
 
 
 @nox.session(name="python", python=parse_supported_python_versions())
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -48,13 +48,6 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      matrix:
-        nox_env:
-          - bandit
-          - black
-          - isort
-          - flake8
-          - mypy
 
     container:
       image: python:3.9-slim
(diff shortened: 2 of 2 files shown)


noxfile.py
```diff
@@ -45,43 +45,27 @@ def parse_supported_python_versions() -> list[str]:
...
-
-
-@nox.session()
-def isort(session: Session) -> None:
-    """Run isort."""
-    install_requirements(session)
-    cmd = "isort --check --diff --color hermeto tests noxfile.py"
-    session.run(*cmd.split(), *session.posargs, silent=True)
+    cmds = [
+        ("bandit", "-c", "pyproject.toml", "-r", "hermeto", "noxfile.py"),
```
a-ovchinnikov: Non-blocking question: what is the motivation behind replacing strings with tuples? The point behind strings was to make the commands easier to read, strings with extra punctuation make a reader exert another little effort, I think it is unnecessary in this specific use case.

noxfile.py
```diff
@@ -45,43 +45,27 @@ def parse_supported_python_versions() -> list[str]:
...
+            fail = True
 
-@nox.session()
-def mypy(session: Session) -> None:
-    """Run mypy."""
-    install_requirements(session)
-    cmd = "mypy --install-types --non-interactive hermeto tests noxfile.py"
-    session.run(*cmd.split(), *session.posargs, silent=True)
+    if fail:
+        session.error()
```
a-ovchinnikov: UX suggestion. With the code written as is on a failure nox terminates with
```
nox > Session lint aborted.
```
Maybe it is just me, but I was confused by this output, initially I thought that my error caused the entire linters sequence to crash midway. Upon further investigation I learned that in reality all linters checked my code, several complained and only _then_ nox reported that a session was aborted because that's apparently what it does when someone calls `session.error()`

I tried a different approach:
```python
     exc = False
     for cmd in cmds:
         try:
             session.run(*cmd, *session.posargs, silent=True)
         except Exception as e:
             exc = e
     if exc:
         raise exc
```
This resulted in
```
nox > Session lint failed.
```
The exact nature of an exception does not matter as long as this is any of the regular session exceptions so I just reraised the last one.

eskultety: I caught the exception, inspected it in the debugger, there was nothing, I simply didn't try re-raising it. Will apply.

noxfile.py
eskultety: - long strings would overflow the line length and there's no nice string wrapping really
- I applied the bot's suggestion and also remembered some past "fun" issues with the podman pip-compile session, so I converted it (I can revert it):
> Avoid cmd.split(): naive splitting breaks on quoted args and is brittle. Pass args as lists.

eskultety:
> Would it make sense to pass specific linter name trough `session.posargs` and filter linters on posargs if present? (I do like to hunt down my mypy errors separately)

We could, but that would require further custom parsing which I wanted to avoid.

noxfile.py
a-ovchinnikov: That all is fair in a general case, but here we are in full control of what we are feeding to split(), be it line length-wise or arguments structure-wise. I won't insist too hard on reverting, but please consider
```
"bandit -c pyproject.toml -r hermeto noxfile.py"
```
vs
```
"bandit", "-c", "pyproject.toml", "-r", "hermeto", "noxfile.py"
```

a-ovchinnikov:
> We could, but that would require further custom parsing which I wanted to avoid.

Fair enough, I tried to come up with a quick solution for the problem and it looks like it is doable, but is more work than it is worth. This is definitely a non-blocker.

slimreaper35 (approved):

slimreaper35:
So this PR is replaced by https://github.com/hermetoproject/hermeto/pull/1084 ?

eskultety:
> So this PR is replaced by #1084 ?

Can be, but doesn't need to be. The other was just a learning experiment and ultimately to test the willingness to adopt it. #1084 doesn't incorporate (deliberately) the inclusion of mypy to the lint session. Depending on the consensus I can close this PR and incorporate it in the other one, or just update #1084 after merging this one. I wouldn't say this one is replaced, more like enhanced :).

eskultety:
> > We could, but that would require further custom parsing which I wanted to avoid.
> 
> Fair enough, I tried to come up with a quick solution for the problem and it looks like it is doable, but is more work than it is worth. This is definitely a non-blocker.

So, I did an experiment with `notify`:
```diff
+@nox.session
+def __install(session: Session) -> None:
+    """Install dependencies in a shared virtual environment for linters."""
+    # Get the linter name from session.posargs
+    install_requirements(session)
+
+    if session.posargs:
+        for task in session.posargs[0]:
+            task(session)
+
+@nox.session(venv_backend="none")
+def black(session: Session) -> None:
+    """Run black."""
+    session.notify("__install", posargs=([_black], *session.posargs))
+
+
+def _black(session) -> None:
+    cmd = "black hermeto/ tests"
+    session.run(*cmd.split(), *session.posargs[1:])
```
That would kind of achieve what you want @a-ovchinnikov, but at the same time `notify` can only be run once, so you could either do `nox -s lint` or `nox -s mypy` but not `nox -s black isort mypy` because only the first item would run. I could do that, but it'd IMO be confusing for anyone not familiar with this hack. Nox has had an open issue for cross-session virtualenv sharing for a while, but to no avail.

noxfile.py
eskultety: Reverted.

a-ovchinnikov:
> That would kind of achieve what you want @a-ovchinnikov, but at the same time notify can only be run once...

I experimented a bit with converting the command strings list to mapping from linter name to linter arguments string and then running either all or just those selected with `-- linter1 ... linterN`, it did not look pretty, likely had some corner cases and felt like an overkill. I am ok with merging it as it is now, I'll post a PR if I ever find the new behavior too upsetting.

eskultety:
> > That would kind of achieve what you want @a-ovchinnikov, but at the same time notify can only be run once...
> 
> I experimented a bit with converting the command strings list to mapping from linter name to linter arguments string and then running either all or just those selected with `-- linter1 ... linterN`, it did not look pretty, likely had some corner cases and felt like an overkill. I am ok with merging it as it is now, I'll post a PR if I ever find the new behavior too upsetting.

Okay, I'll merge then.Hopefully you won't need to if we go with Ruff,  because not only it gets rid of most linters, it's also, unsurprisingly, fast, so it should not get in the way too much :)

Issue #876, opened 2025-03-19
More exceptions classes are needed


a-ovchinnikov:
As it is now cachi2 uses a limited set of classes to signal different exceptional conditions. A lot of contextual information gets recorded as part of an exception message. As a result many tests end up verifying not just an exception that occurred, but also check if a specific message was emitted. This is error-prone and introduces excessive coupling between tests and code, even trivial changes to user-facing messages result in need to change tests.

To resolve this issue it is suggested to introduce a richer set of subclasses for existing exceptions, define error messages in them and amend the test suit to verify exception types instead of verifying exception messages.

a-ovchinnikov:
This is a follow-up to several discussions, the most recent of which took place in https://github.com/hermetoproject/cachi2/pull/867/commits/207f5a471ae33eff2cb17f9ddb7356acd16e599b#r2000983385

brunoapimentel:
One thing that is worth keeping is trying to group the errors into user errors and system errors (or other wider classification categories). Maybe keeping a smaller set of superclasses is a way to achieve this.

The reason for this is to make it easier for filtering logs in log aggregation tools (for instance, I want to see how many actual system errors happened in all pipelines that ran in a single day, but I want to disregard errors caused by user configuration).

derasdf:
One important part of this process should involve examining the already existing use cases of errors and tweaking them appropriately for the new classes of exceptions, as well as maybe creating new test cases for these errors. Additionally, this can help resolve issues like `BaseError` being raised in some files (e.g., `utils.py`, `output.py`...), even though it’s explicitly commented in the `BaseError` that `"Root of the error hierarchy. Don't raise this directly, use more specific error types."`

eskultety:
> One thing that is worth keeping is trying to group the errors into user errors and system errors (or other wider classification categories). Maybe keeping a smaller set of superclasses is a way to achieve this.

@brunoapimentel ^this should be tracked in a separate issue since on the result of the issue at hand will just be more granularity, but in the end the end user is not supposed to see an exception ever, so this is a matter of a formatter ---> separate issue :)

PR #1083, opened 2025-08-26
Add git submodule support for Go vendor directory detection

### LINKED ISSUE #1029: Strange interaction between git submodules and gomod processing
I've got two repositories.

* http://github.com/konflux-ci/yq-container
* http://github.com/konflux-ci/git-clone

Both of them:

* Have git submodules referring to the "real" project.
* Have go dependencies.

However, I can successfully fetch deps for the first one (yq-container) but not for the second one (git-clone).

If I go directly to the upstream for each, I can successfully fetch the deps of each project.

I'm not sure what's going on or why its failing. Any insight from the maintainers here is appreciated.

## yq-container

✔️ Upstream:

```
❯ gh repo clone mikefarah/yq && cd yq/ && hermeto fetch-deps '{"packages": [{"path": ".", "type": "gomod"}], "flags": ["gomod-vendor"]}'
Cloning into 'yq'...
remote: Enumerating objects: 17435, done.
remote: Counting objects: 100% (254/254), done.
remote: Compressing objects: 100% (168/168), done.
remote: Total 17435 (delta 155), reused 86 (delta 86), pack-reused 17181 (from 3)
Receiving objects: 100% (17435/17435), 11.43 MiB | 27.74 MiB/s, done.
Resolving deltas: 100% (11847/11847), done.
2025-07-17 11:36:19,495 WARNING The `gomod-vendor` flag is deprecated and will be removed in future versions. Note that it will no longer perform automatic vendoring, so in case vendoring is needed, it needs to be manually added to the source repository.
2025-07-17 11:36:20,250 INFO Fetching the gomod dependencies at subpath .
2025-07-17 11:36:20,273 INFO go.mod reported versions: '1.24'[go], '1.24.1'[toolchain]
2025-07-17 11:36:20,273 INFO Using Go release: go1.21.0
2025-07-17 11:36:25,049 INFO Downloading the gomod dependencies
2025-07-17 11:36:26,607 INFO Retrieving the list of packages
2025-07-17 11:36:26,939 INFO All dependencies fetched successfully \o/
```

✔️ Downstream:

```
❯ gh repo clone konflux-ci/yq-container && cd yq-container/ && git submodule init && git submodule update && hermeto fetch-deps '{"packages": [{"path": "yq", "type": "gomod"}], "flags": ["gomod-vendor"]}'
Cloning into 'yq-container'...
remote: Enumerating objects: 676, done.
remote: Counting objects: 100% (217/217), done.
remote: Compressing objects: 100% (38/38), done.
remote: Total 676 (delta 203), reused 179 (delta 179), pack-reused 459 (from 1)
Receiving objects: 100% (676/676), 259.16 KiB | 3.46 MiB/s, done.
Resolving deltas: 100% (427/427), done.
Submodule 'yq' (https://github.com/mikefarah/yq) registered for path 'yq'
Cloning into '/home/rbean/code/experiment/yq-container/yq'...
Submodule path 'yq': checked out '2c7cce0878c02afe93986cacb27eab690d3676e0'
2025-07-17 11:37:44,783 WARNING The `gomod-vendor` flag is deprecated and will be removed in future versions. Note that it will no longer perform automatic vendoring, so in case vendoring is needed, it needs to be manually added to the source repository.
2025-07-17 11:37:45,460 INFO Fetching the gomod dependencies at subpath yq
2025-07-17 11:37:45,492 INFO go.mod reported versions: '1.23.0'[go], '-'[toolchain]
2025-07-17 11:37:45,492 INFO Using Go release: go1.21.0
2025-07-17 11:37:49,615 INFO Downloading the gomod dependencies
2025-07-17 11:37:51,062 INFO Retrieving the list of packages
2025-07-17 11:37:51,397 INFO All dependencies fetched successfully \o/
```

## git-clone

✔️ Upstream:

```
❯ gh repo clone tektoncd-catalog/git-clone && cd git-clone && hermeto fetch-deps '{"packages": [{"path": "image/git-init", "type": "gomod"}], "flags": ["gomod-vendor"]}'
Cloning into 'git-clone'...
remote: Enumerating objects: 5422, done.
remote: Counting objects: 100% (173/173), done.
remote: Compressing objects: 100% (98/98), done.
remote: Total 5422 (delta 89), reused 75 (delta 75), pack-reused 5249 (from 1)
Receiving objects: 100% (5422/5422), 8.72 MiB | 19.50 MiB/s, done.
Resolving deltas: 100% (2204/2204), done.
2025-07-17 11:39:08,982 WARNING The `gomod-vendor` flag is deprecated and will be removed in future versions. Note that it will no longer perform automatic vendoring, so in case vendoring is needed, it needs to be manually added to the source repository.
2025-07-17 11:39:09,865 INFO Fetching the gomod dependencies at subpath image/git-init
2025-07-17 11:39:10,072 INFO go.mod reported versions: '1.23.0'[go], '-'[toolchain]
2025-07-17 11:39:10,072 INFO Using Go release: go1.21.0
2025-07-17 11:39:14,380 INFO Vendoring the gomod dependencies
2025-07-17 11:39:20,735 INFO Retrieving the list of packages
2025-07-17 11:39:21,399 INFO All dependencies fetched successfully \o/
```

❌ Downstream:

```
❯ gh repo clone konflux-ci/git-clone && cd git-clone && git submodule init && git submodule update && hermeto fetch-deps '{"packages": [{"path": "git-clone/image/git-init", "type": "gomod"}], "flags": ["gomod-vendor"]}'                           
Cloning into 'git-clone'...
remote: Enumerating objects: 126, done.
remote: Counting objects: 100% (60/60), done.
remote: Compressing objects: 100% (49/49), done.
remote: Total 126 (delta 29), reused 25 (delta 11), pack-reused 66 (from 3)
Receiving objects: 100% (126/126), 71.34 KiB | 2.16 MiB/s, done.
Resolving deltas: 100% (57/57), done.
Submodule 'git-clone' (https://github.com/tektoncd-catalog/git-clone.git) registered for path 'git-clone'
Cloning into '/home/rbean/code/experiment/git-clone/git-clone'...
Submodule path 'git-clone': checked out 'af93731eb19a28e13f967b7cee8269d1f75dfe6b'
2025-07-17 11:40:10,891 WARNING The `gomod-vendor` flag is deprecated and will be removed in future versions. Note that it will no longer perform automatic vendoring, so in case vendoring is needed, it needs to be manually added to the source repository.
2025-07-17 11:40:11,807 INFO Fetching the gomod dependencies at subpath git-clone/image/git-init
2025-07-17 11:40:12,024 INFO go.mod reported versions: '1.23.0'[go], '-'[toolchain]
2025-07-17 11:40:12,024 INFO Using Go release: go1.21.0
2025-07-17 11:40:17,205 INFO Vendoring the gomod dependencies
2025-07-17 11:40:25,897 ERROR GitCommandError: Cmd('git') failed due to: exit code(128)\n  cmdline: git add --intent-to-add --force -- /home/rbean/code/experiment/git-clone/tmp6b8eed46.hermeto-source-copy/git-clone/image/git-init\n  stderr: 'fatal: Pathspec '/home/rbean/code/experiment/git-clone/tmp6b8eed46.hermeto-source-copy/git-clone/image/git-init' is in submodule 'git-clone''
╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ /home/rbean/code/hermeto/hermeto/interface/cli.py:132 in cmd_with_error_handling                                                                                                                                                                                             │
│                                                                                                                                                                                                                                                                              │
│   129 │   @functools.wraps(cmd)                                                                                                                                                                                                                                              │
│   130 │   def cmd_with_error_handling(*args: tuple[Any, ...], **kwargs: dict[str, Any]) -> Non                                                                                                                                                                               │
│   131 │   │   try:                                                                                                                                                                                                                                                           │
│ ❱ 132 │   │   │   cmd(*args, **kwargs)                                                                                                                                                                                                                                       │
│   133 │   │   except BaseError as e:                                                                                                                                                                                                                                         │
│   134 │   │   │   _bail_out_with_error(e)                                                                                                                                                                                                                                    │
│   135 │   │   except Exception as e:                                                                                                                                                                                                                                         │
│                                                                                                                                                                                                                                                                              │
│ /home/rbean/code/hermeto/hermeto/interface/cli.py:323 in fetch_deps                                                                                                                                                                                                          │
│                                                                                                                                                                                                                                                                              │
│   320 │   │   log.debug(f"Removing existing deps directory '{deps_dir}'")                                                                                                                                                                                                    │
│   321 │   │   shutil.rmtree(deps_dir, ignore_errors=True)                                                                                                                                                                                                                    │
│   322 │                                                                                                                                                                                                                                                                      │
│ ❱ 323 │   request_output = resolve_packages(request)                                                                                                                                                                                                                         │
│   324 │                                                                                                                                                                                                                                                                      │
│   325 │   request.output_dir.path.mkdir(parents=True, exist_ok=True)                                                                                                                                                                                                         │
│   326 │   request.output_dir.join_within_root(".build-config.json").path.write_text(                                                                                                                                                                                         │
│                                                                                                                                                                                                                                                                              │
│ /home/rbean/code/hermeto/hermeto/core/resolver.py:48 in resolve_packages                                                                                                                                                                                                     │
│                                                                                                                                                                                                                                                                              │
│   45 │   │   source_backup = copy_directory(original_source_dir.path, Path(temp_dir).resolve(                                                                                                                                                                                │
│   46 │   │                                                                                                                                                                                                                                                                   │
│   47 │   │   request.source_dir = RootedPath(source_backup)                                                                                                                                                                                                                  │
│ ❱ 48 │   │   output = _resolve_packages(request)                                                                                                                                                                                                                             │
│   49 │   │   request.source_dir = original_source_dir                                                                                                                                                                                                                        │
│   50 │   │                                                                                                                                                                                                                                                                   │
│   51 │   │   # Update all project file paths that end up directly in the source repository                                                                                                                                                                                   │
│                                                                                                                                                                                                                                                                              │
│ /home/rbean/code/hermeto/hermeto/core/resolver.py:78 in _resolve_packages                                                                                                                                                                                                    │
│                                                                                                                                                                                                                                                                              │
│   75 │   │   │   solution="But the good news is that we're already working on it!",                                                                                                                                                                                          │
│   76 │   │   )                                                                                                                                                                                                                                                               │
│   77 │   pkg_managers = [_supported_package_managers[type_] for type_ in sorted(requested_typ                                                                                                                                                                                │
│ ❱ 78 │   return sum([pkg_manager(request) for pkg_manager in pkg_managers], RequestOutput.emp                                                                                                                                                                                │
│   79                                                                                                                                                                                                                                                                         │
│   80                                                                                                                                                                                                                                                                         │
│   81 def inject_files_post(from_output_dir: Path, for_output_dir: Path, **kwargs: Any) -> Non                                                                                                                                                                                │
│                                                                                                                                                                                                                                                                              │
│ /home/rbean/code/hermeto/hermeto/core/package_managers/gomod.py:670 in fetch_gomod_source                                                                                                                                                                                    │
│                                                                                                                                                                                                                                                                              │
│    667 │   │   │   go_work = GoWork(main_module_dir)                                                                                                                                                                                                                         │
│    668 │   │   │                                                                                                                                                                                                                                                             │
│    669 │   │   │   try:                                                                                                                                                                                                                                                      │
│ ❱  670 │   │   │   │   resolve_result = _resolve_gomod(                                                                                                                                                                                                                      │
│    671 │   │   │   │   │   main_module_dir, request, Path(tmp_dir), version_resolver, go_work                                                                                                                                                                                │
│    672 │   │   │   │   )                                                                                                                                                                                                                                                     │
│    673 │   │   │   except PackageManagerError:                                                                                                                                                                                                                               │
│                                                                                                                                                                                                                                                                              │
│ /home/rbean/code/hermeto/hermeto/core/package_managers/gomod.py:1021 in _resolve_gomod                                                                                                                                                                                       │
│                                                                                                                                                                                                                                                                              │
│   1018 │                                                                                                                                                                                                                                                                     │
│   1019 │   # Vendor dependencies if the gomod-vendor flag is set                                                                                                                                                                                                             │
│   1020 │   if should_vendor:                                                                                                                                                                                                                                                 │
│ ❱ 1021 │   │   downloaded_modules = _vendor_deps(go, app_dir, bool(go_work), request.mode, run_                                                                                                                                                                              │
│   1022 │   else:                                                                                                                                                                                                                                                             │
│   1023 │   │   log.info("Downloading the gomod dependencies")                                                                                                                                                                                                                │
│   1024 │   │   downloaded_modules = (                                                                                                                                                                                                                                        │
│                                                                                                                                                                                                                                                                              │
│ /home/rbean/code/hermeto/hermeto/core/package_managers/gomod.py:1616 in _vendor_deps                                                                                                                                                                                         │
│                                                                                                                                                                                                                                                                              │
│   1613 │                                                                                                                                                                                                                                                                     │
│   1614 │   cmdscope = "work" if has_workspace else "mod"                                                                                                                                                                                                                     │
│   1615 │   go([cmdscope, "vendor"], run_params)                                                                                                                                                                                                                              │
│ ❱ 1616 │   if _vendor_changed(context_dir, enforcing_mode):                                                                                                                                                                                                                  │
│   1617 │   │   if enforcing_mode == Mode.STRICT:                                                                                                                                                                                                                             │
│   1618 │   │   │   raise PackageRejected(                                                                                                                                                                                                                                    │
│   1619 │   │   │   │   reason=(                                                                                                                                                                                                                                              │
│                                                                                                                                                                                                                                                                              │
│ /home/rbean/code/hermeto/hermeto/core/package_managers/gomod.py:1643 in _vendor_changed                                                                                                                                                                                      │
│                                                                                                                                                                                                                                                                              │
│   1640 │                                                                                                                                                                                                                                                                     │
│   1641 │   repo = git.Repo(repo_root)                                                                                                                                                                                                                                        │
│   1642 │   # Add untracked files but do not stage them                                                                                                                                                                                                                       │
│ ❱ 1643 │   repo.git.add("--intent-to-add", "--force", "--", context_dir)                                                                                                                                                                                                     │
│   1644 │                                                                                                                                                                                                                                                                     │
│   1645 │   try:                                                                                                                                                                                                                                                              │
│   1646 │   │   # Diffing modules.txt should catch most issues and produce relatively useful out                                                                                                                                                                              │
│                                                                                                                                                                                                                                                                              │
│ /home/rbean/.local/lib/python3.13/site-packages/git/cmd.py:986 in <lambda>                                                                                                                                                                                                   │
│                                                                                                                                                                                                                                                                              │
│    983 │   │   """                                                                                                                                                                                                                                                           │
│    984 │   │   if name.startswith("_"):                                                                                                                                                                                                                                      │
│    985 │   │   │   return super().__getattribute__(name)                                                                                                                                                                                                                     │
│ ❱  986 │   │   return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)                                                                                                                                                                                      │
│    987 │                                                                                                                                                                                                                                                                     │
│    988 │   def set_persistent_git_options(self, **kwargs: Any) -> None:                                                                                                                                                                                                      │
│    989 │   │   """Specify command line options to the git executable for subsequent                                                                                                                                                                                          │
│                                                                                                                                                                                                                                                                              │
│ /home/rbean/.local/lib/python3.13/site-packages/git/cmd.py:1598 in _call_process                                                                                                                                                                                             │
│                                                                                                                                                                                                                                                                              │
│   1595 │   │   call.append(dashify(method))                                                                                                                                                                                                                                  │
│   1596 │   │   call.extend(args_list)                                                                                                                                                                                                                                        │
│   1597 │   │                                                                                                                                                                                                                                                                 │
│ ❱ 1598 │   │   return self.execute(call, **exec_kwargs)                                                                                                                                                                                                                      │
│   1599 │                                                                                                                                                                                                                                                                     │
│   1600 │   def _parse_object_header(self, header_line: str) -> Tuple[str, str, int]:                                                                                                                                                                                         │
│   1601 │   │   """                                                                                                                                                                                                                                                           │
│                                                                                                                                                                                                                                                                              │
│ /home/rbean/.local/lib/python3.13/site-packages/git/cmd.py:1388 in execute                                                                                                                                                                                                   │
│                                                                                                                                                                                                                                                                              │
│   1385 │   │   # END handle debug printing                                                                                                                                                                                                                                   │
│   1386 │   │                                                                                                                                                                                                                                                                 │
│   1387 │   │   if with_exceptions and status != 0:                                                                                                                                                                                                                           │
│ ❱ 1388 │   │   │   raise GitCommandError(redacted_command, status, stderr_value, stdout_value)                                                                                                                                                                               │
│   1389 │   │                                                                                                                                                                                                                                                                 │
│   1390 │   │   if isinstance(stdout_value, bytes) and stdout_as_string:  # Could also be output                                                                                                                                                                              │
│   1391 │   │   │   stdout_value = safe_decode(stdout_value)                                                                                                                                                                                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

taylormadore:
Enables Hermeto to detect vendor directory changes when Go modules are located in git submodules. Previously, the vendor change detection only worked for Go modules in the main repository.

Resolves #1029


tests/unit/conftest.py
```diff
@@ -9,6 +10,29 @@
 from hermeto.core.rooted_path import RootedPath
 
 
+def _create_git_repo(path: Path, files: Optional[dict[str, str]] = None) -> git.Repo:
+    """Create a git repository with initial files.
+
+    :param path: Directory to create the repository in
+    :param files: Dictionary of file paths to content (empty string creates empty file)
```
a-ovchinnikov: s/of/mapping?

tests/unit/conftest.py
```diff
@@ -9,6 +10,29 @@
 from hermeto.core.rooted_path import RootedPath
 
 
+def _create_git_repo(path: Path, files: Optional[dict[str, str]] = None) -> git.Repo:
```
a-ovchinnikov: WDYT about
```python
FileContents = str
...
... files: Optional[dict[Path, FileContents]] = None) ...
```
?

tests/unit/conftest.py
```diff
@@ -9,6 +10,29 @@
...
+    :param path: Directory to create the repository in
+    :param files: Dictionary of file paths to content (empty string creates empty file)
+    :return: Initialized git repository
+    """
+    path.mkdir(exist_ok=True)
+    repo = git.Repo.init(path)
+    repo.git.config("user.name", "user")
+    repo.git.config("user.email", "user@example.com")
+
+    if files:
```
a-ovchinnikov: PEP8 alert: s/if files/if files is not None/.

tests/unit/package_managers/test_gomod.py
```diff
@@ -59,6 +59,73 @@
 
 GO_CMD_PATH = "/usr/bin/go"
 
+# Shared test data for vendor_changed tests
```
a-ovchinnikov: Would it make sense to have
```python
shared_parameter = "vendor_before, vendor_changes, expected_change"
```
defined about here? It will tell a reader what the parameters below are supposed to mean.

hermeto/core/scm.py
```diff
@@ -70,6 +71,71 @@ def get_repo_id(repo: Union[str, PathLike[str], Repo]) -> RepoID:
...
+
+    Handles nested submodules by iteratively finding the deepest submodule
+    containing the target path.
+
+    :param repo_root: Root of the main repository
+    :param target_path: Path to operate on
+    :return: Tuple of (repo, relative_path)
+    :raises NotAGitRepo: if target is in an uninitialized submodule
+    """
+    # Convert relative path to absolute for consistent processing
```
a-ovchinnikov: Would the two lines below benefit from becoming a helper function? If yes then I suggest moving them into a helper and removing a comment explaining what the code does, if no then I suggest removing a redundant comment. The same applies to lines 126 and 135.

tests/unit/conftest.py
```diff
@@ -78,3 +79,56 @@ def input_request(tmp_path: Path, request: pytest.FixtureRequest) -> Request:
...
+def repo_with_uninitialized_submodule(
+    repo_with_nested_submodules: git.Repo,
+) -> git.Repo:
+    """Create a test setup with an uninitialized submodule.
+
+    Simulates the state after a fresh clone without --recursive.
+    """
+    main_repo = repo_with_nested_submodules
+    # Get the 'submodule' submodule by name
+    submodule = next(s for s in main_repo.submodules if s.name == "submodule")
```
a-ovchinnikov: There is a `fist_for` utility function that takes a predicate, an iterable and a fallback value which does exactly this under the hood:
```python
submodule = first_for(lambda s: s.name == "submodule", main_repo.submodules, None)
```

tests/unit/test_scm.py
```diff
@@ -117,3 +118,84 @@ def test_clone_as_tarball_wrong_ref(golang_repo_path: Path, tmp_path: Path) -> N
...
+        expected_repo_path: str,
+        expected_relative_path: str,
+        repo_with_nested_submodules: git.Repo,
+    ) -> None:
+        """Test path resolution across nested submodules."""
+        main_repo_root = Path(repo_with_nested_submodules.working_dir)
+
+        # Create the target path (directories or files)
+        full_target_path = main_repo_root / path_from_main_root
+        if "." not in full_target_path.name:  # No extension means it's a directory
```
a-ovchinnikov: I am sorry, I feel very dense, but why are paths with dots in them neglected completely? I can see several in parametrization list.

a-ovchinnikov (commented):
A few observations and small questions, looks good to me otherwise.

tests/unit/conftest.py
eskultety: > `FileContents = str`

Do we need a custom type alias? The function doc string is fairly self explanatory

> `dict[Path,`

This would mean you'd have to always pass a path object from the caller, I guess for any purposes `Union[os.PathLike[str], Path]` might be a more robust type guard, woudln't it?

hermeto/core/scm.py
```diff
@@ -70,6 +71,71 @@ def get_repo_id(repo: Union[str, PathLike[str], Repo]) -> RepoID:
...
+def _get_submodule_repo(submodule: git.Submodule) -> Repo:
+    """Get the repository for a submodule with initialization validation.
+
+    :param submodule: Git submodule to access
+    :return: Git repository for the submodule
+    :raises NotAGitRepo: if submodule is not initialized
+    """
+    try:
+        return submodule.module()
+    except InvalidGitRepositoryError:
```
eskultety, line 97: What does the error string for this exception say? I'm asking whether we have to define our own nice user facing error.

hermeto/core/scm.py
```diff
@@ -70,6 +71,71 @@ def get_repo_id(repo: Union[str, PathLike[str], Repo]) -> RepoID:
...
+    """Find the submodule containing the target path, if any.
+
+    :param repo: Git repository to search in
+    :param repo_root: Root path of the repository
+    :param target_path: Path to find containing submodule for
+    :return: Tuple of (submodule, submodule_path) or None if no submodule contains the path
+    """
+    for submodule in repo.submodules:
+        submodule_path = repo_root / submodule.path
+        if target_path.is_relative_to(submodule_path):
```
eskultety, line 83: `_find_submodule_containing_path` is only called from `get_repo_for_path`. On L121 you made sure target path is absolute. It's also a Path instance, can we just do `target_path == submodule_path` here? In any case `target_path.is_relative_to(submodule_path)` feels incorrect because target_path is the **containing** path, so if my equality test can't be used for some reason, then `submodule_path.is_relative_to(target_path)` should be used.

tests/unit/test_scm.py
eskultety: What do you mean by neglected? Dot means it's a file and we're running mkdir on paths. That said, this is a bad file test - `full_target_path.is_dir()` is the correct check.

hermeto/core/package_managers/gomod.py
```diff
@@ -1635,12 +1635,16 @@ def _vendor_changed(context_dir: RootedPath, enforcing_mode: Mode) -> bool:
...
     """
     repo_root = context_dir.root
-    vendor = context_dir.path.relative_to(repo_root).joinpath("vendor")
+
+    # Get the correct repo context (main or submodule)
+    repo, context_relative_path = get_repo_for_path(repo_root, context_dir.path)
+
+    # Calculate vendor paths relative to the active repo
+    vendor = context_relative_path / "vendor"
     modules_txt = vendor / "modules.txt"
```
eskultety, line 1644: nitpick: Lines 1643-1644 can be squashed.

tests/unit/conftest.py
a-ovchinnikov: > Do we need a custom type alias? 

Yes. It costs nothing and saves one docstring reading. As for the type guard yes, but I would suggest having an alias for that too. `dict[PathToFile, FileContents]` is succinct and very self-explanatory.

hermeto/core/scm.py
```diff
@@ -70,6 +71,71 @@ def get_repo_id(repo: Union[str, PathLike[str], Repo]) -> RepoID:
...
+    # Iteratively find the deepest submodule containing the path
+    while (
+        result := _find_submodule_containing_path(current_repo, current_repo_root, target_path)
+    ) is not None:
+        submodule, submodule_path = result
+        current_repo = _get_submodule_repo(submodule)
+        current_repo_root = submodule_path
+
+    # Calculate relative path from the final repository root
+    relative_path = target_path.relative_to(current_repo_root)
```
eskultety: Same here `target_path.relative_to(current_repo_root)` is odd because it should be vice versa. Swapping them will then result in several test breakages with:
`ValueError: '/tmp/pytest-of-eskultety/pytest-260/test_path_resolution_directory1/main' is not in the subpath of '/tmp/pytest-of-eskultety/pytest-260/test_path_resolution_directory1/main/submodule/src'` 

which is true, but it's IMO because the tests are relying on a wrong implementation of path handling.

hermeto/core/package_managers/gomod.py
```diff
@@ -1635,12 +1635,16 @@ def _vendor_changed(context_dir: RootedPath, enforcing_mode: Mode) -> bool:
...
 
-    repo = git.Repo(repo_root)
     # Add untracked files but do not stage them
-    repo.git.add("--intent-to-add", "--force", "--", context_dir)
+    repo.git.add("--intent-to-add", "--force", "--", str(context_relative_path))
```
eskultety: Does `git` no longer accept Path objects to work correctly here?
Same comment a few lines below.

tests/unit/conftest.py
eskultety: That would do 2 things though:
- add the overhead of a lambda compared to a simple condition
- warrant mypy complaining about submodule being None on L131 where the check for existence should e dropped since it's nonsense because we just created the hierarchy so we know it exists (well, probably there should be a Runtime exception in `repo_with_nested_submodules` signaling that we failed to prep the test environment)

is it necessary?

tests/unit/package_managers/test_gomod.py
```diff
@@ -1651,6 +1651,47 @@ def test_vendor_changed(
...
+    vendor_changes: dict[str, Any],
+    expected_change: Optional[str],
+    repo_with_nested_submodules: git.Repo,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    """Test _vendor_changed when vendor directory is inside a git submodule."""
+    main_repo = repo_with_nested_submodules
+    main_root = Path(main_repo.working_dir)
+
+    # Get the 'submodule' submodule by name
```
eskultety: self-explanatory

tests/unit/package_managers/test_gomod.py
```diff
@@ -1651,6 +1651,47 @@ def test_vendor_changed(
...
+    submodule = next(s for s in main_repo.submodules if s.name == "submodule")
+
+    # Create app_dir inside the submodule instead of main repo
```
eskultety: self-explanatory

tests/unit/package_managers/test_gomod.py
```diff
@@ -1651,6 +1651,47 @@ def test_vendor_changed(
...
+
+    # Post-vendoring changes
+    write_file_tree(vendor_changes, app_dir, exist_ok=True)
+
+    assert _vendor_changed(app_dir, Mode.STRICT) == bool(expected_change)
+    if expected_change:
+        assert expected_change.format(subpath=subpath) in caplog.text
+
+    # The _vendor_changed function should reset the `git add` => added files should not be tracked
+    assert not submodule_repo.git.diff("--diff-filter", "A")
```
eskultety: I'm a bit confused here, what scenario does this check? I'm having a hard time with my imagination.

tests/unit/test_scm.py
```diff
@@ -117,3 +118,84 @@ def test_clone_as_tarball_wrong_ref(golang_repo_path: Path, tmp_path: Path) -> N
...
+
+class TestGetRepoForPath:
+    """Tests for get_repo_for_path and related submodule handling."""
+
+    @pytest.mark.parametrize(
+        "path_from_main_root,expected_repo_path,expected_relative_path",
+        [
+            pytest.param("", ".", ".", id="main_repo_root"),
+            pytest.param("src", ".", "src", id="directory_in_main"),
+            pytest.param("src/utils", ".", "src/utils", id="nested_directory_in_main"),
```
eskultety: Is there a functional difference between testing `src` and `src/utils` from the logic's POV?

tests/unit/test_scm.py
```diff
@@ -117,3 +118,84 @@ def test_clone_as_tarball_wrong_ref(golang_repo_path: Path, tmp_path: Path) -> N
...
+        if expected_repo_path == ".":
+            expected_repo_root = main_repo_root
+        else:
+            expected_repo_root = main_repo_root / expected_repo_path
+
+        # Test with both absolute and relative paths
+        paths_to_test = [
+            full_target_path,  # Absolute path
+            Path(path_from_main_root) if path_from_main_root else Path("."),  # Relative path
+        ]
```
eskultety: redundant comments

tests/unit/test_scm.py
```diff
@@ -117,3 +118,84 @@ def test_clone_as_tarball_wrong_ref(golang_repo_path: Path, tmp_path: Path) -> N
...
+            assert relative_path == Path(expected_relative_path)
+
+    def test_uninitialized_submodule_error(
+        self, repo_with_uninitialized_submodule: git.Repo
+    ) -> None:
+        """Test error handling for uninitialized submodules."""
+        main_repo = repo_with_uninitialized_submodule
+        main_root = Path(main_repo.working_dir)
+
+        # Get the 'submodule' submodule by name
```
eskultety: self-explanatory

tests/unit/conftest.py
```diff
@@ -78,3 +79,56 @@ def input_request(tmp_path: Path, request: pytest.FixtureRequest) -> Request:
...
+    submodule_path = Path(main_repo.working_dir) / submodule.path
+
+    if submodule_path.exists():
+        shutil.rmtree(submodule_path)
+    submodule_path.mkdir()  # Git expects empty directory
```
eskultety: nitpick: I don't think we tend to put commentaries after the statements unless it's a linter hint

tests/unit/conftest.py
```diff
@@ -78,3 +79,56 @@ def input_request(tmp_path: Path, request: pytest.FixtureRequest) -> Request:
         output_dir=tmp_path / "output",
         packages=package_input,
     )
+
+
+@pytest.fixture
+def repo_with_nested_submodules(tmp_path: Path) -> git.Repo:
```
eskultety: So, what would be the line we'd draw after unit tests and prefer a scoped integration test instead (we have a bunch similar to this)? Having all this generated during runtime is plausible, I agree, but I'm wondering if it's worth the hassle and complexity, because I feel like with this AI assistance in tests we're losing control over test maintenance and they'll likely explode (just like `test_resolve_gomod` did even without AI), so wouldn't it make more sense to replicate the setup on a carefully constructed integration test repo?

Looking at the parametrize cases made me wonder about this, yes, it's harder to replicate that matrix with integration tests, but that's the thing I'm genuinely curious about - what do we really need unit tested and where we can say, there's too many test cases for this type of thing, one or two scoped integration tests would achieve the same thing.

tests/unit/conftest.py
eskultety: Well, technically `PathToFile`, `FileContents` and too many aliases (though cheap, yes) will lead to the opposite of

>  saves one docstring reading

because then you won't remember what the type really translates to when coding yourself and so you'll have to read the alias definition anyway.

a-ovchinnikov: > you won't remember what the type really translates to

I have to politely disagree on this one: you don't need to. The whole point is to have a descriptive enough name for that to ensure that no behavior is a surprise. `PathToFile` looks like something that could be PathLike with a high probability, `FileContents` is something that will go into a file, likely some text. This is all about picking sufficiently clear names which will reflect the intent. Random names would have been puzzling, `FooBar` could be pretty much anything and does not provide any hints about the code which follows, but these names are not. While `FileContents` is just an alias to `str` I still think it serves a purpose of telling a reader that an author have spent enough time considering their type system and allowed a wide type deliberately (there are way too many examples when engineers coming from other ecosystems would slap `str` annotation on `email` variable and call it a day despite `email` being much narrower subset of `str` without any chance to ever grow). There is also always the twilight path of Hungarian notation.

eskultety: > I have to politely disagree on this one

Fair enough, I can work with the suggestion despite being in a disagreement, non-blocking for me  :).

tests/unit/conftest.py
slimreaper35: `gomod` integration tests deserve some love (cleanup)

tests/unit/conftest.py
a-ovchinnikov: Overhead like in "takes time to execute" or like in "is harder to read"? If it is the former then it is negligible compared to file system operations which generally should not be taking place in a UT at all. If it is the latter then I'd say it is not fundamentally worse than having `next()` machinery exposed: using `next` to obtain the first element of a generator is _not_ obvious or clear despite being legitimate and convenient way of doing that. This is why `first_for(predicate)` was introduced: it is immediately clear what the function is supposed to do basing on its name and types of its arguments, how it is implemented is irrelevant. Maybe the interface for `first_for` could be amended, back in the day making default value non-optional felt necessary, or another wrapper could be added for it to incorporate some sane default values could be added, but any of this should be better than exposing raw primitive interfaces.

On mypy complaints: it should start complaining right at L129, however it does not complain at all. I am still not convinced that mypy is necessary for tests (or at all for that matter, this is a pretty obvious type mismatch and it just lets it pass through).

I will not block on this one, but I do believe in wrapping low-level elements when possible.

tests/unit/test_scm.py
a-ovchinnikov: What happens when it is a file residing in a subdirectory which does not exist? I would expect all parent directories to be created if missing, and that's what I could not find here.

hermeto/core/scm.py
taylormadore: I'm not sure I follow on this one :thinking: 

We're trying to determine here whether the submodule contains the `target_path`, so we need to check whether `target_path` is inside `submodule_path`.

`submodule_path.is_relative_to(target_path)` seems backwards to me, since we would be checking whether the `submodule_path` could be resolved relative to the target_path:
```
>>> submodule_path = Path("foo/bar")
>>> target_path = Path("foo/bar/vendor")
>>> submodule_path.is_relative_to(target_path)
False  # This should be a match, since the submodule does contain the target_path
```

The equality test would only work if we were always looking for the exact submodule root, but I was trying to make this function more generic for future/other cases where the target_path could be any file or subdirectory.

hermeto/core/scm.py
taylormadore: It's just `git.exc.InvalidGitRepositoryError: No valid repository at <path>`. I thought an error referencing submodules was more helpful :man_shrugging:

tests/unit/package_managers/test_gomod.py
taylormadore: It's part of the original _vendor_changed test logic, so tbh I didn't scrutinize it too much since I didn't change that part: https://github.com/hermetoproject/hermeto/blob/7f2a7b9059dece9ed929ca479d5778accf3f3819/tests/unit/package_managers/test_gomod.py#L1649

Seems to be testing whether this happens: https://github.com/hermetoproject/hermeto/blob/7f2a7b9059dece9ed929ca479d5778accf3f3819/hermeto/core/package_managers/gomod.py#L1668

hermeto/core/package_managers/gomod.py
taylormadore: It can be, but `vendor` is used later on in the function, and I feel like this is more readable than `modules_txt.parent`

I don't care too much though :man_shrugging:

hermeto/core/package_managers/gomod.py
taylormadore: Nope, no change that I am aware of--I was just making the args to these git repo.git calls consistent within the function (it's different a few lines below). I can remove it

tests/unit/conftest.py
taylormadore: I'm fine with this. I wanted comprehensive tests for my own reasons when I was validating the changes to make sure I didn't miss any edge cases. Now that I'm more confident about the fix we can cut them down. What test cases would we want for integration tests? Just one expected pass and one expected fail? Do we want any unit tests for this if we add integration tests?

taylormadore: > `gomod` integration tests deserve some love (cleanup)

What do you mean about this one? Just in-general?

taylormadore: Oh I see it now. I completely forgot we had integration tests for _vendor_changed already

hermeto/core/package_managers/gomod.py
eskultety: Fair enough.

hermeto/core/scm.py
eskultety: Scratch my comment, I don't know what I was thinking, sigh. I must have done some weird local interactive tests that swayed my line of thinking completely.

hermeto/core/scm.py
eskultety: Resolving, because the original comment this depends on was nonsense.

hermeto/core/scm.py
eskultety: So we have https://github.com/hermetoproject/hermeto/issues/876 filed, so from that POV you're thinking correctly. I'm just not sure it's worth going deeper than a generic `GitError` instantiated with the **actual** git error rather than our own string interpretation of the error, that's all.
If we were a library, then maybe, but logically scoped exceptions may very well satisfy our current needs and still provide value.

taylormadore:
Changed in V2:
- Dropped path parameter/return value from _find_submodule_containing_path to use submodule.working_dir instead
- Greatly simplified the unit tests (integration tests commit will follow once agreed in those separate PRs)
- Incorporated various minor feedback/nitpicks

taylormadore:
Rebased on main

tests/unit/conftest.py
taylormadore: Fixed

tests/unit/conftest.py
taylormadore: Fixed

tests/unit/package_managers/test_gomod.py
taylormadore: Dropped this commit

hermeto/core/scm.py
taylormadore: Dropped the comments

tests/unit/conftest.py
taylormadore: Dropped this from the fixture along with the nested submodules

tests/unit/test_scm.py
taylormadore: Dropped from the revised test

tests/unit/conftest.py
taylormadore: Added the type aliases

hermeto/core/package_managers/gomod.py
taylormadore: Removed

tests/unit/package_managers/test_gomod.py
taylormadore: Dropped this--submodule case to be tested via integration test instead

tests/unit/test_scm.py
taylormadore: Removed in the simplified test

tests/unit/test_scm.py
taylormadore: Removed in the simplified test

tests/unit/test_scm.py
taylormadore: Removed in the simplified test

tests/unit/package_managers/test_gomod.py
taylormadore: Test dropped

tests/unit/package_managers/test_gomod.py
taylormadore: Test dropped

taylormadore:
Changed in V3:
- Added integration tests

hermeto/core/scm.py
eskultety: Let's make this an optional comment, since we'll need to do some work in #876 anyway.

eskultety (approved):
no further comments on my side, looking good, ACK.

slimreaper35 (approved):

taylormadore:
> should probably have been moved out of draft

Whoops, I forgot to toggle that with V2 :man_facepalming:

tests/unit/conftest.py
taylormadore: Tests added

slimreaper35:
@taylormadore the version in the `bom.json` here is still *1.4* which does not match expected *1.6* after #1106 🫤

taylormadore:
> @taylormadore the version in the `bom.json` here is still _1.4_ which does not match expected _1.6_ after #1106 🫤

Will fix in https://github.com/hermetoproject/hermeto/pull/1108

PR #1084, opened 2025-08-27
[lint] Adopt Ruff

eskultety:
This effort introduces Ruff as a replacement for our trusty `black`, `flake8`, `isort`, and `bandit` linters/analyzers. By default Ruff would report a few hundred issues (~500), so configuration was added to keep it compatible as much as possible and cause the least amount of churn to our code base.

This effort is a result of some AI comparison experiments using Cursor, Claude CLI, Gemini CLI to see which led to the least amount of disruption to the existing code base. Interestingly enough Cursor which defaults to the Claude Sonet model yielded better results than Claude CLI itself.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,11 +119,10 @@ Note: `--dev-package-managers` is deprecated (no-op) and will be removed; use `x
 
 ### Coding standards
 
-Hermeto's codebase conforms to standards enforced by a collection of formatters, linters and other code checkers:
+Hermeto's codebase relies on Ruff & other code quality tools to conform to common and widely used
+standards:
 
-* [black](https://black.readthedocs.io/en/stable/) (with a line-length of 100) for consistent formatting
-* [isort](https://pycqa.github.io/isort/) to keep imports sorted
-* [flake8](https://flake8.pycqa.org/en/latest/) to (de-)lint the code and ~~politely~~ ask for docstrings
+* [Ruff](https://docs.astral.sh/ruff/rules/) for formatting, import sorting, and linting
 * [mypy](https://mypy.readthedocs.io/en/stable/) for type-checking. Please include type annotations for new code.
 * [pytest](https://docs.pytest.org/en/7.1.x/) to run unit tests and report coverage stats. Please aim for (near) full
   coverage of new code.


hermeto/interface/cli.py
```diff
@@ -155,7 +155,7 @@ def version_callback(value: bool) -> None:
 
 @app.callback()
 @handle_errors
-def main(  # noqa: D103; docstring becomes part of --help message
```
a-ovchinnikov, line 158: I suspect this was just a visual separator. This is a comment after all, it is supposed to be free form. One point from Ruff.

noxfile.py
```diff
@@ -76,6 +76,45 @@ def isort(session: Session) -> None:
...
+        silent=True,
+        external=True,
+    )
+
+
+@nox.session(name="ruff-fix")
+def ruff_fix(session: Session) -> None:
+    """Run ruff with auto-fix for linting and formatting."""
+    install_requirements(session)
+    session.run("ruff", "check", "--fix", "hermeto", "tests", "noxfile.py", *session.posargs)
```
a-ovchinnikov: WDYT about keeping the precedent and splitting a command string instead of having multiline function calls? (`run(*"ruff check --fix hermeto tests noxfile.py".split()` would also work :X)

noxfile.py
a-ovchinnikov, line 1: Comment on the commit in general: amendments to https://github.com/hermetoproject/hermeto/blob/main/.github/workflows/gating.yaml seem to be missing.

eskultety: I didn't bother with that, I was just testing the waters first ;)

noxfile.py
eskultety: Sure. That said, you didn't complain about the existing precedent with:
`session.install("--no-deps", "-r", "requirements-extras.txt")` :-P

noxfile.py
```diff
@@ -22,7 +22,7 @@
 from nox.sessions import Session
 
 # default sessions to run (sorted alphabetically)
-nox.options.sessions = ["lint", "python"]
+nox.options.sessions = ["ruff", "lint", "python"]
```
slimreaper35: > Sessions not found: ruff

I would prefer to have `ruff`, `mypy`, `python` names as default sessions and `ruff-fix` separately for local development.

slimreaper35 (commented):
Missing update to [CONTRIBUTING.md](https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md#coding-standards)

noxfile.py
eskultety: > Sessions not found: ruff

Sigh, some rebase problem.

> I would prefer to have ruff, mypy, python names as default sessions and ruff-fix separately for local development.

But that will bring back the issue of wasting storage for virtual environments which https://github.com/hermetoproject/hermeto/pull/1082 addressed. Honestly, Ruff is so fast that the `lint` session is pretty much just mypy, so even if ruff checks failed (everything should now be grouped and exception properly handled), I don't see much point in keeping mypy separate, at all. `ruff-fix` was added by the AI automatically, but I personally never was a fan, but kept it there for dev convenience if someone wanted to try it, but again, most devs will already have a main `venv` active with Ruff installed, so re-installing it to a fresh new nox venv with another 300M of dependencies just to be able to run Ruff auto fixups doesn't sound like the smartest approach, especially since this isn't even run directly, but is tunneled through nox as a yet another subprocess. Since you mentioned CONTRIBUTING update I could advise adding a bash alias for this.
To summarize, I can see the following options and we need to select based on team consensus:

- keep `lint`, `ruff`, `ruff-fix` as standalone sessions, BUT default to `lint`, `python` **only** and keep the rest solely as convenience sessions for anyone who wishes to waste storage
- only have `lint` and drop the `ruff` session versions to keep the session list less confusing for anyone who does `nox -l`
- okay, we can even drop this PR and not adopt Ruff :D ... but I guess we have very little incentive for not doing that IMO so I don't count that as an option

@a-ovchinnikov, @taylormadore, @ben-alkov, @derasdf, thoughts?

slimreaper35: > only have lint and drop the ruff session versions to keep the session list less confusing for anyone who does nox -l

Okay, works for me. We never had `black-fix` or `mypy-fix`, so I guess we don't need `ruff-fix` either.

eskultety: Anothe alternative that just occurred to me as I wanted to respond could be - not using a venv for the ruff sessions, i.e. since you always have a dev venv anyway with Ruff, it could reuse the top-level one, we'd keep clean venvs for linters and tests though, so that could turn out viable. I'm just mentioning that just in case.

ben-alkov: > only have `lint` and drop the `ruff` session versions

I'm reading between the lines here and assuming that we mean `lint` will simply be running `ruff` with our config, yes?

ben-alkov:
I'm confused and fatigued by the discussion... however - TL;DR I've been using Ruff as my daily for *everything* in VSCode for... years?! :shrug:.

I removed all the linters, black, isort...

I have Ruff configured to autoformat on save (which for me in VSCode means "autosave", "save when switching apps", and even "save when switching editor tabs" (I HATE losing work) - and in reality it formats imports on the fly - and I have yet to push anything (that I can think of) which caused linting to fail.

TL, TL;DR - Please merge this so I can stop thinking about it :rofl:

noxfile.py
eskultety: > I'm reading between the lines here and assuming that we mean lint will simply be running ruff with our config, yes?

Almost - `lint` will also run mypy, otherwise Ruff would have been a standalone session (which I originally proposed as a result of the AI help, but discarded in the latest revision).

eskultety:
FWIW I can revert the `lint` session and split it in 2 - Ruff and mypy if desired, other than that this change should be completely non-intrusive and there's only something to gain rather than lose :) .

eskultety:
> FWIW I can revert the `lint` session and split it in 2 - Ruff and mypy if desired, other than that this change should be completely non-intrusive and there's only something to gain rather than lose :) .

Any preference? Shall I proceed with ^this and waste ~250MB of storage for just mypy env or keep it as it is right now? :)

noxfile.py
slimreaper35: So do we keep `lint` and `ruff-fix` ?

eskultety:
Ignoring the review bot comments, because we either already talked about them (i.e. splitting the string command to a sequence) or it doesn't even work (dropping the D103 noqa rule).

noxfile.py
eskultety: I don't know, I got 1 ACK and no consensus, it's not up to me, I'm fine any way pretty much as long as we adopt it since we found a compatible configuration, I just had to raise concerns on having `mypy` and `lint` as separate venvs.
The moment I get strong and clear preference comments, I'll make the final adjustment (hopefully soon) :) .

a-ovchinnikov: I would have said `keep lint, ruff, ruff-fix as standalone sessions` -- it is not that much of a mental burden after all. `ruff-fix` might prove occasionally helpful, however I won't miss it if others decide that it must go. I think we could just merge and see how it goes, if there are many complaints we can always revert.

eskultety: > `keep lint, ruff,`

You know `lint` will run `ruff`, right? I think you meant `lint` vs `mypy`.

eskultety: > I just had to raise concerns on having mypy and lint as separate venvs.

Just to give another perspective to the argument I'm making, this is what happened to me over the last couple of days as I've been running all CI checks (lint, unit, ITs) locally in a rebase:
```
  359.3 MiB [##############################] /integration-tests
  353.4 MiB [############################# ] /mypy
  337.3 MiB [############################  ] /python-3-13
  337.0 MiB [############################  ] /python-3-12
  334.6 MiB [###########################   ] /isort
  334.6 MiB [###########################   ] /flake8
  334.6 MiB [###########################   ] /black
  334.6 MiB [###########################   ] /bandit
  333.3 MiB [###########################   ] /ruff
  321.2 MiB [##########################    ] /python-3-9
  317.4 MiB [##########################    ] /lint
  298.9 MiB [########################	   ] /__install
  297.9 MiB [########################	   ] /all-integration-tests
   11.7 MiB [                              ] /pip-compile

TOTAL: 4.2 GiB
```

a-ovchinnikov: > You know lint will run ruff, right? I think you meant lint vs mypy.

Sorry, should have not quoted an earlier bit without thinking. I meant that keeping linters and mypy in one session is fine given that it is ruff and there is not much of a delay that way. Having or not having ruff-fix are both fine with me.

ben-alkov (approved):
🌮

noxfile.py
ben-alkov: I think we can add `ruff check --fix .` for actual in-place fixing AFTER we get this merged.

PR #1086, opened 2025-08-29
Tekton: Fix git-clone task reference

slimreaper35:

### DIFF
--- a/.tekton/release.yaml
+++ b/.tekton/release.yaml
@@ -19,7 +19,14 @@ spec:
     tasks:
       - name: fetch-repository
         taskRef:
-          name: git-clone
+          params:
+            - name: name
+              value: git-clone
+            - name: bundle
+              value: quay.io/konflux-ci/tekton-catalog/task-git-clone:0.1@sha256:7939000e2f92fc8b5d2c4ee4ba9000433c5aa7700d2915a1d4763853d5fd1fd4
+            - name: kind
+              value: task
+          resolver: bundles
         workspaces:
           - name: output
             workspace: workspace


.github/workflows/gating.yaml
```diff
@@ -24,8 +24,9 @@ jobs:
       run: |
         # We need to install git inside the container otherwise the checkout action will use Git
         # REST API and the .git directory won't be present which fails due to setuptools-scm
-        apt-get update && apt-get install --no-install-recommends --no-install-suggests -y git
-        pip install --upgrade pip nox
+
+        apt-get update
+        apt-get install --no-install-recommends --no-install-suggests -y git nox
```
a-ovchinnikov: Will it install the same version of nox as pip? Apt was known for keeping rather stale versions of packages in the past.

.github/workflows/gating.yaml
```diff
@@ -57,8 +58,9 @@ jobs:
       run: |
         # We need to install git inside the container otherwise the checkout action will use Git
         # REST API and the .git directory won't be present which fails due to setuptools-scm
-        apt-get update && apt-get install --no-install-recommends --no-install-suggests -y git
-        pip install --upgrade pip nox
+
+        apt-get update
+        apt-get install --no-install-recommends --no-install-suggests -y git nox python3-venv
```
eskultety: Why do you need `python3-venv`? We didn't need it before, IIRC venv is a built in module, so should be shipped with python, isn't so?

.github/workflows/gating.yaml
eskultety: > Will it install the same version of nox as pip? Apt was known for keeping rather stale versions of packages in the past.

It most certainly won't :) . I'm not in favour really because the recommended approach for local development is `pip install -r requirements-extras` which will give you the latest nox which will 99.8% be the exact same version as in our CI, so reproducibility is easier than with mismatching versions in different environments as it would be with `apt`

.github/workflows/gating.yaml
slimreaper35: I encountered this error *only* for the lint session:

```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt install python3.13-venv

You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.
```

.github/workflows/gating.yaml
slimreaper35: Ok, I will change it to pip install.

EDIT: I will drop it, doesn't matter.

a-ovchinnikov (approved):

.github/workflows/gating.yaml
eskultety: Oh, interesting.

eskultety (approved):

PR #1088, opened 2025-09-01
cli: do not error out on long fetch-deps inputs

ptoscano:
As a result of a recently introduced possibility to use a filename/path for the fetch-deps command, the input is first tried as filename, and in case it is not then it is loaded as JSON.

Now, in case the input string is very long, then stat'ing the string results in an OSError exception, which causes the whole application to stack trace. Instead, silently accept that failure and fallback to JSON as previously done.

Fixes commit 01140a3f034dabd5e97fac631a8f8ceaf0bd416d.

### DIFF
--- a/hermeto/interface/cli.py
+++ b/hermeto/interface/cli.py
@@ -210,8 +210,11 @@ def _looks_like_json(value: str) -> bool:
 
 
 def _looks_like_input_file(value: str) -> bool:
-    path = Path(value)
-    return path.exists() and path.is_file()
+    try:
+        path = Path(value)
+        return path.exists() and path.is_file()
+    except OSError:
+        return False
 
 
 class _Input(pydantic.BaseModel, extra="forbid"):
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -357,6 +357,34 @@ class TestFetchDeps:
                     {"type": "gomod", "path": "pkg_b"},
                 ],
             ),
+            pytest.param(
+                dedent(
+                    """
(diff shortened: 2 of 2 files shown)


ptoscano:
cc @slimreaper35

tests/unit/test_cli.py
```diff
@@ -357,6 +357,34 @@ def test_no_packages(self) -> None:
                     {"type": "gomod", "path": "pkg_b"},
                 ],
             ),
+            # very long input string
```
eskultety: nitpick: I know you're just following the established pattern of ours, but we'd like to gradually convert most parametrize cases to use `pytest` IDs, so please use an ID instead of a commentary here.

hermeto/interface/cli.py
```diff
@@ -210,8 +210,11 @@ def _looks_like_json(value: str) -> bool:
```
eskultety, line 210: nitpick:

> Fixes commit https://github.com/hermetoproject/hermeto/commit/01140a3f034dabd5e97fac631a8f8ceaf0bd416d.

No periods after commit digests please, it renders incorrectly in git history (unless viewed in GH UI) :) .
```
Fixes: <digest>
```
is perfectly sufficient.

eskultety (commented):
ACK with 2 minor comments.

eskultety (approved):

slimreaper35 (approved):

ssbarnea:
Any idea when this could be released? Is there any release upcoming?

taylormadore:
@ssbarnea We just released it in https://github.com/hermetoproject/hermeto/releases/tag/0.36.0

PR #1105, opened 2025-09-10
cargo: Change heuristic to use Cargo.lock for Rust root detection

slimreaper35:

### DIFF
--- a/hermeto/core/package_managers/pip/rust.py
+++ b/hermeto/core/package_managers/pip/rust.py
@@ -50,14 +50,19 @@ def _depends_on_rust(extracted_dir: Path) -> bool:
     return False
 
 
-def _get_rust_root_dir(cargo_files: Iterable[Path]) -> Path:
-    # find the top-most Cargo.toml in the package - that's not necessarily the most accurate
-    # solution, but this heuristic has proven to work on the most popular python packages
-    # that have rust dependencies; if this stops working, then we would probably need to check
-    # pyproject toml config section for maturin or setuptools-rust...
-    # More info on this issue in the design doc
-    # https://github.com/hermetoproject/hermeto/blob/e5fa5c0fcd0dff62cf02be5b0d219e04c1ea440c/docs/design/cargo-support.md#L806
-    return min(cargo_files, key=lambda x: len(x.parts)).parent
+def _get_rust_root_dir(cargo_lockfiles: Iterable[Path]) -> Path:
+    """Find the top-most Cargo.lock in the package.
+
+    NOTE: This is not necessarily the most accurate solution, but this heuristic has proven to work
+    on the most popular python packages with rust dependencies. A repository may have multiple
+    Cargo.toml files (e.g. workspace), but we are mostly interested in Cargo.lock file.
+
+    Also, on the same level as Cargo.lock file, there is always a Cargo.toml file.
+
+    If this solution is not accurate enough, then we would probably need to check for the manifest
+    path in pyproject.toml config section for maturin or setuptools-rust.
+    """
+    return min(cargo_lockfiles, key=lambda x: len(x.parts)).parent
 
 
 def filter_packages_with_rust_code(packages: list[dict[str, Any]]) -> list[CargoPackageInput]:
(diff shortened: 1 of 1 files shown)


hermeto/core/package_managers/pip/rust.py
```diff
@@ -80,8 +85,8 @@ def filter_packages_with_rust_code(packages: list[dict[str, Any]]) -> list[Cargo
         extract_dir = pip_deps_dir / filename
         shutil.unpack_archive(package_path, extract_dir=extract_dir, filter=extract_filter)  # type: ignore[arg-type]
 
-        cargo_files = list(extract_dir.rglob("Cargo.toml"))
-        if not cargo_files:
+        cargo_lockfiles = list(extract_dir.rglob("Cargo.lock"))
```
a-ovchinnikov, line 88: I remember seeing some Py+Rust projects which were not locked, cannot find a link to any, but this can happen rarely. Would it make sense to search for both and fall back to ".toml" when there is no ".lock"? That's assuming that we might expand permissive mode to allow unlocked projects (again, I vaguely remember someone asking about that, but never opening an issue).

hermeto/core/package_managers/pip/rust.py
```diff
@@ -50,14 +50,19 @@ def _depends_on_rust(extracted_dir: Path) -> bool:
...
+
+    NOTE: This is not necessarily the most accurate solution, but this heuristic has proven to work
+    on the most popular python packages with rust dependencies. A repository may have multiple
+    Cargo.toml files (e.g. workspace), but we are mostly interested in Cargo.lock file.
+
+    Also, on the same level as Cargo.lock file, there is always a Cargo.toml file.
+
+    If this solution is not accurate enough, then we would probably need to check for the manifest
+    path in pyproject.toml config section for maturin or setuptools-rust.
+    """
```
a-ovchinnikov, line 64: Out of curiosity, why did you change a comment into an docstring? (The update itself is good).

hermeto/core/package_managers/pip/rust.py
slimreaper35: Hmm, if we fall back to `Cargo.toml` that would fail [here](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/cargo/main.py#L113) right?

> That's assuming that we might expand permissive mode to allow unlocked projects (again, I vaguely remember someone asking about that, but never opening an issue).

I remember, I raised this question/comment in one pull request 😄 :
https://github.com/hermetoproject/hermeto/pull/949#pullrequestreview-2877829011

hermeto/core/package_managers/pip/rust.py
slimreaper35: After reading the comment, it felt like it was not connected to the code itself, but rather the reasoning why we have the function and why we try to find a top-level cargo file.

Could be a personal preference, so I don't mind changing it back.

a-ovchinnikov: Well, this comment explains why the code is here. A docstring is something intended for a function user rather than a developer who might want to toch this function, and a comment deals more with implementation details. I won't push for keeping this as a comment, I am fine with this being a docstring as long as it preserves the information.

hermeto/core/package_managers/pip/rust.py
a-ovchinnikov: >  if we fall back to Cargo.toml that would fail here right?

Right. Relaxing this requirement was discussed back in the day and I believe it was also someone else who ran into some misbehaving transitive dependency and asked whether we could consider lifting this strict requirement.

slimreaper35: So, the current behavior should basically save them. No `Cargo.lock` file -> we skip the tarball and continue.

taylormadore (approved):

hermeto/core/package_managers/pip/rust.py
eskultety: Agree on the docstring vs comment reasoning, specifically NOTEs is something that is specific to the code, not the interface that is implementing a particular logic, it's not a deal breaker because we're not a library, but it'd be good to adhere to that distinction in the future :) .

slimreaper35:
I hope we cover 99% of use cases 🙏

PR #1111, opened 2025-09-17
[gomod] Refactor toolchains

eskultety:
This is a refactor of our `Go` class to pick the toolchain version correctly. Since everything in the `gomod.py` module is kind of interconnected at some point I resorted to refactor the `GoWork` class as well.
It should now be more straightforward to make adjustments to either `Go` or `GoWork` classes since they should be more self-contained, see individual patches for details.
Ultimately, this series ditches the `1.20` toolchain we have pre-installed in the image and finally bumps it to the latest release to make us much less dependent on `GOTOOLCHAIN=auto`.

**Depends on: https://github.com/hermetoproject/hermeto/pull/1110**  (is rebased onto)

Closes: https://github.com/hermetoproject/hermeto/issues/787


eskultety:
I did my best to split the changes to the smallest possible and logically enclosed chunks, so hopefully it should not be too much of a PITA to review.

eskultety:
Integration tests pretty much passed (missing one more test data update), but I realized this is not bisectable, i.e. if it turned out something stopped working in our go backend we'd lose the ability to pinpoint the problem with git bisect and we'd only be able to do it manually, so I may need to spend some more time making sure integration tests pass after all commits.

eskultety:
> Integration tests pretty much passed (missing one more test data update), but I realized this is not bisectable, i.e. if it turned out something stopped working in our go backend we'd lose the ability to pinpoint the problem with git bisect and we'd only be able to do it manually, so I may need to spend some more time making sure integration tests pass after all commits.

Luckily, the change to keep the code base bisectable was easy and it also allowed me to split the changes to several PRs to reduce the review fatigue.

FWIW, here's a log of the more exotic use case - we don't have a suitable toolchain version installed and so we have to install a full blown SDK:
```
2025-09-18 14:29:17,129 DEBUG Found installed Go releases:
    - /usr/local/go/go1.20/bin/go
2025-09-18 14:29:17,129 DEBUG Testing for go mod file in /home/eskultety/git/hermeto-integration-tests.git/tmpr4cdbbic.hermeto-source-copy/go.mod
2025-09-18 14:29:18,257 INFO Fetching the gomod dependencies at subpath .
2025-09-18 14:29:18,258 DEBUG Matched Go version 1.22 on go.mod line 2: 'go 1.22
'
2025-09-18 14:29:18,258 DEBUG Matched toolchain 1.22.4 on go.mod line 4: 'toolchain go1.22.4
'
2025-09-18 14:29:18,258 DEBUG go.mod reported versions: '1.22'[go], '1.22.4'[toolchain]
2025-09-18 14:29:18,258 DEBUG Running `/usr/local/go/go1.20/bin/go version`
2025-09-18 14:29:18,264 DEBUG Go release: go version go1.20 linux/amd64

2025-09-18 14:29:18,266 DEBUG Installing Go go1.22.4 toolchain shim from 'golang.org/dl/go1.22.4@latest'
2025-09-18 14:29:18,267 DEBUG Running `/usr/local/go/go1.20/bin/go install golang.org/dl/go1.22.4@latest`
2025-09-18 14:29:28,099 DEBUG Downloading Go go1.22.4 SDK
2025-09-18 14:29:28,100 DEBUG Running `/tmp/hermeto97oyiinwgo-download/bin/go1.22.4 download`
2025-09-18 14:30:00,581 DEBUG Go go1.22.4 toolchain installed at: /home/eskultety/.cache/hermeto/go/go1.22.4
2025-09-18 14:30:00,581 DEBUG Running `/home/eskultety/.cache/hermeto/go/go1.22.4/bin/go version`
2025-09-18 14:30:00,587 DEBUG Go release: go version go1.22.4 linux/amd64

2025-09-18 14:30:00,587 DEBUG Using Go toolchain version '1.22.4'
....
2025-09-18 14:30:02,251 INFO All dependencies fetched successfully \o/
```

hermeto/core/package_managers/gomod.py
```diff
@@ -1221,6 +1222,19 @@ class GoCacheTemporaryDirectory(tempfile.TemporaryDirectory[str]):
...
     """
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize our TemporaryDirectory context manager wrapper.
+
+        Store the Go toolchain version used in this session for the subsequent cleanup.
+        """
+        super().__init__(*args, **kwargs)
+        # store the exact toolchain instance that was used for all actions within the context
+        self._go_instance: Go
```
taylormadore: Should we default this to something?

tests/unit/package_managers/test_gomod.py
```diff
@@ -2133,50 +2139,146 @@
...
+    mock_go: mock.Mock,
+    mock_get_cache_dir: mock.Mock,
+    tmp_path: Path,
+    PATH: Optional[str],
+    file_tree: dict,
+    binary_count: int,
+) -> None:
+    """Test various combinations of PATH, cache, and /usr/local Go installations."""
+    mock_get_cache_dir.return_value = tmp_path
+    mock_go.side_effect = lambda binary: mock_go_class(binary)
```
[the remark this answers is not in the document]
eskultety: Applied locally.

hermeto/core/package_managers/gomod.py
```diff
@@ -618,6 +618,12 @@ def _resolve_package_relative_path(package: ParsedPackage, module: Module) -> st
     return [_create_package(package) for package in parsed_packages]
 
 
+def _clean_go_modcache(go: Go, dir_: Optional[Union[str, os.PathLike[str]]]) -> None:
```
a-ovchinnikov: Optional[suggestion]: `Union[str, os.PathLike[str]]` appears more than once in the code, maybe it is a sign to introduce an alias for it, `PathKind` or `LikeAPath`? (`PathLike` is already claimed by `os`, but that would have been the best name for this type).

hermeto/core/package_managers/gomod.py
a-ovchinnikov: From the looks of it this declares an intent to store an instance, but does not actually store it, or am I missing something?

hermeto/core/package_managers/gomod.py
```diff
@@ -220,6 +220,67 @@ def to_component(self) -> Component:
...
+
+    @classmethod
+    def max(cls) -> "GoVersion":
+        """Instantiate and return a GoVersion object with the maximum supported version of Go."""
+        return cls(cls.MAX_VERSION)
+
+    @cached_property
+    def language(self) -> version.Version:
+        """
+        Language version for the given Go version.
```
a-ovchinnikov, line 279: Naming confusion nitpick: When the class is instantiated we get to work with GoToolchainVersion which we can convert to language version, right? Maybe `to_language_version` would indicate this better?

hermeto/core/package_managers/gomod.py
```diff
@@ -1741,3 +1742,41 @@ def _vendor_changed(context_dir: RootedPath, enforcing_mode: Mode) -> bool:
...
+
+def _list_installed_toolchains() -> set[Go]:
+    """List all Go SDK installations we recognize.
+
+    We look at (in this exact order of precedence):
+        - /usr/local/go/                    container environments (Go pre-installed by us)
+        - $XDG_CACHE_HOME/<APP_NAME>/go     local environments (Go downloaded & cached by us)
+        - $PATH/go                          default system-wide Go installation
+
+    :returns: A list of Go instances corresponding to the installations found
```
a-ovchinnikov: Microscopic nitpick: it returns a set.

hermeto/core/package_managers/gomod.py
```diff
@@ -1741,3 +1742,41 @@ def _vendor_changed(context_dir: RootedPath, enforcing_mode: Mode) -> bool:
...
+            continue
+
+        try:
+            log.debug("Probing %s toolchain...", path)
+            ret.add(Go(binary=bin_path.as_posix()))
+        except Exception as e:
+            # Logging toolchain probing failures due due to [1].
+            # [1] https://bandit.readthedocs.io/en/1.8.3/plugins/b112_try_except_continue.html
+            log.debug("Toolchain %s failed probing: %s, skipping...", path, e)
+            continue
```
a-ovchinnikov: Nitpick: excessive continue, `log.debug()` should be enough.

hermeto/core/package_managers/gomod.py
```diff
@@ -688,6 +688,13 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
     if not subpaths:
         return RequestOutput.empty()
 
+    if not (installed_toolchains := _list_installed_toolchains()):
+        raise FetchError(
+            "Could not find any installed Go toolchains in known locations",
+            solution="Please make sure at least one go toolchain is installed in the system",
+        )
+    go: Optional[Go] = next(iter(installed_toolchains))
```
a-ovchinnikov: Two observations here, first, why `Optional`? _list_installed_toolchains() returns a set of Go objects which are not None by definition. Second, since we already know that `installed_toolchains` is not empty and we don't care about the order why not just `.pop()`? _Edit_: having written that I realized that this is temporary logic. Keeping the comment just in case

hermeto/core/package_managers/gomod.py
```diff
@@ -441,7 +442,7 @@ def _retry(self, cmd: list[str], **kwargs: Any) -> str:
             reraise=True,
         )
         def run_go(_cmd: list[str], **kwargs: Any) -> str:
-            return self._run(_cmd, **kwargs)
+            return Go._run(_cmd, **kwargs)
```
a-ovchinnikov, line 447: Basing on this line doesn't this have to be a classmethod?

hermeto/core/package_managers/gomod.py
```diff
@@ -964,6 +967,83 @@ def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
...
+    :return: a Go instance which matches the go.mod version constraints or None if we could not find
+             a satisfying toolchain version installed
+    """
+    go_max_version = GoVersion.max()
+    go_mod_version_msg = "go.mod reported versions: '%s'[go], '%s'[toolchain]"
+
+    go_version_str, toolchain_version_str = _get_gomod_version(go_mod_file)
+    log.debug(
+        go_mod_version_msg,
+        go_version_str if go_version_str else "-",
```
a-ovchinnikov: WDYT about this:
```diff
- go_version_str if go_version_str else "-",
+ go_version_str or "-",
```

hermeto/core/package_managers/gomod.py
```diff
@@ -964,6 +967,83 @@ def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
...
+                "You may also want to open a feature request on adding support for this version."
+            ),
+        )
+
+    # If we cannot find a matching toolchain, we'll try to fallback to a 1.21 one
+    matching_toolchains = filter(lambda t: t.version >= target_version, installed_toolchains)
+    try:
+        go = min(matching_toolchains)
+        log.debug("Using Go toolchain version '%s'", go.version)
+    except ValueError:
```
a-ovchinnikov, line 979: Non-blocking nitpick: this one is a bit cryptic, maybe a comment like `# no toolchain satisfied the condition` could be added on this same line?

hermeto/core/package_managers/gomod.py
```diff
@@ -964,6 +967,83 @@ def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
...
+        try:
+            go = max(filter(lambda t: t.version >= GoVersion("1.21"), installed_toolchains))
+            log.debug("Best matching Go toolchain version: '%s'", go.version)
+            log.debug("Will use Go toolchain version '%s' [via GOTOOLCHAIN=auto]", target_version)
+        except ValueError:
+            # This is a long shot - we couldn't find a matching toolchain, we don't have a toolchain
+            # that can do GOTOOLCHAIN=auto, so we pick any installed toolchain (we know we have
+            # any) and use it to download a new full-blown SDK for the target version
```
a-ovchinnikov: s/any/some?

hermeto/core/package_managers/gomod.py
```diff
@@ -964,6 +967,83 @@ def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
...
+            log.debug("Installing Go toolchain version '%s'", target_version)
+            release_str = f"go{str(target_version)}"
+            try:
+                work_toolchain = next(iter(installed_toolchains))
+                go = Go.from_missing_toolchain(work_toolchain.binary, release=release_str)
+                log.debug("Using Go toolchain version '%s'", go.version)
+            except StopIteration:
+                # should not happen, we test this in fetch_gomod_source, this is only for unit tests
```
a-ovchinnikov: Why would we want to test this behavior if we know that it is impossible?

hermeto/core/package_managers/gomod.py
```diff
@@ -964,6 +967,83 @@ def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
...
+            log.debug("Best matching Go toolchain version: '%s'", go.version)
+            log.debug("Will use Go toolchain version '%s' [via GOTOOLCHAIN=auto]", target_version)
+        except ValueError:
+            # This is a long shot - we couldn't find a matching toolchain, we don't have a toolchain
+            # that can do GOTOOLCHAIN=auto, so we pick any installed toolchain (we know we have
+            # any) and use it to download a new full-blown SDK for the target version
+            log.debug("Installing Go toolchain version '%s'", target_version)
+            release_str = f"go{str(target_version)}"
+            try:
+                work_toolchain = next(iter(installed_toolchains))
```
a-ovchinnikov, line 992: Would a simple `.pop()` be enough or does the set need to be preserved?

a-ovchinnikov (commented):
This is an impressive feat of structuring a PR! LGTM with several minor observations and minor questions.

hermeto/core/package_managers/gomod.py
```diff
@@ -1741,3 +1742,41 @@ def _vendor_changed(context_dir: RootedPath, enforcing_mode: Mode) -> bool:
         repo.git.reset("--", context_relative_path)
 
     return False
+
+
+def _list_installed_toolchains() -> set[Go]:
+    """List all Go SDK installations we recognize.
+
+    We look at (in this exact order of precedence):
```
taylormadore: I'm a little confused about precedence here given that we are returning something unordered

hermeto/core/package_managers/gomod.py
eskultety: Strictly speaking from unit testing perspective that context is irrelevant. I'll drop that particular test combo, I never liked this exception handling myself anyway.

hermeto/core/package_managers/gomod.py
eskultety: Shame on me for not looking at the `set` docs thoroughly, because that was my first idea, but I thought it's only possible with lists. Also shame on me for not even trying first.

hermeto/core/package_managers/gomod.py
eskultety: I might as well rename the whole class to `GoToolchainVersion` during that process.

hermeto/core/package_managers/gomod.py
eskultety: I didn't want to declare it as optional, because mypy... but then the review bot raised some valid concerns so I already fixed this on my local branch ->resolving.

hermeto/core/package_managers/gomod.py
eskultety: I can add a patch for that, `PathLikeStr` ??

hermeto/core/package_managers/gomod.py
eskultety: Uhm, I always seem to forget that one can do this in python and it won't return a boolean. Still, it's likely I will keep forgetting this feature as it doesn't seem very intuitive to me having worked with other languages, so can we keep this as is for the sake of general readability?

hermeto/core/package_managers/gomod.py
eskultety: > I might as well rename the whole class to GoToolchainVersion during that process.

Actually, looking at the class front and back, renaming it would be more confusing, especially in context of max, so I'll keep the name and rename the helper method as suggested.

hermeto/core/package_managers/gomod.py
eskultety: Correction - we already have `StrPath` redefined in several modules, so I'm just going to define it in a single place and reuse it.

eskultety: https://github.com/hermetoproject/hermeto/pull/1115

hermeto/core/package_managers/gomod.py
eskultety: Will fix anyway.

hermeto/core/package_managers/gomod.py
eskultety: Depends on what angle you're looking at this - the common pattern in Python is to define classmethods when you need alternative constructors for some commonly used scenarios, or common instance configuration if you will. Neither of the methods fit that definition, they're just helpers that have a strong logical connection to the class, so despite the invocation looking like it would deserve being a classmethod I'd like to keep it as is.

hermeto/core/package_managers/gomod.py
eskultety: Huh, there was one thing which I just remembered I did here where I didn't even consider `pop` for lists. I designed the function to be generic enough to take any iterable (I didn't want to tie this to sets particularly, so that if we need to change that aspect, it's always going to be just some iterable for us to pick from), so unless you see a good reason to make that input type change we need to keep this `next` handling.

hermeto/core/package_managers/gomod.py
a-ovchinnikov: Sure, let's keep it as is. (I guess the intuitivity of the feature depends on the family of languages one compares to, looks like the ones which behave this way are a bit more obscure.)

hermeto/core/package_managers/gomod.py
```diff
@@ -881,24 +950,33 @@ def _setup_go_toolchain(go_mod_file: RootedPath) -> Go:
...
-        # 1.21 toolchain will bump the input project's go.mod file to make sure the minimum
-        # required Go version is met across all dependencies. That is a problem, because it'll lead
-        # to fatal build failures forcing everyone to update their build recipes. Note that at some
-        # point they'll have to do that anyway, but until majority of projects in the ecosystem
-        # adopt 1.21, we need a fallback to an older toolchain version.
-        go = Go(release="go1.20")
+    # If we cannot find a matching toolchain, we'll try to fallback to a 1.21 one
+    matching_toolchains = filter(lambda t: t.version >= target_version, installed_toolchains)
+    try:
+        go = min(matching_toolchains)
```
taylormadore, line 977: Why the minimum of the matching ones here?

taylormadore (commented):
Very refactor. Such immutable. Wow. :rocket:

hermeto/core/package_managers/gomod.py
taylormadore: Oh, I see this was already mentioned in another comment

hermeto/core/package_managers/gomod.py
eskultety: True, great observation, when I initially added this piece of code, that was at the beginning of the refactor and I remember I wanted to make clear we'd follow the standard system binary precedence hierarchy, but I agree seeing it this way makes no sense.

hermeto/core/package_managers/gomod.py
eskultety: So the filter returns everything `>= <user preferred toolchain>` and so the premise here is to select the closest one matching, so e.g. if a user requires `1.24.8` and we have `1.24.9, 1.25.0, 1.25.1` then we'd go with `1.24.9` as there should be the least number of surprises.

hermeto/core/package_managers/gomod.py
a-ovchinnikov: Oh, good point, it is typed as an `Iterable`, I managed to miss that. For that case I introduced a [wrapper](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/utils.py#L217) around `next(iter())`. Using it directly would be even more unpleasant though, maybe it is time to wrap a wrapper around that wrapper? Something like `pick_one`? I think using `next(iter())` directly is a bit too low-level and at the same time comes dangerously close to language quirks, somewhat similar to using `or` to pick the first truthy value.

eskultety: > Something like pick_one?

`pick_one` invites for the question "based on what heuristic?" Maybe `pop_from_iterable`? As that clearly means it's `pop` depending on particular backend implementation.

> I think using next(iter()) directly is a bit too low-level and at the same time comes dangerously close to language quirks, somewhat similar to using or to pick the first truthy value.

That said, can we leave it for a follow up?

a-ovchinnikov: > `pick_one` invites for the question "based on what heuristic?" Maybe `pop_from_iterable`?

`pop` is a destructive operation by definition, `next(iter())` is not, you'd want to call it `peek` which is non-destructive and is often understood as a non-destructive companion to `pop`. Maybe `pick_any` or `pick_some` could be more descriptive?

> That said, can we leave it for a follow up?

Yes, absolutely!

eskultety: > pop is a destructive operation by definition, next(iter()) is not

Doh, missed that.

hermeto/core/package_managers/gomod.py
slimreaper35: nitpick: Can we improve the format for this log message? Even at a cost of some duplication. I never liked this construction: `'-'[toolchain]`.

I can imagine splitting into two messages or using something like `No 'toolchain' directive in the go.mod file`

slimreaper35 (commented):
Oh no! Something is broken on macOS 😢 

```
2025-09-29 17:14:19,994 DEBUG Using Go toolchain version '1.25.1'
2025-09-29 17:14:19,995 DEBUG Running `/usr/local/bin/go env GOWORK`
2025-09-29 17:14:19,995 DEBUG Running `/opt/homebrew/bin/go clean -modcache`
2025-09-29 17:14:20,021 ERROR BaseError: '/usr/local/bin/go' executable not found in PATH
Error: BaseError: '/usr/local/bin/go' executable not found in PATH
  Please make sure that the '/usr/local/bin/go' executable is installed in your PATH.
  If you are using hermeto via its container image, this should not happen - please report this bug.
```

hermeto/core/package_managers/gomod.py
eskultety: Addressed.

eskultety:
Since last revision:
- addressed review comments
- addressed the hardcoded path comment from AI bot: https://github.com/hermetoproject/hermeto/pull/1111#pullrequestreview-3240032881
- refactored the code some more to get rid of https://github.com/hermetoproject/hermeto/pull/1111/commits/3a0dcacc77ce4e10083b49e47026f244d6b9f8f2 which deliberately broke local install scenarios; the gomod backend should be fully functional throughout this whole refactor series, **please do try randomly to double check my claim!**
- as a result of ^this, dropped the first 2 commits which were no longer necessary

eskultety:
> Oh no! Something is broken on macOS 😢
> 
> ```
> 2025-09-29 17:14:19,994 DEBUG Using Go toolchain version '1.25.1'
> 2025-09-29 17:14:19,995 DEBUG Running `/usr/local/bin/go env GOWORK`
> 2025-09-29 17:14:19,995 DEBUG Running `/opt/homebrew/bin/go clean -modcache`
> 2025-09-29 17:14:20,021 ERROR BaseError: '/usr/local/bin/go' executable not found in PATH
> Error: BaseError: '/usr/local/bin/go' executable not found in PATH
>   Please make sure that the '/usr/local/bin/go' executable is installed in your PATH.
>   If you are using hermeto via its container image, this should not happen - please report this bug.
> ```

@slimreaper35 this sounds serious, I think it may have been related to the fact I hardcoded the Go path to paths in our container image which don't exist on local system, **try again please** and let me know, I believe I addressed it in this revision because I didn't like that it happened although it should have been temporary only.

slimreaper35:
> try again please and let me know,

It works now.

slimreaper35 (approved):

hermeto/core/package_managers/gomod.py
eskultety: @taylormadore can I resolve this or do you have any further comments on this PR before I merge it? :)

taylormadore: Yeah, we can resolve this

PR #1115, opened 2025-09-22
Move some typing aliases to a new dedicated module

eskultety:
Motivated by a review comment in https://github.com/hermetoproject/hermeto/pull/1111

### DIFF
--- a/hermeto/core/checksum.py
+++ b/hermeto/core/checksum.py
@@ -3,11 +3,11 @@ import hashlib
 import logging
 from collections import defaultdict
 from collections.abc import Iterable
-from os import PathLike
 from pathlib import Path
-from typing import NamedTuple, Optional, Union
+from typing import NamedTuple, Optional
 
 from hermeto.core.errors import PackageRejected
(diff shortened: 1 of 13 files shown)


slimreaper35 (approved):

eskultety:
> > > Based on current Python typing best practices, the recommended approach is to use `TypeAlias` for type aliases to make the intent explicit and avoid ambiguity with regular variable assignments.
> > > Apply this diff to follow Python typing best practices:
> > > ```diff
> > >  import os
> > > -from typing import Union
> > > +from typing import Union, TypeAlias
> > > 
> > >  from semver import Version
> > > 
> > > -StrPath = Union[str, os.PathLike[str]]
> > > -SemverLike = Union[Version, str]
> > > +StrPath: TypeAlias = Union[str, os.PathLike[str]]
> > > +SemverLike: TypeAlias = Union[Version, str]

This is reasonable, however, that's only available since Python 3.10 (https://peps.python.org/pep-0613/), so we need to wait one more month before we bump to 3.10

a-ovchinnikov (approved):

PR #1122, opened 2025-09-24
fix(yarn v4): support project-root patches without workspace locators

### LINKED ISSUE #1085: support patch files w/ yarn v4
The latest RHDH 1.7 builds (based on Backstage 1.39) involve building Backstage using yarn patch files, which are not supported by Hermeto:

```
2025-08-06 18:27:54,540 INFO Processing the request using yarn@4.8.1
2025-08-06 18:31:49,539 ERROR UnsupportedFeature: PatchLocator(package=NpmLocator(scope='yarnpkg', name='plugin-npm', version='3.1.0'), patches=(PosixPath('/.yarn/patches/@yarnpkg-plugin-npm-npm-3.1.0-6533d0f5a1.patch'),), locator=None) is missing an associated workspace locator and {APP_NAME} expects all non-builtin yarn patches to have one
Error: UnsupportedFeature: PatchLocator(package=NpmLocator(scope='yarnpkg', name='plugin-npm', version='3.1.0'), patches=(PosixPath('/.yarn/patches/@yarnpkg-plugin-npm-npm-3.1.0-6533d0f5a1.patch'),), locator=None) is missing an associated workspace locator and {APP_NAME} expects all non-builtin yarn patches to have one
  If you need cachi2 to support this feature, please contact the maintainers.
```

Will patches on dependencies ever be supported? I would think all that Hermeto would need to do is fetch the dep, then apply the patch to achieve the hermetic build environment envisioned by the developers. 

Is this a feature that might be supported one day?

ben-alkov:
Closes #1085

- Adds support for Yarn v4 project-root patches which don't have workspace locators
  - Add two new test functions to verify the new functionality
  - Refactor existing test to reflect new behavior
  - Implement the core fix in `_get_path_patch_url()`

### Problem

Yarn v4 introduced project-root patches (stored in `.yarn/patches/`) which don't
have workspace locators, causing Hermeto to raise `UnupportedFeature`

### Solution

Modified `_get_path_patch_url()` method in
`hermeto/core/package_managers/yarn/resolver.py` to

- Handle patches with `locator=None` (project-root patches)
- Maintain backward compatibility with workspace-bound patches
- Generate correct PURLs for both patch types

### Files Changed

- `hermeto/core/package_managers/yarn/resolver.py` - Core implementation
- `tests/unit/package_managers/yarn/test_resolver.py` - Test coverage

## Tested

- [x] New test verifies project-root patch support
- [x] New test verifies correct PURL generation
- [x] Existing tests updated to reflect new behavior
- [x] All 177 yarn tests pass across Python 3.9-3.13
- [x] mypy type checking passes
- [x] No regressions in existing functionality

### DIFF
--- a/hermeto/core/package_managers/yarn/resolver.py
+++ b/hermeto/core/package_managers/yarn/resolver.py
@@ -458,16 +458,17 @@ class _ComponentResolver:
     def _get_path_patch_url(self, patch_locator: PatchLocator, patch_path: Path) -> str:
         """Return a PURL-style VCS URL qualifier with subpath for a Patch."""
         if patch_locator.locator is None:
-            raise UnsupportedFeature(
-                f"{patch_locator} is missing an associated workspace locator "
-                "and {APP_NAME} expects all non-builtin yarn patches to have one"
-            )
+            # Project-root patch (v4 format) - patch path is relative to project root
+            normalized = self._project.source_dir.join_within_root(patch_path)
+            subpath_from_root = str(normalized.subpath_from_root)
+        else:
+            # Workspace-bound patch - patch path is relative to workspace
+            workspace_path = patch_locator.locator.relpath
+            normalized = self._project.source_dir.join_within_root(workspace_path, patch_path)
+            subpath_from_root = str(normalized.subpath_from_root)
 
         project_path = self._project.source_dir
-        workspace_path = patch_locator.locator.relpath
-        normalized = self._project.source_dir.join_within_root(workspace_path, patch_path)
         repo_url = get_repo_id(project_path.root).as_vcs_url_qualifier()
-        subpath_from_root = str(normalized.subpath_from_root)
 
         return f"{repo_url}#{subpath_from_root}"
 
--- a/tests/unit/package_managers/yarn/test_resolver.py
+++ b/tests/unit/package_managers/yarn/test_resolver.py
@@ -2,7 +2,7 @@ import json
 import re
 import zipfile
 from pathlib import Path
-from typing import Any, NamedTuple, Optional, Union
+from typing import Any, NamedTuple, Optional
 from unittest import mock
 from urllib.parse import quote
 
@@ -995,32 +995,77 @@ def test_get_pedigree(
     assert resolver._pedigree_mapping == expected_pedigree
 
 
-@pytest.mark.parametrize(
-    "patch",
-    [
-        pytest.param(
-            Path("foo.patch"),
-            id="path_patch_without_workspace",
-        ),
-        pytest.param(
-            "builtin<bogus/patch>",
-            id="builtin_patch_from_unknown_plugin",
-        ),
-    ],
-)
 @mock.patch("hermeto.core.package_managers.yarn.resolver.get_repo_id")
 @mock.patch("hermeto.core.package_managers.yarn.resolver.extract_yarn_version_from_env")
-def test_get_pedigree_with_unsupported_locators(
+def test_get_pedigree_with_path_patch_without_workspace_locator(
     mock_get_yarn_version: mock.Mock,
     mock_get_repo_id: mock.Mock,
-    patch: Union[Path, str],
     rooted_tmp_path: RootedPath,
 ) -> None:
+    """Path patches without workspace locators are OK (Yarn v4)."""
     mock_get_yarn_version.return_value = Version(3, 0, 0)
     mock_get_repo_id.return_value = MOCK_REPO_ID
 
-    patch_locators = [PatchLocator(NpmLocator(None, "foo", "1.0.0"), [patch], None)]
+    patch_locators = [
+        PatchLocator(
+            package=NpmLocator(None, "foo", "1.0.0"),
+            patches=[Path("foo.patch")],
+            locator=None,
+        )
+    ]
+    mock_project = mock.Mock(source_dir=rooted_tmp_path.re_root("source"))
+
+    resolver = _ComponentResolver(
+        {}, patch_locators, mock_project, rooted_tmp_path.re_root("output")
+    )
+    assert resolver is not None
+
+
+@mock.patch("hermeto.core.package_managers.yarn.resolver.get_repo_id")
+@mock.patch("hermeto.core.package_managers.yarn.resolver.extract_yarn_version_from_env")
+def test_get_pedigree_with_unsupported_builtin_patch(
+    mock_get_yarn_version: mock.Mock,
+    mock_get_repo_id: mock.Mock,
+    rooted_tmp_path: RootedPath,
+) -> None:
+    """Builtin patches from unknown plugins are NOT OK (raise `UnsupportedFeature`)."""
+    mock_get_yarn_version.return_value = Version(3, 0, 0)
+    mock_get_repo_id.return_value = MOCK_REPO_ID
+
+    patch_locators = [
+        PatchLocator(
+            package=NpmLocator(None, "foo", "1.0.0"),
+            patches=["builtin<bogus/patch>"],
+            locator=None,
+        )
+    ]
     mock_project = mock.Mock(source_dir=rooted_tmp_path.re_root("source"))
 
     with pytest.raises(UnsupportedFeature):
         _ComponentResolver({}, patch_locators, mock_project, rooted_tmp_path.re_root("output"))
+
+
+@mock.patch("hermeto.core.package_managers.yarn.resolver.get_repo_id")
+def test_get_path_patch_url_no_workspace_locator_AND_correct_PURLs(
+    mock_get_repo_id: mock.Mock,
+    rooted_tmp_path: RootedPath,
+) -> None:
+    """_get_path_patch_url should handle patches with `locator=None` and generates correct PURLs."""
+    mock_get_repo_id.return_value = MOCK_REPO_ID
+
+    patch_locator = PatchLocator(
+        package=NpmLocator(None, "test-package", "1.0.0"),
+        patches=[Path(".yarn/patches/test-package-npm-1.0.0-abc123.patch")],
+        locator=None,
+    )
+
+    resolver = _ComponentResolver(
+        {}, [], mock_project(rooted_tmp_path.re_root("source")), rooted_tmp_path.re_root("output")
+    )
+
+    result_purl = resolver._get_path_patch_url(
+        patch_locator, Path(".yarn/patches/test-package-npm-1.0.0-abc123.patch")
+    )
+
+    expected_purl = "git+https://github.com/org/project.git@fffffff#.yarn/patches/test-package-npm-1.0.0-abc123.patch"
+    assert result_purl == expected_purl


hermeto/core/package_managers/yarn/resolver.py
```diff
@@ -458,16 +458,17 @@ def _cache_path_as_rooted(self, cache_path: str) -> RootedPath:
     def _get_path_patch_url(self, patch_locator: PatchLocator, patch_path: Path) -> str:
         """Return a PURL-style VCS URL qualifier with subpath for a Patch."""
         if patch_locator.locator is None:
-            raise UnsupportedFeature(
-                f"{patch_locator} is missing an associated workspace locator "
-                "and {APP_NAME} expects all non-builtin yarn patches to have one"
-            )
+            # Project-root patch (v4 format) - patch path is relative to project root
```
a-ovchinnikov: What would happen if there were project-root patches in v1 or v3?

tildebyte: Ouch. Tunnel vision on my part. I'll have to sync up w/Taylor tomorrow…

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -995,32 +995,77 @@ def test_get_pedigree(
...
+        )
+    ]
     mock_project = mock.Mock(source_dir=rooted_tmp_path.re_root("source"))
 
     with pytest.raises(UnsupportedFeature):
         _ComponentResolver({}, patch_locators, mock_project, rooted_tmp_path.re_root("output"))
+
+
+@mock.patch("hermeto.core.package_managers.yarn.resolver.get_repo_id")
+def test_get_path_patch_url_no_workspace_locator_AND_correct_PURLs(
```
slimreaper35:
```suggestion
def test_get_path_patch_url_no_workspace_locator_and_correct_purls(
```

taylormadore:
The requirements for this (as I see them) are:

- [Project-relative paths](https://github.com/yarnpkg/berry/blob/8ff18d709a4211f92837ff2f59eaf4972ca579c0/packages/plugin-patch/sources/patchUtils.ts#L130) for patch locators were introduced in v4. We need to [preserve](https://github.com/hermetoproject/hermeto/blob/1aebd1160a05e55bd250e43e815959d1db8ec937/hermeto/core/package_managers/yarn/locators.py#L214) the `~/` prefix when parsing the locator Paths so that we can later detect and resolve the complete path relative to the project root.
- The ComponentResolver get_path_patch_url needs to be updated to handle the new project-relative paths while maintaining backwards compatibility with the existing workspace-relative paths
- Preserving the project-relative prefix is going to intersect with the current optional flag handling in process_patch_path. We should probably refactor this to separate flag processing logic from the path processing logic. We can use [how Yarn is extracting the flags](https://github.com/yarnpkg/berry/blob/8ff18d709a4211f92837ff2f59eaf4972ca579c0/packages/plugin-patch/sources/patchUtils.ts#L140) from the patch path as a guide, but we also need to be careful to preserve the (different) [v3 flag parsing behavior](https://github.com/yarnpkg/berry/blob/b6026842dfec4b012571b5982bb74420c7682a73/packages/plugin-patch/sources/patchUtils.ts#L109)

The [patchUtils.ts](https://github.com/yarnpkg/berry/blob/master/packages/plugin-patch/sources/patchUtils.ts#L140) file from Yarn contains most of the logic we need for the parsing part. If only I could read JS a little better :sweat_smile:

hermeto/core/package_managers/yarn/resolver.py
ben-alkov: Fixed

tests/unit/package_managers/yarn/test_resolver.py
ben-alkov: Fixed

ben-alkov:
> The requirements for this (as I see them) are:

Re-written

hermeto/core/package_managers/yarn/resolver.py
```diff
@@ -457,18 +457,31 @@ def _cache_path_as_rooted(self, cache_path: str) -> RootedPath:
 
     def _get_path_patch_url(self, patch_locator: PatchLocator, patch_path: Path) -> str:
         """Return a PURL-style VCS URL qualifier with subpath for a Patch."""
-        if patch_locator.locator is None:
-            raise UnsupportedFeature(
-                f"{patch_locator} is missing an associated workspace locator "
-                "and {APP_NAME} expects all non-builtin yarn patches to have one"
-            )
-
+        rel_path: str = "~/"
```
a-ovchinnikov: Where do these type annotations come from? They are a bit excessive.

ben-alkov: Me. Too much?

ben-alkov:
>  do we want them to be published separately or maybe they should be polished into a proper document

I wasn't really sure what to do with it. I think it's useful to have as a standalone doc. The existing comments, *even the links*, were barely useful as more than breadcrumbs.

Having a fully-fleshed implementation guide for the entire "patch" feature of Yarn was invaluable in getting this done, and stands as technical documentation which could be *very* useful for any future work in this part of Yarn.

Yeah, I get YAGNI, but docs are cheap, and don't really get in the way (unless you end up with 100s in a disorganized heap...).

hermeto/core/package_managers/yarn/resolver.py
a-ovchinnikov: A bit. `rel_path` here is clearly a string by construction. The same holds for L222 in `locators.py` and arguably holds for L220 which tells that we are receiving an integer three times: once with variable name, second time with a built-in function and third time with an annotation. I think we should avoid `public static void` style for as long as we can because it visually pollutes code. I think annotations within a code block make some sense only when mypy cannot handle them on its own, hints about types should be provided by naming in all other cases.

a-ovchinnikov:
> I think it's useful to have as a standalone doc. The existing comments, _even the links_, were barely useful as more than breadcrumbs.

Yes, I fully agree with the links part. For myself I settled with providing a summary of a linked material and then a link, I think just links do not really work. (I also tried introducing a `References` section in modules just for links to reduce clutter in code area, but that did not fly). I do not argue with the necessity of adding supplementary documentation. My only concern is representation of this information. As it is now it looks like a raw dump of some shorthand notes extensively referencing state of Hermeto (which is supposed to change after this PR is merged). It is definitely better than nothing and better than raw links, I only wonder if this could be (or should be?) refined a bit more?

ben-alkov:
> I only wonder if this could be (or should be?) refined a bit more?

I'll have a think. Right now it's good for explaining the why of the implementation.

Maybe I should do a short summary at the top, and have the rest of the doc as an appendix...

a-ovchinnikov:
> Maybe I should do a short summary at the top

Yes, that's a good idea! A short summary is what is missing.

hermeto/core/package_managers/yarn/resolver.py
```diff
@@ -457,18 +457,31 @@ def _cache_path_as_rooted(self, cache_path: str) -> RootedPath:
...
         project_path = self._project.source_dir
-        workspace_path = patch_locator.locator.relpath
-        normalized = self._project.source_dir.join_within_root(workspace_path, patch_path)
         repo_url = get_repo_id(project_path.root).as_vcs_url_qualifier()
-        subpath_from_root = str(normalized.subpath_from_root)
 
+        if patch_path.is_absolute():
+            abs_path = patch_path.resolve()
```
taylormadore: Can't we just use RootedPath directly to get the normalized path?

tests/unit/package_managers/yarn/test_locators.py
```diff
@@ -537,6 +537,58 @@ def test_parse_unknown_protocol(locator_str: str) -> None:
         parse_locator(locator_str)
 
 
+def test_project_relative_patch_without_locator() -> None:
```
taylormadore: Should these individual tests just be integrated into the main test? https://github.com/hermetoproject/hermeto/blob/9c634ace7746695ce4b970e9e59784d70dfa2c6d/tests/unit/package_managers/yarn/test_locators.py#L510

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -1024,3 +1024,88 @@ def test_get_pedigree_with_unsupported_locators(
 
     with pytest.raises(UnsupportedFeature):
         _ComponentResolver({}, patch_locators, mock_project, rooted_tmp_path.re_root("output"))
+
+
+@mock.patch("hermeto.core.package_managers.yarn.resolver.get_repo_id")
+@mock.patch("hermeto.core.package_managers.yarn.resolver.extract_yarn_version_from_env")
+def test_get_pedigree_with_project_relative_patches(
```
taylormadore: I wonder if we could work these cases into the existing tests, since the setup is the same

hermeto/core/package_managers/yarn/resolver.py
ben-alkov: Ahhh, now I see. I hadn't taken a hard look at `RootedPath` before...

hermeto/core/package_managers/yarn/resolver.py
ben-alkov: Fixed

hermeto/core/package_managers/yarn/resolver.py
ben-alkov: Fixed

tests/unit/package_managers/yarn/test_locators.py
ben-alkov: Fixed

tests/unit/package_managers/yarn/test_resolver.py
ben-alkov: Fixed

ben-alkov:
Pushed some minor fixes

ben-alkov:
Refactored `test_get_pedigree` - the data factories were too much.

agent-notes/patch-locator-comparison.md
```diff
@@ -0,0 +1,266 @@
+# Patch Locator Implementation Comparison
```
taylormadore: I'm not sure if we should commit such notes to the repository. It appears like this information is already included in the PR description and the commit message. It'll also be historical/development notes once this PR is merged to close the gaps. The ultimate source-of-truth and reference implementation is Yarn itself, so we'd be taking on the maintenance burden of not just our code but these docs as well.

hermeto/core/package_managers/yarn/resolver.py
```diff
@@ -457,18 +462,32 @@ def _cache_path_as_rooted(self, cache_path: str) -> RootedPath:
...
     def _get_path_patch_url(self, patch_locator: PatchLocator, patch_path: Path) -> str:
         """Return a PURL-style VCS URL qualifier with subpath for a Patch."""
-        if patch_locator.locator is None:
-            raise UnsupportedFeature(
-                f"{patch_locator} is missing an associated workspace locator "
-                "and {APP_NAME} expects all non-builtin yarn patches to have one"
-            )
+        patch_str = str(patch_path)
+        pp_root = self._project.source_dir.root
+        pp_join = self._project.source_dir.join_within_root
```
taylormadore, line 464: Do we need these shortcuts? IMO having them makes the subsequent code less explicit and harder to follow

hermeto/core/package_managers/yarn/resolver.py
```diff
@@ -457,18 +462,32 @@ def _cache_path_as_rooted(self, cache_path: str) -> RootedPath:
...
+        project_relative_prefix = "~/"
+
+        if patch_path.is_absolute():
+            try:
```
taylormadore: I wonder if we need to transform PathOutsideRoot to UnsupportedFeature here? The exception message for PathOutsideRoot is already fairly explicit about both the problem and the rationale for it being an error

agent-notes/patch-locator-comparison.md
ben-alkov: It's meant as an historical reference to the implementation at this point in time, not something which needs to be maintained going forward, but if you don't think it's worth it, I'll remove it.

hermeto/core/package_managers/yarn/resolver.py
ben-alkov: I find it easier to follow *with* them :rofl:

hermeto/core/package_managers/yarn/resolver.py
ben-alkov: So, change to

```python
   except PathOutsideRoot as exc:
       raise exc
```

?

taylormadore: I'd just drop the whole try/except and not handle the exception raised from RootedPath here at all

ben-alkov: :saluting_face:

agent-notes/patch-locator-comparison.md
taylormadore: I believe this is the first example we have of such documentation, so we should probably use it as the basis for a team discussion about how to handle these going forward.

hermeto/core/package_managers/yarn/resolver.py
taylormadore: A non-blocking nit from my side

tests/unit/package_managers/yarn/test_resolver.py
taylormadore: After looking at this more closely, I think I was wrong to suggest this. The test_get_pedigree test is really testing that patches get mapped to the patched package and flattened. It might make sense to test _get_path_patch_url separately. WDYT?

In my mind we need to test these as separate concerns:
- The different path types in _get_path_patch_url
- Flattening nested patches in _get_pedigree_mapping

agent-notes/patch-locator-comparison.md
ben-alkov: I'll drop it from this PR so we don't block.

ben-alkov: Done

hermeto/core/package_managers/yarn/locators.py
```diff
@@ -207,18 +210,29 @@ def _parse_patch_locator(locator: "_ParsedLocator") -> PatchLocator:
...
-        # https://github.com/yarnpkg/berry/blob/b6026842dfec4b012571b5982bb74420c7682a73/packages/plugin-patch/sources/patchUtils.ts#L92
-        patch = patch.removeprefix("~")
         # `optional!' denotes an optional patch in Yarn v4
         # https://github.com/yarnpkg/berry/blob/93a56643ba3c813a87920dcf75c644eaf3b38e6f/packages/plugin-patch/sources/patchUtils.ts#L147
-        patch = patch.removeprefix("optional!")
-        if re.match(r"^builtin<([^>]+)>$", patch):
-            return patch
+        # Flags are currently unused by Hermeto, but must be stripped for path parsing
+        flag_index = patch.rfind("!")
+        if flag_index != -1:
```
a-ovchinnikov: A nitpick: if `flag_index` == -1 then `[flag_index + 1:]` == `[0:]` which is effectively a noop in `foo = foo[0:]`. You can save four vertical lines!

hermeto/core/package_managers/yarn/resolver.py
```diff
@@ -455,18 +459,28 @@ def _cache_path_as_rooted(self, cache_path: str) -> RootedPath:
...
+            normalized = pp_join(project_relative_path)
+        else:
+            locator = patch_locator.locator
 
-        project_path = self._project.source_dir
-        workspace_path = patch_locator.locator.relpath
-        normalized = self._project.source_dir.join_within_root(workspace_path, patch_path)
-        repo_url = get_repo_id(project_path.root).as_vcs_url_qualifier()
-        subpath_from_root = str(normalized.subpath_from_root)
+            # handle locator = None case
```
a-ovchinnikov: Another nitpick: if locator can actually be `None` it is better (as per PEP8) to be explicit about checking if it `is None`. That would also make the comment unnecessary.

tests/unit/package_managers/yarn/test_resolver.py
a-ovchinnikov, line 1: Can https://github.com/hermetoproject/hermeto/pull/1122/commits/0659dc59db83c2177521f3938a66b9a969787f43 and https://github.com/hermetoproject/hermeto/pull/1122/commits/5924af3a7e468e2a9efac65c23b027df1672d820 be swapped in the name of history clarity?

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -929,6 +931,146 @@ def test_create_components_cache_path_reported_but_missing(rooted_tmp_path: Root
...
+    ],
+)
+@mock.patch("hermeto.core.package_managers.yarn.resolver.get_repo_id")
+@mock.patch("hermeto.core.package_managers.yarn.resolver.extract_yarn_version_from_env")
+def test_get_path_patch_url(
+    mock_get_yarn_version: mock.Mock,
+    mock_get_repo_id: mock.Mock,
+    rooted_tmp_path: RootedPath,
+    patch_path_template: str,
+    workspace_locator: Optional[WorkspaceLocator],
```
a-ovchinnikov, line 964: `Optional` is not imported in this module. I also believe that we are supposed to use `|` since we are on 3.10 now.

hermeto/core/package_managers/yarn/locators.py
taylormadore: I like `_, _, path = patch.rpartition("!")`

ben-alkov: NICE, it's the clearest option, really communicates intent

PR #1140, opened 2025-10-03
pip: Prepare for binary filters support

slimreaper35:

### DIFF
--- a/hermeto/core/package_managers/pip/package_distributions.py
+++ b/hermeto/core/package_managers/pip/package_distributions.py
@@ -1,6 +1,7 @@
 """This module provides functionality to process package distributions from PyPI (sdist and wheel)."""
 
 import logging
+from collections.abc import Iterable
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Literal, Optional, cast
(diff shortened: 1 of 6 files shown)


hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -151,32 +172,17 @@ def process_package_distributions(
...
-        project_page = client.get_project_page(name, timeout)
-        packages: list[pypi_simple.DistributionPackage] = project_page.packages
-    except (requests.RequestException, pypi_simple.NoSuchProjectError) as e:
-        raise FetchError(f"PyPI query failed: {e}")
-
-    def _is_valid(pkg: pypi_simple.DistributionPackage) -> bool:
-        return (
-            pkg.version is not None
-            and canonicalize_version(pkg.version) == normalized_version
-            and pkg.package_type is not None
```
eskultety, line 173: Was this predicate unnecessary even prior this work or has it become unnecessary with this work? I'm asking because I don't see the `package_type is not None` check in the new helper.

ralphbean:
Consolidated into #1141 per maintainer request for easier review.

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -109,17 +110,18 @@ def test_process_existing_wheel_only_package(
         None,
         None,
     )
-    artifacts = process_package_distributions(req, rooted_tmp_path, allow_binary=True)
+    artifacts = process_package_distributions(req, rooted_tmp_path, PipBinaryFilters())
```
eskultety: You sure this is correct? IIUC the drop-in replacement for `allow_binary=True` is https://github.com/hermetoproject/hermeto/blob/21bf50c0d954506000053dffd40c86cd0e47896e/hermeto/core/models/input.py#L215

Both UT's and ITs are passing which means we're probably not testing the new enhancement and more specifically its backwards compatibility thoroughly, unless I missed something.
There are multiple occurrences of this pattern.

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: The type of `package_type` is `str | None`.

> `if package.package_type not in allowed_distros:`

This check should cover the `None` case too, right? IOW, if `package_type` is None -> it's not in `allowed_distros` -> skip.

tests/unit/package_managers/pip/test_package_distributions.py
slimreaper35: Oh, I missed that method.

hermeto/core/package_managers/pip/package_distributions.py
eskultety: Right, didn't notice that, but I'm not sure you want this implicit type of None handling. It's non-blocking for me, but we tend to cover None checks explicitly :).

_Edit: if not for anything else, the extraction is not a 1:1 code movement, so getting as much of resemblence as possible is IMO a good thing._

taylormadore (approved):

PR #1143, opened 2025-10-06
Implement pip platform filtering

### LINKED ISSUE #931: [pip] Allow specifying only a subset of wheels to be downloaded
Whenever a user provides the `allow_binary` input JSON option, all Python dependencies are downloaded as binary packages even though many/most of the dependencies are actually distributed as tarballs as well (NB: naturally this only applies to those that **actually** distribute as wheels). In fact, for dependencies that distribute both wheels and sdists we download both. This directly or indirectly leads to 2 problems:

__1. Wasting bandwidth (and storage)__
If one only needs wheels for *some* of the dependencies, but not all, we fetched an unnecessary number of wheels that, yes, will be used during the build, but may not be exactly what the user intended. Conversely, if a user wanted **only wheels**, we would fetch potentially many sdists as well, even though those would not be used anywhere nor would they be reported in the SBOM (because we'd report the wheel counterpart).
Based on https://github.com/hermetoproject/hermeto/issues/813#issuecomment-2630355056 we can't automatically only download wheels for dependencies that do not distribute as sdists because that would be breaking backwards compatibility and a worse user experience (see the comment for more details).

__2. Tedious provenance__
With many wheels downloaded, users may need to request an equal number of exceptions with their provenance system (or a global one) simply because wheels, as binary distribution, poses a potential risk from security POV (i.e. infecting one's built artifact).

The proposal in this issue is to introduce a supplementary input JSON option `allowed_binaries` which, paired with `allow_binary` would supply a list of dependencies for which the user requests a wheel to be downloaded. The rest of the dependencies in user's `requirements.txt` file would be downloaded as sdists. This addresses problem nr. 2. On its own this can mitigate problem nr. 1 only partially, i.e. **we would not download an sdist for any package specified in the list of allowed binaries**. Addressing it completely is out of scope of this issue.

Here are a few sample scenarios for demonstration of the new way of pip deps fetching
```
# deps: foo, bar, baz
allow_binary=true, allowed_binaries=[foo, bar] --> wheels = [foo, bar], sdists = [baz]

# deps: foo, bar, baz (all distribute as both wheels and sdists)
allow_binary=true --> wheels = [foo, bar, baz], sdists = [foo, bar, baz]

# deps: foo, bar, baz (all distribute as sdist only)
allow_binary=true --> wheels = [], sdists = [foo, bar, baz]

# deps: foo, bar (bar distributes as wheel only)
allow_binary=true --> wheels = [foo, bar], sdists = [foo]

# deps: foo, bar (bar distributes as wheel only)
allow_binary=true, allowed_binaries=[foo] --> wheels = [foo, bar]   # note the difference with the previous case

# deps: foo, bar (bar distributes as wheel only)
allow_binary=true, allowed_binaries=[foo] --> error, cannot fetch bar as an sdist

# deps: foo, bar (bar distributes as sdist only)
allow_binary=true, allowed_binaries=[foo, bar] --> wheels = [foo], sdists = [bar]    # do we rather want an error here?

# deps: foo, bar
allow_binary=true, allowed_binaries=[]  --> error, conflicting request, docs update needed

# deps: foo, bar
allow_binary=false, allowed_binaries=[foo, bar] --> error, conflicting request, docs update needed
```

slimreaper35:
Closes #931 

### DIFF
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -611,8 +611,7 @@ def _resolve_pip(
         output_dir, resolved_build_req_files, binary_filters
     )
 
-    # No need to search for Rust code when a user requested just binaries.
-    if binary_filters is not None or get_config().ignore_pip_dependencies_crates:
+    if get_config().ignore_pip_dependencies_crates:
         packages_containing_rust_code = []
     else:
         packages_containing_rust_code = filter_packages_with_rust_code(requires + build_requires)
--- a/hermeto/core/package_managers/pip/rust.py
+++ b/hermeto/core/package_managers/pip/rust.py
@@ -4,13 +4,14 @@ import logging
 import shutil
 from pathlib import Path
 from textwrap import dedent
-from typing import Any, Iterable
+from typing import Any, Iterable, Optional
 
 from pybuild_deps import parsers
 
 from hermeto.core.models.input import CargoPackageInput, Request
 from hermeto.core.models.output import EnvironmentVariable, ProjectFile, RequestOutput
 from hermeto.core.package_managers.cargo import fetch_cargo_source
+from hermeto.core.package_managers.pip.requirements import WHEEL_FILE_EXTENSION
 
 log = logging.getLogger(__name__)
 
@@ -72,16 +73,16 @@ def filter_packages_with_rust_code(packages: list[dict[str, Any]]) -> list[Cargo
     for p in packages:
         # File name and package name may differ e.g. when there is a hyphen in
         # package name it might be replaced by an underscore in a file name.
-        package_path = p.get("path", "")
-        if not package_path:
+        package_path: Optional[Path] = p.get("path")
+        if package_path is None or package_path.suffix == WHEEL_FILE_EXTENSION:
             continue
 
-        filename = Path(Path(package_path).name)
+        filename = Path(package_path.name)
         extract_filter = "data" if filename.suffix != ".zip" else None
         while filename.suffix in (".zip", ".tar", ".gz", ".tgz"):
             filename = filename.with_suffix("")
 
-        pip_deps_dir = Path(package_path).parent
+        pip_deps_dir = package_path.parent
         extract_dir = pip_deps_dir / filename
         shutil.unpack_archive(package_path, extract_dir=extract_dir, filter=extract_filter)  # type: ignore[arg-type]
 


hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -171,17 +182,37 @@ def process_package_distributions(
...
+            allowed_distros = ["sdist", "wheel"]
+            to_process = list(sdists) + wheels_filter.filter(wheels)
+        # only-binary mode
+        elif name in wheels_filter.packages:
+            allowed_distros = ["wheel"]
+            to_process = wheels_filter.filter(wheels)
+        # reverted only-binary mode to no-binary mode
+        else:
+            allowed_distros = ["sdist"]
+            to_process = list(sdists)
```
ben-alkov: I'm not clear on exactly what this does (the rest of it makes perfect sense to me).

Maybe a short comment block just above `if binary_filters is None:` to explain the full logic (as opposed to inline comments)?

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +234,110 @@ def process_package_distributions(
...
+        - multiple values in a field are combined with OR logic
+        - multiple fields are combined with AND logic
+        - if a field is not provided (None), it is treated as `:all:`
+
+        See https://packaging.pypa.io/en/stable/tags.html
+        """
+        valid_arch = (
+            self.arch is None
+            or tag.platform == "any"
+            or any(arch in tag.platform for arch in self.arch)
```
ben-alkov: `arch in tag.platform`: Would "x86" match "linux_x86_64"?

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +234,110 @@ def process_package_distributions(
...
+        )
+        valid_os = (
+            self.os is None or tag.platform == "any" or any(os in tag.platform for os in self.os)
```
ben-alkov: `os in tag.platform`: Likewise (probably trivial 🤷), would "win" match "darwin_arm64"?

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -244,3 +246,56 @@ def process_package_distributions(
...
+
+    def __init__(self, filters: PipBinaryFilters) -> None:
+        """Initialize the filter."""
+        self.packages = self._parse_filter_spec(filters.packages)
+        self.arch = self._parse_filter_spec(filters.arch)
+        self.os = self._parse_filter_spec(filters.os)
+        self.py_version = self._parse_filter_spec(filters.py_version)
+        self.py_impl = self._parse_filter_spec(filters.py_impl)
+
+    def __contains__(self, item: pypi_simple.DistributionPackage) -> bool:
```
ben-alkov, line 320: There's a gap in my Python-fu where I don't understand how this is being called.

I assume it's implicit in e.g. a comparison somewhere, but I'm just not seeing it.

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -338,3 +339,142 @@ def test_process_package_distributions_noncanonical_version(
     assert artifacts[0].package_type == "sdist"
     assert artifacts[0].version == requested_version
     assert all(w.version == requested_version for w in artifacts[1:])
+
+
+class TestWheelsFilter:
+    """Don't let the sound of your own wheels drive you crazy."""
```
ben-alkov: :rofl:

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: Yes, but I think that's not what we want 😞

EDIT: I hope it is a very rare corner case.

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: Here: `return [wheel for wheel in wheels if wheel in self]` - `in` is the keyword

hermeto/core/package_managers/pip/package_distributions.py
ben-alkov: Probably :grimacing:

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +234,110 @@ def process_package_distributions(
...
+        )
+        valid_os = (
+            self.os is None or tag.platform == "any" or any(os in tag.platform for os in self.os)
+        )
+        valid_py_version = self.py_version is None or any(
+            version in tag.interpreter for version in self.py_version
+        )
+        valid_py_impl = (
+            self.py_impl is None
+            or "py" in tag.interpreter
```
slimreaper35: Should we make the generic `"py"` part of the default values or keep it as it is now? Any opinion?

ben-alkov: I might be getting this wrong, but are we saying that the Python implementation value can be [None, "\*py\*"]?

If so, then that means nearly anything (except "cargo" :laughing:) will pass...

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +234,110 @@ def process_package_distributions(
...
+        """
+        valid_arch = (
+            self.arch is None
+            or tag.platform == "any"
+            or any(arch in tag.platform for arch in self.arch)
+        )
+        valid_os = (
+            self.os is None or tag.platform == "any" or any(os in tag.platform for os in self.os)
+        )
+        valid_py_version = self.py_version is None or any(
```
taylormadore: I had to think about this one for a bit, but I think the wheel version is a match if any of the following conditions is met:
- The py_version filter is None
- The generic `py` interpreter is being used and the major-only wheel version is compatible
- The generic `py` interpreter is being used and the wheel version <= py_version
- wheel version == py_version
- The ABI tag == "abi3" and wheel_version <= py_version

slimreaper35: How would that work in practice? A user can specify multiple Python versions, and we will download everything that is `<=` as the largest one in the set?

taylormadore: I think that would apply for the generic `py` interpreter comparison and the abi3 comparison, but not for the version-locked comparison.

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: clarified the comments a bit more

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +234,110 @@ def process_package_distributions(
...
+        self.arch = self._parse_filter_spec(filters.arch)
+        self.os = self._parse_filter_spec(filters.os)
+        self.py_version = self._parse_filter_spec(filters.py_version)
+        self.py_impl = self._parse_filter_spec(filters.py_impl)
+
+    def __contains__(self, item: pypi_simple.DistributionPackage) -> bool:
+        """Check if the wheel matches the user constraints."""
+        _, _, _, tags = parse_wheel_filename(item.filename)
+        # usually `tags` contain only one item
+        return any(self.matches(tag) for tag in tags)
```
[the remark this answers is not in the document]
slimreaper35: added

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +235,149 @@ def process_package_distributions(
...
+                wheel_py_version <= max_py_version and tag.abi in ("abi3", "none")
+            )
+        else:
+            # specific implementation tags require exact version match
+            valid_py_version = self.py_version is None or any(
+                version in tag.interpreter for version in self.py_version
             )
 
-    processed_dpis.extend(wheels)
+        return valid_arch and valid_os and valid_py_impl and valid_py_version
```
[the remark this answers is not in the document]
slimreaper35: duplicate comment to https://github.com/hermetoproject/hermeto/pull/1143#discussion_r2417090382

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -123,6 +123,15 @@ def _sdist_preference(sdist_pkg: DistributionPackageInfo) -> tuple[int, int]:
     return yanked_pref, filetype_pref
 
 
+def _find_the_best_sdist(sdists: list[DistributionPackageInfo]) -> DistributionPackageInfo:
```
eskultety, line 129: nitpick: you most likely want to start the series with this commit in order to not break the logical flow, i.e. introducing a wheels filter and then making use of it in a subsequent patch.

tests/integration/test_data/pip_custom_index/fetch_deps_sha256sums.json
```diff
@@ -1,4 +1,3 @@
 {
-  "pip/requests-2.32.3-py3-none-any.whl": "sha256:70761cfe03c773ceb22aa2f671b4757976145175cdfca038c02654d061d6dcc6",
-  "pip/requests-2.32.3.tar.gz": "sha256:55365417734eb18255590a9ff9eb97e9e1da868d4ccd6402399eaf68af20a760"
+  "pip/requests-2.32.3-py3-none-any.whl": "sha256:70761cfe03c773ceb22aa2f671b4757976145175cdfca038c02654d061d6dcc6"
```
eskultety, line 2: This should be squashed to the previous commit as otherwise we have broken integration tests in between commits for no good reason, IOW we should make our policy stricter by enforcing that integrations tests must pass in between commits to ensure proper bisectability.

hermeto/core/package_managers/pip/package_distributions.py
eskultety: So, `py` will also match `py2` and so if a project (for whatever reason, most likely by the means of private indices) still builds for python2 then this check could cause boolean short-circuit and mark the wheel as matching even if the user wanted a py3X build and we'll hence fetch it. It's not a deal breaker, but if we only want to fetch what's truly matching, we should adjust `py` based on `self.py_impl`, IOW search for `py3` if `py_impl == [3.something]` and fallback to `py` otherwise. Or we could just leave it be and shrug shoulders if it ever happens.

tests/unit/package_managers/pip/test_package_distributions.py
eskultety: I don't we need that commentary necessarily, can we perhaps remove it? :)

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -244,3 +247,95 @@ def process_package_distributions(
...
+
+    def parse_py_impl_and_py_version(self, interpreter: str) -> tuple[str, str]:
+        """
+        Examples:
+        >>> parse_py_impl_and_py_version("cp310")
+        ("cp", "310")
+        >>> parse_py_impl_and_py_version("py310")
+        ("py", "310")
+        >>> parse_py_impl_and_py_version("py3")
+        ("py", "3")
```
eskultety: You can have a compound tag like `py2.py3` which would fail the regex below, leading to wheel rejection instead of accepting it.

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -244,3 +247,95 @@ def process_package_distributions(
...
+        )
+
+        # check Python version compatibility with special handling for generic vs specific tags
+        max_py_version = self.get_max_py_version()
+        wheel_py_impl, wheel_py_version = self.parse_py_impl_and_py_version(tag.interpreter)
+
+        if wheel_py_impl == "py":
+            # generic "py" tags are compatible if version <= max_version and abi allows it
+            valid_py_version = self.py_version is None or (
+                wheel_py_version <= max_py_version and tag.abi in ("abi3", "none")
```
eskultety: Please don't compare strings as integers, that's a nasty and error prone thing to do, there's no implicit int conversion happening and I also don't think it does what you're intending to:
```
"310" <= "3"   False;    we don't stop a user to use "3" for the python value, what should happen next?
"310" <= ""     False;    our condition should semantically evaluate to True and we should decide on ABI only
```

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -244,3 +247,95 @@ def process_package_distributions(
...
+            )
+        else:
+            # specific implementation tags require exact version match
```
eskultety: I'm wondering if this heavily custom comparison logic (speaking about the whole `match` method) could be improved and simplified by utilizing the https://packaging.pypa.io/en/stable/tags.html#packaging.tags.generic_tags function which yields a generator of all tag combination for a given input matrix. IIUC we could then skip parsing a wheel's tag and basically just match according to all the generated tag strings.

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -183,17 +183,37 @@ def process_package_distributions(
...
-    packages = _get_project_packages_from(index_url, name, version)
+    packages = list(_get_project_packages_from(index_url, name, version))
+    sdists = filter(lambda x: x.package_type == "sdist", packages)
+    wheels = filter(lambda x: x.package_type == "wheel", packages)
 
-    for package in packages:
+    # no-binary mode: only process source distributions
+    if binary_filters is None:
+        allowed_distros = ["sdist"]
+        to_process = list(sdists)
```
eskultety: Do we need this conversion to list if we only ever process sdists once? Does mypy complain? Can we perhaps just declare `to_process` to be an `Iterable`?

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -183,17 +183,37 @@ def process_package_distributions(
...
+            allowed_distros = ["sdist", "wheel"]
+            to_process = list(sdists) + wheels_filter.filter(wheels)
+        # only-binary mode: only process wheels for packages in the filter list
+        elif name in wheels_filter.packages:
+            allowed_distros = ["wheel"]
+            to_process = wheels_filter.filter(wheels)
+        # no-binary mode: revert only-binary mode and only process source distributions
+        else:
+            allowed_distros = ["sdist"]
+            to_process = list(sdists)
```
eskultety: What's the realistic path to get to this `else` code branch once we concluded that `binary_filters` was set to some value?

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -215,43 +235,60 @@ def process_package_distributions(
...
-                f"No distributions found for package {name}=={version}",
-                solution=solution,
-                docs=docs,
-            )
+        result = _process_prefer_binary_mode(filtered_sdists, filtered_wheels, name, version)
+
+    return result
+
+
+def _process_no_binary_mode(
```
eskultety, line 256: `s/no_binary/sdist_only`

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -215,43 +235,60 @@ def process_package_distributions(
...
+def _process_prefer_binary_mode(
+    sdists: list[DistributionPackageInfo],
+    wheels: list[DistributionPackageInfo],
+    name: str,
+    version: str,
+) -> list[DistributionPackageInfo]:
+    if len(wheels) > 0:
+        return wheels
+
+    return _process_no_binary_mode(sdists, name, version)
```
eskultety: nitpick: can be shortened to `return wheels if wheels else _process_no_binary_mode` for no loss of readability IMO.

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -215,43 +235,60 @@ def process_package_distributions(
...
+
+    return result
+
+
+def _process_no_binary_mode(
+    sdists: list[DistributionPackageInfo],
+    name: str,
+    version: str,
+) -> list[DistributionPackageInfo]:
+    if len(sdists) == 0:
```
eskultety: nitpick: `if not sdists` is I think very acceptable in this case ,same for the other `_mode` function below

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: If the `packages` field was set to some value. Let's say "x, y", so for "x" and "y" we download only wheels, and for all other packages we download only one sdist.

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: Yes, it complains.

> Can we perhaps just declare to_process to be an Iterable?

We can, but we will need the conversion for `list(sdists) + wheels_filter.filter(wheels)` or use `itertools.chain` 🙆‍♂️

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: That's just to be compatible with native pip CLI options (and design doc). But not a big deal.

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -171,17 +183,37 @@ def process_package_distributions(
...
+            allowed_distros = ["sdist", "wheel"]
+            to_process = list(sdists) + wheels_filter.filter(wheels)
+        # only-binary mode: only process wheels for packages in the filter list
+        elif name in wheels_filter.packages:
+            allowed_distros = ["wheel"]
+            to_process = wheels_filter.filter(wheels)
+        # no-binary mode: revert only-binary mode and only process source distributions
+        else:
+            allowed_distros = ["sdist"]
+            to_process = list(sdists)
```
ben-alkov:
```suggestion
            to_process = list(sdists)
        # no-binary mode:  for some packages, we've already downloaded *only*
        # wheels; for all other packages we downloaded only one sdist
        else:
            allowed_distros = ["sdist"]
            to_process = list(sdists)
```

ben-alkov: Bleh, crappy diff, but hopefully you see what I mean

slimreaper35: But this code has nothing with downloading. It's just processing allowed distributions.
https://github.com/hermetoproject/hermeto/blob/main/docs/design/binary-platform-filtering.md#binary-fetching-strategies

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: I know this syntax is possible. Is it preferred for any reason, or is it the standard way in Python?

eskultety: I don't think there is a standard, I look at this from the semantics POV - is the emphasis on the number of elements or is the emphasis on whether there's **anything**/**nothing** in the list? I marked this as nitpick anyway.

hermeto/core/package_managers/pip/package_distributions.py
eskultety: Design doc is a guide, and naming it in a particular way internally is an implementation detail that doesn't need to have any connection to pip CLI options, this is strictly from a reader's POV. You already have a several comments in the code denoting which code branch corresponds to which handling :) . Not a blocker, that was just an observation from a reviewer.

hermeto/core/package_managers/pip/package_distributions.py
eskultety: The question to ask here is whether we truly need to store a list for anything - we don't, we only need to iterate the values once, so from implementation POV `itertools.chain` sounds like a good fit.

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: AI says `if not x` is more idiomatic.

hermeto/core/package_managers/pip/package_distributions.py
eskultety: Then please reword the commentary accordingly to make it clearer please.

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +235,149 @@ def process_package_distributions(
...
+            self.arch is None
+            or tag.platform == "any"
+            or any(arch in tag.platform for arch in self.arch)
+        )
+        # check operating system compatibility
+        valid_os = (
+            self.os is None or tag.platform == "any" or any(os in tag.platform for os in self.os)
+        )
+        # check Python implementation compatibility
+        valid_py_impl = (
```
eskultety: I'd not call this `valid_`. We're not checking validity, but compatibility, so `is_py_impl_compatible`, `is_platform_compatible` might be more suitable names.

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +235,149 @@ def process_package_distributions(
...
+        )
+
+        # check Python version compatibility with special handling for generic vs specific tags
+        max_py_version = self.get_max_py_version()
+        wheel_py_impl, wheel_py_version = self.parse_py_impl_and_py_version(tag.interpreter)
+
+        if wheel_py_impl == "py":
+            # generic "py" tags are compatible if version <= max_version and abi allows it
+            valid_py_version = self.py_version is None or (
+                wheel_py_version <= max_py_version and tag.abi in ("abi3", "none")
```
eskultety: I don't think this is a correct assessment - `abi3` is shorthand for `cp3\d` and so if python tag is `py*`, abi is irrelevant as there's no compiled code, it's automatically matching, abi handling needs to go the `else` branch and iff the implementation user requested is `cp*`, otherwise only `none` can be used as the ever-matching default value along with `cp*`. This claim (or actually the comment that comes next) is further supported by `packaging.tags.compatible_tags` signature which doesn't take `abi` into consideration unlike its other variants which do.

hermeto/core/package_managers/pip/package_distributions.py
eskultety: Answer - that's not a viable alternative, because it cannot generate ALL possible combinations, tags are generated based on a given platform where a package was built, so not all combinations are knowng beforehand.

hermeto/core/package_managers/pip/package_distributions.py
eskultety: Based on https://discuss.python.org/t/is-there-a-canonical-list-of-all-possible-pep-425-tags-or-patterns/5161/10 and some inspection and experiments I did with `packaging.tags` utilities this is going to be almost impossible to get this spot on correct. Citing from the forum above:

> Python packaging does not predefine a list of supported platforms, but platforms declare themselves to support Python packaging. So packaging tools query the platform to get what tag to use (usually via packaging.tags as you already know), not the other way around.

AND

>> it should be be possible to establish an algorithm to convert a tag back to the information which generated it.

> Not necessarily because a lot of it also comes straight from the OS/build itself, e.g. variables set when generating the Makefile. Like if you look at the generic tag support that’s used by PyPy you will see a lot of it is simply asking the interpreter to tell us something that was set at compile time which can ultimately be anything.

What IMO needs to happen in the `else` branch is:

- is the wheel's python tag (e.g. cp310) exactly matching user input ---> match for all ABIs
- is the python tag version < user input --> match iff wheel's abi is `none` for non-cp implementation and `none or abi3` for cp implementation, otherwise such wheel can't be a match because we don't know the correct ABI.

hermeto/core/package_managers/pip/package_distributions.py
a-ovchinnikov: IIRC ability to use iterables in such expressions is the reason behind adding automatic boolean values to them. Personally I believe that for crystal clear semantics one would rather had introduced `is_empty` predicate or at least a method, also `not` conditions are slightly harder to process (for a human) than positive ones, but `if not x` is indeed the preferred way of expressing this.

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: So what would be the proposed change for the binary field based on https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/ ?

- python
- abi
- split platform to os and architecture

eskultety: No, IMHO if we can all agree on this alternative approach to map the input fields to those compatibility tags, then `os` and `architecture` would have to become `platform` otherwise it's confusing. And that is even at the cost of our binary filter fields not being consistent across the backend range which can support binary filtering, because that's what's currently kind of happening, we're trying to keep the common input fields consistent and add some more for Python where `os` and `architecture` isn't enough.

hermeto/core/package_managers/pip/package_distributions.py
ben-alkov: I don't understand what `# no-binary mode: revert only-binary mode and only process source distributions` means...

eskultety: Answered in this thread: https://github.com/hermetoproject/hermeto/pull/1143#discussion_r2429217635

hermeto/core/package_managers/pip/package_distributions.py
eskultety: @slimreaper35 so, in an ideal world we'd have to get involved in https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#compressed-tag-sets parsing each tag as `.` separated values, but that would grow the complexity easily. Let's keep this simple, always assume a single tag component value and for pure python impls let's only consider `py3` instead of `py` when you do the match. I tried to find a python 2 only package on pypi, but didn't find any, so I think `py3` is IMO a safe bet. We can always document this as a limitation that generic Python 2 only packages wheel matching may or may not work reliably.

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: updated

slimreaper35:
New revision with updated binary options. No change to unit tests so far.

hermeto/core/models/input.py
```diff
@@ -209,8 +209,37 @@ class PipBinaryFilters(BinaryModeOptions):
 
     arch: BinaryFilterStr = "x86_64"
     os: BinaryFilterStr = "linux"
-    py_version: BinaryFilterStr = BINARY_FILTER_ALL
+    py_version: Optional[str] = None
```
slimreaper35: We should probably make it a float or an int.

hermeto/core/models/input.py
```diff
@@ -209,8 +209,37 @@ class PipBinaryFilters(BinaryModeOptions):
...
+        except ValueError:
+            raise ValueError(f"Invalid Python version: {py_version}")
+
+        return py_version
+
+    @pydantic.model_validator(mode="after")
+    def _validate_platform_exclusivity(self) -> Self:
+        has_platform = self.platform is not None
+        has_custom_os = self.os != "linux"
+        has_custom_arch = self.arch != "x86_64"
```
slimreaper35, line 222: What should be the default values for `os` and `arch` after introducing `platform`?

hermeto/core/models/input.py
a-ovchinnikov: Please don't convert `version` into float (unless there is a pressing need to support an exceptionally wide dynamic range of versions).

slimreaper35: I mean to enforce using `3.9` for example, instead of `"3.9"`.

a-ovchinnikov: Floats are suitable for numerical simulations of certain kind, in other places they could cause hilarious bugs. There is no need in floats when integers, fixed points, symbols or tuples suffice.

eskultety: > There is no need in floats when integers, fixed points, symbols or tuples suffice.

^This, if you look at other higher version primitives, it's all pretty much just tuples underneath

hermeto/core/models/input.py
eskultety: I think the original defaults for those are to remain unchanged.

hermeto/core/models/input.py
```diff
@@ -209,8 +209,24 @@ class PipBinaryFilters(BinaryModeOptions):
 
     arch: BinaryFilterStr = "x86_64"
     os: BinaryFilterStr = "linux"
-    py_version: BinaryFilterStr = BINARY_FILTER_ALL
+    py_version: Optional[int] = None
     py_impl: BinaryFilterStr = "cp"
+    abi: BinaryFilterStr = BINARY_FILTER_ALL
+    platform: Optional[str] = None
```
taylormadore, line 216: What do you think about adding a field validator to ensure the platform regex compiles?

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -249,3 +252,96 @@ def process_package_distributions(
...
+    def _compatible_tag_interpreter(self, interpreter: str) -> bool:
+        compatible_py_impl = (
+            self.py_impl is None
+            or "py" in interpreter
+            or any(impl in interpreter for impl in self.py_impl)
+        )
+
+        wheel_py_version = self.parse_py_version(interpreter)
+        compatible_py_version = self.py_version is None or (
+            int(wheel_py_version) <= self.py_version
```
taylormadore: Wheels for earlier python versions are only compatible if the ABI is compatible, so I don't think this condition can be evaluated independently of some additional ABI context. Say for example, the user filtered on `cp39` and didn't filter on ABI. I think it would be incorrect to allow a cp38/py38 wheel through the filter unless the abi tag were either `abi3` or `None`

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -249,3 +252,96 @@ def process_package_distributions(
...
+            self.arch is None or platform == "any" or any(arch in platform for arch in self.arch)
+        )
+        compatible_os = (
+            self.os is None or platform == "any" or any(os in platform for os in self.os)
+        )
+
+        return compatible_arch and compatible_os
+
+    def _compatible_tag_abi(self, abi: str) -> bool:
+        return self.abi is None or any(a in abi for a in self.abi)
```
taylormadore: If I understand the pip install behavior correctly, I believe that `abi3` and `none` are always compatible even if the user does not specify them.

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -183,17 +184,37 @@ def process_package_distributions(
...
+        allowed_distros = ["sdist"]
+        to_process = chain(sdists)
+    else:
+        wheels_filter = WheelsFilter(binary_filters)
+        # process both sdists and wheels if no packages are provided in the binary filters
+        if wheels_filter.packages is None:
+            allowed_distros = ["sdist", "wheel"]
+            to_process = chain(sdists, wheels_filter.filter(wheels))
+        # process only wheels for packages in the binary filters
+        elif name in wheels_filter.packages:
```
taylormadore, line 208: We might want to canonicalize these user-specified package names to ensure we align with the registry names. We already do that for the PipRequirement names.

hermeto/core/package_managers/pip/package_distributions.py
slimreaper35: Probably I misunderstood something. I had this condition before.

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -171,17 +189,37 @@ def process_package_distributions(
...
```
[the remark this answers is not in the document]
slimreaper35: We already do `requirement.package = canonicalize_name(req.name)`

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +236,147 @@ def process_package_distributions(
...
+        "310"
+        >>> parse_py_version("py3")
+        "3"
+        """
+        match = re.fullmatch(r"([a-z]+)(\d+)", interpreter)
+        if match is None:
+            raise ValueError(f"Invalid wheel interpreter: {interpreter}")
+
+        return match.group(2)
+
```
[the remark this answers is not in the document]
slimreaper35: duplicate

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +241,146 @@ def process_package_distributions(
...
+        >>> parse_py_version("pp312")
+        312
+        >>> parse_py_version("py3")
+        3
+        """
+        match = re.fullmatch(r"[a-z]+(\d+)", interpreter)
+        if match is None:
+            raise RuntimeError(f"Invalid wheel interpreter: {interpreter}")
+
+        return int(match.group(1))
```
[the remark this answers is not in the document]
slimreaper35: addressed

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -335,3 +337,190 @@ def test_process_package_distributions_noncanonical_version(
...
+        with pytest.raises(ValueError):
+            PipBinaryFilters(py_version=3.11)
+
+    def test_init_with_platform_and_os_or_arch_fails(self) -> None:
+        with pytest.raises(ValueError):
+            PipBinaryFilters(platform="manylinux.*", os="linux", arch="aarch64")
+
+    def test_init_with_invalid_platform_regex_fails(self) -> None:
+        with pytest.raises(ValueError):
+            PipBinaryFilters(platform="*")
```
eskultety: These failure variations are definitely a parametrize candidate.

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -335,3 +337,190 @@ def test_process_package_distributions_noncanonical_version(
...
+
+    def test_filter_with_whitespace_in_constraints(self) -> None:
```
eskultety: This one belongs to the "init" tests above, it's just another string input variation.

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -335,3 +337,190 @@ def test_process_package_distributions_noncanonical_version(
     assert artifacts[0].package_type == "sdist"
     assert artifacts[0].version == requested_version
     assert all(w.version == requested_version for w in artifacts[1:])
+
```
eskultety, line 317: For all of the test coverage below parametrizing can be used to reduce code redundancy and be consistent with how we approach unit tests elsewhere. These are really simple unit tests, so there's little incentive in not using parametrize IMO and it'll result in slightly fewer LOC.

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -94,14 +94,13 @@ def test_process_non_existing_package_distributions(
...
-    caplog: pytest.LogCaptureFixture,
 ) -> None:
-    package_name = "aiowsgi"
+    package_name = "pkg"
     version = "0.1.0"
     req = mock_requirement(package_name, "pypi", version_specs=[("==", version)])
 
     file_1 = package_name + "-" + version + "-py3-none-any.whl"
-    file_2 = package_name + "-" + version + "-manylinux1_x86_64.whl"
+    file_2 = package_name + "-" + version + "-cp311-cp311-any.whl"
```
eskultety, line 103: This caught my eye, what was the issue with `-manylinux...` here?

slimreaper35: It is an invalid wheel filename, so it would be skipped with a warning, and the test would fail.

tests/unit/package_managers/pip/test_package_distributions.py
slimreaper35: Grouped all test_init tests, but does not seem I reduced LOC drastically.

eskultety: I didn't say it would be a "drastic" improvement :D

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -335,3 +337,190 @@ def test_process_package_distributions_noncanonical_version(
...
+                    "py_impl": ":all:",
+                    "abi": ":all:",
+                },
+                {
+                    "packages": None,
+                    "arch": None,
+                    "os": None,
+                    "py_impl": None,
+                    "abi": None,
+                },
```
eskultety, line 367: nitpick: You might also replace this with `{ key: None for key in ["packages", "arch", "os", "py_impl", "abi"] },` to fit it on a single line.

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -335,3 +337,190 @@ def test_process_package_distributions_noncanonical_version(
...
+    def test_filter_with_invalid_wheel_filename(self) -> None:
+        filters = PipBinaryFilters()
+        wheels_filter = WheelsFilter(filters)
+
+        wheels = [mock_pypi_simple_distribution_package("foo.whl", "1.0.0")]
+
+        result = wheels_filter.filter(wheels)
+        assert result == []
+
+    def test_filter_with_no_matching_wheels(self) -> None:
```
eskultety: FWIW these filtering tests are also pretty much identical minus configuration and so can be consolidated as well.

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -335,3 +337,190 @@ def test_process_package_distributions_noncanonical_version(
...
+        for attr, expected_value in expected.items():
+            assert getattr(wheels_filter, attr) == expected_value
+
+    @pytest.mark.parametrize(
+        "filter_kwargs",
+        [
+            pytest.param(
+                {"py_version": 3.11},
+                id="invalid_py_version_type",
+            ),
```
eskultety: nitpick: This fits on a single line.

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -335,3 +337,190 @@ def test_process_package_distributions_noncanonical_version(
...
+            pytest.param(
+                {"platform": "manylinux.*", "os": "linux", "arch": "aarch64"},
+                id="invalid_platform_os_and_arch_combination",
+            ),
+            pytest.param(
+                {"platform": "*"},
+                id="invalid_platform_regex_syntax",
+            ),
```
eskultety: nitpick: This also fits on a single line, can be placed after the first param.

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -335,3 +337,190 @@ def test_process_package_distributions_noncanonical_version(
...
+                id="all_keyword",
+            ),
+        ],
+    )
+    def test_init_with_valid_values(self, filter_kwargs: dict, expected: dict) -> None:
+        filters = PipBinaryFilters(**filter_kwargs)
+        wheels_filter = WheelsFilter(filters)
+
+        for attr, expected_value in expected.items():
+            assert getattr(wheels_filter, attr) == expected_value
```
eskultety, line 377: I have to say I'm impressed with Claude's "reproducible" behaviour - for different prompts it generated word for word identical diff, that's quite useful.

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -203,44 +238,155 @@ def process_package_distributions(
...
+            or "py" in interpreter
+            or any(impl in interpreter for impl in self.py_impl)
+        )
+
+        wheel_py_version = _parse_py_version(interpreter)
+        compatible_py_version = self.py_version is None or (
+            wheel_py_version <= self.py_version and abi in ("abi3", "none")
+        )
+
+        return compatible_py_impl and compatible_py_version
```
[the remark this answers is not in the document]
eskultety: Ouch, the review bot is very correct here, especially in its second suggestion. I was looking for that in my review and yet I still overlooked it.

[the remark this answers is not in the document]
eskultety: Also, when applying the fix, please follow the line breaks for the long expression as suggested as I forgot to mention that.

tests/unit/package_managers/pip/test_package_distributions.py
a-ovchinnikov: Is this code Claude-assisted? If yes then I believe it should be mentioned in the commit message.

tests/unit/package_managers/pip/test_package_distributions.py
a-ovchinnikov: Or use `defaultdict(types.NoneType)`. At the same time this will make test case less straightforward, test case parameters are probably the only place where having dictionary elements spread over multiple vertical lines makes sense.

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -335,3 +337,190 @@ def test_process_package_distributions_noncanonical_version(
...
+
+        result = wheels_filter.filter(wheels)
+        assert result == []
+
+    def test_filter_with_no_matching_wheels(self) -> None:
+        filters = PipBinaryFilters(arch="aarch64", os="macos")
+        wheels_filter = WheelsFilter(filters)
+
+        wheels = [
+            # wrong arch
```
a-ovchinnikov: Code-shape nitpick (optional, non-blocking): this comment effectively names an object, thus the same rule as for a block of code could be applied:

```python
wrong_arch_wheel = mock_pypi_simple_distribution_package("package-1.0.0-cp39-cp39-macos_10_9_x86_64.whl", "1.0.0")
wrong_os_wheel = mock_pypi_simple_distribution_package("package-1.0.0-cp39-cp39-linux_aarch64.whl", "1.0.0")

wheels = [wrong_arch_wheel, wrong_os_wheel]
```

tests/unit/package_managers/pip/test_package_distributions.py
eskultety: I marked this as nitpick, so even no action is acceptable.

tests/unit/package_managers/pip/test_package_distributions.py
slimreaper35: I reworked the class a bit more and added more cases to address https://github.com/hermetoproject/hermeto/pull/1143#discussion_r2486702659

tests/unit/package_managers/pip/test_package_distributions.py
slimreaper35: I always prefer better code over comments, so I applied that everywhere.

taylormadore (approved):

PR #1149, opened 2025-10-10
models: Silence CodeQL complaints about missing @classmethod decorators

eskultety:
CodeQL started to randomly complain that we use an unusual denominator 'cls' for model methods without being marked as classmethods explicitly. These complaints are annoying but the truth is we weren't exactly following Pydantic's official documentation where they declare validators like this in the examples [1]

[1] https://docs.pydantic.dev/latest/concepts/validators/

Resolves https://github.com/hermetoproject/hermeto/issues/1147

### DIFF
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -134,6 +134,7 @@ class _PackageInputBase(pydantic.BaseModel, extra="forbid"):
     path: Path = Path(".")
 
     @pydantic.field_validator("path")
+    @classmethod
     def _path_is_relative(cls, path: Path) -> Path:
         return check_sane_relpath(path)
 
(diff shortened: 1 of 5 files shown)


slimreaper35:
I see a couple of more cases (unless we don't want to touch _rpm_)
- https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/yarn_classic/workspaces.py#L27
- https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/rpm/redhat.py

eskultety:
> I see a couple of more cases (unless we don't want to touch _rpm_)
> 
>     * https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/yarn_classic/workspaces.py#L27
> 
>     * https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/rpm/redhat.py

Sigh, fixed.

a-ovchinnikov:
To be fair the docs ask for a callable, looks like a `@staticmethod` could have mostly worked as well since `cls` is used only in SPDX (and IIRC it is used to call a static method anyway).

PR #1154, opened 2025-10-13
Drop Containerfile in favor of Dockerfile

slimreaper35:


eskultety:
My preference would be to stick with Containerfile, however, from a practical POV using a Dockerfile instead makes more sense:
- people intuitively look for its presence in the repo
- dependabot doesn't recognize Containerfile we'd have to switch to renovate for dep updates.

We could keep the symlinks, but since it's no longer necessary I'd say it's more confusing than beneficial.

ben-alkov:
I heard, a while ago (sorry, can't remember source), that RHT was de-emphasizing "Containerfile" as it:

- wasn't really gaining traction
- was seen in some circles as actively divisive

We may just have to deal with the Kleenex/Xerox effect, and live with "Dockerfile"

a-ovchinnikov:
Let's drop Containerfile in favor of Dockerfile.

tests/integration/utils.py
```diff
@@ -33,7 +33,6 @@
         Path("hermeto"),
         Path("tests/integration"),
         Path("Dockerfile"),
-        Path("Containerfile"),
```
eskultety, line 36: There's one occurrence in `build_image_and_check_cmd` as well.

slimreaper35: But that should be for integration tests only (in `test_data_dir.joinpath(test_case, "container")`.

eskultety: True.

PR #1155, opened 2025-10-17
Implement bundler platform filtering

### LINKED ISSUE #1075: Implement bundler platform filtering
Using the new input models from https://github.com/hermetoproject/hermeto/issues/1072, implement platform filtering for the bundler backend according to that [section of the design](https://github.com/hermetoproject/hermeto/blob/main/docs/design/binary-platform-filtering.md#bundler). Ensure that the new behavior adheres to the fetching strategies defined [here](https://github.com/hermetoproject/hermeto/blob/main/docs/design/binary-platform-filtering.md#binary-fetching-strategies).

slimreaper35:
Closes #1075

### DIFF
--- a/hermeto/core/package_managers/bundler/parser.py
+++ b/hermeto/core/package_managers/bundler/parser.py
@@ -27,10 +27,7 @@ BundlerDependency = GemDependency | GemPlatformSpecificDependency | GitDependenc
 ParseResult = list[BundlerDependency]
 
 
-def parse_lockfile(
-    package_dir: RootedPath, binary_filters: BundlerBinaryFilters | None = None
-) -> ParseResult:
-    """Parse a Gemfile.lock file and return a list of dependencies."""
+def _ensure_bundler_files_exist(package_dir: RootedPath) -> None:
     lockfile_path = package_dir.join_within_root(GEMFILE_LOCK)
     gemfile_path = package_dir.join_within_root(GEMFILE)
     if not lockfile_path.path.exists() or not gemfile_path.path.exists():
@@ -41,12 +38,19 @@ def parse_lockfile(
         )
         raise PackageRejected(reason=reason, solution=solution)
 
+
+def parse_lockfile(
+    package_dir: RootedPath, binary_filters: BundlerBinaryFilters | None = None
+) -> ParseResult:
+    """Parse a Gemfile.lock file and return a list of dependencies."""
+    _ensure_bundler_files_exist(package_dir)
+
     scripts_dir = Path(__file__).parent / "scripts"
     lockfile_parser = scripts_dir / "lockfile_parser.rb"
     try:
         output = run_cmd(cmd=[str(lockfile_parser)], params={"cwd": package_dir.path})
-    except subprocess.CalledProcessError:
-        raise PackageManagerError(f"Failed to parse {lockfile_path}")
+    except subprocess.CalledProcessError as e:
+        raise PackageManagerError("Failed to parse Gemfile.lock") from e
 
     json_output = json.loads(output)
 
--- a/tests/unit/package_managers/bundler/test_parser.py
+++ b/tests/unit/package_managers/bundler/test_parser.py
@@ -84,7 +84,7 @@ def test_parse_lockfile_os_error(
     with pytest.raises(PackageManagerError) as exc_info:
         parse_lockfile(rooted_tmp_path)
 
-    assert f"Failed to parse {empty_bundler_files[1]}" in exc_info.value.friendly_msg()
+    assert "Failed to parse Gemfile.lock" in exc_info.value.friendly_msg()
 
 
 @mock.patch("hermeto.core.package_managers.bundler.parser.run_cmd")


hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -29,7 +29,20 @@
...
+      s[:type] == 'rubygems' &&
+      s[:source] == parsed_spec[:source]
+    }
+
+    if existing_spec
+      # extend the platforms array
+      existing_spec[:platforms] << parsed_spec[:platforms].first
+    else
+      parsed_specs << parsed_spec
+    end
```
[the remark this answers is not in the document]
slimreaper35: Well, that would be a bug for the bundler parser lib.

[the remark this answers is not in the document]
eskultety: Indeed, actually more of a Gemfile.lock formatter issue, the parser doesn't really have that much of a wiggle room to be borked like this (not impossible though). I would hope that if anything `bundler` would have caught this one way or the other before it gets to us, so I'm inclined to just resolve this.

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -35,43 +38,94 @@ def parse_lockfile(package_dir: RootedPath, allow_binary: bool = False) -> Parse
...
+        return NotImplemented
+
+    def _apply_prefer_binary_mode(self, gem: dict[str, Any]) -> None:
+        if "ruby" in gem["platforms"] and len(gem["platforms"]) > 1:
+            gem["platforms"].remove("ruby")
+
+    def apply_platform_filters(self, gems: list[dict[str, Any]]) -> None:
+        """Update platforms for each gem based on the filter constraints."""
+        for gem in gems:
+            # all packages & all platforms
```
eskultety, line 111: nitpick: Maybe use 'on' instead of '&'

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -394,17 +399,119 @@ def test_parse_gemlock_detects_binaries_and_skips_then_when_instructed_to_skip(
...
+
+    def test_init_with_specific_values_and_all(self) -> None:
+        filters = BundlerBinaryFilters(
+            packages="rails,:all:,rack",
+            platform="x86_64-linux,:all:,x86_64-darwin",
+        )
+        gems_filter = GemsFilter(filters)
+
+        assert gems_filter.packages is None
+        assert gems_filter.platform is None
```
eskultety, line 451: Same comment about parametrize as with the Python counterpart.

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -35,43 +38,94 @@ def parse_lockfile(package_dir: RootedPath, allow_binary: bool = False) -> Parse
...
+        """Initialize the filter."""
+        self.packages = self._parse_filter_spec(filters.packages)
+        self.platform = self._parse_filter_spec(filters.platform)
+
+    def __contains__(self, item: Any) -> bool:
+        return NotImplemented
+
+    def _apply_prefer_binary_mode(self, gem: dict[str, Any]) -> None:
+        if "ruby" in gem["platforms"] and len(gem["platforms"]) > 1:
+            gem["platforms"].remove("ruby")
```
eskultety, line 106: Hmm, do we need to update https://github.com/hermetoproject/hermeto/blob/main/docs/design/binary-platform-filtering.md#bundler? Because the table says:

> Do not prefetch platform-agnostic gems when user requests binaries for specific packages

and then in the very first example it says we need to fetch the 'ruby' version for `nokogiri`. It's true that in that example the package list is empty, but still don't we want only platform-specific gems for ALL packages in that case and ONLY fetch 'ruby' for those that do not match the filter? I'm confused by the example a bit.

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -46,32 +52,76 @@ def parse_lockfile(package_dir: RootedPath, allow_binary: bool = False) -> Parse
...
+class GemsFilter(BinaryPackageFilter):
+    """Filter gems based on the filter constraints."""
+
+    def __init__(self, filters: BundlerBinaryFilters) -> None:
+        """Initialize the filter."""
+        self.packages = self._parse_filter_spec(filters.packages)
+        self.platform = self._parse_filter_spec(filters.platform)
+
+    def __contains__(self, item: Any) -> bool:
+        return NotImplemented
```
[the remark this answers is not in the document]
eskultety: Yes, this concerns me too. Maybe we don't really need to define `__contains__` on the abstract class?

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -35,43 +38,94 @@ def parse_lockfile(package_dir: RootedPath, allow_binary: bool = False) -> Parse
...
+
+    def _apply_prefer_binary_mode(self, gem: dict[str, Any]) -> None:
+        if "ruby" in gem["platforms"] and len(gem["platforms"]) > 1:
+            gem["platforms"].remove("ruby")
+
+    def apply_platform_filters(self, gems: list[dict[str, Any]]) -> None:
```
eskultety, line 108: This is inconsistent with the existing wheels handling leading to the odd `__contains__` handling IMO. The main source of inconsistency here is that this operation is destructive because it modifies the input argument directly without returning a result. I think we want to return a new iterable like we do with wheels and `filter` would sounds more appropriate for what it should do.

In case we want to keep `__contains__` around and not return `NotImplemented` then I think the fact `Gemfile.lock` contains some duplicate entries plays in our favour to actually write something to the magic method and process this entry by entry.

hermeto/core/package_managers/bundler/parser.py
taylormadore: It looks like I may just have missed updating the example after we decided on the `prefer-binary` approach

taylormadore (commented):
I feel like we're going to need to make use of the [Gem:Platform](https://github.com/rubygems/rubygems/blob/8ad4509f95c90cf9523a82ca917b6b842fd37132/lib/rubygems/platform.rb#L10) class to normalize the user-specified platforms and to do the platform compatibility matching.

eskultety:
> I feel like we're going to need to make use of the [Gem:Platform](https://github.com/rubygems/rubygems/blob/8ad4509f95c90cf9523a82ca917b6b842fd37132/lib/rubygems/platform.rb#L10) class to normalize the user-specified platforms and to do the platform compatibility matching.

@tmadore do we need to get our hands dirty with Ruby? Is there a scenario where a package listed multiple times in the lockfile would not be listed in e.g. "Ruby" variant except that it should? Can't we take the lockfile at face value? Can we detect lockfile changes just like we do e.g. with Yarn, let the user correct any discrepancies and then rely fully on the content in the file and then process and match the gems against the filter one by one?

PR #1159, opened 2025-10-30
gomod: Do not expose automatically downloaded toolchains in output dir

eskultety:
We never officially supported this, yet some users out there did make use of cached toolchains. This is problematic from the multi-arch support POV, so stop exposing any toolchains Go may have downloaded via the means of GOTOOLCHAIN=auto.

Closes https://github.com/hermetoproject/hermeto/issues/557

### DIFF
--- a/hermeto/core/package_managers/gomod.py
+++ b/hermeto/core/package_managers/gomod.py
@@ -641,6 +641,16 @@ def _clean_go_modcache(go: Go, dir_: Optional[StrPath]) -> None:
         go(["clean", "-modcache"], {"env": {"GOPATH": dir_, "GOCACHE": dir_}})
 
 
+def _list_toolchain_files(dir_path: str, files: list[str]) -> list[str]:
+    def is_a_toolchain_path(path: Union[str, os.PathLike[str]]) -> bool:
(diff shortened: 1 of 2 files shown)


hermeto/core/package_managers/gomod.py
```diff
@@ -642,6 +642,16 @@ def _clean_go_modcache(go: Go, dir_: Optional[StrPath]) -> None:
         go(["clean", "-modcache"], {"env": {"GOPATH": dir_, "GOCACHE": dir_}})
 
 
+def _ignore_toolchain_files(dir_path: str, files: list[str]) -> list[str]:
```
a-ovchinnikov: Minor naming nitpick: WDYT about replacing `_ignore_toolchain_files` with `_filter_toolchain_files_to_ignore` or even just `_select_toolchain_files`?

eskultety: Went with `_list_toolchain_files` since that's what the function does :) . I didn't want the `select` prefix to be confused with `_select_toolchain` which is a very different thing and part of the core gomod module's logic. This here is more like an auxiliary method.

PR #1161, opened 2025-10-31
Revert "gomod: _resolve_gomod: Move the global go env variables 1 level up"

eskultety:
This reverts commit 82cfd9875858c3dedb485b87c01f18e00827273f.

The original idea was to have a single shared Go environment variable set for the whole execution. This idea had a major flaw - the variable set (a dict) is mutable and was being modified during resolve!

Most importantly, the value of the GOMODCACHE variable changes depending on whether we're processing a Go vendored project or not which led to the GOMODCACHE variable affecting all Go projects on the input. The difference in the variable setting between a regular fetch and the vendoring mode is that during vendoring Go may still download some data which end up in the output cache which is undesirable and so we point GOMODCACHE to a location that is wiped after every input dependency fetch.

Unfortunately, the bug introduced by the commit above means that regular fetches got affected by an incorrect value of the GOMODCACHE variable that was carried over and so all fetched modules were wiped in the end, leading to Go hermetic build errors following this pattern:

    pkg/repo/<module>.go:20:2: github.com/google/<mod>@<ver>:
        reading file:///hermeto/output/deps/gomod/<path_to_module>:
            no such file or directory

### DIFF
--- a/tests/integration/test_data/gomod_e2e_vendor_nonvendor_module_mix_ordering_2/.build-config.yaml
+++ b/tests/integration/test_data/gomod_e2e_vendor_nonvendor_module_mix_ordering_2/.build-config.yaml
@@ -1,0 +1,10 @@
+environment_variables:
+- name: GOCACHE
+  value: ${output_dir}/deps/gomod
+- name: GOMODCACHE
+  value: ${output_dir}/deps/gomod/pkg/mod
+- name: GOPATH
+  value: ${output_dir}/deps/gomod
+- name: GOPROXY
+  value: file://${GOMODCACHE}/cache/download
+project_files: []
--- a/tests/integration/test_gomod.py
+++ b/tests/integration/test_gomod.py
@@ -268,6 +268,36 @@ def test_gomod_packages(
             [""],
             id="gomod_e2e_1.22_workspace_vendoring",
         ),
+        # This test case is about ordering of package subpaths
+        pytest.param(
+            utils.TestParameters(
+                branch="gomod/e2e-vendor-nonvendor-module-mix",
+                packages=(
+                    {"path": "vendored-module", "type": "gomod"},
+                    {"path": "non-vendored-module", "type": "gomod"},
+                ),
+                expected_exit_code=0,
+                expected_output="All dependencies fetched successfully",
+            ),
+            [],  # check using CMD defined in Dockerfile
+            [""],
+            id="gomod_e2e_vendor_nonvendor_module_mix_ordering_1",
+        ),
+        # This test case is about ordering of package subpaths
+        pytest.param(
+            utils.TestParameters(
+                branch="gomod/e2e-vendor-nonvendor-module-mix",
+                packages=(
+                    {"path": "non-vendored-module", "type": "gomod"},
+                    {"path": "vendored-module", "type": "gomod"},
+                ),
+                expected_exit_code=0,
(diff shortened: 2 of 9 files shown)


eskultety:
Opening as draft until I introduce a corresponding integration test.

eskultety:
Integration test added: https://github.com/hermetoproject/integration-tests/pull/30

Currently, this references my fork until the test case is merged. Some integration test failures are expected due to missing branches in my fork, the important test cases being tagged as `gomod/e2e-vendor-nonvendor-module-mix`

tests/integration/test_gomod.py
```diff
@@ -268,6 +268,34 @@ def test_gomod_packages(
             [""],
             id="gomod_e2e_1.22_workspace_vendoring",
         ),
+        pytest.param(
+            utils.TestParameters(
+                branch="gomod/e2e-vendor-nonvendor-module-mix",
+                packages=(
```
a-ovchinnikov, line 275: Just a minor suggestion: please add a comment right here about the importance of ordering,  it might end up optimized out otherwise. Or maybe have it identified as "gomod_e2e_vendor_nonvendor_module_mix_direct_order" and "gomod_e2e_vendor_nonvendor_module_mix_reverse_order".

a-ovchinnikov (commented):
LGTM with a minor suggestion.

tests/integration/test_gomod.py
eskultety: I added an explanatory comment and went with `_ordering_1/2` as `direct_order` and `reverse_order` sound nice, but in the end don't tell you anything as there's no recommended schema for input ordering (as it shouldn't).

tests/integration/test_data/gomod_e2e_vendor_nonvendor_module_mix_ordering_2/.build-config.yaml
```diff
@@ -0,0 +1,10 @@
+environment_variables:
+- name: GOCACHE
+  value: ${output_dir}/deps/gomod
+- name: GOMODCACHE
+  value: ${output_dir}/deps/gomod/pkg/mod
+- name: GOPATH
+  value: ${output_dir}/deps/gomod
+- name: GOPROXY
+  value: file://${GOMODCACHE}/cache/download
+project_files: []
```
slimreaper35, line 10: Could this file also be a symlink?

eskultety: Possibly, not sure. Well, technically symlinks would not prevent the `file://` from working correctly because the resolution happens on the OS level, but whether `go` would complain, that's the thing I don't know. Why?

slimreaper35: I just saw that all files in `gomod_e2e_vendor_nonvendor_module_mix_ordering_2` are symlinks.

eskultety: Right, but that's of no significance either to the test or prefetch itself, because it only affects where the test stores input data and output data for comparison purposes, whatever you see in that `build-config.yaml` file is pretty much irrelevant to the core logic, otherwise our `RootedPath` check would have already complained.

Issue #1018, opened 2025-07-08
Create a folder to host miscellaneous scripts


joelcapitao:
In CoreOS, we have a python script [1] that updates the checksum of the artifacts in `artifacts.lock.yaml`.
I believe this script could be useful to others using the generic driver, but I don't know where would be the best place to push such a script here.
We can maybe consider creating a misc folder e.g: `hack/`, what do you think ?

[1] https://github.com/coreos/coreos-assembler/blob/83fdbc45ba4977b878d4ab73f1ac4205abe7a48a/ci/hermetic/update_artifacts_lockfile

taylormadore:
@jcapiitao Thanks for sharing that script with us--I can definitely see how that could be a valuable tool for users of the generic fetcher. We don't currently have a policy for how/where to handle utility script contributions like this, so we'll need some time to deliberate as a team and decide the best way forward. We'll get back to you!

eskultety:
@jcapiitao thank you for the suggestion. I'd like to clarify on the expectation here - do you expect this to just live in our main repository and be maintained by us or do you expect these helpers to find their way to our container image that we ship and be packaged there too? Understanding this detail will help us decide on our next steps.

joelcapitao:
> [@jcapiitao](https://github.com/jcapiitao) thank you for the suggestion. I'd like to clarify on the expectation here - do you expect this to just live in our main repository and be maintained by us or do you expect these helpers to find their way to our container image that we ship and be packaged there too? Understanding this detail will help us decide on our next steps.

As it's helper scripts, so not core, I wouldn't expect them to be packaged in the container image.
Adding a section in the documentation, explaining that a collection of helper scripts exist and can be used to streamline hermeto use (e.g updating checksums of artifacts lockfile), is sufficient IMHO.

eskultety:
Thank you for the clarification on the expectations, we'll discuss further within the team how to approach this best.

joelcapitao:
What's the status of this request ?

slimreaper35:
Hi @joelcapitao 
Apologies for a delayed reply. I think you can open a PR, and we can go from there. We already have a `hack` directory in the repo.

Our main concern was that we didn't want to make these scripts too official and to be "signed" by maintainers, but rather to be some community helpers. Let's say someone finds a bug in a script and expects that it will be fixed immediately <- that's not what we want. But I hope it can be clarified in a simple README file.

eskultety:
[...]

PR #1165, opened 2025-11-02
Replace gomod.sh with gomod.py

a-ovchinnikov:
This PR introduces a gomod.py script for generating unit tests data for go PM in a more structured way than before.

To verify the correctness of the port I observed the difference between updates to test files generated by the shell script and then by Python port and found no differece:

```
$ ./hack/mock-unittest-data/gomod.sh
...
$ git diff tests/* > /tmp/foo/gomodsh
$ wc -l /tmp/foo/gomodsh 
60153 /tmp/foo/gomodsh
$ git restore tests/unit/data/*
$ ./hack/mock-unittest-data/gomod.py
...
$ git diff tests/* > /tmp/foo/gomodpy
$ diff -qs /tmp/foo/gomodpy /tmp/foo/gomodsh
Files /tmp/foo/gomodpy and /tmp/foo/gomodsh are identical 
```

Resolves: https://github.com/hermetoproject/hermeto/issues/776


eskultety:
@a-ovchinnikov please use one of the recognized keywords for automatic issue linking :) : https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

a-ovchinnikov:
>  one of the recognized keywords

Derp, for some reason I thought 'Addresses' is one of those.

slimreaper35:
WDYT about putting the script into, for example, `tests/unit/scripts` directory and reserving the current `hack` directory for user-facing #1018 ?

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env python3
+import json
+import subprocess
+from collections import ChainMap
```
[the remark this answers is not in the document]
a-ovchinnikov: Done.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,154 @@
...
+
+def generate_auxiliry_dirs(mocked_data_dir):
+    mkdir_p(f"{mocked_data_dir}/non-vendored")
+    mkdir_p(f"{mocked_data_dir}/vendored")
+    mkdir_p(f"{mocked_data_dir}/workspaces")
+    mocked_data_dir_abspath = Path(f"{mocked_data_dir}").absolute()
+
+    tmpdir = Path(gettempdir())
+
+    return mocked_data_dir_abspath, tmpdir
```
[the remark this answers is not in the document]
a-ovchinnikov: Now this one was useful.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,154 @@
...
+        yield pandemo_dir
+    finally:
+        delete_directory_with_contents(pandemo_dir)
+
+
+def inject_cwd(run, pandemo_dir):
+    _, _, fpart = run.__reduce__()
+    f, args, kwargs, _dct = fpart
+    kwargs |= {"cwd": pandemo_dir}
+    run.__setstate__((f, args, kwargs, _dct))
```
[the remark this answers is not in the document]
a-ovchinnikov: No.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,154 @@
...
+    inject_cwd(run, pandemo_dir)
+
+
+def main() -> None:
+    print_banner("Generating mock data for gomod unit tests")
+    mocked_data_dir = Path(argv[1] if len(argv) > 1 else "tests/unit/data/gomod-mocks")
+    mocked_data_dir_abspath, tmpdir = generate_auxiliry_dirs(mocked_data_dir)
+    with pandemonium_dir(tmpdir) as pandemo_dir:
+        params = {"env": environ | {"GOMODCACHE": tmpdir/"hermeto-mock-gomodcache"}, "check": True}
+        run = partial(subprocess.run, **params)
```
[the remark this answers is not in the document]
a-ovchinnikov: It did not.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,154 @@
...
+        run(_c("git switch go-1.22-workspaces"))
+        print(f"generating {mocked_data_dir}/workspaces/go.sum")
+        copyfile(pandemo_dir/"go.sum", Path(f"{mocked_data_dir_abspath}/workspaces/go.sum"))
+        run_redirect = partial(
+            run_with_redirection, run=run, mocked_data_dir=mocked_data_dir,
+            mocked_data_dir_abspath=mocked_data_dir_abspath
+        )
+        run_redirect(where="workspaces/go_list_modules.json", what="go list -m -json")
+        run_redirect(where="workspaces/go_mod_download.json", what="go mod download -json")
+        run_redirect(where="workspaces/go_mod_download.json", what="go mod download -json")
```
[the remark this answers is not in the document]
a-ovchinnikov: This one is right, but for a wrong reason, I missed a string when running conversion.

a-ovchinnikov:
> WDYT about putting the script into ... tests/unit/scripts directory 

I'd rather kept is in a familiar place, at least for now and moved it once it does not make sense to keep it in its present location.

a-ovchinnikov:
Since previous revision I have found a lost banner and replaced it.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+    if vendored_diff or non_vendored_diff:
+        nonvendor = f"{mocked_data_dir}/expected-results/resolve_gomod.json"
+        vendor = f"{mocked_data_dir}/expected-results/resolve_gomod_vendored.json"
+        banner = f"""The mock data changed => the expected unit test results may change.
+            The following files may need to be adjusted manually:
+            {nonvendor if non_vendored_diff else ""}
+            {vendor if vendored_diff else ""}
+        """
+        print_banner(banner)
```
a-ovchinnikov, line 79: This was helpful.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
+#!/usr/bin/env python3
+import json
+import subprocess
+from contextlib import contextmanager
+from functools import partial
+from itertools import chain
+from os import environ
+from pathlib import Path
+from shutil import copyfile
+from shutil import rmtree as delete_directory_with_contents
```
eskultety: You don't need this, just use `tempfile.TemporaryDirectory` directly, see below.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+    return mocked_data_dir_abspath, tmpdir
+
+
+@contextmanager
+def pandemonium_dir(pandemonium_root):
+    pandemo_dir = pandemonium_root / "gomod-pandemonium"
+    try:
+        yield pandemo_dir
+    finally:
+        delete_directory_with_contents(pandemo_dir)
```
eskultety: This looks very unnecessary since you're basically just doing:
```
with tempfile.TemporaryDirectory() as  tmpdir:
  x = Path(tmpdir.name) / "gomod-pandemonium" . mkdir()
```

I would argue that this helper context manager not only doesn't improve readabilty, it also goes against some commonly used pattern (tempfile context manager), reimplementing what you already get for free with `tempfile`, please drop this.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+
+def inject_cwd(run, pandemo_dir):
+    _, _, fpart = run.__reduce__()
+    f, args, kwargs, _dct = fpart
+    kwargs |= {"cwd": pandemo_dir}
+    run.__setstate__((f, args, kwargs, _dct))
+
+
+def _c(command_string):
+    return command_string.split()
```
eskultety: This is incredibly opaque and totally unnecessary. If instead of making `run` a partial alias you just define `run` as a regular wrapper function over `subprocess.run` taking `**kwargs` for subprocess arguments and letting other functions in this module will be able to access the function easily.

As for `_c` just by looking at the name it doesn't tell you what it does at all! You also don't need to do the split like this if you define `run` as a standalone method:
```
def run(command: str, **kwargs) -> subprocess.CalledProcess:
  return subprocess.run(command.split(), **kwargs)
```

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def run_with_redirection(run, where, what, mocked_data_dir, mocked_data_dir_abspath):
```
eskultety: It took me half a minute to connect the name (redirection) with what you meant, likely it's just me.
nitpick: how about `write_to_file` instead?

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+    print(f"generating {mocked_data_dir}/{where}")
+    with open(f"{mocked_data_dir_abspath}/{where}", "w") as f:
+        run(_c(what), stdout=f)
+
+
+def swap_temporary_directories_for_variables(tmpdir, pandemo_dir, mocked_data_dir):
```
eskultety: This name also confused me, because you can't really swap directories and variables, can you ? :) It's a replacement, not a swap.

How about `replace_paths_with_placeholder`?

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+    return Path(dirname).mkdir(exist_ok=True, parents=True)
+
+
+def generate_auxiliary_dirs(mocked_data_dir):
+    mkdir_p(f"{mocked_data_dir}/non-vendored")
+    mkdir_p(f"{mocked_data_dir}/vendored")
+    mkdir_p(f"{mocked_data_dir}/workspaces")
+    mocked_data_dir_abspath = Path(f"{mocked_data_dir}").absolute()
+
+    tmpdir = Path(gettempdir())
```
eskultety: This logic doesn't really fall into the context of the function, does it? It's supposed to create additional dirs, but you're also creating an additional directory. Not only should this happen inside `main`'s context manager but if you use the standard `with tempfile.TemporaryDirectory` pattern, you don't even need to do this and can make this function simpler.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+
+def mkdir_p(dirname):
+    return Path(dirname).mkdir(exist_ok=True, parents=True)
+
+
+def generate_auxiliary_dirs(mocked_data_dir):
+    mkdir_p(f"{mocked_data_dir}/non-vendored")
+    mkdir_p(f"{mocked_data_dir}/vendored")
+    mkdir_p(f"{mocked_data_dir}/workspaces")
+    mocked_data_dir_abspath = Path(f"{mocked_data_dir}").absolute()
```
eskultety: This function is pretty much the only place that works with `mocked_data_dir` as a relative path, the rest of the script uses the absolute path. You can just use the relative paths everywhere, you don't need the absolute path, it'll make the code simpler in terms of reducing the number of arguments passed around, keep reading on I'll comment elsewhere....

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+from pathlib import Path
+from shutil import copyfile
+from shutil import rmtree as delete_directory_with_contents
+from sys import argv
+from tempfile import gettempdir
+
+from yarn import print_banner
+
+
+def mkdir_p(dirname):
```
eskultety: nitpick: `mkdir_p` is easily optically lost IMO, would `mkdir_w_parents` work better?

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+
+def warn_about_potential_manual_intervention(mocked_data_dir):
+    non_vendored_diff = subprocess.run(
+        ["git", "diff", "--", f"{mocked_data_dir}/non-vendored"],
+        capture_output=True, text=True, check=True
+    ).stdout
+    vendored_diff = subprocess.run(
+        ["git", "diff", "--", f"{mocked_data_dir}/vendored"],
+        capture_output=True, text=True, check=True
+    ).stdout
```
eskultety: Once you define `run` as a standalone function, you can remain consistent within the whole script and just call `run` here as well and let subprocess be invoked from a single place. Also, you can define `text` as a common argument within `run` as well, doesn't hurt any invocation. It'll also fit on a single line once you do that, pretty much :) .

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+            {vendor if vendored_diff else ""}
+        """
+        print_banner(banner)
+
+
+def main() -> None:
+    print_banner("Generating mock data for gomod unit tests")
+    mocked_data_dir = Path(argv[1] if len(argv) > 1 else "tests/unit/data/gomod-mocks")
+    mocked_data_dir_abspath, tmpdir = generate_auxiliary_dirs(mocked_data_dir)
+    with pandemonium_dir(tmpdir) as pandemo_dir:
```
eskultety: Like I mentioned earlier, this will be much better suited as:
```
with tempfile.TemporaryDirectory as tmpdir:
  pandemo_dir = tmpdir  / PANDEMO_DIR    # yes, a constant to enable all callers to access it
```

because you get the cleanup for free.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+        params = {"env": environ | {"GOMODCACHE": tmpdir/"hermeto-mock-gomodcache"}, "check": True}
+        run = partial(subprocess.run, **params)
```
eskultety: I already mentioned this, please define `run` as a standalone method.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+        run(_c(f"git clone https://github.com/cachito-testing/gomod-pandemonium {pandemo_dir}"))
+        inject_cwd(run, pandemo_dir)
```
eskultety: If you define `run` as a standalone method, this questionable inject will no longer be necessary. It's IMO not idiomatic compared to overriding `cwd` wherever needed, especially since you have `params` defined above, which you can enhance with `cwd` as an additional arg and then do `run(command, **params)` everywhere.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,175 @@
...
+        run(_c("git switch go-1.22-workspaces"))
+        print(f"generating {mocked_data_dir}/workspaces/go.sum")
+        copyfile(pandemo_dir/"go.sum", Path(f"{mocked_data_dir_abspath}/workspaces/go.sum"))
+        run_redirect = partial(
+            run_with_redirection, run=run, mocked_data_dir=mocked_data_dir,
+            mocked_data_dir_abspath=mocked_data_dir_abspath
+        )
+        run_redirect(where="workspaces/go_list_modules.json", what="go list -m -json")
```
eskultety: I guess I'd prefer a few for loops instead of a partial, but it's readable, so let's keep `run_redirect` as you have it.

eskultety (commented):
Thank you very much for taking up this tedious work!!! I'm including a few comments because I think we can simplify this and make it slightly more readable. 
Some of the things were harder to capture in comments, so please look at the patch below as a reference to the comments I made and as an additional context for whatever I didn't capture in my comments (e.g. reducing the number of args in function signatures and what should be passed around and what doesn't need to be).

_Edit: It also shaves quite a few LOC too :)_ 
[patch.txt](https://github.com/user-attachments/files/23363219/patch.txt)

hack/mock-unittest-data/gomod.py
a-ovchinnikov: Yes, replaced it with `mkdir_with_parents`.

hack/mock-unittest-data/gomod.py
a-ovchinnikov: Good point, I completely forgot about `tempfile`.

hack/mock-unittest-data/gomod.py
a-ovchinnikov: Yes, it reads better with a helper. As for `_c` it was so auxiliary and at the same time so ubiquitous that I opted for a short name (like `_` for internationalization wrapper for strings in some projects) to declutter the screen. Having a longer name for such an alias would have not resulted in any benefits compared to just calling `.split()` directly on a command. In any case, moving this into a wrapper is a much better approach.

hack/mock-unittest-data/gomod.py
a-ovchinnikov: Yes, the name is obscure when one have not converted it from a shell command with redirection. I like `write_to_file` better.

hack/mock-unittest-data/gomod.py
a-ovchinnikov: Maybe `run_with_stdout_redirection` then? I'll update the code and will revert if it ends up being too verbose.

hack/mock-unittest-data/gomod.py
a-ovchinnikov: Yes, I'll take this advice, however the main reason I experimented with injecting into a partial's context was because I wanted to avoid repeatedly passing `**params` everywhere. I decided that having a little obscure state and having to infrequently update it is fine as long as `run` calls themselves remain as short as possible.

a-ovchinnikov:
> It also shaves quite a few LOC too :) patch.txt

Thank you for the patch! I have applied it and the code did indeed become easier to follow.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,156 @@
...
+        d_paths = [v['DiskPath'] for v in json.loads(f.read())['Use']]
+        paths = [Path(f"{pandemo_dir}")/p.removeprefix("./") for p in d_paths if p != '.']
+
+    for p in paths:
+        print(f"preparing per-workspace directory {p}")
+        mkdir_with_parents(f"{mocked_data_dir}/workspaces/{p}")
+        if (gosum:=Path(p)/"go.sum").exists():
+            print(f"generating {mocked_data_dir}/workspaces/{p.name}/go.sum")
+            copyfile(gosum, Path(f"{mocked_data_dir}/workspaces/{p.name}/go.sum"))
+        run_with_stdout_redirect(
```
eskultety: Since the partial only exists in main, you could make kind of an "exception" here and just call `write_to_file` directly.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,156 @@
...
+        subprocess_kwargs = {
+            "env": environ | {"GOMODCACHE": str(gomodcache)},
+            "cwd": pandemo_dir
+        }
+
+        run(f"git clone https://github.com/cachito-testing/gomod-pandemonium {pandemo_dir}")
+        run("git switch go-1.22-workspaces", **subprocess_kwargs)
+        print(f"generating {mocked_data_dir}/workspaces/go.sum")
+        copyfile(pandemo_dir/"go.sum", Path(f"{mocked_data_dir}/workspaces/go.sum"))
+        run_with_stdout_redirect = partial(
```
eskultety: nitpick, we could also try `redirect_to_file` for name. Just an idea...

hack/mock-unittest-data/gomod.py
a-ovchinnikov: I don't want to introduce another alias for an entity in the middle of a module. I'd rather kept the names same.

hack/mock-unittest-data/gomod.py
a-ovchinnikov: Ok, since `redirect_to_file` requires `what` in the same sense as `run_with_stdout_redirect`

hack/mock-unittest-data/gomod.py
eskultety: It's not an alias, you'd literally call the function that is aliased to `run_with_stdout_redirect` in main which you're passing by reference here.

hack/mock-unittest-data/gomod.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def update_workspaces(redirect_to_file, pandemo_dir, mocked_data_dir):
+    with open(f"{mocked_data_dir}/workspaces/go_work.json") as f:
+        d_paths = [v['DiskPath'] for v in json.loads(f.read())['Use']]
+        paths = [Path(f"{pandemo_dir}")/p.removeprefix("./") for p in d_paths if p != '.']
+
+    for p in paths:
+        print(f"preparing per-workspace directory {p}")
+        mkdir_with_parents(f"{mocked_data_dir}/workspaces/{p}")
```
taylormadore: Should this instead be:
```suggestion
        mkdir_with_parents(f"{mocked_data_dir}/workspaces/{p.name}")
```
?

a-ovchinnikov: Good catch, thank you. It does not look like this affects existing test cases after all, but this is definitely a translation omission.

Issue #1148, opened 2025-10-10
Bump our minimum Python version to Python 3.10


eskultety:
Python 3.9 is scheduled for EOL on Oct 31st [1],[2], let's migrate onto 3.10 shortly afterwards.

[1] https://devguide.python.org/versions/
[2] https://endoflife.date/python

eskultety:
Addressed in #1167

PR #1166, opened 2025-11-03
Containerfile: Upgrade the base image to UBI 10

eskultety:
UBI 10 has been available for a while, there's little incentive for us not to upgrade from UBI 9.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/ubi@sha256:3816d303e75dec4da2d10eeb9e8651eef4393721598bea4690c607282635aa57 AS ubi
+FROM registry.access.redhat.com/ubi10@sha256:9bd3aebdfdf6eebb6a18541d838cac9e9a35d2f807aa8e36d9518978cc86371f AS ubi
 FROM mirror.gcr.io/library/golang:1.25.5-alpine AS golang
 FROM mirror.gcr.io/library/node:24.12-bookworm-slim AS node
 
@@ -11,7 +11,7 @@ RUN dnf -y install \
     --nodocs \
     git-core \
     jq \
-    python3.11 \
+    python3.12 \
     rubygem-bundler \
     rubygem-json \
     subscription-manager && \
@@ -28,14 +28,14 @@ RUN dnf -y install \
     gcc \
     # not a build dependency, but we copy the binary to the final image
     cargo \
-    python3.11-devel \
-    python3.11-pip \
-    python3.11-setuptools \
+    python3.12-devel \
+    python3.12-pip \
+    python3.12-setuptools \
     && dnf clean all
 
 # Install dependencies in a separate layer to maximize layer caching
 COPY requirements.txt .
-RUN python3.11 -m venv /venv && \
+RUN python3.12 -m venv /venv && \
     /venv/bin/pip install --upgrade pip && \
     /venv/bin/pip install -r requirements.txt --no-deps --no-cache-dir --require-hashes
 


eskultety:
Let's keep this PR for after the coming release.

slimreaper35:
When I run the container, I see we jumped from Python version `3.9` to `3.12`. What is then our minimum Python version ? (Looking at #1148)

eskultety:
> When I run the container, I see we jumped from Python version `3.9` to `3.12`. What is then our minimum Python version ? (Looking at #1148)

That's a great observation!
```
$ dnf list python3.10*
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

Last metadata expiration check: 0:00:04 ago on Mon Nov  3 15:24:43 2025.
Error: No matching Packages to list
```
UBI10 simply only ships with Python 3.12. It doesn't change anything when it comes to our Python version support claim, it would still be 3.10+, IOW it doesn't matter you ship with 3.12 as long as you remain 3.10 compatible. However, it's just going to get harder to actually prove we haven't broken compatibility. We could either run ITs twice, once for the production image and second time for a 3.10 custom python install image OR we could incorporate `pyupgrade` to the CI checks to make sure we don't include a 3.11+ change (I'm in favour of the latter).

Interestingly enough, there's no reason for us staying on UBI9 beyond the 3.10 bump because:
```
$ cat /etc/redhat-release 
Red Hat Enterprise Linux release 9.5 (Plow)
$ dnf list 'python3\.10*'
Error: No matching Packages to list
```
For some reason 3.10 was completely skipped on RHEL-9 and it was only 3.9, 3.11, or 3.12.

**What I'm going to do is to hold this PR up until after we've bumped to 3.10 (for which I already have some draft patches on my branch).**

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/ubi@sha256:dec374e05cc13ebbc0975c9f521f3db6942d27f8ccdf06b180160490eef8bdbc AS ubi
+FROM registry.access.redhat.com/ubi10@sha256:9bd3aebdfdf6eebb6a18541d838cac9e9a35d2f807aa8e36d9518978cc86371f AS ubi
```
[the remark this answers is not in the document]
eskultety: Closing as irrelevant - I took the FROM line from Red Hat's catalog verbatim.

eskultety:
> We could either run ITs twice, once for the production image and second time for a 3.10 custom python install image OR we could incorporate pyupgrade to the CI checks to make sure we don't include a 3.11+ change (I'm in favour of the latter).

@a-ovchinnikov @slimreaper35 @taylormadore so with:

- https://github.com/hermetoproject/hermeto/blob/70a097936807ab6cd493f58f33d9ab2349b1fb4d/pyproject.toml#L63
- https://github.com/hermetoproject/hermeto/blob/70a097936807ab6cd493f58f33d9ab2349b1fb4d/pyproject.toml#L107

in place that should satisfy my comment above (unless I'm wrong). Any further comments before merging?

slimreaper35:
I am a bit confused now. So we currently install `python3.11` in our Dockerfile, but based on:

> UBI10 simply only ships with Python 3.12

we would have to bump it to 3.12, or ?

I don't see what kind of role `pyupgrade` plays here. We use Python 3.10 syntax, so we should be fine with any higher minor version.

eskultety:
> I am a bit confused now. So we currently install `python3.11` in our Dockerfile, but based on:
> 
> > UBI10 simply only ships with Python 3.12
> 
> we would have to bump it to 3.12, or ?
> 

No, it's just what your image ships, you can't make it ship 3.10, same as UBI9 can't be made to ship with 3.10. The runtime has no impact on our language support promise, i.e. we must not use 3.12 syntax, functions, arguments, etc. anywhere, hence the Ruff and mypy 3.10 version settings in `pyproject.toml` which ought to guard against any accidental 3.11/3.12 stuff inclusion.
To put this into perspective, it was in fact the case even before these image updates, since developers have used Python 3.12 for development on their Fedora machines for a while. `mypy` and potentially other linters we used to have may have checked our compatibility with 3.9. Same thing here, what runtime ends up being installed in the image is irrelevant.

> I don't see what kind of role `pyupgrade` plays here. We use Python 3.10 syntax, so we should be fine with any higher minor version.

`pyupgrade` is nothing but spec compliance validator, IOW with every Python version there are some PEPs that either add, modify or remove some functionality and/or semantics. `pyupgrade` makes sure that you don't end up with invalid semantics due to some PEP or stdlib deprecations.

eskultety:
@slimreaper35 did my last comment clarify the problem domain? :)

slimreaper35:
So to summarize, we will ship our image with Python 3.12, which is already included in the UBI, while still being compatible with 3.10 - continue to run unit tests and pin the 3.10 version for mypy, ruff...

?

eskultety:
> So to summarize, we will ship our image with Python 3.12, which is already included in the UBI, while still being compatible with 3.10 - continue to run unit tests and pin the 3.10 version for mypy, ruff...
> 
> ?

Correct!

slimreaper35:
> Correct!

Ok. I see some merge conflicts and we should then [bump](https://github.com/hermetoproject/hermeto/blob/main/Dockerfile#L38) the Python version.

eskultety:
> > Correct!
> 
> Ok. I see some merge conflicts and we should then [bump](https://github.com/hermetoproject/hermeto/blob/main/Dockerfile#L38) the Python version.

Yep, it's lacking behind. On the other hand, with a downstream in the picture, maybe it's time for another "base image" discussion that can happen in the public space, even on GH.

eskultety:
End of the year is rapidly approaching with many folks taking the time to recharge, let's reserve this major base image change for 2026 and not rush it to the last releases of the year :) .

eskultety:
Merge conflicts resolved.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/ubi@sha256:3816d303e75dec4da2d10eeb9e8651eef4393721598bea4690c607282635aa57 AS ubi
+FROM registry.access.redhat.com/ubi10@sha256:9bd3aebdfdf6eebb6a18541d838cac9e9a35d2f807aa8e36d9518978cc86371f AS ubi
```
slimreaper35, line 1: Was the deletion of "/ubi" intentional?

eskultety: Not really. I just grabbed it from https://catalog.redhat.com/en/software/containers/ubi10/66f2a8d39bdbdc1b74e81460#get-this-image . Weirdly enough going through the catalog, there's:

- https://catalog.redhat.com/en/software/containers/ubi10/ubi/66f2b46b122803e4937d11ae#get-this-image
- https://catalog.redhat.com/en/software/containers/ubi9/ubi/615bcf606feffc5384e8452e#get-this-image
- https://catalog.redhat.com/en/software/containers/ubi9/618326f8c0d15aff4912fe0b#get-this-image

as distinct entries, so that's where this a bit surprising change comes from.

PR #1170, opened 2025-11-06
design: Adding proxy support to Hermeto

a-ovchinnikov:
The added document outlines the scope of problem for adding proxy support to Hermeto.

### DIFF
--- a/docs/design/proxy-support.md
+++ b/docs/design/proxy-support.md
@@ -1,0 +1,228 @@
+# Supporting package managers proxies in Hermeto
+
+## Background
+
+It might be desirable to pull dependencies through a proxy for a variety of
+reasons (builds reproducibility, additional verification, fetch speedup,
+etc). This means that Hermeto has to support proxying packages for
+individual package managers.
+
+## Proxy usage overview
+
+Most supported package managers (with a notable exception of Yarn v1) make
+use of proxy URL and support standard mechanisms for authentication.  Proxy
+usage is transparent to users as long as they provide correct proxy
+address and authentication credentials. Given that a proxy is correctly
+configured a user would just receive their dependencies as usual with an
+underlying package manager handling everything seamlessly.  Thus a way to
+consume proxy URL and credentials from the environment must be added to
+Hermeto.
+
+A proxy might require authentication or not.  Also note that a proxy might only be
+set up for a subset of the supported package managers.  This means that each
+individual package manager will need to handle checking the proxy setting and
+authentication setting. Once these settings are known they must be made
+available to the underlying tool used for fetching the actual dependencies
+either via environment or via configuration options.
+
+
+## Additional considerations
+
+### SBOM enhancement
+
+SBOMs must be marked to indicate whether a proxy was used or not. Currently the
+origin of an artifact is not recorded in an SBOM at all, it is assumed that
+download location is the same as the one reported in the PURL. With addition of
+proxies support this won't be true anymore. Both supported SBOM formats provide
+means for specifying a component's location:
+
+  1. CyclondeDX provides [externalReferences][];
+  1. SPDX provides [PackageDownloadLocation][] and [PackageSourceInfo][] fields.
+
+The most straightforward solution would be to include these fields into SBOMs.
+Note, that PackageDownloadLocation allows `None` as value which could be used
+to indicate that a package was downloaded from a source referenced in a PURL and
+externalReferences, being an array, provides a natural way of indicating this
+with an empty array. The downside of this solution is that some package
+managers support specifying multiple proxies in the environment and not
+reporting which specific proxy was used which means that all proxies must be
+reported.  This is not the best possible outcome, but in practice would result
+in verification of several proxies instead of one.  Unfortunately
+[PackageDownloadLocation][] does not seem to support multiple locations which
+is necessary for storing multiple proxy URLs. [PackageSourceInfo][] is a free-form
+field which can be omitted, it could be used as a marker for proxy usage: if absent
+then no proxy was used, if present it would contain a semicolon-separated list
+of proxy URLs. An alternative to this is to add a custom property to SBOMs:
+
+```
+   proxy: None | [proxy_url]
+```
+
+The extra field will be added to each package/component of an SBOM.
+The extra field will become a property for CycloneDX SBOMs and an annotation
+for SPDX SBOMs. Having this property will allow any SBOM processing tool make
+policy decisions basing on recorded proxy usage: since a proxy could be
+opinionated the mere fact that a proxy was used and the sources came
+from it would be an indicator that any opinion forming logic on the proxy
+side has approved the use of these particular sources. This is less
+important in the case of a simple caching proxy, but nevertheless
+adds a clear record that a proxy was used (which might become important
+when an anomalous build is investigated). Finally, in the rare case when
+a package was pulled from a registry, but remained in cache this
+property would unambiguously indicate where the sources came from.
+The downside of this approach is that the property is custom and will not
+be universally understood by third party tools which could be used to
+process SBOMs.
+
+Basing on the information provided above it is recommended to use
+[PackageSourceInfo][] for SPDX and [externalReferences][] for CycloneDX.
+
+
+### Fallback option
+
+It might be desired to try and retrieve packages from the original source if
+proxy fails for any reason. In other words, with fall-back enabled Hermeto
+would try a proxy first, and if the proxy is non-responsive it would then try
+fetching the sources as usual, as if no proxy settings were present. With
+fall-back disabled Hermeto would fail a fetch if a proxy is not responding. To
+achieve that the following option is proposed for addition to Hermeto config:
+
+```
+    fall-back-to-standard-source = False
+```
+
+It will be called `HERMETO_FALL_BACK_TO_STANDARD_SOURCE` in the environment.
+The default value is proposed to be set to False. Fall-back should not happen
+if there was an authorization error because it would mask it otherwise.
+
+
+## Implementation
+
+All configuration is proposed to be done via environment variables and
+Hermeto config as a fall-back option. All new environment variables
+will be prefixed with `HERMETO_` to disambiguate them from other
+variables with same names. In case a variable is meant to be consumed by
+a package manager then a secondary prefix is added (`PIP_` or `GOMOD_`).
+The resulting variables would look like the following:
+`HERMETO_PIP_PROXY_URL`, `HERMETO_GOMOD_PROXY_AUTH_TOKEN` etc. The values
+associated with these variables could be the same. This would allow for
+fine-grained control of proxy settings for each individual package manager.
+
+Every package manager will receive an instance of a Proxy subclass.
+The class will be responsible for checking the override option, for
+reading and verifying proxy URL from environment and for reading and
+verifying credentials for the package manager:
+
+```python
+    class Proxy:
+        proxy_url = URL | None  # Must be defined if username is defined
+        username  = str | None  # Must be defined if password xor token is defined
+        password  = str | None  # Must not be defined if token is defined
+        token     = str | None  # Must not be defined if password is defined
+
+        @abstractmethod
+        def make_environment_extension(self) -> dict[varname, varvalue]
+
+        @abstractmethod
+        @classmethod
+        def from_env(Proxy) -> Proxy
+```
+
+Each package manager would subclass Proxy and extend it with a translation
+table from field names to variable names which are understood by each
+individual package manager. A subclass must resolve proxy URL variable and
+authentication variables named according to scheme described above to names
+used by its underlying native tool.  These names will be used to instantiate
+the class from environment and later to populate the environment extension
+dictionary. The dictionary will be used by the class' client to populate
+environment of a process that would run the individual package manager.
+
+```python
+    class FooProxy(Proxy):
+        variable_conversion_table = (("HERMETO_FOO_PROXY_URL", "FOO_PROXY), ... )
+        ...
+```
+
+None of the fields is mandatory, the class provides one method to create
+environment extension dictionary and one class method for consuming values from
+environment. If a field is None then it would not be added to an extension
+dictionary. It is possible to receive an empty extension dictionary when the
+variables are not set or when there is an override taking place thus it will be
+always safe to extend environment with this dictionary. This would require
+making sure that every package manager that relies on native tools always
+passes environment to subprocesses. Any package manager that re-implements some
+or all aspects of native tools would need to process Proxy properties on its
+own.
+
+
+## Appendix A. Future enhancements: handling of non-registry dependencies
+
+Note, that the text below explores rather rare corner cases and having
+this functionality is not necessary for the majority of packages out there.
+
+In some cases a project might have a dependency that is not sourced from a standard
+registry for that ecosystem (it could be a git dependency or a dependency
+shared via a non-official registry). It might be desirable to  provide a push back
+mechanism for such cases, for instance when the goal of using a proxy is to
+ensure build reproducibility by storing all intermediate artefacts. Such push back
+should be implemented on individual package managers level: it is up to an individual
+package manager to distinguish between different types of dependencies and to make
+appropriate choices. The following discussion will assume that there is a need for
+a storing cache since this is the most complex scenario and everything else could
+be reduced to it.
+
+To store anything in a cache one would need a scheme for naming entities, a way to
+store the entities and a mechanism for making decisions of whether to try and retrieve
+an entity from a cache or to push it first.
+
+It is proposed to identify non-registry entities by purl and to store them as archived
+blobs. This way any non-registry entity entry in a corresponding lock file would be
+converted to a purl first, then this purl would be used to query a cache pointed to
+via an environment variable. In case there is a cache miss the dependency would be
+downloaded from the source present in the lock file, archived and pushed to the cache,
+then retrieved from the cache again. This double action might be necessary if the
+cache is to provide some additional source-processing service like malware scanning
+and should be controlled with a pair of environment switches, `HERMETO_REPUBLISH_SOURCES_TO_CACHE`,
+and `HERMETO_PULL_BACK_CUSTOM_SOURCES`,
+with default values of `False`. These variable will have to be set to `True` alongside with
+`HERMETO_FALL_BACK_TO_STANDARD_SOURCE` to activate the double-action mechanism.
+
+The possible combinations of the variables and their effects are presented in the table below:
+
+```
+[A] HERMETO_FALL_BACK_TO_STANDARD_SOURCE
+[B] HERMETO_REPUBLISH_SOURCES_TO_CACHE
+[C] HERMETO_PULL_BACK_CUSTOM_SOURCES
++-------+-------+-------+--------------------------------------------------------------+
+|   A   |   B   |   C   | Effect                                                       |
++-------+-------+-------+--------------------------------------------------------------+
+| True  | True  | True  | Double-action: sources are downloaded from regular location, |
+|       |       |       | then pushed to a cache, then fetched from the cache.         |
++-------+-------+-------+--------------------------------------------------------------+
+| True  | True  | False | Simple cache filling, for systems that do not rely on any    |
+|       |       |       | cache-side processing.                                       |
++-------+-------+-------+--------------------------------------------------------------+
+| True  | False | False | Cache circumvention for custom sources.                      |
++-------+-------+-------+--------------------------------------------------------------+
+| True  | False | True  | Forbidden combination.                                       |
++-------+-------+-------+--------------------------------------------------------------+
+| False | Any   | Any   | Effectively blocks custom sources from being used.           |
++-------+-------+-------+--------------------------------------------------------------+
+```
+
+Implementation-wise that would require decorating corresponding download() methods with
+a handler that would do the conversion and dispatch the request to other methods if needed
+basing on the state of the environment.
+
+Finally, it might be desirable to exclude certain dependencies from being archived
+(for example, because they originate from private repositories). For that case
+it is proposed to add  `HERMETO_EXCLUDE_FROM_CACHING` variable containing a semicolon-
+separated list of regular expressions which should match package names to exclude.
+Any package that matches such regexp will always be downloaded directly without
+accessing a cache. Such package will also be always marked accordingly in an SBOM
+which would help ensuring that this mechanism is not abused.
+
+
+[externalReferences]: https://cyclonedx.org/docs/1.6/json/#components_items_externalReferences
+[PackageDownloadLocation]: https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field
+[PackageSourceInfo]: https://spdx.github.io/spdx-spec/v2.3/package-information/#712-source-information-field


docs/design/proxy-support.md
```diff
@@ -0,0 +1,151 @@
+# Supporting package managers proxies in Hermeto
+
+## Background
+
+It might be desirable to pipe dependencies through a proxy for a variety of
```
eskultety: nitpick: `s/pipe/pull`

docs/design/proxy-support.md
```diff
@@ -0,0 +1,151 @@
...
+
+
+## Additional considerations
+
+### SBOM enhancement
+
+SBOM must be marked to indicate whether a proxy was used or not:
+
+```
+   used_proxy: None | proxy_url
```
eskultety: nitpick: maybe just `proxy`

docs/design/proxy-support.md
```diff
@@ -0,0 +1,151 @@
...
+and authentication setting. Once these settings are known they must be
+made available to the underlying tool used for fetching the actual
+dependencies either via environment or via a configuration option.
+
+
+## Additional considerations
+
+### SBOM enhancement
+
+SBOM must be marked to indicate whether a proxy was used or not:
```
eskultety: The way I understand the intended use case combined with a bit of insider knowledge which may or may not be put into practice IIUC we're talking about 2 types of proxies:

1. caching mirrors -> speedups
2. opinionated caching proxies -> reproducible builds, scanning, etc.

We can leave type 1 out of the picture, that one is straight forward. Type 2 proxy can very well introduce a policy to cache artifacts "indefinitely" to accentuate the importance of reproducible builds. Now, it's fine we record usage of a proxy in the SBOM, although I don't have a strong opinion on that and I could imagine not doing it, especially with caching mirrors, but I'd very much like to get educated on why the claim is that _we must record it_, I'm genuinely interested and the response should ideally be recorded in the document's paragraph above.

Now, back to caching indefinitely, if a package got removed from the origin, the PURL is automatically invalid because the package can no longer be located, because the default registry removed it. Except that it CAN be retrieved from the proxy's cache. My question is - shouldn't we describe this scenario in this doc and suggest a way forward, e.g. recording the cache's <ecosystem> registry URL in the `repository_url` PURL qualifier and (optionally) leaving out the `proxy` SBOM property completely in such case?

docs/design/proxy-support.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+### Fallback option
+
+It might be desired to try and retrieve packages from the standard source if
+proxy fails for any reason:
+
+```
+    fall-back-to-standard-source = False
+```
```
eskultety: I'm failing to see the need for ^these. IIUC the implementation suggestion is to query proxy-related configuration information from environment variables rather than through the input JSON, although the doc is pretty vague on this as it doesn't really lay out how many env variables we'd use, what's the suggested naming scheme (e.g. IIUC these would have to be prefixed with the respective backend). Let's just say we have **some** environment variables we recognize and consume their values like this:
```
HERMETO_PROXY_FOO_URL=https://foo.proxy.org
HERMETO_PROXY_FOO_TOKEN=fsjfiu823jksdfjo8rjfdsa5
hermeto fetch-deps gomod
```
What is preventing me from doing this instead:
```
HERMETO_PROXY_FOO_TOKEN=fsjfiu823jksdfjo8rjfdsa5
hermeto fetch-deps gomod
```
Doesn't ^this achieve the exact same thing as `ignore-proxy-settings = False`? I mean I guess I could theoretically see the value of it being syntactic sugar for developers to quickly switch between different debug sessions, but from user's perspective NOT specifying the proxy info should be the preferred go-to approach because it's intuitive whereas specifying the info and then disabling it in a config file or a different variable is IMO unintuitive, confusing and also error prone.

docs/design/proxy-support.md
```diff
@@ -0,0 +1,151 @@
...
+        @classmethod
+        def from_env(Proxy) -> Proxy
+```
+
+Each package manager would subclass Proxy and extend it with a translation
+table from field names to variable names which are understood by each
+individual package manager. A subclass must provide the same name for proxy URL
+variable and authentication variables as used by its underlying native tool.
+These names will be used to instantiate the class from environment and later to
+populate the environment extension dictionary.
```
eskultety: I think this would be the appropriate section to describe HOW we're going to consume the individual proxy configuration settings, since you mentioned these can be per-backend, so it's not like there's always going to be a single proxy instance for all backends set up.

docs/design/proxy-support.md
```diff
@@ -0,0 +1,151 @@
...
+an entity from a cache or to push it first.
+
+It is proposed to identify non-registry entities by purl and to store them as archived
+blobs. This way any non-registry entity entry in a corresponding lock file would be
+converted to a purl first, then this purl would be used to query a cache pointed to
+via an environment variable. In case there is a cache miss the dependency would be
+downloaded from the source present in the lock file, archived and pushed to the cache,
+then retrieved from the cache again (this double action might be necessary if the
+cache is to provide some additional source-processing service like malware scanning
+and generally could be made optional with an environment switch).
```
eskultety: I think this is a very important piece of architecture decision, so might as well highlight it more. Also

> then retrieved from the cache again
> ... and generally could be made optional with an environment switch

disabled (i.e. no double pull) should be the default

docs/design/proxy-support.md
```diff
@@ -0,0 +1,151 @@
...
+
+Implementation-wise that would require decorating corresponding download() methods with
+a handler that would do the conversion and dispatch the request to other methods if needed
+basing on the state of the environment.
+
+Finally, it might be desirable to exclude certain dependencies from being archived
+(for example, because they originate from private repositories). For that case
+it is proposed to add  `HERMETO_EXCLUDE_FROM_CACHING` variable containing a semicolon-
```
eskultety, line 219: This is where I'm confused. On one hand environment variables are suggested as a means of consuming proxy settings (which is totally understandable since this would likely be enforced on the cluster level rather than expecting users to mess with it via the input JSON), however in the same sentence the suggestion is to use a list of regexes to match against packages to be excluded from proxy handling which is something that only the users know and cannot be IMO reliably applied on cluster level unless the only expectation is to put something like `*.rpm` in there to exclude a whole ecosystem, but then I don't know how other ecosystem's packages distributed as gziped tarballs could be properly excluded.

docs/design/proxy-support.md
a-ovchinnikov: To the first question, in short: this records assumptions made by users during prefetch and could be useful when someone investigates build anomalies. A more elaborate answer could be found below:
 
We must record the use of proxy for the sake of any tools that could process this SBOM  (especially if a proxy were of opinionated variety): this is an indication that the sources came from a different location than originally expected. In the positive light an opinionated proxy could have scanned the sources for malware and in this case having an indication that a proxy was used could be interpreted as an indicator of a successful verification. In a negative light a supposedly transparent proxy could become a victim of a sophisticated attack to make it augment sources. In any case this is something that a tool that relies on Hermeto's SBOMs should decide. I think that recording as much context as possible is a good idea: it is easier to cast away excessive data than to reconstruct events basing on partial records. I'll add this reasoning to the document.

To your second question I'd say no: the described situation _should be_ rare. A proxy should not (in theory and in most practical cases) modify sources, and thus the results of prefetch should remain the same proxy or no proxy, so I would rather have keept PURL intact and had a separate indicator for proxy usage. If a repository goes down, but we still use the same cache we would never know that it went down unless we try short-circuiting a cache. Even if this happens a cache (when used) should be considered the source of truth for entities that would have otherwise be retrieved by their PURL.

docs/design/proxy-support.md
a-ovchinnikov: Thank you for pointing out the omission of naming scheme, I'll add it to the document. As for the `ignore-proxy` variable/config element it is a mechanism for temporary switching off proxying. With a token being reused for that it would require us to always retry upstream sources on 403. This could result in someone accidentally breaking their token and not recognizing it until the problem resurfaces much later in a pipeline.

docs/design/proxy-support.md
a-ovchinnikov: My idea is to consume from environment and to delegate this to `from_env()` factory method. I'll elaborate on this a bit more in the next revision.

docs/design/proxy-support.md
a-ovchinnikov: That won't be applied on cluster level, these exclusions (if they are ever needed) will be applied by users on task level. This is solely to ensure that Hermeto continues working the same way for those who prefetch for projects depending on both public and non-public sources. At the moment I am not sure such users exist, but I cannot rule out their existence and that is why I added this mechanism.

docs/design/proxy-support.md
taylormadore: IMO the primary PURL should continue to be the canonical registry URL where the package originated. `repository_url` should  be reserved for cases where the package does not originate from the default registry for the given package type

eskultety: > IMO the primary PURL should continue to be the canonical registry URL where the package originated. `repository_url` should be reserved for cases where the package does not originate from the default registry for the given package type

Yes, absolutely! What I'm arguing though is the scenario where the default registry no longer contains the package and we're solely consuming it from the opinionated caching proxy which may be caching artifacts indefinitely.

docs/design/proxy-support.md
```diff
@@ -0,0 +1,201 @@
...
+
+
+## Additional considerations
+
+### SBOM enhancement
+
+SBOM must be marked to indicate whether a proxy was used or not:
+
+```
+   proxy: None | proxy_url
```
eskultety: A recent observation, we may need to make this a **list** of URLs if this is directly to translate to individual backends, see `GOPROXY` in https://go.dev/ref/mod#environment-variables, it can be a comma separated list of proxies and unfortunately for us you have pretty much no way of telling which one was used unless you sniff the network.

a-ovchinnikov: Yes, this is the best we can provide in such case. In practice this feature is sometimes used to implement a poor man's retry: one would specify the same proxy several times in hopes that if there is a transient error during the first few attempts then it would autoresolve by the time the end of list is reached. I doubt that specifying multiple different proxies is a widespread practice. Even if this happens then those who would need to investigate a build would need to study several proxies instead of not knowing of their existence.

docs/design/proxy-support.md
```diff
@@ -0,0 +1,201 @@
...
+some package managers and do not exist for others. This means that each
+individual package manager will need to handle checking the proxy setting
+and authentication setting. Once these settings are known they must be
+made available to the underlying tool used for fetching the actual
+dependencies either via environment or via a configuration option.
+
+
+## Additional considerations
+
+### SBOM enhancement
```
taylormadore, line 31: I'll add here what we spoke about offline:

I was considering whether an [externalReference](https://cyclonedx.org/docs/1.6/json/#components_items_externalReferences) associated with the component made more sense than a custom property for CycloneDX and whether [PackageDownloadLocation](https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field) might be suitable for SPDX

a-ovchinnikov: Yes, it looks like both options could be used for storing this data. It is not clear to me at the moment how we would represent the fact that no proxy was used? Would it be an absence of the field or would it be always present and a comparison between the field and PURL required to determine that? I don't think it is a good idea to have a volatile field -- it feels easier to miss it when processing when only some SBOMs contain such a field, on the other hand if the field is always there then there is an extra indirect step for comparing it to another field (which, in turn, may need to be parsed). Having a custom property resolves both problems: the new field will be always there and it will immediately answer whether a proxy was used or not. Further, being a custom field it will never interfere with any future changes to SBOM that might be needed.

taylormadore: When I was originally thinking about it, the idea is that we would always provide these fields and they would always point to the actual download location (compared to the PURL, which should be the canonical download location). Any policy we would construct around this would check whether the actual download location was from an approved source.

I was trying to find a way to generalize the concept of using a proxy in an SBOM. Policy doesn't necessarily care that a proxy was used--it cares where the content was sourced from. Our current gap is that we only report the canonical origin.

a-ovchinnikov: Yes, that's a good point, it is of wider scope, but I think it is fine to have it settled here. I'll update the design accordingly.

eskultety: > I was considering whether an [externalReference](https://cyclonedx.org/docs/1.6/json/#components_items_externalReferences)

Curious, which type were you thinking of? `distribution`? https://cyclonedx.org/docs/1.6/json/#components_items_externalReferences_items_type

eskultety: > When I was originally thinking about it, the idea is that we would always provide these fields and they would always point to the actual download location 

Not objecting per se, but isn't the semantics just a tiny bit forward-incompatible? IOW previously we didn't have that field which definitely meant that no proxy was used. Now we're going to always have a field to indicate whether a proxy was or wasn't used, so not only you need to look at the presence of the field, but also branch on its value. Full forward compatibility would IMO be: no field -> no proxy, field -> proxy...no further branching and proper SBOM support (old and new) - this is especially important IMO since we don't version our SBOM content, fields just magically appear with time.

a-ovchinnikov: I am having trouble with `no field -> no proxy` approach: suppose someone for some reason somehow used a proxy before introduction of this feature. Once their SBOM is processed after introduction of this feature it would be incorrectly marked as "no proxy". I afraid that to be precise we need three states: "proxy was used", "proxy was not used" and "unknown" with "unknown" reserved to historical SBOMs and considered a policy violation for anything built with feature enabled. Also maybe we should start versioning SBOMs?

eskultety: > suppose someone for some reason somehow used a proxy before introduction of this feature.

You mean like consuming it with the default `GOPROXY` variable and such? The proxy would have to be recorded to the individual backend lockfiles for it to be implicit. How else would anyone have used the proxy if we don't support connecting yet? IIRC for every backend we carefully evaluate all the configuration options before we implement support so that we override any "non-official" sources of content, so which case am I not thinking of when you say they somehow could have used a proxy before this feature?

a-ovchinnikov: My point is about being explicit about what we known and what we don't know. With three states we can do it, with an implicit "no proxy" we merge two of the states. I do not find my hypothetical scenario likely and am not sure that it is possible without serious abuse of prefetch environment, however I cannot rule it out as impossible and do not want to hide this additional bit of data.

taylormadore: I'm trying to keep these two ideas separate if possible:
- Reporting actual download location
- Reporting "we used a proxy"

For old SBOMs, it isn't that a proxy was or was not used--it's that we weren't reporting the actual download location. We're merely enriching the SBOM with new information, which should be backwards compatible. For the Konflux use-case, do we care about old SBOMs? The policy written around this would have an effective date **after** all of the "not reporting actual download location" versions of the prefetch-dependencies task have already aged out and can no longer be used for release. Historical SBOMs wouldn't be evaluated against the new policy.

docs/design/proxy-support.md
```diff
@@ -0,0 +1,222 @@
...
+Most supported package managers (with a notable exception of Yarn v1) make
+use of proxy URL and support standard mechanisms for authentication.  Proxy
+usage is transparent to users as long as they provide correct proxy
+address and authentication credentials. Given that a proxy is correctly
+configured a user would just receive their dependencies as usual with an
+underlying package manager handling everything seamlessly.  Thus a way to
+consume proxy URL and credentials from the environment must be added to
+Hermeto.
+
+A proxy might require authentication or not.  A proxy might exist for just
```
eskultety: nitpick:
`s/A proxy might exist for just some package managers and do not exist for others/Note that a proxy might only be set up for a subset of the supported package managers.`

docs/design/proxy-support.md
```diff
@@ -0,0 +1,222 @@
...
+some package managers and do not exist for others. This means that each
+individual package manager will need to handle checking the proxy setting
+and authentication setting. Once these settings are known they must be
+made available to the underlying tool used for fetching the actual
+dependencies either via environment or via a configuration option.
```
eskultety: nitpick: s/a configuration option/configuration options.

docs/design/proxy-support.md
```diff
@@ -0,0 +1,222 @@
...
+
+
+## Additional considerations
+
+### SBOM enhancement
+
+SBOMs must be marked to indicate whether a proxy was used or not. Currently the
+origin of an artifact is not recorded in a SBOM at all, it is assumed that
```
eskultety: nitpick:

> origin of an artifact is not recorded in a SBOM at all

The way I understand origin is the actual source of the content which is known and already recorded in the PURL. I think what you mean would be "download location" which could be a caching proxy in this case?

also `s/a SBOM/an SBOM`

docs/design/proxy-support.md
```diff
@@ -0,0 +1,222 @@
...
+download location is the same as one reported in PURL. With addition of
```
eskultety: nitpick: `s/as one/as the one`, `s/in PURL/in the PURL`

docs/design/proxy-support.md
```diff
@@ -0,0 +1,222 @@
...
+proxies support this won't be true anymore. Both supported SBOM formats provide
+means for specifying a component's location: CyclondeDX provides
```
eskultety: nitpick: make this a formatted 2-item unordered list instead for slightly better readability :)

docs/design/proxy-support.md
```diff
@@ -0,0 +1,222 @@
...
+[externalReferences][] field for that, and SPDX provides
+[PackageDownloadLocation][]. The most straightforward solution would be to
+include these fields into SBOMs. The downside of this solution is that some
+package managers support specifying multiple proxies in the environment and not
+reporting which specific proxy was used which means that all proxies must be reported.
+This is not the best possible outcome, but in practice would result in
+verification of several proxies instead of one.  Unfortunately
+[PackageDownloadLocation][] does not seem to support multiple locations which
```
eskultety: The fact that it natively doesn't support it is not a show-stopper IMO because at the end of the day it's just a string, same as `GOPROXY` . We can "just happen" to record multiple urls into the field and choose a delimiter ourselves.

docs/design/proxy-support.md
eskultety: > We're merely enriching the SBOM with new information, which should be backwards compatible.

I argued forward compatibility, backward compatibility is fine as far as we're concerned, so no discussion there :). Anyway, not important to me, consider that resolved from my POV, let's focus on the proper field selection.

> For the Konflux use-case, do we care about old SBOMs? ... Historical SBOMs wouldn't be evaluated against the new policy.

No, but Konflux may not be the only use case, so tayloring the solution based on that project's needs might not be ideal if we can find a way of achieving a similar solution that would allow other consumers who don't have as  strict SBOM policy to run the same logic even on old SBOMs for some reason.

> 
> I'm trying to keep these two ideas separate if possible:
> 
>     Reporting actual download location
>     Reporting "we used a proxy"
> 

IIUC the former should be achieved or somehow derivable from PURL and we need a solution for the latter. Having dug deeper into `PackageDownloadLocation` I'm not 100% sure or convinced it's the best field to record the latter, a separate comment on that below.

eskultety: Originally `PackageDownloadLocation` seemed like the obvious choice to me and I was generally supportive of it, but now that you've concluded that you can't record several locations I read the definition again and I have to say the spec is IMO extremely vague on the field's meaning - is the field supposed to duplicate what PURL is already supposed to record, i.e. the actual authoritative origin location? If so, then it's not a good fit. Is our proxy intent to be authoritative? I don't think so.

Additionally, I also noticed that the field's cardinality is 1..1 which does conform to @taylormadore's intent to always record the field, but then it IMO becomes confusing a bit, because for anyone inspecting the SBOM (including tooling) one cannot tell at first glance whether the download location was from origin or from a caching proxy **unless** it were compared with the PURL, do we want that?  I thought that we wanted to make it absolutely clear whether proxy was or wasn't used which, looking at the field, doesn't achieve for me unless I missed something obvious.

The spec is huge, but skimming through it I also found https://spdx.github.io/spdx-spec/v2.3/package-information/#712-source-information-field which is optional, so the origin always remains recorded in the PURL and if a proxy was used, this field would be populated with the respective proxies and because it's free-form, you can record as many as you want, though yes, the nature of a free-form text makes this suggestion comparable and not that much superior to a custom property.

docs/design/proxy-support.md
eskultety: Okay, so why not omitting any proxy settings from our tool's invocation then? I'm trying to figure out why maintaining a questionable syntactic sugar for testing purposes that would never be used in production is something we'd want to do? IOW I as a developer who comes in contact with many tools would never expect temporary functionality disablement to be done in persistent config, nor would I expect a dedicated env variable for that if I can just omit the respective settings from the invocation directly. Admittedly, that's my modus operandi, but I guess everyone uses a shell with history, so it's not that big of a deal IMO to just remove some options from CLI rather than add something like:
```
HERMETO_XXX_PROXY_URL=some_url \
HERMETO_XXX_PROXY_AUTH_USERNAME=username \
HERMETO_XXX_PROXY_AUTH_PASSWORD=pass \
HERMETO_XXX_PROXY_AUTH_TOKEN=token \     .... this as a replacement for the 2 above it, example only
HERMETO_IGNORE_PROXY_SETTINGS=true

hermeto fetch-deps ...
```
instead of just dropping all the settings and run with "vanilla". This goes for both the override and fallback suggestions.

docs/design/proxy-support.md
```diff
@@ -0,0 +1,222 @@
...
+environment. If a field is None then it would not be added to an extension
+dictionary. It is possible to receive an empty extension dictionary when the
+variables are not set or when there is an override taking place thus it will be
+always safe to extend environment with this dictionary. This would require
+making sure that every package manager that relies on native tools always
+passes environment to subprocesses. Any package manager that re-implements some
+or all aspects of native tools would need to process Proxy properties on its
+own.
+
+### Handling of non-registry dependencies
```
eskultety: @a-ovchinnikov Do you think we could have ^this omitted from the MVP and even in the design mention this as a possible future extension in an appendix rather than the main design doc like we did with other designs? I feel like we could update the design with fresh info on this scenario when it becomes a real thing we need to cover or deal with, how about that?

docs/design/proxy-support.md
eskultety: > intent to always record the field, but then it IMO becomes confusing a bit, because for anyone inspecting the SBOM (including tooling) one cannot tell at first glance whether the download location was from origin or from a caching proxy unless it were compared with the PURL

Correction, `NONE` is a valid value, so ^this becomes a non-issue. My question on the intended usecase for the field vs `PackageSourceInfo` remains, once we're clear on that, I'm good.

a-ovchinnikov: I have documented these considerations in the latest version of the document, I slightly prefer a custom field since it is a solution designed specifically for this problem. I will update the doc with a decision once it is made. @taylormadore, @eskultety please vote: should it be a custom field or should we repurpose a standard one?

docs/design/proxy-support.md
a-ovchinnikov: Yes, I used origin in the sense of download location, will update.

docs/design/proxy-support.md
a-ovchinnikov: Sure, will move this to an appendix.

docs/design/proxy-support.md
eskultety: I personally am not in favour of custom fields if the spec standardizes free-form text fields which is pretty much what we'd do anyway. I find it important to list ALL sources serving the content otherwise it's incomplete information, so from that perspective it's either a combination of `PackageDownloadLocation` with `PackageSourceInfo` listing all the other alternative proxies if any, or just `PackageSourceInfo` alone. I might be slightly biased towards the latter since it doesn't create the illusion of `PackageDownloadLocation` being superior or even authoritative because we simply won't know in some cases which proxy was actually used and `PackageSourceInfo` captures that IMO better. That said, I can live with both if the preference lies with `PackageDownloadLocation`. If you don't make it a custom property, consider this resolved from my POV.

docs/design/proxy-support.md
eskultety: @a-ovchinnikov I don't see this addressed either in the doc or in this thread. Can you elaborate on why having these proxy disablement/override settings are substantial?

a-ovchinnikov: Sorry, I missed this one yesterday, somehow it was not displayed in the "commits" view. I'll update this too.

a-ovchinnikov: After re-reading the comment above this setting is not about omitting a proxy, but about falling back to the original sources _in case a proxy is unavailable for any reason_. I.e. if someone values pulling everything through a proxy more than having a build (maybe because this proxy does stellar job at marking malicious components), then they would set this variable to `False`, if, however, they prefer a completed build to a guarantee of pulling through a cache then they would set it to `True`.

eskultety: My bad, now I see what you meant with the source fallback, I was dense. Maybe you just need a small clarification on what "standard source" means such that we'd respect whatever ecosystem settings user had prior overriding everything with a proxy URL, other than that I'm good with it.

When it comes to the global override, I think that is something that needs to be dealt with on the upper/management layer rather than on our side - we'd only expect proxy settings on input to use it and if a proxy shall not be used I would expect whoever driving hermeto to NOT provide those proxy settings in such case, so to me this is looks very binary kind of situation.

a-ovchinnikov:
In the most recent push I dropped the override option as excessive following the reasoning in https://github.com/hermetoproject/hermeto/pull/1170#discussion_r2606008653

docs/design/proxy-support.md
a-ovchinnikov: Thank you, I updated the design with `PackageSourceInfo` and went ahead to recommend using it since it is both standard and suits our needs.

eskultety: I'd wait for @taylormadore's input as well, then we can resolve, I'm good.

taylormadore: I'm OK with `PackageSourceInfo`, but I want to solicit one additional opinion tomorrow before I give my final ack

taylormadore: Confirmed--let's use `PackageSourceInfo`. Marking this resolved.

PR #1172, opened 2025-11-14
aiohttp: retry on network errors

taylormadore:
Despite comments to the contrary, aiohttp_retry appears to not retry any exceptions by default: https://github.com/inyutin/aiohttp_retry/issues/114

Retry on ClientPayloadError and ClientConnectionError

### DIFF
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -106,6 +106,15 @@ jobs:
     # ubuntu 24.04 and respect the YAML tag (revert the commit that added this)
     runs-on: ubuntu-24.04
     steps:
+      - name: Free up disk space
+        run: |
+          echo "Disk space before cleanup:"
+          df -h /
+          # Android consumes >12G disk space
+          sudo rm -rf /usr/local/lib/android
+          echo "Disk space after cleanup:"
+          df -h /
+
       - name: Install required packages
         run: |
           sudo apt-get update


a-ovchinnikov (approved):

hermeto/core/http_requests.py
```diff
@@ -20,7 +20,7 @@
     "read": 5,
     "connect": 5,
     "backoff_factor": 1.3,
-    "status_forcelist": (500, 502, 503, 504),
+    "status_forcelist": (429, 500, 502, 503, 504),
```
eskultety: if we want to make the following claim:
> This is part of the urllib3 default status_forcelist

to indicate some compatibility with `urllib` which the code doesn't explicitly express today then I believe we should create our forcelist based off of https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry.RETRY_AFTER_STATUS_CODES to make this fact explicit and add whatever extra we need on top of it.

hermeto/core/package_managers/general.py
```diff
@@ -114,6 +114,10 @@ async def async_download_files(
     retry_options = aiohttp_retry.JitterRetry(
         attempts=num_attempts,
         statuses=set(DEFAULT_RETRY_OPTIONS["status_forcelist"]),
+        exceptions={
```
eskultety, line 117: **commit message:**
s/an exceptions/an exception

.github/workflows/gating.yaml
```diff
@@ -106,6 +106,14 @@ jobs:
     # ubuntu 24.04 and respect the YAML tag (revert the commit that added this)
     runs-on: ubuntu-24.04
     steps:
+      - name: Free up disk space
```
eskultety, line 109: I'm missing some context in the commit message for the selection, especially in context of the PR even - were we hitting some CI failures because of this? Or was it some random observation? Due to the missing commit message I feel a bit of a disconnect.

The Android SDK cumulatively taking up to 12G sure is a nice cleanup. Do we want to remove some more cruft maybe to expand on this image cleanup? I inspected the disk usage for anything over 500M and some other good candidates for removal (from our POV) could be:
```
6619444	/usr/local/.ghcup   # Haskell
4171288	/usr/share/dotnet
3283732	/usr/share/swift
1273244	/usr/local/share/powershell
1031664	/usr/local/julia1.12.1
```

None of which I expect us to need any time soon.

.github/workflows/gating.yaml
```diff
@@ -106,6 +106,14 @@ jobs:
...
+        run: |
+          echo "Disk space before cleanup:"
+          df -h /
+          sudo rm -rf /usr/local/lib/android
```
eskultety, line 114: Maybe add a comment here that it's taking >12G of disk space

hermeto/core/http_requests.py
taylormadore: I had actually started with this originally, but I thought it was less explicit overall and felt like a comment was needed to explain it--which then felt like it defeated the benefit. I don't have a strong opinion though, so also fine with adopting this suggestion

.github/workflows/gating.yaml
taylormadore: > were we hitting some CI failures because of this? Or was it some random observation?

Yeah, the CI was failing due to lack of disk space in the runner on some of the e2e tests. I kicked off the run with this fix before I left for the day Friday and before I knew if it would resolve the issue. I'll add more context to the commit message now.

> Do we want to remove some more cruft maybe to expand on this image cleanup?

I had considered removing more unneeded tools, but I didn't want to spend too much time deleting stuff from the filesystem unless we actually needed the space (we can always delete more later). As-is, this already adds almost 3 min to the CI run, so I figured if we don't need more space right now I'd rather not spend more CI time freeing it up.

hermeto/core/http_requests.py
eskultety: > I don't have a strong opinion though, so also fine with adopting this suggestion

It would be unarguably more explicit IMO

.github/workflows/gating.yaml
eskultety: > As-is, this already adds almost 3 min to the CI run, so I figured if we don't need more space right now I'd rather not spend more CI time freeing it up.

That's fair, shaving those 12G did the job -> resolving.

hermeto/core/http_requests.py
taylormadore: After looking at this more deeply, I'm just going to drop this commit adding 429 for now. aiohttp_retry doesn't have similar special handling for Retry-After like urllib3 does. I'm also not sure we want to be unconditionally forcing retries of 413 from the RETRY_AFTER_STATUS_CODES. We can revisit and rework 429 handling for aiohttp later if it becomes an issue. Sorry for the noise.

eskultety: Okay, that makes sense, but doesn't 429 fall under ClientConnectionError? Looking at the `aiohttp_retry` setting changes, we won't deal with too many requests errors at all. We'll at least need to put that into the issue tracker.

taylormadore: I believe 429 would fall under ClientResponseError, which we don't handle generally in favor of handling specific statuses instead

taylormadore: https://github.com/hermetoproject/hermeto/issues/1177

PR #1174, opened 2025-11-18
[gomod] Change the way we query git tags to favour <refspec> instead of `git fetch --tags`

eskultety:
The commit message contains an extensive explanation of the matter. Unfortunately, the origin of this issue is private and originally pointed to a private repo resource. However, the experiments as noted in the commit message were conducted on publicly available resources.
Worth noting that I ran a test with our current image as well as one built off of this PR and
`diff -u <repo>/hermeto-old/bom.json <repo>/hermeto-new/bom.json` and it didn't return anything.

### DIFF
--- a/hermeto/core/package_managers/gomod.py
+++ b/hermeto/core/package_managers/gomod.py
@@ -1312,7 +1312,10 @@ class ModuleVersionResolver:
         repo = git.Repo(repo_path)
         commit = repo.commit(repo.rev_parse("HEAD").hexsha)
         try:
-            repo.remote().fetch(force=True, tags=True)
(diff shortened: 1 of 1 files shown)


eskultety:
I encourage all reviewers to perform their own tests of this PR to confirm/refute my experiments and observations.

eskultety:
\<joke\>and yes, this time around I spent more time recording my notes in the commit message than investigating and coding\</joke\>

hermeto/core/package_managers/gomod.py
a-ovchinnikov, line 1: **Commit message** (apart from it being a very insightful read) optional nitpick: s/whereas some repos it won't./whereas with some repos it won't./ ?

PR #1179, opened 2025-11-24
gomod: Include proxy config in SBOM 

sswguo:
Add ParsedOrigin class to parse Go module origin data from `go mod download`. The Origin field is the indicator that NO proxy was used in fetching a particular module.  And record the proxy as a property into SBOM. 

Issue: https://github.com/hermetoproject/hermeto/issues/577

### DIFF
--- a/hermeto/core/package_managers/gomod.py
+++ b/hermeto/core/package_managers/gomod.py
@@ -41,8 +41,11 @@ from hermeto.core.models.input import Mode, Request
 from hermeto.core.models.output import EnvironmentVariable, RequestOutput
 from hermeto.core.models.property_semantics import PropertySet
 from hermeto.core.models.sbom import (
+    PROXY_COMMENT,
+    PROXY_REF_TYPE,
     Annotation,
     Component,
+    ExternalReference,
     create_backend_annotation,
     spdx_now,
 )
@@ -153,6 +156,8 @@ class Module(NamedTuple):
     main: if this is the main module in the repository subpath that is being processed
     missing_hash_in_file: path (relative to repository root) to the go.sum file which should have
         had a checksum for this module but didn't
+    proxy: the list of custom proxy URLs used to fetch this module, or None if the module was
+        fetched directly from VCS or only via the canonical Go proxy (proxy.golang.org).
     """
 
     name: str
@@ -161,6 +166,7 @@ class Module(NamedTuple):
     version: str
     main: bool = False
     missing_hash_in_file: Path | None = None
+    proxy: list[str] | None = None
 
     @property
     def purl(self) -> str:
@@ -180,11 +186,19 @@ class Module(NamedTuple):
         else:
             missing_hash_in_file = frozenset()
 
+        ref_rest = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
+
+        if self.proxy:
+            refs = [ExternalReference(url=p, **ref_rest) for p in self.proxy]
+        else:
+            refs = None
+
         return Component(
             name=self.name,
             version=self.version,
             purl=self.purl,
             properties=PropertySet(missing_hash_in_file=missing_hash_in_file).to_properties(),
+            external_references=refs,
         )
 
 
@@ -556,6 +570,27 @@ def _get_module_id(module: ParsedModule) -> ModuleID:
     return name, version_or_path
 
 
+def _get_proxy_for_module(module: ParsedModule) -> list[str] | None:
+    """
+    Returns None when the module was fetched directly from VCS (origin is set), or when
+    the configured GOPROXY contains only canonical entries (proxy.golang.org, direct).
+    When custom proxies are present, returns all configured proxy URLs except 'direct',
+    including proxy.golang.org if set alongside custom proxies.
+    """
+    if module.origin:
+        return None
+
+    canonical_proxies = {"direct", "https://proxy.golang.org"}
+    proxy_config = get_config().gomod.proxy_url
+
+    proxies = [p.strip() for p in re.split(r"[|,]", proxy_config) if p.strip()]
+
+    if not any(p not in canonical_proxies for p in proxies):
+        return None
+
+    return [p for p in proxies if p != "direct"]
+
+
 def _create_modules_from_parsed_data(
     main_module: Module,
     main_module_dir: RootedPath,
@@ -573,6 +608,7 @@ def _create_modules_from_parsed_data(
         if not version_or_path.startswith("."):
             version = version_or_path
             real_path = name
+            proxy = _get_proxy_for_module(module)
 
             if mod_id not in modules_in_go_sum:
                 if go_work:
@@ -587,6 +623,7 @@ def _create_modules_from_parsed_data(
             resolved_replacement_path = main_module_dir.join_within_root(version_or_path)
             version = version_resolver.get_golang_version(module.path, resolved_replacement_path)
             real_path = _resolve_path_for_local_replacement(module)
+            proxy = None
 
         return Module(
             name=name,
@@ -594,6 +631,7 @@ def _create_modules_from_parsed_data(
             original_name=original_name,
             real_path=real_path,
             missing_hash_in_file=missing_hash_in_file,
+            proxy=proxy,
         )
 
     def _resolve_path_for_local_replacement(module: ParsedModule) -> str:
--- a/tests/unit/package_managers/test_gomod.py
+++ b/tests/unit/package_managers/test_gomod.py
@@ -22,7 +22,15 @@ from hermeto.core.errors import (
 )
 from hermeto.core.models.input import Flag, Mode, Request
 from hermeto.core.models.output import BuildConfig, EnvironmentVariable, RequestOutput
-from hermeto.core.models.sbom import Annotation, Component, Property, PropertyEnum
+from hermeto.core.models.sbom import (
+    PROXY_COMMENT,
+    PROXY_REF_TYPE,
+    Annotation,
+    Component,
+    ExternalReference,
+    Property,
+    PropertyEnum,
+)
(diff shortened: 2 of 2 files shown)


hermeto/core/package_managers/gomod.py
```diff
@@ -67,6 +67,20 @@ class _ParsedModel(pydantic.BaseModel):
     model_config = pydantic.ConfigDict(alias_generator=to_pascal, populate_by_name=True)
 
 
+class ParsedOrigin(_ParsedModel):
```
a-ovchinnikov, line 84: Could you please explain how this data is used? I might be missing something, but it looks like it is not used anywhere in this PR.

sswguo:
> Thank you for the PR! Could you please share what is the plan regarding proxy URLs, will they be handled in a follow-up PR or will the solution be postponed until some later moment?

Yeah, I am trying to make it in another PR, draft commit here: https://github.com/sswguo/hermeto/commit/3f772d518694fb9b3a00b3e06724434c5d575652  

Since the proxy config supports multiple values, so I think it would be good to make it as a property `hermeto:go:proxy:config` instead of custom qualifier `proxy` as mentioned in the issue: https://github.com/hermetoproject/hermeto/issues/577 .  Not sure if I understand correctly or if this is related with the design Erik mentioned: https://github.com/hermetoproject/hermeto/pull/1170. ; -) thanks.

eskultety:
@a-ovchinnikov @sswguo do we want to keep this one open or close it until the time comes? Alternatively, would draft be more suitable? I'm only asking in context of trying to keep the PR list limited. Works for me either way.

a-ovchinnikov:
Let's convert this to a draft for the time being and see what should be done about it once the more general proxy handling is introduced.

sswguo:
@a-ovchinnikov @eskultety can we continue this based on the design https://github.com/hermetoproject/hermeto/pull/1170 ? or this can be covered in the specific proxy support in gomod  like this https://github.com/hermetoproject/hermeto/pull/1248 ?  if the latter case, we can close this PR IIUC. thanks.

a-ovchinnikov:
> can we continue this based on the design https://github.com/hermetoproject/hermeto/pull/1170 ? or this can be covered in the specific proxy support in gomod like this https://github.com/hermetoproject/hermeto/pull/1248 ?

Let's follow the lead of https://github.com/hermetoproject/hermeto/pull/1248 . It should not change by much now and the same approach should be applied to golang.

sswguo:
> Let's follow the lead of #1248 . It should not change by much now and the same approach should be applied to golang.

I was trying to revise this PR based on that, let's see if the change works. thanks. I will continue on the integration tests failure.

hermeto/core/package_managers/gomod.py
```diff
@@ -174,11 +182,22 @@ def to_component(self) -> Component:
...
 
+        external_refs = None
+        eref_rest = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
+
+        if self.proxy:
+            external_refs = [
+                ExternalReference(url=p.strip(), **eref_rest)
+                for p in self.proxy.split(",")
+                if p.strip() and p.strip() != "direct"
+            ]
```
a-ovchinnikov: A non-blocking nitpick: I would have created a couple of helpers to ensure that the list comprehension fits one line:
```python
proxies = self.proxy.split(",")
is_indirect = lambda p: p.strip() and p.strip() != "direct"
external_refs = [ExternalReference(url=p.strip, **proxy_common) for p in proxies if is_indirect(p)]

**Review by a-ovchinnikov — COMMENTED** [review]
I find the approach reasonable.

**Comment by sswguo** [comment]
> I find the approach reasonable.

Thanks, then mark it `Ready for review`.

File: hermeto/core/package_managers/gomod.py [hunk 3]
```diff
@@ -832,11 +872,19 @@ def _create_main_module_from_parsed_data(
         # Should not happen, since the version is always resolved from the Git repo
         raise RuntimeError(f"Version was not identified for main module at {resolved_subpath}")
 
+    if parsed_main_module.origin:
+        # ParsedOrigin exists = direct VCS fetch
+        proxy = None
+    else:
+        # No ParsedOrigin = fetched via proxy
+        proxy = get_config().gomod.proxy_url
```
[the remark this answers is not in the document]
sswguo: Good catch. Updated.

hermeto/core/package_managers/gomod.py
```diff
@@ -161,6 +164,7 @@ class Module(NamedTuple):
     version: str
     main: bool = False
     missing_hash_in_file: Path | None = None
+    proxy: str | None = None
```
taylormadore: nit: We should probably add a description of this field to the docstring, for both consistency and also given that the format is not immediately apparent

hermeto/core/package_managers/gomod.py
```diff
@@ -180,11 +184,20 @@ def to_component(self) -> Component:
         else:
             missing_hash_in_file = frozenset()
 
+        refs = None
+        ref_rest = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
+
+        if self.proxy:
+            proxies = self.proxy.split(",")
```
taylormadore: See https://go.dev/ref/mod#goproxy-protocol

Pipes can also be used as a delimiter

hermeto/core/package_managers/gomod.py
```diff
@@ -180,11 +184,20 @@ def to_component(self) -> Component:
...
+            is_indirect = lambda p: p.strip() and p.strip() != "direct"
+            refs = [ExternalReference(url=p.strip(), **ref_rest) for p in proxies if is_indirect(p)]
```
taylormadore: Should we deduplicate the proxy urls we report? I remember from the past that specifying a url multiple times in GOPROXY was used to effectuate retries

tests/integration/test_data/gomod_correct_vendor_in_submodule_passes_vendor_check/bom.json
```diff
@@ -145,7 +145,14 @@
       ],
       "purl": "pkg:golang/github.com/hermetoproject/integration-tests/integration-tests@v0.0.0-20250910180432-ddc935adef13?type=module",
       "type": "library",
-      "version": "v0.0.0-20250910180432-ddc935adef13"
+      "version": "v0.0.0-20250910180432-ddc935adef13",
```
taylormadore: I'm not necessarily suggesting we change this behavior, but I did start questioning whether we should be reporting the official GOPROXY proxy_url when it is the only one used. I was reminded of the PURL spec, where you're not really supposed to qualify PURLs with extra information pointing to the canonical registry for a given package type.

eskultety: > where you're not really supposed to qualify PURLs with extra information pointing to the canonical registry for a given package type.

This is a very good argument.

sswguo: Yeah, good point, so I guess this would require either a hardcoded check or an additional configuration to handle the official goproxy URL.

sswguo: When checking the tests, if the config includes both the custom proxy and official one, like 
`https://goproxy.corp.example.com,https://proxy.golang.org,direct`, since there is no direct info to show which proxy gomod used during the fetch,  then it would be wrong if we just record the custom proxy in the SBOM, WDYT ?

taylormadore: Good question. If it's ambiguous, I'd lean towards reporting both. The main thing I would like to avoid is reporting _**all**_ gomod components with the official proxy even when we know no other non-official proxy is being used. @eskultety Do you agree?

eskultety: 100% agreement.

sswguo: Ok, that makes sense. thank you.

hermeto/core/package_managers/gomod.py
```diff
@@ -161,6 +166,7 @@ class Module(NamedTuple):
     version: str
     main: bool = False
     missing_hash_in_file: Path | None = None
+    proxy: set[str] | None = None
```
taylormadore: This is unordered, so we may not be reporting the proxy URLs in the SBOM consistently or in the order that they appear in gomod.proxy_url

hermeto/core/package_managers/gomod.py
```diff
@@ -841,11 +874,14 @@ def _create_main_module_from_parsed_data(
         # Should not happen, since the version is always resolved from the Git repo
         raise RuntimeError(f"Version was not identified for main module at {resolved_subpath}")
 
+    proxy = _get_proxy_for_module(parsed_main_module)
```
taylormadore: Should there be a proxy reported for the main module? I believe that would be the user repo, not something that is fetched via the proxy

taylormadore: Same for workspace/local-replacement dependencies. Actually now I'm wondering why we don't report vcs_url PURL qualifiers for all these cases as well. The same case could be made for direct dependencies. This appears to be a larger issue though and certainly outside the scope of this PR. Filed: https://github.com/hermetoproject/hermeto/issues/1434

sswguo: Remove the setting for main module and local-replacement if I understand those two correctly.

hermeto/core/package_managers/gomod.py
sswguo: Ah, updated that to list then.

PR #1180, opened 2025-11-24
tests: Update the base image for generic_e2e_maven tests

sswguo:
The ibmjava does not support arch `arm64`, which will cause the test failure as follows on MacOS,  and eclipse-temurin java works. 

` 
Error: creating build container: choosing an image from manifest list docker://mirror.gcr.io/ibmjava:11-jdk: no image found in manifest list for architecture "arm64", variant "v8", OS "linux"
`

### DIFF
--- a/tests/integration/test_data/generic_e2e_maven/container/Containerfile
+++ b/tests/integration/test_data/generic_e2e_maven/container/Containerfile
@@ -1,4 +1,4 @@
-FROM mirror.gcr.io/ibmjava:11-jdk
+FROM mirror.gcr.io/eclipse-temurin:11-jre-jammy
 
 # Test disabled network access
 RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi


tests/integration/test_data/generic_e2e_maven/container/Containerfile
```diff
@@ -1,4 +1,4 @@
-FROM mirror.gcr.io/ibmjava:11-jdk
+FROM mirror.gcr.io/eclipse-temurin:11-jdk
```
slimreaper35: nitpick: We could probably find a smaller image.

```suggestion
FROM mirror.gcr.io/eclipse-temurin:11-jdk-jammy
```

eskultety: @sswguo nice to see this replaced, I've had this as part of my other changes to the test images on my branch. @slimreaper35 is right though, smaller footprint is better, so please use the Alpine version of the image this is Java, so we should get away with Alpine just fine if I'm not mistaken.

sswguo: @eskultety  @slimreaper35 +1, let's use the smaller one: [11-jdk-alpine].

slimreaper35: I run the integration test and get the same error:

```
E           RuntimeError: Building image failed. Output:
E           STEP 1/4: FROM mirror.gcr.io/eclipse-temurin:11-jdk-alpine
E           Trying to pull mirror.gcr.io/eclipse-temurin:11-jdk-alpine...
E           Error: creating build container: internal error: unable to copy from source docker://mirror.gcr.io/eclipse-temurin:11-jdk-alpine: choosing an image from manifest list docker://mirror.gcr.io/eclipse-temurin:11-jdk-alpine: no image found in image index for architecture "arm64", variant "v8", OS "linux"
```

on my macOS 🙂

sswguo: Oh, my bad, I have not test that locally. So we should use: [11-jdk-jammy] which supports `linux/arm64/v8`.

sswguo: yes, 11-jdk-alpine only supports `linux/amd64`.

eskultety: Sigh, my bad for not checking the tags when I suggested Alpine. @sswguo for integration tests purposes it should be enough to use the `jre` instead of `jdk`, right?

sswguo: @eskultety yeah, I think so, which requires just the runtime one. I can test it locally, and the jre one is just around 88M.

sswguo: `11-jre-jammy` works in my MacOS. ; -)

PR #1184, opened 2025-11-25
enhancement: allow relative paths for generic lockfiles

derasdf:
Previously, the generic package manager rejected relative paths for lockfiles, requiring absolute paths which are often unknown in CI environments.

This change updates `generic` manager to resolve relative lockfile paths against the source directory.

Resolves: https://github.com/hermetoproject/hermeto/issues/761

### DIFF
--- a/docs/generic.md
+++ b/docs/generic.md
@@ -16,8 +16,9 @@ because the produced SBOM component will not be accurate.
 
 The generic fetcher requires a lockfile `artifacts.lock.yaml` that specifies
 which files to download. This file is expected to be in the source repository.
-Alternatively, it can be supplied as an absolute path via the `lockfile` key in
-the JSON input to hermeto.
+Alternatively, a different filename or location can be supplied via the
+`lockfile` key in the JSON input to hermeto. The value may be either an absolute
+path or a path relative to the package `path`.
 
 Below are sections for each type of supported artifact. Several artifacts of
 different types can be specified in a single lockfile.
@@ -50,8 +51,8 @@ where 'JSON input' is
   // path to the package (relative to the --source directory)
   // defaults to "."
   "path": ".",
-  // option to specify lockfile path, must be an absolute path if specified
-  // defaults to "artifacts.lock.yaml", relative to path
+  // option to specify lockfile path: absolute or relative to package path
+  // defaults to "artifacts.lock.yaml", resolved relative to package path
   "lockfile": "artifacts.lock.yaml",
 }
 ```
@@ -153,9 +154,11 @@ git clone -b sample-app https://github.com/cachito-testing/cachi2-generic.git
 #### Pre-fetch dependencies
 
 In order to retrieve the archive with the tool, either a `artifacts.lock.yaml`
-needs to be in the repository, or an absolute path needs to be supplied in the
-JSON input, pointing to a lockfile. You can find a sample lockfile below. It is
-identical to the one found in the [sample repository][].
+needs to be in the repository, or a lockfile path needs to be supplied in the
+JSON input (either an absolute path or a path relative to the package `path`).
+You can find
+a sample lockfile below. It is identical to the one found in the
+[sample repository][].
 A lockfile for the generic fetcher must contain a `metadata` header and a list
 of artifacts, where each artifact is represented as a pair of URL and a checksum
 string in the format of `"algorithm:checksum"`. Optionally, you can also specify
--- a/hermeto/core/package_managers/generic/main.py
+++ b/hermeto/core/package_managers/generic/main.py
@@ -47,13 +47,20 @@ def _resolve_lockfile_path(
     """
     Return the lockfile path for a package.
     """
+    if lockfile_path and lockfile_path.is_absolute():
+        return lockfile_path
+
     path = source_dir.join_within_root(package_path)
-    lockfile = lockfile_path or path.join_within_root(DEFAULT_LOCKFILE_NAME).path
-    if not lockfile.is_absolute():
+    lockfile_name = lockfile_path or DEFAULT_LOCKFILE_NAME
+    lockfile = path.join_within_root(lockfile_name).path
+
+    if not lockfile.is_relative_to(path.path):
         raise PackageRejected(
-            f"Supplied generic lockfile path '{lockfile}' is not absolute, refusing to continue.",
-            solution="Make sure the supplied path to the generic lockfile is absolute.",
+            f"Supplied generic lockfile path '{lockfile_name}' must be inside the package "
+            f"path '{package_path}'.",
+            solution="Use a lockfile path located within the package path.",
         )
+
     return lockfile
 
 
--- a/tests/unit/package_managers/test_generic.py
+++ b/tests/unit/package_managers/test_generic.py
@@ -13,6 +13,7 @@ from hermeto.core.package_managers.generic.main import (
     DEFAULT_LOCKFILE_NAME,
     _load_lockfile,
     _resolve_generic_lockfile,
+    _resolve_lockfile_path,
     fetch_generic_source,
 )
 from hermeto.core.rooted_path import PathOutsideRoot, RootedPath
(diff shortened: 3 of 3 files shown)


hermeto/core/package_managers/generic/main.py
```diff
@@ -31,7 +31,10 @@ def fetch_generic_source(request: Request) -> RequestOutput:
     components = []
     for package in request.generic_packages:
         path = request.source_dir.join_within_root(package.path)
-        lockfile = package.lockfile or path.join_within_root(DEFAULT_LOCKFILE_NAME).path
+        if package.lockfile:
+            lockfile = path.join_within_root(package.lockfile).path
+        else:
+            lockfile = path.join_within_root(DEFAULT_LOCKFILE_NAME).path
```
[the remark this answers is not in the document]
a-ovchinnikov: The suggestion is questionable,such line would be rather hard to read. A more readable alternative is to use an intermediate name:
```python
lockfile_name = package.lockfile or DEFAULT_LOCKFILE_NAME
lockfile = path.join_within_root(lockfile_name).path
```

tests/unit/package_managers/test_generic.py
```diff
@@ -142,21 +142,24 @@ def test_fetch_generic_source(
     mock_resolve_generic_lockfile.assert_called()
 
 
-def test_fetch_generic_source_relative_lockfile_path() -> None:
+@mock.patch("hermeto.core.package_managers.generic.main._resolve_generic_lockfile")
+def test_fetch_generic_source_relative_lockfile_path(
```
a-ovchinnikov: Both the original test name and the new one could be improved: they do not tell what to expect from the test. The original name does not tell us that relative lockfile must fail, the new one does not tell us that that now it has to succedd. What do you think about renaming test to something like `test_generic_source_can_be_fetched_when_relative_lock_file_is_used`? Such _test_ name would tell a reader at a glance what is about to happen in the test without the need to dive into implementation. (In the same vein a better name for the original test would have been `test_relative_path_to_lock_file_is_rejected`.)

tests/unit/package_managers/test_generic.py
```diff
@@ -142,21 +142,24 @@ def test_fetch_generic_source(
...
+    mock_resolve_generic_lockfile: mock.Mock, rooted_tmp_path: RootedPath
+) -> None:
+    mock_resolve_generic_lockfile.return_value = []
+
     model_input = GenericPackageInput.model_construct(
-        type="generic", lockfile=Path("relative.yaml")
+        type="generic", path=Path("pkg"), lockfile=Path("relative.yaml")
```
a-ovchinnikov: What happens if you don't add a path here?

derasdf: We now resolve a package’s lockfile relative to its own path (just like pip does), and the new test confirms relative lockfile names in subdirectories no longer get rejected.

a-ovchinnikov: Does this mean that this is a non-optional argument anymore?

hermeto/core/package_managers/generic/main.py
```diff
@@ -31,7 +31,8 @@ def fetch_generic_source(request: Request) -> RequestOutput:
     components = []
     for package in request.generic_packages:
         path = request.source_dir.join_within_root(package.path)
-        lockfile = package.lockfile or path.join_within_root(DEFAULT_LOCKFILE_NAME).path
+        lockfile_name = package.lockfile or DEFAULT_LOCKFILE_NAME
+        lockfile = path.join_within_root(lockfile_name).path
         if not lockfile.is_absolute():
             raise PackageRejected(
                 f"Supplied generic lockfile path '{lockfile}' is not absolute, refusing to continue.",
```
slimreaper35: I assume this exception is not needed anymore.

tests/unit/package_managers/test_generic.py
derasdf: No, it is still optional, the test was changed to check the new subdirectory case. Do you want me to add a cover for the default case too?

a-ovchinnikov: Yes, please. If an aspect of a behavior was not changed then a test for it should not be removed. You can parameterize this test with both variants and proper test ids.

eskultety:
@derasdf have you actually verified your change? :)

derasdf:
@eskultety Yes, I’ve checked it with the test and a separate dummy project before, but now after further checks i see what you mean, fixing it...

docs/generic.md
```diff
@@ -50,8 +51,8 @@ where 'JSON input' is
   // path to the package (relative to the --source directory)
   // defaults to "."
   "path": ".",
-  // option to specify lockfile path, must be an absolute path if specified
-  // defaults to "artifacts.lock.yaml", relative to path
+  // option to specify lockfile path: absolute or relative to "path"
```
eskultety: nitpick: instead of just "relative to path" (was confusing at first glance without further code context), say "relative to package path"

docs/generic.md
```diff
@@ -153,9 +154,10 @@ git clone -b sample-app https://github.com/cachito-testing/cachi2-generic.git
 #### Pre-fetch dependencies
 
 In order to retrieve the archive with the tool, either a `artifacts.lock.yaml`
-needs to be in the repository, or an absolute path needs to be supplied in the
-JSON input, pointing to a lockfile. You can find a sample lockfile below. It is
-identical to the one found in the [sample repository][].
+needs to be in the repository, or a lockfile path needs to be supplied in the
+JSON input (either an absolute path or a path relative to `path`). You can find
```
eskultety: Same comment about 'relative to path' vs "relative to package path" as above.

tests/unit/package_managers/test_generic.py
```diff
@@ -142,21 +142,50 @@ def test_fetch_generic_source(
...
+        lockfile=lockfile_value,
     )
 
     mock_request = mock.Mock()
+    mock_request.source_dir = rooted_tmp_path
     mock_request.generic_packages = [model_input]
 
-    with pytest.raises(PackageRejected) as exc_info:
-        fetch_generic_source(mock_request)
+    fetch_generic_source(mock_request)
```
eskultety: While the fix is correct and works, calling `fetch_generic_source` just to verify lockfile doesn't sound correct. A lesson from the gomod and pip backends, we should avoid calling into any `fetch_` entrypoints in unit tests since they can potentially grow beyond being manageable. If you prepend a simple commit to extract the lockfile lookup to a standalone helper, you don't need to do any mocking whatsoever. Should be a couple minute job to add another patch :) .

derasdf: Done. Let me know if this approach feels too redundant anywhere.

tests/unit/package_managers/test_generic.py
```diff
@@ -142,21 +143,17 @@ def test_fetch_generic_source(
     mock_resolve_generic_lockfile.assert_called()
 
 
-def test_fetch_generic_source_relative_lockfile_path() -> None:
-    model_input = GenericPackageInput.model_construct(
-        type="generic", lockfile=Path("relative.yaml")
+def test_lockfile_path_rejects_relative(rooted_tmp_path: RootedPath) -> None:
```
eskultety: It's a bit pointless I think to rename and modify an existing test that is going to be reworked heavily in next commit, it's not very clean. 
The reason I say that is because you're changing the nature of the test - from a negative test to a positive test in the span of 2 commits. In the other commit you just lost a negative test and so I believe the way to go about this patch knowing you'll need test changes eventually is to introduce just the helper in this commit with no test changes yet...

tests/unit/package_managers/test_generic.py
```diff
@@ -143,17 +143,39 @@ def test_fetch_generic_source(
     mock_resolve_generic_lockfile.assert_called()
 
 
-def test_lockfile_path_rejects_relative(rooted_tmp_path: RootedPath) -> None:
-    package = GenericPackageInput.model_construct(
+@pytest.mark.parametrize(
```
eskultety, line 146: ...and then in this commit you'll add this complex positive test and rework **the original negative test** such that you also test if a relative path points outside the package path. I also believe that since you're making these test adjustments we should follow some standard convention and rename the negative test to something like `test_resolve_lockfile_path` and `test_resolve_lockfile_path_fail` (or `_error`).

hermeto/core/package_managers/generic/main.py
```diff
@@ -30,17 +30,25 @@ def fetch_generic_source(request: Request) -> RequestOutput:
...
-                f"Supplied generic lockfile path '{lockfile}' is not absolute, refusing to continue.",
-                solution="Make sure the supplied path to the generic lockfile is absolute.",
-            )
-        components.extend(_resolve_generic_lockfile(lockfile, request.output_dir))
+        lockfile_path = _lockfile_path(package, request.source_dir)
+        components.extend(_resolve_generic_lockfile(lockfile_path, request.output_dir))
     return RequestOutput.from_obj_list(components=components)
 
 
+def _lockfile_path(package: GenericPackageInput, source_dir: RootedPath) -> Path:
```
eskultety: nitpick: I think something like `_resolve_lockfile_path` would be a tiny bit more descriptive.

Additionally, I don't think you need to consume the `GenericPackageInput` model instance. This is a simple helper, it should only take `source_dir` and a `Path` instance (it'll simplify the test in the next commit by not needing to construct a pydantic model at all).

hermeto/core/package_managers/generic/main.py
```diff
@@ -30,17 +30,33 @@ def fetch_generic_source(request: Request) -> RequestOutput:
...
+        )
+        components.extend(_resolve_generic_lockfile(lockfile_path, request.output_dir))
     return RequestOutput.from_obj_list(components=components)
 
 
+def _resolve_lockfile_path(
+    source_dir: RootedPath,
+    package_path: Path,
+    lockfile_path: Path | None,
+) -> Path:
```
slimreaper35, line 46: The diff is not aligned with the commit message, namely:

> and add a focused unit test without calling fetch_generic_source

tests/unit/package_managers/test_generic.py
eskultety: @derasdf I believe the

> that you also test if a relative path points outside the package path. 

and

> test_resolve_lockfile_path_fail

parts are still missing in the latest revision.

hermeto/core/package_managers/generic/main.py
slimreaper35: TBH, I expected the updated unit test in the first commit. But still the commit title is misleading.

PR #1188, opened 2025-12-01
Backport Go 1.25 support to 0.28 branch

eskultety:
This backports Go 1.25 enablement including some outstanding Go fixes we've accumulated to the 0.28 release. This PR also backports several integration test changes to be able to integration test this backport cleanly since the integration tests have been updated in place, thus directly affecting the old releases.


eskultety:
Also not the target branch name...this is the first time we're backporting to an old release's branch, so we have to establish a naming standard.

taylormadore:
> Also not the target branch name...this is the first time we're backporting to an old release's branch, so we have to establish a naming standard.

I think the most common pattern I have seen in the wild is some variation of `release-major.minor`, but anything where the intent is clear is fine by me

slimreaper35:
What is maint ?

tests/integration/test_pip.py
```diff
@@ -245,6 +245,8 @@ def test_e2e_pip(
     :param tmp_path: Temp directory for pytest
     """
     test_case = request.node.callspec.id
+    if test_case == "pip_e2e_rust_extensions":
```
a-ovchinnikov, line 250: For my education: would it not be easier to just drop the parameter? In the unlikely case of backporting Cargo support we could just add it back.

eskultety:
> What is maint ?

"maint" is a commonly used abbreviation for a maintenance branch

eskultety:
> > Also not the target branch name...this is the first time we're backporting to an old release's branch, so we have to establish a naming standard.
> 
> I think the most common pattern I have seen in the wild is some variation of `release-major.minor`, but anything where the intent is clear is fine by me

okay, no problem, I rename the branch and push again.

eskultety:
Branch renamed, PR base updated.

tests/integration/test_pip.py
eskultety: Could be either. My motivation simply was "making it work and touching original code as little as possible" which doesn't make your suggestion any less correct :) !

PR #1196, opened 2025-12-02
Backport github release workflow to the 0.28 branch

eskultety:
This backports the recently adopted GitHub release workflow which builds and pushes a container image to the GitHub registry.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -34,8 +34,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v5
-        with:
-          fetch-tags: true
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3


a-ovchinnikov (approved):

taylormadore:
Do we want to move the `latest` tag for maintenance releases?

eskultety:
> Do we want to move the `latest` tag for maintenance releases?

Definitely not (simply because it's a maintenance release, not a regular release) and I deliberately didn't do so for the 0.28.1 release.

_Edit: Well, hit the send button too soon :) . It depends on what your understanding of "latest" is and in which context you mean it. I presume you mean it in context of container image tags and not the release label on GitHub.
In simple terms the latter makes no sense :).
In case of the former, I'm aware of "latest" tag conundrum in the container world where latest can either be be literally the latest build available OR the latest stable build. I personally never was a fan of the original - latest whatever build - meaning because as a user I simply want  (most of the time) the latest stable build, I don't want to have to look at what exact versions are available and I don't care about nightlies unless I'm a developer of that project. From that POV I'd rather keep "latest" for the latest stable regular release. @taylormadore what are your thoughts on this?_

taylormadore:
> I'd rather keep "latest" for the latest stable regular release. @taylormadore what are your thoughts on this?

This was exactly what I was thinking :100:

PR #1200, opened 2025-12-04
fix: prefer strongest hash when parsing SRI strings

derasdf:
When npm/Yarn classic (v1) integrity fields include multiple hashes, pick the strongest (sha512 > sha384 > sha256 > others) instead of blindly taking the first entry.

Note: Yarn v2+ uses a different integrity format and does not use this helper.

### DIFF
--- a/hermeto/core/checksum.py
+++ b/hermeto/core/checksum.py
@@ -1,4 +1,5 @@
 import base64
+import binascii
 import hashlib
 import logging
 from collections import defaultdict
@@ -26,7 +27,8 @@ class ChecksumInfo(NamedTuple):
 
         https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
 
-        Note: npm and yarn use this format in their lockfiles.
+        Note: npm and Yarn classic (v1) use this format in their lockfiles; newer Yarn
+        versions use a different integrity representation that does not go through this helper.
         """
         bytes_sha = bytes.fromhex(self.hexdigest)
         base64_sha = base64.b64encode(bytes_sha).decode("utf-8")
@@ -37,9 +39,45 @@ class ChecksumInfo(NamedTuple):
 
     @classmethod
     def from_sri(cls, sri: str) -> "ChecksumInfo":
-        """Convert the input Subresource Integrity value to ChecksumInfo."""
-        algorithm, checksum = sri.split("-", 1)
-        return ChecksumInfo(algorithm, base64.b64decode(checksum).hex())
+        """Convert the input Subresource Integrity value to ChecksumInfo.
+
+        The integrity string may contain multiple hashes separated by whitespace according to the
+        SRI specification [https://www.w3.org/TR/sri-2/].
+
+        When multiple hashes are present, we pick the strongest algorithm
+        (sha512 > sha384 > sha256 > others). Note that while the SRI specification only mentions
+        SHA-2 algorithms (SHA-256, SHA-384, SHA-512), in practice "others" can include legacy
+        algorithms like sha1, which may appear when Git commit hashes are referenced in
+        integrity fields.
+        """
+
+        sri_hash_priorities = {"sha512": 3, "sha384": 2, "sha256": 1}
+        best = None
+
+        for part in sri.split():
+            if "-" not in part:
+                continue
+
+            alg, val = part.strip().split("-", 1)
+            alg, val = alg.lower(), val.strip()
+
+            if not alg or not val:
+                continue
+
+            # (priority, algorithm, value)
+            current = (sri_hash_priorities.get(alg, 0), alg, val)
+            if best is None or current > best:
+                best = current
+
+        try:
+            # If best is None, unpacking will raise TypeError
+            _, alg, val = best  # type: ignore[misc]
+            return ChecksumInfo(alg, base64.b64decode(val).hex())
+        except (TypeError, binascii.Error) as e:
+            raise PackageRejected(
+                "Integrity value is empty or malformed",
+                solution="Please check the integrity value in your lockfile.",
+            ) from e
 
     @classmethod
     def from_hash(cls, h: str) -> "ChecksumInfo":
--- a/tests/unit/test_checksum.py
+++ b/tests/unit/test_checksum.py
@@ -11,6 +11,11 @@ FILE_CONTENT = "Beetlejuice! Beetlejuice! Beetlejuice!"
 SHA512 = "da518fe8b800b3325fe35ca680085fe37626414d0916937a01a25ef8f5d7aa769b7233073235fce85eec717e02bb9d72062656cf2d79223792a784910c267b54"
 SHA256 = "ed1f8cd69bfacf0528744b6a7084f36e8841b6128de0217503e215612a0ee835"
 MD5 = "308764bc995153f7d853827a675e6731"
+SHA512_SRI = "sha512-Ml8Hhh4KuIjZBgaxB0/elW/TlU3MTG5Bjb/52KqDQrVQdIFAiDK/qsjkjzRNxlDI3w+BgsAnHtn6IzqjLDKYOQ=="
+SHA512_HEX = (
+    "325f07861e0ab888d90606b1074fde956fd3954dcc4c6e418dbff9d8aa8342b5507481408832bfaac8e48f344"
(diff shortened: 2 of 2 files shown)


hermeto/core/checksum.py
```diff
@@ -37,9 +38,37 @@ def __str__(self) -> str:
...
+        In such cases pick the strongest algorithm (sha512 > sha384 > sha256 > others).
+        """
+
+        parsed: list[tuple[str, str]] = []
+        for part in sri.split():
+            entry = part.strip()
+            if not entry or "-" not in entry:
+                continue
+
+            algorithm, checksum = entry.split("-", 1)
```
eskultety: nitpick: the `not entry or '-' not in entry` check should probably come first and stripping could be done as part of splitting a on L53: `entry.strip().split("=", 1)`.

hermeto/core/checksum.py
```diff
@@ -37,9 +38,37 @@ def __str__(self) -> str:
...
+            algorithm = algorithm.strip().lower()
+            checksum = checksum.strip()
+
+            if not algorithm or not checksum:
+                continue
+
+            parsed.append((algorithm, checksum))
+
+        if not parsed:
+            raise ValueError("Integrity value is empty or malformed")
```
eskultety: nitpick: I think 'malformed' will suffice, in most cases the value either is going to be there or not. If it's there than anything other than a successful parse is malformed.

hermeto/core/checksum.py
```diff
@@ -37,9 +38,37 @@ def __str__(self) -> str:
...
+
+        strongest_algorithm, strongest_checksum = max(
+            parsed, key=lambda item: (_SRI_HASH_PRIORITIES.get(item[0], 0), item[0])
```
eskultety: nitpick: the worst case scenario - multiple values - will pretty much never happen, but do we need to loop over the values twice and define a lambda for that purpose? Wouldn't a simple `best` tuple comparison while looping over the sri strings achieve the same thing since we already **have to** loop over it and can't use a trivial list comprehension? :)

hermeto/core/checksum.py
```diff
@@ -37,9 +38,37 @@ def __str__(self) -> str:
 
     @classmethod
     def from_sri(cls, sri: str) -> "ChecksumInfo":
-        """Convert the input Subresource Integrity value to ChecksumInfo."""
-        algorithm, checksum = sri.split("-", 1)
-        return ChecksumInfo(algorithm, base64.b64decode(checksum).hex())
+        """Convert the input Subresource Integrity value to ChecksumInfo.
+
+        The integrity string may contain multiple hashes separated by whitespace.
```
eskultety: I think it would be worth mentioning the SRI docs (maybe even the schema) to back this docstring.

hermeto/core/checksum.py
```diff
@@ -13,6 +13,7 @@
 
 
 SUPPORTED_ALGORITHMS = hashlib.algorithms_guaranteed
+_SRI_HASH_PRIORITIES = {"sha512": 3, "sha384": 2, "sha256": 1}
```
eskultety: There's no reason to define this outside the SRI method.

hermeto/core/checksum.py
```diff
@@ -13,6 +13,7 @@
 
 
 SUPPORTED_ALGORITHMS = hashlib.algorithms_guaranteed
```
eskultety, line 16: **Commit message:**

> npm/yarn

Might be wort mentioning yarn classic explicitly since Yarn v2+ seems to have a different format for the string and we also don't use this method.

hermeto/core/checksum.py
eskultety: Might be also worth mentioning here what others could be because the SRI docs doesn't mention anything past SHA-2, but in fact it could be SHA-1 if git commits are referenced in that field.

hermeto/core/checksum.py
slimreaper35: This lambda is not very readable. I like what we have in [package_distributions.py](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/pip/package_distributions.py#L108) for sdists preference.

hermeto/core/checksum.py
slimreaper35: Something like `RuntimeError` would fit better than `ValueError`. It is not something users provide.

tests/unit/test_checksum.py
```diff
@@ -175,19 +170,37 @@ def test_convert_hex_checksum_to_sri(checksum: str, algorithm: str, expected: st
...
+    [
+        "",
+        "   ",
+        "sha512",
+        "sha512-",
+        "-noalgo",
+        "sha256-",
+        "sha512-not-base64",
+    ],
+)
```
slimreaper35, line 203: Do we need so many negative test cases for that utility function?

eskultety: Yes, this should be the standard and in fact was flagged by the AI bot which respected its config guide. We might not need all the combinations in the world

hermeto/core/checksum.py
eskultety: > It is not something users provide.

Oh, they do, the SRI value comes from the lockfile which was recorded by the user.

> Something like RuntimeError would fit better than ValueError

I'm not sure I'd agree with that given the official docs: 
> Raised when an error is detected that doesn’t fall in any of the other categories. 

If you're looking for a good alternative, then I think `PackageRejected` would be the way to go.

tests/unit/test_checksum.py
a-ovchinnikov: I would have kept them, they tell a story of what will not work.

hermeto/core/checksum.py
```diff
@@ -37,9 +38,50 @@ def __str__(self) -> str:
...
+        if best is None:
+            raise PackageRejected(
+                "Integrity value is empty or malformed",
+                solution="Please check the integrity value in your lockfile.",
+            )
+
+        _, alg, val = best
+        try:
+            return ChecksumInfo(alg, base64.b64decode(val).hex())
+        except Exception:
```
a-ovchinnikov: This is very broad, surely you would want to catch some specific exception? Maybe `binascii.Error` would be a better candidate for this?

hermeto/core/checksum.py
```diff
@@ -37,9 +39,50 @@ def __str__(self) -> str:
...
+        SRI specification [https://www.w3.org/TR/sri-2/].
+
+        When multiple hashes are present, we pick the strongest algorithm
+        (sha512 > sha384 > sha256 > others). Note that while the SRI specification only mentions
+        SHA-2 algorithms (SHA-256, SHA-384, SHA-512), in practice "others" can include legacy
+        algorithms like sha1, which may appear when Git commit hashes are referenced in
+        integrity fields.
+        """
+
+        _SRI_HASH_PRIORITIES = {"sha512": 3, "sha384": 2, "sha256": 1}
```
slimreaper35:
```suggestion
        sri_hash_priorities = {"sha512": 3, "sha384": 2, "sha256": 1}
```

slimreaper35: the variable is declared as private (seems like it is a leftover from the previous iteration)

hermeto/core/checksum.py
eskultety: @derasdf looks like this hasn't been addressed in the commit message yet :), only in the commentary, thanks.

hermeto/core/checksum.py
```diff
@@ -37,9 +39,50 @@ def __str__(self) -> str:
...
+                solution="Please check the integrity value in your lockfile.",
+            )
+
+        _, alg, val = best
+        try:
+            return ChecksumInfo(alg, base64.b64decode(val).hex())
+        except binascii.Error as e:
+            raise PackageRejected(
+                "Integrity value is malformed",
+                solution="Please check the integrity value in your lockfile.",
```
eskultety, line 79: Hmm, there seems to be some exception redundancy here in these hunks, can we do instead:
```
try:
    _, alg, val = best   # will raise TypeError if best is None
    return ChecksumInfo(alg, base64.b64decode(val).hex())
except (TypeError, binasciiError) as e:
    ...
```


Not a blocker for me if there's not a general consensus on such patterns in our code base.

derasdf: I'm not completely sure either but for now i implemented the suggestion.

PR #1201, opened 2025-12-05
Prepare SBOM for permissive mode annotations

### LINKED ISSUE #936: Taint the SBOM with information on failed checks in permissive mode
It's good that we'd want to allow users to progress further in their build pipelines by relaxing some of our checks via the planned permissive mode toggle, but with provenance checking systems in the picture we should also provide a record/records in the SBOM which checks would have otherwise failed and hence tainting the SBOM in way these provenance checking systems could verify and act upon.

The key here will be figuring out how granular we want this metadata to be which will directly impact the design with regards to the CycloneDX and SPDX formats and their fields respectively. 

**Depends on: https://github.com/hermetoproject/hermeto/issues/934**

slimreaper35:
Closes #936 

### DIFF
--- a/hermeto/core/models/sbom.py
+++ b/hermeto/core/models/sbom.py
@@ -180,6 +180,19 @@ class Sbom(pydantic.BaseModel):
     spec_version: str = pydantic.Field(alias="specVersion", default="1.6")
     version: int = 1
 
+    @pydantic.model_serializer(mode="wrap")
+    def serialize_model(self, handler: pydantic.SerializerFunctionWrapHandler) -> dict[str, object]:
+        """
+        Custom serializer that removes the `annotations` field if it is empty.
+
+        https://docs.pydantic.dev/latest/concepts/serialization/#model-serializers
+        """
+        serialized = handler(self)
+        if not self.annotations:
+            serialized.pop("annotations")
+
+        return serialized
+
     def __add__(self, other: Union["Sbom", "SPDXSbom"]) -> "Sbom":
         if isinstance(other, self.__class__):
             return Sbom(


hermeto/interface/cli.py
a-ovchinnikov, line 1: **Commit message**: we do not support full SPDX, just a subset, could you please reword the message? Something along the lines of "Drop redundant mention of SPDX since we already mention that we suport merging SBOMs generated by Hermeto" would be less confusing.

hermeto/core/models/sbom.py
```diff
@@ -166,6 +166,19 @@ class Sbom(pydantic.BaseModel):
     spec_version: str = pydantic.Field(alias="specVersion", default="1.6")
     version: int = 1
 
+    @pydantic.model_serializer(mode="wrap")
+    def serialize_model(self, handler: pydantic.SerializerFunctionWrapHandler) -> dict[str, object]:
+        """
+        Custom serializer that removes the `annotations` field if it is empty.
```
a-ovchinnikov, line 186: Out of curiosity, why would you want to remove it?

hermeto/core/models/output.py
```diff
@@ -151,6 +151,7 @@ def _unique_project_files(cls, project_files: list[ProjectFile]) -> list[Project
 class RequestOutput(pydantic.BaseModel):
     """Results of processing one or more package managers."""
 
+    annotations: list[Annotation]
```
a-ovchinnikov: WDYT about adding a default value to this? It looks like it gets initialized with an empty list almost everywhere.
```diff
-    annotations: list[Annotation]
+    annotations: list[Annotation] = Field(default_factory=list)
```
[source](https://docs.pydantic.dev/latest/concepts/fields/#default-values)

a-ovchinnikov (commented):
LGTM in general with a few minor suggestions.

hermeto/core/models/sbom.py
slimreaper35: I thought, we dont' want to have empty `annotations` field in almost every SBOM.

hermeto/core/models/output.py
slimreaper35: done

hermeto/core/models/sbom.py
```diff
@@ -23,6 +23,21 @@
 log = logging.getLogger(__name__)
 
 
+Annotator = Literal["organization", "individual", "component", "service"]
+
+
+class Annotation(pydantic.BaseModel):
```
eskultety, line 33: Might want to add a link to the SPDX doc here as well as to the commit message for easy referencing.

eskultety:
> I'm missing some SPDX SBOM test updates exercising annotations (visual verification needed here :) ), but otherwise LGTM.

To clarify on ^this, I know you're adding it here: https://github.com/hermetoproject/hermeto/pull/1204/commits/5f29adb7df2f06a58263783d8145ce726025d4c7 I was just wondering if we could have some generic unit test as well not necessarily connected to the permissive mode effort just to make sure the annotations are visually added correctly.

slimreaper35:
>  I was just wondering if we could have some generic unit test as well not necessarily connected to the permissive mode effort just to make sure the annotations are visually added correctly.

Added.

hermeto/core/models/output.py
```diff
@@ -161,7 +161,9 @@ def generate_sbom(self) -> Sbom:
         Note that RequestOutput may contain duplicated components, we de-duplicate them here
         while merging their `properties`.
         """
-        return Sbom(components=merge_component_properties(self.components))
+        return Sbom(
+            annotations=self.annotations, components=merge_component_properties(self.components)
```
taylormadore, line 165: Do annotations need any special merge handling?

hermeto/core/models/sbom.py
```diff
@@ -23,6 +23,23 @@
...
+    """
+    A comment, note, explanation, or similar textual content
+    which provides additional context to the object(s) being annotated.
+
+    https://cyclonedx.org/docs/1.6/json/#annotations
+    """
+
+    subjects: list[str]
+    annotator: dict[Annotator, dict[str, str]]
+    timestamp: str
```
taylormadore: Just a question: Is there any reason not to use the datetime type internally and have a serializer for the correct format? I see we also use str for SPDX annotationDate.

hermeto/core/models/output.py
slimreaper35: I don't think so. Each annotation is a separate object and so each one should reference separate components.

hermeto/core/models/sbom.py
slimreaper35: We could use datetime type, but we would still need to pass `datetime.datetime.now()`, right ? But without applying ISO format.

hermeto/core/models/output.py
taylormadore: If I use the `merge-sboms` command on two Hermeto-produced SBOMs and each SBOM has an annotation for permissive mode, what should the correct annotation output be for the merged SBOM? One or two annotations? The components used for the bom-ref get merged/deduplicated based on the PURL

hermeto/core/models/sbom.py
a-ovchinnikov: I have linked [this](https://docs.pydantic.dev/2.0/usage/types/datetime/#validation-of-datetime-types) in the other PR, the validator looks very permissive and accepting quite a few options, from the looks of it it will be fine with a string representation.

hermeto/core/models/sbom.py
```diff
@@ -22,6 +22,29 @@
 log = logging.getLogger(__name__)
 
 
+def datetime_to_iso_8601(value: datetime) -> str:
+    """Serialize a datetime to a string in ISO 8601 standard."""
+    # or return value.isoformat(timespec="seconds") + "Z"
+    return value.strftime("%Y-%m-%dT%H:%M:%SZ")
```
a-ovchinnikov: So, which one is better?

hermeto/core/models/output.py
a-ovchinnikov: To elaborate on @slimreaper35's answer (initially I got confused a bit too): [annotations](https://github.com/hermetoproject/hermeto/blob/main/docs/design/permissive-mode.md#approach-1-preferred-top-level-annotation-with-component-references) will be created per violating component, so merging two CycloneDX SBOMs each containing affected components should result in a SBOM with concatenated lists of annotations. Now the question is whether it is possible to get two SBOMs containing identical components? If yes and if such a component happens to require permissive mode then it would probably need deduplication. (Realistically this is likely a very rare corner case with the only side effect of minor data duplication so I am not convinced that it must be handled.)

On the other hand .subjects is a list, so it might make sense to merge `Annotation` objects when they have identical .annotator and .text values. This would results in a loss of one of timestamps which is probably not that important in this case and would also look rather unsavory, at least if attempted through adding \_\_add\_\_().

taylormadore: My comment was based on my initial reading of the design doc where it currently states:

> a single, centralized annotation entry can reference dozens of affected components by their bom-ref, reducing data duplication

Based on this, I assumed that annotations would be per .text field with a list of subject components and not per-component. We may need to clarify that this is actually the case.

hermeto/core/models/sbom.py
slimreaper35: let's go with `value.strftime("%Y-%m-%dT%H:%M:%SZ")` 😄 seems more transparent

hermeto/core/models/output.py
taylormadore: This isn't a current concern/blocker by any means--just a note that if we expand use of annotations in the future, we may need more complex merge behavior. We can consider this resolved from my POV with a comment

PR #1203, opened 2025-12-08
Add multiple exception class in errors.py

shokakucarrier:
Add multiple exception class in errors.py
new errors class: ExecutableNotFound, ChecksumVerificationFailed, LockfileNotFound, InvalidLockfileFormat, ChecksumMissing
Adapted new exception class in code including unit test to replace BaseError useage
including all BaseError reference in generic/main.py, yarn/main.py, yarn_classic/main.py, rpm/main.py; npm.py, gomod.py and output.py
replaced some of the PackageRejected error if it is applicable

Altered all unit test and integration test of those exceptions

Resolves: https://github.com/hermetoproject/hermeto/issues/876

Detailed BaseError and PackageRejected replaced:

Based on the new commit 433c50a, here are the GitHub hyperlinks for all the places where you replaced `BaseError` and `PackageRejected` with new exception classes in the main codebase (excluding tests):

### ExecutableNotFound
- [`hermeto/core/utils.py` (line 54)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/utils.py#L54)

### ChecksumVerificationFailed
- [`hermeto/core/checksum.py` (line 129)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/checksum.py#L129)
- [`hermeto/core/package_managers/rpm/main.py` (line 379)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/rpm/main.py#L379)

### InvalidChecksum
- [`hermeto/core/checksum.py` (lines 86-94)](https://github.com/hermetoproject/hermeto/blob/433c50a952b391fdb2b6736133e8df20fd1056de/hermeto/core/checksum.py#L86-L94)
- [`hermeto/core/package_managers/pip/requirements.py` (lines 635-641)](https://github.com/hermetoproject/hermeto/blob/433c50a952b391fdb2b6736133e8df20fd1056de/hermeto/core/package_managers/pip/requirements.py#L635-L641)
- [`hermeto/core/package_managers/pip/requirements.py` (lines 687-691)](https://github.com/hermetoproject/hermeto/blob/433c50a952b391fdb2b6736133e8df20fd1056de/hermeto/core/package_managers/pip/requirements.py#L687-L691)

### MissingChecksum
- [`hermeto/core/checksum.py` (lines 79-84)](https://github.com/hermetoproject/hermeto/blob/433c50a952b391fdb2b6736133e8df20fd1056de/hermeto/core/checksum.py#L79-L84)
- [`hermeto/core/package_managers/npm.py` (lines 524-529)](https://github.com/hermetoproject/hermeto/blob/433c50a952b391fdb2b6736133e8df20fd1056de/hermeto/core/package_managers/npm.py#L524-L529)
- [`hermeto/core/package_managers/pip/requirements.py` (lines 674-678)](https://github.com/hermetoproject/hermeto/blob/433c50a952b391fdb2b6736133e8df20fd1056de/hermeto/core/package_managers/pip/requirements.py#L674-L678)

### LockfileNotFound
- [`hermeto/core/package_managers/bundler/parser.py` (lines 31-35)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/bundler/parser.py#L31-L35)
- [`hermeto/core/package_managers/cargo/main.py` (lines 179-182)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/cargo/main.py#L179-L182)
- [`hermeto/core/package_managers/generic/main.py` (line 74)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/generic/main.py#L74)
- [`hermeto/core/package_managers/gomod.py` (lines 683-687)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/gomod.py#L683-L687)
- [`hermeto/core/package_managers/npm.py` (lines 717-723)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/npm.py#L717-L723)
- [`hermeto/core/package_managers/pip/main.py` (lines 540-543)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/pip/main.py#L540-L543)
- [`hermeto/core/package_managers/rpm/main.py` (lines 282-284)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/rpm/main.py#L282-L284)
- [`hermeto/core/package_managers/yarn/main.py` (lines 84-85)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/yarn/main.py#L84-L85)
- [`hermeto/core/package_managers/yarn/project.py` (lines 105-111)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/yarn/project.py#L105-L111)
- [`hermeto/core/package_managers/yarn_classic/main.py` (line 219)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/yarn_classic/main.py#L219)
- [`hermeto/core/package_managers/yarn_classic/project.py` (lines 65-71)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/yarn_classic/project.py#L65-L71)
- [`hermeto/core/package_managers/yarn_classic/project.py` (lines 101-107)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/yarn_classic/project.py#L101-L107)

### InvalidLockfileFormat
- [`hermeto/core/package_managers/generic/main.py` (lines 107-110)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/generic/main.py#L107-L110)
- [`hermeto/core/package_managers/generic/main.py` (lines 120-127)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/generic/main.py#L120-L127)
- [`hermeto/core/package_managers/rpm/main.py` (lines 294-297)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/rpm/main.py#L294-L297)
- [`hermeto/core/package_managers/rpm/main.py` (lines 306-308)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/rpm/main.py#L306-L308)
- [`hermeto/core/package_managers/yarn/project.py` (lines 113-119)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/yarn/project.py#L113-L119)
- [`hermeto/core/package_managers/yarn_classic/project.py` (lines 73-79)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/yarn_classic/project.py#L73-L79)
- [`hermeto/core/package_managers/yarn_classic/project.py` (lines 109-114)](https://github.com/hermetoproject/hermeto/blob/6cd9ae023a7873a31ff31fee0e056212b548777f/hermeto/core/package_managers/yarn_classic/project.py#L109-L114)

### DIFF
--- a/hermeto/core/package_managers/yarn_classic/main.py
+++ b/hermeto/core/package_managers/yarn_classic/main.py
@@ -36,16 +36,6 @@ YARN_NETWORK_TIMEOUT_MILLISECONDS = 600000
 _yarn_classic_pattern = "yarn lockfile v1"  # See [yarn_classic_trait].
 
 
-class MissingLockfile(PackageRejected):
-    """Indicate that a lock file is missing."""
-
-    def __init__(self) -> None:
-        """Initialize a Missing Lockfile error."""
-        reason = "Yarn lockfile 'yarn.lock' missing, refusing to continue"
-        solution = "Make sure your repository has a Yarn lockfile (i.e. yarn.lock) checked in"
-        super().__init__(reason, solution=solution)
-
-
 class NotV1Lockfile(PackageRejected):
     """Indicate that a lockfile is of wrong tyrpoe."""
 


hermeto/core/errors.py
```diff
@@ -124,6 +125,162 @@ class UnsupportedFeature(UsageError):
...
+        reason = (
+            f"Supplied lockfile path '{lockfile_name}' must be inside the package "
+            f"path '{package_path}'"
+        )
+        if solution == _argument_not_specified:
+            solution = "Use a lockfile path located within the package path."
+        super().__init__(reason, solution=solution, docs=docs)
+
+
+class CyclicDependency(UsageError):
```
eskultety: This one was added for a single purpose only, the use case seems to be pretty niche, at least the current scenario where this is used. We can always add it back when needed, the point of this PR should be to set a precedent for future additions :). Please drop this one.

hermeto/core/checksum.py
```diff
@@ -6,7 +6,7 @@
 from pathlib import Path
 from typing import NamedTuple
```
eskultety, line 8: **Commit subject:**
s/class/classes

**Commit message:**
- I'd like to see some more description on which error cases are now covered with an exception class, i.e. not the classes names, but brief scenario descriptions
- if this closes an issue (which it does), please mention the URL (i.e. not the issue ID) in the commit message as well, e.g. `Fixes https://github.com/hermetoproject/hermeto/issues/876`

hermeto/core/package_managers/generic/main.py
```diff
@@ -72,12 +67,8 @@ def _resolve_generic_lockfile(lockfile_path: Path, output_dir: RootedPath) -> li
     :param output_dir: the output directory to store the dependencies
     """
     if not lockfile_path.exists():
-        raise PackageRejected(
-            f"{APP_NAME} generic lockfile '{lockfile_path}' does not exist, refusing to continue.",
-            solution=(
-                f"Make sure your repository has {APP_NAME} generic lockfile '{DEFAULT_LOCKFILE_NAME}' "
-                f"checked in to the repository, or the supplied lockfile path is correct."
-            ),
+        raise LockfileNotFound(
```
eskultety: I believe that 
- https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/rpm/main.py#L278
- https://github.com/hermetoproject/hermeto/blob/d02aff98684ecfe4aa0f8986571a9a95fe44761f/hermeto/core/package_managers/yarn/main.py#L81
- https://github.com/hermetoproject/hermeto/blob/d02aff98684ecfe4aa0f8986571a9a95fe44761f/hermeto/core/package_managers/npm.py#L711
- https://github.com/hermetoproject/hermeto/blob/d02aff98684ecfe4aa0f8986571a9a95fe44761f/hermeto/core/package_managers/gomod.py#L681
- https://github.com/hermetoproject/hermeto/blob/d02aff98684ecfe4aa0f8986571a9a95fe44761f/hermeto/core/package_managers/bundler/parser.py#L31

 should all get the exact same treatment.

eskultety: Additionally, I think https://github.com/hermetoproject/hermeto/blob/d02aff98684ecfe4aa0f8986571a9a95fe44761f/hermeto/core/package_managers/yarn_classic/main.py#L39 the `MissingLockfile` exception could be consolidated underneath this newly added one.

hermeto/core/errors.py
```diff
@@ -124,6 +125,162 @@ class UnsupportedFeature(UsageError):
...
+        :param solution: politely suggest a potential solution to the user
+        :param docs: include a link to relevant documentation (if there is any)
+        """
+        reason = f"{APP_NAME} lockfile '{lockfile_path}' format is not valid: {err_details}"
+        if solution == _argument_not_specified:
+            solution = "Check correct syntax in the lockfile."
+        super().__init__(reason, solution=solution, docs=docs)
+
+
+class InvalidPackagePath(PackageRejected):
```
eskultety: I don't think we need this one at all because this scenario is/should be already covered by the existing `PathOutsideRoot` exception.

hermeto/core/errors.py
```diff
@@ -124,6 +125,162 @@ class UnsupportedFeature(UsageError):
...
+        super().__init__(reason, solution=solution, docs=docs)
+
+    default_solution = (
+        "Please make sure that the required executable is installed in your PATH.\n"
+        f"If you are using {APP_NAME} via its container image, this should not happen - "
+        "please report this bug."
+    )
+
+
+class ChecksumVerificationFailed(PackageRejected):
```
eskultety, line 144: Since we're adding a checksum validation specific exception type we might also benefit from a `ChecksumMissing` one, just an idea.

hermeto/core/package_managers/generic/main.py
shokakucarrier: Fixed in all places, do we need to replace more PackageRejected with errors like this?

hermeto/core/package_managers/generic/main.py
```diff
@@ -51,15 +50,11 @@ def _resolve_lockfile_path(
         return lockfile_path
 
     path = source_dir.join_within_root(package_path)
-    lockfile_name = lockfile_path or DEFAULT_LOCKFILE_NAME
+    lockfile_name = lockfile_path if lockfile_path is not None else Path(DEFAULT_LOCKFILE_NAME)
```
a-ovchinnikov: A non-blocking nitpick: this change seems to be unrelated and purely cosmetic.

hermeto/core/errors.py
```diff
@@ -253,34 +253,6 @@ def __init__(
         super().__init__(reason, solution=solution, docs=docs)
 
 
-class CyclicDependency(UsageError):
```
a-ovchinnikov: Could you please amend the previous commit with this chunk? Otherwise we'll have a commit introducing an entity followed by a commit removing the same entity.

hermeto/core/package_managers/generic/main.py
shokakucarrier: fixed

hermeto/core/errors.py
shokakucarrier: removed

hermeto/core/errors.py
shokakucarrier: can we do it later? I didn't find there is any existing reference that can replace with ChecksumMissing error

eskultety: > I didn't find there is any existing reference that can replace with ChecksumMissing error

- https://github.com/hermetoproject/hermeto/blob/1fa9f74d1c14c5889919dee975903c38d8058ac1/hermeto/core/package_managers/npm.py#L518
- https://github.com/hermetoproject/hermeto/blob/1fa9f74d1c14c5889919dee975903c38d8058ac1/hermeto/core/package_managers/pip/requirements.py#L681
- https://github.com/hermetoproject/hermeto/blob/1fa9f74d1c14c5889919dee975903c38d8058ac1/hermeto/core/package_managers/pip/requirements.py#L642C23-L642C38

> can we do it later?

We could, but it's simple enough and since we already have at least 3 occurrences :) ...

hermeto/core/errors.py
shokakucarrier: OK, I'll merge two commits together

shokakucarrier: fixed

hermeto/core/errors.py
shokakucarrier: OK, I'll change these two case, but are we going to replace more of the PackageRejected error? the original error were about just to replace BaseError where the error says not to raise directly

shokakucarrier: fixed this one

hermeto/core/models/output.py
```diff
@@ -82,15 +82,14 @@ def get_placeholders(t: string.Template) -> set[str]:
...
-                raise BaseError(
+                raise PackageRejected(
                     f"Detected a cycle in environment variable expansion of '{self.name}'",
                     solution=(
                         "Inspect all relevant environment variables and make sure their "
                         "expansion doesn't lead to a cycle. Ideally, avoiding nesting of "
                         "variables altogether."
                     ),
                 )
-
```
eskultety, line 93: Unrelated change.

hermeto/core/package_managers/generic/main.py
```diff
@@ -55,10 +54,9 @@ def _resolve_lockfile_path(
     lockfile = path.join_within_root(lockfile_name).path
 
     if not lockfile.is_relative_to(path.path):
-        raise PackageRejected(
-            f"Supplied generic lockfile path '{lockfile_name}' must be inside the package "
-            f"path '{package_path}'.",
-            solution="Use a lockfile path located within the package path.",
+        raise PathOutsideRoot(
+            s_self=str(lockfile),
+            s_root=str(path.path),
```
eskultety: Unrelated change and also incorrect. `PathOutsideRoot` would have been already emitted on L54 if that were the case.
In fact, looking at the check on L56, it is redundant and should be dropped completely (in a standalone commit).

hermeto/core/errors.py
```diff
@@ -124,6 +125,107 @@ class UnsupportedFeature(UsageError):
...
+        docs: str | None = None,
+    ) -> None:
+        """Initialize LockfileNotFound.
+
+        :param lockfile_path: Path where lockfile was expected
+        :param lockfile_name: Name of the expected lockfile
+        :param solution: politely suggest a potential solution to the user
+        :param docs: include a link to relevant documentation (if there is any)
+        """
+        reason = f"{APP_NAME} lockfile '{lockfile_name}' does not exist in '{lockfile_path}', refusing to continue"
```
eskultety: `APP_NAME` resolves to either `hermeto` or `cachi2` and as such shouldn't appear in the reason or solution strings unless necessary, e.g. `hermeto lockfile 'foo.lock' does not exist in...` gives the wrong impressions as if hermeto owned the lockfile which is not true; this applies to all the newly added classes where I can see this pattern as well.

Additionally, consider an alternative, slightly more generic wording (which is more future proof):
`'{lockfile_name}' not found under '{lockfile_path}', refusing to continue`.

eskultety: Based on a later change:
`lockfile 'Gemfile and Gemfile.lock' does not exist in ..., refusing to continue`

is IMO not a very good user-facing error message. ^This showed us that you need to allow `reason` to be an argument to the exception constructor that can be overridden - applies to all newly added exception classes.

hermeto/core/models/output.py
```diff
@@ -82,15 +82,14 @@ def get_placeholders(t: string.Template) -> set[str]:
 
             substituted |= get_placeholders(t_old) & mappings.keys()
             if substituted & p_new:
-                raise BaseError(
+                raise PackageRejected(
```
eskultety: This is not a `PackageRejected` error, this is an internal error caused by our own actions. `BaseError` is well justified, it could also be an `InternalError` but such a fix should happen in a separate patch.

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -28,12 +28,9 @@ def parse_lockfile(package_dir: RootedPath, allow_binary: bool = False) -> Parse
     lockfile_path = package_dir.join_within_root(GEMFILE_LOCK)
     gemfile_path = package_dir.join_within_root(GEMFILE)
     if not lockfile_path.path.exists() or not gemfile_path.path.exists():
-        reason = "Gemfile and Gemfile.lock must be present in the package directory"
-        solution = (
-            "Run `bundle init` to generate the Gemfile.\n"
-            "Run `bundle lock` to generate the Gemfile.lock."
+        raise LockfileNotFound(
+            lockfile_path=lockfile_path.path, lockfile_name="Gemfile and Gemfile.lock"
```
eskultety: See my comment above, this scenario calls for a `reason` argument that could be overridden.

hermeto/core/package_managers/npm.py
```diff
@@ -708,10 +713,9 @@ def _resolve_npm(pkg_path: RootedPath, npm_deps_dir: RootedPath) -> ResolvedNpmP
         if package_lock_path.path.exists():
             break
     else:
-        raise PackageRejected(
-            "The npm-shrinkwrap.json or package-lock.json file must be present for the npm "
-            "package manager",
-            solution="Please double-check that you have specified the correct path to the package directory containing one of those two files",
+        raise LockfileNotFound(
+            lockfile_path=package_lock_path.path,
+            lockfile_name="npm-shrinkwrap.json or package-lock.json",
```
eskultety: Same here, `reason` override is needed here.

hermeto/core/package_managers/yarn_classic/main.py
```diff
@@ -36,16 +36,6 @@
...
-class MissingLockfile(PackageRejected):
-    """Indicate that a lock file is missing."""
-
-    def __init__(self) -> None:
-        """Initialize a Missing Lockfile error."""
-        reason = "Yarn lockfile 'yarn.lock' missing, refusing to continue"
-        solution = "Make sure your repository has a Yarn lockfile (i.e. yarn.lock) checked in"
-        super().__init__(reason, solution=solution)
-
-
```
eskultety, line 48: Usually cleanups like this one are done in separate commits  when the replacement is already in place to keep history clean and diffs manageable.

hermeto/core/checksum.py
```diff
@@ -6,7 +6,7 @@
...
```
eskultety, line 9: **Commit subject:**
Introduce more exception subtypes for common scenarios

**Commit message:**
I believe we can do better with the commit description, how about this:

"""
This introduces better granularity into our existing exceptions by adding more `PackageRejected` exception subtypes to handle the following common error scenarios:
- package checksum verification failures
- missing lockfiles/manifest files
- invalid lockfile/manifest file format
- executables not found

This is to give us the ability to differentiate between various error paths in our tests and hence rid us of the need to do error string comparisons which is error prone in itself.
"""

hermeto/core/package_managers/rpm/main.py
```diff
@@ -275,12 +275,9 @@ def _resolve_rpm_project(
...
     if not source_dir.join_within_root(DEFAULT_LOCKFILE_NAME).path.exists():
-        raise PackageRejected(
-            f"RPM lockfile '{DEFAULT_LOCKFILE_NAME}' missing, refusing to continue.",
-            solution=(
-                "Make sure your repository has RPM lockfile '{DEFAULT_LOCKFILE_NAME}' checked in "
-                "to the repository."
-            ),
+        raise LockfileNotFound(
+            lockfile_path=source_dir.subpath_from_root,
+            lockfile_name=DEFAULT_LOCKFILE_NAME,
```
eskultety: multiple occurrences of this pattern - so you're introducing a new exception class `Checksum missing`, but you're also updating unrelated tests in  this commit, that should have been done in the previous one. In fact, since both the previous commit and this one introduce new exception types, you can squash them to a single commit.

hermeto/core/errors.py
```diff
@@ -175,6 +175,27 @@ def __init__(
...
+    ) -> None:
+        """Initialize ChecksumMissing.
+
+        :param reason: explain what checksum is missing and for which dependency
+        :param solution: politely suggest a potential solution to the user
+        :param docs: include a link to relevant documentation (if there is any)
+        """
+        super().__init__(reason, solution=solution, docs=docs)
+
+    default_solution = "Please ensure that all dependencies have proper checksums/hashes specified in the lockfile."
```
eskultety: I'd like us to avoid overusing the term 'lockfile'. Not all ecosystems call it that way, some call it manifests, some just call it by their name.

suggestion: `Please ensure all dependencies have their checksums listed to verify integrity`

PS: Yeah, and I'm coming with this after having approved the `LockfileNotFound` exception type, but in the case of the latter thatt's just an internal name of a constant, it doesn't and shouldn't necessarily appear in the user-facing error messages like it does in the `default_solution` field :), but in order to defend myself note that I've deliberately omitted the word 'lockfile' from any user facing message suggestions I made in previous commit :) .

hermeto/core/errors.py
```diff
@@ -116,6 +117,127 @@ class UnsupportedFeature(UsageError):
...
+        *,
+        reason: str | None = None,
+        solution: str | None = _argument_not_specified,
+    ) -> None:
+        """Initialize LockfileNotFound.
+
+        :param lockfile_path: Path where lockfile was expected
+        :param lockfile_name: Name of the expected lockfile
+        :param solution: politely suggest a potential solution to the user
+        :param docs: include a link to relevant documentation (if there is any)
```
a-ovchinnikov: There does not seem to be `docs` kwarg anymore, but at the same time there is `reason`.

tests/integration/test_bundler.py
```diff
@@ -18,7 +18,7 @@
                 check_output=False,
                 check_deps_checksums=False,
                 expected_exit_code=2,
-                expected_output="Gemfile and Gemfile.lock must be present in the package directory",
+                expected_output="hermeto lockfile 'Gemfile and Gemfile.lock' does not exist",
```
a-ovchinnikov: I am sorry, these are not Hermeto lock files. `Gemfile.lock` is a lock file used by Bundler (dependency manager for Ruby ecosystem), `Gemfile` is not a lock file at all, it is a list of dependencies alongside with constraints on their versions. A lock file would typically pin a specific version instead of providing a range.

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -62,13 +62,10 @@ def sample_parser_output() -> dict[str, Any]:
...
 def test_parse_lockfile_without_bundler_files(rooted_tmp_path: RootedPath) -> None:
-    with pytest.raises(PackageRejected) as exc_info:
+    with pytest.raises(LockfileNotFound) as exc_info:
         parse_lockfile(rooted_tmp_path)
 
-    assert (
-        "Gemfile and Gemfile.lock must be present in the package directory"
-        in exc_info.value.friendly_msg()
-    )
+    assert "Gemfile or Gemfile.locklockfile' not found under" in exc_info.value.friendly_msg()
```
a-ovchinnikov: Since you are already testing for a specific exception type string check could be dropped.

tests/unit/test_errors.py
```diff
@@ -63,3 +63,28 @@ def test_fetch_error_friendly_msg() -> None:
         """
     ).strip()
     assert err.friendly_msg() == expect_msg
+
+
+def test_checksum_missing_friendly_msg() -> None:
```
a-ovchinnikov: This test does not seem to be testing any busyness logic, it effectively tests whether Python can instantiate objects correctly. I would recommend against adding tests which exercise functionality tested elsewhere.

tests/unit/package_managers/bundler/test_parser.py
slimreaper35: Misspelled string.

hermeto/core/errors.py
```diff
@@ -116,6 +117,127 @@ class UnsupportedFeature(UsageError):
...
+        *,
+        solution: str | None = _argument_not_specified,
+    ) -> None:
+        """Initialize InvalidLockfileFormat.
+
+        :param lockfile_path: Path to the invalid lockfile
+        :param err_details: Details about what is invalid
+        :param solution: politely suggest a potential solution to the user
+        :param docs: include a link to relevant documentation (if there is any)
+        """
```
slimreaper35, line 254: [multiple occurences] `docs:` does not exist anymore

Do we need such detailed docstring for each exception ? It is not a library for users. 

For example, `lockfile_path: Path to the invalid lockfile` does not need any further explanation. Also, I see `:param solution:` explained like 5x.

hermeto/core/errors.py
```diff
@@ -116,6 +117,127 @@ class UnsupportedFeature(UsageError):
...
+        reason = f"Failed to verify {filename} against any of the provided checksums"
+        super().__init__(reason, solution=solution)
+
+    default_solution = (
+        "Verify that the file has not been corrupted and that the expected checksums are correct."
+    )
+
+
+class ChecksumMissing(PackageRejected):
```
slimreaper35:
```suggestion
class MissingChecksum(PackageRejected):
```

shokakucarrier: OK, I made the change with reference as well

tests/unit/test_errors.py
shokakucarrier: Deleted these two case

tests/integration/test_bundler.py
shokakucarrier: Fixed, I forgot to change test string here

hermeto/core/errors.py
shokakucarrier: Dropped `docs:` field
but other param I want to keep them for the format, are we going to delete them?

shokakucarrier:
@a-ovchinnikov can you help me rerun the test? the last problem was fixed now, thanks!

hermeto/core/package_managers/pip/requirements.py
```diff
@@ -629,7 +634,7 @@ def validate_requirements(requirements: list[PipRequirement]) -> None:
                     f"URL requirement must specify exactly one hash, but specifies {n_hashes}: "
                     f"{req.download_line}."
                 )
-                raise PackageRejected(
+                raise MissingChecksum(
```
a-ovchinnikov: I afraid this is not entirely correct. The exception is raised when there is a number of hashes not equal to 1, which could be 2, in this case calling it a "MissingChecksum" would be misleading.

hermeto/core/errors.py
```diff
@@ -116,6 +117,123 @@ class UnsupportedFeature(UsageError):
...
+
+    def __init__(
+        self,
+        reason: str,
+        *,
+        solution: str | None = _argument_not_specified,
+    ) -> None:
+        """Initialize MissingChecksum.
+
+        :param reason: explain what checksum is missing and for which dependency
```
a-ovchinnikov: This statement is ambiguous and results in a rather ambiguous interface. For something that is called "MissingChecksum" I would expect to see just _one_ argument here: the name of an offending dependency. I think the ambiguity stems from the fact that in reality this is a combination of two slightly different exceptions into one. Could you please split it in two? (Also calling it "reason" implies that we know the reason why a checksum is missing, when in reality we can only observe that it is missing and report that.)

hermeto/core/errors.py
```diff
@@ -116,6 +117,123 @@ class UnsupportedFeature(UsageError):
...
+        solution: str | None = _argument_not_specified,
+    ) -> None:
+        """Initialize LockfileNotFound.
+
+        :param lockfile_path: Path where lockfile was expected
+        :param lockfile_name: Name of the expected lockfile
+        :param reason: explain what went wrong
+        :param solution: politely suggest a potential solution to the user
+        """
+        if reason is None:
```
a-ovchinnikov: Of about a dozen instantiations of `LockfileNotFound` only two use `reason` argument, in both cases the provided reason does not differ much from the default one, furthermore, in both cases it could be constructed the same way other reasons are constructed since `lockfile_name` accepts a string. I would suggest dropping user-visible `reason` field altogether and renaming `lockfile_name` accordingly, maybe to something like `affected_files`? (This is not a perfect name either since it hints at an iterator. Now iterator is fine, but would require some additional work on the initializer side: if it is a string then it should be substituted as is, it it is a different type of iterable then its elements must be concatenated as strings with " or ".)

hermeto/core/errors.py
```diff
@@ -116,6 +117,123 @@ class UnsupportedFeature(UsageError):
...
 
 
+class ExecutableNotFound(UsageError):
+    """A required executable was not found in PATH."""
+
+    def __init__(
+        self,
+        executable: str,
+        *,
+        solution: str | None = _argument_not_specified,
```
a-ovchinnikov, line 127: Here and in other classes `solution` is typed as a union of `str` and `None` and then set to a string by default. Are there any instances where it is overridden to `None` anywhere? I believe this is some kind of an artefact from the past that keeps lurking in the codebase and `solution` could be safely typed just as a string. (I would argue that a better way of doing this is assigning `_argument_not_specified` to a dedicated sentinel object and then comparing to it instead of a string, but that is out of scope for this change.)

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -62,13 +62,10 @@ def sample_parser_output() -> dict[str, Any]:
...
 def test_parse_lockfile_without_bundler_files(rooted_tmp_path: RootedPath) -> None:
-    with pytest.raises(PackageRejected) as exc_info:
+    with pytest.raises(LockfileNotFound) as exc_info:
         parse_lockfile(rooted_tmp_path)
 
-    assert (
-        "Gemfile and Gemfile.lock must be present in the package directory"
-        in exc_info.value.friendly_msg()
-    )
+    assert "Gemfile or Gemfile.lock' not found under" in exc_info.value.friendly_msg()
```
a-ovchinnikov: Multiple occasions (in this module and others): do we still need these string checks after introducing proper exception types? My understanding of the issue was that we want narrower types to avoid string matching: string matching in tests is always  brittle and must be done only as a last resort.

eskultety: > Multiple occasions (in this module and others): do we still need these string checks after introducing proper exception types? 

Agreed, we don't.

hermeto/core/errors.py
eskultety: At the time of writing this patch it's consistent with the rest, so either it has to be addressed in a dedicated commit as part of this PR globally or we just leave it as a `my_first_issue` follow up. I'm fine both ways.

hermeto/core/package_managers/yarn_classic/main.py
```diff
@@ -216,7 +216,7 @@ def _reject_if_wrong_lockfile_version(project: Project) -> None:
 def _reject_if_lockfile_is_missing(project: Project) -> None:
     yarnlock_path = _get_path_to_yarn_lock(project)
     if not yarnlock_path.exists():
-        raise MissingLockfile()
```
eskultety, line 219: All `MissingLockfile` references have been dropped, but the exception class definition remained - needs to be cleaned up in a standalone commit (i.e. commit nr. 2 in this PR).

hermeto/core/package_managers/rpm/main.py
```diff
@@ -376,8 +376,8 @@ def _verify_downloaded(metadata: dict[Path, Any]) -> None:
     log.debug("Verification of downloaded files has started.")
 
     def raise_exception(message: str) -> None:
-        raise PackageRejected(
-            f"Some RPM packages or sources weren't verified after being downloaded: '{message}'",
+        raise ChecksumVerificationFailed(
+            filename=message,
```
eskultety: This looks fishy - you're assigning a message string to a filename argument. Not only that, but if you look at this helper's callers you'll see `ChecksumVerificationFailed` is not universally applicable to all error paths.

hermeto/core/errors.py
eskultety: Yeah, the constructor signature needs to be thought through a bit more, something like
```
 def __init__(
          self,
          files: Path | str | Iterable[Path | str],
          *,
          solution: str | None = _argument_not_specified,
      ) -> None:
          # Normalize to list of Paths
          if isinstance(files, (str, Path)):
              self.missing_files = [Path(files)]
          else:
              self.missing_files = [Path(f) for f in files]

          reason = f"Required files not found: {', '.join(str(f) for f in self.missing_files)}"
          if solution == _argument_not_specified:
              solution =  f"Make sure the files exist and are checked into the repository"
 
```

tests/unit/package_managers/pip/test_main.py
```diff
@@ -1040,11 +1040,9 @@ def test_resolve_pip_invalid_req_file_path(
 ) -> None:
     mock_metadata.return_value = ("foo", "1.0")
     invalid_path = Path("foo/bar.txt")
-    expected_error = (
-        f"The requirements file does not exist: {rooted_tmp_path.join_within_root(invalid_path)}"
-    )
+    expected_error = f"Required files not found: '{rooted_tmp_path.path / 'foo/bar.txt'}'"
```
eskultety: We should no longer match the error string in tests, this is one of still many occurrences in commit 1.

hermeto/core/package_managers/yarn_classic/main.py
eskultety: ^This comment hasn't been addressed yet.

hermeto/core/errors.py
eskultety: ^This comment hasn't been addressed yet.

shokakucarrier: fixed

hermeto/core/package_managers/pip/requirements.py
shokakucarrier: replaced with ChecksumVerificationFailed and keeping original remind text

hermeto/core/errors.py
shokakucarrier: fixed, with little change of this constructor

hermeto/core/package_managers/pip/requirements.py
eskultety: Maybe we could try emitting  `ChecksumVerificationFailed` in this case? Non-actionable,  `ChecksumVerificationFailed` isn't quite structured for this particular error as it's currently defined it seems.

shokakucarrier:
It seems everything is ready, wdyt?

hermeto/core/package_managers/rpm/main.py
eskultety: This has not been resolved in its entirety - `ChecksumVerificationFailed` is not the correct error to emit when "Unexpected file size of " occurs.

hermeto/core/errors.py
eskultety: > fixed

You still document 'reason' as an input argument even though it's not part of the constructor signature.

hermeto/core/errors.py
```diff
@@ -116,6 +117,124 @@ class UnsupportedFeature(UsageError):
...
+        "Please ensure all dependencies have their checksums listed to verify integrity"
+    )
+
+
+class LockfileNotFound(PackageRejected):
+    """A required lockfile was not found."""
+
+    def __init__(
+        self,
+        files: Path | str | Iterable[Path | str],
```
a-ovchinnikov, line 220: Either the type of `files` or the code must be adjusted:
```python
>>> from typing import Iterable
>>> files = "foobar"
>>> isinstance(files, Iterable)
True
>>> if not isinstance(files, Iterable):
...     files = [files]
...     
>>> files
'foobar'
>>> ", ".join(str(f) for f in files)
'f, o, o, b, a, r'
```

eskultety: Yep. The type is fine, but the condition is not. Note that in my original proposal this was correctly handled by converting to a Path object first.

hermeto/core/package_managers/pip/requirements.py
```diff
@@ -652,7 +656,8 @@ def validate_requirements_hashes(requirements: list[PipRequirement], require_has
 
     :param list[PipRequirement] requirements: All requirements from a file
     :param bool require_hashes: True if hashes are required for all requirements
-    :raise PackageRejected: If hashes are missing or have invalid format
+    :raise InvalidChecksum: If hashes are missing
+    :raise ChecksumVerificationFailed: If hashes have invalid format
```
a-ovchinnikov: This does not seem to be correct, this function does not raise this error.

hermeto/core/errors.py
shokakucarrier: I added one more condition: or isinstance(files, str) to fix this
or is it better to change test case so Mock will stop complain TypeError?

hermeto/core/package_managers/pip/requirements.py
shokakucarrier: I've changed MissingChecksum to InvalidChecksum to include both situation

hermeto/core/errors.py
```diff
@@ -116,6 +117,121 @@ class UnsupportedFeature(UsageError):
...
+        *,
+        solution: str | None = _argument_not_specified,
+    ) -> None:
+        """Initialize LockfileNotFound.
+
+        :param files: Path(s) where lockfile was expected
+        :param solution: politely suggest a potential solution to the user
+        """
+        if isinstance(files, (str)) or not isinstance(files, Iterable):
```
a-ovchinnikov: Minor nitpick: this could be simplified a bit:
```python
if isinstance(files, (str, Path)):
    files = [files]
```
(since you are already checking against one of two possible non-iterable types anyway).

tests/unit/package_managers/rpm/test_main.py
```diff
@@ -233,9 +235,12 @@ def test_resolve_rpm_project_invalid_lockfile_format(rooted_tmp_path: RootedPath
             },
             f,
         )
-    with pytest.raises(PackageManagerError) as exc_info:
+    with pytest.raises(InvalidLockfileFormat) as exc_info:
         _resolve_rpm_project(rooted_tmp_path, rooted_tmp_path)
-    assert f"RPM lockfile '{DEFAULT_LOCKFILE_NAME}' format is not valid" in str(exc_info.value)
+    assert (
```
a-ovchinnikov: This is an assertion on a string, please remove it (because you check against a narrow exception type on L238).

tests/unit/package_managers/rpm/test_main.py
```diff
@@ -127,29 +131,27 @@ def _has_multiple_options(rpm_models: list[RpmPackageInput]) -> bool:
...
 
     with open(rooted_tmp_path.join_within_root("rpms.lock.yaml"), "w") as f:
         # end of line is missing between items
         f.write("lockfileVendor: redhat lockfileVersion: 1\narches:\n")
-    with pytest.raises(PackageRejected) as exc_info:
+    with pytest.raises(InvalidLockfileFormat) as exc_info:
         _resolve_rpm_project(rooted_tmp_path, rooted_tmp_path)
-    assert f"RPM lockfile '{DEFAULT_LOCKFILE_NAME}' yaml format is not correct" in str(
-        exc_info.value
+    assert (
```
a-ovchinnikov: This is an assertion on a string value, please  remove it as well (because you have already made an assertion about a narrow exception type on L150).

tests/unit/package_managers/rpm/test_main.py
```diff
@@ -8,7 +8,12 @@
 import yaml
 
 from hermeto import APP_NAME
-from hermeto.core.errors import PackageManagerError, PackageRejected
+from hermeto.core.errors import (
+    ChecksumVerificationFailed,
+    InvalidLockfileFormat,
+    LockfileNotFound,
+    PackageRejected,
+)
```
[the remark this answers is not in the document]
a-ovchinnikov: Please remove this unused import.

hermeto/core/package_managers/pip/requirements.py
```diff
@@ -13,7 +13,13 @@
...
 
 from hermeto import APP_NAME
-from hermeto.core.errors import PackageRejected, UnexpectedFormat, UnsupportedFeature
+from hermeto.core.errors import (
+    ChecksumVerificationFailed,
+    InvalidChecksum,
+    PackageRejected,
+    UnexpectedFormat,
+    UnsupportedFeature,
+)
```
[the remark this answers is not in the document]
a-ovchinnikov: Please remove this unused import.

hermeto/core/checksum.py
```diff
@@ -74,7 +74,7 @@ def from_sri(cls, sri: str) -> "ChecksumInfo":
             _, alg, val = best  # type: ignore[misc]
             return ChecksumInfo(alg, base64.b64decode(val).hex())
         except (TypeError, binascii.Error) as e:
-            raise PackageRejected(
+            raise InvalidChecksum(
```
a-ovchinnikov: If I read L183 in `errors.py` correctly then a user would get this error message:
```
Checksum element: 'Integrity value is empty or malformed' is invalid
```
I would have called that a bad user experience, WDYT?

hermeto/core/errors.py
```diff
@@ -116,6 +117,121 @@ class UnsupportedFeature(UsageError):
...
+        "Verify that the file has not been corrupted and that the expected checksums are correct."
+    )
+
+
+class InvalidChecksum(PackageRejected):
+    """Provided checksum/hash/integrity is missing or it is not valid data"""
+
+    def __init__(
+        self,
+        checksum: List[str] | str | None,
```
a-ovchinnikov: The right way of dealing with this would be to introduce a dedicated type for checksums, however this is very out of scope for this PR. Maybe having an optional "reason" field would help? Then you will keep this class and dispatch on (reason, checksum) pair along these lines:
```python
    ...
    *,
    reason: str | None = None,
    ...
match (reason, checksum):
    case (None, None): reason = "Checksum is missing"
    case (None, checksum): reason = "Checksum element: f'{checksum}' is invalid"
    case (reason, None): pass
    case (reason, checksum): raise Exception("Please stop abusing this class")
```

hermeto/core/package_managers/pip/requirements.py
a-ovchinnikov: Hmm, this was apparently fixed after I created the comment. Resolving.

hermeto/core/errors.py
a-ovchinnikov: Please do not change test cases unless you absolutely must! The condition is fine and it could actually be simplified a bit.

hermeto/core/errors.py
shokakucarrier: That won't work for the test case involves Mock.mock(), if I change it to `isinstance(files, (str, Path)):` then the test case will have to change, which one do you think it is better? I don't have idea

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -3,7 +3,7 @@
 import subprocess
 from pathlib import Path
 
-from hermeto.core.errors import PackageManagerError, PackageRejected
```
eskultety, line 7: **Commit subject:**
Something happened to your commit message because the message body got concatenated with the subject, i.e.:
```
$ git log --oneline
ea5ae69a7 (HEAD) Add multiple exception class in errors.py This introduces better granularity into our existing exceptions by adding more PackageRejected exception subtypes to handle the following common error scenarios:
908c20a91 (origin/main, origin/HEAD, main) fix: Upgrade createrepo-c to v1.2.2
...
```
I think there's a missing empty line now.

hermeto/core/errors.py
shokakucarrier: I think it's better not to raise a Exception if both variable is not None tho, wdyt?

a-ovchinnikov: That's an example of what could be done, `(None, None)` might have legitimate meaning, I did not think long enough about it yesterday and added a default clause instead to ensure that the match is complete.

hermeto/core/errors.py
a-ovchinnikov: Oh, I see what the problem is! In this case I would have adjusted the UT a bit:
```diff
- mock_source_dir = mock.Mock()
+ mock_source_dir = mock.MagicMock()
```
This should be enough to make a mocked object look sufficiently str-like to pass the test.

hermeto/core/errors.py
shokakucarrier: I changed it back, maybe Lets keep the simplicity of `checksum` as the only variable to think of, and split the error class if there is more reference in the code and it cannot suit the code

tests/unit/package_managers/rpm/test_main.py
```diff
@@ -127,30 +130,24 @@ def _has_multiple_options(rpm_models: list[RpmPackageInput]) -> bool:
 
 
 def test_resolve_rpm_project_no_lockfile(rooted_tmp_path: RootedPath) -> None:
-    with pytest.raises(PackageRejected) as exc_info:
-        mock_source_dir = mock.Mock()
+    with pytest.raises(LockfileNotFound):
+        mock_source_dir = mock.MagicMock()
```
a-ovchinnikov, line 135: This change is necessary only if a corresponding change to `errors.py` is made, the one where you would do a check for two types instead of one: `isinstance(files, (str, Path))`. You either apply that change and also replace a Mock with a MagicMock or do not apply any of those. I won't insist on any of the options as long as you remain consistent.

If you decide to go with introducing a MagicMock and the corresponding change to `errors.py` could you please add a comment here describing why a MagickMock is necessary ? I.e. something like `# A Path or a str-like object is expected by the tested code, thus using MagicMock`.

hermeto/core/errors.py
```diff
@@ -116,6 +117,124 @@ class UnsupportedFeature(UsageError):
...
+
+        :param checksum: The invalid checksum value(s), or None if checksum is missing entirely
+        :param solution: politely suggest a potential solution to the user
+        """
+        if checksum is None:
+            reason = "Checksum is missing"
+        elif isinstance(checksum, list):
+            reason = f"Invalid checksum elements: {checksum}"
+        else:
+            reason = f"Invalid checksum: {checksum!r}"
```
eskultety: Looking at ^this change in a deeper context of its application across the code base I think we need one more exception subclass:
```
class MissingChecksum(InvalidChecksum)
```
because this introduced a user error regression on a few places (this comment is loosely related to this thread: https://github.com/hermetoproject/hermeto/pull/1203#discussion_r2718122846):
```
                 else:                                                                                                                       │  but the parametrize decorator test IDs suggest this was a mechanical refactor. The test logic is
-                    raise PackageRejected(                                                                                                  │  correct.
-                        f"{info['name']} is missing integrity checksum. It is mandatory"                                                    │
-                        f"for https dependencies.",                                                                                         │  ---
-                        solution="Please double-check provided package-lock.json that"                                                      │  Minor Nits
-                        " your dependencies specify integrity. Try to "                                                                     │
+                    raise InvalidChecksum(                                                                                                  │  - errors.py:250: Redundant type annotation:
+                        None,                                                                                                               │  reason: str = f"lockfile..."  # `: str` is unnecessary
+                        solution="Please double-check provided package-lock.json that "                                                     │  - Some custom solution= arguments duplicate the default_solution content with minor variations—could
+                        "your dependencies specify integrity. Try to "                                                                      │  simplify.
                         "rerun `npm install` on your repository.",                                                                          │
                     )                                                               
```
from L518. So, previously we reported the offending dependency as a reason, but now we don't report anything. If a checksum is missing, don't we want to report the offending entry the same way we report an invalid checksum? I think we do, but the current interface doesn't allow that due to the special meaning of None. I think if we subclass this further to have all the data unless we want to be super creative with the interface.
There's another occurrence in the code of the same pattern (passing None) where we previously reported the offending dependency, but now that information moved to `solution=`. Note that `solution=` is pretty much just noise, it should be approached as something auxiliary only, not a data carrying element.

hermeto/core/package_managers/gomod.py
```diff
@@ -703,9 +704,11 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
     if invalid_gomod_files:
         invalid_files_print = "; ".join(str(file.parent) for file in invalid_gomod_files)
 
-        raise PackageRejected(
-            f"The go.mod file must be present for the Go module(s) at: {invalid_files_print}",
-            solution="Please double-check that you have specified correct paths to your Go modules",
+        raise LockfileNotFound(
+            files=Path(invalid_files_print),
```
eskultety: I don't think this is correct - `invalid_files_print` is a `;` separated list of files that you convert to a Path object. I think we should be passing an iterable in this case.

hermeto/core/errors.py
```diff
@@ -1,5 +1,6 @@
 import textwrap
-from typing import ClassVar
+from pathlib import Path
+from typing import ClassVar, Iterable, List
```
eskultety: nitpick: You don't need to import `List`, we use Python 3.10 and we have converted to Python native keywords for types, i.e. `list`, `tuple`, `dict`, etc.

hermeto/core/errors.py
```diff
@@ -116,6 +117,124 @@ class UnsupportedFeature(UsageError):
...
+        "Verify that the file has not been corrupted and that the expected checksums are correct."
+    )
+
+
+class InvalidChecksum(PackageRejected):
+    """Provided checksum/hash/integrity is missing or it is not valid data."""
+
+    def __init__(
+        self,
+        checksum: List[str] | str | None = None,
```
eskultety: nitpick: Same comment about `List` vs `list` as above.

hermeto/core/errors.py
```diff
@@ -116,6 +117,124 @@ class UnsupportedFeature(UsageError):
...
+        *,
+        solution: str | None = _argument_not_specified,
+    ) -> None:
+        """Initialize InvalidLockfileFormat.
+
+        :param lockfile_path: Path to the invalid lockfile
+        :param err_details: Details about what is invalid
+        :param solution: politely suggest a potential solution to the user
+        """
+        reason: str = f"lockfile '{lockfile_path}' format is not valid: {err_details}"
```
eskultety: nitpick: unnecessary `str` type declaration, it'll be derived automatically.

hermeto/core/errors.py
shokakucarrier: In this case I've added `"{info['name']}"` as variable to `InvalidChecksum`
Then it will hint user Invalid checksum "{info['name']}"
is this acceptable?

eskultety: > Then it will hint user Invalid checksum "{info['name']}" is this acceptable?

I believe ^this is easily solved with a nice error message with my proposal above:

> `class MissingChecksum(InvalidChecksum)`

shokakucarrier: OK, I've added MissingChecksum, and did some adjustment on InvalidChecksum

PR #1204, opened 2025-12-08
gomod: Taint the SBOM with failed checks in permissive mode

slimreaper35:

### DIFF
--- a/hermeto/core/package_managers/gomod.py
+++ b/hermeto/core/package_managers/gomod.py
@@ -32,7 +32,7 @@ from hermeto.core.errors import FetchError, PackageManagerError, PackageRejected
 from hermeto.core.models.input import Mode, Request
 from hermeto.core.models.output import EnvironmentVariable, RequestOutput
 from hermeto.core.models.property_semantics import PropertySet
-from hermeto.core.models.sbom import Component
+from hermeto.core.models.sbom import Annotation, Component, spdx_now
 from hermeto.core.rooted_path import RootedPath
 from hermeto.core.scm import get_repo_for_path, get_repo_id
 from hermeto.core.type_aliases import StrPath
@@ -649,6 +649,29 @@ def _list_toolchain_files(dir_path: str, files: list[str]) -> list[str]:
     return [file for file in files if is_a_toolchain_path(Path(dir_path) / file)]
 
 
+def _update_sbom_annotations(components: list[Component], annotations: list[Annotation]) -> None:
+    """
+    Generate the bom-ref field for provided components because each annotation needs to have
+    a subjects list that references the affected components.
+
+    If the annotations list is empty, a new annotation is created. Otherwise, the existing annotation
+    is extended with more subjects. There is only one permissive mode use case for gomod at the moment.
+    """
+    subjects = [c.update_bom_ref().bom_ref for c in components]
+    if annotations:
+        # NOTE: mypy thinks that the subject argument contains None values
+        annotations[0].subjects.extend(subjects)  # type: ignore
+    else:
+        annotations.append(
+            Annotation(
+                subjects=subjects,
+                annotator={"organization": {"name": "red hat"}},
+                timestamp=spdx_now(),
+                text="hermeto:permissive-mode:gomod:vendor-directory-changed-after-vendoring",
+            )
+        )
+
+
 def fetch_gomod_source(request: Request) -> RequestOutput:
     """
     Resolve and fetch gomod dependencies for a given request.
@@ -680,6 +703,7 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
         )
 
     components: list[Component] = []
+    annotations: list[Annotation] = []
 
     repo_name = _get_repository_name(request.source_dir)
     version_resolver = ModuleVersionResolver.from_repo_path(request.source_dir)
@@ -712,6 +736,19 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
                 log.error("Failed to fetch gomod dependencies")
                 raise
 
+            vendor_changed = _vendor_changed(main_module_dir, request.mode)
+            if vendor_changed and request.mode == Mode.STRICT:
+                raise PackageRejected(
+                    reason=(
+                        "The content of the vendor directory is not consistent with go.mod. "
+                        "Please check the logs for more details."
+                    ),
+                    solution=(
+                        "Please try running `go mod vendor` and committing the changes.\n"
+                        "Note that you may need to `git add --force` ignored files in the vendor/ dir."
+                    ),
+                )
+
             main_module = _create_main_module_from_parsed_data(
                 main_module_dir, repo_name, resolve_result.parsed_main_module
             )
@@ -730,8 +767,14 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
 
             packages = _create_packages_from_parsed_data(modules, resolve_result.parsed_packages)
 
-            components.extend(module.to_component() for module in modules)
-            components.extend(package.to_component() for package in packages)
+            module_components = [module.to_component() for module in modules]
+            package_components = [package.to_component() for package in packages]
+            subpath_components = module_components + package_components
+
+            if vendor_changed:
+                _update_sbom_annotations(subpath_components, annotations)
+
+            components.extend(subpath_components)
 
         tmp_download_cache_dir = Path(tmp_dir.name).joinpath("pkg/mod/cache/download")
         if tmp_download_cache_dir.exists():
@@ -755,12 +798,16 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
     }
     env_vars_template.update(config.gomod.environment_variables)
 
+    if annotations:
+        # Update the bom-ref field for all components for consistency when permissive mode is used.
+        components = [c.update_bom_ref() for c in components]
+
     return RequestOutput.from_obj_list(
         components=components,
         environment_variables=[
             EnvironmentVariable(name=key, value=value) for key, value in env_vars_template.items()
         ],
-        project_files=[],
+        annotations=annotations,
     )
 
 
@@ -1080,7 +1127,7 @@ def _resolve_gomod(
 
     # Vendor dependencies if the gomod-vendor flag is set
     if should_vendor:
-        downloaded_modules = _vendor_deps(go, app_dir, bool(go_work), request.mode, run_params)
+        downloaded_modules = _vendor_deps(go, app_dir, bool(go_work), run_params)
     else:
         log.info("Downloading the gomod dependencies")
         downloaded_modules = (
@@ -1666,7 +1713,6 @@ def _vendor_deps(
     go: Go,
     context_dir: RootedPath,
     has_workspace: bool,
-    enforcing_mode: Mode,
     run_params: dict[str, Any],
 ) -> Iterable[ParsedModule]:
     """
@@ -1683,21 +1729,8 @@ def _vendor_deps(
     :raise UnexpectedFormat: if application fails to parse vendor/modules.txt
     """
     log.info("Vendoring the gomod dependencies")
-
     cmdscope = "work" if has_workspace else "mod"
     go([cmdscope, "vendor"], run_params)
-    if _vendor_changed(context_dir, enforcing_mode):
-        if enforcing_mode == Mode.STRICT:
-            raise PackageRejected(
-                reason=(
-                    "The content of the vendor directory is not consistent with go.mod. "
-                    "Please check the logs for more details."
-                ),
-                solution=(
-                    "Please try running `go mod vendor` and committing the changes.\n"
-                    "Note that you may need to `git add --force` ignored files in the vendor/ dir."
-                ),
-            )
     return _parse_vendor(context_dir)
 
 
--- a/tests/unit/package_managers/test_gomod.py
+++ b/tests/unit/package_managers/test_gomod.py
@@ -1431,45 +1431,18 @@ def test_invalid_local_replacements(tmpdir: Path) -> None:
         _validate_local_replacements(modules, app_path)
 
 
-@pytest.mark.parametrize("enforcing_mode", [Mode.STRICT, Mode.PERMISSIVE])
 @pytest.mark.parametrize("go_vendor_cmd", ["mod", "work"])
 @mock.patch("hermeto.core.package_managers.gomod.Go._run")
-@mock.patch("hermeto.core.package_managers.gomod._vendor_changed")
 def test_vendor_deps(
-    mock_vendor_changed: mock.Mock,
     mock_run_cmd: mock.Mock,
     go_vendor_cmd: str,
-    enforcing_mode: Mode,
     rooted_tmp_path: RootedPath,
 ) -> None:
     app_dir = rooted_tmp_path.join_within_root("some/module")
     run_params = {"cwd": app_dir}
-    mock_vendor_changed.return_value = False
-
-    # Test that vendor-changes == True in permissive mode also leads to a success path
-    # [ruff] - skip formatting because it would remove the parentheses making it less readable
-    mock_vendor_changed.return_value = (enforcing_mode != Mode.STRICT)  # fmt: skip
-
-    _vendor_deps(Go(), app_dir, go_vendor_cmd == "work", enforcing_mode, run_params)
+    _vendor_deps(Go(), app_dir, go_vendor_cmd == "work", run_params)
 
     mock_run_cmd.assert_called_once_with([GO_CMD_PATH, go_vendor_cmd, "vendor"], **run_params)
-    mock_vendor_changed.assert_called_once_with(app_dir, enforcing_mode)
-
-
-@mock.patch("hermeto.core.package_managers.gomod.Go._run")
-@mock.patch("hermeto.core.package_managers.gomod._vendor_changed")
-def test_vendor_deps_fail(
-    mock_vendor_changed: mock.Mock,
-    mock_run_cmd: mock.Mock,
-    rooted_tmp_path: RootedPath,
-) -> None:
-    app_dir = rooted_tmp_path.join_within_root("some/module")
-    run_params = {"cwd": app_dir}
-    mock_vendor_changed.return_value = True
-
-    msg = "The content of the vendor directory is not consistent with go.mod."
-    with pytest.raises(PackageRejected, match=msg):
-        _vendor_deps(Go(), app_dir, False, Mode.STRICT, run_params)
 
 
 def test_parse_vendor(rooted_tmp_path: RootedPath, data_dir: Path) -> None:
@@ -1809,7 +1782,9 @@ def test_missing_gomod_file(
 @mock.patch("hermeto.core.package_managers.gomod.ModuleVersionResolver.from_repo_path")
 @mock.patch("hermeto.core.package_managers.gomod._select_toolchain")
 @mock.patch("hermeto.core.package_managers.gomod._get_go_work_path")
+@mock.patch("hermeto.core.package_managers.gomod._vendor_changed")
 def test_fetch_gomod_source(
+    mock_vendor_changed: mock.Mock,
     mock_get_go_work_path: mock.Mock,
     mock_select_toolchain: mock.Mock,
     mock_version_resolver: mock.Mock,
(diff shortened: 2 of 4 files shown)


hermeto/core/package_managers/gomod.py
```diff
@@ -760,12 +760,27 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
...
+    annotations = []
+    if request.mode == Mode.PERMISSIVE:
+        for c in components:
+            c.update_bom_ref()
+
+        annotations.append(
+            Annotation(
+                subjects=[c.bom_ref for c in components],
+                annotator={"organization": {"name": "red hat"}},
+                timestamp=spdx_now(),
```
a-ovchinnikov: It was a long time ago, but I think there was some timestamp surgery necessary for matching SPDX expectations. I would recommend renaming this function if it is to be used for anything other than SPDX.

hermeto/core/package_managers/gomod.py
```diff
@@ -760,12 +760,27 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
     }
     env_vars_template.update(config.default_environment_variables.get("gomod", {}))
 
+    annotations = []
+    if request.mode == Mode.PERMISSIVE:
```
a-ovchinnikov: Could you please add a comment here, something like `If we reached this spot in PERMISSIVE mode then we can be sure that vendor directory was changed after vendoring`? Otherwise it looks like we taint indiscriminately.

tests/integration/utils.py
```diff
@@ -561,6 +561,25 @@ def _replace_tmp_path_with_placeholder(
         item["abspath"] = "${test_case_tmp_path}/" + str(relative_path)
 
 
+def _normalize_timestamps(json_obj: Any, fixed_timestamp: str = "2025-01-01T00:00:00Z") -> Any:
```
eskultety: nitpick: not sure if normalize is the best word when it's just a straight out string replacement, but I can live with it :).

Other than that - why are you adding fixed_timestamp as an argument if you don't intend to set it (as you shouldn't it should be a constant as far as tests are concerned).

hermeto/core/package_managers/gomod.py
eskultety: Sorry, I don't get the commentary - what is it that is ensuring us that the vendor directory changed after vendoring? I'm missing the clear indicator here once we jumped out of the `GoCacheTemporaryDirectory` context manager.

hermeto/core/models/sbom.py
```diff
@@ -130,11 +130,9 @@ class Metadata(pydantic.BaseModel):
     tools: list[Tool] = [Tool(vendor="red hat", name=f"{APP_NAME}")]
 
 
-def spdx_now() -> str:
-    """Return a time stamp in SPDX-compliant format.
-
-    See https://spdx.github.io/spdx-spec/v2.3/search.html?q=date
-    for details.
+def timestamp_now() -> str:
```
eskultety: I'm not 100% convinced this is the right thing to do if you want to get a timestamp, because you don't need a dedicated function for it per-se. The sheer need for `spdx_now` stems from the fact that we needed an **explicit** string format to comply with the SPDX spec. Therefore, if you need a timestamp outside of SPDX's context, just use `datetime.datetime.now` directly. For me, keeping `spdx_now` is a very clear indicator that we need to something slightly differently compared to just calling `datetime` with its default settings and be done with it.

Calling `timestamp_now` might raise eyebrows for whoever reads the code and already forgot this context and might leave them puzzled about why we need a wrapper to get a timestamp in the first place. So we should either keep `spdx_now` and simply accept the reality OR if you want to rename it, okay, but then you **must** keep the original function doc string and extend it with something like: _Use the same timestamp format in CycloneDX for consistency as well_ . And also the trivia in the commit message.

hermeto/core/package_managers/gomod.py
slimreaper35: Hmm, so what should we do when the permissive mode is enabled but the vendor directory wasn't changed ? The vendoring would pass even in the strict mode.

hermeto/core/models/sbom.py
```diff
@@ -23,6 +23,21 @@
...
+
+class Annotation(pydantic.BaseModel):
+    """
+    A comment, note, explanation, or similar textual content
+    which provides additional context to the object(s) being annotated.
+    """
+
+    subjects: list[str]
+    annotator: dict[Annotator, dict[str, str]]
+    timestamp: str
```
a-ovchinnikov: `srt` is a very wide type for a timestamp. Ideally it should be narrowed down. What about `datetime.datetime`? [Pydantic's view on this](https://docs.pydantic.dev/2.0/usage/types/datetime/)

hermeto/core/models/output.py
a-ovchinnikov, line 1: Comment on 135819f3f45878af6dc8fe5967e66e5350e2e83d's commit message:

Please consider something along the lines of "Extend SBOM model with Annotations". Otherwise it is a bit ambiguous and might look like you annotated the code. Maybe  even adding some redundancy like "This commit makes use of Annotation class introduced earlier and extends SBOM model with 'annotations' field." won't hurt.

hermeto/core/package_managers/gomod.py
a-ovchinnikov: >  so what should we do when the permissive mode is enabled but the vendor directory wasn't changed

At the very least we should not be marking this component as having a changed vendor directory. This would require returning a flag from `_vendor_deps` along with downloaded modules. Ideally we should treat this as an error condition: permissive mode is meant to be a workaround for a specific problem and one should not just use it, however I am not sure that we cannot have a situation when modified vendor directory coexists with non-modified within one request.

tests/unit/package_managers/test_gomod.py
```diff
@@ -1431,45 +1431,18 @@ def test_invalid_local_replacements(tmpdir: Path) -> None:
...
-    _vendor_deps(Go(), app_dir, go_vendor_cmd == "work", enforcing_mode, run_params)
+    _vendor_deps(Go(), app_dir, go_vendor_cmd == "work", run_params)
 
     mock_run_cmd.assert_called_once_with([GO_CMD_PATH, go_vendor_cmd, "vendor"], **run_params)
-    mock_vendor_changed.assert_called_once_with(app_dir, enforcing_mode)
-
-
-@mock.patch("hermeto.core.package_managers.gomod.Go._run")
-@mock.patch("hermeto.core.package_managers.gomod._vendor_changed")
-def test_vendor_deps_fail(
```
a-ovchinnikov, line 1461: We would probably still want to check if a change in vendored directory was processed correctly. On one hand this possible change stems from an external dependency, so it is better be tested with an integration test, on the other a UT could be used as a smoke test for now.

hermeto/core/package_managers/gomod.py
```diff
@@ -712,6 +713,19 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
                 log.error("Failed to fetch gomod dependencies")
                 raise
 
+            vendor_changed = _vendor_changed(main_module_dir, request.mode)
+            if vendor_changed and request.mode == Mode.STRICT:
+                raise PackageRejected(
```
a-ovchinnikov, line 741: Minor suggestion, not a blocker: the exception could be pre-instantiated on the module level and just raised here. That could save some vertical space in the function.

hermeto/core/package_managers/gomod.py
```diff
@@ -730,8 +744,23 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
...
 
-            components.extend(module.to_component() for module in modules)
-            components.extend(package.to_component() for package in packages)
+            module_components = [module.to_component() for module in modules]
+            package_components = [package.to_component() for package in packages]
+
+            components.extend(chain(module_components, package_components))
+
+            if vendor_changed:
+                for c in chain(module_components, package_components):
```
a-ovchinnikov: Minor suggestion, not a blocker: L753-L763 could be tucked away into a helper function.

hermeto/core/package_managers/gomod.py
```diff
@@ -32,7 +32,7 @@
 from hermeto.core.models.input import Mode, Request
```
eskultety, line 32: **Commit message:**
```
s/mutliple/multiple
s/it make/it makes
s/more a bit more/a bit more
```

hermeto/core/package_managers/gomod.py
eskultety: Maybe when we adopt more of these it could make sense to do it on the global level, but I'd keep it like this for now.

hermeto/core/package_managers/gomod.py
eskultety: I don't get the iterable chaining twice bit - `components` is `[]` here, so why do we need to chain the iterables twice instead of just looping over updated components once?

You only ever reference the "updated" components list in the annotations constructor, that's confusing from reader's POV IMO. Is it because `update_bom_ref` is inconvenient and returns None? If so then the standard practice on self-inflicted update methods the convention is to return `self` which did the trick for me. I did this locally and it passed the integration tests:
```
diff --git hermeto/core/models/sbom.py hermeto/core/models/sbom.py                                                                           │  - Potentially network I/O (fetch, clone)
index 9d174cd37..1de081852 100644                                                                                                            │  - Takes milliseconds to seconds
--- hermeto/core/models/sbom.py                                                                                                              │
+++ hermeto/core/models/sbom.py                                                                                                              │  The lambda overhead is ~100-200 nanoseconds. The git operation is ~10-1000 milliseconds. That's a 5-7 orders of magnitude difference.
@@ -101,7 +101,7 @@ class Component(pydantic.BaseModel):                                                                                     │
         """                                                                                                                                 │  You'd need to call _git_operation millions of times in a tight loop to notice. In practice, a typical hermeto run does maybe dozens of git
         return self.purl                                                                                                                    │operations total.
                                                                                                                                             │
-    def update_bom_ref(self) -> None:                                                                                                       │  TL;DR: No practical penalty. Micro-optimization territory that doesn't matter here.
+    def update_bom_ref(self) -> Self:                                                                                                       │
         """                                                                                                                                 │> what about debugging?
         Update the `bom-ref` field of this component.                                                                                       │
                                                                                                                                             │● Lambdas have debugging downsides:
@@ -109,6 +109,7 @@ class Component(pydantic.BaseModel):                                                                                     │
         initialized by default and needs to be updated manually with this method.                                                           │  1. Stack traces show <lambda> instead of meaningful names:
         """                                                                                                                                 │  Traceback (most recent call last):
         self.bom_ref = self.purl                                                                                                            │    File "scm.py", line 45, in _git_operation
+        return self                                                                                                                         │      return func()
                                                                                                                                             │    File "scm.py", line 62, in <lambda>    # ← Not helpful
     @pydantic.field_validator("properties")                                                                                                 │      return self._git_operation("remote", lambda: self._repo.remote(name))
     @classmethod                                                                                                                            │
diff --git hermeto/core/package_managers/gomod.py hermeto/core/package_managers/gomod.py                                                     │  vs our direct approach:
index a0c11b2d2..171846517 100644                                                                                                            │  Traceback (most recent call last):
--- hermeto/core/package_managers/gomod.py                                                                                                   │    File "scm.py", line 130, in get_origin_url
+++ hermeto/core/package_managers/gomod.py                                                                                                   │      origin = self._repo.remote("origin")    # ← Clear context
@@ -750,12 +750,9 @@ def fetch_gomod_source(request: Request) -> RequestOutput:                                                              │
             components.extend(chain(module_components, package_components))                                                                 │  2. Extra stack frames - Stepping through in a debugger goes: remote() → _git_operation() → <lambda> → actual call. More frames to navigate.
                                                                                                                                             │
             if vendor_changed:                                                                                                              │  3. Breakpoints - Can't easily set a breakpoint "inside" a lambda. You'd set it on _git_operation and hit it for every git operation, not ju
-                for c in chain(module_components, package_components):                                                                      │st the one you care about.
-                    c.update_bom_ref()                                                                                                      │
-                                                                                                                                            │  4. Closure inspection - Some debuggers show captured variables less clearly than local variables.
                 annotations.append(                                                                                                         │
                     Annotation(                                                                                                             │  ---
-                        subjects=[c.bom_ref for c in components],                                                                           │  For git operations that fail rarely and are few in number, it's a minor annoyance. But if you're debugging a tricky git issue, the explicit
+                        subjects=[c.bom_ref for c in map(lambda c: c.update_bom_ref(), components)],                                        │ try-except with direct calls gives cleaner traces and easier stepping.
                         annotator={"organization": {"name": "red hat"}},                                                                    │
                         timestamp=spdx_now(),                                                                                               │> /model
                         text="hermeto:permissive-mode:gomod:vendor-directory-changed-after-vendoring",

File: tests/unit/package_managers/test_gomod.py (hunk 6 above)
Comment (eskultety) [reply tests/unit/package_managers/test_gomod.py:1461]: Agreed with the assessment.

File: hermeto/core/package_managers/gomod.py (hunk 7 above)
Comment (a-ovchinnikov) [reply hermeto/core/package_managers/gomod.py:741]: I think `raise PackageRejectedDueToVendoredDirChange` is an easier to read mouthful than a full instantiation with reason and suggestions, but that's a matter of taste, I won't block on this.

File: hermeto/core/package_managers/gomod.py (hunk 8 above)
Comment (slimreaper35) [reply hermeto/core/package_managers/gomod.py]: > I don't get the iterable chaining twice bit - components is [] here, so why do we need to chain the iterables twice instead of just looping over updated components once?

Because, `components` is shared for all "request.gomod_packages", so we report only affected packages.

Alternatively, I could concat `module_components` and `package_components` to a list.

File: tests/unit/package_managers/test_gomod.py (hunk 6 above)
Comment (slimreaper35) [reply tests/unit/package_managers/test_gomod.py:1461]: What kind of integration test ?

File: hermeto/core/package_managers/gomod.py (hunk 8 above)
Comment (eskultety) [reply hermeto/core/package_managers/gomod.py]: > Because, components is shared for all "request.gomod_packages", so we report only affected packages.

Right, but what confuses me is how you actually report only affected packages with multiple subpaths? Simply based on the following line:
```
...
subjects=[c.bom_ref for c in components]
```
where you iterate over the whole components, but a few lines earlier you iterated over a subset of its content.
I found the answer in the meantime (took me a while) on why the `None`s never appeared in the list and even in a multi-project setup this would still work with multiple subpaths where only some would make use of vendoring --> `exclude_none=True` on the SBOM formatter!

For this patch, please make sure you construct `subjects` only from the subset of components you constructed a few lines above, otherwise it's completely opaque to the reader because the fact a bunch of `None` values will not make it to the output is buried somewhere in the `cli.py` module where noone would look for it.

I'd also suggest to construct `subjects` using the `map` proposal I did (minus using components) which will create an immediate context for anyone reading the code, but I'll leave this as an optional suggestion.

Next, for the `bom-ref` design. It's true the cyclonedx spec documents bom-ref as an optional field, but from my POV it's incredibly inconsistent if some components have it and some don't and using `bom-ref` only as an indicator of annotations is an unfortunate design decision IMO. I'd much rather prefer if we specify bom-ref for all components IF we have to use it and not using it if we don't OR just always use the bom-ref in all scenarios, either one works for me. 

Lastly, the fact that `bom-ref == purl` is also questionable, but I agree it was an easy decision to ensure uniqueness, but at the same time it's pretty much just a bookmark in a sense so it can be named anything. We released it yesterday, but since it's just a link the content should not matter that much even if we change it before the next major release, at least I don't think that the field's content is binding for us such that we couldn't use something more readable for the ref ----> also not immediately actionable, just food for thought.

File: tests/unit/package_managers/test_gomod.py (hunk 6 above)
Comment (eskultety) [reply tests/unit/package_managers/test_gomod.py:1461]: I believe one that uses an artificially created dependency that would not have an updated vendor directory?

Comment (eskultety) [reply tests/unit/package_managers/test_gomod.py:1461]: But keeping the unit test is also fine with me for now, the integration test is not a blocker for me, which wasn't clear from my "agreed with the assessment" message.

Comment (slimreaper35) [reply tests/unit/package_managers/test_gomod.py:1461]: We have some tests from 5 years ago - https://github.com/hermetoproject/integration-tests/branches/all?query=vendor&lastTab=overview so we could revisit them.

File: hermeto/core/package_managers/gomod.py (hunk 8 above)
Comment (slimreaper35) [reply hermeto/core/package_managers/gomod.py]: > Right, but what confuses me is how you actually report only affected packages with multiple subpaths? Simply based on the following line:

Ah, I missed that. It was a bug.

> I'd also suggest to construct subjects using the map proposal I did (minus using components) which will create an immediate context for anyone reading the code, but I'll leave this as an optional suggestion.

Done.

> I'd much rather prefer if we specify bom-ref for all components IF we have to use it and not using it if we don't 

Done. Agree.

> Lastly, the fact that bom-ref == purl is also questionable, but I agree it was an easy decision to ensure uniqueness, but at the same time it's pretty much just a bookmark in a sense so it can be named anything.

I wish I had proposed a larger POC alongside the design document. A lot of questions pop up in this and previous PR that could be resolved earlier.

File: tests/unit/package_managers/test_gomod.py (hunk 6 above)
Comment (eskultety) [reply tests/unit/package_managers/test_gomod.py:1461]: Hmm, yep, the `wrong_vendor` test looks like a good candidate to start with, but of course it lacks any description and I haven't looked at the test case.

File: hermeto/core/package_managers/gomod.py [hunk 10]
```diff
@@ -649,6 +649,28 @@ def is_a_toolchain_path(path: str | os.PathLike[str]) -> bool:
...
 
+def _update_sbom_annotations(components: list[Component], annotations: list[Annotation]) -> None:
+    """
+    Generate the bom-ref field for provided components because each annotation needs to have
+    a subjects list that references the affected components.
+
+    If the annotations list is empty, a new annotation is created. Otherwise, the existing annotation
+    is extended with more subjects. There is only one permissive mode use case for gomod at the moment.
+    """
+    updated_components = map(lambda c: c.update_bom_ref(), components)
```
eskultety: you could do `subjects = [c.bom_ref for c in map(lambda c: c.update_bom_ref(), components)]` and then just assign `subjects` further on.

hermeto/core/models/sbom.py
```diff
@@ -101,14 +101,15 @@ def key(self) -> str:
         """
         return self.purl
 
-    def update_bom_ref(self) -> None:
+    def update_bom_ref(self) -> Self:
```
eskultety, line 104: Technically should be a separate commit :) .

hermeto/core/package_managers/gomod.py
```diff
@@ -755,12 +797,16 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
     }
     env_vars_template.update(config.gomod.environment_variables)
 
+    if annotations:
+        # Update the bom-ref field for all components for consistency when permissive mode is used.
+        components = list(map(lambda c: c.update_bom_ref(), components))
```
eskultety: @a-ovchinnikov tends to use a trick to create named lambdas for cases like this one, since we already have several, might be an improvement to re-use it also in `_update_sbom_annotations` :).

eskultety (approved):
nitpicks only,so I'm good FWIW.

hermeto/core/package_managers/gomod.py
a-ovchinnikov: I am sorry, why not just
```python
components = [c.update_bom_ref() for c in components]
```
?
(The same holds for `_update_sbom_annotations` with s/c.update_bom_ref()/c.update_bom_ref().bom_ref/)

eskultety: It's cleaner, works for me.

hermeto/core/package_managers/gomod.py
```diff
@@ -649,6 +649,28 @@ def is_a_toolchain_path(path: str | os.PathLike[str]) -> bool:
...
+    """
+    Generate the bom-ref field for provided components because each annotation needs to have
+    a subjects list that references the affected components.
+
+    If the annotations list is empty, a new annotation is created. Otherwise, the existing annotation
+    is extended with more subjects. There is only one permissive mode use case for gomod at the moment.
+    """
+    subjects = [c.update_bom_ref().bom_ref for c in components]
+    if annotations:
+        annotations[0].subjects.extend(subjects)  # type: ignore
```
a-ovchinnikov, line 663: I would recommend adding a short notice explaining why mypy is unhappy and why it is wrong about being unhappy.

hermeto/core/package_managers/gomod.py
```diff
@@ -730,8 +766,14 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
 
             packages = _create_packages_from_parsed_data(modules, resolve_result.parsed_packages)
 
-            components.extend(module.to_component() for module in modules)
-            components.extend(package.to_component() for package in packages)
+            module_components = (module.to_component() for module in modules)
+            package_components = (package.to_component() for package in packages)
+            subpath_components = list(chain(module_components, package_components))
```
a-ovchinnikov: Non-blocking:
```python
module_components = [module.to_component() for module in modules]
package_components = [package.to_component() for package in packages]
subpath_components = module_components + package_components
```
Reason: the generators get immediately converted to a list, so you might just start with a list.

PR #1206, opened 2025-12-10
Unify Git operations into one place

sswguo:
Implement unified Git exception handling through wrapper classes and the new
git errors:
    
      Git Errors:
      - GitError: base class for Git operations
      - GitRemoteNotFoundError: for missing remotes
      - GitInvalidRevisionError: for invalid commits/branches/tags/refs

      Wrapper classes:
      - GitRepo: wraps git.Repo
      - GitHEAD: wraps git.refs.HEAD
      - GitCommandWrapper: wraps git commands executed through repo.git.* methods

Resolves: https://github.com/hermetoproject/hermeto/issues/905

Assisted-by: Claude Code(Sonnet 4)

### DIFF
--- a/hermeto/core/scm.py
+++ b/hermeto/core/scm.py
@@ -26,6 +26,21 @@ from hermeto.core.type_aliases import StrPath
 log = logging.getLogger(__name__)
 
 
+class GitCommandWrapper(git.Git):
+    """Git command wrapper that converts GitCommandError to GitError."""
+
+    def execute(self, command: Any, *args: Any, **kwargs: Any) -> Any:
+        """Execute git command with unified error handling."""
+        try:
+            return super().execute(command, *args, **kwargs)
+        except GitCommandError as ex:
+            raise GitError(
+                f"Git command failed: {ex}",
+                stderr=ex.stderr,
+                stdout=ex.stdout,
+            ) from ex
+
+
 class GitHEAD(git.HEAD):
     """HEAD reference wrapper with unified error handling."""
 
@@ -86,6 +101,8 @@ class GitHEAD(git.HEAD):
 
 class GitRepo(git.Repo):
     """Git repository wrapper with unified error handling."""
+
+    GitCommandWrapperType = GitCommandWrapper
 
     def __init__(self, path: str | PathLike[str], *args: Any, **kwargs: Any) -> None:
         """Initialize git repository with unified error handling."""
--- a/hermeto/core/package_managers/gomod.py
+++ b/hermeto/core/package_managers/gomod.py
@@ -28,13 +28,19 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
 from hermeto.core.config import get_config
-from hermeto.core.errors import FetchError, PackageManagerError, PackageRejected, UnexpectedFormat
+from hermeto.core.errors import (
+    FetchError,
+    GitError,
+    PackageManagerError,
+    PackageRejected,
+    UnexpectedFormat,
+)
 from hermeto.core.models.input import Mode, Request
 from hermeto.core.models.output import EnvironmentVariable, RequestOutput
 from hermeto.core.models.property_semantics import PropertySet
 from hermeto.core.models.sbom import Annotation, Component, spdx_now
 from hermeto.core.rooted_path import RootedPath
-from hermeto.core.scm import get_repo_for_path, get_repo_id
+from hermeto.core.scm import GitRepo, get_repo_for_path, get_repo_id
 from hermeto.core.type_aliases import StrPath
 from hermeto.core.utils import GIT_PRISTINE_ENV, get_cache_dir, load_json_stream, run_cmd
 from hermeto.interface.logging import EnforcingModeLoggerAdapter
@@ -1355,7 +1361,7 @@ class GoCacheTemporaryDirectory(tempfile.TemporaryDirectory):
 class ModuleVersionResolver:
     """Resolves the versions of Go modules in a git repository."""
 
-    def __init__(self, repo: git.Repo, commit: git.objects.commit.Commit):
+    def __init__(self, repo: GitRepo, commit: git.objects.commit.Commit):
         """Initialize a ModuleVersionResolver for the provided Repo."""
         self._repo = repo
         self._commit = commit
@@ -1363,14 +1369,14 @@ class ModuleVersionResolver:
     @classmethod
     def from_repo_path(cls, repo_path: RootedPath) -> "Self":
         """Fetch tags from a git Repo and return a ModuleVersionResolver."""
-        repo = git.Repo(repo_path)
+        repo = GitRepo(repo_path)
         commit = repo.commit(repo.rev_parse("HEAD").hexsha)
         try:
             # Do not run 'git fetch --tags' because that fetches pretty much everything from the
             # remote. Save bandwidth and storage with an explicit refspec instead.
             # See man 1 git-fetch for the authoritative answer.
             repo.remote().fetch(refspec="+refs/tags/*:refs/tags/*", force=True)
-        except Exception as ex:
+        except GitError as ex:
             raise FetchError(
                 f"Failed to fetch the tags on the Git repository ({type(ex).__name__}) "
                 f"for {repo.working_tree_dir}: "
@@ -1400,7 +1406,7 @@ class ModuleVersionResolver:
         :param all_reachable: True to get all tags on the current commit and all commits preceding
                               it. False to get the tags on the current commit only.
         :return: a list of tag names
-        :raises GitCommandError: if failed to fetch the tags on the Git repository
+        :raises GitError: if failed to fetch the tags on the Git repository
         """
         try:
             if all_reachable:
@@ -1428,7 +1434,7 @@ class ModuleVersionResolver:
                 as_process=False,
                 stdout_as_string=True,
             ).splitlines()
-        except git.GitCommandError:
+        except GitError:
             msg = f"Failed to get the tags associated with the reference {self._commit.hexsha}"
             log.error(msg)
             raise
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -16,10 +16,10 @@ from typing import Any
 import jsonschema
 import requests
 import yaml
-from git import Repo
 
 from hermeto import APP_NAME
 from hermeto.core import resolver
+from hermeto.core.scm import GitRepo
 from hermeto.core.type_aliases import StrPath
 from hermeto.interface.cli import DEFAULT_OUTPUT
 from tests.integration.container_engine import get_container_engine
@@ -223,7 +223,7 @@ def _get_git_commit_from_tarball(tarball: Path) -> str:
         extract_path = str(tarball).replace(".tar.gz", "").replace(".tgz", "")
         _safe_extract(tarfile, extract_path)
 
-    repo = Repo(path=f"{extract_path}/app")
+    repo = GitRepo(path=f"{extract_path}/app")
     commit = f"gitcommit:{repo.commit().hexsha}"
 
     shutil.rmtree(extract_path)
@@ -339,7 +339,7 @@ def _clone_custom_test_repo(tmp_path: Path, repo_url: str, branch: str) -> Path:
     repo_dir = tmp_path / f"{repo_name}_{safe_branch}"
 
     log.info(f"Cloning custom test repository from {repo_url} (branch: {branch})")
-    Repo.clone_from(
+    GitRepo.clone_from(
         url=repo_url,
         to_path=repo_dir,
         branch=branch,
@@ -384,7 +384,7 @@ def fetch_deps_and_check_output(
         )
     else:
         actual_repo_dir = test_repo_dir
-        repo = Repo(actual_repo_dir)
+        repo = GitRepo(actual_repo_dir)
         # Submodule could end up being in detached HEAD state which would
         # result in a cascading failure for all tests that follow. This does
         # not happen always and at the moment of writing it is not clear what
@@ -594,7 +594,7 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
 
 
 def retrieve_changed_files_from_git() -> tuple[Path, ...]:
-    repo = Repo(".", search_parent_directories=True)
+    repo = GitRepo(".", search_parent_directories=True)
     # >>> type(repo.branches)
     # <class 'git.util.IterableList'>
     # Despite the fact stated above mypy does not believe one can use 'in' with
--- a/tests/unit/package_managers/test_gomod.py
+++ b/tests/unit/package_managers/test_gomod.py
@@ -58,6 +58,7 @@ from hermeto.core.package_managers.gomod import (
     fetch_gomod_source,
 )
 from hermeto.core.rooted_path import PathOutsideRoot, RootedPath
+from hermeto.core.scm import GitRepo
 from hermeto.core.utils import GIT_PRISTINE_ENV, load_json_stream
 from tests.common_utils import GIT_REF, write_file_tree
 
@@ -1393,7 +1394,7 @@ def test_get_golang_version(
     module_name = f"github.com/mprahl/test-golang-pseudo-versions{module_suffix}"
 
     module_dir = RootedPath(golang_repo_path)
-    repo = git.Repo(golang_repo_path)
+    repo = GitRepo(golang_repo_path)
     repo.git.checkout(ref)
     version_resolver = ModuleVersionResolver(repo, repo.commit(ref))
 
@@ -1924,7 +1925,7 @@ def test_fetch_tags_fail(repo_remote_with_tag: tuple[RootedPath, RootedPath]) ->
     # The remote_repo itself has no remote configured, so will fail when fetching tags
     remote_repo_path, _ = repo_remote_with_tag
     error_msg = re.escape(
-        f"Failed to fetch the tags on the Git repository (ValueError) for {remote_repo_path}"
+        f"Failed to fetch the tags on the Git repository (GitRemoteNotFoundError) for {remote_repo_path}"
     )
     with pytest.raises(FetchError, match=error_msg):
         ModuleVersionResolver.from_repo_path(remote_repo_path)


hermeto/core/scm.py
```diff
@@ -209,3 +240,108 @@ def _reset_git_head(repo: Repo, ref: str) -> None:
...
+    log.debug("Cloning git repository from %s for %s", url, config.package_manager)
+
+    try:
+        _clone_git_repo(
+            url=url,
+            to_path=to_path,
+            ref=ref,
+            branch=branch,
+        )
+        log.debug("Successfully cloned %s to %s", url, to_path)
```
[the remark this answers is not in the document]
sswguo: Updated.

hermeto/core/scm.py
```diff
@@ -156,39 +157,69 @@ def clone_as_tarball(url: str, ref: str, to_path: Path) -> None:
     :param ref: the revision to check out
     :param to_path: create the tarball at this path
     """
+    with tempfile.TemporaryDirectory(prefix="cachito-") as temp_dir:
```
eskultety: Code extractions to simplify existing functions, isolate functionality to individual helpers for easier testing, etc. should be done in a standalone commit to reduce the diff noise. Note when doing that, please make sure it's code extraction only, no other code modifications (reserve those for other commits) to keep the diff nice & clean and easy to review with `git show --color-moved`, thanks.

hermeto/core/scm.py
```diff
@@ -209,3 +240,88 @@ def _reset_git_head(repo: Repo, ref: str) -> None:
             "Failed on checking out the Git repository. Please verify the supplied reference "
             f'of "{ref}" is valid.'
         )
+
+
+class GitSupportConfig:
```
eskultety: Introduction of this class should also be (including the unit test you have) in a standalone commit to introduce the changes gradually, more granular == easier to review/revert/bisect

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -26,6 +25,12 @@
 log = logging.getLogger(__name__)
 
 
+BUNDLER_GIT_CONFIG = GitSupportConfig(
+    supported_schemes=["git", "git+http", "git+https", "git+ssh", "https", "ssh"],
```
eskultety: Bundler doesn't document supporting VCS-style protocol schemes, from https://bundler.io/guides/git.html:
```
gem 'rack', git: 'https://github.com/rack/rack.git'
gem 'rack', git: 'git@github.com:rack/rack.git'
gem 'rack', git: 'git://github.com/rack/rack.git'
```

hermeto/core/scm.py
```diff
@@ -209,3 +240,88 @@ def _reset_git_head(repo: Repo, ref: str) -> None:
...
+
+        Args:
+            supported_schemes: List of supported git URL schemes (e.g. ['git+https', 'git+ssh'])
+            package_manager: The package manager this configuration is for
+        """
+        self.supported_schemes = supported_schemes
+        self.package_manager = package_manager
+
+
+def validate_git_dependency(url: str, config: GitSupportConfig) -> None:
```
eskultety: I don't think we need this validation. This merely validates URLs, more specifically - URL schemes. Instead, this should be done on the pydantic model validation level and that's what we already do.
However, looking at our existing `AcceptedUrl = Annotated[` definition for bundler, there's a bug we need to fix as it only assumes the `https` URL scheme for git dependency URLs.

eskultety (changes requested):
Unfortunately, I'm not convinced this quite covers what's needed for https://github.com/hermetoproject/hermeto/issues/905. This doesn't look like it's unifying anything really. What I'd envision for the solution for the issue above is something that centralizes **all** git operations to a single place and wraps all possible git failures with a deterministic error-type for easier testing. To give you an idea, if you run
```
$ mkdir repo && cd repo
$ git init
$ python3 -c "from hermeto.core.scm import get_repo_id; get_repo_id('.')"
...
hermeto.core.errors.UnsupportedFeature: hermeto cannot process repositories that don't have an 'origin' remote.
```
In ^this case, let's say this is correct for now, but won't be after https://github.com/hermetoproject/hermeto/pull/1176, but I guess we can agree that even then it's not consistent as a git error, but it's acceptable.

If you do:
```
$ mkdir repo && cd repo
$ git init
$ git remote add origin <someurl>
$ python3 -c "from hermeto.core.scm import get_repo_id; get_repo_id('.')"
...
ValueError: Reference at 'refs/heads/master' does not exist
```
Because no commits exist for this artificial repo. There  are several other possible variants of the `ValueError` git commands can yield and so my idea with the issue at hand was to make sure we reliably wrap all these `ValueError`s such that we can deterministically only ever catch e.g. a custom `GitError` in our unit tests rather that guess what `ValueError` we're not handling yet.

tests/unit/test_scm.py
```diff
@@ -145,3 +153,61 @@ def test_get_repo_for_path(
...
+        with pytest.raises(
+            UnsupportedFeature, match="Unsupported VCS scheme for bundler: git\\+ssh"
+        ):
+            validate_git_dependency("git+ssh://github.com/example/repo.git", config)
+
+
+class TestCloneGitDependency:
+    """Test the clone_git_dependency function."""
+
+    def test_clone_git_dependency_bundler(self, golang_repo_path: Path, tmp_path: Path) -> None:
```
a-ovchinnikov: This calls bundler out specifically, however it does not look like the tested function actually has anything to do with bundler. I would suggest dropping any references to bundler. I would also ask for a more descriptive test name, something along the lines of `test_clone_git_dependency_can_handle_correct_repositories` or similar, but that would also beg for a negative test too.

hermeto/core/scm.py
```diff
@@ -209,3 +240,88 @@ def _reset_git_head(repo: Repo, ref: str) -> None:
...
+
+    try:
+        _clone_git_repo(
+            url=url,
+            to_path=to_path,
+            ref=ref,
+            branch=branch,
+        )
+
+    except Exception as ex:
```
a-ovchinnikov: Please correct me if I am wrong, but it looks like the only exception that could be raised here is `FetchError`. I'd argue that `Exception` is too broad and could be narrowed down.

hermeto/core/scm.py
sswguo: Sure, will make a separate commit for this.

hermeto/core/scm.py
sswguo: Sounds good, will remove this.

hermeto/core/scm.py
sswguo: Yeah, I am checking this doc https://gitpython.readthedocs.io/en/stable/reference.html to see what's the specific exception for each operation.

a-ovchinnikov: `_clone_git_repo` essentially converts `Exception` into `FetchError`. While it might make sense to keep `_clone_git_repo` catch the most generic exception possible here it looks like it should always be `FetchError`.

tests/unit/test_scm.py
sswguo: I see, will update this test as well. thanks.

hermeto/core/package_managers/bundler/gem_models.py
sswguo: Removed this config.

hermeto/core/scm.py
```diff
@@ -239,12 +255,16 @@ def clone_as_tarball(url: str, ref: str, to_path: Path) -> None:
     :param to_path: create the tarball at this path
     """
     with tempfile.TemporaryDirectory(prefix="cachito-") as temp_dir:
-        cloner = GitCloner()
-        repo = cloner.clone(url=url, to_path=Path(temp_dir), ref=ref, filter="blob:none")
+        try:
+            cloner = GitCloner()
```
a-ovchinnikov: A very minor nitpick: please move GitCloner() outside of `try:`, it does not look like it is going to raise any exceptions.

hermeto/core/scm.py
```diff
@@ -147,32 +147,51 @@ def _canonicalize_origin_url(url: str) -> str:
         )
 
 
-def clone_as_tarball(url: str, ref: str, to_path: Path) -> None:
-    """Clone a git repository, check out the specified revision and create a compressed tarball.
+class GitCloner:
```
a-ovchinnikov: I am sorry, I miss the point of this class. It does not preserve any state and serves as a wrapper for two loosely coupled methods, both of which could be either static or class methods or even remain just functions. Could you please elaborate why this needs to be a class?

hermeto/core/scm.py
```diff
@@ -117,7 +126,15 @@ def get_repo_for_path(repo_root: Path, target_path: Path) -> tuple[Repo, Path]:
     if not target_path.is_absolute():
         target_path = repo_root / target_path
 
-    current_repo = Repo(repo_root)
+    try:
+        current_repo = Repo(repo_root)
```
a-ovchinnikov: I think what would make sense here is to have a class that wraps `Repo` and converts git errors internally. I.e. the bulk of code should remain the same, but `Repo` would be an alias to `HermetoRepo` which would handle exception conversions implicitly.

sswguo: Yeah, that's a good point. And should we overwrite the methods of the `Repo` as well ? I mean the ones we used in this file like:

`repo.remote("origin")`
`commit_id = repo.head.commit.hexsha`

a-ovchinnikov: It depends on how the wrapping will be implemented. If `repo` is an instance of `Repo` and if these methods can raise a generic error (like `ValueError`) or if you end up creating a class that is not a subclass of `Repo` then yes. I would have aimed for a change that keeps most of the code intact and implicitly wraps all exceptions.

hermeto/core/scm.py
```diff
@@ -4,21 +4,85 @@
...
 
 from hermeto import APP_NAME
-from hermeto.core.errors import FetchError, NotAGitRepo, UnsupportedFeature
+from hermeto.core.errors import FetchError, GitOperationError, NotAGitRepo, UnsupportedFeature
 from hermeto.core.type_aliases import StrPath
 
 log = logging.getLogger(__name__)
 
 
+class HermetoRepo(Repo):
```
a-ovchinnikov: A minor, non-blocking, optional naming observation: `Repo` is not that well associated with the class that `git.repo` provides. I think you could have saved quite a few changes by just naming this class `Repo` and importing the original class as `BaseRepo`.

hermeto/core/errors.py
```diff
@@ -167,3 +167,24 @@ def __init__(
...
+class GitOperationError(BaseError):
+    """The Application failed to perform a Git operation.
+
+    This encompasses all Git-related failures such as repository access, commit resolution,
+    cloning, and checkout operations. Provides consistent error handling across the application.
+    """
+
+    default_solution = textwrap.dedent(
+        f"""
+        The cause of the Git operation failure could be:
```
a-ovchinnikov: This is a very broad error cause, I wonder if it were possible to be more specific?

hermeto/core/scm.py
slimreaper35: Using `Hermeto` prefix for a class is probably not a good idea. We used to have old `Cachi2Error` exception that had to be renamed (before rebranding) in many places to something that is not coupled to the project name. Furthermore, the prefix `Hermeto` does not seem to be related to the git operations in the class.

hermeto/core/scm.py
```diff
@@ -4,21 +4,85 @@
...
+        """Execute git operation with ValueError/TypeError/BadName exception handling."""
+        try:
+            return func()
+        except (ValueError, TypeError, BadName) as ex:
+            raise GitOperationError(f"Git failed during '{operation_name}': {ex}") from ex
+
+    @property
+    def head(self) -> git.Head:
+        """Get repository HEAD reference with error handling."""
+        return self._git_operation("HEAD access", lambda: super(HermetoRepo, self).head)
```
a-ovchinnikov: The way the actual call is abstracted away is neat, however I think it hides the actual problem. Take "HEAD access", for instance. Suppose it fails with `ValueError`. Can we immediately tell the user what went wrong instead of a generic suggestion to check their repo for correctness? Repositories which our users deal with are not always under their control, moreover there could be additional steps involved before Hermeto gets to processing a repo, that is why it is important to be as precise with reporting errors as possible. When we see a `ValueError` in a certain context we almost always know what went wrong, so we better tell the user right away what their issue is.

WDYT about either not abstracting away the calls to the superclass or alternatively passing along a mapping from expected exceptions to more specific subclasses of `GitOperationException`?

hermeto/core/scm.py
```diff
@@ -46,19 +110,13 @@ def get_repo_id(repo: StrPath | Repo) -> RepoID:
...
-                ),
-            )
+        repo = HermetoRepo(repo, search_parent_directories=True)
+    elif isinstance(repo, Repo) and not isinstance(repo, HermetoRepo):
+        repo = HermetoRepo(repo.working_dir)
 
     try:
         origin = repo.remote("origin")
-    except ValueError:
+    except GitOperationError:
```
a-ovchinnikov: Here `GitOperationError` is less precise than `ValueError`. The latter is understood as a universal indicator that a supplied value is wrong, so one can construe that "origin" is somehow not recognized as a correct value which could mean that it is missing. `GitOperationError` on the other hand, could mean anything (`default_solution` tells us how many thing can go wrong to result in this exception). Moreover, it is raised not only for `ValueError`, but also for `TypeError` and `BadName`. Does `TypeError` mean that a repository is lacking a remote or does it mean something else? I think the better way of dealing with this would be to introduce a correct subclass of `GitOperationError` to signal this specific condition of lacking a remote and raise this exception in `repo.remote()`.

tests/unit/package_managers/pip/test_main.py
```diff
@@ -1588,23 +1588,23 @@ def test_generate_purl_dependencies(dependency: dict[str, Any], expected_purl: s
...
+@mock.patch("hermeto.core.scm.HermetoRepo")
 def test_generate_purl_main_package(
     mock_git_repo: Any, subpath: Path, expected_purl: str, rooted_tmp_path: RootedPath
 ) -> None:
     package = {"name": "foo", "version": "1.0.0", "type": "pip"}
 
     mocked_repo = mock.Mock()
     mocked_repo.remote.return_value.url = "ssh://git@github.com/my-org/my-repo"
-    mocked_repo.head.commit.hexsha = GIT_REF
+    mocked_repo.head_commit.hexsha = GIT_REF
```
a-ovchinnikov: Minor, non-blocking: from aesthetic perspective I would have introduced a tiny wrapper around `Head` to preserve the interface for accessing commits (but I won't block on this one, it is a nice to have).

a-ovchinnikov (commented):
Thank you for the change! Please note, that comments marked with "non-blocking" are mostly about code aesthetics and could be safely skipped, however comments about exceptions precision are blocking.

hermeto/core/scm.py
sswguo: Ah, agree with this good point. ; -)

hermeto/core/scm.py
sswguo: I see, and then I prefer to remove the abstract method and keep the specific function addressing the errors, to avoid to maintaining the exception mapping.

hermeto/core/scm.py
sswguo: Good point here, the code here should be only for ValueError, will use the specific error.

tests/unit/package_managers/pip/test_main.py
sswguo: I checked again and see there are other properties `reference` and functions `reset()` related to `Head`, so I think it's worth wrapping it as well.

hermeto/core/errors.py
sswguo: +1, will try to figure it out.

hermeto/core/scm.py
eskultety: Agreed.

hermeto/core/scm.py
```diff
@@ -4,21 +4,85 @@
...
 from pathlib import Path
-from typing import NamedTuple
+from typing import Any, Callable, NamedTuple
 from urllib.parse import ParseResult, SplitResult, urlparse, urlsplit
 
 import git
-from git.exc import InvalidGitRepositoryError, NoSuchPathError
+from git.exc import GitCommandError, InvalidGitRepositoryError, NoSuchPathError
 from git.repo import Repo
+from gitdb.exc import BadName  # type: ignore
```
eskultety: out of curiosity, where did you see this one?

hermeto/core/errors.py
```diff
@@ -167,3 +167,24 @@ def __init__(
         The output of the failing command should provide more details, please check the logs.
         """
     ).strip()
+
+
+class GitOperationError(BaseError):
```
eskultety: Maybe just `GitError` would be sufficient.

hermeto/core/scm.py
eskultety: Maybe `GitRepository` would be an acceptable name?

hermeto/core/scm.py
```diff
@@ -4,21 +4,85 @@
...
+
+    @classmethod
+    def clone_from(  # type: ignore[override]
+        cls, url: str | PathLike[str], to_path: str | PathLike[str], **kwargs: Any
+    ) -> "HermetoRepo":
+        """Clone repository and return HermetoRepo with error handling."""
+        try:
+            Repo.clone_from(url, to_path, **kwargs)
+            return cls(to_path)
+        except GitCommandError as ex:
```
eskultety, line 127: So cloning is pretty much the only instance of `GitCommandError` you're catching IIUC. I don't think that's sufficient, many `Repo` properties resort to using `self.git` to perform the operation, IOW it's just syntactic sugar and we should have the git command errors handled.

hermeto/core/scm.py
eskultety: Also, you sure this is the right way of importing it btw? https://gitpython.readthedocs.io/en/stable/reference.html#git.exc.BadName

hermeto/core/scm.py
eskultety: The way to handle them would be to use https://gitpython.readthedocs.io/en/stable/reference.html#git.repo.base.Repo.GitCommandWrapperType and provide another custom tiny class, e.g. `GitCommandWrapper` that would catch all `GitCommandErrors` with an `execute` method override.

hermeto/core/scm.py
```diff
@@ -4,21 +4,85 @@
...
+        """Get commit object by revision with error handling."""
+        return self._git_operation(
+            f"Commit access ({rev})", lambda: super(HermetoRepo, self).commit(rev)
+        )
+
+    def remote(self, name: str = "origin") -> git.Remote:
+        """Get remote object by name with error handling."""
+        return self._git_operation(
+            f"Remote access ({name})", lambda: super(HermetoRepo, self).remote(name)
+        )
```
eskultety: I'm ambivalent on this lambda approach, because if a problem arises with a custom Git wrapping logic (and it will) I'm wondering how painful it's going to be to debug, but I guess one can add a breakpoint to the accessor be it a method or a property.

hermeto/core/scm.py
eskultety: Initially what I had in mind was using the composition approach rather than inheritance and control the API aspects of the centralized wrapper tightly, i.e. wrap only what we need and have tighter control, because inheritance could surprise us with more methods, changes to properties AND/OR methods and we may need to adjust. On the other hand composition would lead to more code (still good readability though). I'd personally choose composition (doesn't use `GitCommandWrapperType` unlike inheritance) over inheritance here, but I don't want to be too nitpicky, so I guess we can use inheritance and change to a composition approach if it gives headaches in the future.

hermeto/core/errors.py
eskultety: I believe what you need to do here is to take `stdout`, `stderr` and maybe more. Also you should override the `friendly_msg` to make use of the `stdout`, `stderr` fields to construct a meaningful user error out of the failed `GitCommandError`s. Something like:
```
 def __init__(
        self,
        reason: str,
        *,
        stdout: str | None = None,
        stderr: str | None = None,
        solution: str | None = _argument_not_specified,
    ) -> None:
        self.stdout = stdout
        self.stderr = stderr

    def friendly_msg(self) -> str:
        msg = super().friendly_msg()
        if self.stderr:
           msg += f"\n\nGit stderr:\n{textwrap.indent(self.stderr.strip(), '  ')}"
        if self.stdout:
            msg += f"\n\nGit stdout:\n{textwrap.indent(self.stdout.strip(), '  ')}"
        return msg
```

hermeto/core/scm.py
sswguo: The following cmd raises that:
```
python3 -c "import git; repo = git.Repo('.'); repo.commit('gggggggggggggggggggggggggggggggggggggggg')"  
... ... 
in name_to_object  
    raise BadName(name)  
gitdb.exc.BadName: Ref 'gggggggggggggggggggggggggggggggggggggggg' did not resolve to an object
```

I updated it to git.exc.BadName and works too, so seems GitPython re-exports BadName from gitdb based on the following result:
```
> python3 -c "from git.exc import BadName; from gitdb.exc import BadName as GitdbBadName; print(BadName is GitdbBadName)"
> True
```

hermeto/core/scm.py
sswguo: Yeah, I am trying to remove this abstract and address the specific exceptions in each function.

hermeto/core/scm.py
sswguo: I checked and seems only `HEAD.reset()` and `Repo.clone_from()` raise `GitCommandError`,  which are already handled in the latest fixes.

eskultety: > which are already handled in the latest fixes.

https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/gomod.py#L1719-L1744
https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/gomod.py#L1367

Any of those `repo.git.xyz()` calls we have ^^^there (there are more occurrences) can raise the error. If you look deeper at the gitpython implementation of the `Git` class (https://github.com/gitpython-developers/GitPython/blob/7d06b14f0fe85a350f37aaddc5fce91c98d4f7b3/git/cmd.py#L994) you'll see the following pattern:

1. `repo.git.xyz`
2. ` Git.__getattr__("xyz")`
3. `lambda *args, **kwargs: self._call_process("xyz", *args, **kwargs)`
4. `_call_process`
5. `execute`
6. `raise GitCommandError`

hermeto/core/scm.py
eskultety: Fair enough, I don't expect this error anywhere since for a package to be built out of its recipes and metadata the ref field which we query should be a valid ref, but makes sense.

hermeto/core/scm.py
sswguo: That's true, seems I only checked the file `scm.py` from the beginning. ; -)

hermeto/core/errors.py
```diff
@@ -157,3 +167,59 @@ def __init__(
...
+        stderr: str | None = None,
+        solution: str | None = _argument_not_specified,
+    ) -> None:
+        """Initialize GitError.
+
+        :param reason: explain what went wrong
+        :param stdout: stdout output generated by the git operation
+        :param stderr: stderr output generated by the git operation
+        :param solution: politely suggest a potential solution to the user
+        :param docs: include a link to relevant documentation (if there is any)
```
a-ovchinnikov: This one is not present in kwargs list anymore.

hermeto/core/scm.py
```diff
@@ -4,21 +4,150 @@
...
+                type(ex).__name__,
+                str(ex),
+            )
+            raise GitError(
+                f"Failed to clone repository from {url}",
+                stderr=ex.stderr,
+            ) from ex
+
+    @property
+    def head(self) -> HEAD:
```
a-ovchinnikov: Yep, this way it is a bit more verbose and less universal, but at the same time it is much easier to tell at a glance what is happening here and what could go wrong. Thank you!

hermeto/core/scm.py
sswguo: Implemented the GitCommandWrapper for handling the errors of `repo.git.xyz()`.

hermeto/core/scm.py
```diff
@@ -4,21 +4,150 @@
...
-from typing import NamedTuple
+from typing import Any, NamedTuple
 from urllib.parse import ParseResult, SplitResult, urlparse, urlsplit
 
 import git
-from git.exc import InvalidGitRepositoryError, NoSuchPathError
-from git.repo import Repo
+from git.exc import BadName, GitCommandError, InvalidGitRepositoryError, NoSuchPathError
+from git.refs import HEAD as BaseHEAD
+from git.repo import Repo as BaseRepo
```
eskultety: Please don't do this - that's deliberate shadowing, i.e. you're purposefully aliasing the original classes so that we can name them the same as the original, that's IMO obfuscating and I'd personally prefer if we adopted our own names for the sake of clarity, e.g. `GitHEAD, GitRepo`. I presume this was a trick to avoid having to update potentially many typing references across the repo? 

nitpick: Also, for these 2 specifically, I feels there's little need to import them as standalone symbols, since they're used on a couple of places only and we already do `import git` so we can just reference the necessary namespace without too much inconvenience.

hermeto/core/errors.py
```diff
@@ -157,3 +167,58 @@ def __init__(
...
+    """Invalid Git revision (commits, branches, tags, revision specifiers)."""
+
+    default_solution = textwrap.dedent(
+        """
+        Invalid revision. This could be due to:
+        - revision does not exist in the repository
+        - revision points to wrong object type (expected commit)
+        - repository has no commits
+        Please verify the revision exists and is spelled correctly.
+        """
```
eskultety: Do we need a solution here? Don't `scm.py:41` and `scm.py:44` already take care of that including the original exception? Can you give me an example of an original exception that would not provide a clear hint that justifies a solution in this particular case?

hermeto/core/scm.py
```diff
@@ -4,21 +4,150 @@
...
 
 from hermeto import APP_NAME
```
eskultety: in `scm.py: get_repo_for_path` you now have a redundant `Repo` wrap:
https://github.com/sswguo/hermeto/blob/0b2201550b5ee63dd9a728f53fc3a39c51cc222f/hermeto/core/scm.py#L248

`_get_submodule_repo` already returns the appropriate `Repo` instance (btw. see how confusing the change in this PR has become? By playing with namespaces and aliases during imports the new code silently affects existing logic which would otherwise have to be updated appropriately if we used a custom name for the wrapping classes)

hermeto/core/scm.py
```diff
@@ -120,7 +244,8 @@ def get_repo_for_path(repo_root: Path, target_path: Path) -> tuple[Repo, Path]:
     current_repo = Repo(repo_root)
 
     while (submodule := _find_submodule_containing_path(current_repo, target_path)) is not None:
-        current_repo = _get_submodule_repo(submodule)
+        submodule_repo = _get_submodule_repo(submodule)
+        current_repo = Repo(submodule_repo.working_dir)
```
eskultety: `_get_submodule_repo` already returns a proper object, so this wrap is redundant.

hermeto/core/scm.py
```diff
@@ -197,11 +322,12 @@ def _reset_git_head(repo: Repo, ref: str) -> None:
     try:
         repo.head.reference = repo.commit(ref)  # type: ignore # 'reference' is a weird property
         repo.head.reset(index=True, working_tree=True)
-    except Exception as ex:
+    except GitError as ex:
         log.exception(
-            "Failed on checking out the Git ref %s, exception: %s",
+            "Failed on checking out the Git ref %s, exception: %s, exception-msg: %s",
```
eskultety: Now that we have more granular error handling for Git, does ^this make sense? Shouldn't it be just
`"Failed on checking out the Git ref %s: %s"` and leave our the exception name and other params??

hermeto/core/scm.py
```diff
@@ -10,6 +10,7 @@
 from urllib.parse import ParseResult, SplitResult, urlparse, urlsplit
 
 import git
+from git.cmd import Git
```
eskultety: you don't need this convenience import, right? You used it at exactly a single spot.

hermeto/core/scm.py
a-ovchinnikov: > I presume this was a trick to avoid having to update potentially many typing references across the repo?

Yes, I suggested shadowing the original class and mimicking the interface to have a drop-in replacement.

sswguo: Thanks for the feedbacks. I see two different suggestions here. let’s pick one for now, and I can revisit or refine it if needed when I return from PTO next Tuesday. :-)

hermeto/core/scm.py
sswguo: Yeah, we can use friendly_msg from GitError.

sswguo:
> Additionally, a tiny nitpick on commit messages: please also add `Closes https://github.com/hermetoproject/hermeto/issues/905` to both commit messages to have the complete picture in the git history itself (i.e. not putting it just to the PR description).
> 
> Lastly, @sswguo just a tiny reminder - if there was any AI usage involved in contributing the code, it needs to be clearly marked in the commit message using the `Assisted-by:` tag. If not, no action needed obviously :)

Sure, I just left the `Closes issue ...` and `Assisted-by: ...` in the PR description, will amend the commits with this info too.

hermeto/core/errors.py
sswguo: Yeah, I checked all the usage of GitInvalidRevisionError, and all have included the specific error, let's remove this.

hermeto/core/scm.py
eskultety: @sswguo I apologize for the conflict here, I missed Alexey's earlier comment simply because it's been resolved in the meantime. I don't always look at resolved comments (it would slow down the review process a lot), we're trying our best so it doesn't happen often :).

> Yes, I suggested shadowing the original class and mimicking the interface to have a drop-in replacement.

@a-ovchinnikov let's not do this ever please, it's IMO terribly unintuitive, easily to be missed and hackish, regardless of whether it works. As the code base grows, readability should be our nr.1 expectation for any code contribution.

sswguo: @eskultety All good 🙂 I do the same too as usually focus on unresolved comments. 

And fair enough, let's use name `GitRepo` and `GitHEAD`. ; -)

a-ovchinnikov: > let's not do this ever please

Ok, my earlier suggestion is withdrawn. I am still not convinced that in _this particular case_ it is harmful, but lets continue with a separate name.

hermeto/core/scm.py
eskultety: What I meant was that we only need the `friendly_msg` bit, we don't need the exact exception name for anything I think.

PR #1215, opened 2025-12-16
fix(core): use granular http timeouts (connect/read) 

derasdf:
Downloads can fail immediately when a single request hits the global timeout. Replace the global `total` HTTP timeout with granular `connect` and`read` timeouts. In aiohttp, `total` is a hard cap on the entire request; by setting `total=None`, downloads can run as long as needed as long as the connection is established within `connect=30` and the server never stalls for more than `read=300` seconds between bytes.
The existing `http.timeout=300` setting is kept unchanged in config to avoid breaking existing users, but is deprecated in favor of `http.read_timeout`.

### DIFF
--- a/hermeto/core/config.py
+++ b/hermeto/core/config.py
@@ -29,7 +29,7 @@ _FLAT_FIELD_MIGRATIONS = [
     ("ignore_pip_dependencies_crates", ("pip", "ignore_dependencies_crates")),
     ("goproxy_url", ("gomod", "proxy_url")),
     ("gomod_download_max_tries", ("gomod", "download_max_tries")),
-    ("requests_timeout", ("http", "timeout")),
+    ("requests_timeout", ("http", "read_timeout")),
     ("subprocess_timeout", ("runtime", "subprocess_timeout")),
     ("concurrency_limit", ("runtime", "concurrency_limit")),
 ]
@@ -46,16 +46,16 @@ def _remove_gomod_strict_vendor(data: dict[str, Any]) -> None:
         )
 
 
-def _migrate_flat_to_namespace(
+def _migrate_deprecated_field(
     data: dict[str, Any],
     old_key: str,
     value: Any,
     namespace: str,
     new_key: str,
 ) -> None:
-    """Migrate a legacy flat config field to the new namespaced structure.
+    """Migrate a deprecated config field to its replacement.
 
-    If the namespaced field isn't already set, copies the value and warns about deprecation.
+    If the new field isn't already set, copies the value and warns about deprecation.
     If already set, keeps the existing value and warns about the conflict.
     """
     data.setdefault(namespace, {})
@@ -96,7 +96,8 @@ class GomodSettings(BaseModel, extra="forbid"):
 class HttpSettings(BaseModel, extra="forbid"):
     """HTTP-related settings."""
 
-    timeout: int = 300
+    connect_timeout: int = 30
+    read_timeout: int = 300
 
 
 class RuntimeSettings(BaseModel, extra="forbid"):
@@ -141,13 +142,24 @@ class Config(BaseSettings):
 
         for old_key, (namespace, new_key) in _FLAT_FIELD_MIGRATIONS:
             if old_key in data:
-                _migrate_flat_to_namespace(data, old_key, data.pop(old_key), namespace, new_key)
+                _migrate_deprecated_field(data, old_key, data.pop(old_key), namespace, new_key)
+
+        # Migrate http.timeout -> http.read_timeout
+        http_data = data.get("http")
+        if isinstance(http_data, dict) and "timeout" in http_data:
+            _migrate_deprecated_field(
+                data,
+                "http.timeout",
+                http_data.pop("timeout"),
+                "http",
+                "read_timeout",
+            )
 
         # default_environment_variables.gomod -> gomod.environment_variables
         # (default_environment_variables only ever supported the gomod backend)
         default_gomod_env_vars = data.pop("default_environment_variables", {}).get("gomod")
         if default_gomod_env_vars is not None:
-            _migrate_flat_to_namespace(
+            _migrate_deprecated_field(
                 data,
                 "default_environment_variables",
                 default_gomod_env_vars,
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -41,7 +41,8 @@ def download_binary_file(
     :param int chunk_size: Chunk size param for Response.iter_content()
     :raise FetchError: If download failed
     """
-    timeout = get_config().http.timeout
+    config = get_config()
+    timeout = (config.http.connect_timeout, config.http.read_timeout)
     try:
         resp = pkg_requests_session.get(
             url, stream=True, verify=not insecure, auth=auth, timeout=timeout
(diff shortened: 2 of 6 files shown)


tests/unit/package_managers/test_general.py
```diff
@@ -266,3 +266,27 @@ async def test_async_download_files_exception(
 
     assert f"Unsuccessful download: {url}" in caplog.text
     assert str(exc_info.value) == f"exception_name: Exception, details: {exception_message}"
+
+
+@pytest.mark.asyncio
+async def test_async_download_files_retry_config(tmp_path: Path) -> None:
```
a-ovchinnikov: This test looks like it is testing 3rd party library functionality and not code that we control. I suggest not testing anything that is not maintained by us.

tests/unit/package_managers/test_general.py
```diff
@@ -290,3 +291,31 @@ async def test_async_download_files_retry_config(tmp_path: Path) -> None:
         assert mock_jitter_retry.called
         _, kwargs = mock_jitter_retry.call_args
         assert asyncio.TimeoutError in kwargs.get("exceptions", set())
+
+
+@pytest.mark.asyncio
+async def test_async_download_binary_file_timeouts(tmp_path: Path) -> None:
+    """Test that correct timeouts are passed to session.get."""
```
a-ovchinnikov: Here again I am not sure what exactly is being tested: the fact that `session.get` behaves according to spec? I think such tests belong to IT suit rather than to UT.

slimreaper35: > Here again I am not sure what exactly is being tested

Yeah, I am not sure either, if we need such unit tests where 90% is just mocking.

tests/unit/package_managers/test_general.py
```diff
@@ -266,3 +268,16 @@ async def test_async_download_files_exception(
...
     assert str(exc_info.value) == f"exception_name: Exception, details: {exception_message}"
+
+
+@mock.patch("hermeto.core.package_managers.general.get_config")
+def test_get_aiohttp_timeout(mock_get_config: MagicMock) -> None:
+    """Test that aiohttp timeout is configured correctly."""
+    mock_get_config.return_value.http.connect_timeout = 10
+    mock_get_config.return_value.http.read_timeout = 20
+
+    timeout = _get_aiohttp_timeout()
```
a-ovchinnikov: One might say that unit tests exist to test some busyness logic, i.e. some user-facing contract of a piece of code. By this definition internal functions must not be unit tested in isolation because all they do is making actually important code easier to maintain. As long as a piece of code adheres to its contract it is not important how exactly it is organized (at least to its clients). This test verifies an internal helper function and at the same time this internal function just wraps another a function with a more complex invocation. I think this additional functionality should be tested differently. What you want to ensure is that long-running downloads are not killed by timeout which means that you need to somehow simulate a long-running download and demonstrate that this setting prevents it from being killed. Anything that involves interactions with actual services outside of Hermeto control is a better fit for integration tests.

hermeto/core/package_managers/general.py
a-ovchinnikov, line 1: A general question. Would setting `total` to `connect + read` achieve the same result?

derasdf: No. As I understand it, `total` in `aiohttp.ClientTimeout` is a hard cap on the entire request. By setting `total = None`, the download is allowed to run for as long as needed, as long as the connection is established within `connect = 30` and the server never stalls for more than `read = 300` between bytes.

a-ovchinnikov: Thank you. I would suggest adding this answer as a comment or to the commit message or both for the sake of future maintainers.

hermeto/core/package_managers/general.py
```diff
@@ -117,6 +117,7 @@ async def async_download_files(
         exceptions={
             aiohttp.ClientConnectionError,
             aiohttp.ClientPayloadError,
+            asyncio.TimeoutError,
```
taylormadore: I think we can drop this commit as redundant because ClientTimeout total=None means no direct asyncio.TimeoutError, and socket timeouts are already caught by ClientConnectionError

tests/unit/package_managers/test_general.py
```diff
@@ -32,7 +32,8 @@
 def test_download_binary_file(
     mock_get: Any, auth: AuthBase | None, insecure: bool, chunk_size: int, tmp_path: Path
 ) -> None:
-    timeout = get_config().http.timeout
+    config = get_config()
+    timeout = (config.http.connect_timeout, config.http.read_timeout)
```
taylormadore, line 36: We should use the new settings for the pypi_simple client in the pip backend as well. It also uses requests internally

hermeto/core/config.py
```diff
@@ -97,6 +97,8 @@ class HttpSettings(BaseModel, extra="forbid"):
     """HTTP-related settings."""
 
     timeout: int = 300
```
taylormadore: Since we won't be using this (legacy) setting in our backends anymore, we should probably deprecate+migrate it in favor of http.read_timeout in a similar way to how we handled the legacy flat fields.

hermeto/core/config.py
```diff
@@ -143,6 +144,17 @@ def _normalize_config_structure(cls, data: Any) -> Any:
             if old_key in data:
                 _migrate_flat_to_namespace(data, old_key, data.pop(old_key), namespace, new_key)
 
+        # Migrate http.timeout -> http.read_timeout
+        http_data = data.get("http")
+        if isinstance(http_data, dict) and "timeout" in http_data:
+            _migrate_flat_to_namespace(
```
taylormadore: For clarity, we might want to rename this helper to something more generic now like `_migrate_deprecated_field`

hermeto/core/config.py
taylormadore: Let's not forget to update the [README.md ](https://github.com/hermetoproject/hermeto/blob/1b51efeca5d1bca7e77851a26405ccb15f43b992/README.md?plain=1#L142) with this change as well as the new http.connect_timeout

taylormadore (approved):

hermeto/core/config.py
```diff
@@ -46,16 +46,16 @@ def _remove_gomod_strict_vendor(data: dict[str, Any]) -> None:
         )
 
 
-def _migrate_flat_to_namespace(
+def _migrate_deprecated_field(
```
a-ovchinnikov, line 49: Nitpick: for the best OSS score this should go to a separate commit (since this is a separate refactoring).

tests/unit/test_config.py
```diff
@@ -48,14 +48,21 @@ def test_normalize_config_structure(caplog: pytest.LogCaptureFixture) -> None:
...
-    assert config.http.timeout == 600
+    assert config.http.read_timeout == 600
     assert config.runtime.subprocess_timeout == 7200
     assert config.runtime.concurrency_limit == 10
     assert config.yarn.enabled is False
     assert config.pip.ignore_dependencies_crates is True
     assert "is deprecated" in caplog.text
 
 
+def test_migrate_http_timeout(caplog: pytest.LogCaptureFixture) -> None:
```
a-ovchinnikov, line 59: Nitpick: I am not convinced that this test provides any value since it exercises a utility function and essentially checks that a sequence of steps runs as it is written. I think such behaviors should be tested at a different level. This is not a blocker and not an actionable comment, more of an observation.

derasdf: Fair point, but it's cheap insurance and documents the behavior, so I think I'll just leave it as is for now.

hermeto/core/config.py
derasdf: I could split it out, but for this tiny change I’d probably just keep it as is, unless we end up doing other changes.

PR #1216, opened 2025-12-16
Fix externalReferences field serialization in SBOM

slimreaper35:

### DIFF
--- a/hermeto/core/models/sbom.py
+++ b/hermeto/core/models/sbom.py
@@ -64,9 +64,12 @@ class Component(pydantic.BaseModel):
     properties: list[Property] = pydantic.Field(default_factory=list, validate_default=True)
     type: Literal["library", "file"] = "library"
     external_references: list[ExternalReference] | None = pydantic.Field(
-        serialization_alias="externalReferences", default=None
+        alias="externalReferences", default=None
     )
     pedigree: Pedigree | None = None
+
+    # Aliased fields may be populated by their name as given by the model attribute.
+    model_config = pydantic.ConfigDict(validate_by_name=True, extra="forbid")
 
     def key(self) -> str:
         """Uniquely identifies a package.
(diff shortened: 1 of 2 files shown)


hermeto/core/models/sbom.py
```diff
@@ -63,9 +63,7 @@ class Component(pydantic.BaseModel):
     version: str | None = None
     properties: list[Property] = pydantic.Field(default_factory=list, validate_default=True)
     type: Literal["library", "file"] = "library"
-    external_references: list[ExternalReference] | None = pydantic.Field(
-        serialization_alias="externalReferences", default=None
-    )
+    externalReferences: list[ExternalReference] | None = None
```
taylormadore: Will switching `serialization_alias` --> `alias` also fix this issue?

slimreaper35: I guess I learned a pydantic lesson 😕

> Will switching serialization_alias --> alias also fix this issue?

It did not at first. If an alias is set to a field, than it has to be used to instatiate the `Component`. For the generic backend, we use [snake_case](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/generic/models.py#L122) which did not match the camelCase of an alias so the `external_references` would be `None` and get removed from our data.

I added a `ConfigDict` to the class to allow both, the attribute name and alias.

taylormadore: Pydantic is always a fun adventure :slightly_smiling_face: 

I guess the one downside of this approach is that our model will accept JSON with external_references (snake_case), which is invalid CycloneDX. I think this leniency is OK given that we already have a disclaimer that the merge-sboms command is only meant to be used with SBOMs we generate.

I'd probably just switch `populate_by_name` --> `validate_by_name`, since the pydantic 2.12 documentation mentions that  `populate_by_name` is deprecated and will be removed in V3.

PR #1227, opened 2026-01-05
Proxy support foundational work

a-ovchinnikov:
This is the initial work for adding proxy support to Hermeto.

### DIFF
--- a/docs/design/proxy-support.md
+++ b/docs/design/proxy-support.md
@@ -78,24 +78,6 @@ Basing on the information provided above it is recommended to use
 [PackageSourceInfo][] for SPDX and [externalReferences][] for CycloneDX.
 
 
-### Fallback option
-
-It might be desired to try and retrieve packages from the original source if
-proxy fails for any reason. In other words, with fall-back enabled Hermeto
-would try a proxy first, and if the proxy is non-responsive it would then try
-fetching the sources as usual, as if no proxy settings were present. With
-fall-back disabled Hermeto would fail a fetch if a proxy is not responding. To
-achieve that the following option is proposed for addition to Hermeto config:
-
-```
-    fall-back-to-standard-source = False
-```
-
-It will be called `HERMETO_FALL_BACK_TO_STANDARD_SOURCE` in the environment.
-The default value is proposed to be set to False. Fall-back should not happen
-if there was an authorization error because it would mask it otherwise.
-
-
 ## Implementation
 
 All configuration is proposed to be done via environment variables and


hermeto/core/models/sbom.py
```diff
@@ -230,8 +242,10 @@ def libs_to_packages(libraries: list[Component]) -> list[SPDXPackage]:
                         versionInfo=component.version,
                         externalRefs=[erefdict(component)],
                         annotations=[pAnnotation(comment=mkcomm(p)) for p in component.properties],
+                        packageSourceInfo=None if not (l:=source_infos(component)) else ";".join(l)  # noqa: E741
```
eskultety: You've entered a territory where this line is already becoming a bit hard to read, just use a variable and even Ruff will be happy :)

hermeto/core/config.py
```diff
@@ -15,7 +15,7 @@
 from hermeto import APP_NAME
 from hermeto.core.errors import InvalidInput
 
-# Ascending priority
+# Ascending priority: the closer config is to the tail the gigher priority it gets.
```
eskultety: I don't think this is change is needed, that's exactly what ascending means :), I'd even argue that after reading it, I stumbled upon 'tail' for a second before it clicked for me what you meant, so for me this hunk is inferior to the original.

Anyway, `s/gigher/higher`.

hermeto/core/config.py
```diff
@@ -102,10 +102,20 @@ class HttpSettings(BaseModel, extra="forbid"):
 class RuntimeSettings(BaseModel, extra="forbid"):
     """General runtime execution settings."""
 
+    # To set up a variable you would do like this:
+    # HERMETO_RUNTIME__CONCURRENCY_LIMIT
+    # Note single underscore after application name, then name of the section as it appears
+    # in Config class definition, then double underscore and field name after that.
```
eskultety: Should go to a separate commit I believe, unrelated to the proxy settings stuff being added.

hermeto/core/models/sbom.py
```diff
@@ -420,6 +432,7 @@ def from_package_dict(cls, package: dict[str, Any]) -> "SPDXPackage":
         """Create a SPDXPackage from our dictionary."""
         external_refs = package.get("externalRefs", [])
         annotations = [SPDXPackageAnnotation(**an) for an in package.get("annotations", [])]
+        packageSourceInfo = package.get("packageSourceInfo", None)
```
taylormadore: nit: Is there any reason why this variable name is camelCase while `external_refs` above is snake_case?

hermeto/core/models/sbom.py
```diff
@@ -394,6 +398,14 @@ def __hash__(self) -> int:
...
             + sum(hash(e) for e in self.externalRefs)
             + sum(hash(a) for a in self.annotations)
+            # NOTE: this is problematic since in a rare case when there are
+            # several proxies and a package was downloaded from one proxy and
+            # then from another one in some other context they will be hashed
+            # to the same value and deduplicated. There is a slim chance that
+            # they won't actually be identical. Since we do not have a way to
+            # distinguish between multiple proxies and since this chance is
+            # very small this is something we would need to live with.
+            + hash(self.packageSourceInfo)
```
taylormadore: It took me a minute to understand this comment, so just to make sure I have it correct, let me try to restate it: When you don't know which proxy was actually used, we list all of them, so packages that came from different actual sources will look identical. Is that right?

hermeto/core/models/sbom.py
```diff
@@ -382,6 +382,10 @@ class SPDXPackage(pydantic.BaseModel):
     externalRefs: list[SPDXPackageExternalRefType] = []
     annotations: list[SPDXPackageAnnotation] = []
     downloadLocation: str = "NOASSERTION"
+    # PackageSourceInfo should be present if proxy URL were set for a package
+    # manager. If more than one URL were provided then individual URLs
+    # should be separated with semicolons.
+    packageSourceInfo: str | None = None
```
taylormadore: The commit message for this one calls it PackageSourceReference

a-ovchinnikov:
This is just a rebase, will address comments in a follow-up push.

hermeto/core/models/sbom.py
a-ovchinnikov: Fixed. Thank you for spotting that!

hermeto/core/models/sbom.py
a-ovchinnikov: Yes. This comment is hard to get through, I'll try to reword it.

hermeto/core/models/sbom.py
a-ovchinnikov: Ok, maybe this one is slightly overengineered.

hermeto/core/config.py
a-ovchinnikov: I was stuck with trying to figure out which way a list is supposed ascend. On my screen it actually descends from top to bottom :D. Tail of a list OTOH felt rather natural, like in [Head|Tail] list definition. I believed it to be a wider accepted term than, say, cdr. In any case I'll drop this change.

hermeto/core/models/sbom.py
a-ovchinnikov: No, that's an omission. I have also removed this method from the class since it was not used anywhere in the codebase.

hermeto/core/models/sbom.py
```diff
@@ -697,6 +697,7 @@ def to_cyclonedx(self) -> Sbom:
                 )
                 for an in package.annotations
             ]
+            # sourceInfo morphs into ExternalReference of type SRCDIST
```
taylormadore: I think this was meant to be in the previous commit

hermeto/core/models/sbom.py
```diff
@@ -454,6 +458,13 @@ def __hash__(self) -> int:
             + hash(self.downloadLocation)
             + sum(hash(e) for e in self.externalRefs)
             + sum(hash(a) for a in self.annotations)
+            # NOTE: in a rare case when several proxies exist and a package
+            # ends up being downloaded twice in two different environments each
+            # containing the same set of proxy URLs, there is a small chance
```
eskultety: To better understand this - what does _ends up being downloaded twice in two different environments each containing the same set of proxy URLs_ mean? Can you describe the scenario for me?

hermeto/core/models/sbom.py
```diff
@@ -54,6 +57,8 @@ class ExternalReference(pydantic.BaseModel):
     # ExternalReferences. This type of ExternalReference should be added
     # along with "distribution" ExternalReference.
     # NOTE: CycloneDX.ExternalReference != SPDX.ExternalReference!
+    # mypy complains about the usage of SRCDIST here, so a manual expansion is
+    # used instead to make it happy.
```
eskultety: I wouldn't even notice :), I think we can drop the commentary and some poor soul feels like "optimizing", `mypy` will say "Actually...." ;)

I liked having the `NOTE CDX.ExternalReference != SPDX.ExternalReference` mentioned last more.

hermeto/core/models/sbom.py
```diff
@@ -685,8 +697,18 @@ def to_cyclonedx(self) -> Sbom:
                 )
                 for an in package.annotations
             ]
+            # sourceInfo morphs into ExternalReference of type SRCDIST
+            if package.sourceInfo is not None:
+                actual_download_urls = package.sourceInfo.split(";")
+                ers = [ExternalReference(url=s, type=SRCDIST) for s in actual_download_urls]
```
eskultety: nitpick: `er` and `ers` heavily resemble `err` and `errors`, probably something like `eref` or just `ref/refs` is sufficient, it's just a variable name, the fact it's an "external" reference doesn't make a difference :)

tests/unit/models/test_sbom.py
```diff
@@ -396,28 +396,93 @@ def test_to_spdx(self, mock_spdx_now: str) -> None:
...
+                            "name": "spdx-expression-parse",
+                            "version": "v1.0.0",
+                            "purl": "pkg:npm/spdx-expression-parse@1.0.0",
+                        },
+                        {
+                            "name": "github.com/org/A",
+                            "version": "v1.0.0",
+                            "purl": "pkg:golang/github.com/org/A@v1.0.0",
+                            "external_references": [
+                                {"url": "https://www.fakeproxy.com", "type": "source-distribution"}
```
eskultety: @a-ovchinnikov can you refresh my mind on the `type` decision? Was it by any chance me who suggested that for CycloneDX? :)
Looking at the design doc the specific `type` for the external ref wasn't captured in there (which is good, that's an implementation detail I'd say).

Disclaimer: I will not block on this, but a brief discussion and refreshing my memory would be nice

Looking at the [ CycloneDX docs](https://cyclonedx.org/docs/1.6/json/#components_items_externalReferences_items_type) there's also `distribution` and `distribution-intake`. My question is: How do you understand the _intermediary_ mention in `distribution-intake`, do you think it might describe the proxy slightly better? The reason I'm asking is because for `source-distribution` the docs says:

> The location where the source code distributable can be obtained. This is often an archive format such as zip or tgz.

^This is simply not true for any binary distributables like wheels and for that we'd probably have to differentiate and employ the `distributable` type as well for our SBOM accuracy claim to remain valid. On the other hand:

> distribution-intake: The location where a component was published to

Sounds pretty vague which means we could stuff anything underneath :), @taylormadore any thoughts?

tests/unit/models/test_sbom.py
```diff
@@ -457,6 +522,74 @@ def test_generate_package_annotations(self, mock_spdx_now: str) -> None:
...
+        # implicitly verified elsewhere, here it is just surfaced to make sure
+        # that it is not forgotten or broken.  It would make sense to get rid
+        # of this test once anything else depends on this functionality or once
+        # there is an appropriate integration test which utilises proxies.
+
+        spdx_sbom = sbom.to_spdx("NOASSERTION")
+
+        # mypy insists that it is impossible to check if something is a part of None
+        # however in this particular case it will never be None:
+        # TODO: remove this comment and the pragma below once SPDXID is correctly typed.
```
eskultety: Is it work on our end to make sure SPDXID is correctly typed? Didn't quite get that. Do we need a minor priority issue to track that?

tests/unit/models/test_sbom.py
```diff
@@ -457,6 +522,74 @@ def test_generate_package_annotations(self, mock_spdx_now: str) -> None:
...
+        mock_spdx_now: str,
+        sbom: Sbom,
+        expected_package_source_info: str,
+    ) -> None:
+        # NOTE: this test explicitly verifies an implementation detail. It is
+        # implicitly verified elsewhere, here it is just surfaced to make sure
+        # that it is not forgotten or broken.  It would make sense to get rid
+        # of this test once anything else depends on this functionality or once
+        # there is an appropriate integration test which utilises proxies.
+
```
eskultety: You sure we'll need to delete it? Conversion as an operation is an important test to have, so if anything I'd expect some `TestCDXSPDXConversion` test class that would test this logic thoroughly. :)

hermeto/core/config.py
```diff
@@ -103,6 +103,11 @@ class HttpSettings(BaseModel, extra="forbid"):
 class RuntimeSettings(BaseModel, extra="forbid"):
     """General runtime execution settings."""
 
+    # This is how an environment variable name should look like:
+    #   HERMETO_RUNTIME__CONCURRENCY_LIMIT
+    # Note single underscore after application name, then name of the section
+    # as it appears in Config class definition, then double underscore and
+    # field name after that.
```
eskultety, line 110: I think this is more important for the end-users rather than devs, let's keep it though. In any case I'll be creating an issue on adding a CLI command to dump the available configuration options and envs because the naming can be quite overwhelming and it'll be nicely visible to the end users.

hermeto/core/models/sbom.py
```diff
@@ -261,6 +272,11 @@ def generate_package_annotations(properties: list[Property]) -> list[SPDXPackage
             return result
 
         def libs_to_packages(libraries: list[Component]) -> list[SPDXPackage]:
+            def source_infos(component: Component) -> list[str]:
+                if component.external_references is None:
+                    return []
+                return [er.url for er in component.external_references if er.type == SRCDIST]
```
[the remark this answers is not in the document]
eskultety: I think this is a valid comment from the AI bot.

hermeto/core/models/sbom.py
a-ovchinnikov: The scenario is likely purely hypothetical -- suppose someone has a pipeline which combines results from different fetches, also suppose they deduplicate the results after that. Then _if_ there were multiple proxies (at least golang allows that), _if_ different proxies were used for fetching the same package and _if_ artifacts fetched from these two proxies were different (e.g. one of the proxies was compromised in some non-trivial way) we will not be able to differentiate between these two artifacts because we do not know which exact proxy each of them came from and thus they will be hashed to same value and one of them (we do not know which) will be dropped during deduplication. I do not believe that the scenario is probable, but I mentioned it since it is possible.

hermeto/core/models/sbom.py
a-ovchinnikov: Ok, fair enough. I dropped a comment above, but could not resist and added an in-line comment for now.

hermeto/core/models/sbom.py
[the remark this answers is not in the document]
a-ovchinnikov: I deemed this one to also be from possible, but not probable realm. I'll update the code nevertheless.

hermeto/core/models/sbom.py
```diff
@@ -275,6 +291,7 @@
                 else:
                     human_readable_id = component.name
 
+                source_info = ";".join(source_infos(component))
```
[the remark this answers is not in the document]
a-ovchinnikov: Nope, it is used.

hermeto/core/models/sbom.py
a-ovchinnikov: Fair enough, updated the name here since it lives on more than one line (I kept `er` on L276 since it is used within a proper comprehension and that just cannot confuse a reader). I usually try to leave some hints of what the variable actually refers to even in narrowly scoped names, even when it is just one letter, just to make it easier to connect pieces together.

tests/unit/models/test_sbom.py
a-ovchinnikov: I chose `source-distribution` because of `location where the source code distributable can be obtained` part. Since out main focus is sources I decided that it will describe the situation at hand the most accurately. The other two options could be used as well, but I'd like to note that all three are not a perfect fit, so I went with a good enough fit. I do not like the part about location where a component was published to: in case of some internal proxy no package maintainer has ever published their package to it, so that would be a false statement. OTOH since the remaining two options are comparably vague we could use them too.

tests/unit/models/test_sbom.py
a-ovchinnikov: Maybe not. I remember that I was pretty sure that the test is a stop-gap solution back when I wrote this comment, but I cannot recall why I thought this way. I should probably drop this comment.

tests/unit/models/test_sbom.py
a-ovchinnikov: Sorry, this is a note to myself, it was meant to be dropped once the code is published. The change I mention here was merged a few weeks back, I have even removed this comment at some point (apparently in a branch that did not make it into the PR).

hermeto/core/config.py
a-ovchinnikov: It took me some trial and error to figure out the exact rules for naming variables and underscores placement, so I decided to leave a clue in the code, somewhere where it is not too hard to find. This definitely needs to be shared with the users too.

hermeto/core/models/sbom.py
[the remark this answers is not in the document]
a-ovchinnikov: The only thing that bothers me is that the ordering of URLs might change compared to one present in an environment variable.

tests/unit/models/test_sbom.py
taylormadore: I agree that none are a perfect fit.

The context for `distribution-intake` is that it was added for the Apache Software Foundation and seems to be used to point to where the package was published to in their staging area.

I suppose I lean towards `distribution` since it can be used for both sources and binaries and we are documenting the repository where we actually downloaded the package from.

eskultety: like I said, my only argument is that not everything we fetch is source.

> I do not like the part about location where a component was published to: in case of some internal proxy no package maintainer has ever published their package to it, so that would be a false statement. 

That's fair, so `distribution` then? Seems vague enough to cover everything without too much trouble.

hermeto/core/config.py
eskultety: https://github.com/hermetoproject/hermeto/issues/1253. Could potentially cover the docs part as well -> resolving.

hermeto/core/models/sbom.py
eskultety: Wow, that's quite meta :). The commentary however doesn't quite provide the dept of the example you shared. Not sure I'd need to keep such an hypothetical case commented in the code. Non-blocking, however, there always will be bugs, so I'd be totally fine not bothering with this in any way now and let's just see if it ever happens.

a-ovchinnikov: Back when I was writing this comment it felt important enough to mention. Looking back to it now I don't think it is that important. I think I'll drop this comment in the next revision (or at least shrink it significantly).

hermeto/core/models/sbom.py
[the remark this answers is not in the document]
eskultety: Would that impact the tests? IOW Don't we control that outcome?

tests/unit/models/test_sbom.py
eskultety: @a-ovchinnikov agree^^?

a-ovchinnikov: Yes, sure, I am updating `source-distribution` to `distribution` for the next revision.

hermeto/core/models/sbom.py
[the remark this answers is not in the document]
a-ovchinnikov: Yes, we control it. This is not a major concern though, just something that we should be aware of.

tests/unit/models/test_sbom.py
eskultety: @a-ovchinnikov per our private channel discussion, maybe you'd want to expand (most likely in the commit message) on the reasons for selecting the `distribution` field, its limitations and how we'd like to approach it with the conflict that exists wrt this field in the generic fetcher backend, thanks.

a-ovchinnikov:
The latest revision:
 1. `source-distribution` was replaced with `distribution` plus a comment field;
 2. comments were shrunk down.
I'll keep the PR open for a while in case there is any disagreement on the way external references are marked.

PR #1230, opened 2026-01-07
chore: tests: Force submodules to main in ITs

a-ovchinnikov:
Integration tests could experience a spontaneous cascading failure which was traced back to a presumably race condition somewhere around updating submodules in golang ITs. After some experimentation it turned out that making sure all submodules are at 'main' prior to checking out the right branch resolved the issue.

This change amends the utility which deals with test repository to always check out 'main' first.

### DIFF
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -385,7 +385,12 @@ def fetch_deps_and_check_output(
     else:
         actual_repo_dir = test_repo_dir
         repo = Repo(actual_repo_dir)
-        repo.git.reset("--hard")
+        # Submodule could end up being in detached HEAD state which would
+        # result in a cascading failure for all tests that follow. This does
+        # not happen always and at the moment of writing it is not clear what
+        # exactly triggers such behavior. However ensuring that all submodules
+        # are hard-reset resolves the issue.
+        repo.git.reset("--hard", "--recurse-submodules")
         # remove untracked files and directories from the working tree
         # git will refuse to modify untracked nested git repositories unless a second -f is given
         repo.git.clean("-ffdx")


tests/integration/utils.py
```diff
@@ -389,6 +389,12 @@ def fetch_deps_and_check_output(
         # remove untracked files and directories from the working tree
         # git will refuse to modify untracked nested git repositories unless a second -f is given
         repo.git.clean("-ffdx")
+        # Submodule could end up being in detached HEAD state which would
+        # result in a cascading failure for all tests that follow. This does
+        # not happen always and at the moment of writing it is not clear what
```
eskultety, line 390: > This does not happen always

Sure does for me :)

This has terrorized me for months, but given that it wasn't always the same subset of tests that failed with a similar error like this one:
```
E           git.exc.GitCommandError: Cmd('git') failed due to: exit code(1)
E             cmdline: git checkout gomod/empty-vendor-fails-vendor-check --recurse-submodules
E             stderr: 'error: Entry 'integration-tests/cachito_gomod.go' not uptodate. Cannot merge.
E           error: Submodule 'integration-tests' could not be updated.
E           error: Cannot update submodule:
E               integration-tests
E
E           Aborting'
```
I classified this as a race and just couldn't find the time to investigate and justify this since it appeared I was the only one suffering from this. In any case, **thank you very much for the investigation** you're my hero. Now that you've appointed the function where the problem resides and looking at the error again I noticed we only reset the parent repo, but not the submodules which apparently can be dirtied with some Go tests. The following did the trick for me and now I was able to finish several sequential IT runs successfully \o/

```
diff --git tests/integration/utils.py tests/integration/utils.py
index 1b913d125..5ed74579c 100644
--- tests/integration/utils.py
+++ tests/integration/utils.py
@@ -385,7 +385,7 @@ def fetch_deps_and_check_output(
     else:
         actual_repo_dir = test_repo_dir
         repo = Repo(actual_repo_dir)
-        repo.git.reset("--hard")
+        repo.git.reset("--hard", "--recurse-submodules")
         # remove untracked files and directories from the working tree
         # git will refuse to modify untracked nested git
```

Note that we might need to also clean the submodules recursively, but we don't strictly need it since I was able to get a green run with just the above :) .

a-ovchinnikov: Yep, the proposed change seems to work just as fine and is less messy. Will adopt it.

tests/integration/utils.py
```diff
@@ -385,7 +385,12 @@ def fetch_deps_and_check_output(
     else:
         actual_repo_dir = test_repo_dir
         repo = Repo(actual_repo_dir)
-        repo.git.reset("--hard")
+        # Submodule could end up being in detached HEAD state which would
```
eskultety, line 388: **Commit message:**

> After some
> experimentation it turned out that making sure all submodules are
> at 'main' prior to checking out the right branch resolved the issue.

> This change amends the utility which deals with test repository to
> always check out 'main' first.
> 

These are no longer true with respective the code change.

tests/integration/utils.py
```diff
@@ -385,7 +385,12 @@ def fetch_deps_and_check_output(
...
+        # result in a cascading failure for all tests that follow. This does
+        # not happen always and at the moment of writing it is not clear what
```
eskultety, line 390: I think we may have figured it out with the dirty tree, so I think we can say that it's clearer now?

eskultety (approved):
ACK with the minor nits fixed.

tests/integration/utils.py
a-ovchinnikov: Ouch, I knew I was forgetting something.

tests/integration/utils.py
a-ovchinnikov: Yes, but we have not figured out why it happens sometimes and does not happen at other times. For me the estimated chance of a cascade is about 50%, so about half the times something clears up the mess and another half times it does not and I do not know what exactly is responsible for such behavior. I wanted to capture this unknown in the comment, but maybe this is too much information

eskultety: Fair enough, it's true that I managed to reduce the subset more and more in the past to integration test my changes locally successfully, but it wasn't very deterministic with which test addition to the subset it would start failing --> race.

PR #1232, opened 2026-01-08
[gomod] Enable .netrc authentication for private proxies

eskultety:
Fixes https://github.com/hermetoproject/hermeto/issues/1219

### DIFF
--- a/hermeto/core/package_managers/gomod.py
+++ b/hermeto/core/package_managers/gomod.py
@@ -1021,6 +1021,8 @@ def _go_exec_env(**extra_vars: str) -> dict[str, str]:
     """Build the base environment for go command execution."""
     env = {
         "PATH": os.environ.get("PATH", ""),
+        "HOME": os.environ.get("HOME", Path.home().as_posix()),  # HOME= can be unset, hence Path
+        "NETRC": os.environ.get("NETRC", ""),
     }
     return env | extra_vars
 
--- a/tests/unit/package_managers/test_gomod.py
+++ b/tests/unit/package_managers/test_gomod.py
@@ -2612,20 +2612,30 @@ class TestGoWork:
         assert mock_go.call_args[0][0] == ["work", "edit", "-json"]
 
 
+_ENV_VARS_BASE_INIT = {v: "/some/path" for v in ("PATH", "HOME", "NETRC")}
+
+
 @pytest.mark.parametrize(
     "env, extra_env, expected",
     [
-        pytest.param({"PATH": "/some/path"}, None, {"PATH": "/some/path"}, id="vars_inherited"),
-        pytest.param({}, None, {"PATH": ""}, id="vars_defaults"),
+        pytest.param(_ENV_VARS_BASE_INIT, None, _ENV_VARS_BASE_INIT, id="vars_inherited"),
         pytest.param(
-            {"PATH": "/some/path"},
+            {},
+            None,
+            {"PATH": "", "HOME": "/mocked/home", "NETRC": ""},
+            id="vars_defaults",
+        ),
+        pytest.param(
+            _ENV_VARS_BASE_INIT,
             {"GOPATH": "/tmp/go"},
-            {"PATH": "/some/path"} | {"GOPATH": "/tmp/go"},
+            _ENV_VARS_BASE_INIT | {"GOPATH": "/tmp/go"},
             id="with_extra_env",
         ),
     ],
 )
+@mock.patch("pathlib.Path.home", return_value=Path("/mocked/home"))
 def test_go_exec_env(
+    mock_home: mock.Mock,
     monkeypatch: pytest.MonkeyPatch,
     env: dict[str, str],
     extra_env: dict[str, str] | None,


hermeto/core/package_managers/gomod.py
```diff
@@ -1017,6 +1017,18 @@ def _parse_packages(
     return iter(all_packages)
 
 
+def _go_exec_env(**extra_vars: Any) -> dict[str, str]:
+    """Build the base environment for go command execution."""
```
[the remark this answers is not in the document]
eskultety: The argument is self-explanatory + it's a helper -> resolving.

hermeto/core/package_managers/gomod.py
```diff
@@ -1052,22 +1064,20 @@ def _resolve_gomod(
...
+    go_vars: dict[str, str] = {
+        "GOPATH": str(tmp_dir),
         "GO111MODULE": "on",
-        "GOCACHE": tmp_dir,
-        "PATH": os.environ.get("PATH", ""),
+        "GOCACHE": str(tmp_dir),
         "GOMODCACHE": gomod_cache,
-        "GOSUMDB": "sum.golang.org",
         "GOTOOLCHAIN": "auto",
     }
```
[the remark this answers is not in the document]
eskultety: Oops, serious rebase problem, addressed.

tests/unit/package_managers/test_gomod.py
```diff
@@ -2609,3 +2610,43 @@ def test_get_go_work(self) -> None:
...
+            _ENV_VARS_BASE_INIT,
+            _ENV_VARS_BASE_INIT | {"GOPATH": "/tmp/go"},
+            {"GOPATH": "/tmp/go"},
+            id="with_extra_env",
+        ),
+    ],
+)
+def test_go_exec_env(
+    monkeypatch: pytest.MonkeyPatch,
+    vars: dict[str, str],
```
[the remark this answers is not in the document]
eskultety: Addressed in a different way.

tests/unit/package_managers/test_gomod.py
```diff
@@ -2609,3 +2610,32 @@ def test_get_go_work(self) -> None:
 
         mock_go.assert_called_once()
         assert mock_go.call_args[0][0] == ["work", "edit", "-json"]
+
+
+@pytest.mark.parametrize(
+    "vars, expected, extra_env",
```
slimreaper35: nitpick: `"vars, extra_vars, expected"` / `"env, extra_env, expected"` could be easier to read and would also be consistent with function parameters

tests/unit/package_managers/test_gomod.py
```diff
@@ -2612,30 +2612,36 @@ def test_get_go_work(self) -> None:
...
     extra_env: dict[str, str] | None,
 ) -> None:
     monkeypatch.setattr(os, "environ", vars)
 
-    if extra_env is None:
-        env = _go_exec_env()
-    else:
-        env = _go_exec_env(**extra_env)
-
+    env = _go_exec_env() if extra_env is None else _go_exec_env(**extra_env)
```
slimreaper35: nitpick: this change could be in the first commit :)

hermeto/core/package_managers/gomod.py
```diff
@@ -1017,6 +1017,16 @@ def _parse_packages(
     return iter(all_packages)
 
 
+def _go_exec_env(**extra_vars: str) -> dict[str, str]:
+    """Build the base environment for go command execution."""
```
a-ovchinnikov, line 1021: Technically speaking most of it still gets built in `_resolve_gomod`. This is an observation, not a blocker.

hermeto/core/package_managers/gomod.py
```diff
@@ -1017,6 +1017,16 @@ def _parse_packages(
...
+    env = {
+        "PATH": os.environ.get("PATH", ""),
+    }
+    if extra_vars:
```
a-ovchinnikov: L1025-1027 could be simplified:
```
return env | extra_vars
```
The way the function is defined `extra_vars` will always reference a dictionary, so it is safe to just update env from it. I must also admit that the way this is expected to be typed still mightily confuses me.

a-ovchinnikov (approved):
LGTM with one tiny non-blocking suggestion.

hermeto/core/package_managers/gomod.py
eskultety: I thought about extracting that too, but those variables depend on dynamic inputs and I didn't want to create a helper that "accepts everything" just to make sure, but yes, admittedly this tests just the very basics and doesn't even pretend to cover the go-specific variables. We now have a reasonable config, so we could maybe just pass request around as the top-level object to query data from in many callers - follow-up though if we ever need to get back to this.

hermeto/core/package_managers/gomod.py
eskultety: I'm wondering if I should have gone with a regular dict instead kwargs because kwargs are always weird in typing.

eskultety: Anyway, applied the suggestion.

PR #1242, opened 2026-01-20
fix: Upgrade createrepo-c to v1.2.2

adambkaplan:
Fedora 43 upgrades RPM v6.0.1, which is incompatible with `createrepo-c` v1.2.1. This causes the wheel build of createrepo-c to fail. Upgrading to v1.2.2 fixes the issue.

Assisted-by: Cursor


pyproject.toml
```diff
@@ -18,7 +18,7 @@ dynamic = ["version"]
 dependencies = [
   "aiohttp",
   "aiohttp-retry",
-  "createrepo-c ; sys_platform == 'linux'",
+  "createrepo-c==1.2.2 ; sys_platform == 'linux'",
```
eskultety: This hunk is redundant and can be dropped, a mere pip-compile update does the right thing, i.e.
```
$ >requirements.txt >requirements-extras.txt
$ nox -s pip-compile
```

slimreaper35 (approved):

PR #1243, opened 2026-01-20
doc: Add Troubleshooting to Contributing Guide

adambkaplan:
Setting up a development environment may require local complilation of python wheels - in particular `createrepo-c`. This change documents prerequisite C headers and other tools that need to be installed for the wheels compilation to succeed.

Assisted-by: Cursor

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -282,6 +282,40 @@ re-generate the set of dependencies, run the following in the repository and com
 nox -s pip-compile
 ```
 
+### Troubleshooting
+
+Below are common issues that you may encounter while developing your contribution to Hermeto.
+
+#### Unable to compile `createrepo-c` wheel
+
+In some circumstances your system may try to compile the `createrepo-c` Python wheel from source
+code. Unfortunately this requires several C-headers to be present on your system, which can be
+challenging to chase down.
+
+This situation can usually be resolved in one of two ways:
+
+1. Check if a new version of the `createrepo-c` wheel has been published. If yes, then update the
+   dependency per instructions above.
+2. Otherwise add the required C dependencies to your system. Below are the package names used by
+   Fedora with the necessary dependencies. Note that package names may not be the same on other
+   Linux distributions:
+   - `bzip2-devel` (provides `bzip2` C headers)
+   - `cmake`
+   - `gcc` (C compiler)
+   - `glib2-devel`
+   - `libmodulemd-devel`
+   - `libzstd-devel`
+   - `ninja-build` (build tool)
+   - `python3-devel` (Python development headers)
+   - `rpm-devel`
+   - `sqlite-devel`
+
+On Fedora systems, these can be installed using `dnf`:
+
+```shell
+sudo dnf install -y bzip2-devel cmake gcc glib2-devel libmodulemd-devel libzstd-devel ninja-build python3-devel rpm-devel sqlite-devel
+```
+
 ## Releasing
 
 To release a new version of Hermeto, simply create a [GitHub release](https://github.com/hermetoproject/hermeto/releases).


CONTRIBUTING.md
```diff
@@ -88,7 +88,27 @@ Whenever adding a new feature to Hermeto, it is important to keep these fundamen
...
+- `ninja-build` (build tool)
+- `python3-devel` (Python development headers)
+- `rpm-devel`
+- `sqlite-devel`
+
+On Fedora systems, these can be installed using `dnf`:
+
+```shell
+sudo dnf install -y bzip2-devel cmake gcc glib2-devel libmodulemd-devel libzstd-devel ninja-build python3-devel rpm-devel sqlite-devel
+```
```
eskultety: nitpick: I think this can be dropped, a developer should know how to install software on their particular system, whatever their distro of choice is.

CONTRIBUTING.md
```diff
@@ -88,7 +88,27 @@ Whenever adding a new feature to Hermeto, it is important to keep these fundamen
 
 ### Virtual environment
 
-Set up a virtual environment that has everything you will need for development:
```
eskultety, line 91: **Commit message:**

> Setting up a development environment may require local complilation of
> python wheels - in particular `createrepo-c`.

`createrepo-c` has been a weird dependency giving us headaches for a while. However, it's only a dependency we should never need devel deps to compile it, we should always make sure we can build with an existing wheel.

eskultety: so we either drop this completely or we should adjust the wording such that it doesn't sound like it's a mandatory requirement for a development setup.

adambkaplan: How do we avoid wheel re-compilation then? I hit this issue following the "dev environment setup" instructions word for word.

CONTRIBUTING.md
adambkaplan: Counter points:

- The naming of the packages that provide these headers vary across distros. Adding an example for the distro the community is most familiar with reduces friction.
- Can an AI agent do this? What is the probability that an AI agent, given this document as context, would hallucinate and fail to install the correct packages?

eskultety: > Can an AI agent do this? What is the probability that an AI agent, given this document as context, would hallucinate and fail to install the correct packages?

Why would you even expect anyone to do this, there's no precedent anywhere that such information should be part of docs. Moreover, basic system commands is something the AI agents have been capable of handling for a long time now (I'm definitely not advocating anyone giving an agent that kind of write permission or autonomy to their system, I'm just stating it as a fact).

> The naming of the packages that provide these headers vary across distros. Adding an example for the distro the community is most familiar with reduces friction.

I know, but you mentioned the name of packages as an example for a Fedora system a few lines above that already.

CONTRIBUTING.md
eskultety: > How do we avoid wheel re-compilation then?

You don't, but that's not something that usually happens, and it can happen with literally any project.

> so we either drop this completely or we should adjust the wording such that it doesn't sound like it's a mandatory requirement for a development setup.

What I intended to say with the "re-wording" and in my original comment but didn't explicitly state is that this is still NOT a hard requirement for development of Hermeto and it should not be, this is an unfortunate coincidence of a dependency that got updated before we got our scheduled dep update PR from dependabot. If you make it clear in the paragraph that one MAY need to install additional packages in case they're seeing some random createrepo-c issues during scratch builds/tests I'm good :).

Additionally, I'd also say that covering this in the docs (which is fine with the right wording) is still just treating the symptom rather than the core problem - `createrepo-c` should not be a core dependency same way tools like `go`, `yarn`, `cargo` aren't and unless you intend to build and use hermeto as a local CLI tool rather than consume it in form of a container you won't even notice. I remember having looked at that dependency in the past and solve this problem we have with it and the only thing I could think of was making it an optional dependency based on which backends the user or developer intends to work on. Note `createrepo-c` is also the only dep that prevents us from starting to test upstream with new Python releases in a reasonable time frame.

adambkaplan: > If you make it clear in the paragraph that one MAY need to install additional packages in case they're seeing some random createrepo-c issues during scratch builds/tests I'm good :).

:+1: I may just move this to a "troubleshooting" section, since it sounds like my issue was mainly due to timing wrt to the `createrepo-c` update.

>  - createrepo-c should not be a core dependency same way tools like go, yarn, cargo aren't and unless you intend to build and use hermeto as a local CLI tool 

My use case here was "inner loop" contributor development, where running as a local CLI tool in a Python virtual environment is the fastest way to iterate. It's been a long time since I've done daily Python development (almost a decade), and the tooling ecosystem has definitely evolved since then. Are there mechanisms to declare createrepo-c as a "dev dependency"?

:thinking: for developers on Linux machines, would testing within a [toolbox](https://github.com/containers/toolbox) potentially solve this issue?

eskultety: > My use case here was "inner loop" contributor development, where running as a local CLI tool in a Python virtual environment is the fastest way to iterate.

Totally, virtualenvs rocks for this use case.

> Are there mechanisms to declare createrepo-c as a "dev dependency"?

Well, the problem is `createrepo-c` isn't a dev dependency, not entirely, it's rather a runtime dependency. Unlike `go`, `yarn`, `npm`, etc. which are standalone tools that we don't control installation of in any way, we can consume `createrepo-c` from a wheel and I think that's the problem. IOW if we declare this a dev dependency then we're sending out the message that the tool is not required during runtime (well, as long as one works with the RPM backend) which is incorrect. If we were ever to package hermeto to e.g. PyPI, then it could not be dev dependency because it needs to be part of the built hermeto wheel (I guess or maybe not and just fail).
Alternatively, I thought of making `createrepo-c` an extra pip install target (e.g. `pip install . "[extra]"`) like you would for plugins, but then again, it's not a plugin because a plugin is an optional functionality, whereas in Hermeto's case one would not be able to process RPMs without it. That's where I left of because I simply didn't see a clean and a correct way out of this.

> 🤔 for developers on Linux machines, would testing within a [toolbox](https://github.com/containers/toolbox) potentially solve this issue?

I'm in the strong proponent's camp :), wanna give a one up for this: https://github.com/hermetoproject/hermeto/pull/1070?

adambkaplan: > I'm in the strong proponent's camp :), wanna give a one up for this: https://github.com/hermetoproject/hermeto/pull/1070?

Ha! I would if I had the time. I moved the C-headers bit to a new "Troubleshooting" section that is well below the main dev instructions, which I think is a reasonable compromise.

eskultety: > Ha! I would if I had the time.

@adambkaplan Sorry, wrong words - I meant, wanna upvote the PR? :) The work's already been done, it just needs an approval for us to start delivering a Hermeto toolbox image for devs ;).

PR #1244, opened 2026-01-21
bundler minor improvements

slimreaper35:

### DIFF
--- a/hermeto/core/package_managers/bundler/parser.py
+++ b/hermeto/core/package_managers/bundler/parser.py
@@ -25,10 +25,7 @@ BundlerDependency = GemDependency | GemPlatformSpecificDependency | GitDependenc
 ParseResult = list[BundlerDependency]
 
 
-def parse_lockfile(
-    package_dir: RootedPath, binary_filters: BundlerBinaryFilters | None = None
-) -> ParseResult:
-    """Parse a Gemfile.lock file and return a list of dependencies."""
+def _ensure_bundler_files_exist(package_dir: RootedPath) -> None:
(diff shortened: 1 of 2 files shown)


slimreaper35:
I cherry-picked some commits from my old bundler PR. The proper filtering feature is still to be done.

docs/bundler.md
```diff
@@ -37,16 +37,15 @@ As a result, gems with native extensions will be compiled from the source.
...
 
-```bash
-cd path-to-your-ruby-project
-hermeto fetch-deps '{"type": "bundler", "allow_binary": "true"}'
-```
-
-By default, the `allow_binary` option is disabled.
+> **WARNING**
+>
+> The `binary` field is not fully supported. When the `binary` field is specified,
```
taylormadore: nit: Not fully supported _**yet**_

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -29,7 +29,20 @@
       parsed_spec[:subpath] = spec.source.path
     end
 
-    parsed_specs << parsed_spec
+    existing_spec = parsed_specs.find { |s|
```
taylormadore: nit: Maybe we should consider building the complete hash inside each case branch rather than assembling it in pieces. Right now construction happens before, inside, and after the case. Co-locating it might make each type's structure easier to follow.

The new platform deduplication logic also only applies to the rubygems type, so keeping it inside that branch might make the scope more explicit. Feel free to ignore.

slimreaper35: Done.

PR #1247, opened 2026-01-23
docs: design: Add a design on bearer auth in the generic backend

eskultety:


docs/design/generic-auth.md
```diff
@@ -0,0 +1,342 @@
...
+which `<credential-field>` entries are valid within that block. This gives us
+flexibility to extend authentication support beyond HTTP Bearer Auth in the
+future.
+
+An example of a fully defined `artifacts.lock.yaml` with bearer token
+authentication then looks like:
+
+```yaml
+metadata:
+  version: "2.0" # version bump needed, Pydantic field validation is strict
```
eskultety, line 152: Technically, we could relax this restriction if needed, but old Hermeto's would still fail to parse the new schema.

eskultety:
Given the nature of the effort and the resulting document (mostly research) I don't have a PoC at hand which shouldn't be hard to cook once we have some agreement on the approach.

eskultety:
FWIW I only used the AI bot for native English text and table formatting, the research has been conducted by myself.

docs/design/generic-auth.md
```diff
@@ -0,0 +1,342 @@
...
+    ...
+```
+
+A simplified implementation looks like this (considering the above):
+1. Resolve the `auth` configuration maps for each lockfiles artifact in the
+   generic backend
+2. Read the corresponding environment variables to extract the values based
+   on the `header` field and populate the header information with the `value`
+   field's content
+3. Inject headers into the `aiohttp.ClientSession` context
```
taylormadore: Would we be injecting these headers into the session given that they could be different per-request?

docs/design/generic-auth.md
```diff
@@ -0,0 +1,342 @@
...
+backend's `artifacts.lock.yaml` schema. Alternatively, this could have been
+approached in a uniform manner by extending the input JSON's capabilities to
+carry credential related data. However, the generic backend has a fundamentally
+different requirement - per-URL (instead of per-host) authentication which the
+input JSON simply cannot express and for the rest of the backends realistically
+doesn't even need to.
+
+### Lockfile Schema
+
+The `artifacts.lock.yaml` schema is extended with an optional `auth` per a
```
taylormadore: I was concerned about the potential for repetition of identical auth blocks, but then I remembered that YAML anchors and aliases exist 😅

docs/design/generic-auth.md
```diff
@@ -0,0 +1,342 @@
+# Generic Fetcher - Support for HTTP Authentication
+
+## Background
+
+The generic fetcher downloads artifacts from arbitrary URLs specified in the
+`artifacts.lock.yaml` lockfile. Many artifact hosting services require
+authentication/authorization credentials to access private resources. This
+document describes the type of API authentication mechanism implemented in
```
a-ovchinnikov, line 8: Is it implemented or is it to be implemented?

docs/design/generic-auth.md
```diff
@@ -0,0 +1,342 @@
...
+
+The sections below provide a detailed overview of the most commonly available
+authentication schemes (in no particular order) and the level of Hermeto's
+involvement in the process.
+
+* HTTP Basic authentication
+* HTTP Bearer token
+* OAuth2
+* OIDC
+* other
```
a-ovchinnikov, line 22: Nitpick: `other` could be dropped since there is no actual other section.

docs/design/generic-auth.md
```diff
@@ -0,0 +1,342 @@
...
+Grant][rfc-6749-4-1] which requires interactive browser-based user consent, and
+the [Client Credentials Grant][rfc-6749-4-4] for machine-to-machine
+communication both of which are, in principle, the most relevant to Hermeto's
+use case. However, Hermeto cannot implement OAuth2 flows directly for the
+following reasons:
+1. They are **credential acquisition** mechanisms—they define how tokens are
+   obtained, not how they are attached to requests.
+2. The interactive nature of both the *Authorization Code Grant* flow (i.e. most
+   commonly requiring a browser redirection and user input to obtain a token)
+   and the *Client Credentials Grant* flow where the client has to authenticate
```
a-ovchinnikov: Does the client credentials grant flow also require some manual action? My impression was that it is specifically designed to be non-interactive (at least that's how I read the example in RFC6749 section 4.4.2).

docs/design/generic-auth.md
slimreaper35: Would it make sense to have a top-level auth block as well? -> The auth block for a specific URL would override the top-level one.

docs/design/generic-auth.md
```diff
@@ -0,0 +1,342 @@
...
+Each *type* has its own set of required and optional fields.
+
+##### Types:
+
+**`bearer`** - HTTP header-based token authentication:
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `header` | string | No | HTTP header name (e.g., `PRIVATE-TOKEN`). Defaults to `Authorization` |
+| `value` | string | Yes | HTTP Header value using env variable placeholders and custom strings |
```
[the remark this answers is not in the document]
eskultety: Okay, okay, chill, I'll update it, jeee the bot's on a grammar spree :D :D :D.

docs/design/generic-auth.md
eskultety: Hmm, this is a fantastic question - we may need to batch the generic artifacts into different aiohttp sessions rather than a single one. Question - do you think it's an architectural blocker that needs to be resolved on the design doc level or is it abstract enough that you'd leave this as an implementation detail on how exactly we'd inject the headers (because injecting them is pretty much the only way with bearer auth)?

docs/design/generic-auth.md
eskultety: > Would it make sense to have a top-level auth block as well? -> The auth block for a specific URL would override the top-level one.

I think ^this particular case could be solved with the potential expansion of the input JSON described in the _Input JSON Authentication_ section. @slimreaper35 would that work for you? I mean the global level would reside in the input JSON which would also cover the other backends and if needed the `artifacts.lock.yaml` would override the specific per-URL cases?

docs/design/generic-auth.md
eskultety: I chose the present tense specifically because this is not an ADR but a binding design so to speak. So once the implementation is in place, the design doc narrative will match the reality, how about that?

docs/design/generic-auth.md
eskultety: I didn't want the make the list finite in the sense that I think there are only 4 options.

eskultety: I'll drop it.

docs/design/generic-auth.md
a-ovchinnikov: That makes sense.

docs/design/generic-auth.md
a-ovchinnikov: My only concern here is with you starting the paragraph with "The section [provides detailed overview of these items]" -- it is pretty hard to provide a detailed overview for "other". OTOH adding a sentence along the lines of "The list above is not exhaustive, captures only those mechanisms which are the most important for Hermeto and could be amended in the future" would communicate the same intent unambiguously.

docs/design/generic-auth.md
eskultety: Funnily enough the AI bot raised the same concern. It's poorly worded, I need to change that. What I wanted to convey was - it requires extra work on Hermeto's side to obtain an auth token on behalf of the user and THEN use it instead of just consume the token and inject the headers.

docs/design/generic-auth.md
slimreaper35: Yes

docs/design/generic-auth.md
eskultety: @taylormadore @a-ovchinnikov @slimreaper35 any comments on this one? Are we okay with schema 2.0 rather than 1.1?

docs/design/generic-auth.md
eskultety: Updated.

docs/design/generic-auth.md
taylormadore: I would just word it as though we are injecting the headers into the request rather than the session. The session is an implementation detail whereas those headers will need to be in the request regardless of how we implement it on the backend.

eskultety: Reworded.

eskultety:
All comments addressed, please re-review.

slimreaper35 (approved):

PR #1248, opened 2026-01-23
Proxy support in npm

a-ovchinnikov:
This PR introduces changes necessary to make npm aware of proxies. The change was tested manually with a proxy requiring authentication and was found to work as expected. Automated testing for the feature is TBD.

The feature was tested with a proxy requiring authorization. To test the feature I first ran Hermeto in a venv without specifying a proxy URL:

```
((python-3-12) ) $ env | grep -c HERMETO                                                                                                          
0
((python-3-12) ) $ hermeto fetch-deps npm
...
INFO All dependencies fetched successfully \o/
((python-3-12) ) $ head -15 hermeto-output/bom.json                                                                                               
{                                                                                                                                                 
  "bomFormat": "CycloneDX",                                                                                                                       
  "components": [                                                                                                                                 
    {                                                                                                                                             
      "name": "@types/prop-types",                                                                                                                
      "purl": "pkg:npm/%40types/prop-types@15.7.5",                                                                                               
      "version": "15.7.5",                                                                                                                        
      "properties": [                                                                                                                             
        {                                                                                                                                         
          "name": "hermeto:found_by",                                                                                                             
          "value": "hermeto"                                                                                                                      
        }                                                                                                                                         
      ],                                                                                                                                          
      "type": "library"                                                                                                                           
    },
```

Then I set a proxy URL and repeated the experiment. My requests were directed to the proxy and rejected (as expected):
```
((python-3-12) ) $ export HERMETO_NPM__PROXY_URL='....'
((python-3-12) ) $ env | grep -c HERMETO                                                                                                          
1 
((python-3-12) ) $ hermeto fetch-deps npm
ERROR Unsuccessful download: <*****>/@types/react/-/react-18.2.18.tgz      
ERROR FetchError: exception_name: ClientResponseError, details: 401, message='Unauthorized', url='<*****>/@types/react/-/react-18.2.18.tgz'                                                                                  
Error: FetchError: exception_name: ClientResponseError, details: 401, message='Unauthorized', url='<*****>/@types/react/-/react-18.2.18.tgz'                                                                                                         
  The error might be intermittent, please try again.                                                                                              
  If the issue seems to be on the hermeto side, please contact the maintainers.
```

After that I repeated the same experiment after setting login and password. This resulted in a successful fetch:
```
((python-3-12) ) $ export HERMETO_NPM__PROXY_LOGIN='....'
((python-3-12) ) $ export HERMETO_NPM__PROXY_PASSWORD='....'
((python-3-12) ) $ env | grep -c HERMETO                                                                                                          
3
((python-3-12) ) $ hermeto fetch-deps npm
...
INFO All dependencies fetched successfully \o/
((python-3-12) ) $ head -21 hermeto-output/bom.json                                                                                               
{                                                                                                                                                 
  "bomFormat": "CycloneDX",                                                                                                                       
  "components": [                                                                                                                                 
    {                                                                                                                                             
      "name": "@types/prop-types",                                                                                                                
      "purl": "pkg:npm/%40types/prop-types@15.7.5",                                                                                               
      "version": "15.7.5",                                                                                                                        
      "properties": [                                                                                                                             
        {                                                                                                                                         
          "name": "hermeto:found_by",                                                                                                             
          "value": "hermeto"                                                                                                                      
        }                                                                                                                                         
      ],                                                                                                                                          
      "type": "library",                                                                                                                          
      "externalReferences": [                                                                                                                     
        {                                                                                                                                         
          "url": "<*****>",                                                                       
          "type": "source-distribution",
          "comment": "proxy URL"                                                                                         
        }                                                                                                                                         
      ]                                                                                                                                           
    },
```

Now the sources were fetched from a proxy and that fact was reflected in SBOM.

### DIFF
--- a/hermeto/core/config.py
+++ b/hermeto/core/config.py
@@ -159,6 +159,10 @@ class ProxyMixin(BaseModel):
         return self
 
 
+class NpmSettings(ProxyMixin, extra="forbid"):
+    """Npm settings."""
+
+
 class Config(BaseSettings):
     """Singleton that provides default configuration for the application process."""
 
@@ -173,6 +177,7 @@ class Config(BaseSettings):
 
     pip: PipSettings = PipSettings()
     yarn: YarnSettings = YarnSettings()
+    npm: NpmSettings = NpmSettings()
     gomod: GomodSettings = GomodSettings()
     http: HttpSettings = HttpSettings()
     runtime: RuntimeSettings = RuntimeSettings()
--- a/hermeto/core/package_managers/npm.py
+++ b/hermeto/core/package_managers/npm.py
@@ -4,14 +4,16 @@ import fnmatch
 import functools
 import json
 import logging
+from functools import partial
 from pathlib import Path
 from typing import Any, Literal, NewType, TypedDict
 from urllib.parse import urlparse
 
+import aiohttp
 from packageurl import PackageURL
 
 from hermeto.core.checksum import ChecksumInfo, must_match_any_checksum
-from hermeto.core.config import get_config
+from hermeto.core.config import ProxyUrl, get_config
 from hermeto.core.errors import (
     LockfileNotFound,
     MissingChecksum,
@@ -22,7 +24,7 @@ from hermeto.core.errors import (
 from hermeto.core.models.input import Request
 from hermeto.core.models.output import ProjectFile, RequestOutput
 from hermeto.core.models.property_semantics import PropertySet
-from hermeto.core.models.sbom import Component
+from hermeto.core.models.sbom import PROXY_COMMENT, PROXY_REF_TYPE, Component, ExternalReference
 from hermeto.core.package_managers.general import async_download_files
 from hermeto.core.rooted_path import RootedPath
 from hermeto.core.scm import RepoID, clone_as_tarball, get_repo_id
@@ -50,6 +52,7 @@ class NpmComponentInfo(TypedDict):
     version: str | None
     dev: bool
     bundled: bool
+    external_refs: list[ExternalReference] | None
     missing_hash_in_file: Path | None
 
 
@@ -263,11 +266,14 @@ class PackageLock:
             "dev": False,
             "bundled": False,
             "missing_hash_in_file": None,
+            "external_refs": None,
         }
 
     def get_sbom_components(self) -> list[NpmComponentInfo]:
         """Return a list of dicts with sbom component data."""
         packages = self._packages
+        proxy_common = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
+        proxy_url = get_config().npm.proxy_url
 
         def to_component(package: Package) -> NpmComponentInfo:
             purl = self._purlifier.get_purl(
@@ -275,6 +281,7 @@ class PackageLock:
             ).to_string()
 
             missing_hash_in_file = None
+            external_refs = None
             if package.resolved_url:  # dependency is not bundled
                 resolved_url = _normalize_resolved_url(package.resolved_url)
                 dep_type = _classify_resolved_url(resolved_url)
@@ -282,6 +289,8 @@ class PackageLock:
                 if not package.integrity:
                     if dep_type in ("registry", "https"):
                         missing_hash_in_file = Path(self._lockfile_path.subpath_from_root)
+                if dep_type == "registry" and proxy_url is not None:
+                    external_refs = [ExternalReference(url=str(proxy_url), **proxy_common)]
 
             component: NpmComponentInfo = {
                 "name": package.name,
@@ -290,6 +299,7 @@ class PackageLock:
                 "dev": package.dev,
                 "bundled": package.bundled,
                 "missing_hash_in_file": missing_hash_in_file,
+                "external_refs": external_refs,
             }
 
             return component
@@ -487,6 +497,28 @@ def _clone_repo_pack_archive(
     return download_path
 
 
+def _patch_url_to_point_to_a_proxy(url: NormalizedUrl, proxy_url: ProxyUrl) -> NormalizedUrl:
+    # Convert 'https://registry.npmjs.org/accepts/-/accepts-1.3.8.tgz'
+    # to '<proxyaddress>/accepts/-/accepts-1.3.8.tgz'.
+    s_proxy_url = str(proxy_url)  # mypy becomes really upset when "proxy_url" gets reused here
+    s_proxy_url = s_proxy_url if s_proxy_url[-1] == "/" else s_proxy_url + "/"
+    url_path = urlparse(url).path[1:]  # Don't need the leading / anymore
+    return NormalizedUrl(s_proxy_url + url_path)
+
+
+async def _async_download_tar(files_to_download_list: list[dict[str, dict[str, Any]]]) -> None:
+    ftdl = [e for e in files_to_download_list if e]
+    if not ftdl:
+        return
+    # NOTE: when present proxy auth is the same for all packages accessible
+    # through a proxy.
+    auth = lambda ftd: next(iter(ftd.values()))["proxy_auth"]
+    ftd = lambda ftd: {it["fetch_url"]: it["download_path"] for it in ftd.values()}
+    adf = partial(async_download_files, concurrency_limit=get_config().runtime.concurrency_limit)
+
+    await asyncio.gather(*[adf(files_to_download=ftd(f), auth=auth(f)) for f in ftdl])
+
+
 def _get_npm_dependencies(
     download_dir: RootedPath, deps_to_download: dict[str, dict[str, str | None]]
 ) -> dict[NormalizedUrl, RootedPath]:
@@ -502,9 +534,13 @@ def _get_npm_dependencies(
     """
     files_to_download: dict[str, dict[str, Any]] = {}
     download_paths = {}
+    config = get_config()
+
     for url, info in deps_to_download.items():
         url = _normalize_resolved_url(url)
+        fetch_url = url
         dep_type = _classify_resolved_url(url)
+        proxy_auth = None
 
         if dep_type == "file":
             continue
@@ -516,6 +552,13 @@ def _get_npm_dependencies(
                     "/", "-"
                 )
                 download_paths[url] = download_dir.join_within_root(archive_name)
+                if config.npm.proxy_url is not None:
+                    fetch_url = _patch_url_to_point_to_a_proxy(url, config.npm.proxy_url)
+                    if config.npm.proxy_login and config.npm.proxy_password:
+                        proxy_auth = aiohttp.BasicAuth(
+                            config.npm.proxy_login,
+                            config.npm.proxy_password,
+                        )
             else:  # dep_type == "https"
                 if info["integrity"]:
                     algorithm, digest = ChecksumInfo.from_sri(info["integrity"])
@@ -537,17 +580,17 @@ def _get_npm_dependencies(
                 directory.mkdir(parents=True, exist_ok=True)
 
             files_to_download[url] = {
+                "fetch_url": fetch_url,
                 "download_path": download_paths[url],
                 "integrity": info["integrity"],
+                "proxy_auth": proxy_auth,
             }
 
-    # Asynchronously download tar files
-    asyncio.run(
-        async_download_files(
-            {url: item["download_path"] for (url, item) in files_to_download.items()},
-            get_config().runtime.concurrency_limit,
-        )
-    )
+    files_with_auth = {k: v for k, v in files_to_download.items() if v["proxy_auth"] is not None}
+    files_without_auth = {k: v for k, v in files_to_download.items() if v["proxy_auth"] is None}
+
+    asyncio.run(_async_download_tar([files_with_auth, files_without_auth]))
+
     # Check integrity of downloaded packages
     for url, item in files_to_download.items():
         if item["integrity"]:
@@ -661,6 +704,7 @@ def _generate_component_list(component_infos: list[NpmComponentInfo]) -> list[Co
                 npm_development=component_info["dev"],
                 missing_hash_in_file=missing_hash,
             ).to_properties(),
+            external_references=component_info["external_refs"],
         )
 
     return [to_component(component_info) for component_info in component_infos]
--- a/tests/unit/package_managers/test_npm.py
+++ b/tests/unit/package_managers/test_npm.py
@@ -11,6 +11,7 @@ from packageurl import PackageURL
 
 from hermeto import APP_NAME
 from hermeto.core.checksum import ChecksumInfo
+from hermeto.core.config import NpmSettings
 from hermeto.core.errors import (
     LockfileNotFound,
     PackageRejected,
@@ -646,6 +647,7 @@ class TestPurlifier:
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
                 {
                     "name": "bar",
@@ -654,6 +656,7 @@ class TestPurlifier:
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
             ],
             [
@@ -670,6 +673,7 @@ class TestPurlifier:
                     "bundled": False,
                     "dev": True,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
             ],
             [
@@ -693,6 +697,7 @@ class TestPurlifier:
                     "bundled": True,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
             ],
             [
@@ -716,6 +721,7 @@ class TestPurlifier:
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": Path("path/to/foo/package-lock.json"),
+                    "external_refs": None,
                 },
             ],
             [
@@ -756,6 +762,7 @@ def test_generate_component_list(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     },
                     "dependencies": [
                         {
@@ -765,6 +772,7 @@ def test_generate_component_list(
                             "bundled": False,
                             "dev": False,
                             "missing_hash_in_file": None,
+                            "external_refs": None,
                         }
                     ],
                     "projectfiles": [
@@ -803,6 +811,7 @@ def test_generate_component_list(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     },
                     "dependencies": [
                         {
@@ -812,6 +821,7 @@ def test_generate_component_list(
                             "bundled": False,
                             "dev": False,
                             "missing_hash_in_file": None,
+                            "external_refs": None,
                         }
                     ],
                     "projectfiles": [
@@ -834,6 +844,7 @@ def test_generate_component_list(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     },
                     "dependencies": [
                         {
@@ -843,6 +854,7 @@ def test_generate_component_list(
                             "bundled": False,
                             "dev": False,
                             "missing_hash_in_file": None,
+                            "external_refs": None,
                         }
                     ],
                     "dependencies_to_download": {
@@ -968,6 +980,7 @@ def test_resolve_npm_validation(
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
                 "dependencies": [
                     {
@@ -977,6 +990,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,  # correct since integrity is missing from dependencies but is included in packages section
+                        "external_refs": None,
                     }
                 ],
                 "projectfiles": [
@@ -1038,6 +1052,7 @@ def test_resolve_npm_validation(
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
                 "dependencies": [
                     {
@@ -1047,6 +1062,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     },
                     {
                         "name": "baz",
@@ -1055,6 +1071,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     },
                     {
                         "name": "spam",
@@ -1063,6 +1080,7 @@ def test_resolve_npm_validation(
                         "bundled": True,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     },
                 ],
                 "projectfiles": [
@@ -1104,6 +1122,7 @@ def test_resolve_npm_validation(
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
                 "dependencies": [
                     {
@@ -1113,6 +1132,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     }
                 ],
                 "projectfiles": [
@@ -1154,6 +1174,7 @@ def test_resolve_npm_validation(
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
                 "dependencies": [
                     {
@@ -1163,6 +1184,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     }
                 ],
                 "projectfiles": [
@@ -1212,6 +1234,7 @@ def test_resolve_npm_validation(
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
                 "dependencies": [
                     {
@@ -1221,6 +1244,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     },
                     {
                         "name": "spam",
@@ -1229,6 +1253,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     },
                 ],
                 "projectfiles": [
@@ -1271,6 +1296,7 @@ def test_resolve_npm_validation(
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
                 "dependencies": [
                     {
@@ -1280,6 +1306,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": Path("package-lock.json"),
+                        "external_refs": None,
                     }
                 ],
                 "projectfiles": [
@@ -1315,6 +1342,7 @@ def test_resolve_npm_validation(
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
                 "dependencies": [
                     {
@@ -1324,6 +1352,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": Path("package-lock.json"),
+                        "external_refs": None,
                     }
                 ],
                 "projectfiles": [
@@ -1372,6 +1401,7 @@ def test_resolve_npm_validation(
                     "bundled": False,
                     "dev": False,
                     "missing_hash_in_file": None,
+                    "external_refs": None,
                 },
                 "dependencies": [
                     {
@@ -1381,6 +1411,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": Path("package-lock.json"),
+                        "external_refs": None,
                     },
                     {
                         "name": "baz",
@@ -1389,6 +1420,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     },
                     {
                         "name": "spam",
@@ -1397,6 +1429,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": None,
+                        "external_refs": None,
                     },
                     {
                         "name": "eggs",
@@ -1405,6 +1438,7 @@ def test_resolve_npm_validation(
                         "bundled": False,
                         "dev": False,
                         "missing_hash_in_file": Path("package-lock.json"),
+                        "external_refs": None,
                     },
                 ],
                 "projectfiles": [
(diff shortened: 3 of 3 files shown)


hermeto/core/config.py
```diff
@@ -118,6 +118,14 @@ class ProxySettings(BaseModel, extra="forbid"):
     fallback_to_sources: bool = False
 
 
+class NpmSettings(BaseModel, extra="forbid"):
+    """Npm settings."""
+
+    proxy_url: str | None = None
+    proxy_login: str | None = None
```
taylormadore, line 144: In the npm module we only configure auth if both login and password are set. Is that worth validating when processing the config?

hermeto/core/package_managers/npm.py
```diff
@@ -496,21 +521,31 @@ def _get_npm_dependencies(
     :return: Dictionary of Resolved URL dependencies with downloaded paths
     """
     files_to_download: dict[str, dict[str, Any]] = {}
-    download_paths = {}
+    download_paths: dict[NormalizedUrl, DownloadPaths] = {}  # dropping the hint upsets mypy.
```
taylormadore: Since actual_url isn't used outside of this function, do we need the DownloadPaths type at all?

hermeto/core/config.py
a-ovchinnikov: Yes, that's a good point, thank you.

hermeto/core/package_managers/npm.py
a-ovchinnikov: ~Adding a dictionary subtype with explicitly listed keys was the only way to stop mypy from complaining about values types mismatches. The only alternatives I see so far is a local type definition or constructing some inline monstrosity which would reduce readability even further. _Edit_: type name is rather arbitrary and could be improved.~ _Edit_: after discussing this offline it turned out that the type is not needed after all. The next revision addresses this.

a-ovchinnikov:
Latest revision: a rebase on top of main.

hermeto/core/config.py
a-ovchinnikov: Done.

hermeto/core/config.py
```diff
@@ -112,6 +113,22 @@ class RuntimeSettings(BaseModel, extra="forbid"):
     concurrency_limit: int = 5
 
 
+class NpmSettings(BaseModel, extra="forbid"):
+    """Npm settings."""
+
+    proxy_url: str | None = None
```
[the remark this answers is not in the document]
a-ovchinnikov: This is a fair observation, I'll follow it.

hermeto/core/package_managers/npm.py
```diff
@@ -264,25 +266,30 @@ def get_main_package(self) -> NpmComponentInfo:
...
+            external_refs = None
             if package.resolved_url:  # dependency is not bundled
                 resolved_url = _normalize_resolved_url(package.resolved_url)
                 dep_type = _classify_resolved_url(resolved_url)
 
                 if not package.integrity:
                     if dep_type in ("registry", "https"):
                         missing_hash_in_file = Path(self._lockfile_path.subpath_from_root)
+                if dep_type == "registry" and (proxy_url := get_config().npm.proxy_url):
+                    external_refs = [ExternalReference(url=proxy_url, **eref_rest)]
```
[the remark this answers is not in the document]
a-ovchinnikov: This one is valid, updating.

hermeto/core/package_managers/npm.py
```diff
@@ -488,6 +496,17 @@ def _clone_repo_pack_archive(
...
 
+def _patch_url_to_point_to_a_proxy(url: NormalizedUrl) -> NormalizedUrl:
+    # Convert 'https://registry.npmjs.org/accepts/-/accepts-1.3.8.tgz'
+    # to '<proxyaddress>/accepts/-/accepts-1.3.8.tgz'.
+    # Proxy URL is expected to have a trailing /
+    config = get_config()
+    url_path = urlparse(url).path[1:]  # Don't need the leading /
+    # mypy insists that typing is incompatible despite this helper being called only when
+    # config.npm.proxy_url is a string.
+    return config.npm.proxy_url + url_path  # type: ignore
```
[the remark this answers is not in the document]
a-ovchinnikov: Nope.

hermeto/core/package_managers/npm.py
```diff
@@ -538,15 +561,23 @@ def _get_npm_dependencies(
...
         async_download_files(
-            {url: item["download_path"] for (url, item) in files_to_download.items()},
+            {
+                item["actual_url"]: item["download_path"]
+                for (url, item) in files_to_download.items()
+            },
             get_config().runtime.concurrency_limit,
+            auth=proxy_auth,
         )
     )
```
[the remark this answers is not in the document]
a-ovchinnikov: Nope.

hermeto/core/package_managers/npm.py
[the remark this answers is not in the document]
a-ovchinnikov: Not anymore.

hermeto/core/package_managers/general.py
```diff
@@ -114,6 +118,7 @@ async def async_download_files(
     files_to_download: dict[str, StrPath],
     concurrency_limit: int,
     ssl_context: ssl.SSLContext | None = None,
+    auth: aiohttp.BasicAuth | None = None,
```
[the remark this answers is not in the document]
a-ovchinnikov: Valid.

hermeto/core/config.py
```diff
@@ -112,6 +113,50 @@ class RuntimeSettings(BaseModel, extra="forbid"):
...
+        elif str(self.proxy_url)[-1] == "/":
+            return self
+        else:
+            self.proxy_url = HttpUrl(url=str(self.proxy_url) + "/")
+            return self
+
+    @model_validator(mode="after")
+    def _validate_login_and_password_both_set(self) -> Self:
+        if self.proxy_login is not None and self.proxy_password is None:
+            raise ValueError("Proxy password must be set when proxy login is set")
```
taylormadore: Why ValueError over InvalidInput?

tests/unit/package_managers/test_npm.py
```diff
@@ -1838,3 +1877,36 @@ def test_update_package_json_files(
...
+@pytest.mark.parametrize(
+    "proxy_url",
+    [
+        pytest.param("https://example.com", id="proxy_url_missing_trailing_slash"),
+        pytest.param("https://example.com/", id="proxy_url_having_slash"),
+        pytest.param("https://example.com/some/path", id="proxy_url_having_slash_in_path"),
+        pytest.param("https://example.com/some/path/", id="proxy_url_having_slash_in_path"),
+    ],
+)
+def test_npm_settings_correctly_handles_trailig_slash_in_proxy_urls(
```
taylormadore: Typo in function name

hermeto/core/config.py
```diff
@@ -112,6 +113,50 @@ class RuntimeSettings(BaseModel, extra="forbid"):
     concurrency_limit: int = 5
 
 
+class NpmSettings(BaseModel, extra="forbid"):
```
taylormadore: What do you think about extracting some of the proxy-related Config logic to a separate construct and/or using the Annotated pattern so that we don't need to repeat it for other package managers? I think the proxy_url/proxy_auth should be quite similar for most of them (Go perhaps being the exception). I had Claude whip me up a quick "all-of-the-above" example to try to show what I'm talking about:

```python
from typing import Annotated
from pydantic import AfterValidator, BaseModel, HttpUrl, model_validator
from typing_extensions import Self

# Standalone validators (implementation details omitted)
def _ensure_trailing_slash(url: HttpUrl | None) -> HttpUrl | None: ...
def _reject_embedded_credentials(url: HttpUrl | None) -> HttpUrl | None: ...

# Reusable annotated type - chains validators declaratively
ProxyUrl = Annotated[
    HttpUrl | None,
    AfterValidator(_ensure_trailing_slash),
    AfterValidator(_reject_embedded_credentials),
]

class ProxyAuthMixin(BaseModel):
    proxy_url: ProxyUrl = None
    proxy_login: str | None = None
    proxy_password: str | None = None

    @model_validator(mode="after")
    def _validate_proxy_credentials(self) -> Self:
        # Ensures login/password are both set or both unset
        ...

class NpmSettings(ProxyAuthMixin, BaseModel, extra="forbid"):
    pass

class PipSettings(ProxyAuthMixin, BaseModel, extra="forbid"):
    ignore_dependencies_crates: bool = False
```

hermeto/core/package_managers/npm.py
```diff
@@ -538,15 +559,23 @@ def _get_npm_dependencies(
...
     asyncio.run(
         async_download_files(
-            {url: item["download_path"] for (url, item) in files_to_download.items()},
+            {
+                item["actual_url"]: item["download_path"]
+                for (url, item) in files_to_download.items()
+            },
             get_config().runtime.concurrency_limit,
+            auth=proxy_auth,
         )
```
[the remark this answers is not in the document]
taylormadore: I had the same concern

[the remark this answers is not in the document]
a-ovchinnikov: This is a valid observation.

tests/unit/package_managers/test_npm.py
a-ovchinnikov: Fixed

hermeto/core/config.py
a-ovchinnikov: Yes, this definitely makes sense to do, I was planning to do it once there is more than one `*Settings` class present, but perhaps now is also a good time for that.

hermeto/core/config.py
a-ovchinnikov: No specific reason, `InvalidInput` conveys the message better.

a-ovchinnikov:
Latest revision: a couple of tests added, refactoring for ProxyUrlMixin will be in the next revision.

hermeto/core/config.py
a-ovchinnikov: Done (however I kept my names for now).

hermeto/core/config.py
```diff
@@ -112,6 +113,62 @@ class RuntimeSettings(BaseModel, extra="forbid"):
...
+
+def _ensure_proxy_url_ends_with_a_slash(url: MaybeHttpUrl) -> MaybeHttpUrl:
+    # NOTE: HttpUrl _does not_ normalize URLs with path
+    # (see e.g. https://github.com/pydantic/pydantic/issues/7186
+    # for details on expected behavior), hence this validator.
+    if url is None:
+        return None
+    elif str(url)[-1] == "/":
+        return HttpUrl(url=url)
+    return HttpUrl(url=str(url) + "/")
```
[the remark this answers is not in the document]
a-ovchinnikov: Nope.

a-ovchinnikov:
The latest iteration: abstracted `ProxyMixin` away for future reuse. Will take another look and decide if the second commit is worth splitting.

hermeto/core/package_managers/npm.py
```diff
@@ -503,9 +536,12 @@ def _get_npm_dependencies(
     """
     files_to_download: dict[str, dict[str, Any]] = {}
     download_paths = {}
+    config = get_config()
+
     for url, info in deps_to_download.items():
-        url = _normalize_resolved_url(url)
+        actual_url = url = _normalize_resolved_url(url)
```
taylormadore: nit: I wonder if `fetch_url` might be slightly more clear

hermeto/core/package_managers/npm.py
```diff
@@ -488,6 +496,31 @@ def _clone_repo_pack_archive(
...
+    url_path = urlparse(url).path[1:]  # Don't need the leading /
+    return NormalizedUrl(str(config.npm.proxy_url) + url_path)
+
+
+def _async_download_tar(files_to_download: dict) -> None:
+    if not files_to_download:
+        return
+    # NOTE: when present proxy auth is the same for all packages accessible
+    # through a proxy.
+    auth = next(iter(files_to_download.values()))["proxy_auth"] if files_to_download else None
```
taylormadore: nit: empty files_to_download already checked above

taylormadore (approved):
LGTM

hermeto/core/package_managers/npm.py
a-ovchinnikov: Fixed.

hermeto/core/package_managers/npm.py
a-ovchinnikov: Yes, `fetch_url` is definitely better, updated.

hermeto/core/package_managers/general.py
```diff
@@ -114,11 +118,13 @@ async def async_download_files(
...
```
eskultety, line 121: I guess from a PoC or initial implementation this is okay, but basic auth isn't going to cut it long term. In context of my recent auth investigation work for the generic backend I think we could prep the code base for the bearer token auth (https://github.com/hermetoproject/hermeto/blob/main/docs/design/generic-auth.md#integrating-with-aiohttp) by going with header injection directly even for basic auth - this will cause fewer conflicts (if any) with the future work that is going to be done on the bearer auth front, shouldn't be IMO too hard to rewrite this in the other direction. Basically what I'm trying to avoid is touching the interface too often even though it's an internal one so as not the affect all backends twice, WDYT of the suggestion?

eskultety: Since the `auth` field is only `BasicAuth` we could probably create our own abstraction for our own purposes (extensible to support bearer) and use that in all high-level actions, translating to header injection when we get down to `aiohttp` calls.

hermeto/core/config.py
```diff
@@ -112,6 +113,62 @@ class RuntimeSettings(BaseModel, extra="forbid"):
...
+            "proxy access.",
+        )
+    return url
+
+
+ProxyUrl = Annotated[
+    MaybeHttpUrl,
+    AfterValidator(_ensure_proxy_url_ends_with_a_slash),
+    AfterValidator(_proxy_url_must_not_contain_credentials),
+]
```
eskultety: I think this deserves a standalone tiny introductory commit.

hermeto/core/config.py
```diff
@@ -112,6 +113,62 @@ class RuntimeSettings(BaseModel, extra="forbid"):
     concurrency_limit: int = 5
 
 
+MaybeHttpUrl = HttpUrl | None
```
eskultety: nitpick: we might be stretching it here a bit too much with the maybe type, if we wanted to be consistent we'd have to do this for many other types, personally not a fan of this practice, but won't block on it.

hermeto/core/config.py
```diff
@@ -112,6 +113,62 @@ class RuntimeSettings(BaseModel, extra="forbid"):
...
+    proxy_login: str | None = None
+    proxy_password: str | None = None
+
+    @model_validator(mode="after")
+    def _validate_login_and_password_both_set(self) -> Self:
+        if self.proxy_login is not None and self.proxy_password is None:
+            raise InvalidInput("Proxy password must be set when proxy login is set")
+        if self.proxy_login is None and self.proxy_password is not None:
+            raise InvalidInput("Proxy login must be set when proxy password is set")
+        return self
```
eskultety, line 153: Do we want to extend this to support just a single value - token - right away?

hermeto/core/config.py
```diff
@@ -112,6 +113,62 @@ class RuntimeSettings(BaseModel, extra="forbid"):
     concurrency_limit: int = 5
 
 
+MaybeHttpUrl = HttpUrl | None
+
+
+def _ensure_proxy_url_ends_with_a_slash(url: MaybeHttpUrl) -> MaybeHttpUrl:
+    # NOTE: HttpUrl _does not_ normalize URLs with path
+    # (see e.g. https://github.com/pydantic/pydantic/issues/7186
+    # for details on expected behavior), hence this validator.
```
eskultety: Can you elaborate on this some more for me with an example demonstration? One that exhibits the problem including path.

hermeto/core/package_managers/npm.py
```diff
@@ -488,6 +496,31 @@ def _clone_repo_pack_archive(
...
+    return NormalizedUrl(str(config.npm.proxy_url) + url_path)
+
+
+def _async_download_tar(files_to_download: dict) -> None:
+    if not files_to_download:
+        return
+    # NOTE: when present proxy auth is the same for all packages accessible
+    # through a proxy.
+    auth = next(iter(files_to_download.values()))["proxy_auth"]
+    ftd = {it["fetch_url"]: it["download_path"] for _, it in files_to_download.items()}
```
eskultety: > `for _, it in files_to_download.items()`

`for it in  files_to_download.values()` might be clearer.

hermeto/core/package_managers/npm.py
```diff
@@ -517,6 +553,13 @@ def _get_npm_dependencies(
                     "/", "-"
                 )
                 download_paths[url] = download_dir.join_within_root(archive_name)
+                if config.npm.proxy_url:
+                    fetch_url = _patch_url_to_point_to_a_proxy(url)
+                if config.npm.proxy_login and config.npm.proxy_password:
```
eskultety: If `config.npm.proxy_url == None` you'd send the proxy credentials to the origin registry, effectively leaking them.

hermeto/core/config.py
```diff
@@ -112,6 +113,62 @@ class RuntimeSettings(BaseModel, extra="forbid"):
...
+
+class ProxyMixin(BaseModel):
+    """Proxy URL and authorization mixin."""
+
+    proxy_url: ProxyUrl = None
+    proxy_login: str | None = None
+    proxy_password: str | None = None
+
+    @model_validator(mode="after")
+    def _validate_login_and_password_both_set(self) -> Self:
```
eskultety, line 148: The validator doesn't check the connection credentials <-> proxy URL but because `proxy_url: HttpUrl | None` it should IMO perform this validation only if the URL is set. Or separate the modeling concept of default registry URL vs a proxy, but this way we're checking that we have both creds even though there might be no need for it.

hermeto/core/config.py
```diff
@@ -112,6 +113,62 @@ class RuntimeSettings(BaseModel, extra="forbid"):
...
+        if self.proxy_login is not None and self.proxy_password is None:
+            raise InvalidInput("Proxy password must be set when proxy login is set")
+        if self.proxy_login is None and self.proxy_password is not None:
+            raise InvalidInput("Proxy login must be set when proxy password is set")
+        return self
+
+
+class NpmSettings(ProxyMixin, BaseModel, extra="forbid"):
```
eskultety: Redundant `BaseModel`, `ProxyMixin` already inherits from it.

tests/unit/package_managers/test_npm.py
```diff
@@ -1838,3 +1877,211 @@ def test_update_package_json_files(
...
+        pytest.param("https://example.com/some/path", id="proxy_url_missing_slash_in_path"),
+        pytest.param("https://example.com/some/path/", id="proxy_url_having_slash_in_path"),
+    ],
+)
+def test_npm_settings_correctly_handles_trailing_slash_in_proxy_urls(
+    proxy_url: str,
+) -> None:
+    parsed_url = str(NpmSettings(proxy_url=proxy_url).proxy_url)
+    number_of_trailing_slashes = len(parsed_url) - len(parsed_url.rstrip("/"))
+    msg = f"Got unexpected numner of trailing slashes: {number_of_trailing_slashes} instead of 1"
```
eskultety: `s/numner/number` :)

hermeto/core/config.py
```diff
@@ -112,6 +113,62 @@ class RuntimeSettings(BaseModel, extra="forbid"):
...
+
+
+def _ensure_proxy_url_ends_with_a_slash(url: MaybeHttpUrl) -> MaybeHttpUrl:
+    # NOTE: HttpUrl _does not_ normalize URLs with path
+    # (see e.g. https://github.com/pydantic/pydantic/issues/7186
+    # for details on expected behavior), hence this validator.
+    if url is None:
+        return None
+    elif str(url)[-1] == "/":
+        return HttpUrl(url=url)
```
eskultety: `return url` should suffice in this case.

hermeto/core/config.py
```diff
@@ -112,6 +113,62 @@ class RuntimeSettings(BaseModel, extra="forbid"):
...
+    return HttpUrl(url=str(url) + "/")
+
+
+def _proxy_url_must_not_contain_credentials(url: MaybeHttpUrl) -> MaybeHttpUrl:
+    if url is None:
+        return None
+    if url.username or url.password:
+        raise InvalidInput(
```
eskultety: This used to be a `ValueError` in previous revision, right? `InvalidInput` might seem like it fits, but pydantic expects validators to return `ValueError` or `AssertionError` for error aggregation and detection purposes, so we should IMO be compliant with that expectation: https://docs.pydantic.dev/latest/concepts/validators/#raising-validation-errors

hermeto/core/package_managers/npm.py
```diff
@@ -264,25 +266,30 @@ def get_main_package(self) -> NpmComponentInfo:
             "dev": False,
             "bundled": False,
             "missing_hash_in_file": None,
+            "external_refs": None,
         }
 
     def get_sbom_components(self) -> list[NpmComponentInfo]:
         """Return a list of dicts with sbom component data."""
         packages = self._packages
+        eref_rest = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
```
eskultety: `s/eref_rest/proxy_common` maybe? `eref_rest` is very unintuitive.

hermeto/core/package_managers/npm.py
```diff
@@ -488,6 +496,31 @@ def _clone_repo_pack_archive(
     return download_path
 
 
+def _patch_url_to_point_to_a_proxy(url: NormalizedUrl) -> NormalizedUrl:
+    # Convert 'https://registry.npmjs.org/accepts/-/accepts-1.3.8.tgz'
+    # to '<proxyaddress>/accepts/-/accepts-1.3.8.tgz'.
+    # Proxy URL is expected to have a trailing /
```
eskultety: >     # Proxy URL is expected to have a trailing /

... expected based on the normalization, pydantic or something else?

hermeto/core/package_managers/npm.py
eskultety: `for v in files_to_download.values()` might be more suitable here.

tests/unit/package_managers/test_npm.py
```diff
@@ -1763,7 +1798,11 @@ def test_update_package_lock_with_local_paths(
     expected_lockfile_data: dict[str, Any],
 ) -> None:
     for url, download_path in download_paths.items():
-        download_paths.update({url: rooted_tmp_path.join_within_root(download_path)})
+        download_paths.update(
+            {
+                url: rooted_tmp_path.join_within_root(download_path),
+            }
+        )
```
eskultety: Is this cosmetic change needed?

tests/unit/package_managers/test_npm.py
```diff
@@ -1638,7 +1673,7 @@ def test_get_npm_dependencies(
     mock_async_download_files: mock.Mock,
     rooted_tmp_path: RootedPath,
     deps_to_download: dict[str, dict[str, str | None]],
-    expected_download_subpaths: dict[str, str],
```
eskultety, line 1641: Looks like an unrelated change.

tests/unit/package_managers/test_npm.py
```diff
@@ -1838,3 +1877,211 @@ def test_update_package_json_files(
     package_json_projectfiles = _update_package_json_files(workspaces, rooted_tmp_path)
     for projectfile in package_json_projectfiles:
         assert json.loads(projectfile.template) == expected_file_data
+
+
+@pytest.mark.parametrize(
+    "proxy_url",
+    [
+        pytest.param("https://foo:bar@example.com", id="full_credentials_are_present"),
+        pytest.param("https://:bar@example.com", id="password_is_present"),
```
eskultety: nitpick: `id="password_is_present` I think the opposite is what's more important in context of the test :), i.e. `login_missing` and `password_missing` respectively.

tests/unit/package_managers/test_npm.py
```diff
@@ -1838,3 +1877,211 @@ def test_update_package_json_files(
...
+    mock_clone_as_tarball: mock.Mock,
+    mock_from_sri: mock.Mock,
+    mock_must_match_any_checksum: mock.Mock,
+    mock_async_download_files: mock.Mock,
+    rooted_tmp_path: RootedPath,
+    deps_to_download: dict[str, dict[str, str | None]],
+) -> None:
+    mock_config = mock.Mock()
+    mock_config.npm.proxy_url = "https://fakeproxy.com"
+    # ruff would assume this is a hardcoded password otherwise
```
eskultety, line 1935: Lol. In any case, this should be handled globally in `pyproject.toml`:
```
[tool.ruff.lint.per-file-ignores]
...
```
`S` rules are `bandit` security rules, we disable a bunch, maybe we can disable ALL `bandit` rules for test purposes to make our lives easier because once a PR is opened a potentially leaked secret has been made public, it's too late for `bandit` to report anything :).

hermeto/core/package_managers/npm.py
```diff
@@ -488,6 +496,31 @@ def _clone_repo_pack_archive(
...
+    config = get_config()
+    url_path = urlparse(url).path[1:]  # Don't need the leading /
```
eskultety: Does this slash handling have anything to do with the `HttpUrl` handling you're introducing somewhere else in this commit?

hermeto/core/package_managers/general.py
a-ovchinnikov: I agree with this in general, however I don't think this belongs in this change -- it would require some overhaul of existing code plus some design work when now I am just extending a preexisting interface: the underlying function expects `BasicAuth` object and I am simply passing it through.

hermeto/core/package_managers/npm.py
a-ovchinnikov: Yes, this one is much better, thank you.

hermeto/core/package_managers/npm.py
a-ovchinnikov: By the way I decided to structure the code. I'll add a couple of words to the comment to clarify that it is an internal agreement.

hermeto/core/package_managers/npm.py
a-ovchinnikov: Good point, thank you.

hermeto/core/config.py
a-ovchinnikov: I was not sure if I liked the result or not, I guess I'll remove it after all.

hermeto/core/config.py
a-ovchinnikov: Sure:
```python
>>> HttpUrl("http://www.example.com")
HttpUrl('http://www.example.com/')
>>> HttpUrl("http://www.example.com/")
HttpUrl('http://www.example.com/')
>>> HttpUrl("http://www.example.com/foobar")
HttpUrl('http://www.example.com/foobar')
>>> HttpUrl("http://www.example.com/foobar/")
HttpUrl('http://www.example.com/foobar/')
```
My original expectation based on a quick search was that `HttpUrl` would just deal with the trailing symbol for me, it proved to be wrong, hence a warning comment.

hermeto/core/config.py
a-ovchinnikov: I would disagree on this one: making validator aware of extra fields is an overcomplication. As written now it has a very narrow area of concern: it checks whether two fields come together or not, if not it tells a user to go and check settings. I think this will result in cleaner environments by ensuring that dangling variables are removed. Making it aware of a third field and making validation conditional on that field would make the logic too complicated and also reinforce a bad precedent of lumping separate concerns together. A better way of handling this would be to add a dedicated validator to check that when login is set then url must also be set and to either issue a warning or refuse to preceed altogether. I am not sure such strict approach is necessary though.

hermeto/core/config.py
a-ovchinnikov: Do you mean like in bearer auth? If yes I would have postponed that.

tests/unit/package_managers/test_npm.py
a-ovchinnikov: Danm. Thanks!

tests/unit/package_managers/test_npm.py
a-ovchinnikov: Yes. Removed.

tests/unit/package_managers/test_npm.py
a-ovchinnikov: No, probably a leftover from some earlier iteration. Removed.

tests/unit/package_managers/test_npm.py
a-ovchinnikov: While I firmly believe that tests must be cut quite a lot of slack linter-wise (especially from mypy, I think insisting on making _tests_ type check is counterproductive) this one I would probably kept as is, just in case someone accidentally adds something at some point. ruff complained well before I pushed the change so as long as developers run their tests before pushing this will be caught and dealt with.

hermeto/core/package_managers/npm.py
a-ovchinnikov: Yes. It just has to be handled consistently and I personally find it easier to have reminders spread around the relevant parts of code for that.

hermeto/core/config.py
slimreaper35: You can add doctests now 🙂

hermeto/core/package_managers/general.py
eskultety: Fair enough, whether we do the prep/refactoring work now or later doesn't make much of a difference.

hermeto/core/config.py
eskultety: > A better way of handling this would be to add a dedicated validator to check that when login is set then url must also be set and to either issue a warning or refuse to preceed altogether

This is in full compliance with what I'm asking for, so go for it :) 

> I am not sure such strict approach is necessary though.

I'd put this into the defensive programming practices bucket and I'm quite convinced we an extra mile (sometimes even more) for other stuff, so from that POV I think the code base can handle and extra trivial validator on the model :) .

tests/unit/package_managers/test_npm.py
eskultety: Fair enough. I just hope there won't be many of these.

hermeto/core/package_managers/npm.py
```diff
@@ -264,25 +266,30 @@ def get_main_package(self) -> NpmComponentInfo:
...
             missing_hash_in_file = None
+            external_refs = None
             if package.resolved_url:  # dependency is not bundled
                 resolved_url = _normalize_resolved_url(package.resolved_url)
                 dep_type = _classify_resolved_url(resolved_url)
 
                 if not package.integrity:
                     if dep_type in ("registry", "https"):
                         missing_hash_in_file = Path(self._lockfile_path.subpath_from_root)
+                if dep_type == "registry" and (proxy_url := get_config().npm.proxy_url):
```
eskultety: micro optimization: while `get_config` returns a singleton the call is unnecessary here because it runs in context of every package, better move it up to where `proxy_common` is.

hermeto/core/package_managers/npm.py
```diff
@@ -503,9 +536,12 @@ def _get_npm_dependencies(
     """
     files_to_download: dict[str, dict[str, Any]] = {}
     download_paths = {}
+    config = get_config()
+
     for url, info in deps_to_download.items():
-        url = _normalize_resolved_url(url)
+        fetch_url = url = _normalize_resolved_url(url)
```
eskultety: nitpick: this chained assignment is easy to miss (I missed during the first pass), please split it to separate lines.

hermeto/core/package_managers/npm.py
```diff
@@ -488,6 +496,31 @@ def _clone_repo_pack_archive(
     return download_path
 
 
+def _patch_url_to_point_to_a_proxy(url: NormalizedUrl) -> NormalizedUrl:
+    # Convert 'https://registry.npmjs.org/accepts/-/accepts-1.3.8.tgz'
+    # to '<proxyaddress>/accepts/-/accepts-1.3.8.tgz'.
+    # A proxy URL is assumed to have a trailing / by the implementation.
+    config = get_config()
```
eskultety: This single-purpose helper has a single caller only, no need to access the `config` singleton instance just to query `proxy_url`, you can pass it as an argument from the caller.

hermeto/core/package_managers/npm.py
```diff
@@ -488,6 +496,31 @@ def _clone_repo_pack_archive(
...
+    url_path = urlparse(url).path[1:]  # Don't need the leading /
+    return NormalizedUrl(str(config.npm.proxy_url) + url_path)
+
+
+def _async_download_tar(files_to_download: dict) -> None:
```
eskultety: Don't we want to follow the origin's type definition here? i.e. `dict[str, dict[str, Any]]`

hermeto/core/config.py
```diff
@@ -112,6 +113,55 @@ class RuntimeSettings(BaseModel, extra="forbid"):
...
+
+ProxyUrl = Annotated[
+    HttpUrl | None,
+    AfterValidator(_ensure_proxy_url_ends_with_a_slash),
+    AfterValidator(_proxy_url_must_not_contain_credentials),
+]
+
+
+class ProxyMixin(BaseModel):
+    """Proxy URL and authorization mixin."""
```
eskultety: nitpick: I know it's obvious to us right now, but maybe it would be nice to remind us or new contributors that this is modeling artifactory proxies rather than traditional connect tunnel Squid-like proxies, i.e.
`Artifactory remote repository "proxy" URL and authorization mixin`, how about that? Maybe even (yet another) reminder in the commit message?

hermeto/core/package_managers/npm.py
```diff
@@ -517,6 +553,13 @@ def _get_npm_dependencies(
                     "/", "-"
                 )
                 download_paths[url] = download_dir.join_within_root(archive_name)
+                if config.npm.proxy_url:
+                    fetch_url = _patch_url_to_point_to_a_proxy(url)
+                    if config.npm.proxy_login and config.npm.proxy_password:
+                        proxy_auth = aiohttp.BasicAuth(
```
eskultety, line 558: pure naming nitpick: Did you know that `aiohttp.request` has a named argument `proxy_auth`? https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession.request
Was that a deliberate choice of names by any chance :) ? Because then on L513 you do  basically `auth = proxy_auth` which might not escape and maybe even be confusing to a careful reader.

hermeto/core/package_managers/npm.py
```diff
@@ -533,21 +577,22 @@ def _get_npm_dependencies(
...
-        async_download_files(
-            {url: item["download_path"] for (url, item) in files_to_download.items()},
-            get_config().runtime.concurrency_limit,
-        )
-    )
+    files_with_auth = {k: v for k, v in files_to_download.items() if v["proxy_auth"] is not None}
+    files_without_auth = {k: v for k, v in files_to_download.items() if v["proxy_auth"] is None}
+
+    _async_download_tar(files_with_auth)
+    _async_download_tar(files_without_auth)
```
eskultety: Previously we ran all downloads concurrently, well, at least within the same event loop. This now separates proxied downloads from non-proxied ones. Could we maybe wrap these with another `asyncio.gather` call? Might not require much code at all or at least I wouldn't expect it to.

hermeto/core/config.py
```diff
@@ -112,6 +113,59 @@ class RuntimeSettings(BaseModel, extra="forbid"):
...
+    if url is None:
+        return None
+    if url.username or url.password:
+        raise ValueError(
+            "User-provided proxy URL appears to contain embedded credentials. "
+            "Please ensure that proxy URL does not contain credentials "
+            "supplied as protocol://<login>:<password>@address. Please use "
+            "appropriate environment variables or configuration file entries for that. "
+            "Please refer to documentation for further information on how to set up "
+            "proxy access.",
```
eskultety, line 126: Just an observation, I would have most likely modeled this the same way - I wonder how messy this long message will end up being when it's actually reported by `pydantic` and its validation machinery which is very verbose and not useful at all. Like I said, not sure how much we can do about this, probably another level of try-except somewhere and use this user-friendly error message.

tests/unit/package_managers/test_npm.py
```diff
@@ -1838,3 +1872,211 @@ def test_update_package_json_files(
...
+
+@pytest.mark.parametrize(
+    "proxy_url",
+    [
+        pytest.param("https://foo:bar@example.com", id="full_credentials_are_present"),
+        pytest.param("https://:bar@example.com", id="login_is_missing"),
+        pytest.param("https://foo:@example.com", id="password_is_missing"),
+    ],
+)
+def test_npm_settings_rejects_proxy_urls_containing_credentials(
```
eskultety, line 1885: nitpick: Care enough to wrap these in a e.g. `TestNPMProxySettings` class to make the test names a bit more digestible? (non-actionable if you don't feel like it).

hermeto/core/config.py
eskultety: > Sure:
> 
> ```python
> >>> HttpUrl("http://www.example.com")
> HttpUrl('http://www.example.com/')
> >>> HttpUrl("http://www.example.com/")
> HttpUrl('http://www.example.com/')
> >>> HttpUrl("http://www.example.com/foobar")
> HttpUrl('http://www.example.com/foobar')
> >>> HttpUrl("http://www.example.com/foobar/")
> HttpUrl('http://www.example.com/foobar/')
> ```
> 
> My original expectation based on a quick search was that `HttpUrl` would just deal with the trailing symbol for me, it proved to be wrong, hence a warning comment.

Thank you, that matches my local invocations. The reason I asked in the first place is because the slash handling of the model is truly weird and maybe incorporating something https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.url_preserve_empty_path would make it more deterministic, i.e. no slash on the input -> no slash in the model; slash in the input -> slash in the model. However, regardless of this I'm not entirely convinced that modeling this constraint directly into the `ProxyUrl` model when we pretty much only need to deal with this in `_patch_url_to_point_to_a_proxy` when doing the concatenation is either necessary or desirable. Can't we deal with this situation in the concatenation logic only? Would that not make it more robust?

a-ovchinnikov: I am sorry, I don't get it. This "validator" morphs input data to always match expectations. It is trivial to drop it and shift the burden on a client function, it is trivial to keep it and to keep the client function simple. I prefer "validator" approach slightly more solely because it concentrates the logic in one place, I don't see how this could be a problem in the long run.

hermeto/core/config.py
a-ovchinnikov: Ok, fair enough, I'll add this constraint.

hermeto/core/package_managers/npm.py
a-ovchinnikov: No, I did not. From the cursory look it is hard to tell the difference between then two arguments though. Do you think this is worth renaming?

tests/unit/package_managers/test_npm.py
a-ovchinnikov: I'll skip this for now if you don't mind.

a-ovchinnikov:
Since last revision: reworked `_async_download_tar` to make it run in parallel for files requiring authorization and not requiring authorization, applied minor suggestions. Also I must note that as of today GitHub decided not to show some comments inline, @eskultety please poke me again if I missed something.

hermeto/core/package_managers/npm.py
eskultety: For one more experienced in `aiohttp` it might be, I found out by accident. Uhm, I won't dwell on it since the plan is to cull `auth=auth` (`BasicAuth`) in favour of header injection in a reasonable time frame anyway, I just wanted to mention this as it was an 'aha' moment for me when I found out.

hermeto/core/config.py
eskultety: > I am sorry, I don't get it. This "validator" morphs input data to always match expectations.

>  incorporating something https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.url_preserve_empty_path would make it more deterministic

Mentioning this in the first place only added some noise to my response, it was only something I hoped might remedy the problem, it doesn't really.

> 
I am sorry, I don't get it. This "validator" morphs input data to always match expectations. It is trivial to drop it and shift the burden on a client function, it is trivial to keep it and to keep the client function simple.

Yes, this is a trivial enough case that doesn't require any weird quirks to be done at the place of use (the helper). In contrast, this not only requires a validator that is pretty much hidden and easy to miss by devs but it also requires a reminding comment in the helper function, otherwise the logic with the slash you have there would be opaque.
While we sometimes do modify data in the validators, those are ugly hacks for cases that are non-trivial to solve otherwise, but we should resort to such practices in situations where we considered other ways of achieving the same and we didn't like it. If it can be trivially achieved, then I personally think it's a clear signal :) .

a-ovchinnikov: Ok, in the latest revision I dropped the validator and moved the logic to helper.

eskultety (approved):

PR #1249, opened 2026-01-26
Add experimental package manager support with annotations

sswguo:
Mark the components generated by experimental package managers
(those with `x-` prefix like "x-foo") using document-level
annotations for better processing.

Resolves: https://github.com/hermetoproject/hermeto/issues/688

Assisted-by: Claude Code

### DIFF
--- a/docs/design/package-manager-template.md
+++ b/docs/design/package-manager-template.md
@@ -182,6 +182,11 @@ This section helps the community evaluate the maturity of the package manager. E
 managers use the `x-` prefix in their package manager name. This section should be completed prior
 to the community declaring the package manager "fully supported."
 
+**Note**: Components produced by experimental package managers (those with `x-` prefix) are
+identified through document-level annotations in generated SBOMs. Each experimental package manager
+creates a separate annotation listing the components it produced, with text such as
+`hermeto:backend:experimental:x-foo` and subjects containing the relevant bom-ref values.
+
 ### Current Limitations
 
 Document known limitations of the current implementation:
--- a/hermeto/core/models/sbom.py
+++ b/hermeto/core/models/sbom.py
@@ -158,15 +158,21 @@ def create_backend_annotation(
 ) -> Annotation | None:
     """Create an annotation that tags components with the backend that fetched them.
 
+    Experimental backends (those with an x- prefix) use a distinct label in the annotation text.
+
     Returns None if there are no components to annotate.
     """
     if not components:
         return None
+    if backend_name.startswith("x-"):
+        text = f"{APP_NAME}:backend:experimental:{backend_name}"
+    else:
+        text = f"{APP_NAME}:backend:{backend_name}"
     return Annotation(
         subjects={c.bom_ref for c in components},
         annotator={"organization": {"name": "red hat"}},
         timestamp=spdx_now(),
-        text=f"{APP_NAME}:backend:{backend_name}",
+        text=text,
     )
 
 
--- a/tests/unit/models/test_sbom.py
+++ b/tests/unit/models/test_sbom.py
@@ -24,6 +24,7 @@ from hermeto.core.models.sbom import (
     SPDXRelation,
     SPDXSbom,
     Tool,
+    create_backend_annotation,
     merge_component_properties,
 )
 
@@ -572,6 +573,18 @@ class TestSbom:
         # bar should NOT have the annotation (its bom-ref is NOT in subjects)
         bar_annotation_comments = [a.comment for a in bar_pkg.annotations]
         assert "hermeto:backend:npm" not in bar_annotation_comments
+
+    def test_create_backend_annotation_experimental(self, mock_spdx_now: str) -> None:
+        """Experimental backends (x- prefix) should produce a distinct annotation text."""
+        components = [Component(name="test", purl="pkg:foo/test@1.0.0", version="1.0.0")]
+        annotation = create_backend_annotation(components, "x-foo")
+        sbom = Sbom(annotations=[annotation], components=components)
+
+        spdx_sbom = sbom.to_spdx("NOASSERTION")
+        pkg = next(p for p in spdx_sbom.packages if p.name == "test")
+        annotation_comments = [a.comment for a in pkg.annotations]
+
+        assert f"{APP_NAME}:backend:experimental:x-foo" in annotation_comments
 
     @pytest.mark.parametrize(
         "sbom, expected_package_source_info",


hermeto/core/models/sbom.py
```diff
@@ -73,6 +74,25 @@ class Pedigree(pydantic.BaseModel):
...
+
+    exp_package_types = {
+        package_type.removeprefix("x-")
+        for package_type in package_types
+        if package_type.startswith("x-")
+    }
+    return exp_package_types
+
+
+experimental_package_types = filter_exp_package_types()
```
slimreaper35: Why is this at global namespace ?

tests/unit/models/test_sbom.py
```diff
@@ -82,6 +84,53 @@ def test_default_property(
...
+        if expected_property is not None:
+            assert expected_property in component.properties
+        else:
+            assert FOUND_BY_EXPERIMENTAL_PROPERTY not in component.properties
+
+        assert FOUND_BY_APP_PROPERTY in component.properties
+
+    def test_experimental_package_manager_to_component_property(self) -> None:
+        with mock.patch("hermeto.core.models.sbom.get_args") as mock_get_args:
+            mock_get_args.return_value = ("bundler", "npm", "x-foo", "cargo")
```
slimreaper35: nitpick: mock.patch as a decorator is a bit more readable IMO (less indentation)

tests/unit/models/test_sbom.py
```diff
@@ -82,6 +84,53 @@ def test_default_property(
...
+                    purl="pkg:foo/bar@1.0.0",
+                )
+
+                assert FOUND_BY_APP_PROPERTY in exp_component.properties
+                assert FOUND_BY_EXPERIMENTAL_PROPERTY in exp_component.properties
+
+                component = Component(name="package", purl="pkg:npm/package@1.0.0")
+
+                assert FOUND_BY_APP_PROPERTY in component.properties
+                assert FOUND_BY_EXPERIMENTAL_PROPERTY not in component.properties
```
slimreaper35: This seems unrelated to the test case.

tests/unit/models/test_sbom.py
```diff
@@ -82,6 +84,53 @@ def test_default_property(
...
+        self, purl: str, expected_property: Property | None
+    ) -> None:
+        component = Component(name="test-package", purl=purl)
+
+        if expected_property is not None:
+            assert expected_property in component.properties
+        else:
+            assert FOUND_BY_EXPERIMENTAL_PROPERTY not in component.properties
+
+        assert FOUND_BY_APP_PROPERTY in component.properties
```
a-ovchinnikov: Nitpick: this assertion is arbitrary in the sense that it checks a different fact. I suggest removing it from this test.

tests/unit/models/test_sbom.py
```diff
@@ -82,6 +84,53 @@ def test_default_property(
...
+            ("pkg:foo/C@1.0.0", FOUND_BY_EXPERIMENTAL_PROPERTY),
+            ("pkg:cargo/D@1.0.0", None),
+        ],
+    )
+    def test_component_with_experimental_property(
+        self, purl: str, expected_property: Property | None
+    ) -> None:
+        component = Component(name="test-package", purl=purl)
+
+        if expected_property is not None:
```
a-ovchinnikov: Having branches in tests is an antipattern which unfortunately took root in the codebase a very long time ago and no one has so far managed to focus on weeding it out. I would ask to split the test in two separate tests with clear names:
```
test_component_produced_by_an_experimental_pm_has_experimental_property
```
and
```
test_component_produced_by_fully_supported_pm_does_not_have_experimental_property
```
When named this way the tests would make it immediately clear what assumption did not hold by just presenting their names, with a name as it is now one would need to delve into implementation to understand that. (For even more clarity I would suggest adding an assertion message, something like "Did not find experimental property in a component produced by an experimental PM" and "Found experimental property where it should not be", but won't insist on it since the names are already telling enough.)

tests/unit/models/test_sbom.py
```diff
@@ -82,6 +84,53 @@ def test_default_property(
...
+            assert FOUND_BY_EXPERIMENTAL_PROPERTY not in component.properties
+
+        assert FOUND_BY_APP_PROPERTY in component.properties
+
+    def test_experimental_package_manager_to_component_property(self) -> None:
+        with mock.patch("hermeto.core.models.sbom.get_args") as mock_get_args:
+            mock_get_args.return_value = ("bundler", "npm", "x-foo", "cargo")
+
+            exp_package_types = filter_exp_package_types()
+            assert exp_package_types == {"foo"}
```
a-ovchinnikov: This test case looks like two tests within one, one checking that `filter_exp_package_types()` function works as expected, another one checking that for experimental PM-found component there is a `FOUND_BY_EXPERIMENTAL_PROPERTY` in properties, and that for regular PM-found component there is no such property. The second part looks very similar to what is being tested in `test_component_with_experimental_property()`. I would suggest reworking this test to split it into independent parts and making sure both follow Arrange-Act-Assert pattern.

hermeto/core/models/property_semantics.py
```diff
@@ -18,6 +18,7 @@ class PropertyEnum(str, Enum):
     PROP_CDX_NPM_PACKAGE_BUNDLED = "cdx:npm:package:bundled"
     PROP_CDX_NPM_PACKAGE_DEVELOPMENT = "cdx:npm:package:development"
     PROP_FOUND_BY = f"{APP_NAME}:found_by"
+    PROP_FOUND_BY_EXPERIMENTAL = f"{APP_NAME}:found_by:experimental"
```
a-ovchinnikov: Let's replace `experimental` with `experimental_pm` or even with `experimental_package_manager`, it might confuse a user otherwise if they are not too familiar with Hermeto internal terminology. @eskultety , @taylormadore would you have any other naming suggestions?

hermeto/core/models/property_semantics.py
```diff
@@ -41,6 +42,7 @@ class PropertySet:
 
     bundler_package_binary: bool = False
     found_by: str | None = None
+    found_by_experimental: bool = False
```
a-ovchinnikov: I wonder if it would make sense to have this represented with a string containing the name of experimental PM in question? I.e. to change type to `str | None` and store package manager name without the leading `x-` in it.

hermeto/core/models/sbom.py
```diff
@@ -73,6 +74,25 @@ class Pedigree(pydantic.BaseModel):
 FOUND_BY_APP_PROPERTY: Property = Property(name=PropertyEnum.PROP_FOUND_BY, value=f"{APP_NAME}")
 
 
+def filter_exp_package_types() -> set[str]:
+    """Extract experimental PURL types from PackageManagerType."""
+    package_types = get_args(PackageManagerType)
+
+    exp_package_types = {
```
a-ovchinnikov: A very minor style nitpick: comprehensions are meant to be written on a single line, it is better for readability to use a regular loop otherwise. To get there I would normally use short and very short names for loop variables, I find this acceptable because of extremely narrow scope of such variables (just one line, I believe everyone can keep track of a symbol within a line). I would have compacted this expression to:
```python
exp_package_types = {pt.removeprefix("x-") for pt in package_types if pt.startswith("x-")}
```
or just returned the set comprehension without even naming it. Note, however, that the linter will likely get upset by a too short variable name. In such cases I tell it to ignore types for this particular line and also add a short comment on a line above explaining why I decided to block the check.

tests/unit/models/test_property_semantics.py
```diff
@@ -257,6 +296,16 @@ def test_conversion_from_and_to_properties(
                 PropertySet(pip_package_binary=True),
                 PropertySet(pip_package_binary=True),
             ),
+            (
+                PropertySet(found_by_experimental=False),
```
a-ovchinnikov: Why did you choose this particular half of a merge truth table for this test?

hermeto/core/models/sbom.py
```diff
@@ -73,6 +74,25 @@ class Pedigree(pydantic.BaseModel):
...
+        package_type.removeprefix("x-")
+        for package_type in package_types
+        if package_type.startswith("x-")
+    }
+    return exp_package_types
+
+
+experimental_package_types = filter_exp_package_types()
+
+FOUND_BY_EXPERIMENTAL_PROPERTY: Property = Property(
```
a-ovchinnikov: Two observations here:
1. This is a global object set to `true`;
2. There is no global object set to `false`.

This makes me more convinced that the property should not be boolean, but a string. There is probably no actual need for it to be global too unless I am missing something here.

hermeto/core/models/sbom.py
sswguo: I was thinking to avoid calling `filter_exp_package_types` repeatedly every time a Component is validated. So if there’s a better approach for this?

hermeto/core/models/property_semantics.py
sswguo: Yeah, that makes more sense, but does this overlap with the `purl.type` ?

a-ovchinnikov: A boolean value overlaps too -- it is derived from a purl after all. The thing to note here is that we, as Hermeto developers, always know which PM is experimental, our user who use Hermeto for prefetching might not, or not care enough to dive into docs. Here we will provide a clearly named field at a price of some duplication to make it easier for user to understand what their SBOM actually means.

hermeto/core/models/sbom.py
a-ovchinnikov: It is not that expensive to compute multiple times, but you could use `functools.lru_cache()` decorator for that too.

hermeto/core/models/property_semantics.py
sswguo: Sounds great.

hermeto/core/models/sbom.py
```diff
@@ -73,6 +74,13 @@ class Pedigree(pydantic.BaseModel):
 FOUND_BY_APP_PROPERTY: Property = Property(name=PropertyEnum.PROP_FOUND_BY, value=f"{APP_NAME}")
 
 
+def filter_exp_package_types() -> set[str]:
```
a-ovchinnikov: Non-blocking:
```diff
- def filter_exp_package_types() -> set[str]:
+ @functools.lru_cache
+ def filter_exp_package_types() -> set[str]:
```
That's how you can short-cut the function without needing an explicit variable.

hermeto/core/models/sbom.py
```diff
@@ -119,6 +127,23 @@ def _add_found_by_property(cls, properties: list[Property]) -> list[Property]:
 
         return properties
 
+    @pydantic.model_validator(mode="after")
+    def _add_experimental_property(self) -> Self:
+        try:
```
slimreaper35: I am not sure if logging the exception for every single component is the best solution. If a project has 500 components, and something fails, I can imagine how the logs would look...

hermeto/core/models/sbom.py
```diff
@@ -119,6 +127,23 @@ def _add_found_by_property(cls, properties: list[Property]) -> list[Property]:
...
+        except ValueError as ex:
+            log.warning("Failed to parse PURL: %s, exception: %s", self.purl, str(ex))
+            return self
+
+        if purl.type in filter_exp_package_types():
+            experimental_property = Property(
+                name=PropertyEnum.PROP_FOUND_BY_EXPERIMENTAL, value=purl.type
+            )
+            if experimental_property not in self.properties:
+                self.properties.append(experimental_property)
```
slimreaper35: nitpick: I think properties are deduplicated; so this `if` check might not be needed.

hermeto/core/models/sbom.py
sswguo: Then I prefer to update the level to `debug`, but sure let's see if it's worth to adding the following thread-safe way for this. ; -) 

   
```
_purl_error_lock = threading.Lock()
_logged_purl_errors: set[str] = set()

@pydantic.model_validator(mode="after")
def _add_experimental_property(self) -> Self:
    try:
        purl = PackageURL.from_string(self.purl)
    except ValueError as ex:
        error_key = str(type(ex).__name__ + str(ex))

        with self.__class__._purl_error_lock:
            if error_key not in self.__class__._logged_purl_errors:
                log.warning("Failed to parse PURL (will suppress similar): %s, exception: %s", self.purl, str(ex))
                self.__class__._logged_purl_errors.add(error_key)
        return self
```

hermeto/core/models/property_semantics.py
taylormadore: Apologies, I missed this ping. I don't want to derail this work, but your question made me think: Do we envision any cases where the _package manager_ is not necessarily experimental, but some feature of it is? Should we generalize this to some kind of experimental feature flag rather than associating it specifically with package managers? Just a thought

sswguo:
> @sswguo I'm terribly sorry coming late to the party, especially since you have one ACK already. That said, I have one substantial architectural concern (without going deep into the PR technicalities). 

That's totally fine, I understand that you have a lot on your plate. ; -) 

> This PR has been built around the `hermeto:found_by` custom property precedent. However, custom properties should be something that is inherently bound to each component, IOW you can imagine this problem domain as: _processing context separated from component identity_ VS _processing context mixed into component data_ - custom properties represent the latter. IOW at least in CycloneDX component properties are attributes of the component itself, not a limitation (or behaviour) of its generator. Another thing to consider is the way this is reported - with custom properties the SBOM processor has to iterate over all components, filtering out those that have been generated by an experimental backend to make an informed decision.
>
> I believe we have recently set a different (and IMO correct) precedent in #1201 where we modeled the permissive mode in SBOM using SPDX annotations and CyclonDX's `bom-ref` object references. I think that should become our go-to approach going forward because it allows us to attach contextual metadata about components (yes, in exchange for a slightly more verbose SBOM) rather than polluting the component's own property namespace. Unlike custom properties, annotations limit the subset of components to iterate over to the number of items declared in the annotations field at the top of the SBOM.
> 
> With all the above said one could IMHO say that the permissive flag is fundamentally and effectively in the same bucket.

And thanks for the detailed information, I will go through #1201 as well as learning the formats of both CyclonDX and SPDX first.

docs/design/package-manager-template.md
```diff
@@ -182,6 +182,12 @@ This section helps the community evaluate the maturity of the package manager. E
 managers use the `x-` prefix in their package manager name. This section should be completed prior
 to the community declaring the package manager "fully supported."
 
+**Note**: Components produced by experimental package managers (those with `x-` prefix) are
+identified through document-level annotations in generated SBOMs. Each experimental package manager
+creates a separate annotation listing the components it produced, with text such as
+`hermeto:found_by:experimental_package_manager:foo` and subjects containing the relevant bom-ref
```
eskultety: Based on commit 0004d6539 we'll want to be consistent with the naming here for the experimental backend as well, e.g. `hermeto:backend:experimental:x-foo` might be an option (@taylormadore @a-ovchinnikov we should have a unanimous consensus on the name here since this is public facing)

hermeto/core/models/sbom.py
```diff
@@ -86,6 +88,14 @@ class Pedigree(pydantic.BaseModel):
 FOUND_BY_APP_PROPERTY: Property = Property(name=PropertyEnum.PROP_FOUND_BY, value=f"{APP_NAME}")
 
 
+@functools.lru_cache
+def filter_exp_package_types() -> set[str]:
+    """Extract experimental PURL types from PackageManagerType."""
+    package_types = get_args(PackageManagerType)
+
+    return {pt.removeprefix("x-") for pt in package_types if pt.startswith("x-")}
```
eskultety: nitpick: `removeprefix("x-")` we probably want to remain true to how the backends are referenced on the command line, IOW finding a backend to be reported as experimental only to find out that it's called differently in CLI (needs the `x-` prefix) may be a bad UX.

eskultety (changes requested):
@sswguo Again, sorry for a late reply. So, the work looks very compact now which is awesome and we're converging. You'll need to rebase on top of fresh main. As for the approach I'd suggest having a look at https://github.com/hermetoproject/hermeto/commit/0004d65396519798d288960fbc31e156e4e088e5 which set a good precedent going forward. My first (naive) suggestion from the top of my head would be to just enhance the annotations creation helper to be aware of experimental backends and change the annotation text. We might need a new helper, you'll see how it goes :).

hermeto/core/models/property_semantics.py
eskultety: > Apologies, I missed this ping. I don't want to derail this work, but your question made me think: Do we envision any cases where the _package manager_ is not necessarily experimental, but some feature of it is? Should we generalize this to some kind of experimental feature flag rather than associating it specifically with package managers? Just a thought

Hmm, I clicked submit on my review too soon forgetting about this thread. Yeah, that's a very good point. @taylormadore do you think we could stuff both partial and full functionality under such a flag maintaining clarity on what the true root cause is or would it be better to annotate separately for a backend and for a feature? If it's the former, then we need to rethink this from the grounds up as I do acknowledge this concern. If it's the latter then I think we can leave it for a follow-up, potentially defer to Hermeto 1.0. My personal take on this is that we're kind in need for this SBOM tagging so I'd probably lean towards the latter.

taylormadore: > would it be better to annotate separately for a backend and for a feature?

I'm good with this :+1:

sswguo:
> My first (naive) suggestion from the top of my head would be to just enhance the annotations creation helper to be aware of experimental backends and change the annotation text. We might need a new helper, you'll see how it goes :).

Enhancing the annotations creation helper works, it looks more concise, and avoids iterating over the package managers. ;)

hermeto/core/models/sbom.py
sswguo: +1,  let's keep `x-` here.

docs/design/package-manager-template.md
a-ovchinnikov: `x-` repeats `experimental`, the former is a shorter version of the latter (I won't block on this, as long as the messaging is clear redundancy is not a problem, here I think it is clear in both cases).

Issue #1205, opened 2025-12-08
Prefetch sdists alongside wheels


taylormadore:
When prefetching wheels for multiple target platforms, Hermeto doesn't provide a fallback for platforms that lack compatible wheels on PyPI. The fallback-to-sdist logic in "prefer-binary" mode only triggers when _no_ wheels match the filter at all. If wheels exist for some platforms but not others, Hermeto fetches only the matching wheels and leaves platforms without wheel coverage unable to build.

Example: A user configures binary filters for a multi-arch build:
```
{
  "binary": {
    "arch": "x86_64,ppc64le",
    "os": "linux",
  }
}
```
PyPI has wheels for x86_64 but not ppc64le. Hermeto fetches only the x86_64 wheel and no sdist. The ppc64le build fails because no compatible distribution was fetched.

We should consider either always fetching an sdist when in binary mode or providing a configuration setting to enable doing so.

ckhordiasma:
was able to sort of work around this issue by creating a dummy folder for each architecture that was inconsistent:

```
      [
      {
        "type": "pip", 
        "path": ".", 
        "requirements_files": ["requirements-x86_64.txt"],
        "requirements_build_files": ["requirements-build.txt"],
        "binary": { 
          "arch": "x86_64,aarch64"
        }
      },
      {
        "type": "pip", 
        "path": ".konflux/s390x", 
        "requirements_files": ["requirements.txt"],
        "requirements_build_files": ["requirements-build.txt"],
        "binary": { 
          "arch": "s390x"
        }
      },
      {
        "type": "pip", 
        "path": ".konflux/ppc64le", 
        "requirements_files": ["requirements.txt"],
        "requirements_build_files": ["requirements-build.txt"],
        "binary": { 
          "arch": "ppc64le"
        }
      },
      { "type":"cargo" },
      { "type":"rpm" }
      ]
```

eskultety:
Re-posting my comment (rephrased a tiny bit) from https://github.com/hermetoproject/hermeto/pull/1252 where I got asked to do so:

[...]

PR #1252, opened 2026-01-26
fix(pip): always include sdist in prefer-binary mode

ben-alkov:
What: "prefer-binary" mode now returns the best available sdist along with matching wheels.

Why: Currently, when users specify multiple architectures and the index does not have wheels for all arches, "prefer-binary" mode returns *only* matching wheels (no sdists). Some arches can fail to build under these circumstances.

This commit ensures that arches without wheels can still build from source.

Closes: #1205


ben-alkov:
Two integration tests are failing because they check the *exact* list(s) of packages produced. I'll adjust those tomorrow.

```
tests/integration/test_pip.py::test_pip_packages[pip_custom_index]
tests/integration/test_pip.py::test_e2e_pip[pip_e2e_wheels]
```

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -71,6 +72,110 @@ def test_sdist_sorting() -> None:
...
+
+        result = _process_prefer_binary_mode(
+            sdists=[],
+            wheels=[wheel],
+            name="pkg",
+            version="1.0.0",
+        )
+
+        assert result == [wheel]
+        assert "No sdist available for pkg==1.0.0, returning only wheels" in caplog.text
```
a-ovchinnikov: This looks like a string test, can it be avoided?

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -71,6 +72,110 @@ def test_sdist_sorting() -> None:
...
+        )
+
+        result = _process_prefer_binary_mode(
+            sdists=[sdist_zip, sdist_tar],
+            wheels=[wheel],
+            name="pkg",
+            version="1.0.0",
+        )
+
+        assert result == [wheel, sdist_tar]
```
slimreaper35: We have `test_sdist_sorting` which covers the ordering of all possible sdists.

eskultety:
> We should also update docs to clarify that we _always_ try to download an sdist in the prefer-binary mode.

Glad that ^this was mentioned, because I want to comment on that. I think that the impression or message https://github.com/hermetoproject/hermeto/issues/1205 gives out might be a little unfortunate - it offers a solution to fix a user problem - _fetch an sdist under all circumstances_ - but I believe it was meant as an immediate fix to unblock user builds rather than a long term correct solution.

TL;DR:
--------
Let's proceed with this fix once review comments are addressed, BUT make sure we document this properly for the time being AND that we work out a plan to make the logic conform to the original design and only download sdists for when a wheel isn't available for a particular filter combo again.

Long version
------------------
^This immediate fix returns us back to times prior the binary filtering effort to a point where I'm sort of questioning it's overall usefulness and the time we have spent on this effort once we simply go back the original "cachi2" direction. Admittedly, the wheel availability deviation based on architecture is something I personally didn't consider at all back in the day. That said, the network bandwidth & storage waste that were one of the drivers for the motivation is still very much real, although not really visible in context of a single project. In reality, we're talking hundreds and thousands of builds a day in a scaling pipeline that will download both wheels AND sdists where we're most likely dealing with >90% (I guess even more) wheel coverage match based on the binary filters used.
Storage is cheap, but the network bandwidth is a real issue on that scale, adding to the likelihood of either timed out OR simply failed builds due to many concurrent server connections. Not only that, but we already know for a fact that every file we fetch resides in the kernel page cache, increasing the memory resource requirements of Hermeto which is not limited to a particular backend of ours.
We should therefore think about the coverage matrix based on the filters we expose and implement filter evaluation logic that truly only recommends an sdist for download when a particular binary filter combo doesn't yield a wheel.

tests/unit/package_managers/pip/test_package_distributions.py
ben-alkov: Good point, it's not even needed due to the `assert` just above.

ben-alkov:
@eskultety; Please remove the Long Version rant and paste it into the original issue (#1205).

My understanding of this issue is that it is a Critical hotfix for a customer issue, and that this approach was already decided on.

If you wish to re-litigate it, please do so elsewhere. Doing it here simply adds noise and confusion to what should be a *very* straightforward review process.

tests/unit/package_managers/pip/test_package_distributions.py
ben-alkov: Deleted

ben-alkov:
> We should also update docs to clarify that we _always_ try to download an sdist in the prefer-binary mode.

Added

tests/unit/package_managers/pip/test_package_distributions.py
ben-alkov: Fixed

ben-alkov:
IT data regenerated.

eskultety:
> My understanding of this issue is that it is a Critical hotfix for a customer issue

Well, a workaround exists, if it were critical, it would have been addressed before the EoY.

>, and that this approach was already decided on.

It's extremely hard to predict all aspects of a fix in an issue beforehand as the code can be fairly complex, it's all approximation until the implementation materializes, that's why I often comment on implementation PRs when I finally see all the pieces together, so I would not say "it was already decided on", that's IMO a strong statement, nothing is set in stone as far as I'm concerned.

> If you wish to re-litigate it, please do so elsewhere. Doing it here simply adds noise and confusion to what should be a _very_ straightforward review process.

Well, I disagree, I don't feel there's any confusion caused. If I only commented on the issue (i could), your PR would just close it --> case closed. I tend to believe that there's always a place for a discussion, however brief. As long as we're in an agreement that we need a follow-up work rather than settle for a hotfix permanently (which is exactly how this PR is selling it - the right solution for the problem), I'm okay.

> @eskultety; Please remove the Long Version rant and paste it into the original issue (#1205).

Firstly, I don't think it was a "rant" (you'd know, trust me), it was rather a concerned observation. Secondly, if I stop posting comments inviting for discussion about stability, security, transparency, scalability, etc. for the benefit of the code base I might as well just stop doing reviews altogether ;). I will copy-paste the long version observation to the respective issue, but I will not delete my original comment.

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -272,7 +272,14 @@ def _process_prefer_binary_mode(
     name: str,
     version: str,
 ) -> list[DistributionPackageInfo]:
-    return wheels if wheels else _process_no_binary_mode(sdists, name, version)
+    if not wheels:
```
eskultety, line 284: Please add a big `NOTE` explaining commentary that this is not a permanent solution rather a hotfix, because it's deviating from the design doc, effectively invalidating it. While at it, paste the same reasoning to the commit message (it's a hotfix) as well.

ben-alkov: Added

PR #1263, opened 2026-01-30
feat(cargo): Add support for alternative Cargo registries

### LINKED ISSUE #1001: Support alternative registries for cargo
To use a registry other than [crates.io](https://crates.io/), the name and index URL of the registry must be added to a [.cargo/config.toml file](https://doc.rust-lang.org/cargo/reference/config.html).

See https://doc.rust-lang.org/cargo/reference/registries.html#using-an-alternate-registry for more information.

Since we hide the config file in order to deal with potentially insecure settings during the prefetch, we would only fetch packages from the official registry and report them as such--even though the user may have specified/expect them to be fetched from an alternate registry.

A/C
- Alternative registries are supported.
- The prefetching still works without insecure settings.
- The alternative registry is reported in the package URLs (PURLs) as `repository_url` qualifier.

shokakucarrier:
Change _hidden_cargo_config_file to sanitize: now .cargo/config can keep only safe registry auth fields (index, token, credential-provider)
Add repository_url qualifier to PURLs for packages from alternate registries
Filter out crates.io packages from getting repository_url qualifier
Testing: Add comprehensive unit tests for config sanitization and PURL generation
    
Resolves #1001

### DIFF
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -10,9 +10,10 @@ from typing import Any
 from urllib.parse import urlparse, urlunparse
 
 import tomlkit
+import tomlkit.exceptions
 from packageurl import PackageURL
 
-from hermeto.core.errors import LockfileNotFound, NotAGitRepo, PackageRejected
+from hermeto.core.errors import LockfileNotFound, NotAGitRepo, PackageRejected, UnexpectedFormat
 from hermeto.core.models.input import Mode, Request
 from hermeto.core.models.output import Component, EnvironmentVariable, ProjectFile, RequestOutput
 from hermeto.core.rooted_path import RootedPath
@@ -55,11 +56,19 @@ class CargoPackage:
         if self.checksum is not None:
             qualifiers["checksum"] = self.checksum
 
-        if self.source is not None and self.source.startswith("git+"):
-            parsed_url = urlparse(self.source)
-            commit_id = parsed_url.fragment
-            base_url = urlunparse(parsed_url._replace(query="", fragment=""))
-            qualifiers["vcs_url"] = f"{base_url}@{commit_id}"
+        if self.source is not None:
+            if self.source.startswith("git+"):
+                parsed_url = urlparse(self.source)
+                commit_id = parsed_url.fragment
+                base_url = urlunparse(parsed_url._replace(query="", fragment=""))
+                qualifiers["vcs_url"] = f"{base_url}@{commit_id}"
+            elif self.source.startswith("registry+"):
+                # Extract registry URL from source (format: "registry+https://...")
+                registry_url = self.source.removeprefix("registry+")
+                if "crates.io" not in registry_url:
+                    qualifiers["repository_url"] = registry_url
+            else:
+                raise UnexpectedFormat(f"Unable to construct package URL from '{self.source}'.")
 
         return PackageURL(type="cargo", name=self.name, version=self.version, qualifiers=qualifiers)
 
@@ -130,7 +139,7 @@ def _fetch_dependencies(package_dir: RootedPath, request: Request) -> dict[str, 
     #                    cargo operations however is crucial when it is invoked from pip.
     cmd = ["cargo", "vendor", "--locked", "--versioned-dirs", "--no-delete", str(vendor_dir)]
     log.info("Fetching cargo dependencies at %s", package_dir)
-    with _hidden_cargo_config_file(package_dir):
+    with _sanitized_cargo_config_file(package_dir):
         # Prevent Cargo from invoking rustc
         env = {"CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS": "allow"}
         # The necessary configuration to use the vendored sources will be printed to STDOUT.
@@ -183,11 +192,12 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
 
 
 @contextmanager
-def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None, None]:
-    """Hide the cargo config file if it exists.
+def _sanitized_cargo_config_file(package_dir: RootedPath) -> Generator[None, None, None]:
+    """Replace Cargo config file to keep only alternate registry settings.
 
-    The file may contain various settings that could result in potential attack vectors.
-    Therefore, it is better to "hide" it before running the `cargo vendor` command.
+    The context manager swaps the original config file with one containing only the settings necessary
+    to use alternate registries during prefetch session, or hide it completely if it has no registries,
+    then restores original config file after prefetch complete.
     """
     # There is a slim chance to find an old project with .cargo/config
     # instead of .cargo/config.toml. If found it has to be hidden too since it still
@@ -198,20 +208,74 @@ def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None, 
     # of Cargo. Unlinking a symlink first is safe.
     all_possible_config_names = (".cargo/config", ".cargo/config.toml")
     configs_contents = []
+    processed_paths = set()
 
     for cfgname in all_possible_config_names:
         config = package_dir.join_within_root(cfgname)
-        data = config.path.read_text() if config.path.exists() else None
-        configs_contents.append((config, data))
-        if data is not None:
-            config.path.unlink()
+        if config.path.exists():
+            data = config.path.read_text()
+            sanitized = _sanitize_cargo_config(data)
 
+            if sanitized:
+                absolute_path = (
+                    config.path.readlink().absolute()
+                    if config.path.is_symlink()
+                    else config.path.absolute()
+                )
+                if absolute_path in processed_paths:
+                    continue
+                processed_paths.add(absolute_path)
+                configs_contents.append((config, data))
+                config.path.write_text(sanitized)
+            else:
+                configs_contents.append((config, data))
+                config.path.unlink()
     try:
         yield
     finally:
         for config, data in configs_contents:
             if data is not None:
                 config.path.write_text(data)
+
+
+def _sanitize_cargo_config(config_content: str) -> str:
+    """Extract only the [registries] section from Cargo config, keeping only safe fields.
+
+    Preserves only: index, token, credential-provider fields for each registry.
+    Returns sanitized TOML with only registries and their safe fields, or empty string if none exist.
+    """
+    if not config_content.strip():
+        return ""
+
+    try:
+        parsed = tomlkit.parse(config_content)
+        registries = parsed.get("registries")
+    except (tomlkit.exceptions.TOMLKitError, AttributeError):
+        raise UnexpectedFormat("Cargo config file contains invalid data and cannot be parsed")
+
+    allowed_fields = {"index", "token", "credential-provider"}
+    filtered_registries = tomlkit.table()
+    sanitized = tomlkit.document()
+
+    if registries is None:
+        return ""
+
+    for registry_name, registry_config in registries.items():
+        if not isinstance(registry_config, dict):
+            continue
+        filtered_fields = {
+            field: registry_config[field].strip()
+            for field in registry_config
+            if field in allowed_fields
+        }
+        if filtered_fields:
+            filtered_registries[registry_name] = filtered_fields
+
+    if filtered_registries:
+        sanitized["registries"] = filtered_registries
+        return tomlkit.dumps(sanitized)
+
+    return ""
 
 
 @contextmanager
--- a/tests/unit/package_managers/cargo/test_main.py
+++ b/tests/unit/package_managers/cargo/test_main.py
@@ -1,8 +1,14 @@
+import textwrap
 from typing import Any
 
 import pytest
 
-from hermeto.core.package_managers.cargo.main import CargoPackage, _resolve_main_package
+from hermeto.core.errors import UnexpectedFormat
+from hermeto.core.package_managers.cargo.main import (
+    CargoPackage,
+    _resolve_main_package,
+    _sanitize_cargo_config,
+)
 from hermeto.core.rooted_path import RootedPath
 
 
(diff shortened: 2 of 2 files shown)


hermeto/core/package_managers/cargo/main.py
```diff
@@ -183,32 +197,77 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
...
     all_possible_config_names = (".cargo/config", ".cargo/config.toml")
     configs_contents = []
+    registries: dict[str, str] = {}
 
     for cfgname in all_possible_config_names:
         config = package_dir.join_within_root(cfgname)
         data = config.path.read_text() if config.path.exists() else None
         configs_contents.append((config, data))
+
+        # Extract registry information from the config
```
a-ovchinnikov: Comment on the line 227 describes what code on L228-L241 does and that is good. A slightly better approach is to replace a comment with a function call, like this:
```python
registries |= _extract_registry_information_from_config(data)
```
and to just copy L228-L241 to the helper. Then whoever reads the code would know exactly what this line does and at the same time won't need to digest how exactly it does what it does. Could you please apply this suggestion? (It holds for any bit of code which has a preceding comment describing what the next bit is doing versus why the code is there).

hermeto/core/package_managers/cargo/main.py
```diff
@@ -183,32 +197,77 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
...
+        if data is not None and not registries:
+            try:
+                config_data = tomlkit.parse(data).value
+                # Extract registries section
+                # https://doc.rust-lang.org/cargo/reference/registries.html#using-an-alternate-registry
+                if "registries" in config_data:
+                    for registry_name, registry_config in config_data["registries"].items():
+                        if isinstance(registry_config, dict) and "index" in registry_config:
```
a-ovchinnikov: Optional, non-blocking suggestion:
```diff
- if isinstance(registry_config, dict) and "index" in registry_config:
-    registries[registry_name] = registry_config["index"]
+ match registry_config:
+     case {"index": registry_url, **rest}: registries[registry_name] = registry_url
+     case _: pass
```

hermeto/core/package_managers/cargo/main.py
```diff
@@ -183,32 +197,77 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
...
+                            registries[registry_name] = registry_config["index"]
+            except (tomlkit.exceptions.TOMLKitError, AttributeError, TypeError) as e:
+                raise UnexpectedFormat(
+                    f"{config.path} containing invalid data, cannot read registries: {e}"
+                )
+
+        # Replace original config with sanitized version
```
a-ovchinnikov: A couple of nitpicks: this comment is incorrect when placed here, the lines it matches merely delete the original config without any replacement, the right way to describe the block beneath would be "Delete the orginial config". Second nitpick: since this code block is worth a descriptive comment it is worth to be promoted to a helper function.

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -88,8 +93,130 @@ def test_virtual_workspace_without_workspace_version(rooted_tmp_path: RootedPath
...
+
+def test_sanitized_cargo_config_file_extracts_and_sanitizes_registries(
+    rooted_tmp_path: RootedPath,
+) -> None:
+    """Test that sanitized config contains only registry information."""
+    cargo_dir = rooted_tmp_path.path / ".cargo"
+    cargo_dir.mkdir()
+
+    # Create a config with registries and other potentially insecure settings
+    config_content = """
```
a-ovchinnikov: Indented triple-quote string could be handled with `textwrap.dedent()`.

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -88,8 +93,130 @@ def test_virtual_workspace_without_workspace_version(rooted_tmp_path: RootedPath
...
+        assert "parity" in sanitized_content
+        assert "internal" in sanitized_content
+        assert "https://github.com/paritytech/crates-parity-io" in sanitized_content
+        assert "sparse+https://artifactory.company.com/cargo" in sanitized_content
+        # Verify insecure settings are NOT present
+        assert "[net]" not in sanitized_content
+        assert "[build]" not in sanitized_content
+        assert "rustflags" not in sanitized_content
+
+    # Verify original config is restored after exiting context
```
a-ovchinnikov: This is a good test in general, the only problem is that it tests four different things at once. WDYT about just splitting it into four smaller, more focused tests? You already have good name candidates for them (in comments preceeding each of individual blocks).

shokakucarrier:
@a-ovchinnikov sorry that was just a concept so I raised a Draft, now I have completely rewrite the whole PR but with same concept, now it's almost ready

a-ovchinnikov:
> now I have completely rewrite the whole PR
Sure, no problem!

docs/cargo.md
```diff
@@ -39,6 +39,25 @@ hermeto inject-files --for-output-dir /tmp/hermeto-output hermeto-output
 
 *There are no environment variables that need to be set for the build phase.*
 
+### Using Alternate Registries
+
+Hermeto supports projects that use alternate Cargo registries (not just
+crates.io). If your `.cargo/config.toml` or `.cargo/config` contains a
+`[registries.<name>]` section, Hermeto will preserve these settings during
+dependency etching to ensure dependencies are pulled from the correct registry.
```
a-ovchinnikov: s/etching/fetching:
```diff
- dependency etching to ensure dependencies are pulled from the correct registry.
+ dependency fetching to ensure dependencies are pulled from the correct registry.
```

hermeto/core/package_managers/cargo/main.py
```diff
@@ -55,11 +56,18 @@ def purl(self) -> PackageURL:
...
+                parsed_url = urlparse(self.source)
+                commit_id = parsed_url.fragment
+                base_url = urlunparse(parsed_url._replace(query="", fragment=""))
+                qualifiers["vcs_url"] = f"{base_url}@{commit_id}"
+            elif self.source.startswith("registry+"):
+                # Extract registry URL from source (format: "registry+https://...")
+                registry_url = self.source[len("registry+") :]
+                # Only add repository_url for non-crates.io registries
+                if "crates.io" not in registry_url:
+                    qualifiers["repository_url"] = registry_url
```
a-ovchinnikov, line 69: I would suggest adding a catch-all exception-raising `else` at the end to ensure we don't accidentally swallow an unexpected value.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -183,11 +191,13 @@ def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
...
 @contextmanager
-def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None, None]:
-    """Hide the cargo config file if it exists.
+def _sanitized_cargo_config_file(package_dir: RootedPath) -> Generator[None, None, None]:
+    """Temporarily sanitize Cargo config files, keeping only alternate registry settings.
 
     The file may contain various settings that could result in potential attack vectors.
     Therefore, it is better to "hide" it before running the `cargo vendor` command.
+    Preserves the [registries] section required for non-crates.io registries while
+    removing potentially dangerous configurations during `cargo vendor`.
```
a-ovchinnikov: TBF the previous docstring looked better to me. At the very least I suggest replacing "Temporarily sanitize..." with "Replace Cargo config files with ones keeping only alternate registry settings." The second sentence could be adjusted a bit too: "The context manager swaps the original config file with one containing just the settings necessary to use alternate registries for the duration of a prefetch and restores everything back to its original state afterwards."

hermeto/core/package_managers/cargo/main.py
```diff
@@ -214,6 +229,48 @@ def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None,
...
+def _sanitize_cargo_config(config_content: str) -> str:
+    """Extract only the [registries] section from Cargo config, keeping only safe fields.
+
+    Preserves only: index, token, credential-provider fields for each registry.
+    Returns sanitized TOML with only registries and their safe fields, or empty string if none exist.
+    """
+    if not config_content.strip():
+        return ""
+
+    try:
```
a-ovchinnikov, line 250: This try block is pretty big, could you please consider splitting it? The standard rule of thumb is to wrap only the most relevant line in a try...except leaving non-raising code paths either in else section or completely outside of it.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -214,6 +229,48 @@ def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None,
...
+        parsed = tomlkit.parse(config_content)
+        if registries := parsed.get("registries"):
```
a-ovchinnikov: PEP8 nitpick: `registries` can be `None` here, hence the perfect check would be `if registries:=parsed.get("registries") is not None` (or you could just return the desired value from `.get()` as in `.get("registries", "")`)

hermeto/core/package_managers/cargo/main.py
```diff
@@ -214,6 +229,48 @@ def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None,
...
+            allowed_fields = {"index", "token", "credential-provider"}
+            filtered_registries = tomlkit.table()
+
+            for registry_name, registry_config in registries.items():
+                if not isinstance(registry_config, dict):
+                    continue
+
+                # Keep only allowed fields for this registry
+                filtered = {
```
a-ovchinnikov: A minor naming nitpick: s/filtered/filtered_fields/ This variable lives on multiple lines, it is better to remind a reader what exactly was filtered.

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -88,8 +93,111 @@ def test_virtual_workspace_without_workspace_version(rooted_tmp_path: RootedPath
...
+        ),
+        pytest.param(
+            "",
+            {},
+            id="empty_config",
+        ),
+    ],
+)
+def test_sanitize_cargo_config(config_input: str, expected_registries: dict[str, set[str]]) -> None:
+    import tomlkit
```
a-ovchinnikov: Please move this import to the top of the module.

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -88,8 +93,111 @@ def test_virtual_workspace_without_workspace_version(rooted_tmp_path: RootedPath
...
+
+    result = _sanitize_cargo_config(config_input)
+
+    if not expected_registries:
```
a-ovchinnikov: Please split the test in two. Having branches in tests is an antipattern which is unfortunately still present in the codebase, let's not add more of it. When splitting the tests please consider naming them accordingly, e.g.:
```
test_cargo_config_without_defined_registries_gets_sanitized_completely
test_cargo_config_with_correctly_defined_registries_drops_everything_except_registries
```
(These names are a mouthpiece, but t is OK since these are tests, tests are special).

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -88,8 +93,111 @@ def test_virtual_workspace_without_workspace_version(rooted_tmp_path: RootedPath
...
+        assert result == ""
+    else:
+        parsed = tomlkit.parse(result)
```
a-ovchinnikov: WDYT about not involving tomlkit at all? I mean you could just put expected sanitized value as a parameter and the compare expected vs actual.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -199,13 +209,18 @@ def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None,
     all_possible_config_names = (".cargo/config", ".cargo/config.toml")
     configs_contents = []
 
+    # Backup and sanitize existing configs
```
a-ovchinnikov: Style nitpick: if you feel that a code block deserves a comment describing _what_ it does (vs describing _why_ it exists) then this code block likely belongs to a separate helper function or two.

a-ovchinnikov (commented):
Mostly LGTM, with some minor requests on restructuring tests and nitpicks on code structure.

tests/unit/package_managers/cargo/test_main.py
shokakucarrier: I tried but it's hard to maintain the indentation and turns out sanitize function is not preserving the order of variables, but I'll try fix it again

hermeto/core/package_managers/cargo/main.py
shokakucarrier: It's pretty straight forward I think, hence doesn't need this comment, so I'll remove it

hermeto/core/package_managers/cargo/main.py
shokakucarrier: It won't be None, but I changed the if condition so it will be more simple, and added two test case to ensure it.

hermeto/core/package_managers/cargo/main.py
shokakucarrier: OK, a else condition and raise of UnexpectedFormat is added

hermeto/core/package_managers/cargo/main.py
shokakucarrier: solved, now it's trying only two lines, and thanks for advice, sorry I don't often consider design pattern

tests/unit/package_managers/cargo/test_main.py
shokakucarrier: OK, I removed tomlkit completely from test, now it looks clean and simple

shokakucarrier:
I think this is ready, should I push it Ready for review?

docs/cargo.md
```diff
@@ -39,6 +39,26 @@ hermeto inject-files --for-output-dir /tmp/hermeto-output hermeto-output
...
+Hermeto supports projects that use alternate Cargo registries (not just
+crates.io). If your `.cargo/config.toml` or `.cargo/config` contains a
+`[registries.<name>]` section, Hermeto will preserve these settings during
+dependency fetching to ensure dependencies are pulled from the correct
+registries.
+
+For security, Hermeto sanitizes the Cargo config file during `cargo vendor`,
+keeping only the `[registries]` section (with `index`, `token`, and
+`credential-provider` fields) while removing other potentially dangerous
+configurations.
```
slimreaper35: I think this information was intentionally omitted from the user-facing documentation since it might confuse users and it is our implementation detail.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -55,11 +56,21 @@ def purl(self) -> PackageURL:
...
+                base_url = urlunparse(parsed_url._replace(query="", fragment=""))
+                qualifiers["vcs_url"] = f"{base_url}@{commit_id}"
+            elif self.source.startswith("registry+"):
+                # Extract registry URL from source (format: "registry+https://...")
+                registry_url = self.source.removeprefix("registry+")
+                if "crates.io" not in registry_url:
+                    qualifiers["repository_url"] = registry_url
+            else:
+                raise UnexpectedFormat(
+                    f"Unable to construct package URL from '{self.source}', package source invalid."
```
slimreaper35:
```suggestion
                    f"Unable to construct package URL from '{self.source}'"

File: hermeto/core/package_managers/cargo/main.py [hunk 18]
```diff
@@ -214,6 +236,46 @@ def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None,
...
+            }
+
+            if filtered_fields:
+                filtered_registries[registry_name] = filtered_fields
+
+        if filtered_registries:
+            sanitized["registries"] = filtered_registries
+            return tomlkit.dumps(sanitized)
+
+    return ""
```
slimreaper35: nitpick: IMO it is easier to read code with early returns than multiple nested levels.

```python3
if registries is not None
   return ""

for registry_name, registry_config in registries.items():
```

docs/cargo.md
shokakucarrier: Ok I thought it was important so I added this, will remove them

hermeto/core/package_managers/cargo/main.py
shokakucarrier: fixed, but TBH the check of registries is None is unnecessary, remove the check it will still return empty str

hermeto/core/package_managers/cargo/main.py
shokakucarrier: thanks for the advise, it's redundant.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -198,14 +208,24 @@ def _hidden_cargo_config_file(package_dir: RootedPath) -> Generator[None, None,
...
-        data = config.path.read_text() if config.path.exists() else None
-        configs_contents.append((config, data))
-        if data is not None:
-            config.path.unlink()
-
+        if config.path.exists():
+            real_path = config.path.resolve()
+            if real_path in processed_paths:
+                continue
+            processed_paths.add(real_path)
```
a-ovchinnikov: Suppose both configs are present and the first one is a link to the second one. Then when the first file is processed path to the second file is added to `processed_paths`. After that data is read from the first location (which in reality is the second one) and written back to it -- which is equivalent to writing to the second file, -- or the first location gets unlinked _if there was no registry-related data in the file_. Then the second file is skipped because we have already seen it. Note, that the skipping takes place unconditionally, so it looks like the second file might just remain in place. Here is a small experiment that I ran:
```python
>>> f1, f2 = Path("/tmp/foo/two"), Path("/tmp/foo/one")
>>> f1
PosixPath('/tmp/foo/two')
>>> f2
PosixPath('/tmp/foo/one')
>>> f2.is_symlink()
False
>>> f1.is_symlink()
True
>>> f1.resolve()
PosixPath('/tmp/foo/one')
>>> f2.resolve()
PosixPath('/tmp/foo/one')
>>> f1.read_text()
'suspicious code\n'
>>> f2.read_text()
'suspicious code\n'
>>> f1.unlink()
>>> f2.read_text()
'suspicious code\n'
```
Do I miss something in the analysis?

shokakucarrier: you are right, I'll propose another fix to this.

shokakucarrier: fixed, now it records all absolute path that's been rewrite only, and it will behave as expected of two files if they are symbol link to another

PR #1267, opened 2026-02-03
Add backend annotations to SBOM Components

taylormadore:
Adds `hermeto:backend` annotations to the SBOM to identify which Hermeto backend fetched each Component. This enables SBOM consumers to determine Component provenance.


hermeto/core/models/property_semantics.py
a-ovchinnikov, line 120: PEP8 nitpick:
```diff
- if self.rpm_modularity_label:
+ if self.rpm_modularity_label is not None:
```

tests/unit/models/test_property_semantics.py
```diff
@@ -257,6 +264,11 @@ def test_conversion_from_and_to_properties(
                 PropertySet(pip_package_binary=True),
                 PropertySet(pip_package_binary=True),
             ),
+            (
+                PropertySet(package_managers=frozenset(["npm"])),
+                PropertySet(package_managers=frozenset(["yarn"])),
+                PropertySet(package_managers=frozenset(["npm", "yarn"])),
```
a-ovchinnikov: Has such behavior been seen in the wild or is this a back-up plan for a theoretical possibility of having identical package being requested twice, once via yarn and other time via npm? (`package_managers` is a bit counter-intuitive TBH, I keep expecting just one PM per component.)

taylormadore: Yeah, it's an unlikely but possible edge-case. Components get deduplicated based on PURL and the purpose of the PropertySet is to merge their lists of Property objects.

hermeto/core/models/property_semantics.py
taylormadore: I left the falsy check there purposefully because I didn't think we wanted to potentially create properties for empty strings

a-ovchinnikov: Now I am confused, what is the semantics of `""` vs `None` then? If we plan to distinguish between them (and judging from annotation we do) then they must mean different things, dropping both sort of blurs this distinction.

hermeto/core/package_managers/gomod.py
```diff
@@ -803,6 +808,8 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
     }
     env_vars_template.update(config.gomod.environment_variables)
 
+    if pm_annotation := create_package_manager_annotation(components, "gomod"):
+        annotations.append(pm_annotation)
```
slimreaper35: Wouldn't this syntax be more flexible for other package managers too? If we plan to add more annotations in the future. Compared to:

```python
annotations=[pm_annotation] if pm_annotation else [],
```

hermeto/core/models/sbom.py
```diff
@@ -152,6 +152,23 @@ def spdx_now() -> datetime:
...
+
+    Returns None if there are no components to annotate.
+    """
+    if not components:
+        return None
+    return Annotation(
+        subjects={c.bom_ref for c in components},
+        annotator={"organization": {"name": "red hat"}},
+        timestamp=spdx_now(),
+        text=f"{APP_NAME}:package_manager:{pm_name}",
```
eskultety: nitpick: `package_manager` is how we colloquially refer to these backends, but with ecosystems that formally do not define something that is a package manager (e.g. the pending WIP on huggingface), do you think we could be more conservative with labeling this simply as `backend`?

hermeto/core/models/sbom.py
```diff
@@ -262,7 +265,8 @@ def generate_package_annotations(properties: list[Property]) -> list[SPDXPackage
             )
 
             for annotation in self.annotations:
-                result.append(base_spdx_annotation(comment=annotation.text))
+                if bom_ref is not None and bom_ref in annotation.subjects:
```
eskultety: nitpick: `if bom_ref is not None` will become deadcode in patch 2/3, so we can optimize the check in that commit.

hermeto/core/models/sbom.py
eskultety: Same for the function name probably, but that one isn't user-facing, so up to you.

tests/unit/package_managers/yarn/test_main.py
```diff
@@ -441,15 +442,24 @@ def test_generate_environment_variables(yarn_env_variables: list[EnvironmentVari
     ),
     indirect=["input_request"],
 )
+@mock.patch("hermeto.core.package_managers.yarn.main.create_package_manager_annotation")
```
eskultety: I noticed the pip unit test variant doesn't mock this and runs (with the exception of `_resolve_pip`) the real code including the SBOM handling, was it skipped from mocking this deliberately? Should we mock it or actually compare the annotation data since we filled in everything correctly and have means of doing so?

hermeto/core/package_managers/gomod.py
eskultety: It most certainly would establish good consistency. Unless there are other things to update in the code, not a blocker for me, but a simple nice to have for sure.

eskultety (commented):
A few nits, LGTM.

hermeto/core/models/property_semantics.py
taylormadore: Commit dropped in V2 when we switched from Property --> Annotation

hermeto/core/package_managers/gomod.py
taylormadore: Agreed, will fix

hermeto/core/models/sbom.py
taylormadore: Since this is net-new, I'm fine with this :+1: 

We'll probably have to adjust our naming/terminology more widely at some point

tests/unit/package_managers/yarn/test_main.py
taylormadore: Nope, just an oversight on my part. I knew bundler and cargo didn't have equivalent fetch_source tests, but I  also accidentally skipped this one too :sweat_smile:

taylormadore:
Rebased on main

hermeto/core/package_managers/bundler/main.py
```diff
@@ -48,10 +48,14 @@ def fetch_bundler_source(request: Request) -> RequestOutput:
         _prepare_for_hermetic_build(request.source_dir, request.output_dir, git_paths)
     )
 
+    annotations = []
+    if pm_annotation := create_backend_annotation(components, "bundler"):
+        annotations.append(pm_annotation)
```
slimreaper35: nitpick: this is no longer `pm` (multiple occurrences)

taylormadore: Good point, fixed

taylormadore:
Rebased on main

PR #1270, opened 2026-02-04
Update Rust base image and Cargo.lock error detection

slimreaper35:

### DIFF
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -138,7 +138,7 @@ def _fetch_dependencies(package_dir: RootedPath, request: Request) -> dict[str, 
         config_template = _run_cmd_watching_out_for_lock_mismatch(
             cmd=cmd,
             params={"cwd": package_dir, "env": env},
-            package_dir=package_dir,
+            package_dir=package_dir.path,
             mode=request.mode,
         )
 
@@ -223,13 +223,13 @@ def _temporary_cwd(path_to_new_cwd: Path) -> Generator[None, None, None]:
 
 
 def _run_cmd_watching_out_for_lock_mismatch(
-    cmd: list, params: dict, package_dir: RootedPath, mode: Mode
+    cmd: list, params: dict, package_dir: Path, mode: Mode
 ) -> str:
     warn_about_imminent_update_to_cargo_lock = (
         f"A mismatch between Cargo.lock and Cargo.toml was detected in {package_dir}. "
         "Because of permissive mode Hermeto will now regenerate Cargo.lock "
         "to match expectation and will try to process the package again. This "
-        f"is a violation of reproducibility and must be addressed by {package_dir.path.name} "
+        f"is a violation of reproducibility and must be addressed by {package_dir.name} "
         "maintainers."
     )
     update_cargo_lock_cmd = ["cargo", "generate-lockfile"]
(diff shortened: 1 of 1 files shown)


slimreaper35:
Do we want to keep the "backwards" compatibility or even drop the second marker?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -242,7 +242,7 @@ def _run_cmd_watching_out_for_lock_mismatch(
         # others, thus the only way to check for this specific type of failure is to process
         # stderr. Two parts of a string are used to decrease the likelihood of false positives.
         lock_corruption_marker1 = "failed to sync"
-        lock_corruption_marker2 = "needs to be updated but --locked was passed"
+        lock_corruption_marker2 = "because --locked was passed to prevent this"
```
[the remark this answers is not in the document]
a-ovchinnikov: Unrelated to the comment above: are we planning to always run on the latest cargo? I am slightly concerned about the rare case when someone runs into this issue while also using an older version. Would it make sense to add more band aid here by making the second corruption marker a tuple of strings instead and then checking that any one of those is present in stderr?

slimreaper35:
> are we planning to always run on the latest cargo?

@eskultety disabled automatic updates of Rust image in https://github.com/hermetoproject/hermeto/commit/dc9f4c4871a9dd265fcce2706e2b5533cbfa32be

> I am slightly concerned about the rare case when someone runs into this issue while also using an older version. Would it make sense to add more band aid here by making the second corruption marker a tuple of strings instead and then checking that any one of those is present in stderr?

Basically, that's what I asked [above](https://github.com/hermetoproject/hermeto/pull/1270#issuecomment-3848188541). The problem would be if a user has a lower Cargo version and tries Hermeto locally. That is valid, but since it is only related to permissive mode, which should not be used at all, I am not sure.

eskultety:
> > are we planning to always run on the latest cargo?
> 
> @eskultety disabled automatic updates of Rust image in [dc9f4c4](https://github.com/hermetoproject/hermeto/commit/dc9f4c4871a9dd265fcce2706e2b5533cbfa32be)
> 

I only disabled major/minor updates, not patch releases, just to be safe (well, golang...)

> > I am slightly concerned about the rare case when someone runs into this issue while also using an older version. Would it make sense to add more band aid here by making the second corruption marker a tuple of strings instead and then checking that any one of those is present in stderr?

> Do we want to keep the "backwards" compatibility or even drop the second marker?

Yes, to a point...this is again where we're lacking the platform support claim, IOW saying that we try to support most modern platforms (Debian, Fedora, ...) up to X years old OR X versions old. What that means in practice for us is that we can "reasonably" assume the versions of various toolchains releases.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -241,12 +241,17 @@ def _run_cmd_watching_out_for_lock_mismatch(
         # readily available: cargo returns a generic 101 code on this failure and on multiple
         # others, thus the only way to check for this specific type of failure is to process
         # stderr. Two parts of a string are used to decrease the likelihood of false positives.
-        lock_corruption_marker1 = "failed to sync"
-        lock_corruption_marker2 = "needs to be updated but --locked was passed"
-        if lock_corruption_marker1 in e.stderr and lock_corruption_marker2 in e.stderr:
+        generic_vendor_error_msg = "failed to sync"
+        # Since Cargo version 1.93.0, the error message for an unsynchronized lockfile has changed.
+        old_vendor_error_msg = "needs to be updated but --locked was passed"
```
eskultety: Fair enough, we might end up with a list though eventually :(

eskultety (approved):

hermeto/core/package_managers/cargo/main.py
```diff
@@ -241,9 +241,14 @@ def _run_cmd_watching_out_for_lock_mismatch(
...
+        new_vendor_error_msg = "because --locked was passed to prevent this"
```
a-ovchinnikov: A non-blocking suggestion: just create a tuple right away, this is volatile and adding another element is slightly less work than amending variable names.

slimreaper35: done

PR #1275, opened 2026-02-05
cleanup: removed unused .gitignore entries

ST3V1K:
Removed entries that were copied over from cachito and/or are not project related now.
closes https://github.com/hermetoproject/hermeto/issues/597

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,9 +3,6 @@ __pycache__/
 *.py[cod]
 *$py.class
 
-# C extensions
-*.so
-
 # Distribution / packaging
 .Python
 build/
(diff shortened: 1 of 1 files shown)


.gitignore
a-ovchinnikov: We are not using tox anymore, it should be safe to remove it too.

.gitignore
```diff
@@ -48,40 +45,6 @@ coverage.xml
 .hypothesis/
```
a-ovchinnikov: I do not think we actually use hypothesis (albeit we probably could benefit from it), this should be safe to remove too.

.gitignore
eskultety: I just noticed this one as well, the whole PyInstaller can also be safely deleted, we don't use this.

.gitignore
```diff
@@ -91,10 +52,6 @@ ENV/
 env.bak/
 venv.bak/
 
-# Spyder project settings
-.spyderproject
-.spyproject
-
 # Rope project settings
 .ropeproject
```
eskultety, line 49: I would delete this as well normally, but since this is some kind of Python refactoring tool it's hard to say if users are not using it and hence falls into the same category like vscode --> I told @ST3V1K to keep it.

.gitignore
```diff
@@ -37,51 +34,15 @@ pip-delete-this-directory.txt
 
 # Unit test / coverage reports
 htmlcov/
-.tox/
 .nox/
 .coverage
 .coverage.*
 .cache
 nosetests.xml
```
eskultety: We use pytest, based on the other Python vaguely-related stuff being dropped, this line can go as well.

eskultety (commented):
The problem of our .gitignore in the first place is that it was based on some common online Python gitignore template that was further enhanced with custom entries and so now it's hard to say where do the respective entries coming from really.
@ST3V1K minor nitpicks only, once you update, I'll approve and merge, thank you.

slimreaper35:
> The problem of our .gitignore in the first place is that it was based on some common online Python gitignore template that was further enhanced with custom entries and so now it's hard to say where do the respective entries coming from really.

https://github.com/github/gitignore/blob/main/Python.gitignore

PR #1278, opened 2026-02-06
cli: split --version and package-managers output

ST3V1K:
--version mixed version and package-manager list.
This is non-standard behavior for cli tools.
Separate --package-managers option fixes that.

closes https://github.com/hermetoproject/hermeto/issues/1234

### DIFF
--- a/hermeto/core/resolver.py
+++ b/hermeto/core/resolver.py
@@ -24,8 +24,14 @@ _package_managers: dict[PackageManagerType, Handler] = {
 }
 
 
-# This is *only* used to provide a list for `hermeto --version`
-supported_package_managers = list(_package_managers)
+def get_supported_backends() -> list[PackageManagerType | str]:
+    """List all supported backends sorted alphabetically by name."""
+    return sorted(x for x in _package_managers if not x.startswith("x-"))
+
+
+def get_experimental_backends() -> list[PackageManagerType | str]:
+    """List all experimental backends sorted alphabetically by name."""
+    return sorted(x for x in _package_managers if x.startswith("x-"))
 
 
 def resolve_packages(request: Request) -> RequestOutput:
--- a/hermeto/interface/cli.py
+++ b/hermeto/interface/cli.py
@@ -19,7 +19,12 @@ from hermeto.core.extras.envfile import EnvFormat, generate_envfile
 from hermeto.core.models.input import Flag, Mode, PackageInput, Request, parse_user_input
 from hermeto.core.models.output import BuildConfig
 from hermeto.core.models.sbom import Sbom, SPDXSbom, spdx_now
-from hermeto.core.resolver import inject_files_post, resolve_packages
+from hermeto.core.resolver import (
+    get_experimental_backends,
+    get_supported_backends,
+    inject_files_post,
+    resolve_packages,
+)
 from hermeto.core.rooted_path import RootedPath
 from hermeto.interface.logging import LogLevel, setup_logging
 
@@ -219,6 +224,19 @@ def _looks_like_input_file(value: str) -> bool:
 class _Input(pydantic.BaseModel, extra="forbid"):
     packages: list[PackageInput]
     flags: list[Flag] = list()
+
+
+@app.command()
+@handle_errors
+def list_backends() -> None:
+    """List supported backends."""
+    supported = get_supported_backends()
+    if supported:
+        print("Supported:", ", ".join(supported))
+
+    experimental = get_experimental_backends()
+    if experimental:
+        print("Experimental:", ", ".join(experimental))
 
 
 @app.command(help=FETCH_DEPS_HELP)
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -89,6 +89,16 @@ class TestTopLevelOpts:
         result = invoke_expecting_sucess(app, ["--version"])
         assert expect_version in result.output
 
+    @mock.patch(
+        "hermeto.core.resolver._package_managers",
+        {"foo": mock.Mock(), "bar": mock.Mock(), "x-baz": mock.Mock()},
+    )
+    def test_list_backends_option(self) -> None:
+        result = invoke_expecting_sucess(app, ["list-backends"])
+        lines = result.output.splitlines()
+        assert "bar, foo" in lines[0]
+        assert "x-baz" in lines[1]
+
     @pytest.mark.parametrize(
         "config_values",
         [


hermeto/core/resolver.py
```diff
@@ -24,7 +24,7 @@
 }
```
slimreaper35, line 24: nitpick: we could also sort package managers alphabetically

hermeto/core/resolver.py
```diff
@@ -24,7 +24,7 @@
...
 
 
-# This is *only* used to provide a list for `hermeto --version`
+# This is *only* used to provide a list for `hermeto --package-managers`
```
eskultety: Pre-existing: this commentary has always been odd, we should remove it, nothing that `git grep` on the code base would not reveal :).

hermeto/core/resolver.py
```diff
@@ -24,7 +24,7 @@
...
 supported_package_managers = list(_package_managers)
```
eskultety: Since you're touching this code, this is IMO not very idiomatic, i.e. defining a regular (i.e. not a constant) variable just to import it from a different module. I think we can do better:

- either define `_package_managers` as a constant that you'll then import from `cli.py` OR
- define this as a function, e.g. `list_backends` returning this list and import that one

This would be done in a standalone commit, first in the series.

hermeto/interface/cli.py
```diff
@@ -148,7 +148,15 @@ def version_callback(value: bool) -> None:
...
+    print(importlib.metadata.version("hermeto"))
+    raise typer.Exit()
+
+
+def package_managers_callback(value: bool) -> None:
+    """If --package-managers was used, print the supported package managers and exit."""
+    if not value:
+        return
+
     print("Supported package managers:", ", ".join(supported_package_managers))
```
eskultety: I would drop this whole sentence prefix to keep the output easily machine parseable (however useful it may be).

hermeto/interface/cli.py
```diff
@@ -163,6 +171,13 @@ def main(  # noqa: D103 docstring becomes part of --help message
         is_eager=True,
         help="Show version and exit.",
     ),
+    package_managers: bool = typer.Option(  # noqa: ARG001
+        False,
+        "--package-managers",
```
eskultety: variable naming doesn't matter much, because that's internal and we all know what it means. However, `--package-managers` is user facing. `generic` isn't a package manager it's just an artificial backend construct we had to come up with to effectively support `curl`-like behaviour. Since we're adding a new CLI option I think `--list-backends` is a safer choice here.

hermeto/interface/cli.py
```diff
@@ -148,7 +148,15 @@ def version_callback(value: bool) -> None:
     if not value:
         return
 
-    print(f"{APP_NAME}", importlib.metadata.version("hermeto"))
+    print(importlib.metadata.version("hermeto"))
```
eskultety: You're dropping `{APP_NAME}` here. That's okay to do, but since that is something to clean up the existing version output ---> standalone commit (most likely at the end of the patch series).

tests/unit/test_cli.py
```diff
@@ -87,9 +87,11 @@ class TestTopLevelOpts:
...
         expect_version = importlib.metadata.version(f"{APP_NAME}")
         result = invoke_expecting_sucess(app, ["--version"])
-        lines = result.output.splitlines()
-        assert lines[0] == f"{APP_NAME} {expect_version}"
-        assert lines[1].startswith("Supported package managers: bundler")
+        assert result.output.rstrip() == expect_version
+
+    def test_package_managers_option(self) -> None:
+        result = invoke_expecting_sucess(app, ["--package-managers"])
+        assert result.output.startswith("Supported package managers: bundler")
```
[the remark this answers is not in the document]
eskultety: The test will have to be adjusted anyway, so resolving for now.

hermeto/interface/cli.py
```diff
@@ -148,7 +148,15 @@ def version_callback(value: bool) -> None:
...
+    raise typer.Exit()
+
+
+def package_managers_callback(value: bool) -> None:
```
eskultety: This package manager output inclusion as part of `--version` was explicitly added in 57952f2eb7a239e86bfe2b045af8bb2418dce513 as a nice to have. You could mention this in the commit message, something like:
_This reverts commit 57952f2eb7a239e86bfe2b045af8bb2418dce513_

tests/unit/test_cli.py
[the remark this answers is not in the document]
a-ovchinnikov: Please let's not test for string matches! Such tests don't really tell us much about the tested code and also are very brittle, what they often end up doing is a) repeatedly telling us that Python's string manipulation library still works (which is not our area of responsibility) and b) producing a minor distraction when we add say `apt` to the bunch. If a test feels necessary then maybe a doctest could be a better option? (I would have even considered going without a test for this one since it is not part of busyness logic and is, in fact, a rather trivial combination of elements which has a very limited effect on user experience.)

hermeto/interface/cli.py
eskultety: I hit 'send' too soon. I thought about this some more and I think we might be better off with a new subcommand - `list-backends` - rather than a toggle, because then it would give us greater flexibility, like:

- list supported backends
- list experimental backends
- dump them in a table with some description
... pretty much anything

tests/unit/test_cli.py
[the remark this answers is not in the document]
eskultety: Sure, I don't comment much on tests when there are comments about logic requiring another iteration :). Anyway, we could still do this, but in a different way - we could instead mock the callback and make sure we always call the right callback with every CLI subcommand with the right options.

hermeto/core/resolver.py
ST3V1K: Added sorting in new methods that return the backends.

hermeto/core/resolver.py
ST3V1K: Added two new methods. One for all backends and one for only experimental.

hermeto/interface/cli.py
ST3V1K: Created new subcommand `list-backends` with the option to print only experimental backends.

slimreaper35:
```
hermeto on git i1234 [!] via py v3.13.11 (venv) 
> hermeto list-backends       
bundler, cargo, generic, gomod, npm, pip, rpm, yarn, x-foo

hermeto on git i1234 [!] via py v3.13.11 (venv) 
> hermeto list-backends -x    
x-foo
```

I tried this subcommand, and the `--experimental` flag seems redundant here.

hermeto/interface/cli.py
```diff
@@ -222,6 +226,23 @@ class _Input(pydantic.BaseModel, extra="forbid"):
...
+@handle_errors
+def list_backends(
+    experimental: bool = typer.Option(
+        False,
+        "-x",
+        "--experimental",
+        help="List experimental backends.",
+    ),
+) -> None:
+    """Lists supported backends sorted alphabetically by type."""
```
slimreaper35:
```suggestion
    """List supported backends."""
```

hermeto/interface/cli.py
```diff
@@ -148,8 +153,7 @@ def version_callback(value: bool) -> None:
     if not value:
         return
 
-    print(f"{APP_NAME}", importlib.metadata.version("hermeto"))
-    print("Supported package managers:", ", ".join(supported_package_managers))
+    print(importlib.metadata.version("hermeto"))
```
slimreaper35: If the parsing is the only argument for dropping APP_NAME, I would probably keep it. I think it is a more common format (see `python --version`, `cargo --version`)

hermeto/interface/cli.py
slimreaper35: Also, when creating the issue, I had in mind an explicit notice "Experimental package managers...". For a first-time user, it might not be clear what `x-` is unless he carefully reads our contributing guidelines.

TL;DR: the subcommand could be more verbose, but that's just my nitpicking.

ST3V1K: > I tried this subcommand, and the `--experimental` flag seems redundant here.

Is this what you had in mind?
```sh
$ hermeto list-backends
bundler, cargo, generic, gomod, npm, pip, rpm, yarn

$ hermeto list-backends -x
foo

$ hermeto list-backends -a
bundler, cargo, foo, generic, gomod, npm, pip, rpm, yarn
```

> Also, when creating the issue, I had in mind an explicit notice "Experimental package managers...". For a first-time user, it might not be clear what x- is unless he carefully reads our contributing guidelines.

Removed the `x-` prefix and added a notice to help

hermeto/interface/cli.py
a-ovchinnikov, line 1: _Commit message_: it does not look like this commit reverts a commit from the past, it modifies the behavior to match expectations, but does not apply a diff in reverse.

tests/unit/test_cli.py
```diff
@@ -89,6 +89,37 @@ def test_version_option(self) -> None:
...
+        "all, experimental",
+        [
+            (False, False),
+            (False, True),
+            (True, False),
+            (True, True),
+        ],
+    )
+    def test_list_backends(self, all: bool, experimental: bool) -> None:
+        if all:
```
a-ovchinnikov: This name shadows a built-in. Not a big issue _here_, but something to be aware of. An underscore before or after the name would resolve the issue.

hermeto/interface/cli.py
```diff
@@ -221,6 +227,33 @@ class _Input(pydantic.BaseModel, extra="forbid"):
...
+        help="List backends that are experimental (x-*): technical preview, not for production; no stability or support guarantees.",
+    ),
+) -> None:
+    """List supported backends."""
+    if list_all:
+        backends = get_all_backends(True)
+    elif list_experimental:
+        backends = get_experimental_backends(True)
+    else:
+        backends = get_supported_backends()
```
a-ovchinnikov: I have just checked and hermeto allowed me to specify both flags, this feels a bit untidy, could you please  ensure that just one flag was issued? Another minor suggest: have you considered `match..case` construct instead of `if..elif..else` for this?
```
match list_all, list_experimental:
    case True, False:  backends = get_all_backends(True)
    case False, True:  backends = get_experimental_backends(True)
    case False, False: backends = get_supported_backends()
    case True, True: raise Exception("Please pick one option")
```

tests/unit/test_cli.py
```diff
@@ -89,6 +89,37 @@ def test_version_option(self) -> None:
...
+        elif experimental:
+            patch_target = "hermeto.interface.cli.get_experimental_backends"
+            args = ["list-backends", "--experimental"]
+            backends = ["bar", "foo"]
+        else:
+            patch_target = "hermeto.interface.cli.get_supported_backends"
+            args = ["list-backends"]
+            backends = ["bundler", "cargo"]
+
+        with mock.patch(patch_target) as mocked:
```
a-ovchinnikov: Why don't you patch `_package_managers` to a universal known value instead? E.g. to `foo, x-bar`

tests/unit/test_cli.py
```diff
@@ -89,6 +89,37 @@ def test_version_option(self) -> None:
         result = invoke_expecting_sucess(app, ["--version"])
         assert expect_version in result.output
 
+    @pytest.mark.parametrize(
+        "all, experimental",
+        [
+            (False, False),
+            (False, True),
+            (True, False),
+            (True, True),
```
a-ovchinnikov: Parametrization makes it look like you are executing the flags when in reality you are not executing the flags, but instead you make an arbitrary partial substitution table within the test which tests the same code path twice. This is not the best approach even setting aside the question of general usefulness of such tests (see my earlier point about writing tests for third party tools). I suggest the following changes:
1. parametrizing this test with actual command or at least the relevant part of it;
2. adding expected balues to parameters;
3. removing all branching within the test (it is an antipattern);
4. mocking the most distant common object (in this case it will be the set of PMs).

hermeto/core/resolver.py
```diff
@@ -24,8 +24,21 @@
 }
 
 
-# This is *only* used to provide a list for `hermeto --version`
-supported_package_managers = list(_package_managers)
+def get_all_backends(remove_prefix: bool = True) -> list[PackageManagerType | str]:
+    """List all backends sorted alphabetically by name"""
+    if remove_prefix:
+        return sorted([x.lstrip("x-") for x in _package_managers])
```
a-ovchinnikov: You can skip brackets when list comprehension is an argument:
```diff
- return sorted([x.lstrip("x-") for x in _package_managers])
+ return sorted(x.lstrip("x-") for x in _package_managers)
```

hermeto/interface/cli.py
slimreaper35: I think two options for this command might be overkill. I had in mind this actually:

```bash
$ hermeto list-backends
Supported: bundler, cargo, generic, gomod, npm, pip, rpm, yarn
Experimental: x-foo
```

But currently, we don't have any experimental pkg managers (it will change soon), so the experimental would be empty.

hermeto/interface/cli.py
```diff
@@ -221,6 +227,40 @@ class _Input(pydantic.BaseModel, extra="forbid"):
...
+    ),
+    list_experimental: bool = typer.Option(
+        False,
+        "-x",
+        "--experimental",
+        help="List backends that are experimental (x-*): technical preview, not for production; no stability or support guarantees.",
+    ),
+) -> None:
+    """List supported backends."""
+    backends = None
```
a-ovchinnikov: Minor nitpick: you do not need this initialization.

hermeto/interface/cli.py
```diff
@@ -221,6 +227,40 @@ class _Input(pydantic.BaseModel, extra="forbid"):
...
+    match list_all, list_experimental:
+        case True, False:
+            backends = get_all_backends(True)
+        case False, True:
+            backends = get_experimental_backends(True)
+        case False, False:
+            backends = get_supported_backends()
+        case _:
```
a-ovchinnikov: Minor style nitpick: here you want to guard against a specific case, namely `True, True` (judging by the error message), however you guard against catch-all instead (which is very unlikely to happen here). I would recommend either keeping it simple and checking for `True, True`, or (if you want to guard against extremely unlikely chain of events caused by a bug somewhere else) checking for `True, True`, but also for `_ as impossible` and raising a separate internal error in that case (something along the lines of `CoreMeltdown(f"{impossible} happened")`) -- because it is a different situation to guard against and should not be lumped together with a pattern which we expect and fully know ahead of time.

hermeto/interface/cli.py
```diff
@@ -222,6 +227,39 @@
...
+            raise InvalidInput(
+                "Please pick only one option.",
+                solution="Please select either --all or --experimental, not both.",
+            )
+
+    print(", ".join(backends))
```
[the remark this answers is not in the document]
a-ovchinnikov: Nope.

tests/unit/test_cli.py
```diff
@@ -89,6 +89,42 @@ def test_version_option(self) -> None:
...
+            clear=True,
+        ):
+            result = invoke_expecting_sucess(app, args)
+
+        items = [item.strip() for item in result.output.split(",")]
+        assert items == expected_backends
+
+    def test_list_backends_option_invalid(self) -> None:
+        result = invoke_expecting_invalid_usage(app, ["list-backends", "--all", "--experimental"])
+        assert "Please pick only one option." in result.output
```
a-ovchinnikov: I would again strongly recommend against doing string comparisons in tests: they make tests brittle. In rare cases one has to do that, but I am not convinced that this is the case. Surely one could execute `list_backends(True, True)` in isolation and verify that it raised an exception?

hermeto/interface/cli.py
slimreaper35: This is still unresolved. I think having two CLI flags for this informative command is overkill and adds unnecessary complexity. But if there are more opposite opinions, I guess I am fine.

eskultety: > This is still unresolved. I think having two CLI flags for this informative command is overkill and adds unnecessary complexity. But if there are more opposite opinions, I guess I am fine.

Truth to be told I'm in the same boat. While my original response

> * list supported backends
> 
> * list experimental backends
> 
> * dump them in a table with some description
>   ... pretty much anything

may have indirectly hinted at the need for a dedicated `experimental` flag, the point of that message was the flexibility of a new subcommand in general. Another thing I like about that flexibility is that we could provide easily parseable output thanks to that granularity, but that's something that we should explore only when users request it or when we ourselves feel limited by. This is not the case and a simple combined output (as suggested) seems pretty clean to me :).

hermeto/interface/cli.py
eskultety: wow, TIL about `match` and that it was introduced in 3.10, thumbs up.

tests/unit/test_cli.py
eskultety: Yep.

hermeto/interface/cli.py
a-ovchinnikov: Yes, `match` is almost as good as in languages which are serious about structured matching. Take a loop at [PEP-636](https://peps.python.org/pep-0636/) if you missed it, it is worth it.

hermeto/interface/cli.py
ST3V1K: done

hermeto/interface/cli.py
ST3V1K: removed

hermeto/interface/cli.py
ST3V1K: removed

tests/unit/test_cli.py
ST3V1K: removed

hermeto/core/resolver.py
```diff
@@ -24,8 +24,21 @@
 }
 
 
-# This is *only* used to provide a list for `hermeto --version`
-supported_package_managers = list(_package_managers)
+def get_all_backends(remove_prefix: bool = True) -> list[PackageManagerType | str]:
```
slimreaper35: This function is not used anywhere.

hermeto/core/resolver.py
```diff
@@ -24,8 +24,21 @@
...
+        return sorted(x.lstrip("x-") for x in _package_managers)
+    return sorted(_package_managers)
+
+
+def get_supported_backends() -> list[PackageManagerType | str]:
+    """List all supported backends sorted alphabetically by name."""
+    return sorted(x for x in _package_managers if not x.startswith("x-"))
+
+
+def get_experimental_backends(remove_prefix: bool = False) -> list[PackageManagerType | str]:
```
slimreaper35: What is the intention of the `remove_prefix` parameter?

tests/unit/test_cli.py
```diff
@@ -89,6 +89,26 @@ def test_version_option(self) -> None:
...
 
+    @pytest.fixture
+    def _fake_package_managers(self) -> dict[str, Any]:
+        return {
+            "foo": mock.Mock(return_value=RequestOutput.empty()),
+            "bar": mock.Mock(return_value=RequestOutput.empty()),
+            "x-baz": mock.Mock(return_value=RequestOutput.empty()),
+        }
+
+    def test_list_backends_option(self, _fake_package_managers: dict[str, Any]) -> None:
```
slimreaper35: nitpick: I think this could be simplified with something like:

```
    @mock.patch(
        "hermeto.core.resolver._package_managers",
        {"foo": mock.Mock(), "bar": mock.Mock(), "x-baz": mock.Mock()},
    )
```

hermeto/core/resolver.py
ST3V1K: removed it

hermeto/core/resolver.py
ST3V1K: It was if anyone would need it in the future, but that will probably never happen.
removed it

tests/unit/test_cli.py
ST3V1K: done

PR #1282, opened 2026-02-09
Switched most integration test images to alpine

ST3V1K:
Had to change gomod version 1.18.1 to 1.18.10 due to missing git on 1.18.1 alpine image.

closes https://github.com/hermetoproject/hermeto/issues/1241

### DIFF
--- a/tests/integration/test_data/gomod_e2e_1.18/container/Containerfile
+++ b/tests/integration/test_data/gomod_e2e_1.18/container/Containerfile
@@ -1,7 +1,7 @@
 FROM mirror.gcr.io/golang:1.18.10-alpine
 
 # Test disabled network access
-RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
+RUN if getent hosts www.google.com; then echo "Has network access!"; exit 1; fi
 
 WORKDIR /src
 RUN . /tmp/hermeto.env && go build -o /usr/bin/retrodep
--- a/tests/integration/test_data/multiple_gomod_and_npm/container/Containerfile
+++ b/tests/integration/test_data/multiple_gomod_and_npm/container/Containerfile
@@ -1,7 +1,7 @@
 FROM registry.access.redhat.com/ubi9/ubi@sha256:304b50df1ea4db9706d8a30f4bbf26f582936ebc80c7e075c72ff2af99292a54
 
 # Test disabled network access
-RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
+RUN if getent hosts www.google.com; then echo "Has network access!"; exit 1; fi
 
 # Install dependencies from RPM lockfile
 RUN dnf -y install golang npm
--- a/tests/integration/test_data/yarn_e2e_multiple_packages/container/Containerfile
+++ b/tests/integration/test_data/yarn_e2e_multiple_packages/container/Containerfile
@@ -1,7 +1,7 @@
 FROM mirror.gcr.io/node:18.19.0-alpine
 
 # Test disabled network access
-RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
+RUN if getent hosts www.google.com; then echo "Has network access!"; exit 1; fi
 
 WORKDIR /src/first-pkg
 RUN . /tmp/hermeto.env && yarn install
--- a/tests/integration/test_data/bundler_e2e/container/Containerfile
+++ b/tests/integration/test_data/bundler_e2e/container/Containerfile
@@ -1,7 +1,7 @@
 FROM mirror.gcr.io/ruby:3.3
 
 # Test disabled network access
-RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
+RUN if getent hosts www.google.com; then echo "Has network access!"; exit 1; fi
 
 WORKDIR /src
 
--- a/tests/integration/test_data/bundler_e2e_missing_gemspec/container/Containerfile
+++ b/tests/integration/test_data/bundler_e2e_missing_gemspec/container/Containerfile
@@ -1,7 +1,7 @@
 FROM mirror.gcr.io/ruby:3.3
 
 # Test disabled network access
-RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
+RUN if getent hosts www.google.com; then echo "Has network access!"; exit 1; fi
 
 WORKDIR /src
 
--- a/tests/integration/test_data/cargo_mixed_git_crate_dependency/container/Containerfile
+++ b/tests/integration/test_data/cargo_mixed_git_crate_dependency/container/Containerfile
@@ -1,7 +1,7 @@
 FROM mirror.gcr.io/rust:1-alpine
 
 # Test disabled network access
-RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
+RUN if getent hosts www.google.com; then echo "Has network access!"; exit 1; fi
 
 WORKDIR /src
 RUN . /tmp/hermeto.env && cargo build --offline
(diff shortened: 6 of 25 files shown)


tests/integration/test_data/yarn_e2e_multiple_packages/container/Containerfile
```diff
@@ -1,4 +1,4 @@
-FROM mirror.gcr.io/node:18.19.0
+FROM mirror.gcr.io/node:18.19.0-alpine
```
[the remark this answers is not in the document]
slimreaper35: Hopefully, we can configure Gemini to not duplicate identical comments 16x 💀

slimreaper35 (approved):

tests/integration/test_data/gomod_e2e_1.18/container/Containerfile
```diff
@@ -1,4 +1,4 @@
-FROM mirror.gcr.io/golang:1.18.1
+FROM mirror.gcr.io/golang:1.18.10-alpine
```
eskultety, line 1: **Commit message:**

- the commit message should contain the note from the PR description on why the golang version had to be bumped from 1.18.1 to 1.18.10
- the commit message should mention why the rest of the images either could not be converted or simply N/A, e.g. the Ruby image would fail builds with:
```
STEP 4/4: RUN . /tmp/hermeto.env && bundle install
You need to install git to be able to use gems from git repositories. For help
...
```
Similarly, the pip e2e wheels test would have to be adjusted to not require c++ to build a wheel, etc.

eskultety: Lastly, the `multiple_gomod_and_npm` test can technically be converted too (see the diff below), however, the corresponding test should be changed accordingly, so we can do that in some follow-up work (ideally file an issue). So at least document it in the commit message :).

```diff
diff --git tests/integration/test_data/multiple_gomod_and_npm/container/Containerfile tests/integration/test_data/multiple_gomod_and_npm/cont
ainer/Containerfile
index bf83b1c46..cff88d20a 100644
--- tests/integration/test_data/multiple_gomod_and_npm/container/Containerfile
+++ tests/integration/test_data/multiple_gomod_and_npm/container/Containerfile
@@ -1,10 +1,16 @@
-FROM registry.access.redhat.com/ubi9/ubi@sha256:304b50df1ea4db9706d8a30f4bbf26f582936ebc80c7e075c72ff2af99292a54
+# FROM registry.access.redhat.com/ubi9/ubi@sha256:304b50df1ea4db9706d8a30f4bbf26f582936ebc80c7e075c72ff2af99292a54
+FROM mirror.gcr.io/golang:1.25-alpine as golang
+FROM mirror.gcr.io/node:24-alpine as node
+
+FROM node
+COPY --from=golang /usr/local/go /usr/local/go

 # Test disabled network access
-RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
+RUN if getent hosts www.google.com; then echo "Has network access!"; exit 1; fi
+RUN ln -s /usr/local/go/bin/go /usr/local/bin/go

-# Install dependencies from RPM lockfile
-RUN dnf -y install golang npm
+# # Install dependencies from RPM lockfile
+# RUN dnf -y install golang npm

 WORKDIR /src/gomod-package
 RUN . /tmp/hermeto.env && go build
```

eskultety:
One more thing - this work could also be paired with replacing the `curl` network detection logic with `getent hosts` globally.

tests/integration/test_data/cargo_e2e/container/Containerfile
```diff
@@ -1,4 +1,4 @@
-FROM mirror.gcr.io/rust:1.88-slim
+FROM mirror.gcr.io/rust:1.93-alpine
```
eskultety, line 1: **Commit message:**

> Alpine-based integration tests were failing with the previous runtime
> versions; bumping these images fixes the failures.

I don't think this is a correct statement. This commit converts a bunch of test images to Alpine to have a common container image base across the majority of our test suite and keep the footprint of the downloaded images smaller. However, our integration tests didn't fail at all so it has nothing to do with Alpine.
Was the commit message written by an AI by any chance? :)

tests/integration/test_data/cargo_e2e/container/Containerfile
a-ovchinnikov, line 1: _Commit message_: two signed-off-by lines are present, also it is a bit unclear what the part after `---` means, could you please elaborate on this?

tests/integration/test_data/multiple_gomod_and_npm/container/Containerfile
```diff
@@ -1,10 +1,14 @@
...
+FROM node
+COPY --from=golang /usr/local/go /usr/local/go
 
 # Test disabled network access
 RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
 
-# Install dependencies from RPM lockfile
-RUN dnf -y install golang npm
+# Link go to standard PATH location
+RUN ln -s /usr/local/go/bin/go /usr/local/bin/go
```
a-ovchinnikov: The removed section was actually a part of the test: https://github.com/hermetoproject/hermeto/blob/main/tests/integration/test_multiple.py#L17

The way it was written the test indirectly verified RPM too.

eskultety: Ah, I already forgot that the RPM bits were part of the multiple backend testing, IOW get the toolchains necessary for building from the platform packager in a hermetic manner, makes sense. Maybe we should make it clear in the test scenario though to prevent future updates (or suggestions) like this one :) .

tests/integration/test_data/cargo_e2e/container/Containerfile
ST3V1K: It is about the older versions of runtime failing on Alpine (eg. rust 1.88-alpine) not the old non-alpine tests. I wanted to mention why specifically these were updated and not others.

tests/integration/test_data/cargo_e2e/container/Containerfile
ST3V1K: They are errors that were thrown while trying to run the tests. It's there to show why those specific images were not moved to Alpine

slimreaper35 (approved):

tests/integration/test_data/cargo_e2e/container/Containerfile
eskultety: > It is about the older versions of runtime failing on Alpine (eg. rust 1.88-alpine) not the old non-alpine tests. I wanted to mention why specifically these were updated and not others.

I see, but due to the wording it currently reads is as if we ran Alpine on the mentioned images before this change and some tests were already failing we're fixing that problem. Instead, I think you want to emphasize that while some images require no additional changes and some can't even be switched to Alpine (you already mention this correctly) some images require further runtime adjustments because otherwise some of the tests would not pass for various reasons (see the subtle difference? :) ).

In any case, the commit message needs updating anyway, because

> multiple_gomod_and_npm: golang and node (Alpine)

is no longer true in the latest revision after addressing some of the review comments.

tests/integration/test_data/cargo_e2e/container/Containerfile
a-ovchinnikov, line 1: I am a bit late to the party, but will provide feedback anyway:
_Commit message_: It is still a bit unclear what those errors mean. Right now you know that these paragraphs are messages for errors which you encountered while working on an issues. In half a year from now you will likely forget most of this context. Moreover, no other contributor will have this context. That is why it is important to capture it as precisely as practical. In the case of this commit message I would have suggested adding either a sentence "The list of encountered errors could be found below" or at least by means of punctuation:
```diff
- Rest of the image could not be bumped due to various errors.
+ Rest of the image could not be bumped due to various errors:
```
Note how a colon connects two parts of the message. 

This comment is not actionable, it is just something to keep in mind when writing commit messages: they are for capturing context and explaining (succinctly when possible) to others why this change is important. This is not a big deal in this particular commit, but this is a good habit in general.

PR #1283, opened 2026-02-09
Changes to integration test enviroment variables

ST3V1K:
- **Namespacing**: HERMETO_TEST_* (and prefixed ports) avoid clashes with other tools and future Hermeto runtime env vars and make it clear these are test-only.
- **Booleans**: Using "1" instead of "true" is simpler and more portable across shells and CI.
- **Single configuration path**: A top-level tests/conftest.py defines pytest options (with env fallbacks) and syncs them into os.environ, so existing os.getenv() code works unchanged and supports both pytest --option and env vars.

Closes https://github.com/hermetoproject/hermeto/issues/1238

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -266,12 +266,25 @@ nox -s all-integration-tests
 Note: while developing, you can run the PyPI server with `tests/pypiserver/start.sh &`
 and DNF server with `tests/dnfserver/start.sh &` to speed up the tests.
 
-To run integration-tests with custom image, specify the HERMETO\_IMAGE environment variable. Examples:
+To run integration-tests with custom image, specify the HERMETO\_TEST\_IMAGE environment variable or use the --hermeto-image option. Examples:
 
 ```shell
-HERMETO_IMAGE=ghcr.io/hermetoproject/hermeto:{tag} nox -s integration-tests
-HERMETO_IMAGE=localhost/hermeto:latest nox -s  integration-tests
+HERMETO_TEST_IMAGE=ghcr.io/hermetoproject/hermeto:{tag} nox -s integration-tests
+HERMETO_TEST_IMAGE=localhost/hermeto:latest nox -s integration-tests
+
+nox -s integration-tests -- --hermeto-image=ghcr.io/hermetoproject/hermeto:{tag}
+nox -s integration-tests -- --hermeto-image=localhost/hermeto:latest
 ```
+
+Another way to run integration tests is directly via `pytest`. This approach can be helpful for debugging and gives you more control over which tests are run and how they are executed. Examples:
+
+```shell
+HERMETO_TEST_IMAGE=localhost/hermeto:latest pytest [path/to/integration/test]
+
+pytest --hermeto-image=localhost/hermeto:latest [path/to/integration/test]
+```
+
+All hermeto integration test CLI options as well as the corresponding environment variables can be listed with `pytest --help`.
 
 ### Adding new dependencies to the project
 


tests/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+
+from tests.integration.utils import DEFAULT_INTEGRATION_REPO
+
+
+def pytest_addoption(parser: pytest.Parser) -> None:
+    """Register custom CLI options for Hermeto integration tests."""
+    group = parser.getgroup("hermeto integration", "hermeto integration test options")
+    group.addoption(
+        "--integration-repo",
```
a-ovchinnikov: How are these new options should be used with nox?

ST3V1K: They can be passed at the end of the nox command after -- if I am not missing something
`nox -s integration-tests -- --local-pypiserver`

tests/integration/utils.py
```diff
@@ -27,6 +27,8 @@
 # force IPv4 localhost as 'localhost' can resolve with IPv6 as well
 TEST_SERVER_LOCALHOST = "127.0.0.1"
 
+DEFAULT_INTEGRATION_REPO = "https://github.com/hermetoproject/integration-tests.git"
```
slimreaper35:
```suggestion
DEFAULT_INTEGRATION_TESTS_REPO = "https://github.com/hermetoproject/integration-tests.git"
```

tests/integration/conftest.py
```diff
@@ -10,18 +10,40 @@
...
+from tests.integration.utils import DEFAULT_INTEGRATION_REPO, TEST_SERVER_LOCALHOST
 
 from . import utils
 
 log = logging.getLogger(__name__)
 
 
+def pytest_configure(config: pytest.Config) -> None:
+    """Sync CLI option values to env so existing os.getenv() code sees them."""
+    os.environ["HERMETO_TEST_INTEGRATION_TESTS_REPO"] = config.getoption("--integration-repo")
```
slimreaper35:
```suggestion
    os.environ["HERMETO_TEST_INTEGRATION_TESTS_REPO"] = config.getoption("--integration-tests-repo")
```

same as above, probably more accurate than "integration repo"

tests/conftest.py
```diff
@@ -0,0 +1,71 @@
+import os
```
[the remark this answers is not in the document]
slimreaper35: These license headers are present in just a few random Python modules.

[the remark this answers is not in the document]
a-ovchinnikov: We should either add the license information in one fell swoop to all files or just ignore this rule, having some files conform to it while others don't does not make much sense.

tests/integration/conftest.py
a-ovchinnikov, line 1: _Commit message_ for https://github.com/hermetoproject/hermeto/commit/2638b988b57cd78650a1402e1fd6da707adc0b21: 2s/incosistant/inconsistent/

Also are you sure that options _control_ env variables? WDYT about saying that they _map_ to environment variables?

.github/workflows/gating.yaml
a-ovchinnikov, line 1: _Commit message_ for https://github.com/hermetoproject/hermeto/commit/decfff2dc9c19b6de6a3604a327dcdc3e175def0: 1,2s/enviroment/environment/  also 3s/assosiated/associated/
(TBH I don't think naming inconsistency is that big of a deal in this particular case, but this change won't hurt).

tests/conftest.py
a-ovchinnikov: Thank you. Can available variables be listed somehow?

tests/conftest.py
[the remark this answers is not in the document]
eskultety: This is a new file, so including it makes sense. The fact we don't have it consistent is a different story, but that's for a follow-up, this is totally harmless and I'd argue the current industry standard, so for a new file, totally fine and let's resolve the inconsistency elsewhere :) -> resolving.

tests/conftest.py
eskultety: They're now listed in the pytest's help, other than that I'd say is a larger follow-up, https://github.com/hermetoproject/hermeto/issues/1238 didn't require this.

tests/integration/conftest.py
eskultety: Yes, it's confusing, the logic doesn't even follow that statement as it shouldn't due to the order of precedence!

tests/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+    group.addoption(
+        "--run-all-integration",
+        action="store_true",
+        default=os.getenv("HERMETO_RUN_ALL_INTEGRATION_TESTS") == "1",
+        help="Run all integration tests, disable skip-by-changed-files (env: HERMETO_RUN_ALL_INTEGRATION_TESTS=1)",
+    )
+    group.addoption(
+        "--container-engine",
+        action="store",
+        default=os.getenv("HERMETO_TEST_CONTAINER_ENGINE", "podman").lower(),
```
eskultety: do we even need the `lower` part if we define `choices`?

tests/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+import os
+
+import pytest
+
+from tests.integration.utils import DEFAULT_INTEGRATION_TESTS_REPO
+
+
+def pytest_addoption(parser: pytest.Parser) -> None:
+    """Register custom CLI options for Hermeto integration tests."""
+    group = parser.getgroup("hermeto integration", "hermeto integration test options")
```
eskultety, line 11: I like this approach, pytest has autocomplete enabled.

tests/integration/conftest.py
```diff
@@ -10,18 +10,40 @@
...
 log = logging.getLogger(__name__)
 
 
+def pytest_configure(config: pytest.Config) -> None:
+    """Sync CLI option values to env so existing os.getenv() code sees them."""
+    os.environ["HERMETO_TEST_INTEGRATION_TESTS_REPO"] = config.getoption("--integration-tests-repo")
+    os.environ["HERMETO_TEST_IMAGE"] = config.getoption("--hermeto-image")
+    os.environ["HERMETO_TEST_LOCAL_PYPISERVER"] = (
+        "1" if config.getoption("--local-pypiserver") else "0"
+    )
```
eskultety: Seeing this pattern, what do you think of a small adjustment:
```diff
diff --git tests/integration/conftest.py tests/integration/conftest.py
index bf2253782..f842581f5 100644
--- tests/integration/conftest.py
+++ tests/integration/conftest.py
@@ -18,26 +18,31 @@ from . import utils
 log = logging.getLogger(__name__)


+_ENV_VAR_CLI_MAP = [
+    ("HERMETO_TEST_INTEGRATION_TESTS_REPO", "--integration-tests-repo"),
+    ("HERMETO_TEST_IMAGE", "--hermeto-image"),
+    ("HERMETO_TEST_LOCAL_PYPISERVER", "--local-pypiserver"),
+    ("PYPISERVER_PORT", "--pypiserver-port"),
+    ("HERMETO_TEST_LOCAL_DNF_SERVER", "--local-dnf-server"),
+    ("DNFSERVER_SSL_PORT", "--dnfserver-ssl-port"),
+    ("HERMETO_TEST_NETRC_CONTENT", "--netrc-content"),
+    ("HERMETO_GENERATE_TEST_DATA", "--generate-test-data"),
+    ("HERMETO_RUN_ALL_INTEGRATION_TESTS", "--run-all-integration"),
+    ("HERMETO_TEST_CONTAINER_ENGINE", "--container-engine"),
+]
+
+
 def pytest_configure(config: pytest.Config) -> None:
     """Sync CLI option values to env so existing os.getenv() code sees them."""
-    os.environ["HERMETO_TEST_INTEGRATION_TESTS_REPO"] = config.getoption("--integration-tests-repo")
-    os.environ["HERMETO_TEST_IMAGE"] = config.getoption("--hermeto-image")
-    os.environ["HERMETO_TEST_LOCAL_PYPISERVER"] = (
-        "1" if config.getoption("--local-pypiserver") else "0"
-    )
-    os.environ["PYPISERVER_PORT"] = config.getoption("--pypiserver-port")
-    os.environ["HERMETO_TEST_LOCAL_DNF_SERVER"] = (
-        "1" if config.getoption("--local-dnf-server") else "0"
-    )
-    os.environ["DNFSERVER_SSL_PORT"] = config.getoption("--dnfserver-ssl-port")
-    os.environ["HERMETO_TEST_NETRC_CONTENT"] = config.getoption("--netrc-content")
-    os.environ["HERMETO_GENERATE_TEST_DATA"] = (
-        "1" if config.getoption("--generate-test-data") else "0"
-    )
-    os.environ["HERMETO_RUN_ALL_INTEGRATION_TESTS"] = (
-        "1" if config.getoption("--run-all-integration") else "0"
-    )
-    os.environ["HERMETO_TEST_CONTAINER_ENGINE"] = config.getoption("--container-engine").lower()
+
+    def env_value(cli_opt: str) -> str:
+        value = config.getoption(cli_opt)
+        if isinstance(value, bool):
+            return "1" if value else "0"
+        return value
+
+    for env_var, cli_opt in _ENV_VAR_CLI_MAP:
+        os.environ[env_var] = env_value(cli_opt)
```

tests/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+        action="store",
+        default=os.getenv("HERMETO_TEST_NETRC_CONTENT", ""),
+        help=".netrc content for private PyPI etc. (env: HERMETO_TEST_NETRC_CONTENT)",
+    )
+    group.addoption(
+        "--generate-test-data",
+        action="store_true",
+        default=os.getenv("HERMETO_GENERATE_TEST_DATA") == "1",
+        help="Regenerate expected test data files (env: HERMETO_GENERATE_TEST_DATA=1)",
+    )
```
eskultety, line 59: Do we need to mention `env: HERMETO_TEST_` in the help string for `store` actions? This is what the CLI output looks like:
```
--hermeto-image=HERMETO_IMAGE
                        Hermeto container image reference; build local image if not set (env: HERMETO_TEST_IMAGE)
  --local-pypiserver    Start local pypiserver for pip tests (env: HERMETO_TEST_LOCAL_PYPISERVER=1)
  --pypiserver-port=PYPISERVER_PORT
                        Port for local pypiserver (env: HERMETO_TEST_PYPISERVER_PORT)
  --local-dnf-server    Start local DNF server for RPM tests (env: HERMETO_TEST_LOCAL_DNF_SERVER=1)
```

tests/dnfserver/start.sh
```diff
@@ -7,7 +7,7 @@ DEFAULT_HTTP_PORT=8081
 TEST_RPM="https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/x86_64/baseos/os/Packages/r/redhat-release-9.4-0.5.el9.x86_64.rpm"
 
 HTTP_PORT=${DNFSERVER_HTTP_PORT:-$DEFAULT_HTTP_PORT}
-SSL_PORT=${DNFSERVER_SSL_PORT:-$DEFAULT_SSL_PORT}
+SSL_PORT=${HERMETO_TEST_DNFSERVER_SSL_PORT:-$DEFAULT_SSL_PORT}
```
eskultety, line 10: You'll need to perform the same conversion for the CLI options OR just introduce them with the right name in the respective commit.

eskultety (commented):
The approach LGTM, needs some more tweaks. Note https://github.com/hermetoproject/hermeto/issues/1238 is linked from within a commit, but not on the PR description level, so the issue won't get automatically closed when the PR is merged....this is a GH issue.

eskultety:
Hit send too soon, this change deserves documenting as well, have a look at https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md#running-integration-tests

tests/integration/conftest.py
ST3V1K: changed it to _map to_

tests/conftest.py
ST3V1K: Removed the `lower`. I was not sure about the default option following `choices`.

tests/integration/conftest.py
ST3V1K: done

tests/conftest.py
ST3V1K: The `env: HERMETO_TEST_` refers to the environment variable, while the name in `store` action only refers to the option name and does not have to match with the environment variable.

eskultety: That's an overlook on my side -> resolved.

tests/dnfserver/start.sh
ST3V1K: done

tests/integration/conftest.py
a-ovchinnikov, line 1: _Commit message_ in https://github.com/hermetoproject/hermeto/commit/de9068269b900c7f75e47f2f915db90000805818: 3s/enviroment/environment/

CONTRIBUTING.md
```diff
@@ -266,13 +266,15 @@ nox -s all-integration-tests
 Note: while developing, you can run the PyPI server with `tests/pypiserver/start.sh &`
 and DNF server with `tests/dnfserver/start.sh &` to speed up the tests.
 
-To run integration-tests with custom image, specify the HERMETO\_IMAGE environment variable. Examples:
```
eskultety, line 269: Depending on the exact nature of changes, docs update may or may not be paired with code logic. However, CONTRIBUTING.md isn't part of our docs, so the update should definitely go to a standalone commit.

CONTRIBUTING.md
```diff
@@ -266,13 +266,15 @@ nox -s all-integration-tests
...
+To run integration-tests with custom image, use the --hermeto-image option. Examples:
 
 ```shell
-HERMETO_IMAGE=ghcr.io/hermetoproject/hermeto:{tag} nox -s integration-tests
-HERMETO_IMAGE=localhost/hermeto:latest nox -s  integration-tests
+nox -s integration-tests -- --hermeto-image=ghcr.io/hermetoproject/hermeto:{tag}
+nox -s integration-tests -- --hermeto-image=localhost/hermeto:latest
```
eskultety, line 276: Functionally correct, but environment variables are still available for use, so this update needs to be enhanced to cover both using the new CLI opts as well as environment variables. This is a matter of convenience and we should not force a particular way of running tests upon a developer.

On the same note I'd even argue we should expand this paragraph to also mention an explicit invocation of `pytest` (since you even mention it in the next hunk), because `nox` is just a convenience sugar when one wants to run everything, but once you want scope what's actually run and debug that particular thing, nox is simply not a way to do that.

CONTRIBUTING.md
```diff
@@ -266,13 +266,15 @@ nox -s all-integration-tests
...
 ```
 
+All hermeto integration test options can be listed with `pytest --help`.
```
eskultety: s/test options/test CLI options as well as the corresponding environment variables/

CONTRIBUTING.md
ST3V1K: Done

CONTRIBUTING.md
ST3V1K: Added info about all the ways to run them.

CONTRIBUTING.md
ST3V1K: Done

CONTRIBUTING.md
```diff
@@ -266,13 +266,26 @@ nox -s all-integration-tests
 Note: while developing, you can run the PyPI server with `tests/pypiserver/start.sh &`
 and DNF server with `tests/dnfserver/start.sh &` to speed up the tests.
 
-To run integration-tests with custom image, specify the HERMETO\_IMAGE environment variable. Examples:
+To run integration-tests with custom image, specify the HERMETO\_TEST\_IMAGE environment variable or use the --hermeto-image option. Examples:
 
 ```shell
 HERMETO_IMAGE=ghcr.io/hermetoproject/hermeto:{tag} nox -s integration-tests
-HERMETO_IMAGE=localhost/hermeto:latest nox -s  integration-tests
+HERMETO_IMAGE=localhost/hermeto:latest nox -s integration-tests
```
eskultety: I believe this is now called `HERMETO_TEST_IMAGE` :) .

CONTRIBUTING.md
```diff
@@ -266,13 +266,26 @@ nox -s all-integration-tests
...
+
+nox -s integration-tests -- --hermeto-image=ghcr.io/hermetoproject/hermeto:{tag}
+nox -s integration-tests -- --hermeto-image=localhost/hermeto:latest
 ```
 
+Another way to run integration tests is directly via `pytest`. This approach can be helpful for debugging and gives you more control over which tests are run and how they are executed. Examples:
+
+```shell
+HERMETO_IMAGE=localhost/hermeto:latest pytest [path/to/integration/test]
```
eskultety: ...same here...

CONTRIBUTING.md
ST3V1K: Sorry about that, now it should be correct.

PR #1285, opened 2026-02-09
refactor(pip): introduce CommonArtifact dataclass

### LINKED ISSUE #1150: Introduce data classes for pip to replace "download info"
Currently, the pip package manager code passes around untyped `dict[str, Any]` objects called `download_info(s)` throughout various functions. This approach lacks type safety, making it easy to misspell keys or forget required fields. Introducing typed data classes would provide improved refactoring safety and self-documenting code.

ben-alkov:
Refactors Pip artifact handling to use typed dataclasses, improving code clarity

Closes [#1150](https://github.com/hermetoproject/hermeto/issues/1150)

Why:

Enhances type safety, readability, and maintainability across the Pip package manager's core logic. Currently we have an untyped, `dict[str, Any]` ("download_info"), which is never explicitly defined and is passed throughout the artifact handling. This is exceedingly difficult to work with if one has to dig deeply into the pip plumbing.

What:

Replaces untyped `dict[str, Any]` with a hierarchy of ArtifactDownload dataclasses (PyPIArtifact, VCSArtifact, URLArtifact) for representing downloaded package information. This massively simplifies many parts of the artifact-handling code.

Side-effects:

- public-facing `_resolve_pip` function maintains its dictionary-based return type
- any direct function calls internal to "hermeto/core/package_managers/pip/main.py" which previously expected `dict[str, Any]` would be effected by the new ArtifactDownload return (none that I'm aware of)
- However, further refactoring could (should) be done elsewhere (e.g. "hermeto/core/package_managers/pip/rust.py" in `filter_packages_with_rust_code()` whose `packages` arg still takes `list[dict[str, Any]]` as its type - in particular, this should be an easy fix)

How:

- New Data Models: Introduced "hermeto/core/package_managers/pip/models.py" defining `CommonArtifact` (base), `PyPIArtifact`, `VCSArtifact`, and `URLArtifact` dataclasses. These are Frozen dataclasses
- Type Signature Updates: Functions like `_process_req`, `_process_pypi_req`, `_process_vcs_req`, `_process_url_req`, `_download_dependencies`, and `_download_from_requirement_files` in "hermeto/core/package_managers/pip/main.py" now explicitly return ArtifactDownload types (or lists thereof) instead of `dict[str, Any]`
- Object Instantiation: Inside `_process_req`, dictionary manipulation has been replaced with the instantiation of the appropriate ArtifactDownload subclass based on the requirement kind
- Build Dependency Marking: Build dependencies are now marked using dataclasses.replace on immutable ArtifactDownload instances
- Output Reversion: The `_resolve_pip` function maps the properties of ArtifactDownload objects back to the original dictionary structure for its final return value, preserving external API consistency.
- Testing: New unit tests for the artifact models ("test_models.py") and updates to existing tests in "test_main.py" to reflect the new object structures are included.

Assisted-by: Claude <noreply@anthropic.com>
Signed-off-by: Ben Alkov <ben.alkov@redhat.com>

### DIFF
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
+
 import asyncio
+import dataclasses
 import logging
 import tarfile
 import zipfile
@@ -23,6 +25,12 @@ from hermeto.core.package_managers.general import (
     async_download_files,
     download_binary_file,
     extract_git_info,
+)
+from hermeto.core.package_managers.pip.models import (
+    Artifact,
+    PyPIArtifact,
+    URLArtifact,
+    VCSArtifact,
 )
 from hermeto.core.package_managers.pip.package_distributions import (
     DistributionPackageInfo,
@@ -260,11 +268,14 @@ def _process_req(
     pip_deps_dir: RootedPath,
     download_info: dict[str, Any],
     dpi: DistributionPackageInfo | None = None,
-) -> dict[str, Any]:
-    download_info["kind"] = req.kind
-    download_info["requirement_file"] = str(requirements_file.file_path.subpath_from_root)
-    download_info["missing_req_file_checksum"] = True
-    download_info["package_type"] = ""
+    build_dependency: bool = False,
+) -> Artifact:
+    """Process a downloaded PipRequirement and return an Artifact."""
+    # Extract common fields from download_info dict
+    package_name = download_info["package"]
+    path = download_info["path"]
+    requirement_file = str(requirements_file.file_path.subpath_from_root)
+    missing_req_file_checksum = True
 
     def _checksum_must_match_or_path_unlink(
         path: Path, checksum_info: Iterable[ChecksumInfo]
@@ -276,34 +287,67 @@ def _process_req(
             path.unlink()
             log.warning("Download '%s' was removed from the output directory", path.name)
 
+    artifact: Artifact
     if dpi:
+        # PyPI artifact
         if dpi.req_file_checksums:
-            download_info["missing_req_file_checksum"] = False
+            missing_req_file_checksum = False
         if dpi.checksums_to_match:
             _checksum_must_match_or_path_unlink(dpi.path, dpi.checksums_to_match)
         if dpi.package_type == "sdist":
             _check_metadata_in_sdist(dpi.path)
-        download_info["package_type"] = dpi.package_type
-        download_info["index_url"] = dpi.index_url
+
+        artifact = PyPIArtifact(
+            package=package_name,
+            path=path,
+            requirement_file=requirement_file,
+            missing_req_file_checksum=missing_req_file_checksum,
+            build_dependency=build_dependency,
+            index_url=dpi.index_url,
+            package_type=dpi.package_type,
+            _version=dpi.version,
+        )
     elif req.kind == "vcs":
-        # `missing_req_file_checksum` is *always* True for VCS deps
-        pass
+        # VCS artifact - extract git info from download_info
+        artifact = VCSArtifact(
+            package=package_name,
+            path=path,
+            requirement_file=requirement_file,
+            missing_req_file_checksum=True,  # Always True for VCS deps
+            build_dependency=build_dependency,
+            url=download_info["url"],
+            host=download_info["host"],
+            namespace=download_info["namespace"],
+            repo=download_info["repo"],
+            ref=download_info["ref"],
+        )
+    elif req.kind == "url":
+        # URL artifact
+        hashes = req.hashes or [req.qualifiers.get("cachito_hash", "")]
+        if hashes:
+            missing_req_file_checksum = False
+            _checksum_must_match_or_path_unlink(path, list(map(ChecksumInfo.from_hash, hashes)))
+
+        artifact = URLArtifact(
+            package=package_name,
+            path=path,
+            requirement_file=requirement_file,
+            missing_req_file_checksum=missing_req_file_checksum,
+            build_dependency=build_dependency,
+            original_url=download_info["original_url"],
+            url_with_hash=download_info["url_with_hash"],
+            package_type=download_info.get("package_type"),
+        )
     else:
-        if req.kind == "url":
-            hashes = req.hashes
-            if hashes:
-                download_info["missing_req_file_checksum"] = False
-                _checksum_must_match_or_path_unlink(
-                    download_info["path"], list(map(ChecksumInfo.from_hash, hashes))
-                )
+        raise ValueError(f"Unknown requirement kind: {req.kind}")
 
     log.debug(
         "Successfully processed '%s' in path '%s'",
         req.download_line,
-        download_info["path"].relative_to(pip_deps_dir.root),
+        path.relative_to(pip_deps_dir.root),
     )
 
-    return download_info
+    return artifact
 
 
 def _process_pypi_req(
@@ -312,18 +356,23 @@ def _process_pypi_req(
     index_url: str,
     pip_deps_dir: RootedPath,
     binary_filters: PipBinaryFilters | None = None,
-) -> list[dict[str, Any]]:
-    download_infos: list[dict[str, Any]] = []
+) -> list[Artifact]:
+    """Process a PipRequirement, getting DPIs, then download artifacts.
+
+    Artifacts are filtered and allowed by type. Uses `aiohttp_retry.RetryClient`.
+    """
+    download_artifacts: list[Artifact] = []
 
     artifacts: list[DistributionPackageInfo] = process_package_distributions(
         req, pip_deps_dir, binary_filters, index_url
     )
 
     files: dict[str, StrPath] = {dpi.url: dpi.path for dpi in artifacts if not dpi.path.exists()}
+
     asyncio.run(async_download_files(files, get_config().runtime.concurrency_limit))
 
     for artifact in artifacts:
-        download_infos.append(
+        download_artifacts.append(
             _process_req(
                 req,
                 requirements_file,
@@ -333,12 +382,10 @@ def _process_pypi_req(
             )
         )
 
-    return download_infos
+    return download_artifacts
 
 
-def _process_vcs_req(
-    req: PipRequirement, pip_deps_dir: RootedPath, **kwargs: Any
-) -> dict[str, Any]:
+def _process_vcs_req(req: PipRequirement, pip_deps_dir: RootedPath, **kwargs: Any) -> Artifact:
     return _process_req(
         req,
         pip_deps_dir=pip_deps_dir,
@@ -349,37 +396,36 @@ def _process_vcs_req(
 
 def _process_url_req(
     req: PipRequirement, pip_deps_dir: RootedPath, trusted_hosts: set[str], **kwargs: Any
-) -> dict[str, Any]:
-    result = _process_req(
+) -> Artifact:
+    download_info = _download_url_package(req, pip_deps_dir, trusted_hosts)
+    if req.url.endswith(WHEEL_FILE_EXTENSION):
+        download_info["package_type"] = "wheel"
+
+    return _process_req(
         req,
         pip_deps_dir=pip_deps_dir,
-        download_info=_download_url_package(req, pip_deps_dir, trusted_hosts),
+        download_info=download_info,
         **kwargs,
     )
-    if req.url.endswith(WHEEL_FILE_EXTENSION):
-        result["package_type"] = "wheel"
-
-    return result
 
 
 def _download_dependencies(
     output_dir: RootedPath,
     requirements_file: PipRequirementsFile,
     binary_filters: PipBinaryFilters | None = None,
-) -> list[dict[str, Any]]:
+) -> list[Artifact]:
     """
     Download artifacts of all dependency packages in a requirements.txt file.
 
     :param output_dir: the root output directory for this request
     :param requirements_file: A requirements.txt file
     :param binary_filters: process wheels?
-    :return: Info about downloaded packages; all items will contain "kind" and "path" keys
-        (and more based on kind, see _download_*_package functions for more details)
-    :rtype: list[dict]
+    :return: List of artifact download objects
+    :rtype: list[Artifact]
     """
     options: dict[str, Any] = process_requirements_options(requirements_file.options)
     trusted_hosts = set(options["trusted_hosts"])
-    processed: list[dict[str, Any]] = []
+    processed: list[Artifact] = []
 
     if options["require_hashes"]:
         log.info("Global --require-hashes option used, will require hashes")
@@ -402,29 +448,29 @@ def _download_dependencies(
     for req in requirements_file.requirements:
         log.info("-- Processing requirement line '%s'", req.download_line)
         if req.kind == "pypi":
-            download_infos: list[dict[str, Any]] = _process_pypi_req(
+            download_artifacts: list[Artifact] = _process_pypi_req(
                 req,
                 requirements_file=requirements_file,
                 index_url=options["index_url"] or pypi_simple.PYPI_SIMPLE_ENDPOINT,
                 pip_deps_dir=pip_deps_dir,
                 binary_filters=binary_filters,
             )
-            processed.extend(download_infos)
+            processed.extend(download_artifacts)
         elif req.kind == "vcs":
-            download_info = _process_vcs_req(
+            artifact = _process_vcs_req(
                 req,
                 requirements_file=requirements_file,
                 pip_deps_dir=pip_deps_dir,
             )
-            processed.append(download_info)
+            processed.append(artifact)
         elif req.kind == "url":
-            download_info = _process_url_req(
+            artifact = _process_url_req(
                 req,
                 requirements_file=requirements_file,
                 pip_deps_dir=pip_deps_dir,
                 trusted_hosts=trusted_hosts,
             )
-            processed.append(download_info)
+            processed.append(artifact)
         else:
             # Should not happen
             raise RuntimeError(f"Unexpected requirement kind: '{req.kind!r}'")
@@ -436,7 +482,7 @@ def _download_dependencies(
 
 def _download_vcs_package(requirement: PipRequirement, pip_deps_dir: RootedPath) -> dict[str, Any]:
     """
-    Fetch the source for a Python package from VCS (only git is supported).
+    Clone the source for a Python package from a VCS to a local tarball. Only "git" is supported.
 
     :param PipRequirement requirement: VCS requirement from a requirements.txt file
     :param RootedPath pip_deps_dir: The deps/pip directory in an application request bundle
@@ -461,7 +507,7 @@ def _download_url_package(
     requirement: PipRequirement, pip_deps_dir: RootedPath, trusted_hosts: set[str]
 ) -> dict[str, Any]:
     """
-    Download a Python package from a URL.
+    Download a Python package from a URL (uses `requests`).
 
     :param PipRequirement requirement: URL requirement from a requirements.txt file
     :param RootedPath pip_deps_dir: The deps/pip directory in an application request bundle
@@ -497,18 +543,17 @@ def _download_from_requirement_files(
     output_dir: RootedPath,
     files: list[RootedPath],
     binary_filters: PipBinaryFilters | None = None,
-) -> list[dict[str, Any]]:
+) -> list[Artifact]:
     """
     Download dependencies listed in the requirement files.
 
     :param output_dir: the root output directory for this request
     :param files: list of absolute paths to pip requirements files
     :param binary_filters: process wheels?
-    :return: Info about downloaded packages; see download_dependencies return docs for further
-        reference
+    :return: List of artifact download objects
     :raises PackageRejected: If requirement file does not exist
     """
-    requirements: list[dict[str, Any]] = []
+    requirements: list[Artifact] = []
     for req_file in files:
         if not req_file.path.exists():
             raise LockfileNotFound(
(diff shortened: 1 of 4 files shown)


hermeto/core/package_managers/pip/models.py
```diff
@@ -0,0 +1,114 @@
...
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Literal, Union
+
+
+@dataclass(frozen=True)
+class CommonArtifact:
+    """Base class for representing a packaged Python artifact.
+
+    This replaces the ad-hoc dict[str, Any] usage throughout the pip package manager
```
a-ovchinnikov: A nitpicky question: will the information about ad-hoc class stay relevant after the change is merged?

hermeto/core/package_managers/pip/models.py
```diff
@@ -0,0 +1,114 @@
...
+    kind: Literal["url"] = field(default="url", init=False)
+
+    @property
+    def version(self) -> str:
+        """Return version string for URL artifacts."""
+        return self.url_with_hash
+
+
+# Type alias for any artifact download
+ArtifactDownload = Union[PyPIArtifact, VCSArtifact, URLArtifact]
```
a-ovchinnikov: WDYT about renaming this to `ArtifactToDownload` or `DownloadableArtifact`?

tests/unit/package_managers/pip/test_models.py
```diff
@@ -0,0 +1,376 @@
...
+    URLArtifact,
+    VCSArtifact,
+)
+
+
+class TestPyPIArtifact:
+    """Test PyPIArtifact dataclass."""
+
+    def test_creation(self) -> None:
+        """Test basic PyPIArtifact creation."""
```
a-ovchinnikov: What is the purpose of this test? I would suspect that this behavior has already been extensively tested [here](https://github.com/python/cpython/blob/main/Lib/test/test_dataclasses/__init__.py).

tests/unit/package_managers/pip/test_models.py
```diff
@@ -0,0 +1,376 @@
...
+            url_with_hash="https://example.com/package.tar.gz#cachito_hash=sha256:abc123",
+        )
+
+        # Test that all types are accepted by the union
+        artifacts: list[ArtifactDownload] = [pypi_artifact, vcs_artifact, url_artifact]
+
+        assert len(artifacts) == 3
+        assert isinstance(artifacts[0], PyPIArtifact)
+        assert isinstance(artifacts[1], VCSArtifact)
+        assert isinstance(artifacts[2], URLArtifact)
```
a-ovchinnikov: Which scenario are these assertions for? Take L188. It asserts a seemingly trivial fact: that an object of type URLAtifact is an object of type URLArtifact -- `artifacts` contains a reference to URLArtifact in third position _by definition_ (i.e. L172 via L183). What could happen in a program flow (outside of external events like bit flips) that effectively a pointer to a struct suddenly becomes a pointer to a different structure without any manipulation of this pointer? I completely miss why this has to be checked.

tests/unit/package_managers/pip/test_models.py
```diff
@@ -0,0 +1,376 @@
...
+                requirement_file="requirements.txt",
+                missing_req_file_checksum=False,
+                build_dependency=False,
+                original_url="https://example.com/package.tar.gz",
+                url_with_hash="https://example.com/package.tar.gz#cachito_hash=sha256:abc123",
+            ),
+        ]
+
+    def test_all_artifacts_have_common_fields(self, artifacts: list[ArtifactDownload]) -> None:
+        """Test that all artifacts have the common fields from CommonArtifact."""
```
a-ovchinnikov: Artifacts inherit from CommonArtifact. This test looks like it verifies that CPython implements inheritance correctly.

tests/unit/package_managers/pip/test_models.py
```diff
@@ -0,0 +1,376 @@
...
+        assert pypi_artifact.kind == "pypi"
+        assert vcs_artifact.kind == "vcs"
+        assert url_artifact.kind == "url"
+
+    def test_build_dependency_values(self, artifacts: list[ArtifactDownload]) -> None:
+        """Test build_dependency field values."""
+        pypi_artifact, vcs_artifact, url_artifact = artifacts
+
+        assert pypi_artifact.build_dependency is False
+        assert vcs_artifact.build_dependency is True
```
a-ovchinnikov: Why? Is this a universal properly of vcs artifacts?

tests/unit/package_managers/pip/test_models.py
```diff
@@ -0,0 +1,376 @@
...
+            path=Path("/tmp/numpy-1.21.0.tar.gz"),
+            requirement_file="requirements.txt",
+            missing_req_file_checksum=False,
+            build_dependency=False,
+            index_url="https://pypi.org/simple/",
+            package_type="sdist",
+            version="1.21.0",
+        )
+        with pytest.raises(FrozenInstanceError):
+            artifact.build_dependency = True  # type: ignore[misc]
```
a-ovchinnikov: Isn't this tested [elsewhere](https://github.com/python/cpython/blob/main/Lib/test/test_dataclasses/__init__.py#L3055)?

tests/unit/package_managers/pip/test_models.py
```diff
@@ -0,0 +1,376 @@
...
+                url="",
+                host="github.com",
+                namespace="user",
+                repo="myproject",
+                ref="abc123",
+            )
+
+    def test_empty_ref_raises(self) -> None:
+        """Test that empty ref raises ValueError."""
+        with pytest.raises(ValueError, match="non-empty ref"):
```
a-ovchinnikov: Direct string comparisons in tests are an antipattern. They should be avoided and when present replaced with checking for specific exception type.

hermeto/core/package_managers/pip/models.py
ben-alkov: No

hermeto/core/package_managers/pip/models.py
ben-alkov: 🤔 Good point, probably unecessary

hermeto/core/package_managers/pip/models.py
ben-alkov: That was very terse, but let me just say that this was conceived of in mid-2024 (at the latest!), and the ONE thing I'm absolutely certain of by all criteria, after spending *hours* thinking about this[1], is that I chose the correct name.

[1] In fact, aside from the actual pain of working with that untyped dict, the next most painful canker was the proliferation of names, all referring to the same thing, but with no real logic to them or their hierarchy.

ben-alkov:
More work needed on tests, obviously.

hermeto/core/package_managers/pip/models.py
a-ovchinnikov: The reason I suggested a rename is in how I read `ArtifactDownload`. Here I see `Download` as the main component of the name and `Artifact` as a qualifier that narrows down the main component. Further, I read `Download` as a noun describing a process, like in `FileDownload` or `ContentDownload`, a process which results in some artifact being retrieved from some network location. In GoF parlance I would have called this a strategy or a command judging from the name alone. Problem is, this is not the case, this is a dataclass containing data necessary to perform a download, not an encapsulation of download action and I think the name misleads a reader. I could be wrong and would be very interested in learning the reasoning behind the chosen name (and I fully agree with namines proliferation part).

ben-alkov: `DownloadArtifact` is even *worse*; `Download` is a superposition of noun and verb, definitely indeterminate. `Artifact` (last I checked) is a noun only.

IIRC, I considered `DownloadedArtifact`, but it's too ugly to live.

Suggestions welcomed :shrug:

ben-alkov: WDYT, @a-ovchinnikov, did Claude Code just nail it?

"""
Suggestion: Rename ArtifactDownload → ResolvedArtifact
Rationale: Matches resolver.py terminology; "Download" implies action, not
data
"""

a-ovchinnikov: IDK, `Resolved` reads strange when attached to `Artifact`. Were there any gaps that had to be filled? No, we have just extracted a bit of data from a bigger piece of data. Is the `Artifact` determined to do something? Likely not. I would like to learn what's wrong with `ArtifactToDownload`. There the stress in on `Artifact`, which we are about to download (the name conveys that too). Am I missing something or reading it wrong? (For the record, I won't block on this, naming is hard, especially here.)

ben-alkov: OK, I put on my big-boy pants and put in the effort to review/remember *exactly* how this works.

You can't simply look at the data model, you have to look at how/where it's *used*.

It CAN'T be an `Artifact*To*Download`, because at the point where it's instantiated... it's already been downloaded (simplest example is VCS):

```python
def _process_vcs_req(
    req: PipRequirement, pip_deps_dir: RootedPath, **kwargs: Any
) -> ArtifactDownload:
    return _process_req(
        req,
        pip_deps_dir=pip_deps_dir,
        download_info=_download_vcs_package(req, pip_deps_dir),  # <== Download occurs here
        **kwargs,
    )
```

The best description I can come up with is "Metadata of/about a *downloaded* artifact" (literally true; by the type any `CommonArtifact` type is instantiated, the artifact is already on the local filesystem).

Now I have a headache :grin:

P.S. And as I said above: "IIRC, I considered DownloadedArtifact, but it's too ugly to live" (also, it's not accurate, as we're NOT working with the downloaded artifact - we're working with data *about* it)

ben-alkov: I mean... They could just be `Artifacts`... :shrug:

a-ovchinnikov: > They could just be Artifacts...

Yep, this one is definitely better!

a-ovchinnikov: Actually it should be just `Artifact`, `s` implies iteration. This type is a union of other types i.e. an Artifact is either PyPIArtifact, or VCSArtifact or URLArtifact.

tests/unit/package_managers/pip/test_models.py
```diff
@@ -0,0 +1,96 @@
...
+
+from hermeto.core.package_managers.pip.models import (
+    Artifact,
+    PyPIArtifact,
+    URLArtifact,
+    VCSArtifact,
+)
+
+
+class TestVCSArtifact:
```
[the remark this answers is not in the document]
ben-alkov: Do not write tests which compare log messages.

hermeto/core/package_managers/pip/main.py
```diff
@@ -606,40 +648,41 @@ def resolve_req_files(req_files: list[Path] | None, devel: bool) -> list[RootedP
...
-        for dep in (requires + build_requires)
-    ]
+        if isinstance(artifact, PyPIArtifact):
+            dep["index_url"] = artifact.index_url
+            dep["package_type"] = artifact.package_type
+        elif isinstance(artifact, URLArtifact):
+            dep["package_type"] = artifact.package_type
+        else:
+            # VCS artifacts have no package_type or index_url
+            dep["package_type"] = None
```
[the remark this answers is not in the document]
ben-alkov: Explicit is better than implicit.

Should `VCSArtifact` ever gain `package_type`, this would become a bug.

Yes, unlikely, but many unlikely things happen every day.

hermeto/core/package_managers/pip/main.py
```diff
@@ -315,18 +358,20 @@ def _process_pypi_req(
...
+) -> list[Artifact]:
+    download_artifacts: list[Artifact] = []
 
     artifacts: list[DistributionPackageInfo] = process_package_distributions(
         req, pip_deps_dir, binary_filters, index_url
     )
 
     files: dict[str, StrPath] = {dpi.url: dpi.path for dpi in artifacts if not dpi.path.exists()}
+
+    # Do the actual download (aiohttp)
```
a-ovchinnikov: My usual nitpick here: "what"-style comments are a sign of a well-named function being needed.

hermeto/core/package_managers/pip/main.py
```diff
@@ -315,18 +358,20 @@ def _process_pypi_req(
     index_url: str,
     pip_deps_dir: RootedPath,
     binary_filters: PipBinaryFilters | None = None,
-) -> list[dict[str, Any]]:
-    download_infos: list[dict[str, Any]] = []
+) -> list[Artifact]:
```
a-ovchinnikov, line 359: This reads so much better now!

hermeto/core/package_managers/pip/main.py
```diff
@@ -455,6 +497,7 @@ def _download_vcs_package(requirement: PipRequirement, pip_deps_dir: RootedPath)
     download_to = pip_deps_dir.join_within_root(_get_external_requirement_filepath(requirement))
     download_to.path.parent.mkdir(exist_ok=True, parents=True)
 
+    # Do the actual download (git clone)
```
a-ovchinnikov: L500 and L535 essentially duplicate lines that follow. Any IDE and any serious enough editor can provide this context at a click/command. I strongly recommend removing these comments.

hermeto/core/package_managers/pip/main.py
a-ovchinnikov: Explicit type checks make me a bit uneasy. WDYT about handling this with getattr?
```python
dep["package_type"] = getattr(artifact, "package_type", None)
if (index_url:=getattr(artifact, "index_url", None)) is not None:
    dep["index_url"] = index_url
```

hermeto/core/package_managers/pip/models.py
```diff
@@ -0,0 +1,126 @@
...
+from pathlib import Path
+from typing import Any, Literal, Union
+
+
+@dataclass(frozen=True)
+class CommonArtifact(ABC):
+    """Base class for representing a packaged Python artifact."""
+
+    package: str
+    """The package name."""
```
a-ovchinnikov: Two non-blocking questions out of curiosity: why strings and not comments and why below and not above a field?

tests/unit/package_managers/pip/test_models.py
```diff
@@ -0,0 +1,96 @@
...
+            package="custom-package",
+            path=Path("/tmp/custom-package-abc123.tar.gz"),
+            requirement_file="requirements.txt",
+            missing_req_file_checksum=False,
+            build_dependency=False,
+            original_url="https://example.com/package.tar.gz",
+            url_with_hash="https://example.com/package.tar.gz#cachito_hash=sha256:abc123",
+        )
+
+        # Test that all types are accepted by the union
```
a-ovchinnikov: How does this actually test that a union accepts all three types? Testing something requires a specific assertion, annotations are (optionally) checked elsewhere. As a slightly exaggerated example consider this:
```python
>>> a: list[int] = ["annotations", "are", "immaterial"]
>>> assert len(a) == 3
>>> # assertion did not raise AssertionError
```

I also don't think we should be checking CPython and mypy internals in our test suit.

a-ovchinnikov (approved):
LGTM with a couple of general questions and a couple of non-blocking nitpicks.

hermeto/core/package_managers/pip/main.py
ben-alkov: :thinking:

hermeto/core/package_managers/pip/models.py
ben-alkov: That... is a very good question. All of the design and most of the code ( particularly the model) predates Claude Code by years. All I can figure is that it reformatted the source at some point. What's even weirder is that I didn't notice :grimacing:.

tests/unit/package_managers/pip/test_models.py
ben-alkov: I agree, I discarded most of the tests Claude Code proposed, but felt weird having so few... (axe comes out again).

hermeto/core/package_managers/pip/main.py
ben-alkov: Good point. Moved into docstrings.

hermeto/core/package_managers/pip/main.py
ben-alkov: > Any IDE and any serious enough editor can provide this context at a click/command

Point taken. Moved into docstrings.

hermeto/core/package_managers/pip/models.py
ben-alkov: Fixed.

tests/unit/package_managers/pip/test_models.py
ben-alkov: Deleted.

hermeto/core/package_managers/pip/main.py
ben-alkov: I prefer the explicit version.

I've got an even better compromise (I don't know why I didn't do this to start with - it's literally the sort of thing `kind` was designed to address):

```python
        if artifact.kind == "pypi":
            dep["index_url"] = artifact.index_url
            dep["package_type"] = artifact.package_type
        elif artifact.kind == "url":
            dep["package_type"] = artifact.package_type
        else:
            # VCS artifacts have neither package_type nor index_url
            dep["package_type"] = None
```

tests/unit/package_managers/pip/test_models.py
a-ovchinnikov: Yes, sometimes things are literally tested elsewhere. This is a refactoring, so I'd say that the existing test suit should be enough: if it does not fail then the refactoring is successful, i.e. the new, better code does the same thing the old code did (at least according to existing tests).

hermeto/core/package_managers/pip/main.py
```diff
@@ -276,34 +287,67 @@ def _checksum_must_match_or_path_unlink(
...
+            build_dependency=build_dependency,
+            url=download_info["url"],
+            host=download_info["host"],
+            namespace=download_info["namespace"],
+            repo=download_info["repo"],
+            ref=download_info["ref"],
+        )
+    elif req.kind == "url":
+        # URL artifact
+        hashes = req.hashes or [req.qualifiers.get("cachito_hash", "")]
```
slimreaper35, line 326: The `cachito_hash` "support" was deleted in #1360.

hermeto/core/package_managers/pip/models.py
```diff
@@ -0,0 +1,127 @@
...
+        if not self.url:
+            raise ValueError("VCS artifact must have non-empty url")
+        if not self.ref:
+            raise ValueError("VCS artifact must have non-empty ref")
+        if not self.host:
+            raise ValueError("VCS artifact must have non-empty host")
+        if not self.namespace:
+            raise ValueError("VCS artifact must have non-empty namespace")
+        if not self.repo:
+            raise ValueError("VCS artifact must have non-empty repo")
```
slimreaper35, line 97: Is it possible that those values can be empty?

hermeto/core/package_managers/pip/models.py
```diff
@@ -0,0 +1,127 @@
...
+    requirement_file: str
+
+    # Whether this artifact is missing checksums in the requirements file.
+    missing_req_file_checksum: bool
+
+    # Whether this is a build-time dependency (from requirements-build.txt).
+    build_dependency: bool
+
+    # Discriminator set by each subclass.
+    kind: Literal["pypi", "url", "vcs"] = field(init=False)
```
slimreaper35, line 30: I believe that was one of the reasons why I didn't like the previous design. If we have concrete classes, do we need the "kind" attribute? I don't see it anywhere else in other package managers' backends.

hermeto/core/package_managers/pip/models.py
```diff
@@ -0,0 +1,127 @@
...
+    def version(self) -> str:
+        """Version identifier for the artifact."""
+
+    def to_filter_dict(self) -> dict[str, Any]:
+        """Convert to dict format expected by filter_packages_with_rust_code."""
+        return {
+            "package": self.package,
+            "path": self.path,
+            "kind": self.kind,
+        }
```
slimreaper35, line 43: The filter_packages_with_rust_code function uses `packages: list[dict[str, Any]]` argument, and it reads only the path from the dictionary.

It just introduces the same approach we want to remove in this PR.

hermeto/core/package_managers/pip/models.py
```diff
@@ -0,0 +1,127 @@
...
+    """Base class for representing a packaged Python artifact."""
+
+    # The package name.
+    package: str
+
+    # Path where the artifact is stored locally.
+    path: Path
+
+    # Path to the requirements file that specified this dependency.
+    requirement_file: str
```
slimreaper35, line 21: nitpick: I think it is better to use longer, more descriptive names than short ones with comments. For example, `package` could be `name`.

hermeto/core/package_managers/pip/main.py
```diff
@@ -461,7 +507,7 @@ def _download_url_package(
     requirement: PipRequirement, pip_deps_dir: RootedPath, trusted_hosts: set[str]
 ) -> dict[str, Any]:
     """
-    Download a Python package from a URL.
+    Download a Python package from a URL (uses `requests`).
```
slimreaper35, line 510: nitpick: Unnecessary comment.

hermeto/core/package_managers/pip/main.py
```diff
@@ -579,41 +624,42 @@ def resolve_req_files(req_files: list[Path] | None, devel: bool) -> list[RootedP
         output_dir, resolved_build_req_files, binary_filters
     )
 
-    if get_config().pip.ignore_dependencies_crates:
+    # Mark build dependencies by creating new frozen instances
+    build_requires_marked = [
+        dataclasses.replace(dependency, build_dependency=True) for dependency in build_requires
+    ]
```
slimreaper35, line 630: Maybe we could set `build_dependency=True` when creating the objects, e.g., propagate the information when we are processing the `requirements-build.txt` or get that information later.

Right now, `process_req` has a default param `build_dependency: bool = False,` and it is not overwritten anywhere if I didn't miss something.

hermeto/core/package_managers/pip/models.py
ben-alkov: If *nothing else*, it avoids us having to use `isinstance` all over the place. It also generally simplifies, and makes it easier to reason about, the processing flow.

hermeto/core/package_managers/pip/models.py
ben-alkov: > I think it is better to use longer, more descriptive names than short ones with comments

Good point

hermeto/core/package_managers/pip/main.py
ben-alkov: Reasonable people can disagree :) . It took me way too long to figure out (by inspection), *exactly* where and how downloads were happening.

I added these comments for future me, and other future readers who aren't even as familiar with pip as I am.

hermeto/core/package_managers/pip/models.py
a-ovchinnikov: Technically `name` is even shorter than `package`. Here comments reside within a class that introduces the names, I'd say as long as one does not need to comment on a name at the spot where it is used it is fine to have comments (and these names all seem to be striking the right balance between brevity and expressiveness)

slimreaper35: > For example, package could be name.

EDIT: I meant `package_name`. The names are subjective, so it's a nitpick.

ben-alkov: > EDIT: I meant package_name

I figured :wink:

hermeto/core/package_managers/pip/models.py
ben-alkov: e.g., the canonical example: main.py, ll. 644 `for artifact in all_artifacts:`.

What *would be* `if isinstance(PyPIArtifact):` is more direct and readable (IMO): `if artifact.kind == "pypi":`

hermeto/core/package_managers/pip/models.py
ben-alkov: You're right, they can't be; none of them are Optional, so they'll throw on instantiation

slimreaper35:
Implemented in https://github.com/hermetoproject/hermeto/pull/1585

PR #1286, opened 2026-02-10
fix(build): container image reporting wrong version

ST3V1K:
Ensure the image reports the actual image tag (e.g. 0.20.0) instead of a generic or post-release version string when running --version inside the container.

closes https://github.com/hermetoproject/hermeto/issues/822

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,27 @@
-*
+# Python cache
+__pycache__/
+*.pyc
 
-!.git/
-!.git_archival.txt
-!hermeto/
-!pyproject.toml
-!requirements.txt
-!MANIFEST.in
+# Environments
+venv/
+.venv/
+.env
+*.env
+
+# Distribution / packaging
+*.egg-info/
+build/
+dist/
+site/
+
+# Test and tool caches
+.nox/
+.pytest_cache/
+.mypy_cache/
+.coverage
+coverage.xml
+
+# Project outputs
+dest/
+hermeto-output/
+cachi2-output/


.dockerignore
```diff
@@ -1,8 +1,8 @@
...
-!requirements.txt
-!MANIFEST.in
+__pycache__/
+*.pyc
+venv/
+.venv/
+.nox/
+.pytest_cache/
+.mypy_cache/
+*.egg-info/
```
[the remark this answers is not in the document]
eskultety: I guess the suggestion and reasoning is fair, @ST3V1K can you update please?

.dockerignore
```diff
@@ -1,8 +1,8 @@
-*
```
eskultety, line 1: **Commit message:**

> Ensure the image reports the actual image tag (e.g. 0.20.0) instead of
> a generic or post-release version string when running --version inside
> the container.

Knowing the full history of this I'd like to see a more descriptive WHY on this change :), i.e.:

- current `.dockerignore` doesn't mount some of th egit tracked objects, e.g. tests
- missing git tracked files show up as `deleted` in the container image build environment
- setuptools-scm derives version based on the git repo state, in this case - a dirty repo state (a bunch of "deleted" files)
- defining the `.dockerignore` file such that everything tracked in git (i.e. IOW inverts the negative logic to use the common convention) is mounted to the container build fixes the issue for good, hopefully.

.dockerignore
```diff
@@ -1,8 +1,27 @@
...
+build/
+dist/
+site/
+
+# Test and tool caches
+.nox/
+.pytest_cache/
+.mypy_cache/
+.coverage
+coverage.xml
```
eskultety, line 22: nitpick: I see this has been inspired with our .gitignore :) . I noticed we don't use coverage.xml anymore as the tool is configured in our `pyproject.toml`:
```
[tool.coverage.report]
skip_covered = true
show_missing = true
fail_under = 90
exclude_lines = [
  "def __repr__",
  "if __name__ == .__main__.:",
  "if TYPE_CHECKING:",
  "return NotImplemented",
]
```

non-actionable.

eskultety (approved):

PR #1287, opened 2026-02-10
feat(hack): add script to update generic lockfile checksums

ST3V1K:
Generic fetcher users must keep artifacts.lock.yaml checksums in sync when upstreams change. Hosting this script in hack/ lets them do that in one step instead of editing the lockfile by hand.

License details: [wiki.geant.org](https://wiki.geant.org/spaces/GSD/pages/419758342/Reference+Information+about+OSS+Licences+and+Tools#ReferenceInformationaboutOSSLicencesandTools-LicenceCompatibilityMatricesandCheckers)

closes https://github.com/hermetoproject/hermeto/issues/1018

### DIFF
--- a/hack/script/update_artifacts_lockfile
+++ b/hack/script/update_artifacts_lockfile
@@ -5,6 +5,9 @@
 # (https://github.com/coreos/coreos-assembler)
 # (https://github.com/coreos/coreos-assembler/blob/c239624efd2e411b9bd3b2694e246f61f84c205e/ci/hermetic/update_artifacts_lockfile)
 
+# Modification notice:
+# Changed ruamel.yaml to PyYAML to avoid dependency on ruamel.yaml.
+
 """
 This script reads a YAML file containing a list of artifacts.
 For each artifact, it downloads the file from the specified 'download_url',
@@ -18,7 +21,7 @@ of dictionaries. Each dictionary should have at least 'download_url',
 '<hash_algorithm>:<hash_value>'.
 
 Dependencies:
-- ruamel.yaml: For reading and writing YAML files while preserving style.
+- PyYAML: For reading and writing YAML files.
 - requests: For downloading files from URLs.
 """
 
@@ -26,7 +29,7 @@ import hashlib
 from typing import Any, Dict, List, Optional
 
 import requests
-from ruamel.yaml import YAML
+import yaml
 
 
 def calculate_checksum(file_content: bytes, algorithm_name: str) -> Optional[str]:
@@ -63,18 +66,14 @@ def update_artifact_checksums(yaml_file_path: str) -> None:
     Args:
         yaml_file_path (str): The path to the input YAML file.
     """
-    yaml_parser = YAML()
-    yaml_parser.default_flow_style = False
-    yaml_parser.preserve_quotes = True  # Keep existing quotes (single/double)
-    yaml_parser.indent(mapping=2, sequence=4, offset=2)
     data: Dict[str, Any] = {}
     try:
         with open(yaml_file_path, "r", encoding="utf-8") as f:
-            data = yaml_parser.load(f)
+            data = yaml.safe_load(f)
     except FileNotFoundError:
         print(f"Error: YAML file not found at '{yaml_file_path}'")
         return
-    except Exception as e:  # ruamel.yaml can raise different exceptions
+    except yaml.YAMLError as e:
         print(f"Error parsing YAML file: {e}")
         return
 
@@ -141,8 +140,13 @@ def update_artifact_checksums(yaml_file_path: str) -> None:
 
     try:
         with open(yaml_file_path, "w", encoding="utf-8") as f:
-            # ruamel.yaml's dump method handles preserving formatting
-            yaml_parser.dump(data, f)
+            yaml.dump(
+                data,
+                f,
+                default_flow_style=False,
+                sort_keys=False,
+                indent=2,
+            )
         print(f"\nSuccessfully updated checksums in '{yaml_file_path}'.")
     except Exception as e:
         print(f"Error writing updated YAML file: {e}")
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,9 +49,6 @@ test = [
   "pytest-cov",
   "pytest-env",
 ]
-hack = [
-  "ruamel.yaml",
-]
 [project.scripts]
 cachi2 = "hermeto.interface.cli:app"
 hermeto = "hermeto.interface.cli:app"
@@ -126,10 +123,6 @@ ignore_missing_imports = true
 [[tool.mypy.overrides]]
 module = "noxfile"
 disallow_untyped_decorators = false
-
-[[tool.mypy.overrides]]
-module = "ruamel.*"
-ignore_missing_imports = true
 
 [tool.coverage.report]
 skip_covered = true
--- a/requirements-extras.txt
+++ b/requirements-extras.txt
@@ -1616,10 +1616,6 @@ rpds-py==0.30.0 \
     # via
     #   jsonschema
     #   referencing
-ruamel-yaml==0.19.1 \
-    --hash=sha256:27592957fedf6e0b62f281e96effd28043345e0e66001f97683aa9a40c667c93 \
-    --hash=sha256:53eb66cd27849eff968ebf8f0bf61f46cdac2da1d1f3576dd4ccee9b25c31993
-    # via hermeto (pyproject.toml)
 ruff==0.15.0 \
     --hash=sha256:238a717ef803e501b6d51e0bdd0d2c6e8513fe9eec14002445134d3907cd46c3 \
     --hash=sha256:5298d518e493061f2eabd4abd067c7e4fb89e2f63291c94332e35631c07c3662 \


hack/script/update_artifacts_lockfile
```diff
@@ -0,0 +1,156 @@
...
+
+    if "artifacts" not in data or not isinstance(data["artifacts"], list):
+        print("Error: YAML file must contain an 'artifacts' list.")
+        return
+
+    # Ensure mypy can infer the type of data['artifacts'] as
+    # List[Dict[str, Any]]
+    artifacts: List[Dict[str, Any]] = data["artifacts"]
+
+    for i, artifact in enumerate(artifacts):
```
a-ovchinnikov, line 88: Nitpick: `i` is a great name for a variable that lives on one line of code, two maximum. If, however, your loop does not fit on a screen you will be better off with a more expressive name, maybe `art_number` or something like that.

hack/script/update_artifacts_lockfile
```diff
@@ -0,0 +1,156 @@
...
+        download_url: Optional[str] = artifact.get("download_url")
+        current_checksum: Optional[str] = artifact.get("checksum")
+        filename: Optional[str] = artifact.get("filename")
+
+        if not download_url:
+            print(f"Warning: Artifact {i + 1} is missing 'download_url'. Skipping.")
```
a-ovchinnikov, line 94: Two observations: 1) `enumerate` takes `start` argument, you can start counting from 1 and 2) `artifact 42` is not very user-friendly way of mentioning issues, surely there could be something else present, filename or sha?

hack/script/update_artifacts_lockfile
```diff
@@ -0,0 +1,156 @@
...
+
+        try:
+            response = requests.get(download_url, stream=True, timeout=10)
+            response.raise_for_status()  # Raise an HTTPError for bad responses
+
+            # Read content in chunks to handle large files efficiently
+            file_content_chunks: List[bytes] = []
+            for chunk in response.iter_content(chunk_size=8192):
+                file_content_chunks.append(chunk)
+            file_content: bytes = b"".join(file_content_chunks)
```
a-ovchinnikov, line 119: Have you considered using hermeto for prefetching generic artifacts?

hack/script/update_artifacts_lockfile
```diff
@@ -0,0 +1,156 @@
...
+Dependencies:
+- ruamel.yaml: For reading and writing YAML files while preserving style.
+- requests: For downloading files from URLs.
+"""
+
+import hashlib
+from typing import Any, Dict, List, Optional
+
+import requests
+from ruamel.yaml import YAML
```
a-ovchinnikov: Why do we need another yaml dependency?

hack/script/update_artifacts_lockfile
ST3V1K: The whole file is only a copy of [[1]](https://github.com/coreos/coreos-assembler/blob/83fdbc45ba4977b878d4ab73f1ac4205abe7a48a/ci/hermetic/update_artifacts_lockfile) as requested in https://github.com/hermetoproject/hermeto/issues/1018. Should i modify it?

[1] https://github.com/coreos/coreos-assembler/blob/83fdbc45ba4977b878d4ab73f1ac4205abe7a48a/ci/hermetic/update_artifacts_lockfile

a-ovchinnikov: If it is just a direct copy then please disregard my comments on any code style. The only thing that matters in this case is the extra dependency which I would suggest to avoid.

hack/script/update_artifacts_lockfile
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python3
+
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
a-ovchinnikov: @eskultety WDYT, can we copy files over and modify license information?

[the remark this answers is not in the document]
eskultety: Absolutely not, not even with permissive licenses like Apache, not without an explicit permission from the authors. The things to watch out for when using multiple licenses in a project:

1.  what does the license permit you to do, can you include it in your project? What are the conditions for redistribution? etc.
2. license compatibility, i.e. is the license you're adopting compatible with the license under which your project is released?

Prime example of brainless AI slop.

hack/script/update_artifacts_lockfile
eskultety: So, if there are any design, logical, styling, etc. comments, then either of these can happen:

1. merge the PR as is, apply all comments in a follow-up PR (unnecessary IMO)
2. make this a 2 commit change:
   - commit 1 introduces the file verbatim, no changes, including the correct license SPDX ID
   - commit 2 addresses all outstanding comments

eskultety: Upon further inspection of https://www.apache.org/licenses/LICENSE-2.0 what **has to** happen with a modification is to add a notice that the file was modified, although the wording doesn't explicitly state we have to document every change explicitly like in a changelog:

> You must cause any modified files to carry prominent notices stating that You changed the files; and

I guess based on the vague and loose wording it can be pretty much anything.

hack/script/update_artifacts_lockfile
```diff
@@ -0,0 +1,156 @@
...
+# Derived from coreos-assembler ci/hermetic/update_artifacts_lockfile
+# (https://github.com/coreos/coreos-assembler)
+# (https://github.com/coreos/coreos-assembler/blob/c239624efd2e411b9bd3b2694e246f61f84c205e/ci/hermetic/update_artifacts_lockfile)
```
eskultety, line 6: If this is supposed to serve as the Apache 2.0 modification notice then I think:

- it should be a bit more visible, e.g. "mofification notice: original work from ..."
- the github link should appear in the commit message itself.

hack/script/update_artifacts_lockfile
eskultety: > So, if there are any design, logical, styling, etc. comments, then either of these can happen:
> 
>     1. merge the PR as is, apply all comments in a follow-up PR (unnecessary IMO)
> 
>     2. make this a 2 commit change:
>        
>        * commit 1 introduces the file verbatim, no changes, including the correct license SPDX ID
>        * commit 2 addresses all outstanding comments

Well, it could obviously all be done in a single commit provided we include the modification notice, but for bookkeeping and review reasons it's always cleaner if we just introduce something like this verbatim first and then make the modifications, because it reduces the context window for the modifications.

hack/script/update_artifacts_lockfile
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python3
```
eskultety, line 1: **Commit subject:**

anything under hack cannot be a feature, because hack is not a functional part of our distributable logic :) .

hack/script/update_artifacts_lockfile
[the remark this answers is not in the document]
a-ovchinnikov: My understanding is that Apache2.0 is a weaker license than GPLv3.0, however I am not sure how exactly this is supposed to work, especially in the case of user-provided scripts. Are they supposed to be governed by the same license as the bulk of the project? On one hand yes, they live in the project tree, so any license in the root applies, on the other hand these are effectively third-party contributions which the team is not really responsible for. To make things more complicated this particular script already has a license header coming from its repository of origin, to the best of my knowledge such headers are meant for dealing with precisely such situations -- when someone copies some code over from one project to another. Does GPLv3.0 allow one to include code distributed under other licenses? I am not a lawyer, but I doubt it. I think this PR sets an interesting precedent that has to be carefully thought through.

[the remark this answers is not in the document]
eskultety: TL;DR - we're totally fine this time, no license infringement with  honoring the Apache 2.0 license.

> Are they supposed to be governed by the same license as the bulk of the project?

Yes.

> To make things more complicated this particular script already has a license header coming from its repository of origin, to the best of my knowledge such headers are meant for dealing with precisely such situations 

Correct. It's a good thing, because you can't accidentally forget the root license of the origin project when doing this.

> Does GPLv3.0 allow one to include code distributed under other licenses?

Yes, that's why I mentioned license compatibility in my response above and I asked @ST3V1K to do his due diligence, hence the compatibility matrix link to wiki he included, i.e. consuming Apache in GPL is fine, because a permissive license allows stronger copyleft license restrictions, but you can't do it vice-versa, because you'd suddenly sublicense a file in a more permissive way than what e.g. GPL license would allow for redistributing hence in this case (and you can see it in the matrix) this is called "one way compatibility".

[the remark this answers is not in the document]
a-ovchinnikov: Thank you, TIL!

hack/script/update_artifacts_lockfile
eskultety: @a-ovchinnikov do you stand by your comments given that this was originally proposed as plain adoption of a user script? If so then we most likely need a 2 commit work here, but since this is just `hack`, i.e. no guarantees or anything I think we're good. If you agree with me I think we can resolve the comments.

a-ovchinnikov: No, sorry, let's resolve. I've mentioned in another thread that I did not realize at first that this is a copy, and since this is just a copy then all my style suggestions immediately become optional.

hack/script/update_artifacts_lockfile
ST3V1K: Hopefully mentioning modification notice in a commit message and in a comment on top of the file should be enough.

hack/script/update_artifacts_lockfile
ST3V1K: Added a notice under this.

hack/script/update_artifacts_lockfile
ST3V1K: Changed it to `chore`

PR #1288, opened 2026-02-10
add Nexus module for proxy integration testing

taylormadore:
Configures a local Nexus repository server for running integration tests and handles the container lifecycle. Supports configurable repositories and can be used standalone for debugging.

This will be integrated into our integration test framework in future PRs.

### DIFF
--- a/tests/nexusserver/start.py
+++ b/tests/nexusserver/start.py
@@ -1,0 +1,483 @@
+#!/usr/bin/env python3
+"""Start and initialize a Nexus Repository Server for integration tests."""
+
+import contextlib
+import logging
+import signal
+import subprocess
+import time
+from types import FrameType
+from typing import Any, NoReturn
+
+import requests
+import typer
+from requests.auth import HTTPBasicAuth
+from tenacity import (
+    RetryError,
+    retry,
+    retry_if_exception_type,
+    stop_after_delay,
+    wait_fixed,
+)
+
+if __package__:
+    from .repositories import DEFAULT_REPOSITORIES, ProxyRepositoryConfig
+else:
+    from repositories import (  # type: ignore[import-not-found,no-redef]
+        DEFAULT_REPOSITORIES,
+        ProxyRepositoryConfig,
+    )
+
+log = logging.getLogger(__name__)
+
+app = typer.Typer()
+
+DEFAULT_SUBPROCESS_TIMEOUT = 300
+DEFAULT_NEXUS_IMAGE = "docker.io/sonatype/nexus3:latest"
+DEFAULT_NEXUS_HOST = "127.0.0.1"
+DEFAULT_NEXUS_PORT = 8082
+DEFAULT_NEXUS_ADMIN_PASSWORD = "admin123"  # noqa: S105
+DEFAULT_NEXUS_CONTAINER_NAME = "hermeto-nexus"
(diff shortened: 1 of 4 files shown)


tests/nexusserver/start.py
```diff
@@ -0,0 +1,452 @@
...
+        nonlocal _shutdown
+        _shutdown = True
+
+    signal.signal(signal.SIGINT, cleanup_on_exit)
+    signal.signal(signal.SIGTERM, cleanup_on_exit)
+
+    print("\n" + "=" * 80)
+    print("Nexus Repository Server is ready!")
+    print(f"  URL: {client.base_url}")
+    print(f"  Admin credentials: admin / {admin_password}")
```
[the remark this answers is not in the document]
taylormadore: This is for local, ephemeral integration testing. The password is already hard-coded, so no risk to logging it

tests/nexusserver/start.py
```diff
@@ -0,0 +1,452 @@
...
+        DEFAULT_REPOSITORIES,
+        ProxyRepositoryConfig,
+    )
+
+log = logging.getLogger(__name__)
+
+app = typer.Typer()
+
+DEFAULT_SUBPROCESS_TIMEOUT = 300
+DEFAULT_NEXUS_IMAGE = "sonatype/nexus3:latest"
```
eskultety: nitpick: you might want to use the fully qualified path - `docker.io/sonatype/nexus3` to prevent podman interactively asking from which registry it should try pulling the image.

tests/nexusserver/start.py
```diff
@@ -0,0 +1,452 @@
...
+    with contextlib.ExitStack() as stack:
+        container = NexusContainer(
+            image=image,
+            container_name=container_name,
+            volume_name=volume_name,
+            port=port,
+            subprocess_timeout=subprocess_timeout,
+        )
+        container.start()
+        stack.enter_context(container)
```
eskultety: nitpick: would it be slightly more correct to first enter the context before starting the container or am I missing something?

tests/nexusserver/start.py
```diff
@@ -0,0 +1,452 @@
...
+try:
+    from tests.nexusserver.repositories import (
+        DEFAULT_REPOSITORIES,
+        ProxyRepositoryConfig,
+    )
+except ImportError:
+    from repositories import (  # type: ignore[import-not-found,no-redef]
+        DEFAULT_REPOSITORIES,
+        ProxyRepositoryConfig,
+    )
```
eskultety: nitpick: would
```
if __package__:
  from tests.nexusserver.... import
else:
  from repositories import
```
be a tiny bit clearer or am I missing some important detail?

tests/nexusserver/start.py
taylormadore: Good point. Fixed :+1:

tests/nexusserver/start.py
taylormadore: I thought I fixed this already :man_facepalming: . Realized I never pulled in a fixup from the franken-branch.

Fixed, along with some other non-functional items

tests/nexusserver/start.py
taylormadore: Fixed

taylormadore:
Rebased on main

PR #1291, opened 2026-02-11
fix(sbom): coerce empty Component version to None

ST3V1K:
Component used None to mean "version not specified". An empty string from package data was not normalized and could appear in the SBOM. Add a field validator that coerces empty/falsy version values to None so only None is used for unspecified versions and empty strings do not propagate.

Closes https://github.com/hermetoproject/hermeto/issues/990

### DIFF
--- a/hermeto/core/models/sbom.py
+++ b/hermeto/core/models/sbom.py
@@ -122,6 +122,14 @@ class Component(pydantic.BaseModel):
         """
         return self.purl
 
+    @pydantic.field_validator("version")
+    @classmethod
+    def _remove_empty_version(cls, version: str | None) -> str | None:
+        if not version:
+            return None
+
+        return version
+
     @pydantic.field_validator("properties")
     @classmethod
     def _add_found_by_property(cls, properties: list[Property]) -> list[Property]:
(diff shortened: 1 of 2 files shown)


hermeto/core/models/sbom.py
```diff
@@ -124,6 +124,14 @@ def update_bom_ref(self) -> Self:
         self.bom_ref = self.purl
         return self
 
+    @pydantic.field_validator("version")
+    @classmethod
+    def _remove_empty_version(cls, version: str | None) -> str | None:
+        if not version:
```
a-ovchinnikov, line 128: Non-blocker: for the best explicit type precision I would have expanded this to `if version is not None and not version` because `None` is a special signaling object whose appearance warrants special handling (checking if something is this unique object). OTOH this is a rather fine hair splitting so I suggest keeping the code as is and remembering that `None` is not just another synonym to `False`.

slimreaper35:
Does it make sense to have a basic unit test for this ?

hermeto/core/models/sbom.py
eskultety: Right, in this instance I believe this is a very correct and very compact way of checking -> resolving.

PR #1292, opened 2026-02-11
Revert "gomod: moved from getting the Go toolchain with 'go version' …

slimreaper35:


hermeto/core/package_managers/gomod.py
```diff
@@ -406,9 +406,27 @@ def version(self) -> GoVersion:
         return GoVersion(self._get_release())
 
     def _get_release(self) -> str:
-        output = self(["env", "GOVERSION"], params={"env": {"GOTOOLCHAIN": "local"}})
+        output = self(["version"], params={"env": {"GOTOOLCHAIN": "local"}})
```
eskultety, line 409: **Commit message:**
You're missing a reason for the revert - This is breaking our latest release because we found a Go toolchain that reports some unexpected characters (or format if you will) in the `GOVERSION` variable:
```
[root@d2e3e6323d9b /]# /usr/local/go/go_latest/bin/go env GOVERSION
go1.25.5 (Red Hat 1.25.5-1.el10_1)
```

PR #1293, opened 2026-02-11
Align tests with exit codes and exception types

### LINKED ISSUE #1250: Define a set of exit codes and match them with our exceptions
We've introduced a handful of new granular exception types to make our unit tests more stable and less dependent on the exact error strings which we match/matched against. We still do that with integration tests, especially the negative ones. Since Hermeto is predominantly a CLI tool, a good CLI experience includes a variety of exit codes ("errno" if you will) to be paired with the `stderr` produced allowing one to silently redirect all output to a sink, but still be able to check for the error code whichever way the user feels like.

- go through the exceptions we have defined
- define a list of numeric exit codes for each distinct exception type
- make sure we don't match against any error strings in the negative integration tests

ST3V1K:
Exit codes were previously 0, 1, or 2 (usage error), so callers had to parse
stderr to tell failure reasons apart.

Assign each error type a dedicated exit code (1–22), centralize the mapping in
errors.py with validation, and have the CLI use e.exit_code. Scripts and CI can
rely on specific codes (e.g. LockfileNotFound=13, PackageRejected=5) without
parsing output.

Integration and unit tests now assert by error type (expected_error=...) instead
of expected_exit_code and fragile output snippets, so tests stay stable and
failures show the error name.

Closes https://github.com/hermetoproject/hermeto/issues/1250

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -246,6 +246,12 @@ could be to manually create an internal network (but you'll need root
 privileges): `sudo podman network create --internal isolated-network; sudo
 podman build --network isolated-network ...`.
 
+## Exit codes
+
+Hermeto uses a set of exit codes to signal different error conditions. These are
+internal only and serve informational purposes and hence may change in between
+releases, please do NOT depend on them!
+
 [buildah#4227]: https://github.com/containers/buildah/issues/4227
 [CycloneDX v1.6]: https://cyclonedx.org/docs/1.6/json
 [limited set]: https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/models/sbom.py#L7-L13
--- a/hermeto/core/errors.py
+++ b/hermeto/core/errors.py
@@ -1,17 +1,82 @@
 # SPDX-License-Identifier: GPL-3.0-only
 import textwrap
+from enum import IntEnum, unique
 from pathlib import Path
-from typing import ClassVar, Iterable
+from typing import Any, ClassVar, Iterable
 
 from hermeto import APP_NAME
 
 _argument_not_specified = "__argument_not_specified__"
 
 
-class BaseError(Exception):
+@unique
+class ExitError(IntEnum):
+    """Hermeto process exit codes."""
+
+    ERR_OK = 0
+    ERR_BASE = 1
+    ERR_USAGE = 2
+    ERR_PATH_OUTSIDE_ROOT = 3
+    ERR_INVALID_INPUT = 4
+    ERR_PACKAGE_REJECTED = 5
+    ERR_NOT_A_GIT_REPO = 6
+    ERR_UNEXPECTED_FORMAT = 7
+    ERR_UNSUPPORTED_FEATURE = 8
+    ERR_EXECUTABLE_NOT_FOUND = 9
+    ERR_CHECKSUM_VERIFICATION_FAILED = 10
+    ERR_INVALID_CHECKSUM = 11
+    ERR_MISSING_CHECKSUM = 12
+    ERR_LOCKFILE_NOT_FOUND = 13
+    ERR_INVALID_LOCKFILE_FORMAT = 14
+    ERR_FETCH = 15
+    ERR_PACKAGE_MANAGER = 16
+    ERR_GIT = 17
+    ERR_GIT_REMOTE_NOT_FOUND = 18
+    ERR_GIT_INVALID_REVISION = 19
+    ERR_PACKAGE_WITH_CORRUPT_LOCKFILE_REJECTED = 20
+    ERR_UNSATISFIABLE_ARCHITECTURE_FILTER = 21
+    ERR_NOT_V1_LOCKFILE = 22
+
+
+class ErrorRegistryMeta(type):
+    """Meta class for registering errors."""
+
+    registry: ClassVar[dict[ExitError, "ErrorRegistryMeta"]] = {}
+
+    def __init__(
+        cls, name: str, bases: tuple[type, ...], namespace: dict[str, Any], **kwargs: Any
+    ) -> None:
+        """Register error subclasses that declare their own exit code.
+
+        Subclasses that set ``_exit_error`` explicitly are registered and
+        checked for uniqueness.  Subclasses that omit it simply inherit
+        the parent's exit code — this is intentional for internal /
+        intermediary exceptions that are caught in code and never
+        propagated to the user.
+
+        :param name: name of the error class
+        :param bases: base classes of the error class
+        :param namespace: namespace of the error class
+        :param kwargs: keyword arguments
+        """
+        super().__init__(name, bases, namespace, **kwargs)
+        # Subclasses without _exit_error inherit their parent's code and
+        # are not registered
+        if "_exit_error" in namespace:
+            exit_error = namespace["_exit_error"]
+            if exit_error in ErrorRegistryMeta.registry:
+                raise RuntimeError(
+                    f"{name} and {ErrorRegistryMeta.registry[exit_error].__name__} "
+                    f"both use {exit_error!r}"
+                )
+            ErrorRegistryMeta.registry[exit_error] = cls
+
+
+class BaseError(Exception, metaclass=ErrorRegistryMeta):
     """Root of the error hierarchy. Don't raise this directly, use more specific error types."""
 
     is_invalid_usage: ClassVar[bool] = False
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_BASE
     default_solution: ClassVar[str | None] = None
 
     def __init__(
@@ -31,6 +96,11 @@ class BaseError(Exception):
         else:
             self.solution = solution
 
+    @property
+    def exit_code(self) -> int:
+        """Return the exit code for this error."""
+        return self._exit_error.value
+
     def friendly_msg(self) -> str:
         """Return the user-friendly representation of this error."""
         msg = str(self)
@@ -43,10 +113,13 @@ class UsageError(BaseError):
     """Generic error for "Hermeto was used incorrectly." Prefer more specific errors."""
 
     is_invalid_usage: ClassVar[bool] = True
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_USAGE
 
 
 class PathOutsideRoot(UsageError):
     """Afer joining a subpath, the result is outside the root of a rooted path."""
+
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_PATH_OUTSIDE_ROOT
 
     def __init__(
         self,
@@ -75,6 +148,8 @@ class PathOutsideRoot(UsageError):
 class InvalidInput(UsageError):
     """User input was invalid."""
 
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_INVALID_INPUT
+
 
 class PackageRejected(UsageError):
     """The Application refused to process the package the user requested.
@@ -82,6 +157,8 @@ class PackageRejected(UsageError):
     a) The package appears invalid (e.g. missing go.mod for a Go module).
     b) The package does not meet our extra requirements (e.g. missing checksums).
     """
+
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_PACKAGE_REJECTED
 
     def __init__(self, reason: str, *, solution: str | None) -> None:
         """Initialize a Package Rejected error.
@@ -97,9 +174,13 @@ class PackageRejected(UsageError):
 class NotAGitRepo(PackageRejected):
     """A package turned out to be not a git repository."""
 
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_NOT_A_GIT_REPO
+
 
 class UnexpectedFormat(UsageError):
     """The Application failed to parse a file in the user's package (e.g. requirements.txt)."""
+
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_UNEXPECTED_FORMAT
 
     default_solution = (
         "Please check if the format of your file is correct.\n"
@@ -113,6 +194,8 @@ class UnsupportedFeature(UsageError):
     The requested feature might be valid, but application doesn't implement it.
     """
 
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_UNSUPPORTED_FEATURE
+
     default_solution = (
         f"If you need {APP_NAME} to support this feature, please contact the maintainers."
     )
@@ -120,6 +203,8 @@ class UnsupportedFeature(UsageError):
 
 class ExecutableNotFound(UsageError):
     """A required executable was not found in PATH."""
+
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_EXECUTABLE_NOT_FOUND
 
     def __init__(
         self,
@@ -145,6 +230,8 @@ class ExecutableNotFound(UsageError):
 class ChecksumVerificationFailed(PackageRejected):
     """Checksum verification failed for a file."""
 
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_CHECKSUM_VERIFICATION_FAILED
+
     def __init__(
         self,
         filename: Path | str,
@@ -166,6 +253,8 @@ class ChecksumVerificationFailed(PackageRejected):
 
 class InvalidChecksum(PackageRejected):
     """Provided checksum/hash/integrity is not valid data."""
+
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_INVALID_CHECKSUM
 
     def __init__(
         self,
@@ -189,6 +278,8 @@ class InvalidChecksum(PackageRejected):
 
 class MissingChecksum(InvalidChecksum):
     """Provided checksum/hash/integrity is missing"""
+
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_MISSING_CHECKSUM
 
     def __init__(
         self,
@@ -216,6 +307,8 @@ class MissingChecksum(InvalidChecksum):
 class LockfileNotFound(PackageRejected):
     """A required lockfile was not found."""
 
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_LOCKFILE_NOT_FOUND
+
     def __init__(
         self,
         files: Path | str | Iterable[Path | str],
@@ -240,6 +333,8 @@ class LockfileNotFound(PackageRejected):
 class InvalidLockfileFormat(PackageRejected):
     """Lockfile format is invalid or cannot be parsed."""
 
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_INVALID_LOCKFILE_FORMAT
+
     def __init__(
         self,
         lockfile_path: Path | str,
@@ -262,6 +357,8 @@ class InvalidLockfileFormat(PackageRejected):
 class FetchError(BaseError):
     """The Application failed to fetch a dependency or other data needed to process a package."""
 
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_FETCH
+
     default_solution = (
         "The error might be intermittent, please try again.\n"
         f"If the issue seems to be on the {APP_NAME} side, please contact the maintainers."
@@ -274,6 +371,8 @@ class PackageManagerError(BaseError):
     Maybe some configuration is invalid, maybe the package manager was unable to fetch a dependency,
     maybe the error is intermittent. We don't really know, but we do at least log the stderr.
     """
+
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_PACKAGE_MANAGER
 
     def __init__(
         self,
@@ -304,6 +403,8 @@ class PackageManagerError(BaseError):
 
 class GitError(BaseError):
     """Base class for Git operation failures."""
+
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_GIT
 
     def __init__(
         self,
@@ -342,6 +443,10 @@ class GitError(BaseError):
 class GitRemoteNotFoundError(GitError):
     """A Git remote with the specified name does not exist."""
 
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_GIT_REMOTE_NOT_FOUND
+
 
 class GitInvalidRevisionError(GitError):
     """Invalid Git revision (commits, branches, tags, revision specifiers)."""
+
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_GIT_INVALID_REVISION
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -7,7 +7,7 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import cached_property
 from pathlib import Path
-from typing import Any, NamedTuple
+from typing import Any, ClassVar, NamedTuple
 from urllib.parse import urlparse, urlunparse
 
 import tomlkit
@@ -18,6 +18,7 @@ from hermeto import APP_NAME
 from hermeto.core.config import get_config
 from hermeto.core.constants import Mode
 from hermeto.core.errors import (
+    ExitError,
     LockfileNotFound,
     NotAGitRepo,
     PackageManagerError,
@@ -45,6 +46,8 @@ class CargoVendorResult(NamedTuple):
 
 class PackageWithCorruptLockfileRejected(PackageRejected):
     """Package lock file does not match package config."""
+
+    _exit_error: ClassVar[ExitError] = ExitError.ERR_PACKAGE_WITH_CORRUPT_LOCKFILE_REJECTED
 
     def __init__(self, package_path: str) -> None:
         """Initialize the error."""
--- a/hermeto/interface/cli.py
+++ b/hermeto/interface/cli.py
@@ -123,7 +123,7 @@ def _bail_out_with_error(e: BaseError) -> None:
     """Report and error and set correct exit code."""
     log.error("%s: %s", type(e).__name__, str(e).replace("\n", r"\n"))
     print(f"Error: {type(e).__name__}: {e.friendly_msg()}", file=sys.stderr)
-    raise typer.Exit(2 if e.is_invalid_usage else 1)
+    raise typer.Exit(e.exit_code)
 
 
 def handle_errors(cmd: Callable[..., None]) -> Callable[..., None]:
--- a/tests/integration/test_bundler.py
+++ b/tests/integration/test_bundler.py
@@ -4,6 +4,8 @@ from pathlib import Path
 
 import pytest
 
+from hermeto.core.errors import ExitError
+
 from . import utils
 
 log = logging.getLogger(__name__)
@@ -18,7 +20,7 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "bundler"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=2,
+                expected_error=ExitError.ERR_LOCKFILE_NOT_FOUND,
                 expected_output="Required files not found:",
             ),
             id="bundler_missing_gemfile",
@@ -29,7 +31,7 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "bundler"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=2,
+                expected_error=ExitError.ERR_LOCKFILE_NOT_FOUND,
                 expected_output="Required files not found:",
             ),
             id="bundler_missing_lockfile",
@@ -40,7 +42,7 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "bundler"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=1,
+                expected_error=ExitError.ERR_PACKAGE_MANAGER,
                 expected_output="Failed to parse",
             ),
             id="bundler_missing_git_revision",
@@ -72,8 +74,6 @@ def test_bundler_packages(
                 packages=({"path": ".", "type": "bundler", "binary": {}},),
                 check_output=True,
                 check_deps_checksums=False,
-                expected_exit_code=0,
-                expected_output="",
             ),
             [],  # No additional commands are run to verify the build
             [],
@@ -85,8 +85,6 @@ def test_bundler_packages(
                 packages=({"path": ".", "type": "bundler", "binary": {}},),
                 check_output=True,
                 check_deps_checksums=False,
-                expected_exit_code=0,
-                expected_output="",
             ),
             [],  # No additional commands are run to verify the build
             [],
--- a/tests/integration/test_cargo.py
+++ b/tests/integration/test_cargo.py
@@ -4,6 +4,8 @@ from pathlib import Path
 
 import pytest
 
+from hermeto.core.errors import ExitError
+
 from . import utils
 
 log = logging.getLogger(__name__)
@@ -18,8 +20,6 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "cargo"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=0,
-                expected_output="",
             ),
             id="cargo_just_a_crate_dependency",
         ),
@@ -29,8 +29,6 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "cargo"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=0,
-                expected_output="",
             ),
             id="cargo_just_a_git_dependency",
         ),
@@ -40,8 +38,6 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "cargo"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=0,
-                expected_output="",
             ),
             id="cargo_mixed_git_crate_dependency",
         ),
@@ -51,8 +47,6 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "cargo"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=0,
-                expected_output="",
             ),
             id="cargo_uses_resolver_v3",
         ),
@@ -62,7 +56,7 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "cargo"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=2,
+                expected_error=ExitError.ERR_LOCKFILE_NOT_FOUND,
                 expected_output="Cargo.lock not found",
             ),
             id="cargo_missing_lockfile",
@@ -74,7 +68,6 @@ log = logging.getLogger(__name__)
                 global_flags=["--mode", "permissive"],
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=0,
                 expected_output="All dependencies fetched successfully",
             ),
             id="cargo_missing_lockfile_permissive_mode",
@@ -106,8 +99,6 @@ def test_cargo_packages(
                 packages=({"path": ".", "type": "cargo"},),
                 check_output=True,
                 check_deps_checksums=False,
-                expected_exit_code=0,
-                expected_output="",
             ),
             [],  # No additional commands are run to verify the build
             [],
@@ -119,7 +110,6 @@ def test_cargo_packages(
                 packages=({"path": ".", "type": "cargo"},),
                 check_output=True,
                 check_deps_checksums=False,
-                expected_exit_code=0,
                 expected_output="All dependencies fetched successfully",
             ),
             ["foo"],
--- a/tests/integration/test_yarn_classic.py
+++ b/tests/integration/test_yarn_classic.py
@@ -4,6 +4,8 @@ from pathlib import Path
 
 import pytest
 
+from hermeto.core.errors import ExitError
+
 from . import utils
 
 log = logging.getLogger(__name__)
@@ -18,7 +20,6 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "yarn"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=0,
                 expected_output="Processing the request using yarn@1.22.",
             ),
             id="yarn_classic_corepack_ignored",
@@ -29,7 +30,6 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "yarn"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=0,
                 expected_output="Processing the request using yarn@1.22.",
             ),
             id="yarn_classic_yarn_path_ignored",
@@ -40,7 +40,7 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "yarn"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=1,
+                expected_error=ExitError.ERR_PACKAGE_MANAGER,
                 expected_output='Integrity check failed for "@colors/colors"',
             ),
             id="yarn_classic_invalid_checksum",
@@ -51,7 +51,7 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "yarn"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=1,
+                expected_error=ExitError.ERR_PACKAGE_MANAGER,
                 expected_output="Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.",
             ),
             id="yarn_classic_updating_frozen_lockfile_fails",
@@ -62,7 +62,6 @@ log = logging.getLogger(__name__)
                 packages=({"path": ".", "type": "yarn"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=0,
                 expected_output="All dependencies fetched successfully",
             ),
             id="yarn_classic_lifecycle_scripts",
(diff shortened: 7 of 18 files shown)


hermeto/core/package_managers/cargo/main.py
```diff
@@ -23,22 +28,6 @@
 log = logging.getLogger(__name__)
 
 
-class PackageWithCorruptLockfileRejected(PackageRejected):
```
a-ovchinnikov, line 46: This exception is very specific, it can happen only to cargo and only in a very special context, that is why it was defined within cargo/main.py. errors.py are meant to hold shared exception types.

hermeto/core/errors.py
```diff
@@ -1,15 +1,19 @@
 import textwrap
 from pathlib import Path
-from typing import ClassVar, Iterable
+from typing import Any, ClassVar, Iterable
 
 from hermeto import APP_NAME
 
 _argument_not_specified = "__argument_not_specified__"
 
+INVALID_USAGE_FLAG = 1 << 6
```
a-ovchinnikov: Where does this value come from? Is it inspired by `sysexits.h` or that's just a coincidence? Such things should be documented in comments and briefly in commit message.

hermeto/core/errors.py
```diff
@@ -37,15 +41,28 @@ def friendly_msg(self) -> str:
...
 class UsageError(BaseError):
     """Generic error for "Hermeto was used incorrectly." Prefer more specific errors."""
 
     is_invalid_usage: ClassVar[bool] = True
 
 
+class CliParameterError(BaseError):
+    """A parameter passed to the CLI was invalid."""
+
+    _error_code = 2
```
a-ovchinnikov: How are these codes assigned? Do they have any meaning or are they just arbitrary numbers? Why 2 is repeated twice and other numbers are apparently unique? I would have suggested automating error numbers assignment by means of a stateful decorator, but the codes are user-visible and should better remain stable, any automatic solution will be either prone to accidents due to code reshuffling or too complicated.

hermeto/core/errors.py
```diff
@@ -37,15 +41,28 @@ def friendly_msg(self) -> str:
             msg += f"\n{textwrap.indent(self.solution, prefix='  ')}"
         return msg
 
+    @classmethod
+    def exit_code(cls) -> int:
```
a-ovchinnikov: Optional suggestion: making this into a property-like entity could be a fun exercise (optional because that would require messing with metaclass since this is a classmethod). Just consider how more aesthetically pleasing would a point of call look:
```diff
- raise typer.Exit(e.exit_code())
+ raise typer.Exit(e.exit_code)
```

hermeto/core/errors.py
```diff
@@ -341,6 +376,44 @@ def friendly_msg(self) -> str:
...
+        reason = f"{package_path} not a Yarn v1"
+        super().__init__(reason, solution=None)
+
+    _error_code = 20
+
+
+class UnsatisfiableArchitectureFilter(PackageRejected):
+    """RPM architecture filter constraints cannot be satisfied by lockfile architectures."""
+
+    _error_code = 21
```
a-ovchinnikov: What will happen when someone eventually adds 64th distinct exception?

tests/integration/utils.py
```diff
@@ -420,13 +421,15 @@ def fetch_deps_and_check_output(
         entrypoint=entrypoint,
         podman_flags=podman_flags,
     )
-    assert exit_code == test_params.expected_exit_code, (
+    expected_exit_code = test_params.expected_error.exit_code() if test_params.expected_error else 0
```
a-ovchinnikov: PEP8 nitpick: since expected_error could be `None` you should explicitly check if it is `None` or not.

tests/integration/test_yarn_classic.py
```diff
@@ -107,8 +111,7 @@ def test_yarn_classic_packages(
             utils.TestParameters(
                 branch="yarn-classic/e2e",
                 packages=({"path": ".", "type": "yarn"},),
-                expected_exit_code=0,
-                expected_output="All dependencies fetched successfully",
+                expected_error=None,
```
a-ovchinnikov: expected_error defaults to `None`. WDYT about not specifying it unless it is something other than `None`?

hermeto/core/errors.py
```diff
@@ -74,6 +92,8 @@ def __init__(
 class InvalidInput(UsageError):
     """User input was invalid."""
 
+    _error_code = 3
```
a-ovchinnikov: Since these codes are very user-visible they must be documented.

hermeto/core/package_managers/cargo/main.py
ST3V1K: I was worried we’d forget about the error types defined outside errors.py (e.g. in package-manager modules) and end up reusing the same exit code for different errors. 
By centralizing the error→exit-code mapping and adding checks for duplicate codes and for unregistered error classes, that risk is much lower now.

hermeto/core/errors.py
ST3V1K: Any overlap with `sysexits.h` is coincidental. I kept exit codes in the 0-127 range because, as I understand from this page [1], 128-255 are reserved for signaled exits.

> What will happen when someone eventually adds 64th distinct exception?

But because of this, I removed the flag and the maximum errors we can now have (without duplicate exit codes) is 127.

[1] https://tldp.org/LDP/abs/html/exitcodes.html

docs/exit-codes.md
```diff
@@ -0,0 +1,77 @@
...
+| 15 | `FetchError` | Failed to fetch data or perform a network operation. |
+| 16 | `PackageManagerError` | The package manager reported an error or could not complete its task. |
+| 17 | `GitError` | Generic error raised when a Git command fails. |
+| 18 | `GitRemoteNotFoundError` | The specified Git remote does not exist in this repository. |
+| 19 | `GitInvalidRevisionError` | The given revision identifier is not valid for this repository. |
+| 20 | `PackageWithCorruptLockfileRejected` | Lockfile for this package is corrupted and cannot be used. |
+| 21 | `UnsatisfiableArchitectureFilter` | Unable to satisfy the architecture constraints specified for RPM packages. |
+| 22 | `NotV1Lockfile` | The Yarn Classic lockfile is not version 1 and cannot be used. |
+
+## Usage in scripts and CI
```
a-ovchinnikov: I would suggest dropping this entire section on the grounds of exit codes being a commonly understood mechanism.

docs/exit-codes.md
```diff
@@ -0,0 +1,77 @@
+# Exit codes
+
+Hermeto uses **stable exit codes per error type** so that scripts and CI can
+detect failure reasons without parsing stderr. The CLI exits with `0` on success
+and with one of the codes below on failure.
+
+| Code | Error type | Description |
+| ---- | ---------- | ----------- |
```
a-ovchinnikov: Error type is an implementation detail, it does not belong to user-facing docs.

docs/exit-codes.md
```diff
@@ -0,0 +1,77 @@
...
+| 1 | `BaseError` | Root of the error hierarchy. Don't raise this directly, usemore specific error types. |
+| 2 | `UsageError` | Generic error for "Hermeto was used incorrectly." Prefer more specific errors. |
+| 3 | `PathOutsideRoot` | After joining a subpath, the result is outside the root of a rooted path. |
```
a-ovchinnikov: Note, that class docstrings as they are now are meant for developers, not end users. Some of them will make very little sense when taken out of context, just like the line above. Is this something that a user can fix? Is this something they must report as a bug? It is hard to tell.

hermeto/core/errors.py
```diff
@@ -6,11 +6,47 @@
...
+    "FetchError": 15,
+    "PackageManagerError": 16,
+    "GitError": 17,
+    "GitRemoteNotFoundError": 18,
+    "GitInvalidRevisionError": 19,
+    "PackageWithCorruptLockfileRejected": 20,  # hermeto/core/package_managers/cargo/main.py::PackageWithCorruptLockfileRejected
+    "UnsatisfiableArchitectureFilter": 21,  # hermeto/core/package_managers/rpm/binary_filters.py::UnsatisfiableArchitectureFilter
+    "NotV1Lockfile": 22,  # hermeto/core/package_managers/yarn_classic/main.py::NotV1Lockfile
+}
+if len(_exit_codes) != len(set(_exit_codes.values())):
```
a-ovchinnikov: My concern here is that we will be dynamically checking static data. I would strongly recommend against this check.

hermeto/core/errors.py
```diff
@@ -6,11 +6,47 @@
...
+    raise ValueError("Duplicate exit codes found")
+
+
+def get_error_name_from_code(code: int) -> str | None:
```
a-ovchinnikov: Depending on usage pattern for this function it might benefit from `@functools.cache` decorator.

_Edit_: an alternative solution is to construct an inverse mapping once:
```python
_codes_to_names = dict(zip(_exit_codes.values(), _exit_codes.keys()))
```

ST3V1K: Added constructing the inverse mapping once.

docs/exit-codes.md
```diff
@@ -0,0 +1,30 @@
...
+Hermeto uses **stable exit codes per error type** so that scripts and CI can
+detect failure reasons without parsing stderr. The CLI exits with `0` on success
+and with one of the codes below on failure.
+
+| Code | Description |
+| ---- | ----------- |
+| 1 | An unexpected system error occurred. Contact the maintainers. |
+| 2 | The command usage was incorrect. Review your input syntax. |
+| 3 | The file path sits outside the allowed directory. Ensure files remain within the source or output directories. |
+| 4 | The input arguments are invalid. Check your arguments. |
```
a-ovchinnikov: How does this differ from 2?

docs/exit-codes.md
```diff
@@ -0,0 +1,30 @@
...
+| 5 | The package was rejected. Check package validity. If the error persists, contact the maintainers. |
```
a-ovchinnikov: It is a bit unclear which maintainers we are suggesting to contact here, package maintainers or Hermeto maintainers?

docs/exit-codes.md
```diff
@@ -0,0 +1,30 @@
...
+| 6 | The package is not a Git repository. Ensure the directory is a valid Git repository. |
```
a-ovchinnikov: This error demonstrates the problem I have with this fine-grained approach. Suppose it has somehow happened when a user was working with hermeto. They will simultaneously know too much and too little: first, they will know that something is not a git repo which is a lot of details, second, they will not know what exactly is not a git repo. There is no space in an error code to point a user to whatever broke down, thus it looks like an exposure of inner workings of the tool (which is usually a bad idea).

The exception could get raised in `get_repo_id()` which is used by most if not all PMs when processing dependencies, one would definitely need to consult logs to figure out what went wrong. Does it really make sense to expose a user to this information? (The same question holds for all other specialized exception, e.g. for 7: which file? Can a user tell at a glance? What about from the error code alone? If not then is it really necessary?) I would strongly recommend re-reading this documentation change as a user: pretend that you know nothing about the project's internals and see if this would make any sense to you.

docs/exit-codes.md
```diff
@@ -0,0 +1,30 @@
...
+| 7 | The file format is incorrect. Verify the file structure. If the issue persists, contact the maintainers. |
+| 8 | The feature is unavailable. Contact the maintainers to request this feature. |
+| 9 | A required executable is missing. Install the executable in your PATH. |
+| 10 | The file verification failed. Verify the file is not corrupt and checksums are correct. |
+| 11 | The checksum data is invalid. Ensure the checksum matches the expected format. |
+| 12 | The integrity checksum is missing. Ensure the checksum exists. |
+| 13 | A required lockfile is missing. Ensure the file exists and is checked into the repository. |
+| 14 | The lockfile format is invalid. Check the syntax in the lockfile. |
+| 15 | The system failed to fetch data. Try again. If the issue persists, contact the maintainers. |
```
a-ovchinnikov: This one, by the way, _might_ deserve both a dedicated error code and a mention in docs since it likely happens due to network errors.

a-ovchinnikov (commented):
The code part LGTM, but I have serious doubts regarding the way it is documented. I am also not sure if we want that many distinct error codes. I can see the benefit of such granularity for testing, but from a user's perspective this looks confusing. At the very least I'd recommend grouping most of the codes together and naming the groups in the vein of "dependency format errors" or "dependency lockfile problems" without going into specifics in the table and suggesting inspecting logs instead.

docs/exit-codes.md
ST3V1K: 2 is usually directly from Typer app - wrong command, subcommand, unknown option...
4 is from Hermeto when there was an error with argument but it is valid in Typer.

docs/exit-codes.md
ST3V1K: Rewrote the table into groups of similar errors and point users to output/logs.

docs/exit-codes.md
```diff
@@ -0,0 +1,71 @@
...
+
+The command was used incorrectly. Wrong syntax, subcommand or argument.
+
+Review your input and the documentation.
+
+## Paths and files
+
+**Codes:** 3, 7, 9, 10, 11, 12
+
+A path sits outside the allowed directory, a file format is wrong, a required
```
a-ovchinnikov: This has the same problem as the original table: which path? which file has a wrong format? what is missing? It is physically impossible to pack all this data into eight bits of exit code alone. Imagine a regular user seeing an exit code of 9 for the first time and then consulting this page. Will it help them? I doubt that. I am kind of ok with creating multiple exit codes for internal purposes, but I don't find partially exposing them to users a good idea.

Maybe this document could be restructured to slightly shift focus off individual codes? What I mean is something along these lines:

```
Hermeto uses a rich set of exit codes which can signal different internal
conditions. The codes are described in a table below. Please note, that 
the codes are not meant to be used alone, they are intended to be used as 
merely an aid for automated systems which were found to sometimes benefit
from such set. Most errors require additional information to be corrected,
this information is always saved to logs.

| Group | Code           | Meaning
+-------+----------------+---------
|     1 |              0 | No errors
|     2 |              1 | Unexpected internal error
|     3 |           2, 4 | Hermeto CLI usage error
|     4 | 3, 7, 9, 10-12 | Filesystem error
....

Recommended course of action for each group:

| Group | Action
+-------+--------
|     1 | Do nothing, the was no error
|     2 | Collect run logs and stderr, share them with Hermeto maintainers
|     3 | Verify commands you issued to Hermeto
|     4 | Refer to logs to find out which filesystem object caused the trouble
...
```

Note, that I moved the mention of logs from the bottom of the document closer to the top, the way the document is written now a user would need to go through all codes first and only then read a very important bit of information. (To be completely transparent I wrote my proposal first and discovered that you actually mentioned the logs only after that. I believe a user reaching out for error code reference will be in a hurry, possibly frustrated and even less intending to carefully read the document top to bottom.) I also switched back to your original format of tables, just slightly adjusted, I like it more because it makes using this document much easier. I think the goal of this document is to provide a quick reference, and tables allow one to get important information at a glance. And as always, this is not the ultimate truth, comments and counterarguments are welcome!

ST3V1K: Thanks for the detailed feedback! I've pushed an update addressing your suggestions. Hopefully this is THE version :)

hermeto/core/errors.py
```diff
@@ -7,11 +7,46 @@
...
+    "GitRemoteNotFoundError": 18,
+    "GitInvalidRevisionError": 19,
+    "PackageWithCorruptLockfileRejected": 20,  # hermeto/core/package_managers/cargo/main.py::PackageWithCorruptLockfileRejected
+    "UnsatisfiableArchitectureFilter": 21,  # hermeto/core/package_managers/rpm/binary_filters.py::UnsatisfiableArchitectureFilter
+    "NotV1Lockfile": 22,  # hermeto/core/package_managers/yarn_classic/main.py::NotV1Lockfile
+}
+_codes_to_names: dict[int, str] = dict(zip(_exit_codes.values(), _exit_codes.keys()))
+
+
+def get_error_name_from_code(code: int) -> str | None:
```
a-ovchinnikov: This function can be safely dropped now, all it does is wrapping dict.__getitem__

hermeto/core/errors.py
```diff
@@ -31,6 +66,13 @@ def __init__(
         else:
             self.solution = solution
 
+    def __init_subclass__(cls) -> None:
+        class_name = cls.__name__
+        if class_name not in _exit_codes:
+            raise ValueError(f"No exit code found for {class_name}")
```
a-ovchinnikov: I wonder if it would make sense to add an `InternalLogicError` as a subclass of `ValueError` with only purpose of highlighting that we forgot to add a class to the table?

hermeto/core/errors.py
a-ovchinnikov, line 1: _Commit_ https://github.com/hermetoproject/hermeto/commit/9f2376a3bd20d012fab6dfbd84dbe4d48980330e: I think it should be split in at least two: one will introduce the new exit codes, another one will make use of them. An even smoother flow could be achieved by separating the introduction of the new mechanism of checking a scenario from updating scenarios and finally dropping outdated checks in yet another commit.

docs/exit-codes.md
```diff
@@ -0,0 +1,36 @@
...
+| Group | Code                       | Meaning                      |
+| ----- | -------------------------- | ---------------------------- |
+| 1     | 0                          | No errors                    |
+| 2     | 1                          | Unexpected internal error    |
+| 3     | 2, 4                       | Hermeto CLI usage error      |
+| 4     | 3, 7, 9&ndash;12           | Filesystem error             |
+| 5     | 5, 13, 14, 16, 20&ndash;22 | Packages and lockfiles       |
+| 6     | 6, 17&ndash;19             | Git error                    |
+| 7     | 15                         | Network / fetch error        |
+| 8     | 8                          | Unsupported feature          |
```
a-ovchinnikov: A minor style nitpick: numbers are usually right-aligned when in tables (text is left-aligned though).

docs/exit-codes.md
```diff
@@ -0,0 +1,36 @@
...
+Most errors require additional information to be corrected, this information is
+always saved to logs.
+
+| Group | Code                       | Meaning                      |
+| ----- | -------------------------- | ---------------------------- |
+| 1     | 0                          | No errors                    |
+| 2     | 1                          | Unexpected internal error    |
+| 3     | 2, 4                       | Hermeto CLI usage error      |
+| 4     | 3, 7, 9&ndash;12           | Filesystem error             |
+| 5     | 5, 13, 14, 16, 20&ndash;22 | Packages and lockfiles       |
```
a-ovchinnikov: Nitpitck: s/Packages and lockfiles/Packages or lockfiles error/ to keep it aligned with other rows.

tests/integration/test_cargo.py
a-ovchinnikov, line 1: This change belongs in the previous commit

a-ovchinnikov (commented):
LGTM in general, please find several suggestions about structuring the PR inline.

hermeto/core/errors.py
ST3V1K: After that the `_codes_to_names` dict would have to be "public"

hermeto/core/errors.py
ST3V1K: I don't think it makes much sense only for this. But there might some other use case for this in the future so I added it.

hermeto/core/errors.py
a-ovchinnikov: Yes, and that is fine when one can be reasonable sure that future maintainers will think about what they are doing. One day in the distant future someone might even change this to a [frozendict](https://peps.python.org/pep-0814/)

hermeto/core/errors.py
```diff
@@ -7,11 +7,53 @@
...
+    :param code: Exit code (e.g. from a process or from BaseError.exit_code).
+    :return: The corresponding error class name, or None if the code is not registered.
+    """
+    return _codes_to_names.get(code)
+
+
+class InternalLogicError(ValueError):
+    """An error that should not happen during normal operation.
+
+    Used only to signal a bug in the code."""
```
a-ovchinnikov: Wording: this is not about signaling a bug, but rather that code integrity was accidentally broken by a change, i.e. that one part of code does not follow from another one anymore. Maybe "to signal inconsistency in code"? This exception could be useful in all cases when we raise on reaching a branch that should be unreachable.

hermeto/core/errors.py
```diff
@@ -7,11 +7,53 @@
...
+def get_error_name_from_code(code: int) -> str | None:
+    """Return the error class name for the given exit code.
+
+    :param code: Exit code (e.g. from a process or from BaseError.exit_code).
+    :return: The corresponding error class name, or None if the code is not registered.
+    """
+    return _codes_to_names.get(code)
+
+
+class InternalLogicError(ValueError):
```
a-ovchinnikov: Since this is an exception which theoretically could happen and propagate all the way to the user would it need an error code too?

hermeto/interface/cli.py
a-ovchinnikov, line 1: _Commit_message_ in https://github.com/hermetoproject/hermeto/commit/27a1b11c4143cc324b5541f47a2e3a13c51207ac: s/excepected/expected/

a-ovchinnikov (commented):
LGTM in general, just a couple small inline suggestions.

hermeto/interface/cli.py
ST3V1K: done

hermeto/core/errors.py
ST3V1K: done

hermeto/core/errors.py
ST3V1K: It technically exits with an exit code 1 as it is the default one and 1 is already some form of unexpected error for the user. But if we want to specify the error code, it would have to extends the `BaseError` class and not `ValueError`. Then it would have to be in the error codes dictionary and would need a unique code.

a-ovchinnikov (approved):

eskultety:
Please wait for my review, I'd like to have a look.

mkdocs.yml
```diff
@@ -35,6 +35,7 @@ theme:
 nav:
   - Home: index.md
   - Usage: usage.md
+  - Exit codes: exit-codes.md
```
slimreaper35: I am not sure if the exit codes fit the other website content.

hermeto/core/errors.py
```diff
@@ -7,11 +7,54 @@
 
 _argument_not_specified = "__argument_not_specified__"
```
eskultety, line 9: **Commit message:**
`s/Assited-by/Assisted-by`

The trailer is probably worth automating ;) .

hermeto/core/errors.py
```diff
@@ -7,11 +7,54 @@
...
+    "InvalidLockfileFormat": 14,
+    "FetchError": 15,
+    "PackageManagerError": 16,
+    "GitError": 17,
+    "GitRemoteNotFoundError": 18,
+    "GitInvalidRevisionError": 19,
+    "PackageWithCorruptLockfileRejected": 20,  # hermeto/core/package_managers/cargo/main.py::PackageWithCorruptLockfileRejected
+    "UnsatisfiableArchitectureFilter": 21,  # hermeto/core/package_managers/rpm/binary_filters.py::UnsatisfiableArchitectureFilter
+    "NotV1Lockfile": 22,  # hermeto/core/package_managers/yarn_classic/main.py::NotV1Lockfile
+}
```
eskultety: Having a string-keyed dictionary with class names mapped to integers this way is fragile:

- any class rename will immediately break this and will result in an `InternalLogicError`
- there's no check for duplicate error code values

Since these will be now public API (i.e. not just for our test usage although that was the initial motivation), the solution must be modeled with `IntEnum` which prevents the above by design. One more thing, when you do that, please make sure to avoid using [`auto` ](https://docs.python.org/3/library/enum.html#enum.auto) because that would risk breaking ABI if a developer would add a new constant somewhere in the middle of the enum list.

I'd also advise to choose a consistent naming for the codes (e.g. using `ERR_` as prefix) to make it easy to grep the code base for an error you just can't remember the name of using a simple pattern to narrow down the matching space significantly :)

hermeto/core/errors.py
```diff
@@ -7,11 +7,54 @@
 
 _argument_not_specified = "__argument_not_specified__"
 
+_exit_codes: dict[str, int] = {
```
eskultety: Since this doesn't explicitly list 0 as success (which is good), then `error_codes` is IMO a better fitting name. Also, we'll be importing this directly from other places, so should be a public member.

hermeto/core/errors.py
eskultety: This function won't be needed at all when we model this around Enum.

hermeto/core/errors.py
```diff
@@ -31,6 +74,13 @@ def __init__(
         else:
             self.solution = solution
 
+    def __init_subclass__(cls) -> None:
+        class_name = cls.__name__
+        if class_name not in _exit_codes:
+            raise InternalLogicError(f"No exit code found for {class_name}")
```
eskultety: While I can definitely imagine a generic error like this in the future, its current single-use scenario begs the question - what value does a custom exception bring over a plain `assert` or `RuntimeError` if this is supposed to be a fatal initialization error?
I have a slight preference for `RuntimeError` since asserts can be disabled on python CLI and with this particular scenario I'm confident that it would never appear in production because the CI would always catch it. That said, I can definitely live with an assert as well and we could then drop it by tweaking our Dockerfile build recipe.

hermeto/core/errors.py
```diff
@@ -7,11 +7,54 @@
...
+
+    Used to signal inconsistency in code. Raise this when reaching a branch
+    that should be unreachable (e.g. after a change)."""
+
 
 class BaseError(Exception):
     """Root of the error hierarchy. Don't raise this directly, use more specific error types."""
 
     is_invalid_usage: ClassVar[bool] = False
+    exit_code: ClassVar[int] = 1
```
eskultety: I'd also mandate that each class explicitly defines its exit code rather than trying to derive it from its name (a name can change, that's an internal impl. detail)

hermeto/core/errors.py
```diff
@@ -31,6 +74,13 @@ def __init__(
         else:
             self.solution = solution
 
+    def __init_subclass__(cls) -> None:
+        class_name = cls.__name__
+        if class_name not in _exit_codes:
+            raise InternalLogicError(f"No exit code found for {class_name}")
+        cls.exit_code = _exit_codes[class_name]
+        super().__init_subclass__()
+
```
eskultety, line 103: This `__init_subclass__` is a nice pattern! First time I'm seeing it. I'd extend this in the following way to catch or handle more scenarios as this currently only validates explicit declarations:

1. Allow child exceptions to simply inherit parent's code, e.g. `BaseError`, that is a legitimate use case for extremely tailored module-level exceptions that won't even propagate to the user, yet have to subclass `BaseError` rather than `ValueError` or similar.
2. Check the error code membership in the Enum list
3. Decline an explicit error code duplication, i.e. if a class doesn't decide to inherit from its parent and also not to define its own code, that smells and probably wasn't intended and so we should flag it.

hermeto/core/errors.py
```diff
@@ -46,7 +96,7 @@ class UsageError(BaseError):
 
 
 class PathOutsideRoot(UsageError):
-    """Afer joining a subpath, the result is outside the root of a rooted path."""
+    """After joining a subpath, the result is outside the root of a rooted path."""
```
eskultety: Unrelated change, but since you'll be updating these with enum codes, you can reserve this hunk for that particular change.

tests/integration/test_bundler.py
```diff
@@ -4,6 +4,8 @@
 
 import pytest
 
+from hermeto.core.errors import LockfileNotFound, PackageManagerError
+
```
eskultety: I don't think this is the expected outcome of this effort - importing particular errors requires listing the correct symbols which is more accentuated when you import an error from a particular backend module rather than from a centralized error location, that I'd say is an anti-pattern. With enums, you only ever import a single Enum class - clear, neat & consistent. :)

tests/integration/test_bundler.py
```diff
@@ -18,8 +20,7 @@
                 packages=({"path": ".", "type": "bundler"},),
                 check_output=False,
                 check_deps_checksums=False,
-                expected_exit_code=2,
-                expected_output="Required files not found:",
+                expected_error=LockfileNotFound,
```
eskultety: We should still be expecting an error code, not a particular error.

docs/exit-codes.md
eskultety: I beg to differ on this docs composition, I think simplicity is key here and would require less maintenance from our side. Let me expand on that.

1. Grouping error codes on its own is not a problem, but it's IMO only a tiny visual aid that requires manual updates to keep the docs up-to-date, that's maintenance burden and easy to forget due to human factor. The problem with grouping this way occurs the moment you introduce a bunch of unrelated exceptions and then you need to add a new exception to an existing category: you need to remember to update the list of codes in the table and do it manually (e.g.  `3, 7, 9, 10-12, 17-18, 25`). That would IMO be sort of a docs anti-pattern, because I believe that API related docs, should be mostly or fully generated.
2. We should keep grouping, but the group numbers don't map to anything and so they IMO divert user's focus to something that cannot be even caught in their scripts, so groupings could remain only as categories, duplicated across a flat table of codes
3. >  I am kind of ok with creating multiple exit codes for internal purposes, but I don't find partially exposing them to users a good idea.

In general, very true, but at the same time, who reads API pages top-to-bottom by scrolling the page? Probably noone, so while we should be careful about what  exit codes we actually create, dumping a simple flat table of all exit codes we use is both user friendly - Ctrl-F is already as convenient as it can be :) - and automatable.

4. Combining all: error codes, error category names, meaning, and actions in a single flat table would allow us to not only generate it within docs, but also programatically sort the codes by any key we see fit.
5. Automation - there's a [mkdocs macros plugin ](https://mkdocs-macros-plugin.readthedocs.io/en/latest/) that would allow us to create a simple python hook that would populate a placeholder in the markdown with a table of codes sorted to our liking, containing all the information we need.

eskultety (changes requested):
The changes I'm requesting here revolve solely around the fact that the moment we merge and release this it becomes both a binding API contract (through documenting this in user facing docs) AND ABI contract (users handling specific codes in their scripts) that we're stuck with until Hermeto v1.0 so we need to approach this extra carefully. Needless to say this is still an extremely important work to do :) !

ST3V1K:
I originally had the `ExitError` enum register a tuple with each error's category and meaning, but that info belongs on the error class itself. Moving it there means the docs build needs to import those classes. Since it only picks up what's in errors.py, all user-facing errors now live there. Internal errors that never propagate to the user can still be defined elsewhere.

mkdocs.yml
ST3V1K: I'm not really sure where else to put it.

First idea is to have it directly in the Usage section
Second is to have like a Reference section at the end
```diff
nav:                                                                                                                                                                                                              
   - Home: index.md                                                                                                                                                                                               
   - Usage: usage.md                                                                                                                                                                                              
-  - Exit codes: exit-codes.md                                                                                                                                                                                    
   - Package managers:                                                                                                                                                                                            
       - bundler: bundler.md
       - cargo: cargo.md                                                                                                                                                                                          
       - generic: generic.md
       - gomod: gomod.md                                                                                                                                                                                          
       - npm: npm.md
       - pip: pip.md
       - rpm: rpm.md                                                                                                                                                                                              
       - yarn classic: yarn_classic.md                                                                                                                                                                            
       - yarn berry: yarn.md                                                                                                                                                                                      
+  - Reference:                                                                                                                                                                                                   
+      - Exit codes: exit-codes.md
```

slimreaper35: I think the usage is fine for this.

> I am not sure if the exit codes fit the other website content.

I didn't mean the actual content, but the item in the navbar.

hermeto/core/errors.py
a-ovchinnikov: Not to object to the rewrite itself, but are we sure we won't want or need to reuse some of the exit codes? Right now every exception has its own code, imagine we add a dozen more PMs, won't we run out of exit codes at some point? I also don't think it is a good idea to make an impression that this is something that a user would want to use: this is an internal signaling mechanism after all, all user should care about is zero vs non-zero values, maybe this should be stressed out more in the docs change?

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov, line 1: _Commit message_ for https://github.com/hermetoproject/hermeto/commit/8d468c0fd7bef3c1a9a40b09025deb78fd638c52 : is `__init_subclass__`  still used directly anywhere?

hermeto/core/errors.py
```diff
@@ -1,18 +1,92 @@
...
+    ) -> None:
+        """Initialize ErrorRegistryMeta.
+
+        :param name: name of the error class
+        :param bases: base classes of the error class
+        :param namespace: namespace of the error class
+        :param kwargs: keyword arguments
+        """
+        super().__init__(name, bases, namespace, **kwargs)
+        if "_exit_error" in namespace:
```
a-ovchinnikov, line 65: What if `_exit_error` is not in namespace? Are we OK with having exceptions without a defined error code? In both cases this is something that should be mentioned in the docstring or in a comment.

hermeto/core/errors.py
eskultety: > Not to object to the rewrite itself, but are we sure we won't want or need to reuse some of the exit codes?

In my previous review I asked for support for exit_code reuse through inheritance, so rather than reusing an exit code mistakenly, we'd pull it in through subclassing the desired error class (which will further subclass BaseError). This is just my POV on the design as I'm trying to avoid the scenario where someone reuses an exit code during POC implementation and nobody notices during review. Subclassing a particular error "may" (not will) trigger a receptor during code review.

> Right now every exception has its own code

The plan isn't to have this many public exceptions. The current list is what is propagated to the user. Subclassing BaseError or a different error without an explicit exit code (inherit the parent) should be a supported pattern per my review comments.

> won't we run out of exit codes at some point?

Theoretically, but 256 values should give us plenty of headroom until the next major version where we can revisit this topic and re-structure if we conclude this was a bad decision.

> this is an internal signaling mechanism after all, all user should care about is zero vs non-zero values

I admit this was my expectation initially when I filed the issue. During the first rounds of review, however, there was a request to document this for the user. The moment you provide this information to the user, while informative, you immediately make it a public API IMO. I am okay both ways, my review was meant to be ON TOP of whatever was already suggested to not go against other reviewers. That said, I can imagine users reacting to particular exit codes in particular ways in their CI/CD solutions to attempt to recover from some errors (however unlikely).

@a-ovchinnikov if you feel strongly about keeping this solely as an internal signaling mechanism, that is fine with me, however, in that case we should not document it in the user docs, because I don't think the user should care about those codes at all, like you said it's either 0 or error, so please state your preference :) .

hermeto/core/errors.py
eskultety: We should avoid crashes. Exceptions without codes are fine, but those really should not be propagated to the user, e.g. if you subclass `ValueError` with no other overrides just to use it as an intermediary signal somewhere in the code where it's handled appropriately.

hermeto/core/package_managers/cargo/main.py
ST3V1K: I forgot to change it. Thanks

mkdocs.yml
ST3V1K: Like this?

hermeto/core/errors.py
ST3V1K: > What if _exit_error is not in namespace? Are we OK with having exceptions without a defined error code? In both cases this is something that should be mentioned in the docstring or in a comment.

I have added a comment to clarify that

hermeto/core/errors.py
a-ovchinnikov: My thinking was along these lines: there is some user-observable behavior, in order to ensure that users do not ask whether it is intentional or if something is leaking we should mention it in the docs, but at the same time warn users that this is not a reliable mechanism for them to determine the exact nature of a failure. Being explicit about admitting this behavior and at the same time not recommending it would serve another purpose: to serve as a warning for those users who prefer ad-hoc structures in their pipelines and who empirically find that some code corresponds to a certain failure condition, build some automation around that and then discover that after an update to Hermeto it does not work anymore. Maybe the table in the docs is not needed after all, I do not remember if it was my idea to add it or not, if yes then I am sorry for the confusion. I definitely did not intend it to become a binding contract.

mkdocs.yml
slimreaper35: No, just have a text reference to `exit-codes.md` file.

hermeto/core/errors.py
ST3V1K: Undid this change

hermeto/core/errors.py
ST3V1K: I have added a warning to the exit code docs about them NOT being a public API and that they can change

mkdocs.yml
ST3V1K: > I didn't mean the actual content, but the item in the navbar.

Sorry, you confused me with this :)

hermeto/core/errors.py
```diff
@@ -11,7 +11,7 @@
 
 @unique
 class ExitError(IntEnum):
-    """Stable Hermeto process exit codes (public API).
+    """Hermeto process exit codes.
```
eskultety: This hunk should be squashed to the previous commit.

tests/integration/utils.py
```diff
@@ -468,13 +469,26 @@ def fetch_deps_and_check_output(
...
+    def _fmt(code: int) -> str:
+        if code == 0:
+            return "0"
+        try:
+            return f"{code} ({ExitError(code).name})"
+        except ValueError:
+            return f"{code} (unknown exit code)"
+
+    expected_exit_code = (
+        int(test_params.expected_error) if test_params.expected_error is not None else 0
```
eskultety: nitpick: If you simply add `ERR_OK` to the enum and make that the default instead of None you can get rid of this line.

docs/exit-codes.md
```diff
@@ -0,0 +1,16 @@
...
+    These codes are **not a stable API** and may change between releases.
+    Do not build automation that relies on specific exit code values. For
+    error handling, treat any non-zero exit code as a failure and consult
+    the logs for details.
+
+{{ exit_codes_table }}
+
+For bugs or unsupported features, [contact the Hermeto maintainers][].
+
+[contact the Hermeto maintainers]: https://github.com/hermetoproject/hermeto/issues
```
eskultety: Based on recent comments by @a-ovchinnikov and myself, we concluded that it's clear we don't desire making this public API after all, this piece of docs is no longer needed (except for the usage.md bit) and all related changes can be dropped. Nice experiment though, we definitely converged to something neat even in the docs! ;)

docs/usage.md
```diff
@@ -260,6 +262,12 @@ could be to manually create an internal network (but you'll need root
 privileges): `sudo podman network create --internal isolated-network; sudo
 podman build --network isolated-network ...`.
 
+## Exit codes
+
+Hermeto uses different exit codes to signal different error conditions. These
+codes are informational and may change between releases. See the
+[exit codes](exit-codes.md) page for more details.
```
eskultety: I would keep this bit in `usage.md`! Let's reword slightly to:
_Hermeto uses a set of exit codes to signal different error conditions. These are internal only and serve informational purposes and hence may change in between releases, please do NOT depend on them!_

hermeto/core/package_managers/cargo/main.py
```diff
@@ -7,7 +7,7 @@
 from dataclasses import dataclass
 from functools import cached_property
 from pathlib import Path
-from typing import Any, ClassVar, NamedTuple
```
eskultety: Refresh my mind please, do we still need this patch if we don't need to generate the exit codes table?

hermeto/core/errors.py
```diff
@@ -1,18 +1,100 @@
...
+        propagated to the user.
+
+        :param name: name of the error class
+        :param bases: base classes of the error class
+        :param namespace: namespace of the error class
+        :param kwargs: keyword arguments
+        """
+        super().__init__(name, bases, namespace, **kwargs)
+        # Subclasses without _exit_error inherit their parent's code and
+        # are not registered; they serve as internal signals only
```
eskultety: _they serve as internal signals only_

Since the conclusion was not to make these public after all, all of them are internal signals in some shape or form :), I think we can drop this bit.

hermeto/core/errors.py
```diff
@@ -1,18 +1,100 @@
...
+class ExitErrorCategory(Enum):
+    """Hermeto process exit code categories."""
+
+    INTERNAL = "Internal"
+    CLI_USAGE = "CLI usage"
+    FILESYSTEM = "Filesystem"
+    NETWORK = "Network"
+    PACKAGE = "Package"
+    GIT = "Git"
+    UNSUPPORTED = "Unsupported"
```
eskultety: Grepping the code base for category, it doesn't look like `category` is actualy used anywhere after this PR is merged which means we can drop this and leave it as an optional follow-up work if we ever feel the need for it.

hermeto/core/errors.py
ST3V1K: Done

tests/integration/utils.py
ST3V1K: Done

docs/exit-codes.md
ST3V1K: Removed

docs/usage.md
ST3V1K: Done

hermeto/core/package_managers/cargo/main.py
ST3V1K: I wanted to keep the tables and add info about them not being stable. But since we're dropping the docs about them, this can be reverted as well.

hermeto/core/errors.py
ST3V1K: Done

hermeto/core/errors.py
ST3V1K: This was used for generating the table. 
Removed

eskultety (approved):

mkdocs.yml
slimreaper35: seems like this is not valid anymore

slimreaper35 (approved):

eskultety:
@a-ovchinnikov I know you already approved in the past, do you have any last minute comments on the latest revision before we merge?

a-ovchinnikov (approved):
Still LGTM, I do not have any outstanding comments

PR #1295, opened 2026-02-13
feat(yarn): support git dependencies in Yarn Berry projects

mprahl:
# PR Description:

Adds similar functionality that already exists in the npm package manager for git dependencies by not executing any packaging code during the packaging process.

It scans yarn.lock for git-resolved dependencies, clones them as tarballs, and writes yarn resolutions so that yarn install uses the local copies instead of fetching from git. The lockfile is verified for consistency before any modifications, and `enableImmutableInstalls` is restored after the resolution update.

PURLs for overridden git deps carry a vcs_url qualifier pointing back to the original repository and commit.

Unsupported variants (patched git deps, workspace+commit selectors) are rejected with actionable error messages.

Related Jiras:
https://issues.redhat.com/browse/RHOAIENG-48192
https://issues.redhat.com/browse/AIPCC-8214
https://issues.redhat.com/browse/KFLUXSPRT-6907

# Red Hat Context

I am working on getting hermetic builds working for MLflow. It uses Yarn Berry and has some dependencies from GitHub.

Here is a reproducer with the current code:

```bash
#!/bin/bash
set -e

# Clean up any prior runs
rm -rf /tmp/mlflow-e2e /tmp/mlflow-rebuild /tmp/hermeto-output /tmp/hermeto-venv /tmp/corepack-local

# Clone the repo
git clone --depth 1 https://github.com/red-hat-data-services/mlflow.git /tmp/mlflow-e2e

# Install hermeto from the branch under review
python3 -m venv /tmp/hermeto-venv
source /tmp/hermeto-venv/bin/activate
pip install -e .

# Ensure corepack/yarn are available
npm install --prefix /tmp/corepack-local corepack
export PATH="/tmp/corepack-local/node_modules/.bin:$PATH"

# Run hermeto fetch-deps
hermeto fetch-deps \
  --source /tmp/mlflow-e2e \
  --output /tmp/hermeto-output \
  '{"type": "yarn", "path": "mlflow/server/js"}'

# Verify: apply the project files to a fresh copy and run yarn install
git clone --depth 1 https://github.com/red-hat-data-services/mlflow.git /tmp/mlflow-rebuild
python3 -c "
import json
from string import Template

bc = json.load(open('/tmp/hermeto-output/.build-config.json'))
for pf in bc['project_files']:
    path = pf['abspath'].replace('/tmp/mlflow-e2e/', '/tmp/mlflow-rebuild/')
    content = Template(pf['template']).safe_substitute(output_dir='/tmp/hermeto-output')
    open(path, 'w').write(content)
    print(f'Wrote: {path}')
"

cd /tmp/mlflow-rebuild/mlflow/server/js
YARN_ENABLE_GLOBAL_CACHE=false \
YARN_ENABLE_IMMUTABLE_CACHE=false \
YARN_ENABLE_MIRROR=true \
YARN_GLOBAL_FOLDER=/tmp/hermeto-output/deps/yarn \
  yarn install --mode skip-build
```

### DIFF
--- a/hermeto/core/package_managers/yarn/main.py
+++ b/hermeto/core/package_managers/yarn/main.py
@@ -1,13 +1,33 @@
 # SPDX-License-Identifier: GPL-3.0-only
+import copy
+import json
 import logging
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+from urllib.parse import parse_qs
 
 import semver
+import yaml
 
 from hermeto import APP_NAME
-from hermeto.core.errors import LockfileNotFound, PackageManagerError, PackageRejected
-from hermeto.core.models.input import Request
-from hermeto.core.models.output import Component, EnvironmentVariable, RequestOutput
+from hermeto.core.errors import (
+    LockfileNotFound,
+    PackageManagerError,
+    PackageRejected,
+    UnexpectedFormat,
+)
+from hermeto.core.models.input import Mode, Request
+from hermeto.core.models.output import (
+    Component,
+    EnvironmentVariable,
+    ProjectFile,
+    RequestOutput,
+)
 from hermeto.core.models.sbom import create_backend_annotation
+from hermeto.core.package_managers.js_utils import clone_repo_pack_archive
+
+from hermeto.core.package_managers.yarn.locators import parse_raw_locator
 from hermeto.core.package_managers.yarn.project import (
     Plugin,
     Project,
@@ -26,15 +46,29 @@ from hermeto.core.rooted_path import RootedPath
 log = logging.getLogger(__name__)
 
 
+@dataclass(frozen=True)
+class GitDep:
+    """A git dependency extracted from yarn.lock."""
+
+    name: str
+    clone_url: str
+    ref: str
+
+
 def fetch_yarn_source(request: Request) -> RequestOutput:
     """Process all the yarn source directories in a request."""
-    components = []
+    components: list[Component] = []
+    project_files: list[ProjectFile] = []
 
     for package in request.yarn_packages:
         path = request.source_dir.join_within_root(package.path)
         project = Project.from_source_dir(path)
 
-        components.extend(_resolve_yarn_project(project, request.output_dir))
+        pkg_components, pkg_project_files = _resolve_yarn_project(
+            project, request.output_dir, request.mode
+        )
+        components.extend(pkg_components)
+        project_files.extend(pkg_project_files)
 
     annotations = []
     if backend_annotation := create_backend_annotation(components, "yarn"):
@@ -42,7 +76,7 @@ def fetch_yarn_source(request: Request) -> RequestOutput:
     return RequestOutput.from_obj_list(
         components=components,
         environment_variables=_generate_environment_variables(),
-        project_files=[],
+        project_files=project_files,
         annotations=annotations,
     )
 
@@ -87,10 +121,9 @@ def _check_zero_installs(project: Project) -> None:
 
 
 def _check_lockfile(project: Project) -> None:
-    lockfile_filename = project.yarn_rc.get("lockfileFilename", "yarn.lock")
-    if not project.source_dir.join_within_root(lockfile_filename).path.exists():
+    if not project.lockfile_path.path.exists():
         raise LockfileNotFound(
-            files=project.source_dir.join_within_root(lockfile_filename).path,
+            files=project.lockfile_path.path,
         )
 
 
@@ -100,23 +133,62 @@ def _verify_repository(project: Project) -> None:
     _check_lockfile(project)
 
 
-def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Component]:
+def _resolve_yarn_project(
+    project: Project, output_dir: RootedPath, mode: Mode
+) -> tuple[list[Component], list[ProjectFile]]:
     """Process a request for a single yarn source directory.
 
     :param project: the directory to be processed.
     :param output_dir: the directory where the prefetched dependencies will be placed.
+    :param mode: the processing mode (strict or permissive).
+    :return: a tuple of (components, project_files)
     :raises PackageManagerError: if fetching dependencies fails
+    :raises PackageRejected: if git deps are found in strict mode
     """
     log.info(f"Fetching the yarn dependencies at the subpath {project.source_dir}")
 
     version = _configure_yarn_version(project)
     _verify_repository(project)
 
-    _set_yarnrc_configuration(project, output_dir, version)
-    packages = resolve_packages(project.source_dir)
-    _fetch_dependencies(project.source_dir)
+    git_deps = _parse_lockfile_git_deps(project)
 
-    return create_components(packages, project, output_dir)
+    if not git_deps:
+        _set_yarnrc_configuration(project, output_dir, version)
+        packages = resolve_packages(project.source_dir)
+        _fetch_dependencies(project.source_dir)
+        return create_components(packages, project, output_dir), []
+
+    # When git deps are present the processing order is intentionally inverted
+    # compared to the normal path above: we must fetch first (to update the
+    # lockfile with local tarball references) and then resolve (to parse
+    # package data from the now-modified lockfile).
+    if mode != Mode.PERMISSIVE:
+        raise PackageRejected(
+            "Git dependencies in Yarn Berry projects cannot be processed in strict mode "
+            "because the lockfile must be modified to reference local tarballs.",
+            solution="Consider whether permissive mode is acceptable for your use case.",
+        )
+
+    project_files, tarball_vcs_url_map = _clone_and_resolve_git_deps(
+        project, git_deps, output_dir
+    )
+
+    _set_yarnrc_configuration(project, output_dir, version)
+    project.yarn_rc["enableImmutableInstalls"] = False
+    project.yarn_rc.write()
+    try:
+        _fetch_dependencies(project.source_dir)
+    finally:
+        project.yarn_rc["enableImmutableInstalls"] = True
+        project.yarn_rc.write()
+
+    project_files.append(_build_lockfile_project_file(project, output_dir))
+
+    packages = resolve_packages(project.source_dir)
+    return (
+        create_components(packages, project, output_dir, tarball_vcs_url_map),
+        project_files,
+    )
 
 
 def _configure_yarn_version(project: Project) -> semver.Version:
@@ -262,3 +334,170 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir: 
         )
 
     log.info("Processing the request using yarn@%s", installed_yarn_version)
+
+
+def _parse_lockfile_git_deps(project: Project) -> list[GitDep]:
+    """Scan yarn.lock for git-resolved dependencies.
+
+    Unsupported variants (patched git deps, workspace+commit) are silently
+    skipped here and will be reported later by resolve_packages.
+
+    :param project: the Project whose yarn.lock will be scanned.
+    :return: list of GitDep instances.
+    """
+    with project.lockfile_path.path.open("r") as f:
+        lockfile_data: dict[str, Any] = yaml.safe_load(f) or {}
+
+    git_deps: list[GitDep] = []
+
+    for key, entry in lockfile_data.items():
+        if key == "__metadata":
+            continue
+
+        resolution = entry.get("resolution") if isinstance(entry, dict) else None
+        if not resolution:
+            continue
+
+        try:
+            parsed = parse_raw_locator(resolution)
+            ref = parsed.parsed_reference
+        except UnexpectedFormat:
+            log.debug("Skipping lockfile entry %r: could not parse locator", key)
+            continue
+
+        protocol = ref.protocol.removesuffix(":") if ref.protocol else None
+
+        # Skip patched git deps and workspace+commit git deps; these unsupported
+        # variants will be reported later by resolve_packages/parse_locator.
+        if protocol == "patch":
+            continue
+
+        selector_qs = parse_qs(ref.selector)
+        if "commit" not in selector_qs:
+            continue
+        if "workspace" in selector_qs:
+            continue
+
+        commit = selector_qs["commit"][0]
+        clone_url = _build_clone_url(protocol, ref.source)
+
+        name = parsed.name
+        if parsed.scope:
+            name = f"@{parsed.scope}/{name}"
+
+        git_deps.append(GitDep(name=name, clone_url=clone_url, ref=commit))
+
+    return git_deps
+
+
+def _build_clone_url(protocol: str | None, source: str | None) -> str:
+    """Build a clone-friendly URL from a Berry git locator's protocol and source.
+
+    Given protocol="https" and source="//host/path.git", returns "https://host/path.git".
+    For SCP-style locators like protocol="git@host", source="ns/repo.git", returns
+    "git@host:ns/repo.git".
+
+    The ``git+`` prefix (e.g. ``git+ssh``, ``git+https``) is stripped so the result
+    is a URL that git can clone directly and that clone_as_tarball can apply its
+    ssh-to-https fallback to correctly.
+    """
+    if not protocol or not source:
+        raise PackageRejected(
+            f"Cannot construct clone URL from protocol={protocol!r}, source={source!r}",
+            solution="Ensure the git dependency in yarn.lock has a valid URL.",
+        )
+
+    # Strip git+ prefix (e.g. git+ssh -> ssh, git+https -> https) so the URL
+    # is directly usable by git clone and the ssh-to-https fallback in clone_as_tarball.
+    protocol = protocol.removeprefix("git+")
+
+    return f"{protocol}:{source}"
+
+
+def _build_vcs_url(dep: GitDep) -> str:
+    """Build a canonical vcs_url qualifier for PURL generation."""
+    return f"git+{dep.clone_url}@{dep.ref}"
+
+
+def _clone_and_resolve_git_deps(
+    project: Project,
+    git_deps: list[GitDep],
+    output_dir: RootedPath,
+) -> tuple[list[ProjectFile], dict[str, str]]:
+    """Clone git deps, write resolutions to package.json, and return ProjectFiles.
+
+    :param project: the Project whose package.json will be modified
+    :param git_deps: list of GitDep instances
+    :param output_dir: base output directory
+    :return: tuple of (project_files, tarball_vcs_url_map) where tarball_vcs_url_map maps
+        tarball absolute path strings to vcs_url qualifiers for PURL generation.
+    :raises PackageRejected: if two git deps share the same name but different sources
+    """
+    seen_names: dict[str, tuple[str, str]] = {}
+    for dep in git_deps:
+        key = (dep.clone_url, dep.ref)
+        if dep.name in seen_names and seen_names[dep.name] != key:
+            raise PackageRejected(
+                f"Multiple git dependencies share the name '{dep.name}' but resolve to "
+                f"different sources. This cannot be expressed in a single yarn resolution.",
+                solution=(
+                    "Ensure all git dependencies with the same package name point to the same "
+                    "repository and commit."
+                ),
+            )
+        seen_names[dep.name] = key
+
+    yarn_deps_dir = output_dir.join_within_root("deps", "yarn")
+    tarball_info: dict[str, tuple[RootedPath, Path]] = {}
+    cloned_sources: dict[tuple[str, str], RootedPath] = {}
+    tarball_vcs_url_map: dict[str, str] = {}
+
+    for dep in git_deps:
+        source_key = (dep.clone_url, dep.ref)
+
+        if source_key in cloned_sources:
+            tarball_rooted = cloned_sources[source_key]
+        else:
+            tarball_rooted = clone_repo_pack_archive(dep.clone_url, dep.ref, yarn_deps_dir)
+            cloned_sources[source_key] = tarball_rooted
+
+        rel_to_output = tarball_rooted.path.relative_to(output_dir.path)
+        tarball_info[dep.name] = (tarball_rooted, rel_to_output)
+        tarball_vcs_url_map[str(tarball_rooted.path)] = _build_vcs_url(dep)
+
+    resolutions = project.package_json.data.get("resolutions", {})
+    for name, (tarball_rooted, _) in tarball_info.items():
+        resolutions[name] = f"file:{tarball_rooted.path}"
+    project.package_json["resolutions"] = resolutions
+    project.package_json.write()
+
+    template_data = copy.deepcopy(project.package_json.data)
+    for name, (_, rel_to_output) in tarball_info.items():
+        template_data["resolutions"][name] = f"file:${{output_dir}}/{rel_to_output}"
+
+    package_json_path = project.source_dir.join_within_root("package.json").path
+    project_files = [
+        ProjectFile(
+            abspath=package_json_path.resolve(),
+            template=json.dumps(template_data, indent=2) + "\n",
+        )
+    ]
+
+    return project_files, tarball_vcs_url_map
+
+
+def _build_lockfile_project_file(project: Project, output_dir: RootedPath) -> ProjectFile:
+    """Read the updated yarn.lock and build a ProjectFile with templated paths.
+
+    Replaces any occurrence of the output directory path with ${output_dir}.
+    """
+    lockfile_path = project.lockfile_path.path
+    lockfile_content = lockfile_path.read_text()
+
+    output_dir_str = str(output_dir.path)
+    lockfile_content = lockfile_content.replace(output_dir_str, "${output_dir}")
+
+    return ProjectFile(
+        abspath=lockfile_path.resolve(),
+        template=lockfile_content,
+    )
--- a/hermeto/core/package_managers/yarn/resolver.py
+++ b/hermeto/core/package_managers/yarn/resolver.py
@@ -177,9 +177,18 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
 
 
 def create_components(
-    packages: list[Package], project: Project, output_dir: RootedPath
+    packages: list[Package],
+    project: Project,
+    output_dir: RootedPath,
+    tarball_vcs_url_map: dict[str, str] | None = None,
 ) -> list[Component]:
-    """Create SBOM components for all the packages parsed from the 'yarn info' output."""
+    """Create SBOM components for all the packages parsed from the 'yarn info' output.
+
+    :param tarball_vcs_url_map: optional mapping of tarball absolute path strings to vcs_url
+        qualifiers for git deps that were redirected to local tarballs via yarn resolutions.
+        When provided, file dependencies whose locator path matches a key will use the git
+        vcs_url qualifier instead of the default file-path qualifier.
+    """
     package_mapping: dict[Locator, Package] = {}
     patch_locators: list[PatchLocator] = []
 
@@ -191,7 +200,9 @@ def create_components(
         else:
             package_mapping[package.parsed_locator] = package
 
-    component_resolver = _ComponentResolver(package_mapping, patch_locators, project, output_dir)
+    component_resolver = _ComponentResolver(
+        package_mapping, patch_locators, project, output_dir, tarball_vcs_url_map
+    )
     return [component_resolver.get_component(package) for package in package_mapping.values()]
 
 
@@ -222,10 +233,17 @@ class _ComponentResolver:
         patch_locators: list[PatchLocator],
         project: Project,
         output_dir: RootedPath,
+        tarball_vcs_url_map: dict[str, str] | None = None,
     ) -> None:
+        """Initialize the component resolver.
+
+        :param tarball_vcs_url_map: optional mapping of tarball absolute path strings to vcs_url
+            qualifiers for git deps that were redirected to local tarballs via resolutions.
+        """
         self._project = project
         self._output_dir = output_dir
         self._package_mapping = package_mapping
+        self._tarball_vcs_url_map = tarball_vcs_url_map or {}
         self._pedigree_mapping = self._get_pedigree_mapping(patch_locators)
 
     def _get_pedigree_mapping(self, patch_locators: list[PatchLocator]) -> dict[Locator, Pedigree]:
@@ -285,8 +303,7 @@ class _ComponentResolver:
             pedigree=self._pedigree_mapping.get(package.parsed_locator),
         )
 
-    @staticmethod
-    def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> str:
+    def _generate_purl_for_package(self, package: _ResolvedPackage, project: Project) -> str:
         """Create a purl for a package based on its protocol.
 
         :param package: the resolved package to be used in the purl generation.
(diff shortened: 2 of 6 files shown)


mprahl:
@slimreaper35 @eskultety @a-ovchinnikov could you please help with a PR review?

slimreaper35:
Hi @mprahl. Thank you. Currently, the team is busy with other important workstreams, so the review of this PR might be slightly delayed.

eskultety:
> Hi @mprahl. Thank you. Currently, the team is busy with other important workstreams, so the review of this PR might be slightly delayed.

I'd just add that the common practice is to ping a PR for reviews after a week or so due to no activity, exactly for the reasons mentioned by @slimreaper35.
@mprahl in the meantime, may I kindly ask you to tweak your GH profile (I suppose you contribute directly via the webUI) to include the your real name and real email? Looking at our contributing guide, we sadly didn't go into any specifics for Hermeto (during the rebranding) + you hold over 200 contributions :) that can be used as precedent. 
Until the contributing guide addresses this formality, the least you could do is to put real email into the author field, I believe each contribution with good intent (in general) to a legitimate project should be done with accountability. Lastly, there's DCO that should pretty much address ^this, but even DCO doesn't mandate real names and real emails to be used (effectively rendering the whole concept useless IMO), so there's that.

Additionally, while we appreciate the detailed context you provided, please don't link/use private resources that cannot be opened by the public (i.e. only https://issues.redhat.com/browse/AIPCC-8214 is a public JIRA dashboard).

Thank you for understanding and we'll do our best to get to your PR in a timely manner.

mprahl:
I've updated my commits to use my full name and email (Matt Prahl <mprahl@redhat.com>) as the author. I'm not sure what you mean by updating my GitHub profile.

Sorry about the private JIRA links. I thought they were all public, but I'll double check next time.

Regarding the ping, there wasn't a tool such as Prow to auto-assign a reviewer, so I tagged directly to make sure the PR was visible. Happy to follow whatever process the team prefers for that in the future.

Thanks for offering to review! I know you don't have capacity to review right now, but approving CI to run would be helpful so I can iterate on CI fixes.

tests/integration/test_yarn.py
```diff
@@ -29,7 +29,7 @@
                 check_output=False,
                 check_deps_checksums=False,
                 expected_exit_code=2,
-                expected_output="UnsupportedFeature: Found 8 unsupported dependencies, more details in the logs.",
+                expected_output="Multiple git dependencies share the name",
```
mprahl: I had to update this test because the branch used on the integration test triggers a new error that gets raised earlier. Let me know how you'd like to proceed.

ben-alkov: I know that (a least) a couple of the guys will say that you shouldn't compare strings in tests :grin:

hermeto/core/package_managers/yarn/main.py
```diff
@@ -100,23 +133,47 @@ def _verify_repository(project: Project) -> None:
...
+    git_deps = _parse_lockfile_git_deps(project)
+
+    if git_deps:
+        project_files = _clone_and_resolve_git_deps(project, git_deps, output_dir)
 
-    return create_components(packages, project, output_dir)
+        # Immutable installs must be off so yarn can update the lockfile
+        # to reference the local tarballs we wrote to package.json.
+        _set_yarnrc_configuration(project, output_dir, version)
+        project.yarn_rc["enableImmutableInstalls"] = False
```
taylormadore: At a minimum, I think this would need to be gated by Hermeto's `mode=permissive`. We won't be able to guarantee that we are downloading what is specified by the committed yarn.lock

hermeto/core/package_managers/yarn/resolver.py
```diff
@@ -317,15 +328,19 @@ def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> s
             subpath = str(workspace_path)
 
         elif isinstance(package.locator, (FileLocator, LinkLocator, PortalLocator)):
-            project_path = project.source_dir
-            workspace_path = package.locator.locator.relpath
-            package_path = package.locator.relpath
+            # Check if this file dep was originally a git dep that we overrode
+            if package.name in self._git_purl_map:
```
taylormadore: One thing to be careful of is that the `git_purl_map` is keyed by the name the parent project's lockfile gives to the dependency, but the lookup uses the name from the dependency's own package.json. These don't have to and may not match

taylormadore: Another thing is that package.name is populated by `_read_name_from_cache `for file dependencies. That function appears to assume a single package .json and so will break for a monorepo git dependency with workspaces

mprahl:
Thanks for the first pass @taylormadore! I took a stab at addressing your comments.

Also, good to see you again @ben-alkov ! :smile:

ben-alkov:
> Thanks for the first pass @taylormadore! I took a stab at addressing your comments.
> 
> Also, good to see you again @ben-alkov ! 😄

:wave: :grin:

eskultety:
> Thanks @mprahl for the contribution. I have only partially reviewed this so far. My main focus was verifying that it avoids arbitrary code execution. As far as I can tell, that is the case 👍

I can confirm:
```
[root@35754756c964 workspace]# rm -rf /tmp/js-lifecycle/ /tmp/node-compile-cache/
[root@35754756c964 workspace]# ls /tmp/
[root@35754756c964 workspace]#

[root@35754756c964 workspace]# hermeto --log-level=DEBUG fetch-deps yarn
2026-03-03 11:41:20,837 DEBUG Copying /workspace to /workspace/tmp36146nmf.hermeto-source-copy using fast in-kernel copy.
2026-03-03 11:41:20,871 DEBUG Fast copying failed, falling back to standard copy.
2026-03-03 11:41:21,103 INFO Fetching the yarn dependencies at the subpath /workspace/tmp36146nmf.hermeto-source-copy
2026-03-03 11:41:21,261 INFO Processing the request using yarn@3.6.1
2026-03-03 11:41:21,309 ERROR PackageRejected: Git dependencies in Yarn Berry projects require permissive mode because the lockfile must be modified to reference local tarballs.
Error: PackageRejected: Git dependencies in Yarn Berry projects require permissive mode because the lockfile must be modified to reference local tarballs.
  Rerun with the --mode permissive flag to process git dependencies.

[root@35754756c964 workspace]# hermeto --log-level=DEBUG --mode=permissive fetch-deps yarn
2026-03-03 11:41:55,773 DEBUG Copying /workspace to /workspace/tmpw21au0cm.hermeto-source-copy using fast in-kernel copy.
2026-03-03 11:41:55,805 DEBUG Fast copying failed, falling back to standard copy.
2026-03-03 11:41:55,980 INFO Fetching the yarn dependencies at the subpath /workspace/tmpw21au0cm.hermeto-source-copy
2026-03-03 11:41:56,102 INFO Processing the request using yarn@3.6.1
2026-03-03 11:41:56,105 DEBUG Cloning the Git repository from https://github.com/chmeliik/js-lifecycle-scripts.git
2026-03-03 11:41:58,282 DEBUG berryscary@workspace:.: reading package version from package.json
2026-03-03 11:41:58,284 DEBUG js-lifecycle-scripts@file:/workspace/hermeto-output/deps/yarn/github.com/chmeliik/js-lifecycle-scripts/js-lifecycle-scripts-external-gitcommit-c31b39bfa27355ca5583184dfa21257c54938e1a.tgz::locator=berryscary%40workspace%3A.: reading package name from deps/yarn/cache/js-lifecycle-scripts-file-fa88504cdd-8.zip
2026-03-03 11:41:58,359 INFO All dependencies fetched successfully \o/

[root@35754756c964 workspace]# ls /tmp/
node-compile-cache
[root@35754756c964 workspace]# strings /tmp/node-compile-cache/*/bdc67299
@yarnpkg/plugin-exec
RaRsQ
factory
module
.factory`j
ad!Q
/workspace/tmpw21au0cm.hermeto-source-copy/.yarn/plugins/@yarnpkg/plugin-exec.cjs
require
__filename
__dirname
 SdC0

==================== HERMETIC BUILD =======================
[root@4da9a6192c76 workspace]# getent hosts www.google.com
[root@4da9a6192c76 workspace]# echo $?
2
[root@4da9a6192c76 workspace]# .yarn/releases/yarn-3.6.1.cjs install
➤ YN0065: Yarn will periodically gather anonymous telemetry: https://yarnpkg.com/advanced/telemetry
➤ YN0065: Run yarn config set --home enableTelemetry 0 to disable

➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0013: │ fecha@npm:4.2.3 can't be found in the cache and will be fetched from the remote registry
➤ YN0013: │ js-lifecycle-scripts@file:/tmp/hermeto/deps/yarn/github.com/chmeliik/js-lifecycle-scripts/js-lifecycle-scripts-external-gitcommit-c31b
➤ YN0000: └ Completed in 0s 350ms
➤ YN0000: ┌ Link step
➤ YN0007: │ js-lifecycle-scripts@file:/tmp/hermeto/deps/yarn/github.com/chmeliik/js-lifecycle-scripts/js-lifecycle-scripts-external-gitcommit-c31b39bfa27355ca5583184dfa21257c54938e1a.tgz::locator=berryscary%40workspace%3A. must be built because it never has been before or the last one failed
➤ YN0000: └ Completed
➤ YN0000: Done in 0s 484ms
[root@4da9a6192c76 workspace]# echo $?
0

# install lifecycle scripts ran in the hermetic build (.txt file side-effects)
[root@4da9a6192c76 workspace]# ls /tmp/js-lifecycle/yarnberry/
install.txt  postinstall.txt  preinstall.txt
```

mprahl:
I rebased and attempted to address the CI failures. Could you please approve the workflows to run again?

tests/unit/package_managers/test_js_utils.py
```diff
@@ -0,0 +1,52 @@
+# SPDX-License-Identifier: GPL-3.0-only
+"""Tests for common JS git clone utilities in js_utils.py."""
+
+import pytest
+
+from hermeto.core.errors import PackageRejected
+from hermeto.core.package_managers.js_utils import extract_git_url_parts
+
+
+class TestExtractGitUrlParts:
```
eskultety, line 14: These new tests belong to a separate commit, this one is only about moving content. Additionally, given how uniform the tests are and `parametrize` wasn't utilized, were these tests generated by an AI by any chance?

hermeto/core/package_managers/yarn/main.py
```diff
@@ -26,23 +48,36 @@
 log = logging.getLogger(__name__)
 
 
+class GitDep(TypedDict):
```
eskultety: nitpick: Nice inconsistency you found there! I can see this to be used in the npm backend, but we have since moved pretty much to frozen dataclasses, so any new code we introduce should already follow the new pattern IMO.

hermeto/core/package_managers/npm.py
```diff
@@ -456,7 +457,7 @@ def _extract_git_info_npm(vcs_url: NormalizedUrl) -> dict[str, str]:
     namespace_repo = url.path.strip("/").removesuffix(".git")
 
     # Everything up to the last '/' is namespace, the rest is repo
-    namespace, _, repo = namespace_repo.partition("/")
+    namespace, _, repo = namespace_repo.rpartition("/")
```
eskultety, line 460: This bugfix should go to a standalone commit, it's not directly related to the code movement/extraction.

hermeto/core/package_managers/js_utils.py
```diff
@@ -0,0 +1,69 @@
...
+    tarball_relpath = Path(
+        parts["host"],
+        parts["namespace"],
+        parts["repo"],
+        f"{parts['repo']}-external-gitcommit-{ref}.tgz",
+    )
+    tarball_path = deps_dir.join_within_root(str(tarball_relpath))
+
+    tarball_path.path.parent.mkdir(parents=True, exist_ok=True)
+    clone_as_tarball(clone_url, ref, tarball_path.path)
```
eskultety, line 67: Unfortunately this commit is not a clean extraction, the diff is mixed with new changes. When moving code/extracting, please always do it in a 1:1 fashion, then you can propose improvements, like adding the new extra  `extract_git_url_parts` helper.

hermeto/core/package_managers/yarn/main.py
```diff
@@ -1,13 +1,35 @@
...
+    EnvironmentVariable,
+    ProjectFile,
+    RequestOutput,
+)
 from hermeto.core.models.sbom import create_backend_annotation
+from hermeto.core.package_managers.js_utils import clone_repo_pack_archive
+
+# The public parse_locator rejects git locators; we need the lower-level
+# parser to inspect them for git dependency detection.
+from hermeto.core.package_managers.yarn.locators import _parse_locator
```
eskultety: This sounds odd - if we're adding support for git deps why don't we change the "public" locator logic? I must be missing something obvious in the way how we're adding the support, don't I?

hermeto/core/package_managers/yarn/main.py
eskultety: Agreed, testing for permissive mode itself is not enough, the SBOM part of reporting this needs t o be added as well.

hermeto/core/package_managers/yarn/main.py
```diff
@@ -100,23 +135,55 @@ def _verify_repository(project: Project) -> None:
...
     version = _configure_yarn_version(project)
     _verify_repository(project)
 
-    _set_yarnrc_configuration(project, output_dir, version)
-    packages = resolve_packages(project.source_dir)
-    _fetch_dependencies(project.source_dir)
+    git_deps = _parse_lockfile_git_deps(project)
 
-    return create_components(packages, project, output_dir)
+    if git_deps:
```
eskultety: nitpick: for better (or more fluent) readability the condition should be inverted to make sure the `else` clause is always longer.

eskultety: Additionally, git deps invert the order of how we process the project:

Git deps
- fetch --> to updaet the lockfile
- resolve --> to parse the package data

VS

No git deps
- resolve --> to parse package data from a frozen lockfile
- fetch -> to download

There should be a commentary warning about this is intentional.

hermeto/core/package_managers/yarn/main.py
eskultety: Do the patched git deps and workspace+commit cases have anything to do with it? Would those make it more complex for us if we wanted to adjust the logic of the public locator parser?

hermeto/core/package_managers/yarn/main.py
```diff
@@ -262,3 +329,173 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+    For SCP-style locators like protocol="git@host", source="ns/repo.git", returns
+    "git@host:ns/repo.git".
+
+    The ``git+`` prefix (e.g. ``git+ssh``, ``git+https``) is stripped so the result
+    is a URL that git can clone directly and that clone_as_tarball can apply its
+    ssh-to-https fallback to correctly.
+    """
+    if not protocol or not source:
+        raise PackageRejected(
+            f"Cannot build clone URL from protocol={protocol!r}, source={source!r}",
```
eskultety: nitpick: s/build/construct

hermeto/core/package_managers/yarn/main.py
```diff
@@ -262,3 +329,173 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+            solution="Ensure the git dependency in yarn.lock has a valid URL.",
+        )
+
+    # Strip git+ prefix (e.g. git+ssh -> ssh, git+https -> https) so the URL
+    # is directly usable by git clone and the ssh→https fallback in clone_as_tarball.
```
eskultety: Hmm, was this '→' unicode symbol intentional or is it a byproduct of the code snippet having been generated?

hermeto/core/package_managers/yarn/main.py
```diff
@@ -262,3 +329,173 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+    cloned_sources: dict[tuple[str, str], RootedPath] = {}
+    git_purl_map: dict[str, str] = {}
+
+    for dep in git_deps:
+        source_key = (dep["clone_url"], dep["ref"])
+
+        if source_key not in cloned_sources:
+            tarball_rooted = clone_repo_pack_archive(dep["clone_url"], dep["ref"], yarn_deps_dir)
+            cloned_sources[source_key] = tarball_rooted
+        else:
```
eskultety: nitpick: the conditional logic should be swapped so that 'else' is always longer.

hermeto/core/package_managers/yarn/main.py
```diff
@@ -1,13 +1,35 @@
...
+    Component,
+    EnvironmentVariable,
+    ProjectFile,
+    RequestOutput,
+)
 from hermeto.core.models.sbom import create_backend_annotation
+from hermeto.core.package_managers.js_utils import clone_repo_pack_archive
+
+# The public parse_locator rejects git locators; we need the lower-level
+# parser to inspect them for git dependency detection.
```
a-ovchinnikov: This sounds like a reason to make it public.

hermeto/core/package_managers/yarn/main.py
```diff
@@ -100,23 +135,55 @@ def _verify_repository(project: Project) -> None:
     _check_lockfile(project)
 
 
-def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Component]:
+def _resolve_yarn_project(
+    project: Project, output_dir: RootedPath, mode: Mode
```
a-ovchinnikov: You should not need to pass mode explicitly, it is accessible from config:
```python
from hermeto.core.config import get_config
mode = get_config().mode
```

hermeto/core/package_managers/yarn/main.py
```diff
@@ -100,23 +135,55 @@ def _verify_repository(project: Project) -> None:
...
-    _set_yarnrc_configuration(project, output_dir, version)
-    packages = resolve_packages(project.source_dir)
-    _fetch_dependencies(project.source_dir)
+    git_deps = _parse_lockfile_git_deps(project)
 
-    return create_components(packages, project, output_dir)
+    if git_deps:
+        if mode != Mode.PERMISSIVE:
+            raise PackageRejected(
+                "Git dependencies in Yarn Berry projects require permissive mode "
```
a-ovchinnikov: This makes it look like permissive mode is just a usual mode of operation when in reality is a workaround for limitations of actual projects. WDYT about rephrasing this as follows:
```diff
- "Git dependencies in Yarn Berry projects require permissive mode "
+ "Git dependencies in Yarn Berry projects cannot be processed in strict mode"
```
?
I would also have either dropped solution completely or at least reworded it in order not to encourage permissive mode usage.

hermeto/core/package_managers/yarn/main.py
```diff
@@ -262,3 +329,173 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+        if key == "__metadata":
+            continue
+
+        resolution = entry.get("resolution") if isinstance(entry, dict) else None
+        if not resolution:
+            continue
+
+        try:
+            parsed = _parse_locator(resolution)
+        except (UnexpectedFormat, UnsupportedFeature):
```
a-ovchinnikov: Is it worth it to at least log skipped resolutions?

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -1101,3 +1101,74 @@ def test_get_pedigree_with_unsupported_locators(
...
+
+        resolver = _ComponentResolver(
+            {file_locator: mock.Mock()},
+            [],
+            mock_project,
+            output_dir,
+            git_purl_map={tarball_path: GIT_VCS_URL},
+        )
+
+        from hermeto.core.package_managers.yarn.resolver import _ResolvedPackage
```
a-ovchinnikov: Could you please move imports to the top of the module?

a-ovchinnikov (commented):
LGTM in general, a couple of small questions and requests inline.

hermeto/core/package_managers/yarn/main.py
```diff
@@ -262,3 +329,173 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+            continue
+
+        ref = parsed.parsed_reference
```
eskultety: `parsed_reference` is a cached property, but otherwise similar to `_parse_locator` in that can raise `UnexpectedFormat`. The `_parse_lockfile_git_deps` isn't guarded by a try-except block anywhere, so theoretically it could leak. @mprahl do you agree with me that this call would be better placed within the try-except clause a few lines above?

hermeto/core/package_managers/yarn/main.py
eskultety: `_parse_locator` cannot raise `UnuspportedFeature`, that's the "public" variant that can do so? Is this some remnant of previous implementation attempt? Back to my earlier question in a different thread - isn't it actually better to adjust and consume the public version of the locator parser?

hermeto/core/package_managers/yarn/main.py
```diff
@@ -262,3 +329,173 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+def _parse_lockfile_git_deps(project: Project) -> list[GitDep]:
+    """Scan yarn.lock for git-resolved dependencies.
+
+    Unsupported variants (patched git deps, workspace+commit) are silently
+    skipped here and will be reported later by resolve_packages.
+
+    :param project: the Project whose yarn.lock will be scanned.
+    :return: list of GitDep dicts with keys: name, clone_url, ref.
+    """
+    lockfile_filename = project.yarn_rc.get("lockfileFilename", "yarn.lock")
```
eskultety: This hardcoded yarn.lock path construction is duplicated in here and in `_build_clone_url`. I think now would be a good time to make `yarn.lock` a property of the `Project` class which already holds references to yarnrc and package.json.

hermeto/core/package_managers/yarn/main.py
```diff
@@ -262,3 +329,173 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+    lockfile_path = project.source_dir.join_within_root(lockfile_filename).path
+    lockfile_content = lockfile_path.read_text()
+
+    output_dir_str = str(output_dir.path)
+    lockfile_content = lockfile_content.replace(output_dir_str, "${output_dir}")
+
+    return ProjectFile(
+        abspath=lockfile_path.resolve(),
+        template=lockfile_content,
+    )
```
eskultety, line 503: It would be nice to separate some of these helpers to a standalone introductory commit, solely for review purposes, there's quite an amount of code happening that one needs to jump back and forth, not a blocker though, I managed, but if you end up separating changes in this PR, consider this comment as well :) .

hermeto/core/package_managers/yarn/main.py
```diff
@@ -262,3 +329,173 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+    project: Project,
+    git_deps: list[GitDep],
+    output_dir: RootedPath,
+) -> tuple[list[ProjectFile], dict[str, str]]:
+    """Clone git deps, write resolutions to package.json, and return ProjectFiles.
+
+    :param project: the Project whose package.json will be modified
+    :param git_deps: list of GitDep dicts
+    :param output_dir: base output directory
+    :return: tuple of (project_files, git_purl_map) where git_purl_map maps
```
eskultety: naming suggestion: `tarball_vcs_url_map` or something similar everywhere.

hermeto/core/package_managers/yarn/resolver.py
```diff
@@ -317,15 +334,21 @@ def _generate_purl_for_package(package: _ResolvedPackage, project: Project) -> s
...
 
         elif isinstance(package.locator, (FileLocator, LinkLocator, PortalLocator)):
-            project_path = project.source_dir
-            workspace_path = package.locator.locator.relpath
-            package_path = package.locator.relpath
+            tarball_path = (
+                str(package.locator.relpath) if isinstance(package.locator, FileLocator) else None
+            )
+            if tarball_path and tarball_path in self._git_purl_map:
+                qualifiers["vcs_url"] = self._git_purl_map[tarball_path]
```
eskultety: So we need this purl map that is is propagated all the way down here. That is very opaque and it took me a while to understand and trace the code. Alternative suggestion (disclaimer: I know shamefully little about Yarn) - what if we defined a new locator type `GitFileLocator` that would be identical to `FileLocator` except it would carry the VCS attribute we need?
Unless I'm missing something we could inject this `GitFileLocator` in the `_resolve_package` method of the `ComponentResolver` before returning a `_ResolvedPackage`. That way this method can remain a static helper method and the resolver would remain the sole "owner" of the parsed locator data including the git dependency resolution.

eskultety: Might be a good thing to convert Locators to dataclasses instead of named tuples (unless we unpack or index somewhere in the code) to support inheritance and proper type-based equality checks.

tests/unit/package_managers/yarn/test_main.py
```diff
@@ -490,3 +505,256 @@ def test_fetch_yarn_source(
...
+    )
+    # git+ prefix is stripped so clone_as_tarball's ssh→https fallback works
+    assert _build_clone_url("git+ssh", "//git@github.com/owner/repo.git") == (
+        "ssh://git@github.com/owner/repo.git"
+    )
+    assert _build_clone_url("git+https", "//github.com/owner/repo.git") == (
+        "https://github.com/owner/repo.git"
+    )
+    with pytest.raises(PackageRejected, match="Cannot build clone URL"):
+        _build_clone_url(None, "//github.com/owner/repo.git")
```
eskultety: Please separate positive and negative test cases for better readability. Additionally, this should be handled with a parametrize as the native mechanism to automatically generate cases based on variable inputs.

tests/unit/package_managers/yarn/test_main.py
```diff
@@ -490,3 +505,256 @@ def test_fetch_yarn_source(
...
+
+
+def test_build_vcs_url() -> None:
+    dep: GitDep = {"name": "foo", "clone_url": "https://github.com/owner/foo.git", "ref": "abc123"}
+    assert _build_vcs_url(dep) == "git+https://github.com/owner/foo.git@abc123"
+
+    dep = {"name": "bar", "clone_url": "git@github.com:owner/bar.git", "ref": "def456"}
+    assert _build_vcs_url(dep) == "git+git@github.com:owner/bar.git@def456"
```
eskultety: Same here --> pytest parametrize.

tests/integration/test_yarn.py
eskultety: >  triggers a new error that gets raised earlier.

We're in the middle of defining a set of error codes and the corresponding errors, so in foreseeable future we should convert all these `expected_output`s to just error codes. @mprahl let's just skip output comparison in the test.

tests/integration/test_yarn.py
```diff
@@ -27,10 +27,11 @@
             utils.TestParameters(
                 branch="yarn/disallowed-protocols",
                 packages=({"path": ".", "type": "yarn"},),
+                global_flags=["--mode", "permissive"],
```
eskultety: Hold on, this test no longer tests what the original did, now we're running it in the permissive mode. We ought to keep the original test pretty much untouched and to verify the git deps we'd need a new test that would exercise the "happy" path using the permissive mode.

hermeto/core/package_managers/js_utils.py
```diff
@@ -0,0 +1,69 @@
...
+                solution="Ensure the git dependency has a valid URL.",
+            )
+        host = match.group(1)
+        path = match.group(2).removesuffix(".git")
+
+    namespace, _, repo = path.rpartition("/")
+    return {"host": host, "namespace": namespace, "repo": repo}
+
+
+def clone_repo_pack_archive(
```
eskultety, line 42: This patch doesn't cover this function with a unit test. I myself are trying to justify the existence of such test, so I'll leave this as an open question if anyone has a stronger opinion.

hermeto/core/package_managers/yarn/resolver.py
mprahl: The `tarball_vcs_url_map` approach matches the npm package manager. I think a new locator type is a larger refactor outside of the scope of this PR.

hermeto/core/package_managers/yarn/main.py
mprahl: @eskultety could you add more context of what you mean about the SBOM reporting? I've tried to keep it the same as the npm package manager where the VCS URL is carried over into the PURL.

hermeto/core/package_managers/yarn/resolver.py
mprahl: The first comment in the thread should be fixed. For the second, `_parse_lockfile_git_deps` skips `workspace` Git deps.

mprahl:
@eskultety @a-ovchinnikov @taylormadore thank for your the PR reviews! I believe I've addressed or replied to all the comments. Could you please take another look?

hermeto/core/package_managers/yarn/resolver.py
eskultety: > matches the npm package manager

That's a fair argument, although old patterns should not be binding :).

> I think a new locator type is a larger refactor outside of the scope of this PR.

Uhm, I don't think it is, but fair enough, since this is an external contribution, we can file a follow-up issue on this. @mprahl since there are merge conflicts which you need to resolve, can you at least add a commentary on the `tarball_vcs_url_map` on why it exists and where is the intended consumer - the PURL generator, everything else is just a transit station as the variable is passed around until it's saved here. Any type of commentary, like a `TODO`, will make it much easier for us to identify the changes needed in the follow-up work I just mentioned, especially since I don't know when we'll get to it, thanks.

tests/integration/test_yarn.py
eskultety: @mprahl I don't see the git deps "happy" path integration test being added in this PR. We can create a branch here https://github.com/hermetoproject/integration-tests/ so you can add one.

eskultety (commented):
Apart from the merge conflicts we have a strong bias towards merging features without integration tests. Unless I missed something obvious (please correct me) then we'll need a positive one for the git dependencies. ~~LGTM otherwise.~~

hermeto/core/package_managers/yarn/main.py
eskultety: @mprahl Oh, sorry, I missed this comment completely (the scrollback is long :) ). 

> what you mean about the SBOM reporting? 

Since you need permissive mode to go through successfully, we adopted SBOM annotations to report permissive mode to-be violations, e.g. here: https://github.com/hermetoproject/hermeto/blob/d9d4463340791e35c033b0e8fdc8f09a2a011418/hermeto/core/package_managers/gomod.py#L657

jacobbaungard:
> Based on my comment [#1295 (comment)](https://github.com/hermetoproject/hermeto/pull/1295#discussion_r2923839135) we'll need the SBOM annotations tweaked, @mprahl please have a look at how we do it for both SPDX and CycloneDX. This and a positive integration tests are the 2 last blocking items here for me.

@eskultety , I'll try to see if I can get this over the line, with @mprahl having found a different solution.

About the SBOM annotation, is this the right approach?

```
{
  "bomFormat": "CycloneDX",
  "annotations": [
    {
      "subjects": [
         # all deps listed here
        "pkg:npm/%40grafana/async-query6.0",
        "pkg:npm/get-document@1.0.0?vcs_url=git%2Bhttps://github.com/webmodules/get-document.git%40a04ccb499d6e0433a368c3bb150b4899b698461f",
        .....
      ],
      "annotator": {
        "organization": {
          "name": "red hat"
        }
      },
      "timestamp": "2026-03-13T12:45:18Z",
      "text": "hermeto:backend:yarn"
    },
    {
      "subjects": [
         # only git deps listed
        "pkg:npm/get-document@1.0.0?vcs_url=git%2Bhttps://github.com/webmodules/get-document.git%40a04ccb499d6e0433a368c3bb150b4899b698461f",
      ],
      "annotator": {
        "organization": {
          "name": "red hat"
        }
      },
      "timestamp": "2026-03-13T12:45:18Z",
      "text": "hermeto:permissive-mode:yarn:using-git-dependencies"
    }
  ],
  "components": [
  ......
```

jacobbaungard:
> Based on my comment [#1295 (comment)](https://github.com/hermetoproject/hermeto/pull/1295#discussion_r2923839135) we'll need the SBOM annotations tweaked, @mprahl please have a look at how we do it for both SPDX and CycloneDX. This and a positive integration tests are the 2 last blocking items here for me.

Ok, I tried here (I could not add my commits to this PR): https://github.com/hermetoproject/hermeto/pull/1405

mprahl:
This PR is superseded by #1405.

PR #1296, opened 2026-02-17
Parallelized integration tests

ST3V1K:
Include worker_id in the temp dir name so each xdist worker gets its own clone. Avoids multiple workers sharing one repo and conflicting on git checkout.

Start local pypi and dnf server only once in "master" worker in pytest_configure. Avoids creating multiple servers since session scope fixtures run once for every worker.

Closes https://github.com/hermetoproject/hermeto/issues/1207

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -286,6 +286,16 @@ pytest --hermeto-image=localhost/hermeto:latest [path/to/integration/test]
 
 All hermeto integration test CLI options as well as the corresponding environment variables can be listed with `pytest --help`.
 
+To run integration tests in parallel, use the `-n` option + the number of workers. To utilize all CPU cores, use `auto` as the number of workers. Examples:
+
+```shell
+nox -s integration-tests -- -n 4
+nox -s integration-tests -- -n auto
+
+pytest [path/to/test] -n 4
+pytest [path/to/test] -n auto
+```
+
 ### Adding new dependencies to the project
 
 Sometimes when working on adding a new feature you may need to add a new dependency to the project.


eskultety (commented):
Design suggestion before I get to full review: parallel cloning will waste both network and storage (although with `depth=1` this won't be significant), especially for cases like running them inside an interactive rebase. Can you try experimenting with git worktrees first? I can already foresee some issues when it comes to testing the gomod backend where I think we'd simply fail quite early because we'd not be able to fetch tags on a worktree, but I'd like to see some data on how many failures and how serious we'd see if we tried it. I think that would be a nice experiment (unless you've already done it).

An alternative to worktrees if they turn out to be completely unusable for our use case you could either clone the once and all subsequent repos as 'local' rather than wasting network bandwidth to communicate with GitHub OR just clone once and then do filesystem copying manually.

ST3V1K:
> Can you try experimenting with git worktrees first? I can already foresee some issues when it comes to testing the gomod backend where I think we'd simply fail quite early because we'd not be able to fetch tags on a worktree, but I'd like to see some data on how many failures and how serious we'd see if we tried it. 

As you said I have ran into issues with gomod and fetching tags. I was able to solve this but that ran into other problem with how we make a copy of the repository. After this I have scrapped the idea of worktrees and moved to the alternative solution.
 
> either clone the once and all subsequent repos as 'local' rather than wasting network bandwidth to communicate with GitHub

With this I have ran into problem with getting remote branches from a local shallow clone.

>  just clone once and then do filesystem copying manually.

Finally I have done this and it seems that it's working fine for now.

tests/integration/conftest.py
a-ovchinnikov, line 1: _Commit message_ for
https://github.com/hermetoproject/hermeto/commit/9cdc076a9c11e33db8bf0d733edafb81d2696bb3: 5s/copies it's/copies its/ ?

.github/workflows/gating.yaml
```diff
@@ -189,4 +189,4 @@ jobs:
           HERMETO_TEST_LOCAL_DNF_SERVER: '1'
         run: |
           git config --global --add safe.directory "*"
-          /var/tmp/venv/bin/nox -s integration-tests
+          /var/tmp/venv/bin/nox -s integration-tests -- -n auto
```
eskultety, line 192: Commit message nitpicking: This "topic" should be probably something like `.github` or `.github: workflows` or something similar ;)

CONTRIBUTING.md
```diff
@@ -286,6 +286,16 @@ pytest --hermeto-image=localhost/hermeto:latest [path/to/integration/test]
 
 All hermeto integration test CLI options as well as the corresponding environment variables can be listed with `pytest --help`.
 
+To run integration tests in parallel, use the `-n` option + the number of workers. To utilize all CPU cores, use `auto` as the number of workers. Examples:
```
eskultety, line 289: **Commit message:**
s/contributin.md/contributing.md

tests/integration/conftest.py
```diff
@@ -43,17 +47,26 @@ def env_value(cli_opt: str) -> str:
...
+    if worker_id == "master":
+        stack = ExitStack()
+        try:
+            stack.enter_context(_pypiserver_context())
+            stack.enter_context(_dnfserver_context())
+        except Exception:
+            stack.close()
+            raise
+        setattr(config, "_hermeto_exit_stack", stack)
+
```
eskultety, line 63: Spinning up the servers from the controller is a change that deserves a standalone intro commit, IOW it makes sense regardless of the parallelization effort.

tests/integration/conftest.py
```diff
@@ -1,16 +1,17 @@
 # SPDX-License-Identifier: GPL-3.0-only
-import contextlib
```
eskultety, line 2: **Commit message:**

You're lacking references to some resources, e.g. 

- https://pytest-xdist.readthedocs.io/en/stable/how-it-works.html
- https://pytest-xdist.readthedocs.io/en/stable/how-to.html#identifying-the-worker-process-during-a-test

tests/integration/conftest.py
```diff
@@ -103,9 +122,9 @@ def local_pypiserver() -> Iterator[None]:
         yield
         return
 
-    pypiserver_dir = Path(__file__).parent.parent / "pypiserver"
+    pypiserver_dir = Path(__file__).parents[1] / "pypiserver"
```
eskultety: Unrelated change.

tests/integration/conftest.py
```diff
@@ -120,13 +139,17 @@ def local_pypiserver() -> Iterator[None]:
             except requests.RequestException as e:
                 log.debug(e)
         else:
+            _terminate_proc(proc)
             raise RuntimeError("pypiserver didn't start fast enough")
 
+    try:
         yield
+    finally:
+        _terminate_proc(proc)
```
eskultety: This all seems redundant if you pass this handler as the stack context callback.

tests/integration/conftest.py
```diff
@@ -90,11 +103,17 @@ def hermeto_image() -> utils.HermetoImage:
...
-# test output.
-@pytest.fixture(autouse=True, scope="session")
-def local_pypiserver() -> Iterator[None]:
+def _terminate_proc(proc: subprocess.Popen[bytes]) -> None:
+    proc.terminate()
+    try:
+        proc.wait(timeout=5)
+    except subprocess.TimeoutExpired:
+        proc.kill()
+        proc.wait()
```
eskultety: Similarly here I think this terminate handler should be introduced standalone as it's an actual bug fix to our integration test suite.
How did you arrive to the timeout magic value? Shouldn't we be a little more graceful, e.g. 30s especially since we now also spin up a nexus instance? We won't realistically be waiting that long, but it's enough to gracefully clean up resources like databases and such.

eskultety:
> With this I have ran into problem with getting remote branches from a local shallow clone.

Care enough to explain? :) I was able to make this work (easter egg: there's a legitimate permissions related bug in the ITs code), but it wasn't as clean of a solution as just using file system copies. 
I also benchmarked both solutions. Here are my findings: 
- on a tmpfs (/tmp default for many distros and default for pytest) local git clone was marginally faster ~5%
- on a regular FS (btrfs in my case) filesystem copies were faster ~10% - I presume the numbers would be similar or equal on something like XFS or EXT

It was a nice experiment for me :), let's proceed with FS copies as you already have.

ST3V1K:
> Care enough to explain? 

I don't really remember all the details about it, in short I had problem with getting remote branches from the local clone (`fatal: invalid reference: ...`.) I was stuck on it for many hours and no matter what I did, I couldn't make it work. Well, I must have been doing something really wrong since you got it working.

tests/integration/conftest.py
ST3V1K: > This all seems redundant if you pass this handler as the stack context callback.

What exactly should I do then? Should I remove it because it is redundant or split it into new commit?

eskultety: `context.callback(_terminate_proc, proc)` is all you should need IMO for the auxiliary test servers:

- raising an exception doesn't prevent the callback to be run on context exit
- guarding `yield` is pointless because it's still in the context, once the context is over, the callback will run

That is unless I missed something somewhere.

tests/integration/conftest.py
```diff
@@ -177,3 +175,23 @@ def _check_ssl_configuration() -> None:
             raise RuntimeError("DNF server didn't start fast enough")
 
         yield
+
+
+def pytest_collection_modifyitems(
```
eskultety: This change is new, why is it now done during the collection time? I seem to remember this being done in `pytest_configure`? Or am I misremembering? Are we missing some data preventing us from doing it in `pytest_configure`?

ST3V1K: > I still don't see the Hermeto image being built for integr. tests purposes once only, i.e. only a single worker building it for everyone else.

I have some issues with making it run faster than without it. I will update it tomorrow. 

> This change is new, why is it now done during the collection time? I seem to remember this being done in pytest_configure? Or am I misremembering? Are we missing some data preventing us from doing it in pytest_configure?

It was like this before, but I was a little bit confused as well when I saw. I believe it has to do something with how are the hooks ran on workers/master. I will find more details about it tomorrow.

ST3V1K: > It was like this before, but I was a little bit confused as well when I saw. I believe it has to do something with how are the hooks ran on workers/master. I will find more details about it tomorrow.

Turns out that attribute `_tmp_path_factory` on pytest config does not exist in `pytest_configure` yet. So it has to be done in the `pytest_collection_modifyitems` hook. (Maybe, it can be done in some other hooks as well.)

ST3V1K: > I still don't see the Hermeto image being built for integr. tests purposes once only, i.e. only a single worker building it for everyone else.

Should be done. @eskultety

tests/integration/conftest.py
```diff
@@ -177,3 +201,23 @@ def _check_ssl_configuration() -> None:
...
```
eskultety: https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_sessionstart feels like a better fit for the job than test collection hook which is supposed to only filter the collected tests.

tests/integration/conftest.py
```diff
@@ -73,18 +72,43 @@ def top_level_test_dir() -> Path:
...
         utils.build_image(repo_root, tag=image_ref)
 
+    env_image = os.getenv("HERMETO_TEST_IMAGE")
+    image_ref = env_image or "localhost/hermeto:latest"
+
+    root_tmp_dir = tmp_path_factory.getbasetemp().parent
+
+    if not env_image:
+        if worker_id == "master":
+            _build_image()
```
eskultety: The controller process doesn't need to be the one building the image, all of the workers block on the lock regardless, so it can be just one of the workers doing it. Doing that you'll free the controller process and simplify the logic.

tests/integration/conftest.py
```diff
@@ -73,18 +72,43 @@ def top_level_test_dir() -> Path:
...
+        else:
+            fn = root_tmp_dir / "hermeto_image_ref.txt"
+            with FileLock(str(fn) + ".lock"):
+                if fn.is_file():
+                    image_ref = fn.read_text().strip()
+                else:
+                    _build_image()
+                    fn.write_text(image_ref)
```
eskultety: Do we need to write anything to the lock file instead of just `""`? I mean IIUC the workers are just waiting for the file to be written, that's all it's a synchronization gate not a source of any information.

tests/integration/conftest.py
eskultety: Still applies.

tests/integration/conftest.py
```diff
@@ -73,18 +72,43 @@ def top_level_test_dir() -> Path:
...
+
+    root_tmp_dir = tmp_path_factory.getbasetemp().parent
+
+    if not env_image:
+        if worker_id == "master":
+            _build_image()
+        else:
+            fn = root_tmp_dir / "hermeto_image_ref.txt"
+            with FileLock(str(fn) + ".lock"):
+                if fn.is_file():
```
eskultety: you can just do `if not fn.is_file()` and only use the code from the `else` branch.

eskultety: And as I'm looking at the code, once you optimize out "master" handling here and simplify the `FileLock` logic, you may as well stuff both pulling AND building under the same lock and decide based on `env_image` what you should do next - pulling or building. I have not tried, but should work IMO.

tests/integration/conftest.py
```diff
@@ -177,3 +201,23 @@ def _check_ssl_configuration() -> None:
...
+    This function implements a standard pytest hook. Please refer to pytest
+    docs for further information.
+    """
+    test_repo_url = os.getenv(
+        "HERMETO_TEST_INTEGRATION_TESTS_REPO",
+        DEFAULT_INTEGRATION_TESTS_REPO,
+    )
+    tmp_path_factory: pytest.TempPathFactory = getattr(config, "_tmp_path_factory")
+    base: Path = tmp_path_factory.getbasetemp()
+    repo_dir = base.parent / "integration-tests"
```
eskultety: I don't think the logic is correct as it breaks the single-process mode:
```
$ pytest tests/integration/test_generic.py
...
hermeto.core.errors.NotAGitRepo: The provided path /tmp/pytest-of-eskultety/pytest-N/integration-tests cannot be processed as a valid git repository
```
^this is because in single-process mode `base.parent` will clone it to /tmp/pytest-of-XYZ/integration-tests and so both the controller process and one of the workers clone the repo.

tests/integration/conftest.py
```diff
@@ -42,17 +47,26 @@ def env_value(cli_opt: str) -> str:
...
-    # https://pytest.org/en/latest/reference/reference.html#tmp-path-factory-factory-api
-    repo_dir = tmp_path_factory.mktemp("integration-tests", False)  # type: ignore
-    Repo.clone_from(url=test_repo_url, to_path=repo_dir, depth=1, no_single_branch=True)
-    return repo_dir
+def test_repo_dir(tmp_path_factory: pytest.TempPathFactory, worker_id: str) -> Path:
+    base = tmp_path_factory.getbasetemp()
+    target = base / "integration-tests"
+    if worker_id == "master":
+        return target
+    return copy_directory(base.parent / "integration-tests", target)
```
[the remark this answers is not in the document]
eskultety: This was somewhat correct, but not entirely ->resolving.

tests/integration/conftest.py
ST3V1K: done

tests/integration/conftest.py
ST3V1K: `worker_id` will be `master` only when tests are ran in the non-parallelized mode and it will _never_ be `master` with parallelization

tests/integration/conftest.py
ST3V1K: changed it to `fn.touch()`

tests/integration/conftest.py
ST3V1K: done

tests/integration/conftest.py
ST3V1K: should be fixed

tests/integration/conftest.py
eskultety: My last comment would be to add a commentary to all these `master` worker id checks to let everyone reading the code know that that code path only ever executes in the single-process mode.

tests/integration/conftest.py
```diff
@@ -177,3 +194,24 @@ def _check_ssl_configuration() -> None:
             raise RuntimeError("DNF server didn't start fast enough")
 
         yield
+
+
+def pytest_sessionstart(session: pytest.Session) -> None:
+    """Clone the integration tests repository.
+
+    This function implements a standard pytest hook. Please refer to pytest
+    docs for further information.
```
a-ovchinnikov, line 234: I think it is worth it to post a link to a referenced document.

tests/integration/conftest.py
```diff
@@ -177,3 +194,24 @@ def _check_ssl_configuration() -> None:
...
+    """
+    if os.getenv("PYTEST_XDIST_WORKER", "master") != "master":
+        return
+
+    test_repo_url = os.getenv(
+        "HERMETO_TEST_INTEGRATION_TESTS_REPO",
+        DEFAULT_INTEGRATION_TESTS_REPO,
+    )
+    tmp_path_factory: pytest.TempPathFactory = getattr(session.config, "_tmp_path_factory")
+    base: Path = tmp_path_factory.getbasetemp()
```
a-ovchinnikov: L212-L213: the linter is fine without these annotations, I suggest removing them.

tests/integration/conftest.py
```diff
@@ -44,15 +46,12 @@ def env_value(cli_opt: str) -> str:
...
-def test_repo_dir(tmp_path_factory: pytest.FixtureRequest) -> Path:
-    test_repo_url = os.environ.get(
-        "HERMETO_TEST_INTEGRATION_TESTS_REPO",
-        DEFAULT_INTEGRATION_TESTS_REPO,
-    )
-    # https://pytest.org/en/latest/reference/reference.html#tmp-path-factory-factory-api
-    repo_dir = tmp_path_factory.mktemp("integration-tests", False)  # type: ignore
-    Repo.clone_from(url=test_repo_url, to_path=repo_dir, depth=1, no_single_branch=True)
-    return repo_dir
+def test_repo_dir(tmp_path_factory: pytest.TempPathFactory, worker_id: str) -> Path:
```
a-ovchinnikov, line 66: This fixture could benefit from a short comment explaining where exactly the contents of `base/"integration-tests"` will go.

tests/integration/conftest.py
```diff
@@ -120,6 +120,15 @@ def _build_and_pull_image() -> utils.HermetoImage:
     return hermeto
 
 
+def _terminate_proc(proc: subprocess.Popen[bytes]) -> None:
+    proc.terminate()
+    try:
+        proc.wait(timeout=30)
+    except subprocess.TimeoutExpired:
+        proc.kill()
```
a-ovchinnikov, line 140: Can a process remain stuck after this second signal?

tests/integration/conftest.py
ST3V1K: done

tests/integration/conftest.py
ST3V1K: done

tests/integration/conftest.py
ST3V1K: done

tests/integration/conftest.py
ST3V1K: Removed the type annotation from L213, but `tmp_path_factory` would be type `Any`. It wouldn't break linters but it would remove autocomplete in IDEs.

tests/integration/conftest.py
ST3V1K: done

tests/integration/conftest.py
ST3V1K: In some edge-case it could have remained stuck. I added a timeout to the second `wait` call so that it doesn't hang indefinitely and tests can finish.

eskultety: > In some edge-case it could have remained stuck

How? Can you elaborate? `proc.kill` sends SIGKILL which can't be masked, so the process is gone.

tests/integration/conftest.py
eskultety: > but it would remove autocomplete in IDEs.

Only if your IDE mandates a strict type compliance which mypy would check later on. Syntactic and even semantic autocomplete should still work, so I'd say it's a particular IDE thing.

tests/integration/conftest.py
ST3V1K: From what I have read in the short time I was researching this, I got to the point that the process can be in a so called D state, where it is waiting for some IO operation that hangs and will not be killed (SIGKILL) until that operation is done or timed out.
The kernel supposedly marks the process as killed but it will remain like this until it is possible to actually kill the process.

tests/integration/conftest.py
ST3V1K: I tried both VS Code (Cursor) and PyCharm, syntactic autocomplete works, but semantic doesn't.

tests/integration/conftest.py
eskultety: >  The kernel supposedly marks the process as killed but it will remain like this until it is possible to actually kill the process.

Well, that can be indefinitely though, so any **additional** wait on our side is IMO fairly redundant, because we can't do anything about a D-state process. IF such a process even resumes, it'll be killed right away after and its state will be collected by init, so I still don't think this requires any further action on our side, SIGKILL is enough.

tests/integration/conftest.py
eskultety: Fair enough, let's keep your original typing proposal then if it makes IDE experience smoother :) .

tests/integration/conftest.py
ST3V1K: The `wait` was there mainly for processes that would be killed almost immediately. Any process in D-state would just break it, hence I later added the timeout. But it's true that we don't have to wait for any of the processes. 

Removed it

PR #1298, opened 2026-02-18
chore: Adds license information to Python modules

a-ovchinnikov:
Style guide for code assitants demands SPDX license header for all new files, however existing files lack this information. This commit adds license information as suggested in the style guide to all existing *.py modules in hermeto/ and tests/ directories.

This change is generated with a trivial sed script.

### DIFF
--- a/hermeto/__init__.py
+++ b/hermeto/__init__.py
@@ -1,3 +1,4 @@
+# SPDX-License-Identifier: GPL-3.0-only
 import importlib.util as imputil
 import sys
 from pathlib import Path
--- a/hermeto/core/binary_filters.py
+++ b/hermeto/core/binary_filters.py
@@ -1,3 +1,4 @@
+# SPDX-License-Identifier: GPL-3.0-only
 """Base classes for binary package filtering."""
 
 from abc import ABC, abstractmethod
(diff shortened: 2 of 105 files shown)


a-ovchinnikov:
Second revision: some filers had '# SPDX-License-Identifier: GPL-3.0-or-later' in them and ended up with two headers, removed all occurrences of the old header.

hermeto/core/package_managers/pip/main.py
```diff
@@ -1,4 +1,4 @@
-# SPDX-License-Identifier: GPL-3.0-or-later
+# SPDX-License-Identifier: GPL-3.0-only
```
eskultety: You can't do this at will I'm afraid. Not without an explicit consent of everyone who ever modified this file. It's impractical and doesn't make much of a difference. Let's keep this as is. Same for the other case.

a-ovchinnikov: That would be annoyingly inconsistent, to the point that I am almost willing to do a poll and somehow commit the results. I'll roll back the change for these files.

a-ovchinnikov:
Latest revision: rolled changes to preexisting headers back.

PR #1301, opened 2026-02-19
Report outdated Cargo.lock in permissive mode

slimreaper35:
One of the two use cases for permissive mode is a lockfile mismatch, e.g., an out-of-sync Cargo.lock with Cargo.toml.

Taint the SBOM with annotations as we did for gomod (https://github.com/hermetoproject/hermeto/pull/1204).

### DIFF
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -13,10 +13,11 @@ import tomlkit
 import tomlkit.exceptions
 from packageurl import PackageURL
 
+from hermeto import APP_NAME
 from hermeto.core.errors import LockfileNotFound, NotAGitRepo, PackageRejected, UnexpectedFormat
 from hermeto.core.models.input import Mode, Request
-from hermeto.core.models.output import Component, EnvironmentVariable, ProjectFile, RequestOutput
-from hermeto.core.models.sbom import create_backend_annotation
+from hermeto.core.models.output import Annotation, Component, ProjectFile, RequestOutput
+from hermeto.core.models.sbom import create_backend_annotation, spdx_now
 from hermeto.core.rooted_path import RootedPath
 from hermeto.core.scm import get_repo_id
 from hermeto.core.utils import run_cmd
@@ -123,8 +124,8 @@ class LocalCargoPackage:
 def fetch_cargo_source(request: Request) -> RequestOutput:
     """Fetch the source code for all cargo packages specified in a request."""
     components: list[Component] = []
-    environment_variables: list[EnvironmentVariable] = []
     project_files: list[ProjectFile] = []
+    annotations: list[Annotation] = []
 
     for package in request.cargo_packages:
         package_dir = request.source_dir.join_within_root(package.path)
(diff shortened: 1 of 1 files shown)


tests/integration/test_pip.py
```diff
@@ -137,7 +137,6 @@
                 branch="pip/rust_extension_lock_and_config_mismatch",
                 packages=({"path": ".", "type": "pip"},),
                 global_flags=["--mode", "permissive"],
-                check_output=False,
```
a-ovchinnikov, line 141: I am not going to block on this one now, but this is yet another example of verbatim check as opposed to rules-based checks. Some day we should abandon all verbatim checks except those few ones which are imposed by external tools.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -114,30 +124,58 @@ def to_component(self) -> Component:
...
+        existing.subjects.update(subjects)
+    else:
+        annotations.append(
+            Annotation(
+                subjects=subjects,
+                annotator={"organization": {"name": "red hat"}},
+                timestamp=spdx_now(),
+                text=annotation_text,
+            )
+        )
```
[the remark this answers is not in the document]
a-ovchinnikov: Is there a guarantee that there will be at least one element matching the `if` clause? If not you might want to add a default fallback value.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -137,19 +138,43 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
+    components: list[Component],
+) -> None:
+    """Update permissive mode SBOM annotation with subjects from the provided components."""
+    annotation_text = f"{APP_NAME}:permissive-mode:cargo:generated-lockfile"
+    subjects = {c.bom_ref for c in components}
+    if annotations:
+        existing = next(a for a in annotations if a.text == annotation_text)
+        existing.subjects.update(subjects)
+    else:
+        annotations.append(
```
a-ovchinnikov: This function mutates an object in enclosing scope instead of returning a value which can lead to very fun bugs. I will not block on this, but would recommend reconsidering this approach.

hermeto/core/package_managers/cargo/main.py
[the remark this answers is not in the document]
slimreaper35: I applied the suggestion from Gemini. It should be clear and robust enough now.

tests/integration/test_pip.py
slimreaper35: I don't think I understand.

slimreaper35: Hmm, so the fact that we regenerate the lockfile leads to undeterministic behavior, since cargo pulls out the latest crates and SBOM comparison will fail 😞

a-ovchinnikov: What I mean is the original change was essentially string match on steroids. String matches are brittle by nature and should be avoided. Our entire IT suit is based on string matches, I don't know exactly why, but I think because such matches are straightforward to implement and the maintenance price is spread over years. In reality we want to ensure that SBOMs follow certain general rules and that's what we should be testing. This is way out of scope for the PR though and also the latest revision seems to drop the change I initially commented on.

slimreaper35: > also the latest revision seems to drop the change I initially commented on.

Yes, because of my comment above. Testing SBOM changes for this use case does not make sense.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -137,19 +138,44 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
...
     )
 
 
+def _update_permissive_mode_annotation(
+    annotations: list[Annotation],
+    components: list[Component],
+) -> None:
+    """Update permissive mode SBOM annotation with subjects from the provided components."""
+    text = f"{APP_NAME}:permissive-mode:cargo:generated-lockfile"
+    subjects = {c.bom_ref for c in components}
```
a-ovchinnikov: Non-blocking nitpick: IIRC the general project preference is against set comprehensions written in compact style (I.e. `set(k for k in foo)` is preferred to `{k for k in foo}`)

hermeto/core/package_managers/cargo/main.py
```diff
@@ -25,6 +24,15 @@
 log = logging.getLogger(__name__)
 
 
+class CargoVendorResult(NamedTuple):
+    """
+    Important information after running cargo vendor command.
```
a-ovchinnikov: Non-blocking suggestion: WDYT about s/Important information.*$/Cargo vendoring result./ ?

hermeto/core/package_managers/cargo/main.py
slimreaper35: yeah, I remember

PR #1303, opened 2026-02-21
ci(docs): deploy docs from published release tags

### LINKED ISSUE #1284: [docs] Build & deploy docs out of released versions instead of main
When we skip the weekly release schedule (per our guidelines) for any reason it may happen that we may have adopted a change that enhanced the docs and was deployed instantly. That can be very confusing to our users.
Make sure we build docs only out of released versions and not from the main branch immediately after merging a PR.
Optionally we could deploy a `dev` version (or `next`) of docs out of the main branch.

mahmoudnasser1561:
fixes [ #1284 ](https://github.com/hermetoproject/hermeto/issues/1284)

Build and deploy docs only from published releases (tags), not from pushes to main.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -2,18 +2,17 @@
 
 name: Publish documentation
 on:
-  push:
-    branches:
-      - main
-
+  release:
+    types: [released]
 permissions:
   contents: write
-
 jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
+        with:
+          ref: refs/tags/${{ github.event.release.tag_name }}
       - name: Configure Git Credentials
         run: |
           git config user.name github-actions[bot]
@@ -28,4 +27,16 @@ jobs:
           path: .cache
           restore-keys: mkdocs-material-
       - run: pip install mkdocs-material
+      - name: Check if this is the latest release
+        id: latest
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          LATEST_TAG=$(gh release view --json tagName -q .tagName)
+          if [ "$LATEST_TAG" = "${{ github.event.release.tag_name }}" ]; then
+            echo "is_latest=true" >> $GITHUB_OUTPUT
+          else
+            echo "is_latest=false" >> $GITHUB_OUTPUT
+          fi
       - run: mkdocs gh-deploy --force --strict
+        if: steps.latest.outputs.is_latest == 'true'


mahmoudnasser1561:
Hi @eskultety
could you please review this PR
it's supposed to fix that doc build Issue

eskultety:
@mahmoudnasser1561 would you mind extending this also for a 'dev' version that is built out of main to have the current behaviour covered?

mahmoudnasser1561:
Hi @eskultety 

I updated the docs publishing so the root site is deployed only from published release tags, and added a separate `/dev/` site built from `main` to keep the current behavior. 

The deploys are non-destructive (`keep_files`) and I used concurrency to avoid races.

I appreciate your feedback

.github/workflows/docs.yaml
```diff
@@ -5,11 +5,16 @@ on:
   release:
     types: [published]
 
+concurrency:
+  group: docs-pages
+  cancel-in-progress: false
+
```
eskultety: Why exactly is concurrency needed? What problem are we facing when it comes to publishing docs?

.github/workflows/docs.yaml
```diff
@@ -29,4 +34,11 @@ jobs:
           path: .cache
           restore-keys: mkdocs-material-
       - run: pip install mkdocs-material
-      - run: mkdocs gh-deploy --force --strict
+      - run: mkdocs build --strict
+      - name: Deploy stable docs (release tag) to GitHub Pages root
+        uses: peaceiris/actions-gh-pages@v4
```
eskultety: The use of a 3rd-party action isn't disclosed anywhere, it's sneaking it in, please don't do that.

eskultety (changes requested):
This is not the right way of achieving multiple revisions/versions in our docs. This topic requires careful investigation on the options, experimenting and then discussing the possible approaches in the PR.  In its current shape or form this work unfortunately cannot be accepted.

eskultety:
@mahmoudnasser1561 maybe we should get back to just building the docs on tagged releases and leave the dev version to a follow-up. Provided you comply with all our contributing guidelines and you show us how you tested the publish work on your fork so that we can verify the change.

.github/workflows/docs.yaml
mahmoudnasser1561: I did that because now we have multiple doc release pipelines so I wanna make sure none would block the other or stop it.

that's why thought about  concurrency group to handle that

but if docs release happen at different spaced times then there is no need for concurrency

mahmoudnasser1561:
@eskultety 
I trimed the commits related to the `dev version built out of main` and opened a follow up Issue to keep it on the todo list

also we shoud  discuss the approach there to be on the same page

eskultety (commented):
@mahmoudnasser1561 this would trigger the docs build and deploy also for pre-releases and also releases which were not marked as latest which is not desirable.

mahmoudnasser1561:
@eskultety 
prereleases are now skipped and only the latest release deploys.
could you take another look ?

.github/workflows/docs.yaml
```diff
@@ -11,6 +11,9 @@ permissions:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+
+    if: ${{ !github.event.release.prerelease }}
+
```
eskultety: Extra new lines that you then remove again in the next patch...

.github/workflows/docs.yaml
```diff
@@ -2,9 +2,8 @@
 
 name: Publish documentation
 on:
-  push:
-    branches:
-      - main
+  release:
+    types: [published]
```
eskultety: This is not the correct release event type, this fires on pre-releases as well.

eskultety (changes requested):
@mahmoudnasser1561 Apart from my comment there is no need for this to be split into 3 commits, this is logically enclosed work so all commits should be squashed. You'll need to provide a nice commit message which links the issue this closes (in this case the issue provides the whole WHY) and apart from that it should briefly explain why you did id the way you did, link docs, etc.
Lastly, your DCO and git author is still missing your full name, please fix that in the next revision.

mahmoudnasser1561:
@eskultety 
I have changed the release type so that it doesn't fire on pre-releases.
also removed the redundant line breaks.
also removed the conditional check as there is no use for it with the new release type.
also squshed the commits into one.

please have a look

eskultety:
> You'll need to provide a nice commit message which links the issue this closes (in this case the issue provides the whole WHY) and apart from that it should briefly explain why you did id the way you did, link docs, etc.
> Lastly, your DCO and git author is still missing your full name, please fix that in the next revision.

@mahmoudnasser1561 I don't think you have so far complied with ^this part of my comment. Have you utilized any AI agent that needs to be disclosed in the commit message (see our AI contributing policy)? ACK to the code change.

mahmoudnasser1561:
@eskultety I addressed your feedback and updated the commit message accordingly.

Could you please confirm whether the PR is now ready to merge, or if there’s anything else I should adjust?

.github/workflows/docs.yaml
```diff
@@ -28,4 +27,16 @@ jobs:
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          LATEST_TAG=$(gh release view --json tagName -q .tagName)
+          if [ "$LATEST_TAG" = "${{ github.event.release.tag_name }}" ]; then
+            echo "is_latest=true" >> $GITHUB_OUTPUT
+          else
+            echo "is_latest=false" >> $GITHUB_OUTPUT
+          fi
       - run: mkdocs gh-deploy --force --strict
+        if: steps.latest.outputs.is_latest == 'true'
```
eskultety:
```
+        if: steps.latest.outputs.is_latest == 'true'
\ No newline at end of file
```

@mahmoudnasser1561 please fix that.

eskultety (commented):
ACK with thew new line fixed.

eskultety (approved):

a-ovchinnikov (approved):

Issue #1290, opened 2026-02-10
[github] Figure out a way to test changes to the GH release workflow


eskultety:
Context: https://github.com/hermetoproject/hermeto/pull/1289#pullrequestreview-3779860998

We need to find a way to test ANY changes to the release workflow to prevent tagging a release in order to find out that the pipeline failed to build and release the image.
In general, testing the release workflow would need to look at the following  3 aspects (well, technically just 2):

**1. Event trigger, was this triggered by a manual or automatic even trigger (e.g. release)?**
If someone tweaks the release trigger, we won't know about whether it's correct until the release time.
**2. Container image build**
We already know we can build the image successfully, because we do that for every PR, so every PR tests that for us already, any failure of building the image in our workflow is an infra issue that can be (currently) retried with the manual trigger of the `release` workflow. This is a NOP
**3.  Pushing the container image to the container registry**
Pushing an image should not fail, unless someone accidentally tweaks the registry credentials and it goes unnoticed.

Linting CI actions, such as the https://github.com/hermetoproject/hermeto/issues/1404 proposal may alleviate the problem at hand ever so slightly, but most linters won't be able to catch semantic errors, i.e. like a missing context attribute (e.g. `github.event.release.latest`).

Demiserular:
Hey @eskultety, I'd like to work on this! Based on the issue description, I'm planning to make `release.yaml `a reusable workflow (adding a `workflow_call`trigger with a `dry_run` boolean input), then create a new `test-release.yaml` that fires on any PR touching `release.yaml`and calls it with `dry_run: true` —this avoids duplicating logic and keeps a single source of truth.

A few quick clarifications before I submit:

Point 3 (registry push) : Is a dry-run build (image built but not pushed) sufficient to close this, or would you also want the test to validate an actual push to a throwaway/PR-scoped tag on `ghcr.io`? The latter would require `packages: write` on every PR.

The release workflow builds for both `linux/amd64` and `linux/arm64`. Should the dry-run test do the same (full coverage, ~10–15 min) or just `linux/amd64` (faster, ~3–5 min)? Given `gating.yaml `already tests `amd64 `on every PR, the latter might be redundant.

Should `test-release.yaml` also trigger on Dockerfile changes, or strictly on `release.yaml` changes only? Since the image build is already covered by `gating.yaml`'s `build-image` job, I'd lean toward `release.yaml` only unless you feel otherwise.

[...]

PR #1305, opened 2026-02-22
bug : (related to #1290) testing GH release workflow

uchiha-vivek:
What is fixed ?

Tried to fix  #1289, #1290
the latest tag is now applied with a single rule: on release events only when the release is marked "Latest" on GitHub and is not a prerelease

Included a `Test Release workflow` , trying to run the same build as the release workflow (caution - no push) .

**What are the tests done ?**

I have tried to do the following local testing

1. Docker build Test

```bash
docker build -f Dockerfile -t hermeto:release-test .
docker run --rm hermeto:release-test --help
```

The image build happened successfully

2. Testing Yaml workflow linting with the `actionlint`

```bash
actionlint .github/workflows/release-test.yaml
```

The test has been successfully ran on all the workflow files.

actionlint-testing

The specific release pipeline test on this Pull request is waiting for the maintainer approval . 
The result can be determined once the `release-test.yaml` is checked.

If the test `release-test` workflow succeed then we are good to go

@eskultety  Sir can you review this once.

### DIFF
--- a/.github/workflows/release-test.yaml
+++ b/.github/workflows/release-test.yaml
@@ -1,0 +1,49 @@
+# Test the release pipeline without pushing. Use this to validate changes to
+# release.yaml or the Dockerfile before doing a real release (see issue #1290).
+#
+# Runs: same build steps as release.yaml, but with push: false.
+# Triggers: PRs that touch release workflow or Dockerfile, or manual run.
+name: Release pipeline test
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/release.yaml'
+      - '.github/workflows/release-test.yaml'
+      - 'Dockerfile'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  build-only:
+    name: Build only (no push)
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      # Same metadata action as release; on PR ref we only get a test tag (no semver)
+      - name: Extract metadata (tags, labels)
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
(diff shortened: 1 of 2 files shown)


eskultety:
@uchiha-vivek thank you for the contribution. Can you walk me through how this actually resolves the problem which led to the release workflow test issue? The issue doesn't provide much context and hence it wasn't labeled as 'my first issue'. We're confident that builds are going to pass just fine, because we build the image and run integration tests against it with every PR + if a build has failed due to infra issues during release we can always re-run the job. So I'm wondering problem exactly does this solve other than the problem behind #1290?
Was this completely vibe-coded by any chance?

uchiha-vivek:
> @uchiha-vivek thank you for the contribution. Can you walk me through how this actually resolves the problem which led to the release workflow test issue? The issue doesn't provide much context and hence it wasn't labeled as 'my first issue'. We're confident that builds are going to pass just fine, because we build the image and run integration tests against it with every PR + if a build has failed due to infra issues during release we can always re-run the job. So I'm wondering problem exactly does this solve other than the problem behind #1290? Was this completely vibe-coded by any chance?

@eskultety  The `release-test.yaml` file is only built for the testing purpose. It dont hinder the work of `release.yaml` during ci run.

The `release-test.yaml` file does the following things :

1. check whether the Dockerfile builds correctly
2. check if ci changes break
3.  There is no push operation, it only builds
4. it creates another tag `release-test`

IMPORTANT 

When does `release-test.yaml` run ?
Answer - When a pr modifies _release.yaml_,  _release-test.yaml_ or _Dockerfile_ . Its more of validation check on the yaml file.

Since the image is build and all integrations tests are ran on every pr so the test file can be less releveant but provides significant check.
@eskultety  Please let me know your view.

eskultety:
@uchiha-vivek you're contradicting yourself when it comes to describing the value of the workflow in the very same comment you posted.

eskultety:
This is not the right fix to the problem. Closing due to inactivity.

Issue #1156, opened 2025-10-22
Split npm.py module into smaller ones


slimreaper35:
We have already split most other package managers into smaller modules. Each one usually contains a single piece of functionality. For consistency and potentially easier maintenance, we could split `npm` too.

EDIT: This includes splitting the corresponding unit tests module as well.

vanshaj2023:
Hi! @eskultety @slimreaper35 

I'm new to this codebase but have been going through the repo to understand how things work. 
I'd love to work on this issue if possible could you please assign it to me?

Looking forward to contributing!

PuneetKumar1790:
Hi! I’d like to work on this one 🙂

I’ve been looking at how the pip and yarn backends are structured, and I noticed they both split responsibilities into smaller modules  parsing, fetching, SBOM logic, orchestration, etc.

Right now npm.py seems to handle several concerns in a single file (lockfile parsing, dependency fetching, SBOM/purl generation, and project file updates), so I was thinking of restructuring it into a small npm/ package to separate those responsibilities.

Rough idea:

main.py → orchestration

package_lock.py → lockfile parsing + data models

fetch.py → artifact fetching and URL helpers

sbom.py → purl generation + SBOM components

This would make it more consistent with pip/yarn and keep each module focused and easier to test independently.

Does this direction align with what you had in mind? Happy to adjust the boundaries before starting the refactor.

MonishPuttu:
I am currently working on this issue and made the first commit by moving the npm.py into the npm/npm_legacy.py and using a temp shim to resolve the imports, i have made a step by step plan to keep the commits separated by logic and that the commits are atomic, i am running tests after and before each and every commit to make sure that nothing is breaking.

and here is my plan:
commit 1: To convert npm.py to a package
commit 2: extract project_files module by cutting three functions
commit 3: extract fetch module
commit 4: Cut all models + URL helpers into package_lock.py
commit 5: Cut fetch_npm_source + _resolve_npm into main.py
commit 6: remove _npm_legacy.py, finalize __init__.py
commit 7: split test_npm.py into per-module test files

MonishPuttu:
[...]

PR #1306, opened 2026-02-22
Fix/pip rust missing cargo lock warning 1297

### LINKED ISSUE #1297: Handle Rust dependencies for Python sdists even if Cargo.lock is missing
When [detecting the python packages that have Rust dependencies](https://github.com/hermetoproject/hermeto/blob/ad1694ab578175201643ef1d4120b63cf1e5d0dc/hermeto/core/package_managers/pip/rust.py#L70), Hermeto skips those that don't have a `Cargo.lock` file. This leads to user confusion.

In strict mode, one would expect the absence of the lock file to cause errors. Given the heuristic nature of the Rust+Python support, this may be tricky. If it's not possible to do reliably, maybe Hermeto should at least log a warning for packages that [depend on rust](https://github.com/hermetoproject/hermeto/blob/ad1694ab578175201643ef1d4120b63cf1e5d0dc/hermeto/core/package_managers/pip/rust.py#L28) and don't have a Cargo.lock file.

In permissive mode, one would expect Hermeto to fetch the dependencies anyway and generate the lockfile, just like it does for lockfiles that are out of sync with Cargo.toml. Hermeto doesn't do this by design. To quote @a-ovchinnikov:

> There is a conceptual difference between a lock out of sync and no lock at all. Locks are essential for reproducible builds. A lock out of sync means that there was an attempt to lock a package which ultimately failed for some reason, but at least package maintainers recognize that dependencies should be locked. Here lock regeneration is a band-aid until maintainers fix the lock issue. When a package has never been locked to begin with then we cannot provide any guarantees and any band aid will become the permanent solution which will also make build reproducibility fundamentally out of reach. I still think that such issues must be opened with offending packages.

However, this nuance is lost on the vast majority of Hermeto users.

It may be worth to extend the "permissiveness" of permissive mode to cover missing Cargo.lock as well.
If not, it would be good to document the design decision (and also log a warning message when such packages are found) to reduce user confusion.

mahmoudnasser1561:
Fixes #1297.

This PR improves visibility when processing Python sdists with Rust build dependencies.

Changes:
- In `pip` Rust detection, when a package appears Rust-based but has no `Cargo.lock`,
  Hermeto now logs a warning instead of silently skipping Rust prefetch.
- Keeps current reproducibility behavior unchanged: missing `Cargo.lock` is still not auto-generated.
- Adds unit tests for the new warning behavior.
- Updates `docs/pip.md` to clarify:
  - missing `Cargo.lock` handling
  - permissive mode only regenerates lockfiles that exist but are out of sync.

also I reproduced that situation to verify the new behavior on my machine and took a screenshot for reference here
image


.github/workflows/docs.yaml
a-ovchinnikov, line 1: Please drop https://github.com/hermetoproject/hermeto/commit/92f0385322fe51f87d15c86be53f2a6e480a0667 from this PR, it does not belong here, it is a separate issue.

docs/pip.md
a-ovchinnikov, line 1: __Commit message__: when contributing please ensure it follows project standard of descriptiveness.

hermeto/core/package_managers/pip/rust.py
```diff
@@ -104,6 +99,18 @@ def filter_packages_with_rust_code(packages: list[dict[str, Any]]) -> list[Cargo
         if not _depends_on_rust(source_dir):
             shutil.rmtree(extract_dir, ignore_errors=True)
             continue
+
+        cargo_lockfiles = list(extract_dir.rglob("Cargo.lock"))
+        if not cargo_lockfiles:
+            log.warning(
+                "Detected Rust-based python package %s without Cargo.lock. "
+                "Skipping Rust dependency prefetch for reproducibility. "
```
a-ovchinnikov, line 107: Could you please elaborate what does this line mean (from a user's prespective)? I afraid that this won't provide UX we are striving for.

tests/unit/package_managers/pip/test_rust.py
```diff
@@ -35,3 +36,52 @@ def test_the_shortest_path_in_cargo_package_is_inferred_as_root(
...
+        source_dir.mkdir(parents=True, exist_ok=True)
+
+    monkeypatch.setattr(rust.shutil, "unpack_archive", _mock_unpack_archive)
+    monkeypatch.setattr(rust, "_depends_on_rust", lambda _source_dir: True)
+    caplog.set_level("WARNING", logger=rust.log.name)
+
+    packages = rust.filter_packages_with_rust_code([{"path": package_path}])
+
+    assert packages == []
+    assert f"package {package_path.name} without Cargo.lock" in caplog.text
```
a-ovchinnikov, line 62: What is the purpose of  this assertion?

tests/unit/package_managers/pip/test_rust.py
```diff
@@ -35,3 +36,52 @@ def test_the_shortest_path_in_cargo_package_is_inferred_as_root(
...
+    assert not (tmp_path / "foo-1.0.0").exists()
```
a-ovchinnikov, line 63: Why did you choose negation vs object identity here?

tests/unit/package_managers/pip/test_rust.py
```diff
@@ -2,6 +2,7 @@
 
 import pytest
 
+from hermeto.core.package_managers.pip import rust
 from hermeto.core.package_managers.pip.rust import _get_rust_root_dir
```
slimreaper35, line 6: Why do we need two import lines from the same module?

slimreaper35 (commented):
Just added one more comment in addition to the ones above. Again, this PR does not provide the actual fix for the issue.

hermeto/core/package_managers/pip/rust.py
mahmoudnasser1561: well I thought if rust cargo files are missing maybe It would be useful to log out clearer warning logs to the user

tests/unit/package_managers/pip/test_rust.py
mahmoudnasser1561: I always used to check such cases with negation, I think they are easier to read
If the project has standard regarding that I will follow the guides

tests/unit/package_managers/pip/test_rust.py
mahmoudnasser1561: my intention was to try to filter the rust package file
so I am testing if if those package files are missing

tests/unit/package_managers/pip/test_rust.py
mahmoudnasser1561: I pulled in both because I wanted to avoid having to reference the module (rust._get_rust_root_dir) everywhere.

to be able to use _get_rust_root_dir with the rust prefix

mahmoudnasser1561:
@slimreaper35 @a-ovchinnikov 
well I guess I was supposed to ask clarifying questions to get the scope and domain of the referenced Issue 
I thought the Issue was just add some UX logs and fix docs
I thought It was that easy as I saw that my-first-issue label so I rushed the process

sorry for that

mahmoudnasser1561:
the feedback is really useful as I was about to take on some refactoring Issues like #1156 
so I will try to discuss the approach and propose my view of the solution with you before moving on

Thank you for your time
I really appreciate you feedback

tests/unit/package_managers/pip/test_rust.py
eskultety: > I pulled in both because I wanted to avoid having to reference the module (rust._get_rust_root_dir) everywhere.

Right, makes sense. Do you mean in the single instance where we actually call `_get_rust_root_dir` in the module?

mahmoudnasser1561: yes without the module name prefix

a-ovchinnikov: > yes without the module name prefix

Please point me to the line you are talking about.

a-ovchinnikov:
> as I was about to take on some refactoring Issues

Do you mean that you are not planning to continue working on this PR?

slimreaper35:
@mahmoudnasser1561 We would appreciate if you start working only on one or a maximum of two issues at a time. Many PRs require too much context switching for maintainers, and in the end, the review might be delayed.

> the feedback is really useful as I was about to take on some refactoring Issues like https://github.com/hermetoproject/hermeto/issues/1156

I think that issue is more suitable for newcomers, so if you want to work on that, it's fine. I can take over this one as I have more context, and it is on the edge as "my first issue".

mahmoudnasser1561:
> I think that issue is more suitable for newcomers, so if you want to work on that, it's fine. I can take over this one as I have more context, and it is on the edge as "my first issue".

@slimreaper35 
I wanted to work on the referenced Issue because it deals with refactoring and maybe that would help me get more familiar with the code through hands-on

a-ovchinnikov:
> that would help me get more familiar with the code through hands-on

@mahmoudnasser1561 do I understand you correctly and https://github.com/hermetoproject/hermeto/issues/1297 (i.e. the issue you are dealing with in this PR) is not hands-on enough?

mahmoudnasser1561:
> > that would help me get more familiar with the code through hands-on
> 
> @mahmoudnasser1561 do I understand you correctly and #1297 (i.e. the issue you are dealing with in this PR) is not hands-on enough?

@a-ovchinnikov it is hands-on but I am quite new to the codebase and don't have much context about the Issue I already rushed that PR and the reviewers told me that the PR actually doesn't fix the Issue.

unlike to the other one I know where the problem is and have clear context about the file contains ~800 lines of code
if you want me to stick to the current PR and only work on #1297 then I will stick to it but @slimreaper35 offered to take it as he said he has enough context

Issue #1300, opened 2026-02-19
[tests][integration][gomod] Add a test exercising multi-toolchain environment


eskultety:
Our container image no longer ships with multiple Go toolchains installed (we ship with pretty much latest Go), but one can locally execute Hermeto with several Go toolchains installed or cached. It'd be nice if we had an integration test covering this scenario.
Note the challenge here is how we can get an alternative toolchain to the integration test environment which runs in context of the Hermeto container image which only ships with a single toolchain, so the integration test executor will likely need to take care of that, however, it needs to do it in a way that supports the correct underlying architecture on which the test suite is executing, i.e. not injecting an x86_64 Go toolchain on aarch64 hosts.

LeeFred3042U:
@eskultety can i work on this issue?

notjivi:
Hey, I've been poking around the integration test setup and I'd love to take a crack at this.

I see what you mean about the arch mismatch risk, I was planning to use platform.machine() to make sure we're grabbing the right Go binary for the host so it doesn't break on ARM/aarch64. I'm also planning to bake in a SHA-256 check for the download to keep it secure and keep the 'side-loading' logic clean so we don't bloat the base image.

I've already got the repo cloned and nox running locally. Can I be assigned to this?

vanshaj2023:
Hi @eskultety 

I'm relatively new to the codebase but have been going through the repo to understand how things work.

I'd love to work on this issue! My initial thought on the approach:

- Set up a small script in the integration test executor that downloads the appropriate Go toolchain based on the host architecture (using `GOARCH`/`runtime.GOARCH`) before running the test, and then write the integration test to verify Hermeto handles a multi-toolchain environment correctly.

Could you assign this issue to me? Also, does this approach sound right, or is there something I'm missing or a better way to handle it?

Thanks!

eskultety:
**Depends on** https://github.com/hermetoproject/hermeto/pull/1717 which will make integration test contributions easier.

Priyanshu-u07:
@eskultety #1717 is merged now. Should I take this on?

eskultety:
AFAIK nobody is working on this item. Beware that solving this will require some deep thinking on the approach, at this point in time I am not very inclined towards downloading toolchains on demand - each Go SDK is about ~80MBs of data that needs to be fetched, unpacked and mounted which will prolong test executions. I'd say the "easyfix" solution, while straightforward should IMO be our last resort solution if there's no other option.

PR #1308, opened 2026-02-23
[tests][integration][gomod] Add a test exercising multi-toolchain environment

notjivi:
Addresses #1300. Adds platform-aware downloading and checksum verification for side-loading alternate Go toolchains in integration tests.

### DIFF
--- a/tests/unit/package_managers/test_gomod.py
+++ b/tests/unit/package_managers/test_gomod.py
@@ -2812,3 +2812,44 @@ def test_parsed_module_with_origin(input_json: dict, expected_origin: dict | Non
         assert module.origin.hash == expected_origin["hash"]
         assert module.origin.tag_sum == expected_origin["tag_sum"]
         assert module.origin.ref == expected_origin["ref"]
+
+
+@mock.patch.dict(os.environ, {"PATH": ""})
+@mock.patch("shutil.which")
+@mock.patch("hermeto.core.package_managers.gomod.Go._get_release")
+def test_multi_toolchain_detection(
+    mock_get_release: mock.Mock,
+    mock_which: mock.Mock,
+    tmp_path: Path,
+) -> None:
+    """
+    Test that Hermeto can successfully detect a multi-toolchain environment
+    by pointing the search paths to a temporary directory with fake binaries.
+    """
+    expected_go_releases = ["go1.20.1", "go1.21.0-custom.build.info"]
+    system_go_dir = tmp_path / "usr" / "local" / "go"
+    cache_go_dir = tmp_path / ".cache" / "hermeto" / "go" / "go1.21"
+
+    for go_dir in (system_go_dir, cache_go_dir):
+        bin_dir = go_dir / "bin"
+        bin_dir.mkdir(parents=True)
+        (bin_dir / "go").touch()
+        (bin_dir / "go").chmod(0o755)
+
+    mock_which.side_effect = lambda cmd, *args, **kwargs: cmd if Path(cmd).is_absolute() else None
+
+    with (
+        mock.patch("hermeto.core.package_managers.gomod.HERMETO_GO_INSTALL_DIR", system_go_dir),
+        mock.patch(
+            "hermeto.core.package_managers.gomod.get_cache_dir",
+            return_value=tmp_path / ".cache" / "hermeto",
+        ),
+    ):
+        mock_get_release.side_effect = expected_go_releases
+        installed_toolchains = _list_installed_toolchains()
+
+    expected = {GoVersion(release) for release in expected_go_releases}
+    actual = {go.version for go in installed_toolchains}
+
+    assert expected == actual
+    assert mock_get_release.call_count == len(expected_go_releases)


tests/integration/utils.py
```diff
@@ -771,3 +771,50 @@ def determine_integration_tests_to_skip() -> Any:
...
+    machine = platform.machine().lower()
+    
+    # 1. Ask the host computer what hardware it has
+    arch_map = {"x86_64": "amd64", "amd64": "amd64", "aarch64": "arm64", "arm64": "arm64"}
+    go_arch = arch_map.get(machine)
+    
+    if not go_arch:
+        raise ValueError(f"Unsupported architecture: {machine}")
+
+    version = "1.21.6" # The older version we want to test
```
a-ovchinnikov: Where does this specific version come from?

tests/integration/utils.py
```diff
@@ -771,3 +771,50 @@ def determine_integration_tests_to_skip() -> Any:
...
+    
+    tar_path = dest_dir / filename
+    extract_path = dest_dir / f"go-alt-{version}"
+
+    # Only download if we haven't already done it
+    if not extract_path.exists():
+        log.info(f"Downloading alternative Go toolchain from {url}...")
+        urllib.request.urlretrieve(url, tar_path)
+        
+        # 2. The Cybersecurity Check! (Verifying the fingerprint)
```
a-ovchinnikov: Two problems with this. First, if you feel compelled to write a comment describing what a piece of code does vs why it is there this is a sign you need a helper function. Second, could you please elaborate how exactly this is a security check, all this code does is computing and logging a hash.

eskultety: Was this entirely vibe coded?

notjivi: @eskultety Yeah, that's a good catch, I used AI to bootstrap the draft so that I don't have to start coding from scratch, I wanted to use github's testing servers to test the core idea of volume mounting actually worked, instead of spending hours polishing the code locally only to find out the core logic was flawed. However, I moved way too fast and definitely should have manually verified the generated cryptographic checks before pushing. That's on me. I'm fixing the core logic now.

notjivi: @a-ovchinnikov Thank you for the detailed review, you're right for all points. 

-I will rewrite this to actually pass in an expected SHA-256 string and raise an error if the downloaded tarball doesn't match, instead of just logging it passively.
-I'll move the downloading and hashing logic into its own separate helper function to keep the main test clean.
-I used 1.21.6 as a quick baseline to test fallback, but I will look into extracting the required version dynamically.
-and I'll run the linters, fix the formatting, and amend the commit message to follow the contributor guidelines (including the AI declaration).

a-ovchinnikov: > Thank you for the detailed review, you're right for all points.
> ...

FYI this reads like a machine-generated response, especially the second point since the code we are discussing is **not** a part of a test, but rather a helper. Please review https://github.com/hermetoproject/hermeto/blob/main/AI_CONTRIBUTION_POLICY.md especially the part about understanding the proposed change and interaction with maintainers.

eskultety: > instead of spending hours polishing the code locally only to find out the core logic was flawed

^This is actually the only right way of going about this, otherwise it just generates traffic and noise in upstream, please keep that in mind and if you wish to complete this work, you'll need to refine it locally first to the point you're confident the tests are passing for you.

notjivi:
Hey, my bad for the messy draft earlier. You guys were right, I definitely moved too fast without testing it locally first which might've ended up creating a bunch of noise.
I took a step back, took a few days and fixed everything up properly:
I moved the toolchain download stuff into utils.py with actual SHA-256 hash checking. It throws a ValueError if something doesn't match instead of just logging it.
Added a _safe_extract helper to stop any path traversal (CVE-2007-4559) when extracting the tarball.
Dropped the hardcoded 1.21.6 version. It now reads the required Go version directly from the test project's go.mod.
Added the actual gomod_multi_toolchain dummy project to test_data so Hermeto actually has a target to scan.
Fixed the volume mounts and podman flags so GOROOT and PATH point to the right alternate toolchain inside the container.
Got rid of all the weird comments, ran everything through ruff to fix the linting errors and squashed it all into one clean commit.
Also I made sure the Assisted-by and DCO sign-offs are in the commit message asw. Tested everything locally and the tests are passing smoothly now. Thanks for bearing with me on this one.

tests/integration/test_gomod.py
```diff
@@ -327,3 +327,56 @@ def test_e2e_gomod(
         expected_cmd_output,
         hermeto_image,
     )
+    
+def test_issue_1300_multi_toolchain_environment(
```
a-ovchinnikov: Can you see any downsides in including an issue ID into any object name?

tests/integration/test_gomod.py
```diff
@@ -327,3 +327,56 @@ def test_e2e_gomod(
...
+        """Test using an alternate Go toolchain specified in the environment.
+
+        This test prepares an alternate Go toolchain, mounts it into the container,
+        and sets GOROOT and PATH environment variables to ensure Hermeto uses it
+        for dependency resolution. This addresses issue #1300.
+        """
+        test_case_name = "gomod_multi_toolchain"
+        test_case_path = test_data_dir / test_case_name
+
+        # Parse required Go version and prepare the alternate toolchain
```
a-ovchinnikov: Why did you choose to add this comment?

tests/integration/utils.py
```diff
@@ -771,3 +773,77 @@ def determine_integration_tests_to_skip() -> Any:
     elif just_some_package_managers_were_affected_by(changes):
         return SUPPORTED_PMS - affected_package_managers(changes)
     return set()
+
+EXPECTED_GO_HASHES = {
```
a-ovchinnikov: Now this is something that would require an adequate comment. As I have mentioned earlier comments should tell why something is done, not how. Here you would want to tell your reader why you added these specific numbers to the codebase.

tests/integration/utils.py
```diff
@@ -771,3 +773,77 @@ def determine_integration_tests_to_skip() -> Any:
...
+    """Identify host architecture and setup verified Go toolchain."""
+    machine = platform.machine().lower()
+    if machine in ("x86_64", "amd64"):
+        go_arch = "amd64"
+    elif machine in ("aarch64", "arm64"):
+        go_arch = "arm64"
+    else:
+        go_arch = None
+
+    if not go_arch:
```
a-ovchinnikov: PEP8 alert: please use idiomatic `is None` for checking if something is None

tests/integration/utils.py
```diff
@@ -771,3 +773,77 @@ def determine_integration_tests_to_skip() -> Any:
...
+    tar_path = dest_dir / filename
+    extract_path = dest_dir / f"go-alt-{version}"
+
+    if not extract_path.exists():
+        _download_and_verify_toolchain(url, tar_path, expected_hash)
+
+        log.info("Extracting verified toolchain to %s", extract_path)
+        extract_path.mkdir(parents=True, exist_ok=True)
+        with TarFile.open(tar_path, "r:gz") as tar:
+            # Mitigation for CVE-2007-4559 (path traversal)
```
a-ovchinnikov: Is this something added by this change?

tests/integration/utils.py
```diff
@@ -771,3 +773,77 @@ def determine_integration_tests_to_skip() -> Any:
...
+    elif machine in ("aarch64", "arm64"):
+        go_arch = "arm64"
+    else:
+        go_arch = None
+
+    if not go_arch:
+        raise ValueError(f"Unsupported architecture: {machine}")
+
+    expected_hash = EXPECTED_GO_HASHES.get(version, {}).get(go_arch)
+    if not expected_hash:
```
a-ovchinnikov: Can this realistically happen?

tests/integration/utils.py
```diff
@@ -771,3 +773,77 @@ def determine_integration_tests_to_skip() -> Any:
...
+    if actual_hash != expected_hash:
+        dest_path.unlink(missing_ok=True)
+        raise ValueError(
+            f"SECURITY ALERT: Hash mismatch for {url}!\n"
+            f"Expected: {expected_hash}\nActual: {actual_hash}"
+        )
+
+def prepare_alt_go_toolchain(dest_dir: Path, version: str) -> Path:
+    """Identify host architecture and setup verified Go toolchain."""
+    machine = platform.machine().lower()
```
a-ovchinnikov: For better reader experience I would recommend placing L819-L828 into an appropriately named helper.

tests/integration/test_gomod.py
```diff
@@ -327,3 +327,56 @@ def test_e2e_gomod(
...
+
+        test_params = utils.TestParameters(
+            branch="main",
+            # This path is relative to the repo root inside the container
+            packages=({"path": test_case_name, "type": "gomod"},),
+            expected_exit_code=0,
+            expected_output="All dependencies fetched successfully",
+        )
+
+        # Execute the scan with custom toolchain mounts and environment overrides
```
a-ovchinnikov: Why did you choose to add this comment?

tests/integration/test_gomod.py
notjivi: Yes, issue IDs can become confusing if issue trackers change. I'll rename the function to test_multi_toolchain_environment.

tests/integration/test_gomod.py
notjivi: My bad, the comment is just explaining what the code is doing instead of why. I'll strip these out as the code is kinda self explanatory

tests/integration/test_gomod.py
slimreaper35: > Yes, issue IDs can become confusing if issue trackers change.

Can we add the URL to the function name somehow?

tests/integration/utils.py
notjivi: Understood. I'll add a comment explaining that these official hashes from go.dev verify the toolchains we're downloading to keep the tests secure and stable.

tests/integration/utils.py
notjivi: Updating to use the idiomatic is None check.

tests/integration/utils.py
notjivi: Yes, I used the _safe_extract in this to prevent path traversal when extracting the downloaded tarball.

tests/integration/utils.py
notjivi: You're right. That's just some leftover from when I was trying out a few different ways to do this. I'll clean it up.

tests/integration/utils.py
notjivi: Alright, I'll move the architecture detection into a separate helper function to keep this part cleaner.

tests/integration/test_gomod.py
notjivi: I'll pull that out. It's just some leftover noise from an earlier draft that isn't really needed since the code is clear enough on its own.

tests/integration/test_gomod.py
notjivi: The python syntax doesn't allow it but I can put it in the function's docstring.

tests/integration/utils.py
```diff
@@ -576,3 +595,261 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+                return line.split()[1].strip()
+
+    raise ValueError(f"No 'go' version directive found in {go_mod_path}")
+
+def _download_and_verify_toolchain(url: str, dest_path: Path, expected_hash: str) -> None:
+    """Download toolchain and verify SHA-256 checksum."""
+    log.info("Downloading alternative Go toolchain from %s", url)
+    if not url.startswith("https://"):
+        raise ValueError(f"URL scheme not allowed: {url}")
+    urllib.request.urlretrieve(url, dest_path) # noqa: S310
```
a-ovchinnikov: What is this noqa for?

tests/integration/utils.py
```diff
@@ -576,3 +595,261 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+
+    sha256 = hashlib.sha256()
+    with open(dest_path, "rb") as f:
+        for chunk in iter(lambda: f.read(4096), b""):
+            sha256.update(chunk)
+
+    actual_hash = sha256.hexdigest()
+    if actual_hash != expected_hash:
+        dest_path.unlink(missing_ok=True)
+        raise ValueError(
```
a-ovchinnikov: If I understand the code correctly a binary will be downloaded using https, why do we need another verification?

tests/integration/utils.py
```diff
@@ -576,3 +595,261 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+        raise FileNotFoundError(f"Could not find go.mod in {project_path}")
+
+    with open(go_mod_path, "r") as f:
+        for line in f:
+            if line.startswith("go "):
+                return line.split()[1].strip()
+
+    raise ValueError(f"No 'go' version directive found in {go_mod_path}")
+
+def _download_and_verify_toolchain(url: str, dest_path: Path, expected_hash: str) -> None:
```
a-ovchinnikov: What is the reason for making url an argument when it is hardcoded just a few lines below?

tests/integration/utils.py
```diff
@@ -576,3 +595,261 @@ def _validate_expected_dep_file_contents(dep_contents_file: Path, output_dir: Pa
...
+    elif machine in ("aarch64", "arm64"):
+        return "arm64"
+    else:
+        return None
+
+def prepare_alt_go_toolchain(dest_dir: Path, version: str) -> Path:
+    """Identify host architecture and setup verified Go toolchain."""
+    go_arch = _get_go_arch()
+
+    if go_arch is None:
```
a-ovchinnikov: If you raise in `_get_go_arch()` there won't be any need in this check.

tests/integration/utils.py
a-ovchinnikov: What I meant is this comment reads as if some pre-exisitng problem existed and the added line mitigates it when `_safe_extract()` have been around for almost three years, with CVE-2007-4559 being mentioned in docstring. I recommend  dropping this comment.

tests/integration/test_gomod.py
```diff
@@ -327,3 +327,54 @@ def test_e2e_gomod(
...
+                mounts=[
+                    # Mount the full Go directory to satisfy internal lib/src dependencies
+                    (str(alt_go_root), container_go_root),
+                    # Mount the test project into the container's temporary repo
+                    (str(test_case_path), str(test_repo_dir / test_case_name))
+                ],
+                podman_flags=[
+                    "-e", f"GOROOT={container_go_root}",
+                    "-e", f"PATH={container_bin}:/usr/local/bin:/usr/bin:/bin"
+                ],
```
eskultety: Why are we setting these? I'm curious why we want to force use of this injected toolchain?

tests/integration/test_data/gomod_multi_toolchain/main.go
```diff
@@ -0,0 +1,5 @@
+package main
```
eskultety: Integration tests live in a different repo - https://github.com/hermetoproject/integration-tests/

notjivi: Understood. I'll move the main.go and project files over to the integration tests repo and update this PR to reference them from over there.

tests/integration/utils.py
notjivi: The security linter saw urllib part and was showing problems because it thought I might use a sketchy URL. Since I'm manually forcing https://, there's no risk attached, which is why i used noqa.

tests/integration/utils.py
notjivi: Yea the hashing felt like overkill, i've removed the hashing part to keep it simple

tests/integration/utils.py
notjivi: Fixed! I got rid of that extra helper function as the verification part is gone now, and just put the URL directly where it's needed.

tests/integration/utils.py
notjivi: Fixed this.

tests/integration/test_gomod.py
notjivi: I put these to force it to see the custom toolchain we just downloaded, otherwise the test would just use the default Go installed in the container, which won't be testing the fix for #1300.

notjivi:
Apologies for the overlapping push, Mr. Erik! I was in the middle of fixing the code which addresses Mr. Alexey's feedback when you posted your comment.

I’ve pushed the real toolchain version now as a baseline. I agree that the CI run is too expensive. I'm happy to pivot to a mocked approach for the version detection test I’ll wait for Mr. Alexey to weigh in so I can follow the project's preferred direction for this.

tests/integration/test_gomod.py
eskultety: How do you intend to verify the correct toolchain was actually used in the test? :) The only way I can think of is to inspect the logs and to do string matching. But once we enter the string matching territory then I'm asking why do we need to bother forcing a toolchain or even using a real one instead of mocking this since we don't really have a good other option. If we had a different way of checking, I'd say go for it, but I'm really not sure we have. :)

a-ovchinnikov: If we are ok with pinning a toolchain then we could come up with a small program that won't build with this specific toolchain and use this as a negative proof. This, however, sounds like an overkill to me.

tests/integration/utils.py
a-ovchinnikov: Could you please add a one-line comment to the line above mentioning this? Hermeto treats any such pragmas as exceptional and requiring comments for future maintainers.

a-ovchinnikov:
> I think we can get a bit more creative with the test - why don't we just define a bunch of test python scripts that emulate go env GOVERSION and return various strings?

If we are ok with simulated `go env` output then this becomes a unit test with extra steps, and in that case I'd say a UT is enough. I wonder how much overhead does a full integration test add? Will it still be a problem once tests are parallelized? If yes should we consider adding an appropriate skip condition to it?

eskultety:
> If we are ok with simulated `go env` output then this becomes a unit test with extra steps, and in that case I'd say a UT is enough.

Uhm, not really, not in general, we always have to evaluate on a case by case basis, look at `test_resolve_gomod` that test IMO tests exactly nothing given all the mocks it has. That said, after double checking, you're right, a UT in this case should be enough as only need to mock a single property. That would most certainly cover the bug we had.

>  I wonder how much overhead does a full integration test add? Will it still be a problem once tests are parallelized? 

Yes, it will because the long tests are still going to be long. IOW, yes, we can get to ~7 minutes out of ~17mins due to parallelization, but if you skipped the cargo+pip multi-backend one, you can get to under 4 minutes, that's quite insane how much a single test takes.

> If yes should we consider adding an appropriate skip condition to it?

Uhm, we're talking e2es really, those are the long ones, so you can already skip those, so adding another skip condition about a test beign expensive doesn't IMO bring much value, that's the nature of e2es.

In this particular case it wasn't even a correct test because it never exercised the multi-toolchain path, it just injected something into the image and made it the only toolchain to be used (this is btw. why I'm so worried about fully agentic workflows). Once you start thinking real multi-toolchain, then we don't really have a good means of verifying (a new toolchain can always build an old project and you wouldn't be able to tell) apart from log scanning which if compared to a cheap integration test (i.e. not an e2e) makes no difference to the value brought. That said, @a-ovchinnikov you convinced me that in this particular case a unit test should do just fine and should not be very complex.

notjivi:
@eskultety  Thanks for the feedback! 
Honestly, I got a bit fixated on the 'challenge' part of the issue description 

>Note the challenge here is how we can get an alternative toolchain to the integration test environment which runs in context of the Hermeto container image which only ships with a single toolchain

and 

>however, it needs to do it in a way that supports the correct underlying architecture on which the test suite is executing, i.e. not injecting an x86_64 Go toolchain on aarch64 hosts.

These parts. 

I spent so much time making sure it could download and relocate toolchain location correctly, that I overlooked the fact that forcing the container to use a single injected toolchain doesn't actually test if Hermeto can navigate a multi-toolchain environment.

I’ll get rid of the download logic, and get a unit test instead. Thank you for the guidance on the architecture, helped me learn a lot about the 'cost vs. value' side of things here.

notjivi:
@eskultety  Thanks for the feedback! 
Honestly, I got a bit fixated on the 'challenge' part of the issue description 

>Note the challenge here is how we can get an alternative toolchain to the integration test environment which runs in context of the Hermeto container image which only ships with a single toolchain



I spent so much time making sure it could download and relocate toolchain location correctly, that I overlooked that forcing the container to use a single injected toolchain doesn't actually test if Hermeto can navigate a multi-toolchain environment.

I’ll get rid of the download logic, and get a unit test instead. Thank you for the guidance on the architecture, helped me learn a lot about the 'cost vs. value' side of things here.

eskultety:
@notjivi there hasn't been any update on this PR, do you still plan on working on this?

notjivi:
Hi @eskultety, yes, I absolutely plan on finishing this.

I sincerely apologize for the silence over the last two weeks. I had my college exams going on, and I could allocate very little time apart from it which I strictly dedicated to researching and writing my GSoC proposal.

Now that my exams are done and the first draft of my proposal is submitted, I am jumping right back into this PR.

I will have the updated code pushed within the next 24 hours. Thank you for your patience and for pinging me on this!

notjivi:
Hi @eskultety and @a-ovchinnikov,

As promised, I removed the expensive integration test and replaced it with a unit test.

On Alexey's advice, this skips the part where we download real binaries. Instead, I used tmp_path to simulate toolchains existing in the expected discovery folders, and mocks to the `go env goversion` output to return 1.20.1 and 1.21.0.

Let me know if this goes along with what you had in mind.

tests/unit/package_managers/test_gomod.py
```diff
@@ -2664,3 +2664,45 @@ def test_go_exec_env(
...
+        (bin_dir / "go").chmod(0o755)
+
+    mock_which.side_effect = lambda cmd, *args, **kwargs: cmd if Path(cmd).is_absolute() else None
+
+    with mock.patch("hermeto.core.package_managers.gomod.HERMETO_GO_INSTALL_DIR", system_go_dir), \
+         mock.patch("hermeto.core.package_managers.gomod.get_cache_dir", return_value=tmp_path / ".cache" / "hermeto"):
+
+        mock_get_release.side_effect = [
+            "go1.20.1",
+            "go1.21.0"
```
eskultety: This is correct, yes, but part of the issue we had where our mechanism failed were extra build info in the toolchain release string, so I'd advise adding some extra bits to one of ^these version strings.

eskultety: Additionally, can we declare these as a list in a dedicated variable?

tests/unit/package_managers/test_gomod.py
```diff
@@ -2664,3 +2664,45 @@ def test_go_exec_env(
...
+        ]
+
+        installed_toolchains = _list_installed_toolchains()
+
+    assert len(installed_toolchains) == 2
+
+    detected_versions = {str(go.version) for go in installed_toolchains}
+    assert "1.20.1" in detected_versions
+    assert "1.21.0" in detected_versions
```
eskultety: this works, but such asserts are very basic and prone to even minor test updates. It would be best to use a list of the release strings above, iterate over them and compare on L2704 based on `GoVersion` rather than on strings. Couple the asserts visually together, right now they're delimited by empty lines and the computed `detected_versions`.

tests/unit/package_managers/test_gomod.py
notjivi: Done, I moved it to `expected_go_releases` and added `go1.21.0-custom.build.info`

tests/unit/package_managers/test_gomod.py
notjivi: Updated, I changed it to iterating over `expected_go_releases` and comparing `GoVersion` objects. Also removed the empty lines to couple the asserts visually together as suggested.

tests/unit/package_managers/test_gomod.py
```diff
@@ -2664,3 +2664,44 @@ def test_go_exec_env(
...
+            return_value=tmp_path / ".cache" / "hermeto",
+        ),
+    ):
+        mock_get_release.side_effect = expected_go_releases
+        installed_toolchains = _list_installed_toolchains()
+
+    detected_versions = [go.version for go in installed_toolchains]
+    assert len(installed_toolchains) == len(expected_go_releases)
+    for release_string in expected_go_releases:
+        assert GoVersion(release_string) in detected_versions
```
eskultety: Why don't we condense this and L2703 like this:
```
expected = {GoVersion(release_string) for release_string in expected_go_releases}
actual =  = {go.version for go in installed_toolchains}
assert expected == actual
```

notjivi: Alright, I updated it. Resolved the merged conflicts as well.

eskultety:
ping @a-ovchinnikov I think we're good to go here.

tests/unit/package_managers/test_gomod.py
```diff
@@ -2812,3 +2812,44 @@ def test_parsed_module_with_origin(input_json: dict, expected_origin: dict | Non
...
+    for go_dir in (system_go_dir, cache_go_dir):
+        bin_dir = go_dir / "bin"
+        bin_dir.mkdir(parents=True)
+        (bin_dir / "go").touch()
+        (bin_dir / "go").chmod(0o755)
+
+    mock_which.side_effect = lambda cmd, *args, **kwargs: cmd if Path(cmd).is_absolute() else None
+
+    with (
+        mock.patch("hermeto.core.package_managers.gomod.HERMETO_GO_INSTALL_DIR", system_go_dir),
```
a-ovchinnikov, line 2842: Why did you choose to embed these mock contexts instead of moving them out of the test?

tests/unit/package_managers/test_gomod.py
```diff
@@ -2812,3 +2812,44 @@ def test_parsed_module_with_origin(input_json: dict, expected_origin: dict | Non
...
+        mock.patch(
+            "hermeto.core.package_managers.gomod.get_cache_dir",
+            return_value=tmp_path / ".cache" / "hermeto",
+        ),
+    ):
+        mock_get_release.side_effect = expected_go_releases
+        installed_toolchains = _list_installed_toolchains()
```
a-ovchinnikov, line 2849: Do I understand it right and this line is the "Act/When" part of the test?

notjivi: Yes, everything above is just the setup. This line is where the actual detection happens, and the results are checked right after this.

tests/unit/package_managers/test_gomod.py
notjivi: I used `with` because `system_go_dir` is dependent on `tmp_path`. Since decorators at the top of the function are evaluated as soon as the file is loaded, if we declare it outside the function, the test wouldn't know what `tmp_path` is, as `tmp_path` isn't created until the function starts executing, resulting in a `NameError`.
This is why I kept it inside the function and mocked it after `tmp_path` is created.

PR #1311, opened 2026-02-23
[gomod] Add GOSUMDB=off to the generated env variable set

### LINKED ISSUE #1187: [gomod] Add GOSUMDB=off to the generated env variable set
We've encountered a pipeline build problem with https://github.com/openstack-k8s-operators/openstack-operator/tree/18.0-fr4 including the following diff on top:
```
diff --git a/.gitignore b/.gitignore
index c25a43231d5c3b1564a0b292e27aa88e11a16a29..8141b57438c675a0e02fdace7dcb9fc37caf4ab5 100644
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,6 @@ config/operator/deployment/kustomization.yaml
 CI_TOOLS_REPO
 
 # generated workspace file
-go.work
 go.work.sum
 
 # docs
diff --git a/go.work b/go.work
new file mode 100644
index 0000000000000000000000000000000000000000..f597fdcd7650299e1f8045b9610b0a4b7a6683c0
--- /dev/null
+++ b/go.work
@@ -0,0 +1,8 @@
+go 1.24.4
+
+toolchain go1.24.9
+
+use (
+	.
+	./apis
+)
```

Sometimes `go` tries to contact the internet and fetch sumdb data for some modules, sometimes it didn't, this wasn't 100% reproducible.
Technically, we should have all the sum files already updated **so we need to investigate if it's safe to disable sumdb checking in general for all cases.**

kunalsdas:
Fixes #1187

**Problem**

I looked into the issue with the openstack-operator 18.0-fr4 branch where Go sometimes tries to hit sum.golang.org during the hermetic build. It's not 100% reproducible, but when it happens, the build fails because there's no network access.

The reason is that we never set GOSUMDB in the generated build-phase env vars, so Go just uses its default and tries to reach the sumdb.

**What I found**

We already set GOSUMDB=sum.golang.org in _resolve_gomod() for the fetch phase — that part is fine, checksums get verified when we download modules. But the env_vars_template that gets written out for the build phase doesn't include GOSUMDB at all. Since the build uses GOPROXY=file://... and everything is already cached locally, Go shouldn't need to contact the sumdb at that point.

**Fix**

Added GOSUMDB=off to env_vars_template so the build phase explicitly skips sumdb lookups. The fetch phase stays as-is.

**Testing**

 Unit tests pass (nox -s python — 1382 passed)
 Lint passes (nox -s lint)
 Validate against openstack-operator build (18.0-fr4 branch with go.work)

### DIFF
--- a/AI_CONTRIBUTION_POLICY.md
+++ b/AI_CONTRIBUTION_POLICY.md
@@ -1,0 +1,134 @@
+# AI Contribution Policy
+
+## Purpose
+
+This policy establishes guidelines for contributions to Hermeto that involve the
+use of AI tools, including but not limited to Large Language Models (LLMs), code
+generation assistants, and similar technologies. It aims to balance openness to
+modern development workflows with the project's need for high-quality,
+well-understood contributions.
+
+## Position
+
+Hermeto takes a **permissive approach** toward AI-assisted contributions. We
+recognize AI tools as legitimate development aids that can improve productivity
+and code quality when used responsibly.
+
+This position is consistent with the [Linux Foundation's Generative AI
+Policy](https://www.linuxfoundation.org/legal/generative-ai), which states that
+code generated in whole or in part using AI tools can be contributed to open
+source projects, **provided that licensing and intellectual property
+considerations are properly addressed**.
+
+## Contributor accountability
+
+Regardless of AI involvement, **the contributor is fully responsible for every
+aspect of their submission**. This includes correctness, security, adherence to
+project standards, and licensing compliance.
+
+Hermeto is a security-sensitive and security-oriented project: it prefetches
+dependencies, validates checksums, and produces SBOMs that downstream consumers
+rely on. AI-generated code that is submitted without thorough human review and
+understanding poses a direct risk to its mission.
+
+### What we expect
+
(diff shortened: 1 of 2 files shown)


eskultety:
@kunalsdas your PR doesn't provide any reasoning behind the change (except for the linked issue), the commit message is empty and doesn't conform to our contributing guidelines. Additionally, the point I raised in my comment in the linked issue doesn't seem to be covered or described anywhere in this PR which sadly leads me to think this was entirely AI generated? Please read up on our https://github.com/hermetoproject/hermeto/blob/main/AI_CONTRIBUTION_POLICY.md before proposing a contribution, thank you.

kunalsdas:
@eskultety,  I am sorry for the mistake I made while creating the PR. I updated the comment. Can u pls check and let me know your thoughts. Since I would love to explain more about the changes or ready to share screenshot.

a-ovchinnikov:
@kunalsdas I would suggest addressing all concerns raised in https://github.com/hermetoproject/hermeto/pull/1311#issuecomment-3945988743. Expanding PR description was just one of those.

Issue #1157, opened 2025-10-22
Split gomod.py module into smaller ones


slimreaper35:
We have already split most other package managers into smaller modules. Each one usually contains a single piece of functionality. For consistency and potentially easier maintenance, we could split `gomod` too.

adityai0:
@eskultety can i work on this issue as my first issue?

harshitbansal184507:
Here is how I propose the split , (only for the core folder , not for unit tests yet )

gomod/
├── __init__.py     
├── main.py          _resolve_gomod, _vendor_deps, _vendor_changed,  _select_toolchain, other helpers 
├── go.py            Go CLI wrapper class
├── gowork.py       GoWork class 
├── models.py       all data models
├── version.py      GoVersion, ModuleVersionResolver
├── parser.py       parsing functions     

@slimreaper35 
@eskultety would you mind taking a look when you have a moment and letting me know if this structure makes sense ?

Jay2006sawant:
Hi @eskultety I am a new contributor here,I would love to contribute here !

PR #1316, opened 2026-02-24
Refactor: Split gomod module handling into structured components

harshitbansal184507:
Related to #1157 
Previously, all Go module handling logic was implemented inside a single gomod.py file. This made the file large and responsibilities somewhat mixed.

This PR restructures the implementation by splitting it into smaller, focused modules:

gomod/
├── __init__.py
├── main.py     
├── go.py          Go CLI  class
├── gowork.py      GoWork class
├── models.py       data models
├── version.py     # GoVersion, ModuleVersionResolver etc.
├── parser.py      # Parsing functions

//import changes in test_gomod.py (unit tests)
Ran the existing unit tests after the refactor.

There are no intended functional changes — this is purely a structural refactor.

This is my first larger structural change in this codebase, so I would really appreciate feedback  and whether this aligns well with the project's architectural direction. Happy to iterate based on review.

### DIFF
--- a/AI_CONTRIBUTION_POLICY.md
+++ b/AI_CONTRIBUTION_POLICY.md
@@ -1,0 +1,134 @@
+# AI Contribution Policy
+
+## Purpose
+
+This policy establishes guidelines for contributions to Hermeto that involve the
+use of AI tools, including but not limited to Large Language Models (LLMs), code
+generation assistants, and similar technologies. It aims to balance openness to
+modern development workflows with the project's need for high-quality,
+well-understood contributions.
+
+## Position
+
+Hermeto takes a **permissive approach** toward AI-assisted contributions. We
+recognize AI tools as legitimate development aids that can improve productivity
+and code quality when used responsibly.
+
+This position is consistent with the [Linux Foundation's Generative AI
+Policy](https://www.linuxfoundation.org/legal/generative-ai), which states that
+code generated in whole or in part using AI tools can be contributed to open
+source projects, **provided that licensing and intellectual property
+considerations are properly addressed**.
+
+## Contributor accountability
+
+Regardless of AI involvement, **the contributor is fully responsible for every
+aspect of their submission**. This includes correctness, security, adherence to
+project standards, and licensing compliance.
+
+Hermeto is a security-sensitive and security-oriented project: it prefetches
+dependencies, validates checksums, and produces SBOMs that downstream consumers
+rely on. AI-generated code that is submitted without thorough human review and
+understanding poses a direct risk to its mission.
+
+### What we expect
+
+- **Understand your code.** You must be able to explain every line of your
+    submission, defend design decisions during review, and respond to reviewer
+    feedback with substance.
+
+- **Review AI output critically.** AI tools can produce code that is
+    syntactically correct, but semantically or logically flawed. Treat all AI
+    output as untrusted input that requires careful validation.
+
+- **Follow project standards.** AI-assisted contributions must meet the same
+    coding standards, test coverage, and review quality as any other
+    contribution. See [CONTRIBUTING.md](CONTRIBUTING.md).
+
+### What we will reject
+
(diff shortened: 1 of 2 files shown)


hermeto/core/package_managers/gomod/parser.py
a-ovchinnikov, line 1: Welcome to the project! As a new contributor you must have noticed that this module violates standard Python conventions which we strive to always follow. To continue collaboration please identify those.

harshitbansal184507: thank you  , I'm reviewing and will update the pr .

a-ovchinnikov: No no, this is just a question, you can answer it right here first. I would even insist on that.

harshitbansal184507: Ok , I wanted to ask 😅 but hesitated bcz , I thought I missed few things which were very basic . What I am able to identify is that there are 
--  few unnecessary imports which I have done in few files 
-- order of imports is not according to the conventions in python. 
--The split I have , I took the help of AI to suggest me how the split should happen , which suggested me to do split in 10-12 different files . Some of them I felt were unnecessary so I combined few .
-- Naming is according to the conventions .

hermeto/core/package_managers/gomod/go.py
```diff
@@ -0,0 +1,240 @@
...
+        finally:
+            super().__exit__(exc, value, tb)            
+
+
+
+
+def _clean_go_modcache(go: Go, dir_: StrPath | None) -> None:
+    # It's easier to mock a helper when testing a huge function than individual object instances
+    if dir_ is not None:
+        go(["clean", "-modcache"], {"env": {"GOPATH": dir_, "GOCACHE": dir_}})
```
eskultety: This would not belong to this module, this is a single-purpose helper that should live with its caller.

hermeto/core/package_managers/gomod/models.py
```diff
@@ -0,0 +1,188 @@
...
+            missing_hash_in_file = frozenset()
+
+        return Component(
+            name=self.name,
+            version=self.version,
+            purl=self.purl,
+            properties=PropertySet(missing_hash_in_file=missing_hash_in_file).to_properties(),
+        )    
+    
+class Package(NamedTuple):
```
eskultety: These are not models, these are essentially just data classes.

hermeto/core/package_managers/gomod/parser.py
```diff
@@ -0,0 +1,190 @@
...
+
+    modules: frozenset[ModuleID] = frozenset()
+
+    for go_sum_file in go_sum_files:
+        modules = modules | _parse_go_sum(go_sum_file)
+
+    return modules
+
+
+def _get_go_sum_files(
```
eskultety: This doesn't feel like it's parsing anything.

hermeto/core/package_managers/gomod/parser.py
```diff
@@ -0,0 +1,190 @@
...
+                fail_for_unexpected_format(f"package has no parent module: {line}")
+            module_has_packages[-1] = True
+        elif not line.startswith("##"):  # marker line
+            fail_for_unexpected_format(f"unexpected format: {line!r}")
+
+    return (module for module, has_packages in zip(modules, module_has_packages) if has_packages)
+
+
+
+def _list_toolchain_files(dir_path: str, files: list[str]) -> list[str]:
```
eskultety: This doesn't parse anything either.

eskultety (changes requested):
@harshitbansal184507 you went kinda overboard with the split resulting in odd placement of several classes and functions. Less is often more.
I will point out that you must have missed our contributing guidelines, most notably the section on pull requests because this PR doesn't meet several of the requirements, if not for anything +1,972-1,905 is not okay in a single commit, at least not from a review POV.
If you re-adjust the split and introduce e.g. a `utils.py` module for some helpers, this could work.

PR #1318, opened 2026-02-24
Add local Nexus proxy mode for integration tests

taylormadore:
Adds support for running integration tests through a local Sonatype Nexus proxy to verify that SBOM proxy External References and URLs are emitted correctly. When enabled, the test harness spins up a Nexus container, routes package fetches through it, then validates that each registry component has the expected proxy external references before stripping them for comparison against baseline test data.

Proxy mode is off by default. Individual tests can opt out with the no_proxy_mode pytest marker.

To exercise proxy mode:
`HERMETO_TEST_LOCAL_NEXUS_PROXY=true nox -s integration-tests`
This is also enabled automatically by `nox -s all-integration-tests`

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -123,6 +123,7 @@ def all_integration_tests(session: Session) -> None:
         {
             "HERMETO_TEST_LOCAL_PYPISERVER": "1",
             "HERMETO_TEST_LOCAL_DNF_SERVER": "1",
+            "HERMETO_TEST_LOCAL_NEXUS_PROXY": "1",
         },
     )
 
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -11,7 +11,9 @@ import pytest
 import requests
 from git import Repo
 
+from tests.integration.proxy import TEST_NEXUS_PORT, is_local_nexus_proxy_enabled
 from tests.integration.utils import DEFAULT_INTEGRATION_TESTS_REPO, TEST_SERVER_LOCALHOST
+from tests.nexusserver import DEFAULT_NEXUS_HOST, initialize_nexus
 
 from . import utils
 
@@ -26,6 +28,8 @@ _ENV_VAR_CLI_MAP = [
     ("HERMETO_TEST_DNFSERVER_SSL_PORT", "--hermeto-dnfserver-ssl-port"),
     ("HERMETO_TEST_GENERATE_DATA", "--hermeto-generate-test-data"),
     ("HERMETO_TEST_CONTAINER_ENGINE", "--hermeto-container-engine"),
+    ("HERMETO_TEST_LOCAL_NEXUS_PROXY", "--hermeto-local-nexus-proxy"),
+    ("HERMETO_TEST_LOCAL_NEXUS_NO_CLEANUP", "--hermeto-local-nexus-no-cleanup"),
 ]
 
 
@@ -176,3 +180,35 @@ def local_dnfserver(top_level_test_dir: Path) -> Iterator[None]:
             raise RuntimeError("DNF server didn't start fast enough")
 
         yield
+
+
+@pytest.fixture(autouse=True, scope="session")
+def local_nexusserver(top_level_test_dir: Path) -> Iterator[None]:
+    if (os.getenv("CI") and os.getenv("GITHUB_ACTIONS")) or not is_local_nexus_proxy_enabled():
+        yield
+        return
+
+    with contextlib.ExitStack() as context:
+        container, client = initialize_nexus(host=DEFAULT_NEXUS_HOST, port=TEST_NEXUS_PORT)
+        context.enter_context(client)
+
+        # Allows inspection of running Nexus container after test completion
+        if os.getenv("HERMETO_TEST_LOCAL_NEXUS_NO_CLEANUP") == "1":
+            log.info("HERMETO_TEST_LOCAL_NEXUS_NO_CLEANUP=1, Nexus server will NOT be cleaned up")
+        else:
+            context.enter_context(container)
+
+        log.info("Nexus server ready at %s", client.base_url)
+        yield
+
+
+def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
+    """Skip tests that are incompatible with proxy mode."""
+    if not is_local_nexus_proxy_enabled():
+        return
+
+    for item in items:
+        if item.get_closest_marker("no_proxy_mode"):
+            item.add_marker(
+                pytest.mark.skip(reason="Test incompatible with local Nexus proxy mode")
+            )
--- a/tests/integration/proxy.py
+++ b/tests/integration/proxy.py
@@ -1,0 +1,161 @@
+# SPDX-License-Identifier: GPL-3.0-only
+"""Utilities for testing registry proxy mode."""
+
+import copy
+import os
+import re
+from collections import defaultdict
+from collections.abc import Mapping
+from typing import Any
+
+from packageurl import PackageURL
+
+from hermeto import APP_NAME
+from hermeto.core.models.property_semantics import PropertyEnum
+from hermeto.core.models.sbom import (
+    BACKEND_ANNOTATION_PREFIX,
+    PROXY_COMMENT,
+    PROXY_REF_TYPE,
+    Annotation,
+    Component,
+    ExternalReference,
+)
+from tests.nexusserver import DEFAULT_NEXUS_HOST, DEFAULT_NEXUS_PORT
+
+TEST_NEXUS_PORT = int(os.getenv("NEXUS_PORT", str(DEFAULT_NEXUS_PORT)))
+_NEXUS_BASE_URL = f"http://{DEFAULT_NEXUS_HOST}:{TEST_NEXUS_PORT}"
+
+_PROXY_URL_ENV_PATTERN = re.compile(rf"^{APP_NAME}_([A-Z0-9_]+)__PROXY_URL$", re.IGNORECASE)
+DEFAULT_LOCAL_NEXUS_PROXY_ENV: dict[str, str] = {
+    f"{APP_NAME}_NPM__PROXY_URL": f"{_NEXUS_BASE_URL}/repository/npm-proxy/",
+}
+
+_DIRECT_SOURCE_QUALIFIERS = frozenset({"vcs_url", "download_url", "repository_url"})
+
+
+def is_local_nexus_proxy_enabled() -> bool:
+    """Return True when local Nexus proxy defaults should be enabled."""
+    return os.getenv("HERMETO_TEST_LOCAL_NEXUS_PROXY") == "1"
+
+
+def parse_proxy_env(env: Mapping[str, str]) -> dict[str, str]:
+    """Extract a {backend: proxy_url} mapping from Hermeto env vars.
+
+    Example: HERMETO_YARN_CLASSIC__PROXY_URL -> {"yarn-classic": <url>}
+    """
+    backend_proxy_urls = {}
+    for env_var, proxy_url in env.items():
+        if match := _PROXY_URL_ENV_PATTERN.match(env_var):
+            backend = match.group(1).lower().replace("_", "-")
+            backend_proxy_urls[backend] = proxy_url
+
+    return backend_proxy_urls
+
+
+def _get_bom_ref_backends(sbom: dict[str, Any]) -> dict[str, set[str]]:
+    """Return a mapping of component bom-refs to backend names."""
+    backends_by_bom_ref: dict[str, set[str]] = defaultdict(set)
+    for raw_annotation in sbom.get("annotations", []):
+        annotation = Annotation(**raw_annotation)
+        if not annotation.text.startswith(BACKEND_ANNOTATION_PREFIX):
+            continue
+
+        # Can be backend:foo or backend:experimental:foo
+        if not (backend_name := annotation.text.split(":")[-1]):
+            continue
+
+        for subject in annotation.subjects:
+            backends_by_bom_ref[subject].add(backend_name)
+
+    return dict(backends_by_bom_ref)
+
+
+def _is_registry_component(component: Component) -> bool:
+    """Return True if the component was fetched from a registry (not bundled or direct-source)."""
+    is_bundled = any(
+        p.name == PropertyEnum.PROP_CDX_NPM_PACKAGE_BUNDLED and p.value == "true"
+        for p in component.properties
+    )
+    if is_bundled:
+        return False
+
+    purl = PackageURL.from_string(component.purl)
+    return not (purl.qualifiers.keys() & _DIRECT_SOURCE_QUALIFIERS)
+
+
+def _partition_proxy_external_refs(
+    component: Component,
+) -> tuple[list[ExternalReference], list[ExternalReference]]:
+    """Partition external refs into (proxy_refs, non_proxy_refs)."""
+    if component.external_references is None:
+        return [], []
+
+    is_proxy_ref = lambda ref: ref.type == PROXY_REF_TYPE and ref.comment == PROXY_COMMENT
+    proxy_refs: list[ExternalReference] = []
+    non_proxy_refs: list[ExternalReference] = []
+    for ref in component.external_references:
+        (proxy_refs if is_proxy_ref(ref) else non_proxy_refs).append(ref)
+
+    return proxy_refs, non_proxy_refs
+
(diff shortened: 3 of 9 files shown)


tests/integration/utils.py
```diff
@@ -47,6 +47,34 @@
 log = logging.getLogger(__name__)
 container_engine = get_container_engine()
 
+
+def _resolve_hermeto_env(
+    run_defaults: Mapping[str, str] | None = None,
+    test_overrides: Mapping[str, str] | None = None,
+    call_overrides: Mapping[str, str] | None = None,
+    unset_hermeto_env: Collection[str] | None = None,
```
a-ovchinnikov: Why did you choose `Mapping` and `Collection` over `dict` and `Iterable` respectively? Also tuples are immutable so it is safe to default to an empty tuple instead of None on L55 (this will also simplify L63).

tests/integration/utils.py
```diff
@@ -47,6 +47,34 @@
...
+    return resolved
+
+
+def _env_to_engine_flags(env: Mapping[str, str] | None) -> list[str]:
+    """Convert env var dict to container engine ``-e`` flags."""
+    if not env:
+        return []
+    flags: list[str] = []
+    for name, value in env.items():
+        flags.extend(["-e", f"{name}={value}"])
```
a-ovchinnikov: Optional compacting suggestion:
```diff
-    flags: list[str] = []
-    for name, value in env.items():
-        flags.extend(["-e", f"{name}={value}"]) 
-   return flags
+   return sum(["-e", f"{name}={value}"] for name, value in env.items(), [])
```

tests/integration/utils.py
```diff
@@ -47,6 +47,34 @@
...
+) -> dict[str, str]:
+    """Resolve effective Hermeto env for one test invocation."""
+    resolved = {
+        **(run_defaults or {}),
+        **(test_overrides or {}),
+        **(call_overrides or {}),
+    }
+    for key in unset_hermeto_env or ():
+        resolved.pop(key, None)
+    return resolved
```
a-ovchinnikov: Optional declarative suggestion:
```diff
-    for key in unset_hermeto_env or ():
-        resolved.pop(key, None)
-    return resolved
+    return {k, v for k, v  in resolved.items() if k not in unset_hermeto_env}
```

Also maybe s/unset_hermeto_env/unset_hermeto_vars/?

tests/integration/utils.py
```diff
@@ -47,6 +47,34 @@
...
 
+
+def _resolve_hermeto_env(
+    run_defaults: Mapping[str, str] | None = None,
+    test_overrides: Mapping[str, str] | None = None,
+    call_overrides: Mapping[str, str] | None = None,
+    unset_hermeto_env: Collection[str] | None = None,
+) -> dict[str, str]:
+    """Resolve effective Hermeto env for one test invocation."""
+    resolved = {
```
a-ovchinnikov, line 56: The way `resolved` is constructed is pretty neat!

tests/integration/utils.py
```diff
@@ -47,6 +47,34 @@
...
+        **(call_overrides or {}),
+    }
+    for key in unset_hermeto_env or ():
+        resolved.pop(key, None)
+    return resolved
+
+
+def _env_to_engine_flags(env: Mapping[str, str] | None) -> list[str]:
+    """Convert env var dict to container engine ``-e`` flags."""
+    if not env:
```
a-ovchinnikov: PEP8 alert: s/not env/env is not None/

tests/integration/utils.py
```diff
@@ -494,11 +558,17 @@ def build_image_and_check_cmd(
...
+    :param podman_flags: Additional flags to be passed to podman
+    :param hermeto_env_overrides: Highest-precedence env var overrides for this call
     :param fetch_output_dirname: Name of the directory where the fetch output is stored
     :param env_vars_filename: Name of the file where the environment variables are stored
     :return: None
     """
     output_dir = tmp_path.joinpath(fetch_output_dirname)
+    merged_env = _resolve_hermeto_env(
+        call_overrides=hermeto_env_overrides,
+    )
```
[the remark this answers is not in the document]
a-ovchinnikov: This actually might warrant a comment.

tests/integration/conftest.py
```diff
@@ -154,20 +156,40 @@
...
+    # do not try to skip by code-change heuristic if a keyword or marker is specified
+    skip_by_change = not (config.getoption("-k") or config.getoption("-m"))
 
-    This function implements a standard pytest hook. Please refer to pytest
-    docs for further information.
-    """
-    # do not try to skip tests if a keyword or marker is specified
-    if config.getoption("-k") or config.getoption("-m"):
-        return
+    tests_to_skip = utils.determine_integration_tests_to_skip() if skip_by_change else set()
```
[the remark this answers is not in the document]
a-ovchinnikov: Looks like a very wide miss to me.

tests/integration/proxy.py
```diff
@@ -0,0 +1,161 @@
...
+    "HERMETO_NPM__PROXY_URL": f"{_NEXUS_BASE_URL}/repository/npm-proxy/",
+}
+
+_BACKEND_ANNOTATION_PREFIX = f"{APP_NAME}:backend:"
+_DIRECT_SOURCE_QUALIFIERS = frozenset({"vcs_url", "download_url", "repository_url"})
+
+
+def is_local_nexus_proxy_enabled() -> bool:
+    """Return True when local Nexus proxy defaults should be enabled."""
+    return os.environ.get("HERMETO_TEST_LOCAL_NEXUS_PROXY", "false").lower() == "true"
```
a-ovchinnikov: Once merged https://github.com/hermetoproject/hermeto/pull/1283 will switch from string boolean representation to numeric one.

tests/integration/proxy.py
```diff
@@ -0,0 +1,161 @@
...
+    proxy_refs = []
+    non_proxy_refs = []
+    if component.external_references is not None:
+        for ref in component.external_references:
+            if ref.type == PROXY_REF_TYPE and ref.comment == PROXY_COMMENT:
+                proxy_refs.append(ref)
+            else:
+                non_proxy_refs.append(ref)
+
+    return proxy_refs, non_proxy_refs
```
a-ovchinnikov, line 99: Optional suggestion (will require adding a new dependency):
```python
is_proxy_ref = lambda ref: ref.type == PROXY_REF_TYPE and ref.comment == PROXY_COMMENT
poxy_refs, non_proxy_refs = more_itertools.partition(is_proxy_ref, component.external_refs) if component.external_refs is not None else [], []
return list(poxy_refs), list(non_proxy_refs)
```

This requires [more-itertools](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.partition)

tests/integration/proxy.py
```diff
@@ -0,0 +1,161 @@
...
+    backend_proxy_urls: Mapping[str, str],
+) -> set[str]:
+    """Return the set of proxy URLs expected for a component, empty if none apply."""
+    component_backends = bom_ref_backends.get(component.bom_ref, set())
+    proxy_enabled_backends = component_backends & backend_proxy_urls.keys()
+
+    if not proxy_enabled_backends or not _is_registry_component(component):
+        return set()
+
+    return {backend_proxy_urls[backend] for backend in proxy_enabled_backends}
```
a-ovchinnikov: The consensus seems to be in favor of explicit set() invocations instead set comprehensions, could you please update them?

tests/integration/proxy.py
```diff
@@ -0,0 +1,161 @@
...
+def validate_and_strip_proxy_refs(
+    sbom: dict[str, Any], backend_proxy_urls: Mapping[str, str]
+) -> dict[str, Any]:
+    """Validate proxy refs exist and have correct URLs, return SBOM with them stripped."""
+    sbom_copy = copy.deepcopy(sbom)
+    bom_ref_backends = _get_bom_ref_backends(sbom_copy)
+    mismatches: list[dict[str, Any]] = []
+
+    for component_raw in sbom_copy.get("components", []):
+        component = Component.model_validate(component_raw)
```
a-ovchinnikov: Can `Component(**comoponent_raw)` be used instead?

tests/integration/conftest.py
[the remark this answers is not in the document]
taylormadore: It's impressive, really :sweat_smile:

tests/integration/utils.py
taylormadore: It was a reaction to the issue I fixed in dec8956. I was just making it explicit (and enforced via type-checking) that no mutations of the input were occurring. Perhaps it was an overreaction 😅

tests/integration/utils.py
taylormadore: Ha, I actually started with a slightly different nested list comprehension here but changed it 🤣

a-ovchinnikov: If we agree on bringing in `more-itertools` we could leverage [intersperse](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.intersperse) here.

tests/integration/utils.py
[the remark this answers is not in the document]
taylormadore: Ugh, this was leftover from an earlier revision. I chose not to add the Hermeto env_vars to the build_image_and_check_cmd function because none are needed by the non-fetch Hermeto commands used there. We also don't currently propagate the TestParameters there because they are also not needed. We can always add them later if they become necessary.

tests/integration/proxy.py
taylormadore: That's a neat library. I'm ok with including it

tests/integration/utils.py
taylormadore: Is that simpler than a regular nested list comprehension? We'd have to prepend the first `-e`

taylormadore:
Rebased on main

tests/integration/proxy.py
```diff
@@ -0,0 +1,159 @@
...
+def _partition_proxy_external_refs(
+    component: Component,
+) -> tuple[list[ExternalReference], list[ExternalReference]]:
+    """Partition external refs into (proxy_refs, non_proxy_refs)."""
+    is_proxy_ref = lambda ref: ref.type == PROXY_REF_TYPE and ref.comment == PROXY_COMMENT
+    non_proxy_refs, proxy_refs = (
+        more_itertools.partition(is_proxy_ref, component.external_references)
+        if component.external_references is not None
+        else ([], [])
+    )
```
a-ovchinnikov: For the sake of readability I suggest importing `partition` directly and introducing an alias for `component.external_references`:
```python
ext_ref = component.external_references
non_proxy_refs, proxy_refs = partition(is_proxy_ref, ext_ref) if ext_ref is not None else [], []
```
The syntax is like list comprehension, it works best when it all fits on one line.

tests/integration/proxy.py
```diff
@@ -0,0 +1,159 @@
...
+    return not (purl.qualifiers.keys() & _DIRECT_SOURCE_QUALIFIERS)
+
+
+def _partition_proxy_external_refs(
+    component: Component,
+) -> tuple[list[ExternalReference], list[ExternalReference]]:
+    """Partition external refs into (proxy_refs, non_proxy_refs)."""
+    is_proxy_ref = lambda ref: ref.type == PROXY_REF_TYPE and ref.comment == PROXY_COMMENT
+    non_proxy_refs, proxy_refs = (
+        more_itertools.partition(is_proxy_ref, component.external_references)
```
slimreaper35: If this is the only usage of `more-itertools`, I would consider rewriting it to a simple utility function, e.g.:

```python
def _partition(iterable, pred):
    t, f = [], []
    for x in iterable:
        (t if pred(x) else f).append(x)
    return t, f
```

Just to reduce noise from dependabot in the future.

a-ovchinnikov: `more-itertools` is a pretty handy expansion of `itertools`, I would suggest using it _more_ across the codebase in order not to keep reinventing the wheel.

taylormadore: Yeah, that was my thought as well after looking at the library. I wasn't considering adding it for just this case alone

slimreaper35: ok

eskultety: > I would suggest using it more across the codebase in order not to keep reinventing the wheel.

That's fine, but feels like a candidate for a wider refactor to prove its value and need, otherwise it's out of scope of the PR IMO.

a-ovchinnikov: > out of scope of the PR IMO.

100%!

taylormadore: Since we have differing opinions here, let me try to summarize and propose next steps. It seems like we're all open to the potential the more-itertools library offers, but we don't have consensus about including it now for this single use-case.

I propose that we drop it for now, but file an issue to investigate adding adding the library and searching out places in our codebase where it could be useful. How does that sound?

eskultety: > I propose that we drop it for now, but file an issue to investigate adding adding the library and searching out places in our codebase where it could be useful. How does that sound?

Sounds good to me!

noxfile.py
```diff
@@ -127,6 +124,7 @@ def all_integration_tests(session: Session) -> None:
             "HERMETO_TEST_RUN_ALL_INTEGRATION_TESTS": "1",
             "HERMETO_TEST_LOCAL_PYPISERVER": "1",
             "HERMETO_TEST_LOCAL_DNF_SERVER": "1",
+            "HERMETO_TEST_LOCAL_NEXUS_PROXY": "1",
```
eskultety, line 126: With https://github.com/hermetoproject/hermeto/pull/1283 merged we might want to also add a pytest custom CLI option since pytest does autocomplete.

tests/integration/test_pip.py
```diff
@@ -126,6 +126,10 @@
                 packages=({"path": ".", "type": "pip", "binary": {}},),
                 expected_exit_code=0,
                 expected_output="All dependencies fetched successfully",
+                netrc_content=os.getenv(
+                    "HERMETO_TEST_NETRC_CONTENT",
+                    "machine 127.0.0.1 login cachi2-user password cachi2-pass",
+                ),
```
eskultety: this is where the CI is failing due to https://github.com/hermetoproject/hermeto/blob/eab41ec99362bb9efd9080862584a2f1bec48153/tests/conftest.py#L51

tests/integration/conftest.py
```diff
@@ -180,20 +182,40 @@ def _check_ssl_configuration() -> None:
...
+    if (os.getenv("CI") and os.getenv("GITHUB_ACTIONS")) or not is_local_nexus_proxy_enabled():
+        yield
+        return
+
+    with contextlib.ExitStack() as context:
+        container, client = initialize_nexus(host=DEFAULT_NEXUS_HOST, port=TEST_NEXUS_PORT)
+        context.enter_context(client)
+
+        # Allows inspection of running Nexus container after test completion
+        if os.getenv("HERMETO_TEST_NEXUS_NO_CLEANUP") == "1":
```
eskultety: naming nitpick: maybe `HERMETO_TEST_LOCAL_NEXUS_NO_CLEANUP`? To keep the root of the name the same? non-blocking though.

It's so cool that you used `--replace` in the start script so that the old instance doesn't hang around in between executions.

taylormadore: Good point, will fix

noxfile.py
taylormadore: Agreed

tests/integration/test_pip.py
taylormadore: I think the right move is just to drop this option for now and hard-code the netrc in the test params.

1. This is only currently used by one test
2. The credentials are already currently hard-coded in the startup script for the server and so can't really be configured this way as-is anyway
3. I want to migrate the dnfserver and pypiserver tests to just use the nexusserver at some point soon anyway. Why have three server fixtures when we can just extend one. We can add something more flexible back if needed at that time

eskultety: 1. and 2. - hard YES!

> 3. Why have three server fixtures when we can just extend one. 

Because the DNF server is doing TLS cert auth, can you simulate the same thing with Nexus?

taylormadore: We might need a small spike first, but I think it will work

eskultety: Ok, I'd be interested to see whether we can make it so that we don't need to always assume the proxy mode for that to keep testing the non-proxy code paths (however applicable)

Issue #1208, opened 2025-12-12
Platform-specific gem downloads fail due to platform normalization mismatch


taylormadore:
When Hermeto parses a Gemfile.lock containing platform-specific gems, it may fail to download gems from the registry because the platform string used in the download URL doesn't match the platform under which the gem was originally published. A comment in [gem_models.py](https://github.com/hermetoproject/hermeto/blob/c68c6affe4ef6ee14b0da6be5f4e1a5769ae6388/hermeto/core/package_managers/bundler/gem_models.py#L99) attributes this to a Ruby limitation, but the issue is that Hermeto doesn't preserve the original platform string from the lockfile.

Hermeto uses Bundler's [LockfileParser](https://github.com/ruby/rubygems/blob/8c351fab41142a2cbe68776e41aff09c6902c85e/bundler/lib/bundler/lockfile_parser.rb#L6) to parse Gemfile.lock files. During parsing, platform strings are converted to [Gem::Platform](https://github.com/ruby/rubygems/blob/8c351fab41142a2cbe68776e41aff09c6902c85e/lib/rubygems/platform.rb#L8) objects, which normalize the platform representation. When Hermeto later [constructs the download URL](https://github.com/hermetoproject/hermeto/blob/c68c6affe4ef6ee14b0da6be5f4e1a5769ae6388/hermeto/core/package_managers/bundler/gem_models.py#L89) for a platform-specific gem, it uses this normalized platform string. This doesn't account for the fact that gems are published to registries with their original platform names. When normalization changes the platform string, the download fails because the requested gem variant doesn't exist.

Bundler handles this scenario correctly. The [RemoteSpecification](https://github.com/ruby/rubygems/blob/8c351fab41142a2cbe68776e41aff09c6902c85e/bundler/lib/bundler/remote_specification.rb#L8) class maintains both representations:
- **original_platform**: the exact string from the lockfile, used for fetching
- **platform**: normalized Gem::Platform object, used for matching/comparison

When downloading, Bundler uses original_platform to ensure the URL matches what's in the registry.

eskultety:
Based on the discussion in #1321, most notably https://github.com/hermetoproject/hermeto/pull/1321#issuecomment-4057009512, closing this issue as no longer needed.

PR #1321, opened 2026-02-25
fix(bundler): preserve original platform strings for gem downloads

shokakucarrier:
Hermeto was using normalized platform strings from Gem::Platform when constructing download URLs for platform-specific gems. This caused downloads to fail because gems are published with their original platform names (e.g., "aarch64-linux-gnu" vs normalized "aarch64-linux").

The lockfile parser now tracks both the normalized platform (for matching) and the original platform string (for downloads), matching Bundler's RemoteSpecification behavior.

Fixes: https://github.com/hermetoproject/hermeto/issues/1208

### DIFF
--- a/hermeto/core/package_managers/bundler/parser.py
+++ b/hermeto/core/package_managers/bundler/parser.py
@@ -57,7 +57,7 @@ def parse_lockfile(
     result: ParseResult = []
     for dep in dependencies:
         if dep["type"] == "rubygems":
-            for platform in dep["platforms"]:
+            for platform, original_platform in zip(dep["platforms"], dep["original_platforms"]):
                 if platform == "ruby":
                     result.append(GemDependency(**dep))
                 else:
@@ -68,7 +68,12 @@ def parse_lockfile(
                             "Will download binary dependency %s because 'binary' field is set",
                             full_name,
                         )
-                        result.append(GemPlatformSpecificDependency(platform=platform, **dep))
+                        result.append(
+                            GemPlatformSpecificDependency(
+                                platform=original_platform,
+                                **dep,
+                            )
+                        )
                     else:
                         # No need to force a platform if we skip the packages.
                         log.warning(
--- a/hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
+++ b/hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
@@ -3,8 +3,28 @@
 require 'bundler'
 require 'json'
 
+class HermetoLockfileParser < Bundler::LockfileParser
+  attr_reader :original_platform
+
+  def initialize(lockfile)
+    @original_platform = {}
+    super
+  end
+
+  private
+
+  def parse_spec(line)
+    super
+
+    return unless line =~ NAME_VERSION
+    if @current_spec && $1.size == 4
+      @original_platform[@current_spec] = $4 || Gem::Platform::RUBY
+    end
+  end
+end
+
 lockfile_content = File.read("Gemfile.lock")
-lockfile_parser = Bundler::LockfileParser.new(lockfile_content)
+lockfile_parser = HermetoLockfileParser.new(lockfile_content)
 
 parsed_specs = []
 
@@ -16,7 +36,8 @@ lockfile_parser.specs.each do |spec|
         version: spec.version,
         type: 'rubygems',
         source: spec.source.remotes.first,
-        platforms: [spec.platform]
+        platforms: [spec.platform],
+        original_platforms: [lockfile_parser.original_platform[spec]]
       }
 
       existing_spec = parsed_specs.find { |s|
@@ -27,8 +48,9 @@ lockfile_parser.specs.each do |spec|
       }
 
       if existing_spec
-        # extend the platforms array
+        # extend the platforms arrays
         existing_spec[:platforms] << parsed_spec[:platforms].first
+        existing_spec[:original_platforms] << parsed_spec[:original_platforms].first
       else
         parsed_specs << parsed_spec
       end
--- a/tests/unit/package_managers/bundler/test_parser.py
+++ b/tests/unit/package_managers/bundler/test_parser.py
@@ -152,6 +152,7 @@ def test_parse_gemlock(
             "type": "rubygems",
             "source": "https://rubygems.org/",
             "platforms": ["ruby"],
+            "original_platforms": ["ruby"],
             **base_dep,
         },
     ]
@@ -346,6 +347,7 @@ def test_parse_gemlock_detects_binaries_and_adds_to_parse_result_when_allowed_to
             "type": "rubygems",
             "source": "https://rubygems.org/",
             "platforms": ["i8080_cpm"],
+            "original_platforms": ["i8080_cpm"],
             **base_dep,
         },
     ]
@@ -384,6 +386,7 @@ def test_parse_gemlock_detects_binaries_and_skips_then_when_instructed_to_skip(
             "type": "rubygems",
             "source": "https://rubygems.org/",
             "platforms": ["i8080_cpm"],
+            "original_platforms": ["i8080_cpm"],
             **base_dep,
         },
     ]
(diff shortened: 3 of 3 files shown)


hermeto/core/package_managers/bundler/parser.py
```diff
@@ -68,7 +68,12 @@ def parse_lockfile(
                             "Will download binary dependency %s because 'binary' field is set",
                             full_name,
                         )
-                        result.append(GemPlatformSpecificDependency(platform=platform, **dep))
+                        result.append(
+                            GemPlatformSpecificDependency(
+                                platform=original_platform,
+                                **dep,
+                            )
+                        )
```
slimreaper35, line 76: So we will need only the "original platform" for pre-compiled gems?

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -3,8 +3,28 @@
...
+    super
+  end
+
+  private
+
+  def parse_spec(line)
+    super
+
+    line =~ NAME_VERSION
+    if @current_spec && $1.size == 4
```
taylormadore, line 20: If `NAME_VERSION `doesn't match, `$1` might be undefined or whatever the Ruby equivalent is. I see the parent handles that here: https://github.com/ruby/rubygems/blob/8c351fab41142a2cbe68776e41aff09c6902c85e/bundler/lib/bundler/lockfile_parser.rb#L269

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -3,8 +3,28 @@
...
+    @original_platform = {}
+    super
+  end
+
+  private
+
+  def parse_spec(line)
+    super
+
+    line =~ NAME_VERSION
```
taylormadore: Much of the reference lockfile_parser.rb has been stable for many years, but I do worry about borrowing too many of the parsing internals. I'm curious what others think

hermeto/core/package_managers/bundler/parser.py
taylormadore: I think so. The standard ruby gems don't have a platform defined in Gemfile.lock at all IIRC

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
shokakucarrier: OK, I thought it will not execute if parent method returned, will change this

hermeto/core/package_managers/bundler/parser.py
shokakucarrier: yes the original_platform in most case will be exactly the same string as platform since these are extracted from cmd output.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
shokakucarrier: fixed

hermeto/core/package_managers/bundler/parser.py
slimreaper35: The `platforms` array is now only used for checking if it is equal to `ruby`. Maybe it could be possible to drop it from the script.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
shokakucarrier: That might need overwrite the parse_spec method instead of just add these couple of lines, I'm OK with it just don't know how the team think

hermeto/core/package_managers/bundler/parser.py
shokakucarrier: I am trending to keep platform as ruby bundler intended to, maybe they will be used in future record/report etc.

slimreaper35: > I am trending to keep platform as ruby bundler intended to, maybe they will be used in future record/report etc.

Future potential needs is not a strong argument. But I don't have any suggested code snippet because it is Ruby 😄

harshakumar25:
One thing I noticed — this fixes the download URL but the SBOM PURL still gets built from the normalized platform string, so it'll report `aarch64-linux` even when the gem fetched is `aarch64-linux-gnu`. Filed #1400 to track that separately.

eskultety:
As I was reviewing this I admittedly hit a point where I simply gave up on reading Ruby as the dots stopped connecting for me when I went back and forth to the rubygems code base, so I fired up Claude. Here's the gist and correct me wherever I'm wrong. #1208 was filed back when we shipped with UBI9, we now ship with UBI10.
```
# UBI9
$ dnf install ruby
Installing:
 ruby  x86_64       >>>>3.0.7-165.el9_5<<<<       ubi-9-appstream-rpms
...
```
```
# UBI10
$ dnf install ruby
Installing:
 ruby  x86_64   >>>>>>3.3.10-11.el10_1<<<<<<   ubi-10-for-x86_64-appstream-rpms
...
```
I then took the changes from this PR and applied the parser on the following fake Gemfile.lock in pristine Ruby 3.0 and Ruby 3.3 environments:
```
$ cat /tmp/Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    fakegem (1.0.0)
    fakegem (1.0.0-i686-linux)
    fakegem (1.0.0-aarch64-linux-gnu)
    fakegem (1.0.0-aarch64-linux-musl)
    fakegem (1.0.0-x86_64-darwin)

PLATFORMS
  ruby

DEPENDENCIES
  fakegem

BUNDLED WITH
   2.5.10

############################# WITH RUBY 3.0
$ podman run --rm -v /tmp/Gemfile:/app/Gemfile:Z -v /tmp/Gemfile.lock:/app/Gemfile.lock:Z -v /tmp/lockfile_parser.rb:/app/lockfile_parser.rb:Z ruby:3.0-buster ruby lockfile_parser.rb
{
  "bundler_version": "2.5.10",
  "dependencies": [
    {
      "name": "fakegem",
      "version": "1.0.0",
      "type": "rubygems",
      "source": "https://rubygems.org/",
      "platforms": [
        "ruby",
        "aarch64-linux", <<<<<<<<<<<<<<<<
        "aarch64-linux-musl",
        "x86-linux",
        "x86_64-darwin"
      ],
      "original_platforms": [
        "ruby",
        "aarch64-linux-gnu", <<<<<<<<<<<<<<<
        "aarch64-linux-musl",
        "i686-linux",
        "x86_64-darwin"
      ]
    }
  ]
}


############################### WITH RUBY 3.3
$  podman run --rm -v /tmp/Gemfile:/app/Gemfile:Z -v /tmp/Gemfile.lock:/app/Gemfile.lock:Z -v /tmp/lockfile_parser.rb:/app/lockfile_parser.rb:Z -w /app ruby:3.3-bookworm ruby lockfile_parser.rb
{
  "bundler_version": "2.5.10",
  "dependencies": [
    {
      "name": "fakegem",
      "version": "1.0.0",
      "type": "rubygems",
      "source": "https://rubygems.org/",
      "platforms": [
        "ruby",
        "aarch64-linux-gnu", <<<<<<<<<<<<<<<
        "aarch64-linux-musl",
        "x86-linux",
        "x86_64-darwin"
      ],
      "original_platforms": [
        "ruby",
        "aarch64-linux-gnu", <<<<<<<<<<<<<<
        "aarch64-linux-musl",
        "i686-linux",
        "x86_64-darwin"
      ]
    }
  ]
}
```
It looks like Ruby 3.3's Bundler parser no longer does this normalization, so I dug deeper. I found https://github.com/ruby/rubygems/commit/7e976d790a6780bb8c642b1a1182f05ddbb99a55 which seemingly removes the `-gnu` rule. The related PR for that commit is https://github.com/ruby/rubygems/pull/5852. That bundler parser change landed in Bundler 2.4.0 which seems to have started distributing with Ruby 3.2:
```
$  podman run -it --rm ruby:3.2-bookworm bundler --version
Bundler version 2.4.19
```

@taylormadore @slimreaper35 My questions therefore are:

- based on the fact that we ship with Ruby 3.3 and Bundler 2.5.X and the fact that #1321 was filed only for the `-gnu` problem, do we still need this PR since Bundler now correctly reports linux GNU platform variants?
- it is true that not even the latest Bundler version can handle `iX86` 32bit platform variants and those are still normalized, BUT I only found a handful of packages with 32bit platform tags which are old (2009-2012) and technically bundler on Ruby 3.3+ isn't even capable of creating a lockfile for these 32bit packages, because they simply don't match the host:
```
$ cat Gemfile
source "https://rubygems.org"
gem "camcapture"
$  podman run --rm -v /tmp/Gemfile:/app/Gemfile:Z -w /app ruby:3.3-bookworm sh -c 'rm -f Gemfile.lock && bundle lock --add-platform i686-linux && cat Gemfile.lock'
Fetching gem metadata from https://rubygems.org/.
Could not find gems matching 'camcapture' valid for all resolution platforms
(x86_64-linux, x86-linux) in rubygems repository https://rubygems.org/ or
installed locally.

The source contains the following gems matching 'camcapture':
  * camcapture-0.1.0-x86-linux
  * camcapture-0.2.0-x86-linux
  * camcapture-0.3.0-x86-linux
```

shokakucarrier:
@eskultety I noticed that when I tested platform string formation, this is actually fixed upstream, so could you please discuss this PR with team cuz I cannot join the meeting now.

slimreaper35:
Well, that's a strong justification.

taylormadore:
@eskultety  Nice digging :100: 

I'm fine if we drop this :+1:

eskultety:
Based on the reactions in https://github.com/hermetoproject/hermeto/pull/1321#issuecomment-4073761228 and https://github.com/hermetoproject/hermeto/pull/1321#issuecomment-4118467006 --> closing as not needed anymore. We can reopen as necessary.

PR #1323, opened 2026-02-25
Handle Rust dependencies for Python sdists if Cargo.lock is missing

slimreaper35:
Closes https://github.com/hermetoproject/hermeto/issues/1297

### DIFF
--- a/tests/integration/test_cargo.py
+++ b/tests/integration/test_cargo.py
@@ -56,6 +56,29 @@ log = logging.getLogger(__name__)
             ),
             id="cargo_uses_resolver_v3",
         ),
+        pytest.param(
+            utils.TestParameters(
+                branch="cargo/missing-lockfile",
+                packages=({"path": ".", "type": "cargo"},),
+                check_output=False,
+                check_deps_checksums=False,
+                expected_exit_code=2,
+                expected_output="Cargo.lock not found",
+            ),
+            id="cargo_missing_lockfile",
+        ),
+        pytest.param(
+            utils.TestParameters(
+                branch="cargo/missing-lockfile",
+                packages=({"path": ".", "type": "cargo"},),
+                global_flags=["--mode", "permissive"],
+                check_output=False,
+                check_deps_checksums=False,
+                expected_exit_code=0,
+                expected_output="All dependencies fetched successfully",
+            ),
+            id="cargo_missing_lockfile_permissive_mode",
+        ),
     ],
 )
 def test_cargo_packages(


tests/unit/package_managers/pip/test_main.py
```diff
@@ -1623,7 +1621,6 @@ def test_fetch_pip_source_correctly_reraises_when_there_is_a_dependency_cargo_lo
         returncode=101,
         stderr="... failed to sync ... because --locked was passed to prevent this ...",
     )
-    mock_verify_lockfile_present.return_value = None
```
eskultety, line 1626: We should also adjust our integration test suite to include this scenario, right now we only test for Cargo.lock and Cargo.toml mismatch with a pip extension.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -189,18 +188,6 @@ def _resolve_main_package(package_dir: RootedPath) -> tuple[str, str | None]:
     return name, version
 
 
-def _verify_lockfile_is_present_or_fail(package_dir: RootedPath) -> None:
```
mahmoudnasser1561, line 229: could you please tell me why that function was deleted and how did it contribute to the original Bug ?

just reviewing to learn

hermeto/core/package_managers/cargo/main.py
```diff
@@ -189,18 +188,6 @@ def _resolve_main_package(package_dir: RootedPath) -> tuple[str, str | None]:
...
-    # Most packages will be locked, however metapackages (i.e. those, which
-    # contain just a workspace and could even lack a name) could arrive without
-    # a lock file. A user could try and fix this by explicitly locking the
-    # package first.
-    if not (package_dir.path / "Cargo.lock").exists():
-        raise LockfileNotFound(
-            files=package_dir.path / "Cargo.lock",
-            solution="Please lock it first by running 'cargo generate-lockfile'",
```
a-ovchinnikov, line 237: In the hindsight this was a poor messaging: a user should have reached to maintainers to ensure _they_ lock the package. WDYT about retaining the check, but making it conditional on permissive mode? I.e. if the mode is permissive we just regenerate the missing lock and issue a warning, if the mode is strict we fail as usual.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: To @mahmoudnasser1561 : Please note, that the issue has never been a bug. As stated in its description some users found this behavior overly strict and requested an official workaround. Now to answer your original question I would suggest running Hermeto on a sample cargo project with and without lockfile, first when `main` is checked out then when this PR is checked out and comparing results. You'll get four datapoints. Once you have this data and if you still have questions then please feel free to reach out

hermeto/core/package_managers/cargo/main.py
taylormadore: I was going to add the same comment. I think I prefer to fail-fast if we aren't in permissive mode :+1:

If we take this route, we should probably expand the error message to include mention of permissive mode

hermeto/core/package_managers/pip/rust.py
```diff
@@ -53,7 +53,11 @@ def _depends_on_rust(extracted_dir: Path) -> bool:
     return False
 
 
-def _get_rust_root_dir(cargo_lockfiles: Iterable[Path]) -> Path:
+def _shortest_path_parent(files: Iterable[Path]) -> Path:
+    return min(files, key=lambda x: len(x.parts)).parent
+
+
+def _get_rust_root_dir(source_dir: Path) -> Path | None:
     """Find the top-most Cargo.lock in the package.
```
taylormadore: Docstring needs an update

hermeto/core/package_managers/cargo/main.py
eskultety: The error we'd see from a plain `cargo vendor` is also not very user friendly compared to the original check:
```
    Updating crates.io index
    Updating git repository `https://github.com/rust-random/rand`
error: failed to sync

Caused by:
  failed to load lockfile for /home/eskultety/git/hermeto-integration-tests.git/tmp8ufdsxdk.hermeto-source-copy

Caused by:
  cannot create the lock file /home/eskultety/git/hermeto-integration-tests.git/tmp8ufdsxdk.hermeto-source-copy/Cargo.lock because --locked was passed to prevent this
  help: to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
2026-02-25 16:19:05,356 ERROR PackageWithCorruptLockfileRejected: /home/eskultety/git/hermeto-integration-tests.git/tmp8ufdsxdk.hermeto-source-copy contains a Cargo.lock that does not match the corresponding Cargo.toml
Error: PackageWithCorruptLockfileRejected: /home/eskultety/git/hermeto-integration-tests.git/tmp8ufdsxdk.hermeto-source-copy contains a Cargo.lock that does not match the corresponding Cargo.toml
  Consider reaching out to maintainer of the dependency in question to address inconsistencies between Cargo.lock and Cargo.toml
```

The only thing ^this would achieve is "bugs" being filed on how we interact with cargo.
+1 on failing fast.

hermeto/core/package_managers/cargo/main.py
taylormadore, line 370: We might need to update this warning message

hermeto/core/package_managers/cargo/main.py
slimreaper35: > if the mode is permissive we just regenerate the missing lock and issue a warning, if the mode is strict we fail as usual.

Okay, makes sense

> In the hindsight this was a poor messaging: a user should have reached to maintainers to ensure they lock the package.

But it can also happen that a user is the maintainer and just somehow forgot to update the lockfile. So maybe a generic message is more suitable with a suggestion to use permissive mode.

hermeto/core/package_managers/cargo/main.py
taylormadore, line 381: If we don't keep the fail-fast check for Cargo.lock, this error message needs an update

hermeto/core/package_managers/pip/rust.py
```diff
@@ -65,7 +69,13 @@ def _get_rust_root_dir(cargo_lockfiles: Iterable[Path]) -> Path:
...
     path in pyproject.toml config section for maturin or setuptools-rust.
     """
-    return min(cargo_lockfiles, key=lambda x: len(x.parts)).parent
+    if lockfiles := list(source_dir.rglob("Cargo.lock")):
+        return _shortest_path_parent(lockfiles)
+
+    if tomlfiles := list(source_dir.rglob("Cargo.toml")):
+        return _shortest_path_parent(tomlfiles)
+
+    return None
```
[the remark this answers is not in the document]
slimreaper35: Not really, just converted the iterator of Cargo.lock files to list + nested try block is ugly

slimreaper35:
Thank you for the feedback. I understand the concern. Once the design of the flow is fully complete , i will try to come up with the loom video with missing Cargo.lock execution. Looking forward to your thoughts on it.

tests/unit/package_managers/pip/test_main.py
slimreaper35: https://github.com/hermetoproject/integration-tests/pull/32

hermeto/core/package_managers/cargo/main.py
slimreaper35: done

tests/unit/package_managers/pip/test_rust.py
```diff
@@ -33,6 +33,25 @@
...
-
+    inferred_rust_root_dir = _shortest_path_parent(cargo_files)
     assert inferred_rust_root_dir == expected_rust_root_dir
+
+
+def test_get_rust_root_dir_returns_none_if_no_rust_files_exist(tmp_path: Path) -> None:
+    assert _get_rust_root_dir(tmp_path) is None
+
+
+def test_get_rust_root_dir_falls_back_to_cargo_toml(tmp_path: Path) -> None:
```
a-ovchinnikov, line 44: These names are neat!

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov, line 1: _Commit message_ (https://github.com/hermetoproject/hermeto/commit/01eb9af2ee4be8dfb7b6d9317f4549a5d9fb9179): I won't call it "too strict", it was OK strict, I'd say the problem lies with users who do not care if their builds are reproducible or not.

a-ovchinnikov (approved):

hermeto/core/package_managers/cargo/main.py
slimreaper35: rephrased

PR #1325, opened 2026-02-27
docs(sbom): documented sbom output of hermeto

### LINKED ISSUE #1012: Document our SBOM
We don't provide any user facing documentation on our SBOM. We simply assume that everyone already knows what it is and what the format is, but the truth is we don't really disclose upfront the following which may be a source of confusion for anyone who dives into it and tries to understand & process it:

- what kind of scheme we use (flat list vs nested tree of components)
- what custom fields we report
- how we report components (e.g. we report all? things as a library, is there a better fit perhaps?)
- our perception of the "main" package and its versioning
- why don't we sometimes report a version for a package even though there technically might be a value for us to report
- how do we map Cyclonedx values to SPDX?
- if we support 2 standards concurently, can we convert between those 2 arbitrarily? etc.

ST3V1K:
This should provide a basic overview for users of hermeto about the created SBOM. Until now we just assumed that everyone knows what SBOMs are and what is the format we output.

Closes https://github.com/hermetoproject/hermeto/issues/1012

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,8 @@ hermeto fetch-deps \
 The `fetch-deps` command fetches your project's dependencies and stores them on
 your disk. Hermeto also produces a detailed SBOM containing information about
 all the project's components and packages. You can find the SBOM in the output
-directory.
+directory. See `docs/sbom.md` for SBOM structure, custom fields, and
+CycloneDX/SPDX mapping.
 
 See `docs/usage.md` for a more detailed, practical example of Hermeto usage.
 
--- a/docs/sbom.md
+++ b/docs/sbom.md
@@ -1,0 +1,198 @@
+# SBOM
+
+Hermeto produces a Software Bill of Materials (SBOM) describing the dependencies
+it fetches. This document explains how Hermeto structures the SBOM, what custom
+fields it uses, how it maps between supported formats, and how conversion works.
+
+Hermeto supports the following SBOM formats:
+
+- **CycloneDX** [v1.6][] (default)
+- **SPDX** [v2.3][]
+
+The output format can be selected with `--sbom-output-type cyclonedx` or
+`--sbom-output-type spdx` when running `hermeto fetch-deps` or
+`hermeto merge-sboms`.
+
+Hermeto represents SBOMs internally as CycloneDX and can convert them to SPDX.
+Both formats use only a subset of their respective specifications; the subsets
+map into each other. The mapping is documented in
+[CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping).
+
+## Scheme
+
+Hermeto produces a flat list of components (CycloneDX) or packages (SPDX). There
+is no nested tree of dependencies in either format.
+
+### CycloneDX
+
+All packages processed by Hermeto are represented by CycloneDX components. In a
+SBOM produced by Hermeto components form a single flat top-level `components` array,
+no hierarchy is preserved withing it. Of all possible component types available
+in CycloneDX spec Hermeto uses only two component types: `library` (default, used
+for components produced by all package managers except generic) and `file` (used
+for URL-based generic artifacts such as archives when the generic package manager
+fetches a zip/tarball).
+
+### SPDX
+
+At the time of writing, SBOMs are represented internally as CycloneDX. SPDX is
+produced with a converter (see [CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)).
+SPDX SBOMs provided by Hermeto have the following properties:
+
+- The output has a `packages` array and a `relationships` array.
+- `packages`: the first entry is a synthetic root with
+  `SPDXID`:&nbsp;`SPDXRef-DocumentRoot-File-`
+  (empty `name` and `versionInfo`); the rest are SPDX package entries for actual
+  packages. The root is required by the SPDX specification.
+- `relationships`: the document describes the root
+  (`SPDXRef-DOCUMENT` &rarr; `SPDXRef-DocumentRoot-File-` via `DESCRIBES`);
+  every real package is linked to the root via `CONTAINS`.
+- Unlike CycloneDX SPDX does not have direct equivalents to "library" or "file"
+  component types; all packages are listed without that distinction. As a result,
+  a `file` component in CycloneDX that is converted to SPDX and back will become
+  a `library` component &mdash; this is a known limitation.
+
+## Custom data Hermeto reports
+
+### Dependency properties
+
+Hermeto attaches properties from a fixed set to all dependencies. These
+properties carry metadata about dependencies they are attached to. These
+properties are stored in the CycloneDX `components[].properties` array; in SPDX
+they are represented as package annotations (see
+[CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)). Property names use
+`hermeto:` as a prefix, except for the two [standard CycloneDX npm properties][]
+which use the `cdx:` prefix defined by the CycloneDX specification. The table
+below lists all custom properties:
+
+| Property name | Meaning | When it appears |
+| ------------- | ------- | --------------- |
+| `hermeto:found_by` | Tool that added the dependency | Always |
+| `hermeto:missing_hash:in_file` | Path to a lockfile with a missing checksum | When there is a lockfile with some checksums missing |
+| `hermeto:bundler:package:binary` | Bundler gem is a platform-specific binary | When the gem is a platform-specific binary |
+| `cdx:npm:package:bundled` | Npm package is bundled | When the package is bundled within another |
+| `cdx:npm:package:development` | Npm package is a dev dependency | When the package is a development dependency |
+| `hermeto:pip:package:binary` | Pip package is a binary wheel | When the package is a binary wheel |
+| `hermeto:pip:package:build-dependency` | Pip package is a build-time dependency | When the package is required at build time |
+| `hermeto:rpm_modularity_label` | RPM modularity label | When the RPM has a modularity label |
+| `hermeto:rpm_summary` | RPM summary | [When you set `include_summary_in_sbom` to `true`](rpm.md#basic-options) |
+
+### Backend and source information
+
+Unlike the dependency properties above, which are stored as flat key/value pairs
+in `components[].properties` (CycloneDX) or package annotations (SPDX), backend
+and source information use other SBOM fields because they describe where a
+component came from, not arbitrary key/value metadata on the component.
+
+**Backends** &mdash; lets consumers know which backend (e.g. gomod, pip, npm)
+produced each dependency. Experimental backends (name starting with `x-`) are
+tagged as experimental. Supported SBOM types represent them like follows:
+
+- **CycloneDX**: Top-level annotations. Text format `hermeto:backend:<backend_name>`
+  (e.g. `hermeto:backend:gomod`); for experimental backends,
+  `hermeto:backend:experimental:<backend_name>`. Each annotation references a
+  component by `bom-ref` (which equals the component's `purl`).
+- **SPDX**: Those backend tags become package annotations on the matching
+  packages (see [CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)).
+
+**Actual source URL** &mdash; When a dependency was downloaded via an artifact
+repository manager (e.g. JFrog Artifactory or Sonatype Nexus), Hermeto records
+the real download location as well as the usual distribution reference. If
+several URLs are known, all are recorded; the SBOM does not say which URL was
+actually used (some backends cannot provide that). It is represented as follows:
+
+- **CycloneDX**: An external reference with `type`:&nbsp;`"distribution"`,
+  `comment`:&nbsp;`"proxy URL"`, and `url` set to the download location. Multiple
+  URLs are multiple such references.
+- **SPDX**: The same information is mapped to the package field `sourceInfo`
+  (semicolon-separated when there are several).
+
+## Main package representation
+
+The SBOM does **not** mark which dependency is the main package. All are
+listed as peers. The SBOM alone does not distinguish "the app" from its
+dependencies.
+
+## Versioning
+
+Hermeto provides several versioning mechanisms in SBOMs:
+
+- **Version**: Each dependency can have a version, taken from the package
+  manager's data (lockfile, metadata, or similar). Some dependencies have no
+  version; Hermeto omits the field in that case. For further details see
+  [Reasons for lack of versions in some dependencies](#reasons-for-lack-of-versions-in-some-dependencies).
+- **PURL**: The main identifier for each dependency is its PURL (package URL).
+  The PURL can still include version-like information (e.g. a Git commit) even
+  when the version field is missing.
+- **SBOM format version**: The SBOM file itself carries a format version
+  &mdash; the schema version, not the application's version.
+  CycloneDX:&nbsp;`specVersion` (e.g. `"1.6"`) and a top-level integer `version`
+  (BOM revision). SPDX:&nbsp;`spdxVersion` (e.g. `"SPDX-2.3"`).
+
+### Reasons for lack of versions in some dependencies
+
+Some dependencies in the SBOM may have no version; when and why depends on the
+ecosystem:
+
+- **Go (gomod)**: The standard library version is tied to the Go runtime, but
+  Hermeto omits it because the Go version used during pre-fetching may differ
+  from the one used in the actual hermetic build; recording it could lead
+  vulnerability analyzers to incorrectly flag versions and to incorrectly apply
+  a particular provenance policy. Modules referenced via `replace` directives
+  pointing to local paths may also lack a version; other modules get one when
+  resolved.
+- **Yarn**: Workspace and linked packages often have no version.
+- **Cargo**: Missing if a local path dependency omits it, or if a workspace
+  member inherits from `[workspace.package]` and that field is unset.
+- **npm**: Version is taken from the lock file, however in rare cases it could
+  be missing.
+- **pip**: PyPI packages always have a version; VCS or local path sources may
+  not. May also be absent if the main project uses dynamic versioning
+  (e.g. setuptools-scm).
+
+## Merging SBOMs
+
+Merging is only meant for **SBOMs produced by Hermeto** (for example from
+`fetch-deps` or an earlier successful merge) due to the fact that Hermeto
+supports only subsets of SBOM formats. Input files must match Hermeto's
+CycloneDX or SPDX model; arbitrary SBOMs from other tools are not supported and
+will be rejected as invalid.
+
+## CycloneDX to SPDX mapping
+
+Hermeto forms the SBOM as CycloneDX first; this section describes the
+translation used when SPDX output is requested (e.g. via
+`--sbom-output-type spdx`).
+
+| CycloneDX | SPDX |
+| --------- | ---- |
+| `Component` | `SPDXPackage` |
+| Component identity | `SPDXID`: `SPDXRef-Package-{name}-{version}-{hash}` (or `SPDXRef-Package-{name}-{hash}`). The hash is the SHA-256 hex digest of a JSON object with keys `name`, `version`, and `purl`, serialized with keys sorted. The idstring is sanitized per SPDX rules (only letters, digits, `.`, `-`). `versionInfo` is set to the component's version or omitted when `None`. |
+| `purl` | `externalRefs`: one entry with `referenceCategory=PACKAGE-MANAGER`, `referenceType=purl`, `referenceLocator=<purl>` |
+| `properties` | Package `annotations`: each property encoded as JSON `{"name":"...","value":"..."}` in `comment`; `annotator` is `Tool: hermeto:jsonencoded`. |
+| Top-level `annotations` (by bom-ref) | Per-package annotations containing the same values as CycloneDX top-level annotations: `comment` = annotation text |
+| ExternalReference (`type=distribution`, `comment=proxy URL`) | `sourceInfo` (semicolon-separated if multiple) |
+| (no root in CycloneDX) | A synthetic root package `SPDXRef-DocumentRoot-File-` is created with `name=""` and `versionInfo=""`. The document describes the root; the root contains every package. |
+| `metadata.tools` | `creationInfo.creators` (`Tool:` / `Organization:`) |
+
+## SPDX to CycloneDX
+
+When SPDX is converted to CycloneDX (e.g. when merging an SPDX SBOM with a
+CycloneDX one), the following applies:
+
+- Each SPDX package becomes one or more CycloneDX components. Because CycloneDX
+  allows only one PURL per component, an SPDX package with multiple PURLs in
+  `externalRefs` becomes multiple components.
+- `versionInfo` is passed through as the component's `version`.
+- Annotations whose `annotator` ends with `:jsonencoded` are parsed as
+  properties; others are stored as top-level annotations.
+- `sourceInfo` is converted back to ExternalReferences with `type=distribution`
+  and `comment=proxy URL`.
+
+**Limitation**: An SPDX package that has multiple PURLs in `externalRefs` becomes
+multiple CycloneDX components when converted to CycloneDX, because CycloneDX does
+not support multiple PURLs on one component.
+
+[v1.6]: https://cyclonedx.org/docs/1.6/json
+[v2.3]: https://spdx.github.io/spdx-spec/v2.3/
+[standard CycloneDX npm properties]: https://github.com/CycloneDX/cyclonedx-property-taxonomy/blob/main/cdx/npm.md#cdxnpmpackage-namespace-taxonomy
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -6,7 +6,9 @@ config:
   MD009:
     br_spaces: 0  # Non-breaking spaces are forbidden
     strict: true  # Strict no-trailing spaces
-  MD013: {code_blocks: false}  # Ignore line-length for code blocks
+  MD013:
+    code_blocks: false # Ignore line-length for code blocks
+    tables: false # Ignore line-length for tables
   MD026: {punctuation: ".,;:。，；："}  # Allow "!"
   MD027: {list_items: false}  # Only a single space after blockquote symbol (">")
   # Please do not use "em" and "strong" *outside of* <details> or <summary> blocks
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -52,8 +52,9 @@ hermeto fetch-deps \
 - `--output` the path to the directory where Hermeto will write all output
   `[default: ./hermeto-output]`
 - `--sbom-output-type` the format of generated SBOM, supported values are
-  `cyclonedx` (outputs [CycloneDX v1.6][]) and `spdx` (outputs [SPDX v2.3][])
-  `[default: cyclonedx]`
+  `cyclonedx` (outputs [CycloneDX v1.6][]) and `spdx` (outputs [SPDX v2.3][]).
+  See [SBOM documentation](sbom.md) for structure, custom fields, and format
+  mapping. `[default: cyclonedx]`
 - `{JSON}` specifies a *package* (a directory) within the repository to process
 
 Note that Hermeto does not auto-detect which package managers your project uses.
@@ -145,7 +146,8 @@ hermeto merge-sboms <hermeto_sbom_1.json> ... <hermeto_sbom_n.json>
 
 The subcommand expects at least two SBOMs, all produced by Hermeto, and will
 exit with error otherwise. The reason for this is that Hermeto supports a
-[limited set][] of component [properties][], and it validates that no other
+[limited set][] of component [properties][]
+(documented in [SBOM documentation](sbom.md)), and it validates that no other
 properties exist in the SBOM. By default the result of a merge will be printed
 to stdout. To save it to a file use `-o` option
 
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -35,6 +35,7 @@ theme:
 nav:
   - Home: index.md
   - Usage: usage.md
+  - SBOM: sbom.md
   - Package managers:
       - bundler: bundler.md
       - cargo: cargo.md


docs/sbom.md
```diff
@@ -0,0 +1,194 @@
+# SBOM documentation
+
+Hermeto produces a Software Bill of Materials (SBOM) describing the dependencies
+it fetches. This document explains how our SBOM is structured, what custom
+fields we use, how we map between supported formats, and how conversion works.
```
a-ovchinnikov: 4s/our/the/ ?

@ben-alkov WDYT about converting this sentence to passive voice? It could be just me, but active voice feels a bit misplaced here.

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
...
+
+- **CycloneDX** [v1.6][] (default)
+- **SPDX** [v2.3][]
+
+You can choose the output format with `--sbom-output-type cyclonedx` or
+`--sbom-output-type spdx` when running `hermeto fetch-deps` or `hermeto merge-sboms`.
+
+## Scheme
+
+Hermeto uses a **flat list** of components (CycloneDX) or packages (SPDX). There
```
a-ovchinnikov: I would suggest against using bold font **until** you absolutely **must** grab **attention**. A much less stressful higlighting could be achieved with _italics_. What is the reason for highlighting "flat list"?

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
+# SBOM documentation
+
+Hermeto produces a Software Bill of Materials (SBOM) describing the dependencies
+it fetches. This document explains how our SBOM is structured, what custom
+fields we use, how we map between supported formats, and how conversion works.
+
+Supported formats:
```
a-ovchinnikov:
```diff
- Supported formats:
+ Hermeto supports the following SBOM formats:
```

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
...
+
+- **CycloneDX** [v1.6][] (default)
+- **SPDX** [v2.3][]
+
+You can choose the output format with `--sbom-output-type cyclonedx` or
+`--sbom-output-type spdx` when running `hermeto fetch-deps` or `hermeto merge-sboms`.
+
+## Scheme
+
```
a-ovchinnikov, line 22: 17s/Hermeto uses/Hermeto produces/ WDYT?

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
...
+- The output has a single top-level `components` array.
+- There is no `dependencies` array and no parent/child hierarchy.
+- All components are peers in one list.
+
+### SPDX
+
+- The output has a `packages` array and a `relationships` array.
+- `packages`: the first entry is a synthetic root with
+  `SPDXID`:&nbsp;`SPDXRef-DocumentRoot-File-`
+  (empty `name` and `versionInfo`); the rest are the real packages.
```
a-ovchinnikov: 31s/the real/actual/ ?

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
...
+- `relationships`: shows that every real package in the list is directly linked
+  to the root, using simple "contains" and "describes" relationships without
+  referencing the exact field values.
+
+## Custom fields we report
+
+### Component properties
+
+We attach a fixed set of custom properties to components. All are stored in
```
a-ovchinnikov: 40s/components/dependencies/ Motivation: components are called components in CyDX, in SPDX they are packages, dependencies is a schema-neutral way of referencing entities we deal with.

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
...
+the CycloneDX `components[].properties` array; in SPDX they are represented
+as package annotations (see [CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)).
+Property names use the application name as prefix (e.g. `hermeto:*`; when the
+tool is run as `cachi2`, the prefix is `cachi2:*`).
```
a-ovchinnikov: What do you mean by "when the tool is run as cachi2"?

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
...
+| `cdx:npm:package:bundled` | Npm package is bundled | When the package is bundled within another |
+| `cdx:npm:package:development` | Npm package is a dev dependency | For development dependencies |
+| `<app>:pip:package:binary` | Pip package is a binary wheel | For binary wheel distributions |
+| `<app>:pip:package:build-dependency` | Pip package is a build-time dependency | For packages required at build time |
+| `<app>:rpm_modularity_label` | RPM modularity label | For RPM components when applicable. |
+| `<app>:rpm_summary` | RPM summary | For RPM components when included in SBOM. |
+
+### Annotations
+
+We use CycloneDX top-level **annotations** to tag which backend
```
a-ovchinnikov: I suggest inverting this and the next subsections: WDYT about starting with _what_ we add to a SBOM (information of which PM was used to process a dependency and actual source, i.e. the fact that a dependency was downloaded via a proxy respectively) and then describing _how_ it is done in cases of different supported formats?

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
...
+
+The SBOM does **not** mark which component (CycloneDX) or package (SPDX) is the
+main application. All are listed as peers. You cannot tell from the SBOM alone
+which one is "the app" and which are its dependencies.
+
+### Versioning
+
+- **Version**: Each component can have a version, taken from the package
+  manager’s data (lockfile, metadata, or similar).
+- **Missing version**: If we don’t have a version for a component, we
```
a-ovchinnikov: WDYT about dropping "missing version" bullet completely and adding "Some dependencies can have no version, we report no version in such case. For further details see [the corresponding section][when a component does not have a version]"?

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
...
+| `purl` | `externalRefs`: one entry with `referenceCategory=PACKAGE-MANAGER`, `referenceType=purl`, `referenceLocator=<purl>` |
+| `properties` | Package `annotations`: each property as JSON `{"name":"...","value":"..."}` in `comment`; `annotator` is `Tool: <app_name>:jsonencoded` (e.g. `hermeto` or `cachi2`). |
+| Top-level `annotations` (by bom-ref) | Same package annotations: `comment` = annotation text |
+| ExternalReference (type=distribution, comment=proxy URL) | `sourceInfo` (semicolon-separated if multiple) |
+| (no root in CycloneDX) | We create a synthetic root package `SPDXRef-DocumentRoot-File-` with `name=""`, `versionInfo=""`; DOCUMENT DESCRIBES root; root CONTAINS every package. |
+| `metadata.tools` | `creationInfo.creators` (Tool: / Organization:) |
+
+### SPDX to CycloneDX (reverse mapping)
```
a-ovchinnikov: Why is this a reverse mapping?

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
...
+- Annotations whose `annotator` ends with `:jsonencoded` are parsed as
+properties; others are stored as top-level annotations.
+- `sourceInfo` is converted back to ExternalReferences with type
+“distribution” and comment “proxy URL”.
+
+## Two standards and conversion
+
+- We support **CycloneDX 1.6** and **SPDX 2.3**.
+- Conversion is **bidirectional**: `Sbom.to_spdx()` and `SPDXSbom.to_cyclonedx()`.
+- **Round-trip** (CycloneDX → SPDX → CycloneDX, or SPDX → CycloneDX → SPDX)
```
a-ovchinnikov: L179 -- this was already stated at the top of the document;
L180 -- I'd say this was already mentioned just above;
L181-L182 -- This is assumed and sometimes an implementation detail, I would not recommend putting a spotlight on the fact that something trivial is tested (this could result in a false impression that everything else is not tested which is obviously not true)

docs/sbom.md
```diff
@@ -0,0 +1,194 @@
...
+Hermeto uses a **flat list** of components (CycloneDX) or packages (SPDX). There
+is no nested tree of dependencies in either format.
+
+### CycloneDX
+
+- The output has a single top-level `components` array.
+- There is no `dependencies` array and no parent/child hierarchy.
+- All components are peers in one list.
+
+### SPDX
```
a-ovchinnikov, line 36: I have probably missed that, but do you mention that we support only a subset of SPDX?

a-ovchinnikov (commented):
Please find some comments which I had after the first pass inline.

docs/sbom.md
ST3V1K: Removed it on multiple places. There was pretty much no reason to add that highlight.

docs/sbom.md
ST3V1K: When you run `cachi2 fetch-deps ...` it has the `cachi2` prefix in SBOM so I thought I'd mention it. But it was probably unnecessary and is now removed.

docs/sbom.md
ST3V1K: Sounds much better. Changed it

docs/sbom.md
ST3V1K: Since it's not used that much (if I'm not mistaken), I wanted to "show" it somehow. But it doesn't really make sense the more I think about it. 

I changed it to be the same header as CycDX -> SPDX and removed the reverse mapping

docs/sbom.md
ST3V1K: Removed all of those and kept only the info about merging

docs/sbom.md
ST3V1K: No, it wasn't there. Now I have added it at the top of the file in the supported formats.

docs/sbom.md
ST3V1K: > 4s/our/the/ ?

Changed this.

docs/sbom.md
a-ovchinnikov: Thank you!

docs/sbom.md
a-ovchinnikov: It is not supposed to be run as cachi2, at least not by new users. Some old ones might have some automation in place which they do not want to or cannot easily modify, but they probably already know about SBOMs.

docs/sbom.md
ST3V1K: > ben-alkov WDYT about converting this sentence to passive voice? It could be just me, but active voice feels a bit misplaced here.

Should I change this or is it fine as it is? I'm not really sure from the thumbs up

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+  missing if neither specifies it.
+- **npm**: Version comes from the lockfile; rarely missing.
+- **pip**: PyPI dependencies have a version; dependencies from VCS or other
+  sources may not. The main project’s version comes from its metadata
+  (pyproject.toml, setup.py, etc.).
+
+## CycloneDX to SPDX mapping
+
+When we convert from CycloneDX to SPDX (e.g. via `--sbom-output-type spdx`),
+we use the following mapping.
```
slimreaper35: That seems like our internal (bad) implementation. Not something for users.

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+| `properties` | Package `annotations`: each property as JSON `{"name":"...","value":"..."}` in `comment`; `annotator` is `Tool: <app_name>:jsonencoded` (e.g. `hermeto` or `cachi2`). |
+| Top-level `annotations` (by bom-ref) | Same package annotations: `comment` = annotation text |
+| ExternalReference (type=distribution, comment=proxy URL) | `sourceInfo` (semicolon-separated if multiple) |
+| (no root in CycloneDX) | We create a synthetic root package `SPDXRef-DocumentRoot-File-` with `name=""`, `versionInfo=""`; DOCUMENT DESCRIBES root; root CONTAINS every package. |
+| `metadata.tools` | `creationInfo.creators` (Tool: / Organization:) |
+
+## SPDX to CycloneDX
+
+When converting SPDX back to CycloneDX (e.g. when merging an SPDX SBOM with a
+CycloneDX one):
```
slimreaper35: Same comment as above.

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+“distribution” and comment “proxy URL”.
+
+**Limitation**: An SPDX package that has multiple purls in `externalRefs`
+becomes multiple CycloneDX components when converted to CycloneDX, because
+CycloneDX does not support multiple purls on one component.
+
+## Merging SBOMs
+
+When you merge SBOMs, we merge in CycloneDX form. If you pass
+an SPDX SBOM to `merge-sboms`, it is converted to CycloneDX first, then
```
slimreaper35: Same comment as above.

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
+# SBOM documentation
```
slimreaper35:
```suggestion
# SBOM
```

docs/sbom.md
ST3V1K: Is this about the sentence or the whole category? Because info about mapping between CycDX and SPDX was requested in the [issue](https://github.com/hermetoproject/hermeto/issues/1012). Or have I misinterpreted the request?
> how do we map Cyclonedx values to SPDX?

slimreaper35: Just the sentence. The category is fine :)

docs/sbom.md
a-ovchinnikov: yThumbs-up were for the changed part, I myself am still unsure about the passive voice part. @ben-alkov we need your opinion, please weigh in _Edit_: or at the very least it could be switched to third person singular vs second plural (i.e. "Hermeto uses" vs "we use")

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+
+### SPDX
+
+- The output has a `packages` array and a `relationships` array.
+- `packages`: the first entry is a synthetic root with
+  `SPDXID`:&nbsp;`SPDXRef-DocumentRoot-File-`
+  (empty `name` and `versionInfo`); the rest are actual packages.
+- `relationships`: shows that every real package in the list is directly linked
+  to the root, using simple "contains" and "describes" relationships without
+  referencing the exact field values.
```
a-ovchinnikov: I am sorry, which exact field values are not referenced?

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+| `cdx:npm:package:bundled` | Npm package is bundled | When the package is bundled within another |
+| `cdx:npm:package:development` | Npm package is a dev dependency | For development dependencies |
+| `<app>:pip:package:binary` | Pip package is a binary wheel | For binary wheel distributions |
+| `<app>:pip:package:build-dependency` | Pip package is a build-time dependency | For packages required at build time |
+| `<app>:rpm_modularity_label` | RPM modularity label | For RPM dependencies when applicable. |
+| `<app>:rpm_summary` | RPM summary | For RPM dependencies when included in SBOM. |
+
+### Backend and source information we add
+
+We attach two kinds of custom information to dependencies:
```
a-ovchinnikov: It is not clear how these two kinds of custom information differ from nine kinds of custom information attached in the previous subsection. IIRC these two kinds are reported similarly, and this data is missing. As it is now it is not clear why two subsections are needed.

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+
+**CycloneDX**: We use only two component types in the `components` array:
+
+- `library` (default): used for almost all components (gomod, npm, pip,
+  cargo, bundler, rpm, etc.).
+- `file`: used for generic artifacts such as archives (e.g. when the
+  generic package manager fetches a zip/tarball).
+
+We do not use other CycloneDX component types (e.g. `application`, `container`,
+`firmware`).
```
a-ovchinnikov: Why do we need examples of things which we do not use?

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+
+**SPDX**: The SPDX output does not encode an
+equivalent “library” vs “file” type per package; all packages are listed
+without that distinction.
```
a-ovchinnikov: This sounds like a bug to me: judging from this description if A is a CyDX with file then A.toSpdx().toCyDx() =/= A. Is it so?

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+- **Package manager**: The backend annotations become package annotations on the
+  corresponding packages (see [CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)).
+- **Proxy URL**: Mapped to the package field `sourceInfo` (semicolon-separated
+  if there are several).
+
+## How we report dependencies (library vs file)
+
+**CycloneDX**: We use only two component types in the `components` array:
+
+- `library` (default): used for almost all components (gomod, npm, pip,
```
a-ovchinnikov: s/for almost all components(.*)/used for components produced by all package managers except generic/

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+  [the corresponding section](#when-a-dependency-has-no-version).
+- **Purl**: The main identifier for each dependency is its purl (package URL).
+  The purl can still include version-like information (e.g. a Git
+  commit) even when the version field is missing.
+- **SBOM format version**: The SBOM file itself has a format version
+  &mdash; the schema version, not your application’s version.
+  CycloneDX:&nbsp;`specVersion` (e.g. `"1.6"`) and a top-level integer `version`
+  (BOM revision). SPDX:&nbsp;`spdxVersion` (e.g. `"SPDX-2.3"`).
+
+### When a dependency has no version
```
a-ovchinnikov: "When a dependency has no version" reads as an opening before some continuation: "when the roof is on fire call 911", "when there are to many PRs in the queue send the bat signal" etc. Here you just list possible reasons, the title should be modified accordingly. I would have used a simple approach: "Reasons for lack of versions in dependencies" or, if you insist on starting with "when": "When would a dependency lack a version?" (but that's actually worse than the simple version)

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+
+- **The ecosystem doesn’t provide one**: e.g. Go’s standard library packages
+  have no version in the Go toolchain.
+- **We don’t treat the value as a version**: e.g. Yarn uses a placeholder for
+  workspace or linked packages; we omit it so the SBOM doesn’t show a fake
+  version.
+- **The source doesn’t define it**: e.g. a dependency installed from a VCS URL
+  might not have a tagged version.
+
+Per ecosystem:
```
a-ovchinnikov: This section feels redundant: you already mention general rules with some examples, now you repeat yourself with more examples of essentially the same

docs/sbom.md
```diff
@@ -0,0 +1,198 @@
...
+- **Yarn**: Workspace and linked packages often have no version; we omit
+  placeholder values.
+- **Cargo**: Version comes from the package definition or workspace; it can be
+  missing if neither specifies it.
+- **npm**: Version comes from the lockfile; rarely missing.
+- **pip**: PyPI dependencies have a version; dependencies from VCS or other
+  sources may not. The main project’s version comes from its metadata
+  (pyproject.toml, setup.py, etc.).
+
+## CycloneDX to SPDX mapping
```
a-ovchinnikov, line 161: Isn't this something described in a design doc? Won't it be more reliable to send them off to that document instead of copying parts of it over? When we update this mapping in the future we'll have just one place to fix instead of two

docs/sbom.md
ben-alkov: One could make the *technical* distinction between the Project ("we use, we do"), and Hermeto itself ("Hermeto uses"), but I think sticking with Hermeto (*and avoiding passive voice*) is a good rule-of-thumb.

ST3V1K: Used Hermeto throughout the document

docs/sbom.md
ST3V1K: There used to be a sentence after that but was removed in one of the edits and I forgot to remove this.
Removed the "without referencing the exact field values" part

docs/sbom.md
ST3V1K: Removed them

docs/sbom.md
ST3V1K: Added a sentence clarifying why are they in two subsections.

docs/sbom.md
ST3V1K: Don't know if I would call it a bug, but it is definitely something to address in the docs.
Added a note about this behavior.

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done. Changed to "Reasons for lack of versions in dependencies".

docs/sbom.md
ST3V1K: Kept only the info about specific ecosystem, since the first part was only mentioning them and not adding anything different.

docs/sbom.md
ST3V1K: I am not really sure which docs you have in mind. If you mean the ones in `docs/design` I haven't found anything about that + as far as I know they are not included in the final documentation that is published.

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+fields it uses, how it maps between supported formats, and how conversion works.
+
+Hermeto supports the following SBOM formats:
+
+- **CycloneDX** [v1.6][] (default)
+- **SPDX** [v2.3][] &mdash; Hermeto supports only a *subset* of the SPDX
+  specification; the output includes the fields and structure described in this
+  document.
+
+The output format is selected with `--sbom-output-type cyclonedx` or
```
a-ovchinnikov: WDYT about s/is/could be/? With `is` the sentence reads as if one must select a format when in reality that is just an option.

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+## Scheme
+
+Hermeto produces a flat list of components (CycloneDX) or packages (SPDX). There
+is no nested tree of dependencies in either format.
+
+### CycloneDX
+
+- The output has a single top-level `components` array.
+- There is no `dependencies` array and no parent/child hierarchy.
+- All components are peers in one list.
```
a-ovchinnikov: Another optional suggestion. To make CyDX section easier to read it could be reformatted from a list to a sentence:
```
For CycloneDX that means that the output has just a single top-level 
`components` array, there is no `dependencies` array and no p
arent/child hierarchy, all components are peers in one list.
```

Same suggestion holds for SPDX, I think that flowing text is easier to read and comprehend than bullet lists, at least in these cases. (This is open to debate though!)

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+- The output has a `packages` array and a `relationships` array.
+- `packages`: the first entry is a synthetic root with
+  `SPDXID`:&nbsp;`SPDXRef-DocumentRoot-File-`
+  (empty `name` and `versionInfo`); the rest are actual packages. The root is
+  required by the SPDX specification. Every SPDX document must have a document
+  root element that the rest of its contents are related to.
+- `relationships`: every real package is linked directly to the root via
+  `CONTAINS` and `DESCRIBES` relationships.
+
+## Custom fields Hermeto reports
```
a-ovchinnikov: s/fields/data/ ?

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Dependency properties
+
+Hermeto attaches a fixed set of custom properties to dependencies. All are stored
```
a-ovchinnikov: Possible wording improvement:

```diff
- Hermeto attaches a fixed set of custom properties to dependencies. All are stored
+ Hermeto attaches properties from a fixed set to all dependencies. These properties carry metadata about dependencies they are attached to. These properties are stored
```

Reason: the sentence reads a bit like the entire fixed set is attached every time (which is not true), plus the suggested version provides some immediate context to set expectations.

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+in the CycloneDX `components[].properties` array; in SPDX they are represented
+as package annotations (see [CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)).
+Property names use `hermeto:` as a prefix, except for the two
+[standard CycloneDX npm properties][] which use the `cdx:` prefix defined by the
+CycloneDX specification.
+
+| Property name | Meaning | When it appears |
+| ------------- | ------- | --------------- |
+| `hermeto:found_by` | Tool that added the dependency | Always |
+| `hermeto:missing_hash:in_file` | Lockfile path where a hash was missing | When checksum was missing in that file |
```
a-ovchinnikov:
```diff
- Lockfile path where a hash was missing
+ Path to a lockfile with a missing checksum
```

```diff
- When checksum was missing in that file
+ When there is a lock file with some checksums missing
```

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+| `hermeto:bundler:package:binary` | Bundler gem is a platform-specific binary | For platform-specific binary gems |
```
a-ovchinnikov: There is a slight discontinuation: column head says "when X appears", some lines say "For Y" instead of "When Y". This is not critical, but for the best reader's experience I suggest synchronizing these values.

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+| `hermeto:rpm_modularity_label` | RPM modularity label | For RPM dependencies when applicable |
+| `hermeto:rpm_summary` | RPM summary | For RPM dependencies when included in SBOM |
+
+### Backend and source information
+
+Unlike the dependency properties above, which are stored as flat key/value pairs
+in `components[].properties` (CycloneDX) or package annotations (SPDX), the
+following two kinds of information use different SBOM constructs because they
+describe the component's provenance rather than its attributes. See
+[How it is represented](#how-it-is-represented) below for the exact fields used.
```
a-ovchinnikov: Now this is a great passage that makes a lot of sense! My only suggestion would be to drop the last sentence ("See [How..."), it breaks the story and if moved to a spot where it does not interfere (e.g. L81) it becomes redundant.

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+1. **Which package manager processed the dependency**: Consumers can identify
+   which backend (e.g. gomod, pip, npm) produced each dependency. Experimental
+   backends (name starting with `x-`) are tagged as experimental.
+
+2. **Actual source (proxy URL)**: When a dependency was downloaded via a proxy,
+   Hermeto records the actual download location (proxy URL) in addition to any
+   normal distribution reference. Multiple proxy URLs are represented when
+   applicable.
+
+#### How it is represented
```
a-ovchinnikov: s/How it is represented/Backend and source representation in SBOMs/ 

Reason: this way a reader does not need to maintain any additional context when reading table of contents.

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+describe the component's provenance rather than its attributes. See
+[How it is represented](#how-it-is-represented) below for the exact fields used.
+
+Hermeto attaches two kinds of provenance information to dependencies:
+
+1. **Which package manager processed the dependency**: Consumers can identify
+   which backend (e.g. gomod, pip, npm) produced each dependency. Experimental
+   backends (name starting with `x-`) are tagged as experimental.
+
+2. **Actual source (proxy URL)**: When a dependency was downloaded via a proxy,
```
a-ovchinnikov: "Proxy" is unfortunately an overloaded term, thus this passage easily ends up being confusing. In the context of actual source a registry proxy is meant, essentially a cache for packages of certain type. Maybe it would make more sense to replace "proxy" with "artifact repository manager (like JFrog Artifactory or Sonatype Nexus)?

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+   Hermeto records the actual download location (proxy URL) in addition to any
+   normal distribution reference. Multiple proxy URLs are represented when
+   applicable.
```
a-ovchinnikov: "Multiple proxy URLs are represented when applicable." -- this one sounds fishy and also misses an important point. If there are several URLs all end up in the SBOM, but there is no indication _which_ was used (because some PMs cannot provide us with this info). If multiple URLs are mentioned then this has to be mentioned as well.

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+  references.
+
+**SPDX**:
+
+- **Package manager**: The backend annotations become package annotations on the
+  corresponding packages (see [CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)).
+- **Proxy URL**: Mapped to the package field `sourceInfo` (semicolon-separated
+  if there are several).
+
+## How Hermeto reports dependencies (library vs file)
```
a-ovchinnikov: Why is this an individual section? WDYT about splitting it in CyDX and SPDX parts and moving them to appropriate places around L30?

docs/sbom.md
a-ovchinnikov: Hmm, for some reason I thought there was a document predating SPDX support introduction and referenced that imaginary document without checking, sorry for the confusion! (I am aware that I sound like a chat-bot now and that does not make me comfortable.) _Edit_: the doc is [not a dream](https://github.com/konflux-ci/architecture/blob/main/ADR/0044-spdx-support.md) after all! It was referenced by the original contributor of the feature, but it was not the part of Hermeto design process. In this case it is fine to keep this information here

a-ovchinnikov (commented):
Nice job! The document is getting very close to an approvable shape, it is mostly LGTM now with several optional suggestions.

docs/sbom.md
ST3V1K: For CyDX, I agree that a sentence is much better, but I wanted to keep the same style for both. And I don't think it is a good idea for SPDX with all those keys and values to put it in one sentence. 
Edit: 
https://github.com/hermetoproject/hermeto/pull/1325#discussion_r2941590190
> `## How Hermeto reports dependencies (library vs file)`
> Why is this an individual section? WDYT about splitting it in CyDX and SPDX parts and moving them to appropriate places around L30?

With this, I believe it makes even more sense to keep it a list and only add the lib/file section to it

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done (https://github.com/hermetoproject/hermeto/pull/1325#discussion_r2940688191)

docs/sbom.md
```diff
@@ -0,0 +1,195 @@
...
+- There is no `dependencies` array and no parent/child hierarchy.
+- All components are peers in one list.
+- Hermeto uses only two component types: `library` (default, used for components
+  produced by all package managers except generic) and `file` (used for generic
+  artifacts such as archives when the generic package manager fetches a
+  zip/tarball).
+
+### SPDX
+
+- The output has a `packages` array and a `relationships` array.
```
eskultety, line 42: This might deserve a brief note that we don't support SPDX natively (hence "a subset") but instead we synthesize it from CycloneDX. I agree that's an internal impl. detail that should be transparent to users, however, then the "CycloneDX to SPX mapping" mention in the next section feels a bit as a surprise.

docs/sbom.md
```diff
@@ -0,0 +1,195 @@
...
+they are represented as package annotations (see
+[CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)). Property names use
+`hermeto:` as a prefix, except for the two [standard CycloneDX npm properties][]
+which use the `cdx:` prefix defined by the CycloneDX specification.
+
+| Property name | Meaning | When it appears |
+| ------------- | ------- | --------------- |
+| `hermeto:found_by` | Tool that added the dependency | Always |
+| `hermeto:missing_hash:in_file` | Path to a lockfile with a missing checksum | When there is a lockfile with some checksums missing |
+| `hermeto:bundler:package:binary` | Bundler gem is a platform-specific binary | When the gem is a platform-specific binary |
```
eskultety, line 72: nitpick: I don't think we need to enumerate all of ^these, we could just use a placeholder for the backend, since it's the same for all backends (where applicable). Short snippet examples are IMO worth more than descriptions and enumerating all possible combinations :).

docs/sbom.md
```diff
@@ -0,0 +1,195 @@
...
+  `type`:&nbsp;`"distribution"`, `comment`:&nbsp;`"proxy URL"`, and the `url`
+  set to the actual download location. Multiple URLs are represented as multiple
+  such external references.
+
+**SPDX**:
+
+- **Package manager**: The backend annotations become package annotations on the
+  corresponding packages (see [CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)).
+- **Actual source URL**: Mapped to the package field `sourceInfo`
+  (semicolon-separated if there are several).
```
eskultety: This section reads a bit weird after the previous one. I think the reason for that is that it basically translates the human readable text into the actual technical details which feels to me like a good candidate for a merged section.

docs/sbom.md
```diff
@@ -0,0 +1,195 @@
...
+  (BOM revision). SPDX:&nbsp;`spdxVersion` (e.g. `"SPDX-2.3"`).
+
+### Reasons for lack of versions in dependencies
+
+Some dependencies in the SBOM may have no version; when and why depends on the
+ecosystem:
+
+- **Go (gomod)**: The standard library has no version. Modules referenced via
+  `replace` directives pointing to local paths may also lack one; other modules
+  get one when resolved.
```
eskultety: I don't think this statement about Go is correct - the standard library's version is directly dependent on the Go installed runtime (or language version if you will). The reason why we don't version the stdlib modules in the SBOM (although we could) is because the Go runtime version we use will typically be different than what the hermetic build would use and so the resulting SBOM might be incorrectly interpreted by vulnerability analyzers without some cumbersome SBOM post-process filtering after a build in layered solutions, that's the actual reason for not recording stdlib module versions in the SBOM.

Please summarize some of ^this in your reasoning in the doc to make it factually correct.

docs/sbom.md
```diff
@@ -0,0 +1,195 @@
...
+CycloneDX one), the following applies:
+
+- Each SPDX package becomes one or more CycloneDX components. Because CycloneDX
+  allows only one purl per component, an SPDX package with multiple purls in
+  `externalRefs` becomes multiple components.
+- `versionInfo` is passed through as the component's `version`.
+- Annotations whose `annotator` ends with `:jsonencoded` are parsed as
+  properties; others are stored as top-level annotations.
+- `sourceInfo` is converted back to ExternalReferences with `type=distribution`
+  and `comment=proxy URL`.
```
eskultety, line 190: These 2 sections (although referenced) still come a bit by surprise IMO. I think after reading these there's this elephant in the room - why does Hermeto convert to SPDX from CycloneDX? Therefore, I think we should mention that we translate one to the other at the beginning where you mention that we only support a "subset". ^This is the reason for that - we don't have a native support for SPDX, yet. To be hones, none of this should really appear in user docs and the fact that we use conversion of one to the other should be completely transparent to the user, but given the situation it looks like it's probably better to mention it than not describing any of these mappings (or maybe I'm wrong and we really shouldn't mention this, @a-ovchinnikov what's your opinion on mentioning CDX->SPDX conversions?).

docs/sbom.md
```diff
@@ -0,0 +1,195 @@
...
+**Limitation**: An SPDX package that has multiple purls in `externalRefs` becomes
+multiple CycloneDX components when converted to CycloneDX, because CycloneDX does
+not support multiple purls on one component.
+
+## Merging SBOMs
+
+When SBOMs are merged, merging is performed in CycloneDX form. If an SPDX SBOM
+is passed to `merge-sboms`, it is converted to CycloneDX first, then merged.
+The merged result can be output as either CycloneDX or SPDX via
+`--sbom-output-type`.
```
eskultety: We should mention that we only support merging our own SBOMs, that is a very important detail for the user to know.

docs/sbom.md
a-ovchinnikov: Thank you. I have my reservations about lists in text, but won't push this particular change.

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Sorry for that. I have summarized some of it.

docs/sbom.md
ST3V1K: Merged them

docs/sbom.md
ST3V1K: I don't think they are that similar, at least on the property naming level. The only similar IMO are bundler and pip.

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: I changed the intro to explain upfront that we don't have a native SPDX writer and always build CycloneDX internally, then map to SPDX. I also added the same context at the top of the SPDX scheme section and the mapping section.

docs/sbom.md
```diff
@@ -0,0 +1,200 @@
...
+dependencies.
+
+### Versioning
+
+- **Version**: Each dependency can have a version, taken from the package
+  manager's data (lockfile, metadata, or similar). Some dependencies have no
+  version; Hermeto omits the field in that case. For further details see
+  [Reasons for lack of versions in dependencies](#reasons-for-lack-of-versions-in-dependencies).
+- **Purl**: The main identifier for each dependency is its purl (package URL).
+  The purl can still include version-like information (e.g. a Git commit) even
```
eskultety: s/purl/PURL (multiple occurrences, not applicable to code references) per the official acronym https://github.com/package-url/purl-spec#package-url-purl-specification .

docs/sbom.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Reasons for lack of versions in dependencies
+
+Some dependencies in the SBOM may have no version; when and why depends on the
+ecosystem:
+
+- **Go (gomod)**: The standard library version is tied to the Go runtime, but
+  Hermeto omits it because the Go version used during pre-fetching may differ
+  from the one used in the actual hermetic build; recording it could lead
+  vulnerability analyzers to flag incorrect versions. Modules referenced via
```
eskultety: nitpick:
`s/to flag/ to flag (or NOT to flag)`
`s/incorrect versions./incorrect versions, to apply (or NOT to apply) a particular provenance policy.`

docs/sbom.md
eskultety: > I don't think they are that similar

I specifically tried to only point out the `<backend>:package:binary` property, but fair enough, let's tackle the problem when it becomes one :) .

docs/sbom.md
```diff
@@ -0,0 +1,200 @@
...
+## CycloneDX to SPDX mapping
+
+Hermeto forms the SBOM as CycloneDX first; this section describes the
+translation used when SPDX output is requested (e.g. via
+`--sbom-output-type spdx`).
+
+| CycloneDX | SPDX |
+| --------- | ---- |
+| `Component` | `SPDXPackage` |
+| Component identity | `SPDXID`: `SPDXRef-Package-{name}-{version}-{hash}` (or `SPDXRef-Package-{name}-{hash}` when version is `None`). The hash is the SHA-256 hex digest of the JSON-serialized object `{"name", "version", "purl"}` with keys sorted. The idstring is sanitized per SPDX rules (only letters, digits, `.`, `-`). `versionInfo` is set to the component's version or omitted when `None`. |
```
eskultety: `s/when version is `None`//`

This whole table doesn't read very nicely because it dives too deep into the conversion rules, so were I a new collaborator to Hermeto, this would be complete gibberish to me without reading the whole code logic. Therefore I'm wondering how much of it we want documented and to what detail. I do not have a good idea at the moment, so I'm willing to just accept this as is and not block the PR further with the hope of improving it over time.

eskultety (commented):
nitpicks mostly, ACK.

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: Done

docs/sbom.md
ST3V1K: > s/when version is None//

Done 

> This whole table doesn't read very nicely because it dives too deep into the conversion rules, so were I a new collaborator to Hermeto, this would be complete gibberish to me without reading the whole code logic. Therefore I'm wondering how much of it we want documented and to what detail. I do not have a good idea at the moment, so I'm willing to just accept this as is and not block the PR further with the hope of improving it over time.

My only idea is to drop a bunch of the details from the table and have only something like this:
| CycloneDX | SPDX |
| --- | --- |
| Component | SPDXPackage |
| Component identity | Unique package ID derived from name, version, and purl |
| purl | External reference |
| ... | |

eskultety: Let's ask @a-ovchinnikov, he's our SPDX guy. Any guidance on ^this thread?

In any case, I'm going to approve on my side because, ^this could IMO be handled in a follow-up easily, HAVING documentation is already a big improvement, so I'd personally be okay with that.

eskultety (approved):

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+Hermeto produces a Software Bill of Materials (SBOM) describing the dependencies
+it fetches. This document explains how Hermeto structures the SBOM, what custom
+fields it uses, how it maps between supported formats, and how conversion works.
+
+Hermeto supports the following SBOM formats:
+
+- **CycloneDX** [v1.6][] (default)
+- **SPDX** [v2.3][] &mdash; Hermeto does not use a separate SPDX-native writer.
+  It always builds a CycloneDX-shaped SBOM internally and, when SPDX is
+  requested, maps that representation to SPDX. Only a *subset* of the SPDX
```
a-ovchinnikov: Non-blocking: ~a subset is an important bit of information, I recommend moving it as close to *SPDX* as possible. The reason for that is that an average reader skimming through the document will see **SPDX** and some text that follows, chances that they will not reach the second sentence are high. I think
either "**SPDX** (subset)" or even "**SPDX (subset)**" would convey the message better.~ _Edit_: after thinking more about it I came to a conclusion that a) being a subset is not important (since it is a subset, not a superset) and b) CyDX is also a subset (as clearly stated in *Scheme* section). Now when we know that both supported formats are subsets the question arises -- why do we want to tell our users that we support only subsets? The reason for stressing the subset nature back in the implementation days was an idea that Hermeto might be used to merge SBOMs produced by it and by some other tool in which case this partial support could bite. However when Hermeto is used to only _produce_ SBOMs or to produce and merge its own SBOMs then it is not a problem and not something a user should be concerned about at all. Thus I would suggest rephrasing the sentence and moving it out of the second bullet point in the list, i.e. to get something like:
```
Hermeto generates SBOMs which utilize only subsets of both SPDX and CyDX, the chosen subsets map into each other and Hermeto can perform this conversion, the mapping is documented in [CycloneDX to SPDX mapping]```

File: docs/sbom.md [hunk 46]
```diff
@@ -0,0 +1,201 @@
...
+## Scheme
+
+Hermeto produces a flat list of components (CycloneDX) or packages (SPDX). There
+is no nested tree of dependencies in either format.
+
+### CycloneDX
+
+- The output has a single top-level `components` array.
+- There is no `dependencies` array and no parent/child hierarchy.
+- All components are peers in one list.
```
a-ovchinnikov: I am sorry, I am still having trouble with this list, at least with the first three bullets. It feels like they convey the same point: the first states that there is only one top-level array, second says that there are no other arrays, third states that all components are placed into one array. At the same time the term "components" is just introduced as a self-obvious concept. I think that having something along the lines of "All packages processed by Hermeto are represented by CyDX components. In a SBOM produced by Hermeto Components form a single flat top-level "components" array, no hierarchy is preserved withing it. Of all possible component types available in CyDX spec Hermeto uses only ..."

This would make the document easier on new readers. A list looks tidy at the first glance, but only when one already knows what this is all about. I do not think it is clear for newcomers, and making it clear for them is the point of writing this document, everyone else could read the code.

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+- Hermeto uses only two component types: `library` (default, used for components
+  produced by all package managers except generic) and `file` (used for generic
+  artifacts such as archives when the generic package manager fetches a
+  zip/tarball).
+
+### SPDX
+
+SPDX output comes from mapping the internal CycloneDX representation to SPDX
+(see [CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)); there is no
+separate SPDX-native writer. Specifically:
```
a-ovchinnikov: What you are saying is that _internal representation_ of SBOM happens to be CyDX, and that's an implementation detail which might or might not change. The part about SPDX-native writer is not clear: how do we save SPDX SBOMs on disk if we don't have such a writer? I think what you meant to say is "At the time of writing SBOMs are represented internally as CyDX, SPDX is produced with a converter (see [mapping]). SPDX SBOMs provided by Hermeto have the following properties:"

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+
+- The output has a `packages` array and a `relationships` array.
+- `packages`: the first entry is a synthetic root with
+  `SPDXID`:&nbsp;`SPDXRef-DocumentRoot-File-`
+  (empty `name` and `versionInfo`); the rest are actual packages. The root is
```
a-ovchinnikov: Here "packages" mean "packages entities as defined by SPDX 2.3 spec", outside of this sentence "packages" might as well mean "archives of specific ecosystem-defined structure downloaded from certain location on the Internet". I would suggest replacing "packages" with "SPDX packages entries for actual packages" (or with something that is less of a mouthful).

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+  required by the SPDX specification. Every SPDX document must have a document
+  root element that the rest of its contents are related to.
+- `relationships`: every real package is linked directly to the root via
+  `CONTAINS` and `DESCRIBES` relationships.
+- SPDX does not have an equivalent "library" vs "file" component type; all
```
a-ovchinnikov: `does not have an equivalent "library" vs "file" component type` like what? Remember, a user might start reading this document from this line having zero experience with terminology. I think either a link to previous section, or soft link "Unlike CyDX SPDX does not" is necessary.

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+  packages are listed without that distinction. As a result, a `file` component
+  in CycloneDX that is converted to SPDX and back will become a `library`
+  component — this is a known limitation of the round-trip conversion.
```
a-ovchinnikov: Could you please refresh my memory and point to a location which states this limitation? I'd also recommend dropping "of the round-trip conversion" (for the sole reason that realistically none of the users will ever engage in it, it is a tool for internal diagnostics).

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Dependency properties
+
+Hermeto attaches properties from a fixed set to all dependencies. These
+properties carry metadata about dependencies they are attached to. These
+properties are stored in the CycloneDX `components[].properties` array; in SPDX
+they are represented as package annotations (see
+[CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)). Property names use
+`hermeto:` as a prefix, except for the two [standard CycloneDX npm properties][]
+which use the `cdx:` prefix defined by the CycloneDX specification.
```
a-ovchinnikov: From ergonomics perspective it is better to have some linkage between this paragraph and the table. "The table below lists all custom properties:" would do.

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+| Property name | Meaning | When it appears |
+| ------------- | ------- | --------------- |
+| `hermeto:found_by` | Tool that added the dependency | Always |
+| `hermeto:missing_hash:in_file` | Path to a lockfile with a missing checksum | When there is a lockfile with some checksums missing |
+| `hermeto:bundler:package:binary` | Bundler gem is a platform-specific binary | When the gem is a platform-specific binary |
+| `cdx:npm:package:bundled` | Npm package is bundled | When the package is bundled within another |
+| `cdx:npm:package:development` | Npm package is a dev dependency | When the package is a development dependency |
+| `hermeto:pip:package:binary` | Pip package is a binary wheel | When the package is a binary wheel |
+| `hermeto:pip:package:build-dependency` | Pip package is a build-time dependency | When the package is required at build time |
+| `hermeto:rpm_modularity_label` | RPM modularity label | When applicable for RPM dependencies |
```
a-ovchinnikov: I am sorry, but what does "When applicable for RPM dependencies" mean?

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+| `hermeto:rpm_summary` | RPM summary | When the RPM summary is included in the SBOM |
+
+### Backend and source information
+
+Unlike the dependency properties above, which are stored as flat key/value pairs
+in `components[].properties` (CycloneDX) or package annotations (SPDX), backend
+and download provenance use other SBOM fields because they describe where a
```
a-ovchinnikov: Is "download provenance" same as "source information"?

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+component came from, not arbitrary key/value metadata on the component.
+
+**Which package manager produced the dependency** &mdash; Consumers can tell
```
a-ovchinnikov: It took me some time to figure out that by "Which package manager produced the dependency" one must understand "Backend". The label on the box says "backend", a box must contain "backend". The section promises backends, it must be describing backends. I recommend replacing **the screaming bold line that grabs all attention first** with just **backends**

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+
+- **CycloneDX**: An external reference with `type`:&nbsp;`"distribution"`,
+  `comment`:&nbsp;`"proxy URL"`, and `url` set to the download location. Multiple
+  URLs are multiple such references.
+- **SPDX**: The same information is mapped to the package field `sourceInfo`
+  (semicolon-separated when there are several).
+
+## Main package and versioning
+
+### No designated "main" package in the SBOM
```
a-ovchinnikov: Again, the box says "Main package and versioning", but there is no "Main package" in the box. Something has to be aligned.

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+
+The SBOM does **not** mark which dependency is the main application. All are
+listed as peers. The SBOM alone does not distinguish "the app" from its
+dependencies.
+
+### Versioning
+
```
a-ovchinnikov, line 119: Please avoid lists without any introductory statements, this makes a text look very untidy and uncared for. At the very least "Hermeto provides several versioning mechanisms in SBOMs:" would provide some introduction.

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+  [Reasons for lack of versions in dependencies](#reasons-for-lack-of-versions-in-dependencies).
+- **PURL**: The main identifier for each dependency is its PURL (package URL).
+  The PURL can still include version-like information (e.g. a Git commit) even
+  when the version field is missing.
+- **SBOM format version**: The SBOM file itself carries a format version
+  &mdash; the schema version, not the application's version.
+  CycloneDX:&nbsp;`specVersion` (e.g. `"1.6"`) and a top-level integer `version`
+  (BOM revision). SPDX:&nbsp;`spdxVersion` (e.g. `"SPDX-2.3"`).
+
+### Reasons for lack of versions in dependencies
```
a-ovchinnikov: I would move it into a subsection of "Versioning" (i.e. s/###/####/).

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+
+Some dependencies in the SBOM may have no version; when and why depends on the
+ecosystem:
+
+- **Go (gomod)**: The standard library version is tied to the Go runtime, but
+  Hermeto omits it because the Go version used during pre-fetching may differ
+  from the one used in the actual hermetic build; recording it could lead
+  vulnerability analyzers to flag (or NOT to flag) incorrect versions, to apply
```
a-ovchinnikov: Optional: s/to flag (or NOT to flag) incorrect versions/to incorrectly flag versions/, s/to apply (or NOT to apply)/to incorrectly apply/

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+  (or NOT to apply) a particular provenance policy. Modules referenced via
+  `replace` directives pointing to local paths may also lack a version; other
+  modules get one when resolved.
+- **Yarn**: Workspace and linked packages often have no version; Hermeto omits
+  placeholders.
+- **Cargo**: Missing if a local path dependency omits it, or if a workspace
+  member inherits from `[workspace.package]` and that field is unset.
+- **npm**: Taken from the lockfile; rarely missing.
```
a-ovchinnikov: Optional: s/Taken from the lockfile; rarely missing/Version is taken from the lock file, however in rare cases it could be missing/

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+
+- **Go (gomod)**: The standard library version is tied to the Go runtime, but
+  Hermeto omits it because the Go version used during pre-fetching may differ
+  from the one used in the actual hermetic build; recording it could lead
+  vulnerability analyzers to flag (or NOT to flag) incorrect versions, to apply
+  (or NOT to apply) a particular provenance policy. Modules referenced via
+  `replace` directives pointing to local paths may also lack a version; other
+  modules get one when resolved.
+- **Yarn**: Workspace and linked packages often have no version; Hermeto omits
+  placeholders.
```
a-ovchinnikov: Optional: I suggest omitting "Hermeto omits placeholders", the phrase poses more questions than it answers.

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+translation used when SPDX output is requested (e.g. via
+`--sbom-output-type spdx`).
+
+| CycloneDX | SPDX |
+| --------- | ---- |
+| `Component` | `SPDXPackage` |
+| Component identity | `SPDXID`: `SPDXRef-Package-{name}-{version}-{hash}` (or `SPDXRef-Package-{name}-{hash}`). The hash is the SHA-256 hex digest of the JSON-serialized object `{"name", "version", "purl"}` with keys sorted. The idstring is sanitized per SPDX rules (only letters, digits, `.`, `-`). `versionInfo` is set to the component's version or omitted when `None`. |
+| `purl` | `externalRefs`: one entry with `referenceCategory=PACKAGE-MANAGER`, `referenceType=purl`, `referenceLocator=<purl>` |
+| `properties` | Package `annotations`: each property encoded as JSON `{"name":"...","value":"..."}` in `comment`; `annotator` is `Tool: hermeto:jsonencoded`. |
+| Top-level `annotations` (by bom-ref) | Same package annotations: `comment` = annotation text |
```
a-ovchinnikov: Is it "same _package annotations_" or "_same package_ annotations"? Maybe replace this with "Per-package comments containing same values as CyDX top-level annotations"? Or is it "same as above"?

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+  and `comment=proxy URL`.
+
+**Limitation**: An SPDX package that has multiple PURLs in `externalRefs` becomes
+multiple CycloneDX components when converted to CycloneDX, because CycloneDX does
+not support multiple PURLs on one component.
+
+## Merging SBOMs
+
+Merging is only meant for **SBOMs produced by Hermeto** (for example
+from `fetch-deps` or an earlier successful merge). Input files must match
```
a-ovchinnikov: Optional addition (for the sake of those who skip to the part of the document which interests them the most): "due to the fact that Hermeto supports only subsets of SBOM formats".

docs/sbom.md
a-ovchinnikov: Sorry for the late response, I would have kept the conversion section, but ensured that it is clearly marked as some form of advanced Hermeto knowledge not necessary for a day-to-day operation, maybe an appendix for the curious? I think both CyDX->SPDX and SPDX->CyDX should become subsections of this section.

a-ovchinnikov (commented):
Sorry for a delayed review, this PR slipped from my attention for a while. Please see my comments inline.

docs/sbom.md
a-ovchinnikov: Just noticed that I have already been pinged in this section, sorry for the delay! I think we can keep these details, but mark them as an extra knowledge, not something that a user actually needs to know to run Hermeto, but something they might want to know if they are doing something very special with it. Maybe move this entire section into an appendix?

README.md
```diff
@@ -94,7 +94,8 @@ hermeto fetch-deps \
 The `fetch-deps` command fetches your project's dependencies and stores them on
 your disk. Hermeto also produces a detailed SBOM containing information about
 all the project's components and packages. You can find the SBOM in the output
-directory.
+directory. For SBOM structure, custom fields, and CycloneDX/SPDX mapping, see
```
derasdf: This link works in https://hermetoproject.github.io/hermeto/ but is broken on GitHub. README.md is at the repo root in GitHub, the file is docs/sbom.md. Leave it like in other .md files unlinked.

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+separate SPDX-native writer. Specifically:
+
+- The output has a `packages` array and a `relationships` array.
+- `packages`: the first entry is a synthetic root with
+  `SPDXID`:&nbsp;`SPDXRef-DocumentRoot-File-`
+  (empty `name` and `versionInfo`); the rest are actual packages. The root is
+  required by the SPDX specification. Every SPDX document must have a document
+  root element that the rest of its contents are related to.
+- `relationships`: every real package is linked directly to the root via
+  `CONTAINS` and `DESCRIBES` relationships.
```
derasdf: DESCRIBES links SPDXRef-DOCUMENT to the root. Real packages get CONTAINS from the root only, not both. I think this should read something like:

```
- `relationships`: the document describes the root
  (`SPDXRef-DOCUMENT` -> `SPDXRef-DocumentRoot-File-` via `DESCRIBES`);
  every real package is linked to the root via `CONTAINS`.
```

See [`create_root_relationship`](https://github.com/hermetoproject/hermeto/blob/455a7e0a0e6e83f612aa32b2c37a8fd70986b423/hermeto/core/models/sbom.py#L266-L272) (single DESCRIBES) and [`link_to_root`](https://github.com/hermetoproject/hermeto/blob/455a7e0a0e6e83f612aa32b2c37a8fd70986b423/hermeto/core/models/sbom.py#L274-L281) (CONTAINS only).

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+
+Hermeto produces a flat list of components (CycloneDX) or packages (SPDX). There
+is no nested tree of dependencies in either format.
+
+### CycloneDX
+
+- The output has a single top-level `components` array.
+- There is no `dependencies` array and no parent/child hierarchy.
+- All components are peers in one list.
+- Hermeto uses only two component types: `library` (default, used for components
```
derasdf: "used for generic artifacts such as archives" is slightly broad. Maven artifacts fetched via the generic PM use [`library`](https://github.com/hermetoproject/hermeto/blob/455a7e0a0e6e83f612aa32b2c37a8fd70986b423/hermeto/core/package_managers/generic/models.py#L212-L217). Only URL-based fetches (zip/tarball) get [`file`](https://github.com/hermetoproject/hermeto/blob/455a7e0a0e6e83f612aa32b2c37a8fd70986b423/hermeto/core/package_managers/generic/models.py#L120-L124). I think "used for URL-based generic artifacts such as archives" would be more precise.

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+not support multiple PURLs on one component.
+
+## Merging SBOMs
+
+Merging is only meant for **SBOMs produced by Hermeto** (for example
+from `fetch-deps` or an earlier successful merge). Input files must match
+Hermeto's CycloneDX or SPDX model; arbitrary SBOMs from other tools are not
+supported and will be rejected as invalid.
+
+When SBOMs are merged, merging is performed in CycloneDX form. If an SPDX SBOM
```
derasdf: "merging is performed in CycloneDX form" is only true for the default output type. With `--sbom-output-type spdx` the [merge happens in SPDX form](https://github.com/hermetoproject/hermeto/blob/455a7e0a0e6e83f612aa32b2c37a8fd70986b423/hermeto/interface/cli.py#L517-L524). Maybe:

```
The merge format matches the requested output type. With cyclonedx
(default), inputs are converted to CycloneDX. With spdx, inputs are
converted to SPDX.
```

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+| (no root in CycloneDX) | A synthetic root package `SPDXRef-DocumentRoot-File-` is created with `name=""` and `versionInfo=""`. The document describes the root; the root contains every package. |
+| `metadata.tools` | `creationInfo.creators` (`Tool:` / `Organization:`) |
+
+## SPDX to CycloneDX
+
+When SPDX is converted to CycloneDX (e.g. when merging an SPDX SBOM with a
+CycloneDX one), the following applies:
+
+- Each SPDX package becomes one or more CycloneDX components. Because CycloneDX
+  allows only one PURL per component, an SPDX package with multiple PURLs in
```
derasdf, line 184: nitpick: "others are stored as top-level annotations" isn't quite right. The [else branch](https://github.com/hermetoproject/hermeto/blob/455a7e0a0e6e83f612aa32b2c37a8fd70986b423/hermeto/core/models/sbom.py#L730-L736) creates component Properties too. Only matters for non-hermeto SBOMs though, so low risk.

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+  `externalRefs` becomes multiple components.
+- `versionInfo` is passed through as the component's `version`.
+- Annotations whose `annotator` ends with `:jsonencoded` are parsed as
+  properties; others are stored as top-level annotations.
+- `sourceInfo` is converted back to ExternalReferences with `type=distribution`
```
derasdf, line 189: nitpick: this Limitation paragraph says the same thing as the bullet at 163-165. One can go.

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+
+Hermeto forms the SBOM as CycloneDX first; this section describes the
+translation used when SPDX output is requested (e.g. via
+`--sbom-output-type spdx`).
+
+| CycloneDX | SPDX |
+| --------- | ---- |
+| `Component` | `SPDXPackage` |
+| Component identity | `SPDXID`: `SPDXRef-Package-{name}-{version}-{hash}` (or `SPDXRef-Package-{name}-{hash}`). The hash is the SHA-256 hex digest of the JSON-serialized object `{"name", "version", "purl"}` with keys sorted. The idstring is sanitized per SPDX rules (only letters, digits, `.`, `-`). `versionInfo` is set to the component's version or omitted when `None`. |
+| `purl` | `externalRefs`: one entry with `referenceCategory=PACKAGE-MANAGER`, `referenceType=purl`, `referenceLocator=<purl>` |
```
derasdf, line 171: nitpick: `{"name", "version", "purl"}` reads like a Python set. Maybe "a JSON object with keys `name`, `version`, and `purl`".

docs/sbom.md
```diff
@@ -0,0 +1,201 @@
...
+
+- **CycloneDX** [v1.6][] (default)
+- **SPDX** [v2.3][] &mdash; Hermeto does not use a separate SPDX-native writer.
+  It always builds a CycloneDX-shaped SBOM internally and, when SPDX is
+  requested, maps that representation to SPDX. Only a *subset* of the SPDX
+  specification is covered; the output matches the fields and structure
+  described in this document. The mapping is documented in
+  [CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping).
+
+The output format could be selected with `--sbom-output-type cyclonedx` or
```
derasdf: nitpick: `s/could be selected/can be selected/` "could" is conditional, "can" states a capability that exists right now.

derasdf (commented):
Just few more nitpick comments, mostly OK tho.

docs/sbom.md
ST3V1K: https://github.com/hermetoproject/hermeto/pull/1325#discussion_r2927728141
> This sounds like a bug to me: judging from this description if A is a CyDX with file then A.toSpdx().toCyDx() =/= A. Is it so?

There is not really a point that says this is a "limitation", but this is IMO the best way to say it in docs

docs/sbom.md
ST3V1K: I meant whenever the RPM had the modularity label.
I have rephrased both RPM-related property details.

docs/sbom.md
ST3V1K: Yes, but I have now aligned the text with the header

docs/sbom.md
ST3V1K: I have split "Main package and versioning" section into two.

docs/sbom.md
ST3V1K: Based on [this](https://github.com/hermetoproject/hermeto/pull/1325#discussion_r3131081554) it now is a subsection

docs/sbom.md
ST3V1K: Since Hermeto works only with its own SBOMs, this isn't necessary

docs/sbom.md
ST3V1K: IMO, removing it from either place will create either confusion, or users could miss the info

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+At the time of writing, SBOMs are represented internally as CycloneDX. SPDX is
+produced with a converter (see [CycloneDX to SPDX mapping](#cyclonedx-to-spdx-mapping)).
+SPDX SBOMs provided by Hermeto have the following properties:
+
+- The output has a `packages` array and a `relationships` array.
+- `packages`: the first entry is a synthetic root with
+  `SPDXID`:&nbsp;`SPDXRef-DocumentRoot-File-`
+  (empty `name` and `versionInfo`); the rest are SPDX package entries for actual
+  packages. The root is required by the SPDX specification. Every SPDX document
+  must have a document root element that the rest of its contents are related to.
```
a-ovchinnikov: Non-blocking nitpick: "The root is rqeuired..." and "Every SPDX document must have..." are essentially the same sentence, I would have just dropped the second one (or merged them somehow).

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+- `relationships`: the document describes the root
+  (`SPDXRef-DOCUMENT` &rarr; `SPDXRef-DocumentRoot-File-` via `DESCRIBES`);
+  every real package is linked to the root via `CONTAINS`.
+- Unlike CycloneDX SPDX does not have an equivalent "library" vs "file" component
```
a-ovchinnikov: Nitpick: s/have an equivalent "library" vs "file" component type/have direct equivalents to "library" or "file" component types/ ("vs" is very surprising here, you  are not trying to contrast them, are you?)

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+| `hermeto:rpm_summary` | RPM summary | [When you set `include_summary_in_sbom` to `true`](rpm.md#basic-options) |
+
+### Backend and source information
+
+Unlike the dependency properties above, which are stored as flat key/value pairs
+in `components[].properties` (CycloneDX) or package annotations (SPDX), backend
+and source information use other SBOM fields because they describe where a
+component came from, not arbitrary key/value metadata on the component.
+
+**Backends** &mdash; Consumers can tell which backend (e.g. gomod, pip, npm)
```
a-ovchinnikov: Non-blocking nitpick: s/Consumers can tell/lets consumers know/

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+produced each dependency. Experimental backends (name starting with `x-`) are
+tagged as experimental.
```
a-ovchinnikov: Nitpick (non blocking): for better cohesion consider adding a linking sentense, something along the lines of "Supported SBOM types represent them like follows:"

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+several URLs are known, all are recorded; the SBOM does not say which URL was
+actually used (some backends cannot provide that).
+
+- **CycloneDX**: An external reference with `type`:&nbsp;`"distribution"`,
+  `comment`:&nbsp;`"proxy URL"`, and `url` set to the download location. Multiple
+  URLs are multiple such references.
+- **SPDX**: The same information is mapped to the package field `sourceInfo`
+  (semicolon-separated when there are several).
+
+## No designated "main" package in the SBOM
```
a-ovchinnikov: Nitpick, non-blocking: please consider renaming this section to "Main package representation", it sticks out in the table of contents otherwise.

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+dependencies.
+
+## Versioning
+
+Hermeto provides several versioning mechanisms in SBOMs:
+
+- **Version**: Each dependency can have a version, taken from the package
+  manager's data (lockfile, metadata, or similar). Some dependencies have no
+  version; Hermeto omits the field in that case. For further details see
+  [Reasons for lack of versions in dependencies](#reasons-for-lack-of-versions-in-dependencies).
```
a-ovchinnikov: Nitpick, non blocking: s/Reasons for lack of versions in dependencies/Reasons for lack of versions in some dependencies/, it might make a casual reader think that all dependencies lack versions otherwise.

docs/sbom.md
```diff
@@ -0,0 +1,199 @@
...
+- **Yarn**: Workspace and linked packages often have no version.
+- **Cargo**: Missing if a local path dependency omits it, or if a workspace
+  member inherits from `[workspace.package]` and that field is unset.
+- **npm**: Version is taken from the lock file, however in rare cases it could
+  be missing.
+- **pip**: PyPI packages always have a version; VCS or local path sources may
+  not. May also be absent if the main project uses dynamic versioning
+  (e.g. setuptools-scm).
+
+## CycloneDX to SPDX mapping
```
a-ovchinnikov, line 161: WDYT about moving this to appendix and clearly marking this as an advanced topic?

a-ovchinnikov (commented):
LGTM, several non-blocking nitpick, one semi-blocking question. Will approve once there is an agreement on it.

docs/sbom.md
ST3V1K: I'm not against it, but I would suggest moving it after the section about merging without any appendix/header. I don't think that appendix would fit the document.

a-ovchinnikov: Since I think that an appendix would make perfect sense for the document (this section is very optional and needed by very advanced users only, I think that's what appendices are for) let's summon @eskultety to this thread to break the tie: what's your preference, to have appendix or to keep the section as it is?

eskultety: > to this thread to break the tie: what's your preference, to have appendix or to keep the section as it is?

TL;DR - move the section.

At some point we started using appendices for designs where it made sense at the time, but we don't use it for user-facing docs, so the question is - do we want to set a precedent? What's the motivation or intended goal of that? I, as a user, would find it a bit confusing to be honest, it's either something I'm interested in or I continue scrolling, if I happen to be interested in such detail or depth, I will stop and read and I think most tech people would do the same, so I'd simply move the section to the end of file so that the table can potentially grow or provide more detailed description.

ST3V1K: Change this:

_CycloneDX to SPDX mapping
SPDX to CycloneDX
Merging SBOMs_

To this:

_Merging SBOMs
CycloneDX to SPDX mapping
SPDX to CycloneDX_

a-ovchinnikov (approved):
Great job, thank you!

derasdf (approved):

Issue #1097, opened 2025-09-03
Hermeto injects .cargo/config.toml for pip and cargo and breaks cargo builds


slimreaper35:
This bug appears when a repository has:

1. **Rust** project
2. **Python** project with dependencies that have **Rust** extensions

It is necessary to inject `.cargo/config.toml` for each project with specific configuration to allow hermetic builds.

**config file directly in a Rust project**
```
[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "..."
```

**config file in the hermeto output directory (for Python projects and Rust extensions )**
```
[source.crates-io]
replace-with = "local"

[source.local]
directory = "..."
```

The latter works with setting the environment variable [CARGO_HOME](https://github.com/hermetoproject/hermeto/blob/7efebe0e1dcbfae84859cf10e9da2f76bf904d51/hermeto/core/package_managers/pip/rust.py#L153).

Having both of these files will cause an error during a cargo build:
```
error: source `local` defines source dir /cachi2/output/deps/cargo, but that source is already defined by `vendored-sources`
      note: Sources are not allowed to be defined multiple times.
```

Cargo source code error:
https://github.com/rust-lang/cargo/blame/05bf93cf8f5ab1a1f4ba8c3c015db4058779c090/src/cargo/sources/config.rs#L252-L265

kunalsdas:
Hi @slimreaper35, I'd like to work on this bug.

I traced through the code and I think I understand the root cause:

The cargo backend writes source_dir/.cargo/config.toml with replace-with = "vendored-sources" (cargo/main.py:442)
The pip rust backend writes output_dir/.cargo/config.toml with replace-with = "local" (pip/rust.py:118-131) and sets CARGO_HOME=${output_dir}/.cargo (line 160)
Both configs point to the same ${output_dir}/deps/cargo directory, but use different source names (vendored-sources vs local). When Cargo walks the config hierarchy during a build, it finds both and errors because crates-io can't have two replacements.

The simplest fix I see would be to make pip/rust.py's _config_data() use the same "vendored-sources" source name that cargo vendor generates, since they point to the same directory anyway. That way even if Cargo finds both configs, they're compatible.

However, I want to also consider whether we should handle this at the RequestOutput merge level (output.py) instead, to be more robust for future backends. Would you prefer I open a discussion with a brief proposal first, or go straight to a PR?

Could you assign this to me?

eskultety:
[...]

PR #1326, opened 2026-02-27
fix(cargo): unify pip cargo source alias to vendored-sources and add pip+cargo combined integration test

Siva-Sainath:
Fixes the issue: #1097

Use local as the source alias for the pip/rust config which caused cargo to reject builds that combined a native rust project with pip packages that had rust extensions, as the configs point to the same vendor directory location but with two different names, hence unifying the pip cargo source alias to vendored-sources ensures that the error doesn't occur as the prev_name!=name if statement in 'match registry_sources.by_path.insert' will reject the duplicate. 

Also adds a new test case in the test_multiple.py file for cargo + pip with expected test output(bom.json) and a related build-config.yaml file. Test was built on top of integration-tests/tree/pip/e2e_rust_extensions branch's forked version and the code there will be shifted to integration-tests/tree/multiple/cargo-and-pip branch once verified by the maintainers.


Siva-Sainath:
Could I please get a review for the PR @eskultety @slimreaper35

a-ovchinnikov:
Thank you for your contribution!

> Could I please get a review for the PR

May I enquire about the reason for such urgency?

To speed up the review process please describe how you reproduced the issue and tested your change.

slimreaper35:
I think this is a good candidate for an integration test with multiple package managers - https://github.com/hermetoproject/hermeto/blob/main/tests/integration/test_multiple.py

Siva-Sainath:
Hello @a-ovchinnikov I didn't want to convey urgency I am sorry it came out that way just wanted to get eyes on the PR as a first time contributor for this repo. 

I setup the repo on my laptop and then reproduced the error by dropping a rust project and rust extension python package in the same directory, ran hermeto so that i could write the two config files, then run cargo build to produce the error. 

After reproducing the issue I traced the generation of .cargo/config.toml and then realized that when CARGO_HOME is set it reads the output config in addition to the repo's own config and hence we have two source entries pointing towards the same vendored directory. 

I used the source collision check to reject the duplicate path and am changing the alias as this check works only when both of them have the same alias.

For validation I ran unit tests and then pip/cargo tests locally before committing. 

I apologize for replying late.

Siva-Sainath:
Hello @slimreaper35 I added a commit to this PR that adds the test parameter in test_multiple.py for this native cargo crate and a python package case and a ContainerFile as well. In the integration tests repo could you create an empty multiple/rust-and-pip branch in hermetoproject/integration-tests so I can open a PR for this multiple integration test? because I am not able to add my new branch to the existing repo is there any process to doing this I assume it's restricted? Can you also please approve workflows so CI can run check on the new commit in this PR

slimreaper35:
@Siva-Sainath - https://github.com/hermetoproject/integration-tests/tree/multiple/cargo-and-pip

Siva-Sainath:
@slimreaper35  thank you for opening the new branch I have raised a PR over there and an integration test has failed in this PR due to the branch name reference being wrong changed it our new branch="multiple/cargo-and-pip" from "multiple/rust-and-pip"(previous), please approve test workflows again

Siva-Sainath:
@slimreaper35 The CI failures  aren’t due to bugs in the code but due to the multiple/cargo-and-pip branch in the upstream integration-tests repo being empty as we have not merged the PR over there, after reviewing and merging the PR in the branch I believe the integration test would pass here

slimreaper35:
@Siva-Sainath, You can temporarily update the integration test repo URL for a specific test case: https://github.com/hermetoproject/hermeto/blob/main/tests/integration/utils.py#L77

Siva-Sainath:
thank you for the guidance @slimreaper35, added the temp URL hopefully the integration test should run successfully now could I please get the workflows approved

Siva-Sainath:
Hello @slimreaper35, in the multiple_rust_and_pip folder the expected output files were not added and since test suite uses a pytest into look up the expected outputs, it crashed with a FileNotFoundError for the .build-config.yaml. I have added the expected output files and the build-config file as well in the latest commit. 

Apologize for this will do a better job in future work on understanding the flow of the integration tests. Asking for workflows to be approved again.

The integration tests also failed due to a yarn error(npm_yarn_registry_lockfile3): ContentLengthError: Not enough data to satisfy content length header, is this something related to the hermeto project or a issue with the Yarn registry?

Siva-Sainath:
Hello @slimreaper35 the UBI9 base image came with python3-3.9.21-2.el9 pre installed but the RPM repo's packaged needed  python3-3.9.21-2.el9_5 specifically which is why dnf refused to installed and it caued the container to fail in building, I have added a new commit that adds --allowerasing in the Containerfile so that the df can replace or downgrade pre installed python to meet its dependency requirements.
Please approve workflows

tests/integration/test_multiple.py
```diff
@@ -26,6 +26,25 @@
...
+                    # using RPM to provide cargo and python in the image
+                    {"type": "rpm"},
+                ),
+                flags=[],
+                expected_exit_code=0,
+                expected_output="All dependencies fetched successfully",
+            ),
+            [],
+            [],
+            id="multiple_rust_and_pip",
```
slimreaper35: The id is now different from the branch name.

tests/integration/test_data/multiple_rust_and_pip/container/Containerfile
```diff
@@ -0,0 +1,25 @@
+# A corresponding rpms.lock.yaml was built for UBI9
+FROM registry.access.redhat.com/ubi9/ubi@sha256:304b50df1ea4db9706d8a30f4bbf26f582936ebc80c7e075c72ff2af99292a54
```
slimreaper35: maybe you could use a newer ubi10 image

slimreaper35 (commented):
Please squash the commits. We don't do squash merging.

tests/integration/test_data/multiple_rust_and_pip/container/Containerfile
Siva-Sainath: Moved from ubi9 to ubi10 and regenerated the expected test files after changing to ubi10

tests/integration/test_multiple.py
Siva-Sainath: Set the id to the correct branch name multiple_cargo_and_pip

slimreaper35:
@Siva-Sainath Where did you get the digest from ? [This](https://catalog.redhat.com/en/software/containers/ubi10/ubi/66f2b46b122803e4937d11ae#get-this-image) one should be the latest.

Siva-Sainath:
@slimreaper35 I pulled the image locally on my Mac with apple silicon , which resolved to the arm64 variant's digest instead of the correct digest. I have updated it to the digest to the correct amd64 digest that was provided.

tests/integration/test_multiple.py
```diff
@@ -26,6 +26,25 @@
             [],
             id="multiple_gomod_and_npm",
         ),
+        pytest.param(
+            utils.TestParameters(
```
slimreaper35, line 30: Hmm, I would set `check_deps_checksums` to False. It is set to False for all cargo integration tests.

slimreaper35 (commented):
@Siva-Sainath you need to rebase because of recent changes in the repo

tests/integration/test_multiple.py
Siva-Sainath: realized that a few mins ago didn't notice this comment, will do so and commit

Siva-Sainath: set check_deps_checksums to False and rebased as well @slimreaper35

Siva-Sainath:
@slimreaper35 test failed due to local paths being present in the build_config.project_files changed them to the ${test_case_tmp_path} placeholder and amended the existing commit

Siva-Sainath:
@slimreaper35 project-files array had the wrong order for repos, we had to keep the top level repo's at the end but it was at the first, rewrote it to the correct order and amended the commit

Siva-Sainath:
@slimreaper35 Set my fork commit which has the test data to the same SHA as the one the current test fixture is expecting, Fork branch: https://github.com/Siva-Sainath/integration-tests/tree/multiple/cargo-and-pip
Commit SHA: d28c0c334858702f99e1ec2b4bd9b4af20d4df9e, CI test should be able to find the rust crate and python pkg files now

Siva-Sainath:
@slimreaper35 the CI generated the wrong SBOM using the the wrong vcs_url the one that was set to whatever commit the forked branch was pointing to instead of the one with the latest files, hence I added logic(explicit repo_commit parameter) in utils.py and test_multiple for it to use the specific commit in the forked repo that has the right files hopefully the vcs_url points and takes the right commit this time

Siva-Sainath:
Hello @slimreaper35 I have built on top my forked branch of the e2e_rust_extensions and generated a new bom.js file using a x86 pod and then tested our test on a local pod again(x86) the test passed I hope it runs and executes in the CI as well I have used cryptography package from the existing implementation in that branch and modified it for our test cargo and pip use case, please approve test workflows

Siva-Sainath:
Thank you for approving the workflows @slimreaper35 looks like all the tests have passed, how should we proceed next, changing the repo_url back to the cargo_and_pip branch with the files present on the forked rust_e2e_extensions branch?

slimreaper35:
A lot of unrelated AI-generated content with disabled network access. Closing.

PR #1331, opened 2026-03-01
fix: use prefix instead of suffix for TemporaryDirectory in resolver

Sanchit2662:
## Summary

While reviewing the temporary directory handling in `resolver.py`, I noticed a subtle but incorrect usage of Python's `TemporaryDirectory` class. The code was passing the directory name pattern as a positional argument, which maps to the `suffix` parameter rather than `prefix`. This resulted in temp directories named like `tmpXXXXXXXX.hermeto-source-copy` instead of the intended `.hermeto-source-copy-XXXXXXXX`.

The fix is straightforward—adding the `prefix=` keyword argument—and brings this usage in line with every other `TemporaryDirectory` call in the codebase.

---

## Fix

**File:** `hermeto/core/resolver.py` (line 47)

The positional argument `f".{APP_NAME}-source-copy"` was being interpreted as a suffix. By explicitly using `prefix=`, the temp directory now gets a recognizable name that starts with `.hermeto-source-copy-` followed by the random suffix.

```python
# Before
with TemporaryDirectory(f".{APP_NAME}-source-copy", dir=".") as temp_dir:

# After  
with TemporaryDirectory(prefix=f".{APP_NAME}-source-copy-", dir=".") as temp_dir:
```

I also updated the corresponding test assertions in `tests/unit/test_resolver.py` to verify the correct prefix behavior:

```python
# Updated assertions
assert tmp_dir_name.startswith(".hermeto-source-copy-")
```

This change makes debugging easier when temp cleanup fails or when inspecting the working directory, and maintains consistency with how `TemporaryDirectory` is used elsewhere in the project (see `scm.py` and `gomod.py`).

### DIFF
--- a/hermeto/core/resolver.py
+++ b/hermeto/core/resolver.py
@@ -44,7 +44,7 @@ def resolve_packages(request: Request) -> RequestOutput:
     """
     original_source_dir = request.source_dir
 
-    with TemporaryDirectory(f".{APP_NAME}-source-copy", dir=".") as temp_dir:
+    with TemporaryDirectory(prefix=f".{APP_NAME}-source-copy-", dir=".") as temp_dir:
         source_backup = copy_directory(original_source_dir.path, Path(temp_dir).resolve())
 
         request.source_dir = RootedPath(source_backup)
--- a/tests/unit/test_resolver.py
+++ b/tests/unit/test_resolver.py
@@ -115,8 +115,7 @@ def test_source_dir_copy(
 
         # assert a temporary directory is being used
         assert tmp_dir_name != tmp_path.name
-        assert tmp_dir_name.startswith("tmp")
-        assert tmp_dir_name.endswith(".hermeto-source-copy")
+        assert tmp_dir_name.startswith(".hermeto-source-copy-")
 
         return RequestOutput.empty()
 
@@ -150,7 +149,7 @@ def test_project_files_fix_for_work_copy(
     def _resolve_packages(request: Request) -> RequestOutput:
         # assert request is based on a copy of the source dir
         assert request.source_dir.path != tmp_path
-        assert request.source_dir.path.name.endswith(".hermeto-source-copy")
+        assert request.source_dir.path.name.startswith(".hermeto-source-copy-")
 
         abspath = request.source_dir.path if with_path_replacement else request.output_dir.path
         return RequestOutput(


a-ovchinnikov:
Hi @Sanchit2662, thank you for the contribution! 

> I noticed a subtle but incorrect usage of Python's TemporaryDirectory class

Please elaborate why do you think it is incorrect?

Sanchit2662:
Hi @a-ovchinnikov , thanks for the review!

  So the thing is, tempfile.TemporaryDirectory takes suffix as its first positional argument, not prefix:

  tempfile.TemporaryDirectory(suffix=None, prefix=None, dir=None, ...)

  Right now the code passes .hermeto-source-copy positionally, which means it ends up as a suffix — so the temp dirs
  look like tmpXXXXXXXX.hermeto-source-copy instead of .hermeto-source-copy-XXXXXXXX.

  I noticed the rest of the codebase already uses prefix= explicitly:
  - scm.py → prefix="cachito-"
  - gomod.py → prefix=f"{APP_NAME}" and prefix=f"{APP_NAME}-"

  So this just brings resolver.py in line with those. It's a small thing but makes the temp dir names more recognizable
  when debugging.

a-ovchinnikov:
> the thing is, tempfile.TemporaryDirectory takes suffix as its first positional argument, not prefix

I still fail to see how this is incorrect. To quote your own words:

> I noticed a subtle but incorrect usage of Python's TemporaryDirectory class.

What you share above is neither subtle nor incorrect, it is inconsistent at best. "Subtle" means "difficult to understand or perceive", when someone talks about a piece of code being subtle they usually mean that it has behavior which can elude a reader at a cursory glance _and_ has a potential of leading to errors in runtime (consider, for instance `int* i, j` declaration in C). To claim that a piece of code is incorrect one must provide an example when this piece of code results in a behavior that drastically differs from expectations. Having a prefix and suffix swapped in a directory that no one realistically ever looks at and which exists for the duration of a program execution is **not** an incorrect behavior. I am not aware of any debugging attempt that has ever been hindered by placement of "hermeto" in a suffix and not in a prefix, I cannot imagine that `tmpXXXXXXXX.hermeto-source-copy` would cause any problem as compared to `.hermeto-source-copy-XXXXXXXX`, personally I would have recommended a person who finds the first name significantly more confusing than the first one to avoid debugging any code, however if that has significantly interfered with your debugging of any issue then please share your experience. Without that this PR is void of any substance.

> This change makes debugging easier when temp cleanup fails 

I am also very interested in this statement of yours. What would a failure to cleanup temp mean? Please structure your answer in dactylic hexameter

Sanchit2662:
@a-ovchinnikov ,
Yeah fair point, I may have overstated it by calling it "incorrect." You're right that functionally it works fine , the temp dir gets created, used, and cleaned up either way.                                                                                                                                                                                                 Honestly my main motivation was just consistency. Every other TemporaryDirectory call in the codebase uses prefix=explicitly, and this one was the odd one out passing it positionally   as suffix. I figured aligning it makes the code easier to read at a glance , you see prefix= and immediately know what the string is for, without having to remember the argument order.                                                                                                                                                                                                                                                                                                                                                                          I don't have a concrete debugging horror story to point to. If you feel this isn't worth the change, I totally understand , happy to close this.

PR #1339, opened 2026-03-02
fix(input): improve user-facing validation error messages

### LINKED ISSUE #1016: Improve input validation error messages
Imagine you do the following:
```
$ hermeto fetch-deps asdf
```
The corresponding error message:
```
2025-07-08 11:33:07,452 ERROR InvalidInput: 1 validation error for user input\npackages -> 0\n  Input tag 'asdf' found using 'type' does not match any of the expected tags: 'bundler', 'cargo', 'generic', 'gomod', 'npm', 'pip', 'yarn'
packages -> 0
  Input tag 'asdf' found using 'type' does not match any of the expected tags: 'bundler', 'cargo', 'generic', 'gomod', 'npm', 'pip', 'yarn'
```

From UX POV this is wrong on 2 levels:

1. Why is the error reported twice in the log? It's because the logger can report any error, not just ours and so we end up reporting the problem twice in this pydantic gibberish - **different issue!**
2. pydantic isn't particularly nice to deal with parsing its `ValidationError` structures and so we opted in to a simple concatenation of some of pydantic's `ValidationError` fields, but left it at that.

The problem with the latter specifically is that it may not always tell the user what's wrong in context of their input rather than in context of our internal validation, e.g. how is `packages -> 0` going to help end users at all? As for "tags" I myself had to dig into the docs and code to understand why it refers to "tags" in the first place.

User facing error messages must be concise & clear. One way of going about this with `pydantic` without ripping one's hair could be error message translation (at least for the most common errors we can already identify): https://docs.pydantic.dev/latest/errors/errors/#customize-error-messages

abhayrajjais01:
Improve user-facing input validation error messages

Fixes #1016

## What changed

Rewrites [_present_user_input_error()](cci:1://file:///c:/Users/abhay/hemerto/hermeto/hermeto/core/models/input.py:70:0-101:33) in [hermeto/core/models/input.py](cci:7://file:///c:/Users/abhay/hemerto/hermeto/hermeto/core/models/input.py:0:0-0:0)
to stop leaking pydantic internals into error messages shown to the user.

## Changes

- Add [_simplify_location()](cci:1://file:///c:/Users/abhay/hemerto/hermeto/hermeto/core/models/input.py:82:0-102:29) to strip pydantic-internal prefixes from
  error location paths (package indices, union type names). Users see
  [path](cci:1://file:///c:/Users/abhay/hemerto/hermeto/hermeto/core/models/input.py:138:4-141:39) instead of `packages -> 0 -> gomod -> path`.
- Rewrite `union_tag_invalid` errors from pydantic jargon into plain
  English ("Unknown package manager x. Valid options are: ...").
- Drop the `"N validation error for user input"` header for single-error
  cases — show the message directly.
- Suppress duplicate `log.error()` in [_bail_out_with_error](cci:1://file:///c:/Users/abhay/hemerto/hermeto/hermeto/interface/cli.py:120:0-125:52) for
  [InvalidInput](cci:2://file:///c:/Users/abhay/hemerto/hermeto/hermeto/core/errors.py:74:0-75:33) — the `print(stderr)` already displays the message.
- Add [TestSimplifyLocation](cci:2://file:///c:/Users/abhay/hemerto/hermeto/tests/unit/models/test_input.py:37:0-56:50) with parametrized unit tests for the helper.

## Testing

```sh
nox -s python-3.10 -- tests/unit/models/test_input.py tests/unit/test_cli.py
```
Used Gemini (Antigravity) to validiate and test the changes

### DIFF
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -3,7 +3,7 @@ import logging
 import re
 from collections.abc import Callable
 from pathlib import Path
-from typing import TYPE_CHECKING, Annotated, Any, Literal, TypeVar, Union
+from typing import TYPE_CHECKING, Annotated, Any, Literal, TypeVar, Union, get_args
 
 import pydantic
 from typing_extensions import Self
@@ -67,40 +67,8 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
         raise InvalidInput(_present_user_input_error(e)) from e
 
 
-def _present_user_input_error(validation_error: pydantic.ValidationError) -> str:
-    """Make a slightly nicer representation of a pydantic.ValidationError.
-
-    Compared to pydantic's default message:
-    - don't show the model name, just say "user input"
-    - don't show the underlying error type (e.g. "type=value_error.const")
-    """
-    errors = validation_error.errors()
-    n_errors = len(errors)
-
-    def show_error(error: "ErrorDict") -> str:
-        location = " -> ".join(map(str, error["loc"]))
-        if error.get("type") != "union_tag_invalid":
-            message = error["msg"]
-        else:
-            # Handle regular union tag errors (i.e. errors which stem from
-            # a missing package manager implementation). Errors in experimental
-            # package managers are handled elsewhere.
-            ctx = error.get("ctx", {})
-            raw = ctx.get("expected_tags", "")
-            expected = [t.strip(" '") for t in raw.split(",")]
-            quoted = ", ".join(f"'{t}'" for t in sorted(expected))
-            message = f"Requested backend type '{ctx.get('tag', '<unknown>')}' doesn't match expected ones: {quoted}"
-
-        if location != "__root__":
-            message = f"{location}\n  {message}"
-
-        return message
-
-    header = f"{n_errors} validation error{'' if n_errors == 1 else 's'} for user input"
-    details = "\n".join(map(show_error, errors))
-    return f"{header}\n{details}"
-
-
+# The canonical list of supported package manager types, used as the discriminator
+# for the PackageInput union. Experimental backends use the 'x-' prefix.
 PackageManagerType = Literal[
     "bundler",
     "cargo",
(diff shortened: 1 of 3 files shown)


hermeto/interface/cli.py
```diff
@@ -119,8 +119,9 @@ class SBOMFormat(str, enum.Enum):
 
 
 def _bail_out_with_error(e: BaseError) -> None:
-    """Report and error and set correct exit code."""
-    log.error("%s: %s", type(e).__name__, str(e).replace("\n", r"\n"))
+    """Report an error and set correct exit code."""
+    if not isinstance(e, InvalidInput):
```
a-ovchinnikov: Do I understand your intent correctly and you suggest scrubbing `InvalidInput`s from logs completely? What if someone sets up a logger that writes them to a file?

abhayrajjais01: thank you for your feedback, my initial goal was to avoid the double-output on stderr, but suppressing log.error() entirely for InvalidInput would break file-based logging setups
now I will revert the isinstance guard-log.error() then it will runs unconditionally again

hermeto/core/models/input.py
```diff
@@ -68,36 +68,76 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
...
+_KNOWN_BACKEND_TYPES = {
+    "bundler",
+    "cargo",
+    "generic",
+    "gomod",
+    "npm",
+    "pip",
+    "rpm",
+    "yarn",
+}
```
eskultety: Could we maybe construct this programatically rather than defining it statically? I feel like this just adds to the number of places we must not forget updating when adding a new backend.

hermeto/core/models/input.py
```diff
@@ -68,36 +68,76 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
...
+
+    Removes leading 'packages', numeric indices, and discriminated-union type names
+    so that users see the field name rather than the full internal path.
+
+    Examples:
+        ("packages", 0, "gomod", "path") -> "path"
+        ("packages", 0, "gomod", "what") -> "what"
+        ("packages", 0)                  -> ""   (no meaningful sub-field)
+        ("packages",)                    -> ""   (no meaningful sub-field)
+        ("what",)                        -> "what"
```
eskultety: Looking at this I would suggest replacing this with doctests using actual pydantic real world data as input for validation. I see you added this as a pytest unit test, that's fine, but doctest would be faster and less overhead since it's just slicing and string work.

hermeto/core/models/input.py
```diff
@@ -68,36 +68,76 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
...
+    """
+    parts = [str(p) for p in loc]
+    # Remove leading "packages" and any numeric index that follows it
+    while parts and (parts[0] == "packages" or parts[0].isdigit()):
+        parts.pop(0)
+    # Remove discriminated-union type name if it is the next remaining part
+    if parts and parts[0] in _KNOWN_BACKEND_TYPES:
+        parts.pop(0)
+    return " -> ".join(parts)
```
eskultety: This feels inefficient. Can't we do just:
```python
while i < len(loc) and (loc[i] == "packages" or isinstance(loc[i], int)):
    i += 1
if i < len(loc) and loc[i] in _KNOWN_BACKEND_TYPES:
    i += 1
return " -> ".join(str(p) for p in loc[i:])

File: hermeto/core/models/input.py [hunk 5]
```diff
@@ -68,36 +68,76 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
...
             # Handle regular union tag errors (i.e. errors which stem from
             # a missing package manager implementation). Errors in experimental
             # package managers are handled elsewhere.
             ctx = error.get("ctx", {})
             raw = ctx.get("expected_tags", "")
             expected = [t.strip(" '") for t in raw.split(",")]
             quoted = ", ".join(f"'{t}'" for t in sorted(expected))
-            message = f"Requested backend type '{ctx.get('tag', '<unknown>')}' doesn't match expected ones: {quoted}"
+            tag = ctx.get("tag", "<unknown>")
+            return f"Unknown package manager '{tag}'. Valid options are: {quoted}"
```
eskultety: Why the early return? Hmm, thinking outloud - can you remind me what the error message would look like for `union_tag_invalid`? I feel like we're stripping valuable information for the end user in multi-package inputs because if either reports a `union_tag_invalid` the user might not know which one, because of the early return.

eskultety (changes requested):
The second commit needs to be squashed to the first one. It's true this PR cosmetically improves the error message, but does not resolve the linked issue because the error duplication is still in place.

eskultety:
@abhayrajjais01 are you planning on incorporating the review feedback or shall we close this?

abhayrajjais01:
Hi @eskultety,
I will back with the fix soon

hermeto/core/models/input.py
```diff
@@ -67,56 +67,104 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
...
+
+def _simplify_location(loc: tuple[str | int, ...]) -> str:
+    """Strip pydantic-internal prefixes from an error location, keeping only meaningful parts.
+
+    Removes leading 'packages', numeric indices, and discriminated-union type names
+    so that users see the field name rather than the full internal path.
+
+    >>> _simplify_location(("packages", 0, "gomod", "path"))
+    'path'
+    >>> _simplify_location(("packages", 0, "gomod", "what"))
```
a-ovchinnikov: Nitpick: this test case is identical to one above.

hermeto/core/models/input.py
```diff
@@ -67,56 +67,104 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
...
+    'requirements_files -> 0'
+    """
+    i = 0
+    # Skip leading "packages" key and any numeric package index that follows
+    while i < len(loc) and (loc[i] == "packages" or isinstance(loc[i], int)):
+        i += 1
+    # Skip the discriminated-union type name (e.g. "gomod", "pip") if present
+    if i < len(loc) and loc[i] in _KNOWN_BACKEND_TYPES:
+        i += 1
+    return " -> ".join(str(p) for p in loc[i:])
```
a-ovchinnikov: Two observations here: first, the comments, they tell what a piece of code is doing, that is either a sign that a piece of code must be moved to a separate, properly named function or that logic is too convoluted and could be simlified. Why did you chose to add these comments here, by the way? Second, the code is written in C-style, it takes additional effort to understand it. WDYT about something along these lines:
```python
def _simplify_location(loc):                                                                  
    while loc:                                                                  
        head = loc[0]                                                           
        if head == "packages" or head in _KNOWN_BACKEND_TYPES or isinstance(head, int):            
            loc = loc[1:]                                                       
        else:                                                                   
            break                                                               
    return " -> ".join(str(e) for e in loc)
```
Here the idea is the same, but without the need to explicitly deal with a pointer into an iterable.

hermeto/core/models/input.py
```diff
@@ -67,56 +67,104 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
...
+            # Include the package index so users know which entry is invalid
+            # when multiple packages are specified in the same request.
+            location = _simplify_location(error["loc"])
+            if location:
+                return f"{location}\n  {msg}"
+            return msg
+
+        location = _simplify_location(error["loc"])
+        message = error["msg"]
+        if location:
```
a-ovchinnikov: Lines 149-152 and lines 154-158 are essentially the same, could you please drop the duplication?

hermeto/core/models/input.py
```diff
@@ -67,56 +67,104 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
...
             message = f"{location}\n  {message}"
-
         return message
 
-    header = f"{n_errors} validation error{'' if n_errors == 1 else 's'} for user input"
+    if n_errors == 1:
+        return show_error(errors[0])
```
a-ovchinnikov: Why did you choose to drop the header here?

hermeto/core/models/input.py
```diff
@@ -67,56 +67,104 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
...
+        i += 1
+    # Skip the discriminated-union type name (e.g. "gomod", "pip") if present
+    if i < len(loc) and loc[i] in _KNOWN_BACKEND_TYPES:
+        i += 1
+    return " -> ".join(str(p) for p in loc[i:])
+
+
 def _present_user_input_error(validation_error: pydantic.ValidationError) -> str:
-    """Make a slightly nicer representation of a pydantic.ValidationError.
+    """Make a user-friendly representation of a pydantic.ValidationError.
```
a-ovchinnikov: That's a bold claim. Can you tell off the top of your head what `"Error: InvalidInput: Unable to extract tag using discriminator 'type'",` in the context of ` '[{"type": "gomod"}, {}]'` means? I, for instance, cannot, I have to think for some time to figure it out. Actually this is how this should be done eventually: the error message should tell the user what went wrong and where, maybe instead of dropping the path it should be used to reconstruct the offending part of the input? I.e. instead of "Unable to use foo discriminator for bar" Hermeto should say "foo was expected to be found in the input in highlighted position {quux: {>>>foo<<<: bar}} but was not found there". I suggest at least thinking through and discussing this approach.

a-ovchinnikov (commented):
The most important comment is https://github.com/hermetoproject/hermeto/pull/1339#discussion_r3592116347, I suggest addressing it first, it might make other comments obsolete. _Edit_: corrected link to comment

abhayrajjais01:
@a-ovchinnikov ,
You are completely right; this error is still pydantic jargon (hence why I softened the docstring's claim)
 Reconstructing the offending part of the input structure is a great idea, and I'd love to track it as a separate enhancement issue so we can discuss the implementation details

a-ovchinnikov:
@abhayrajjais01 why do you think that a separate issue is necessary to figure out the best approach to resolving the current issue?

abhayrajjais01:
@a-ovchinnikov,
after thinking about this i think no separate issue needed,
i will  look into reconstructing the offending input structure from the error location path and include it directly in the error message within this PR

a-ovchinnikov:
@abhayrajjais01 hold on, please note, I am not telling you what to do with your PR, I am inviting you to discuss this approach since you chose to work on the issue. My suggestion looks like a better UX to me, but right now you have a better context of what exactly is going on in this particular area of code so you likely have some thoughts on how hard it could be to implement, whether there is already enough data at place etc. I would be very interested in hearing your opinion. Your previous response is an opinion, what I think it lacks is an explanation of why you think that work does not fit here. Please share your thoughts.

abhayrajjais01:
@a-ovchinnikov,
i was just trying to think in other way but now i have implemented that in this pr itself please take a look and tell me if something needs to be clear or change

a-ovchinnikov:
>  now i have implemented that in this pr itself please take a look and tell me if something needs to be clear or change

Yes, @abhayrajjais01  please share your thoughts on better UX in the thread about better UX.

PR #1346, opened 2026-03-02
fix: retry HTTP 429 responses and respect Retry-After header in async path

### LINKED ISSUE #1177: aiohttp_retry configuration does not retry HTTP 429
The HTTP 429 client error response is currently not retried by our aiohttp_retry configuration. We should investigate whether aiohttp_retry can be configured to respect and handle the `Retry-After` header like urllib3 does:

- https://github.com/urllib3/urllib3/blob/b20c8368ec02421ecc4e49d796cd5eff1e04f325/src/urllib3/util/retry.py#L404
- https://github.com/urllib3/urllib3/blob/b20c8368ec02421ecc4e49d796cd5eff1e04f325/src/urllib3/util/retry.py#L358

farhann-saleem:
## Summary   
                                                                                                                                                                                   
  Closes #1177                                     
               
HTTP 429 (Too Many Requests) responses were not being retried by either the sync or async request paths. Additionally, `aiohttp_retry.JitterRetry.get_timeout()` [accepts but does not read](https://github.com/inyutin/aiohttp_retry/blob/master/aiohttp_retry/retry_options.py#L128-L133) the `response` parameter (`# noqa: ARG002`  unused argument), so  the `Retry-After` header is never checked. Retry timing is computed solely from the attempt number and jitter randomization, ignoring the server's explicit backoff instruction.

  ## What changed

  **`hermeto/core/http_requests.py`**
  - Added `429` to `DEFAULT_RETRY_OPTIONS["status_forcelist"]`. This covers both the sync path (`urllib3.Retry`, which already respects `Retry-After` natively once 429 is in the  list) and feeds into the async path's status set.
  - Added `RetryAfterJitterRetry`, a subclass of `aiohttp_retry.JitterRetry` that overrides `get_timeout()` to check for a `Retry-After` header before falling back to the default  jitter backoff. Parsing supports both RFC 7231 formats: integer seconds (`Retry-After: 120`) and HTTP-date (`Retry-After: Fri, 31 Dec 1999 23:59:59 GMT`). The parsed value is  capped at `max_timeout` to prevent indefinite stalling.

  **`hermeto/core/package_managers/general.py`**
  - Replaced `JitterRetry` with `RetryAfterJitterRetry` in `async_download_files`.

  **`tests/unit/test_http_requests.py`** (new file)
  - 19 unit tests covering: integer parsing, whitespace handling, HTTP-date parsing, max_timeout cap, zero/negative/empty/invalid/float fallback to jitter, missing header, `None`
  response (connection error), and `DEFAULT_RETRY_OPTIONS` content.

  ## Why this approach

  `aiohttp_retry` has no built-in `Retry-After` support ([aiohttp_retry source](https://github.com/inyutin/aiohttp_retry)). The two realistic options were:

  1. **Subclass `JitterRetry` and override `get_timeout()`** : minimal surface area, uses the existing retry machinery, no monkey-patching.
  2. **Wrap the entire retry loop manually**  :more code, duplicates what `aiohttp_retry` already handles (attempt counting, exception filtering, status matching), higher  maintenance burden.

  Option 1 is the obvious choice. The subclass only touches timeout calculation; everything else (attempt limits, status filtering, exception handling) stays with the parent class   unchanged.

  For the sync path (`urllib3.Retry`): no code change needed beyond adding 429 to the status list. urllib3 already reads `Retry-After` when `respect_retry_after_header=True` (the  default).

  ## Verification

  - All 19 new unit tests pass
  - `ruff check` clean on all changed files
  - `mypy` clean on all changed files (no new type errors)
  - Existing test suite unaffected "DEFAULT_RETRY_OPTIONS` is consumed by reference, so adding 429 flows through to all callers without signature changes

  ## AI disclosure

  I wrote this change and the code. Claude Code was used as a verification tool to check consistency across the sync/async paths, confirm test coverage gaps, and validate that  nothing breaks. The architectural decisions (subclassing vs wrapping, combining 429 + Retry-After) are mine.

### DIFF
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -1,7 +1,9 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import asyncio
+import email
 import logging
 import ssl
+import time
 import types
 from collections.abc import Mapping
 from types import TracebackType
@@ -28,10 +30,69 @@ _pkg_requests_session: requests.Session | None = None
 
 SAFE_REQUEST_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE"})
 BACKOFF_FACTOR = 1.3
-STATUS_FORCELIST = (500, 502, 503, 504)
+STATUS_FORCELIST = (429, 500, 502, 503, 504)
 
 
 log = logging.getLogger(__name__)
+
+
+class RetryAfterJitterRetry(aiohttp_retry.JitterRetry):
+    """JitterRetry that uses the Retry-After header value when present."""
+
+    def __init__(
+        self,
+        max_timeout: float = 3600.0,  # 1 hour should be enough for this use case
+        **kwargs: Any,
+    ) -> None:
+        """Initialize the RetryAfterJitterRetry.
+
+        :param max_timeout: The maximum timeout in seconds.
+        :param kwargs: Keyword arguments for the JitterRetry constructor.
+        """
+        kwargs["max_timeout"] = max_timeout
+        super().__init__(**kwargs)
+
+    def get_timeout(
+        self,
+        attempt: int,
+        response: aiohttp_retry.ClientResponse | None = None,
+    ) -> float:
+        """Return the retry delay, preferring the Retry-After header when present."""
+        default = super().get_timeout(attempt, response)
+
+        if response is None:
+            return default
+
+        retry_after = response.headers.get("Retry-After")
+        if retry_after is None:
+            return default
+
+        parsed = self._parse_retry_after(retry_after)
+        if parsed is None:
+            log.warning("Unparseable Retry-After header: '%s'", retry_after)
+            return default
+        if parsed < 0:
+            log.warning("Retry-After resolved to negative delay (%.1fs): '%s'", parsed, retry_after)
+            return default
+
+        return min(parsed, self._max_timeout)
+
+    @staticmethod
+    def _parse_retry_after(retry_after: str) -> float | None:
+        """Parse a Retry-After header value into seconds, or None if unparseable."""
+        # Try numeric seconds first (most common for 429 responses)
+        try:
+            return float(retry_after)
+        except ValueError:
+            pass
+
+        # Try HTTP-date format
+        parsed_date = email.utils.parsedate_tz(retry_after)
+        if parsed_date is not None:
+            retry_timestamp = email.utils.mktime_tz(parsed_date)
+            return retry_timestamp - time.time()
+
+        return None
 
 
 class SyncLoggingRetry(Retry):
@@ -255,7 +316,7 @@ async def async_download_files(
     max_retries = get_config().http.max_retries
     # aiohttp uses n calls (1 call, n-1 retries).
     max_retries = max_retries + 1
-    retry_options = aiohttp_retry.JitterRetry(
+    retry_options = RetryAfterJitterRetry(
         start_timeout=BACKOFF_FACTOR,
         attempts=max_retries,
         statuses=set(STATUS_FORCELIST),
--- a/tests/unit/package_managers/test_general.py
+++ b/tests/unit/package_managers/test_general.py
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import asyncio
+import email
 import random
+import time
 from pathlib import Path
 from typing import Any
 from unittest import mock
@@ -17,6 +19,7 @@ from hermeto.core.config import get_config
 from hermeto.core.errors import FetchError
 from hermeto.core.package_managers import general
 from hermeto.core.package_managers.general import (
+    RetryAfterJitterRetry,
     _async_download_binary_file,
     _get_pkg_requests_session,
     async_download_files,
@@ -295,3 +298,75 @@ async def test_async_download_preserves_redirect_url_encoding(tmp_path: Path) ->
         download_path = tmp_path / "artifact"
         await async_download_files({url: str(download_path)}, concurrency_limit=1)
         assert b"text%2Fplain" in download_path.read_bytes()
+
+
+class TestRetryAfterJitterRetry:
+    _FALLBACK_TIMEOUT = 1.0
+
+    @pytest.fixture()
+    def retry(self) -> RetryAfterJitterRetry:
+        return RetryAfterJitterRetry()
+
+    @pytest.fixture()
+    def mock_response(self) -> mock.Mock:
+        resp = mock.Mock(spec=aiohttp_retry.ClientResponse)
+        resp.headers = {}
+        return resp
+
+    @pytest.mark.parametrize(
+        "retry_after, expected",
+        [
+            pytest.param("5", 5.0, id="integer"),
+            pytest.param("0", 0.0, id="zero"),
+            pytest.param("2.5", 2.5, id="float"),
+        ],
+    )
+    def test_get_timeout_valid_seconds(
+        self,
+        retry: RetryAfterJitterRetry,
+        mock_response: mock.Mock,
+        retry_after: str,
+        expected: float,
+    ) -> None:
+        mock_response.headers = {"Retry-After": retry_after}
+        assert retry.get_timeout(attempt=1, response=mock_response) == expected
+
+    def test_get_timeout_valid_http_date(
+        self,
+        retry: RetryAfterJitterRetry,
+        mock_response: mock.Mock,
+    ) -> None:
+        future = time.time() + 10
+        mock_response.headers = {"Retry-After": email.utils.formatdate(future, usegmt=True)}
+        timeout = retry.get_timeout(attempt=1, response=mock_response)
+        assert timeout == pytest.approx(10, abs=1)
+
+    def test_get_timeout_capped_at_max_timeout(self, mock_response: mock.Mock) -> None:
+        retry = RetryAfterJitterRetry(max_timeout=10.0)
+        mock_response.headers = {"Retry-After": "60"}
+        assert retry.get_timeout(attempt=1, response=mock_response) == 10.0
+
+    @pytest.mark.parametrize(
+        "headers",
+        [
+            pytest.param({}, id="no-header"),
+            pytest.param({"Retry-After": "-5"}, id="negative"),
+            pytest.param({"Retry-After": "not-a-number"}, id="unparseable"),
+        ],
+    )
+    @mock.patch("aiohttp_retry.JitterRetry.get_timeout", return_value=_FALLBACK_TIMEOUT)
+    def test_get_timeout_falls_back_to_default(
+        self,
+        mock_get_timeout: mock.Mock,
+        retry: RetryAfterJitterRetry,
+        mock_response: mock.Mock,
+        headers: dict[str, str],
+    ) -> None:
+        mock_response.headers = headers
+        assert retry.get_timeout(attempt=1, response=mock_response) == self._FALLBACK_TIMEOUT
+
+    @mock.patch("aiohttp_retry.JitterRetry.get_timeout", return_value=_FALLBACK_TIMEOUT)
+    def test_get_timeout_no_response(
+        self, mock_get_timeout: mock.Mock, retry: RetryAfterJitterRetry
+    ) -> None:
+        assert retry.get_timeout(attempt=1, response=None) == self._FALLBACK_TIMEOUT


farhann-saleem:
@taylormadore   @slimreaper35 @a-ovchinnikov 
would love your feedback on the implementation

tests/unit/test_http_requests.py
ST3V1K, line 1: This file feels a bit overengineered. Most of the tests test the same thing, and it's usually stdlib behavior. I suggest dropping a lot of these tests and keeping only a few from each category (number, http date, invalid). Then the rest could be parametrized using `@pytest.mark.parametrize`.

tests/unit/test_http_requests.py
```diff
@@ -0,0 +1,156 @@
...
+    """Tests for DEFAULT_RETRY_OPTIONS configuration."""
+
+    def test_status_forcelist_includes_429(self) -> None:
+        """Verify HTTP 429 (Too Many Requests) is in the retry status codes."""
+        assert 429 in DEFAULT_RETRY_OPTIONS["status_forcelist"]
+
+    def test_status_forcelist_includes_server_errors(self) -> None:
+        """Verify standard server error codes are in the retry status codes."""
+        for status in (500, 502, 503, 504):
+            assert status in DEFAULT_RETRY_OPTIONS["status_forcelist"]
```
ST3V1K: This only asserts hardcoded values. I would remove it.

hermeto/core/http_requests.py
```diff
@@ -20,10 +24,52 @@
...
+    @staticmethod
+    def _parse_retry_after(retry_after: str) -> float | None:
+        """Parse a Retry-After header value into seconds, or None if unparseable."""
+        retry_after = retry_after.strip()
+        if not retry_after:
+            return None
+
+        # Try integer seconds first (most common for 429 responses)
+        try:
+            return float(int(retry_after))
```
ST3V1K: I think we could allow float values since the intent is still clear enough, even tho it's not really in the standard. Using the default jitter implementation because of that is a bit extreme.

hermeto/core/http_requests.py
```diff
@@ -20,10 +24,52 @@
...
+class RetryAfterJitterRetry(JitterRetry):
+    """JitterRetry that uses the Retry-After header value when present."""
+
+    def get_timeout(
+        self,
+        attempt: int,
+        response: ClientResponse | None = None,
+    ) -> float:
+        """Return the retry delay, preferring the Retry-After header when present."""
+        if response is not None:
```
ST3V1K: Early returns might be better.

hermeto/core/http_requests.py
```diff
@@ -20,10 +24,52 @@
...
+            retry_after = response.headers.get("Retry-After")
+            if retry_after is not None:
+                parsed = self._parse_retry_after(retry_after)
+                if parsed is not None and parsed > 0:
+                    timeout = min(parsed, self._max_timeout)
+                    log.debug("Using Retry-After header value: %s seconds", timeout)
+                    return timeout
```
ST3V1K: IMO, we should warn when the header resolves to negative seconds.

hermeto/core/http_requests.py
```diff
@@ -20,10 +24,52 @@
...
+        try:
+            return float(int(retry_after))
+        except ValueError:
+            pass
+
+        # Try HTTP-date format
+        parsed_date = email.utils.parsedate_tz(retry_after)
+        if parsed_date is not None:
+            retry_timestamp = email.utils.mktime_tz(parsed_date)
+            return max(0.0, retry_timestamp - time.time())
```
ST3V1K: I would either:
 - return negative values — match how you handle seconds

or

 - return `None` when seconds are negative

I'd lean more towards returning negative values since you already handle that in `get_timeout`.

hermeto/core/http_requests.py
```diff
@@ -13,5 +18,48 @@
...
+
+        if response is None:
+            return default
+
+        retry_after = response.headers.get("Retry-After")
+        if retry_after is None:
+            return default
+
+        parsed = self._parse_retry_after(retry_after)
+        if parsed is None or parsed < 0:
```
derasdf: `parsed is None` (unparseable garbage) and `parsed < 0` (valid parse but time already passed) are semantically different. One is a server bug worth investigating, the other is a timing edge case. Splitting the warning helps debugging:

```python
if parsed is None:
    log.warning("Unparseable Retry-After header: '%s'", retry_after)
    return default
if parsed < 0:
    log.warning("Retry-After resolved to negative delay (%.1fs): '%s'", parsed, retry_after)
    return default
```

tests/unit/test_http_requests.py
```diff
@@ -0,0 +1,80 @@
...
+
+import pytest
+from aiohttp import ClientResponse
+
+from hermeto.core.http_requests import RetryAfterJitterRetry
+
+_FALLBACK_TIMEOUT = 1.0
+
+
+@pytest.fixture(scope="session")
```
derasdf: scope="session" shares one instance across all tests, but the fallback tests patch JitterRetry.get_timeout globally, that patch leaks into other tests using the same instance. Just drop the scope, the constructor is cheap.

tests/unit/test_http_requests.py
```diff
@@ -0,0 +1,80 @@
...
+def mock_response() -> mock.Mock:
+    resp = mock.Mock(spec=ClientResponse)
+    resp.headers = {}
+    return resp
+
+
+@pytest.mark.parametrize(
+    "retry_after, expected",
+    [
+        pytest.param("0", 0.0, id="zero"),
```
derasdf: nitpick: Parametrize covers "0" and "2.5" but not a plain integer like "120", which is the primary RFC 7231 Retry-After format.

hermeto/core/http_requests.py
```diff
@@ -13,5 +18,48 @@
...
+            log.warning("Invalid Retry-After header value: '%s'", retry_after)
+            return default
+
+        return min(parsed, self._max_timeout)
+
+    @staticmethod
+    def _parse_retry_after(retry_after: str) -> float | None:
+        """Parse a Retry-After header value into seconds, or None if unparseable."""
+        # Try integer (float) seconds first (most common for 429 responses)
```
derasdf: nitpick: s/integer (float)/numeric/

hermeto/core/http_requests.py
ST3V1K: Done

tests/unit/test_http_requests.py
ST3V1K: Done

tests/unit/test_http_requests.py
ST3V1K: Done

hermeto/core/http_requests.py
ST3V1K: Done

hermeto/core/http_requests.py
```diff
@@ -13,5 +18,51 @@
...
+
+        parsed = self._parse_retry_after(retry_after)
+        if parsed is None:
+            log.warning("Unparseable Retry-After header: '%s'", retry_after)
+            return default
+        if parsed < 0:
+            log.warning("Retry-After resolved to negative delay (%.1fs): '%s'", parsed, retry_after)
+            return default
+
+        return min(parsed, self._max_timeout)
```
taylormadore: `_max_timeout` is only 30s by default. That might not be long enough for this. I'm not sure what the right number is though. There is no documented max. urllib3 uses 6 hours, which feels quite lengthy...

hermeto/core/http_requests.py
```diff
@@ -13,5 +18,51 @@
...
+
+    @staticmethod
+    def _parse_retry_after(retry_after: str) -> float | None:
+        """Parse a Retry-After header value into seconds, or None if unparseable."""
+        # Try numeric seconds first (most common for 429 responses)
+        try:
+            return float(retry_after)
```
taylormadore: This is a non-negative integer (not a float) in the RFC: https://httpwg.org/specs/rfc9110.html#rfc.section.10.2.3

hermeto/core/http_requests.py
```diff
@@ -13,5 +18,51 @@
...
+
+        retry_after = response.headers.get("Retry-After")
+        if retry_after is None:
+            return default
+
+        parsed = self._parse_retry_after(retry_after)
+        if parsed is None:
+            log.warning("Unparseable Retry-After header: '%s'", retry_after)
+            return default
+        if parsed < 0:
```
taylormadore: If this is in the past, why not just delay zero?

hermeto/core/http_requests.py
ST3V1K: This is actually something the original author had, but I think that this would be better for a few reasons:
1) Even though floats are not the standard. Rejecting, let's say `1.5`, and falling back to jitter implementation is worse IMO 
1) It wouldn't break the standard; it would only superset it
1) Return value must be a float either way
1) HTTP date can already be resolved to a float value, so I'd argue that floats are still used in the standard 

My original comment:
> I think we could allow float values since the intent is still clear enough, even tho it's not really in the standard. Using the default jitter implementation because of that is a bit extreme.

hermeto/core/http_requests.py
ST3V1K: If it's only a few seconds in the past (due to a lag or something), it could just delay 0 seconds. However, when it's e.g. 100 years in the past, what exactly is the intent there? I think there still should be some guard with a threshold.

farhann-saleem: kinda sorry . i was super busy in exams and couldnt get to do any code changes . thanks @ST3V1K for taking this .

hermeto/core/http_requests.py
ST3V1K: Yes, thank you. I completely missed that. I think we could use 1 hour — nobody would ever wait more than that in the context of Hermeto.

hermeto/core/package_managers/general.py
```diff
@@ -16,15 +16,16 @@
...
+)
 from hermeto.core.scm import get_repo_id
 from hermeto.core.type_aliases import StrPath
 
 _pkg_requests_session: requests.Session | None = None
 
-SAFE_REQUEST_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE"})
-BACKOFF_FACTOR = 1.3
-STATUS_FORCELIST = (500, 502, 503, 504)
-
```
slimreaper35, line 32: Let's not revert the commit and add back dead code.

I would suggest moving things from `general.py` to `http_requests.py` or vice versa. (It seems that most of the `general.py` module is already related to HTTP requests.)

slimreaper35 (changes requested):

hermeto/core/package_managers/general.py
ST3V1K: Done, moved to `general.py`

hermeto/core/package_managers/general.py
```diff
@@ -28,12 +30,71 @@
...
 
 log = logging.getLogger(__name__)
 
 
+class RetryAfterJitterRetry(aiohttp_retry.JitterRetry):
+    """JitterRetry that uses the Retry-After header value when present."""
+
+    def __init__(
+        self,
+        max_timeout: float = 3600.0,  # 1 hour should be enough for this use case
```
a-ovchinnikov, line 44: One hour is a lot, I wonder where did this value come from?

hermeto/core/package_managers/general.py
```diff
@@ -28,12 +30,71 @@
...
+        if response is None:
+            return default
+
+        retry_after = response.headers.get("Retry-After")
+        if retry_after is None:
+            return default
+
+        parsed = self._parse_retry_after(retry_after)
+        if parsed is None:
+            log.warning("Unparseable Retry-After header: '%s'", retry_after)
```
a-ovchinnikov, line 72: Do we know at this point who sent an unparseable header? Probably if we care enough to warn that a header cannot be parsed we should tell users who sent them this header.

hermeto/core/package_managers/general.py
```diff
@@ -28,12 +30,71 @@
...
+    def _parse_retry_after(retry_after: str) -> float | None:
+        """Parse a Retry-After header value into seconds, or None if unparseable."""
+        # Try numeric seconds first (most common for 429 responses)
+        try:
+            return float(retry_after)
+        except ValueError:
+            pass
+
+        # Try HTTP-date format
+        parsed_date = email.utils.parsedate_tz(retry_after)
```
a-ovchinnikov, line 90: Why `email` and not, say, `dateutils.parser.parse`? The comment on line 89 tells a reader what the code does without explaining why it does that, I suggest replacing it with a sentence stating that in some cases it could be HTTP-date.

tests/unit/package_managers/test_general.py
```diff
@@ -295,3 +298,75 @@ async def target_handler(request: web.Request) -> web.Response:
...
+        mock_get_timeout: mock.Mock,
+        retry: RetryAfterJitterRetry,
+        mock_response: mock.Mock,
+        headers: dict[str, str],
+    ) -> None:
+        mock_response.headers = headers
+        assert retry.get_timeout(attempt=1, response=mock_response) == self._FALLBACK_TIMEOUT
+
+    @mock.patch("aiohttp_retry.JitterRetry.get_timeout", return_value=_FALLBACK_TIMEOUT)
+    def test_get_timeout_no_response(
```
a-ovchinnikov, line 369: Do you mean that you are testing a scenario where no response from a server results in a retry after a default timeout? I.e.
`test_no_response_from_server_results_in_default_timeout_for_the_next_attempt()`

tests/unit/package_managers/test_general.py
```diff
@@ -295,3 +298,75 @@ async def target_handler(request: web.Request) -> web.Response:
...
+        mock_response.headers = {"Retry-After": retry_after}
+        assert retry.get_timeout(attempt=1, response=mock_response) == expected
+
+    def test_get_timeout_valid_http_date(
+        self,
+        retry: RetryAfterJitterRetry,
+        mock_response: mock.Mock,
+    ) -> None:
+        future = time.time() + 10
+        mock_response.headers = {"Retry-After": email.utils.formatdate(future, usegmt=True)}
```
a-ovchinnikov, line 340: This test essentially checks that `email.utils.parse(email.utils.format)` chain works. It also uses varying, effectively random data without any strategy for replicating it (which is probably not a problem in this particular case, but should be avoided in general).

hermeto/core/package_managers/general.py
ST3V1K: https://github.com/hermetoproject/hermeto/pull/1346#discussion_r3413871178
Taylor pointed out in this comment that urllib3 uses 6 hours, which is quite long. So I just picked an hour. 
Maybe 15 minutes could be fine.

hermeto/core/package_managers/general.py
ST3V1K: I didn't know about that. It makes it cleaner, thanks!

tests/unit/package_managers/test_general.py
ST3V1K: You could say the same about the valid-seconds test -- that’s mostly `float()`. I’d argue both are checking that `RetryAfterJitterRetry` passes the header through correctly. 
On the non-determinism: fair, though for a relative “~10s from now” assertion it shouldn’t matter much.

hermeto/core/package_managers/general.py
ST3V1K: Done

tests/unit/package_managers/test_general.py
ST3V1K: Done. But I modified your suggested name to `test_get_timeout_no_response_falls_back_to_jitter` to match the other tests.

a-ovchinnikov: That's fine, as long as it tells a reader what it is about to test.

hermeto/core/package_managers/general.py
a-ovchinnikov: Fair enough, one hour is just an eye-catching number.

pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
   "pydantic",
   "pydantic-settings",
   "pypi-simple",
+  "python-dateutil",
```
a-ovchinnikov: TIL that `dateutil` is not built-in. Now I will need to figure out how come it is available system-wide for me. I won't block on this, but instead of adding a new dependency a helper could be defined along the lines of `datetime.strptime(formatdate(time()), "%a, %d %b %Y %H:%M:%S %z")`

a-ovchinnikov (dismissed):

pyproject.toml
eskultety: No new dependency for parsing dates please, essentially for a single function, `datetime` should have all the magic we'd need when it comes to representing and parsing dates.

ST3V1K: Done, used `email.utils.parsedate_to_datetime` which is stdlib

hermeto/core/package_managers/general.py
```diff
@@ -2,8 +2,10 @@
 import asyncio
 import logging
 import ssl
+import time
 import types
 from collections.abc import Mapping
+from email.utils import parsedate_to_datetime
```
a-ovchinnikov: Non-blocking, not affecting my approval: an import from `email` stands out because the code below is not dealing with emails. To make it stand out less I would have at least laundered it with a wrapper in hermeto's utils or just used `datetime`'s parser.

a-ovchinnikov:
Hi @farhann-saleem, please note, that the changes you pushed have merge conflicts and are not following contributing guidelines.

a-ovchinnikov:
Hi @farhann-saleem, please note, that your latest push rolled back all the progress made in this PR. It also introduced merge conflicts. There was no activity from your side for two weeks after that. If you still have any plans for continuing working on this PR please do it within this week, otherwise I would ask you to not interfere with the overtaken PR.

_Edit_: for some reason I cannot revoke my approval from the PR, but it must be considered not approved by me until it is fixed.

hermeto/core/package_managers/general.py
```diff
@@ -29,12 +33,80 @@
...
+            return default
+
+        retry_after = response.headers.get("Retry-After")
+        if retry_after is None:
+            return default
+
+        parsed = self._parse_retry_after(retry_after)
+        if parsed is None:
+            log.warning(
+                "Unparseable Retry-After header from %s: '%s'", response.url.host, retry_after
```
a-ovchinnikov, line 76: @farhann-saleem why did you choose to specifically log just the host part of the URL?

hermeto/core/package_managers/general.py
```diff
@@ -29,12 +33,80 @@
...
+            return default
+
+        return min(parsed, self._max_timeout)
+
+    @staticmethod
+    def _parse_retry_after(retry_after: str) -> float | None:
+        """Parse a Retry-After header value into seconds, or None if unparseable."""
+        try:
+            seconds = float(retry_after)
+            return seconds if math.isfinite(seconds) else None
```
a-ovchinnikov, line 95: @farhann-saleem could you please elaborate why do you think it is realistic to receive a request for an infinite number of seconds to wait?

tests/unit/package_managers/test_general.py
```diff
@@ -283,3 +287,155 @@ async def target_handler(request: web.Request) -> web.Response:
         download_path = tmp_path / "artifact"
         await async_download_files({url: str(download_path)}, concurrency_limit=1)
         assert b"text%2Fplain" in download_path.read_bytes()
+
+
+class TestRetryAfterJitterRetry:
+    _FALLBACK_TIMEOUT: float = 1.0
```
a-ovchinnikov, line 293: This is over-annotation. The point of annotation is to disambiguate. In this case it is perfectly clear that _FALLBACK_TIMEOUT is a float. No need in adding extra visual noise when it could be avoided.

tests/unit/package_managers/test_general.py
```diff
@@ -283,3 +287,155 @@ async def target_handler(request: web.Request) -> web.Response:
...
+        "retry_after",
+        [
+            "Sun, 06 Nov 1994 08:49:37 GMT",
+            "Sunday, 06-Nov-94 08:49:37 GMT",
+            "Sun Nov  6 08:49:37 1994",
+            "Sun, 06 Nov 1994 08:49:37 -0000",
+        ],
+    )
+    @mock.patch("hermeto.core.package_managers.general.time.time", return_value=784111767.0)
+    def test_http_dates_use_utc(
```
a-ovchinnikov, line 377: @farhann-saleem please correct me if I am wrong, but the patch does not implement date parsing, it relies on a standard library parser. this parser is tested [here](https://github.com/python/cpython/blob/main/Lib/test/test_email/test_utils.py). Why did you decide to test third-party library in Hermeto tests?

tests/unit/package_managers/test_general.py
```diff
@@ -283,3 +287,155 @@ async def target_handler(request: web.Request) -> web.Response:
...
+        mock_response.headers = {"Retry-After": retry_after}
+        retry.get_timeout(attempt=1, response=mock_response)
+        assert "example.org" in caplog.text
+        assert "token=secret" not in caplog.text
+
+
+@pytest.mark.asyncio
+async def test_async_download_retries_429(tmp_path: Path) -> None:
+    from aiohttp import web
+    from aiohttp.test_utils import TestServer
```
a-ovchinnikov, line 411: '@farhann-saleem please follow the project guidelines and import entities properly.

tests/unit/package_managers/test_general.py
```diff
@@ -283,3 +287,155 @@ async def target_handler(request: web.Request) -> web.Response:
...
+    requests_seen = 0
+
+    async def handler(request: web.Request) -> web.Response:
+        nonlocal requests_seen
+        requests_seen += 1
+        if requests_seen == 1:
+            return web.Response(status=429, headers={"Retry-After": "0"})
+        return web.Response(body=b"artifact")
+
+    app = web.Application()
```
a-ovchinnikov, line 422: @farhann-saleem please note, that the way you defined this test it is not a unit-test but an integration test. Furthermore, it is clearly testing third-party libraries again since this change essentially provides a thin  wrapper around them.

tests/unit/package_managers/test_general.py
```diff
@@ -283,3 +287,155 @@ async def target_handler(request: web.Request) -> web.Response:
...
+    assert requests_seen == 2
+    assert download_path.read_bytes() == b"artifact"
+
+
+def test_sync_session_retries_429_without_retry_after(monkeypatch: pytest.MonkeyPatch) -> None:
+    monkeypatch.setattr(general, "_pkg_requests_session", None)
+    session = _get_pkg_requests_session()
+    try:
+        adapter = session.get_adapter("https://example.org")
+        assert isinstance(adapter, HTTPAdapter)
```
a-ovchinnikov, line 438: @farhann-saleem could you please share what makes you think that adapter won't be an instance of HTTPAdapter in this particular case?

tests/unit/package_managers/test_general.py
```diff
@@ -283,3 +287,155 @@ async def target_handler(request: web.Request) -> web.Response:
...
+
+    @pytest.fixture()
+    def retry(self) -> RetryAfterJitterRetry:
+        return RetryAfterJitterRetry()
+
+    @pytest.fixture()
+    def mock_response(self) -> mock.Mock:
+        resp = mock.Mock(spec=aiohttp_retry.ClientResponse)
+        resp.headers = {}
+        resp.url = URL("https://example.org/artifact?token=secret")
```
a-ovchinnikov, line 303: @farhann-saleem please point to the part of pip backend which makes you think that tokens will ever be embedded into URLs.

a-ovchinnikov (changes requested):
Hi @farhann-saleem , please find my comments inline. Please note, that adding tests which make little sense does not improve PRs, quite the contrary. The same applies to guards against inputs which cannot exist in practice.

hermeto/core/package_managers/general.py
farhann-saleem: My reasoning was that the hostname identifies the responding server without including the full URL. However, that also loses diagnostic context, and I haven’t demonstrated a need for that trade-off here. I’ll restore the previous logging.

tests/unit/package_managers/test_general.py
farhann-saleem: Understood. I added it for explicitness, but the literal already makes the type clear. I’ll remove the annotation.

tests/unit/package_managers/test_general.py
farhann-saleem: I wasn’t basing this on pip embedding tokens itself. I was considering query parameters in download URLs, but I haven’t established a concrete case in this backend. That makes the fixture speculative. I’ll remove the token-specific test and keep the logging consistent with the previously reviewed implementation.

tests/unit/package_managers/test_general.py
farhann-saleem: I added this to verify that our configured retry policy is actually used during a download. However, using a local server does make it an integration-style test, and much of what it exercises belongs to the client library. I’ll remove it from the unit suite and focus the coverage on our timeout-selection logic.

tests/unit/package_managers/test_general.py
farhann-saleem: That assertion was for type narrowing, not because I expected another adapter at runtime. I see why it reads as an unnecessary behavioral check. I’ll remove this additional test, since it mostly verifies the library’s behavior.

tests/unit/package_managers/test_general.py
farhann-saleem: The intention was to test our timezone handling after parsing, rather than the parser itself. A naive datetime passed to timestamp() uses the machine’s local timezone, so our conversion affects the resulting delay. I agree that parametrizing several date formats mixes those concerns. I’ll mock the parser result and narrow the test to our UTC conversion.

hermeto/core/package_managers/general.py
farhann-saleem: I wasn’t expecting a server to intentionally request an infinite wait. The guard was because float() accepts values such as NaN and Infinity, which aren’t usable retry delays. That said, I don’t have an observed case motivating it here. I’ll remove this defensive handling and the associated tests to keep the patch focused.

eskultety:
Superseded by https://github.com/hermetoproject/hermeto/pull/1786, closing.

PR #1349, opened 2026-03-03
docs(usage.md): improve git workaround readability

harshitbansal184507:
Split Git workaround command into separate lines for better readability.
very minor change but looks good and improves readability

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -35,7 +35,9 @@ Make sure to set the remote as well.
 
 <!-- markdownlint-disable-next-line no-inline-html -->
 <pre>
-git init && git add -A && git commit -m "initial commit" && \
+git init
+git add -A
+git commit -m "initial commit"
 git remote add origin https://github.com/someorg/somerepo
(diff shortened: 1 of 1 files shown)


eskultety:
When commit https://github.com/hermetoproject/hermeto/pull/942/changes/2aded5b9587120318c2598d97e67c0ceb75e8bd0 introduced this it was a direct response to https://github.com/hermetoproject/hermeto/pull/856/. The backstory and intent there was to have a single easily copy-pastable command from the docs. Moreover, this workaround will soon hopefully not be needed after https://github.com/hermetoproject/hermeto/pull/1176.

This PR makes the situation only marginally more readable and in general I'm not in favour of accepting changes for the sake of making a change, there should always be some value behind a contribution. Based on my comments - NACK to this change, but if someone from the team feels otherwise I'm keeping this open for a while.

a-ovchinnikov:
> looks good and improves readability

That's purely a matter of taste. I prefer the original version.

PR #1351, opened 2026-03-03
fix(gomod): correct Go version regex typo in go.mod parsing

abhayrajjais01:
**Description**
The version parsing regex in [gomod.py](cci:7://file:///c:/Users/abhay/hemerto/hermeto/tests/integration/test_gomod.py:0:0-0:0) for `go.mod` and `go.work` files contained a character typo: `(:?\.\d+)` instead of the intended non-capturing group `(?:\.\d+)`. 
The former syntax inaccurately defined an optional capturing group that expects an optional colon character, inadvertently allowing malformed Go version strings (like `1.21:.0`) to be accepted.

This PR corrects the regex by properly defining a non-capturing group (`(?:\.\d+)`), ensuring malformed version components are correctly rejected.

**Changes**
* Fixed the `version_str_regex` assignment in [hermeto/core/package_managers/gomod.py](cci:7://file:///c:/Users/abhay/hemerto/hermeto/hermeto/core/package_managers/gomod.py:0:0-0:0) (Line 865).
* Added a new negative test case (`"go 1.21:.0"`) to `INVALID_VERSION_STRINGS` in [tests/unit/package_managers/test_gomod.py](cci:7://file:///c:/Users/abhay/hemerto/hermeto/tests/unit/package_managers/test_gomod.py/c:/Users/abhay/hemerto/hermeto/tests/unit/package_managers/test_gomod.py:0:0-0:0).

**Testing**
* Explictly verified the negative test case guarantees `go 1.21:.0` now throws a failure as expected.
* Verified that all previous valid baseline tests run successfully with zero regressions (`nox -s python` / `pytest`).

### DIFF
--- a/hermeto/core/package_managers/gomod.py
+++ b/hermeto/core/package_managers/gomod.py
@@ -862,7 +862,7 @@ def _get_gomod_version(go_mod_file: RootedPath) -> tuple[str | None, str | None]
     # - 'go 1.21.0'
     # - '   go 1.21.0rc4'
     # - 'go 1.21beta1//commentary'
-    version_str_regex = r"(?P<ver>\d+\.\d+(:?\.\d+)?(?:[a-zA-Z]+\d+)?)"
+    version_str_regex = r"(?P<ver>\d+\.\d+(?:\.\d+)?(?:[a-zA-Z]+\d+)?)"
     post_version_chars_regex = r"\s*(?:\/\/.*)?"
     go_version_regex = rf"^\s*go\s+{version_str_regex}{post_version_chars_regex}$"
     toolchain_version_regex = rf"^\s*toolchain\s+go{version_str_regex}{post_version_chars_regex}$"
--- a/tests/unit/package_managers/test_gomod.py
+++ b/tests/unit/package_managers/test_gomod.py
@@ -1973,6 +1973,7 @@ INVALID_VERSION_STRINGS = [
     "go 1.21prerelease",  # pre-release with no number
     "go 1.21prerelease_4",  # pre-release with non-alphanum character
     "toolchain 1.21",  # missing 'go' prefix for the toolchain spec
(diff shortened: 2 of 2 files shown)


eskultety:
@abhayrajjais01 can you please describe in what scenario you encountered this problem, what the error was?

abhayrajjais01:
> @abhayrajjais01 can you please describe in what scenario you encountered this problem, what the error was?

I found this edge-issue during code review in regex parser for go versions in gomod.py, the parsing regex accidentally used (:?\.\d+) instead of (?:\.\d+),  (?:) defines a non-capturing regex group, but because the (:) and (?) were switched, it created a capturing group that matched an optional valid colon instead, no colon should ever appear in a Go version string, so this accepts input it should reject
I did not hit any production issue but it hardens strict version validation

a-ovchinnikov:
@abhayrajjais01 thank you for your contribution. While the change looks formally correct it makes little to no sense in practice. I think that you (as someone who has studied the codebase carefully) can clearly see the reasons why.

abhayrajjais01:
> @abhayrajjais01 thank you for your contribution. While the change looks formally correct it makes little to no sense in practice. I think that you (as someone who has studied the codebase carefully) can clearly see the reasons why.

sorry if i missed something let me go through this

abhayrajjais01:
> @abhayrajjais01 thank you for your contribution. While the change looks formally correct it makes little to no sense in practice. I think that you (as someone who has studied the codebase carefully) can clearly see the reasons why.

now i understand,
a go.mod file is maintained by the Go toolchain which will never generate a malformed version like 1.21:.0. If a user manually edited it to be invalid, the project wouldn't even build under standard Go tooling anyway. further, since _get_gomod_version is slated for removal once Go 1.20 support is dropped, putting effort into perfectly hardening this obsolete regex fallback doesn't make practical sense
I will go ahead and close this PR. thanks for the quick review and guidance

Issue #950, opened 2025-05-15
Test-exclusion logic does not handle multi-PM tests correctly


a-ovchinnikov:
When a change is made to PM that is indirectly invoked by another PM (e.g. Cargo and Pip) then a corresponding Pip test is not automatically run. See https://github.com/hermetoproject/hermeto/actions/runs/15054329940/job/42316534191?pr=949 as an example.

eskultety:
This was a nice experiment, provided some worth for local use cases, but also came at a cost (and some confusion and bugs). A dev should be comfortable with filtering the tests to run with the `-k` pytest CLI option locally. This automatic logic loses its shine even more so with https://github.com/hermetoproject/hermeto/issues/1207 underway, so let's just drop the logic and keep it simple.

Jay2006sawant:
Hi @eskultety I am a new contributor here,I would love to contribute here !

PR #1354, opened 2026-03-04
tests(integration): drop automatic test-exclusion logic

sammy200-ui:
The test-exclusion logic that automatically skipped integration tests based on changed files did not handle multi-package-manager tests correctly. Developers can use pytest's -k option to filter tests locally instead. 

Closes: #950

Assisted-by: Claude


sammy200-ui:
the only added change is the old docstring as it was now incorrect after removing the filtering logic

a-ovchinnikov:
@sammy200-ui this PR needs to be rebased on main, LGTM otherwise

sammy200-ui:
@a-ovchinnikov done rebase , do have another look pls 😄

noxfile.py
```diff
@@ -124,7 +124,6 @@ def all_integration_tests(session: Session) -> None:
     _run_integration_tests(
         session,
         {
-            "HERMETO_TEST_RUN_ALL_INTEGRATION_TESTS": "1",
             "HERMETO_TEST_LOCAL_PYPISERVER": "1",
             "HERMETO_TEST_LOCAL_DNF_SERVER": "1",
         },
```
slimreaper35, line 129: I wonder if we still need the `all_integration_tests` nox session just for these two env variables. @a-ovchinnikov ?

a-ovchinnikov: To me it will act as a wrapper around setting these two vars, so I am ok with keeping the session. _Edit_: at least for now.

slimreaper35: ok, not a big deal

PR #1356, opened 2026-03-04
Added config subcommand for pretty printing config

### LINKED ISSUE #1253: [RFE][cli] Add 'config' subcommand for runtime configuration introspection
Our configuration options, i.e. file sections and settings, let alone the corresponding environment variable names for nested configuration (e.g., `HERMETO_GOMOD__PROXY_URL`) are hard to remember. FWIW we don't even have our config options documented properly in the online docs at the time of writing. Having a CLI-based config options reference can not only solve most of the problem, it provides some extra benefits:

  1. Quick lookup — get env var names and available options without context-switching to a browser to consult the online docs
  2. Live diff — see which values are currently overridden vs defaults alongside the reference itself — something static (online) docs cannot provide

**Goal**

  Add a hermeto `config` subcommand that:
  1. Dumps the current effective configuration (merged from defaults, config files, env vars)
  2. Shows corresponding environment variable names for each setting
  3. Highlights values that differ from defaults

_Note:  a dedicated subcommand allows the flexibility to enhance or tune it with more options in the future vs a simple `--config-something` toggle_

**Scope**

  - CLI - new Typer subcommand, output format options (YAML/table)
  - config introspection: Programmatic access to `Config` schema, field metadata, and default values via Pydantic's
  model_fields
  - env var name generation: Reconstruct env var names from pydantic's `Config` model representation
  - diff logic: compare `get_config()` against defaults to identify user-modified values

24aysh:
Closes #1253

## Overview

Pretty print the config

## Changes

- Added `config` command for quick lookup of ENV variables 
- Outputs `config` in YAML or JSON
- --diff flag to only display the overidden env vars 

## Demonstration

```bash
$ hermeto git:cmd-config ❯ HERMETO_RUNTIME__CONCURRENCY_LIMIT=10 HERMETO_GOMOD__DOWNLOAD_MAX_TRIES=12 hermeto config --diff
{
  "runtime": {
    "concurrency_limit": 10
  },
  "gomod": {
    "download_max_tries": 12
  }
}
```

```bash
$ hermeto git:cmd-config ❯ hermeto config --format yaml        
gomod:
  download_max_tries: 5
  environment_variables: {}
  proxy_url: https://proxy.golang.org,direct
http:
  connect_timeout: 30
  read_timeout: 300
npm:
  proxy_login: null
  proxy_password: null
  proxy_url: null
pip:
  ignore_dependencies_crates: false
runtime:
  concurrency_limit: 5
  subprocess_timeout: 3600
yarn:
  enabled: true 
```

```bash
$ hermeto git:cmd-config ❯ HERMETO_RUNTIME__CONCURRENCY_LIMIT=10 HERMETO_GOMOD__DOWNLOAD_MAX_TRIES=12 hermeto config --format json
{
  "pip": {
    "ignore_dependencies_crates": false
  },
  "yarn": {
    "enabled": true
  },
  "npm": {
    "proxy_url": null,
    "proxy_login": null,
    "proxy_password": null
  },
  "gomod": {
    "proxy_url": "https://proxy.golang.org,direct",
    "download_max_tries": "12 (*)",
    "environment_variables": {}
  },
  "http": {
    "connect_timeout": 30,
    "read_timeout": 300
  },
  "runtime": {
    "subprocess_timeout": 3600,
    "concurrency_limit": "10 (*)"
  }
}
```

### DIFF
--- a/hermeto/interface/cli.py
+++ b/hermeto/interface/cli.py
@@ -110,6 +110,13 @@ class SBOMFormat(str, enum.Enum):
     spdx = "spdx"
 
 
+class OutputFormat(str, enum.Enum):
+    """Output format for the config command."""
+
+    json = "json"
+    yaml = "yaml"
+
+
 SBOM_TYPE_OPTION = typer.Option(
     SBOMFormat.cyclonedx,
     "--sbom-output-type",
@@ -244,10 +251,10 @@ def list_backends() -> None:
 @app.command()
 @handle_errors
 def config(
-    format: str = typer.Option(
-        "json",
+    format: OutputFormat = typer.Option(
+        OutputFormat.json,
         "--format",
-        help="Default: JSON, Supported: JSON, YAML",
+        help="Output format.",
     ),
     diff: bool = typer.Option(
         False,
@@ -296,17 +303,24 @@ def config(
                 result[field_name] = f"{masked} (*)" if field_name in overridden else masked
         return result
 
+    def _postprocess(s: str) -> str:
+        """Format (*) markers in JSON and YAML output."""
+        import re
+        s = re.sub(r' \(\*\)"', '"   (*)', s)       # JSON double-quoted strings
+        s = re.sub(r"'(.*?) \(\*\)'", r"'\1'   (*)", s)  # YAML single-quoted strings
+        return s
+
     cfg = _get_diff(get_config()) if diff else _get_annotated_config(get_config())
 
+    if diff and not cfg:
+        print("No overrides — all values are at their defaults.")
+        return
+
     dumpers = {
-        "json": lambda data: json.dumps(data, indent=2),
-        "yaml": yaml.dump,
+        OutputFormat.json: lambda d: _postprocess(json.dumps(d, indent=2, default=str)),
+        OutputFormat.yaml: lambda d: _postprocess(yaml.dump(d)),
     }
-    dumper = dumpers.get(format)
-    if dumper:
-        print(dumper(cfg))
-    else:
-        raise typer.BadParameter(f"Invalid format: {format}. Supported: {', '.join(dumpers.keys())}")
+    print(dumpers[format](cfg))
 
 @app.command(help=FETCH_DEPS_HELP)
 @handle_errors


24aysh:
@eskultety  ptal in this
additionally, do i need to add ENV VAR names too like you showed in your issue description

hermeto/interface/cli.py
```diff
@@ -239,6 +241,57 @@ def list_backends() -> None:
...
+    if diff and not cfg:
+        print("No overrides — all values are at their defaults.")
+        return
+
+    if format == "yaml":
+        print(yaml.dump(cfg))
+    elif format == "json":
+        print(json.dumps(cfg,indent=2))
+    else:
+        raise typer.BadParameter(f"Invalid format: {format}. Supported: json, yaml")
```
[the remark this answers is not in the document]
a-ovchinnikov: This is a very poor advice, there is a better mechanism for that.

hermeto/interface/cli.py
```diff
@@ -239,6 +241,57 @@ def list_backends() -> None:
...
+            iterateIn = model.model_fields_set
+        else:
+            iterateIn = model.model_fields
+            
+        for field_name in iterateIn:
+            value = getattr(model, field_name)
+            if isinstance(value, pydantic.BaseModel):
+                result[field_name] = _annotate_overrides(value, diff)
+            else:
+                result[field_name] = value if diff else (f"{value} (*)" if field_name in overridden else value)
```
a-ovchinnikov: This embeds `(*)` into a value, from a cursory glance at the example provided in the issue it is not supposed to do that.

hermeto/interface/cli.py
```diff
@@ -239,6 +241,57 @@ def list_backends() -> None:
     if experimental:
         print("Experimental:", ", ".join(experimental))
 
+@app.command()
+@handle_errors
+def config(
+    format: str = typer.Option(
```
a-ovchinnikov: How would you characterize `str` type? Are you familiar with any techniques for making it more suitable for specifying a finite set of options?

hermeto/interface/cli.py
```diff
@@ -257,6 +257,19 @@ def config(
 ) -> None:
     """Show current configuration. Values overridden from defaults are marked with (*)."""
 
+    _SENSITIVE_FIELDS = {
+        "proxy_password",
+        "proxy_login",
+    }
+
+    def _mask(field_name: str, value: object) -> object:
```
a-ovchinnikov, line 272: Please describe a usage scenario when masking is necessary.

a-ovchinnikov (commented):
A few drive-by comment.

hermeto/interface/cli.py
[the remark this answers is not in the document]
24aysh: ohh, i think we can make outformat class and implement based on it
high level overview
```python
  class OutputFormat(str, Enum):
        json = "json"
        yaml = "yaml"
   ...
   dumpers = {
          OutputFormat.json: lambda d: json.dumps(d, indent=2),
          OutputFormat.yaml: yaml.dump,
      }
      
      print(dumpers[format](cfg))
  ```
  
  should i make this change @a-ovchinnikov ?

File: hermeto/interface/cli.py (hunk 2 above)
Comment (24aysh) [reply hermeto/interface/cli.py]: yeah that's the current issue, but fixing would clutter the code, and if we  want that, then  we should create a new helper.py with the actual function implementation, and would keep the actual config cmd clean
Shall i proceed?  @a-ovchinnikov

File: hermeto/interface/cli.py (hunk 3 above)
Comment (24aysh) [reply hermeto/interface/cli.py]: so it would be resolved with https://github.com/hermetoproject/hermeto/pull/1356#discussion_r2888234915, just like SBOM and log level, 
top level overview
```python
def config(
    format: OutputFormat = typer.Option(
        OutputFormat.json,
        "--format",
        help="Output format.",
    ),
) -> None:
```

24aysh:
@a-ovchinnikov i need approval of my plan on the issues that you pointed, ptal

hermeto/interface/cli.py
a-ovchinnikov: While you did not answer any of the questions directly you provided a code snippet that shows a better approach. Please apply it.

hermeto/interface/cli.py
a-ovchinnikov: WDYT about the reason _why_ the example is provided the way it is provided?

24aysh:
> Could you please ensure that any refactors _within_ one sequence of commits are moved into a single commit introducing a change?

i'll squash the commits when you fully approve the pr, as it's difficult for me to have a fully acceptable pr in one run

a-ovchinnikov:
> i'll squash the commits

Please do not squash commits, only those parts which get modified more than once within the chain.

> difficult for me to have a fully acceptable pr in one run

Sure, just keep in mind that you will be asked to restructure your commits to ensure that changes are logically isolated.

hermeto/interface/cli.py
24aysh: I can't think of an immediate use-case in this repo's own CI (this repo workflows don't use proxy credentials), but the fields are real and setable `HERMETO_NPM__PROXY_LOGIN` and `HERMETO_NPM__PROXY_PASSWORD` are accepted env vars with a validator enforcing they're both set together. And anyone running hermeto against a corporate Artifactory proxy would set these, and then hermeto config would expose them in plain text in logs,so  masking is low-effort and defensive — preventative rather than reactive. so there's no harm considering these things.

However i can change it if you ask me to..

a-ovchinnikov: I am sorry, how is this response supposed to make any sense? What does "but the fields are real and settable" even mean? I introduced these fields myself just four weeks ago, are you implying that I do not remember what I did in the project? Furthermore, anyone capable of running `hermeto config describe` in an environment where these variables are set will also be capable of running `env`. Last, but not least: which logs are you talking about? This data goes straight to stdout, I double-checked your submission, I cannot see any mentions of logging this data there. Did I miss something and if yes could you please point me to a line that introduces logging of this specific data?

hermeto/interface/cli.py
24aysh: just to provide a way to check the overridden vars? honestly
i'll just update it to the way it was supposed to be in the issue example. If it works for you, we can keep it , otherwise i'll try to find a work around for it

24aysh:
i assisted with claude, the results are what was supposed to be in the example

@a-ovchinnikov  ready to review  (i'll rewrite the commit history in the end)

```bash
$ hermeto git:cmd-config ❯ HERMETO_NPM__PROXY_URL=https://myproxy.example.com \                                                                                                   
HERMETO_NPM__PROXY_LOGIN=myuser \
HERMETO_NPM__PROXY_PASSWORD=mysecret \
hermeto config --format json       
{
  "pip": {
    "ignore_dependencies_crates": false
  },
  "yarn": {
    "enabled": true
  },
  "npm": {
    "proxy_url": "https://myproxy.example.com/"   (*),
    "proxy_login": "***"   (*),
    "proxy_password": "***"   (*)
  },
  "gomod": {
    "proxy_url": "https://proxy.golang.org,direct",
    "download_max_tries": 5,
    "environment_variables": {}
  },
  "http": {
    "connect_timeout": 30,
    "read_timeout": 300
  },
  "runtime": {
    "subprocess_timeout": 3600,
    "concurrency_limit": 5
  }
}
```

```bash
$ hermeto git:cmd-config ❯ HERMETO_NPM__PROXY_URL=https://myproxy.example.com \                                                                                                   
HERMETO_NPM__PROXY_LOGIN=myuser \
HERMETO_NPM__PROXY_PASSWORD=mysecret \
hermeto config --format yaml
gomod:
  download_max_tries: 5
  environment_variables: {}
  proxy_url: https://proxy.golang.org,direct
http:
  connect_timeout: 30
  read_timeout: 300
npm:
  proxy_login: '***'   (*)
  proxy_password: '***'   (*)
  proxy_url: https://myproxy.example.com/ (*)
pip:
  ignore_dependencies_crates: false
runtime:
  concurrency_limit: 5
  subprocess_timeout: 3600
yarn:
  enabled: true
  ```

File: hermeto/interface/cli.py [hunk 5]
```diff
@@ -239,6 +241,57 @@ def list_backends() -> None:
     if experimental:
         print("Experimental:", ", ".join(experimental))
 
+@app.command()
+@handle_errors
+def config(
+    format: str = typer.Option(
+        "json",
```
a-ovchinnikov, line 248: What is the point of havin `json` format, especially as the default one?

hermeto/interface/cli.py
```diff
@@ -296,17 +303,24 @@ def _get_annotated_config(model: pydantic.BaseModel) -> dict:
                 result[field_name] = f"{masked} (*)" if field_name in overridden else masked
         return result
 
+    def _postprocess(s: str) -> str:
```
a-ovchinnikov, line 306: Could you please explain what is this function and why you deem it necessary?

a-ovchinnikov:
> i assisted with claude

Yes, that could be clearly seen from the contribution, thank you for confirming.

> the results are what was supposed to be in the example

I am sorry, but no, there is a quite big difference between an example in the issue and in a comment above.

hermeto/interface/cli.py
a-ovchinnikov: > i'll just update it to the way it was supposed to be in the issue example

This reads a lot like you have neither an understanding of what is proposed to be done in the issue nor an interest in gaining one. I hope that it is just poor wording because otherwise it is a terrible attitude for any opensource project, not just Hermeto. May I ask you why you decided to start contributing to Hermeto?

> otherwise i'll try to find a work around for it

This part just doesn't make any sense. I did not ask about workarounds, I asked about your thought process and about why _you think_ certain things are done the way they are done. Understanding the reasons behind changes and requests for features is of extreme importance, this is not a competition to churn as many lines of code as one could after all.

24aysh:
> I am sorry, but no, there is a quite big difference between an example in the issue and in a comment above.

can you specifically tell me which diff are you talking about other than #ENV_VAR names? i didn't added it because i was waiting for approval from  [eskultety](https://github.com/eskultety) in https://github.com/hermetoproject/hermeto/pull/1356#issuecomment-4000790284

the other thing you might be talking about that --diff doesn't show the default values of the vars, that is need to be added

hermeto/interface/cli.py
24aysh: `(*)` was getting embedded in the value, like "<changed val>" was showed  like `"<changed val>  (*)"`,  specially in the json

hermeto/interface/cli.py
24aysh: i chose it default arbitrarily, and added it from my side alongside yaml as i thought it would be good to keep json alonside yaml

hermeto/interface/cli.py
24aysh: sorry, i  apologize for imprecise wording in my justification. i added masking for general caution and pointed out the term logging but i think it messed up my logic, it was meant to be preventive
Apologies from my side

hermeto/interface/cli.py
24aysh: i was interested to learn about debian package ecosystem, and also gsoc, and found this issue a way to contribute to hermeto.
Apologies if i couldn't make a meaningful contribution,  and also now i don't think you would allow me to further work on this.

a-ovchinnikov: Thank you, please do not be alarmed and bear with me a bit longer. What do you think GSoC is?

24aysh: i see gsoc as a way to learn about different techs that one might not be able to see as an individual learner (ofc stipend is one of the driving factor, i won't lie about it). Over the past few months, i learned k8s, docker, golang and now about hermeto. I was already contributing to kgateway , you can see [here](https://github.com/kgateway-dev/kgateway/pulls?q=is%3Apr+is%3Aclosed+author%3A24aysh). But recently  i got to know about hermeto project (deb/ubuntu package ecosystem) and i'm personally using linux from past 2 years so i was more interested in it than in my previous org. 
I apologize to  you again for my behaviour.

a-ovchinnikov: Thank you. The most important aspect of GSoC from participating organisations (and theoretically students as well) is the learning opportunity it provides. Producing code is a simple activity, understanding why it is necessary, what problems it exists to deal with is much harder, and understanding these is crucial for producing good code, manually or with a help of a LLM. This requires learning to ask why something is important, what problems does some tool solve and what problems arise from any particular solution. This is why I ask so many questions about reasons behind certain changes: code which is produced without thinking some problem through is very easy to spot and it does not help any project, quite the contrary. If it were not the case maintainers would have already closed all issues by unleashing AI agents on them on their own. Understanding what is needed and why is what most maintainers in most projects look for and what is crucial for success of any non-trivial fix. That's why I keep asking whether there is any special meaning in provided examples. To answer this question you would need to carefully dissect an example and think what would happen if some parts were removed from it (or added).

Last, but not least, and I'll stress this out again, no one expects perfection from GSoC student's contributions right away (or any new contributor for that matter). It is impossible to arrive into any serious project and hit a perfect score from the first try unless one has very serious experience.

Let's get back to the original question: why is the example provided the way it is provided? What could be modified in it to make it look more like the output your fix provides and what that would change in the meaning of the example?

24aysh: yeah thanks, i thought i lost it, heading to original question
the way it is provided to show the current config and highlight what's been changed from defaults.
In the example it shows env vars too (my implementation is lacking this as of now), and that is helpful for letting someone which env var to set 
All of this at one place was the purpose

a-ovchinnikov: Let me show you what I meant by asking questions about what you see.

You look at an example and see some strings. They have some structure. What is the structure? It looks like yaml. Is it a requirement to be a yaml? No, it is not (the example is labeled as *not binding*). Why is it a yaml then? Likely because it is compact and expressive. What does tha mean? It means that it is easy for humans to read. Who else can read yaml easily? Machines. Does that mean that a correct yaml document should be interprettable by both humans and machines? Yes. Does that mean that a yaml could be copied from stdout to a file and be processed by a program? Yes. What could be seen on the example? Config values are printed to stdout. Are they printed randomly? No, they are printed in order. What is the order? They are structured to follow Config object's structure. Are there any peculiarities in values representation? Yes, some values are marked with stars to indicate changes from defaults. Are stars just added to lines in the output? Not exactly, they are enclosed by parentheses and prepended by a sharp sign. Does that make any difference? I don't know. How could you learn that? I could run a test. Please do.

```python3
>>> import yaml                                                                 
>>> from pathlib import Path                                                    
>>> one = Path("one.yml")                                                       
>>> one.read_text()                                                             
'---\nfoo:\n  bar: "http://www.example.com"\n'                                  
>>> with open(one) as f: yaml.safe_load(f)                                      
...                                                                             
{'foo': {'bar': 'http://www.example.com'}}                                      
>>> two = Path("two.yml")                                                       
>>> two.read_text()                                                             
'---\nfoo:\n  bar: "http://www.example.com"  # (*)\n'                           
>>> with open(two) as f: yaml.safe_load(f)                                      
...                                                                             
{'foo': {'bar': 'http://www.example.com'}}                                      
>>> three = Path("three.yml")                                                   
>>> three.read_text()                                                           
'---\nfoo:\n  bar: "http://www.example.com"  (*)\n'                             
>>> with open(three) as f: yaml.safe_load(f)
... 
  File "<python-input-10>", line 1, in <module>
    with open(three) as f: yaml.safe_load(f)
                           ~~~~~~~~~~~~~~^^^
...
yaml.parser.ParserError: while parsing a block mapping
  in "three.yml", line 3, column 3
expected <block end>, but found '<scalar>'
  in "three.yml", line 3, column 34
>>>
```
Apparently sharp sign is invisible to a parser and also makes everything after it invisible as well. When it is present a human can easily detect a line in a config that was modified, but a program cannot; such config could be just copied over. Leaving it out, however, breaks the parser; a config without sharp signs in front of stars cannot be just copied over and must always be processed before being submitted to Hermeto.

Producing some code is easy, understanding how to come up with a chain of questions like above is the important part. Please note, that all my other questions in this PR require the same approach to arrive to an answer.

PR #1362, opened 2026-03-06
update test data for gomod/with-deps test migration

derasdf:
This should be merged with the [PR](https://github.com/hermetoproject/integration-tests/pull/38) merge in integration tests

### DIFF
--- a/tests/integration/test_data/gomod_with_deps/bom.json
+++ b/tests/integration/test_data/gomod_with_deps/bom.json
@@ -11,8 +11,8 @@
         "pkg:golang/cmp?type=package",
         "pkg:golang/errors?type=package",
         "pkg:golang/fmt?type=package",
-        "pkg:golang/github.com/hermetoproject/integration-tests@v0.0.0-20201001160613-4c65d49cae6b?type=module",
-        "pkg:golang/github.com/hermetoproject/integration-tests@v0.0.0-20201001160613-4c65d49cae6b?type=package",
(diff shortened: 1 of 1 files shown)


tests/integration/test_data/gomod_with_deps/bom.json
```diff
@@ -11,8 +11,8 @@
         "pkg:golang/cmp?type=package",
         "pkg:golang/errors?type=package",
         "pkg:golang/fmt?type=package",
-        "pkg:golang/github.com/hermetoproject/integration-tests@v0.0.0-20201001160613-4c65d49cae6b?type=module",
-        "pkg:golang/github.com/hermetoproject/integration-tests@v0.0.0-20201001160613-4c65d49cae6b?type=package",
+        "pkg:golang/github.com/hermetoproject/integration-tests@v0.0.0-20260305223514-980620988bdf?type=module",
```
a-ovchinnikov: This is the current revert, shouldn't this be pointing to revert of revert?

derasdf: I think you are right, but then it would be breaking everything right now anyway, so should i merge right now first the [pull/38](https://github.com/hermetoproject/integration-tests/pull/38) and then fix this ?

a-ovchinnikov: I think that if your revert my revert and merge it, then regenerate this SBOM and push the change everything will be back to normal. If you merge it right now it won't hurt, but would create an excessive commit.

derasdf: Should be mergable now after the checks

PR #1370, opened 2026-03-07
tests: remove check_deps_checksums from TestParameters

### LINKED ISSUE #959: Revisit attributes of TestParameters class for integration tests
Revisit and investigate some attributes of the `TestParameters` class inside [hermeto/tests/integration/utils.py](https://github.com/hermetoproject/hermeto/blob/main/tests/integration/utils.py).

For example:
1. One of the class attributes is `check_vendor_checksums: bool`, but the actual usage of it is debatable. As a result, this attribute creates bunch of empty `vendor_sha256sums.json` files.

```bash
$ cd ~/hermeto
$ git grep "check_vendor_checksums=True" | wc -l
0
$ git grep "check_vendor_checksums=False" | wc -l
75
```

2. Another class attribute is `check_deps_checksums: bool`. Now it is 50:50, but still far from a consistency. The result of this attribute are `fetch_deps_sha256sums.json` inside directory for integration tests that has set it to True.

```bash
$ cd ~/hermeto
$ git grep "check_deps_checksums=True" | wc -l
2
$ git grep "check_deps_checksums=False" | wc -l
37
```

Let's decide on:
- Are they actually still needed or are they needed for only certain package managers ?
- What should be the default value ?
- How to keep it consistent ?

A/C
Document the attributes in the class.

sammy200-ui:
Remove `check_deps_checksums` from `TestParameters` and replace with auto-detection
 
The issue had three questions about `TestParameters` attributes: are they still needed, what should the defaults be, and how to keep things consistent.

check_vendor_checksums was already gone

`check_deps_checksums` was still present but the pattern was clear: it was set to False for every test that expected a non-zero exit code (nothing to verify) and for every package manager whose output is non-deterministic (bundler, cargo). The True value appeared only twice, and the attribute default was also True  
The attribute wasn't documenting anything meaningful

What changed

Removed check_deps_checksums from the dataclass entirely. The runner now determines whether to verify checksums automatically:

If `expected_exit_code != 0`,skip (no deps were fetched)
Otherwise , look for `fetch_deps_sha256sums.json` in the test-data directory; verify if it exists, skip if it doesn't
This means non-deterministic package managers simply don't have a checksum file committed. No attribute needed.
 
Resolves: #959


sammy200-ui:
umm.... the AI-agents left some comments on it, and i would like some help on these reviews - so the first co-pilot review suggested documenting the internal special cases (gitcommit  tarballs, sumdb entries etc.) in the TestParameters docstring. My thinking is that - is this belongs in the function's own docstring, not here TestParameters only needs to tell you `when` to use True/False, not `how` the checksums are internally computed. 

the last review - Copilot said that `check_output=False` isn't only for failing tests. initially i thought that then after verifying again it might be correct   for example `rpm/repo-file` has `exit_code=0` but still sets `check_output=False`. Same for `yarn-classic/corepack-ignored` and `yarn-classic/yarn-path-ignored`.

so i was thinking to word it like this - 
`Set to False when output file comparison is intentionally skipped, either because the fetch is expected to fail and no output is produced, or because the test intentionally skips output verification.`

a-ovchinnikov (commented):
This PR does not resolve the issue. The issue is about figuring out whether these parameters are needed at all and likely getting rid of them, adding a docstring with Claude-generated summary won't help.

eskultety:
> the last review - Copilot said that check_output=False isn't only for failing tests. initially i thought that then after

@sammy200-ui may I ask how you got Copilot to perform a review on your PR? Did you select it from the list of reviewers on your own?

sammy200-ui:
Hello @a-ovchinnikov, you're completely right. I leaned too hard on AI for those docstrings and totally missed the actual part of the issue. My bad on that!

did recheck and found that `check_vendor_checksums ` is not being used anywhere anymore anymore , 
the `check_deps_checksums`is something that i am still in the middle of reviewing. So far, I have noticed the False cases mostly fall into two buckets- tests expecting a non-zero exit code (nothing gets fetched), and non-deterministic package managers (like cargo and bundler).

I haven't fully figured out the safest way to refactor this second attribute yet without breaking how the tests read the .json files. Let me finish up with these , then i will update here

sammy200-ui:
Hello @eskultety , i am not sure how co-pilot was able to review my pr , i am sure on that it was not something done "manually" by me , maybe i have the github pro thats why this feature exists?? , but i am not sure fully

sammy200-ui:
hello @a-ovchinnikov , following up on the pr , the pr has been updated since the last review , would really appreciate  review once again

eskultety:
Closing in favour of https://github.com/hermetoproject/hermeto/pull/1557

Issue #1372, opened 2026-03-09
bug(gomod): Snap-installed Go breaks toolchain detection due to symlink resolution


farhann-saleem:
### Description

I was testing hermeto's gomod backend against a couple of Go projects on my Ubuntu machine and ran into a failure. Go is installed via snap on my system and works fine from the command line, but hermeto's toolchain detection resolves the snap symlink, which breaks the Go binary path. Screenshots are attached below:

### Reproduction

**Environment:**
- Ubuntu 24.04 (with snap)
- Go installed via snap (`snap install go --classic`)
- Go version: 1.25.7
- Python: 3.12

Go works correctly on the system:
```bash
$ go version
go version go1.25.7 linux/amd64

$ which go
/snap/bin/go

$ go env GOVERSION
go1.25.7
```

**Test 1  [junegunn/fzf](https://github.com/junegunn/fzf):**
```bash
$ git clone https://github.com/junegunn/fzf
$ hermeto fetch-deps --source ./fzf --output /tmp/fzf-out gomod

2026-03-09 14:24:52,423 INFO Fetching the gomod dependencies at subpath .
2026-03-09 14:24:52,479 ERROR The command "/usr/bin/snap env GOVERSION" failed
2026-03-09 14:24:52,479 ERROR STDERR:
error: unknown command "env", see 'snap help'.
2026-03-09 14:24:52,485 ERROR PackageManagerError: Go execution failed: `/usr/bin/snap env GOVERSION` failed with rc=64
```

**Test 2  [hashicorp/terraform](https://github.com/hashicorp/terraform):**
```bash
$ git clone https://github.com/hashicorp/terraform
$ hermeto fetch-deps --source ./terraform --output /tmp/terraform-out gomod

2026-03-09 14:32:35,318 INFO Fetching the gomod dependencies at subpath .
2026-03-09 14:32:35,377 ERROR The command "/usr/bin/snap env GOVERSION" failed
2026-03-09 14:32:35,377 ERROR STDERR:
error: unknown command "env", see 'snap help'.
2026-03-09 14:32:35,601 ERROR PackageManagerError: Go execution failed: `/usr/bin/snap env GOVERSION` failed with rc=64
```

Both projects fail with the same error, confirming the issue is in hermeto's toolchain detection, not in any specific repository.

<img width="1798" height="623" alt="Image" src="https://github.com/user-attachments/assets/8a0860c6-f1fe-420b-ac18-939c3f8f2b01" />

[...]

PR #1373, opened 2026-03-09
fix(gomod): preserve symlink names in toolchain detection

farhann-saleem:
Description:                                                                                                                                   
  Resolves: #1372                                  
                                                                                                                                                 
  ### Problem                              
                                                                                                                                                 
  `_get_available_go_toolchains()` uses `Path.resolve()` on Go binary paths to  filter out duplicates. On systems where Go is installed via snap, this resolves  `/snap/bin/go` → `/usr/bin/snap`, since snap binaries are symlinks to the snap                                                                 
  dispatcher.                                                                                                                                    
                                                                                                                                                 
  Snap determines which application to run based on `argv[0]` (the symlink name).                                                                
  Once resolved, the binary name is lost and hermeto ends up executing                                                                           
  `/usr/bin/snap env GOVERSION`, which snap does not recognize as a valid command.                                                               
                                                                                                                                                 
  ### Fix

  Resolved paths are now used only as deduplication keys. The original (unresolved)  path is preserved and passed to the `Go` constructor, so snap dispatch continues  to work correctly.

  ### Testing

<img width="1765" height="311" alt="Screenshot from 2026-03-09 15-19-49" src="https://github.com/user-attachments/assets/fc35886b-10ad-462e-a8c9-896b6dfdf798" />

  - Verified the fix locally on Ubuntu with snap-installed Go (1.25.7)
  - Tested against [junegunn/fzf](https://github.com/junegunn/fzf) and  [hashicorp/terraform](https://github.com/hashicorp/terraform) both complete successfully after the change
  - All 1391 unit tests pass, including the existing `filter_symlinked_path`    deduplication test
  - Linter (ruff + mypy) passes

### DIFF
--- a/hermeto/core/package_managers/gomod.py
+++ b/hermeto/core/package_managers/gomod.py
@@ -1809,9 +1809,16 @@ def _list_installed_toolchains() -> set[Go]:
     for path in (HERMETO_GO_INSTALL_DIR, get_cache_dir()):
         paths |= {p.resolve().parent for p in Path(path).rglob("bin/go")}
 
-    # filter out symlinked paths
-    go_binary_paths = set([Path(p, "go").resolve() for p in paths if Path(p, "go").exists()])
-    for bin_path in go_binary_paths:
+    # Resolve paths for deduplication only; preserve original symlink names for the binary.
+    # Tools like snap rely on argv[0] for dispatch, resolving /snap/bin/go into /usr/bin/snap
+    # which breaks Go execution.
+    go_binary_paths: dict[Path, Path] = {}
+    for p in paths:
+        bin_path = Path(p, "go")
+        if bin_path.exists():
+            go_binary_paths.setdefault(bin_path.resolve(), bin_path)
+
+    for bin_path in go_binary_paths.values():
         try:
             log.debug("Probing %s toolchain...", bin_path)
             ret.add(Go(binary=bin_path.as_posix()))


hermeto/core/package_managers/gomod.py
```diff
@@ -1809,9 +1809,18 @@ def _list_installed_toolchains() -> set[Go]:
...
+    # Resolve paths for deduplication only; preserve original symlink names for the binary.
+    # Tools like snap rely on argv[0] for dispatch resolving /snap/bin/go into /usr/bin/snap
+    # breaks Go execution.
+    seen_resolved: dict[Path, Path] = {}
+    for p in paths:
+        bin_path = Path(p, "go")
+        if bin_path.exists():
+            resolved = bin_path.resolve()
+            if resolved not in seen_resolved:
+                seen_resolved[resolved] = bin_path
```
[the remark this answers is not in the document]
eskultety: Agreed.

hermeto/core/package_managers/gomod.py
```diff
@@ -1809,9 +1809,18 @@ def _list_installed_toolchains() -> set[Go]:
     for path in (HERMETO_GO_INSTALL_DIR, get_cache_dir()):
         paths |= {p.resolve().parent for p in Path(path).rglob("bin/go")}
 
-    # filter out symlinked paths
-    go_binary_paths = set([Path(p, "go").resolve() for p in paths if Path(p, "go").exists()])
-    for bin_path in go_binary_paths:
+    # Resolve paths for deduplication only; preserve original symlink names for the binary.
+    # Tools like snap rely on argv[0] for dispatch resolving /snap/bin/go into /usr/bin/snap
+    # breaks Go execution.
+    seen_resolved: dict[Path, Path] = {}
```
eskultety: nitpick: I would keep the old name `go_binary_paths` as it's more descriptive than `seen_resolved` when you actually loop over it a few lines below.

hermeto/core/package_managers/gomod.py
```diff
@@ -1809,9 +1809,18 @@ def _list_installed_toolchains() -> set[Go]:
     for path in (HERMETO_GO_INSTALL_DIR, get_cache_dir()):
         paths |= {p.resolve().parent for p in Path(path).rglob("bin/go")}
 
-    # filter out symlinked paths
```
eskultety, line 1812: Commit message:
you want to mention the issue in the commit as well.

farhann-saleem: setdefault Done
go_binary_paths rename  Done
 Commit message   Done

hermeto/core/package_managers/gomod.py
```diff
@@ -1809,9 +1809,16 @@ def _list_installed_toolchains() -> set[Go]:
...
-    go_binary_paths = set([Path(p, "go").resolve() for p in paths if Path(p, "go").exists()])
-    for bin_path in go_binary_paths:
+    # Resolve paths for deduplication only; preserve original symlink names for the binary.
+    # Tools like snap rely on argv[0] for dispatch resolving /snap/bin/go into /usr/bin/snap
+    # breaks Go execution.
+    go_binary_paths: dict[Path, Path] = {}
+    for p in paths:
+        bin_path = Path(p, "go")
+        if bin_path.exists():
+            go_binary_paths.setdefault(bin_path.resolve(), bin_path)
```
a-ovchinnikov, line 1819: Do I read this right: if multiple links exist to a single filesystem object then only the first one will be preserved?

hermeto/core/package_managers/gomod.py
```diff
@@ -1809,9 +1809,16 @@ def _list_installed_toolchains() -> set[Go]:
     for path in (HERMETO_GO_INSTALL_DIR, get_cache_dir()):
         paths |= {p.resolve().parent for p in Path(path).rglob("bin/go")}
 
-    # filter out symlinked paths
-    go_binary_paths = set([Path(p, "go").resolve() for p in paths if Path(p, "go").exists()])
-    for bin_path in go_binary_paths:
+    # Resolve paths for deduplication only; preserve original symlink names for the binary.
+    # Tools like snap rely on argv[0] for dispatch resolving /snap/bin/go into /usr/bin/snap
+    # breaks Go execution.
```
a-ovchinnikov: A minor nitpick: it feels like a word is missing, did you mean "tools like snap rely on argv[0] for dispatch resolving /snap/bin/go into /usr/bin/snap which breaks Go execution."?

hermeto/core/package_managers/gomod.py
a-ovchinnikov, line 1: _Commit message_: could you please add problem description to the commit message too? Problem description will end up separated from the repository otherwise.

a-ovchinnikov (commented):
LGTM in general, a question and a couple of small nitpickish requests are inline.

hermeto/core/package_managers/gomod.py
eskultety: Yes.

hermeto/core/package_managers/gomod.py
farhann-saleem: setdefault question: Yes, the first path encountered for given resolved target is kept.
missing word: Done
commit message: Done

eskultety: > Commit message: could you please add problem description to the commit message too? Problem description will end up separated from the repository otherwise.

Does it? I mean yeah, you have to open a browser, but I personally tend to be lenient on context in commit messages if the linked issue description is satisfactory and is well described. In this case it is very well described.
That said, @farhann-saleem I noticed you only mentioned the issue number in the commit message, please always use full URLs when referencing issues in commit messages - you can click on the link locally without having to search for the issue manually and GH will render it correctly -> win-win.

farhann-saleem: done  . added proper link referencing the issue .

a-ovchinnikov: > Does it?

Yes, with a (shortened) description in a commit message the repository remains self-contained and will remain self contained in the case of migration from Github. It is not that I am expecting such migration to happen, but I think this particular preparation could be achieved at no cost.

a-ovchinnikov (approved):

PR #1374, opened 2026-03-09
fix: raise exception on pip checksum verification failure

Sanchit2662:
## Summary

I discovered a critical silent failure bug in pip dependency resolution. When a downloaded artifact fails checksum verification, the file gets deleted but the build succeeds anyway , the dependency is still recorded in the SBOM as if nothing went wrong.

This is a supply chain security issue. Here's what happens:

1. Checksum verification fails on a downloaded artifact
2. The file is deleted and only a WARNING is logged
3. No exception is raised , execution continues normally
4. Build reports success (exit code 0) with missing artifacts
5. SBOM lists the dependency as present, but `deps/pip/` is empty

When someone later tries to use these fetched dependencies, they get a confusing "file not found" error disconnected from the real cause.

The root cause is in `_checksum_must_match_or_path_unlink` , it catches `PackageRejected` but swallows it entirely.

## Fix

Added one line to re-raise after cleanup:

```python
except PackageRejected:
    path.unlink()
    log.warning("Download '%s' was removed", path.name)
    raise  # fail the build on checksum mismatch
```

Now corrupted artifacts are cleaned up AND the build fails immediately. CI pipelines will correctly catch integrity issues instead of green-lighting broken builds.

Also updated two unit tests that were asserting the old broken behavior , they now expect exceptions to be raised on checksum failures.

Verification: All 1376 tests passing, DCO compliant, minimal 1-line fix.

### DIFF
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -275,6 +275,7 @@ def _process_req(
         except PackageRejected:
             path.unlink()
             log.warning("Download '%s' was removed from the output directory", path.name)
+            raise
 
     if dpi:
         if dpi.req_file_checksums:
--- a/tests/unit/package_managers/pip/test_main.py
+++ b/tests/unit/package_managers/pip/test_main.py
@@ -628,7 +628,6 @@ class TestDownload:
                 None,  # sdist_download
                 None,  # wheel_0_download - checksums OK
                 PackageRejected("", solution=None),  # wheel_1_download - checksums NOK
-                PackageRejected("", solution=None),  # wheel_2_download - no checksums to verify
             ]
         else:
             mock_must_match_any_checksum.side_effect = [
@@ -637,13 +636,21 @@ class TestDownload:
         # </setup>
 
         # <call>
-        found_downloads = pip._download_dependencies(rooted_tmp_path, req_file, binary_filters)
-        assert found_downloads == expected_downloads
+        if binary_filters is not None:
+            # wheel_1 fails checksum verification -> must raise
+            with pytest.raises(PackageRejected):
+                pip._download_dependencies(rooted_tmp_path, req_file, binary_filters)
+        else:
+            found_downloads = pip._download_dependencies(
+                rooted_tmp_path, req_file, binary_filters
+            )
+            assert found_downloads == expected_downloads
         assert pip_deps.path.is_dir()
         # </call>
 
         # <check calls that must always be made>
-        mock_check_metadata_in_sdist.assert_called_once_with(sdist_DPI.path)
+        if binary_filters is None:
+            mock_check_metadata_in_sdist.assert_called_once_with(sdist_DPI.path)
         mock_process_package_distributions.assert_called_once_with(
             req, pip_deps, binary_filters, expect_index_url
         )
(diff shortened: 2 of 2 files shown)


hermeto/core/package_managers/pip/main.py
```diff
@@ -275,6 +275,7 @@ def _checksum_must_match_or_path_unlink(
         except PackageRejected:
             path.unlink()
             log.warning("Download '%s' was removed from the output directory", path.name)
```
a-ovchinnikov, line 277: IIRC this was a design decision, @slimreaper35 might be able to provide some comment on this.

a-ovchinnikov:
@Sanchit2662 thank you for your contribution!

> I discovered a critical silent failure bug in pip dependency resolution.

Please share how exactly did you discover it? Was it some specific package that triggered the problem?

hermeto/core/package_managers/pip/main.py
slimreaper35: I think it was, but I don't remember the reasoning behind it.

eskultety: The original design when it comes to fetching was very simple and straightforward - just fetch **everything**. The user, however, would only really list a subset of acceptable hashes for any of the distribution artifacts corresponding to a dependency version (i.e. a wheel hash or an sdist hash). Pip normally verifies these hashes before installation. If Hermeto (cachi2 back then) fetches incompatible wheels, then pip would never consider them during install. Therefore, the decision was to eliminate all non-matching artifacts from the output directory due to hash checking and continue with other artifacts where at least one should match the expected hashes instead of failing right away which would have been wrong because WE DO the fetching, not the user or pip.

eskultety:
> @Sanchit2662 thank you for your contribution!
> 
> > I discovered a critical silent failure bug in pip dependency resolution.
> 
> Please share how exactly did you discover it? Was it some specific package that triggered the problem?

I'd also be interested in knowing how this was encountered and why we should address it.

> This is a supply chain security issue. Here's what happens:

"supply chain issue" - yes
"security" - definitely not

>     1. Checksum verification fails on a downloaded artifact

This means that hash-checking is ON, yet none of the hashes could be met with any of the archives we fetched meaning that someone tampered with the source repository, that's a user problem, not a supply chain attack, at least not a smart one. If an attacker managed to poison the source repo such that no matching dependency would be fetched because of this, then the subsequent hermetic build would fail. I'm tempted to close this, but I'll wait for a better explanation @Sanchit2662

Sanchit2662:
Hi @a-ovchinnikov ,   Honestly, I found it through code review  not from a production incident. I was reading through the pip fetching logic and the except PackageRejected block in  _checksum_must_match_or_path_unlink caught my eye because it deletes the file but doesn't re-raise. At first glance it looked like a textbook swallowed exception, so I dug deeper to understand the downstream impact.

Sanchit2662:
Thanks for the detailed context @eskultety , I genuinely appreciate you explaining the original design rationale. That clears up a lot.

  You're right about the PyPI case. When Hermeto fetches multiple distribution artifacts for a single dependency (sdists, wheels), removing the ones that don't match and keeping the ones that do makes total sense. The user only listed a subset of hashes, Hermeto fetched everything, and pip will only consider the matching ones during install. I didn't fully appreciate that flow when I opened this that's on me.

  I do want to flag one thing though, and I'd genuinely like your take on it. The same _checksum_must_match_or_path_unlink function is also used in the URL dependency path (req.kind == "url" in _process_req, around line 296-302). For URL deps, there's only a single artifact  no alternative sdist or wheel to fall back to. When that one file fails checksum and gets deleted, _process_req still returns the download_info dict pointing to the now-deleted file. The dependency still shows up in the SBOM, the build exits 0, and the missing file only surfaces later when someone tries pip install --find-links against the output.

  For PyPI I now agree the current behavior is intentional and correct. But for the URL dep path  is that also intentional? It feels like a different situation since there's no pool of artifacts to filter from.

  If you think the URL case is also fine as-is, I'm happy to close this. But if it's worth addressing, I can rework the PR to only raise for URL deps where there's no fallback. Let me know either way.

  And fair point on the "security" framing  that was too strong. "Supply chain correctness issue" would've been more accurate.

PR #1377, opened 2026-03-10
 fix(cargo): wrap generic subprocess errors in PackageManagerError

farhann-saleem:
- When `cargo vendor` fails for reasons other than a lockfile mismatch, the cargo backend re-raises the raw `CalledProcessError`, leaking a  Python traceback to the user
  - Other backends (gomod, yarn, bundler) wrap all subprocess errors in `PackageManagerError` consistently
  - This change applies the same pattern to the cargo backend's fallback error path

  Resolves: https://github.com/hermetoproject/hermeto/issues/1376
  ## Changes

  - Added `PackageManagerError` to the imports in `cargo/main.py`
  - Replaced the bare `raise` with `PackageManagerError` wrapping, including the command, return code, and stderr (same pattern as gomod/yarn)

  ## Testing

  - Tested on [rust-lang/mdBook](https://github.com/rust-lang/mdBook)  before: raw traceback, after: clean `PackageManagerError`

Screenshot from 2026-03-10 15-00-29

  - All cargo unit tests pass (16/16)
  - Linting and type checking clean

### DIFF
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -15,7 +15,13 @@ import tomlkit.exceptions
 from packageurl import PackageURL
 
 from hermeto import APP_NAME
-from hermeto.core.errors import LockfileNotFound, NotAGitRepo, PackageRejected, UnexpectedFormat
+from hermeto.core.errors import (
+    LockfileNotFound,
+    NotAGitRepo,
+    PackageManagerError,
+    PackageRejected,
+    UnexpectedFormat,
+)
 from hermeto.core.models.input import Mode, Request
 from hermeto.core.models.output import Annotation, Component, ProjectFile, RequestOutput
 from hermeto.core.models.sbom import create_backend_annotation, spdx_now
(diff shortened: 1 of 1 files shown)


hermeto/core/package_managers/cargo/main.py
```diff
@@ -15,7 +15,13 @@
 from packageurl import PackageURL
 
 from hermeto import APP_NAME
-from hermeto.core.errors import LockfileNotFound, NotAGitRepo, PackageRejected, UnexpectedFormat
+from hermeto.core.errors import (
```
eskultety, line 18: **Commit message:**
Please try to keep the lines wrapped at ~72-80 characters (72 is the old recommended standard).

farhann-saleem: done

eskultety: Nope, your manual line wrap didn't work at all if you carefully look at it. This is the correct wrap:
```
fix(cargo): wrap generic subprocess errors in PackageManagerError

When cargo vendor fails for reasons other than a lockfile mismatch, the
bare raise leaks a raw CalledProcessError traceback to the user. Other
backends (gomod, yarn, bundler) wrap all subprocess errors in
PackageManagerError consistently. Apply the same pattern to cargo.

Resolves: https://github.com/hermetoproject/hermeto/issues/1376

Signed-off-by: Farhan Saleem <chaudaryfarhann@gmail.com>
```

PR #1378, opened 2026-03-10
Replace dict with Mapping for download files parameter

slimreaper35:


hermeto/core/package_managers/general.py
```diff
@@ -2,6 +2,7 @@
 import asyncio
 import logging
 import ssl
+from collections.abc import Mapping
```
eskultety, line 5: > A caller of this function always had to explicitly annotate the argument
> as `dict[str, StrPath]` to silence mypy.

Wouldn't it be actually better to drop those annotations to prove the point and correctness of the patch? Don't get me wrong the change is harmless, but it'd be nice to have a practical proof of that claim. :)

hermeto/core/package_managers/general.py
a-ovchinnikov, line 1: _Commit message_: Are you sure an abstract class is more precise than a concrete implementation?

hermeto/core/package_managers/general.py
slimreaper35: I can add a temporary commit that reverts the change and removes an annotation.

hermeto/core/package_managers/general.py
slimreaper35: ah, bad wording...

hermeto/core/package_managers/general.py
eskultety: Cool, so given the TEST commit which makes the linters fails, can we use the relevant bits, i.e. squash
```diff
diff --git hermeto/core/package_managers/pip/main.py hermeto/core/package_managers/pip/main.py
index a982f78f3..db0717d60 100644
--- hermeto/core/package_managers/pip/main.py
+++ hermeto/core/package_managers/pip/main.py
@@ -45,7 +45,6 @@ from hermeto.core.package_managers.pip.rust import (
 )
 from hermeto.core.rooted_path import RootedPath
 from hermeto.core.scm import clone_as_tarball, get_repo_id
-from hermeto.core.type_aliases import StrPath

 log = logging.getLogger(__name__)

@@ -319,7 +318,7 @@ def _process_pypi_req(
         req, pip_deps_dir, binary_filters, index_url
     )

-    files: dict[str, StrPath] = {dpi.url: dpi.path for dpi in artifacts if not dpi.path.exists()}
+    files = {dpi.url: dpi.path for dpi in artifacts if not dpi.path.exists()}
     asyncio.run(async_download_files(files, get_config().runtime.concurrency_limit))

     for artifact in artifacts:
```
to the main commit? :)

slimreaper35: done

eskultety (approved):

hermeto/core/package_managers/general.py
```diff
@@ -115,7 +116,7 @@ async def _async_download_binary_file(
 
 
 async def async_download_files(
-    files_to_download: dict[str, StrPath],
+    files_to_download: Mapping[str, StrPath],
```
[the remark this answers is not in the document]
a-ovchinnikov: TBF this does not look like a good advice: `Mapping` is abstract, `dict` is the only concrete type that will bee passed in practice. The change as is exists to placate mypy, I think the update to docstring muddies the water a tiny bit more. (I won't block on this though)

PR #1385, opened 2026-03-11
fix(npm): raise InvalidLockfileFormat on malformed JSON for consistency with yarn

farhann-saleem:
## Summary                                                                                                                                     
  - Align npm backend's JSON error handling with the yarn backend
  - Wrap `json.load()` calls in `PackageLock.from_file()` and `_update_package_json_files()` to raise `InvalidLockfileFormat` instead of leaking 
  raw `json.JSONDecodeError` tracebacks                                                                                                          
  - The yarn backend already handles this consistently ([yarn/project.py                                                                         
  L113-117](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/yarn/project.py#L113-L117))
                                         
  ## Test plan
  - [x] Verified raw `json.JSONDecodeError` traceback with malformed `package-lock.json` before fix
  - [x] Verified clean `InvalidLockfileFormat` error with solution string after fix
  - [x] All 1376 unit tests pass

  Resolves: https://github.com/hermetoproject/hermeto/issues/1384

### DIFF
--- a/hermeto/core/package_managers/npm.py
+++ b/hermeto/core/package_managers/npm.py
@@ -16,6 +16,7 @@ from packageurl import PackageURL
 from hermeto.core.checksum import ChecksumInfo, must_match_any_checksum
 from hermeto.core.config import ProxyUrl, get_config
 from hermeto.core.errors import (
+    InvalidLockfileFormat,
     LockfileNotFound,
     MissingChecksum,
     PackageRejected,
@@ -209,8 +210,7 @@ class PackageLock:
     @classmethod
     def from_file(cls, lockfile_path: RootedPath) -> "PackageLock":
         """Create a PackageLock from a package-lock.json file."""
-        with lockfile_path.path.open("r") as f:
-            lockfile_data = json.load(f)
+        lockfile_data = _load_json_file(lockfile_path.path)
 
         lockfile_version = lockfile_data.get("lockfileVersion")
         if lockfile_version not in (2, 3):
@@ -392,6 +392,20 @@ class _Purlifier:
 
 
 NormalizedUrl = NewType("NormalizedUrl", str)
+
+
+def _load_json_file(file_path: Path) -> Any:
+    """Load and parse a JSON file, raising an appropriate error on decode failure."""
+    try:
+        with file_path.open("r") as f:
+            return json.load(f)
+    except json.JSONDecodeError as e:
+        if "lock" in file_path.name:
+            raise InvalidLockfileFormat(
+                lockfile_path=file_path,
+                err_details=str(e),
+            ) from e
+        raise UnexpectedFormat(f"The {file_path.name} file must contain valid JSON: {e}") from e
 
 
 def _normalize_resolved_url(resolved_url: str) -> NormalizedUrl:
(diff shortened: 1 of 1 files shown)


hermeto/core/package_managers/npm.py
```diff
@@ -209,8 +210,18 @@ def _purlifier(self) -> "_Purlifier":
...
+        try:
+            with lockfile_path.path.open("r") as f:
+                lockfile_data = json.load(f)
+        except json.JSONDecodeError as e:
+            raise InvalidLockfileFormat(
+                lockfile_path=lockfile_path.path,
+                err_details=str(e),
+                solution=(
+                    "The package-lock.json file must contain valid JSON. "
+                    "Refer to the parser error and fix the contents of the file."
```
a-ovchinnikov: This is a bad advice since a broken lock file is not something a user can fix.

hermeto/core/package_managers/npm.py
```diff
@@ -209,8 +210,18 @@ def _purlifier(self) -> "_Purlifier":
     @classmethod
     def from_file(cls, lockfile_path: RootedPath) -> "PackageLock":
         """Create a PackageLock from a package-lock.json file."""
-        with lockfile_path.path.open("r") as f:
-            lockfile_data = json.load(f)
+        try:
```
a-ovchinnikov: I would suggest creating a helper for this.

hermeto/core/package_managers/npm.py
farhann-saleem: Done 
updated  solution to recommend npm install to regenerate  file

hermeto/core/package_managers/npm.py
farhann-saleem: Done 
extracted a shared _load_json_file() helper used by both call sites

hermeto/core/package_managers/npm.py
```diff
@@ -394,6 +394,20 @@ def get_purl(
 NormalizedUrl = NewType("NormalizedUrl", str)
 
 
+def _load_json_file(file_path: Path, file_description: str) -> Any:
+    """Load and parse a JSON file, raising InvalidLockfileFormat on decode errors."""
+    try:
+        with file_path.open("r") as f:
+            return json.load(f)
+    except json.JSONDecodeError as e:
+        raise InvalidLockfileFormat(
```
a-ovchinnikov: This is a tricky one: in one case it is a a lcok file, in another it is not a lock file. You should not be raising `InvalidLockFfileFormat` for a non-lockfile. I think you could dispatch basing on whether "lock" is present in filename (or use file_desctiption for that, file name could be easily deduced from a path) and pick an exception basing on that. As for re-generating the file -- this is not something a typical user should be doing, if a package contains a lock file so corrupted that it cannot be parsed it should be resolved on package maintainer's side, not by a user. I would say solution should be "Please reach out to package {package_name} maintainers about this issue" at the very least, maybe just skipping solution altogether is a reasonable approach here.

farhann-saleem: Done 
 lockfiles now raise InvalidLockfileFormat (no custom solution)
 non-lockfiles raise UnexpectedFormat
 Dispatched based on filename

slimreaper35:
@farhann-saleem Please run at least unit tests/linters **before** you submit a pull request or you make some new changes.

farhann-saleem:
@slimreaper35   Fixed, sorry about that

PR #1388, opened 2026-03-12
fix(sbom): standardize __hash__ to prevent data loss and collisions.

notjivi:
### Description
While reviewing the SPDX models, I identified several unsafe `__hash__` implementations that cause silent hash collisions. Because Hermeto uses `set()` operations to merge relationships and deduplicate data (e.g., `list(set(self.relationships + processed_other))`), these collisions lead to silent data loss of valid dependency information.

This PR standardizes `SPDXPackage`, `SPDXRelation`, and `SPDXSbom` to use native Python tuple hashing (aligning them with the existing correct implementation already used in `SPDXPackageExternalRef`).

### Specific Fixes
1. **String Concatenation Collisions (`SPDXRelation` and `SPDXSbom`)**
   Hashing concatenated strings erases variable boundaries. For example, `pkg-a` + `123` yields the same hash as `pkg-a123` + `""`. If both exist in a merged SBOM, `set()` will silently delete one.
2. **Class Hashing Bug (`SPDXSbom`)**
   The method called `hash(SPDXCreationInfo)` instead of hashing the instance data. This means two SBOMs generated at different times evaluated to the exact same hash, completely ignoring timestamps and creators. 
3. **Commutative Addition Erasure (`SPDXPackage`)**
   Summing integer hashes (`hash(a) + hash(b)`) destroys list ordering and drastically increases collision risk.

<details>
<summary>Proof of Concept: Silent Data Loss in SPDXRelation</summary>

Here is a minimal reproduction showing how the previous concatenation method silently drops dependencies during a `set()` merge due to boundary erasure:

```python
class SPDXRelation:
    def __init__(self, spdxElementId, relatedSpdxElement, relationshipType, comment=""):
        self.spdxElementId = spdxElementId
        self.relatedSpdxElement = relatedSpdxElement
        self.relationshipType = relationshipType
        self.comment = comment

    # Previous broken implementation
    def __hash__(self) -> int:
        return hash(
            hash(self.spdxElementId + self.relatedSpdxElement + self.relationshipType)
            + hash(self.comment)
        )

    def __eq__(self, other):
        return type(self) == type(other) and hash(self) == hash(other)

# Two fundamentally different dependencies
rel_A = SPDXRelation("SPDXRef-Package-X", "-v1", "DEPENDS_ON")
rel_B = SPDXRelation("SPDXRef-Package-X-v", "1", "DEPENDS_ON")

# Simulate Hermeto merging the SBOMs using a set (like line 699 of sbom.py)
merged_relationships = list(set([rel_A, rel_B]))

print(f"Total relationships after merge: {len(merged_relationships)}")
# Output: 1  <--- FATAL: Data Loss! The second dependency was silently erased.

### DIFF
--- a/hermeto/core/models/sbom.py
+++ b/hermeto/core/models/sbom.py
@@ -501,14 +501,15 @@ class SPDXPackage(pydantic.BaseModel):
 
     def __hash__(self) -> int:
         return hash(
-            hash(self.SPDXID)
-            + hash(self.name)
-            + hash(self.versionInfo)
-            + hash(self.downloadLocation)
-            + sum(hash(e) for e in self.externalRefs)
-            + sum(hash(a) for a in self.annotations)
-            # NOTE: in an exceptionally rare case there could be a slim chance of collision here.
-            + hash(self.sourceInfo)
+            (
+                self.SPDXID,
+                self.name,
+                self.versionInfo,
+                self.downloadLocation,
+                tuple(self.externalRefs),
+                tuple(self.annotations),
+                self.sourceInfo,
+            )
         )
 
     @staticmethod
@@ -556,8 +557,7 @@ class SPDXRelation(pydantic.BaseModel):
 
     def __hash__(self) -> int:
         return hash(
-            hash(self.spdxElementId + self.relatedSpdxElement + self.relationshipType)
-            + hash(self.comment)
+            (self.spdxElementId, self.relatedSpdxElement, self.relationshipType, self.comment)
         )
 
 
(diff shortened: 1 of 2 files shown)


a-ovchinnikov:
@notjivi please refresh my memory, what is the effective type of `relatedSpdxElement`? Without looking at the code I think it is typed as `str`, but it is effective range is much narrower.

notjivi:
yea it's a `str`. the format is pretty predictable so an actual collision is rare, but doing `hash(a + b + c)` still felt kinda risky.

I switched it to a standard tuple `hash((a, b, c))` instead to be safe from any weird edge cases. it also matches what `SPDXPackageExternalRef` is already doing in the file keeping it consistent.

notjivi:
Looking at this again, you're right to point it out, `relatedSpdxElement` values are sanitized through `sanitize_spdxid()`, so the practical collision possibility of it happening is way lower than my initial claim. This fix is then primarily about consistency with `SPDXPackageExternalRef` (ie the tuple based hashing).
The stronger fix in this PR is the `hash(SPDXCreationInfo)` bug as it hashes the class itself. Which makes it so that every `SPDXSbom` instance gets the same value, regardless of the time or who created it. Which means set based deduplication can silently collapse distinct SBOMs.

a-ovchinnikov:
> The stronger fix in this PR is the hash(SPDXCreationInfo) bug as it hashes the class itself.

Yes, this is a genuine bug

notjivi:
Since the core hashing logic was addressed and merged in #1423, I'm going to close this PR to keep the queue clean! Glad we got this bug fixed.

PR #1390, opened 2026-03-12
Allow testing release workflow without requiring a release tag

MonishPuttu:
Fixes: #1290
Tested the release workflow using workflow_dispatch and reproduced the failure where the build step errored because no tag was generated without a release tag.

docker/metadata-action produces no tags outside a release event while docker/build-push-action still attempts to push an image.

Changed push: true to `push: ${{ github.event_name == 'release' }}` and added a fallback test tag for manual runs.

This allows the same workflow to be executed for testing without requiring a release.

Verified by running the workflow manually and confirming the build succeeds with a test tag and no image push.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,7 +17,7 @@ on:
   workflow_dispatch:
     inputs:
       push_latest:
-        description: 'Also push as latest tag'
+        description: "Also push as latest tag"
         required: true
         type: boolean
         default: false
@@ -60,6 +60,7 @@ jobs:
           tags: |
             type=semver,pattern={{version}}
             type=raw,value=latest,enable=${{ (github.event_name == 'workflow_dispatch' && inputs.push_latest) || (github.event_name == 'release' && !github.event.release.prerelease) }}
+            type=raw,value=test,enable=${{ github.event_name == 'workflow_dispatch' }}
 
       # Multi-arch build setup
       - name: Build and push container image
@@ -68,6 +69,6 @@ jobs:
           context: .
           file: ./Dockerfile
           platforms: linux/amd64,linux/arm64
-          push: true
+          push: ${{ github.event_name == 'release' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}


.github/workflows/release.yaml
```diff
@@ -17,7 +17,7 @@ on:
   workflow_dispatch:
     inputs:
       push_latest:
-        description: 'Also push as latest tag'
+        description: "Also push as latest tag"
```
eskultety, line 20: Unrelated change, please drop.

.github/workflows/release.yaml
```diff
@@ -54,20 +54,21 @@ jobs:
...
       # Multi-arch build setup
       - name: Build and push container image
-        uses: docker/build-push-action@v7
+        uses: docker/build-push-action@v6
         with:
           context: .
           file: ./Dockerfile
           platforms: linux/amd64,linux/arm64
-          push: true
+          push: ${{ github.event_name == 'release' }}
```
eskultety, line 72: This directly contradicts the intent of the manual workflow denoted at the top of the doc: https://github.com/hermetoproject/hermeto/blob/main/.github/workflows/release.yaml#L7-L16

This only pushes the built image with the release event, but not on a manual workflow run.

.github/workflows/release.yaml
```diff
@@ -54,20 +54,21 @@ jobs:
...
       - name: Extract metadata (tags, labels)
         id: meta
-        uses: docker/metadata-action@v6
+        uses: docker/metadata-action@v5
         with:
           images: ghcr.io/${{ github.repository }}
           tags: |
             type=semver,pattern={{version}}
             type=raw,value=latest,enable=${{ (github.event_name == 'workflow_dispatch' && inputs.push_latest) || (github.event_name == 'release' && !github.event.release.prerelease) }}
+            type=raw,value=test,enable=${{ github.event_name == 'workflow_dispatch' }}
```
eskultety, line 63: So what this does in combination of the line above is it'll tag the throwaway image as latest + test and that's it, do I understand correctly?

MonishPuttu:
@eskultety Thanks for the detailed explanation. I realize now that I misunderstood the problem described in #1290.

While investigating, I reproduced the failure when running the workflow manually without a tag and assumed the issue was related to the build failing due to missing tags generated by docker/metadata-action. To verify this, I triggered the workflow via workflow_dispatch and observed the build error, which led me to propose a workaround that added a fallback tag and restricted pushing to release events.

Based on your clarification, I understand that the real concern is validating the correctness of the release workflow logic itself for example incorrect GitHub context usage like (github.event.release.latest), rather than testing the container build or handling missing tags.

I'll close this PR and revisit the issue with that understanding, exploring approaches such as adding workflow validation using actionlint to catch these problems earlier. I'll follow up with a cleaner proposal that addresses the actual problem.

Thanks again for the feedback.

MonishPuttu:
@eskultety Before proceeding further, I want to confirm that I understood the issue correctly.

Based on your suggestion about tools like actionlint, I’m considering adding a CI workflow that runs actionlint on changes under .github/workflows/ to catch invalid expressions or incorrect GitHub context usage, for example non-existent fields like `github.event.release.latest` before they reach the release workflow.

Would this be the direction you had in mind for addressing the issue? If so, I’d be happy to open a PR.

MonishPuttu:
@eskultety Opened this PR to test the proposed approach using actionlint to validate GitHub workflow expressions. Please let me know if this aligns with the direction you had in mind or if there are improvements you would recommend.

PR #1391, opened 2026-03-12
fix(pip): support URL fragments in wheel detection

notjivi:
This PR ensures that pip wheel packages are correctly identified even when the requirement URL contains a fragment (e.g., #sha256=...).

The current logic uses .endswith(".whl") on the full URL string. While #cachito_hash was recently deprecated (as noted in #1387), standard PEP-compliant URLs frequently use fragments for integrity checks (e.g., #sha256=...).

Currently, any URL with a fragment causes the wheel detection to fail, because it misidentifies the package as a source distribution (sdist) and fails to set the binary flag in the SBOM.

The Fix
Refactored _process_url_req in hermeto/core/package_managers/pip/main.py to use urlparse(url).path instead of a raw string check. This isolates the filename from any query parameters or fragments.

Updated unit tests in tests/unit/package_managers/pip/test_main.py to use standard #sha256 fragments.

Related Issues
Supersedes #1387.

### DIFF
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -356,7 +356,8 @@ def _process_url_req(
         download_info=_download_url_package(req, pip_deps_dir, trusted_hosts),
         **kwargs,
     )
-    if req.url.endswith(WHEEL_FILE_EXTENSION):
+    parsed_url = urlparse.urlparse(req.url)
+    if parsed_url.path.endswith(WHEEL_FILE_EXTENSION):
         result["package_type"] = "wheel"
 
     return result
--- a/tests/unit/package_managers/pip/test_main.py
+++ b/tests/unit/package_managers/pip/test_main.py
@@ -1643,3 +1643,29 @@ def test_fetch_pip_source_correctly_reraises_when_there_is_a_dependency_cargo_lo
 
     with pytest.raises(PackageWithCorruptLockfileRejected):
         pip.fetch_pip_source(request)
+
+
+@pytest.mark.parametrize(
+    ("test_url", "expected_type"),
+    [
+        ("https://example.com/pkg-1.0-py3-none-any.whl", "wheel"),
+        ("https://example.com/pkg-1.0-py3-none-any.whl#sha256=08695f5ad7", "wheel"),
+        ("https://example.com/pkg-1.0-py3-none-any.whl?v=1.0#sha256=08695f5ad7", "wheel"),
+        ("https://example.com/pkg-1.0.tar.gz", ""),
+        ("https://example.com/pkg-1.0.tar.gz#sha256=08695f5ad7", ""),
+        ("https://example.com/pkg-1.0.tar.gz?v=1.0#sha256=08695f5ad7", ""),
+    ],
+)
+@mock.patch("hermeto.core.package_managers.pip.main._download_url_package")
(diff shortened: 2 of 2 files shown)


hermeto/core/package_managers/pip/main.py
```diff
@@ -355,7 +356,8 @@ def _process_url_req(
         download_info=_download_url_package(req, pip_deps_dir, trusted_hosts),
         **kwargs,
     )
-    if req.url.endswith(WHEEL_FILE_EXTENSION):
+    parsed_url = urlparse.urlparse(req.url)
```
farhann-saleem, line 359: Is there a real-world requirements.txt where a .whl URL has a #sha256 fragment? 
pip usually puts hashes in --hash options, not URL fragments.

notjivi:
Apologies for the misclick which resulted in accidental closing of the pr.



@farhann-saleem 
>_download_url_package already uses urlparse on req.url

this is what made it necessary for _process_url_req to be using urlparse for the url asw. 
the download phase was working okay but the processing phase was labelling it as sdist.

>was this inconsistency causing an actual SBOM issue for a project you tested?

yea it was as explained in #1387
i'll explain it here again

So, if we're using a url like 
https://.../pkg.whl#sha=sha256:123...
the current endwith function would label it as sdist. Resulting in an sbom with a missing hermeto:pip:package:binary
image

After the fix, for the same link, the sbom correctly displays the hermeto:pip:package:binary
image


>Is there a real-world requirements.txt where a .whl URL has a #sha256 fragment?
pip usually puts hashes in --hash options, not URL fragments.

A url with a #sha256 fragment is an industry standard (PEP503). A simple example would be:
pypi.org/simple/requests/
you will see that every single distribution link, both wheels and tarballs, includes the hash in the fragment.

hermeto/core/package_managers/pip/main.py
a-ovchinnikov, line 1: _Commit message_: please add some background information to the commit message for the sake of future maintainers. A short description of what was the problem, why it was a problem (with a link to the PEP) would be very helpful here.

slimreaper35:
@notjivi Do you have a simple requirements.txt file with such a URL? Maybe even with the steps how to generate it.

hermeto/core/package_managers/pip/main.py
notjivi: Done.

notjivi:
@slimreaper35 
Alright I'll explain how to reproduce the issue

requirements.txt
image
requests @ https://files.pythonhosted.org/packages/70/8e/0e2d847013cb52cd35b38c009bb167a1a26b2ce6cd6965bf26b47bc0bf44/requests-2.31.0-py3-none-any.whl#sha256=942c5a0f9b87940265e317231075621a14982d1b13e44434459da44dc69d4fb4 --hash=sha256:942c5a0f9b87940265e317231075621a14982d1b13e44434459da44dc69d4fb4

bom.json
image
there is no hermeto:pip:package:binary listed.

but after the fix
this is what bom.json looks like
image

Let me know if this reproduction is clear enough or if you'd like me to provide anything else.

slimreaper35:
> Let me know if this reproduction is clear enough or if you'd like me to provide anything else.

Okay, I get the same result.

notjivi:
Hi @a-ovchinnikov, just wanted to follow up on this. I have updated the commit message with PEP-503 context and added the background info as requested, and slimreaper35 has already approved, let me know if there is anything else needed.

PR #1402, opened 2026-03-13
cargo: add CARGO_NET_OFFLINE=true to build-phase env vars

### LINKED ISSUE #1398: cargo: CARGO_NET_OFFLINE=true missing from generated build-phase env vars
Noticed that the build environment hermeto generates for Cargo doesn't include `CARGO_NET_OFFLINE=true`.

The gomod backend sets `GOPROXY=file://...,off` to hard-block any outbound connections during the build. npm does the same with `npm_config_offline=true`. But for Cargo, the only network guard in place is the vendor config in `.cargo/config.toml`:

```toml
[source.crates-io]
replace-with = "vendored-sources"
[source.vendored-sources]
directory = "..."
```

That handles normal crate resolution, but it doesn't stop Cargo from making network calls from a `build.rs` script, or from trying to fetch a crate that slipped through the vendor step. With `CARGO_NET_OFFLINE=true` set, Cargo will fail immediately if anything tries to reach the network rather than silently succeeding.

Fix should just be adding it to the `environment_variables` block in the cargo build config output, same pattern as gomod. Want to confirm this is the right place to add it before I send a PR.

sujal-jain-activities:
Fixes #1398

## What
Added `CARGO_NET_OFFLINE=true` to the environment variables generated by
the Cargo backend during the build phase, using module-level constants
for the variable name and value.

## Why
The Cargo backend was relying solely on `.cargo/config.toml` with
vendored-sources to block network access. This doesn't prevent network
calls from `build.rs` scripts or crates that slipped through vendoring.

Other backends already handle this correctly:
- gomod sets `GOPROXY=file://...,off`
- npm sets `npm_config_offline=true`

`CARGO_NET_OFFLINE=true` makes Cargo fail immediately on any network
attempt instead of silently succeeding.

## Changes
- Defined `CARGO_OFFLINE_ENV_VAR_NAME` and `CARGO_OFFLINE_ENV_VAR_VALUE`
  as module-level constants in `cargo/main.py`
- Used constants in the `env_vars` block inside `fetch_cargo_source()`
- Updated test to import and assert against the same constants

## Testing
- Added unit test asserting `CARGO_NET_OFFLINE=true` is present in
  generated env vars
- All tests passing:
  `nox -s python-3.13 -- tests/unit/package_managers/cargo/test_main.py`

### DIFF
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -37,6 +37,9 @@ from hermeto.core.utils import run_cmd
 
 log = logging.getLogger(__name__)
 
+CARGO_OFFLINE_ENV_VAR_NAME = "CARGO_NET_OFFLINE"
+CARGO_OFFLINE_ENV_VAR_VALUE = "true"
+
 
 class CargoVendorResult(NamedTuple):
     """
@@ -161,7 +164,7 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
     if backend_annotation := create_backend_annotation(components, "cargo"):
         annotations.append(backend_annotation)
     env_vars = [
-        EnvironmentVariable(name="CARGO_NET_OFFLINE", value="true"),
+        EnvironmentVariable(name=CARGO_OFFLINE_ENV_VAR_NAME, value=CARGO_OFFLINE_ENV_VAR_VALUE),
     ]
     return RequestOutput.from_obj_list(
         components=components,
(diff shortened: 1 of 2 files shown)


harshakumar25:
Thanks @sujal-jain-activities for raising pr,
personlly i have raised this issue so i think i am good to work on this .
but i don't looking for blindly ai coded please close this pr.

sujal-jain-activities:
can u tell me how could u say that's blind ai coded 
i have taken some help where i got a doubt but that's not blind ai coded called

eskultety:
`CARGO_NET_OFFLINE=true` would make sense to add if there was evidence that cargo tries to contact the internet even though vendored sources were requested (our override). If that happened, it could either be that cargo is broken OR it were by design in which case we'd have to add it to unblock failed hermetic builds, however, we have so far not see any failed cargo build that would originate at cargo trying to connect to the internet. Closing with the same reasoning as #1398 .

PR #1403, opened 2026-03-13
Add Actionlint workflow to validate GitHub Actions workflows

MonishPuttu:
Fixes: #1290
This PR adds a CI workflow that runs actionlint.

The goal is to catch invalid expressions or incorrect GitHub context usage for example non-existent fields like `github.event.release.latest` during PR validation, before such changes reach workflows like the release pipeline.

This PR is mainly intended to verify that this approach works. Since the workflow is newly added, it needs to run at least once before it appears in the Actions UI. Opening this PR allows the workflow to be triggered and verified. If this aligns with the intended solution, I’d be happy to refine or extend the implementation based on feedback.

Hope this approach would solve the problems mentioned in the discussion #1390.

### DIFF
--- a/.github/workflows/actionlint.yaml
+++ b/.github/workflows/actionlint.yaml
@@ -1,11 +1,11 @@
 name: Lint workflows
 
 on:
+  workflow_dispatch:
   pull_request:
     paths:
       - ".github/workflows/*.yml"
       - ".github/workflows/*.yaml"
-
   push:
     branches:
       - main


eskultety:
> The goal is to catch invalid expressions or incorrect GitHub context usage for example non-existent fields like github.event.release.latest during PR validation, before such changes reach workflows like the release pipeline.

@MonishPuttu have you verified the action lint fixes the problem?

MonishPuttu:
@eskultety I initially thought the workflow needed to run once so that it would appear in the Actions tab and I could verify the implementation. However, after your question I tried validating whether actionlint would actually catch the incorrect field.

To test this, I created a small temporary workflow referencing github.event.release.latest and ran actionlint locally, but it did not report any issue. It seems that webhook payload fields like this are not validated by actionlint, so this check would not prevent the problem described in https://github.com/hermetoproject/hermeto/issues/1290.

At the moment I’m not aware of another straightforward way to automatically detect this type of mistake in workflow expressions. It might be useful to document this finding in the original issue so that future contributors can continue investigating possible approaches.

MonishPuttu:
The VS Code GitHub Actions extension does flag it with "Context access might be invalid: latest". 
However I am not able to find a straightforward CI-level enforcement for this kind of bug

So the situation is: 
-  VS Code extension catches it at edit time locally
-  actionlint does not catch it in CI or locally

image

eskultety:
> To test this, I created a small temporary workflow referencing github.event.release.latest and ran actionlint locally, but it did not report any issue. It seems that webhook payload fields like this are not validated by actionlint

Exactly! Note what I said here: https://github.com/hermetoproject/hermeto/pull/1390#pullrequestreview-3936228644

> Adopting something like https://github.com/rhysd/actionlint could alleviate some of the pain and resolve part of the problem, but definitely not the core of the issue

I knew that linter would not catch it because it's a semantic problem, but I still do find value in linting though.

> At the moment I’m not aware of another straightforward way to automatically detect this type of mistake in workflow

Neither am I at the moment that's why #1290 was written the way it was.

MonishPuttu:
Thanks for confirming. I understand what you mean. I will explore it further and let you know if I find anything.

eskultety:
Superseded by https://github.com/hermetoproject/hermeto/pull/1416

PR #1405, opened 2026-03-13
feat(yarn): support git dependencies in Yarn Berry projects

jacobbaungard:
Replaces: https://github.com/hermetoproject/hermeto/pull/1295

## Summary

Adds Yarn Berry support for git dependencies, mirroring the approach already used by the npm backend: clone git sources as tarballs and avoid executing packaging code during prefetch.

The backend scans `yarn.lock` for git-resolved dependencies, clones them as tarballs, and writes yarn resolutions so that `yarn install` uses the local copies instead of fetching from git. The lockfile is verified for consistency before any modifications, and `enableImmutableInstalls` is restored after the resolution update.

PURLs for overridden git deps carry a `vcs_url` qualifier pointing back to the original repository and commit.

Unsupported variants (patched git deps, workspace+commit selectors) are rejected with actionable error messages.

## Motivation

This was motivated by MLflow's Yarn Berry frontend (`mlflow/server/js`), which declares git dependencies from GitHub.

End-to-end validation with the current code:

```shell
#!/bin/bash
set -e

# Clean up any prior runs
rm -rf /tmp/mlflow-e2e /tmp/mlflow-rebuild /tmp/hermeto-output /tmp/hermeto-venv /tmp/corepack-local

# Clone the repo
git clone --depth 1 https://github.com/red-hat-data-services/mlflow.git /tmp/mlflow-e2e

# Install hermeto from the branch under review
python3 -m venv /tmp/hermeto-venv
source /tmp/hermeto-venv/bin/activate
pip install -e .

# Ensure corepack/yarn are available
npm install --prefix /tmp/corepack-local corepack
export PATH="/tmp/corepack-local/node_modules/.bin:$PATH"

# Run hermeto fetch-deps
hermeto fetch-deps \
  --source /tmp/mlflow-e2e \
  --output /tmp/hermeto-output \
  '{"type": "yarn", "path": "mlflow/server/js"}'

# Verify: apply the project files to a fresh copy and run yarn install
git clone --depth 1 https://github.com/red-hat-data-services/mlflow.git /tmp/mlflow-rebuild
python3 -c "
import json
from string import Template

bc = json.load(open('/tmp/hermeto-output/.build-config.json'))
for pf in bc['project_files']:
    path = pf['abspath'].replace('/tmp/mlflow-e2e/', '/tmp/mlflow-rebuild/')
    content = Template(pf['template']).safe_substitute(output_dir='/tmp/hermeto-output')
    open(path, 'w').write(content)
    print(f'Wrote: {path}')
"

cd /tmp/mlflow-rebuild/mlflow/server/js
YARN_ENABLE_GLOBAL_CACHE=false \
YARN_ENABLE_IMMUTABLE_CACHE=false \
YARN_ENABLE_MIRROR=true \
YARN_GLOBAL_FOLDER=/tmp/hermeto-output/deps/yarn \
  yarn install --mode skip-build
  ```

### CODE REVIEW

File: tests/unit/package_managers/javascript/npm/test_resolver.py [hunk 1]
```diff
@@ -417,7 +417,7 @@ def test_npm_settings_rejects_proxy_urls_containing_credentials(
 @mock.patch("hermeto.core.package_managers.javascript.npm.resolver.async_download_files")
 @mock.patch("hermeto.core.package_managers.javascript.npm.resolver.must_match_any_checksum")
 @mock.patch("hermeto.core.checksum.ChecksumInfo.from_sri")
-@mock.patch("hermeto.core.package_managers.javascript.npm.resolver.clone_as_tarball")
+@mock.patch("hermeto.core.package_managers.javascript.js_utils.clone_as_tarball")
```
a-ovchinnikov, line 424: UTs fail on ea8dde88: AttributeError: module 'hermeto.core.package_managers.javascript' has no attribute 'js_utils'

hermeto/core/package_managers/javascript/npm/resolver.py
```diff
@@ -44,25 +44,15 @@ def _clone_repo_pack_archive(
...
     :raise FetchError: If download failed
     """
     info = extract_git_info_npm(vcs)
-    download_path = download_dir.join_within_root(
-        info["host"],  # host
-        info["namespace"],
-        info["repo"],
-        f"{info['repo']}-external-gitcommit-{info['ref']}.tgz",
+    return clone_repo_pack_archive(
+        info["url"], info["ref"], download_dir, info["host"], info["namespace"], info["repo"]
```
a-ovchinnikov: Won't it be easier to just pass `info` at this point? (Or to  at least unpack it?) Alternatively this looks like a good candidate to become an `info` method.

hermeto/core/package_managers/javascript/js_utils.py
```diff
@@ -0,0 +1,42 @@
...
+"""Common utilities shared between JavaScript package managers (npm, yarn)."""
+
+from pathlib import Path
+
+from hermeto.core.rooted_path import RootedPath
+from hermeto.core.scm import clone_as_tarball
+
+
+def clone_repo_pack_archive(
+    clone_url: str,
```
a-ovchinnikov: Type-wise, WDYT about being more specific and using `pydantic.HttpUrl` instead of `str` here? It might also make sense to introduce a `pydantic.constr` for `ref` while we are here.

hermeto/core/package_managers/javascript/js_utils.py
```diff
@@ -1,38 +1,69 @@
...
+            host = path = None
+
+    if not host or not path:
+        raise PackageRejected(
+            f"Cannot parse git URL: {clone_url}",
+            solution="Ensure the git dependency has a valid URL.",
+        )
+
+    namespace, _, repo = path.rpartition("/")
+    return {"host": host, "namespace": namespace, "repo": repo}
```
a-ovchinnikov: Not actionable, just wondering why you decided to return a dictionary instead of a tuple? A tuple could be unpacked with little change to the original code.

hermeto/core/package_managers/javascript/js_utils.py
```diff
@@ -1,38 +1,69 @@
...
-    namespace: str,
-    repo: str,
 ) -> RootedPath:
     """Clone a git repository at a specific ref and pack it as a tarball.
 
     The tarball path follows the convention:
         {deps_dir}/{host}/{namespace}/{repo}/{repo}-external-gitcommit-{ref}.tgz
 
-    :param clone_url: the git clone URL
+    :param clone_url: the git clone URL (protocol-based or SCP-style)
```
a-ovchinnikov: Once again, this information is better be passed as a type: `clone_url: HttpUrl| SCPUrl`

tests/unit/package_managers/test_js_utils.py
```diff
@@ -0,0 +1,94 @@
...
+                "git@gitlab.com:org/subgroup/repo.git",
+                {"host": "gitlab.com", "namespace": "org/subgroup", "repo": "repo"},
+            ),
+            (
+                "http://bitbucket.org/ns/repo.git",
+                {"host": "bitbucket.org", "namespace": "ns", "repo": "repo"},
+            ),
+        ],
+    )
+    def test_valid_urls(self, url: str, expected: dict[str, str]) -> None:
```
a-ovchinnikov: WDYT about making this a doctest instead?

tests/unit/package_managers/test_js_utils.py
```diff
@@ -0,0 +1,94 @@
...
+        ],
+    )
+    def test_invalid_urls(self, url: str) -> None:
+        with pytest.raises(PackageRejected, match="Cannot parse git URL"):
+            extract_git_url_parts(url)
+
+
+class TestCloneRepoPackArchive:
+    @mock.patch("hermeto.core.package_managers.javascript.js_utils.clone_as_tarball")
+    def test_creates_tarball_at_expected_path(self, mock_clone: mock.Mock, tmp_path: Path) -> None:
```
a-ovchinnikov: Do we need these tests? It looks like ITs cover these scenarios.

hermeto/core/package_managers/javascript/js_utils.py
```diff
@@ -1,38 +1,69 @@
...
+
+    Handles both protocol-based URLs (https://host/ns/repo.git) and SCP-style
+    URLs (git@host:ns/repo.git).
+
+    :param clone_url: the git clone URL
+    :return: dict with keys: host, namespace, repo
+    :raises PackageRejected: if the URL cannot be parsed
+    """
+    if "://" in clone_url:
+        parsed = urlparse(clone_url)
```
a-ovchinnikov: `clone_url` being a `HttpUrl` would remove the need to parse it explicitly -- once constructed successfully it will carry references to elements of the original URL:
```python3
>>> HttpUrl("https://gitlab.com/org/subgroup/repo.git")
HttpUrl('https://gitlab.com/org/subgroup/repo.git')   # Note __repr__() done right here
>>> u = _
>>> u.host
'gitlab.com'
>>> u.path
'/org/subgroup/repo.git'
```

hermeto/core/package_managers/javascript/js_utils.py
```diff
@@ -1,38 +1,69 @@
...
+        host = parsed.hostname
+        path = parsed.path.strip("/").removesuffix(".git")
+    else:
+        # SCP-style: git@host:ns/repo.git
+        match = re.match(r"^[^@]*@([^:]+):(.+)$", clone_url)
+        if match:
+            host = match.group(1)
+            path = match.group(2).removesuffix(".git")
+        else:
+            host = path = None
```
a-ovchinnikov: This is effectively a constructor for `SCPUrl`

hermeto/core/package_managers/javascript/yarn/locators.py
```diff
@@ -285,35 +296,64 @@ def __str__(self) -> str:
...
     def get_param(self, param_name: str) -> str | None:
+        """Return a binding parameter value by name.
+
+        :param param_name: the binding key to look up
+        :return: the parameter value, or None if absent
+        :raises UnexpectedFormat: if the parameter is present more than once
+        """
         if not self.params or not (param_value := self.params.get(param_name)):
             return None
         if len(param_value) != 1:
```
a-ovchinnikov, line 347: Not related to your change as it stands, but this delayed validation feels weird: wrong input should be either rejected at construction time or returned as is, validation on `get` looks like an unexpected surprise to me.

tests/unit/package_managers/javascript/yarn/test_locators.py
```diff
@@ -579,7 +579,11 @@ def test_parse_locator(locator_str: str, expect_locator: Locator) -> None:
 @pytest.mark.parametrize("locator_str", UNSUPPORTED_LOCATORS)
 def test_parse_unsupported_locator(locator_str: str) -> None:
     with pytest.raises(
-        UnsupportedFeature, match=f"{APP_NAME} does not support Git or Exec dependencies"
+        UnsupportedFeature,
+        match=(
```
a-ovchinnikov: This is a good opportunity to get rid of a string match in a test.

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -29,27 +50,42 @@
...
-    components = []
+    components: list[Component] = []
+    project_files: list[ProjectFile] = []
 
     for package in request.yarn_packages:
         path = request.source_dir.join_within_root(package.path)
         project = Project.from_source_dir(path)
 
-        components.extend(_resolve_yarn_project(project, request.output_dir, package.workspaces))
+        pkg_components, pkg_project_files, _git_deps = _resolve_yarn_project(
```
a-ovchinnikov: It does not look like you use `_git_deps` anywhere, could you please use `_` to name the third element of return value to highlight that?

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -110,37 +145,73 @@ def _verify_repository(project: Project) -> None:
...
+    :param mode: the processing mode (strict or permissive).
     :param workspaces: optional list of workspace names to focus on (Yarn v4 only).
+    :return: a tuple of (components, project_files, git_deps)
     :raises PackageManagerError: if fetching dependencies fails
+    :raises PackageRejected: if git deps are found in strict mode
     """
     log.info(f"Fetching the yarn dependencies at the subpath {project.source_dir}")
 
     version = _configure_yarn_version(project)
+    _verify_repository(project)
```
a-ovchinnikov: Is it possible to verify a project at construction time? I.e. have an error when Project is instantiated vs at some later point.

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -338,3 +428,181 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+        if "workspace" in selector_qs:
+            # Raises UnsupportedFeature for #workspace=…&commit=…
+            parse_locator(resolution)
+
+        commit = selector_qs["commit"][0]
+        clone_url = _build_clone_url(protocol, ref.source)
+
+        name = parsed.name
+        if parsed.scope:
+            name = f"@{parsed.scope}/{name}"
```
a-ovchinnikov: Lines 457-484 read like they are a part of GitDeps constructor. WDYT about making it a pydantic model and let pydantic do all the heavy lifting?

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -338,3 +428,181 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+    project: Project,
+    git_deps: list[GitDep],
+    output_dir: RootedPath,
+) -> tuple[list[ProjectFile], dict[str, str]]:
+    """Clone git deps, write resolutions to package.json, and return ProjectFiles.
+
+    :param project: the Project whose package.json will be modified
+    :param git_deps: list of GitDep instances
+    :param output_dir: base output directory
+    :return: tuple of (project_files, tarball_vcs_url_map) where tarball_vcs_url_map maps
```
a-ovchinnikov: This information could be mostly conveyed by a correct type alias, like below, but better:
```python3
VCSUrlToTarMap = dict[VCSUrl, TarballPath]
...
-> tuple[list[ProjectFiles], VCSUrlToTarMap]
```

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -338,3 +428,181 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+        dep.ref,
+    ).as_vcs_url_qualifier()
+
+
+def _clone_and_resolve_git_deps(
+    project: Project,
+    git_deps: list[GitDep],
+    output_dir: RootedPath,
+) -> tuple[list[ProjectFile], dict[str, str]]:
+    """Clone git deps, write resolutions to package.json, and return ProjectFiles.
```
a-ovchinnikov: Nitpick: this describes three functions, not one, it might make sense to split `_clone_and_resolve_git_deps()` further.

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -338,3 +428,181 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+    :param project: the Project whose package.json will be modified
+    :param git_deps: list of GitDep instances
+    :param output_dir: base output directory
+    :return: tuple of (project_files, tarball_vcs_url_map) where tarball_vcs_url_map maps
+        tarball absolute path strings to vcs_url qualifiers for PURL generation.
+    :raises PackageRejected: if two git deps share the same name but different sources
+    """
+    seen_names: dict[str, tuple[str, str]] = {}
+    for dep in git_deps:
+        key = (dep.clone_url, dep.ref)
```
a-ovchinnikov: Why is it called `key` when it is used as value? It looks more like `url_ref_pair` to me

tests/unit/package_managers/javascript/yarn/test_main.py
```diff
@@ -344,26 +364,327 @@ def test_strip_workspace_scripts(rooted_tmp_path: RootedPath) -> None:
...
+
+
+@pytest.mark.parametrize(
+    "input_request",
+    [pytest.param([{"type": "yarn", "path": "."}], id="single_package_with_project_files")],
+    indirect=["input_request"],
+)
+@mock.patch("hermeto.core.package_managers.javascript.yarn.main._resolve_yarn_project")
+@mock.patch("hermeto.core.package_managers.javascript.yarn.project.Project.from_source_dir")
+def test_fetch_yarn_source_with_project_files(
```
a-ovchinnikov: Why is this a unit test and not an integration test?

tests/unit/package_managers/javascript/yarn/test_main.py
```diff
@@ -344,26 +364,327 @@ def test_strip_workspace_scripts(rooted_tmp_path: RootedPath) -> None:
...
+# --- Tests for git dependency support ---
+
+
+def _make_project_with_lockfile(
+    tmp_path: Path, lockfile_data: dict, package_json_data: dict | None = None
+) -> Project:
+    source_dir = RootedPath(tmp_path)
+    lockfile_path = tmp_path / "yarn.lock"
+    with lockfile_path.open("w") as f:
+        yaml.safe_dump(lockfile_data, f)
```
a-ovchinnikov: Unit tests which touch external services are considered an anti-pattern: one is supposed to be testing individual, logically isolated units. Tests which require e.g. file system interactions are integration tests.

tests/unit/package_managers/javascript/yarn/test_main.py
```diff
@@ -344,26 +364,327 @@ def test_strip_workspace_scripts(rooted_tmp_path: RootedPath) -> None:
...
+        "resolution": "ccto-wo-deps@git@github.com:cachito-testing/cachito-npm-without-deps.git#commit=2f0ce1d7",
+    },
+    "@databricks/json-bigint@https://github.com/databricks/json-bigint.git#commit=a1defaf9": {
+        "version": "0.2.3",
+        "resolution": "@databricks/json-bigint@https://github.com/databricks/json-bigint.git#commit=a1defaf9",
+    },
+}
+
+
+class TestParseLockfileGitDeps:
```
a-ovchinnikov, line 426: I think most of these tests could be removed if a proper model is used for parsing the lock files.

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -68,20 +73,39 @@ def fetch_yarn_source(request: Request) -> RequestOutput:
...
 
     annotations = []
     if backend_annotation := create_backend_annotation(components, "yarn"):
         annotations.append(backend_annotation)
+    if git_deps:
+        vcs_urls = {_build_vcs_url(d) for d in git_deps}
+        subjects = [
+            c.bom_ref
+            for c in components
+            if (PackageURL.from_string(c.purl).qualifiers or {}).get("vcs_url") in vcs_urls
```
a-ovchinnikov: I would recommend lifting the `PackageURL.from...` part into a helper:
```python3
component_in_vcs_urls = lambda c: (PackageURL.from_string(c.purl).qualifiers or {}).get("vcs_url") in vcs_urls
subjects = [c.bom_ref for c in components if component_in_vcs_urls(c)]
```
Generally if an iterable comprehension sits on more than one line it is a sign that something has to be done about it, multiline comprehensions defeat the purpose of iterable comprehensions (which is to make code easier to read).

tests/unit/package_managers/javascript/yarn/test_main.py
```diff
@@ -406,6 +406,42 @@ def test_fetch_yarn_source_with_project_files(
...
+            purl="pkg:npm/foo@1.0.0?vcs_url=git+https://github.com/owner/foo.git@abc123",
+            version="1.0.0",
+        )
+    ]
+    git_deps = [GitDep(name="foo", clone_url="https://github.com/owner/foo.git", ref="abc123")]
+    mock_resolve_yarn.return_value = (components, [], git_deps)
+
+    output = fetch_yarn_source(input_request)
+
+    git_ann = next(
```
a-ovchinnikov: Nitpick: there is `first_for()` for this in `core.utils`

tests/unit/package_managers/javascript/npm/test_resolver.py
ST3V1K: Done

hermeto/core/package_managers/javascript/npm/resolver.py
ST3V1K: I did a bigger refactor that touched this, more related to other comments

hermeto/core/package_managers/javascript/js_utils.py
ST3V1K: I introduced a new class that handles all the url variants

hermeto/core/package_managers/javascript/js_utils.py
ST3V1K: It was there before I started working on this, so I just kept it. Now I return a new class instance

hermeto/core/package_managers/javascript/js_utils.py
ST3V1K: Done

tests/unit/package_managers/test_js_utils.py
ST3V1K: Done

tests/unit/package_managers/test_js_utils.py
ST3V1K: Removed

hermeto/core/package_managers/javascript/js_utils.py
ST3V1K: Refactored

hermeto/core/package_managers/javascript/js_utils.py
ST3V1K: Done

tests/unit/package_managers/javascript/yarn/test_locators.py
ST3V1K: Removed from this test

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: Done

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: Done

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: I don't think a pydantic model fits this use case, but I created a helper git dep parser method to make the lockfile parser method cleaner

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: Done

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: I split the function into 3 and kept it only as a slim wrapper around them.

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: Renamed

tests/unit/package_managers/javascript/yarn/test_main.py
ST3V1K: Fair, I removed it since it's already tested by the new IT

tests/unit/package_managers/javascript/yarn/test_main.py
ST3V1K: Refactored most of the newly introduced tests to use in-memory dics and mocks instead of file system access.

tests/unit/package_managers/javascript/yarn/test_main.py
ST3V1K: Since I haven't introduced the pydantic model I haven't touched these

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: I introduced a new helper function to extract the vcs url

tests/unit/package_managers/javascript/yarn/test_main.py
ST3V1K: Done

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -386,3 +508,194 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+            tarball_rooted = cloned_sources[source_key]
+        else:
+            tarball_rooted = clone_repo_pack_archive(clone_url, yarn_deps_dir)
+            cloned_sources[source_key] = tarball_rooted
+
+        rel_to_output = tarball_rooted.path.relative_to(output_dir.path)
+        tarball_info[dep.name] = (tarball_rooted, rel_to_output)
+        tarball_vcs_url_map[str(tarball_rooted.path)] = _build_vcs_url(dep)
+
+    return tarball_info, tarball_vcs_url_map
```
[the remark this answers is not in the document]
ST3V1K: Done

hermeto/core/package_managers/javascript/npm/utils.py
tnevrlka, line 1: ad PR description:

I would reconsider whether necessary to mention a company (Red Hat) in the PR description

hermeto/core/package_managers/javascript/js_utils.py
```diff
@@ -1,45 +1,181 @@
...
+    ...
+    hermeto.core.errors.UnexpectedFormat: ...
+    >>> parse_git_clone_url("ftp://github.com/example-org/example-repo.git")
+    Traceback (most recent call last):
+    ...
+    hermeto.core.errors.UnexpectedFormat: ...
+    >>> parse_git_clone_url("git@github.com/example-org/example-repo")
+    Traceback (most recent call last):
+    ...
+    hermeto.core.errors.UnexpectedFormat: ...
```
tnevrlka: Isn't this a bit overkill for the docstring? I would imagine that this should be in a unit test instead

tests/unit/package_managers/javascript/yarn/test_main.py
a-ovchinnikov: Ok, that's fair, resolving.

tests/unit/package_managers/javascript/yarn/test_main.py
a-ovchinnikov: Thank you!

hermeto/core/package_managers/javascript/yarn/locators.py
```diff
@@ -285,35 +296,64 @@ def __str__(self) -> str:
...
+        selector: the selector identifying the dependency within its protocol
+        params: query-string bindings after '::'
+    """
+
     protocol: str | None
     source: str | None
     selector: str
     params: dict[str, list[str]] | None
 
     def get_param(self, param_name: str) -> str | None:
```
a-ovchinnikov, line 338: https://github.com/hermetoproject/hermeto/commit/373bf7dfd15ba8ba5308bc8405c9e35bc2777c8a : 

Non-blocking, not necessarily actionable right now: note, that this method again raises an exception that should have been raised on construction time. Other than that I wonder why `get_param()` was chosen instead of `__getitem__()`, and to think of it why this is a named tuple rather than, say, a regular class with `__iter__()` defined to assist with unpacking.

tests/unit/package_managers/javascript/yarn/test_resolver.py
```diff
@@ -109,9 +109,9 @@ def test_validate_unsupported_locators(
         resolve_packages(rooted_tmp_path)
 
     assert caplog.messages == [
-        f"{APP_NAME} does not support Git or Exec dependencies for Yarn Berry: is-positive@git@github.com:kevva/is-positive.git#commit=97edff6f525f192a3f83cea1944765f769ae2678",
-        f"{APP_NAME} does not support Git or Exec dependencies for Yarn Berry: is-positive@git@github.com:kevva/is-positive.git#commit=97edff6f525f192a3f83cea1944765f769ae2678",
-        f"{APP_NAME} does not support Git or Exec dependencies for Yarn Berry: holy-hand-grenade@exec:./generate-holy-hand-grenade.js#./generate-holy-hand-grenade.js::hash=3b5cbd&locator=berryscary%40workspace%3A.",
+        f"{APP_NAME} cannot process this Yarn Berry Git dependency: is-positive@git@github.com:kevva/is-positive.git#commit=97edff6f525f192a3f83cea1944765f769ae2678",
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/f7e3757ea600b77408f92ce2f3ba1e20e9b44700 : 

Non-blocking: this is the reason why direct string matches are usually a bad idea for UTs. Also here we test Python's logging capabilities, not our code, I would suggest just removing this assertion.

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -22,105 +50,137 @@
...
+    except UnexpectedFormat:
+        log.debug("Skipping resolution %r: could not parse locator", resolution)
+        return None
+
+    protocol = ref.protocol.removesuffix(":") if ref.protocol else None
+
+    # Validate patches with parse_locator (nested git fails; npm patches pass),
+    # then skip — they are not git deps to rewrite.
+    if protocol == "patch":
+        parse_locator(resolution)
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/38239b862e1b2efe0619eadc9f703847f7f2ba43 : 

WDYT about a wrapper around `parse)locator()` named `validate_resolution()`?

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -22,105 +50,137 @@
...
-        components=components,
-        environment_variables=_generate_environment_variables(),
-        project_files=[],
-        annotations=annotations,
-    )
+    if git_deps:
 
+        def get_vcs_url(c: Component) -> str | None:
+            purl = PackageURL.from_string(c.purl)
+            return purl.qualifiers.get("vcs_url") if purl.qualifiers else None
```
a-ovchinnikov, line 152: Non-blocking, non-actionable: this is an interesting round-trip, it might be hinting that something is off in our models.

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -387,3 +508,205 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+
+    for dep in git_deps:
+        clone_url = parse_git_clone_url(dep.clone_url, dep.ref)
+        host, namespace, repo, ref = (
+            clone_url.host,
+            clone_url.namespace,
+            clone_url.repo,
+            clone_url.ref,
+        )
+        if not host or not namespace or not repo or not ref:
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/38239b862e1b2efe0619eadc9f703847f7f2ba43 : 

Non-blocking nitpick:
```diff
- if not host or not namespace or not repo or not ref:
+ if not all((host, namespace, repo, ref)):
```

or (slightly more disturbing):

```diff 
- host, namespace, repo, ref = (
-     clone_url.host,
-     clone_url.namespace,
-     clone_url.repo,
-     clone_url.ref,
- )
- if not host or not namespace or not repo or not ref:
+ clone_url_params = host, namespace, repo, ref = (
+     clone_url.host,
+     clone_url.namespace,
+     clone_url.repo,
+     clone_url.ref,
+ )
+ if not all(clone_url_params):
```

or even:

```python3
class GitCloneUrl(...):
    ...
    def __iter__(self):
        yield self.host
        yield self.namespace
        yield self.repo
        yield self.ref
... ... ...
clone_url_params = host, namespace, repo, ref = get_git_clone_url(...)
if not all(clone_url_params):
    ...
```

hermeto/core/package_managers/javascript/yarn/project.py
```diff
@@ -125,7 +139,57 @@ def from_source_dir(cls, source_dir: RootedPath) -> "Project":
...
+        project.yarn_rc.get("installStatePath"): "installStatePath",
+        project.yarn_rc.get("patchFolder"): "patchFolder",
+        project.yarn_rc.get("virtualFolder"): "virtualFolder",
+    }
+
+    for path in paths_conf_opts:
+        if path is not None:
+            try:
+                project.source_dir.join_within_root(path)
+            except Exception:
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/38239b862e1b2efe0619eadc9f703847f7f2ba43:

`Exception` is a bit too broad, could you please narrow this down?

tests/unit/package_managers/javascript/yarn/test_main.py
```diff
@@ -338,26 +355,338 @@ def test_strip_workspace_scripts(rooted_tmp_path: RootedPath) -> None:
...
+                    "resolution": "npm-lifecycle-scripts@https://github.com/chmeliik/js-lifecycle-scripts.git#workspace=my-workspace&commit=0e786c88d5aca79a68428dadaed4b096bf2ae3e0",
+                },
+            },
+        ],
+        ids=["patched-git", "workspace-git"],
+    )
+    def test_rejects_unsupported_git_deps(self, lockfile: dict) -> None:
+        with pytest.raises(
+            UnsupportedFeature,
+            match="cannot process this Yarn Berry Git dependency",
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/38239b862e1b2efe0619eadc9f703847f7f2ba43:
Is this match absolutely necessary?

tests/unit/package_managers/javascript/yarn/test_main.py
```diff
@@ -338,26 +355,338 @@ def test_strip_workspace_scripts(rooted_tmp_path: RootedPath) -> None:
...
+        result = _git_deps_from_lockfile(lockfile)
+        assert len(result) == 1
+        assert result[0].name == "foo"
+
+
+@pytest.mark.parametrize(
+    "protocol, source, expected",
+    [
+        ("https", "//github.com/owner/repo.git", "https://github.com/owner/repo.git"),
+        ("git@github.com", "cachito-testing/repo.git", "git@github.com:cachito-testing/repo.git"),
```
a-ovchinnikov, line 548: https://github.com/hermetoproject/hermeto/commit/38239b862e1b2efe0619eadc9f703847f7f2ba43:
Non-blocking, non actionable: cachito-testing? I have not heard that name in years.

tests/unit/package_managers/javascript/yarn/test_main.py
```diff
@@ -338,26 +355,338 @@ def test_strip_workspace_scripts(rooted_tmp_path: RootedPath) -> None:
...
+                clone_url="https://github.com/owner/my-dep.git",
+                ref="abc123",
+            )
+        ]
+
+        output_dir = RootedPath("/fake/output")
+
+        with (
+            _make_project() as (project, _, _),
+            pytest.raises(PackageRejected, match="cannot be processed in strict mode"),
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/38239b862e1b2efe0619eadc9f703847f7f2ba43:
Is a match necessary here? If the exception is not specific enough it should be subclassed.

tests/unit/package_managers/javascript/yarn/test_main.py
a-ovchinnikov, line 1: https://github.com/hermetoproject/hermeto/commit/4ebf0acafe6f48d8a75e49eef906289deaac6e13:

_Commit message_: states that the commit "adds an annotation for git-resolved yarn dependencies.", at the same time the only changed file is tests/unit/package_managers/javascript/yarn/test_main.py. I think something must be adjusted, either the message or the diff.

a-ovchinnikov (commented):
Almost there! The outstanding comments are the one  about a broad exception, one about adjusting one commit message and two about string matches in UTs, everything else is observations/nice to haves.

tests/unit/package_managers/javascript/yarn/test_resolver.py
ST3V1K: Removed the `caplog` assertion

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: Added. It makes it more understandable.

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: I did a variant of the second suggestion.

hermeto/core/package_managers/javascript/yarn/project.py
ST3V1K: Done

tests/unit/package_managers/javascript/yarn/test_main.py
ST3V1K: No, removed

tests/unit/package_managers/javascript/yarn/test_main.py
ST3V1K: No, removed

tests/unit/package_managers/javascript/yarn/test_main.py
ST3V1K: I'm sorry for that. Moved the annotation change back to its commit.

tests/unit/package_managers/javascript/yarn/test_main.py
ST3V1K: I didn't even realize that when I was refactoring this old PR implementation :D

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -387,3 +488,200 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+    for dep in git_deps:
+        clone_url = parse_git_clone_url(dep.clone_url, dep.ref)
+        source_key = (clone_url.host, clone_url.namespace, clone_url.repo, clone_url.ref)
+        if not all(source_key):
+            raise UnexpectedFormat(
+                f"Cannot parse git URL: {dep.clone_url}",
+                solution="Ensure the git dependency has a valid URL and commit ref.",
+            )
+
+        source_key = cast(tuple[str, str, str, str], source_key)
```
a-ovchinnikov, line 649: https://github.com/hermetoproject/hermeto/commit/a058508984b65d4c0360eddad04663629f7f1d12: 

Non-blocking, non-actionable: having `cast` in a _dynamically_ typed language code feels like having sand in a sandwich. This likely indicates that something is not handled correctly, but rather quickly patched over with tools at hand (which, arguably, is the point of dynamically typed languages). I won't block on this, but please remember, that this is a sign that something is missing.

a-ovchinnikov (approved):

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: `cast` is only for mypy, the runtime check is fine. I had the unpack + chained-or guards before. Per [your earlier suggestion](https://github.com/hermetoproject/hermeto/pull/1405#discussion_r3752334746) I switched to `not all(...)`, but unfortunately mypy doesn't narrow tuple elements after that.

a-ovchinnikov: Yes, sorry, this was more to do with `mypy` behavior than your code. Casts are very unnatural and the fact that `mypy` sometimes needs them is rather upsetting, at least to me. A better type granularity is a topic way out of scope for this PR.

hermeto/core/package_managers/javascript/js_utils.py
a-ovchinnikov: I suggested converting a UT to doctests. The function is a thin wrapper aound `GitCloneUrl` which is a subclass of `BaseModel`. I'd say a couple examples would be enough for it.

ST3V1K: Trimmed the doctest examples. 3 positive cases (HTTPS, SCP-style, `git+` with ref) and 1 negative.

hermeto/core/package_managers/javascript/npm/utils.py
ST3V1K: Updated it a little

jacobbaungard:
Thanks for the work in brining this to the finish line. Merging seems blocked since a few conversations must be resolved - I believe the coderabbit ones. Are we OK to resolve these?

a-ovchinnikov:
> Merging seems blocked since a few conversations must be resolved - I believe the coderabbit ones. Are we OK to resolve these?

@jacobbaungard the change needs a second reviewer, that's the main blocker now, let me see what could be done about this.

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -134,14 +199,52 @@ def _resolve_yarn_project(
...
+            "Git dependencies in Yarn Berry projects cannot be processed in strict mode "
+            "because the lockfile must be modified to reference local tarballs."
+        )
 
     _set_yarnrc_configuration(project, output_dir, version)
+    if workspaces:
+        # Focus cannot use --mode skip-build, and enableScripts:false does not
+        # cover workspace scripts
+        # https://github.com/yarnpkg/berry/blob/7744e6678de126a2ca2398d4123e3f7e009256b8/packages/docusaurus/static/configuration/yarnrc.json#L252
+        _strip_workspace_scripts(project.source_dir)
```
derasdf: Blocking: this strip is undone before the first `yarn workspaces focus` runs, which reopens the exact hole it exists to close.

`_strip_workspace_scripts` here rewrites `package.json` on disk without `scripts`, but `_apply_git_dep_resolutions` then writes `project.package_json` (which still holds `scripts` in memory) straight back over it at [L663](https://github.com/hermetoproject/hermeto/blob/be659424b6710e5e1212c47676677dbe2adbd882/hermeto/core/package_managers/javascript/yarn/main.py#L663). So the first `yarn workspaces focus` at [L237](https://github.com/hermetoproject/hermeto/blob/be659424b6710e5e1212c47676677dbe2adbd882/hermeto/core/package_managers/javascript/yarn/main.py#L237) runs root `postinstall`/`prepare`.

That is arbitrary code execution during prefetch, which is exactly what principle #1 forbids: `focus` cannot use `--mode skip-build` and is not covered by `enableScripts: false`, per your own note at [L454](https://github.com/hermetoproject/hermeto/blob/be659424b6710e5e1212c47676677dbe2adbd882/hermeto/core/package_managers/javascript/yarn/main.py#L454). The re-strip at [L247](https://github.com/hermetoproject/hermeto/blob/be659424b6710e5e1212c47676677dbe2adbd882/hermeto/core/package_managers/javascript/yarn/main.py#L247) runs too late, the `focus` already executed them.

IMO strip the in-memory `project.package_json` so the resolutions write cannot undo it. And please add a workspaces + git-deps test, nothing covers that combination today.

derasdf (changes requested):

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -134,14 +174,52 @@ def _resolve_yarn_project(
...
+    # package data from the now-modified lockfile).
+    project_files, tarball_vcs_url_map = _clone_and_resolve_git_deps(project, git_deps, output_dir)
+
+    project.yarn_rc["enableImmutableInstalls"] = False
+    project.yarn_rc.write()
+    try:
+        _fetch_dependencies(project.source_dir, workspaces)
+    finally:
+        project.yarn_rc["enableImmutableInstalls"] = True
+        project.yarn_rc.write()
```
a-ovchinnikov: These lines are a good candidate to become a context manager:
```python3
with immutable_installs_disabled():
    _fetch_dependencies(project.source_dir, workspaces)
```

hermeto/core/package_managers/javascript/yarn/main.py
a-ovchinnikov: Good catch! This makes me think that stripping workspaces should happen in fetch.

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: Done

hermeto/core/package_managers/javascript/yarn/main.py
ST3V1K: Confirmed. Resolutions were writing `scripts` back before the first `focus`. 
`_fetch_dependencies` now strips from disk right before `focus`. No extra test; strip isn't a separate step git deps can undo anymore.

derasdf (approved):

hermeto/core/package_managers/javascript/js_utils.py
```diff
@@ -1,45 +1,160 @@
...
+from hermeto.core.errors import UnexpectedFormat
 from hermeto.core.rooted_path import RootedPath
 from hermeto.core.scm import clone_as_tarball
 
+_SCHEME_URL = Annotated[
+    AnyUrl,
+    UrlConstraints(allowed_schemes=["http", "https", "git", "ssh"]),
+]
+_SCHEME_URL_ADAPTER = TypeAdapter(_SCHEME_URL)
+_SCP_URL_RE = re.compile(r"^([^/@]+@)?[^/:]+:[^/].*$")
```
a-ovchinnikov: Please consider adding a comment to regexp.

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -387,3 +522,200 @@ def _verify_corepack_yarn_version(expected_version: semver.Version, source_dir:
...
+
+
+def _apply_git_dep_resolutions(
+    project: Project,
+    tarball_info: dict[str, tuple[RootedPath, Path]],
+) -> list[ProjectFile]:
+    """Write file: resolutions into package.json and return a templated ProjectFile."""
+    resolutions = project.package_json.data.get("resolutions", {})
+    for name, (tarball_rooted, _) in tarball_info.items():
+        resolutions[name] = f"file:{tarball_rooted.path}"
```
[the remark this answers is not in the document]
a-ovchinnikov: @coderabbitai provide a minimal lock file that demonstrates this problem.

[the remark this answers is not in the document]
a-ovchinnikov: @coderabbitai why would a project depend on two versions of the same package from different sources?

Issue #1406, opened 2026-03-14
Pip backend: canonicalize versions before PURL / SBOM dedupe


PuneetKumar1790:
### Summary
The pip backend currently uses `canonicalize_version()` when filtering distributions from PyPI, but the *raw* version string from the requirements is later propagated into the SBOM PURLs without normalization.
As a result, different textual representations of the same version (for example `1.0`, `1.0.0`, `1.00`, `01.0`) generate distinct PURLs and thus distinct `Component.key()` values, so they are treated as separate SBOM components even though they refer to the same logical version.
### Observed behavior
- In `_get_project_packages_from`, versions are compared via  
  `canonicalize_version(p.version) == canonicalize_version(version)`  
  which correctly matches distributions like `1.0` vs `1.0.0`.
- However, once a distribution is selected, its `version` field is preserved as the raw specifier from the requirement.
- That raw version is then threaded through into the `dependencies` list in `_resolve_pip()` and finally into `_generate_purl_dependency()`.
- `_generate_purl_dependency()` passes this unnormalized value directly to `PackageURL(version=...)`.
- `Component.key()` returns `self.purl`, so any textual differences in the version segment make components distinct for sorting/deduplication.
### Steps to reproduce
Using the real code paths inside hermeto:
1. Create a temporary `requirements.txt`:
   ```text
   requests==2.31.0
   requests==2.31
   requests==2.31.00
   requests==02.31
Run fetch_pip_source on that directory as a pip package.
Inspect the resulting SBOM components where name == "requests".
Observed output (summarized):

requests 2.31.0 → pkg:pypi/requests@2.31.0
requests 2.31 → pkg:pypi/requests@2.31
requests 2.31.00 → pkg:pypi/requests@2.31.00
requests 02.31 → pkg:pypi/requests@02.31
All four become separate components because their PURLs differ, even though canonicalize_version() normalizes each of these strings to the same canonical version.

A small synthetic test that constructs dependency dicts and Component objects directly via _generate_purl_dependency() and merge_component_properties() shows the same behavior: four variants in, four components out after merge.

Expected behavior
It seems more intuitive (and more friendly to downstream tools that consume PURLs) if versions are normalized at or before PURL construction, so that different textual spellings of the same version do not produce distinct PURLs and duplicate SBOM components.

Possible approaches
Canonicalize the version string once when constructing the dependency dict (e.g. in _resolve_pip() for kind == "pypi") using canonicalize_version(version).
[...]

PR #1407, opened 2026-03-14
Normalize pip dependency versions in PURLs

PuneetKumar1790:
## Summary

This PR normalizes pip dependency versions before constructing PURLs so that different textual representations of the same version (e.g. `1.0`, `1.0.0`, `1.00`, `01.0`) no longer produce distinct PURLs and duplicate SBOM components.

## Details

- Update `_generate_purl_dependency` for `kind == "pypi"` to:
  - Use `canonicalize_version(package["version"])` for the PURL `version` field.
  - Fall back to the raw version and log a warning if `InvalidVersion` is raised.
- Leave VCS and URL dependencies unchanged (still use the raw `version` string).
- Adjust existing PURL unit tests to expect canonicalized versions in PyPI PURLs.
- Add a unit test that verifies all of `["1.0", "1.0.0", "1.00", "01.0"]` yield the same PURL.

This addresses the behavior described in #1406, where SBOM deduplication based on `Component.key()` (PURL) treated these variants as different components.

## Testing

- `pytest tests/unit/package_managers/pip/test_main.py`
- `pytest tests/unit`
- (Full `pytest` run also performed; integration errors are unrelated to this change and due to missing integration environment.)

### DIFF
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -10,7 +10,8 @@ from urllib import parse as urlparse
 
 import pypi_simple
 from packageurl import PackageURL
-from packaging.utils import canonicalize_name
+from packaging.utils import canonicalize_name, canonicalize_version
+from packaging.version import InvalidVersion
 
 from hermeto.core.checksum import ChecksumInfo, must_match_any_checksum
 from hermeto.core.config import get_config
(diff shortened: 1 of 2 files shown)


hermeto/core/package_managers/pip/main.py
```diff
@@ -156,7 +157,16 @@ def _generate_purl_dependency(package: dict[str, Any]) -> str:
     qualifiers: dict[str, str] | None = None
 
     if dependency_kind == "pypi":
-        version = package["version"]
+        raw_version = package["version"]
+        try:
+            version = canonicalize_version(raw_version)
```
eskultety, line 162: This must use `strip_trailing_zero=False` (https://packaging.pypa.io/en/stable/utils.html#packaging.utils.canonicalize_version) otherwise it'll normalize all valid semver strings ending with a 0, e.g. `1.0.0` becomes 1 which would affect all our pip backend PURLs as indicated in the unit test changes. That is I think unacceptable for us.

eskultety (changes requested):
The linked issue is IMO quite hypothetical because tools like `pip-compile` and similar would not generate the mentioned version inconsistencies, that would have to be done manually and hence I'm inclined to reject this PR. That said, if `canonicalize_version` is used correctly this could bring **marginal only** improvement.

eskultety:
Closing due to inactivity.

PR #1413, opened 2026-03-15
fix(cargo): handle workspace-inherited versions in PURL generation

notjivi:
# Summary
Fixed a fatal error where the Cargo backend crashed when encountering workspace-inherited versions during PURL generation.

**Workspace Crash (ValueError: Invalid purl)**
* **The issue:** Modern Rust monorepos frequently use `version.workspace = true`. `tomlkit` parses this as a dictionary (`{'workspace': True}`) instead of a string. Passing this dictionary directly to the PURL generator causes a `ValueError` and crashes Hermeto.
* **The fix:** Updated the parser logic to safely detect dictionary-valued versions using `isinstance(version, dict)` and correctly fall back to the actual version defined in the `[workspace.package]` root (or handle missing/orphaned versions gracefully).

## Steps to Reproduce 

```bash
mkdir -p workspace-crash/src
cd workspace-crash

cat << 'EOF' > Cargo.toml
[workspace]
[workspace.package]
version = "2.0.0"

[package]
name = "workspace-crash"
version.workspace = true
edition = "2021"

[dependencies]
#Remote dependency included to bypass the KeyError and isolate the PURL crash
log = "0.4"
EOF

echo 'fn main() {}' > src/main.rs
cargo generate-lockfile
hermeto fetch-deps --source . --output ./hermeto-output cargo
```

Output:
image

### DIFF
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -218,17 +218,16 @@ def _resolve_main_package(package_dir: RootedPath) -> tuple[str, str | None]:
     parsed_toml = _parse_toml_project_file(package_dir.path / "Cargo.toml")
 
     package_info = parsed_toml.get("package", {})
-    workspace_info = parsed_toml.get("workspace", {})
+    workspace_package_info = parsed_toml.get("workspace", {}).get("package", {})
 
     # use default values if the project is a virtual workspace without any package information
     name = package_info.get("name", package_dir.path.stem)
-    version = package_info.get("version", None)
+    version = package_info.get("version")
 
     # check for a workspace package version
     # https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
-    if version is None:
-        version = workspace_info.get("package", {}).get("version")
-
+    if isinstance(version, dict) or version is None:
+        version = workspace_package_info.get("version")
     return name, version
 
 
--- a/tests/unit/package_managers/cargo/test_main.py
+++ b/tests/unit/package_managers/cargo/test_main.py
@@ -34,22 +34,61 @@ def test_standard_package_with_name_and_version(rooted_tmp_path: RootedPath) -> 
     assert version == "1.2.3"
 
 
-def test_virtual_workspace_with_workspace_package_version(rooted_tmp_path: RootedPath) -> None:
-    write_cargo_toml(
-        rooted_tmp_path,
-        """
-        [workspace]
-        members = ["a", "b", "c"]
+@pytest.mark.parametrize(
+    "toml_content, expected_name, expected_version",
+    [
+        pytest.param(
+            """
+            [workspace]
+            members = ["a", "b", "c"]
 
-        [workspace.package]
-        version = "1.2.3"
-        """,
-    )
+            [workspace.package]
+            version = "1.2.3"
+            """,
+            None,
+            "1.2.3",
+            id="workspace_package_version",
+        ),
+        pytest.param(
+            """
+            [workspace.package]
+            version = "2.5.0"
 
-    expected_name = rooted_tmp_path.path.name
+            [package]
+            name = "inherited-pkg"
+            version.workspace = true
+            """,
+            "inherited-pkg",
+            "2.5.0",
+            id="workspace_inheritance_shorthand",
+        ),
+        pytest.param(
+            """
+            [workspace]
+            members = ["a"]
+
+            [package]
+            name = "orphaned-pkg"
+            version.workspace = true
+            """,
+            "orphaned-pkg",
+            None,
+            id="workspace_inheritance_missing_root_version",
+        ),
+    ],
+)
+def test_virtual_workspace_with_workspace_package_version(
+    rooted_tmp_path: RootedPath,
+    toml_content: str,
+    expected_name: str | None,
+    expected_version: str | None,
+) -> None:
+    write_cargo_toml(rooted_tmp_path, toml_content)
+
     name, version = _resolve_main_package(rooted_tmp_path)
-    assert name == expected_name
-    assert version == "1.2.3"
+
+    assert name == (expected_name or rooted_tmp_path.path.name)
+    assert version == expected_version
 
 
 def test_virtual_workspace_without_workspace_version(rooted_tmp_path: RootedPath) -> None:


a-ovchinnikov:
A few things to note here. First, please do no bundle multiple unrelated fixes in one commit (one PR is fine though). Second, have you seen these crashes in actual projects which someone might decide to depend upon or build? Especially the first one. Third, the second issue needs a bit more details: why can it happen and why haven't it been seen yet?

notjivi:
>have you seen these crashes in actual projects which someone might decide to depend upon or build?

I reproduced the second one against https://github.com/drone-os/drone and it was throwing the same error

image
image


>the second issue needs a bit more details: why can it happen 

since rust 1.64 was released, more and more monorepos have started to use `version.workspace=true` to keep versions in sync across many crates.
what tomlkit does is, when it sees `version.workspace= true` it returns a dictionary `{'workspace' : True}` and that gets passed on as it is to the purl generator, resulting in a ValueError

>and why haven't it been seen yet?

Most older or smaller rust projects still use version strings like how it was before 1.64. This is a relatively newer approach to use 'inherited versioning' and is becoming a standard (like in https://github.com/drone-os/drone or https://github.com/valence-rs/valence)


>Especially the first one.

I agree that it is an edge case for manual use. I'm mainly trying to ensure consistency across the backends.
Since other backends in Hermeto can handle empty states without crashing, I thought it would be a good idea to make the Cargo one do the same (I looked into the Pip backend to ensure this aligns with Hermeto's internal standards). This way, if it’s ever part of an automated pipeline (like in Konflux), it won't break the build just because a repository happens to be minimal or standard-library only.

notjivi:
@a-ovchinnikov Hi, I've updated the PR to split the fixes into separate commits and added the unit tests. I’ve also provided the explanation above for why `version.workspace = true` is becoming more prevalent in modern Rust monorepos (like in https://github.com/drone-os/drone or https://github.com/valence-rs/valence) and how Hermeto breaks in a relevant scenario. Ready for another look whenever you have the time.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -466,8 +469,13 @@ def _generate_sbom_components(package_dir: RootedPath) -> list[Component]:
...
 def _swap_sources_directory_for_subsitution_slot(template: str) -> dict:
     toml_template = tomlkit.parse(template).value
+
+    if "source" not in toml_template:
+        return toml_template
+
     # Absolute path has to be replaced with relative path for sources relocation to work:
-    toml_template["source"]["vendored-sources"]["directory"] = "${output_dir}/deps/cargo"
+    if "vendored-sources" in toml_template["source"]:
+        toml_template["source"]["vendored-sources"]["directory"] = "${output_dir}/deps/cargo"
```
slimreaper35: > Second, have you seen these crashes in actual projects which someone might decide to depend upon or build?

Not answered yet.

hermeto/core/package_managers/cargo/main.py
slimreaper35, line 231:
```diff
     package_info = parsed_toml.get("package", {})
-    workspace_info = parsed_toml.get("workspace", {})
+    workspace_package_info = parsed_toml.get("workspace", {}).get("package", {})
 
     # use default values if the project is a virtual workspace without any package information
     name = package_info.get("name", package_dir.path.stem)
@@ -226,8 +226,8 @@ def _resolve_main_package(package_dir: RootedPath) -> tuple[str, str | None]:
 
     # check for a workspace package version
     # https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
-    if version is None:
-        version = workspace_info.get("package", {}).get("version")
+    if version == {"workspace": True}:
+        version = workspace_package_info.get("version", None)
 
     return name, version
```

hermeto/core/package_managers/cargo/main.py
notjivi: @slimreaper35 I actually replied to this in my comment above to a-ovchinnikov last week, but GitHub might have collapsed the thread. I'll quickly recap those responses for you:

For the Workspace crash : Yes, this actively breaks real-world projects. I ran hermeto against [drone-os/drone](https://github.com/drone-os/drone) and [valence-rs/valence](https://github.com/valence-rs/valence) and it crashed in both cases. You can find the verification for the drone-os in my previous comment [here](https://github.com/hermetoproject/hermeto/pull/1413#issuecomment-4065344182). Modern Rust monorepos are increasingly using version.workspace = true, which is parsed by tomlkit parses as a dictionary instead of a string, crashing the PURL generator.

For the Zero-Deps crash (Issue 1): No, I haven't seen this one in an actual upstream project. I triggered it locally while boundary-testing to ensure consistency across Hermeto's backends. Since other backends (like Pip) handle empty states gracefully, I wanted to ensure the Cargo backend wouldn't randomly crash an automated pipeline (like Konflux) if it ever comes across a minimal or standard-library-only repo.

That being said, if the team prefers not to merge fixes for theoretical edge cases, I am completely happy to drop the zero-deps commit and just proceed with the workspace fix! Let me know what you prefer.

hermeto/core/package_managers/cargo/main.py
notjivi: Thank you for the suggestion, I'll apply these changes to the branch

hermeto/core/package_managers/cargo/main.py
slimreaper35: > For the Workspace crash : Yes, this actively breaks real-world projects. I ran hermeto against [drone-os/drone](https://github.com/drone-os/drone) and [valence-rs/valence](https://github.com/valence-rs/valence)

Yes, I reproduced it. That's fine.

> For the Zero-Deps crash (Issue 1): No, I haven't seen this one in an actual upstream project.

That's what I was asking. Because why would someone try to use Hermeto with their project that has no dependencies? When I run `cargo vendor` in an empty cargo project, I see: `There is no dependency to vendor in this project.`

notjivi: >Because why would someone try to use Hermeto with their project that has no dependencies?

Agreed, it is an edge case and I haven't seen it in a major upstream project yet. My goal wasn't to say that it's common.

But since Hermeto's other backends can handle empty states without crashing, shouldn't we have the Cargo backend do the same in case if Hermeto ever hits a standard-library-only repo in an automated run? This will make sure that it returns a clean empty result instead of a python error.

That said, if the team prefers to keep the code focused strictly on proven real-world crashes, I’m happy to drop the zero-deps commit. I'll follow your lead on this.

notjivi:
Hi, @slimreaper35 
I updated the PR based on your feedback, applied the suggested changes for workspace inheritance, dropped the zero dependency check logic and resolved merge conflicts. Ready for another look.

notjivi:
`version == {"workspace": True}` triggered an error in the workspace tests where version is None so I updated the logic to if `version == {"workspace": True} or version is None` to cover both cases and verified the tests locally. Ready for review

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -284,3 +284,22 @@ def test_use_vendored_sources(
 
     assert result_toml["source"]["crates-io"]["replace-with"] == "vendored-sources"
     assert result_toml["source"]["vendored-sources"]["directory"] == "${output_dir}/deps/cargo"
+
+
+def test_resolve_main_package_with_workspace_inheritance(rooted_tmp_path: RootedPath) -> None:
```
slimreaper35: @notjivi Can you merge this unit test with the one above?
https://github.com/notjivi/hermeto/blob/13f95fb2cf2d0a6cbd4312d548b8aa8883f68d31/tests/unit/package_managers/cargo/test_main.py#L37 and use `pytest.param` with ids

hermeto/core/package_managers/cargo/main.py
slimreaper35: I am looking at the [Cargo.toml](https://github.com/drone-os/drone/blob/master/Cargo.toml) in the project you mentioned. The project is defined as a workspace, but in the same file, I see a project section that inherits keys from the workspace, which is different from cargo [documentation](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table), which was my source when implementing this function.

tests/unit/package_managers/cargo/test_main.py
notjivi: Done, merged the tests using `pytest.param` with ids, verified locally and fixed the linting. Ready for another look.

slimreaper35 (approved):

hermeto/core/package_managers/cargo/main.py
```diff
@@ -218,16 +218,16 @@ def _resolve_main_package(package_dir: RootedPath) -> tuple[str, str | None]:
...
 
     # use default values if the project is a virtual workspace without any package information
     name = package_info.get("name", package_dir.path.stem)
     version = package_info.get("version", None)
 
     # check for a workspace package version
     # https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table
-    if version is None:
-        version = workspace_info.get("package", {}).get("version")
+    if version == {"workspace": True} or version is None:
```
derasdf: I think `isinstance(version, dict)` would be safer here than matching a specific dict shape. This file already does the same thing at L410 for dep_info:

```python
if isinstance(version, dict) or version is None:
    version = workspace_package_info.get("version")
```

If Cargo ever adds another key to that table, the equality check silently falls through and you get a dict leaking into the PURL again aka the same class of bug this PR is fixing.

Also for `.get("version", None)` the `None` is redundant, `.get()` returns `None` by default.

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -34,22 +34,50 @@ def test_standard_package_with_name_and_version(rooted_tmp_path: RootedPath) ->
...
+    rooted_tmp_path: RootedPath,
+    toml_content: str,
+    expected_name: str | None,
+    expected_version: str,
+) -> None:
+    write_cargo_toml(rooted_tmp_path, toml_content)
 
-    expected_name = rooted_tmp_path.path.name
     name, version = _resolve_main_package(rooted_tmp_path)
+    if expected_name is None:
```
derasdf: nitpick: the `if expected_name is None: expected_name = ...` after the function call reads a bit oddly. Could simplify to:

```python
assert name == (expected_name or rooted_tmp_path.path.name)
```

Or just keep the two scenarios as separate test functions since they exercise different code paths (no `[package]` vs dict-valued version).

derasdf:
**Commit message:** the subject is good but the body is empty (just DCO). For a bug fix, please add the WHY into the commit body. tomlkit parses `version.workspace = true` as `{"workspace": True}` instead of a string, which crashes PURL generation.

PR description still mentions both fixes (zero-deps + workspace) but the zero-deps fix was dropped. Worth updating so it matches what's actually here.

Also, what happens when `version.workspace = true` but there's no `[workspace.package]` version? You get `None` back, same as the virtual workspace case. That's fine, but worth a test to document the contract:

```toml
[workspace]
members = ["a"]

[package]
name = "orphaned-pkg"
version.workspace = true
```

notjivi:
I've updated the commit, updated the PR description, updated the logic and added a test case for the orphaned workspace version scenario as suggested. 
Thank you for the detailed feedback!

Issue #1404, opened 2026-03-13
[github] Lint our workflows with an action linter


eskultety:
While there are many action linters out there on the GH marketplace, we should be very careful about adopting 3rd party stuff due to many projects simply being an "abandonware" (that's the inherent toll of open source).

The investigation should start with e.g. https://github.com/rhysd/actionlint and consider other alternatives out there. Part of this should be a comparison of possible approaches for us to make an educated decision and hence adopt the best solution at hand.

Kasturijadhav14:
Hi, I would like to work on this issue.

I read the description and understood that we need to explore action linters (like actionlint) and compare options before choosing one. I’m new to this but really interested in learning how GitHub Actions workflows are linted.

I can start by checking actionlint and a few alternatives and share what I find. Please let me know if that sounds good or if I should follow a specific direction.

PR #1416, opened 2026-03-16
Add GitHub Actions workflow linting using actionlint

MonishPuttu:
Fixes: #1404

I looked into a few options for linting GitHub Actions workflows, including Super-Linter, zizmor, and actionlint.
Super-Linter felt too heavy for this use case, and zizmor focuses more on security analysis than workflow linting. Since the issue already suggested actionlint, and it is lightweight, actively maintained, and purpose-built for GitHub Actions, it seemed like the best one I could find.

I implemented a workflow that installs and runs actionlint, the workflow downloads the official binary directly instead of relying on additional wrapper actions.

I tested this locally by running actionlint against the existing workflows to ensure they pass, and also verified that it correctly detects errors using a deliberately broken workflow file.

I Hope that this satisfies the requirements of the issue by adding lightweight CI linting for GitHub Actions workflows while keeping the dependency surface small.

### DIFF
--- a/.github/workflows/actionlint.yaml
+++ b/.github/workflows/actionlint.yaml
@@ -1,0 +1,17 @@
+name: Lint GitHub Actions
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/**"
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Run actionlint
+        uses: docker://rhysd/actionlint:1.7.11
+        with:
+          args: -color
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/setup-python@v6
         with:
           python-version: "3.10"
-      - run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV
+      - run: echo "cache_id=$(date --utc '+%V')" >> "$GITHUB_ENV"
       - uses: actions/cache@v5
         with:
           key: mkdocs-material-${{ env.cache_id }}
@@ -34,9 +34,9 @@ jobs:
         run: |
           LATEST_TAG=$(gh release view --json tagName -q .tagName)
           if [ "$LATEST_TAG" = "${{ github.event.release.tag_name }}" ]; then
-            echo "is_latest=true" >> $GITHUB_OUTPUT
+            echo "is_latest=true" >> "$GITHUB_OUTPUT"
           else
-            echo "is_latest=false" >> $GITHUB_OUTPUT
+            echo "is_latest=false" >> "$GITHUB_OUTPUT"
           fi
       - run: mkdocs gh-deploy --force --strict
         if: steps.latest.outputs.is_latest == 'true'
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -20,28 +20,28 @@ jobs:
       image: python:${{ matrix.python-version }}-slim
 
     steps:
-    - name: Install dependencies
-      run: |
-        # We need to install git inside the container otherwise the checkout action will use Git
-        # REST API and the .git directory won't be present which fails due to setuptools-scm
-        apt-get update && apt-get install --no-install-recommends --no-install-suggests -y git
-        pip install --upgrade pip nox
+      - name: Install dependencies
+        run: |
+          # We need to install git inside the container otherwise the checkout action will use Git
+          # REST API and the .git directory won't be present which fails due to setuptools-scm
+          apt-get update && apt-get install --no-install-recommends --no-install-suggests -y git
+          pip install --upgrade pip nox
 
-    - uses: actions/checkout@v6
-      with:
-        fetch-depth: 0
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
 
-    - name: Test with nox
-      run: |
-        # Disable Git's safe.directory mechanism as some unit tests do clone repositories
-        git config --global --add safe.directory '*'
-        nox -s python-${{ matrix.python-version }}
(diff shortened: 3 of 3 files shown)


victor-adebowale-momodu:
Hey @MonishPuttu! I’m also looking to contribute to Hermeto and have been diving into this issue. I would love to offer a few suggestions on your implementation to help make it more robust.

First, I want to mention the `bash` installation method.

> I implemented a workflow that installs and runs actionlint, the workflow downloads the official binary directly instead of relying on additional wrapper actions.

While I understand the goal is to keep the setup lightweight, there are issues that can arise from this approach:

1. Because the script pulls the "latest" version, any breaking change or update to actionlint is executed immediately. This could cause the CI to fail unexpectedly even when no changes were made to our actual code.
2. Also, relying on a remote script is generally discouraged in CI/CD. It risks executing code whose effect you are not in control of.

I would suggest that you use the official wrapper and pin it to a specific version (or commit hash), so that any decision to upgrade or change is explicit.

MonishPuttu:
@victor-adebowale-momodu, really appreciate the feedback

You’re right on both points. I originally went with the bash install approach to avoid introducing a third-party action wrapper, especially since the issue had concerns about abandonware and external dependencies. Piping a remote script into bash without pinning a version is definitely an anti-pattern, and I should have seen that.

I’ll switch this over to the official Docker image pinned to a specific version. That would keep things simple and avoids the remote script issue and makes version updates explicit.

I’ll push a fix shortly

MonishPuttu:
> I would suggest that you use the official wrapper and pin it to a specific version (or commit hash), so that any decision to upgrade or change is explicit.

I was waiting for the maintainer's pov before making any changes

> I’ll switch this over to the official Docker image pinned to a specific version. That would keep things simple and avoids the remote script issue and makes version updates explicit.

For now I will go with this approach

.github/workflows/actionlint.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - ".github/workflows/**"
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Run actionlint
+        uses: docker://rhysd/actionlint:1.7.11
```
eskultety, line 15: Not a particular fan of pulling fixed image versions without monitoring this with robots like dependabot, but we'll need to think of a better way of doing this. Theoretically we could consume latest, but at this point I'm not sure we can trust the stability of these, so let's keep the version fixed.

.github/workflows/actionlint.yaml
```diff
@@ -0,0 +1,21 @@
+name: Lint GitHub Actions workflows
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/**"
+  push:
+    branches: [main]
+    paths:
+      - ".github/workflows/**"
```
eskultety: Why do we need to re-run the actionlint linter on every push to the main branch if we already ran it on a PR? Was this an AI suggestion?

.github/workflows/actionlint.yaml
MonishPuttu: I will keep the version pinned for now. I agree it would be better to have something like Dependabot for keeping it up to date over time, so it doesn’t go stale, if you want I can probably take that up as a follow up.

.github/workflows/actionlint.yaml
MonishPuttu: my initial thought was to have it in place if something gets pushed directly to main without going through a PR. that is something I usually include while writing workflows, but you are right in a normal workflow where everything goes through PRs, this is redundant.

eskultety: Well, the industry standard in real projects the common security pattern is to have a branch protection rule which I think is the default now for all new GH projects, that's why I asked because I found it redundant.

eskultety:
@MonishPuttu some of the PR description should go directly to the commit message as well for reasoning purposes. Also, do you have it deployed on your fork so that we can see it in action? :)
LGTM otherwise.

eskultety:
>Also, do you have it deployed on your fork so that we can see it in action?

Oh, nevermind, it ran from the PR :D, so yeah, it works.

.github/workflows/actionlint.yaml
```diff
@@ -0,0 +1,17 @@
+name: Lint GitHub Actions workflows
```
eskultety: Let's drop the word 'workflows' from the name

MonishPuttu: I have updated the commit message to include the reasoning and dropped “workflows” from name.

MonishPuttu:
> Oh, nevermind, it ran from the PR :D, so yeah, it works.

Thank you, :D

eskultety:
@MonishPuttu Well, since the previous CI run failed, you will also need to provide a fix to the failures before we can merge instead of just applying our review comments :) .

MonishPuttu:
@eskultety actionlint was failing due to existing shellcheck issues in current workflows. should I update the workflow to ignore shellcheck for now so CI passes without expanding scope. 
Happy to follow up separately to address those.

eskultety:
> @eskultety actionlint was failing due to existing shellcheck issues in current workflows. should I update the workflow to ignore shellcheck for now so CI passes without expanding scope. Happy to follow up separately to address those.

@MonishPuttu I think there might be some misunderstanding still, so I re-run the CI for you so you can see it. I know the problems are in existing workflows but we cannot merge a PR with a failing CI (that's a hard repo settings check).

MonishPuttu:
@eskultety I misunderstood, thanks for clarifying. I will fix the shellcheck issues in the existing workflows so CI passes.

MonishPuttu:
I think I have addressed the remaining shellcheck issues in workflows.

MonishPuttu:
@eskultety Sorry about that, I missed updating the commit message earlier. I have updated it now to include the workflow fixes as well.

PR #1419, opened 2026-03-16
feat: log which requirements files are used during pip dependency resolution

### LINKED ISSUE #1418: feat: log which requirements files are used during dependency resolution
## Summary

Hermeto does not log which requirements files it picks up during dependency 
resolution. When a project does not use the default requirements.txt naming, 
hermeto fetches zero dependencies and reports success.

## Reproducing

Setup a project with requirements files under a non-default path:
```
myproject/
├── pyproject.toml
└── requirements/
    ├── default.txt
    └── build.txt
```

After running hermeto without specifying files:

```
INFO All dependencies fetched successfully 
```

Zero dependencies fetched. No warning. No indication why.

## Expected
```
INFO Using requirements files: []
INFO Using build requirements files: []
INFO All dependencies fetched successfully
```

Empty list immediately tells the user no requirements files were found.

## Proposed Fix
  ```
       resolved_req_files = resolve_req_files(requirement_files, False) 
+    log.info("Using requirements files: %s", [str(f.subpath_from_root) for f in resolved_req_files])

     resolved_build_req_files = resolve_req_files(build_requirement_files, True)
+    log.info("Using build requirements files: %s", [str(f.subpath_from_root) for f in resolved_build_req_files])
```

### PR DESCRIPTION
**Pull request by harshitbansal184507**
Log the resolved requirements and build requirements files at the start
of dependency resolution so it is immediately visible when no files
were found.

Resolves #1418

### DIFF
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -598,7 +598,22 @@ def _resolve_pip(
         return resolved
 
     resolved_req_files = resolve_req_files(requirement_files, False)
+    if not resolved_req_files:
+        log.warning("No requirements files found, no dependencies will be fetched")
+    else:
+        log.info(
+            "Using requirements files: %s",
+            ", ".join(str(f.subpath_from_root) for f in resolved_req_files),
+        )
+
     resolved_build_req_files = resolve_req_files(build_requirement_files, True)
+    if not resolved_build_req_files:
+        log.info("No build requirements files found")
+    else:
+        log.info(
+            "Using build requirements files: %s",
+            ", ".join(str(f.subpath_from_root) for f in resolved_build_req_files),
+        )
 
     requires = _download_from_requirement_files(output_dir, resolved_req_files, binary_filters)
     build_requires = _download_from_requirement_files(

### CODE REVIEW

File: hermeto/core/package_managers/pip/main.py [hunk 1]
```diff
@@ -571,7 +599,19 @@ def resolve_req_files(req_files: list[Path] | None, devel: bool) -> list[RootedP
...
     resolved_req_files = resolve_req_files(requirement_files, False)
+    if not resolved_req_files and requirement_files is None:
+        log.warning("No requirements files found, no dependencies will be fetched")
+    else:
+        log.info(
+            "Using requirements files: %s", [str(f.subpath_from_root) for f in resolved_req_files]
+        )
+
     resolved_build_req_files = resolve_req_files(build_requirement_files, True)
+    if not resolved_build_req_files and build_requirement_files is None:
```
a-ovchinnikov: `not resolved_build_req_files and build_requirement_files is None` resolves to True when `build_requirement_files` is `None` and also when `resolved_build_req_files` is an object with boolean equivalence of `False`, e.g. an empty list. Iterating over an empty list produces an empty list on L613. Are you sure you want to log an empty list only when there is an empty list on the input?

hermeto/core/package_managers/pip/main.py
```diff
@@ -571,7 +599,18 @@ def resolve_req_files(req_files: list[Path] | None, devel: bool) -> list[RootedP
...
 
     resolved_req_files = resolve_req_files(requirement_files, False)
+    if not resolved_req_files:
+        log.warning("No requirements files found, no dependencies will be fetched")
+    else:
+        log.info(
+            "Using requirements files: %s", [str(f.subpath_from_root) for f in resolved_req_files]
+        )
+
     resolved_build_req_files = resolve_req_files(build_requirement_files, True)
```
a-ovchinnikov, line 609: Why did you decide to drop the check?

harshitbansal184507: that condition is leftover from an earlier version where I was warning for both runtime and build requirements. I felt there no need to warn against empty build requirements , but forgot to clean up . for dropped check , resolve_req_files() already handles it, when requirement_files is None and the default file does not 
exist it returns an empty list, so checking not resolved_req_files alone is sufficient.

hermeto/core/package_managers/pip/main.py
```diff
@@ -571,7 +599,18 @@ def resolve_req_files(req_files: list[Path] | None, devel: bool) -> list[RootedP
...
+    log.info(
+        "Using build requirements files: %s",
+        [str(f.subpath_from_root) for f in resolved_build_req_files],
+    )
```
ST3V1K: > I felt there no need to warn against empty build requirements.

I agree with not warning about it, but I'd still log a different message, at an info level, when no build requirements files are found. I feel like it's weird to log an empty list.

harshitbansal184507: When a user specifies requirements files explicitly (e.g. requirements-extra.txt), the default requirements files are skipped. Am I just overthinking scenarios here, or is this a case worth handling?

ST3V1K: I think this is the intended behavior. When the user wants only `requirements-extra.txt`, he can specify that. If he wants `requirements-extra.txt` and the default one, he would have to specify both. In both cases you logic will only report the specified files, which is IMO correct.

hermeto/core/package_managers/pip/main.py
```diff
@@ -598,7 +598,21 @@ def resolve_req_files(req_files: list[Path] | None, devel: bool) -> list[RootedP
         return resolved
 
     resolved_req_files = resolve_req_files(requirement_files, False)
+    if not resolved_req_files:
+        log.warning("No requirements files found, no dependencies will be fetched")
+    else:
```
derasdf, line 603: The list-as-%s produces `['requirements.txt']` in the log. Rest of the pip module uses `', '.join()` . I'd match that:

```python
log.info(
    "Using requirements files: %s",
    ", ".join(str(f.subpath_from_root) for f in resolved_req_files),
)
```

Same for the build requirements log. Otherwise LGTM.

derasdf (approved):

eskultety (approved):

Issue #1389, opened 2026-03-12
Any docs on why `npm ci --ignore-scripts` isn't used to fetch npm deps ?


gsaslis:
Hello there folks, 

I am trying to find some backing documentation as to why the current hermeto [implementation](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/npm.py) doesn't use npm itself (e.g. something like `npm ci --ignore-scripts`) but rather prefers python code to parse the dependencies. 

Was there an arbitrary code execution concern with that, or was that because of other concerns (e.g. SBOM generation, etc) ?

Any chance there's a public doc somewhere that might explain that ? 

Thanks in advance! 🙏

brunoapimentel:
I don't think we have any public docs on the npm design, unfortunately 😞 

I was checking some notes we made while implementing it, and I think we can trace back the decision to these factors:
- `npm install` will generate the cache, but we still need to figure out what went into it to generate the SBOM. That means, we'd either have to parse the lockfile anyways, or have some way to get the package info from the cache (using [cacache](https://www.npmjs.com/package/cacache) for cache operations was considered at the time).
- npm doc's clearly [state](https://docs.npmjs.com/cli/v11/commands/npm-cache#a-note-about-the-caches-design) that the cache is not something to be relied as a persistent data store. We didn't want to tie our solution to something that could eventually change and break.

chmeliik:
>  * npm doc's clearly [state](https://docs.npmjs.com/cli/v11/commands/npm-cache#a-note-about-the-caches-design) that the cache is not something to be relied as a persistent data store. We didn't want to tie our solution to something that could eventually change and break.

Also note that Hermeto would be running `npm install` with a different version of `npm` than the user's containerfile, which makes the unreliability of the cache drastically worse.

And back when we implemented it, the cache didn't support some types of dependencies at all (e.g. git dependencies)

gsaslis:
Thanks both! Really appreciate the prompt responses! 

Always so helpful. 🙇🏼 🙏
 
I was primarily interested in whether arbitrary code execution was a concern and it sounds like that wasn't the case here (as it was with other package managers that I recall <cough>bundler</bundler> discussing with you all).  I'll go ahead and close this issue. 👌

PR #1420, opened 2026-03-16
📝 Add pnpm design document

### LINKED ISSUE #1251: Support for pnpm package manager
Currently, Hermeto supports `npm` (via `package-lock.json`) and `yarn` (via `yarn.lock`), but it lacks support for **pnpm**.

Many modern JavaScript/TypeScript projects utilize pnpm for its efficiency and strict dependency management. Users currently have to maintain a separate `package-lock.json` or `yarn.lock` solely for Hermeto compatibility, or temporarily switch package managers during the build process. This adds friction to the workflow and negates the benefits of using pnpm in the development cycle.

I would like Hermeto to add native support for `pnpm` by parsing `pnpm-lock.yaml`, cause [OCP Console](https://github.com/openshift/console) needs to do this migration.

The implementation should:
1. Detect `pnpm-lock.yaml` in the project root.
2. Parse the lockfile to identify all dependency URLs and checksums (similar to how it handles `package-lock.json`).
3. Pre-fetch these artifacts into the `hermeto` cache to enable hermetic, offline builds.

### Additional context
Support for pnpm is critical for modern tooling stacks due to significant advantages over npm and Yarn:

**1. Speed & Efficiency (The "Hard Links" Advantage)**
* **Disk Space:** Unlike npm/Yarn, which copy files for every project, pnpm uses a content-addressable store. Dependencies are stored once on the disk and hard-linked into projects.
* **Install Speed:** Benchmarks consistently show pnpm is significantly faster than npm and Yarn (often 2-3x faster) because it avoids duplicating IO operations.

**2. Security & Correctness (Phantom Dependencies)**
* **Strict Isolation:** pnpm uses a non-flat `node_modules` structure. It prevents "phantom dependencies" (accessing packages that are not listed in `package.json` but happen to be hoisted by the package manager).
* **Supply Chain Security:** By enforcing strict dependency trees, pnpm ensures that the build only has access to exactly what was declared, aligning perfectly with Hermeto's goal of reproducible, secure builds.

Adding support for `pnpm-lock.yaml` would make Hermeto viable for a much larger segment of the modern JS ecosystem.

slimreaper35:
Closes https://github.com/hermetoproject/hermeto/issues/1251

### DIFF
--- a/docs/design/pnpm.md
+++ b/docs/design/pnpm.md
@@ -1,0 +1,516 @@
+# pnpm
+
+## Background
+
+This design document introduces support for the pnpm package manager in Hermeto.
+pnpm is a fast, disk space-efficient package manager for JavaScript and TypeScript.
+It is a great candidate for Hermeto in addition to `npm` and `yarn berry` since
+many projects are slowly migrating from ancient `yarn classic` to a different
+package manager, and pnpm's popularity has been growing [exponentially](https://npmtrends.com/pnpm)
+in the last couple of years.
+
+## Project structure
+
+A simple pnpm project could have the following structure:
+
+```text
+my-pnpm-project
+  ├── node_modules
+  ├── pnpm-lock.yaml
+  ├── package.json
+  └── src/
+      └── index.js
+```
+
+pnpm projects have a `package.json` file with an additional `packageManager` field
+that specifies the package manager and version used to install the dependencies.
+
+```json
+{
+  "name": "foo",
+  "version": "1.0.0",
+  "packageManager": "pnpm@x.y.z"
+}
+```
+
+## Features
+
+### [Symlinked node_modules structure](https://pnpm.io/symlinked-node-modules-structure)
+
+pnpm uses a unique hard links and symbolic links approach to organize `node_modules`,
+which is very different from the traditional flat npm structure. All package files
+are stored internally only once in the local file system.
+
+Only direct dependencies of a project appear as top-level directories inside
+`node_modules`. By using this approach, it is impossible to use modules that are
+not specified in the `package.json` file.
+
+### [Package sources](https://pnpm.io/package-sources)
+
+pnpm uses the [npm registry](https://www.npmjs.com) by default, but also supports
+other package sources:
+
+- [JSR registry](https://jsr.io) (JavaScript Registry)
+- workspace packages
+- local file system
+- remote tarballs
+- git repositories
+
+### [Workspaces](https://pnpm.io/workspaces)
+
+pnpm supports monorepos via the `pnpm-workspace.yaml` file. A simple workspace
+could have the following structure:
+
+```text
+my-pnpm-workspace
+  ├── package.json
+  ├── pnpm-lock.yaml
+  ├── pnpm-workspace.yaml
+  └── packages/
+      └── my-pkg/
+          └── package.json
+```
+
+`pnpm-workspace.yaml` defines the root of the workspace and enables the
+inclusion/exclusion of directories from the workspace.
+
+## [Configuration](https://pnpm.io/npmrc)
+
+pnpm uses the same configuration format as npm. All settings can be configured
+via `.npmrc` files. Naturally, respective environment variables for each configuration
+option take precedence over the `.npmrc` file.
+
+Authentication settings can be handled via Hermeto configuration
+(YAML or environment variables). See the [Proxy support](#proxy-support) section.
+Further support for other configuration options is out of scope for now.
+
+## Lockfile format
+
+The format is versioned. The latest version is 9.0, which was introduced in
+[pnpm v9.0.0](https://github.com/pnpm/pnpm/releases/tag/v9.0.0). The pnpm GitHub
+organization occasionally maintains the various specifications in the [pnpm/spec](https://github.com/pnpm/spec)
+GitHub repository. The latest commit (at the time of writing) from 2024 points to
+a briefly described format of the lockfile version 9.0 - [https://github.com/pnpm/spec/blob/master/lockfile/9.0.md](https://github.com/pnpm/spec/blob/master/lockfile/9.0.md).
+
+Hermeto will target the lockfile version 9.0. The lockfile version does not seem
+to be directly tied to the pnpm version. pnpm major versions 10 or 11 work with
+version 9.0.
+
+The root keys and their order are defined in the main pnpm source repository:
+[sortLockfileKeys.ts](https://github.com/pnpm/pnpm/blob/main/lockfile/fs/src/sortLockfileKeys.ts).
+
+- `lockfileVersion` (ComVer string, not SemVer - `'9.0'`)
+- `settings` (metadata about how the lockfile was generated)
+- `patchedDependencies` (mapping of package selectors to patch file metadata)
+- `importers` (dependency declarations - specifiers and resolved versions)
+- `packages` (mapping of dependency ID to dependency object)
+- `snapshots` (dependency graphs and optional/peer metadata per package)
+
+<details>
+<summary>Example of a pnpm-lock.yaml file with a single dependency</summary>
+
+```yaml
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+  .:
+    dependencies:
+      lodash:
+        specifier: ^4.17.23
+        version: 4.17.23
+
+packages:
+  lodash@4.17.23:
+    resolution: {integrity: sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w==}
+
+snapshots:
+  lodash@4.17.23: {}
+```
+
+</details>
+
+## Implementation
+
+A brief overview of the workflow for Hermeto:
+
+1. Perform basic validation of the project files.
+2. Parse the `pnpm-lock.yaml` file and extract data from the `packages` section.
+3. Process all packages from the lockfile.
+4. Asynchronously download tarballs and verify their checksums.
+5. Generate SBOM components from the parsed data.
+6. Patch the lockfile for hermetic builds.
+
+The parsing of the lockfile will be the core of the implementation. Each non-registry
+package should have already resolved its URL. For registry packages, the URL must
+be constructed from the package scope, name, and version (`https://registry.npmjs.org/{scope}/-/{name}-{version}.tgz`).
+
+### Why not use the pnpm binary?
+
+Hermeto could run the [pnpm list](https://pnpm.io/cli/list) command and parse the
+JSON output from STDOUT.
+
+```shell
+pnpm list --lockfile-only --json --depth Infinity --no-color
+```
+
+This approach could be useful for resolving the dependency tree and generating
+SBOM components. Each entry from the JSON output has a name, version, and resolved
+URL. Transitive development, optional, and peer dependencies are included in that
+tree, which reduces custom parsing and surface area for errors.
+
+<details>
+<summary>Example of JSON output</summary>
+
+```jsonc
+[
+  {
+    "name": "foo",
+    "version": "1.0.0",
+    "path": "/path/to/foo",
+    "private": false,
+    "dependencies": {
+      "lodash": {
+        "from": "lodash",
+        "version": "4.17.23",
+        "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.23.tgz",
+        "path": ".pnpm/lodash@4.17.23/node_modules/lodash"
+      }
+    },
+    "devDependencies": {
+      "vue": {
+        "from": "vue",
+        "version": "3.5.29",
+        "resolved": "https://registry.npmjs.org/vue/-/vue-3.5.29.tgz",
+        "path": ".pnpm/vue@3.5.29/node_modules/vue",
+        "dependencies": {
+          "@vue/compiler-dom": {
+            "from": "@vue/compiler-dom",
+            "version": "3.5.29",
+            "resolved": "https://registry.npmjs.org/@vue/compiler-dom/-/compiler-dom-3.5.29.tgz",
+            "path": ".pnpm/@vue+compiler-dom@3.5.29/node_modules/@vue/compiler-dom"
+          }
+          // more dependencies...
+        }
+      }
+    }
+  }
+]
+```
+
+</details>
+
+However, there are several reasons why using the pnpm binary outweighs its benefits:
+
+1. `pnpm list` output JSON does not include the `resolution.integrity` (SRI) values
+stored in the lockfile. Hermeto needs them to verify downloaded tarballs.
+
+2. JSON output size - on a large workspace, `pnpm list` with full depth can produce
+enormous JSON (from local testing, a lockfile with tens of thousands of lines expands
+to millions of lines of output). This can lead to buffer overflows and other performance
+issues.
+
+3. Extra dependency in our image that the maintainers would have to work around.
+
+4. Running any pnpm command exposes the risk of arbitrary code execution in the
+form of pnpmfile hooks or scripts. These can be avoided by environment variables,
+but still, there is a risk that pnpm could introduce something else in the future.
+
+5. Hermeto has to parse the `pnpm-lock.yaml` anyway to substitute the resolved URLs
+with local files for offline build.
+
+### Why not use the node_modules directory instead?
+
+`node_modules` is whatever `pnpm install` produced. Optional and platform-specific
+packages differ, and the real files live under `.pnpm` with symlinks in `node_modules`,
+so walking the directory does not give a straightforward portable output of tarballs.
+
+There are no supported environment variables that relocate `node_modules` to a stable
+path Hermeto could rely on.
+
+### Optional dependencies
+
+These packages appear in the `importers` section in the `optionalDependencies`
+field and in the `snapshots` section with the `optional: true` flag.
+
+```yaml
+importers:
+  .:
+    optionalDependencies:
+      lodash:
+        specifier: ^4.17.23
+        version: 4.17.23
+
+snapshots:
+  lodash@4.17.23:
+    optional: true
+```
+
+### Development dependencies
+
+They appear in the `packages` section as regular dependencies.
+
+They can be determined by checking if a dependency is listed in the `devDependencies`
+field of a package.json file, though. Transitive development dependencies can be
+determined by parsing the lockfile `snapshots` section and applying the breadth-first
+search algorithm.
+
+```json
+  "devDependencies": {
+    "vue": "^3.5.29"
+  }
+```
+
+```yaml
+...
+snapshots:
+  vue@3.5.30:
+    dependencies:
+      '@vue/compiler-dom': 3.5.30
+      '@vue/compiler-sfc': 3.5.30
+      '@vue/runtime-dom': 3.5.30
+      '@vue/server-renderer': 3.5.30(vue@3.5.30)
+      '@vue/shared': 3.5.30
+```
+
+### Peer dependencies
+
+A package declares `peerDependencies` when it expects the host (or another
+dependency) to provide a given package, for example, a React plugin expects React.
+
+They appear only as metadata (`peerDependencies` on packages, `transitivePeerDependencies`
+in snapshots) and do not have their own entries in the `packages` section. When
+a peer is satisfied, the satisfying package appears in `packages` as a regular dependency.
+Hermeto will prefetch everything in the `packages` section, so no special handling
+is needed for peers.
+
+### Patched dependencies
+
+Patched dependencies allow applying custom modifications (patches) to a specific
+version of a dependency without forking the package or waiting for an upstream fix.
+They are declared in the top-level patchedDependencies key of the `pnpm-lock.yaml`
+file.
+
+```yaml
+patchedDependencies:
+  lodash@4.17.23:
+    hash: 5877a18891ec19fc2a2e4eaf39284abc77fe7f8a910708f424aac3434b0813aa
+    path: patches/lodash-4.17.23.patch
+```
+
+CycloneDX specification defines patches for a component as a list of `patch` objects.
+This approach is already used for the yarn berry backend. See [cyclonedx.org/docs](https://cyclonedx.org/docs/1.6/json/#metadata_tools_oneOf_i0_components_items_pedigree_patches).
+
+SPDX specification defines a similar way to report patches as `externalRefs` object.
+See [spdx-spec/v2.3](https://spdx.github.io/spdx-spec/v2.3/how-to-use/#k17-linking-to-a-code-fix-for-a-security-issue)
+
+### SBOM
+
+Components will be generated from the parsed packages. Each package will get
+a package URL per the [purl-spec](https://github.com/package-url/purl-spec/blob/main/types-doc/npm-definition.md).
+
+All components will use the `npm` type - `pkg:npm/namespace/name@version`. The namespace
+is empty for unscoped packages. Other qualifiers and properties vary depending
+on the package source. The PURLs are automatically encoded by the `python-packageurl`
+library using `to_string()` method on package URL objects without any additional
+configuration.
+
+The official CycloneDX property [cdx:npm:package:development](https://github.com/CycloneDX/cyclonedx-property-taxonomy/blob/main/cdx/npm.md)
+will be set for development dependencies and transitive development dependencies
+when detected.
+
+The `resolution.integrity` field in the dependency object is an SRI (Subresource
+Integrity) string used to verify the checksum of the tarball. Currently, Hermeto
+does not have a unified approach to checksum handling in the SBOM. See [hermetoproject/hermeto/issues/852](https://github.com/hermetoproject/hermeto/issues/852).
+Let's not add it to the SBOM for now. Removing it later could be considered as
+a breaking change.
+
+### npm registry
+
+```yaml
+packages:
+  # unscoped package
+  express@5.2.1:
+    resolution: {integrity: sha512-hIS4idWWai69NezIdRt2xFVofaF4j+6INOpJlVOLDO8zXGpUVEVzIYk12UUi2JzjEzWL3IOAxcTubgz9Po0yXw==}
+    engines: {node: '>= 18'}
+
+  # scoped package
+  '@vue/compiler-core@3.5.30':
+    resolution: {integrity: sha512-s3DfdZkcu/qExZ+td75015ljzHc6vE+30cFMGRPROYjqkroYI5NV2X1yAMX9UeyBNWB9MxCfPcsjpLS11nzkkw==}
+
+```
+
+**PURLs:**
+
+- `pkg:npm/express@5.2.1`
+- `pkg:npm/vue/compiler-core@3.5.30` (scoped package)
+
+### JSR registry
+
+```yaml
+packages:
+  '@jsr/hono__hono@4.12.4':
+    resolution: {integrity: sha512-fxOSxFkBlmt2dNAvz2vIj8UmihKrkiyjsvU/A8yvHtd9Pt53WBUxsRbpzl8GXodwA0/63gXGxDFDb7XXcPQXbA==, tarball: https://npm.jsr.io/~/11/@jsr/hono__hono/4.12.4.tgz}
+```
+
+JSR packages use the `@jsr/scope__name` naming convention in the lockfile.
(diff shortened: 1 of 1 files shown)


docs/design/pnpm.md
```diff
@@ -0,0 +1,535 @@
...
+        }
+      }
+    }
+  }
+]
+```
+
+</details>
+
+### Why not use the pnpm store instead?
```
a-ovchinnikov: Why not use the store when? During fetch or during build? If it is the latter then why not? Can we simulate the store?

docs/design/pnpm.md
```diff
@@ -0,0 +1,535 @@
...
+
+packages:
+
+  lodash@4.17.23:
+    resolution: {integrity: sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w==}
+```
+
+The example shows what a `pnpm-lock.yaml` file looks like, but a better approach
+would be to parse the `pnpm-workspace.yaml` or `package.json` files for information
+about the workspace packages using glob patterns. The `version` field could be
```
a-ovchinnikov: Won't this result in non-reproducible builds because package versions are not pinned to an exact version?

docs/design/pnpm.md
```diff
@@ -0,0 +1,535 @@
...
+  lodash@4.17.23:
+    resolution:
+      integrity: sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w==
+      tarball: file:///tmp/hermeto-output/deps/pnpm/lodash-4.17.23.tgz
+```
+
+### Proxy support
+
+Hermeto can use the same proxy configuration as the npm backend
+(`config.pnpm.proxy_url`, `config.pnpm.proxy_login`, `config.pnpm.proxy_password`).
```
a-ovchinnikov: Would this require any additional settings if we decide to use pnpm itself instead of providing a parser/downloader?

a-ovchinnikov (commented):
A few initial questions before a more thorough review.

docs/design/pnpm.md
slimreaper35: So for the workspaces, we could parse the `pnpm-workspace.yaml` file and retrieve all workspace paths. It can contain the following content.

```yaml
packages:
  - packages/*
  - other
```
That's why I mentioned globs. And for each workspace package, we would parse the respective `package.json` for information about name and version. I believe we use a similar approach for yarn classic now.

docs/design/pnpm.md
slimreaper35: We could probably just set env variables - https://pnpm.io/npmrc#proxy-settings

docs/design/pnpm.md
slimreaper35: > During fetch or during build?

Instead of manually downloading everything.

docs/design/pnpm.md
```diff
@@ -0,0 +1,519 @@
...
+
+    @classmethod
+    def from_file(cls, path: Path) -> "PnpmLock":
+        with path.open("r") as f:
+            data = yaml.safe_load(f)
+            return cls(path, data)
+
+```
+
+#### 2. pnpm binary
```
slimreaper35: I am not 100% about this part. But I didn't find any other threats except *.cjs files.

docs/design/pnpm.md
```diff
@@ -0,0 +1,519 @@
...
+
+The second option is to run the `pnpm list` command and parse the JSON output from
+STDOUT and add the pnpm binary from corepack to the container image.
+
+This approach could be extremely useful for downloading dependencies and generating
+SBOM components, though. Each entry from the JSON output has a `name`, `version`,
+and a URL. In addition, transitive development, optional, and peer dependencies
+are also resolved. Which means less custom parsing logic and less error-prone code.
+
+##### Arbitrary code execution
```
taylormadore: It looks like we can avoid `.pnpmfile.cjs` execution using env in pnpm v11 beta, but not in pnpm v10:
```
root@f0c72f188b90:/src# npm install -g pnpm@latest-10

added 1 package in 1s

root@f0c72f188b90:/src# pnpm_config_ignore_pnpmfile=true pnpm list
*** PNPMFILE EXECUTED — arbitrary code runs here ***
root@f0c72f188b90:/src# npm install -g pnpm@11.0.0-beta.2

changed 1 package in 966ms

root@f0c72f188b90:/src# pnpm_config_ignore_pnpmfile=true pnpm list
root@f0c72f188b90:/src# 
```

And it looks like we can do the same using `.npmrc` lin pnpm v10 but not pnpm v11 beta:
```
root@64cbcacfa1bc:/src# cat .npmrc 
ignore-pnpmfile=true
root@64cbcacfa1bc:/src# npm install -g pnpm@latest-10

added 1 package in 2s

root@64cbcacfa1bc:/src# pnpm list
root@64cbcacfa1bc:/src# 
root@64cbcacfa1bc:/src# npm install -g pnpm@11.0.0-beta.2

changed 1 package in 596ms

root@64cbcacfa1bc:/src# pnpm list
*** PNPMFILE EXECUTED — arbitrary code runs here ***
root@64cbcacfa1bc:/src# 
```

docs/design/pnpm.md
```diff
@@ -0,0 +1,519 @@
...
+### Why not use the node_modules directory instead?
+
+The [npm cache documentation](https://docs.npmjs.com/cli/v9/commands/npm-cache#a-note-about-the-caches-design)
+also explicitly states that the cache is not meant to be a "persistent and reliable
+data store". See discussion in [hermetoproject/hermeto#1389](https://github.com/hermetoproject/hermeto/issues/1389).
+
+In addition, there are no environment variables that can be used to configure the
+location of the node_modules directory.
+
+### Optional dependencies
```
taylormadore, line 234: Why skip prefetching optionalDependencies?

slimreaper35: Hmm, I thought we skip them in other backends, but apparently [not](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/npm.py#L39).

slimreaper35: Do we need to report them in the SBOM somehow?

eskultety: Yes, because once we download them, we have no control over what the build will actually use, it's like building with different flags (e.g. vim) resulting in different binaries. It will affect the accuracy though, but hey, that's why there's the notion of an analyzed SBOM which should be combined with our source SBOM.

slimreaper35: I don't think we have a precedent for optional dependencies in the SBOM.

taylormadore: I believe we download and report them in the SBOM like any other dependency. We have an integration test for it as well:
- https://github.com/hermetoproject/integration-tests/tree/npm/dev-optional-peer-deps
- https://github.com/hermetoproject/hermeto/blob/main/tests/integration/test_data/npm_dev_optional_peer_deps/bom.json

docs/design/pnpm.md
```diff
@@ -0,0 +1,519 @@
...
+Read the `pnpm-lock.yaml` file and extract data from the `packages` section.
+
+#### 2. pnpm binary
+
+Run the `pnpm list` command and parse the JSON output from STDOUT. Hermeto would
+use corepack to add the pnpm binary to our container image. Currently, pnpm requires
+Node.js v18.12 or higher to be installed on the system. See [https://pnpm.io/installation#compatibility](https://pnpm.io/installation#compatibility)
+
+This approach is useful for resolving the dependency tree and generating
+SBOM components. Each entry from the JSON output has a name, version, and URL.
```
taylormadore: Does `pnpm list` give us the integrity hashes?

docs/design/pnpm.md
```diff
@@ -0,0 +1,519 @@
...
+
+To ignore the pnpmfile, it is possible to set the [ignore-pnpmfile](https://pnpm.io/pnpmfile#ignorepnpmfile)
+option to `true` in the `.npmrc` file or via an environment variable - `npm_config_ignore_pnpmfile=true`.
+
+The pnpm configuration is loaded in [`config/config/src/index.ts`](https://github.com/pnpm/pnpm/blob/v10.33.0/config/config/src/index.ts#L242).
+Initial options are taken only from the `npmConfig.get()` function for every key,
+so settings such as `ignore-pnpmfile` must use npm's convention `npm_config_ignore_pnpmfile=true`.
+
+```shell
+npm_config_ignore_pnpmfile=true pnpm list --lockfile-only --json --depth Infinity --no-color
```
taylormadore: Noting that `--lockfile-only` only exists for pnpm versions 10.23.0 and later: https://pnpm.io/cli/list#--lockfile-only

docs/design/pnpm.md
```diff
@@ -0,0 +1,519 @@
...
+```
+
+#### pnpm v11
+
+pnpm v10 doesn't support prefix `pnpm_config_*` for environment variables (added
+in v11 via [#10029](https://github.com/pnpm/pnpm/pull/10029)).
+
+```shell
+pnpm_config_ignore_pnpmfile=true pnpm list --lockfile-only --json --depth Infinity --no-color
```
slimreaper35: I did a bit more testing, and I am not as convinced about the `pnpm list` command.

1. Corepack downloads the exact pnpm version declared in the `package.json`.
2. When running the full command in a large project. The `pnpm-lock.yaml` with ~11k lines has turned into ~4.4 million lines printed to STDOUT. Because `--depth Infinity` adds a full list of direct and transitive dependencies for **all** dependencies in the lockfile.

taylormadore: Just stating for the record since we spoke offline about this: we've decided to move forward with the parsing `pnpm-lock.yaml` approach. Some additional reasons were:

- We need to parse the `pnpm-lock.yaml` file anyway to substitute the resolved URLs with local files for the offline build
- The `pnpm list` command doesn't give all of the information we need (like integrity hashes)
- It adds an additional external dependency that we need to work around

slimreaper35: Exactly, it is captured in the doc for future reference, why we decided not to take a certain approach.

docs/design/pnpm.md
```diff
@@ -0,0 +1,490 @@
...
+
+A package declares `peerDependencies` when it expects the host (or another
+dependency) to provide a given package, for example, a React plugin expects React.
+
+They appear only as metadata (`peerDependencies` on packages, `transitivePeerDependencies`
+in snapshots) and do not have their own entries in the `packages` section. When
+a peer is satisfied, the satisfying package appears in `packages` as a regular dependency.
+Hermeto will prefetch everything in the `packages` section, so no special handling
+is needed for peers.
+
```
taylormadore, line 289: We should probably have some mention in this doc about handling of patched dependencies: https://pnpm.io/cli/patch

We already have an open bug related to reporting patched dependencies in the SBOM, so this isn't a blocker for me right now.

taylormadore (approved):
I think we have enough details captured here to get started.

docs/design/pnpm.md
```diff
@@ -0,0 +1,508 @@
...
+  # scoped package
+  '@vue/compiler-core@3.5.30':
+    resolution: {integrity: sha512-s3DfdZkcu/qExZ+td75015ljzHc6vE+30cFMGRPROYjqkroYI5NV2X1yAMX9UeyBNWB9MxCfPcsjpLS11nzkkw==}
+
+```
+
+**PURLs:**
+
+- `pkg:npm/express@5.2.1`
+- `pkg:npm/vue/compiler-core@3.5.30`
```
derasdf: `s/pkg:npm\/vue\/compiler-core/pkg:npm\/%40vue\/compiler-core/`

The `@` in scoped packages must be percent-encoded per the [npm PURL type definition](https://github.com/package-url/purl-spec/blob/master/types-doc/npm-definition.md). yarn-v1.md:90 gets this right. Same fix needed for the JSR PURL at line 356.

docs/design/pnpm.md
```diff
@@ -0,0 +1,508 @@
...
+will be set for development dependencies and transitive development dependencies
+when detected.
+
+The `resolution.integrity` field in the dependency object is an SRI (Subresource
+Integrity) string used to verify the checksum of the tarball. Currently, Hermeto
+does not have a unified approach to checksum handling in the SBOM. See [hermetoproject/hermeto/issues/852](https://github.com/hermetoproject/hermeto/issues/852).
+Let's not add it to the SBOM for now. Removing it later could be considered as
+a breaking change.
+
+### npm registry
```
derasdf, line 331: For npm registry packages the lockfile has only name@version + integrity, no tarball URL. How does Hermeto construct the download URL? Is it always `https://registry.npmjs.org/{name}/-/{name}-{version}.tgz`? What about scoped packages? What if `.npmrc` points to a private registry?

This is the most common package source. The implementation can't proceed without this decision documented.

docs/design/pnpm.md
```diff
@@ -0,0 +1,508 @@
...
+a briefly described format of the lockfile version 9.0 - [https://github.com/pnpm/spec/blob/master/lockfile/9.0.md](https://github.com/pnpm/spec/blob/master/lockfile/9.0.md).
+
+Hermeto will target the lockfile version 9.0. The lockfile version does not seem
+to be directly tied to the pnpm version. pnpm major versions 10 or 11 work with
+version 9.0.
+
+The root keys and their order are also defined in the main pnpm source repository:
+[sortLockfileKeys.ts](https://github.com/pnpm/pnpm/blob/main/lockfile/fs/src/sortLockfileKeys.ts).
+
+- `lockfileVersion` (ComVer string, not SemVer - `'9.0'`)
```
derasdf, line 102: The list references `sortLockfileKeys.ts` but omits `catalogs`, `overrides`, `packageExtensionsChecksum`, and `pnpmfileChecksum`. `snapshots` isn't in the `ROOT_KEYS` array either, it's handled separately in the function body.

2 options: make the list match the source, or note that these are "keys relevant to Hermeto" rather than the full set from `sortLockfileKeys.ts`.

docs/design/pnpm.md
```diff
@@ -0,0 +1,508 @@
...
+4. Running any pnpm command exposes the risk of arbitrary code execution in the
+form of pnpmfile hooks or scripts. These can be avoided by environment variables,
+but still, there is a risk that pnpm could introduce something else in the future.
+
+5. Hermeto has to parse the `pnpm-lock.yaml` anyway to substitute the resolved URLs
+with local files for offline build.
+
+### Why not use the node_modules directory instead?
+
+The [npm cache documentation](https://docs.npmjs.com/cli/v9/commands/npm-cache#a-note-about-the-caches-design)
```
derasdf: Section title says "node_modules" but the body cites the npm cache docs and issue #1389 (which discusses `npm ci` and the npm cache, not node_modules). These are different things. Either retitle to match the evidence or find a node_modules-specific justification.

docs/design/pnpm.md
```diff
@@ -0,0 +1,508 @@
...
+```text
+my-pnpm-project
+  ├── node_modules
+  ├── pnpm-lock.yaml
+  ├── package.json
+  └── src/
+      └── index.js
+```
+
+Compared to an npm project, the `package.json` file has an additional `packageManager`
```
derasdf: nitpick: `packageManager` is a [corepack standard](https://nodejs.org/api/corepack.html), not pnpm-specific. npm and yarn projects use it too. Maybe "pnpm projects typically include the `packageManager` field" rather than "compared to an npm project, the package.json has an additional `packageManager` field."

docs/design/pnpm.md
```diff
@@ -0,0 +1,508 @@
...
+  '@jsr/hono__hono@4.12.4':
+    resolution: {integrity: sha512-fxOSxFkBlmt2dNAvz2vIj8UmihKrkiyjsvU/A8yvHtd9Pt53WBUxsRbpzl8GXodwA0/63gXGxDFDb7XXcPQXbA==, tarball: https://npm.jsr.io/~/11/@jsr/hono__hono/4.12.4.tgz}
+```
+
+JSR packages use the `@jsr/scope__name` naming convention in the lockfile.
+The double underscore separates the scope and name of the package.
+
+**PURLs:**
+
+- `pkg:npm/hono/hono@4.12.4?repository_url=https://npm.jsr.io`
```
derasdf: nitpick: The JSR PURL maps `@jsr/hono__hono` to `hono/hono`, unpacking the double underscore and dropping `@jsr`. Is this intentional? If so, a sentence explaining the mapping logic would help the implementer. If not, should be `pkg:npm/%40jsr/hono__hono@4.12.4?repository_url=https://npm.jsr.io`.

docs/design/pnpm.md
```diff
@@ -0,0 +1,508 @@
...
+packages:
+  is-positive@https://codeload.github.com/kevva/is-positive/tar.gz/97edff6f525f192a3f83cea1944765f769ae2678:
+    resolution: {tarball: https://codeload.github.com/kevva/is-positive/tar.gz/97edff6f525f192a3f83cea1944765f769ae2678}
+    version: 3.1.0
+    engines: {node: '>=0.10.0'}
+```
+
+**PURLs:**
+
+- `pkg:npm/is-positive@3.1.0?vcs_url=git+https://github.com/kevva/is-positive.git#97edff6f525f192a3f83cea1944765f769ae2678`
```
derasdf, line 459: nitpick: The git PURL requires transforming `codeload.github.com` into `github.com` and appending `.git`. Worth noting whether this is GitHub-specific or if there's a general rule for other forges (GitLab, Bitbucket, self-hosted).

docs/design/pnpm.md
```diff
@@ -0,0 +1,508 @@
...
+Authentication settings can be handled via Hermeto configuration
+(YAML or environment variables). See the [Proxy support](#proxy-support) section.
+Further support for other configuration options is out of scope for now.
+
+## Lockfile format
+
+The format is versioned. The latest version is 9.0, which was introduced in
+[pnpm v9.0.0](https://github.com/pnpm/pnpm/releases/tag/v9.0.0). The pnpm GitHub
+organization occasionally maintains the various specifications in the [pnpm/spec](https://github.com/pnpm/spec)
+GitHub repository. The latest commit (at the time of writing) from 2024 points to
```
derasdf, line 92: nitpick: "The latest commit (at the time of writing) from 2024" will rot. Link to the specific commit SHA or just drop the temporal qualifier.

docs/design/pnpm.md
```diff
@@ -0,0 +1,508 @@
...
+a briefly described format of the lockfile version 9.0 - [https://github.com/pnpm/spec/blob/master/lockfile/9.0.md](https://github.com/pnpm/spec/blob/master/lockfile/9.0.md).
+
+Hermeto will target the lockfile version 9.0. The lockfile version does not seem
+to be directly tied to the pnpm version. pnpm major versions 10 or 11 work with
```
derasdf, line 96: nitpick: pnpm v11 is pre-release only as of today. "pnpm major versions 10 or 11" reads as if both are stable. Maybe just "pnpm v10" for now.

derasdf (commented):
Commit subject: `s/desing/design/`. Also could add the `docs:` prefix per project convention, e.g. `docs(pnpm): Add pnpm design document`.

The template has a "Build Environment Config" section that's missing here. "Offline build" partially covers it but doesn't specify whether the patched lockfile overwrites the original, or what generate-env / inject-files would produce.

I'd also suggest adding a "Current Limitations" section that consolidates the scattered out-of-scope mentions (no lockfile < 9.0, no pnpmfile.cjs, no binary platform filtering) into one place.

docs/design/pnpm.md
slimreaper35: Did AI suggest this comment?

docs/design/pnpm.md
slimreaper35: Did AI suggest this comment?

derasdf: Yes, I'm testing my review helper (so if any comments are bad let me know and tell why they are not relevant, trying to make this thing usable and reasonable...)

docs/design/pnpm.md
slimreaper35: not important IMO

docs/design/pnpm.md
slimreaper35: I prefer to keep the current ones as they are easier to read.

docs/design/pnpm.md
slimreaper35: rephrased a bit

docs/design/pnpm.md
```diff
@@ -0,0 +1,512 @@
...
+
+### [Symlinked node_modules structure](https://pnpm.io/symlinked-node-modules-structure)
+
+pnpm uses a unique hard links and symbolic links approach to organize `node_modules`,
+which is very different from the traditional flat npm structure. All package files
+are stored only once in a global content-addressable store.
+
+Only direct dependencies of a project appear as top-level directories inside
+`node_modules`. By using this approach, it is impossible to use modules that are
+not specified in the `package.json` file.
```
a-ovchinnikov, line 46: I would suggest briefly mentioning transitive dependencies here, even a reference for a section below would do, it lets the issue hang otherwise.

docs/design/pnpm.md
slimreaper35: added

docs/design/pnpm.md
```diff
@@ -0,0 +1,512 @@
...
+
+### [Package sources](https://pnpm.io/package-sources)
+
+pnpm uses the [npm registry](https://www.npmjs.com) by default, but also supports
+other package sources.
```
a-ovchinnikov: Nitpick. WDYT of s/./:/ for clarity?

docs/design/pnpm.md
```diff
@@ -0,0 +1,512 @@
...
+
+`pnpm-workspace.yaml` defines the root of the workspace and enables the
+inclusion/exclusion of directories from the workspace.
+
+## [Configuration](https://pnpm.io/npmrc)
+
+pnpm uses the same configuration format as npm. All settings must be configured
+via `.npmrc` files. Hermeto's npm package manager backend ignores the `.npmrc`
+files completely. Each configuration option can be set via an environment variable.
+Naturally, environment variables take precedence over the `.npmrc` file.
```
a-ovchinnikov: So, "must be configured" or "could be configured"? If env takes precedence then why do we need to care about the file? I also think that adding a line explicitly stating that all variables remain the same could help clarifying the situation.

docs/design/pnpm.md
```diff
@@ -0,0 +1,512 @@
...
+[pnpm v9.0.0](https://github.com/pnpm/pnpm/releases/tag/v9.0.0). The pnpm GitHub
+organization occasionally maintains the various specifications in the [pnpm/spec](https://github.com/pnpm/spec)
+GitHub repository. The latest commit (at the time of writing) from 2024 points to
+a briefly described format of the lockfile version 9.0 - [https://github.com/pnpm/spec/blob/master/lockfile/9.0.md](https://github.com/pnpm/spec/blob/master/lockfile/9.0.md).
+
+Hermeto will target the lockfile version 9.0. The lockfile version does not seem
+to be directly tied to the pnpm version. pnpm major versions 10 or 11 work with
+version 9.0.
+
+The root keys and their order are also defined in the main pnpm source repository:
```
a-ovchinnikov: Non-blocking, non-binding: I think the code is the source of truth here, "also" gives precedence to something else (when in reality the tool will behave like written in the code and not like written in the spec unless the code is generated from the spec). I'd drop "also".

docs/design/pnpm.md
```diff
@@ -0,0 +1,512 @@
...
+packages:
+  lodash@4.17.23:
+    resolution: {integrity: sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w==}
+
+snapshots:
+  lodash@4.17.23: {}
+```
+
+</details>
+
```
a-ovchinnikov, line 135: It looks like pnpm stores a mapping from package name-version pair to a file system location internally. WDYT about mentioning this briefly somewhere in the first part of the document and also saying that this mapping will need to be produced/updated by Hermeto?

docs/design/pnpm.md
```diff
@@ -0,0 +1,512 @@
...
+5. Generate SBOM components from the parsed data.
+6. Patch the lockfile for hermetic builds.
+
+The parsing of the lockfile will be the core of the implementation. Each non-registry
+package should have already resolved its URL. For registry packages, the URL must
+be constructed from the package scope, name, and version (`https://registry.npmjs.org/{scope}/-/{name}-{version}.tgz`).
+
+### Why not use the pnpm binary?
+
+Hermeto would run the [pnpm list](https://pnpm.io/cli/list) command and parse the
```
a-ovchinnikov: s/would/could/ ?

docs/design/pnpm.md
```diff
@@ -0,0 +1,512 @@
...
+file.
+
+```yaml
+patchedDependencies:
+  lodash@4.17.23:
+    hash: 5877a18891ec19fc2a2e4eaf39284abc77fe7f8a910708f424aac3434b0813aa
+    path: patches/lodash-4.17.23.patch
+```
+
+CycloneDX specification defines patches for a component as a list of `patch` objects.
```
a-ovchinnikov, line 304: SPDX docs suggest a [couple](https://spdx.github.io/spdx-spec/v2.3/how-to-use/#k17-linking-to-a-code-fix-for-a-security-issue) of [ways](https://spdx.github.io/spdx-spec/v2.3/relationships-between-SPDX-elements/) to report patches, as relationships or as external refs, WDYT about mentioning those here as well?

a-ovchinnikov (commented):
Some questions inline, LGTM otherwise

docs/design/pnpm.md
slimreaper35: > I would suggest briefly mentioning transitive dependencies here

in which way ?

docs/design/pnpm.md
slimreaper35: sure

docs/design/pnpm.md
slimreaper35: This section has only informational purposes. I just picked some lines from the official documentation. I am fine with even dropping it.

a-ovchinnikov: I thought about something like a footnote in style: "Note, however, that this does not mean that indirect dependencies cannot be accounted for, this should be done using a different mechanism, see [Section N] for details". I won't block on this, this is just a minor thing that stood out when I was reading the document.

a-ovchinnikov (approved):

docs/design/pnpm.md
derasdf: I don't think this is purely cosmetic, the link above states `The npm scope @ sign prefix is always percent encoded, as it was in the early days of npm scope.` Every integration test bom.json in the repo and the npm backend (test_project.py:448) also use `%40`. Affects three PURLs in this doc:
- L341: `s/pkg:npm\/vue\/compiler-core/pkg:npm\/%40vue\/compiler-core/`
- L356: `s/pkg:npm\/hono\/hono/pkg:npm\/%40hono\/hono/`
- L457: git PURL `vcs_url` uses `git+` and `#commit` but yarn-v1.md:101
and all integration test data use `git%2Bhttps://...%40commitsha`
(i.e. `+` is `%2B`, commit ref uses `%40` not `#` which is the purl subpath separator)

slimreaper35: Yes, the PURLs are encoded. But that's done automatically by [packageurl-python](https://github.com/package-url/packageurl-python) library that we use by `to_string()` method. I wouldn't know the specific characters from the top of my head.

Would this ⏫ comment in the design be enough?

derasdf: The library handles encoding, agreed. But the examples should match actual output:

```python
>>> PackageURL(type='npm', namespace='@vue', name='compiler-core', version='3.5.30').to_string()
'pkg:npm/%40vue/compiler-core@3.5.30'

>>> PackageURL(type='npm', namespace='vue', name='compiler-core', version='3.5.30').to_string()
'pkg:npm/vue/compiler-core@3.5.30'

>>> p1 == p2
False
```

`pkg:npm/vue/...` and `pkg:npm/%40vue/...` are different PURLs (different namespace). It's not about knowing encoding rules, it's about the examples matching what `to_string()` actually produces.

slimreaper35: The PURLs are just examples for humans, not the actual code. That's why I said it's easier to read. The value of having exact output in the document is minimal.

derasdf: Hmm, alright it is fine then.

PR #1421, opened 2026-03-16
[tests][integration] Fix Nexus server spawning & misc

eskultety:


eskultety:
I don't feel too strongly about the last commit, so I can be convinced to drop it if needed.

pyproject.toml
```diff
@@ -139,7 +139,7 @@ exclude_lines = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "--doctest-modules"
+addopts = "--doctest-modules -n auto"
```
slimreaper35, line 142: The parallelization of integration tests has not been working reliably on my macOS system. I haven't started investigating why, so I would appreciate if the option is not default for integration tests.

slimreaper35: EDIT:

The `pip_e2e_rust_extension` test is stuck :harold: So it has nothing to do with parallelization.

eskultety: So, does that mean I can keep the change? :)

slimreaper35: yes

tests/integration/conftest.py
```diff
@@ -249,32 +237,39 @@ def pytest_sessionstart(session: pytest.Session) -> None:
...
         return
 
-    with contextlib.ExitStack() as context:
-        container, client = initialize_nexus(host=DEFAULT_NEXUS_HOST, port=TEST_NEXUS_PORT)
-        context.enter_context(client)
+    # Unlike PyPI and DNF servers - child processes killed when the parent - exits, Nexus runs
+    # as a detached podman container that survives process termination. No explicit teardown is
+    # performed here.
+    container, client = initialize_nexus(host=DEFAULT_NEXUS_HOST, port=TEST_NEXUS_PORT)
+    log.info("Nexus server ready at %s", client.base_url)
```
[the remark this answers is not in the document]
eskultety: Uhm, yes, both AI comments are legitimate concerns, I pushed a fixup for both -> resolving

PR #1426, opened 2026-03-17
Fix duplicate imports from the same module

slimreaper35:
CodeQL reports these kinds of "warnings" (notes).

Generated-by: Cursor

### DIFF
--- a/hermeto/core/package_managers/npm.py
+++ b/hermeto/core/package_managers/npm.py
@@ -2,10 +2,9 @@
 import asyncio
 import copy
 import fnmatch
-import functools
 import json
 import logging
-from functools import partial
+from functools import cached_property, partial
 from pathlib import Path
 from typing import Any, Literal, NewType, TypedDict
 from urllib.parse import urlparse
@@ -202,7 +201,7 @@ class PackageLock:
                 return True
         return False
 
-    @functools.cached_property
+    @cached_property
     def _purlifier(self) -> "_Purlifier":
         pkg_path = self._lockfile_path.join_within_root("..")
         return _Purlifier(pkg_path)
@@ -341,7 +340,7 @@ class _Purlifier:
         """Init a purlifier for the package at pkg_path."""
         self._pkg_path = pkg_path
 
-    @functools.cached_property
+    @cached_property
     def _repo_id(self) -> RepoID:
         return get_repo_id(self._pkg_path.root)
 
--- a/hermeto/core/http_requests.py
+++ b/hermeto/core/http_requests.py
@@ -2,8 +2,8 @@
 import logging
 from typing import Any
 
-import requests
 from requests import Session
+from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


hermeto/core/package_managers/npm.py
```diff
@@ -537,7 +536,9 @@ async def _async_download_tar(files_to_download_list: list[dict[str, dict[str, A
     # through a proxy.
     auth = lambda ftd: next(iter(ftd.values()))["proxy_auth"]
     ftd = lambda ftd: {it["fetch_url"]: it["download_path"] for it in ftd.values()}
-    adf = partial(async_download_files, concurrency_limit=get_config().runtime.concurrency_limit)
+    adf = functools.partial(
```
a-ovchinnikov: Please don't do this. The point of importing a name directly is to make code easier for a human reader to follow, it is bad enough that questionable annotation practices have overtaken the repository, let's not javify it further!

slimreaper35: That is not the point of this PR.

a-ovchinnikov: I am sorry, what is not the point? I see multiple names being removed from global namespace in favor of accessing them via a module name. I think this makes the code harder to follow and that the other way around should be preferred. Am I missing something?

slimreaper35: But there are many cases where we only use `import x` and access everything via `x`. It seems quite subjective and that it depends on the length of the module name.

a-ovchinnikov: That's correct, however please note, that not all precedents are fortunate and at the same time reworking them is not really affordable. gomod.py and its tests are a good example of that. The only way forward in such cases is a gradual rework, relaxing imports structure when it makes sense is one of the ways of doing this, ensuring LCs and function calls don't span multiple lines is another one. Python is not Java (fun fact, back in the day Java was less Java than it is now), and no matter how much effort is being spent lately on making it indistinguishable from Java this effort is misdirected. Readability counts and it is _the_ selling point of Python, I think it should not be sacrificed.

slimreaper35: I am not against it. It would just feel weird not to have the module name before some functions. I think some modules from the standard library are in 99% cases imported as a whole in the wild as well (`import os`, `import re`). Because you can easily tell where a function/class is coming from. For example, replacing `re.compile("...")` with `compile("...")`.

a-ovchinnikov: `re.compile` is actually a great example for when it makes sense to use a module prefix (because it will shadow a built-in otherwise)

slimreaper35: Updated. It's a very minor improvement, let's not overcomplicate it :)

a-ovchinnikov: > let's not overcomplicate it

I have to politely disagree. Names are important, style is important. If this was not so we would have still been writing in assembly with six characters per label limit or in Java. I fully understand that my views on what constitutes a good style might sometimes look controversial and require explanation, that is why I think discussing style is important, especially in PRs which deal with style changes

slimreaper35: > I think discussing style is important, especially in PRs which deal with style changes

100%

Maybe this specific thing could be solved by adding an instruction for Gemini.

PR #1429, opened 2026-03-19
fix(input): improve error UX and clean logging 

Gitjay11:
### Description
It addresses issue  #1016  by transforming internal Pydantic validation errors into clear, actionable, and user-friendly messages.

As there was no update in the existing pr there so I implemented it. 
It supersedes the work in PR [#1339 ] and incorporates all feedback provided by maintainer @eskultety, including:

1. Clean Log Routing: Resolved the duplicate console output issue by implementing a native logging.Filter ( SuppressConsoleFilter). This allows log.error() to correctly capture failures in FileHandler setups while suppressing redundant terminal output for the end-user.

2. Dynamic Type Management: Refactored the backend type list to be programmatically derived from PackageManagerType using get_args, ensuring the error messages stay up-to-date automatically as the project grows.

3. Efficient Path Simplification: Re-implemented the _simplify_location helper using a high-performance while loop to strip Pydantic-internal array indices (like packages -> 0) from user-facing strings.

4. Integrated Documentation: Added executable doctests within the source code to ensure the string-slicing logic is always verified alongside the implementation.

5. Context preservation: Rewrote the union_tag_invalid handler to provide a clear, plain-English "Unknown package manager" message while preserving the package index for multi-package inputs.

### Key Changes

- hermeto/core/models/input.py: Intercepts Pydantic ValidationError structures and translates them into concise descriptions.
- hermeto/interface/logging.py: Minimal addition of a console-specific log filter.
- hermeto/interface/cli.py: Cleaned up the error handler to utilize the new log filter and improved formatting.
- tests/unit/models/test_input.py: Comprehensive new test suite covering edge cases for single and multi-package inputs.

### Testing & Verification:

Before: 

```
2026-03-19 08:07:04,423 ERROR InvalidInput: 1 validation error for user input\npackages -> 0\n  Requested backend type 'asdf' doesn't match expected ones: 'bundler', 'cargo', 'generic', 'gomod', 'npm', 'pip', 'rpm', 'yarn'
packages -> 0
 Requested backend type 'asdf' doesn't match expected ones: 'bundler', 'cargo', 'generic', 'gomod', 'npm', 'pip', 'rpm', 'yarn'
```

After: 
Screenshot 2026-03-19 080544

Automated Tests: Ran ``` pytest tests/unit/models/test_input.py  ``` and project-wide nox suites. All functional test cases and doctests are passing.

### DIFF
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -3,8 +3,9 @@ import enum
 import logging
 import re
 from collections.abc import Callable
+from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Annotated, Any, Literal, TypeVar, Union
+from typing import TYPE_CHECKING, Annotated, Any, Literal, TypeVar, Union, get_args
 
 import pydantic
 from typing_extensions import Self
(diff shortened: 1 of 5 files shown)


tests/unit/models/test_input.py
```diff
@@ -21,18 +21,51 @@
...
+    def test_union_tag_invalid_plain_english(self) -> None:
+        """union_tag_invalid errors should be rewritten to plain English."""
+        adapter: pydantic.TypeAdapter[PackageInput] = pydantic.TypeAdapter(PackageInput)
+        with pytest.raises(pydantic.ValidationError) as exc_info:
+            adapter.validate_python({"type": "asdf"})
+        # Verify the raw pydantic error is union_tag_invalid
+        assert exc_info.value.errors()[0]["type"] == "union_tag_invalid"
+
+        # Now test through our wrapper
+        with pytest.raises(InvalidInput, match=r"Unknown package manager 'asdf'"):
```
a-ovchinnikov: A drive-by comment: we are trying to move away from string comparisons of any kind in tests because this makes tests brittle. Please consider using alternative ways to make assertions.

Gitjay11: Hi @a-ovchinnikov, okay I totally agree to it.
I've updated the PR to remove the string-matching regex from the unit tests. I refactored the error generation to use a new structured UserFriendlyError dataclass under the hood. The tests now assert directly against the object's deterministic properties instead of verifying raw strings, which makes them much less brittle. It's alright now.

tests/unit/models/test_input.py
```diff
@@ -21,18 +21,47 @@
...
+class TestPresentUserInputError:
+    """Test the structured translation of validation errors."""
+
+    def test_single_error_no_header(self) -> None:
+        """Single errors should be translated cleanly."""
+        try:
+            parse_user_input(GomodPackageInput.model_validate, {"type": "go-package"})
+        except InvalidInput as e:
+            # The exact string is formatted later, we just ensure it was raised cleanly
+            assert e is not None
```
eskultety: Even with the commentary I don't really see what this is testing in a meaningful way.

hermeto/core/models/input.py
```diff
@@ -68,40 +69,125 @@ def parse_user_input(to_model: Callable[[T], ModelT], input_obj: T) -> ModelT:
...
+        type_ = error.get("type", "")
+        msg = error.get("msg", "")
+        
+        pkg_idx = None
+        if len(loc) >= 2 and loc[0] == "packages" and isinstance(loc[1], int):
+            pkg_idx = loc[1]
+            
+        unknown_tag = None
+        expected_tags = None
+        if type_ == "union_tag_invalid":
```
eskultety: Why do we need this new class as an improvement over the original work in #1339 ? What could we have not achieved that required this new abstraction and a new pre-processing pass?

tests/unit/models/test_input.py
```diff
@@ -21,18 +21,47 @@
...
 
 
+class TestPresentUserInputError:
+    """Test the structured translation of validation errors."""
+
+    def test_single_error_no_header(self) -> None:
+        """Single errors should be translated cleanly."""
+        try:
+            parse_user_input(GomodPackageInput.model_validate, {"type": "go-package"})
+        except InvalidInput as e:
```
a-ovchinnikov: This is an unusual choice of a way to verify that an exception occurred. Why did you pick it instead of the standard approach?

tests/unit/models/test_input.py
Gitjay11: yea i did the changes now
that was a weak test and didn't verify anything useful. I've completely removed the TestPresentUserInputError class. Instead, I've added a focused TestSimplifyLocation suite with parametrized tests that deterministically verify the _simplify_location helper logic.

hermeto/core/models/input.py
Gitjay11: You're right. I've removed the UserFriendlyError dataclass and the two-pass processing. I've reverted to the much simpler single-pass _present_user_input_error approach, which achieves the exact same result. i did the changes now

tests/unit/models/test_input.py
Gitjay11: I was over-correcting trying to avoid the brittle string-matching regex assertions, but you're absolutely right that try/except is an anti-pattern here. I've removed these tests entirely and replaced them with standard parametrized unit tests for the _simplify_location function to test the logic deterministically.

ST3V1K:
Hi, thank you for the contribution. 
Could you please rebase this?

eskultety:
Closing due to inactivity.

PR #1430, opened 2026-03-19
Converting SPDX to CycloneDX SBOM failed

ST3V1K:
## Problem
  
  Converting a CycloneDX SBOM with top-level annotations to SPDX and back to CycloneDX failed because:
  - Non-JSON values were being treated as JSON, causing parse errors
  - SPDX package annotations were not properly differentiated between top-level annotations (which backend found the component) and component properties - this meant that converting back to CyDX was changing the SBOM

## Fix
 
  - Introducing distinct tool identifiers (`hermeto:backend` vs `hermeto:jsonencoded`) to differentiate annotation types in SPDX
  - Adding proper annotation processing logic that handles backend annotations, JSON-encoded properties, and custom annotations separately
  - Implementing `merge_component_annotations()` to properly merge backend annotations during SBOM merging
  - Preserving backend annotations during round-trip CycloneDX → SPDX → CycloneDX conversion

### DIFF
--- a/hermeto/core/models/sbom.py
+++ b/hermeto/core/models/sbom.py
@@ -5,6 +5,7 @@ import logging
 import re
 from collections import defaultdict
 from collections.abc import Iterable
+from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from functools import cached_property, partial, reduce
 from itertools import chain, groupby
@@ -17,12 +18,17 @@ from packageurl import PackageURL
 from typing_extensions import Self
 
 from hermeto import APP_NAME
+from hermeto.core.errors import UnexpectedFormat
 from hermeto.core.models.property_semantics import Property, PropertyEnum, PropertySet
 from hermeto.core.utils import first_for
 
 log = logging.getLogger(__name__)
 
 BACKEND_ANNOTATION_PREFIX = f"{APP_NAME}:backend:"
+
+ANNOTATOR_BACKEND = f"{APP_NAME}:backend"
+ANNOTATOR_JSON = f"{APP_NAME}:jsonencoded"
+_ANNOTATOR_ORG = {"organization": {"name": "red hat"}}
 
 
 def datetime_to_iso_8601(value: datetime) -> str:
@@ -32,6 +38,70 @@ def datetime_to_iso_8601(value: datetime) -> str:
 
 ISODatetime = Annotated[datetime, pydantic.PlainSerializer(datetime_to_iso_8601)]
 SortedSet = Annotated[set[str], pydantic.PlainSerializer(sorted, return_type=list[str])]
+
+
+@dataclass
+class _PerBackendAccumulator:
+    _max_timestamp: datetime
+    subjects: set[str] = field(default_factory=set)
+
+    @property
+    def max_timestamp(self) -> datetime:
+        return self._max_timestamp
+
+    @max_timestamp.setter
+    def max_timestamp(self, value: datetime) -> None:
+        if value > self._max_timestamp:
+            self._max_timestamp = value
+
+    def accumulate(self, subjects: Iterable[str], at: datetime) -> None:
+        """Add subjects and update the max timestamp."""
+        self.subjects.update(subjects)
+        self.max_timestamp = at
+
+    @staticmethod
+    def to_annotations(
+        accumulators: dict[str, "_PerBackendAccumulator"],
+    ) -> "list[Annotation]":
+        """Convert accumulated backends to CycloneDX annotations."""
+        return [
+            Annotation(
+                subjects=SortedSet(acc.subjects),
+                annotator=_ANNOTATOR_ORG,
+                timestamp=acc.max_timestamp,
+                text=name,
+            )
+            for name, acc in accumulators.items()
+        ]
+
+
+a_backend_indicator = object()
+
+
+def convert_annotation_to_property(
+    annotation: "SPDXPackageAnnotation",
+) -> Property | object:
+    """Convert a SPDX annotation to a CycloneDX property or a backend indicator.
+
+    Returns a Property for JSON-encoded or custom annotations,
+    or `a_backend_indicator` for backend annotations.
+    """
+    if annotation.annotator.endswith(ANNOTATOR_JSON):
+        try:
+            return Property(**json.loads(annotation.comment))
+        except json.JSONDecodeError:
+            raise UnexpectedFormat(
+                f"Invalid JSON in annotation: {annotation.comment!r}",
+                solution=(
+                    "The annotation comment should be a valid JSON object"
+                    ' with "name" and "value" keys,'
+                    ' e.g. {"name": "property_name", "value": "property_value"}.'
+                ),
+            )
+    elif annotation.annotator.endswith(ANNOTATOR_BACKEND):
+        return a_backend_indicator
+    else:
+        return Property(name=annotation.annotator, value=annotation.comment)
 
 
 class Annotation(pydantic.BaseModel):
@@ -180,7 +250,7 @@ def create_backend_annotation(
         text = f"{BACKEND_ANNOTATION_PREFIX}{backend_name}"
     return Annotation(
         subjects={c.bom_ref for c in components},
-        annotator={"organization": {"name": "red hat"}},
+        annotator=_ANNOTATOR_ORG,
         timestamp=spdx_now(),
         text=text,
     )
@@ -232,7 +302,9 @@ class Sbom(pydantic.BaseModel):
                 # NOTE: We might consider deduplicating annotations based on the annotation text
                 # in the future. It is a very rare corner case, though, and it only matters when
                 # merging multiple CycloneDX SBOMs.
-                annotations=self.annotations + other.annotations,
+                annotations=merge_component_annotations(
+                    chain.from_iterable(s.annotations for s in [self, other])
+                ),
                 components=merge_component_properties(
                     chain.from_iterable(s.components for s in [self, other])
                 ),
@@ -289,19 +361,31 @@ class Sbom(pydantic.BaseModel):
             result = []
             base_spdx_annotation = partial(
                 SPDXPackageAnnotation,
-                annotator=f"Tool: {APP_NAME}:jsonencoded",
                 annotationDate=spdx_now(),
                 annotationType="OTHER",
             )
 
             for annotation in self.annotations:
                 if bom_ref in annotation.subjects:
-                    result.append(base_spdx_annotation(comment=annotation.text))
+                    tool = ANNOTATOR_JSON
+                    if annotation.text.startswith(BACKEND_ANNOTATION_PREFIX):
+                        tool = ANNOTATOR_BACKEND
+
+                    result.append(
+                        base_spdx_annotation(
+                            annotator=f"Tool: {tool}",
+                            annotationDate=annotation.timestamp,
+                            comment=annotation.text,
+                        )
+                    )
 
             for property in properties:
                 result.append(
                     base_spdx_annotation(
-                        comment=json.dumps(dict(name=f"{property.name}", value=f"{property.value}"))
+                        annotator=f"Tool: {ANNOTATOR_JSON}",
+                        comment=json.dumps(
+                            dict(name=f"{property.name}", value=f"{property.value}")
+                        ),
                     )
                 )
 
@@ -724,17 +808,26 @@ class SPDXSbom(pydantic.BaseModel):
 
     def to_cyclonedx(self) -> Sbom:
         """Convert a SPDX SBOM to a CycloneDX SBOM."""
+
         components = []
+        backend_accumulators: dict[str, _PerBackendAccumulator] = {}
+        get_backend_name = lambda ann: ann.comment
         eref_rest = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
         for package in self.packages:
-            properties = [
-                (
-                    Property(**json.loads(an.comment))
-                    if an.annotator.endswith(":jsonencoded")
-                    else Property(name=an.annotator, value=an.comment)
-                )
-                for an in package.annotations
-            ]
+            purls = _extract_purls(package.externalRefs)
+
+            properties = []
+            for ann in package.annotations:
+                result = convert_annotation_to_property(ann)
+                if result is a_backend_indicator:
+                    accumulator = backend_accumulators.setdefault(
+                        get_backend_name(ann),
+                        _PerBackendAccumulator(ann.annotationDate),
+                    )
+                    accumulator.accumulate(purls, ann.annotationDate)
+                else:
+                    properties.append(result)
+
             # sourceInfo morphs into ExternalReference of type PROXY_REF_TYPE
             # with PROXY_COMMENT comment
             if package.sourceInfo is not None:
@@ -749,7 +842,6 @@ class SPDXSbom(pydantic.BaseModel):
                 properties=properties,
                 external_references=exrefs,
             )
-            purls = _extract_purls(package.externalRefs)
 
             # cyclonedx doesn't support multiple purls, therefore
             # new component is created for each purl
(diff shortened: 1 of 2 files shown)


hermeto/core/models/sbom.py
```diff
@@ -17,13 +17,17 @@
 from typing_extensions import Self
 
 from hermeto import APP_NAME
+from hermeto.core.errors import UnexpectedFormat
```
eskultety, line 21: Commit message:

It would help to also mention a snippet of the error log that led to this patch.

hermeto/core/models/sbom.py
```diff
@@ -774,12 +815,48 @@ def to_cyclonedx(self) -> Sbom:
                 tools.append(Tool(vendor=vendor, name=name))
                 name, vendor = None, None
 
+        annotations = [
+            Annotation(
+                subjects=SortedSet(purls),
+                annotator={"organization": {"name": "red hat"}},
+                timestamp=spdx_now(),
```
eskultety: SPDX should have a valid timestamp in the `annotationDate` field, this CDX reconstruction wipes the original information which from the correctness POV is IMO wrong. Was that desired?

tests/unit/models/test_sbom.py
```diff
@@ -500,6 +552,7 @@ def test_cyclonedx_sbom_can_be_converted_to_spdx_and_back_without_loosing_any_da
         )
         cyclonedx_sbom = sbom.to_spdx("NOASSERTION").to_cyclonedx()
 
+        sbom.annotations = merge_component_annotations(sbom.annotations)
```
eskultety: Ironically, the test is called 'to_spdx_and_back_without_loosing_any_data' which isn't really true due to always generating a new timestamp I mentioned above. The test itself should be just fine afterwards, since the timestamps are static.

hermeto/core/models/sbom.py
```diff
@@ -17,13 +17,17 @@
...
+from hermeto.core.errors import UnexpectedFormat
 from hermeto.core.models.property_semantics import Property, PropertyEnum, PropertySet
 from hermeto.core.utils import first_for
 
 log = logging.getLogger(__name__)
 
 BACKEND_ANNOTATION_PREFIX = f"{APP_NAME}:backend:"
 
+BACKEND_ANNOTATION_TOOL = f"{APP_NAME}:backend"
+JSON_ENCODED_TOOL = f"{APP_NAME}:jsonencoded"
```
eskultety: An `ANNOTATOR_` prefix might be desirable with these new constants. 

Also, a nitpick (non actionable) - would referencing the backend annotation tool constant from the prefix constant make the string redundancy desirable? :)

hermeto/core/models/sbom.py
```diff
@@ -289,19 +295,28 @@ def generate_package_annotations(
...
             )
 
             for annotation in self.annotations:
                 if bom_ref in annotation.subjects:
-                    result.append(base_spdx_annotation(comment=annotation.text))
+                    tool = (
+                        BACKEND_ANNOTATION_TOOL
+                        if annotation.text.startswith(BACKEND_ANNOTATION_PREFIX)
+                        else JSON_ENCODED_TOOL
+                    )
```
eskultety: We gain nothing from this multiline "oneliner" assignment, just opencode it as:
```
tool = JSON_ENCODED_TOOL
if annotation.text.startswith(...):
  tool = BACKEND_ANNOTATION_TOOL
```

hermeto/core/models/sbom.py
```diff
@@ -774,12 +815,48 @@ def to_cyclonedx(self) -> Sbom:
...
+            subjects=SortedSet(subjects),
+            annotator={"organization": {"name": "red hat"}},
+            timestamp=spdx_now(),
+            text=text,
+        )
+        for text, subjects in backend_annotations.items()
+    ]
+
+    return sorted(
+        all_annotations, key=lambda ann: (ann.annotator, ann.text, ann.timestamp, ann.subjects)
```
eskultety: `(ann.annotator, ann.text, ann.timestamp, ann.subjects)` - I'm afraid this sorting is unstable:

- `ann.annotator` is a dict, `<>` will raise a `TypeError` on dicts, luckily our backends only ever use a static annotator dict so the comparing logic will fall through which bears the question if we should be sorting based on that field at all? We should at least serialize it to stabilize the sorting algorightm
- `ann.subjects` is a set, `<>` have a completely different meaning to standard comparisons, it is about whether `one` is a subset of `other` so although it won't crash, it will not produce expected results

tests/unit/models/test_sbom.py
```diff
@@ -405,6 +406,23 @@ def test_to_spdx(self, mock_spdx_now: str) -> None:
...
+                        Annotation(
+                            subjects=[
+                                "pkg:golang/github.com/org/A@v1.0.0",
+                                "pkg:golang/github.com/org/A@v1.1.0",
+                            ],
+                            annotator={"organization": {"name": "red hat"}},
+                            timestamp="2026-01-01T00:00:00Z",
+                            text="hermeto:backend:golang",
+                        ),
+                    ],
```
eskultety: This same hunk is copy-pasted in the test params 3 times, shouldn't we therefore extract it to a helper or something? :)

eskultety (changes requested):
Apart from the comments I'm under the impression that patch 2/2 could be easily squashed to 1/2 even though tests pass for both; I think they're logically co-located, am I mistaken?

hermeto/core/models/sbom.py
ST3V1K: There could be multiple timestamps to choose from. Instead of selecting one, I used the current time.

So,
> Was that desired?

Yes

Is it correct?
Probably not, I should select the oldest/newest.

eskultety: Hold on, the SBOMs we're merging were ours, I'm rusty about all the SBOM format details, but how could there be multiple timestamps from us? Can you give me an example so that I don't have to go digging in the code?

ST3V1K: This part converts SPDX to CyDX. SPDX has per-package annotations, thus having one timestamp for each one of them. On the other hand, CyDX has one top-level annotation that combines these SPDX per-package annotations into one and has only one timestamp.

The SPDX timestamps will usually be the same (or in margin of seconds), so it shouldn't really matter which to choose.

SPDX:
```
{
    {
      ...
      "annotations": [
        {
          "annotator": "Tool: hermeto:jsonencoded",
          "annotationDate": "2026-04-09T07:24:16Z",
          "annotationType": "OTHER",
          "comment": "hermeto:backend:pip"
        },
        {
          "annotator": "Tool: hermeto:jsonencoded",
          "annotationDate": "2026-04-09T07:24:16Z",
          "annotationType": "OTHER",
          "comment": "{\"name\": \"hermeto:found_by\", \"value\": \"hermeto\"}"
        }
      ],
      ...
    },
    {
      ...
      "annotations": [
        {
          "annotator": "Tool: hermeto:jsonencoded",
          "annotationDate": "2026-04-09T07:24:16Z",
          "annotationType": "OTHER",
          "comment": "hermeto:backend:pip"
        },
        {
          "annotator": "Tool: hermeto:jsonencoded",
          "annotationDate": "2026-04-09T07:24:16Z",
          "annotationType": "OTHER",
          "comment": "{\"name\": \"hermeto:found_by\", \"value\": \"hermeto\"}"
        }
      ],
      ...
    },
```

CyDX:
```
    {
      "subjects": [
        "pkg:pypi/aiohappyeyeballs@2.6.1",
        "pkg:pypi/aiohttp-retry@2.9.1",
        ...
      ],
      "annotator": {
        "organization": {
          "name": "red hat"
        }
      },
      "timestamp": "2026-04-09T07:24:35Z",
      "text": "hermeto:backend:pip"
    },
```

ST3V1K: This is only a problem for the `hermeto:backend` annotation. Others like `hermeto:found_by` are unique for each component.

hermeto/core/models/sbom.py
ST3V1K: Done

tests/unit/models/test_sbom.py
ST3V1K: Should now preserve the timestamp, so I left the name unchanged

hermeto/core/models/sbom.py
ST3V1K: > An ANNOTATOR_ prefix might be desirable with these new constants.

Done

hermeto/core/models/sbom.py
ST3V1K: Done

hermeto/core/models/sbom.py
ST3V1K: It should now be fixed

tests/unit/models/test_sbom.py
ST3V1K: Added a helper function that constructs the annotations

hermeto/core/models/sbom.py
eskultety: > The SPDX timestamps will usually be the same (or in margin of seconds), so it shouldn't really matter which to choose.

In general I'd lean more towards "timestamp represents when the annotation was made, not when a conversion happened" kind of approach.

hermeto/core/models/sbom.py
```diff
@@ -34,6 +40,21 @@ def datetime_to_iso_8601(value: datetime) -> str:
...
 
 
+@dataclass
+class _BackendMerge:
+    _max_timestamp: datetime | None = None
+    subjects: set[str] = field(default_factory=set)
+
+    @property
+    def max_timestamp(self) -> datetime:
+        return self._max_timestamp or datetime.min
```
eskultety: `datetime.min` isn't timezone aware, so later when you sort annotations and compare (also) against timestamps, having failed to set a timestamp (i.e. propagating the default) would crash:
```
>>> datetime.min < datetime.now(timezone.utc)
TypeError: can't compare offset-naive and offset-aware datetimes
```
The default value doesn't luckily propagate currently, but a change in the current logic could open up that path. We could either require `_max_timestamp` at `BackendMerge` construction rather than leaving it optional OR the default could be epoch time: `datetime.fromtimestamp(0, tz=timezone.utc)`. The latter only addresses the crash but still leaves the door open to propagate to the resulting SBOM silently. It's an easier fix though, so I won't complain :) .

hermeto/core/models/sbom.py
```diff
@@ -774,12 +841,51 @@ def to_cyclonedx(self) -> Sbom:
...
+    """Merge component annotations."""
+    backend_annotations: dict[str, _BackendMerge] = defaultdict(_BackendMerge)
+    other_annotations = []
+    for ann in annotations:
+        if ann.text.startswith(BACKEND_ANNOTATION_PREFIX):
+            backend_merge = backend_annotations[ann.text]
+            backend_merge.subjects.update(ann.subjects)
+            backend_merge.max_timestamp = ann.timestamp
+        else:
+            other_annotations.append(ann)
```
eskultety, line 891: nitpick: the if-else branches could be swapped for better optical purposes.

eskultety (commented):
I have a minor comment on the datetime default being introduced, otherwise LGTM.

hermeto/core/models/sbom.py
```diff
@@ -724,17 +759,50 @@ def to_spdx(self, *a: Any, **k: Any) -> Self:  # noqa: ARG002
 
     def to_cyclonedx(self) -> Sbom:
         """Convert a SPDX SBOM to a CycloneDX SBOM."""
+
+        def process_annotation(annotation: SPDXPackageAnnotation) -> Property | None:
```
a-ovchinnikov: Nitpick: this is a nested helper function that gets defined during run time, are you sure you want a detailed docstring for it? (Especially given the type hints.) Additionally if you name it, say, `convert_annotation_to_property` it will be immediately clear what it does.

hermeto/core/models/sbom.py
```diff
@@ -724,17 +759,50 @@ def to_spdx(self, *a: Any, **k: Any) -> Self:  # noqa: ARG002
...
+
+            properties = []
+            for ann in package.annotations:
+                processed = process_annotation(ann)
+                if processed is not None:
+                    properties.append(processed)
+                else:
+                    backend_merge = backend_annotations[ann.comment]
+                    backend_merge.subjects.update(purls)
+                    backend_merge.max_timestamp = ann.annotationDate
```
a-ovchinnikov: Here is an experiment with naming:

```python
a_backend_indicator = object()

def convert_annotation_to_property(... -> Property | Literal[a_backend_indicator]:
...

backend_accumulators = defaultdict(_PerBackendAccumulator)
get_backend_name = lambda ann: ann.comment

for ann in package.annotations:
    result = convert_annotation_to_property(ann)
    if result is a_backend_indicator:
        backend_accumulator = backend_accumulators[get_backend_name(ann)]
        backend_accumulator.accumulate(purls, at=ann.annotationDate)
    else:
        properties.append(result)
```

Same logic, different names, some helpers and aliases for clarity. WDYT? 

_Edit_: this piece of code is also suspiciously similar to `merge_component_annotations()`, could it be reused here?

tests/unit/models/test_sbom.py
```diff
@@ -500,6 +532,7 @@ def test_cyclonedx_sbom_can_be_converted_to_spdx_and_back_without_loosing_any_da
         )
         cyclonedx_sbom = sbom.to_spdx("NOASSERTION").to_cyclonedx()
 
+        sbom.annotations = merge_component_annotations(sbom.annotations)
```
a-ovchinnikov: I am sorry, but why is this call necessary? The code either allows cyclic conversion or not. The original test did verify that, this modification makes me think that `sbom.to_spdx("NOASSERTION").to_cyclonedx()` does not really work because now after a conversion one needs to explicitly do a merge. Could it be tucked under `.to_spdx()` to preserve the interface?

hermeto/core/models/sbom.py
ST3V1K: I've made the `_max_timestamp` required

hermeto/core/models/sbom.py
ST3V1K: Done

hermeto/core/models/sbom.py
ST3V1K: Yes, this is much better. I couldn't come up with names for these classes and variables.

tests/unit/models/test_sbom.py
ST3V1K: This is actually used only for the final sorting. The same applies to components. 
I have moved it under the comment to make it clearer.

hermeto/core/models/sbom.py
ST3V1K: I have moved it from top-level to a nested helper function and forgot to remove the docstring.
But, based on your other review, this was moved and renamed.

eskultety (approved):

tests/unit/models/test_sbom.py
```diff
@@ -504,6 +536,7 @@ def test_cyclonedx_sbom_can_be_converted_to_spdx_and_back_without_loosing_any_da
         # TODO: Stop storing components/properties as a list and use sets; that, however, will
         # cause issues with the JSON encoder/decoder.
         sbom.components = merge_component_properties(sbom.components)
+        sbom.annotations = merge_component_annotations(sbom.annotations)
```
a-ovchinnikov, line 539: Ok, now I remember where the need for this call stems. Not for this PR, but maybe it is better to introduce a test helper to verify that lists contain same elements? (something like `Counter(l1) == Counter(l2)`)

hermeto/core/models/sbom.py
```diff
@@ -724,17 +819,24 @@ def to_spdx(self, *a: Any, **k: Any) -> Self:  # noqa: ARG002
...
-                for an in package.annotations
-            ]
+            purls = _extract_purls(package.externalRefs)
+
+            properties = []
+            for ann in package.annotations:
+                result = convert_annotation_to_property(ann)
+                if result is a_backend_indicator:
+                    _PerBackendAccumulator.accumulate_for(
+                        backend_accumulators, get_backend_name(ann), purls, ann.annotationDate
```
a-ovchinnikov: Non-blocker, nitpick: judging from the number of arguments this method probably does more than it should, I think in this case it would have been fine to keep interactions with the defaultdict outside the helper.

a-ovchinnikov (approved):

hermeto/core/models/sbom.py
ST3V1K: Fair, I removed the helper function and kept the dictionary interaction separate

tests/unit/models/test_sbom.py
ST3V1K: Yes, that would be better. It could be done in a follow-up

hermeto/core/models/sbom.py
```diff
@@ -232,7 +302,9 @@ def __add__(self, other: Union["Sbom", "SPDXSbom"]) -> "Sbom":
                 # NOTE: We might consider deduplicating annotations based on the annotation text
                 # in the future. It is a very rare corner case, though, and it only matters when
                 # merging multiple CycloneDX SBOMs.
-                annotations=self.annotations + other.annotations,
+                annotations=merge_component_annotations(
+                    chain.from_iterable(s.annotations for s in [self, other])
```
a-ovchinnikov, line 306: Sorry, just noticed this: why did you choose `chain.from_iterable(<comprehentsion>)` instead of just a sum? (The sum of course should be passed as an argument to `merge_component_annotations`.) Purely from the ease of reading the code perspective `a.quux + b.quux` looks better to me than `foo.bar_baz(s.quux for s in [a, b])`.

a-ovchinnikov (approved):
Still LGTM with one minor question inline

hermeto/core/models/sbom.py
ST3V1K: I haven't really thought about it and copied what is being done with components.
Should I change it to a simple sum, or keep it as it is (consistent with components)?

a-ovchinnikov: I see, thank you. That's up to you, if you are going to push another revision then please replace this with a sum, if not then it is fine to keep it as is. I have already approved the change and won't block on this.

ST3V1K: Okay, I'll keep it this way unless someone else has any other comments.

PR #1431, opened 2026-03-19
refactor: Use a single PackageJson class

slimreaper35:
Having multiple identical classes is just a bad practise. Another one would arrive during the implementation of the **pnpm** backend.

### DIFF
--- a/hermeto/core/package_managers/common.py
+++ b/hermeto/core/package_managers/common.py
@@ -1,0 +1,48 @@
+# SPDX-License-Identifier: GPL-3.0-only
+import json
+from collections import UserDict
+from pathlib import Path
+from typing import Any
+
+from hermeto.core.errors import InvalidLockfileFormat, LockfileNotFound
+
+
+class PackageJson(UserDict):
+    """Class representing package.json files."""
+
+    def __init__(self, path: Path, data: dict[str, Any]) -> None:
+        """Initialize a PackageJson object."""
+        self.path = path
+        super().__init__(data)
+
+    @classmethod
+    def from_file(cls, path: Path) -> "PackageJson":
+        """Create a PackageJson object from a package.json file."""
+        if not path.exists():
+            raise LockfileNotFound(
+                path,
+                solution="Make sure the package.json file exists in the specified directory.",
+            )
+
+        try:
+            with path.open("r") as f:
+                data = json.load(f)
+        except json.decoder.JSONDecodeError as e:
+            raise InvalidLockfileFormat(
+                lockfile_path=path,
+                err_details=str(e),
+                solution="The package.json file must contain valid JSON.",
+            )
+
+        return cls(path, data)
+
+    @classmethod
+    def from_dir(cls, path: Path) -> "PackageJson":
+        """Create a PackageJson object from a package.json file."""
+        return cls.from_file(path.joinpath("package.json"))
+
+    def write(self) -> None:
+        """Write the data to the package.json file."""
+        with self.path.open("w") as f:
+            json.dump(self.data, f, indent=2)
+            f.write("\n")
--- a/hermeto/core/package_managers/yarn_classic/project.py
+++ b/hermeto/core/package_managers/yarn_classic/project.py
@@ -6,7 +6,6 @@ It also provides basic utility functions. The main logic to resolve and prefetch
 the dependencies should be implemented in other modules.
 """
 
-import json
 import logging
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
@@ -15,6 +14,7 @@ from typing import Any
 from pyarn import lockfile  # type: ignore
 
 from hermeto.core.errors import InvalidLockfileFormat, LockfileNotFound, PackageRejected
+from hermeto.core.package_managers.common import PackageJson
 from hermeto.core.rooted_path import RootedPath
 
 log = logging.getLogger(name=__name__)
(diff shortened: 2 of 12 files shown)


hermeto/core/package_managers/common.py
```diff
@@ -0,0 +1,48 @@
+# SPDX-License-Identifier: GPL-3.0-only
```
eskultety, line 1: I don't think 'common.py' is a good name, I believe we already have a module for this - `general.py` which is truly common to all backends. Additionally, `PackageJson` is only really relevant to Yarn backends (originally I was tempted to mention NPM as well, but we only consume the `-lock.json` there).
Lastly, there seems to already be somewhat of a  coupling between the Yarn backends:
https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/yarn_classic/main.py#L14-L17.
If this also included the NPM backend then I guess I'd be more inclined towards this Yarn common module, but if it's just about a single class then I don't mind the existing Yarn coupling - there are similarities after all! So, you have 2 options:

1. Extract the already common bits to a module named differently than `common.py` OR
2. Just make the change in the Yarn Berry backend and import it from Classic (this is my slight preference)

If you go with 2. then please make sure these symbols are properly exported via Berry's `__init__.py` so that we get rid of the `yarn.utils` submodule import.

slimreaper35: > If this also included the NPM backend then I guess I'd be more inclined towards this Yarn common module, but if it's just about a single class then I don't mind the existing Yarn coupling - there are similarities after all!

I would like the class to be reusable for pnpm and maybe even npm (I don't know why we are not using it). I don't mind the actual location, whether it is a new file, an existing file, or in the pkg manager backend.

eskultety: > (I don't know why we are not using it). 

Because we primarily pull everything from package-lock.json

eskultety: Maybe we could stuff it under js-utils following the idea from https://github.com/hermetoproject/hermeto/pull/1295/changes#diff-b07d08370945d8eb8e800aaf0d41ea4d299ac782925bc0de3d403e8dda5c8a09

eskultety: OR now that we have several JS backends and we'll have a new python backend soon-ish, maybe we can introduce an additional <language> directory level - a virtual subpackage if you will - and stuff all JS backends in there and do the same for Python later, i.e.
```
$  tree -L2 hermeto/core/package_managers
...
├── javascript
│   ├── metayarn.py
│   ├── npm.py
│   ├── yarn
│   └── yarn_classic
...
```

slimreaper35: works for me

taylormadore: Yeah, I like this idea as well. We should group the JS backends together so that they can logically share commons

hermeto/core/package_managers/common.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __init__(self, path: Path, data: dict[str, Any]) -> None:
+        """Initialize a PackageJson object."""
+        self.path = path
+        super().__init__(data)
+
+    @classmethod
+    def from_file(cls, path: Path) -> "PackageJson":
+        """Create a PackageJson object from a package.json file."""
+        if not path.exists():
```
taylormadore, line 21: nit: Is this preferred to just catching the FileNotFoundError when trying to open? Easier to ask forgiveness than permission

hermeto/core/package_managers/common.py
```diff
@@ -0,0 +1,48 @@
...
+                lockfile_path=path,
+                err_details=str(e),
+                solution="The package.json file must contain valid JSON.",
+            )
+
+        return cls(path, data)
+
+    @classmethod
+    def from_dir(cls, path: Path) -> "PackageJson":
+        """Create a PackageJson object from a package.json file."""
```
taylormadore, line 41: nit: docstring is identical to the method above

hermeto/core/package_managers/yarn_classic/project.py
```diff
@@ -138,13 +100,15 @@ def is_pnp_install(self) -> bool:
         - the presence of an expanded node_modules directory
         For more details on PnP, see: https://classic.yarnpkg.com/en/docs/pnp
         """
-        install_config_pnp_enabled = self.package_json.install_config.get("pnp", False)
+        install_config = self.package_json.data.get("installConfig", {})
```
taylormadore, line 103: nit: We don't need to access this via .data with the UserDict anymore, do we? There are some other places where this occurs too

hermeto/core/package_managers/yarn_classic/project.py
taylormadore, line 24: We should probably drop this now since it isn't being shared by anything else anymore

hermeto/core/package_managers/yarn_classic/workspaces.py
```diff
@@ -22,6 +22,8 @@ class Workspace(pydantic.BaseModel):
         package_json: Content of package.json file.
     """
 
+    model_config = pydantic.ConfigDict(arbitrary_types_allowed=True)
```
taylormadore, line 25: Implementing `__get_pydantic_core_schema__` on the common PackageJson class might be the preferred way to resolve this. It could be very simple like returning `core_schema.is_instance_schema(cls)`

slimreaper35:
I will open a fresh PR with the PackageJson class in the `javascript` directory.

PR #1435, opened 2026-03-20
fix(tests): isolate git-reliant tests from developer config

### LINKED ISSUE #1059: [tests] Isolate our git-reliant tests from the effects of developer gitconfig files
If one adopts some new global git configuration it may happen that doing so would break some of our unit tests in that development environment. I actually tried preventing git from picking up user `gitconfig` configuration, but that didn't go as planned due to some `gitcredential` issue: https://github.com/hermetoproject/hermeto/pull/844#issuecomment-2720440229 . The only correct way of dealing with the problem rather than treating the symptom (by forcing git to use a Null configuration) is to make sure our tests are not 100% dependent on particular `stdout/stderr` from git.

Original work: https://github.com/hermetoproject/hermeto/pull/844

Gitjay11:
Fixes #1059

### Description
Currently, several Git-reliant unit tests randomly fail if developers have customized certain configurations in their global ~/.gitconfig. Most notably:

1. [diff] noprefix = true breaks test_vendor_changed (which incorrectly asserts the exact git diff output formatting including 
a/ prefixes).
2. [init] defaultBranch = main and [tag] gpgSign = true can break various internal repository operations.

Previous PR attempts to solve similar problems used env=GIT_PRISTINE_ENV (which forces GIT_CONFIG_GLOBAL=/dev/null). However, as noted in the issue, completely suppressing the global config this way caused unintended side-effects with git credentials on integration runs.

### Verify
 verify the robustness of this test isolation by creating a highly disruptive dummy config and forcing tests to read it:

1. Create dummy_gitconfig:

``` [tag]
    gpgSign = true
[diff]
    noprefix = true
[init]
    defaultBranch = main
[commit]
    gpgSign = true
```

2. Run the gomod test suite:

```
GIT_CONFIG_GLOBAL=$(pwd)/dummy_gitconfig pytest tests/unit/package_managers/test_gomod.py 
```

### BEFORE:  

Screenshot 2026-03-20 100149

### AFTER:

Screenshot 2026-03-20 095846

Screenshot 2026-03-20 095802

(Note regarding test output: Test runs may currently emit PytestDeprecationWarning: asyncio_default_fixture_loop_scope is unset and pkg_resources.declare_namespace('google') deprecation warnings. These are pre-existing warnings from third-party libraries interacting with Python 3.12 and are entirely unrelated to the git configuration fixes in this PR).

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -32,6 +32,8 @@ def _create_git_repo(path: Path, files: dict[StrPath, FileContents] | None = Non
     repo = git.Repo.init(path)
     repo.git.config("user.name", "user")
     repo.git.config("user.email", "user@example.com")
+    repo.git.config("commit.gpgSign", "false")
+    repo.git.config("tag.gpgSign", "false")
 
     if files is not None:
         for file_path, content in files.items():
--- a/tests/unit/package_managers/gomod/test_main.py
+++ b/tests/unit/package_managers/gomod/test_main.py
@@ -68,8 +68,8 @@ from hermeto.core.package_managers.gomod.main import (
 )
 from hermeto.core.rooted_path import PathOutsideRoot, RootedPath
 from hermeto.core.scm import GitRepo, RepoID
-from hermeto.core.utils import GIT_PRISTINE_ENV
 from tests.common_utils import GIT_REF, write_file_tree
+from tests.unit.conftest import _create_git_repo
 from tests.unit.package_managers.gomod.helpers import get_mocked_data
 
 GO_CMD_PATH = "/usr/bin/go"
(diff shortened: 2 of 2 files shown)


ST3V1K:
Could you please rebase this and drop the changes to `GIT_PRISTINE_ENV` in `gomod.py`/`utils.py` and to `_create_git_repo`? The issue is about making the tests not dependent on git config - i.e. making sure they don't rely on particular stdout/stderr from git, not removing `GIT_PRISTINE_ENV`.

eskultety:
@Gitjay11 CI is failing with the latest changes.

Gitjay11:
@eskultety All checks are passing now.

tests/unit/package_managers/gomod/test_main.py
```diff
@@ -1184,9 +1156,11 @@ def test_vendor_changed(
 
     write_file_tree(vendor_changes, app_dir, exist_ok=True)
 
-    assert _vendor_changed(app_dir) == bool(expected_change)
-    if expected_change:
-        assert expected_change.format(subpath=subpath) in caplog.text
+    assert _vendor_changed(app_dir) == bool(expected_changed_files)
+    if expected_changed_files:
+        caplog_text = caplog.text.replace("\\", "/")
```
ST3V1K: Since Hermeto is Linux only, I'd remove this.

ST3V1K (changes requested):
Looks good, only a few small issues.

Small nit on the commit message 3c0483db02a750cb398601289665f2019e021b0c:
1. Could you use the full issue URL instead of `#1059`? -- it's a project convention so links work from any context.
2. Could you wrap the body to shorter lines so it reads well in git log?

tests/unit/conftest.py
```diff
@@ -29,9 +29,11 @@ def _create_git_repo(path: Path, files: dict[StrPath, FileContents] | None = Non
     :return: Initialized git repository
     """
     path.mkdir(exist_ok=True)
-    repo = git.Repo.init(path)
+    repo = git.Repo.init(path, initial_branch="master")
```
slimreaper35: Why `master` ?

Gitjay11: yea, the branch name is irrelevant to the test logic here. Removed initial_branch so the repo uses git's configured default.

tests/unit/package_managers/gomod/test_main.py
Gitjay11: okay, I did the changes now

tests/unit/conftest.py
slimreaper35: I don't see it.

Gitjay11: sorry, it's done now

a-ovchinnikov:
Hello @Gitjay11, the change has merge conflicts, could you please either resolve them or close the PR properly?

slimreaper35:
Closing due to inactivity and assigning the issue to @ajko-denai

Issue #1432, opened 2026-03-19
bug: "_use_vendored_sources" silently discards existing .cargo/config.toml content in the cargo main.py


Harsh-2005d:
## Description

In `_use_vendored_sources`, the existing cargo config is parsed and merged with the vendoring template, but the merged result is never used. Only the vendoring template is written to disk, silently dropping any pre-existing user configuration.

## Affected code

```python
original_content = _parse_toml_project_file(config_path) if config_path.exists() else {}
original_content.update(config_template)  # merged result is computed...
return ProjectFile(abspath=config_path, template=tomlkit.dumps(config_template))  # ...but never used
```

## Steps to reproduce

Install tomlkit:
```
pip install tomlkit
```

Run the following script:

```python
import tomlkit
import string

existing_config_toml = """
[build]
target = "x86_64-unknown-linux-gnu"

[net]
retry = 3
"""

config_template = tomlkit.loads("""
[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "${output_dir}/deps/cargo"
""")

original_content = tomlkit.loads(existing_config_toml)
original_content.update(config_template)

# Reproduces the bug: original_content is discarded
template_str = tomlkit.dumps(config_template)
output = string.Template(template_str).safe_substitute(output_dir="/hermeto/output")

print(output)
print("[build] present:", "[build]" in output)
print("[net] present:", "[net]" in output)
```

## Expected behavior

```toml
[build]
target = "x86_64-unknown-linux-gnu"

[net]
retry = 3

[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "/hermeto/output/deps/cargo"
```

## Actual behavior

```toml
[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "/hermeto/output/deps/cargo"
```

## Impact

Any project with a pre-existing `.cargo/config.toml` containing custom configuration (build targets, network settings, registry overrides) will silently lose that configuration when processed by hermeto.

## Fix

```python
# Before
return ProjectFile(abspath=config_path, template=tomlkit.dumps(config_template))

# After
return ProjectFile(abspath=config_path, template=tomlkit.dumps(original_content))
```

eskultety:
This is a very well written bug report.

PR #1436, opened 2026-03-20
fix(cargo): preserve existing config when vendoring sources

Harsh-2005d:
Resolves: #1432

## Description

In `_use_vendored_sources`, the existing cargo config is parsed and merged with the vendoring template, but the merged result is never used. Only the vendoring template is written to disk, silently dropping any pre-existing user configuration.

## Affected code

```python
original_content = _parse_toml_project_file(config_path) if config_path.exists() else {}
original_content.update(config_template)  # merged result is computed...
return ProjectFile(abspath=config_path, template=tomlkit.dumps(config_template))  # ...but never used
```

## Fix

```python
# Before
return ProjectFile(abspath=config_path, template=tomlkit.dumps(config_template))

# After
return ProjectFile(abspath=config_path, template=tomlkit.dumps(original_content))
```

### DIFF
--- a/tests/unit/package_managers/cargo/test_main.py
+++ b/tests/unit/package_managers/cargo/test_main.py
@@ -3,12 +3,14 @@ import textwrap
 from typing import Any
 
 import pytest
+import tomlkit
 
 from hermeto.core.errors import UnexpectedFormat
 from hermeto.core.package_managers.cargo.main import (
     CargoPackage,
     _resolve_main_package,
     _sanitize_cargo_config,
+    _use_vendored_sources,
 )
 from hermeto.core.rooted_path import RootedPath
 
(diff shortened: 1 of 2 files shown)


slimreaper35:
@Harsh-2005d Can you write one unit test for that? Otherwise, LGTM.

slimreaper35:
Please squash commits; we don't do merge commits.

Harsh-2005d:
does this work?

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -234,3 +236,58 @@ def test_cargo_config_without_registries_gets_sanitized(config_input: str) -> No
...
+
+@pytest.mark.parametrize(
+    "existing_config, config_template, expected_keys",
+    [
+        pytest.param(
+            None,
+            {
+                "source": {
+                    "crates-io": {"replace-with": "vendored-sources"},
+                    "vendored-sources": {"directory": "${output_dir}/deps/cargo"},
```
slimreaper35: The config template param is the same, so it can be defined in the test.

eskultety:
Nicely done!

PR #1439, opened 2026-03-21
docs: use ENTRYPOINT instead of CMD in pip.md Dockerfile example

aryanchalotraa:
The example Dockerfile in docs/pip.md uses CMD without ENTRYPOINT
for a container that has a single fixed purpose.

Best practice for single-purpose containers is to use ENTRYPOINT
in exec form so the application process runs as PID 1 and handles
OS signals (SIGTERM, SIGINT) correctly. Using CMD alone means the
default command can be silently overridden at runtime by anyone
passing an argument to docker run.

Changed CMD to ENTRYPOINT in the example Dockerfile.

Signed-off-by: Aryan Chalotra <aryanchalotra96@gmail.com>

### DIFF
--- a/docs/pip.md
+++ b/docs/pip.md
@@ -848,7 +848,7 @@ RUN source /tmp/hermeto.env \
     && python3 -m pip install --use-pep517 -r requirements.txt \
     && python3 -m pip install --use-pep517 .
 
-CMD ["python3", "-m", "simple_color_output"]
+ENTRYPOINT ["python3", "-m", "simple_color_output"]
 ```
 
 After mounting the required Hermeto data, we can then build the image!


docs/pip.md
```diff
@@ -848,7 +848,7 @@ RUN source /tmp/hermeto.env \
     && python3 -m pip install --use-pep517 -r requirements.txt \
     && python3 -m pip install --use-pep517 .
 
-CMD ["python3", "-m", "simple_color_output"]
+ENTRYPOINT ["python3", "-m", "simple_color_output"]
```
a-ovchinnikov, line 851: > CMD alone can be silently overridden at runtime by passing arguments to docker run.

I am sorry, but this is an _example_ which serves the purpose of showing a possible way of doing something, not necessarily the best way of doing something. If someone decides to just copy it without thinking to their prod that's hardly Hermeto's problem, or am I wrong in my assessment?

eskultety: Yes, this change is completely unjustified, closing.

eskultety:
This change is unjustified. See the threaded discussion above.

PR #1440, opened 2026-03-21
feat(generic): add bearer token auth for downloads

### LINKED ISSUE #1224: [Feature Request] Support authentication for generic backend downloads in prefetch-dependencies
### Problem Description
Currently, the **generic backend** in `prefetch-dependencies` does not support authentication.

I am trying to run a hermetic build where `artifacts.lock.yaml` contains a `download_url` pointing to a private generic archive (e.g., a ZIP file on a private GitLab instance).

Despite configuring a `.netrc` secret and mounting it to the `prefetch-dependencies` task, the task attempts to download the file anonymously. This results in an authentication failure (receiving an HTML login page instead of the artifact), causing a checksum mismatch.

### Proposed Solution
Please add support for authentication in the generic backend handler (Cachi2). It should respect the provided credentials (e.g., via `.netrc` workspace) when fetching remote artifacts defined by a raw `download_url`.

IronwallxR5:
This implements bearer token authentication for the generic backend, allowing artifacts to be downloaded from private registries.

It introduces a new v2.0 lockfile schema with an 'auth' field and extends the download pipeline to inject per-URL authorization headers.

Closes #1224 
Assisted-by: Claude

### DIFF
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -71,9 +71,9 @@ async def _async_download_binary_file(
     session: aiohttp_retry.RetryClient,
     url: str,
     download_path: StrPath,
-    auth: aiohttp.BasicAuth | None = None,
     ssl_context: ssl.SSLContext | None = None,
     chunk_size: int = 8192,
+    headers: dict[str, str] | None = None,
 ) -> None:
     """
     Download a binary file (such as a TAR archive) from a URL using asyncio.
@@ -81,8 +81,8 @@ async def _async_download_binary_file(
     :param aiohttp_retry.RetryClient session: Aiohttp interface for making HTTP requests.
     :param str url: URL for file download
     :param str download_path: File path location
-    :param aiohttp.BasicAuth auth: Authentication for the URL
     :param int chunk_size: Chunk size param for Response.content.read()
+    :param dict[str, str] headers: Custom HTTP headers.
     :raise FetchError: If download failed
     """
     try:
@@ -94,9 +94,9 @@ async def _async_download_binary_file(
         async with session.get(
             url,
             timeout=timeout,
-            auth=auth,
             raise_for_status=True,
             ssl=ssl_context,
+            headers=headers,
         ) as resp:
             with open(download_path, "wb") as f:
                 while True:
@@ -115,18 +115,32 @@ async def _async_download_binary_file(
     log.debug(f"Download completed - {url}")
 
 
+def _merge_headers(
+    global_headers: dict[str, str] | None,
+    url_headers: dict[str, str] | None,
+) -> dict[str, str] | None:
+    """Merge global headers and url-specific headers."""
+    if not global_headers and not url_headers:
+        return None
+    merged = dict(global_headers or {})
+    merged.update(url_headers or {})
+    return merged
+
+
 async def async_download_files(
     files_to_download: Mapping[str, StrPath],
     concurrency_limit: int,
     ssl_context: ssl.SSLContext | None = None,
-    auth: aiohttp.BasicAuth | None = None,
+    headers: dict[str, str] | None = None,
+    headers_by_url: Mapping[str, dict[str, str]] | None = None,
 ) -> None:
     """Asynchronous function to download files.
 
     :param files_to_download: Mapping of URLs and file paths to download.
     :param concurrency_limit: Max number of concurrent tasks (downloads).
     :param ssl_context: Optional SSL context for the requests.
-    :param auth: Optional authorization data for proxies.
+    :param headers: Optional custom HTTP headers to apply to all requests.
+    :param headers_by_url: Optional mapping of URL to custom HTTP headers.
     """
     trace_config = aiohttp.TraceConfig()
     num_attempts: int = int(DEFAULT_RETRY_OPTIONS["total"])
@@ -163,6 +177,10 @@ async def async_download_files(
                         t.cancel()
                     raise
 
+            # Get the headers specifically for this URL and merge them with global headers
+            url_specific_headers = headers_by_url.get(url) if headers_by_url else None
+            merged_headers = _merge_headers(headers, url_specific_headers)
+
             tasks.add(
                 asyncio.create_task(
                     _async_download_binary_file(
@@ -170,7 +188,7 @@ async def async_download_files(
                         url,
                         download_path,
                         ssl_context=ssl_context,
-                        auth=auth,
+                        headers=merged_headers,
                     )
                 )
             )
--- a/hermeto/core/package_managers/generic/main.py
+++ b/hermeto/core/package_managers/generic/main.py
@@ -14,7 +14,7 @@ from hermeto.core.models.input import Request
 from hermeto.core.models.output import RequestOutput
 from hermeto.core.models.sbom import Component, create_backend_annotation
 from hermeto.core.package_managers.general import async_download_files
-from hermeto.core.package_managers.generic.models import GenericLockfileV1
+from hermeto.core.package_managers.generic.models import GenericLockfileV1, LockfileArtifactUrl
 from hermeto.core.rooted_path import RootedPath
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


hermeto/core/package_managers/generic/main.py
```diff
@@ -83,13 +84,31 @@ def _resolve_generic_lockfile(lockfile_path: Path, output_dir: RootedPath) -> li
...
+        if isinstance(artifact, LockfileArtifactUrl):
+            auth_header = artifact.resolve_auth_header()
+            if auth_header:
+                host = urlparse(url).hostname
+                header_names = ", ".join(auth_header.keys())
+                log.warning(
+                    f"Sending auth header(s) [{header_names}] to host '{host}' "
+                    f"for artifact '{artifact.filename}'. "
+                    f"Ensure that this host is trusted."
+                )
```
eskultety: Such a warning only raises suspicion, please drop this, a user following documentation should be aware of the risks.

hermeto/core/package_managers/generic/models.py
```diff
@@ -24,12 +25,62 @@
...
+        var_name = match.group(1) or match.group(3)
+        env_value = os.environ.get(var_name)
+        if env_value is None:
+            missing.append(var_name)
+            return match.group(0)
+        return env_value
+
+    resolved = ENV_VAR_PATTERN.sub(_replace, value)
+    if missing:
+        raise PackageManagerError(f"Required environment variable(s) not set: {', '.join(missing)}")
```
eskultety: What's the point of this error if you'll discard it later?

hermeto/core/package_managers/general.py
```diff
@@ -163,6 +167,7 @@ async def async_download_files(
                         t.cancel()
                     raise
 
+            url_headers = headers_by_url.get(url) if headers_by_url else None
```
eskultety: This addition deserves an explanatory commentary.

hermeto/core/package_managers/general.py
```diff
@@ -120,13 +122,15 @@ async def async_download_files(
     concurrency_limit: int,
     ssl_context: ssl.SSLContext | None = None,
     auth: aiohttp.BasicAuth | None = None,
+    headers_by_url: Mapping[str, dict[str, str]] | None = None,
```
eskultety, line 135: One could probably say this is to complement the basic auth, but we don't need to distinguish between types of auth like aiohttp does.

eskultety (changes requested):
- the last 2 commits should be squashed to the first one
- the PR needs to be split into multiple commits logically - the current split is not logical and is pretty much a single commit created entirely by AI.
- the PR does not follow our contributing guidelines.
- header based auth needs to replace and enhance (or wrap if you will) our current auth logic not complement it.

victor-adebowale-momodu:
Hey @IronwallxR5, I had previously opened a PR that was heavily based on your work here without coordinating with you or getting your consent, that was wrong of me. This was pointed out to me by one of the maintainer and they are right to do so. The PR has been closed and I have deleted the branch.

I do have some suggestions and would be glad to share them as review comments here if that would be useful to you. No obligation at all. 

Sorry again for the misstep.

IronwallxR5:
> Hey @IronwallxR5, I had previously opened a PR that was heavily based on your work here without coordinating with you or getting your consent, that was wrong of me. This was pointed out to me by one of the maintainer and they are right to do so. The PR has been closed and I have deleted the branch.
> 
> I do have some suggestions and would be glad to share them as review comments here if that would be useful to you. No obligation at all.
> 
> Sorry again for the misstep.

@victor-adebowale-momodu No problem, thanks for getting in touch. I’d like to hear your thoughts. If you want to leave those suggestions as comments on the PR, I can include them once the maintainers approve it.

IronwallxR5:
> * the last 2 commits should be squashed to the first one
> * the PR needs to be split into multiple commits logically - the current split is not logical and is pretty much a single commit created entirely by AI.
> * the PR does not follow our contributing guidelines.
> * header based auth needs to replace and enhance (or wrap if you will) our current auth logic not complement it.

Thanks for the review. I will break this PR into smaller, logical commits according to the contributing guidelines. I will also modify the auth logic to wrap the current implementation instead of adding to it. I will push an updated, clean branch asap.

hermeto/core/package_managers/generic/main.py
IronwallxR5: Ok, I will remove this warning.

hermeto/core/package_managers/general.py
IronwallxR5: Sure , I will add a clear explanatory comment.

hermeto/core/package_managers/general.py
IronwallxR5: I will update the code so we handle all authentication the same way.

hermeto/core/package_managers/generic/models.py
IronwallxR5: I will review how this error is handled and make sure we aren't ignoring it by mistake.

IronwallxR5:
Hey @eskultety, I have made all the changes according to your review. I also cleaned up the commit history.

victor-adebowale-momodu:
Hey @IronwallxR5! Thanks for being open to collaboration, I appreciate it.

I took a closer look at the authentication structure i.e. `AuthConfig` + `BearerAuth`, and I think your approach is clean and works well for the current scope.

I was looking at extensibility though, i.e. adding authentication schemes like AWS SigV4 and found that with the current structure,

```python
class AuthConfig(BaseModel):
    bearer: BearerAuth
```

adding new auth types would require making field optional, ensuring that only one auth type is selected. 

One alternative approach I was thinking about was treating auth like a dict and resolving it dynamically based on its key. That way, adding a new auth type wouldn’t require modifying the schema; instead, each auth type could handle its own validation and resolution, while the lockfile interface only needs to deal with the final output (i.e. the resolved headers).

``` python
def resolve_auth_headers(auth: dict | None) -> dict[str, str]:
    if auth is None:
        return {}

    auth_type, config = next(iter(auth.items()))
    if auth_type == "bearer":
        ...
```

For example, an auth configuation in the lockfile could look like

For example, an auth configuration in the lockfile could look like:

```yaml
auth:
  bearer:
    header: Authorization
    value: $TOKEN
```

which would then be interpreted in Python as:
```python
{
    "bearer": {
        "header": "Authorization",
        "value": "$TOKEN",
    }
}
```

I think Pydantic can still fit well in this approach, by instantiating a Pydantic model on config within the resolver for the headers for each auth type, but scoped to each auth implementation rather than at the top level. This way we can still benefit from validation and type safety.

That said, I do see the benefit of your current Pydantic-based approach in terms of validation and type safety.

I would be glad to hear your thoughts on this.

IronwallxR5:
> Hey @IronwallxR5! Thanks for being open to collaboration, I appreciate it.
> 
> I took a closer look at the authentication structure i.e. `AuthConfig` + `BearerAuth`, and I think your approach is clean and works well for the current scope.
> 
> I was looking at extensibility though, i.e. adding authentication schemes like AWS SigV4 and found that with the current structure,
> 
> ```python
> class AuthConfig(BaseModel):
>     bearer: BearerAuth
> ```
> 
> adding new auth types would require making field optional, ensuring that only one auth type is selected.
> 
> One alternative approach I was thinking about was treating auth like a dict and resolving it dynamically based on its key. That way, adding a new auth type wouldn’t require modifying the schema; instead, each auth type could handle its own validation and resolution, while the lockfile interface only needs to deal with the final output (i.e. the resolved headers).
> 
> ```python
> def resolve_auth_headers(auth: dict | None) -> dict[str, str]:
>     if auth is None:
>         return {}
> 
>     auth_type, config = next(iter(auth.items()))
>     if auth_type == "bearer":
>         ...
> ```
> 
> For example, an auth configuation in the lockfile could look like
> 
> For example, an auth configuration in the lockfile could look like:
> 
> ```yaml
> auth:
>   bearer:
>     header: Authorization
>     value: $TOKEN
> ```
> 
> which would then be interpreted in Python as:
> 
> ```python
> {
>     "bearer": {
>         "header": "Authorization",
>         "value": "$TOKEN",
>     }
> }
> ```
> 
> I think Pydantic can still fit well in this approach, by instantiating a Pydantic model on config within the resolver for the headers for each auth type, but scoped to each auth implementation rather than at the top level. This way we can still benefit from validation and type safety.
> 
> That said, I do see the benefit of your current Pydantic-based approach in terms of validation and type safety.
> 
> I would be glad to hear your thoughts on this.

I completely understand your pov on extensibility. If @eskultety or another maintainer agrees to that approach, I would be more than happy to update the implementation .

eskultety:
The brainstorming in the issue is great!, however you're forgetting one important factor - how is this to be tested? :) We don't have a paid AWS account to verify the auth on a continuous basis in our integration tests. There is a reason why I mentioned AWS SigV4 only as a "potential" future extension to the work if the demand is there (there is not). 
Guys, if you think you can just stuff this additional bit to an AI to dump the code and we'll happily merge, then I'm sorry, that's not going to happen, because we'll need fundamentals to be met first, not just syntax.

eskultety:
Superseded by https://github.com/hermetoproject/hermeto/pull/1525

PR #1441, opened 2026-03-21
fix(scm): strip credentials from username-only HTTPS remote URLs

### LINKED ISSUE #1394: _canonicalize_origin_url   fails to strip credentials from URLs with username-only (no password)
The  _canonicalize_origin_url function in  hermeto/core/scm.py is responsible for stripping credentials from git remote origin URLs before they are included in the SBOM output. However, it fails when a URL contains only a username (no password), which is a common pattern for token-based authentication.

When a URL like https://my-token@github.com/org/repo.git is parsed:

-  parsed.username = "my-token"
- parsed.password = None
- parsed.netloc = "my-token@github.com"

The f-string on line 272 builds the search pattern "my-token:None@", which does not match the actual netloc "my-token@github.com", so str.replace() is a no-op and the credential is not stripped.

##** Impact**

- Credential leak: Auth tokens (e.g. GitHub PATs, CI tokens) embedded as username-only in origin URLs will end up in the generated SBOM/PURL output, potentially exposing secrets.

- Some CI configurations use https://<token>@github.com/... (username-only), which triggers this bug.

*##*Steps to Reproduce**

1. Set up a git repo with a remote origin URL that has a username but no password:

```git remote set-url origin https://my-secret-token@github.com/org/repo.git```

3. Run hermeto fetch-deps and inspect the generated SBOM.
4. The SBOM will contain https://my-secret-token@github.com/org/repo.git instead of https://github.com/org/repo.git.

##**Expected Behavior**
The credential should be stripped, producing: https://github.com/org/repo.git

##**Additional Context**
The existing test in  tests/unit/test_scm.py (line 31) only covers the user:password@ case but never tests a username-only URL, so this bug has gone undetected.

### PR DESCRIPTION
**Pull request by victor-adebowale-momodu**
_canonicalize_origin_url strips credentials from git remote URLs before writing them to the SBOM. It does this by building a search string as "username:password@" and calling .replace() on the netloc. When a token is used as the username with no password (e.g. https://TOKEN@github.com/org/repo.git), urlparse returns password=None. Python's f-string converts None to the literal text "None", producing the search string "TOKEN:None@" which does not exist in the URL. .replace() returns the original string unchanged with no error and the credential leaks into the SBOM output.

Fix by rebuilding netloc from parsed.hostname and parsed.port only, scoped to http/https schemes. Credentials are excluded by construction rather than by pattern matching, handling all credential formats correctly regardless of whether username, password, or both are present.

SSH URLs are explicitly excluded — the username in ssh://user@host is a system user (e.g. "git"), not a credential, and must be preserved.

A warning is logged when credentials are detected, using only the hostname in the message to avoid the warning itself leaking the token.

Add two parametrize cases to TestRepoID.test_get_repo_id that fail before this fix and pass after.

Closes #1394

Assisted-by: Claude

### DIFF
--- a/hermeto/core/scm.py
+++ b/hermeto/core/scm.py
@@ -267,11 +267,22 @@ def get_repo_for_path(repo_root: Path, target_path: Path) -> tuple[GitRepo, Path
 def _canonicalize_origin_url(url: str) -> str:
     if "://" in url:
         parsed: ParseResult = urlparse(url)
-        cleaned_netloc = parsed.netloc.replace(
-            f"{parsed.username}:{parsed.password}@",
-            "",
-        )
-        return parsed._replace(netloc=cleaned_netloc).geturl()
+        if parsed.scheme == "ssh":
+            return url
+
+        if parsed.scheme in ("https", "http"):
+            if parsed.username or parsed.password:
+                log.warning(
+                    "Credentials detected in git remote origin URL for host '%s'. "
+                    "They will be stripped before the URL is written to the SBOM. "
+                    "Consider removing credentials from your remote URL.",
+                    parsed.hostname,
+                )
+            host = parsed.hostname or ""
+            port = f":{parsed.port}" if parsed.port else ""
+            cleaned_netloc = f"{host}{port}"
+            return parsed._replace(netloc=cleaned_netloc).geturl()
+        return url
     # scp-style is "only recognized if there are no slashes before the first colon"
     elif re.match("^[^/]*:", url):
         parts = url.split("@", 1)
--- a/tests/unit/test_scm.py
+++ b/tests/unit/test_scm.py
@@ -30,6 +30,14 @@ class TestRepoID:
                 "https://student:password@github.com/student/repo.git",
                 "https://github.com/student/repo.git",
             ),
+            (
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: hermeto/core/scm.py [hunk 1]
```diff
@@ -267,11 +267,20 @@ def get_repo_for_path(repo_root: Path, target_path: Path) -> tuple[GitRepo, Path
...
+                    "They will be stripped before the URL is written to the SBOM. "
+                    "Consider removing credentials from your remote URL.",
+                    parsed.hostname,
+                )
+            host = parsed.hostname or ""
+            port = f":{parsed.port}" if parsed.port else ""
+            cleaned_netloc = f"{host}{port}"
+            return parsed._replace(netloc=cleaned_netloc).geturl()
+        # ssh:// — preserve the username, it is a system user not a credential
+        return url
```
eskultety: nitpick: I would short-circuit the SSH case to make this easier to read.

victor-adebowale-momodu:
Hey @eskultety, I added a line atop that function that returns the URL if the parsed scheme is `ssh`

eskultety:
We're ready to go with this one, we just need one more ACK.

PR #1443, opened 2026-03-23
fix: handle dict type explicitly and propagate exist_ok in write_file…

aryanchalotraa:
- Replace bare else with explicit elif isinstance(value, dict) for clarity
- Add TypeError for unsupported entry types instead of silently failing
- Fix missing exist_ok passthrough in recursive call

### DIFF
--- a/tests/common_utils/__init__.py
+++ b/tests/common_utils/__init__.py
@@ -31,4 +31,4 @@ def write_file_tree(tree_def: dict, rooted_at: StrPath, exist_ok: bool = False) 
             entry_path.write_text(value)
         else:
             entry_path.mkdir(exist_ok=exist_ok)
-            write_file_tree(value, entry_path, exist_ok=exist_ok)
+            write_file_tree(value, entry_path, exist_ok=exist_ok) 


tests/common_utils/__init__.py
```diff
@@ -30,6 +30,8 @@ def write_file_tree(tree_def: dict, rooted_at: StrPath, exist_ok: bool = False)
             os.symlink(value, entry_path)
         elif isinstance(value, str):
             entry_path.write_text(value)
-        else:
+        elif isinstance(value, dict):
             entry_path.mkdir(exist_ok=exist_ok)
-            write_file_tree(value, entry_path)
+            write_file_tree(value, entry_path, exist_ok=exist_ok)
+        else:
```
a-ovchinnikov: Which types do you expect to see here?

aryanchalotraa: Good question right now the function is only meant to handle Symlink, str, and dict. So anything reaching the else branch would be considered an unexpected or invalid input.
Instead of letting it fail silently, I added a TypeError there so issues can be caught and debugged early.

a-ovchinnikov: Could you please answer my question: which types do you expect to see here? Additionally could you please describe a scenario when such a failure could take place?

tests/common_utils/__init__.py
```diff
@@ -30,6 +30,8 @@ def write_file_tree(tree_def: dict, rooted_at: StrPath, exist_ok: bool = False)
             os.symlink(value, entry_path)
         elif isinstance(value, str):
             entry_path.write_text(value)
-        else:
+        elif isinstance(value, dict):
             entry_path.mkdir(exist_ok=exist_ok)
-            write_file_tree(value, entry_path)
+            write_file_tree(value, entry_path, exist_ok=exist_ok)
```
eskultety: I would accept the `exist_ok` fix.

tests/common_utils/__init__.py
```diff
@@ -30,6 +30,8 @@ def write_file_tree(tree_def: dict, rooted_at: StrPath, exist_ok: bool = False)
...
+        else:
+            raise TypeError(f"Unsupported entry type: {type(value)}")
```
eskultety: I don't think any of the rest (beyond `exist_ok`) is necessary. At best, since this is a type-related thing, I'd look at whether we can enhance the type in the function signature reasonably and leave it to completely to mypy. This is for tests anyway.

tests/common_utils/__init__.py
```diff
@@ -31,4 +31,4 @@ def write_file_tree(tree_def: dict, rooted_at: StrPath, exist_ok: bool = False)
             entry_path.write_text(value)
         else:
             entry_path.mkdir(exist_ok=exist_ok)
-            write_file_tree(value, entry_path, exist_ok=exist_ok)
+            write_file_tree(value, entry_path, exist_ok=exist_ok)
```
a-ovchinnikov, line 34: This won't pass linters.

eskultety:
@aryanchalotraa do you not plan on working on this anymore?

PR #1447, opened 2026-03-23
refactor: replace all cachito refs with hermeto org

derasdf:
Replaced references in 2 hack test generators and regenerated the test data, updated unit tests and docs.

Only remaining refs to Cachito are in Integration tests and they will be gone soon too.

Closes: https://github.com/hermetoproject/hermeto/issues/866

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ builds.
 - [Basic usage](#basic-usage)
 - [Configuration](#configuration)
 - [Package managers](#package-managers)
-- [Project status](#project-status)
 
 ## Goals
 
@@ -268,10 +267,6 @@ project successfully. If missing, you can easily generate one by running
 
 See docs/yarn.md for more details.
 
-## Project status
-
-Hermeto was derived from (but is not a direct fork of) [Cachito][].
-
 [^pip-download-example]: See for example this [python.org discussion][]
 
 [^2]: Hermeto expects to use a specific version of the `go` command
@@ -291,7 +286,6 @@ Hermeto was derived from (but is not a direct fork of) [Cachito][].
   to our gomod documentation for more details.
 
 [bundler]: https://bundler.io
-[Cachito]: https://github.com/containerbuildsystem/cachito
 [Cargo.lock]: https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
 [cargo]: https://doc.rust-lang.org/cargo
 [config.py]: https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/config.py
--- a/docs/design/yarn-v1.md
+++ b/docs/design/yarn-v1.md
@@ -34,7 +34,7 @@ already being used to install Yarn in the Hermeto Yarn v2+ implementation, so a 
 
 Corepack installs the latest known-good Yarn v1.x release by default if no other version is specified explicitly. It seems reasonable to always
 use the latest 1.x release for the prefetch, since 1.x is no longer under active development and we never previously allowed users to specify a specific
-Yarn version to process their request in Cachito. We could add more flexibility with a later enhancement if needed.
+Yarn version to process their request. We could add more flexibility with a later enhancement if needed.
 
 Corepack could be configured to ignore user project configuration and use the global Yarn 1.x default by specifying the following
 [environment variable](https://github.com/nodejs/corepack#environment-variables):
@@ -69,19 +69,17 @@ Yarn projects using the [Plug'n'Play (PnP)](https://classic.yarnpkg.com/lang/en/
 Any additional configuration settings in the repository via [.yarnrc](https://classic.yarnpkg.com/en/docs/yarnrc) or .npmrc
 files should be ignored (at least initially) for the prefetch by specifying [`--no-default-rc`](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L73) for the yarn install command.
 
-We never previously honored user-defined settings with Cachito, but this can be investigated and implemented in Hermeto in a follow-up. At minimum, the
+We do not currently honor user-defined settings, but this can be investigated and implemented in Hermeto in a follow-up. At minimum, the
 limitation should be documented.
 
 ### SBOM Generation
 The Yarn 1.x CLI doesn't provide enough information to generate an SBOM, so the yarn.lock and package.json files will need to
-be parsed in order to gather the necessary data. The code used to process these files will need to be
-[imported from or inspired by Cachito](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py)
-and refactored/improved.
+be parsed in order to gather the necessary data. The code used to process these files will need to be refactored/improved.
 
 #### Key Points
- - The yarn.lock file can be parsed with [pyarn](https://github.com/containerbuildsystem/pyarn) like in cachito
- - Workspaces will need to be [processed separately](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L95) from package.json since they do not appear in the yarn.lock file
- - Dev dependencies can be [determined similiarly](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L140) to cachito
+ - The yarn.lock file can be parsed with [pyarn](https://github.com/containerbuildsystem/pyarn)
+ - Workspaces will need to be processed separately from package.json since they do not appear in the yarn.lock file
+ - Dev dependencies can be determined by checking if a dependency is listed in the `devDependencies` of any `package.json` file, or if it is a transitive dependency of a dev dependency
  - We will need to ensure that there are no collisions in the names of the archive files being added to the offline mirror
 
 #### PURL Examples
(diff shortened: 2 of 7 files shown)


hack/mock-unittest-data/gomod.py
```diff
@@ -12,7 +12,7 @@
 from yarn import print_banner
 
 
-PANDEMO_DIR = "gomod-pandemonium"
+PANDEMO_DIR = "integration-tests"
```
eskultety: it's no longer `PANDEMO_DIR` is it? :) I know you're trying to keep the changes to a minimum, but this may end up being more confusing, so a variable renaming is justified.

docs/design/yarn-v1.md
```diff
@@ -34,7 +34,7 @@ already being used to install Yarn in the Hermeto Yarn v2+ implementation, so a
 
 Corepack installs the latest known-good Yarn v1.x release by default if no other version is specified explicitly. It seems reasonable to always
 use the latest 1.x release for the prefetch, since 1.x is no longer under active development and we never previously allowed users to specify a specific
-Yarn version to process their request in Cachito. We could add more flexibility with a later enhancement if needed.
```
eskultety, line 37: For this and the other occurrences of Cachito in the design docs - when these were introduced to (already) Hermeto, I argued we should not make these references, but my suggestion wasn't accepted because the sentiment was that we should make the project's origin very clear. I still personally disagree with that because this project is something very different now what Cachito used to be, so the references are IMO only confusing rather than helpful. I support this change, but I'm mentioning the context here just so you know you might be asked to drop or rephrase these.

docs/generic.md
```diff
@@ -142,67 +142,35 @@ See the [Example](#example) below for a complete walkthrough of Hermeto usage.
...
 otherwise. This image will provide [OWASP Dependency check][] tool, which is available
-to install from GitHub releases page. Get the repo if you want to try for
-yourself
+to install from GitHub releases page.
+
+Get the repo if you want to try for yourself:
 
 ```shell
-git clone -b sample-app https://github.com/cachito-testing/cachi2-generic.git
+git clone https://github.com/hermetoproject/doc-examples.git --branch=generic-basic && cd doc-examples
```
eskultety, line 152: It seems to me that not all of the changes made in this hunk really belong to this commit, but it's trivial enough, so let's keep it as is.

docs/generic.md
```diff
@@ -142,67 +142,35 @@ See the [Example](#example) below for a complete walkthrough of Hermeto usage.
...
-a sample lockfile below. It is identical to the one found in the
-[sample repository][].
-A lockfile for the generic fetcher must contain a `metadata` header and a list
-of artifacts, where each artifact is represented as a pair of URL and a checksum
-string in the format of `"algorithm:checksum"`. Optionally, you can also specify
-an output `filename` for the artifact. If not specified, it will be derived from
-the url.
+As mentioned above in
+[Specifying artifacts to fetch](#specifying-artifacts-to-fetch),
+Hermeto pre-fetches using the `fetch-deps` command.
```
eskultety, line 159: This rewording doesn't IMO belong to this commit in its entirety. Hmm, either the changes need to be split further OR just reword the commit message so that it's not so confining, i.e. 'replace cachito-testing with hermetoproject' because that's simply not true when one looks at the changes.

derasdf: updated commit message

docs/design/yarn-v1.md
derasdf: I wrote about what is left over of Cachito in PR description, I got rid of some doc references to it but for now left in ones where it is stated that Hermeto is derived from Cachito.
If we want to fully get rid of it, to not have ANY Cachito refs by the end of this, I would be ok with it and I also can update syft.boms I suppose.

slimreaper35: > For this and the other occurrences of Cachito in the design docs - when these were introduced to (already) Hermeto, I argued we should not make these references, but my suggestion wasn't accepted because the sentiment was that we should make the project's origin very clear.

I agree; referencing Cachito in the package manager documentation is unnecessary. Maybe we can just keep this part: https://github.com/hermetoproject/hermeto?tab=readme-ov-file#project-status for nostalgic reasons 😄

derasdf:
Looked into why [syft.bom](https://github.com/hermetoproject/hermeto/blob/main/tests/unit/data/sboms/syft.bom.json) with all the cachito-testing refs still exists and here is the [reason](https://github.com/hermetoproject/hermeto/pull/598#discussion_r1715488090). Still not sure how to proceed, finding out how to regenerate it seems like an overkill to me just for [test1](https://github.com/hermetoproject/hermeto/blob/dbcaf72f35089fcdfc46701e7a89478ef1ed35ac/tests/unit/test_cli.py#L1039) and [test2](https://github.com/hermetoproject/hermeto/blob/dbcaf72f35089fcdfc46701e7a89478ef1ed35ac/tests/unit/test_cli.py#L1132). [SPDX](https://github.com/hermetoproject/hermeto/blob/main/tests/unit/data/sboms/syft.bom.spdx.json) version of it definitely can be swapped out and deleted in the test2 but the CycloneDX one from test1 not really, would have to either delete the test or generate something for it in some way. Or we can just not touch it. What do you guys think?

eskultety:
> Still not sure how to proceed, finding out how to regenerate it seems like an overkill to me just for [test1](https://github.com/hermetoproject/hermeto/blob/dbcaf72f35089fcdfc46701e7a89478ef1ed35ac/tests/unit/test_cli.py#L1039) and [test2](https://github.com/hermetoproject/hermeto/blob/dbcaf72f35089fcdfc46701e7a89478ef1ed35ac/tests/unit/test_cli.py#L1132). [SPDX](https://github.com/hermetoproject/hermeto/blob/main/tests/unit/data/sboms/syft.bom.spdx.json) version of it definitely can be swapped out and deleted in the test2 but the CycloneDX one from test1 not really, would have to either delete the test or generate something for it in some way.

Easiest thing you can try is to just manually delete the cachito-testing entries, shouldn't take much time to get the tests working. Of course, the right fix would be to regenerate everything, but that's out of scope for this PR. Since I'm already personally fed up with the omnipresent cachito strings and this purge becoming a neverending story I'd personally go with the more radical approach as I mentioned - manually delete the cachito entries from those SBOMs, we don't even test the output, just the fact we can dump one, so we don't care about the contents. And because we don't even test the merged contents we might not even need a follow-up issue to generate new SBOMs, up to you, I'm ambivalent.

docs/design/yarn-v1.md
derasdf: Hmm maybe, oooor we just get rid of it fully everywhere and forget it ever existed 😇

eskultety: > and forget it ever existed

I think we reached a stage in this project where it so completely different, that the roots are no longer important at all, because, yes, there are still snippets of code which are still 1:1 reimplementation of Cachito, but we've slowly replaced most of it already :) .

docs/design/yarn-v1.md
```diff
@@ -69,19 +69,17 @@ Yarn projects using the [Plug'n'Play (PnP)](https://classic.yarnpkg.com/lang/en/
 Any additional configuration settings in the repository via [.yarnrc](https://classic.yarnpkg.com/en/docs/yarnrc) or .npmrc
 files should be ignored (at least initially) for the prefetch by specifying [`--no-default-rc`](https://github.com/yarnpkg/yarn/blob/7cafa512a777048ce0b666080a24e80aae3d66a9/src/cli/index.js#L73) for the yarn install command.
 
-We never previously honored user-defined settings with Cachito, but this can be investigated and implemented in Hermeto in a follow-up. At minimum, the
+We never previously honored user-defined settings, but this can be investigated and implemented in Hermeto in a follow-up. At minimum, the
```
taylormadore: nit: Perhaps "We do not currently honor user-defined settings"

docs/design/yarn-v1.md
```diff
@@ -69,19 +69,17 @@ Yarn projects using the [Plug'n'Play (PnP)](https://classic.yarnpkg.com/lang/en/
...
-and refactored/improved.
+be parsed in order to gather the necessary data. The code used to process these files will need to be refactored/improved.
 
 #### Key Points
- - The yarn.lock file can be parsed with [pyarn](https://github.com/containerbuildsystem/pyarn) like in cachito
- - Workspaces will need to be [processed separately](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L95) from package.json since they do not appear in the yarn.lock file
- - Dev dependencies can be [determined similiarly](https://github.com/containerbuildsystem/cachito/blob/63b8ec0ea615d114ccfa0d08dc0bec49e60e6a75/cachito/workers/pkg_managers/yarn.py#L140) to cachito
+ - The yarn.lock file can be parsed with [pyarn](https://github.com/containerbuildsystem/pyarn)
+ - Workspaces will need to be processed separately from package.json since they do not appear in the yarn.lock file
+ - Dev dependencies can be determined similarly
```
taylormadore: We should pull the context into this document if we are removing the reference. Namely:

https://github.com/hermetoproject/hermeto/blob/ea75e4eadc3ba3132d5dc0a1e2cca33d77d5b589/hermeto/core/package_managers/yarn_classic/utils.py#L27-L29

eskultety:
LGTM. I'll give my final ACK after the merge conflicts are resolved since that affects nox results when run locally.

README.md
```diff
@@ -270,8 +270,6 @@ See docs/yarn.md for more details.
 
 ## Project status
 
-Hermeto was derived from (but is not a direct fork of) [Cachito][].
-
```
slimreaper35: Let's delete the empty `# Project status` section as well.

PR #1451, opened 2026-03-24
Fix sh format enum value

slimreaper35:


hermeto/core/extras/envfile.py
```diff
@@ -14,7 +14,7 @@ class EnvFormat(str, Enum):
 
     json = "json"
     env = "env"
-    sh = "env"
+    sh = "sh"
```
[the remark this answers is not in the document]
slimreaper35: okay, so looking at the tests, it seems like by design

eskultety, line 17: What's the difference between `sh` and `env`? Isn't the latter be just an alias to the former? Commit 6ab3e7e5ae that introduced this precedes my time with the team and there's no context provided on this enum in the commit message.

slimreaper35: Maybe. The `--format` does not accept `sh` value, but `--output` accepts e.g., `hermeto.sh`, so I got confused.

PR #1452, opened 2026-03-24
 fix(yarn_classic): parse npm aliases correctly to prevent offline mirror collisions

### LINKED ISSUE #927: Yarn 1.x resolutions used in conjunction with aliases are incorrectly reported as offline mirror collisions
When you have a Yarn 1.x project with a [dependency](https://github.com/Kuadrant/kuadrant-console-plugin/blob/01730bd44cb7ea4f0eba60ad08a174a1d68ad1e7/package.json#L110) and then additionally add a resolution for the [same dependency](https://github.com/Kuadrant/kuadrant-console-plugin/blob/01730bd44cb7ea4f0eba60ad08a174a1d68ad1e7/package.json#L30) under an aliased name, Hermeto reports a collision in the offline mirror:

```
ERROR PackageManagerError: Tarball collision in the offline mirror: esbuild-wasm-0.25.3.tgz (2x)
Error: PackageManagerError: Tarball collision in the offline mirror: esbuild-wasm-0.25.3.tgz (2x)
```

This is because the resolution version selector in yarn.lock appears on the [same line](https://github.com/Kuadrant/kuadrant-console-plugin/blob/01730bd44cb7ea4f0eba60ad08a174a1d68ad1e7/yarn.lock#L4263) as the aliased version selector and breaks an assumption in the pyarn library that version selectors on the same line will share the same name: https://github.com/containerbuildsystem/pyarn/blob/ae472b6bc26f850cf7fe606b60650c1ff5021cc7/pyarn/lockfile.py#L84

If this is a valid use of Yarn resolutions, we should parse/handle this case correctly in Hermeto.
If this is not a valid use of Yarn resolutions, we current error is obscuring the actual issue and we should detect and report this error correctly.

References:
Yarn aliases: https://classic.yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-alias
Yarn resolutions: https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/
Yarn resolutions RFC: https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-selective-versions-resolutions.md

Gitjay11:
Fixes #927

### Description
This PR addresses the "Tarball collision in the offline mirror" `PackageManagerError` that occurs when a Yarn 1.x project contains resolutions overriding dependencies with an `npm:` alias (e.g., `"my-alias": "npm:esbuild-wasm@^0.25.3"`). 
When these scenarios happen, Yarn merges both the main dependency and the aliased resolution into a single comma-separated block in the [yarn.lock](cci:7://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/tmp-integration-tests/yarn.lock:0:0-0:0) file. Due to [pyarn](cci:1://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/hermeto/core/package_managers/yarn_classic/resolver.py:194:4-279:13)'s native regex rules, it evaluates only the primary selector in compound strings, silently stripping the aliased name from the [YarnClassicPackage](cci:2://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/hermeto/core/package_managers/yarn_classic/resolver.py:186:0-279:13). As a result, Hermeto attempts to store two identically sourced tarballs under two different names, triggering a false-positive mirror collision crash.

### How to Test
1. Unit Tests: Execute `pytest tests/unit/package_managers/yarn_classic/test_resolver.py`. The new tests will explicitly prove that compound aliased blocks properly iterate and extract the correct subset name instead of falling back to Pyarn's limited Regex.
2. Integration Tests:  Run `pytest tests/integration/test_yarn_classic.py`. The offline mirror collision validations successfully pass.

### DIFF
--- a/hermeto/core/package_managers/yarn_classic/resolver.py
+++ b/hermeto/core/package_managers/yarn_classic/resolver.py
@@ -139,7 +139,7 @@ class FilePackage(_BasePackage, _PathMixin):
             name=self.name,
             version=self.version,
             qualifiers=qualifiers,
-            subpath=str(self.path.subpath_from_root),
+            subpath=self.path.subpath_from_root.as_posix(),
         ).to_string()
 
 
@@ -157,7 +157,7 @@ class WorkspacePackage(_BasePackage, _PathMixin):
             name=self.name,
             version=self.version,
             qualifiers=qualifiers,
-            subpath=str(self.path.subpath_from_root),
+            subpath=self.path.subpath_from_root.as_posix(),
         ).to_string()
 
 
@@ -175,7 +175,7 @@ class LinkPackage(_BasePackage, _PathMixin):
             name=self.name,
             version=self.version,
             qualifiers=qualifiers,
-            subpath=str(self.path.subpath_from_root),
+            subpath=self.path.subpath_from_root.as_posix(),
         ).to_string()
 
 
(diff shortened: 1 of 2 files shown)


a-ovchinnikov:
Hello @Gitjay11, if you intend to continue working on this PR could you please resolve merge conflicts?

Gitjay11:
@a-ovchinnikov okay sure, I will resolve them

PR #1453, opened 2026-03-25
fix(generate-env): dependent variables get stale values because mappings are built before path resolution

### LINKED ISSUE #1424: generate-env resolves dependent variables using pre-resolved values, leading to incorrect paths
## Summary

While looking at `generate-env`, it seems that environment variables which reference other variables can be resolved using stale values.

This happens because the substitution `mappings` are built before `kind="path"` variables are resolved.

---

## Example

From the gomod backend:

- `GOMODCACHE` (kind="path"):
```
deps/gomod/pkg/mod
```

- `GOPROXY`:
```
file://${GOMODCACHE}/cache/download
```

---

## What happens

`generate_envfile` builds mappings like:
```python
mappings = {var.name: var.value for var in build_config.environment_variables}
```

At this point:
```
GOMODCACHE = deps/gomod/pkg/mod
```

Later, during `resolve_value`, `GOMODCACHE` is updated to:
```
${output_dir}/deps/gomod/pkg/mod
```

But `GOPROXY` is still resolved using the original value from `mappings`.

---

## Result
```
GOPROXY = file://deps/gomod/pkg/mod/cache/download
```

instead of:
```
GOPROXY = file:///tmp/output/deps/gomod/pkg/mod/cache/download
```

---

## Impact

The generated env file can contain relative paths instead of absolute ones. For gomod, this means `GOPROXY` may point to a non-existent location, causing module resolution to fail during the build phase.

---

## Possible Direction

Resolve `kind="path"` variables before constructing the mappings, or build mappings from already-resolved values so dependent variables see the correct paths.

aaradhychinche-alt:
### Context
In the hermeto codebase, the `generate_env.py` file contains a `generate_envfile` function
that has a variable resolution ordering bug.

### The Bug
The function builds a substitution `mappings` dict from raw variable values BEFORE resolving
`kind="path"` variables. This means when a variable like `GOPROXY` references `${GOMODCACHE}`,
it gets substituted with the raw unresolved value of `GOMODCACHE` (e.g. `deps/gomod/pkg/mod`)
instead of its fully resolved absolute path (e.g. `${output_dir}/deps/gomod/pkg/mod`).

**Current broken flow:**
1. Build `mappings = {var.name: var.value for var in build_config.environment_variables}`
   → At this point `GOMODCACHE = "deps/gomod/pkg/mod"` (relative, not yet path-resolved)
2. Resolve `kind="path"` variables → `GOMODCACHE` becomes `/tmp/output/deps/gomod/pkg/mod`
3. Resolve `GOPROXY` using stale mappings → `GOPROXY = "file://deps/gomod/pkg/mod/cache/download"` 

**Expected result:**
`GOPROXY = "file:///tmp/output/deps/gomod/pkg/mod/cache/download"` 

### Fix Required
In `hermeto/core/generate_env.py`, fix the `generate_envfile` function to build the
substitution `mappings` dict AFTER all `kind="path"` variables have been resolved,
not before. The correct order is:

1. First pass: resolve all `kind="path"` variables (prepend `output_dir`)
2. Then build `mappings` from the now-resolved values
3. Then resolve all remaining variable references using these correct mappings

### Implementation guidance
- Find where `mappings` is constructed (likely a dict comprehension over
  `build_config.environment_variables`)
- Find where `kind="path"` resolution happens (likely a call to `resolve_value` or
  similar that prepends `output_dir` to path-kind vars)
- Reorder so path resolution happens first, then mappings are built from the
  already-resolved values
- Ensure the fix handles the case where `kind="path"` variables reference OTHER
  `kind="path"` variables (topological ordering may be needed if such chains exist;
  if not, a single-pass reorder is sufficient)
- Add or update unit tests in the corresponding test file to cover:
  - A `kind="path"` variable (`GOMODCACHE`) with a relative value
  - A dependent variable (`GOPROXY`) that references the path variable via `${GOMODCACHE}`
  - Assert the dependent variable resolves to the absolute path, not the relative one

### Files to look at
- `hermeto/core/generate_env.py` — primary fix location
- `tests/unit/test_generate_env.py` (or similar) — add regression test
- `hermeto/package_managers/gomod.py` — for understanding how GOMODCACHE/GOPROXY
  env vars are declared (kind="path" vs plain), for test fixtures

### Acceptance criteria
- `GOPROXY` (and any variable referencing a `kind="path"` variable) resolves to an
  absolute path including the `output_dir` prefix
- No existing tests are broken
- New test explicitly covers the stale-mapping regression case

fixes #1424

### DIFF
--- a/hermeto/core/extras/envfile.py
+++ b/hermeto/core/extras/envfile.py
@@ -1,12 +1,130 @@
 # SPDX-License-Identifier: GPL-3.0-only
 import json
+import re
 import shlex
+from collections import deque
 from enum import Enum
 from pathlib import Path
+from typing import Optional
 
 from hermeto import APP_NAME
 from hermeto.core.errors import UnsupportedFeature
-from hermeto.core.models.output import BuildConfig
+from hermeto.core.models.output import BuildConfig, EnvironmentVariable
+
+# Pattern that matches ${VAR_NAME} placeholders (brace-delimited only, never bare).
+_REF_PATTERN = re.compile(r"\$\{(\w+)\}")
+
+
+def _extract_refs(value: str) -> list[str]:
+    """Return all variable names referenced as ${NAME} in *value*."""
+    return _REF_PATTERN.findall(value)
+
+
+def _substitute(value: str, mappings: dict[str, str]) -> str:
+    """Replace every ${NAME} in *value* using *mappings* (brace-delimited only).
+
+    Unknown placeholders are left untouched so callers can detect them later.
+    Output_dir is treated as a regular mapping entry; pass it in *mappings* if needed.
+    """
+
+    def _replace(m: re.Match) -> str:  # type: ignore[type-arg]
+        name = m.group(1)
+        return mappings.get(name, m.group(0))
+
+    return _REF_PATTERN.sub(_replace, value)
+
+
+def _detect_cycles(
+    all_vars: list[EnvironmentVariable],
+    name_to_var: dict[str, EnvironmentVariable],
+) -> None:
+    """Raise ValueError if a cycle exists in the ${VAR} reference graph.
+
+    Uses iterative DFS; detected cycles are reported as a readable chain:
+      "Circular variable reference detected: A → B → A"
+    """
+    # Build adjacency list restricted to variables we know about.
+    adj: dict[str, list[str]] = {
+        v.name: [r for r in _extract_refs(v.value) if r in name_to_var] for v in all_vars
+    }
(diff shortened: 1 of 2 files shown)


aaradhychinche-alt:
@eskultety  the pr is ready to be reviewed !!!!

eskultety:
Closing since this PR was superseded by https://github.com/hermetoproject/hermeto/pull/1612 and then covered by https://github.com/hermetoproject/hermeto/pull/1617

PR #1458, opened 2026-03-27
Add support for package manager agnostic lockfile pylock.toml

ST3V1K:
# Add PEP 751 pylock.toml Support

## Overview

This PR adds support for PEP 751 `pylock.toml` lockfiles in Hermeto's pip package manager.

## Background

Users increasingly generate lockfiles with tools like `uv`, `poetry`, and `pipenv`, which produce incompatible formats. Currently, Hermeto only supports `requirements.txt`, forcing users to:
- Generate `requirements.txt` as an intermediate step
- Lose attestation/provenance information
- Maintain duplicate files

[PEP 751](https://peps.python.org/pep-0751/) defines a standardized lockfile format that addresses these issues.

## Key Design Decisions

1. **Source-only mode by default** - Consistent with requirements.txt behavior, avoiding wheel/sdist mismatches
2. **Reuse PipRequirement infrastructure** - PyLockPackage extends existing base class for downloads, hash validation, and Rust handling
3. **Support all package kinds** - PyPI, VCS, URL, and directory packages
4. **Preserve attestations** - Store PEP 740 attestation-identities in SBOM

## Review Process

**First commit**: Design document only - please review the design before implementation proceeds.

Implementation commits will be added after design approval.

### DIFF
--- a/docs/design/pylock.md
+++ b/docs/design/pylock.md
@@ -1,0 +1,277 @@
+# PEP 751 pylock.toml Support
+
+## Background
+
+Hermeto currently supports `requirements.txt` for Python dependencies. However,
+users increasingly generate lockfiles with tools like poetry, pipenv, and uv,
+which produce incompatible formats (`poetry.lock`, `Pipfile.lock`, `uv.lock`).
+
+This forces Hermeto users to:
+
+- Generate `requirements.txt` from their tool's lockfile as an intermediate
+  step
+- Lose metadata that requirements.txt cannot represent (e.g. per-artifact
+  hashes)
+- Manually maintain two files (lockfile + requirements.txt)
+
+Additionally, using `requirements.txt` as a pseudo-lockfile has limitations:
+
+- Hashes are line-based, not per-artifact (cannot distinguish wheel vs sdist
+  hashes)
+
+### Why Add pylock.toml Support
+
+Adding pylock.toml as a supported format alongside requirements.txt is
+valuable for several reasons:
+
+- New projects may prefer an ecosystem-agnostic lockfile that will
+  ultimately be supported by all packaging tools, rather than
+  vendor-locking themselves to a particular tool — including Hermeto
+  itself.
+- Existing projects still using requirements files might switch directly
+  to pylock.toml as the natural choice when transitioning to a more
+  modern way of declaring their dependencies.
+- The format follows a well-defined, versioned schema (unlike
+  tool-specific lockfiles), so the parser is unlikely to break due to
+  unannounced upstream changes.
+
+### Ecosystem Readiness
+
+[PEP 751](https://peps.python.org/pep-0751/) was accepted on 2025-03-31 with
+status Final. Current tool support:
+
+- **pip**: can generate and install from pylock.toml
+- **uv**: can generate and install pylock.toml
+- **PDM**: can generate, export, and install from pylock.toml
+- **Poetry**: support planned but not yet implemented
+
+> [!NOTE]
+> There isn't a wide-spread adoption of the agnostic lockfile at the time of
+> writing and most projects do not ship a pylock.toml yet, and
+> users will likely generate one from their existing toolchain. Given that
+> Hermeto would strictly adhere to the accepted spec and the format is
+> versioned, we can implement this as a supported (non-experimental) feature.
+
+### PEP 751 Standard
+
+[PEP 751](https://peps.python.org/pep-0751/) defines `pylock.toml`, a
+standardized ecosystem-agnostic lockfile format. Key features Hermeto needs:
+
+1. Format versioning: `lock-version` field enables Hermeto to handle
+   schema evolution
+2. Per-artifact hashes: Hashes nested in distribution objects (wheel,
+   sdist, archive)
+3. Multiple package sources: PyPI, VCS, direct URLs, local directories
+4. TOML structure: Easy to parse, better than line-based text
+
+See the [PEP 751 specification](https://peps.python.org/pep-0751/) for full
+field definitions and TOML schema.
+
+### Lockfile Structure
+
+```toml
+lock-version = "1.0"
+
+[[packages]]
+name = "foo"
+version = "1.0.0"
+sdist = { url = "https://...", hashes = { "sha256" = "..." } }
+wheels = [
+    { url = "https://...", hashes = { "sha256" = "..." } }
+]
+```
+
+### Package Kinds Hermeto Must Support
+
+#### PyPI Packages
+
+```toml
+[[packages]]
+name = "foo"
+version = "1.0.0"
+index = "https://pypi.org/simple/"  # Optional, defaults to PyPI
+sdist = { url = "...", hashes = { "sha256" = "..." } }
+wheels = [{ url = "...", hashes = { "sha256" = "..." } }]
+```
+
+#### VCS Packages
+
+```toml
+[[packages]]
+name = "foo"
+
+[packages.vcs]
+type = "git"
+url = "https://github.com/user/repo.git"
+commit-id = "abc123"
+subdirectory = "subpkg"  # optional, project root within the repo
+```
+
+#### URL Packages
+
+```toml
+[[packages]]
+name = "foo"
+archive = { url = "https://example.com/pkg.tar.gz", hashes = { ... } }
+```
+
+#### Directory Packages
+
+```toml
+[[packages]]
+name = "foo"
+directory = { path = "./local-pkg" }
+```
+
+#### Attestation Metadata
+
+Packages can include `attestation-identities` which is signed provenance
+metadata linking published artifacts to their source repository and build
+system. Verifying or otherwise acting on these attestations is out of scope.
+
+## Design
+
+### Parser Architecture
+
+Introduce a new pylock parser for TOML parsing while reusing the existing
+pip download and processing infrastructure (hash verification, async
+downloads, Rust extension detection, SBOM generation). The new parser
+produces the same data structures that the download pipeline expects,
+so the fetching and verification logic remains shared.
+
+Two other alternatives were considered:
+
+- **Extending the requirements.txt parser**: TOML and line-based formats
+  are fundamentally different, nested hash extraction becomes awkward,
+  and it mixes two unrelated parsing responsibilities without a shared
+  abstraction to justify it.
+- **Separate package manager with dedicated download logic**: introduces
+  more code to maintain with no clear benefit over reusing the existing
+  pip download pipeline.
+
+### Key Design Decisions
+
+#### 1. Lockfile Selection
+
+With multiple Python toolchains and a standardized format that is still
+early in adoption, we cannot safely assume which lockfile the user intends
+Hermeto to consume. A project may have both a tool-specific lockfile and
+`pylock.toml`, or may be experimenting with the format. Rather than
+adding format-specific input fields (`pylock_files`, `uv_lock_files`,
+etc.), two new generic fields are introduced: `lockfile` and `lockfile_extras`
+for additional files such as separately locked build dependencies.
+
+When the lockfile uses a default name (`pylock.toml`, `uv.lock`,
+`poetry.lock`, `pdm.lock`, etc.), Hermeto auto-detects the packaging
+tool from the filename. A `packaging_tool` field is available as an
+override for cases where the lockfile has a non-default name (e.g. a
+renamed or generated file).
+
+The existing `requirements_files` and `requirements_build_files` fields
+are preserved for backwards compatibility. The new fields take
+precedence when both are specified. The old fields should be deprecated
+in a future release once migration is complete.
+
+Example input:
+
+```json
+{
+  "type": "pip",
+  "lockfile": "pylock.toml",
+  "lockfile_extras": ["pylock.build.toml"],
+  "packaging_tool": "pylock"
+}
+```
+
+#### 2. Version Tolerance
+
+Hermeto needs to be aware of the `lock-version` field. An unknown major
+version should yield a fatal error.
+
+#### 3. Hash Verification
+
+pylock.toml nests hashes per distribution object (sdist, wheels, archive),
+so Hermeto always knows which artifact it is fetching and can verify against
+exactly that artifact's hashes.
+
+### Markers
+
+Markers are optional in pylock.toml and attach environment conditions
+(OS, Python version, etc.) to packages. They could be used to skip
+fetching packages not needed for a given target platform. However,
+platform-specific artifact selection is already handled by Hermeto's
+binary filters, which use wheel filename tags
+([PEP 491](https://peps.python.org/pep-0491/)) to select matching
+wheels. Markers operate at the package level rather than the artifact
+level, but since Hermeto does not know the target platform at fetch
+time, all packages are fetched unconditionally.
+
+> [!NOTE]
+> Given the pipeline stage where Hermeto operates and due to the existing binary
+> filtering logic environment markers are out of scope.
+
+### Build Dependencies
+
+Hermeto distinguishes between runtime and build dependencies. This mirrors
+how requirements.txt handles build dependencies via a separate
+`requirements-build.txt` file. Build dependencies are:
+
+- Fetched and verified the same way as runtime dependencies
+- Included in the output and SBOM
+- Kept separate so Hermeto can distinguish them in the build config
+
+PEP 751 does not solve locking build requirements for sdists (this was
+[explicitly deferred](https://peps.python.org/pep-0751/#locking-build-requirements-for-sdists)
+from the PEP). Users must generate a separate lockfile for their build
+dependencies. PEP 751 supports named lockfiles (`pylock.<name>.toml`),
+so a natural convention could be `pylock.build.toml`.
+
+**NOTE:** Users can generate this file using
+[pybuild-deps](https://github.com/hermetoproject/pybuild-deps), which
+resolves build dependencies from `pyproject.toml`. pybuild-deps needs to be
+enhanced to support pylock.toml output, generating a named lockfile
+like `pylock.build.toml`. Hermeto consumes it via the `lockfile_extras`
+input field.
+
+### `created-by`
+
+pylock.toml records the generating tool in the `created-by` field.
+Different installers use different environment variables for build
+configuration (e.g. `PIP_FIND_LINKS` for pip). Hermeto can use this
+field to select the appropriate installer and its corresponding
+environment variables when configuring the build environment.
+
+### Integration
+
+- Input: Add `lockfile`, `lockfile_extras`, and `packaging_tool` to
+  `PipPackageInput`
+- Downloads: PyPI, VCS, and URL packages reuse the existing download logic.
+  Directory packages require new handling or rejection since the current
+  implementation does not support them.
+- SBOM: Preserve attestation-identities when present
+
+### Security Considerations
+
+1. Path traversal: Validate directory package paths stay within repo
+2. Hash verification: All downloads verified against the specific
+   artifact's hashes from the lockfile
+3. Attestation preservation: Store in SBOM (verification is future work)
+4. Trusted sources: VCS/directory are trusted; URL requires hashes
+
+## Follow-up
+
+- **Extras / dependency group filtering**: pylock.toml supports extras
+  and dependency groups, allowing lockfiles to cover multiple installation
+  profiles. Users should be able to specify which to fetch via the input
+  JSON to avoid fetching unnecessary packages.
+- **Attestation metadata**: pylock.toml supports `attestation-identities`
+  ([PEP 740](https://peps.python.org/pep-0740/)) as package metadata.
+  Hermeto could preserve these in SBOM output. Adoption of PEP 740 is
+  still early, and existing conversions from other formats to pylock.toml
+  may not preserve this metadata.
+
+## References
+
+- [PEP 751](https://peps.python.org/pep-0751/) - Lockfile specification
+- [PEP 740](https://peps.python.org/pep-0740/) - Attestation identities
+- [uv documentation](https://docs.astral.sh/uv/) - Reference implementation


docs/design/pylock.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### 2. Rust Extension Handling
+
+Pylock handles Rust extensions the same way as requirements.txt: vendor Cargo
+dependencies from source distributions, and fail if Cargo.lock is out of sync.
+
+Behavior:
+
+- Download source distributions (sdists) by default
+- Detect Rust extensions (setuptools-rust, ruff)
```
taylormadore: I think ruff here should be maturin instead

docs/design/pylock.md
```diff
@@ -0,0 +1,269 @@
...
+#### Option 3: Hybrid approach (preferred)
+
+- Pro: Reuses download/processing logic from PipRequirement
+- Pro: Clean separation for TOML parsing
+- Pro: Attestations are built into the TOML format
+- Con: More code than extending requirements.txt parser
+
+### Key Design Decisions
+
+#### 1. Binary Mode Default
```
taylormadore: What are we deciding here? Same question for the one below

docs/design/pylock.md
```diff
@@ -0,0 +1,269 @@
...
+  parsing pylock.toml TOML format.
+
+This design allows pylock to reuse existing pip infrastructure (download
+handling, hash verification, Rust extension detection) while providing a
+separate parser for the TOML-based format.
+
+### Components
+
+1. `PyLockFile` - Parses pylock.toml, validates version, extracts packages
+2. `PyLockPackage` - Extends PipRequirement with pylock-specific parsing
```
taylormadore: Which aspects of PipRequirement are helpful for us to reuse? It looks pretty tightly coupled to the old requirements format to me

docs/design/pylock.md
```diff
@@ -0,0 +1,269 @@
...
+- Downloads: Route packages by kind (pypi, vcs, url, directory)
+- SBOM: Preserve attestation-identities
+
+### Processing Flow
+
+1. `fetch_pip_source()` checks for pylock files
+2. If pylock files exist:
+   - Call `_download_from_pylock_files()`
+   - Route packages by kind to appropriate processors:
+     - PyPI packages -> `_process_pypi_req()`
```
taylormadore: Does processing registry packages the same way as for pip make sense with the new information we have in pylock.toml? We have fully resolved URLs for sdists and wheels already, for example. Do we need to query pypi like _process_pypi_req does?

docs/design/pylock.md
```diff
@@ -0,0 +1,269 @@
...
+- C. Merge both
+  - Con: Defeats lockfile purpose
+  - Con: Complex semantics
+
+Preferred approach: B. Pylock takes precedence
+
+Rationale:
+
+- Users adding pylock.toml want hermeto to use it
+- Clear migration: add pylock.toml - hermeto auto-detects it
```
a-ovchinnikov: What about those users who are not in control of a package they are building? A package maintainers are likely unaware of Hermeto.

docs/design/pylock.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### 5. Hash Extraction
+
+Challenge: Pylock nests hashes in distribution objects (wheels, sdist,
+archive), unlike requirements.txt where hashes are top-level.
+
+Approach:
+
+- Extract hashes from all distribution types
+- Flatten into `PipRequirement.hashes` list
```
a-ovchinnikov: This one is not clear. Basing on what is written above I assumed that .toml contains a mapping from objects to their hashes, this flattening seems to loose this data. Is it not necessary?

ST3V1K: It's the same situation as with requirements.txt. We can have multiple hashes per requirement and we don't know which belongs to which artifact either. We just need to verify that the hash of the downloaded artifact is in the list.

docs/design/pylock.md
ST3V1K: Package maintainers don't need to know about Hermeto. If the project doesn't ship a pylock.toml, users can generate one themselves from whatever lockfile the project provides. If that's not an option, requirements.txt still works as before.

docs/design/pylock.md
ST3V1K: If a package has an sdist, we could process it as a URL package since we already have the resolved URL. But some tools like pip won't generate pylock with sdists - they only generate wheels. And whenever there are multiple wheels, there's no way to select the correct one without querying pypi for distribution metadata. So processing registry packages through `_process_pypi_req()` is still necessary.

docs/design/pylock.md
ST3V1K: PipRequirement is indeed tightly coupled to the requirements format, especially the parsing methods (from_line(), etc.). Most of the fields (package name, version, hashes, kind, qualifiers) will be the same in PyLockPackage, but the methods won't really be reused - PyLockPackage parses from TOML dicts, not requirement lines. The main reuse value is in the processing functions: _process_pypi_req(), _process_vcs_req(), and _process_url_req(), which handle downloading, hash validation, and SBOM generation regardless of the input format. Ideally, PipRequirement and PyLockPackage should share a common superclass that holds the shared fields, with each subclass owning its own parsing logic. That would require refactoring the pip processing pipeline. It might be nice to do as a follow-up.

docs/design/pylock.md
ST3V1K: You're right, these aren't decisions. I ran into issues with binary filters and Rust handling while experimenting with pylock.toml - the easy fix would have been to change the processing, but I realized the correct approach is to keep the behavior consistent with requirements.txt. I documented them to capture that reasoning, but they don't belong in the design doc since there's nothing pylock-specific about them.
Removed both sections.

docs/design/pylock.md
ST3V1K: Good catch, the wording was a bit confusing. Hermeto detects all packages with Rust code, so `ruff` should still be there.
This section was removed based on the comment below.

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
+# PEP 751 pylock.toml Support
+
+## Problem
+
+Hermeto currently supports `requirements.txt` for Python dependencies. However,
+users increasingly generate lockfiles with tools like poetry, pipenv, and uv,
+which produce incompatible formats (`poetry.lock`, `Pipfile.lock`, `uv.lock`).
```
eskultety, line 7: This is true, by all means, but it also begs the question - if this is what our users use, why don't we add support for these individual backends instead if they're so widely used? I mean, the doc should IMO have a stronger argument on why exactly `pylock.toml` instead of e.g. `poetry.lock`.

Hint: it's standardized and eventually all ecosystems should embrace it and support it fully.

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+
+This forces hermeto users to:
+
+- Generate `requirements.txt` from their tool's lockfile as an intermediate
+  step
+- Lose attestation/provenance information in the conversion
```
eskultety: How exactly do they lose attestation & provenance? I presume this is about PEP 740 mentioned below in the doc?
I'm wondering how many of the widely used ecosystems actually make use of PEP 740. In any case, if you maintain both the original file and `requirements.txt` in your repo, does it count as invalidating OR losing PEP 740 metadata due to conversion?

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+- Manually maintain two files (lockfile + requirements.txt)
+
+Additionally, using `requirements.txt` as a pseudo-lockfile has limitations:
+
+- No format versioning (hermeto cannot detect schema changes)
+- Hashes are line-based, not per-artifact (cannot distinguish wheel vs sdist
```
eskultety, line 19: Hmm, does this pose a real practical issue for Hermeto? When we verify a fetched artifact against the hashes, the hashes are stored in a map so O(1), so I'm wondering why is a flat list of hashes attached to a plain dep (not a distributible) a limitation :) .

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+  hashes)
+- No native attestation support (PEP 740)
+
+## Goals
+
+Support a standardized lockfile format that tools can generate directly,
+preserving all metadata.
```
eskultety: Note, **not ALL** metadata can be preserved, because some ecosystem manifest lockfiles have a slightly higher expressive power (e.g. `uv.lock` vs `pylock.toml`).

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+
+## PEP 751 Standard
+
+[PEP 751](https://peps.python.org/pep-0751/) defines `pylock.toml`, a
+standardized lockfile format. Key features hermeto needs:
```
eskultety: "standardized ecosystem-agnostic" (or toolchain agnostic)

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+
+#### Directory Packages
+
+```toml
+[[packages]]
+name = "foo"
+directory = { path = "./local-pkg" }
+```
+
+#### Attestation Support
```
eskultety: This is listed under package types. Is it really a package type? To me attestation is just metadata. Why does Hermeto need to support attestation metadata right away? I haven't studied PEP 740 nor haven looked deep into its integration to `pylock.toml`, so my question is - is there anything in particular about this metadata that we ought to support right away or is there anything that would make our SBOM generating process significantly easier and cheaper than using our current approach taken over from pip?

Additionally, you mention that conversions lose attestation metadata in case of requirements.txt. Note that we're still far away from wide adoption of this new lockfile format and so we should look into whether existing conversions from <anything> -> `pylock.toml` actually CAN preserve the metadata, because that's what's going to be the common use case for a little while IMO.

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+]
+```
+
+Hermeto should add these to SBOM output.
+
+## Design Considerations
+
+### Parser Architecture
+
+#### Option 1: Extend requirements.txt parser
```
eskultety: Shouldn't there also be some "separation of intent" in the logic? Extending the existing parser without a superclass to inherit from would IMO become weird to read and harder to maintain, don't have a good supporting argument for my claim.

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+- Pro: Clean separation, follows PEP 751 spec naturally
+- Pro: Attestations are built into the TOML format
+- Con: More code to maintain
+
+#### Option 3: Hybrid approach (preferred)
+
+- Pro: Reuses download/processing logic from PipRequirement
+- Pro: Clean separation for TOML parsing
+- Pro: Attestations are built into the TOML format
+- Con: More code than extending requirements.txt parser
```
eskultety: I think this is a natural expectation, but fair enough.

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Key Design Decisions
+
+#### 1. Lockfile Priority
+
+Question: When both pylock.toml and requirements.txt exist, which to use?
```
eskultety: I think this is quite simple to answer - `pylock.toml` has a much bigger expressive power than `requirements.txt`, is standardized and more modern :). However that's only part of the picture. The other question to ask is - What do we do if we process projects from different Python ecosystems that we'll support in the future (e.g. uv) and we'll find both e.g. `uv.lock` and `pylock.toml`? `uv.lock` has a slightly stronger expressive power (though it's questionable Hermeto would benefit from it). The answer is that we should probably not assume any defaults at all and simply require the user to specify which "lockfile" to use (with a new input JSON field!).

docs/design/pylock.md
eskultety: > Package maintainers don't need to know about Hermeto. If the project doesn't ship a pylock.toml, users can generate one themselves from whatever lockfile the project provides. If that's not an option, requirements.txt still works as before.

See my comment above, with so many Python ecosystems and now a standardized format to rule them all (with little to no adoption) we can't assume anything safely IMO, the project might just be experimenting with the format to see if it's going to work for their use case --> we'll need a new input JSON config field.

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+
+```python
+SUPPORTED_LOCK_MAJOR = 1
+SUPPORTED_LOCK_MINOR = 0
+
+if version.major != SUPPORTED_LOCK_MAJOR:
+    raise InvalidLockfileFormat(...)
+if version.minor != SUPPORTED_LOCK_MINOR:
+    log.warning("Unknown minor version, continuing...")
+```
```
eskultety: I think this section doesn't need any technical details or examples, just say we need to be aware and major versions should yield a fatal error.

docs/design/pylock.md
eskultety: > We just need to verify that the hash of the downloaded artifact is in the list.

But in case of `pylock.toml` we should know what type of redistributible we're fetching, right? Why do we need to flatten the list if the hashes are natively nested under the redistributible types in the lockfile?

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+
+- `PipRequirement`: Base class providing download, hash validation, and
+  processing logic. Has a `from_line()` method for parsing requirements.txt
+  format.
+- `PyLockPackage`: Extends PipRequirement with a `from_dict()` method for
+  parsing pylock.toml TOML format.
+
+This design allows pylock to reuse existing pip infrastructure (download
+handling, hash verification, Rust extension detection) while providing a
+separate parser for the TOML-based format.
```
eskultety: I think this is an implementation detail, I don't think a design needs to mandate a particular code path towards accomplishment because the code can change easily through a refactor, invalidating the doc instantly which is the harder thing to keep relevant.

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Components
+
+1. `PyLockFile` - Parses pylock.toml, validates version, extracts packages
+2. `PyLockPackage` - Extends PipRequirement with pylock-specific parsing
+3. `_download_pylock_dependencies()` - Entry point for pylock processing
```
eskultety: Same here, implementaiton details.

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Integration
+
+- Input: Add `pylock_files` and `pylock_build_files` to `PipPackageInput`
```
eskultety: This is the first mention of `pylock_build_files`. I think the doc should cover handling of build dependencies in a dedicated section to a bit more depth.

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+2. If pylock files exist:
+   - Call `_download_from_pylock_files()`
+   - Route packages by kind to appropriate processors:
+     - PyPI packages -> `_process_pypi_req()`
+     - VCS packages -> `_process_vcs_req()`
+     - URL packages -> `_process_url_req()`
+     - Directory packages -> `_process_directory_req()`
+3. If no pylock files, call `_download_from_requirement_files()`
+4. Both paths converge at `filter_packages_with_rust_code()`
+5. Return `RequestOutput`
```
eskultety: Implementation details.

docs/design/pylock.md
```diff
@@ -0,0 +1,235 @@
...
+sdist = { url = "...", hashes = { "sha256" = "..." } }
+wheels = [{ url = "...", hashes = { "sha256" = "..." } }]
+```
+
+#### VCS Packages
+
+```toml
+[[packages]]
+name = "foo"
+vcs = { url = "https://github.com/user/repo.git", rev = "abc123" }
```
eskultety: Does this follow the PEP actually? https://peps.python.org/pep-0751/#packages-vcs-commit-id
`rev` should IMO be `commit-id`. Maybe also mention https://peps.python.org/pep-0751/#packages-vcs-subdirectory and https://peps.python.org/pep-0751/#packages-vcs-type?

eskultety (changes requested):
A few more notes I did not (could not raise inline):

- the doc should probably follow the structure of our existing design docs a bit closer (section naming, composition, etc.)
- I'm missing mentions and their relevance to Hermeto of the following bits: 
  * markers (relevance to binary filtering?),
  * environments (relevance to other future backends, can be mentioned as a follow-up item to explore in the future!),
  * `created-by` (relevance to build env variable generating, e.g. `PIP_FIND_LINKS`),
  * extras / dependency groups (which ones does the user need/want to fetch for their build?)
  * named lockfiles, e.g. `pylock.<name>.toml`?
  * spec limitations relevance to Hermeto, e.g. build deps have not been solved yet: https://peps.python.org/pep-0751/#locking-build-requirements-for-sdists, how do users specify their build deps with pylock? We should think of a way to enhance the `build-deps` tool to generate a named lockfile `pylock.build.toml`

Last thing that should be mentioned is actual support of this PEP across Python ecosystems, i.e. which toolchains can generate, read, an install from `pylock.toml`. Are we too early, meaning that is there anything that could force us to introduce this support as experimental only or are we good to implement straight away as a supported enhancement because we'd just adhere to the spec?
^These are all aspects I'd like to see covered in a design doc. Note I'm not saying we must implement everything from day one, but answering all ^these may further shape the exact direction we'll take when we proceed with an implementation.

docs/design/pylock.md
ST3V1K: Added a section about this

docs/design/pylock.md
ST3V1K: Yes, this was about the PEP 740 and specifically about `[[packages.attestation-identities]]` field. Currently, it's not widely used, so I added it as a follow-up topic to look at in the future.

docs/design/pylock.md
ST3V1K: Hashes are actually a list[str], not a map, but that's not the point.
The limitation is about precision, not performance. With requirements.txt, all hashes are in a flat list - you don't know which belongs to the sdist and which to a wheel. When Hermeto downloads a specific artifact, it just checks if the hash is somewhere in that list. With pylock.toml, hashes are nested per distribution object, so we verify against exactly the artifact we're fetching. Both approaches work, but per-artifact hashes are strictly more precise.

docs/design/pylock.md
ST3V1K: Yes, that's true, but this was mostly about the metadata already captured in the `pylock.toml`
I removed this section, so this is now besides the point.

docs/design/pylock.md
ST3V1K: Done

docs/design/pylock.md
ST3V1K: Moved attestation support/metadata as a follow-up

docs/design/pylock.md
ST3V1K: Yes, only extending the existing parser is a bad way to do this. This would probably have to include some superclass for both the old and the new parser. 
I still prefer the hybrid approach, where there would be a completely separate parser, and it would generate a compatible output with the current processing structure.

docs/design/pylock.md
ST3V1K: Removed it

docs/design/pylock.md
ST3V1K: I have rewritten the section about lockfile selection. This should now allow the users to select any "lock-file" type (considering it is supported).

docs/design/pylock.md
ST3V1K: Removed the details

docs/design/pylock.md
ST3V1K: Removed

docs/design/pylock.md
ST3V1K: Removed

docs/design/pylock.md
ST3V1K: Created a whole section about build dependencies

docs/design/pylock.md
ST3V1K: Removed

docs/design/pylock.md
ST3V1K: Yes, thank you. It should have been `commit-id`

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+configuration (e.g. `PIP_FIND_LINKS` for pip). Hermeto can use this
+field to select the appropriate installer and its corresponding
+environment variables when configuring the build environment.
+
+### Integration
+
+- Input: Rename `requirements_files` and `requirements_build_files` to
+  `lockfiles` and `build_lockfiles`, add `lockfile_format` to
+  `PipPackageInput`
+- Downloads: Route packages by kind (PyPI, VCS, URL, directory)
```
eskultety: Can you elaborate on ^this routing for downloads?

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Implementation Scoping
+
+Phased approach:
+
+1. **Phase 1**: Core parser, PyPI package fetching, hash verification,
+   lockfile selection via input JSON
+2. **Phase 2**: VCS, URL, and directory package support
+3. **Phase 3**: Attestation metadata preservation in SBOM, build
+   dependency support via `build_lockfiles`
```
eskultety: This is a design, it should not mandate a particular implemnetation approach. We should only ever clearly state what's "out of scope" for delivering the feature, in this case - attestation metadata preservation.

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Follow-up
+
+- **Environments**: pylock.toml can list environment markers indicating
+  which platforms and Python versions the lockfile targets. Not relevant to
```
eskultety: So, knowing what environment markers are, how could those be useful to Hermeto actually? Based on my comment above I think we can drop this bit.

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+
+- Fetched and verified the same way as runtime dependencies
+- Included in the output and SBOM
+- Kept separate so Hermeto can distinguish them in the build config
+
+PEP 751 does not solve locking build requirements for sdists (this was
+[explicitly deferred](https://peps.python.org/pep-0751/#locking-build-requirements-for-sdists)
+from the PEP). Users must generate a separate lockfile for their build
+dependencies. PEP 751 supports named lockfiles (`pylock.<name>.toml`),
+so a natural convention is `pylock.build.toml`.
```
eskultety: `s/convention is/convention could be`

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+
+Users can generate this file using
+[pybuild-deps](https://github.com/hermetoproject/pybuild-deps), which
+resolves build dependencies from `pyproject.toml`. pybuild-deps should
```
eskultety: `s/should be enhanced/needs to be enhanced`

eskultety: Also, this IMO deserves a nice rendered Note field.

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+Markers are optional in pylock.toml and attach environment conditions
+(OS, Python version, etc.) to packages. They could be used to skip
+fetching packages not needed for a given target platform. However,
+platform-specific artifact selection is already handled by Hermeto's
+binary filters, which use wheel filename tags
+([PEP 491](https://peps.python.org/pep-0491/)) to select matching
+wheels. Markers operate at the package level rather than the artifact
+level, but since Hermeto does not know the target platform at fetch
+time, all packages are fetched unconditionally. This is consistent
+with the requirements.txt behavior.
```
eskultety: _This is consistent with the requirements.txt behavior._

Can't help myself, but this feels like a typical AI sentence. Can we just say something like "Given the pipeline stage where Hermeto operates and due to the existing binary filtering logic environment markers are out of scope."

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+
+With multiple Python toolchains and a standardized format that is still
+early in adoption, we cannot safely assume which lockfile the user intends
+Hermeto to consume. A project may have both a tool-specific lockfile and
+`pylock.toml`, or may be experimenting with the format. Rather than
+adding format-specific input fields (`pylock_files`, `uv_lock_files`,
+etc.), the existing `requirements_files` and `requirements_build_files`
+fields should be renamed to generic `lockfiles` and `build_lockfiles`.
+A new `lockfile_format` field allows the user to specify the format
+explicitly, since lockfiles can be arbitrarily named.
```
eskultety: 1. _fields should be renamed to generic `lockfiles` and `build_lockfiles`._
Renaming fields is a backwards incompatible change, we should add new files and potentially consider if the old ones need to be deprecated
2. `lockfiles` - does it need to be plural? Can we keep the old fields around and reserve this field only for the other packaging tools? Looking at the rest of the tools there always only seems to be a single file - `uv.lock`, `pylock.toml`, `poetry.lock`, `pdm.lock`, etc.
3. `build_lockfiles` - probably an unfortunate name IMO, while build deps is definitely a problem none of the tools solves, maybe we can call this more defensively just `lockfile_extras`? We don't need to find a perfect name in a design, it's not binding in that aspect.
4. `lockfile_format` - so what's the intent here? Having just TOML, YAML, JSON, TXT (also, txt is not a format) ? If so, then how do you differentiate between which parser should be used for `pylock` and `uv` if the file doesn't follow standard naming convention? I think what we might want instead is a `packaging_tool` field to help us out which parser to utilize if `lockfile` doesn't use the default name in which case we should detect it automatically.

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+pylock.toml (PEP 751). Several major tools (pip, uv, PDM) already support it,
+and others (Poetry) have it planned. One implementation will cover the ecosystem
+both now and as new tools emerge.
+
+### Ecosystem Readiness
+
+[PEP 751](https://peps.python.org/pep-0751/) was accepted on 2025-03-31 with
+status Final. Current tool support:
+
+- **pip**: can generate and install from pylock.toml
```
eskultety, line 43: Can pip really install from pylock? https://github.com/pypa/pip/issues/13864#issuecomment-4119524922

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+- **uv**: can generate pylock.toml
```
eskultety: UV seems to claim "some" install support as well: https://docs.astral.sh/uv/concepts/projects/layout/#relationship-to-pylocktoml

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+- **PDM**: can generate, export, and install from pylock.toml
+- **Poetry**: support planned but not yet implemented
+
+Adoption is still early. Most projects do not ship a pylock.toml yet, and
```
eskultety: _Adoption is still early_

"There isn't a wide-spread adoption of the agnostic lockfile at the time of writing and most projects..."

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+- No native attestation support
+
+### Why pylock.toml Instead of Individual Backend Lockfiles
+
+Rather than writing and maintaining N parsers for N backend-specific lockfile
+formats (poetry.lock, Pipfile.lock, uv.lock, pdm.lock, etc.), each with its
+own schema that can change without notice, we implement a single parser against
+pylock.toml (PEP 751). Several major tools (pip, uv, PDM) already support it,
+and others (Poetry) have it planned. One implementation will cover the ecosystem
+both now and as new tools emerge.
```
eskultety: When I said "I'd like to see a section on why pylock.toml" I should have been more clear in my expectations. This isn't quite what I had in mind, because what it states is simply not true - I guess we'll sooner implement those N parsers than `pylock.toml` becomes widely adopted and with some tools it won't ever become the preferred option because its expressive power currently is less than what the other tools' files are providing.

> Rather than writing and maintaining N parsers for N backend-specific lockfile
> formats (poetry.lock, Pipfile.lock, uv.lock, pdm.lock, etc.), each with its
> own schema that can change without notice, we implement a single parser against
> pylock.toml

This also feels like written by an AI (yes, I did prompt it and it spat something very similar)

My take on "Why adding pylock.toml support" which is kinda similar but from a slightly different angle:
- emerging projects may prefer an agnostic file that will ultimately be supported by all packaging tools over vendor-locking themselves to particular tooling, **including Hermeto itself**
- existing projects still using requirements files might also switch directly to pylock.toml as the obvious choice when transitioning to a more modern way of declaring one's dependencies
- the file follows a well defined versioned schema unlike the rest so we can be confident in our parser not breaking easily (you have this one covered)

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+```toml
+[[packages]]
+name = "foo"
+directory = { path = "./local-pkg" }
+```
+
+#### Attestation Metadata
+
+Packages can include `attestation-identities` (PEP 740) as metadata.
+Hermeto preserves these in SBOM output when present. Attestation
```
eskultety: > Hermeto preserves these in SBOM output when present.

If it's out of scope, then just briefly describe what PEP 740 is about  instead of just mentioning the PEP (the PEP is quite long, save readers some time :) ) and then say it's out of scope (without initial implementation...) . Our design docs are living documents, so they should always reflect the current state (except when we're introducing support and have literally nothing).

docs/design/pylock.md
```diff
@@ -0,0 +1,283 @@
...
+- Con: Mixes two unrelated parsing responsibilities in one place without
+  a shared abstraction to justify it
+
+#### Option 2: Separate pylock parser
+
+- Pro: Clean separation, follows PEP 751 spec naturally
+- Pro: Attestations are built into the TOML format
+- Con: More code to maintain
+
+#### Option 3: Hybrid approach (preferred)
```
eskultety: Visually these all appear with the same level of importance. I'd start with the last one and then just mention that other alternatives were considered and list the cons that led us to pick a particular solution.

eskultety (changes requested):

docs/design/pylock.md
ST3V1K: Each package will be downloaded in a different way, same/similar to the current implementation of downloading requirements (from `requirements.txt`). The only difference is the `directory` kind. This will have to be processed differently or rejected.

docs/design/pylock.md
ST3V1K: As of pip 26.1, it can directly install from a pylock.toml with `pip install -r pylock.toml`. However, this is still an experimental feature.

https://pip.pypa.io/en/stable/cli/pip_lock/#cmdoption-r

eskultety: Heh, when I wrote my comment 26.1 was only an RC: https://discuss.python.org/t/announcement-pip-26-1-release/107108 happened ~17h ago ;) .

docs/design/pylock.md
ST3V1K: Removed

docs/design/pylock.md
ST3V1K: Removed

docs/design/pylock.md
ST3V1K: Done

docs/design/pylock.md
ST3V1K: 1) Yes, thank you. Changed it to introduce new fields instead of renaming the old ones.
2) It doesn't have to be. I wanted it to match the `requirements_files`.
3) Yes, `lockfile_extras` is better.
4) The intended behavior was exactly what you specified by the field `packaging_tool`, so I just renamed it and added an explanation.

docs/design/pylock.md
ST3V1K: I may have read that in some early docs or something, but I knew that I saw it somewhere :)

docs/design/pylock.md
ST3V1K: Thank you, I missed that part in docs

docs/design/pylock.md
ST3V1K: Sorry for that, it made sense to me and looked like a valid reason.

docs/design/pylock.md
```diff
@@ -0,0 +1,275 @@
...
+
+- Generate `requirements.txt` from their tool's lockfile as an intermediate
+  step
+- Lose metadata that requirements.txt cannot represent (e.g. per-artifact
+  hashes)
+- Manually maintain two files (lockfile + requirements.txt)
+
+Additionally, using `requirements.txt` as a pseudo-lockfile has limitations:
+
+- No format versioning (Hermeto cannot detect schema changes)
```
a-ovchinnikov: Are there supposed to be any version changes? I'd say the process being an ad-hoc manual intervention is a justification enough for replacing it.

docs/design/pylock.md
```diff
@@ -0,0 +1,275 @@
...
+- Hashes are line-based, not per-artifact (cannot distinguish wheel vs sdist
+  hashes)
+
+### Why Add pylock.toml Support
```
a-ovchinnikov, line 22: Non-blocking, just something to consider: does this section add much value to the doc? As I mentioned above just saying that this will save at least some users from doing extra steps should be a good enough reason for the addition.

docs/design/pylock.md
```diff
@@ -0,0 +1,275 @@
...
+
+Packages can include `attestation-identities` which is signed provenance
+metadata linking published artifacts to their source repository and build
+system. Verifying or otherwise acting on these attestations is out of scope.
+
+## Design
+
+### Parser Architecture
+
+Introduce a new pylock parser for TOML parsing while reusing the existing
```
a-ovchinnikov, line 136: Wording nitpick: a parser is not supposed to be downloading or processing anything except for a character stream. Did you mean that you want to introduce a generic Python package manager which could work with any project that has the necessary lock present?

docs/design/pylock.md
```diff
@@ -0,0 +1,275 @@
...
+pip download and processing infrastructure (hash verification, async
+downloads, Rust extension detection, SBOM generation). The new parser
+produces the same data structures that the download pipeline expects,
+so the fetching and verification logic remains shared.
+
+Two other alternatives were considered:
+
+- **Extending the requirements.txt parser**: TOML and line-based formats
+  are fundamentally different, nested hash extraction becomes awkward,
+  attestations would be bolted on, and it mixes two unrelated parsing
```
a-ovchinnikov: L130 makes one think that attestations are not a big problem, they are supposed to be ignored, aren't they?

docs/design/pylock.md
```diff
@@ -0,0 +1,275 @@
...
+  responsibilities without a shared abstraction to justify it.
+- **Fully separate parser with no shared infrastructure**: introduces
```
a-ovchinnikov: Parser is usually just one (interchangeable) element in a chain, did you mean "package manager"?

docs/design/pylock.md
```diff
@@ -0,0 +1,275 @@
...
+how requirements.txt handles build dependencies via a separate
+`requirements-build.txt` file. Build dependencies are:
+
+- Fetched and verified the same way as runtime dependencies
+- Included in the output and SBOM
+- Kept separate so Hermeto can distinguish them in the build config
+
+PEP 751 does not solve locking build requirements for sdists (this was
+[explicitly deferred](https://peps.python.org/pep-0751/#locking-build-requirements-for-sdists)
+from the PEP). Users must generate a separate lockfile for their build
```
a-ovchinnikov, line 225: Can Hermeto do it for them?

docs/design/pylock.md
```diff
@@ -0,0 +1,275 @@
...
+enhanced to support pylock.toml output, generating a named lockfile
+like `pylock.build.toml`. Hermeto consumes it via the `lockfile_extras`
+input field.
+
+### `created-by`
+
+pylock.toml records the generating tool in the `created-by` field.
+Different installers use different environment variables for build
+configuration (e.g. `PIP_FIND_LINKS` for pip). Hermeto can use this
+field to select the appropriate installer and its corresponding
```
a-ovchinnikov, line 241: Will it select basing on lock file name? If the lock file is universal would it be possible to use a different PM with it, e.g. have it generated by poetry and then use uv to install it? In this case wouldn't it make sense to ask users what they intend to use in build phase?

docs/design/pylock.md
eskultety: No, `pybuild-deps` should though.

docs/design/pylock.md
ST3V1K: It's true that it doesn't add any value, and since the format is the same and will most likely never change, it can be removed.

docs/design/pylock.md
ST3V1K: Since @eskultety requested it in https://github.com/hermetoproject/hermeto/pull/1458#discussion_r3119012282 I would just keep it

docs/design/pylock.md
ST3V1K: I wanted to say that the parser will return an object that could reuse the already existing system for downloading and processing pip requirements.

docs/design/pylock.md
ST3V1K: Yes, sorry. I forgot about this while moving the attestation support to follow-ups.

docs/design/pylock.md
ST3V1K: I meant that the pylock parser could produce its own type with new download/processing logic instead of reusing the existing one. This could technically count as a new package manager.

docs/design/pylock.md
ST3V1K: It should select from the file name if the file uses a default name (`pylock.toml` - pylock, `uv.lock` - uv etc.). Then there would be a `packaging_tool` CLI field in the JSON passed to `fetch-deps`. And if `pylock.toml` is used, then it could use the `created-by` field as well.

Part of this was suggested by @eskultety in https://github.com/hermetoproject/hermeto/pull/1458#pullrequestreview-4098405051
> I'm missing mentions and their relevance to Hermeto of the following bits: 
    ...
    `created-by` (relevance to build env variable generating, e.g. `PIP_FIND_LINKS`),
    ...

In the selection of the installer, we could treat `created-by` as the default, and we could add a new CLI field for users to override it if they want to use a different tool than the one that generated the lockfile. 

I hope I understood the comment correctly.

docs/design/pylock.md
a-ovchinnikov: Ok, let's keep it then.

docs/design/pylock.md
```diff
@@ -0,0 +1,274 @@
...
+- **pip**: can generate and install from pylock.toml
+- **uv**: can generate and install pylock.toml
+- **PDM**: can generate, export, and install from pylock.toml
+- **Poetry**: support planned but not yet implemented
+
+There isn't a wide-spread adoption of the agnostic lockfile at the time of
+writing and most projects do not ship a pylock.toml yet, and
+users will likely generate one from their existing toolchain. Given that
+Hermeto would strictly adhere to the accepted spec and the format is
+versioned, we can implement this as a supported (non-experimental) feature.
```
eskultety: This would be probably better formatted like this: https://github.com/hermetoproject/hermeto/blob/46b56a4dcec9b392ed56415e6280878217d65d4b/docs/adr/0002-yarn-v4.md?plain=1#L246

docs/design/pylock.md
```diff
@@ -0,0 +1,274 @@
...
+(OS, Python version, etc.) to packages. They could be used to skip
+fetching packages not needed for a given target platform. However,
+platform-specific artifact selection is already handled by Hermeto's
+binary filters, which use wheel filename tags
+([PEP 491](https://peps.python.org/pep-0491/)) to select matching
+wheels. Markers operate at the package level rather than the artifact
+level, but since Hermeto does not know the target platform at fetch
+time, all packages are fetched unconditionally. Given the pipeline stage
+where Hermeto operates and due to the existing binary filtering logic
+environment markers are out of scope.
```
eskultety: The last sentence also feels like it could be more visible with: https://github.com/hermetoproject/hermeto/blob/46b56a4dcec9b392ed56415e6280878217d65d4b/docs/adr/0002-yarn-v4.md?plain=1#L246

eskultety (commented):
A couple more nitpicks, but I'm good -> ACK. Now, we don't necessarily need to merge this immediately, we can proceed with the implementation and adjust this if needed based on how the implementation will be going.
If the overall consensus is to merge this ASAP before the implementation, fine with me.

eskultety (approved):

eskultety:
FWIW I like how we introduced the yarn v4 workspace focus, the implementation and design were happening in parallel :) .

docs/design/pylock.md
ST3V1K: Done

ST3V1K:
> Now, we don't necessarily need to merge this immediately, we can proceed with the implementation and adjust this if needed based on how the implementation will be going.
If the overall consensus is to merge this ASAP before the implementation, fine with me.

I agree, I planned to implement it in the same PR. But since this turned out to be much longer than I expected, I will create a new PR with the implementation. I might even change this PR to draft.

docs/design/pylock.md
ST3V1K: Done

docs/design/pylock.md
a-ovchinnikov: Thank you, I either misread the sentence or it became clearer at some point, resolving.

a-ovchinnikov (approved):

PR #1463, opened 2026-03-28
fix(bundler): handle JSONDecodeError and preserve stderr in parse_lockfile

### LINKED ISSUE #1462: bundler: incomplete error handling in parse_lockfile() loses stderr and misses JSONDecodeError
### Description

```markdown

`parse_lockfile()` in `hermeto/core/package_managers/bundler/parser.py` has two error handling gaps:

**1. `json.loads()` is outside the try-except block (line 51)**

The `json.loads(output)` call sits after the try-except that catches `CalledProcessError`, so if the Ruby lockfile parser emits non-JSON output (warnings, version mismatch messages, encoding issues), hermeto crashes with a raw `json.JSONDecodeError` traceback instead of a clean `PackageManagerError`.

Current code:

```python
try:
    output = run_cmd(cmd=[str(lockfile_parser)], params={"cwd": package_dir.path})
except subprocess.CalledProcessError as e:
    raise PackageManagerError("Failed to parse Gemfile.lock") from e

json_output = json.loads(output)  # unhandled JSONDecodeError
```

Every other backend that parses JSON handles this:
- npm: `_load_json_file()` catches both `JSONDecodeError` and `FileNotFoundError`
- yarn: catches `json.JSONDecodeError` and wraps it in `InvalidLockfileFormat`

**2. stderr is discarded when wrapping CalledProcessError (line 49)**

When the Ruby script fails, `run_cmd()` raises `CalledProcessError` which contains the actual error in `e.stderr`. The current code discards it:

```python
except subprocess.CalledProcessError as e:
    raise PackageManagerError("Failed to parse Gemfile.lock") from e
    # e.stderr is never passed along
```

The user only sees a generic "Failed to parse Gemfile.lock" with no indication of what actually went wrong. Other backends preserve this context:
- cargo: inspects `e.stderr` for specific error messages
- gomod: wraps `CalledProcessError` in `PackageManagerError` with function context

### Steps to reproduce

**For issue 1 (JSONDecodeError):**
1. Add `puts "WARNING: something"` before the JSON output in `lockfile_parser.rb`
2. Run `hermeto fetch-deps` on any Ruby project with a Gemfile.lock
3. Result: raw `json.JSONDecodeError` traceback

**For issue 2 (lost stderr):**
1. Use a Ruby project where `lockfile_parser.rb` fails (e.g., incompatible bundler version, missing gem)
2. Run `hermeto fetch-deps` on it
3. Result: error says "Failed to parse Gemfile.lock" with no details about WHY the Ruby script failed

### Tested with

- jekyll/jekyll (50k stars) with injected parser failures
- Python 3.12.3, Ruby 3.2.3, Bundler 4.0.8
- hermeto 0.47.1.dev50

### Expected behavior

Both cases should produce a clean `PackageManagerError` with the relevant context, consistent with how npm, yarn, cargo, and gomod handle similar failures.

Pr coming right up .

farhann-saleem:
## Description
Resolves #1462

```markdown
## What

Fixes two error handling gaps in the bundler backend's `parse_lockfile()`.


## Why

1. `json.loads(output)` was placed **outside** the try-except block. If the Ruby lockfile parser emits non-JSON output (warnings, version mismatch messages, encoding issues), hermeto crashes with a raw `json.JSONDecodeError` traceback instead of a user-friendly `PackageManagerError`.

2. When `run_cmd()` raises `CalledProcessError`, the `stderr` field (containing the actual Ruby error) was discarded. Users only saw a generic "Failed to parse Gemfile.lock" with no context about what went wrong.

Every other backend handles these cases:
- npm: `_load_json_file()` catches `JSONDecodeError`
- yarn: catches `JSONDecodeError` and wraps in `InvalidLockfileFormat`
- cargo/gomod: preserve `stderr` when wrapping `CalledProcessError`

## Changes

- **`bundler/parser.py`**: Move `json.loads()` inside the try-except block, add `json.JSONDecodeError` handler, pass `stderr=e.stderr` when catching `CalledProcessError`
- **`errors.py`**: Add `friendly_msg()` override to `PackageManagerError` so stored stderr is actually displayed to the user (follows the existing `GitError` pattern)
- **`test_parser.py`**: Update existing subprocess error test to verify stderr preservation, add new test for invalid JSON output

## How to test

```bash
python -m pytest tests/unit/package_managers/bundler/test_parser.py -v
```
```

### DIFF
--- a/hermeto/core/package_managers/bundler/parser.py
+++ b/hermeto/core/package_managers/bundler/parser.py
@@ -45,10 +45,12 @@ def parse_lockfile(
     lockfile_parser = scripts_dir / "lockfile_parser.rb"
     try:
         output = run_cmd(cmd=[str(lockfile_parser)], params={"cwd": package_dir.path})
-    except subprocess.CalledProcessError as e:
-        raise PackageManagerError("Failed to parse Gemfile.lock") from e
-
-    json_output = json.loads(output)
+        json_output = json.loads(output)
+    except (subprocess.CalledProcessError, json.JSONDecodeError) as e:
+        details = e.stderr if isinstance(e, subprocess.CalledProcessError) else output
+        raise PackageManagerError(
+            f"Failed to parse Gemfile.lock{f': {details}' if details else ''}"
+        ) from e
 
     bundler_version: str = json_output["bundler_version"]
     log.info("Package %s is bundled with version %s", package_dir.path.name, bundler_version)
--- a/hermeto/core/package_managers/bundler/parser.py
+++ b/hermeto/core/package_managers/bundler/parser.py
@@ -73,7 +73,10 @@ def _run_lockfile_parser(package_dir: Path) -> dict[str, Any]:
         except subprocess.CalledProcessError as e:
             raise PackageManagerError("Failed to parse Gemfile.lock") from e
 
-    return json.loads(output)
+    try:
+        return json.loads(output)
+    except json.JSONDecodeError as e:
+        raise PackageManagerError("Failed to load JSON output from parsed Gemfile.lock") from e
 
 
 def parse_lockfile(
--- a/tests/unit/package_managers/bundler/test_parser.py
+++ b/tests/unit/package_managers/bundler/test_parser.py
@@ -61,10 +61,21 @@ def test_parse_lockfile_os_error(
 ) -> None:
     mock_run_cmd.side_effect = subprocess.CalledProcessError(returncode=1, cmd="cmd")
 
-    with pytest.raises(PackageManagerError) as exc_info:
+    with pytest.raises(PackageManagerError):
         parse_lockfile(rooted_tmp_path)
 
-    assert "Failed to parse Gemfile.lock" in exc_info.value.friendly_msg()
+
+@mock.patch("hermeto.core.package_managers.bundler.parser._ensure_bundler_files_exist")
+@mock.patch("hermeto.core.package_managers.bundler.parser.run_cmd")
+def test_parse_lockfile_invalid_json(
+    mock_run_cmd: mock.MagicMock,
+    mock_ensure_bundler_files_exist: mock.MagicMock,
+    rooted_tmp_path: RootedPath,
+) -> None:
+    mock_run_cmd.return_value = "not valid json"
+
+    with pytest.raises(PackageManagerError):
+        parse_lockfile(rooted_tmp_path)
 
 
 @mock.patch("hermeto.core.package_managers.bundler.parser._ensure_bundler_files_exist")
--- a/tests/unit/package_managers/bundler/test_parser.py
+++ b/tests/unit/package_managers/bundler/test_parser.py
@@ -88,10 +88,20 @@ def test_run_lockfile_parser_raises_exception_on_os_error(
     tmp_path: Path,
 ) -> None:
     mock_run_cmd.side_effect = subprocess.CalledProcessError(returncode=1, cmd="cmd")
-    with pytest.raises(PackageManagerError) as exc_info:
+
+    with pytest.raises(PackageManagerError):
         _run_lockfile_parser(tmp_path)
 
-    assert "Failed to parse Gemfile.lock" in exc_info.value.friendly_msg()
+
+@mock.patch("hermeto.core.package_managers.bundler.parser.run_cmd")
+def test_run_lockfile_parser_raises_exception_on_invalid_json(
+    mock_run_cmd: mock.MagicMock,
+    tmp_path: Path,
+) -> None:
+    mock_run_cmd.return_value = "not valid json"
+
+    with pytest.raises(PackageManagerError):
+        _run_lockfile_parser(tmp_path)
 
 
 @mock.patch("hermeto.core.package_managers.bundler.parser._ensure_bundler_files_exist")

### CODE REVIEW

File: tests/unit/package_managers/bundler/test_parser.py [hunk 1]
```diff
@@ -59,7 +59,25 @@ def test_parse_lockfile_os_error(
...
-    mock_run_cmd.side_effect = subprocess.CalledProcessError(returncode=1, cmd="cmd")
+    mock_run_cmd.side_effect = subprocess.CalledProcessError(
+        returncode=1, cmd="cmd", stderr="ruby error details"
+    )
+
+    with pytest.raises(PackageManagerError) as exc_info:
+        parse_lockfile(rooted_tmp_path)
+
+    assert "Failed to parse Gemfile.lock" in exc_info.value.friendly_msg()
+    assert exc_info.value.stderr == "ruby error details"
```
eskultety: No string comparisons in tests please, verifying an exception was raised is enough.

hermeto/core/errors.py
```diff
@@ -291,6 +291,13 @@ def __init__(
         self.stderr = stderr
         super().__init__(reason, solution=solution)
 
+    def friendly_msg(self) -> str:
+        """Return the user-friendly representation of this error with stderr if available."""
+        msg = super().friendly_msg()
+        if self.stderr:
+            msg += f"\n\nCommand output (stderr):\n{textwrap.indent(self.stderr.strip(), '  ')}"
+        return msg
+
```
eskultety: Was this one added just to be able to use it in tests?

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -45,10 +45,11 @@ def parse_lockfile(
...
     try:
         output = run_cmd(cmd=[str(lockfile_parser)], params={"cwd": package_dir.path})
+        json_output = json.loads(output)
     except subprocess.CalledProcessError as e:
-        raise PackageManagerError("Failed to parse Gemfile.lock") from e
-
-    json_output = json.loads(output)
+        raise PackageManagerError("Failed to parse Gemfile.lock", stderr=e.stderr) from e
+    except json.JSONDecodeError as e:
+        raise PackageManagerError("Failed to parse Gemfile.lock", stderr=output) from e
```
eskultety: Could these maybe be condensed to a single except clause since the root of the message is the same, it's just the details that differ?

eskultety (changes requested):

tests/unit/package_managers/bundler/test_parser.py
farhann-saleem: done 
removed string assertions and simplified it

hermeto/core/package_managers/bundler/parser.py
farhann-saleem: Done. Condensed to single except clause

hermeto/core/errors.py
farhann-saleem: No its needed to display stderr to the user
PackageManagerError stores  stderr but BaseError.friendly_msg() doesnt know about it,so without this override stderr never appears in the error output. 
GitError uses same pattern(lines 334-338 in errors.py)

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -45,10 +45,10 @@ def parse_lockfile(
...
     try:
         output = run_cmd(cmd=[str(lockfile_parser)], params={"cwd": package_dir.path})
-    except subprocess.CalledProcessError as e:
-        raise PackageManagerError("Failed to parse Gemfile.lock") from e
-
-    json_output = json.loads(output)
+        json_output = json.loads(output)
+    except (subprocess.CalledProcessError, json.JSONDecodeError) as e:
+        stderr = e.stderr if isinstance(e, subprocess.CalledProcessError) else output
+        raise PackageManagerError("Failed to parse Gemfile.lock", stderr=stderr) from e
```
eskultety: Except that a library doesn't produce `stderr` so it's logically wrong to treat it that way IMO. I'd envisioned something simple as `raise Exception(f"my_error_string: {details}")` where `details` is one of the two.

hermeto/core/errors.py
eskultety: Right. The fact that `friendly_msg` ignores `stderr` is a wider issue in our code base as I'm grepping it, out of scope of this PR. It's out of scope because we IMO need something a bit more robust than ^this and so it's also the reason why I prefer a simple f-string concat in this case, see my other comment - not great, not terrible, a ton of precedent all around our code base.
Definitely space for improvement though.

hermeto/core/package_managers/bundler/parser.py
farhann-saleem: Understood
Switched to simple f-string concatenation for the error details instead, 
Removed the errors.py changes - keeping this focused on bundler backend only

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -45,10 +45,15 @@ def parse_lockfile(
     lockfile_parser = scripts_dir / "lockfile_parser.rb"
     try:
         output = run_cmd(cmd=[str(lockfile_parser)], params={"cwd": package_dir.path})
-    except subprocess.CalledProcessError as e:
```
eskultety, line 48: **Commit message:**
- why is the whole message including commit subject indented?
- missing an empty line between commit subject and message body
- message body references changes from a past iteration, not the current state of the changes

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -45,10 +45,15 @@ def parse_lockfile(
...
-    json_output = json.loads(output)
+        json_output = json.loads(output)
+    except (subprocess.CalledProcessError, json.JSONDecodeError) as e:
+        details = e.stderr if isinstance(e, subprocess.CalledProcessError) else output
+        msg = (
+            f"Failed to parse Gemfile.lock: {details}"
+            if details
+            else "Failed to parse Gemfile.lock"
+        )
+        raise PackageManagerError(msg) from e
```
eskultety:
```
 raise PackageManagerError(
      f"Failed to parse Gemfile.lock{f': {details}' if details else ''}"
  ) from e
```

eskultety (changes requested):

hermeto/core/package_managers/bundler/parser.py
farhann-saleem: fixed

eskultety (approved):

eskultety (approved):

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -45,10 +45,12 @@ def parse_lockfile(
...
-    except subprocess.CalledProcessError as e:
-        raise PackageManagerError("Failed to parse Gemfile.lock") from e
-
-    json_output = json.loads(output)
+        json_output = json.loads(output)
+    except (subprocess.CalledProcessError, json.JSONDecodeError) as e:
+        details = e.stderr if isinstance(e, subprocess.CalledProcessError) else output
+        raise PackageManagerError(
+            f"Failed to parse Gemfile.lock{f': {details}' if details else ''}"
+        ) from e
```
slimreaper35, line 53: When I try to artificially test it by modifying our lockfile parser, the error message contains a hardcoded `\n` character:
`2026-04-27 11:19:48,457 ERROR PackageManagerError: Failed to parse Gemfile.lock: not valid\n`

Also, it feels like we are handling an error in the bundler JSON library.

slimreaper35 (commented):
I doubt this PR has any value.

a-ovchinnikov:
This PR has outstanding comments and has not been worked on for a long while. @farhann-saleem do you plan to continue working on it?

farhann-saleem:
yes  @a-ovchinnikov  . im on it . sorry for the delay

hermeto/core/package_managers/bundler/parser.py
```diff
@@ -70,10 +70,17 @@ def _run_lockfile_parser(package_dir: Path) -> dict[str, Any]:
...
-    return json.loads(output)
+            json_output = json.loads(output)
+        except (subprocess.CalledProcessError, json.JSONDecodeError) as e:
+            if isinstance(e, subprocess.CalledProcessError):
+                details = e.stderr.strip() if e.stderr else None
+            else:
+                details = str(e)
+            raise PackageManagerError(
+                f"Failed to parse Gemfile.lock{f': {details}' if details else ''}"
+            ) from e
```
slimreaper35: That looks extremely complicated for such a corner case.

What do you think about separating both except clauses?

```python
    env = {"PATH": os.environ.get("PATH")}
    with _with_hidden_bundle_directory(package_dir):
        try:
            output = run_cmd(cmd=[str(lockfile_parser)], params={"cwd": package_dir, "env": env})
        except subprocess.CalledProcessError as e:
            raise PackageManagerError("Failed to parse Gemfile.lock") from e

    try:
        return json.loads(output)
    except json.JSONDecodeError as e:
        raise PackageManagerError("Failed to load JSON output from parsed Gemfile.lock") from e
```

slimreaper35 (changes requested):

hermeto/core/package_managers/bundler/parser.py
farhann-saleem: Separated both except clauses as suggested

slimreaper35 (approved):

a-ovchinnikov (approved):

PR #1466, opened 2026-03-28
Fix(cargo): fixed vcs metdata in sbom for workspace members

### LINKED ISSUE #1455: bug(cargo) :Cargo workspace members are not recognized as local dependencies in SBOM (missing VCS metadata)
## Description

Hermeto does not recognize Cargo workspace members as local dependencies when generating SBOMs. Workspace crates are emitted as external CargoPackage components without VCS metadata (vcs_url, subpath), even though they are part of the same repository.

## Expected behavior

Workspace members should be treated as local packages and include VCS metadata, similar to the root package:
```sh
pkg:cargo/a@... ?vcs_url=...
```
## Actual behavior

Workspace members are emitted as external packages:

```sh
pkg:cargo/a@...
pkg:cargo/b@...
```
Only the root package includes VCS metadata.

## Impact
- Incorrect SBOM provenance
- Local code misclassified as external dependency
- Loss of repository traceability
- Inconsistent behavior within same workspace
## Root cause
_find_local_packages does not consider [workspace.members], only path-based dependencies and patches.

Harsh-2005d:
Resolves: #1455
## Description
_generate_sbom_components only considered root package, explicit path dependencies, and registry packages when generating SBOM components. Workspace members not declared as path dependencies had no source field in Cargo.lock but were still emitted as external CargoPackage components without VCS metadata, resulting in incorrect SBOM provenance.
### Affected code
```python
# Workspace members fall through to else branch, losing VCS metadata
else:
    components.append(
        CargoPackage(
            name=pkg_name,
            version=pkg_version,
            source=pkg.get("source"),
            checksum=pkg.get("checksum"),
        ).to_component()
    )
```
## Fix
```python
# Before — workspace members misclassified as external
else:
    components.append(CargoPackage(...).to_component())

# After — sourceless packages treated as local workspace members
elif pkg.get("source") is None:
    components.append(
        LocalCargoPackage(name=pkg_name, version=pkg_version, vcs_url=vcs_url, subpath=pkg_name)
        .to_component()
    )
else:
    components.append(CargoPackage(...).to_component())
```

## Tests
Added two unit tests for the changed behaviour.

## Assumptions
- The workspace is defined at the root, with `Cargo.lock` also at the root (already assumed).
- Workspace member directory names match their package names (`subpath=pkg_name`). Members where
  the directory name differs from the package name will produce incorrect subpaths in the SBOM.
- All packages without a source field in Cargo.lock are local i.e. workspace or path deps. it might change in future lock versions.

### DIFF
--- a/hack/mock-unittest-data/gomod.py
+++ b/hack/mock-unittest-data/gomod.py
@@ -11,7 +11,6 @@ from tempfile import TemporaryDirectory
 
 from yarn import print_banner
 
-
 PANDEMO_DIR = "gomod-pandemonium"
 
 
@@ -38,7 +37,7 @@ def write_to_file(where, what, mocked_data_dir, **subprocess_kwargs):
 
 def replace_paths_with_placeholder(pandemo_dir, gomodcache, mocked_data_dir):
     subdirs_of_interest = ("non-vendored", "vendored", "workspaces")
-    fs_objects_to_check = [(mocked_data_dir/x).glob("**/*") for x in subdirs_of_interest]
+    fs_objects_to_check = [(mocked_data_dir / x).glob("**/*") for x in subdirs_of_interest]
     files_to_update = [o for o in chain.from_iterable(fs_objects_to_check) if o.is_file()]
     for fpath in files_to_update:
         data = fpath.read_text()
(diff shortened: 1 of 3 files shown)


Harsh-2005d:
Sorry for the unrelated formatting changes these were automatically applied by ruff when formatting. No logic changes involved.

PR #1468, opened 2026-03-28
pip: fix missing separators in error messages

### LINKED ISSUE #1467: [pip] Missing separator in error message strings
Two `solution` strings in `hermeto/core/package_managers/pip/requirements.py` are missing a separator between sentences due to Python implicit string concatenation.

- Line 638 (`validate_requirements`):

`f"URL requirement must specify exactly one hash, but specifies {n_hashes}"`
`"Please specify the expected hashes for all plain URLs using "`

Produces: `...specifies 0Please specify...`

- Line 673 (`validate_requirements_hashes`):

`f"Hash is required, dependency does not specify any: {req.download_line}"`
`"Please specify the expected hashes for all dependencies"`

Produces: `...{download_line}Please specify...`

**The correct pattern with `.\n` is already used in the same function at line 607:**

`"Please pin all packages as <name>==<version>\n"`
`"You may wish to use a tool such as pip-compile to pin automatically."`

**Fix: Add `.\n` at the end of the first string in both cases to match the existing pattern.**

Contribution
I will be working on a fix for this.

Priyanshu-u07:
Two `solution` strings in `validate_requirements` and `validate_requirements_hashes` were missing a `.\n` separator between sentences. 
Python implicit string concatenation joined them without any separator, producing output like:
`...specifies 0Please specify...`

Added `.\n` to match the existing pattern used in the same file (line 607).

Fixes #1467

### DIFF
--- a/hermeto/core/package_managers/pip/requirements.py
+++ b/hermeto/core/package_managers/pip/requirements.py
@@ -635,7 +635,7 @@ def validate_requirements(requirements: list[PipRequirement]) -> None:
                 raise InvalidChecksum(
                     checksum=req.hashes,
                     solution=(
-                        f"URL requirement must specify exactly one hash, but specifies {n_hashes}"
+                        f"URL requirement must specify exactly one hash, but specifies {n_hashes}.\n"
                         "Please specify the expected hashes for all plain URLs using "
                         "--hash options (one --hash for each)"
                     ),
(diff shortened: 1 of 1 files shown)


Priyanshu-u07:
@eskultety I have updated git authorship and DCO. Please take another look.

eskultety:
> Two `solution` strings in `validate_requirements` and `validate_requirements_hashes` were missing a `.\n` separator between sentences. Python implicit string concatenation joined them without any separator, producing output like: `...specifies 0Please specify...`
> 
> Added `.\n` to match the existing pattern used in the same file (line 607).
> 
> Fixes #1467

I forgot to mention in my review that this info should also be part of the commit you're adding as way of reasoning, thanks.

ACK to the change, I'll approve once ^this is sorted out.

Priyanshu-u07:
@eskultety I have addressed the suggested changes. Thanks!

PR #1470, opened 2026-03-29
Revert 4 feat/x uv source policy

mithlohar144:


eskultety:
This is not how one is supposed to go about a GSoC project, there's no prior design, no discussion, not even bare minimum requirements met for the PR according to our contributing guidelines -> closing.

PR #1471, opened 2026-03-29
bug(cargo): Handle missing [source] section in cargo vendor config for workspace-only projects

Harsh-2005d:
## Description

`_swap_sources_directory_for_subsitution_slot` unconditionally accessed `toml_template["source"]`, assuming `cargo vendor` always produces a `[source]` section in its config output. For projects with only workspace members and no registry dependencies, `cargo vendor` produces no `[source]` section, causing a `KeyError` crash.

## Affected code

```python
def _swap_sources_directory_for_subsitution_slot(template: str) -> dict:
    toml_template = tomlkit.parse(template).value
    # crashes when no registry deps — no [source] section produced by cargo vendor
    toml_template["source"]["vendored-sources"]["directory"] = "${output_dir}/deps/cargo"
    return toml_template
```

## Fix

```python
# Before — unconditional access
toml_template["source"]["vendored-sources"]["directory"] = "${output_dir}/deps/cargo"

# After — guard against missing [source] section
if "source" in toml_template:
    toml_template["source"]["vendored-sources"]["directory"] = "${output_dir}/deps/cargo"
```

### DIFF
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -141,10 +141,11 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
         vendor_result = _fetch_dependencies(package_dir, request)
         # cargo allows to specify configuration per-package
         # https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure
-        config_template = _swap_sources_directory_for_subsitution_slot(
-            vendor_result.config_template
-        )
-        project_files.append(_use_vendored_sources(package_dir, config_template))
+        if vendor_result.config_template:
+            config_template = _swap_sources_directory_for_subsitution_slot(
+                vendor_result.config_template
+            )
+            project_files.append(_use_vendored_sources(package_dir, config_template))
         package_components = _generate_sbom_components(package_dir)
 
         if vendor_result.lockfile_was_generated:


ST3V1K:
Small correction: `cargo vendor` doesn't produce TOML with a missing `[source]` section - its stdout is completely empty. So config_template is an empty string, not an incomplete TOML.

Given that, I think the fix belongs one level up in `fetch_cargo_source`. When there's nothing to vendor, we shouldn't be calling the `_swap_sources_directory_for_subsitution_slot` function at all. It has a clear contract: it receives TOML with a `[source]` section and swaps the directory. Silently returning `{}` masks problems, and the empty dict still gets unnecessarily passed to `_use_vendored_sources`.

Harsh-2005d:
Thanks for the review, i have updated the fix to be in the `fetch_cargo_source` . Now both `_swap_sources_directory_for_subsitution_slot` and `use_vendored_sources` will be skipped in empty config case.

ST3V1K:
LGTM. However, I don't have the permission to approve workflows. This will still have to wait for 2 more people.

eskultety:
This now needs a rebase after recent CI fixes.

eskultety (approved):
The PR description doesn't match the changes. Changes look sound. ACK. We can probably live without a unit test for this instance since no fetch_dependencies test currently exists and testing an empty template would require some unnecessary mocking (we're trying to consolidate our tests).

a-ovchinnikov (approved):

Harsh-2005d:
Rebased the branch.

> The PR description doesn't match the changes.

not sure what you were referring to, but i tried to explain more in the commit. Let me know if it needs further changes.

PR #1472, opened 2026-03-29
Feat/x uv fetch artifacts

mithlohar144:


eskultety:
Same argument as https://github.com/hermetoproject/hermeto/pull/1470#issuecomment-4150314053

PR #1475, opened 2026-03-30
pip: Replace pybuild-deps parsers with Hermeto ones

slimreaper35:


eskultety:
ACK to the reasoning for the time being, however, long term I think we should heavily refine pybuild-deps (desperately needs love) to a point where it's not just a CLI tool, but also a library where individual bits could be consumed from. That way, we as maintainers of both would not need to maintain 2 implementations of project files parsers, it would only live in a single place.

slimreaper35:
Rebased + applied suggestions from Gemini.

> ACK to the reasoning for the time being, however, long term I think we should heavily refine pybuild-deps (desperately needs love) to a point where it's not just a CLI tool, but also a library where individual bits could be consumed from. That way, we as maintainers of both would not need to maintain 2 implementations of project files parsers, it would only live in a single place.

Ok. But pybuild-deps needs to parse only build requirements, so other functionality that we have in Hermeto would be dead code there.

eskultety:
> Ok. But pybuild-deps needs to parse only build requirements, so other functionality that we have in Hermeto would be dead code there.

It was just an idea, so it's fair to say it might get rejected. That said, if it were a library, then it also makes sense to provide a complete parser even if the CLI tooling were to consume the build reqs only, so I would not call it dead code, only for the CLI use case, but once you're building a library, then it's IMO not a fair assessment.

hermeto/core/package_managers/pip/rust.py
```diff
@@ -27,27 +27,19 @@ def _has_rust_build_deps(raw_build_dependencies: list[str]) -> bool:
...
+        "setup.cfg": SetupCFG(root).get_setup_requires,
+        "setup.py": SetupPY(root).get_setup_requires,
     }
-    for file_name, parser in file_parser_map.items():
-        file_path = extracted_dir / file_name
-        if not file_path.exists():
-            continue
 
-        # The file is decoded as utf-8-sig because plain utf-8 has proven to be problematic with certain sources from pypi:
-        # https://github.com/hermetoproject/pybuild-deps/blob/4dc40ffabddb8aad1279978b8741111fb64452e6/src/pybuild_deps/finder.py#L45-L51
```
eskultety, line 41: This commentary is dropped in this PR whilst providing some good insight (even the expanded commentary in pybuild-deps) on why we need this particular encoding, e.g. it was previously applied to all files, after your changes it's only applied to the setup.py parser...and this commentary tell you why.

hermeto/core/package_managers/pip/project_files.py
```diff
@@ -206,6 +215,19 @@ def get_version(self) -> str | None:
...
+        if setup_requires is None:
+            return []
+
+        return [
+            r.strip()
+            for req in setup_requires.strip().splitlines()
+            for r in re.split(";|,", req.strip())
+            if r.strip()
+        ]
+
```
eskultety: nested loop multi-line comprehensions don't look very nice, I think we can condense this somewhat.

eskultety: When I said "condense" I meant to get rid of of the nested loop but still possibly keep the benefit of a list comprehension, e.g. we're splitting on newlines first, then on some characters, I think we can combine that. We also call `strip()` twice, once in the condintion, then when adding the member, so that would translate to:
` return [s for r in re.split(r"[;,\n]", setup_requires.strip()) if (s := r.strip())]`, would that be too much, too ambitious? :)

PR #1480, opened 2026-03-31
feat(yarn): add workspace focus support for Yarn v4

### LINKED ISSUE #1508: feat(yarn): Support workspace focus for Yarn v4 monorepos
## Description

When using hermeto with Yarn v4 monorepos, all dependencies across all workspaces are currently prefetched and included in the output. This causes inaccurate SBOMs for projects that only need dependencies from specific workspaces.

We need the ability to specify which workspaces to focus on during dependency prefetching, so that:
- Only the dependencies for the focused workspaces are prefetched (via `yarn workspaces focus`)
- The output/SBOM accurately reflects only the packages actually installed for the focused set, including transitive workspace dependencies

### Motivation

This is needed to resolve SBOM accuracy issues for Yarn v4 monorepo consumers.

### Proposed solution

Allow users to specify a `workspaces` field in the package input:

```json
{"path": ".", "type": "yarn", "workspaces": ["my-app"]}
```

When `workspaces` is provided, use `yarn workspaces focus` to install only the relevant subset. Since `yarn workspaces focus` does not modify the lockfile, a filtering step is needed to compute the transitive closure of required workspaces and exclude unrelated packages from the output.

When `workspaces` is omitted, existing behavior should remain unchanged.

### Related PRs

- #1480 — feat(yarn): add workspace focus support for Yarn v4
- #1485 — docs: extend Yarn v4 design doc with workspace focus

cc @jonkoops

jonkoops:
Allow users to specify workspaces in the package input so that only the dependencies for those workspaces are prefetched via [`yarn workspaces focus`](https://yarnpkg.com/cli/workspaces/focus). The output reflects only the packages actually installed for the focused workspaces, including their transitive workspace dependencies.

Each specified workspace is queried via `yarn workspace <name> info --recursive --cache --json`, and results are deduplicated across workspaces. When `workspaces` is omitted, existing behavior is unchanged.

Since `yarn workspaces focus` does not support `--mode skip-build`, and [`enableScripts: false` does not apply to workspace scripts](https://github.com/yarnpkg/berry/pull/4781), the `scripts` field is stripped from workspace `package.json` files before running the command to prevent lifecycle script execution.

### Usage

To focus on a single workspace, add the `workspaces` field to the package input:

```json
{"path": ".", "type": "yarn", "workspaces": ["my-app"]}
```

Multiple workspaces can be specified:

```json
{"path": ".", "type": "yarn", "workspaces": ["my-app", "my-lib"]}
```

### Input validation

- `workspaces: []` is rejected (omit the field instead to fetch all dependencies)
- `workspaces` requires Yarn v4 or later

### Before merging

This PR depends on [hermetoproject/integration-tests#69](https://github.com/hermetoproject/integration-tests/pull/69) which adds the integration test fixture.

Closes #1508

### DIFF
--- a/docs/yarn.md
+++ b/docs/yarn.md
@@ -7,6 +7,7 @@
   - [Dealing with Yarn Zero-Installs](#dealing-with-yarn-zero-installs)
   - [Dealing with plugins](#dealing-with-plugins)
 - [Specifying packages to process](#specifying-packages-to-process)
+  - [Workspace focus](#workspace-focus)
   - [Controlling Yarn's behavior](#controlling-yarns-behavior)
   - [Downloading dependencies](#downloading-dependencies)
   - [Known pitfalls](#known-pitfalls)
@@ -114,6 +115,43 @@ or more simply by just invoking `hermeto fetch-deps yarn`.
 
 For complete example of how to pre-fetch dependencies, see
 [Example: Pre-fetch dependencies](#pre-fetch-dependencies).
+
+### Workspace focus
+
+For monorepo projects using [Yarn workspaces][], you can instruct Hermeto to
+fetch only the dependencies needed for specific workspaces rather than the
+entire project. This is done by specifying the `workspaces` field in the JSON
+input:
+
+```js
+{
+  "type": "yarn",
+  "path": ".",
+  // Only fetch dependencies for the "my-app" workspace
+  "workspaces": ["my-app"]
+}
+```
+
+Hermeto will fetch only the dependencies needed for the specified workspaces
+and their transitive workspace dependencies. Note that `devDependencies` are
+also included and will appear in the SBOM.
+
+You can specify multiple workspace names:
+
+```shell
+hermeto fetch-deps \
+  --source ./my-monorepo \
+  --output ./hermeto-output \
+  '{"type": "yarn", "workspaces": ["app-frontend", "lib-shared"]}'
+```
+
+> **NOTE**
+>
+> This feature requires **Yarn v4**. The workspace names must match the `name`
+> field in each workspace's `package.json`.
+
+When `workspaces` is not specified, Hermeto fetches dependencies for all
+workspaces.
 
 ### Controlling Yarn's behavior
 
@@ -269,5 +307,6 @@ podman build . \
 [project]: https://github.com/hermetoproject/doc-examples/tree/yarn-basic
 [yarn install]: https://yarnpkg.com/getting-started/usage/#installing-all-the-dependencies
 [Yarn protocols]: https://yarnpkg.com/protocols
+[Yarn workspaces]: https://yarnpkg.com/features/workspaces
 [yarn]: https://yarnpkg.com
 [ZIP archives]: https://yarnpkg.com/features/pnp/#packages-are-stored-inside-zip-archives-how-can-i-access-their-files
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -396,6 +396,14 @@ class YarnPackageInput(_PackageInputBase):
     """Accepted input for a yarn package."""
 
     type: Literal["yarn"]
+    workspaces: list[str] | None = None
+
+    @pydantic.field_validator("workspaces")
+    @classmethod
+    def _workspaces_not_empty(cls, workspaces: list[str] | None) -> list[str] | None:
+        if workspaces is not None and len(workspaces) == 0:
+            raise ValueError("'workspaces' must not be an empty list, omit the field instead")
+        return workspaces
 
 
 PackageInput = Annotated[
--- a/hermeto/core/package_managers/yarn/main.py
+++ b/hermeto/core/package_managers/yarn/main.py
@@ -9,14 +9,20 @@ from hermeto.core.errors import LockfileNotFound, PackageManagerError, PackageRe
 from hermeto.core.models.input import Request
 from hermeto.core.models.output import Component, EnvironmentVariable, RequestOutput
 from hermeto.core.models.sbom import create_backend_annotation
+from hermeto.core.package_managers.yarn.locators import WorkspaceLocator
 from hermeto.core.package_managers.yarn.project import (
+    PackageJson,
     Plugin,
     Project,
     YarnRc,
     get_semver_from_package_manager,
     get_semver_from_yarn_path,
 )
-from hermeto.core.package_managers.yarn.resolver import create_components, resolve_packages
+from hermeto.core.package_managers.yarn.resolver import (
+    Package,
+    create_components,
+    resolve_packages,
+)
 from hermeto.core.package_managers.yarn.utils import (
     VersionsRange,
     extract_yarn_version_from_env,
@@ -35,7 +41,7 @@ def fetch_yarn_source(request: Request) -> RequestOutput:
         path = request.source_dir.join_within_root(package.path)
         project = Project.from_source_dir(path)
 
-        components.extend(_resolve_yarn_project(project, request.output_dir))
+        components.extend(_resolve_yarn_project(project, request.output_dir, package.workspaces))
 
     annotations = []
     if backend_annotation := create_backend_annotation(components, "yarn"):
@@ -101,21 +107,38 @@ def _verify_repository(project: Project) -> None:
     _check_lockfile(project)
 
 
-def _resolve_yarn_project(project: Project, output_dir: RootedPath) -> list[Component]:
+def _resolve_yarn_project(
+    project: Project,
+    output_dir: RootedPath,
+    workspaces: list[str] | None = None,
+) -> list[Component]:
     """Process a request for a single yarn source directory.
 
     :param project: the directory to be processed.
     :param output_dir: the directory where the prefetched dependencies will be placed.
+    :param workspaces: optional list of workspace names to focus on (Yarn v4 only).
     :raises PackageManagerError: if fetching dependencies fails
     """
     log.info(f"Fetching the yarn dependencies at the subpath {project.source_dir}")
 
     version = _configure_yarn_version(project)
+
+    if workspaces and version < semver.Version.parse("4.0.0"):
+        raise PackageRejected(
+            f"Workspace focus requires Yarn v4 or later, but this project uses Yarn {version}",
+            solution="Either upgrade to Yarn v4 or remove the 'workspaces' field from the input.",
+        )
+
     _verify_repository(project)
 
     _set_yarnrc_configuration(project, output_dir, version)
-    packages = resolve_packages(project.source_dir)
-    _fetch_dependencies(project.source_dir)
+
+    packages = resolve_packages(project.source_dir, workspaces)
+
+    if workspaces:
+        _strip_workspace_scripts(project.source_dir, packages)
+
+    _fetch_dependencies(project.source_dir, workspaces)
 
     return create_components(packages, project, output_dir)
 
@@ -241,14 +264,44 @@ def _set_yarnrc_configuration(
     yarn_rc.write()
 
 
-def _fetch_dependencies(source_dir: RootedPath) -> None:
-    """Fetch dependencies using 'yarn install'.
+def _strip_workspace_scripts(source_dir: RootedPath, packages: list[Package]) -> None:
+    """Remove scripts from workspace package.json files.
+
+    yarn workspaces focus does not support --mode skip-build, and enableScripts: false
+    does not apply to workspace scripts (https://github.com/yarnpkg/berry/pull/4781).
+    Stripping the scripts field prevents lifecycle scripts from executing during focus.
+
+    :param source_dir: the project source directory.
+    :param packages: packages returned by ``resolve_packages``, used to find workspace paths.
+    """
+    for pkg in packages:
+        locator = pkg.parsed_locator
+        if not isinstance(locator, WorkspaceLocator):
+            continue
+        pkg_json_path = source_dir.join_within_root(locator.relpath, "package.json")
+        if not pkg_json_path.path.exists():
+            continue
+        pkg_json = PackageJson.from_file(pkg_json_path)
+        if "scripts" in pkg_json:
+            del pkg_json["scripts"]
+            pkg_json.write()
+
+
+def _fetch_dependencies(source_dir: RootedPath, workspaces: list[str] | None = None) -> None:
+    """Fetch dependencies using 'yarn install' or 'yarn workspaces focus'.
+
+    When workspaces are specified, only the dependencies of those workspaces (and their
+    transitive workspace dependencies) are installed via 'yarn workspaces focus'.
 
     :param source_dir: the directory in which the yarn command will be called.
-    :raises PackageManagerError: if the 'yarn install' command fails.
+    :param workspaces: optional list of workspace names to focus on (Yarn v4 only).
+    :raises PackageManagerError: if the yarn command fails.
     """
     try:
-        run_yarn_cmd(["install", "--mode", "skip-build"], source_dir)
+        if workspaces:
+            run_yarn_cmd(["workspaces", "focus", *workspaces], source_dir)
+        else:
+            run_yarn_cmd(["install", "--mode", "skip-build"], source_dir)
     except PackageManagerError as e:
         # TODO: this follows a precedent set in resolver. Either a more robust way for
         # dealing with this must be found or a comment provided that such methods do not exist.
--- a/hermeto/core/package_managers/yarn/resolver.py
+++ b/hermeto/core/package_managers/yarn/resolver.py
@@ -141,36 +141,25 @@ class _YarnInfoEntry(pydantic.BaseModel):
     children: _YarnInfoChildren
 
 
-def resolve_packages(source_dir: RootedPath) -> list[Package]:
+def resolve_packages(source_dir: RootedPath, workspaces: list[str] | None = None) -> list[Package]:
     """Fetch and parse package data from the 'yarn info' output.
+
+    When *workspaces* is provided, runs ``yarn workspace $name info`` for each
+    workspace and deduplicates the results. Otherwise, runs ``yarn info --all``
+    to report the full dependency graph.
 
     This function also performs validation to ensure that the current yarn project can be
     processed.
 
+    :param source_dir: the directory containing the yarn project.
+    :param workspaces: optional list of workspace names to scope the query to.
     :raises UnsupportedFeature: if an unsupported locator type is found in 'yarn info' output
     :raises PackageManagerError: if the 'yarn info' command fails.
     """
-    try:
-        # --all: report dependencies of all workspaces, not just the active workspace
-        # --recursive: report transitive dependencies, not just direct ones
-        # --cache: include info about the cache entry for each dependency
-        result = run_yarn_cmd(["info", "--all", "--recursive", "--cache", "--json"], source_dir)
-    except PackageManagerError as e:
-        if e.stderr and "isn't supported by any available resolver" in e.stderr:
-            raise UnsupportedFeature(
-                "Found an unsupported dependency, more details in the logs.",
-                solution=dedent(
-                    f"""
-                    Please note that {APP_NAME} disables all Yarn plugins, which might be needed for
-                    the correct processing of a dependency. This is done to avoid arbitrary code
-                    execution, which would affect the accuracy of the SBOM.
-                    """
-                ),
-            )
-        raise
-
-    # the result is not a valid json list, but a sequence of json objects separated by line breaks
-    packages = [Package.from_info_string(info) for info in result.splitlines()]
+    if workspaces:
+        packages = _resolve_workspace_packages(source_dir, workspaces)
+    else:
+        packages = _resolve_all_packages(source_dir)
 
     n_unsupported = 0
     for package in packages:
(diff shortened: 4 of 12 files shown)


hermeto/core/package_managers/yarn/main.py
```diff
@@ -241,14 +265,116 @@ def _set_yarnrc_configuration(
...
 
     :param source_dir: the directory in which the yarn command will be called.
-    :raises PackageManagerError: if the 'yarn install' command fails.
+    :param workspaces: optional list of workspace names to focus on (Yarn v4 only).
+    :raises PackageManagerError: if the yarn command fails.
     """
     try:
-        run_yarn_cmd(["install", "--mode", "skip-build"], source_dir)
+        if workspaces:
+            run_yarn_cmd(["workspaces", "focus", *workspaces], source_dir)
```
nickboldt, line 302: don't we have to pass in both a source dir and the specific plugin folder under that source dir? 

eg., `workspaces/orchestrator` and `workspaces/orchestrator/plugins/orchestrator-backend/` ?

jonkoops: No, that is what the `workspaces` are for. The code is able to resolve that path, also multiple workspaces could be active at the same time.

eskultety:
Thank you for the contribution. Please refer to our contributing guidelines on how to submit good candidate PRs. From a quick look `yarn focus` on workspaces is a path we have not evaluated in https://github.com/hermetoproject/hermeto/blob/main/docs/adr/0002-yarn-v4.md. That would be the start for us I'd say. Is the feature proposed here limited to only yarn v4 or any modern Yarn version, i.e. ` 1< yarn_version <=4 `?

hermeto/core/package_managers/yarn/main.py
chmeliik, line 302: Drive-by security observation: the missing `--mode skip-build` is concerning, does `workspaces focus` not support it? I think that means the command will execute lifecycle scripts (`preinstall`, `postinstall` etc.) which Hermeto has to avoid.

If no better option is available, a workaround could be to delete the `scripts` attribute from all local package.jsons before proceeding. Would also have to confirm that `enableScripts: false` in .yarnrc still blocks the scripts of non-workspace dependencies as expected

jonkoops: Good catch, yes, these scripts will execute when using focus, it seems that this is an intentional design decision (see https://github.com/yarnpkg/berry/pull/4781), as [per the documentation](https://yarnpkg.com/configuration/yarnrc#enableScripts):

> If `false` (the default), Yarn will not execute the `postinstall` scripts from third-party packages when installing the project (workspaces will still see their `postinstall` scripts evaluated, as they're assumed to be safe if you're running an install within them).

So whilst this would execute the workspaces' scripts, it would never execute any of the scripts of their dependencies. Would that be sufficient, or is it an explicit goal to avoid executing any of these scripts?

I did see there has been some effort to introduce `--mode=skip-build` for focus (see https://github.com/yarnpkg/berry/issues/3524, https://github.com/yarnpkg/berry/pull/4187), but this work seems to have been stalled, and left without further review. If this were to land, I think it would eliminate the problem entirely.

jonkoops:
@eskultety This is an extension of existing Yarn v4 support using a first-party Yarn command, not a new package manager or major architectural change. If you want I definitely can write up an ADR for you, but it would be rather narrow.

> Is the feature proposed here limited to only yarn v4 or any modern Yarn version, i.e. `1< yarn_version <=4`?

This feature is also supported in Yarn 3, but requires the `@yarnpkg/plugin-workspace-tools` plugin to be installed. It should 'just work', but I chose not to advertise it here, simply because there is no need from our team, as we are on Yarn 4.

hermeto/core/package_managers/yarn/main.py
eskultety: > So whilst this would execute the workspaces' scripts, it would never execute any of the scripts of their dependencies. Would that be sufficient, or is it an explicit goal to avoid executing any of these scripts?

That's a good question, our arbitrary code execution posture comes primarily from the outside rather than the inside. That said, we explicitly don't even allow a user-supplied yarn binary from within the input repository itself , i.e. `.yarn/releases/yarn-<ver>.cjs` which I understand is standard practice in this ecosystem to have it checked in the repo. Executing the workspace scripts falls IMO in the same category. So that's again a potential supply chain attack vector.

eskultety:
> This is an extension of existing Yarn v4 support using a first-party Yarn command, not a new package manager or major architectural change.

I know, but our designs are to track all the research we did so we have some deeper context apart from git history when we question the implementation in the future. Oh, and I linked an ADR, do not get confused by the notion of ADRs, we don't do ADRs, we do just designs, this here is just a leftover mess we need to consolidate under the `docs/designs` directory. In fact we might even encourage you to extend the existing doc with the relevant content.

Edit: And specifically for Yarn v3, we can file an issue to track this in the backlog so that we don't forget, but I wanted to make sure where we actually stand in terms of support of this functionality on Yarn's side.

hermeto/core/package_managers/yarn/main.py
chmeliik: Yeah, it also comes from the desire to really know about everything that went into the build. Everything that Hermeto fetches should be accounted for in the SBOM, and if we execute a user-defined script along the way, that significantly weakens the guarantees.

taylormadore: Hermeto should not be allowing any arbitrary code execution during the prefetch, regardless of whether it comes from the workspace source or dependencies. Non-trusted (from a Hermeto perspective) code could compromise the integrity of the reported SBOM.

jonkoops:
> In fact we might even encourage you to extend the existing doc with the relevant content.

Sure, no problem, I would actually prefer amending the existing documentation, so I will get right on that :+1:

jonkoops:
Created a PR to document the design for this functionality under #1485

hermeto/core/package_managers/yarn/main.py
jonkoops: Considering that this functionality doesn't exist in Yarn at the moment, I agree that stripping the scripts is probably the best way to go. I'll go ahead and implement that.

jonkoops:
@chmeliik @eskultety @taylormadore  I went ahead and implemented the `script` stripping we discussed, it is covered by both unit and integration tests. Could I ask you for another round of reviews?

hermeto/core/package_managers/yarn/main.py
jonkoops: Implemented the `script` stripping (https://github.com/hermetoproject/hermeto/pull/1480#issuecomment-4198581701)

chmeliik:
Added a comment on the design, I think there _might_ be a way to use yarn itself for the filtering

nickboldt (approved):
I approve but I'm not a hermeto committer so we really need buy-in from that team.

jonkoops:
I found a correctness gap in `_resolve_focused_workspaces()`, when a workspace declares `peerDependencies`, Yarn reports it as a [virtual locator](https://github.com/yarnpkg/berry/blob/c0274d6d7ba5939f447e78aaf16e456a00cf0bd1/packages/yarnpkg-core/sources/structUtils.ts#L186-L188) (e.g. `tooling@virtual:<hash>#workspace:packages/tooling`), which the transitive closure lookup failed to match. This caused such workspaces and their transitive deps to be silently excluded from the focused set.

Fixed by [devirtualizing](https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-core/sources/structUtils.ts#L206-L211) dependency locators before lookup. Covered by new unit tests, and the integration test project now includes a `tooling` workspace with `peerDependencies` to check this path.

chmeliik:
> I found a correctness gap in `_resolve_focused_workspaces()`, when a workspace declares `peerDependencies`, Yarn reports it as a [virtual locator](https://github.com/yarnpkg/berry/blob/c0274d6d7ba5939f447e78aaf16e456a00cf0bd1/packages/yarnpkg-core/sources/structUtils.ts#L186-L188) (e.g. `tooling@virtual:<hash>#workspace:packages/tooling`), which the transitive closure lookup failed to match. This caused such workspaces and their transitive deps to be silently excluded from the focused set.

It would be worth looking into https://github.com/hermetoproject/hermeto/pull/1485#discussion_r3045132657, it may be possible to have yarn do the filtering and avoid problems like this

jonkoops:
@chmeliik I went ahead and explored that option, and it does yield a significant cleanup without diverging code paths too much. It also allows us to get rid of a lot of Python code, and rely more on the output of Yarn.

I went ahead and pushed a commit to the branch to see what this implementation would look like, let me know if this is agreeable to you, and then I will amend the proposal and squash the commits.

hermeto/core/package_managers/yarn/resolver.py
chmeliik, line 187: This bit may also be relevant for the `_resolve_workspace_packages` code path

chmeliik (commented):
Apart from a minor inconsistency, this approach LGTM 👍 

(Note that I'm not a Hermeto maintainer anymore though, so I won't be able to give an authoritative ACK)

hermeto/core/package_managers/yarn/resolver.py
jonkoops: Good point, there is actually more logic that could be shared between the two. I went ahead and chunked these functions up a bit more, and improved the test coverage as well.

jonkoops:
@taylormadore @eskultety could I also get your review on this new implementation? If you find it agreeable, I will update the design document under #1480 to reflect the new implementation, and squash the commits in this PR so we can get things in a mergable state.

hermeto/core/models/input.py
```diff
@@ -407,6 +407,7 @@ class YarnPackageInput(_PackageInputBase):
     """Accepted input for a yarn package."""
 
     type: Literal["yarn"]
+    workspaces: list[str] | None = None
```
eskultety, line 399: Won't this allow empty `workspaces: []` to be passed from the user which is nonsensical input? Shouldn't we add a validator for that and default to `None` if that happens?

hermeto/core/package_managers/yarn/resolver.py
```diff
@@ -167,22 +195,35 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
...
+    for ws_name in workspaces:
+        # Run info scoped to a single workspace (without --all), so yarn only reports
+        # the transitive dependencies of that workspace rather than the full project.
+        packages = _get_packages_from_yarn_info(
+            ["workspace", ws_name, "info", "--recursive", "--cache", "--json"], source_dir
         )
+        # Deduplicate across workspaces: overlapping deps appear only once
+        for pkg in packages:
+            if pkg.raw_locator not in seen:
+                seen[pkg.raw_locator] = pkg
```
eskultety: This functionality is built in (and expected as such by, presumably you, the consumers) in Yarn v4, but not in Yarn v3 which requires some further modifications, that's fine, but we'll need a guard for this functionality to refuse proceeding if users specify `workspaces: ` in the input JSON, but we find out we're dealing with Yarn v3.

eskultety (commented):
Commit 2 is the real deal. I played with this a bit using a completely artificial setup and even explored parallel https://yarnpkg.com/cli/workspaces/foreach since that wasn't mentioned anywhere in the linked design doc PR. Turns out, running `yarn workspace info --recursice` sequentially is still much more performant and more straightforward than `foreach` due to increased JSON dedup that needs to happen.
Draft LGTM.

hermeto/core/models/input.py
jonkoops: Agreed, this should be an invalid input. It doesn't make any sense to provide an empty list of workspace, so I have made this a validation error.

hermeto/core/package_managers/yarn/resolver.py
jonkoops: Good catch, I added a validation so that only workspaces on Yarn v4 are accepted.

jonkoops:
@eskultety I have processed your review, can I get another pass? I have also squashed the commits and updated the PR description to reflect the implementation based on @chmeliik's suggestion. @chmeliik @taylormadore if possible, could I also ask you for another review?

I think this is in a good place now, I will update the design proposal under #1485 to reflect the latest changes. Before merging this in, we should merge the integration tests (https://github.com/hermetoproject/integration-tests/pull/69) so that I can point this PR to the right commit upstream, rather than at my fork.

jonkoops:
Went ahead and added a full end-to-end test that verifies the focused workspace can be installed and run in a network-isolated container using only the prefetched deps, as requested by @eskultety (https://github.com/hermetoproject/integration-tests/pull/69#discussion_r3094201366).

jonkoops:
The design proposal under #1485 has been updated with the latest as well.

jonkoops:
Now that the integration tests and design proposal are merged, the tests now point to the upstream branch, rather than my fork. I think this is now in a good enough shape for a proper review, marking as ready.

docs/yarn.md
```diff
@@ -115,6 +116,44 @@ or more simply by just invoking `hermeto fetch-deps yarn`.
 For complete example of how to pre-fetch dependencies, see
 [Example: Pre-fetch dependencies](#pre-fetch-dependencies).
 
+### Workspace focus
+
+For monorepo projects using [Yarn workspaces][], you can instruct Hermeto to
+fetch only the dependencies needed for specific workspaces rather than the
+entire project. This is done by specifying the `workspaces` field in the JSON
```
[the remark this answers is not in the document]
jonkoops: Even though the style guide outlines this, this whole document already uses indicative mood. I am not going to fix this in this PR and introduce a bunch of changes, or a mixed style.

docs/yarn.md
```diff
@@ -115,6 +116,44 @@ or more simply by just invoking `hermeto fetch-deps yarn`.
...
+  "workspaces": ["my-app"]
+}
+```
+
+Under the hood, Hermeto uses [yarn workspaces focus][]
+instead of `yarn install` to install only the specified workspaces and their
+transitive workspace dependencies. The resulting output will contain only the
+dependencies that were actually prefetched.
+
+You can specify multiple workspace names:
```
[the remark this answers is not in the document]
jonkoops: https://github.com/hermetoproject/hermeto/pull/1480#discussion_r3123986247

docs/yarn.md
```diff
@@ -115,6 +116,44 @@ or more simply by just invoking `hermeto fetch-deps yarn`.
...
+  '{"type": "yarn", "workspaces": ["app-frontend", "lib-shared"]}'
+```
+
+> **NOTE**
+>
+> This feature requires **Yarn v4**. The workspace names must match the `name`
+> field in each workspace's `package.json`.
+
+When `workspaces` is not specified, Hermeto fetches dependencies for all
+workspaces as before.
```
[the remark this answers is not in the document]
jonkoops: https://github.com/hermetoproject/hermeto/pull/1480#discussion_r3123986247

hermeto/core/package_managers/yarn/resolver.py
```diff
@@ -170,22 +198,35 @@ def resolve_packages(source_dir: RootedPath) -> list[Package]:
...
 
-    return packages
+    return list(seen.values())
```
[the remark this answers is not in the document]
chmeliik: If this actually works, it would be a solid improvement IMO

chmeliik (approved):
One of Gemini's suggestions would be worth trying, otherwise LGTM. The doc suggestions would reduce clarity IMO.

hermeto/core/package_managers/yarn/resolver.py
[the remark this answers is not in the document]
jonkoops: This is wrong, `yarn info` and `yarn workspace <name> info` are different, and you are confusing them. Since `yarn workspace <name> info` already targets a specific workspace, this doesn't actually work the way you expect.

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -1269,3 +1295,98 @@ def test_path_patch_raises_without_repo_in_permissive_mode(
...
+        "shared@workspace:packages/shared",
+        "left-pad@npm:1.3.0",
+    ]
+    assert mock_yarn_cmd.call_count == 2
+    mock_yarn_cmd.assert_any_call(
+        ["workspace", "app", "info", "--recursive", "--cache", "--json"], rooted_tmp_path
+    )
+    mock_yarn_cmd.assert_any_call(
+        ["workspace", "lib", "info", "--recursive", "--cache", "--json"], rooted_tmp_path
+    )
```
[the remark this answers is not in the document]
jonkoops: https://github.com/hermetoproject/hermeto/pull/1480#discussion_r3124048304

jonkoops:
@chmeliik I investigated the suggestion from Gemini. However, running `yarn info app lib` on the integration test doesn't work the way Gemini suggested. The positional arguments to `yarn info` are package name patterns (for filtering output), not workspace selectors.

jonkoops:
I went ahead and fixed the linting/formatting and other issues that came from the last rebase on `main`, should be good now.

tests/unit/package_managers/yarn/test_main.py
```diff
@@ -492,3 +496,69 @@ def test_fetch_yarn_source(
...
+
+@mock.patch("hermeto.core.package_managers.yarn.main._configure_yarn_version")
+def test_workspace_focus_rejected_for_yarn_v3(
+    mock_configure_version: mock.Mock, rooted_tmp_path: RootedPath
+) -> None:
+    """Workspace focus is rejected when the project uses Yarn v3."""
+    mock_configure_version.return_value = semver.Version.parse("3.6.1")
+    project = mock.Mock(source_dir=rooted_tmp_path)
+
+    with pytest.raises(PackageRejected, match="Workspace focus requires Yarn v4 or later"):
```
eskultety: No string matching in unit tests please :) (multiple occurrences).

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -220,6 +225,34 @@ def test_resolve_packages(mock_run_yarn_cmd: mock.Mock, rooted_tmp_path: RootedP
...
+    mock_run_yarn_cmd: mock.Mock, rooted_tmp_path: RootedPath
+) -> None:
+    """PackageManagerErrors unrelated to resolver support are re-raised as-is."""
+    error = PackageManagerError("something else went wrong", stderr="network timeout")
+    mock_run_yarn_cmd.side_effect = error
+
+    with pytest.raises(PackageManagerError) as exc_info:
+        resolve_packages(rooted_tmp_path)
+
+    assert exc_info.value is error
```
eskultety: What does this test realistically test? It's just another `PackageManagerError` test without string matching (which we don't want anyway), so I'd lean towards dropping this entirely.

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -1165,7 +1198,6 @@ def test_create_components_permissive_mode_without_vcs_url(
     components = create_components([mocked_package.package], mock_project(project_dir), output_dir)
 
     assert len(components) == 1
-    # vcs_url should not be present when repo_id is None
```
eskultety, line 1168: What's with this line being dropped? How is the change related to the PR's scope? I thing I'm missing something.

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -1269,3 +1300,98 @@ def test_path_patch_raises_without_repo_in_permissive_mode(
...
+    mock_yarn_cmd.assert_any_call(
+        ["workspace", "app", "info", "--recursive", "--cache", "--json"], rooted_tmp_path
+    )
+    mock_yarn_cmd.assert_any_call(
+        ["workspace", "lib", "info", "--recursive", "--cache", "--json"], rooted_tmp_path
+    )
+
+
+@mock.patch("hermeto.core.package_managers.yarn.resolver.run_yarn_cmd")
+def test_resolve_packages_single_workspace(
```
eskultety: What value does this test variant bring over the dedup one above? It is a strict subset of the other test IMO.

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -1269,3 +1300,98 @@ def test_path_patch_raises_without_repo_in_permissive_mode(
...
+    mock_yarn_cmd: mock.Mock, rooted_tmp_path: RootedPath
+) -> None:
+    """Unsupported resolver errors are raised as UnsupportedFeature for workspace packages."""
+    mock_yarn_cmd.side_effect = PackageManagerError(
+        "yarn info failed",
+        stderr="foo@exec:./gen.js: isn't supported by any available resolver",
+    )
+
+    with pytest.raises(UnsupportedFeature, match="Found an unsupported dependency"):
+        resolve_packages(rooted_tmp_path, workspaces=["app"])
```
eskultety: Isn't this already caught in `test_resolve_packages_unsupported_resolver` via a different sub-entry point? Since you're entering with `resolve_packages` it doesn't really matter which use case triggered it, does it?

docs/yarn.md
```diff
@@ -115,6 +116,44 @@ or more simply by just invoking `hermeto fetch-deps yarn`.
...
+  "path": ".",
+  // Only fetch dependencies for the "my-app" workspace
+  "workspaces": ["my-app"]
+}
+```
+
+Under the hood, Hermeto uses [yarn workspaces focus][]
+instead of `yarn install` to install only the specified workspaces and their
+transitive workspace dependencies. The resulting output will contain only the
+dependencies that were actually prefetched.
```
eskultety: This is an implementation detail that should be transparent to our users. Instead, we might probably mention that this fetches also dev deps which typically are not needed for production builds, but will appear in the SBOM.

docs/yarn.md
eskultety: `s/as before//`

eskultety (commented):
A few small comments, but this is pretty much ready to go.

tests/unit/package_managers/yarn/test_main.py
jonkoops: Removed. However, this isn't new, there are various places in the test suite that match error messages.

tests/unit/package_managers/yarn/test_resolver.py
jonkoops: This is from a bad rebase, put the original comments back.

tests/unit/package_managers/yarn/test_resolver.py
jonkoops: I added this to have higher code coverage. It verifies that an error is handled properly outside the happy path. This code existed before this PR without full coverage, removed as requested.

tests/unit/package_managers/yarn/test_resolver.py
jonkoops: True, this doesn't really add any value. Dropped.

tests/unit/package_managers/yarn/test_resolver.py
jonkoops: Good point, removed.

docs/yarn.md
jonkoops: Done.

docs/yarn.md
jonkoops: Removed references to the Yarn command, and added the note about the devdeps.

tests/unit/package_managers/yarn/test_main.py
eskultety: Yes and we are actively getting rid of those :) but sometimes it requires some extra work to be able to do so, e.g. we recently introduced a set of explicit exit codes that allowed us to ditch most of those matching in integration tests :) .

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -220,6 +225,21 @@ def test_resolve_packages(mock_run_yarn_cmd: mock.Mock, rooted_tmp_path: RootedP
...
+    mock_run_yarn_cmd.side_effect = PackageManagerError(
+        "yarn info failed",
+        stderr="foo@exec:./gen.js: isn't supported by any available resolver",
+    )
+
+    with pytest.raises(UnsupportedFeature):
+        resolve_packages(rooted_tmp_path)
+
+
+
```
eskultety: nitpick: Extra blank line :)

tests/unit/package_managers/yarn/test_resolver.py
```diff
@@ -1269,3 +1289,62 @@ def test_path_patch_raises_without_repo_in_permissive_mode(
...
+    assert mock_yarn_cmd.call_count == 2
+    mock_yarn_cmd.assert_any_call(
+        ["workspace", "app", "info", "--recursive", "--cache", "--json"], rooted_tmp_path
+    )
+    mock_yarn_cmd.assert_any_call(
+        ["workspace", "lib", "info", "--recursive", "--cache", "--json"], rooted_tmp_path
+    )
+
+
+
```
eskultety: 3 extra trailing blank lines at the end of the file.

eskultety (approved):
@jonkoops 2 more nitpicks (just for consistent formatting reasons) and we merge, great work! Thank you for the time and dedication you put into this and congrats on your first contribution to Hermeto!

Edit: Oh, so the linter complained anyway and made this a hard failure, that means we have to fix it.

tests/unit/package_managers/yarn/test_resolver.py
jonkoops: Fixed.

tests/unit/package_managers/yarn/test_resolver.py
jonkoops: Removed.

PR #1495, opened 2026-04-02
tests: dump effective hermeto test env variables

ST3V1K:
With an increasing number of environment variables, debugging tests became harder. Sometimes knowing the exact test environment can help. Dumping relevant environment variables solves that.

Closes https://github.com/hermetoproject/hermeto/issues/1450

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -284,3 +284,14 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
             item.add_marker(
                 pytest.mark.skip(reason="Test incompatible with local Nexus proxy mode")
             )
+
+
+def pytest_report_header(config: pytest.Config) -> list[str]:
+    """Report effective Hermeto test configuration at the top of the test session."""
+    lines = ["Effective Hermeto test environment:"]
+    for env_var, cli_opt in _ENV_VAR_CLI_MAP:
+        value = config.getoption(cli_opt)
+        if isinstance(value, bool):
+            value = "1" if value else "0"
+        lines.append(f"  {env_var}={value}")
+    return lines


tests/conftest.py
```diff
@@ -70,3 +83,14 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default=os.getenv("HERMETO_TEST_LOCAL_NEXUS_NO_CLEANUP") == "1",
         help="Keep Nexus container running after tests (env: HERMETO_TEST_LOCAL_NEXUS_NO_CLEANUP=1)",
     )
+
+
+def pytest_report_header(config: pytest.Config) -> list[str]:
```
taylormadore: I believe that all of these environment variables are only relevant to integration tests. Do we want to be reporting them for unit tests as well?

ST3V1K: Yes, reporting them only in integration tests makes more sense. 
Moved it to integration/conftest.py

taylormadore (approved):

tests/integration/conftest.py
```diff
@@ -284,3 +284,14 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
             item.add_marker(
                 pytest.mark.skip(reason="Test incompatible with local Nexus proxy mode")
             )
+
```
eskultety, line 287: **Commit message:**

I think a reference to both https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.hookspec.pytest_report_header and https://doc.pytest.org/en/latest/example/simple.html#adding-info-to-test-report-header is worth adding. Especially the latter since it proves that what you're adding is backed by official docs examples and not just some undocumented experiment.

ST3V1K: Added both links to the commit message

Issue #1494, opened 2026-04-02
[Bug]: Pip Wheel Filter incorrectly rejects multi-version Python wheel tags for valid target environments


Anujkumar9081:
**### Description:**

In Hermeto's pip package manager, the logic used to filter wheels by Python version fails when handling wheels that support multiple Python releases compressed into a single tag. This violates [PEP 425](https://peps.python.org/pep-0425/#python-tag), which explicitly permits dot-separated tags (e.g., _py38.py39.py310_) and states a wheel should match if any of the tags match the target environment.

Presently,  __parse_py_version_   in   _hermeto/core/package_managers/pip/package_distributions.py_   extracts all target versions from the interpreter string but reduces them down to only return the maximum version
```
def _parse_py_version(interpreter: str) -> int:
    ...
    # Extracts all versions but returns only the maximum!
    return max(versions)

```
Consequently, for a wheel tag like _py38.py39.py310_, __parse_py_version_ returns 310. Later, in _WheelsFilter._compatible_tag_interpreter_, this single maximum version is compared against the requested Python version (_self.py_version_).
If a user configures Hermeto to resolve dependencies for Python 3.8 (_self.py_version = 38_), _compatible_py_version_ evaluates if _310 == 38_ (False) or _310 < 38_ (False). As a result, the wheel is incorrectly rejected, even though py38 is explicitly defined in the wheel's tag

**Impact**
This is a high-impact functional bug. It forces Hermeto to unnecessarily fall back to source distributions—causing significantly slower builds, compilation overhead, and potential build failures—or fails package resolution outright if _only-binary_ mode is enforced.

**Steps To Reproduce**

1. Configure _PipBinaryFilters_ with a specific _py_version_ (e.g., _38_).
2. Provide a wheel that has a multi-version tag such as _py38.py39.py310-none-any.whl._
3. Process it utilizing _WheelsFilter._
4. Notice that __matches_filter_constraints(tag)_ evaluates to _False_ and excludes the compatible wheel.

**Expected behavior**
Hermeto should conform to PEP 425 by parsing all tags in a multi-version string and accepting the wheel if any of the explicit Python versions satisfy the environment's Python version constraints.

**Proposed Solution**
1. Refactor __parse_py_version_ to __parse_py_versions_, returning a _list[int]_ of all defined versions.
2. Update __compatible_tag_interpreter_ to use _any()_ across the returned list to verify if at least one tag meets the requirement.

[...]

PR #1496, opened 2026-04-02
fix(pip): support multi-version python wheel tags

Anujkumar9081:
This PR addresses issue #1494.
Previously, _parse_py_version was extracting all valid versions from a multi-version Python wheel tag (like py38.py39.py310) but then forcefully reducing them down to just the max(versions). This caused Hermeto to incorrectly evaluate compatibility against a single version, accidentally rejecting perfectly valid, PEP-425-compliant wheels if the environment was set to one of the "lower" versions in the tag list.

**What I changed:**

- Refactored _parse_py_version in package_distributions.py to return the full list[int] of parsed versions.
-Updated _compatible_tag_interpreter to use any() so it evaluates compatibility across all the returned Python versions correctly.
-Added comprehensive unit tests in test_package_distributions.py to assert that multiple tags are correctly parsed as lists, and that parametric filtering logic properly accepts matching multi-version wheel tags.

### DIFF
--- a/hermeto/core/package_managers/pip/package_distributions.py
+++ b/hermeto/core/package_managers/pip/package_distributions.py
@@ -368,26 +368,27 @@ class WheelsFilter(BinaryPackageFilter):
             or any(impl in interpreter for impl in self.py_impl)
         )
 
-        wheel_py_version = _parse_py_version(interpreter)
-        compatible_py_version = (
+        wheel_py_versions = _parse_py_version(interpreter)
+        compatible_py_version = any(
             self.py_version is None
-            or wheel_py_version == self.py_version
-            or (wheel_py_version < self.py_version and abi in ("abi3", "none"))
+            or v == self.py_version
+            or (v < self.py_version and abi in ("abi3", "none"))
+            for v in wheel_py_versions
         )
 
         return compatible_py_impl and compatible_py_version
 
 
-def _parse_py_version(interpreter: str) -> int:
+def _parse_py_version(interpreter: str) -> list[int]:
     """
     >>> _parse_py_version("cp312")
-    312
+    [312]
     >>> _parse_py_version("pp312")
-    312
+    [312]
     >>> _parse_py_version("py3")
-    3
+    [3]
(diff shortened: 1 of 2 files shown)


Anujkumar9081:
Hi @eskultety / @a-ovchinnikov,

Just sending a gentle ping on this PR whenever you have a free moment.

I've pushed the fix we discussed over in #1494 to address the multi-version wheel tags. It looks like the CI workflows are currently awaiting maintainer approval to run. Please let me know if you have any feedback or if there are any changes you'd like me to make before we run the checks.

Thanks!

slimreaper35:
I remember this was one of my review comments when I was implementing the pip binary filtering.

eskultety:
@Anujkumar9081 your PR doesn't comply with our contributing guidelines. You didn't even run unit tests locally which is the least a developer can do given how simple it is for Hermeto. I have a very strong feeling that this was completely AI generated looking at this particular piece of code that is btw. wrong:
```
compatible_py_version = any(
            self.py_version is None
            or v == self.py_version
            or (v < self.py_version and abi in ("abi3", "none"))
            for v in wheel_py_versions
        )
```

I firmly stand by what I said in #1494 that this the scenario is completely fabricated, you didn't provide a real life use case and instead you insisted to be conformant with a particular PEP where the only real documented use case was Python2 and I told you we don't support Python 2 (it's been EOL'd 6 years ago). Note this is not how we intend to treat this project, we won't be merging just about any hypothetical change for the sake of getting a contribution. Based on the above I'm closing the PR.

Anujkumar9081:
Hi @eskultety, thank you for the honest feedback. You're absolutely right — I should have run the tests locally and ensured the code was correct before submitting. I apologize for not providing a concrete real-world example to back up the issue.

Regarding the AI concern — the code changes were done by me, though I did use AI to help polish the PR description and comments for better clarity. I understand how that can raise concerns, and I'll make sure to write everything in my own words going forward so my understanding comes through clearly.

I'll take these lessons seriously and make sure any future contributions are thoroughly tested and grounded in real problems. Thank you for your time and patience.

PR #1502, opened 2026-04-05
feat(cli): add config subcommand to show effective configuration

### LINKED ISSUE #1253: [RFE][cli] Add 'config' subcommand for runtime configuration introspection

Priyanshu-u07:
Closes #1253
Adds `hermeto config` to display effective configuration with env var names and highlight overridden values.

- `--diff`  show only values that differ from defaults
### Note

- YAML markers `(# (*))` use comments so output remains parseable.
- Env var names derived from `Config.model_config` and not hardcoded.
- Core logic in `config_show.py`, following the `envfile.py` pattern.

hermeto10
hermeto11


eskultety:
Why was this closed, I think we'd still want to consider this. It's just not a priority item right now. We'll get to it eventually, hopefully by the end of this quarter.

Priyanshu-u07:
Thank you for reopening this. I closed it thinking it was no longer being considered. I'm available to address any feedback whenever this gets reviewed.

tnevrlka:
Thank you for your contribution and patience, @Priyanshu-u07! 

Can you confirm whether the unit tests you introduce in the PR work for you? 
I am experiencing failures in multiple environments.

Furthermore, there seems to be a conflict with main now due to main getting updated while the PR was open. Could you please **rebase** this branch on top of current main? 
And finally, please drop the [Merge commit](https://github.com/hermetoproject/hermeto/pull/1502/changes/5377784cc853ca3221e2d2d04ff1516a38863203) from your PR. 

Thank you!

Priyanshu-u07:
Thanks for the review @tnevrlka! I've rebased onto current main, dropped the merge commit and fixed test failures.

Regarding test failures, the` Config` now includes a `mode` field (a Mode `enum`) and a `pnpm` section that didn't exist when this PR was first written. Updated `config_show.py` to handle both nested Pydantic model sections and scalar fields and updated tests accordingly.All tests pass locally.
**Ready for review!**

hermeto/interface/cli.py
tnevrlka, line 1: I would consider whether an additional test for CLI would be in order.
I would expect something similar to https://github.com/hermetoproject/hermeto/blob/2f91fbd06990bc2e40177c473ca3d3f1e1564040/tests/unit/test_cli.py#L97-L105

tests/unit/extras/test_config_show.py
```diff
@@ -0,0 +1,254 @@
...
+            ("gomod", "download_max_tries", "HERMETO_GOMOD__DOWNLOAD_MAX_TRIES"),
+            ("http", "connect_timeout", "HERMETO_HTTP__CONNECT_TIMEOUT"),
+            ("http", "read_timeout", "HERMETO_HTTP__READ_TIMEOUT"),
+            ("runtime", "subprocess_timeout", "HERMETO_RUNTIME__SUBPROCESS_TIMEOUT"),
+            ("runtime", "concurrency_limit", "HERMETO_RUNTIME__CONCURRENCY_LIMIT"),
+            ("pip", "ignore_dependencies_crates", "HERMETO_PIP__IGNORE_DEPENDENCIES_CRATES"),
+            ("yarn", "enabled", "HERMETO_YARN__ENABLED"),
+            ("npm", "proxy_url", "HERMETO_NPM__PROXY_URL"),
+            ("npm", "proxy_login", "HERMETO_NPM__PROXY_LOGIN"),
+            ("npm", "proxy_password", "HERMETO_NPM__PROXY_PASSWORD"),
```
tnevrlka, line 44: It looks like passwords are simply exposed in the subcommand.

I can imagine a usecase where the config would be displayed in a CI pipeline and thus passwords would be exposed in logs. 

I would consider whether an approach similar to kubectl would be appropriate, see excerpt from `kubectl config view --help`
```
# Show merged kubeconfig settings
kubectl config view

# Show merged kubeconfig settings, raw certificate data, and exposed secrets
kubectl config view --raw
```

Example:
```
❯ kubectl config view
- name: foo  
  user:
    token: REDACTED
```
```
❯ kubectl config view --raw
- name: foo  
  user:
    token: <actual token>
```

tnevrlka (commented):
Thank you for the contribution! The code looks pretty good to me overall, few notes

tests/unit/extras/test_config_show.py
eskultety: This is an excellent comment! Pydantic can often be a PITA to work with, but I have to say, `pydantic-settings` is hell of a help to back your configuration with. I just checked and they have us covered here too - https://pydantic.dev/docs/validation/latest/concepts/json_schema/#json-schema-types - it should be as easy as declaring the fields with the `SecretStr` type.

hermeto/interface/cli.py
Priyanshu-u07: Done. Added 4 CLI tests.

tests/unit/extras/test_config_show.py
Priyanshu-u07: Added `--raw` flag following the kubectl pattern. Sensitive fields like `proxy_password` are now redacted by default.
Regarding `SecretStr`, that's a good idea for the long term. For this PR, I've kept the current field types unchanged to avoid a broader refactor that would touch validators and proxy logic. Thanks!

eskultety: > Added `--raw` flag following the kubectl pattern. Sensitive fields like `proxy_password` are now redacted by default. Regarding `SecretStr`, that's a good idea for the long term. For this PR, I've kept the current field types unchanged to avoid a broader refactor that would touch validators and proxy logic. Thanks!

This is a new feature this PR is introducing. We won't accept a half-baked solution that primes the code for another refactor, sorry. The design of the feature will need to be re-adjusted such that we're happy with it on day 1 to make us want to maintain it, IOW knowing there's refactor waiting for us the moment we introduce a feature isn't the right development workflow in open source.

Priyanshu-u07: Thanks Erik, I understand. Updated to use `SecretStr` for `proxy_password` in `ProxyMixin`. Redaction is now built into the type system rather than a separate post-processing step. Updated all package managers to call `.get_secret_value()` where the actual password is needed. PTAL

tests/unit/extras/test_config_show.py
```diff
@@ -0,0 +1,288 @@
...
+                assert section_values["proxy_password"] != REDACTED
+
+    def test_non_sensitive_fields_unchanged(self) -> None:
+        config = Config()
+        default_output = get_effective_config(config)
+        raw_output = get_effective_config(config, raw=True)
+        assert (
+            default_output["gomod"]["download_max_tries"]
+            == raw_output["gomod"]["download_max_tries"]
+        )
```
tnevrlka: What is this test supposed to test? I'm not sure I understand what gomod download_max_tries should prove

tests/unit/extras/test_config_show.py
```diff
@@ -0,0 +1,288 @@
...
+                value = section_values["proxy_password"]
+                assert value is None or value == REDACTED
+
+    def test_proxy_password_revealed_with_raw(self) -> None:
+        config = Config()
+        effective = get_effective_config(config, raw=True)
+        for section_name, section_values in effective.items():
+            if isinstance(section_values, dict) and "proxy_password" in section_values:
+                assert section_values["proxy_password"] is None
+                assert section_values["proxy_password"] != REDACTED
```
tnevrlka: I don't think these tests work properly. I believe they depend on the environment, so if no `proxy_password` is set up, they will always pass even if the redacting logic doesn't work.

I believe you have the `_clean_hermeto_env` fixture for that?

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,243 @@
...
+
+import enum
+from typing import Any
+
+import yaml
+from pydantic import SecretStr
+
+from hermeto.core.config import Config
+
+REDACTED = "**********"
```
tnevrlka: optional: I would consider avoiding the `REDACTED` constant and just rely on Pydantic

> When the secret value is nonempty, it is displayed as '**********' instead of the underlying value in calls to repr() and str(). If the value is empty, it is displayed as ''.

https://pydantic.dev/docs/validation/latest/api/pydantic/types/#pydantic.types.SecretStr

tests/unit/test_cli.py
```diff
@@ -104,6 +104,29 @@ def test_list_backends_option(self) -> None:
...
+    def test_config_redacts_passwords_by_default(self) -> None:
+        result = invoke_expecting_sucess(app, ["config"])
+        for line in result.output.splitlines():
+            if "proxy_password:" in line and not line.strip().startswith("#"):
+                assert "**********" in line or "null" in line
+
+    def test_config_raw_shows_secrets(self) -> None:
+        result = invoke_expecting_sucess(app, ["config", "--raw"])
+        assert result.exit_code == 0
+        assert "**********" not in result.output
```
tnevrlka, line 152: Wouldn't it be simpler to set a password and check if it's in the config or not?

tnevrlka (commented):
Thank you, I think this is getting pretty close, left few comments and also CI has to be fixed before merging.
A lot of unit tests depend on the user environment and don't work properly if no relevant environment variables are set up.
Also, the commit message is now imo a bit outdated

tests/unit/extras/test_config_show.py
Priyanshu-u07: I agree with you, replaced with `test_raw_flag_does_not_affect_non_sensitive_fields`, now sets a real password and verifies `proxy_login` and `proxy_url` stay identical between default and raw modes.

tests/unit/extras/test_config_show.py
Priyanshu-u07: Fixed. Both tests now use `_clean_hermeto_env` and set an explicit `proxy_password="s3cret`" via the `Config()` constructor, so the redaction logic is always actually exercised

hermeto/core/extras/config_show.py
Priyanshu-u07: Done, removed the constant. Now using `str(value)` which delegates to Pydantic's `SecretStr.__str__()`.

tests/unit/test_cli.py
Priyanshu-u07: Agreed. Updated both tests now set `HERMETO_GOMOD__PROXY_PASSWORD=s3cret-value` via env and check if it appears in the output or not. Also added a config singleton reset to avoid stale caching.

Priyanshu-u07:
I've addressed all the review comments. Also fixed CI failures and updated the commit message. Ready for another look. Thanks!

tnevrlka (commented):
Overall looks pretty good, I think we're really close. 

Unfortunately I seem to hit unit tests failures when I have env vars set up. 
I believe unit tests should not depend on user's environment, please set some env vars for yourself and check. I think it should be a simple fix by just adding more `_clean_hermeto_env` fixtures. 

An example reproducer for inspiration:
```
HERMETO_GOMOD__PROXY_URL="https://override.example.com" \
HERMETO_GOMOD__PROXY_LOGIN="testuser" \
HERMETO_GOMOD__PROXY_PASSWORD="testpass"\
HERMETO_HTTP__READ_TIMEOUT=999 \
venv/bin/python3 -m pytest \
tests/unit/extras/test_config_show.py
```

Note that CI doesn't catch this because the runners don't have any env vars set up.

tnevrlka (commented):
Also, please take a look at the type annotations.

For example `dict[str, Any] | Any` evaluates to just `Any` as far as I know

Priyanshu-u07:
I've addressed both review comment. PTAL. Thanks

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,241 @@
...
+    effective: dict[str, Any],
+    defaults: dict[str, Any],
+) -> dict[str, dict[str, tuple[Any, Any]] | tuple[Any, Any]]:
+    """Compare effective config against defaults.
+
+    Returns only sections and fields where values differ.
+    For nested sections, each changed field maps to a tuple of (current, default).
+    For scalar fields, the section itself maps to a tuple of (current, default).
+    """
+    diff: dict[str, dict[str, tuple[Any, Any]] | tuple[Any, Any]] = {}
```
a-ovchinnikov: I am sorry, but what does this type annotation mean? Judging from it `{'a-ovchinnikov': (lambda: 42, dict.get)}` is a valid diff, are you sure that is correct? Please consider defining types or at least aliases which explain your intent, otherwise type hints become an annoying distraction instead of being a helping tool.

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,241 @@
...
+    deferred type annotations.
+    """
+    result: dict[str, Any] = {}
+    for name, field in Config.model_fields.items():
+        default = field.default
+        if _is_section(default):
+            result[name] = _get_section_dict(default)
+        else:
+            result[name] = _serialize_value(default)
+    return result
```
a-ovchinnikov, line 64: Why 
```python3
return Config().dict()
```
won't work?

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,241 @@
...
+
+    Example: ("gomod", "proxy_url") -> "HERMETO_GOMOD__PROXY_URL"
+    """
+    prefix = Config.model_config.get("env_prefix", "")
+    delimiter = Config.model_config.get("env_nested_delimiter", "__")
+    return f"{prefix}{section.upper()}{delimiter}{field.upper()}"
+
+
+def _serialize_value(value: Any, *, raw: bool = False) -> Any:
+    """Serialize a config value into a YAML/JSON-friendly representation."""
```
a-ovchinnikov: A well-defined Python object has a serializer defined. Why do you need to define an additional serializer for everything?

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,241 @@
...
+    SecretStr fields are redacted unless raw=True.
+    """
+    result: dict[str, Any] = {}
+    for name in Config.model_fields:
+        value = getattr(config, name)
+        if _is_section(value):
+            result[name] = _get_section_dict(value, raw=raw)
+        else:
+            result[name] = _serialize_value(value, raw=raw)
+    return result
```
a-ovchinnikov: Would
```python3
return Config().dict()
```
work here?

a-ovchinnikov (commented):
Please find a few drive-by comments inline.

hermeto/core/extras/config_show.py
Priyanshu-u07: Okay, I'll add type aliases to make the intent clear. Thanks!

hermeto/core/extras/config_show.py
Priyanshu-u07: `Config` extends `BaseSettings` (pydantic-settings), so `Config()` reads from env vars and config files. It won't return pure defaults. That's why I use `field.default` directly to get the schema defaults without any environment influence.

hermeto/core/extras/config_show.py
Priyanshu-u07: You're right, `model_dump(mode='json')` handles the default (redacted) case. The custom serializer was needed for the `--raw` flag which reveals SecretStr values, Pydantic doesn't have a built-in "reveal secrets" mode. I'm thinking of using `model_dump(mode='json')` as the base and only adding custom handling for `raw=True`. WDYT?

hermeto/core/extras/config_show.py
Priyanshu-u07: `Config().dict()` would create a new instance and ignore the `config` parameter passed to this function. But `config.model_dump(mode='json')` would work for the default (redacted) case. I'll refactor to use that as the base, with custom handling only for `--raw`.

hermeto/core/extras/config_show.py
a-ovchinnikov: If I understood you correctly this will be a better approach, secret string looks like the only class that needs special handling.

Priyanshu-u07:
Hi @a-ovchinnikov, I've addressed all the review comments. PTAL. Thanks!

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,222 @@
...
+
+    Derives the prefix and nested delimiter from Config.model_config rather than
+    hardcoding them, so that changes to the configuration schema are automatically
+    reflected.
+
+    Example: ("gomod", "proxy_url") -> "HERMETO_GOMOD__PROXY_URL"
+    """
+    prefix = Config.model_config.get("env_prefix", "")
+    delimiter = Config.model_config.get("env_nested_delimiter", "__")
+    return f"{prefix}{section.upper()}{delimiter}{field.upper()}"
```
a-ovchinnikov: Please add some doctests to this function, they will be more helpful than UTs.

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,222 @@
...
+                if isinstance(field_val, SecretStr):
+                    result[name][field_name] = field_val.get_secret_value()
+
+
+def get_effective_config(config: Config, *, raw: bool = False) -> dict[str, Any]:
+    """Get the current effective configuration as a nested dict.
+
+    Uses Pydantic's model_dump(mode='json') for serialization, which handles
+    SecretStr redaction, Enum conversion, HttpUrl stringification, etc.
+    SecretStr fields are redacted unless raw=True.
```
a-ovchinnikov, line 44: WDYT about using serializers:
```python3
class Foo(BaseModel):                                                                               
    foo: str                                                                                        
    bar: SecretStr                                                                                  
                                                                                                    
    @field_serializer("bar")                                                                        
    def reveal(self, v, i):                                                                      
        if i.context and i.context.get("reveal", "False"):                                       
            return v.get_secret_value()                                                                
        return "******" 
```
When I run it I see this:
```python3
$ python3 -i secret.py
>>> f = Foo(foo="foo", bar="bar")                                                                                                                                            
>>> f                                                                                                                                                                        
Foo(foo='foo', bar=SecretStr('**********'))                                                                                                                                  
>>> f.model_dump(mode="json")                                                                                                                                                
{'foo': 'foo', 'bar': '******'}                                                                                                                                                                                                                                                                                                 
>>> f.model_dump(mode="json", context={"reveal": True})                                                                                                                      
{'foo': 'foo', 'bar': 'bar'}                                                                                                                                                 
```

hermeto/core/extras/config_show.py
Priyanshu-u07: Thanks! Added doctests.

hermeto/core/extras/config_show.py
Priyanshu-u07: Completely agree with you. Added `field_serializer` on `ProxyMixin` with context-based reveal and removed the custom `_reveal_secrets` helper.

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,211 @@
...
+
+        if not isinstance(section_values, dict):
+            if section_values != default_values:
+                diff[section_name] = (section_values, default_values)
+            continue
+
+        section_defaults = default_values if isinstance(default_values, dict) else {}
+        section_diff: SectionDiff = {}
+
+        for field_name, current_value in section_values.items():
```
a-ovchinnikov: What would happen if someone adds more nesting to configs in the future? WDYT about making this a properly recursive walker? (This also applies to `get_default_config()`)

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+    if not diff:
+        return "# All values are at their defaults"
+
+    lines: list[str] = [
+        "# Only showing values that differ from defaults",
+        "",
+    ]
+
+    for section_name, section_diff in diff.items():
```
a-ovchinnikov: This again looks like something that could benefit from a generalized walker.

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,211 @@
...
+    return "\n".join(lines)
+
+
+def _format_yaml_value(value: ConfigValue) -> str:
+    """Format a single value for inline YAML representation.
+
+    Uses yaml.dump for correctness (handles quoting, special chars, etc.)
+    and strips trailing newlines/document markers.
+    """
+    if value is None:
```
a-ovchinnikov: WDYT about using a match here:
```python3
match value:
    case None: ...
    case bool(): ...
    case dict() | list(): ...
    case _: ...

File: tests/unit/extras/test_config_show.py [hunk 15]
```diff
@@ -0,0 +1,322 @@
...
+
+class TestGetEffectiveConfig:
+    """Tests for dumping current effective configuration."""
+
+    @pytest.mark.usefixtures("_clean_hermeto_env")
+    def test_returns_all_sections(self) -> None:
+        config = Config()
+        effective = get_effective_config(config)
+
+        expected_sections = {"pip", "yarn", "npm", "pnpm", "gomod", "http", "runtime", "mode"}
```
a-ovchinnikov: What happens once someone adds another section or modifies defaults? These tests couple implementation of any feature to config display and will break without no good reason, we should look for a better way of testing the feature.

tests/unit/extras/test_config_show.py
```diff
@@ -0,0 +1,322 @@
...
+        config = Config()
+        effective = get_effective_config(config)
+
+        expected_order = list(Config.model_fields.keys())
+        actual_order = list(effective.keys())
+        assert actual_order == expected_order
+
+    @pytest.mark.usefixtures("_clean_hermeto_env")
+    def test_field_order_within_sections_matches_model(self) -> None:
+        """Field order within each section must match the settings class definition."""
```
a-ovchinnikov, line 73: Could you please remind me, why was it decided to preserve the order?

tests/unit/extras/test_config_show.py
```diff
@@ -0,0 +1,322 @@
...
+
+        diff = get_config_diff(effective, defaults)
+        assert diff == {}
+
+
+class TestFormatYamlOutput:
+    """Tests for YAML output formatting."""
+
+    @pytest.mark.usefixtures("_clean_hermeto_env")
+    def test_output_is_valid_yaml(self) -> None:
```
a-ovchinnikov: These sort of tests better never be done directly because that makes them brittle by effectively tying them to specific implementations. A better way of verifying this would be to ensure that it is possible to read a config back from a full dump and that the config remains the same.

tests/unit/extras/test_config_show.py
```diff
@@ -0,0 +1,322 @@
...
+        defaults = {
+            "gomod": {"proxy_url": "https://proxy.golang.org,direct", "download_max_tries": 5},
+        }
+
+        output = format_yaml_output(effective, defaults)
+        # The changed value should have a star marker
+        assert "# (*)" in output
+        # The unchanged value should NOT have a star marker
+        lines = output.splitlines()
+        for line in lines:
```
a-ovchinnikov: In this particular test you know that you'll have just one star and you also know which line should have the star. Thus you could check these properties instead: this way you will be asking questions about something you control within arrange section of the test.

tests/unit/extras/test_config_show.py
```diff
@@ -0,0 +1,322 @@
...
+            if "download_max_tries" in line and not line.strip().startswith("#"):
+                assert "# (*)" not in line
+
+
+class TestFormatDiffOutput:
```
a-ovchinnikov, line 205: These tests are pretty hard to comprehend, WDYT about making them into doctests? I am not convinced these particular tests are needed, they will be brittle by definition, but if we are to introduce them maybe we should at least make it obvious what we are expecting to see.

hermeto/core/extras/config_show.py
Priyanshu-u07: Done, refactored `get_config_diff()` to a recursive walker. For `get_default_config()`, it already handles arbitrary nesting through Pydantic's` model_dump(mode='json')` so no change was needed there.

hermeto/core/extras/config_show.py
Priyanshu-u07: Done, refactored `format_diff_output()` and `format_yaml_output()` to use recursive walkers (`_walk_diff_lines` and `_walk_yaml_lines`).

hermeto/core/extras/config_show.py
Priyanshu-u07: Done.

tests/unit/extras/test_config_show.py
Priyanshu-u07: Agreed, replaced hardcoded section sets with `set(Config.model_fields.keys())` and removed the per-section default-value tests (`test_gomod_defaults`, etc.) since `test_matches_effective_when_no_overrides` already covers all defaults.

tests/unit/extras/test_config_show.py
Priyanshu-u07: Order is preserved so the CLI output is deterministic and predictable for users. This makes it easy to diff output across runs, only actual value changes show up, not field reordering noise. Pydantic's `model_dump()` already preserves model field order, so this test just ensures we don't accidentally break that property.

tests/unit/extras/test_config_show.py
Priyanshu-u07: Done, replaced with a roundtrip test (`test_yaml_roundtrip_matches_effective_config`) that parses the YAML output back and asserts `parsed == effective`.

tests/unit/extras/test_config_show.py
Priyanshu-u07: Done, now asserts exactly 1 star marker exists and it's on the `proxy_url` line.

tests/unit/extras/test_config_show.py
Priyanshu-u07: Added a doctest to `format_diff_output()` and replaced the brittle format-matching tests with a single roundtrip test. Kept `test_empty_diff` for the edge case.

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,219 @@
...
+import yaml
+
+from hermeto.core.config import Config
+
+# Type aliases for configuration diff structures.
+ConfigValue = str | int | float | bool | None | dict[str, Any] | list[Any]
+FieldDiff = tuple[ConfigValue, ConfigValue]  # (current_value, default_value)
+SectionDiff = dict[str, FieldDiff]
+# Values can also be nested ConfigDiff dicts for arbitrarily deep configs.
+ConfigDiff = dict[str, SectionDiff | FieldDiff]
```
a-ovchinnikov: Neat! _Now_ this tells a story! It tells a reader what to expect from the rest of the code. Thank you!

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,219 @@
...
+        default_value = defaults.get(key)
+
+        if isinstance(current_value, dict) and isinstance(default_value, dict):
+            sub_diff = get_config_diff(current_value, default_value)
+            if sub_diff:
+                diff[key] = sub_diff
+        elif current_value != default_value:
+            diff[key] = (current_value, default_value)
+
+    return diff
```
a-ovchinnikov, line 89: Look at how straightforward this code turned out to be!

tests/unit/extras/test_config_show.py
a-ovchinnikov: Ok, makes sense, thank you.

tests/unit/extras/test_config_show.py
a-ovchinnikov: Thank you.

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,219 @@
...
+            return str(value).lower()
+        case dict() | list():
+            if not value:
+                return "{}" if isinstance(value, dict) else "[]"
+            return yaml.dump(value, default_flow_style=True, width=float("inf")).strip()
+        case str():
+            # Use yaml.dump to handle quoting correctly
+            dumped = yaml.dump(value, default_flow_style=True, width=float("inf")).strip()
+            # yaml.dump adds "...\n" for simple strings, strip document end marker
+            if dumped.endswith("\n..."):
```
a-ovchinnikov: Looks like a typo: did you mean "...\n"? Also for Python 3.10+ you could use `.removesuffix()`:
```python3
>>> yaml.dump("foobar")
'foobar\n...\n'
>>> yaml.dump("foobar").removesuffix("...\n")
'foobar\n'
```

a-ovchinnikov (commented):
Almost there! Please fix the linter errors and make sure ITs pass and the change is good to go from my perspective.

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,218 @@
...
+            return str(value).lower()
+        case dict() | list():
+            if not value:
+                return "{}" if isinstance(value, dict) else "[]"
+            return yaml.dump(value, default_flow_style=True, width=float("inf")).strip()
+        case str():
+            # Use yaml.dump to handle quoting correctly
+            dumped = yaml.dump(value, default_flow_style=True, width=float("inf"))
+            # yaml.dump adds "...\n" for simple strings, strip document end marker
+            dumped = dumped.removesuffix("...\n").strip()
```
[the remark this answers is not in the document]
Priyanshu-u07: I think no current config field contain newlines. If multiline config values are added in the future, we can address this then.

hermeto/core/extras/config_show.py
Priyanshu-u07: Thanks! Credit goes to your review feedback.

hermeto/core/extras/config_show.py
Priyanshu-u07: Done

tnevrlka (approved):
Thanks, there seems to be one last linter issue, otherwise lgtm

Priyanshu-u07:
Thanks @a-ovchinnikov, @tnevrlka. Recent changes:

- Updated `bundler` and `cargo` to use `proxy_password.get_secret_value()`, both were missed since they landed after this branch was based. bundler was the mypy failure; cargo was the cause of the 401s (`SecretStr` interpolating as `**********` into the auth token).
- `_get_env_var_name` uses `or` defaults. Those keys are `str | None`, so `.get(key, default)` stayed optional and would break on an explicit `null`.
- `ConfigDiff` is now a recursive alias so the type matches what `get_config_diff` actually produces, which nests arbitrarily deep. `SectionDiff` is removed with it, since a section dict is already a valid `ConfigDiff`.
- `test_config_diff_command` now sets a real override and parses the output, asserting `runtime.concurrency_limit == 6` and that an unchanged section is absent. It previously accepted either possible output string, so it couldn't fail.
- `test_npm_proxy_credentials_propagate_to_registry_hosts` now asserts the actual `Authorization` header rather than just its presence, it would previously have passed with a redacted password.

PTAL when you get a chance.

a-ovchinnikov (approved):

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,217 @@
...
+    diff: ConfigDiff,
+) -> str:
+    """Format only changed values, showing current and default values.
+
+    Produces valid, parseable YAML with default values shown in comments.
+
+    >>> format_diff_output({"http": {"read_timeout": (600, 300)}})
+    '# Only showing values that differ from defaults\\n\\nhttp:\\n  read_timeout: 600  # default: 300\\n'
+    >>> format_diff_output({})
+    '# All values are at their defaults'
```
eskultety, line 159: Looking at https://github.com/hermetoproject/hermeto/pull/1502#discussion_r3686616916 I'm convinced in this particular case ^this brings much value due to the escaping, so I still think it's quite brittle and mostly just testing the "header" which may change and also could be skipped in the future to just let hermeto output 0 if there's no difference or the YAML if there is difference.
Given that it's been explicitly asked during review, I won't conflict with it, so let's just go through with this as is.

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,217 @@
...
+
+    _walk_yaml_lines(effective, defaults, lines, depth=0, env_parts=[])
+
+    return "\n".join(lines)
+
+
+def _walk_yaml_lines(
+    data: dict[str, Any],
+    defaults: dict[str, Any],
+    lines: list[str],
```
eskultety, line 118: This function takes a list of lines by reference which it mutates --> side-effect. I think the lines it we walked should be explicitly returned as an object instead of None to create a self-documenting interface IMO.

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,217 @@
...
+    ]
+
+    _walk_diff_lines(diff, lines, depth=0)
+
+    return "\n".join(lines)
+
+
+def _walk_diff_lines(
+    diff: ConfigDiff,
+    lines: list[str],
```
eskultety, line 176: same comment about mutating `lines` directly instead of returning it as an object as with the "walk" variant.

tests/unit/test_cli.py
```diff
@@ -104,6 +104,53 @@ def test_list_backends_option(self) -> None:
         assert "bar, foo" in lines[0]
         assert "x-baz" in lines[1]
 
+    def test_config_command(self) -> None:
+        result = invoke_expecting_sucess(app, ["config"])
+        assert "gomod:" in result.output
+        assert "# HERMETO_GOMOD__PROXY_URL" in result.output
```
eskultety, line 113: I think this also suffers from exec env leaking into this test. I think all the tests introduced in this module should get the same treatment with ` _clean_hermeto_env` as the other tests in test_config_show did.
For that though the fixture will have to move to `conftest` which I think might be beneficial for other tests in the future anyway.

hermeto/core/extras/config_show.py
```diff
@@ -0,0 +1,217 @@
+# SPDX-License-Identifier: GPL-3.0-only
```
eskultety, line 1: **Commit message:**

Just a tiny reminder - if any of the code changes have been assisted by an AI tool, please disclose it per our AI policy markdown, thank you.

eskultety (commented):
Despite the minor comments I made I'm very pleased with how this improvement has turned out, good job! This being a new feature I believe we should add a small section somewhere to the user docs here: https://hermetoproject.github.io/hermeto/latest/configuration/, just to let users know they can display the active configuration when needed which may get handy when debugging. So, if I can ask for one more respin, that will be great and I'll approve as well, LGTM!

tests/unit/test_cli.py
```diff
@@ -104,6 +104,58 @@ def test_list_backends_option(self) -> None:
         assert "bar, foo" in lines[0]
         assert "x-baz" in lines[1]
 
+    @pytest.mark.usefixtures("_clean_hermeto_env")
+    def test_config_command(self) -> None:
+        config_file.config = None
+        result = invoke_expecting_sucess(app, ["config"])
+        config_file.config = None
```
eskultety, line 111: why do we need ^this extra reset?

eskultety: Scratch this, it's because it's a global singleton.

tests/unit/test_cli.py
```diff
@@ -104,6 +104,58 @@ def test_list_backends_option(self) -> None:
...
+        assert "gomod:" in result.output
+        assert "# HERMETO_GOMOD__PROXY_URL" in result.output
+
+    def test_config_diff_command(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        for key in list(os.environ):
+            if key.startswith("HERMETO_") and not key.startswith("HERMETO_TEST_"):
+                monkeypatch.delenv(key)
```
eskultety: shouldn't this for loop be replaced with the cleaning fixture as well? Isn't it a bit more readable than this loop? :)

Priyanshu-u07: Thanks for the quick review! Replaced it with `_clean_hermeto_env`.  The test now only sets the single env var it needs.

hermeto/core/extras/config_show.py
Priyanshu-u07: Done

eskultety (approved):

tnevrlka:
Thanks @Priyanshu-u07!

a-ovchinnikov:
@Priyanshu-u07 congrats! It was a long journey, but you made it! Thank you!

Priyanshu-u07:
Thank you everyone for the reviews and for merging this! Looking forward to contributing more!!

PR #1510, opened 2026-04-10
Fix SSRF Vulnerability in HTTP Dependency Downloads

MohammadBinAftab:
This PR fixes a severe Server-Side Request Forgery (SSRF) vulnerability. When Hermeto fetches dependencies via HTTP, it passes user-provided URLs from package lock files (requirements.txt, artifacts.lock.yaml) directly into requests.get() and aiohttp without any network-level validation.

Why is this dangerous?
If an attacker or a compromised lockfile points a dependency to an internal system (for example, http://169.254.169.254/latest/meta-data/ on AWS runners), Hermeto running in a CI/CD environment will blindly fetch the cloud metadata. Cloud credentials will then be written into the deps/ output directory, bundling our secrets directly into the hermetic build outputs where an attacker can extract them.

The Fix
I added an is_safe_url() checker utility to hermeto/core/package_managers/general.py. By using Python's native socket and ipaddress libraries, it resolves the provided URL hostname and ensures the resulting IP address does NOT belong to:

A private subnet (10.x.x.x, 192.168.x.x, 172.16.x.x)
Loopbacks (127.x.x.x)
Link-local (Cloud metadata IPs like 169.254.169.254)
Both the synchronous and asyncio versions of the binary download functions will now safely error out before attempting to establish a connection with an internal server. I'm a new contributor here, so any feedback on this fix is highly appreciated!

### DIFF
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -22,6 +22,30 @@ from hermeto.core.http_requests import (
 )
 from hermeto.core.type_aliases import StrPath
 
+from urllib3.util import connection as urllib3_connection
+from aiohttp.resolver import DefaultResolver
+
+_orig_create_connection = urllib3_connection.create_connection
+
+def safe_create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT, source_address=None, socket_options=None):
+    host, port = address
+    ip = socket.gethostbyname(host)
+    ip_obj = ipaddress.ip_address(ip)
+    if ip_obj.is_private or ip_obj.is_loopback or ip_obj.is_link_local:
+        raise ValueError(f"SSRF blocked: host {host} resolved to internal IP {ip}")
+    return _orig_create_connection((ip, port), timeout, source_address, socket_options)
+
+urllib3_connection.create_connection = safe_create_connection
+
+class SafeResolver(DefaultResolver):
+    async def resolve(self, host, port, family):
+        resolved = await super().resolve(host, port, family)
+        for r in resolved:
+            ip_obj = ipaddress.ip_address(r['host'])
+            if ip_obj.is_private or ip_obj.is_loopback or ip_obj.is_link_local:
+                raise ValueError(f"SSRF blocked: resolved to internal IP {r['host']}")
+        return resolved
+
 pkg_requests_session = get_requests_session(retry_options={"allowed_methods": SAFE_REQUEST_METHODS})
 
 log = logging.getLogger(__name__)
@@ -70,7 +94,7 @@ def download_binary_file(
     timeout = (config.http.connect_timeout, config.http.read_timeout)
     try:
         resp = pkg_requests_session.get(
-            url, stream=True, verify=not insecure, auth=auth, timeout=timeout
+            url, stream=True, verify=not insecure, auth=auth, timeout=timeout, allow_redirects=False
         )
         resp.raise_for_status()
     except requests.RequestException as e:
@@ -124,6 +148,7 @@ async def _async_download_binary_file(
             auth=auth,
             raise_for_status=True,
             ssl=ssl_context,
+            allow_redirects=False,
         ) as resp:
             with open(download_path, "wb") as f:
                 while True:
@@ -165,11 +190,14 @@ async def async_download_files(
             aiohttp.ClientPayloadError,
         },
     )
+    custom_session = aiohttp.ClientSession(
+        connector=aiohttp.TCPConnector(resolver=SafeResolver()),
+        trust_env=True,
+    )
     retry_client = aiohttp_retry.RetryClient(
+        client_session=custom_session,
         retry_options=retry_options,
         trace_configs=[trace_config],
-        # respect proxy settings and .netrc
-        trust_env=True,
     )
 
     async with retry_client as session:
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -51,25 +51,6 @@ pkg_requests_session = get_requests_session(retry_options={"allowed_methods": SA
 log = logging.getLogger(__name__)
 
 
-def is_safe_url(url: str) -> bool:
-    """Check if the resolved IP of a URL is public to prevent SSRF attacks."""
-    try:
-        hostname = urlparse(url).hostname
-        if not hostname:
-            return False
-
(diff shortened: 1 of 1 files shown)


MohammadBinAftab (commented):
Thanks for the rigorous review, These were excellent catches. I have completely refactored the SSRF mitigation to address all four of those security and performance vectors:

DNS Rebinding (TOCTOU): I removed the redundant, separate is_safe_url step. Instead, I am now hooking perfectly into the network stack at resolution time.
Handle IPv6 & Multiple IPs (requests): I created a wrapper around urllib3.util.connection.create_connection. It now uses socket.getaddrinfo to resolve the host natively, loops to verify every single candidate IP (both IPv4 and IPv6) isn't on a private subnet, and then passes the validated IP directly to the original connector—meaning the underlying socket avoids re-resolving the hostname entirely.
Handle Event Loop Blocking (aiohttp): For our async workflows, I created a custom SafeResolver inheriting from DefaultResolver. It intercepts the resolution internally during the aiohttp lifecycle (avoiding any blocking of the main event loop) and asserts that none of the returned resolution targets fall within restricted subnets before passing them down.
Preventing Open Redirect Pivot Attacks: I have explicitly set allow_redirects=False in both the requests and aiohttp HTTP calls so that attackers cannot use external landing pages to pivot back into internal servers.
I've pushed these changes to the PR.

Priyanshu-u07:
Hi @MohammadBinAftab , DCO check is failing on this PR. You need to sign off your commits.
Please go through this: [CONTRIBUTING.md](https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md)
Thanks!

hermeto/core/package_managers/general.py
```diff
@@ -20,6 +22,41 @@
...
+
+urllib3_connection.create_connection = safe_create_connection
+
+class SafeResolver(DefaultResolver):
+    async def resolve(self, host, port, family):
+        resolved = await super().resolve(host, port, family)
+        for r in resolved:
+            ip_obj = ipaddress.ip_address(r['host'])
+            if ip_obj.is_private or ip_obj.is_loopback or ip_obj.is_link_local:
+                raise ValueError(f"SSRF blocked: resolved to internal IP {r['host']}")
```
eskultety, line 57: I'm sorry but this is a very naive implementation, because it dismisses the possibility of public A records in the resolved pool and instead just stops and raises on the first private address it finds, same for the global urllib monkey-patch above.

hermeto/core/package_managers/general.py
```diff
@@ -97,6 +134,7 @@ async def _async_download_binary_file(
             auth=auth,
             raise_for_status=True,
             ssl=ssl_context,
+            allow_redirects=False,
```
eskultety, line 137: This breaks legitimate redirects, even with legitimate services (all occurrences of this) with no handling offered.

hermeto/core/package_managers/general.py
```diff
@@ -20,6 +22,41 @@
 )
 from hermeto.core.type_aliases import StrPath
 
+from urllib3.util import connection as urllib3_connection
+from aiohttp.resolver import DefaultResolver
+
+_orig_create_connection = urllib3_connection.create_connection
```
eskultety, line 28: `create_connection` is not part of the public API:
https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html
https://github.com/urllib3/urllib3/blob/c21b1c16cd8c08dd26876e223bc198a2b4de4fb3/src/urllib3/util/__init__.py

hermeto/core/package_managers/general.py
```diff
@@ -20,6 +22,41 @@
...
+            return _orig_create_connection((ip, port), timeout, source_address, socket_options)
+        except Exception as e:
+            err = e
+            continue
+            
+    if err is not None:
+        raise err
+    raise socket.error("getaddrinfo returned an empty list")
+
+urllib3_connection.create_connection = safe_create_connection
```
eskultety, line 49: This has a global side-effect - this replaces the function for EVERY single connection that is opened in the process, not just deps fetching, everything, any connection that Hermeto opens unaffected by user actions or input now denies contacting private IP addresses.

eskultety (commented):
@MohammadBinAftab First things first - regardless of the fact that the PR doesn't follow our contributing guidelines, by not following our documented security process via Github advisories **were this a real serious Hermeto vulnerability** you publicly disclosed and created an N-day vulnerability. I appreciate the intent, but I hope from now on you'll treat security with a more sensitive approach!

While SSRF is a serious problem, Hermeto (as an application) is not the place for the changes introduced by the PR. The PR not only wraps link-local IPs, but also all private IPs. Hermeto was designed to operate within any infrastructure - public OR private - and so private IPs on their own do not pose a threat. Note Hermeto is a simple (and pretty dumb) build pipeline tool rather than a web service and so it is desirable to fetch from all locations pointed to as we are not in a position to verify whether a given location is allowed/trusted or not.
Therefore, any SSRF mitigations should be done on the infra level, this would include but would not be limited to:

- strict network policies
- firewall rules
- strict IAM roles for all job runners

Conclusion, this report & severity assessment would belong to a particular service integrating Hermeto not Hermeto itself.
Unless I overlooked some serious aspect, please speak up otherwise I'll close the PR by the end of the week.

hermeto/core/package_managers/general.py
MohammadBinAftab: Hi @eskultety,

Thank you so much for taking the time to write out such a detailed review. First and foremost, you are completely right, I sincerely apologize for opening a public PR for a potential vulnerability. It was a lapse in judgment on my part as a newer contributor, and I will strictly use GitHub Security Advisories and responsible disclosure channels moving forward.

Regarding the architecture, your point makes perfect sense. I was looking at this through the lens of a web service SSRF, but you're absolutely right that a dumb pipeline tool should faithfully execute fetches to any IP, otherwise, we would break legitimate pulls from private local registries (like internal Artifactory/Nexus networks). Blocking private subnets at the application level here is the wrong paradigm, and it correctly belongs at the infrastructure/IAM/firewall level.

Thank you again for the fair and educational feedback (and for pointing out the flaws in the urllib global patch). I am closing this PR now since the underlying premise does not align with Hermeto's operational design. I look forward to contributing in more productive (and properly process-compliant) ways in the future!

PR #1513, opened 2026-04-13
fix(pip): support multi-version python wheel tags

### LINKED ISSUE #1494: [Bug]: Pip Wheel Filter incorrectly rejects multi-version Python wheel tags for valid target environments

Anujkumar9081:
Fixes #1494

This PR updates the pip wheel filter logic to properly support multi-version compressed tags (e.g., `py38.py39.py310`), commonly used by active projects like `semgrep`.

**Changes made:**

-Refactored `_parse_py_version` to `_parse_py_versions` to return a list of all parsed versions instead of just the maximum version.
-Updated `_compatible_tag_interpreter` to evaluate compatibility using `any()` across the returned list.
-Fixed the logic flaw from my previous attempt by keeping the `self.py_version is None` check safely outside the iteration loop.
-Added a specific test case in `test_package_distributions.py` simulating the `semgrep` artifact to explicitly verify multi-version tag compatibility.

**Verification:**

-Verified locally that nox -s python-3.10 passes 100% of tests.
-Formatted and checked with ruff.

Thanks to @eskultety for the feedback on the previous attempt and for helping me get the logic and testing right this time around.

### DIFF
--- a/hermeto/core/package_managers/pip/package_distributions.py
+++ b/hermeto/core/package_managers/pip/package_distributions.py
@@ -368,26 +368,32 @@ class WheelsFilter(BinaryPackageFilter):
             or any(impl in interpreter for impl in self.py_impl)
         )
 
-        wheel_py_version = _parse_py_version(interpreter)
-        compatible_py_version = (
-            self.py_version is None
-            or wheel_py_version == self.py_version
-            or (wheel_py_version < self.py_version and abi in ("abi3", "none"))
+        wheel_py_versions = _parse_py_versions(interpreter)
+        compatible_py_version = self.py_version is None or any(
+            v == self.py_version or (v < self.py_version and abi in ("abi3", "none"))
+            for v in wheel_py_versions
         )
 
         return compatible_py_impl and compatible_py_version
 
 
-def _parse_py_version(interpreter: str) -> int:
-    """
-    >>> _parse_py_version("cp312")
-    312
-    >>> _parse_py_version("pp312")
-    312
-    >>> _parse_py_version("py3")
-    3
-    >>> _parse_py_version("py2.py3")
-    3
+def _parse_py_versions(interpreter: str) -> list[int]:
+    """Parse all Python version numbers from a wheel interpreter tag.
+
+    Supports multi-version compressed tags per PEP 425 (e.g. py38.py39.py310).
+    Some real-world packages on PyPI use these compressed tags, for example
+    semgrep (https://pypi.org/simple/semgrep/).
+
+    >>> _parse_py_versions("cp312")
+    [312]
+    >>> _parse_py_versions("pp312")
+    [312]
+    >>> _parse_py_versions("py3")
+    [3]
+    >>> _parse_py_versions("py2.py3")
+    [2, 3]
+    >>> _parse_py_versions("py38.py39.py310")
+    [38, 39, 310]
     """
     parts = interpreter.split(".")
     versions = []
@@ -400,4 +406,4 @@ def _parse_py_version(interpreter: str) -> int:
     if not versions:
         raise RuntimeError(f"Invalid wheel interpreter: {interpreter}")
 
-    return max(versions)
+    return versions
--- a/tests/unit/package_managers/pip/test_package_distributions.py
+++ b/tests/unit/package_managers/pip/test_package_distributions.py
@@ -10,7 +10,6 @@ from hermeto.core.errors import FetchError, PackageRejected
 from hermeto.core.models.input import PipBinaryFilters
 from hermeto.core.package_managers.pip.package_distributions import (
     WheelsFilter,
-    _parse_py_version,
     _process_prefer_binary_mode,
     _sdist_preference,
     process_package_distributions,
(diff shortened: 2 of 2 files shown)


hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -368,26 +368,28 @@ def _compatible_tag_interpreter(self, interpreter: str, abi: str) -> bool:
...
+    Supports multi-version compressed tags per PEP 425 (e.g. py38.py39.py310).
+
+    >>> _parse_py_versions("cp312")
+    [312]
+    >>> _parse_py_versions("pp312")
+    [312]
+    >>> _parse_py_versions("py3")
+    [3]
+    >>> _parse_py_versions("py2.py3")
+    [2, 3]
```
a-ovchinnikov, line 394: Could you please add a link to the package that demonstrates the need for such change as either a comment or a part of docstring?

tests/unit/package_managers/pip/test_package_distributions.py
```diff
@@ -485,10 +485,11 @@ def test_init_with_invalid_values_fails(self, filter_kwargs: dict) -> None:
             PipBinaryFilters(**filter_kwargs)
 
     def test_parse_py_version_from_interpreter(self) -> None:
```
a-ovchinnikov: Do we still need the unit test as long as there are doctests present? I would recommend expanding doctests and dropping the UT.

hermeto/core/package_managers/pip/package_distributions.py
a-ovchinnikov, line 1: _Commit message_: please add a few lines explaining this commit, the general descriptive sentence from the PR is a good starting point. You would want to tell your reader what was the problem before this commit and briefly mention how you fixed it. Briefly is the key word here: the rest of your commit describes how you fixed the issue in detail. This is also a good spot to mention a package which could not be processed properly with the previous iteration of the code.

hermeto/core/package_managers/pip/package_distributions.py
Anujkumar9081: Done! I've added a reference to the semgrep package (https://pypi.org/simple/semgrep/) directly in the docstring of `_parse_py_versions`, along with an expanded doctest covering the multi-version case (`py38.py39.py310`).

tests/unit/package_managers/pip/test_package_distributions.py
Anujkumar9081: Good call - I've expanded the doctests to cover the multi-version tag scenario (py38.py39.py310 -> [38, 39, 310]) and removed the standalone unit test. The integration-level coverage for multi-version filtering is still retained via the multi_version_tag_matches parametrized case in test_filter_with_specific_py_version.

hermeto/core/package_managers/pip/package_distributions.py
Anujkumar9081: Updated! The commit message now explains the problem (only the maximum version was returned), mentions the semgrep package as a real-world example that couldn't be processed correctly, and briefly describes the fix. Thanks for the guidance on making the commit self-documenting.

a-ovchinnikov (approved):

eskultety (approved):
I wonder how many more wheels like semgrep using the tag compression scheme this way could be found. Anyway, turns out I was wrong in my "fabricated" scenario claim in the end, apologies.
The change itself is compact and non-intrusive -> ACK.

Anujkumar9081:
Thank you, @eskultety! No apologies needed at all — your skepticism actually pushed me to dig deeper and find real-world evidence with `semgrep`, which made the fix and the test case much stronger. That's exactly the kind of rigorous review that makes a project better, and I genuinely appreciate it.

I'm really enjoying contributing to Hermeto and learning from the review process here. Looking forward to picking up more issues! 🚀

PR #1428, opened 2026-03-18
fix: improve error messages in pip requirements

harshitbansal184507:
Resolves #1427
While going through `requirements.py` to understand the pip module, I noticed that when a git dependency is used alongside hashed dependencies, hermeto raises a misleading error telling the user to add a `--hash` to a git dependency.
## Changes

1) Detect VCS requirements in `validate_requirements_hashes()` and raise a specific error explaining that git dependencies do not support pip's   `--hash` option, pointing to the https URL alternative
2)Fix two missing `\n` separators in  solution strings that caused sentences to concatenate without any separator

PR #1514, opened 2026-04-13
docs(pip): fix external dependency output  structure

harshitbansal184507:
The docs showed external deps stored in subdirectories (external-dockerfile-parse/), but is stored flat in deps/pip/ alongside PyPI deps. Updated the directory tree and description

### DIFF
--- a/docs/pip.md
+++ b/docs/pip.md
@@ -613,18 +613,17 @@ an example,
 
 ```diff
 - dockerfile-parse @ https://github.com/.../2.0.0.tar.gz \
-+ dockerfile-parse @ file:///absolute-path/hermeto-output/deps/pip/.../dockerfile-parse-...tar.gz
++ dockerfile-parse @ file:///absolute-path/hermeto-output/deps/pip/dockerfile-parse-....tar.gz
 ```
 
-External dependencies are stored a bit further down the deps/pip tree to avoid
-mixing them with PyPI dependencies. The path and filename is an implementation
-detail.
+External dependencies are stored flat alongside PyPI dependencies in
+deps/pip/. The filename includes the checksum hash. The exact filename
+format is an implementation detail.
 
 ```text
 hermeto-output/deps/pip
 ├── ...
-├── external-dockerfile-parse
-│   └── dockerfile-parse-external-sha256-36e4469abb0d96b0e3cd656284d5016e8a674cd57b8ebe5af64786fe63b8184d.tar.gz
+├── dockerfile-parse-36e4469abb0d96b0e3cd656284d5016e8a674cd57b8ebe5af64786fe63b8184d.tar.gz
 └── ...
 ```
 


docs/pip.md
```diff
@@ -612,19 +612,17 @@ We can look at the `git diff` to see what the package remapping looks like. As
 an example,
 
 ```diff
-- dockerfile-parse @ https://github.com/.../2.0.0.tar.gz \
-+ dockerfile-parse @ file:///absolute-path/hermeto-output/deps/pip/.../dockerfile-parse-...tar.gz
+- dockerfile-parse @ https://github.com/containerbuildsystem/dockerfile-parse/archive/refs/tags/2.0.0.tar.gz \
++ dockerfile-parse @ file:///tmp/hermeto-output/deps/pip/dockerfile-parse-36e4469abb0d96b0e3cd656284d5016e8a674cd57b8ebe5af64786fe63b8184d.tar.gz \
```
derasdf: The `+` line uses `\` continuation with `--hash=` on a separate line, but `PipRequirement.__str__` generates single-line output. No continuation. The `--hash=` line also lacks a `+` prefix (reads as unchanged context in diff format).

Did you verify by running hermeto on a project with URL deps?

docs/pip.md
```diff
@@ -612,19 +612,17 @@ We can look at the `git diff` to see what the package remapping looks like. As
...
+      --hash=sha256:36e4469abb0d96b0e3cd656284d5016e8a674cd57b8ebe5af64786fe63b8184d
 ```
 
-External dependencies are stored a bit further down the deps/pip tree to avoid
-mixing them with PyPI dependencies. The path and filename is an implementation
-detail.
+External dependencies are stored flat alongside PyPI dependencies in deps/pip/. The filename includes the checksum hash. The exact filename format is an implementation detail.
```
derasdf: nitpick: ~175-char line. Rest of the doc wraps at ~80.

```
External dependencies are stored flat alongside PyPI dependencies in
deps/pip/. The filename includes the checksum hash. The exact filename
format is an implementation detail.
```

derasdf (changes requested):
`s/docs(pip): fix external dependency output  structure/docs: pip: Fix external dependency output structure/` - convention is colon-separated prefixes, capitalized verb. Double space typo too :)

Commit body is missing the WHY. When was this introduced? How did you find it? (7d069288 changed this ~10 months ago.) Put it in the commit message.

Also: `s/but is actually stored/but they are actually stored/`

nitpick: 3 open PRs (#1419, #1428, #1514), two for a month. I'd suggest slowing down and polishing each PR before moving to the next, it makes reviews smoother for everyone :)

eskultety:
> nitpick: 3 open PRs (https://github.com/hermetoproject/hermeto/pull/1419, https://github.com/hermetoproject/hermeto/pull/1428, https://github.com/hermetoproject/hermeto/pull/1514), two for a month. I'd suggest slowing down and polishing each PR before moving to the next, it makes reviews smoother for everyone :)

Yes, but trivial changes, especially to docs like this one, is not a problem IMO, docs should always be a first-class citizen. That said, yes, the PR pool is still massive and we'll only slowly going to chew through it.

docs/pip.md
harshitbansal184507: To be honest, no,  while working on my proposal, I referenced the output structure from the docs without paying close attention to the actual pip output directly. Later, when I was testing uv integration, I wanted to mirror pip's output directory structure for uv, so I was directly using the generated output. That's when I noticed the actual structure didn't match what the docs described and I quickly created this PR. I'll correct it.

derasdf: Fair, now it is better, just fix few typos in commit

1. Double space: `output  structure`
2. `7d06928 .` in the body has a stray space before the period.

eskultety (approved):

derasdf (approved):

PR #1516, opened 2026-04-14
Implement experimental maven backend

slimreaper35:


hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,235 @@
...
+        ],
+        project_files=[
+            ProjectFile(
+                abspath=request.output_dir.path / "settings.xml", template=SETTINGS_XML_TEMPLATE
+            ),
+        ],
+    )
+
+
+def _resolve_maven_project(project_dir: Path, deps_dir: Path) -> list[Component]:
```
brunoapimentel, line 80: We also need to take into account the `pom` key that lives exclusively on root level of the lockfile. It represents the POM of the project that is being processed, so we don't need to fetch it, but we need to fetch any BOMs or parentPoms it might contain.

brunoapimentel: Here's an example:
```json
{
  "artifactId": "model-mesh",
  "groupId": "com.ibm.watson",
  "version": "0.4.2-SNAPSHOT",
  "pom": {
    "groupId": "com.ibm.watson",
    "artifactId": "model-mesh",
    "version": "0.4.2-SNAPSHOT",
    "relativePath": "pom.xml",
    "checksumAlgorithm": "SHA-256",
    "checksum": "e60f7568677d66e8b3563a03fb84e1f0deac563395cf9c560ccfca0de3b77644",
    "boms": [
      {
        "groupId": "io.grpc",
        "artifactId": "grpc-bom",
        "version": "1.63.2",
        "resolved": "https://repo.maven.apache.org/maven2/io/grpc/grpc-bom/1.63.2/grpc-bom-1.63.2.pom",
        "repositoryId": "central",
        "checksumAlgorithm": "SHA-256",
        "checksum": "7dd51b2af8afbf1b228ba003ef0119d247362f2eb1e9273aa61346560bf4099a"
      },
      {
        "groupId": "io.netty",
        "artifactId": "netty-bom",
        "version": "4.1.125.Final",
        "resolved": "https://repo.maven.apache.org/maven2/io/netty/netty-bom/4.1.125.Final/netty-bom-4.1.125.Final.pom",
        "repositoryId": "central",
        "checksumAlgorithm": "SHA-256",
        "checksum": "c4a17c2b555c756e35c6c250cf57528c3d2ec8601e45e00e1c45d3a1f1cf0216",
        "parent": {
          "groupId": "org.sonatype.oss",
          "artifactId": "oss-parent",
          "version": "7",
          "resolved": "https://repo.maven.apache.org/maven2/org/sonatype/oss/oss-parent/7/oss-parent-7.pom",
          "repositoryId": "central",
          "checksumAlgorithm": "SHA-256",
          "checksum": "b51f8867c92b6a722499557fc3a1fdea77bdf9ef574722fe90ce436a29559454"
        }
      }
    ]
  },
//...
}
````

File: hermeto/core/package_managers/maven/main.py [hunk 2]
```diff
@@ -0,0 +1,235 @@
...
+        environment_variables=[
+            EnvironmentVariable(
+                name="MAVEN_OPTS",
+                value="-Dmaven.repo.local=${output_dir}/deps/maven",
+            ),
+            EnvironmentVariable(
+                name="MAVEN_ARGS",
+                value="-s ${output_dir}/settings.xml",
+            ),
+        ],
```
brunoapimentel: I believe MAVEN_OPTS and MAVEN_ARGS here are redundant. The "MAVEN_OPT" we're setting is the same thing that the `<localRepository>` key in `settings.xml` is doing. We can probably drop MAVEN_OPTS.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,235 @@
...
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.2.0 https://maven.apache.org/xsd/settings-1.2.0.xsd">
+  <localRepository>${output_dir}/deps/maven</localRepository>
+  <mirrors>
+    <mirror>
+      <id>hermeto-local</id>
+      <mirrorOf>*</mirrorOf>
+      <url>file://${output_dir}/deps/maven</url>
+    </mirror>
+  </mirrors>
```
brunoapimentel, line 36: I am not sure why we need to set the `mirror` here, from the tests I was doing, setting `localRepository` was enough to persuade Maven to use the local cache. 🤔

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,235 @@
...
+    get_checksum_algorithm,
+)
+
+log = logging.getLogger(__name__)
+
+DEFAULT_LOCKFILE = "lockfile.json"
+
+SETTINGS_XML_TEMPLATE = """\
+<?xml version="1.0" encoding="UTF-8"?>
+<settings xmlns="http://maven.apache.org/SETTINGS/1.2.0"
```
brunoapimentel, line 25: I believe it might be beneficial to set the `offline` option here too (see the [docs](https://maven.apache.org/settings.html#simple-values)). I don't have any concrete case to back up my claim, though, but I do remember that I saw a project that behaved differently when it was built with `mvn -o`. 

Maybe we can skip this for now and investigate it futher later.

brunoapimentel (commented):
Overall, it looks solid to me, we can iterate on top of this as the lockfile gets to the point we need to build simple projects.

hermeto/core/models/input.py
```diff
@@ -112,6 +112,7 @@ def show_error(error: "ErrorDict") -> str:
     "rpm",
     "yarn",
     # Add experimental package managers here with x- prefix, e.g. "x-foo"
+    "x-maven",
```
eskultety, line 115: Can we at least get some reasoning in the commit message? Some commit message would be nice. This is not our typical backend implementation where we're working with a well-structured ecosystem, so I get the part we don't have a prior design and we're experimenting, but I'd still like to know at least the current direction before we converge to something final and official.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,235 @@
...
+    """Create a _remote.repositories file for each artifact."""
+    now = datetime.now(timezone.utc).strftime("%a %b %d %H:%M:%S %Z %Y")
+    for artifact in artifacts:
+        artifact_dir_abs_path = deps_dir / artifact.artifact_relative_dir
+        remote_repos_file = artifact_dir_abs_path.joinpath("_remote.repositories")
+        lines = [
+            "#NOTE: This is a Maven Resolver internal implementation file, its format can be changed without prior notice.\n",
+            f"#{now}\n",
+        ]
+        lines.extend(f"{artifact.filename}>{derive_repository_id(artifact.url)}=\n")
```
eskultety: `extend` should be replaced with `append`, extend would take the string and extend lines with individual symbols.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,235 @@
...
+            ProjectFile(
+                abspath=request.output_dir.path / "settings.xml", template=SETTINGS_XML_TEMPLATE
+            ),
+        ],
+    )
+
+
+def _resolve_maven_project(project_dir: Path, deps_dir: Path) -> list[Component]:
+    """Resolve and fetch Maven artifacts for the given project."""
+    lockfile = MavenLockfile.from_file(project_dir / DEFAULT_LOCKFILE)
```
eskultety: We should incorporate a basic check on an absent lockfile and emit `LockfileNotFound` before trying to parse it.

hermeto/core/package_managers/maven/models.py
```diff
@@ -0,0 +1,102 @@
...
+    @classmethod
+    def from_file(cls, path: Path) -> "MavenLockfile":
+        """Create a MavenLockfile object from the provided path."""
+        with path.open() as f:
+            data = json.load(f)
+
+        return cls(path, data)
+
+
+class MavenArtifact(UserDict):
```
eskultety: We usually don't model with `UserDict` unless we may need to update the object. The field enumeration below are informative only, but lack any kind of schema enforcement, we usually either use dataclasses or Pydantic models, what was the motivation here behind `UserDict`, I can only see `artifact.get("scope", "compile")` in `main.py:117` that warrants it, but is it really needed? Shouldn't we be more strict in this representation unless we expect to be modifying it?

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,235 @@
...
+    result: list[Component] = []
+    for artifact in artifacts:
+        purl = PackageURL(
+            type="maven",
+            namespace=artifact.group_id,
+            name=artifact.artifact_id,
+            version=artifact.version,
+        )
+
+        name = f"{artifact.group_id}.{artifact.artifact_id}"
```
eskultety: Isn't the standard convention `groupId:artifactId`? Or would it collide with a particular SBOM construct or our generator logic?

hermeto/core/package_managers/maven/utils.py
```diff
@@ -0,0 +1,50 @@
...
+from hermeto.core.errors import PackageRejected
+
+JAVA_TO_PYTHON_CHECKSUM_ALGORITHMS = {
+    "SHA-256": "sha256",
+    "SHA-1": "sha1",
+    "SHA-512": "sha512",
+    "SHA-224": "sha224",
+    "SHA-384": "sha384",
+    "MD5": "md5",
+}
```
eskultety: Sounds like a simple `replace` and `lower` approach rather than having a static map in memory :) .

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+async def _download_optional_file(session: aiohttp.ClientSession, url: str, path: Path) -> None:
+    suffixes = (".sha1", ".md5", ".sha256", ".sha512", ".sha224", ".sha384")
+    async with session.get(url, raise_for_status=False) as response:
+        if response.status == 404:
+            log.debug("Skipping %s", url)
+            return
+
+        content = await response.read()
```
eskultety: What about other response codes? This sounds like we'd write other HTTP errors as checksum content.

eskultety (commented):
I think some basic unit test coverage would be nice to have if we wish to build on the proposed archicture (which I think is the whole point).

brunoapimentel:
@slimreaper35 The latest [commit](https://github.com/chains-project/maven-lockfile/commit/5ef77f3778f4d6bcc6d287bc6e9bfb4540223cda) of the lockfile plugin already produces a lockfile with the parent POMs, not sure if you prefer to include it in this PR or to expand on it as a follow up.

With this, I believe we can already have a e2e test scenario. Just note that it was not oficially released yet, so you need to clone and build the project locally to use the latest version.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,235 @@
...
+    )
+
+
+def _resolve_maven_project(project_dir: Path, deps_dir: Path) -> list[Component]:
+    """Resolve and fetch Maven artifacts for the given project."""
+    lockfile = MavenLockfile.from_file(project_dir / DEFAULT_LOCKFILE)
+    deps = parse_maven_dependencies(lockfile)
+    plugins = parse_maven_plugins(lockfile)
+    boms = parse_maven_boms(deps)
+
```
derasdf, line 85: `deps + plugins + boms` are concatenated without deduplication. If the same artifact appears as both a direct dependency and a transitive dependency of a plugin, it gets downloaded twice and shows up twice in the SBOM. Is that intentional? Other backends (e.g. gomod, npm) deduplicate components before generating the SBOM.

hermeto/core/package_managers/maven/main.py
brunoapimentel: Yes, the standard is [here](https://maven.apache.org/repositories/artifacts.html). But that is meant to be used as a way to identify an artifact in a Maven repository. We don't necessarily need to use it in the SBOM.

I'm not sure what's the better choice here, though, but here's how the [Maven SPDX plugin](https://github.com/spdx/spdx-maven-plugin) does it:

```
{
    "SPDXID" : "SPDXRef--212e7d1b0",
    "copyrightText" : "UNSPECIFIED",
    "description" : "Core Jackson processing abstractions (aka Streaming API), implementation for JSON",
    "downloadLocation" : "NOASSERTION",
    "externalRefs" : [ {
      "referenceCategory" : "PACKAGE-MANAGER",
      "referenceLocator" : "pkg:maven/tools.jackson.core/jackson-core@3.0.2",
      "referenceType" : "purl"
    } ],
    "filesAnalyzed" : false,
    "homepage" : "https://github.com/FasterXML/jackson-core",
    "licenseConcluded" : "NOASSERTION",
    "licenseDeclared" : "Apache-2.0",
    "name" : "Jackson-core",
    "originator" : "Organization:FasterXML",
    "summary" : "Core Jackson processing abstractions (aka Streaming API), implementation for JSON",
    "versionInfo" : "3.0.2"
  }
```

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,235 @@
...
+    artifact_dir_abs_path = deps_dir / artifact.artifact_relative_dir
+    artifact_dir_abs_path.mkdir(parents=True, exist_ok=True)
+    return artifact_dir_abs_path / artifact.filename
+
+
+def _download_maven_artifacts(deps_dir: Path, artifacts: list[MavenArtifact]) -> None:
+    """Download Maven dependencies."""
+    config = get_config()
+
+    download_paths = {a.url: _create_artifact_dir(deps_dir, a) for a in artifacts}
```
derasdf, line 131: The lockfile has checksums for every artifact but downloaded content is never verified against them. `_create_checksums_files` writes the expected hash to a sidecar file, which means a tampered artifact passes through with a correct-looking checksum next to it.

hermeto/core/package_managers/maven/models.py
```diff
@@ -0,0 +1,163 @@
...
+    'sha256'
+    >>> _get_checksum_algorithm("SHA-512")
+    'sha512'
+    >>> _get_checksum_algorithm("MD5")
+    'md5'
+    """
+    return java_algorithm.replace("-", "").lower()
+
+
+def _extract_pom_chain(pom: dict[str, Any] | None, result: list[MavenArtifact]) -> None:
```
a-ovchinnikov, line 105: My main concern here is that all `_extract...` helpers rely on modification of objects in enclosing scope. I would recommend avoiding this unless absolutely necessary, especially when there will be nested calls that modify the same object. (I won't push too hard since the bulk of code is already there.)

hermeto/core/package_managers/maven/models.py
slimreaper35: 🤔 It might have been a leftover. I changed it to the Pydantic base class.

hermeto/core/package_managers/maven/main.py
eskultety: That sounds official :). Can we reference the SPDX plugin here to clear out the confusion? We can then resolve.

slimreaper35: Shouldn't the name be just `artifactId` to match the PURL params? (namespace=groupId, name=artifactId, version=version)

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,199 @@
...
+        name=name,
+        purl=purl.to_string(),
+        version=version,
+        properties=[Property(name=PropertyEnum.PROP_MAVEN_SCOPE, value="compile")],
+    )
+
+
+def _create_artifact_dir(deps_dir: Path, artifact: MavenArtifact) -> Path:
+    """Create a directory for the Maven artifact to download."""
+    artifact_dir_abs_path = deps_dir / artifact.artifact_relative_dir
```
eskultety, line 123: This will be a uniquely named dir, correct?

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,199 @@
...
+        components.append(component)
+
+    return components
+
+
+def _generate_main_component(lockfile: MavenLockfile) -> Component:
+    """Get the main component from the Maven lockfile."""
+    group_id = lockfile.data["groupId"]
+    artifact_id = lockfile.data["artifactId"]
+    version = lockfile.data["version"]
```
eskultety: `MavenLockfile` is pretty much an envelope over a dict, but these fields are coming from a user file. I suppose at least some of them are mandatory by the spec. Can we enforce them rather than letting a `KeyError` to propagate? We could have a parsemodel Pydantic model for the lockfile since we need these fields for PURL generation.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,199 @@
...
+
+    name = f"{group_id}.{artifact_id}"
+    purl = PackageURL(type="maven", namespace=group_id, name=artifact_id, version=version)
+
+    return Component(
+        name=name,
+        purl=purl.to_string(),
+        version=version,
+        properties=[Property(name=PropertyEnum.PROP_MAVEN_SCOPE, value="compile")],
```
eskultety: what is `scope=compile`? I'm just curious, since we're using a magic string literal here. Let's start with what Maven scope is, I want a TIL :) .

README.md
```diff
@@ -222,6 +223,16 @@ some cases drastically so.
 
 See docs/gomod.md for more details.
 
+### [maven][]
+
+Hermeto supports Maven by reading `lockfile.json` in the project directory. To generate
+this file, you can use the [maven-lockfile][] plugn, which records resolved dependencies,
```
eskultety: `s/plugn/plugin`

eskultety (commented):
That's probaly as much as I can provide on this backend, Maven is very much out of my domain, so based on this backend's strong experimental nature, the rest LGTM.

hermeto/core/package_managers/maven/main.py
brunoapimentel: Here's the official docs:
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope

slimreaper35: Referenced the docs in the class.

hermeto/core/package_managers/maven/main.py
```diff
@@ -0,0 +1,197 @@
...
+            type="maven",
+            namespace=artifact.group_id,
+            name=artifact.artifact_id,
+            version=artifact.version,
+        )
+        component = Component(
+            name=artifact.artifact_id,
+            purl=purl.to_string(),
+            version=artifact.version,
+            properties=[Property(name=PropertyEnum.PROP_MAVEN_SCOPE, value=artifact.scope)],
```
taylormadore: Not blocking, but: Do we need to provide this metadata right now? Is anyone downstream looking at it? The reason I ask is because it is another hermeto custom property and CycloneDX appears to have something similar/related in the spec: https://github.com/CycloneDX/cyclonedx-property-taxonomy/blob/caf4fa1f0acd17d9d704c430ca602f2ed645c452/cdx/maven.md?plain=1#L19

Maybe we can defer this for now and make a decision later? It's always easier to add something new later rather than remove something that already exists.

slimreaper35: I believe this property was part of the very first vibe-coded design/PR. I am fine with dropping it.

> Is anyone downstream looking at it?

No idea. @brunoapimentel ?

brunoapimentel: I think this is a strong requirement, but since this is still an experimental phase, we can drop it for now. We just need to file an issue or something so we don't forget to get back to it.

slimreaper35: Dropped.

brunoapimentel (approved):

taylormadore (approved):
LGTM as experimental

slimreaper35:
just dropped old new lines from property_semantics.py for git blame purposes

PR #1519, opened 2026-04-15
yarn: enforce safe defaults for state paths to prevent validation bypass

notjivi:
# Description
While I was looking into how Hermeto handles Yarn dependencies, I noticed that the `_verify_yarnrc_paths` function uses `RootedPath` to block absolute paths (like `/tmp/something`). It works great for literal strings. However, it evaluates the raw string before Yarn actually runs.
If a user puts an environment variable into a path, like `${TMPDIR:-/tmp}/hermeto_state.gz`, the Python validator sees a string starting with `$`. It assumes this is just a weird relative folder name and lets it pass. But when the Yarn binary actually executes, it resolves that variable to the absolute path and writes the file to the host system.
 
## Impact Assessment
The practical impact of this bypass is significantly limited by several factors:
* In standard production deployments, Hermeto runs within a container. Any file written outside the workspace is still trapped within the container and wiped upon completion.
* Yarn writes these state files as binary GZIP data, so there's no risk of data exfiltration, no RCE, no privilege escalation, and the content isn't chosen by the attacker.
* It requires a clear, readable change to the `.yarnrc.yml` file, which is easily caught during standard PR code reviews or by any AI review bots.

## How to reproduce it locally
```bash
mkdir -p hermeto-yarn-repro && cd hermeto-yarn-repro
git init --bare /tmp/local-origin.git
echo '{"name": "sandbox-escape", "version": "1.0.0", "packageManager": "yarn@3.6.4"}' > package.json

# Enable corepack and generate a valid yarn.lock FIRST
corepack enable
yarn install

# Inject the malicious Yarn configuration with the environment variable payload
cat << 'EOF' > .yarnrc.yml
installStatePath: "${TMPDIR:-/tmp}/hermeto_state.gz"
EOF

git init
git config user.name "Hermeto Tester"
git config user.email "test@example.com"
git remote add origin file:///tmp/local-origin.git
git add .
git commit -m "chore: setup workspace with malicious yarnrc"
git push origin HEAD

# Ensure the host system's /tmp is clean of previous test files
rm -f /tmp/hermeto_state.gz

hermeto fetch-deps --source . --output ./repro_output yarn

# Verify the sandbox escape (File is written to /tmp)
ls -lh /tmp/hermeto_state.gz
```

## Expected behavior
Hermeto should either block the `${...}` syntax during validation, or explicitly override the path to yarn defaults so that `hermeto_state.gz` is written inside the sandbox only and is deleted upon pre-fetch completion.

## Actual behavior
The validator allows the path to pass, and the file is written outside the intended workspace directory within the container.

## Changes
* Overrode `installStatePath`, `pnpUnpluggedFolder`, and `pnpDataPath` to their official Yarn defaults based on yarn docs.
* Removed these from `_verify_yarnrc_paths` since they are now explicitly set by Hermeto.

### DIFF
--- a/hermeto/core/package_managers/yarn/main.py
+++ b/hermeto/core/package_managers/yarn/main.py
@@ -48,33 +48,6 @@ def fetch_yarn_source(request: Request) -> RequestOutput:
     )
 
 
-def _verify_yarnrc_paths(project: Project) -> None:
-    paths_conf_opts = {
-        # pnpDataPath is only configurable in Yarn v3
-        project.yarn_rc.get("pnpDataPath"): "pnpDataPath",
-        project.yarn_rc.get("pnpUnpluggedFolder"): "pnpUnpluggedFolder",
-        project.yarn_rc.get("installStatePath"): "installStatePath",
-        project.yarn_rc.get("patchFolder"): "patchFolder",
-        project.yarn_rc.get("virtualFolder"): "virtualFolder",
-    }
-
-    for path in paths_conf_opts:
-        if path is not None:
-            try:
-                project.source_dir.join_within_root(path)
-            except Exception:
-                raise PackageRejected(
-                    (
-                        f"YarnRC '{paths_conf_opts[path]}={path}' property: path points "
-                        "outside of the source directory"
-                    ),
-                    solution=(
-                        "Make sure that all Yarn RC configuration options specifying a path "
-                        "point to a relative location inside the main repository"
-                    ),
-                )
-
-
 def _check_zero_installs(project: Project) -> None:
     if project.is_zero_installs:
         raise PackageRejected(
@@ -96,7 +69,6 @@ def _check_lockfile(project: Project) -> None:
 
 
 def _verify_repository(project: Project) -> None:
-    _verify_yarnrc_paths(project)
     _check_zero_installs(project)
     _check_lockfile(project)
 
@@ -223,6 +195,13 @@ def _set_yarnrc_configuration(
     yarn_rc["enableScripts"] = False
     yarn_rc["enableGlobalCache"] = True
     yarn_rc["globalFolder"] = str(output_dir.join_within_root("deps", "yarn"))
+    yarn_rc["installStatePath"] = "./.yarn/install-state.gz"
+    yarn_rc["pnpUnpluggedFolder"] = "./.yarn/unplugged"
+    yarn_rc["patchFolder"] = "./.yarn/patches"
+    yarn_rc["virtualFolder"] = "./.yarn/__virtual__"
+    # pnpDataPath is only configurable in Yarn v3
+    if version in VersionsRange("3.0.0", "4.0.0-rc1"):
+        yarn_rc["pnpDataPath"] = "./.pnp.data.json"
 
     config = get_config()
     if (proxy_url := config.yarn.proxy_url) is not None:
--- a/tests/unit/package_managers/yarn/test_main.py
+++ b/tests/unit/package_managers/yarn/test_main.py
@@ -2,7 +2,6 @@
 import re
 from enum import Enum
 from itertools import chain, zip_longest
-from pathlib import Path
 from unittest import mock
 
 import pytest
@@ -25,7 +24,6 @@ from hermeto.core.package_managers.yarn.main import (
     _generate_environment_variables,
     _set_yarnrc_configuration,
     _verify_corepack_yarn_version,
-    _verify_yarnrc_paths,
     fetch_yarn_source,
 )
 from hermeto.core.package_managers.yarn.project import PackageJson, Plugin, YarnRc
@@ -335,28 +333,20 @@ def test_set_yarnrc_configuration(
         "unsafeHttpWhitelist": [],
         "pnpMode": "strict",
         "plugins": expected_plugins,
+        "installStatePath": "./.yarn/install-state.gz",
+        "pnpUnpluggedFolder": "./.yarn/unplugged",
+        "patchFolder": "./.yarn/patches",
+        "virtualFolder": "./.yarn/__virtual__",
     }
+
+    if yarn_version in VersionsRange("3.0.0", "4.0.0-rc1"):
+        expected_data["pnpDataPath"] = "./.pnp.data.json"
 
     if yarn_version in VersionsRange("4.0.0-rc1", "5.0.0"):
         expected_data["enableConstraintsChecks"] = False
 
     assert yarn_rc.data == expected_data
     mock_write.assert_called_once()
-
-
-@mock.patch("hermeto.core.package_managers.yarn.main.get_semver_from_package_manager")
-def test_verify_yarnrc_paths(
-    mock_get_semver: mock.Mock,
-    rooted_tmp_path: RootedPath,
-) -> None:
-    output_dir = rooted_tmp_path.join_within_root("output")
-    yarn_rc = YarnRc(rooted_tmp_path.join_within_root(".yarnrc.yml"), {})
-    project = mock.Mock()
-    project.yarn_rc = yarn_rc
-    project.package_json = mock.MagicMock()
-
-    _set_yarnrc_configuration(project, output_dir, semver.Version.parse("3.0.0"))
-    _verify_yarnrc_paths(project)
 
 
 def test_check_missing_lockfile(rooted_tmp_path: RootedPath) -> None:
@@ -366,29 +356,6 @@ def test_check_missing_lockfile(rooted_tmp_path: RootedPath) -> None:
 
     with pytest.raises(LockfileNotFound):
         _check_lockfile(project)
-
-
-@pytest.mark.parametrize(
-    "opt_path",
-    [
-        pytest.param("/custom/path", id="installStatePath"),
-        pytest.param("/custom/path", id="patchFolder"),
-        pytest.param("/custom/path", id="pnpDataPath"),
-        pytest.param("/custom/path", id="pnpUnpluggedFolder"),
-        pytest.param("/custom/path", id="virtualFolder"),
-    ],
-)
-def test_verify_yarnrc_paths_fail(
-    request: pytest.FixtureRequest, tmp_path: Path, opt_path: str
-) -> None:
-    project = mock.Mock()
-    project.source_dir = tmp_path
-    project.yarn_rc = YarnRc(
-        RootedPath(tmp_path / ".yarnrc.yml"), {request.node.callspec.id: opt_path}
-    )
-
-    with pytest.raises(PackageRejected):
-        _verify_yarnrc_paths(project)
 
 
 def test_generate_environment_variables(yarn_env_variables: list[EnvironmentVariable]) -> None:


eskultety:
> Following @eskultety's recommendation, this PR makes three changes:
> 

>     - Drops installStatePath from _verify_yarnrc_paths to remove the now-redundant validation.
>     - Explicitly sets yarn_rc["installStatePath"] = str(output_dir.join_within_root("install-state.gz")) in _set_yarnrc_configuration.
>     - Adds YARN_INSTALL_STATE_PATH to _generate_environment_variables to maintain consistency and ensure portability of the generated environment.


Well, since this went now public I'd like to distance myself from actually suggesting any of ^this :) . What I did in fact recommend in our private discussion was:

> Hermeto already overrides a bunch of config settings in yarnrc, so at best I'd say we could explicitly set installStatePath in the config to write to a Hermeto-controlled location that is going to be wiped anyway

which is a very different thing, because this install state is only something that Yarn does when running some commands, but it is something that is both irrelevant to users' builds' and incorrect for us to inject to their builds. It's something that should either go to `/dev/null` OR to the temporary working copy of the repo that will be wiped at the end of the fetch. Therefore, setting a build time variable is also something that users definitely are not interested in, in fact, it might even lead to problems.

notjivi:
My apologies. That is completely my fault for clumsily adding the bot's suggestion into the description under your name. It makes sense now that the state file shouldn't make its way into the users' actual build environment.

I will revert the `_generate_environment_variables` change right now, and I'll update the PR description to remove the misattribution. Pushing the fix up now

notjivi:
Force-pushed to improve the commit message body by adding the WHY for the git log. The code itself remains exactly the same as reverted yesterday.

hermeto/core/package_managers/yarn/main.py
```diff
@@ -223,6 +222,7 @@ def _set_yarnrc_configuration(
     yarn_rc["enableScripts"] = False
     yarn_rc["enableGlobalCache"] = True
     yarn_rc["globalFolder"] = str(output_dir.join_within_root("deps", "yarn"))
+    yarn_rc["installStatePath"] = str(output_dir.join_within_root("install-state.gz"))
```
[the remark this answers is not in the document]
eskultety: What a pile of rubbish!

hermeto/core/package_managers/yarn/main.py
```diff
@@ -53,7 +53,6 @@ def _verify_yarnrc_paths(project: Project) -> None:
         # pnpDataPath is only configurable in Yarn v3
         project.yarn_rc.get("pnpDataPath"): "pnpDataPath",
         project.yarn_rc.get("pnpUnpluggedFolder"): "pnpUnpluggedFolder",
-        project.yarn_rc.get("installStatePath"): "installStatePath",
```
eskultety, line 56: Don't the other paths in this check need the same treatment? This PR needs to consult the Yarn docs.

hermeto/core/package_managers/yarn/main.py
eskultety: Why `output_dir`? Why do we want to deliberately write the install state to the deps folder and risk user's Yarn offline build would read the install state coming from our environment? Why don't we force the path to its default location based on the official Yarn docs?

eskultety (changes requested):
This needs some more work and investigate if other paths beyond `installStatePath` might need addressing as well.

Furthermore, on the PR description:

> This PR addresses an edge case discussed in GHSA-vmxr-c8c2-m55c

^these are private, so they can't provide any context here, so please don't mention them.

>  However, because Yarn resolves environment variables natively after our validation, dynamic payloads could bypass this check.

Can you please expand on some details? Since both:
- we concluded in your original report the security impact of the problem at hand is extremely limited
- you just hinted at a potential security problem with this vague statement

there's really no reason NOT TO elaborate a bit on some of the original report's details.

hermeto/core/package_managers/yarn/main.py
notjivi: I researched into the yarn docs, found out and verified that along with `installStatePath`, `pnpDataPath` and `pnpUnpluggedFolder` also need the same treatment and to be routed to their defaults explicitly.
I left `patchFolder` and `virtualFolder` in as they are user configured.

hermeto/core/package_managers/yarn/main.py
notjivi: Makes perfect sense, routing this to `output_dir` was a bad call that risked polluting the user's offline build.

I've updated the code to force [installStatePath](https://yarnpkg.com/configuration/yarnrc#installStatePath), [pnpUnpluggedFolder](https://yarnpkg.com/configuration/yarnrc#pnpUnpluggedFolder) and [pnpDataPath](https://v3.yarnpkg.com/configuration/yarnrc#pnpDataPath) directly to their official default locations as per the Yarn docs.

notjivi:
Thank you for the review.
I expanded the description to explain the bypass mechanism and clarified how the impact is significantly limited, and I addressed the remaining feedback for the code in the latest commit.
Ready for another look

hermeto/core/package_managers/yarn/main.py
eskultety: > I left patchFolder and virtualFolder in as they are user configured.

I didn't get this part, can you clarify?

hermeto/core/package_managers/yarn/main.py
```diff
@@ -223,6 +219,10 @@ def _set_yarnrc_configuration(
     yarn_rc["enableScripts"] = False
     yarn_rc["enableGlobalCache"] = True
     yarn_rc["globalFolder"] = str(output_dir.join_within_root("deps", "yarn"))
+    yarn_rc["installStatePath"] = "./.yarn/install-state.gz"
+    yarn_rc["pnpUnpluggedFolder"] = "./.yarn/unplugged"
+    # pnpDataPath is only configurable in Yarn v3
+    yarn_rc["pnpDataPath"] = "./.pnp.data.json"
```
eskultety: What happens when one tries to do this on Yarn v4?

hermeto/core/package_managers/yarn/main.py
notjivi: The distinction is based on whether the path points to an internal state file or the actual project.

`installStatePath`, `pnpUnpluggedFolder`, and `pnpDataPath` are purely internal files used by Yarn. We can safely force these to their Yarn defaults because the final build doesn't care where these scratchpad files were during the fetch phase.

`patchFolder` is used by devs for bug-fix patches and are committed directly to Git. While many use the default folder, some projects choose custom locations for organization. In monorepos like [Neos UI](https://github.com/neos/neos-ui/blob/9.2/.yarnrc.yml), these paths are often part of a shared workspace. If Hermeto forcefully overrides this to the default, Yarn will look in the wrong place, find zero patches, and cause the fetch to use unpatched dependencies.

While manual customization of `virtualFolder` is really rare, it is still a part of the project's generated structure. I've kept it in `_verify_yarnrc_paths` for consistency with `patchFolder` and to avoid breaking any non-standard setups (it still is an edge case considering the rarity of that exact scenario).

To support these while hardening the sandbox, I plan to introduce a validator which will block `$` to ensure the same issue don't happen for these. I'll push this implementation shortly.

hermeto/core/package_managers/yarn/main.py
notjivi: Doing it on a v4 project will not be a problem as Yarn [silently ignores unsupported settings](https://github.com/yarnpkg/berry/blob/master/CHANGELOG.md#configuration) to maintain compatibility with different versions and plugins. Yarn will simply treat `pnpDataPath` as an unknown key and skip it. The override only has an effect on v3 projects.

hermeto/core/package_managers/yarn/main.py
eskultety: > patchFolder is used by devs for bug-fix patches and are committed directly to Git. While many use the default folder, some projects choose custom locations for organization. In monorepos like [Neos UI](https://github.com/neos/neos-ui/blob/9.2/.yarnrc.yml), these paths are often part of a shared workspace. If Hermeto forcefully overrides this to the default, Yarn will look in the wrong place, find zero patches, and cause the fetch to use unpatched dependencies.

Is it so? Have you actually looked at how Hermeto handles Yarn patches? There might be a misunderstanding in how that folder setting actually works and what it's for.

> While manual customization of virtualFolder

Same ^here, I believe there's some level of misunderstanding of what these are really for.

eskultety: > To support these while hardening the sandbox, I plan to introduce a validator which will block $ to ensure the same issue don't happen for these. I'll push this implementation shortly.

Please don't.

notjivi: I stepped back and looked into `locators.py` and `resolver.py` as you suggested, and the misunderstanding is clear now.
I see now that `_parse_patch_locator` extracts the patch file paths directly from the `lockfile` locator strings . Then, in `resolver.py`, `_get_path_patch_url` resolves those paths manually by joining the workspace `relpath` with the patch path . Because Hermeto performs its own resolution of these files independent of the `.yarnrc.yml` settings,  overriding `patchFolder` and `virtualFolder`, will not be a problem. I realize now that these are just caches for the Yarn subprocess and Hermeto’s audit logic relies on `.patch`.

My previous distinction between internal state and project structure and understand of those config was wrong.

I'm thinking of removing the validator and just explicitly override both `patchFolder` and `virtualFolder` to their Yarn defaults in `_set_yarnrc_configuration` along with the other state files. Does that sound like the right path forward?

eskultety: > I see now that _parse_patch_locator extracts the patch file paths directly from the lockfile locator strings . Then, in resolver.py, _get_path_patch_url resolves those paths manually by joining the workspace relpath with the patch path . Because Hermeto performs its own resolution of these files independent of the .yarnrc.yml settings, overriding patchFolder and virtualFolder, will not be a problem. I realize now that these are just caches for the Yarn subprocess and Hermeto’s audit logic relies on .patch.

I wouldn't consider ^this the correct reasoning or assessment of the problem, was this potentially assessed by AI using the wrong prompt? Anyway, to speed up the review process, let's not discuss ^this further.

> I'm thinking of removing the validator and just explicitly override both patchFolder and virtualFolder to their Yarn defaults in _set_yarnrc_configuration along with the other state files. Does that sound like the right path forward?

Yes.

hermeto/core/package_managers/yarn/main.py
eskultety: Can I see a proof? i.e. comparing debug logs (snippets!) from 2 independent runs on a simple (even artificial) project using both Yarn v3 and v4? The reason for that is that I ...

> Doing it on a v4 project will not be a problem as Yarn [silently ignores unsupported settings](https://github.com/yarnpkg/berry/blob/master/CHANGELOG.md#configuration) to maintain compatibility with different versions and plugins.

... read the CHANGELOG a bit differently - it is specifically written in context of older versions of Yarn, so I want to be sure that the logic applies in reverse as well, i.e. a setting that existed in an older version and cannot be explicitly configured in a newer Yarn version doesn't cause Yarn v4+ to complain.

notjivi: You were right to be cautious. I tested it and Yarn v4 it throws an error. My initial assumption about the changelog was incorrect.
I've made it conditional on v3 only, keeping it consistent with the existing comment. 
Pushing now along with the `patchFolder` and `virtualFolder` overrides.

image

hermeto/core/package_managers/yarn/main.py
notjivi: Done

Issue #937, opened 2025-05-06
[yarn v1] Make the backend expectations on user input more permissive


eskultety:
Investigate which current error paths inside this backend module can be made more permissive, hence relaxing our requirements.

**Depends on: https://github.com/hermetoproject/hermeto/issues/934**

Harsh-2005d:
Hey, I've been looking into the issue and working on making the missing lockfile check permissive, similar to how it is done for Cargo.

In permissive mode, instead of failing when `yarn.lock` is absent, we warn and continue. I've verified locally on yarn 4.13.0 that `yarn install --mode skip-build` automatically generates `yarn.lock` if it's missing.

However, I noticed that `_set_yarnrc_configuration`
```python 
def _set_yarnrc_configuration(
    project: Project, output_dir: RootedPath, version: semver.Version
) -> None:
    ...
    yarn_rc["enableImmutableInstalls"] = True
    ...
```
sets `enableImmutableInstalls = True` in `.yarnrc.yml` before `_fetch_dependencies` runs, which would cause yarn to abort if the lockfile needs to be modified.

My approach: in permissive mode, skip setting `enableImmutableInstalls = True` (or set it to `False`) so that yarn install can freely generate the lockfile when absent. The rest would follow the cargo pattern warn instead of hard fail, and add a permissive mode SBOM annotation when the lockfile was generated.

Is this the right approach?

eskultety:
Closing this parent issue (milestone remains), it's not doing us any good. Originally it were supposed to track all permissive cases for the particular backend which haven't identified yet. From now on, we'll only create issues for immediate permissive mode candidates.

PR #1522, opened 2026-04-17
feat(yarn-classic): implement permissive mode for missing lockfiles 

Harsh-2005d:
Based on #937 

---
### Description
This PR implements **Permissive Mode** support for Yarn v1 (`yarn_classic`) projects. Previously, a missing `yarn.lock` file resulted in a `LockfileNotFound` error. Under the new logic, when `--mode permissive` is toggled, Hermeto will warn the user and generate a lockfile to proceed with dependency resolution.

### Key Architectural Changes:
The changes are inspired from the current implementation of missing lockfile in cargo.
1.  Merged lockfile existence and version checks into a single `_verify_lockfile` function. 
2.  The verification result (`lockfile_generated: bool`) is explicitly passed through functions.
3.  Updated `_fetch_dependencies` to toggle the Yarn `--frozen-lockfile` flag. If a lockfile was missing and permissive mode is active.
4.  Integrated with `hermeto.core.models.sbom` to annotate the generated SBOM. Components resolved via a generated lockfile are now tagged with `hermeto:permissive-mode:yarn-classic:generated-lockfile`.
5. Updated unit test mock assertions to the updated function signatures. 

### Verification Results
* **Strict Mode:** Verified that Hermeto correctly raises `LockfileNotFound` when `yarn.lock` is missing.
* **Permissive Mode:** Verified that Hermeto logs a warning, resolves dependencies successfully and includes the correct subjects in the SBOM annotations.
*  All existing tests passed.

### DIFF
--- a/hermeto/core/package_managers/yarn_classic/main.py
+++ b/hermeto/core/package_managers/yarn_classic/main.py
@@ -7,10 +7,10 @@ from typing import Any
 
 from hermeto import APP_NAME
 from hermeto.core.errors import LockfileNotFound, PackageManagerError, PackageRejected
-from hermeto.core.models.input import Request
-from hermeto.core.models.output import Component, EnvironmentVariable, RequestOutput
+from hermeto.core.models.input import Mode, Request
+from hermeto.core.models.output import Annotation, Component, EnvironmentVariable, RequestOutput
 from hermeto.core.models.property_semantics import PropertySet
-from hermeto.core.models.sbom import create_backend_annotation
+from hermeto.core.models.sbom import create_backend_annotation, spdx_now
 from hermeto.core.package_managers.yarn.utils import (
     VersionsRange,
     extract_yarn_version_from_env,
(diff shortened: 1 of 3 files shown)


eskultety:
I'm not convinced generating missing lockfiles is something we want to get into, because I can already tell it won't stop at Yarn v1 and it will cause an avalanche of PRs just like this one. I am not dismissing the possibility of having to do this in the future, we're mainly driven by need. I believe Yarn (any version) is an established ecosystem with fairly well documented patterns and good practices. I think this PR lacks the understanding of the background that led to the changes we have made to the Cargo backend - we didn't do it because of native Rust projects, we had to do it because we found transitive Python Rust extention dependencies that made the primary projects unbuildable with no control over the extensions.
We haven't seen such cases with Yarn and I definitely would not like to make this a precedent because Hermeto simply does NOT benefit from this permission mode in any way, we only ever add stuff when we have to.
I would maybe accept this change if it only took care of the stale lockfile case, not a missing one, then again, we added Yarn v1 support fairly reluctantly because of some downstream needs, but we actually want to get rid of it as soon as possible.

I'd like to see good evidence of existence of reputable (i.e. popular) projects that do not have the lockfile committed that we could assess as a team and change my/our opinion, but right now I'm not convinced we'd want to take the change in.

slimreaper35:
> I'm not convinced generating missing lockfiles is something we want to get into, because I can already tell it won't stop at Yarn v1 and it will cause an avalanche of PRs just like this one.

+1

Harsh-2005d:
@eskultety Thanks for the context behind the cargo implementation, I was not aware of it. I will close this PR.

Meanwhile I will look into whether there are real use cases for stale lockfiles.

PR #1358, opened 2026-03-05
gomod: convert module to subpackage

harshitbansal184507:
Closes #1157 
Commits- 
Commit 1 - convert to subpackage
Commit 2- extract go.py + test_go.py
Commit 3 -extract utils.py + test_utils.py

Sorry for reopening this PR. The previous one was closed automatically because I was pushing through the main branch and ran into issues maintaining the branch locally. I've since switched to a dedicated feature branch (gomod-split) to keep things clean and avoid similar issues going forward.

PR #1526, opened 2026-04-20
refactor(tests): consolidate and remove redundant unit tests

### LINKED ISSUE #1506: Revisit unit tests for redundancy and consolidation
We should revisit unit tests to reduce redundancy, improve clarity, and speed up the suite.

  - near-duplicate tests that can be consolidated into parametrized tests
  - tests that mock everything and assert nothing meaningful                                                                                                                 
  - test files with excessive test counts that could be simplified
  - outdated tests for removed or refactored features                                                                                                                        
   
  Current unit test counts                                                                                                                                                   
                  
  - pip: 348                                                                                                                                                                 
  - yarn: 185     
  - gomod: 180
  - models: 131
  - yarn_classic: 109
  - npm: 105                                                                                                                                                                 
  - cli: 99
  - rpm: 45
  - checksum: 27
  - bundler: 24
  - other: 152

Total: 1405  at the time of creating this issue.

derasdf:
Remove redundant unit tests.

- Parametrize copy-pasted test functions into single tests with explicit IDs
- Simplify over-mocked test setups to minimum needed data
- Delete tests that duplicate other unit tests
- Delete 21 over-mocked wiring tests whose scenarios are covered by integration tests

Each deleted test was verified against the integration test that replaces it.
Guard tests, security tests, and logic tests were kept.

Let me know if there are more tests that I (and AI) have missed.

Closes https://github.com/hermetoproject/hermeto/issues/1506


slimreaper35:
I think we might want to get rid of the 90% threshold for unit test coverage:
https://github.com/hermetoproject/hermeto/blob/main/pyproject.toml#L133

The CI is green, but there is a "warning" that says `FAIL Required test coverage of 90.0% not reached. Total coverage: 89.99%`. It has been slowly dropping anyway, so it was just a matter of time before someone would encounter it.

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -158,22 +152,19 @@ def test_prepare_for_hermetic_build_injects_necessary_variable_into_existing_con
...
-        BUNDLER_NONEXISTENT_VARIABLE: "true"
-        """
-    )
 
     assert not expected_config_location.exists(), "Unexpected .bundle/config in rooted_tmp_path"
-    assert not expected_config_location.parent.exists(), "Unexpected .bundle/ in rooted_tmp_path"
 
-    expected_config_location.parent.mkdir()
-    expected_config_location.write_text(existing_preamble)
+    if existing_preamble is not None:
```
a-ovchinnikov: This is precisely the antipattern the previous construct was dealing with: branching in tests. Tests are **not** regular code, they just happen to look like regular code. A certain level of repetition and copy and paste should be acceptable by design: the point of a test is to be sufficiently self-contained and compact to clearly describe a scenario that is being tested. A unit test is supposed to have "Arrange-Act-Assert" structure (a.k.a. "Given-When-Then"), observe the lack of "If" or "Or Else" in any of those mnemonics, they are not there for a reason. Overly abstract and tightly packed tests defeat the purpose of a unit test. A test should not be written in such a way that one might need to go and debug it later. Note also that tests (unlike code) are not supposed to be read and modified often, but when read they should be as clear as possible. Linear code is the clearest of them all.

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -401,20 +399,17 @@ def test_get_gomod_version(
...
-    "toolchain 1.21",  # missing 'go' prefix for the toolchain spec
-]
-
-
 @pytest.mark.parametrize(
     "go_mod_file",
-    [pytest.param(_, id=_) for _ in INVALID_VERSION_STRINGS],
+    [
+        pytest.param("go1.21", id="missing_space"),
+        pytest.param("go 1.21.0.100", id="non_conforming"),
```
a-ovchinnikov: "non_conforming" carries less information than "non-conforming to the X.Y(.Z)? versioning template", I suggest either keeping everything, or keeping the comments, or even not changing this bit at all.

tests/unit/package_managers/gomod/test_main.py
```diff
@@ -1138,15 +1138,11 @@ def test_go_list_deps(mock_run_cmd: mock.Mock, pattern: Literal["all", "./..."])
...
 @mock.patch("hermeto.core.package_managers.gomod.go.run_cmd")
-def test_go_list_deps_fail(
-    mock_run_cmd: mock.Mock,
-) -> None:
+def test_go_list_deps_fail(mock_run_cmd: mock.Mock) -> None:
     mock_run_cmd.side_effect = subprocess.CalledProcessError(1, cmd="foo")
-    expect_error = "Go execution failed: `go list -e -m -json` failed"
 
-    with pytest.raises(PackageManagerError) as ex:
+    with pytest.raises(PackageManagerError, match="Go execution failed:"):
```
a-ovchinnikov: Please do not introduce additional matches to tests unless you absolutely must, this makes them brittle.

tests/unit/package_managers/npm/test_resolver.py
```diff
@@ -1081,9 +1081,8 @@ def test_npm_proxy_credentials_propagate_to_registry_hosts(
 
     _get_npm_dependencies(rooted_tmp_path, deps_to_download)
 
-    msg = "Not found credentials where they should be!"
     for call in mock_async_download_files.mock_calls:
-        assert call.kwargs["auth"] is not None, msg
+        assert call.kwargs["auth"] is not None, "Not found credentials where they should be!"
```
a-ovchinnikov: Why?

tests/unit/package_managers/pip/test_main.py
```diff
@@ -92,67 +92,32 @@ def mock_requirements_file(requirements: list | None = None, options: list | Non
...
-    assert version == "0.1.0"
-
-    # check logs
-    assert "Checking setup.cfg for metadata" in caplog.messages
-    assert f"Resolved name {name} for package at {rooted_tmp_path}" in caplog.messages
-    assert f"Resolved version {version} for package at {rooted_tmp_path}" in caplog.messages
+        name, version = pip._get_pip_metadata(rooted_tmp_path)
+        assert name == "foo"
+        assert version == "0.1.0"
+        assert log_message in caplog.messages
```
a-ovchinnikov: I strongly recommend getting rid of all log checks while we are here. Having these checks is a bad idea for several reasons: first, such strings are volatile and thus create unnecessary binding between tests and code (you change a string, now you have to also change tests _despite the fact that there were no changes to logic_), second, log messages are normally _not_ a part of contract for a function and are just a helpful side-effect, UTs are supposed to be testing just the contract, third, as long as a log line is present in the code we are sure that it will appear in some logger namespace at some point solely because we rely on interpreter being sufficiently tested elsewhere, in other words as its users we must not test our language stdlib.

tests/unit/package_managers/pip/test_main.py
```diff
@@ -365,133 +330,277 @@ def test_ignored_and_rejected_options(self, caplog: pytest.LogCaptureFixture) ->
...
+                id="vcs_short_ref",
+            ),
+            pytest.param(
+                {
+                    "package": "eggs",
+                    "kind": "vcs",
+                    "url": f"git+https://github.com@{GIT_REF}/spam/eggs",
+                    "download_line": f"eggs @ git+https://github.com@{GIT_REF}/spam/eggs",
+                },
+                PackageRejected,
```
a-ovchinnikov: If there is a need to differentiate between different `PackageRejected` types then `PackageRejected` must be subclassed.

tests/unit/package_managers/pip/test_main.py
```diff
@@ -365,133 +330,277 @@ def test_ignored_and_rejected_options(self, caplog: pytest.LogCaptureFixture) ->
...
+                {
+                    "package": "foo",
+                    "kind": "url",
+                    "url": "https://example.org/file.wheel",
+                    "download_line": "foo @ https://example.org/file.wheel",
+                },
+                PackageRejected,
+                "URL for requirement does not contain any recognized file extension:",
+                id="url_wheel_extension",
+            ),
```
a-ovchinnikov, line 462: With a table approach that I have been promoting lately this entry could have occupied just three lines instead of eleven and would have also let one see the entire test case with all necessary data at a glance:
```
|url_wheel_extension|{"package": "foo", "kind": "url",                       | PackageRejected |
|                   | "url": "https://example.org/file.wheel",               |                 |
|                   | "download_line": "foo @ https://example.org/file.wheel}|                 |
```

tests/unit/package_managers/pip/test_main.py
```diff
@@ -969,34 +1078,33 @@ def test_resolve_pip_no_deps(mock_metadata: mock.Mock, rooted_tmp_path: RootedPa
...
     invalid_path = Path("foo/bar.txt")
-    build_requirement_files = [invalid_path]
+    output_dir = rooted_tmp_path.join_within_root("output")
     with pytest.raises(LockfileNotFound):
-        pip._resolve_pip(
-            package_path=rooted_tmp_path,
-            output_dir=rooted_tmp_path.join_within_root("output"),
-            build_requirement_files=build_requirement_files,
-        )
+        if file_kwarg == "requirement_files":
```
a-ovchinnikov: This test is more complex than the previous pair: it uses parametrization for a single parameter which can take just two values, it has branching. I would suggest keeping it as is.

tests/unit/package_managers/pip/test_main.py
```diff
@@ -1118,45 +1226,60 @@ def test_get_external_requirement_filepath(component_kind: str, url: str) -> Non
...
+        pytest.param("myapp-0.1.zip", id="zip"),
     ],
 )
-def test_check_metadata_from_sdist(sdist_filename: str, data_dir: Path) -> None:
+def test_check_metadata_from_sdist(
+    sdist_filename: str,
+    data_dir: Path,
+) -> None:
     sdist_path = data_dir / "archives" / sdist_filename
     pip._check_metadata_in_sdist(sdist_path)
```
a-ovchinnikov: I would be very suspicious of any test that does not have an explicit assertion. What happens here?

tests/unit/package_managers/pip/test_main.py
```diff
@@ -1118,45 +1226,60 @@ def test_get_external_requirement_filepath(component_kind: str, url: str) -> Non
...
 def test_metadata_check_fails_from_sdist(
-    sdist_filename: Path, expected_error: str, data_dir: Path
+    sdist_filename: str,
+    exc_type: type[Exception],
+    expected_error: str,
+    data_dir: Path,
 ) -> None:
-    sdist_path = data_dir / "archives" / sdist_filename
-    with pytest.raises(PackageRejected, match=expected_error):
+    if sdist_filename == "myapp-0.2.tar.ZZZ":
```
a-ovchinnikov: Putting aside the fact that UTs are not supposed to be touching file system (that's a bigger issue which should be resolved separately), why not just `touch data_dir/archives/myapp-0.2.tar.ZZZ` and save some branching? (Or even just construct the path object, it should be rejected before even reaching to file system anyway.)

tests/unit/package_managers/yarn_classic/test_resolver.py
```diff
@@ -470,15 +460,18 @@ def test_successful_name_extraction(rooted_tmp_path: RootedPath) -> None:
...
-    tarball_path = mock_tarball(path=rooted_tmp_path, package_json_content={"key": "foo"})
-    with pytest.raises(ValueError, match="No 'name' field found"):
+@pytest.mark.parametrize(
+    "content, error_match",
+    [
+        pytest.param({}, "No package.json found", id="no_package_json"),
+        pytest.param({"key": "foo"}, "No 'name' field found", id="missing_name_field"),
+    ],
+)
+def test_read_name_from_tarball_missing_or_invalid(
```
a-ovchinnikov: I think that test names must briefly describe what is being tested. Consider
`test_read_name_from_tarball_missing_or_invalid[missing_name_field]` name. Can you tell at a glance what is being tested? What the author planned to verify with this test? Probably not, at least I cannot. Now consider `test_read_name_from_tarball_rejects_broken_tarballs[missing_name_field]` -- how does it fare? Does it convey the same information, or more, or less? I think this is a great opportunity to take a look at names too and adjust them when it makes sense. 

(For the sake of completeness:
```
>>> len("test_read_name_from_tarball_rejects_broken_tarballs[missing_name_field]    FAIL")
79
```
it is not that big even with the proposed change.)

tests/unit/test_errors.py
```diff
@@ -1,23 +1,58 @@
...
+                  The error might be intermittent, please try again.
+                  If the issue seems to be on the {APP_NAME} side, please contact the maintainers.
+                """
+            ).strip(),
+            id="fetch_error",
+        ),
+    ],
+)
+def test_error_friendly_msg(error: errors.BaseError, expect_msg: str) -> None:
+    assert error.friendly_msg() == expect_msg
```
a-ovchinnikov: This entire block should be dropped unless there are reasons to believe that CPython does not behave as documented (but then again this is the wrong place to verify that).

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -204,28 +204,6 @@ def test_prepare_for_hermetic_build_injects_necessary_variable_into_existing_alt
     assert result.template == existing_preamble + expected_alternate_config_contents
 
 
-def test_prepare_for_hermetic_build_ignores_a_directory_in_place_of_config(
```
a-ovchinnikov, line 216: How did you determine that this test is redundant?

tests/unit/package_managers/bundler/test_main.py
a-ovchinnikov, line 1: https://github.com/hermetoproject/hermeto/commit/4de41f485c9a27c321e29520e898dc68b4da7ced is huge! It touches multiple PMs and sometimes justification for a change seems missing. It would be nice if it could be split into multiple commits per PM with justification elaborated in commit messages.

a-ovchinnikov (commented):
This is a one solid change, long overdue. Please find some initial observations inline

tests/unit/models/test_sbom.py
```diff
@@ -33,29 +33,43 @@
...
+                "1 validation error for SPDXPackage\nname\n  Field required",
+                id="spdx_package_missing_name",
             ),
         ],
     )
-    def test_invalid_components(self, input_data: dict[str, str], expect_error: str) -> None:
+    def test_invalid_components(
+        self, model_cls: type, input_data: dict[str, str], expect_error: str
+    ) -> None:
         with pytest.raises(pydantic.ValidationError, match=expect_error):
```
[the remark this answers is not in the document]
derasdf: Fixing this is a separate effort across all pydantic validation tests.

tests/unit/package_managers/pip/test_main.py
derasdf: Hmm fair, but i'd say first we need to see what we think about this as a team on your PR #1528

tests/unit/package_managers/pip/test_main.py
derasdf: It's testing that the test data files in data_dir/archives/ exist and are valid, could be dropped i guess.

tests/unit/package_managers/pip/test_main.py
derasdf: Hm, what about suggestion of gemini ? Or just reverting is better ?

tests/unit/package_managers/pip/test_main.py
derasdf: No existing subclasses for unpinned/vcs-ref/extension errors. Should I add them in this PR or later?

tests/unit/package_managers/bundler/test_main.py
derasdf: I think you tagged wrong commit here (should be last one and not one before it, or did you really mean this commit you tagged ? because I feel like last one is way worse), will do, for now pushed all other changes except this one.
Also, how granular do you want the split? One commit per PM, or is grouping the small ones (bundler+rpm+npm) fine?

tests/unit/package_managers/npm/test_resolver.py
```diff
@@ -1081,9 +1081,8 @@ def test_npm_proxy_credentials_propagate_to_registry_hosts(
 
     _get_npm_dependencies(rooted_tmp_path, deps_to_download)
 
-    msg = "Not found credentials where they should be!"
     for call in mock_async_download_files.mock_calls:
-        assert call.kwargs["auth"] is not None, msg
+        assert call.kwargs["auth"] is not None
```
a-ovchinnikov: Why do you drop the message? A message makes an assertion more specific.

tests/unit/models/test_sbom.py
```diff
@@ -65,29 +65,43 @@ def _sbom_for_cyclonedx_spdx_roundtrip(components: list[dict[str, object]]) -> S
...
+            pytest.param(
+                SPDXPackage,
+                {"SPDXID": "Not-an-id", "versionInfo": "some-version"},
+                "1 validation error for SPDXPackage\nname\n  Field required",
+                id="spdx_package_missing_name",
             ),
         ],
     )
-    def test_invalid_components(self, input_data: dict[str, str], expect_error: str) -> None:
+    def test_invalid_components(
```
a-ovchinnikov: WDYT about s/test_invalid_components/test_invalid_components_dont_pass_validation/?

tests/unit/models/test_sbom.py
```diff
@@ -928,47 +892,84 @@ def test_sort_and_dedupe_packages(self, mock_spdx_now: str) -> None:
...
+            ),
+            pytest.param(
+                "OTHER",
+                "anythingcangohere",
+                "some-id",
+                id="unsupported_other_category",
+            ),
+        ],
+    )
+    def test_package_external_ref_invalid_reference(
```
a-ovchinnikov: WDYT about s/test_package_external_ref_invalid_reference/test_package_with_invalid_reference_in_external_ref_rejected?

tests/unit/package_managers/pip/test_main.py
a-ovchinnikov: Since this is a tests cleanup effort I'd say they should be added in this PR (as a separate commit).

tests/unit/package_managers/pip/test_main.py
a-ovchinnikov: Correct, I am raising awareness of the approach.

tests/unit/package_managers/rpm/test_main.py
```diff
@@ -149,166 +149,109 @@ def test_resolve_rpm_project_no_lockfile(rooted_tmp_path: RootedPath) -> None:
...
+            "lockfileVendor: redhat\nlockfileVersion: 1\narches\n",
+            id="missing_colon",
+        ),
+        pytest.param(
+            "lockfileVendor: redhat lockfileVersion: 1\narches:\n",
+            id="missing_newline",
+        ),
+    ],
+)
+def test_resolve_rpm_project_invalid_yaml_format(
```
a-ovchinnikov: s/test_resolve_rpm_project_invalid_yaml_format/test_resolve_rpm_project_rejects_invalid_yaml_format/?

tests/unit/package_managers/rpm/test_main.py
```diff
@@ -149,166 +149,109 @@ def test_resolve_rpm_project_no_lockfile(rooted_tmp_path: RootedPath) -> None:
...
-            f,
-        )
-    with pytest.raises(InvalidLockfileFormat) as exc_info:
-        _resolve_rpm_project(rooted_tmp_path, rooted_tmp_path)
-
+            id="empty_packages",
+        ),
+    ],
+)
+def test_resolve_rpm_project_arch_empty(rooted_tmp_path: RootedPath, lockfile_data: dict) -> None:
```
a-ovchinnikov: This test and the next one does not look too different from the ones you have already merged, would it make sense to merge them too?

tests/unit/package_managers/yarn/test_project.py
```diff
@@ -143,14 +143,16 @@ def _prepare_package_json_file(rooted_tmp_path: RootedPath, data: str) -> Packag
...
-    package_json = _prepare_package_json_file(rooted_tmp_path, EMPTY_JSON_FILE)
-    assert package_json.get("packageManager") is None
+@pytest.mark.parametrize(
+    "data, expected",
+    [
+        pytest.param(VALID_PACKAGE_JSON_FILE, "yarn@3.6.1", id="valid"),
+        pytest.param(EMPTY_JSON_FILE, None, id="empty"),
+    ],
+)
+def test_parse_package_json(rooted_tmp_path: RootedPath, data: str, expected: str | None) -> None:
```
a-ovchinnikov: Here the test does not check the entire parser but rather the fact that a package manager is correctly determined.

tests/unit/package_managers/npm/test_resolver.py
derasdf: Hm before it was AI thing, now I thought dropping the string messages would be good since string matching is bad but now the file changes don't make sense with the PR, so sure can drop.

a-ovchinnikov: This is a different mechanism, `assert expression,"message"` checks if expression evaluates to `True`, if not then it prints out `message`. What I regularly suggest to reconsider is having `match` clause in `pytest.raises(foo, match="some regexp")` and making any _assertions_ on strings (i.e. when a string is a part of an expression). Message is there for a user friendlier tests output: when there is no message a user sees something along the lines of
```python
>>> assert foo ==  bar
  File "<python-input-2>", line 1, in <module>
    assert foo ==  bar
           ^^^^^^^^^^^
AssertionError
```
with a message however:
```python
>>> assert foo ==  bar, "Secure keys mismatch"
  File "<python-input-3>", line 1, in <module>
    assert foo ==  bar, "Secure keys mismatch"
           ^^^^^^^^^^^
AssertionError: Secure keys mismatch
```
) That's what I meant by making assertion more specific.

tests/unit/package_managers/bundler/test_main.py
derasdf: It's almost identical to `test_prepare_for_hermetic_build_injects_necessary_variable_into_empty_config`. The only difference is an extra precondition assert on .parent.exists() which is always true in a fresh tmp dir anyway. The test name suggests it should create .bundle/config as a directory to test is_file() returning False on a directory, but it never does.

tests/unit/package_managers/bundler/test_main.py
derasdf: Still not sure what and how this should be done.

tests/unit/models/test_sbom.py
```diff
@@ -65,29 +65,43 @@ def _sbom_for_cyclonedx_spdx_roundtrip(components: list[dict[str, object]]) -> S
...
+            pytest.param(
+                SPDXPackage,
+                {"SPDXID": "Not-an-id", "versionInfo": "some-version"},
+                "1 validation error for SPDXPackage\nname\n  Field required",
+                id="spdx_package_missing_name",
             ),
         ],
     )
-    def test_invalid_components(self, input_data: dict[str, str], expect_error: str) -> None:
+    def test_invalid_components_dont_pass_validation(
```
a-ovchinnikov, line 88: I am still questioning the need for this test a bit. On one hand we might want to document some examples of broken input that we want to reject, on the other hand this test now effectively verifies pydantic and not our code -- we check that pydantic models raise exceptions and also verify messages which we do not control. The former can pass as a form of a smoke test for how we determined the models, but do we need the latter? In other words, WDYT about dropping the match here?

tests/unit/package_managers/yarn_classic/test_resolver.py
```diff
@@ -454,15 +444,18 @@ def test_successful_name_extraction(rooted_tmp_path: RootedPath) -> None:
...
+    [
+        pytest.param({}, "No package.json found", id="no_package_json"),
+        pytest.param({"key": "foo"}, "No 'name' field found", id="missing_name_field"),
+    ],
+)
+def test_read_name_from_tarball_rejects_broken_tarballs(
+    rooted_tmp_path: RootedPath, content: dict, error_match: str
+) -> None:
+    tarball_path = mock_tarball(path=rooted_tmp_path, package_json_content=content)
+    with pytest.raises(ValueError, match=error_match):
```
a-ovchinnikov, line 406: Do these messages we match on come from our code or from Python's standard library? If it is the latter then let's drop this check now.

tests/unit/package_managers/bundler/test_main.py
a-ovchinnikov: What I suggest is: to take large commits which touch multiple package managers and split them in smaller commits each touching just one package manager. Such smaller commits could have a bit more detailed commit messages about why certain changes were made to ensure future readers understand the justification behind changes. I won't be insisting too hard on it, but it makes reviewing the changes much easier.

tests/unit/models/test_sbom.py
derasdf: Sure i'll drop it here.

tests/unit/package_managers/yarn_classic/test_resolver.py
derasdf: These come from our code [yarn_classic/resolver.py#L368](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/yarn_classic/resolver.py#L368), [yarn_classic/resolver.py#L377](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/yarn_classic/resolver.py#L377), so keeping them.

hermeto/core/errors.py
```diff
@@ -171,6 +171,18 @@ def __init__(self, reason: str, *, solution: str | None) -> None:
...
+
+
+class InvalidVCSReference(PackageRejected):
+    """A VCS requirement does not have a valid reference."""
+
+
+class UnrecognizedFileExtension(PackageRejected):
+    """A URL requirement has an unrecognized file extension."""
+
+
```
slimreaper35, line 191: Should we add these exceptions to the exit code error [enum](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/errors.py#L13) ?

eskultety: At least the first 2 are indeed generic enough that would justify a dedicated enum, not convinced about the last one, seems pretty niche.

derasdf: Added

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -154,7 +129,11 @@ def test_retry(
     @mock.patch("hermeto.core.package_managers.gomod.go.run_cmd")
     @mock.patch("time.sleep")
     def test_retry_failure(
-        self, mock_sleep: Any, mock_run: Any, mock_config: Any, caplog: pytest.LogCaptureFixture
+        self,
+        mock_sleep: mock.Mock,
+        mock_run: mock.Mock,
+        mock_config: mock.Mock,
+        caplog: pytest.LogCaptureFixture,
```
eskultety: Did ruff do this or the AI agent? IOW if ruff complained about line being too long we would have fixed it already, so this is an unrelated change in the already large diff, please drop the hunk.

tests/unit/package_managers/bundler/test_main.py
eskultety: Agreed, it would really help.

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -28,52 +28,52 @@ def write_cargo_lock(rooted_path: RootedPath, content: str) -> None:
...
+    rooted_tmp_path: RootedPath,
+    cargo_toml: str,
+    expected_name: str | None,
+    expected_version: str | None,
+) -> None:
+    write_cargo_toml(rooted_tmp_path, cargo_toml)
     name, version = _resolve_main_package(rooted_tmp_path)
-    assert name == expected_name
-    assert version is None
+    assert name == (expected_name or rooted_tmp_path.path.name)
```
eskultety, line 75: nitpick: I would only expect `assert name == expected_name` here as a reader with the conditional assignment happening at the beginning of the function. Matter of taste I guess.

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -401,20 +321,24 @@ def test_get_gomod_version(
...
+        pytest.param("1.21", id="missing_go_prefix"),
+        # extra characters after version string
+        pytest.param("go 1.21 foo", id="extra_characters"),
+        # pre-release with no number
+        pytest.param("go 1.21prerelease", id="prerelease_no_number"),
+        # pre-release with non-alphanum character
+        pytest.param("go 1.21prerelease_4", id="prerelease_non_alphanum"),
+        # missing 'go' prefix for the toolchain spec
+        pytest.param("toolchain 1.21", id="toolchain_missing_go_prefix"),
+    ],
```
eskultety, line 331: So, instead of translating the original comments to pytest IDs, we now have both commentaries and IDs. Descriptive IDs should be enough. I realize this comment is going against an earlier suggestion during review, because I think we can do better.

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -401,20 +321,24 @@ def test_get_gomod_version(
...
-    "toolchain 1.21",  # missing 'go' prefix for the toolchain spec
-]
-
-
 @pytest.mark.parametrize(
     "go_mod_file",
-    [pytest.param(_, id=_) for _ in INVALID_VERSION_STRINGS],
+    [
+        # missing space between go and version number
+        pytest.param("go1.21", id="missing_space"),
```
eskultety: `id="no_space_after_go_keyword"`

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -401,20 +321,24 @@ def test_get_gomod_version(
...
+        # non-conforming to the X.Y(.Z)? versioning template
+        pytest.param("go 1.21.0.100", id="non_conforming"),
+        # missing 'go' at the beginning
+        pytest.param("1.21", id="missing_go_prefix"),
```
eskultety: `id="missing_go_keyword"`

tests/unit/package_managers/gomod/test_go.py
eskultety: How about `id="invalid_version_scheme"`?

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -401,20 +321,24 @@ def test_get_gomod_version(
...
+        # extra characters after version string
+        pytest.param("go 1.21 foo", id="extra_characters"),
```
eskultety: `id="extra_characters_in_version_str"`

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -401,20 +321,24 @@ def test_get_gomod_version(
...
+        # pre-release with no number
+        pytest.param("go 1.21prerelease", id="prerelease_no_number"),
+        # pre-release with non-alphanum character
+        pytest.param("go 1.21prerelease_4", id="prerelease_non_alphanum"),
+        # missing 'go' prefix for the toolchain spec
+        pytest.param("toolchain 1.21", id="toolchain_missing_go_prefix"),
```
eskultety: `id="missing_go_in_toolchain_spec"`

tests/unit/package_managers/gomod/test_main.py
```diff
@@ -1138,15 +830,11 @@ def test_go_list_deps(mock_run_cmd: mock.Mock, pattern: Literal["all", "./..."])
...
-    mock_run_cmd: mock.Mock,
-) -> None:
+def test_go_list_deps_fail(mock_run_cmd: mock.Mock) -> None:
     mock_run_cmd.side_effect = subprocess.CalledProcessError(1, cmd="foo")
-    expect_error = "Go execution failed: `go list -e -m -json` failed"
 
-    with pytest.raises(PackageManagerError) as ex:
+    with pytest.raises(PackageManagerError):
         _go_list_deps(Go(), "./...", {})
-        assert expect_error in str(ex)
```
eskultety, line 1149: This isn't consolidation, this is adhering to our "no string matching in tests rule" --> standalone commit.

tests/unit/models/test_sbom.py
```diff
@@ -65,29 +65,31 @@ def _sbom_for_cyclonedx_spdx_roundtrip(components: list[dict[str, object]]) -> S
 
 class TestComponent:
     @pytest.mark.parametrize(
-        "input_data, expect_error",
```
eskultety, line 68: **Commit message:**
 _Parametrize duplicate test cases_

There's a lot more going on this commit and it'd be best to split into more digestible chunks for review purposes.

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -213,28 +213,6 @@ def test_prepare_for_hermetic_build_injects_necessary_variable_into_existing_alt
...
```
eskultety, line 216: **Commit message:**
_- duplicate of lines  33-34 in same file_
_- copy-paste duplicate of line 190_

^These need to name the exact test which was being duplicated rather than lines. Mentioning lines is not a good way of referencing anything.

eskultety: While at that you might want to add some blank lines in between the bullets to increase the visual separation and hence -> reading experience

eskultety (commented):
I presume we'll end up culling more, but for a vol.1 this is good work! A few nits here and there, but we're close.

tests/unit/package_managers/gomod/test_go.py
derasdf: Done

tests/unit/package_managers/cargo/test_main.py
derasdf: I'll leave it be

tests/unit/package_managers/gomod/test_go.py
derasdf: Fair, done

tests/unit/package_managers/gomod/test_go.py
derasdf: Done

tests/unit/package_managers/gomod/test_go.py
derasdf: Done

tests/unit/package_managers/gomod/test_go.py
derasdf: Done

tests/unit/package_managers/gomod/test_go.py
derasdf: Done

tests/unit/package_managers/gomod/test_main.py
derasdf: Done

tests/unit/models/test_sbom.py
derasdf: I agree that splitting it into per-PM commits will make it easier to review,  but I would say for the purposes of clearer commit history preserving it how it is right now makes more sense IMO, and if still you would like me to split these I am not sure which way I should group them then, any ideas?

tests/unit/package_managers/bundler/test_main.py
derasdf: Done

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -226,51 +226,46 @@ def test_from_missing_toolchain(
...
-        [
-            pytest.param(None, False, id="bundled_go"),
-            pytest.param("go1.20", True, id="custom_release_installed"),
-            pytest.param("go1.21.0", True, id="custom_release_needs_installation"),
-        ],
+        "retry", [pytest.param(False, id="no_retry"), pytest.param(True, id="retry")]
     )
-    @mock.patch("hermeto.core.package_managers.gomod.go.get_config")
     @mock.patch("hermeto.core.package_managers.gomod.go.Go._run")
+    @mock.patch("hermeto.core.package_managers.gomod.go.get_config")
```
eskultety: What's the reason for swapping the mock order? (multiple occurrences) within the go test module.

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -226,51 +226,46 @@ def test_from_missing_toolchain(
         assert result_go.binary == str(target_binary)
 
     @pytest.mark.parametrize(
-        "release, retry",
-        [
-            pytest.param(None, False, id="bundled_go"),
-            pytest.param("go1.20", True, id="custom_release_installed"),
-            pytest.param("go1.21.0", True, id="custom_release_needs_installation"),
```
eskultety, line 233: Why are these 2 test cases being dropped? I'm sorry I must have missed something.

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -226,51 +226,46 @@ def test_from_missing_toolchain(
...
     @mock.patch("subprocess.run")
+    @mock.patch("hermeto.core.package_managers.gomod.go.get_config")
     def test_call_failure(
         self,
-        mock_run: mock.Mock,
         mock_get_config: mock.Mock,
+        mock_run: mock.Mock,
         retry: bool,
     ) -> None:
+        opts = ["mod", "download"]
```
eskultety: Why did this assignment move from a few lines below?

tests/unit/package_managers/pip/test_main.py
```diff
@@ -92,67 +95,27 @@ def mock_requirements_file(requirements: list | None = None, options: list | Non
...
-    mock_setup_cfg: mock.Mock,
+@pytest.mark.parametrize(
+    "mock_target",
+    [
+        pytest.param("PyProjectTOML", id="pyproject_toml"),
+        pytest.param("SetupPY", id="setup_py"),
+        pytest.param("SetupCFG", id="setup_cfg"),
+    ],
+)
+def test_get_pip_metadata_from_config_file(
```
eskultety: nitpick: s/config_file/project_file

tests/unit/package_managers/pip/test_main.py
```diff
@@ -365,133 +328,260 @@ def test_ignored_and_rejected_options(self, caplog: pytest.LogCaptureFixture) ->
         assert log_msg in caplog.text
 
     @pytest.mark.parametrize(
-        "version_specs",
-        [
-            [],
-            [("<", "1")],
-            [("==", "1"), ("<", "2")],
-            [("==", "1"), ("==", "1")],  # Probably no reason to handle this?
-        ],
```
eskultety, line 374: This was really ugly, I like seeing ^this addressed.

tests/unit/package_managers/pip/test_main.py
```diff
@@ -1158,47 +1241,36 @@ def test_get_external_requirement_filepath(component_kind: str, url: str) -> Non
...
-        "myapp-0.1.tar.xz",
-        "myapp-0.1.zip",
-    ],
-)
-def test_check_metadata_from_sdist(sdist_filename: str, data_dir: Path) -> None:
-    sdist_path = data_dir / "archives" / sdist_filename
-    pip._check_metadata_in_sdist(sdist_path)
-
-
-def test_skip_check_on_tar_z(caplog: pytest.LogCaptureFixture) -> None:
```
eskultety, line 1175: Merging of this one is explicitly mentioned in the commit message...

tests/unit/package_managers/pip/test_main.py
```diff
@@ -1158,47 +1241,36 @@ def test_get_external_requirement_filepath(component_kind: str, url: str) -> Non
...
+    "sdist_filename, exc_type, expected_error",
     [
-        "myapp-0.1.tar",
-        "myapp-0.1.tar.bz2",
-        "myapp-0.1.tar.gz",
-        "myapp-0.1.tar.xz",
-        "myapp-0.1.zip",
-    ],
-)
-def test_check_metadata_from_sdist(sdist_filename: str, data_dir: Path) -> None:
```
eskultety, line 1170: ...this one isn't mentioned in the commit message at all, I don't see the parametrized values merged anywhere? What have I missed, why are we dropping this positive test?

tests/unit/test_errors.py
```diff
@@ -37,30 +22,3 @@ def test_unsupported_feature_default_friendly_msg() -> None:
...
-          - something is wrong with your repository
-          - communication with an external service failed (please try again)
-          The output of the failing command should provide more details, please check the logs.
-        """
-    ).strip()
-    assert err.friendly_msg() == expect_msg
-
-
-def test_fetch_error_friendly_msg() -> None:
-    err = errors.FetchError("Failed to fetch something")
```
eskultety, line 58: What are all of these duplicated by that we're deleting them right away? The commit message isn't explicit about these.

eskultety: Shouldn't these be moved to commit 5/6 instead?

tests/unit/models/test_sbom.py
eskultety: >  but I would say for the purposes of clearer commit history preserving it how it is right now makes more sense IMO

Commits 5/6 and 6/6 definitely, those are simple deletions, i.e. 1 operation N times. When we talk about mixed updates like test consolidation, updating tested values, renames, moving code logic etc. those diffs get confusing really quick and it's time-consuming to review, so splitting based on backends and having more explicit commit messages would be preferable.

eskultety (commented):
I crunched through commit 2/6, so no need to split, but for future reference, please do so. Most of my remaining comments are things that stem from the fact there was IMO too much happening in 2/6. Some may require an action, some you can try defending :). 5/6 and 6/6 are some satisfying diffs.
1 more iteration and I think I'll approve.

tests/unit/package_managers/gomod/test_go.py
derasdf: AI thing I missed, this whole test is deleted in commit 5 so reverted.

tests/unit/package_managers/gomod/test_go.py
derasdf: `release` was never passed to `Go()` in the test body, dead code since it was written in #1358, the tests themselves just check a 4-line if/else router (Go.__call__) already covered by `test_retry/test_retry_failure`, so deleted them in commit 5.

tests/unit/package_managers/gomod/test_go.py
derasdf: Cosmetic AI stuff, reverted.

tests/unit/package_managers/pip/test_main.py
derasdf: Done

tests/unit/package_managers/pip/test_main.py
derasdf: Moved to commit 5. No assertions in this test, it just checks test data files open without raising. The function is already exercised by test_metadata_check_fails_from_sdist and pip integration tests.

tests/unit/test_errors.py
derasdf: Yep, moved

eskultety (approved):
The diffs are now cleaner, ACK. I hope we won't need to ever deal with bisect pointing to commit 2/6.

a-ovchinnikov (approved):

PR #1527, opened 2026-04-20
Implement pnpm-lock.yaml parsing

slimreaper35:


hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -0,0 +1,137 @@
...
+    >>> _construct_npm_registry_tarball_url("", "vue", "1.0.0")
+    'https://registry.npmjs.org/vue/-/vue-1.0.0.tgz'
+    >>> _construct_npm_registry_tarball_url("vue", "core", "1.0.0")
+    'https://registry.npmjs.org/@vue/core/-/core-1.0.0.tgz'
+    """
+    if scope:
+        # The URL for scoped packages must include the '@' prefix.
+        return f"{NPM_REGISTRY_URL}/@{scope}/{name}/-/{name}-{version}.tgz"
+
+    return f"{NPM_REGISTRY_URL}/{name}/-/{name}-{version}.tgz"
```
derasdf, line 158: Lockfile v9 omits `tarball` from `resolution` for registry packages by default. The fallback hardcodes `NPM_REGISTRY_URL`:

```python
>>> _resolve_package_tarball_url('mycompany', 'lib', '1.0.0', {'integrity': 'sha512-xxx'})
'https://registry.npmjs.org/@mycompany/lib/-/lib-1.0.0.tgz'  # 404 for private registries
```

The npm backend handles this via `config.npm.proxy_url` + `_patch_url_to_point_to_a_proxy()`. Will pnpm get the same (`PnpmSettings(ProxyMixin)`) as a follow-up?

tests/unit/package_managers/pnpm/test_project.py
```diff
@@ -0,0 +1,53 @@
...
+                "repo@https://codeload.github.com/org/repo/tar.gz/abc123": {
+                    "version": "1.0.0",
+                    "resolution": {"tarball": "https://codeload.github.com/org/repo/tar.gz/abc123"},
+                },
+            },
+        },
+    )
+    packages = _parse_packages(lockfile)
+    assert packages[0].version == "1.0.0"
+    assert packages[0].url == "https://codeload.github.com/org/repo/tar.gz/abc123"
```
derasdf: No test cases for scoped packages, JSR, or _construct_npm_registry_tarball_url.

derasdf (commented):
Mostly OK with few observations.

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -0,0 +1,26 @@
+# SPDX-License-Identifier: GPL-3.0-only
```
eskultety, line 1: nitpick: can we already nest this new backend under another subdir `javascript` or `js` or whatever feels appropriate such that we can move the other backends underneath soon after? Of course we could do that in a batch as a follow-up, it's just with every code movement git blame becomes weird :) , your call.

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -0,0 +1,136 @@
...
+from dataclasses import dataclass
+from functools import cached_property
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+from hermeto.core.errors import LockfileNotFound, UnsupportedFeature
+
+NPM_REGISTRY_URL = "https://registry.npmjs.org"
```
eskultety: This should probably be defined in the `npm` backend (then used in the `CNAME_REGISTRY` constant or what the name is) and imported from there rather than redefining it here.

tests/unit/package_managers/pnpm/test_project.py
eskultety: nitpick: would wrapping these in a `TestPnpmLock` class make it slightly easier to follow/identify?

hermeto/core/package_managers/pnpm/project.py
```diff
@@ -0,0 +1,136 @@
...
+    name: str
+    version: str
+    url: str
+    integrity: str | None = None  # git and URL dependencies don't have an integrity field
+
+
+def _ensure_lockfile_version_is_supported(lockfile: PnpmLock) -> None:
+    """Ensure Hermeto supports the lockfile version in the pnpm-lock.yaml file."""
+    version = str(lockfile.get("lockfileVersion"))
+    if version != "9.0":
```
eskultety: Are we sure we want to compare version as a string literal instead of trying to covert it to a version representation? e.g. would '9' be a valid version specifier in pnpm.lock?

hermeto/core/package_managers/pnpm/project.py
```diff
@@ -0,0 +1,136 @@
...
+        if "__" in full_name:
+            scope, name = full_name.split("__", 1)
+        else:
+            scope = ""
+            name = full_name
+
+        return scope, name, version
+
+    # Scoped format
+    if id.count("@") == 2:
```
eskultety: Is this guaranteed to only ever contain 2 `@` in the metadata? Isn't that fragile? Could you e.g. have a scoped git/URL package?

hermeto/core/package_managers/pnpm/project.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @classmethod
+    def from_dir(cls, path: Path) -> "PnpmLock":
+        """Create a PnpmLock object from a directory containing a pnpm-lock.yaml file."""
+        return cls.from_file(path.joinpath("pnpm-lock.yaml"))
+
+    @cached_property
+    def packages(self) -> dict[str, dict[str, Any]]:
+        """Return the 'packages' key from the pnpm-lock.yaml file."""
+        return self.get("packages", {})
```
eskultety: Unlike with `GoWork` (the only other `UserDict`) subclass with a cached property, is there a chance of overwriting/replacing the whole `packages` key?

hermeto/core/package_managers/pnpm/main.py
```diff
@@ -2,6 +2,11 @@
 from hermeto.core.models.input import Request
 from hermeto.core.models.output import Annotation, Component, ProjectFile, RequestOutput
 from hermeto.core.models.sbom import create_backend_annotation
+from hermeto.core.package_managers.pnpm.project import (
+    PnpmLock,
+    _ensure_lockfile_version_is_supported,
+    _parse_packages,
```
eskultety: These should be public functions then.

eskultety (commented):
commit 1 (can't post a comment there):
_Irrelevant to the other commits in the PR. I discovered this, when I
forgot to add the init file for pnpm unit tests._
^This only makes sense in context of the PR, but that context is lost when merged and isn't reflected in git history.

hermeto/core/package_managers/javascript/pnpm/project.py
slimreaper35: Can we leave it after all JavaScript backends are in a shared directory?

eskultety: It'll easily be forgotten IMO given how simple the change is. IMO the easiest would be to define the constant now and consume it and when we introduce a new level of directory nesting we'd just export properly then as we'll likely need to take care of other stuff when it comes to importing and consuming common JS stuff :) .

hermeto/core/package_managers/pnpm/project.py
slimreaper35: Can we use the `semver` lib here and check the major version? Even though pnpm does not use a patch version.

eskultety: `semver` won't parse it:
```
ValueError: 1.22 is not valid SemVer string
```

The only other thing we use for versions is `packaging.version.Version` and we wrap that one in `GoVersion` due to exactly this problem.

slimreaper35: `optional_minor_and_patch` param should fix it

```python
In [3]: Version.parse("1.22", optional_minor_and_patch=True)
Out[3]: Version(major=1, minor=22, patch=0, prerelease=None, build=None)
```

eskultety: OK.

eskultety: In that case I would probably suggest to also update `GoVersion` to use `semver` instead of `packaging.Version` as long as it's compatible with that logic there :) .

tests/unit/package_managers/pnpm/test_project.py
slimreaper35: I chose doctests for them instead.

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -0,0 +1,139 @@
...
+
+    # Scoped format
+    if id.startswith("@"):
+        scope, full_name = id.split("/", maxsplit=1)
+        name, version = full_name.split("@", maxsplit=1)
+        return scope.removeprefix("@"), name, version
+
+    # Unscoped format
+    if "@" not in id:
+        raise ValueError(f"Invalid package id: {id}")
```
eskultety: This `ValueError` is unhandled meaning it'll be propagated to the user, neither is this a Pydantic validation error, don't we want a nicer error for the user?

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -0,0 +1,26 @@
...
+    """Process all pnpm source directories in the given request."""
+    components: list[Component] = []
+    project_files: list[ProjectFile] = []
+    annotations: list[Annotation] = []
+
+    deps_dir = request.output_dir.path.joinpath("deps", "pnpm")
+    deps_dir.mkdir(parents=True, exist_ok=True)
+
+    for package in request.pnpm_packages:
+        pass
```
eskultety: I think we can drop this hunk from this commit (bce761cc4d77) since you immediately modify it in the next commit.

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def fetch_pnpm_source(request: Request) -> RequestOutput:
+    """Process all pnpm source directories in the given request."""
+    components: list[Component] = []
+    project_files: list[ProjectFile] = []
+    annotations: list[Annotation] = []
+
+    deps_dir = request.output_dir.path.joinpath("deps", "pnpm")
+    deps_dir.mkdir(parents=True, exist_ok=True)
```
eskultety: `deps_dir` will be unused by the end of the PR. We should introduce it when it's time for the change, doing it now as a skeleton is dead code (that didn't turn out so well in the past for us).

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -14,7 +19,10 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
     deps_dir.mkdir(parents=True, exist_ok=True)
 
     for package in request.pnpm_packages:
-        pass
+        project_dir = request.source_dir.join_within_root(package.path)
+        lockfile = PnpmLock.from_dir(project_dir.path)
+        ensure_lockfile_version_is_supported(lockfile)
+        parse_packages(lockfile)
```
eskultety, line 16: `parse_packages` result is thrown away. We should introduce it when it's time for the change - good news is, it'll simplify this commit and the PR quite a bit until the we introduce the heart of the logic.

hermeto/core/package_managers/npm/utils.py
```diff
@@ -20,12 +20,20 @@ def normalize_resolved_url(resolved_url: str) -> NormalizedUrl:
     return NormalizedUrl(resolved_url)
 
 
+def is_from_npm_registry(url: str) -> bool:
+    """Return True if a package URL is from the NPM or Yarn registry."""
+    hostname = urlparse(url).hostname
+    return hostname is not None and any(
+        hostname in url for url in (NPM_REGISTRY_URL, YARN_REGISTRY_URL)
```
eskultety: Was the substring checking intentional? Previously we matched against full hostname string literals. Shouldn't have this been `hostname in (urlparse(NPM_REGISTRY_URL).hostname, urlparse(YARN_REGISTRY_URL).hostname)` instead?

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -0,0 +1,139 @@
...
+    url: str
+    integrity: str | None = None  # git and URL dependencies don't have an integrity field
+
+
+def ensure_lockfile_version_is_supported(lockfile: PnpmLock) -> None:
+    """Ensure Hermeto supports the lockfile version in the pnpm-lock.yaml file."""
+    raw_version = str(lockfile.get("lockfileVersion"))
+    version = Version.parse(raw_version, optional_minor_and_patch=True)
+    if version.major != 9:
+        raise UnsupportedFeature(f"Unsupported lockfileVersion: '{raw_version}'")
```
eskultety: Is this something the caller should care about? Isn't this an a class detail that the instantiation should take care about?

eskultety: Additioanlly, we're relying on the presence of the 'lockfileVersion' field. Is that field mandatory?Should we error out when it's missing?
How much of the lockfile is at least semi-standardized? I know we don't do any validation for YarnRC nor PackageJson, but I'm wondering if it made sense to at least do something like `ParsedGoWork`?

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -0,0 +1,139 @@
...
+    @classmethod
+    def from_file(cls, path: Path) -> "PnpmLock":
+        """Create a PnpmLock object from a pnpm-lock.yaml file."""
+        if not path.exists():
+            raise LockfileNotFound(
+                path, solution="Run 'pnpm install' to generate the pnpm-lock.yaml file."
+            )
+
+        with path.open() as f:
+            data = yaml.safe_load(f) or {}
```
eskultety: YAML parsing could fail.

hermeto/core/package_managers/javascript/pnpm/project.py
slimreaper35: > don't we want a nicer error for the user?

A different exception class?

hermeto/core/package_managers/javascript/pnpm/project.py
slimreaper35: I don't really know. Trying a quick test with the `pnpm install` command and a lockfile without a specified version, and looks like pnpm modifies the file itself and adds the field anyway.

The function is mainly for us, meaning if a new lockfile version is released with some breaking changes, we can investigate it before we support it. If that does not make sense, I am fine with dropping it.

eskultety: > Trying a quick test with the pnpm install command and a lockfile without a specified version, and looks like pnpm modifies the file itself and adds the field anyway.

The problem with that I see is that you need to run the tool first, i.e. without the field we don't really know if we're working with a version we can or want to support hence we're open to breakages, correct?
Wouldn't it be actually better for us to start mandating the field on the premise `pnpm install` formats it by default? I never intended to hint at dropping a check that is crucial to our support scope.

eskultety: Moreover, relaxing a check later on is always easier and praised upon rather than vice versa.

hermeto/core/package_managers/javascript/pnpm/project.py
eskultety: If none of the existing ones can be reused for this and it's something we want to check in our tests than a module-scoped exception should be fine.

slimreaper35: > If none of the existing ones can be reused for this and it's something we want to check 

Probably not. One could argue that we don't check for other characters before splitting.

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -0,0 +1,149 @@
...
+    """Ensure Hermeto supports the lockfile version in the pnpm-lock.yaml file."""
+    raw_version = lockfile.get("lockfileVersion")
+    if raw_version is None:
+        raise InvalidLockfileFormat(
+            lockfile.path,
+            err_details=None,
+            solution="The pnpm-lock.yaml file must contain a lockfileVersion field.",
+        )
+
+    version = Version.parse(raw_version, optional_minor_and_patch=True)
```
eskultety, line 76: You should stringify `raw_version` because YAML will parse `9.0` as a float and `9` as an int unless the original file declares it as `'9.0'` and so it would lead to:
```
>>> type(yaml.safe_load('version: 9.0')['version'])
float
>>> Version.parse(9.0)
TypeError: not expecting type '<class 'float'>'
```

hermeto/core/package_managers/npm/utils.py
```diff
@@ -20,12 +20,18 @@ def normalize_resolved_url(resolved_url: str) -> NormalizedUrl:
     return NormalizedUrl(resolved_url)
 
 
+def is_from_npm_registry(url: str) -> bool:
+    """Return True if a package URL is from the NPM or Yarn registry."""
+    hostname = urlparse(url).hostname
+    return hostname in (urlparse(NPM_REGISTRY_URL).hostname, urlparse(YARN_REGISTRY_URL).hostname)
```
eskultety: When I suggested ^this, I only meant it as a readable stub to support my point :) , however, this parses the URL constants for every package, so probably worth having a module-level constants for the hostnames.

tests/unit/package_managers/javascript/pnpm/test_project.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def test_ensure_lockfile_version_is_unsupported_fails(self, tmp_path: Path) -> None:
+        lockfile = PnpmLock(path=tmp_path / "pnpm-lock.yaml", data={"lockfileVersion": "10.0"})
+        with pytest.raises(UnsupportedFeature):
+            ensure_lockfile_version_is_supported(lockfile)
+
+    def test_ensure_lockfile_version_is_missing_fails(self, tmp_path: Path) -> None:
+        lockfile = PnpmLock(path=tmp_path / "pnpm-lock.yaml", data={})
+        with pytest.raises(InvalidLockfileFormat):
+            ensure_lockfile_version_is_supported(lockfile)
```
eskultety: Sounds like `parametrize` for these 2

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -0,0 +1,149 @@
...
+    if version.major != 9:
+        raise UnsupportedFeature(f"Unsupported lockfileVersion: '{raw_version}'")
+
+
+def parse_packages(lockfile: PnpmLock) -> list[PnpmPackage]:
```
eskultety, line 81: This is dead-code for now. Is this the full logic? No changes to the function needed? In any case, please drop it for now and only introduce it when you need it so that the change is logically paired.
If you happen to need to update it in a follow-up work then what was the point of adding it now? :)

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -0,0 +1,149 @@
...
+        version = data.get("version") or version_from_id
+        resolution = data.get("resolution", {})
+        url = _resolve_package_tarball_url(scope, name, version, resolution)
+        integrity = resolution.get("integrity")
+        result.append(PnpmPackage(id, scope, name, version, url, integrity))
+
+    return result
+
+
+def _parse_package_metadata(id: str) -> tuple[str, str, str]:
```
eskultety: Same here, because the only caller is `parse_packages` this is dead code. I'd suggest waiting with this change until we're ready to wire it all in the `fetch` entrypoint and we no longer discard the results.

hermeto/core/package_managers/npm/utils.py
slimreaper35: or maybe just function variables?

eskultety: Why function variables?

hermeto/core/package_managers/javascript/pnpm/project.py
slimreaper35: I want to use it in the main module because it's all related to parsing the lockfile. And I want to keep the PRs smaller. If we need some changes later, we can update the function as usually.

Is this acceptable?

```
    for package in request.pnpm_packages:
        project_dir = request.source_dir.join_within_root(package.path)
        lockfile = PnpmLock.from_dir(project_dir.path)
        ensure_lockfile_version_is_supported(lockfile)
        parse_packages(lockfile)
```

(without creating deps directory)

tests/unit/package_managers/javascript/pnpm/test_project.py
slimreaper35: The tests raise a different exception type, so I would prefer this approach.

hermeto/core/package_managers/npm/utils.py
slimreaper35: Because they won't be imported to other backends, only the function will be imported.

tests/unit/package_managers/javascript/pnpm/test_project.py
eskultety: It doesn't really matter, the exception is just a parametrized value, the point of parametrize is to reuse the same test logic using a variation of arguments which is exactly this case. By the same logic we should stop using parametrize altogether and we should open-code all positive tests based on a different input value to a parameter.

hermeto/core/package_managers/javascript/pnpm/project.py
eskultety: > I want to use it in the main module because it's all related to parsing the lockfile. And I want to keep the PRs smaller

I understand, that's however not a reason to introduce dead code that might or might not get used in the future, we've been bitten by that pattern in the past. Why can't package parsing happen when you introduce resolving? It's a minor thing that you can enable in a single commit, so I'm not sure it would make a difference in the follow-up PR you plan on contributing. This is really not a matter of throwing results away or having dead code in THIS PR, it's about some principles we should follow to keep a good contributing hygiene.

hermeto/core/package_managers/npm/utils.py
eskultety: now I see your point, sure!

hermeto/core/package_managers/javascript/pnpm/project.py
slimreaper35: It won't be dead code if the `fetch_pnpm_source`function parses the lockfile. I can use that for manual testing. The implementation is just in progress.

If I remove `parse_packages`, then I can remove other unused utility functions, and the whole PR becomes pointless.

eskultety: Fair enough, I'll resolve this thread, but then the question becomes, if by omitting some changes a PR becomes pointless, why weren't other aspects dealt with in the first place? :) What problem does this PR granularity really solve?

slimreaper35: > What problem does this PR granularity really solve?

I thought it would be easier for reviewers. At least for me. Less code, fewer comments, etc. IIRC, we use this granular approach when implementing other backends.

tests/unit/package_managers/javascript/pnpm/test_project.py
slimreaper35: Ok, done.

hermeto/core/package_managers/javascript/pnpm/project.py
eskultety: > Less code, fewer comments, etc. IIRC, we use this granular approach when implementing other backends.

Granular approach is not a problem in general as long as the changes are logically enclosed, that's not what's happening here. While less code might be easier for review, the extra granular approach kills context - without seeing e.g. how parsed packages translate to the resolver logic complicates review (at least for me) because I don't see the end result, or at least the "local maximum". What I'm trying to argue against is granular approach for the sake of granularity because it seemingly allows ultra-parallel development on some feature, I've always been quite skeptical about that approach because it just leads to a mess :) .

eskultety (approved):

derasdf (approved):

hermeto/core/package_managers/javascript/pnpm/main.py
a-ovchinnikov, line 1: Nitpick on Commit message of https://github.com/hermetoproject/hermeto/commit/c7cfcfe22107a5261e6add503c3bfbbe2e08c0be : the change also adds a skeleton of a fetch function, might be worth mentioning it. (At the same time are you sure we need an extra directory level for JS?)

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -9,6 +14,12 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:  # noqa
     project_files: list[ProjectFile] = []
     annotations: list[Annotation] = []
 
+    for package in request.pnpm_packages:
+        project_dir = request.source_dir.join_within_root(package.path)
+        lockfile = PnpmLock.from_dir(project_dir.path)
+        ensure_lockfile_version_is_supported(lockfile)
```
a-ovchinnikov: Minor suggestion: could this check be a part of constructor?

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -0,0 +1,150 @@
...
+
+from hermeto.core.errors import InvalidLockfileFormat, LockfileNotFound, UnsupportedFeature
+
+NPM_REGISTRY_URL = "https://registry.npmjs.org"
+
+
+class PnpmLock(UserDict):
+    """Class representing a pnpm-lock.yaml file."""
+
+    def __init__(self, path: Path, data: dict[str, Any]) -> None:
```
a-ovchinnikov, line 17: Non-blocking, strictly optional suggestion: it looks like the class is supposed to be initialized through classmethods, it might be worth it to block (or soft-block) direct instantiations (e.g. via a default argument) to be explicit that it is not supposed to be instantiated directly.

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -0,0 +1,150 @@
...
+        version = data.get("version") or version_from_id
+        resolution = data.get("resolution", {})
+        url = _resolve_package_tarball_url(scope, name, version, resolution)
+        integrity = resolution.get("integrity")
+        result.append(PnpmPackage(id, scope, name, version, url, integrity))
+
+    return result
+
+
+def _parse_package_metadata(id: str) -> tuple[str, str, str]:
```
a-ovchinnikov: For the sake of future readers let's either use a namedtuple or some type aliases for this:
```python3
p_scope: TypeAlias = str
p_name: TypeAlias = str
p_version: TypeAlias = str
def _parse_package_metadata(id: str) -> tuple[p_scope, p_name, p_id]:
    ...
```

tests/unit/package_managers/javascript/pnpm/test_project.py
```diff
@@ -0,0 +1,62 @@
...
+from hermeto.core.errors import InvalidLockfileFormat, LockfileNotFound, UnsupportedFeature
+from hermeto.core.package_managers.javascript.pnpm.project import (
+    PnpmLock,
+    ensure_lockfile_version_is_supported,
+    parse_packages,
+)
+
+
+class TestPnpmLock:
+    def test_from_dir_with_missing_pnpm_lock(self, tmp_path: Path) -> None:
```
a-ovchinnikov: Minor naming suggestion: s/test_from_dir_with_missing_pnpm_lock/instantiating_PnpmLock_from_dir_without_lockfile_fails/

tests/unit/package_managers/javascript/pnpm/test_project.py
```diff
@@ -0,0 +1,62 @@
...
+        with pytest.raises(LockfileNotFound):
+            PnpmLock.from_dir(tmp_path)
+
+    def test_ensure_lockfile_version_is_supported(self, tmp_path: Path) -> None:
```
a-ovchinnikov: A test without assertions looks fishy: do you really need this one or can it be indirectly tested someplace else? (If you follow my suggestion and make it constructor's responsibility to verify the version you won't need this test since any test that deals with successful parsing will be testing this method.)

tests/unit/package_managers/javascript/pnpm/test_project.py
```diff
@@ -0,0 +1,62 @@
...
+        [
+            pytest.param(
+                {"lockfileVersion": "1000.0"}, UnsupportedFeature, id="unsupported_version_str"
+            ),
+            pytest.param(
+                {"lockfileVersion": 1000.0}, UnsupportedFeature, id="unsupported_version_float"
+            ),
+            pytest.param({}, InvalidLockfileFormat, id="missing_version"),
+        ],
+    )
```
a-ovchinnikov, line 33: I'll misuse this opportunity to promote an approach. What would you say if parametrization could be done like this:
```python
@pytest.mark.parametrize(
make_parameters("""
| id                        | data                        | exception            |
+---------------------------+-----------------------------+----------------------+
| unsupported_version_str   |{"lockfileVersion": "1000.0"}| UnsupportedFeature   |
| unsupported_version_float |{"lockfileVersion": 1000.0}  | UnsupportedFeature   |
| missing_version           |{}                           | InvalidLockfileFormat|
"""))
def test_ensure_lockfile_version_is_unsupported_fails(...)
```

a-ovchinnikov (approved):
LGTM with minor optional suggestions and a couple of questions.

tests/unit/package_managers/javascript/pnpm/test_project.py
slimreaper35: Currently, we don't have such a construction anywhere. Also, I am not sure if typing all these `-`, `+` characters is a viable approach.

hermeto/core/package_managers/javascript/pnpm/main.py
slimreaper35: > At the same time are you sure we need an extra directory level for JS?

That was requested somewhere above by @eskultety. I believe the plan is to move all JS backends and Python backends to a shared directory.

hermeto/core/package_managers/javascript/pnpm/project.py
slimreaper35: This direct initialization is used in testing.

tests/unit/package_managers/javascript/pnpm/test_project.py
a-ovchinnikov: > Currently, we don't have such a construction anywhere

Correct. Let's imagine for the sake of experiment that we have it, furthermore, let's imagine that there is a Vim/<preferred IDE> function to autoformat the table. What would you say about having test parameters defined in a table?

hermeto/core/package_managers/javascript/pnpm/main.py
a-ovchinnikov: Thank you!

tests/unit/package_managers/javascript/pnpm/test_project.py
a-ovchinnikov: @taylormadore this is the parameters shape that I have been talking about lately.

eskultety: I'm going to resolve this thread on the basis that this should be discussed where this is proposed: https://github.com/hermetoproject/hermeto/pull/1528

tests/unit/package_managers/javascript/pnpm/test_project.py
slimreaper35: Ok, done.

hermeto/core/package_managers/javascript/pnpm/project.py
slimreaper35: I remember a similar thread from the past about defining aliases for built-in types, e.g., `do_nothing = None`, and I believe it was collectively rejected.

eskultety: Agreed, I vaguely remember that thread too, let's not do this for scalars. Namedtuples work for me.

slimreaper35: Added.

PR #1528, opened 2026-04-21
👨‍🔬🧪📐 Experiment with tests

a-ovchinnikov:
# Problem statement

While auditing the codebase I noticed that parameters lists in some unit tests are very long, so long that that can easily confuse a maintainer and they will forget while scrolling which variables are being substituted — an obvious UX problem!

# Fix

Since tables are easier for humans to grasp at a glance than lists I went ahead and replaced one of the lists with a table. As a result the test became more compact and easier to comprehend and this increased the efficiency of screen real estate utilization by up to 70%. (I also had to code a helper because unfortunately pytest refused to recognize the benefits of tables otherwise.)

# Changes

 * Parameter list is replaced with a parameter table;
 * Some glue code is added to keep it in place.

# Future work

If this is something we want I'll replace more lists with tables.

### DIFF
--- a/tests/unit/package_managers/cargo/test_main.py
+++ b/tests/unit/package_managers/cargo/test_main.py
@@ -76,56 +76,93 @@ def test_virtual_workspace_without_workspace_version(rooted_tmp_path: RootedPath
     assert version is None
 
 
+# This is something I wanted to try for a long time. Long parameter lists in
+# tests are hard to comprehend, tables are much better for that.  My initial
+# plan was to use standalone .md files with doctests which turned out to be too
+# cumbersome: resulting "examples" became really hard to follow real fast.
+# Gherkin-based test cases do not cut it either (or I cannot figure out how to
+# use the language correctly). As always in such cases I tried to invent my own
+# ad-hoc format.  What I got so far is a (for now) rigidly formatted table (see
+# parameters section in test_cargo_package_purl_generation() below) each visual
+# line group in which represents a test case. It is not pretty as it is now:
+# turns out some entries in such tables could be a mouthful, borders undulate
+# and I do not have a good idea of what to do about that. Maybe untabulate()
+# could accept **kwargs to substitute at least the longest fields in the table
+# (however that would defeat the goal of a visually compact and uninterrupted
+# test case description). The helper itself is essentially a MWP, a stream of
+# conciousness with a lot to improve, I would prefer to get some feedback on
+# the approach first and then polish it and separate into logical pieces.  This
+# location is pretty random, if the idea is appreciated then I'll look for a
+# better place for the helper.
+def untabulate(table: str) -> tuple:
+    # ruff is upset about name 'l' which is local to the next line
+    lines = [ls for l in table.split("\n") if (ls := l.strip())]  # noqa: E741
+    header, separator, raw_examples = lines[0], lines[1], lines[2:]
+    # Want to preserve empty spaces from the table, thus no stripping before filtering
+    examples = [[k.strip() for k in e.split("|") if k] for e in raw_examples]
+    fields = [fs for f in header.split("|") if (fs := f.strip())]
+    if not fields[0] == "id":
+        raise ValueError("id must be present in the table and specified for every test case")
+    if not (set(separator) <= set("+-|")):
+        raise ValueError(
+            f"Separator must consist of +-| symbols only, no idea how to interpret this: {separator}"
+        )
+    all_results = []
+    # mypy is ridiculous
+    current_result: dict = dict(zip(fields, [list() for _ in fields]))
+    while examples:
+        current_line = examples.pop(0)
+        id_, rest_of_line = current_line[0], current_line[1:]
+        if id_:
+            if current_result["id"]:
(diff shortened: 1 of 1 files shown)


a-ovchinnikov:
@derasdf, https://github.com/hermetoproject/hermeto/pull/1526 inspired me on this.

a-ovchinnikov:
The tests also fail in a much (visually) nicer manner now:
```shell
pkg = {'checksum': 'abc123', 'name': 'foo', 'source': 'registry+https://my-crates.io-mirror.example.com/index', 'version': '0.1.0'}, expected_purl = 'pkg:cargo/foo@0.1.0?checksum=abc1234'

    @pytest.mark.parametrize(
        *untabulate(
            """
    | id                                        | pkg                                                               | expected_purl                         |
    +-------------------------------------------+-------------------------------------------------------------------+---------------------------------------+
    | simple_package                            |{"name": "foo",                                                    | "pkg:cargo/foo@0.1.0"                 |
    |                                           | "version": "0.1.0"}                                               |                                       |
    | package_with_registry_source_and_checksum |{"name": "foo",                                                    | "pkg:cargo/foo@0.1.0?checksum=abc123" |
    |                                           | "version": "0.1.0",                                               |                                       |
    |                                           | "source": "registry+https://github.com/rust-lang/crates.io-index",|                                       |
    |                                           | "checksum": "abc123"}                                             |                                       |
    |                                           |                                                                   |                                       |
    | package_with_git_source                   |{"name": "foo", "version": "0.1.0",                                | "pkg:cargo/foo@0.1.0?vcs_url=git%2Bhttps://github.com/rust-random/rand%40abc123"|
    |                                           | "source": "git+https://github.com/rust-random/rand?rev=abc123#abc123"}|                                   |
    |                                           |                                                                   |                                       |
    | package_with_alternate_registry           |{"name": "foo","version": "0.1.0",                                 | "pkg:cargo/foo@0.1.0?repository_url=https://my-registry.example.com/index" |
    |                                           | "source": "registry+https://my-registry.example.com/index"}       |
    |                                           |                                                                   |                                       |
    | package_with_crates_io_in_subdomain_no_repository_url |{"name": "foo","version": "0.1.0",                     | "pkg:cargo/foo@0.1.0?checksum=abc1234" |
    |                                           | "source": "registry+https://my-crates.io-mirror.example.com/index",|                                      |
    |                                           | "checksum": "abc123"}                                             |                                       |
    """
        )
    )
    def test_cargo_package_purl_generation(pkg: dict[str, Any], expected_purl: str) -> None:
        package = CargoPackage(**pkg)
>       assert package.purl.to_string() == expected_purl
E       AssertionError: assert 'pkg:cargo/fo...ecksum=abc123' == 'pkg:cargo/fo...cksum=abc1234'
E         
E         - pkg:cargo/foo@0.1.0?checksum=abc1234
E         ?                                    -
E         + pkg:cargo/foo@0.1.0?checksum=abc123

tests/unit/package_managers/cargo/test_main.py:169: AssertionError

```

eskultety:
This IMO feels like locking ourselves to a completely custom solution which, while the idea might be appealing, apparently will require some additional formatting gymnastics. More importantly though - we'd deviate from an industry-standardized solution of relying on plain pytest syntax and constructs, however verbose they may be. The question I would ask - if the verbosity of the test parametrization declaration is the main motivator here - can we do something about the test structure then to reduce the optical fatigue inflicted by the declarative and verbose nature of some tests? Would more granularity in tests help, combined with our effort to get rid of some behemoth unit tests and replace them with integration tests instead?

a-ovchinnikov:
I don't think granularity would help much, we will still be bound by the number of different combinations we would need to go through, moreover, the part of tests overhaul work is consolidation of _more_ test combinations into the parameters. One of my concerns is separation of actual values from their names, in some tests it is hard to tell at a glance what is an expected value and what is an input.

I am not sure I agree with the industry-standard argument, it was not that long ago when industry-standard meant Java and nothing else, but here we are, writing almost exclusively in Python; the fact that everyone uses something does not necessarily mean that they have explored all options, sometimes it is just inertia. I think the same industry-standard argument could be made about tables, tables are universally understood as well, as long as the machinery processing them is well hidden they should work fine for conveying the meaning of a test.
                                                                                
On this being custom -- yes, it is, but PoC for processing tables is quite compact, that's why I decided to suggest it in the first place. Manual formatting is avoidable -- I have not published it yet, but I cobbled together yet another PoC that takes a Python module as an input, detects all parametrizing decorators and replaces them with (mostly) nicely formatted tables (there are edge cases which I decided not to address in the PoC, those have to be touched up by hand. Or just left as is.).

slimreaper35:
What are the plans with this PR?

a-ovchinnikov:
> What are the plans with this PR?

I suggest keeping it around until the moment when the aesthetics of the proposed approach becomes apparent to everyone. (I am also hoping to publish a PoC of an automated converter from the current format to the proposed one one day, but sadly can't make it a high enough priority, the code needs some clean-up.)

a-ovchinnikov:
Closing due to lack of interest

PR #1530, opened 2026-04-21
refactor: implement batching of downloads for PyPI requirements

### LINKED ISSUE #1456: more parallelization for pip dependency fetching
In the pip module, each requirement is processed sequentially before moving on to the next. While the file downloads for a single requirement do happen in parallel, the requirements themselves are processed one at a time. If possible, we should attempt to batch downloads into a single parallel operation in order to improve performance.

ST3V1K:
## Summary
- Implement batching of downloads for PyPI requirements
- Add unit test coverage for the batching behavior

## Context
Based on the work in #1478 by @victor-adebowale-momodu.

Closes: https://github.com/hermetoproject/hermeto/issues/1456
Closes #1478

### DIFF
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -5,7 +5,7 @@ import tarfile
 import zipfile
 from collections.abc import Iterable, Iterator
 from pathlib import Path
-from typing import Any
+from typing import Any, NamedTuple
 from urllib import parse as urlparse
 
 import pypi_simple
@@ -52,6 +52,11 @@ log = logging.getLogger(__name__)
 
 DEFAULT_BUILD_REQUIREMENTS_FILE = "requirements-build.txt"
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
+
+
+class _PyPIArtifact(NamedTuple):
+    requirement: PipRequirement
+    dpi: DistributionPackageInfo
 
 
 def fetch_pip_source(request: Request) -> RequestOutput:
@@ -321,34 +326,20 @@ def _process_req(
     return download_info
 
 
-def _process_pypi_req(
-    req: PipRequirement,
+def _process_pypi_reqs(
     requirements_file: PipRequirementsFile,
-    index_url: str,
     pip_deps_dir: RootedPath,
-    binary_filters: PipBinaryFilters | None = None,
+    pypi_artifacts: list[_PyPIArtifact],
 ) -> list[dict[str, Any]]:
-    download_infos: list[dict[str, Any]] = []
+    files = {dpi.url: dpi.path for _, dpi in pypi_artifacts if not dpi.path.exists()}
+    if files:
+        log.info("Downloading %d PyPI artifacts", len(files))
+        asyncio.run(async_download_files(files, get_config().runtime.concurrency_limit))
 
-    artifacts: list[DistributionPackageInfo] = process_package_distributions(
-        req, pip_deps_dir, binary_filters, index_url
-    )
-
-    files = {dpi.url: dpi.path for dpi in artifacts if not dpi.path.exists()}
-    asyncio.run(async_download_files(files, get_config().runtime.concurrency_limit))
-
-    for artifact in artifacts:
-        download_infos.append(
-            _process_req(
-                req,
-                requirements_file,
-                pip_deps_dir,
-                artifact.download_info,
-                dpi=artifact,
-            )
-        )
-
-    return download_infos
+    return [
+        _process_req(req, requirements_file, pip_deps_dir, dpi.download_info, dpi=dpi)
+        for req, dpi in pypi_artifacts
+    ]
 
 
 def _process_vcs_req(
@@ -415,17 +406,18 @@ def _download_dependencies(
     pip_deps_dir: RootedPath = output_dir.join_within_root("deps", "pip")
     pip_deps_dir.path.mkdir(parents=True, exist_ok=True)
 
+    pypi_artifacts: list[_PyPIArtifact] = []
     for req in requirements_file.requirements:
         log.info("-- Processing requirement line '%s'", req.download_line)
         if req.kind == "pypi":
-            download_infos: list[dict[str, Any]] = _process_pypi_req(
+            dpis: list[DistributionPackageInfo] = process_package_distributions(
                 req,
-                requirements_file=requirements_file,
-                index_url=options["index_url"] or pypi_simple.PYPI_SIMPLE_ENDPOINT,
-                pip_deps_dir=pip_deps_dir,
-                binary_filters=binary_filters,
+                pip_deps_dir,
+                binary_filters,
+                options["index_url"] or pypi_simple.PYPI_SIMPLE_ENDPOINT,
             )
-            processed.extend(download_infos)
+            pypi_artifacts.extend(_PyPIArtifact(req, dpi) for dpi in dpis)
+            continue
         elif req.kind == "vcs":
             download_info = _process_vcs_req(
                 req,
@@ -445,7 +437,10 @@ def _download_dependencies(
             # Should not happen
             raise RuntimeError(f"Unexpected requirement kind: '{req.kind!r}'")
 
-        log.info("-- Finished processing requirement line '%s'\n", req.download_line)
+        log.info("-- Finished processing requirement line '%s'", req.download_line)
+
+    pypi_download_infos = _process_pypi_reqs(requirements_file, pip_deps_dir, pypi_artifacts)
+    processed.extend(pypi_download_infos)
 
     return processed
 
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -1,9 +1,10 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import asyncio
+import functools
 import logging
 import tarfile
 import zipfile
-from collections.abc import Iterable, Iterator
+from collections.abc import Callable, Iterable, Iterator
 from pathlib import Path
 from typing import Any, NamedTuple
 from urllib import parse as urlparse
@@ -369,6 +370,16 @@ def _process_url_req(
     return result
 
 
+async def _resolve_pypi_distributions(
+    reqs: list[PipRequirement],
+    resolve_callback: Callable[[PipRequirement], list[DistributionPackageInfo]],
+) -> list[list[DistributionPackageInfo]]:
+    """Resolve PyPI distributions for all requirements concurrently."""
+    loop = asyncio.get_running_loop()
+    tasks = [loop.run_in_executor(None, resolve_callback, req) for req in reqs]
+    return await asyncio.gather(*tasks)
+
+
 def _download_dependencies(
     output_dir: RootedPath,
     requirements_file: PipRequirementsFile,
@@ -406,17 +417,11 @@ def _download_dependencies(
     pip_deps_dir: RootedPath = output_dir.join_within_root("deps", "pip")
     pip_deps_dir.path.mkdir(parents=True, exist_ok=True)
 
-    pypi_artifacts: list[_PyPIArtifact] = []
+    pypi_reqs: list[PipRequirement] = []
     for req in requirements_file.requirements:
         log.info("-- Processing requirement line '%s'", req.download_line)
         if req.kind == "pypi":
-            dpis: list[DistributionPackageInfo] = process_package_distributions(
-                req,
-                pip_deps_dir,
-                binary_filters,
-                options["index_url"] or pypi_simple.PYPI_SIMPLE_ENDPOINT,
-            )
-            pypi_artifacts.extend(_PyPIArtifact(req, dpi) for dpi in dpis)
+            pypi_reqs.append(req)
             continue
         elif req.kind == "vcs":
             download_info = _process_vcs_req(
@@ -439,8 +444,18 @@ def _download_dependencies(
 
         log.info("-- Finished processing requirement line '%s'", req.download_line)
 
-    pypi_download_infos = _process_pypi_reqs(requirements_file, pip_deps_dir, pypi_artifacts)
-    processed.extend(pypi_download_infos)
+    pypi_artifacts: list[_PyPIArtifact] = []
+    if pypi_reqs:
+        resolve_callback = functools.partial(
+            process_package_distributions,
+            pip_deps_dir=pip_deps_dir,
+            binary_filters=binary_filters,
+            index_url=options["index_url"] or pypi_simple.PYPI_SIMPLE_ENDPOINT,
+        )
+        pypi_dpis = asyncio.run(_resolve_pypi_distributions(pypi_reqs, resolve_callback))
+        reqs_dpis_zipped = zip(pypi_reqs, pypi_dpis)
+        pypi_artifacts = [_PyPIArtifact(req, dpi) for req, dpis in reqs_dpis_zipped for dpi in dpis]
+        processed.extend(_process_pypi_reqs(requirements_file, pip_deps_dir, pypi_artifacts))
 
     return processed
 
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -5,7 +5,7 @@ import tarfile
 import zipfile
 from collections.abc import Iterable, Iterator
 from pathlib import Path
-from typing import Any
+from typing import Any, NamedTuple
 from urllib import parse as urlparse
 
 import pypi_simple
@@ -52,6 +52,11 @@ log = logging.getLogger(__name__)
 
 DEFAULT_BUILD_REQUIREMENTS_FILE = "requirements-build.txt"
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
+
+
+class _PyPIArtifact(NamedTuple):
+    requirement: PipRequirement
+    dpi: DistributionPackageInfo
 
 
 def fetch_pip_source(request: Request) -> RequestOutput:
@@ -321,34 +326,19 @@ def _process_req(
     return download_info
 
 
-def _process_pypi_req(
-    req: PipRequirement,
+def _process_pypi_reqs(
     requirements_file: PipRequirementsFile,
-    index_url: str,
     pip_deps_dir: RootedPath,
-    binary_filters: PipBinaryFilters | None = None,
+    pypi_artifacts: list[_PyPIArtifact],
 ) -> list[dict[str, Any]]:
-    download_infos: list[dict[str, Any]] = []
+    files = {dpi.url: dpi.path for _, dpi in pypi_artifacts if not dpi.path.exists()}
+    if files:
+        asyncio.run(async_download_files(files, get_config().runtime.concurrency_limit))
 
-    artifacts: list[DistributionPackageInfo] = process_package_distributions(
-        req, pip_deps_dir, binary_filters, index_url
-    )
-
-    files = {dpi.url: dpi.path for dpi in artifacts if not dpi.path.exists()}
-    asyncio.run(async_download_files(files, get_config().runtime.concurrency_limit))
-
-    for artifact in artifacts:
-        download_infos.append(
-            _process_req(
-                req,
-                requirements_file,
-                pip_deps_dir,
-                artifact.download_info,
-                dpi=artifact,
-            )
-        )
-
-    return download_infos
+    return [
+        _process_req(req, requirements_file, pip_deps_dir, dpi.download_info, dpi=dpi)
+        for req, dpi in pypi_artifacts
+    ]
 
 
 def _process_vcs_req(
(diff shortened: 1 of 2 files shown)


hermeto/core/package_managers/pip/main.py
```diff
@@ -415,17 +416,17 @@ def _download_dependencies(
...
-                requirements_file=requirements_file,
-                index_url=options["index_url"] or pypi_simple.PYPI_SIMPLE_ENDPOINT,
-                pip_deps_dir=pip_deps_dir,
-                binary_filters=binary_filters,
+                pip_deps_dir,
+                binary_filters,
+                options["index_url"] or pypi_simple.PYPI_SIMPLE_ENDPOINT,
             )
-            processed.extend(download_infos)
+            pypi_artifacts.extend(PyPIArtifact(req, dpi) for dpi in dpis)
```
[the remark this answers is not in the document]
eskultety: Unbelievable that Gemini finally caught this! AFAIK this would require some deeper refactor, so I wouldn't mind deferring to a follow-up unless this ends up being just a few more lines of code.

hermeto/core/package_managers/pip/main.py
```diff
@@ -447,6 +448,9 @@ def _download_dependencies(
 
         log.info("-- Finished processing requirement line '%s'\n", req.download_line)
```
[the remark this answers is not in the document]
eskultety: Valid.

hermeto/core/package_managers/pip/main.py
```diff
@@ -447,6 +448,9 @@ def _download_dependencies(
...
 
+    pypi_download_infos = _process_pypi_reqs(requirements_file, pip_deps_dir, pypi_artifacts)
+    processed.extend(pypi_download_infos)
```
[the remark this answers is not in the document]
eskultety: SBOM is by nature unordered, this is nonsense.

hermeto/core/package_managers/pip/main.py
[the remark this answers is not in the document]
ST3V1K: Done

hermeto/core/package_managers/pip/main.py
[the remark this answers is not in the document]
ST3V1K: Not sure if this would actually be any better.
```py
loop = asyncio.get_event_loop()
tasks = [loop.run_in_executor(None, resolve, req) for req in reqs]
return await asyncio.gather(*tasks)
```

The follow-up would have to use an alternative pypi_simple that uses aiohttp.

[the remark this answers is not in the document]
eskultety: Yes, the speedup would there would most likely be marginal only.

hermeto/core/package_managers/pip/main.py
```diff
@@ -380,6 +381,16 @@ def _process_url_req(
     return result
 
 
+async def _resolve_pypi_distributions(
+    reqs: list[PipRequirement],
+    resolve: Callable[[PipRequirement], list[DistributionPackageInfo]],
+) -> list[list[DistributionPackageInfo]]:
+    """Resolve PyPI distributions for all requirements concurrently."""
+    loop = asyncio.get_event_loop()
```
eskultety: nitpick: ^this is fine, but the docs on our currently supported Python version 3.10 says that `get_running_loop` is the preferred choice: https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.get_event_loop although we woulnd't suffer from the complex behaviour problems mentioned in there since this is run directly from `asyncio.run`.

hermeto/core/package_managers/pip/main.py
```diff
@@ -54,6 +54,13 @@
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
 
 
+class PyPIArtifact(NamedTuple):
+    """A class representing a PyPI artifact."""
```
eskultety: nitpick: the tuple is self explanatory, the docstring may be dropped.

hermeto/core/package_managers/pip/main.py
```diff
@@ -5,7 +5,7 @@
 import zipfile
 from collections.abc import Iterable, Iterator
 from pathlib import Path
-from typing import Any
+from typing import Any, NamedTuple
```
eskultety, line 9: Commit subject nitpick: the PR says refactor, but the commit says `feat`, this is not a user-facing feature, it's an internal improvement, same thing for commit 3.

hermeto/core/package_managers/pip/main.py
```diff
@@ -321,33 +328,28 @@ def _checksum_must_match_or_path_unlink(
...
                 requirements_file,
                 pip_deps_dir,
-                artifact.download_info,
-                dpi=artifact,
+                dpi.download_info,
+                dpi=dpi,
             )
         )
-
+        log.info("-- Finished processing requirement line '%s'\n", req.download_line)
```
eskultety: This will report the same message for each artifact of a requirement:
```
$ cat requirements-test.txt
pydantic==2.12.5 \
    --hash=sha256:4d351024c75c0f085a9febbb665ce8c0c6ec5d30e903bdb6394b7ede26aebb49 \
    --hash=sha256:e561593fccf61e8a20fc46dfc2dfe075b8be7d0188df33f221ad1f0139180f9d
$  hermeto fetch-deps '{"type":"pip", "requirements_files": ["requirements-test.txt"], "binary": {}}'
[...]
2026-04-22 10:50:23,921 INFO -- Processing requirement line 'pydantic==2.12.5'
2026-04-22 10:50:24,295 INFO Verifying checksums of pydantic-2.12.5-py3-none-any.whl
>>>>2026-04-22 10:50:24,296 INFO -- Finished processing requirement line 'pydantic==2.12.5'

2026-04-22 10:50:24,296 INFO Verifying checksums of pydantic-2.12.5.tar.gz
>>>>2026-04-22 10:50:24,317 INFO -- Finished processing requirement line 'pydantic==2.12.5'

...
```
Bonus points - when you move the log line out of the loop then you can condense the loop to a single return comprehension.

hermeto/core/package_managers/pip/main.py
eskultety: _and validating checksums for each requirement_

and validating checksums for each requirement **sequentially**

> - resolve all PyPI requirements
>  - batch download all requirements using the existing async download  implementation
>  - verify the checksums for all PyPI artifacts by calling the existing `_process_req`     function

1. resolve all ... **sequentially**
2. batch ...
3. verify checksums for all PyPI artifacts after all have been downloaded

tests/unit/package_managers/pip/test_main.py
```diff
@@ -673,15 +720,21 @@ def test_download_dependencies_pypi(
...
         # <check basic logging output>
-        assert f"-- Processing requirement line '{req.download_line}'" in caplog.text
+        assert f"-- Processing requirement line '{foo_req.download_line}'" in caplog.text
+        assert (
+            f"Successfully processed '{foo_req.download_line}' in path 'deps/pip/foo-1.0.tar.gz'"
+        ) in caplog.text
+        assert f"-- Processing requirement line '{bar_req.download_line}'" in caplog.text
         assert (
-            f"Successfully processed '{req.download_line}' in path 'deps/pip/foo-1.0.tar.gz'"
+            f"Successfully processed '{bar_req.download_line}' in path 'deps/pip/bar-2.0.tar.gz'"
```
eskultety: Since you're updating the test I think you can add one more patch at the beginning of the series to get rid of these string comparisons.

hermeto/core/package_managers/pip/main.py
```diff
@@ -450,6 +455,18 @@ def _download_dependencies(
         if req.kind != "pypi":
             log.info("-- Finished processing requirement line '%s'\n", req.download_line)
 
+    pypi_artifacts: list[PyPIArtifact] = []
+    if pypi_reqs:
+        resolve = functools.partial(
```
eskultety: nitpick: `resolve_callback` or `resolve_partial` might be more descriptive names for a partial.

hermeto/core/package_managers/pip/main.py
```diff
@@ -450,6 +455,18 @@ def _download_dependencies(
...
+            pip_deps_dir=pip_deps_dir,
+            binary_filters=binary_filters,
+            index_url=options["index_url"] or pypi_simple.PYPI_SIMPLE_ENDPOINT,
+        )
+        pypi_dpis = asyncio.run(_resolve_pypi_distributions(pypi_reqs, resolve))
+        for req, dpis in zip(pypi_reqs, pypi_dpis):
+            pypi_artifacts.extend(PyPIArtifact(req, dpi) for dpi in dpis)
+
     pypi_download_infos = _process_pypi_reqs(requirements_file, pip_deps_dir, pypi_artifacts)
     processed.extend(pypi_download_infos)
```
eskultety: nitpick: would the following snippet hinder readability?
```
if pypi_reqs:
    ...
    pypi_dpis = ...
    reqs_dpis_zipped = zip(pypi_reqs, pypi_dpis)
    pypi_artifacts = [PyPIArtifact(req, dpi) for req, dpis in reqs_dpis_zipped for dpi in dpis]
    processed.extend(_process_pypi_reqs(requirements_file, pip_deps_dir, pypi_artifacts)
```
We can neatly stuff all that login under the conditional and get rid of those 2 lines at the end of the function.

tests/unit/package_managers/pip/test_main.py
ST3V1K: I'm not really sure what I should patch that would help me get rid of the string comparison

eskultety: What I meant was introducing a new commit (making it the 1st in the series to avoid the need for changes twice) and just drop all ^these string comparison asserts :), that's all.

hermeto/core/package_managers/pip/main.py
ST3V1K: I only read the docstring and did not realise `get_running_loop` is preferred for Python 3.10.
Changed it.

hermeto/core/package_managers/pip/main.py
ST3V1K: Done

hermeto/core/package_managers/pip/main.py
ST3V1K: Done

hermeto/core/package_managers/pip/main.py
ST3V1K: I haven't found a point in the code to log exactly when the processing of a requirement finishes; this would require a larger refactor. So I chose to log after all got processed.

> Bonus points - when you move the log line out of the loop then you can condense the loop to a single return comprehension.

Done

tests/unit/package_managers/pip/test_main.py
ST3V1K: Is this what you meant?

hermeto/core/package_managers/pip/main.py
ST3V1K: Done

hermeto/core/package_managers/pip/main.py
ST3V1K: Yes, this reads better.

tests/unit/package_managers/pip/test_main.py
```diff
@@ -517,20 +519,24 @@ def test_download_dependencies_pypi(
...
-        req = mock_requirement(
+        foo_req = mock_requirement(
             "foo", "pypi", download_line="foo==1.0", version_specs=[("==", "1.0")]
         )
         # match sdist hash, match wheel0 hash, mismatch wheel1 hash, no hash
         # for wheel2
-        req.hashes = ["sha256:abcdef", "sha256:defabc", "sha256:feebaa"]
+        foo_req.hashes = ["sha256:abcdef", "sha256:defabc", "sha256:feebaa"]
+
+        bar_req = mock_requirement(
```
derasdf, line 534: There's no test for what happens when one resolution fails during concurrent `asyncio.gather`. Worth one case with a mixed side_effect:

```python
mock_process_package_distributions.side_effect = [
    [foo_sdist_DPI],
    FetchError("PyPI query failed"),
]
with pytest.raises(FetchError, match="PyPI query failed"):
    pip._download_dependencies(rooted_tmp_path, req_file)
```

Without this, someone could add `return_exceptions=True` to the gather call and silently swallow failures.

hermeto/core/package_managers/pip/main.py
```diff
@@ -415,17 +416,11 @@ def _download_dependencies(
...
         if req.kind == "pypi":
-            download_infos: list[dict[str, Any]] = _process_pypi_req(
-                req,
-                requirements_file=requirements_file,
-                index_url=options["index_url"] or pypi_simple.PYPI_SIMPLE_ENDPOINT,
-                pip_deps_dir=pip_deps_dir,
-                binary_filters=binary_filters,
-            )
-            processed.extend(download_infos)
+            pypi_reqs.append(req)
```
derasdf, line 424: Adding `continue` here would let you drop the `if req.kind != "pypi"` guard at line 443:

```python
if req.kind == "pypi":
    pypi_reqs.append(req)
    continue
elif req.kind == "vcs":
    ...
```

Then the "Finished processing" log at the end of the loop applies to all non-pypi cases naturally, no re-check needed.

tests/unit/package_managers/pip/test_main.py
ST3V1K: Added a new test

hermeto/core/package_managers/pip/main.py
```diff
@@ -445,7 +436,14 @@ def _download_dependencies(
...
 
-        log.info("-- Finished processing requirement line '%s'\n", req.download_line)
+        log.info("-- Finished processing requirement line '%s'", req.download_line)
+
+    pypi_download_infos = _process_pypi_reqs(requirements_file, pip_deps_dir, pypi_artifacts)
+    processed.extend(pypi_download_infos)
+
+    for req in requirements_file.requirements:
+        if req.kind == "pypi":
+            log.info("-- Finished processing requirement line '%s'", req.download_line)
```
a-ovchinnikov: How does this message differ from one on L439?

tests/unit/package_managers/pip/test_main.py
```diff
@@ -728,6 +729,30 @@ def test_download_dependencies_pypi(
 
         # </check calls to checksum verification method>
 
+    @mock.patch("hermeto.core.package_managers.pip.main.process_package_distributions")
+    def test_download_dependencies_pypi_concurrent_resolution_error(
```
a-ovchinnikov: Test name should serve as a helper for a maintainer that runs in a test failure. Imagine you are new to this area of codebase, do a change, run UTs and see:
```
test_download_dependencies_pypi_concurrent_resolution_error      FAIL
```
Can you tell at a glance what exactly your change has broken? I afraid no, you would need to dive into the codebase and look at the test implementation. If, however, you add some words (which are already present in the docstring, by the way!) to the name it could become much more helpful. Now do the same thought experiment with:
```
test_pypi_download_dependencies_concurrent_resolution_propagates_error      FAIL
```
Now it should be immediately clear that an error got swallowed somewhere, no need to dive into yet another module! The change is small, but such things tend to accumulate, I firmly believe that every opportunity to improve ergonomics must be taken since it saves time and effort in the long run.

tests/unit/package_managers/pip/test_main.py
```diff
@@ -728,6 +729,30 @@ def test_download_dependencies_pypi(
...
+        )
+        req_file = mock_requirements_file(requirements=[foo_req, bar_req])
+
+        foo_sdist_DPI = mock_distribution_package_info("foo")
+        mock_process_package_distributions.side_effect = [
+            [foo_sdist_DPI],
+            FetchError("PyPI query failed"),
+        ]
+
+        with pytest.raises(FetchError, match="PyPI query failed"):
```
a-ovchinnikov: Please don't do string matches in tests unless you absolutely must. If exception type is not specific enough then it must be subclassed.

tests/unit/package_managers/pip/test_main.py
```diff
@@ -728,6 +729,30 @@ def test_download_dependencies_pypi(
...
+            pip._download_dependencies(rooted_tmp_path, req_file)
```
a-ovchinnikov: Out of curiosity, why do you expect that the error might not propagate? Does it have anything to do with the way our code is written?

hermeto/core/package_managers/pip/main.py
a-ovchinnikov, line 1: _Commit_ https://github.com/hermetoproject/hermeto/commit/ffac38e735ed86dc37d90c65f3d9c46a0f1f1298 : Commit message states that you are updating a test, half of the commit is about modifications to `main.py`. I think it should be split in two. Alternatively the message should be reworded to shift focus towards changes in resolution logic.

tests/unit/package_managers/pip/test_main.py
eskultety: To be fair I'm not really a fan of these overly long test names, because the names can be truncated in terminal output and more often than not one has to open the test code anyway, so I would not dwell on these long test names, it is pattern that started appearing at some point in the code base. I don't complain because that's bikeshedding, at the same time though I would not enforce this beyong simple "test_foo" & "test_foo_fail" or "test_foo_error". Besides, it's also quite likely that we can parametrize the failing test, so why the long name when the test ID is what is more important in that case.
I'd love to be pragmatic about this going forward, i.e. acknowledge naming is hard and we all struggle with it, follow only the very basic conventions "_pass" or "_fail" and if needed, provide a doc string for the test because the chances are the maintainer will have to open and inspect the test code anyway.

tests/unit/package_managers/pip/test_main.py
eskultety: Why didn't the stupid AI bot complain about this? It's literally in its instructions to flag string matching in tests.

hermeto/core/package_managers/pip/main.py
ST3V1K: It doesn't. It just logs at different times. For non-PyPI requirements it handles them in sequence, so it logs right after processing. But for PyPI requirements, it has to log differently, and the actual message won't change.

tests/unit/package_managers/pip/test_main.py
ST3V1K: The more I think about it, it really doesn't make sense that the error wouldn't propagate. It only tests if the `asyncio.gather` function works, which we shouldn't have to test.

hermeto/core/package_managers/pip/main.py
ST3V1K: Sorry, I have put the changes into the wrong commit. A bunch of the changes should have been in commit https://github.com/hermetoproject/hermeto/commit/7144079245db6eb3e7160f9a58a5948015b1d3b0

tests/unit/package_managers/pip/test_main.py
ST3V1K: Fair point on the naming, but based on this comment https://github.com/hermetoproject/hermeto/pull/1530#discussion_r3147129081 I would drop the test completely.

tests/unit/package_managers/pip/test_main.py
ST3V1K: Done

tests/unit/package_managers/pip/test_main.py
```diff
@@ -506,8 +506,10 @@ def test_malformed_hash(self, requirement_kind: str) -> None:
     @mock.patch.object(Path, "unlink")
     @mock.patch("hermeto.core.package_managers.pip.main.async_download_files")
     @mock.patch("hermeto.core.package_managers.pip.main._check_metadata_in_sdist")
+    @mock.patch("hermeto.core.package_managers.pip.main.log")
     def test_download_dependencies_pypi(
         self,
+        mock_log: mock.Mock,
```
eskultety: This mock is no longer used, a remnant from previous iterations perhaps?

tests/unit/package_managers/pip/test_main.py
eskultety: yes.

hermeto/core/package_managers/pip/main.py
eskultety: I think this logging might be a material for a larger refactor (maybe even a global one). The question is with concurrent handling how much value do these log lines add individually, so we mihgt want to rethink that as well.

eskultety (commented):
Looks good, 1 more nitpick from me, otherwise ACK.

tests/unit/package_managers/pip/test_main.py
ST3V1K: Yes, thank you.
Done

ST3V1K:
> I think this logging might be a material for a larger refactor (maybe even a global one). The question is with concurrent handling how much value do these log lines add individually, so we mihgt want to rethink that as well.

I agree, I wasn't a fan of having to log every pypi requirement like this, but I wanted to keep the same style of logging. Now I changed it to log only "Downloading N PyPI artifacts".

eskultety (approved):

derasdf (approved):

tests/unit/package_managers/pip/test_main.py
a-ovchinnikov: My improved example is precisely 80 symbols long (including spaces and a "FAIL" mark), most terminals nowdays are wide enough to accommodate that :)

a-ovchinnikov (approved):

PR #1537, opened 2026-04-25
fix(pip): gracefully skip sdist artifacts on checksum mismatch

### LINKED ISSUE #1523: bug(pip): FileNotFoundError crash after checksum mismatch for sdist packages
## Problem

When Hermeto downloads an sdist package and the actual downloaded file's
hash does not match the expected checksum, instead of raising a clean
`PackageRejected` error, Hermeto **crashes with an unhandled
`FileNotFoundError`**.

This is a security-relevant robustness issue. A corrupted network
response or a man-in-the-middle attack that serves a tampered artifact
causes Hermeto to crash completely rather than defensively rejecting the
artifact and reporting the mismatch clearly.

## Root Cause

In `hermeto/core/package_managers/pip/main.py`, the `_process_req`
function calls `_checksum_must_match_or_path_unlink` on line 282,
followed unconditionally by `_check_metadata_in_sdist` on line 284:

```python
if dpi.checksums_to_match:
    _checksum_must_match_or_path_unlink(dpi.path, dpi.checksums_to_match)
if dpi.package_type == "sdist":
    _check_metadata_in_sdist(dpi.path)  # called even if the file was deleted above!
```

The helper _checksum_must_match_or_path_unlink catches the PackageRejected exception, deletes the file, and returns normally without re-raising:

```python
def _checksum_must_match_or_path_unlink(path, checksum_info):
    try:
        must_match_any_checksum(path, checksum_info)
    except PackageRejected:
        path.unlink()  # file is deleted here
        log.warning(...)
        # exception is silently swallowed — returns None
```

Because the caller is never told the checksum failed, execution falls through to `_check_metadata_in_sdist`, which tries to open a file that no longer exists, crashing with a raw `FileNotFoundError`.

**Steps to Reproduce**

1. Add an sdist package to `requirements.txt` with a valid hash that matches what PyPI reports (so `checksums_to_match` is non-empty and `should_download()` returns `True`).
2. Simulate a corrupted or tampered download — the actual bytes on disk after download do not match the expected hash.
3. `must_match_any_checksum` raises `PackageRejected`.
4. The file is deleted by `path.unlink()`.
5. `_check_metadata_in_sdist(dpi.path)` tries to open the deleted file.
6. Python raises `FileNotFoundError` — Hermeto crashes.

**Actual Behavior**

```
FileNotFoundError: [Errno 2] No such file or directory:
'/output/deps/pip/flask-3.0.0.tar.gz'
```

**Expected Behavior**

```
PackageRejected: Checksum mismatch detected for flask-3.0.0.tar.gz.
The downloaded artifact does not match the expected hash and has been removed.
```

**Proposed Fix**

```python
Re-raise the exception from `_checksum_must_match_or_path_unlink` so the caller is aware the file was rejected:
def _checksum_must_match_or_path_unlink(path, checksum_info):
    try:
        must_match_any_checksum(path, checksum_info)
    except PackageRejected:
        path.unlink()
        raise  # re-raise so _check_metadata_in_sdist is never reached
```

Alternatively, guard `_check_metadata_in_sdist` with an existence check:
```python
if dpi.package_type == "sdist" and dpi.path.exists():
    _check_metadata_in_sdist(dpi.path)
```

**Security Impact**
In Konflux CI's air-gapped, SLSA-compliant build environment, a checksum mismatch is a critical security signal — it may indicate a tampered or corrupted artifact. The current behavior crashes Hermeto with an opaque error instead of surfacing a clean, auditable rejection, making it significantly harder to detect and respond to potential supply chain attacks.

Anujkumar9081:
### Problem
When a PyPI sdist package fails checksum validation, it is removed from disk by `_checksum_must_match_or_path_unlink`. However, its processing continued unconditionally to metadata extraction, leading to an unhandled `FileNotFoundError` crash.

### Solution
Following the design that an invalid artifact should be skipped rather than crashing the tool (or returning a silent success with missing files), this PR updates `_process_req` to cleanly return `None` when checksum validation fails and the file is unlinked. The callers are updated to filter out these skipped artifacts so they do not proceed to metadata extraction or SBOM generation.

Fixes #1523

### DIFF
--- a/hermeto/core/package_managers/pip/package_distributions.py
+++ b/hermeto/core/package_managers/pip/package_distributions.py
@@ -369,10 +369,23 @@ class WheelsFilter(BinaryPackageFilter):
         )
 
         wheel_py_versions = _parse_py_versions(interpreter)
-        compatible_py_version = self.py_version is None or any(
-            v == self.py_version or (v < self.py_version and abi in ("abi3", "none"))
-            for v in wheel_py_versions
+
+        # we compare strings because e.g. '312' is all we get from parsing wheel interpreter tag
+        is_same_major_ver = lambda other, this: str(other)[0] == str(this)[0]
+        is_versions_match = lambda other, this: other == this
+        is_compatible_abi = lambda abi: abi in ("abi3", "none")
+        is_version_superseded = lambda other, this: (other < this) and is_compatible_abi(abi)
+        is_compatible_candidate = lambda other, this: (
+            is_same_major_ver(other, this)
+            and (is_versions_match(other, this) or is_version_superseded(other, this))
         )
+
+        this = self.py_version
+        some_candidate_is_compatible = any(
+            is_compatible_candidate(v, this) for v in wheel_py_versions
+        )
+
+        compatible_py_version = self.py_version is None or some_candidate_is_compatible
 
         return compatible_py_impl and compatible_py_version
 
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -281,7 +281,7 @@ def _process_req(
     pip_deps_dir: RootedPath,
     download_info: dict[str, Any],
     dpi: DistributionPackageInfo | None = None,
-) -> dict[str, Any]:
+) -> dict[str, Any] | None:
     download_info["kind"] = req.kind
     download_info["requirement_file"] = str(requirements_file.file_path.subpath_from_root)
     download_info["missing_req_file_checksum"] = True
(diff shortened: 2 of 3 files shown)


eskultety (dismissed):
This needs rebasing on main properly, e.g. main already contains commit 1, so we'd only need commit 2 (after the rebase).

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -370,7 +370,8 @@ def _compatible_tag_interpreter(self, interpreter: str, abi: str) -> bool:
 
         wheel_py_versions = _parse_py_versions(interpreter)
         compatible_py_version = self.py_version is None or any(
-            v == self.py_version or (v < self.py_version and abi in ("abi3", "none"))
+            str(v)[0] == str(self.py_version)[0]
```
a-ovchinnikov: A drive-by comment: this will benefit from a couple of simple inline helpers so much!

```python
is_same_major_ver = lambda other, this: str(other)[0] == str(this)[0]
is_versions_match = lambda other, this: other == this
is_compatible_abi = lambda abi: abi in ("abi3", "none")
is_version_superseded = lambda other, this: (other < this) and is_compatible_abi(abi)
is_compatible_candidate = lambda other, this: is_same_major_version(v, this) and (is_versions_match(v, this) or is_version_superseded(v, this))

this = self.py_version
some_candidate_is_compatible = any(is_compatible_candidate(v, this) for v in wheel_py_versions)


is_compatible_py_version = self.py_version is None or some_candidate_is_compatible
```

Code is the same as yours, but more verbose and telling a reader what is going on. At the one-time price of having to type more your readers will not need to decipher a multiline logic expression within a multiline list comprehension within an assignment.

Anujkumar9081: Yeah this reads way better, I was trying to keep it compact but it ended up being hard to parse. Refactored it with the named helpers, pushed in ad7ac6e. Thanks for the suggestion!

eskultety:
@Anujkumar9081 We recently merged some work to speed up pip downloads, so there are some merge conflicts to resolve here.

eskultety:
@Anujkumar9081 there are still lint failures

eskultety:
@a-ovchinnikov I'm taking over this PR, I rebased this onto main and fixed the linter problem, please re-review.

slimreaper35 (approved):

a-ovchinnikov (approved):

eskultety:
Rebase onto fresh main with CI fixes.

PR #1540, opened 2026-04-27
gomod: Replace packaging.Version with semver.Version

slimreaper35:


hermeto/core/package_managers/gomod/go.py
```diff
@@ -436,7 +425,7 @@ def _select_toolchain(go_mod_file: RootedPath, installed_toolchains: Iterable[Go
     else:
         target_version = go_mod_toolchain_version
 
-    if target_version.to_language_version() > go_max_version.to_language_version():
+    if target_version > go_max_version:
```
slimreaper35, line 428: I hope this is still compatible. The default patch version is 0.

eskultety: No, it's not, e.g. with `go_max_version == 1.26` and `target_version == 1.26.3` the original code truncates both to `1.26` and the code continues as normal, your change results in an exception which is incorrect behaviour in this case.

hermeto/core/package_managers/gomod/go.py
```diff
@@ -97,25 +97,14 @@ def __init__(self, version_str: str) -> None:
...
+        parsed = semver.Version.parse(ver, optional_minor_and_patch=True)
+        super().__init__(parsed.major, parsed.minor, parsed.patch)
 
     @classmethod
     def max(cls) -> "GoVersion":
         """Instantiate and return a GoVersion object with the maximum supported version of Go."""
         return cls(cls.MAX_VERSION)
 
-    @cache
-    def to_language_version(self) -> version.Version:
```
eskultety, line 110: This must not be dropped, it's essential.

hermeto/core/package_managers/gomod/go.py
```diff
@@ -48,28 +48,28 @@ class _ParsedModel(pydantic.BaseModel):
...
     >>> v = GoVersion("go1.25.7-asdf-xyz")
-    >>> v.major, v.minor, v.micro
+    >>> v.major, v.minor, v.patch
     (1, 25, 7)
 
     >>> v = GoVersion("1.21")
-    >>> str(v.to_language_version())
-    '1.21'
+    >>> str(v)
+    '1.21.0'
```
eskultety, line 68: This behaviour introduces a regression:
```
# old code
In [2]: f"go{str(GoVersion("1.20"))}"
Out[2]: 'go1.20'

# new code
In [8]: f"go{str(GoVersion("1.20"))}"
Out[8]: 'go1.20.0'
```
^This is what we use to determine which base Go toolchain release to download if needed. Go <1.21 didn't use the micro version for initial releases, so downloading `go1.20.0` would fail as non-existent: https://go.dev/dl/

hermeto/core/package_managers/gomod/go.py
slimreaper35: Okay, so it seems there is no need to switch from packaging to semver.

eskultety: Agree, thanks for the experiment, let's keep things as they are for GoVersion.

PR #1549, opened 2026-05-04
CONTRIBUTING: Reshuffle and update

eskultety:
- reword some rules ever so slightly (minor polishing)
- format as a simple-to-follow DOs/DONTs list
- a couple of rules where essentially mentioned twice in different contexts
- explicitly forbid use of gitmojis

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -150,27 +150,33 @@ nox
 
 Observe the following guidelines when submitting a pull request for review
 
-* Write clear and informative *commit messages*. If you want to provide further context, use the PR's description
-* [Sign off on](https://developercertificate.org) all commits
+**Do:**
+
+* Write clear and informative *commit messages*, explain the "WHY"
+* Use PR description to provide further explanation or context about the PR topic
+* Sign off all your commits ([DCO](https://developercertificate.org))
 * Use your full name and a valid, deliverable email address for commit
   authorship. The notion of a "known identity" is open to interpretation. We simply require a real full name
-  rather than a pseudonym or handle (sorry, no anonymous contributions). The same applies to the
-  [DCO sign-off](https://developercertificate.org) which is required on all commits. If you set
+  rather than a pseudonym or handle (sorry, no anonymous contributions). The same applies to your DCO sign-off. If you set
   your `user.name` and `user.email` git configuration options, you can sign your commits
   automatically with `git commit -s`
-* Please use the PR's description to provide further explanation of the pull request's title
-* Split changes into multiple commits such that each commit addresses a clear and concise problem
-* Avoid PRs which are too large — split the work into multiple PRs if necessary
-* Amend existing commits rather than add new commits to fix issues introduced in the same pull request
-* Keep your branch up-to-date using `rebase` — we don't use merge commits
-* Verify the coding standards by running the configured linters
-* Ensure that every single commit passes CI — this is mandatory
-* Feel free to use Github comments to clarify implementation points and consider adding the same comments to the code
-* Feel free to use diagrams, sample code, or links to specific parts of external documentation — they are highly encouraged
-* Please respond to inline comments made in pull requests: it makes it easier to track what has been done and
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -150,27 +150,33 @@ nox
...
-* For trivial reviewers' requests it is ok to implement it, respond with "Done" and resolve the thread.
+* Split changes into several self-contained commits
+* Rebase frequently
+* Amend existing commits rather than adding fixup commits during review
+* Run linters & unit tests before opening a PR
+* Ensure every single commit passes CI
+
+**Don't:**
+
+* Use gitmojis anywhere — they only add noise
```
slimreaper35: Oh no! Why?

a-ovchinnikov (approved):

CONTRIBUTING.md
eskultety: Let me answer with a question - what's the added value of gitmojis other than being a gimmick that requires everyone who uses a terminal to either have the "right" terminal out of the box or at least make sure they have the necessary fonts installed on the system and the terminal configured correctly to not display mojibakes.
I tend to believe when it comes to SW ASCII is all professionals need to communicate & work efficiently.

slimreaper35: https://github.com/hermetoproject/hermeto/pull/1528/commits

What if gitmojis are only part of the PR name, not the actual commit message?

eskultety: Additionally, there might be other ticketing systems connected to a project's GH and so that's another friction area, can we confidently say that all of those systems will render those emojis correctly? The experience should be identical everywhere, besides, one needs to learn and remember a new emoji -> meaning mapping, isn't it actually faster to read 'fix'/'feature'/'docs' instead? :)

eskultety: > What if gitmojis are only part of the PR name, not the actual commit message?

Doesn't matter, same principle.

slimreaper35 (approved):

CONTRIBUTING.md
```diff
@@ -150,27 +150,33 @@ nox
...
+* Squash all changes into a single massive commit unless it is a simple or self-contained change
+
+**Tips:**
+
+* Feel free to use GitHub comments to clarify implementation points. Consider adding the same clarifications to the code as well
+* Diagrams, sample code, and links to specific parts of external documentation are highly encouraged
+* Respond to inline comments in pull requests — it makes it easier to track progress and speeds up the review process. "Done" is often enough to indicate that you have acted on a comment
```
derasdf: LGTM, but am i the only one who at this point gets triggered when i see an em dash ?

derasdf (approved):

CONTRIBUTING.md
eskultety: If you look at the original it was infested with em dashes throughout already ;), I only tried to remain consistent with my changes when I reshuffled and updated the file. 
That said, AI use within markdown files should be more than welcome IMO given the simple and straightforward use case:
- consistent text formatting
- natural English language (no grammar mistakes?)

eskultety: @derasdf would you prefer me to un-AIify the file? Although, that would take a while :)

derasdf: Hmmm, I would appreciate that, maybe just get rid of unnecessary word bloat in places like this, where it sounds really AI, and you can make AI word it better, for example just generated this:

```
GitHub comments are a good place to explain implementation decisions. If something in the code needs context, add it there too.
Diagrams, code snippets, and links to the relevant parts of external docs help a lot. 
When reviewers leave inline comments on your pull request, reply to them even if it's just 'Done' so everyone can see what's been addressed.
```

eskultety: @derasdf Done for the part I'm touching in this PR. There are other changes I did on my branch but given this already has 2 ACKs and the whole CONTRIBUTING overhaul wasn't my plan I'd keep that for a follow-up.
_Edit: For the parts in question I shortened the text as much as possible to convey the information and kept the imperative tone (who has time to read all the fluff :) )._

PR #1550, opened 2026-05-04
Dockerfile: Use COPY --link for multi-stage copies

eskultety:
COPY --link creates each layer independently of previous layers, which improves cache reuse [1] when the base image or earlier instructions change. From the docs:

    This also means you can easily rebase your images when the base
    images receive updates, without having to execute the whole build
    again.
    ...
    When using --link the COPY/ADD commands are not allowed to read any
    files from the previous state. This means that if in previous state
    the destination directory was a path that contained a symlink,
    COPY/ADD can not follow it.

Since none of the destination paths resolve through symlinks (/usr/local/bin, /usr/bin, etc.), this is safe to use here.

[1] https://docs.docker.com/reference/dockerfile/#copy

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,11 +48,11 @@ FROM base
 LABEL maintainer="Red Hat"
 
 # copy Go SDK and Node.js installation from official images
-COPY --from=golang /usr/local/go /usr/local/go
-COPY --from=node /usr/local/lib/node_modules/corepack /usr/local/lib/corepack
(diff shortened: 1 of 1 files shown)


eskultety:
Sigh, podman on the CI runners is too old (support added in podman 5.6.0):
- https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md
- https://github.com/containers/buildah/issues/4325#issuecomment-3251167324
- https://packages.ubuntu.com/noble/podman

Closing.

PR #1554, opened 2026-05-04
pnpm: Download tarballs for hermetic build

slimreaper35:


hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -10,13 +25,83 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
...
+    """
+    >>> _patch_url_to_point_to_proxy('https://registry.npmjs.org/foo/-/foo-1.0.0.tgz', 'http://proxy.com/npm/registry')
+    'http://proxy.com/npm/registry/foo/-/foo-1.0.0.tgz'
+    >>> _patch_url_to_point_to_proxy('https://registry.npmjs.org/foo/-/foo-1.0.0.tgz', 'http://proxy.com/npm/registry/')
+    'http://proxy.com/npm/registry/foo/-/foo-1.0.0.tgz'
+    """
+    str_proxy_url = str(proxy_url)
+    str_proxy_url = str_proxy_url if str_proxy_url[-1] == "/" else str_proxy_url + "/"
+    url_path = urlparse(url).path.removeprefix("/")
+    return str_proxy_url + url_path
```
eskultety: Shouldn't ^this be shared at least with the npm backend?

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -10,13 +26,85 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
...
+            files_with_auth[actual_url] = tarball_path
+        else:
+            files_without_auth[package.url] = tarball_path
+
+    async def download_all() -> None:
+        concurrency_limit = config.runtime.concurrency_limit
+        await asyncio.gather(
+            async_download_files(files_with_auth, concurrency_limit, auth=proxy_auth),
+            async_download_files(files_without_auth, concurrency_limit),
+        )
```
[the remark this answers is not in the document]
eskultety: Addressing this suggestion from the bot makes the current approach inconsistent with what the NPM backend does.

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -10,13 +25,83 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
...
+        if proxy_auth is not None and package.url.startswith(NPM_REGISTRY_URL):
+            actual_url = _patch_url_to_point_to_proxy(package.url, config.pnpm.proxy_url)
+            files_with_auth[actual_url] = tarball_path
+        else:
+            files_without_auth[package.url] = tarball_path
+
+    async def download_all() -> None:
+        concurrency_limit = config.runtime.concurrency_limit
+        await async_download_files(files_with_auth, concurrency_limit, auth=proxy_auth)
+        await async_download_files(files_without_auth, concurrency_limit)
```
eskultety: This means we download concurrently, but only per-batch, each batch is processed sequentially. I guess that's an acceptable trade-off. I was thinking if we could have a proportional concurrency limit based on the size of each batch, but that looking at my prototype it really looks like a micro-optimization given how elegant ^this is.
Suggestion: please make this a shared helper such that we could ditch the `_async_download_tar` counterpart in the npm backend due to the cryptic variable naming and named lambdas.

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -10,13 +25,83 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
...
+        proxy_auth = aiohttp.BasicAuth(config.pnpm.proxy_login, config.pnpm.proxy_password)
+
+    files_with_auth = {}
+    files_without_auth = {}
+    for package in packages:
+        tarball_path = deps_dir / package.tarball_name
+        if proxy_auth is not None and package.url.startswith(NPM_REGISTRY_URL):
+            actual_url = _patch_url_to_point_to_proxy(package.url, config.pnpm.proxy_url)
+            files_with_auth[actual_url] = tarball_path
+        else:
```
eskultety: nitpick: always make the `else` branch visually longer, it's easier to read.

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -10,13 +25,83 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
...
+    if config.pnpm.proxy_login is not None and config.pnpm.proxy_password is not None:
+        proxy_auth = aiohttp.BasicAuth(config.pnpm.proxy_login, config.pnpm.proxy_password)
+
+    files_with_auth = {}
+    files_without_auth = {}
+    for package in packages:
+        tarball_path = deps_dir / package.tarball_name
+        if proxy_auth is not None and package.url.startswith(NPM_REGISTRY_URL):
+            actual_url = _patch_url_to_point_to_proxy(package.url, config.pnpm.proxy_url)
+            files_with_auth[actual_url] = tarball_path
```
eskultety: This is inconsistent with the npm backend - `ProxyMixin` actually allows non-authenticated proxies, but what happens here is that if `proxy_auth is None` the item goes straight to the `files_without_auth` batch and skips proxy URL patching, that's a bug.

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -10,13 +25,83 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
...
+    return str_proxy_url + url_path
+
+
+def _prepare_lockfile_for_hermetic_build(
+    lockfile: PnpmLock, packages: list[PnpmPackage]
+) -> ProjectFile:
+    for package in packages:
+        data = lockfile.packages[package.id]
+        resolution = data.setdefault("resolution", {})
+        resolution["tarball"] = f"file://${{output_dir}}/deps/pnpm/{package.tarball_name}"
```
eskultety: Is this in-place modification desired? `_resolve_npm` deepcopies the lockfile before modifying so that it can use the original data for SBOM reporting. Since SBOM isn't being taken care of in this iteration it's not visible, but we'll need to fix it eventually before this goes live.

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -10,13 +25,83 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
...
+
+    for package in packages:
+        if package.integrity is not None:
+            must_match_any_checksum(
+                file_path=deps_dir / package.tarball_name,
+                expected_checksums=[ChecksumInfo.from_sri(package.integrity)],
+            )
+
+
+def _patch_url_to_point_to_proxy(url: str, proxy_url: ProxyUrl) -> str:
```
eskultety: Looks very similar to `_patch_url_to_point_to_a_proxy` in the npm backend. We should unify these before we forget about it.

eskultety (changes requested):

hermeto/core/package_managers/javascript/pnpm/main.py
slimreaper35: Oh, good catch!

hermeto/core/package_managers/javascript/pnpm/main.py
slimreaper35: Yes.

slimreaper35: Do we want to eventually move those shared utility functions to the new `javascript` directory?

eskultety: Eventually, yes, absolutely! For now I'd move it probably to the NPM backend instead of having 2 impls and import from there to remain consistent with the other couple of imports.

Or just open another PR to do that switch now since you're touching the code (or anyone else can do it for that matter, it's something that could get merged within a day) and rebase this on top of that.

hermeto/core/package_managers/javascript/pnpm/main.py
slimreaper35: Fixed.

hermeto/core/package_managers/javascript/pnpm/main.py
slimreaper35: Can't remember the details at this point.

hermeto/core/package_managers/javascript/pnpm/main.py
slimreaper35: > Suggestion: please make this a shared helper such that we could ditch the _async_download_tar counterpart in the npm backend due to the cryptic variable naming and named lambdas.

Done.

hermeto/core/package_managers/javascript/pnpm/main.py
slimreaper35: > Eventually, yes, absolutely! For now I'd move it probably to the NPM backend instead of having 2 impls and import from there to remain consistent with the other couple of imports.

Agree.

tests/unit/package_managers/npm/test_resolver.py
```diff
@@ -1081,9 +1085,12 @@ def test_npm_proxy_credentials_propagate_to_registry_hosts(
...
 
-    msg = "Not found credentials where they should be!"
-    for call in mock_async_download_files.mock_calls:
-        assert call.kwargs["auth"] is not None, msg
+    for call in mock_async_download_files.call_args_list:
+        files_to_download = call.args[0] if call.args else {}
+        if not files_to_download:
+            continue
+
+        assert call.kwargs.get("auth") is not None, "Found credentials where they should not be!"
```
eskultety: you have this backwards :), i.e. this fires when `auth is None` in which case the message doesn't match, a copy-paste of the hunk above perhaps?

hermeto/core/package_managers/javascript/pnpm/main.py
eskultety: We should keep being consistent with the NPM backend, we might forget doing that later on when we get to SBOM reporting, so better make the change now I think.

tests/unit/package_managers/javascript/pnpm/test_main.py
```diff
@@ -0,0 +1,143 @@
...
+    mock_parse_packages.return_value = [remote, local]
+
+    _resolve_pnpm_project(tmp_path, mock.Mock())
+    mock_download_resolved_packages.assert_called_once_with([remote], tmp_path)
+    mock_prepare_lockfile_for_hermetic_build.assert_called_once_with(mock.ANY, [remote])
+
+
+@mock.patch("hermeto.core.package_managers.javascript.pnpm.main.get_config")
+@mock.patch("hermeto.core.package_managers.javascript.pnpm.main.async_proxy_download")
+@mock.patch("hermeto.core.package_managers.general.async_download_files")
```
eskultety: This is called inside `async_proxy_download` so this mock is unnecessary.

tests/unit/package_managers/javascript/pnpm/test_main.py
```diff
@@ -0,0 +1,143 @@
...
+        files_with_auth={
+            "http://proxy.com/npm/registry/pkg/-/pkg-1.0.0.tgz": tmp_path / "pkg-1.0.0.tgz",
+        },
+        auth=aiohttp.BasicAuth("login", "password"),
+    )
+
+
+@mock.patch("hermeto.core.package_managers.javascript.pnpm.main.get_config")
+@mock.patch("hermeto.core.package_managers.javascript.pnpm.main.async_proxy_download")
+@mock.patch("hermeto.core.package_managers.general.async_download_files")
```
eskultety: same comment about the mock as above.

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -10,13 +27,81 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
...
+
+    asyncio.run(
+        async_proxy_download(
+            files_without_auth=files_without_auth, files_with_auth=files_with_auth, auth=auth
+        )
+    )
+
+    for package in packages:
+        if package.integrity is not None:
+            must_match_any_checksum(
```
eskultety, line 88: This doesn't appear to be covered by unit tests, shouldn't we cover this scenario with a single simple test?

tests/unit/package_managers/npm/test_resolver.py
slimreaper35: oops

hermeto/core/package_managers/javascript/pnpm/main.py
slimreaper35: added assert call for both `test_download_resolved_packages*`

eskultety (approved):

hermeto/core/package_managers/npm/resolver.py
```diff
@@ -78,17 +77,22 @@ def patch_url_to_point_to_proxy(url: str, proxy_url: ProxyUrl) -> str:
...
-async def _async_download_tar(files_to_download_list: list[dict[str, dict[str, Any]]]) -> None:
-    ftdl = [e for e in files_to_download_list if e]
-    if not ftdl:
-        return
-    # NOTE: when present proxy auth is the same for all packages accessible
-    # through a proxy.
-    auth = lambda ftd: next(iter(ftd.values()))["proxy_auth"]
-    ftd = lambda ftd: {it["fetch_url"]: it["download_path"] for it in ftd.values()}
-    adf = partial(async_download_files, concurrency_limit=get_config().runtime.concurrency_limit)
+async def async_proxy_download(
```
taylormadore: naming nit: Is there anything inherently proxy-related about the downloading here? It's just downloading passed URLs with and without auth

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -10,13 +28,81 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
...
+    lockfile: PnpmLock, packages: list[PnpmPackage]
+) -> ProjectFile:
+    for package in packages:
+        data = lockfile.packages[package.id]
+        resolution = data.setdefault("resolution", {})
+        resolution["tarball"] = f"file://${{output_dir}}/deps/pnpm/{package.tarball_filename}"
+
+    return ProjectFile(
+        abspath=lockfile.path,
+        template=yaml.safe_dump(copy.deepcopy(lockfile.data), sort_keys=False),
```
taylormadore: Shouldn't we copy this before we modify it rather than after?

hermeto/core/package_managers/javascript/pnpm/main.py
```diff
@@ -10,13 +28,81 @@ def fetch_pnpm_source(request: Request) -> RequestOutput:
...
+    auth = None
+    if proxy_login is not None and proxy_password is not None:
+        auth = aiohttp.BasicAuth(proxy_login, proxy_password)
+
+    files_with_auth = {}
+    files_without_auth = {}
+    for package in packages:
+        tarball_path = deps_dir / package.tarball_filename
+
+        if not package.url.startswith(NPM_REGISTRY_URL):  # non-registry packages
```
taylormadore: We have tests for the cases where the proxy is used (both with and without auth). Should we have a case for the normal path (non-proxy and URL deps)?

taylormadore (approved):

hermeto/core/package_managers/javascript/pnpm/main.py
eskultety: Ouch, great observation! I totally overlooked this in my review.

slimreaper35: 🌮

slimreaper35: fixed, ptal one more time

eskultety (approved):

PR #1557, opened 2026-05-07
Remove checksum verification in integration tests

### LINKED ISSUE #959: Revisit attributes of TestParameters class for integration tests

tnevrlka:
Some integration tests currently contain a file `fetch_deps_sha256sums.json` in their test data.
The file is supposed to contain checksums that get generated when the integration tests are run with the parameter 
`--hermeto-generate-test-data`.
This means that the checksums break whenever upstream dependencies change, requiring [regeneration](https://github.com/hermetoproject/hermeto/commit/d9d4463340791e35c033b0e8fdc8f09a2a011418).

The origin of the checksums is a commit [Verify checksum of external NPM dependencies](https://github.com/containerbuildsystem/cachito/commit/6acff84d41e4a2de241a1bfc103e006a1ded89ef).
The commit references a Red Hat-internal Jira issue with the following description:
>If a user has an npm dependency from an HTTP(S) URL in their package-lock.json, Cachito will use the "integrity" value of the dependency in the package-lock.json file to create a unique version for the dependency in the format of "<version>-external-<algorithm>-<checksum>" before uploading it to Nexus. The next time Cachito sees this dependency in a package-lock.json, it will use the one stored in Nexus rather than downloading it from the HTTP(S) URL again.
>
>Cachito needs to verify the "integrity" value matches the downloaded tarball before uploading it to Nexus to ensure that the dependency the user claims this is is accurate. This would prevent malicious code being uploaded to Nexus and being used by another Cachito user.
>
>In the highlighted code, "upload_non_registry_dependency" needs to be modified to accept an optional hash algorithm and the expected hash so that after the dependency is downloaded, the hash can be verified:
>https://github.com/release-engineering/cachito/blob/5ed7b31a619aa711d2f889af1766c9d8cdec8865/cachito/workers/pkg_managers/npm.py#L195-L210

Unlike Cachito, Hermeto does not upload dependencies to a shared artifact store, so there is no need for additional hash verification considering its maintenance burden.

Hash verification should be left to the package manager — Hermeto should not duplicate it, nor add it where the package manager itself does not.

At the moment, whether this check is executed depends on a test util parameter.
Out of 91 total TestParameters, 44 explicitly set the parameter to `False`. 
Some integration test cases contain `fetch_deps_sha256sums.json` files, but don't even check the checksums.

**AI disclaimer**: The author knows how to write em-dashes themselves, Claude Opus 4.6 was used for the investigation in the cachito codebase.

Closes #959


eskultety:
> Hermeto does not upload dependencies

Correction - we don't upload yet :-/ with the artifactory work we might need to start doing it at some point, but only for stuff not natively supported by the artifactory solutions, e.g. the generic backend.

Even if that is going to be the case, we'll add a new implementation.

eskultety (approved):
Nice analysis and background description! Nice commit split, kudos!

Can you please link https://github.com/hermetoproject/hermeto/issues/959 in the PR description as well as one of the commit messages?

a-ovchinnikov (approved):

tnevrlka:
> Can you please link https://github.com/hermetoproject/hermeto/issues/959 in the PR description as well as one of the commit messages?

Should be done.

Also found one more file in `tests/integration/test_data/yarn_e2e_workspace_focus/fetch_deps_sha256sums.json` my git-fu somehow missed ¯\\\_(ツ)\_/¯

PR #1563, opened 2026-05-13
feat(logging): add colored log levels and --log-color CLI option

MartinBasti:
Colorize log level names in terminal output (DEBUG=gray, INFO=blue, WARNING=orange, ERROR/CRITICAL=red). Only the level label is colored, not the entire message. Add --log-color option (auto/on/off) to control behavior; auto (default) enables color only when stderr is a TTY.

### DIFF
--- a/hermeto/interface/cli.py
+++ b/hermeto/interface/cli.py
@@ -184,13 +184,18 @@ def main(  # noqa: D103 -- docstring becomes part of --help message
         case_sensitive=False,
         help="Set log level.",
     ),
+    color: bool | None = typer.Option(
+        None,
+        "--color/--no-color",
+        help="Force colored log output on or off (default: auto - terminal is a TTY).",
+    ),
     mode: Mode = typer.Option(  # noqa: ARG001
         Mode.STRICT,
         "--mode",
         help="Treat input requirements violations as errors or warnings (may affect SBOM accuracy).",
     ),
 ) -> None:
-    setup_logging(log_level)
+    setup_logging(log_level, color=color)
     if config_file:
         config = set_config(config_file)
     else:
--- a/hermeto/interface/logging.py
+++ b/hermeto/interface/logging.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-only
 import enum
 import logging
+import os
 from collections.abc import Iterable
 from typing import Any, TextIO
 
@@ -18,18 +19,33 @@ LEVEL_COLORS: dict[int, str] = {
 RESET = "\033[0m"
 
 
+def _is_env_set(name: str) -> bool:
+    return bool(os.environ.get(name))
+
+
+def _resolve_color_from_env(stream: TextIO | None) -> bool:
+    """Determine whether to use color based on NO_COLOR/FORCE_COLOR env vars and TTY status.
+
+    Precedence: FORCE_COLOR > NO_COLOR > TTY auto-detection.
+    See https://no-color.org/ and https://force-color.org/
+    """
+    if _is_env_set("FORCE_COLOR"):
+        return True
+    if _is_env_set("NO_COLOR"):
+        return False
+    return stream is not None and stream.isatty()
+
+
 class ColoredFormatter(logging.Formatter):
     """Formatter that colorizes the log level name when the output stream supports color."""
 
     def __init__(self, fmt: str, stream: TextIO | None = None, color: bool | None = None) -> None:
         """Initialize the formatter, deciding whether to use color based on stream and mode."""
         super().__init__(fmt)
-        if color is True:
-            self._use_color = True
-        elif color is False:
-            self._use_color = False
+        if color is not None:
+            self._use_color = color
         else:
-            self._use_color = stream is not None and stream.isatty()
+            self._use_color = _resolve_color_from_env(stream)
 
     def format(self, record: logging.LogRecord) -> str:
         """Format the record, wrapping the level name in ANSI color codes if enabled."""
--- a/hermeto/interface/logging.py
+++ b/hermeto/interface/logging.py
@@ -2,11 +2,42 @@
 import enum
 import logging
 from collections.abc import Iterable
-from typing import Any
+from typing import Any, TextIO
 
 from hermeto.core.constants import Mode
 
 LOG_FORMAT = "%(asctime)s %(levelname)s %(message)s"
+
+LEVEL_COLORS: dict[int, str] = {
+    logging.DEBUG: "\033[90m",  # gray
+    logging.INFO: "\033[34m",  # blue
+    logging.WARNING: "\033[33m",  # orange/yellow
+    logging.ERROR: "\033[31m",  # red
+    logging.CRITICAL: "\033[1;31m",  # bold red
+}
+RESET = "\033[0m"
+
+
+class ColoredFormatter(logging.Formatter):
+    """Formatter that colorizes the log level name when the output stream supports color."""
+
+    def __init__(self, fmt: str, stream: TextIO | None = None, color: bool | None = None) -> None:
+        """Initialize the formatter, deciding whether to use color based on stream and mode."""
+        super().__init__(fmt)
+        if color is True:
+            self._use_color = True
+        elif color is False:
+            self._use_color = False
+        else:
+            self._use_color = stream is not None and stream.isatty()
+
+    def format(self, record: logging.LogRecord) -> str:
+        """Format the record, wrapping the level name in ANSI color codes if enabled."""
+        if self._use_color:
+            color = LEVEL_COLORS.get(record.levelno, "")
+            record = logging.makeLogRecord(record.__dict__)
+            record.levelname = f"{color}{record.levelname}{RESET}"
+        return super().format(record)
 
 
 class LogLevel(str, enum.Enum):
@@ -44,10 +75,14 @@ class EnforcingModeLoggerAdapter(logging.LoggerAdapter):
             self.error(msg, *args, **kwargs)
 
 
-def setup_logging(level: LogLevel, additional_modules: Iterable[str] = ()) -> None:
+def setup_logging(
+    level: LogLevel,
+    color: bool | None = None,
+    additional_modules: Iterable[str] = (),
+) -> None:
     """Set up logging. By default, enables only the application root logger."""
     handler = logging.StreamHandler()
-    handler.setFormatter(logging.Formatter(LOG_FORMAT))
+    handler.setFormatter(ColoredFormatter(LOG_FORMAT, stream=handler.stream, color=color))
 
     for module in ["hermeto", *additional_modules]:
         logger = logging.getLogger(module)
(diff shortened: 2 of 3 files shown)


MartinBasti:
image

MartinBasti:
Gemini asks for contradicting fixes, I'll ignore it

hermeto/interface/logging.py
```diff
@@ -8,6 +8,47 @@
...
+        elif color_mode == ColorMode.OFF:
+            self._use_color = False
+        else:
+            self._use_color = hasattr(stream, "isatty") and stream.isatty()
+
+    def format(self, record: logging.LogRecord) -> str:
+        """Format the record, wrapping the level name in ANSI color codes if enabled."""
+        if self._use_color:
+            color = LEVEL_COLORS.get(record.levelno, "")
+            record = logging.makeLogRecord(record.__dict__)
```
slimreaper35, line 54: I think this line is redundant.

hermeto/interface/logging.py
```diff
@@ -8,6 +8,47 @@
...
+    AUTO = "auto"
+    ON = "on"
+    OFF = "off"
+
+
+class ColoredFormatter(logging.Formatter):
+    """Formatter that colorizes the log level name when the output stream supports color."""
+
+    def __init__(
+        self, fmt: str, stream: object = None, color_mode: ColorMode = ColorMode.AUTO
```
slimreaper35: Looking at the type of `handler.stream`, the type should be `TextIO`.

hermeto/interface/logging.py
```diff
@@ -8,6 +8,47 @@
 
 LOG_FORMAT = "%(asctime)s %(levelname)s %(message)s"
 
+LEVEL_COLORS: dict[int, str] = {
+    logging.DEBUG: "\033[90m",  # gray
+    logging.INFO: "\033[34m",  # blue
+    logging.WARNING: "\033[33m",  # orange/yellow
+    logging.ERROR: "\033[31m",  # red
+    logging.CRITICAL: "\033[31m",  # red
```
slimreaper35: nitpick: Maybe the critical could be purple?

hermeto/interface/logging.py
MartinBasti: it's not, it prevents changing original object.
In case we are ok with updating original object that propagates outside of the function, I can remove it

hermeto/interface/logging.py
MartinBasti: fixed

hermeto/interface/logging.py
MartinBasti: I added bold red for critical

hermeto/interface/logging.py
slimreaper35: What is the tradeoff if we modify the objects?

MartinBasti: in case of parallel processing or exporting logs into different reporters, it could do some nasty side effects, like putting colors into reporter where you don't want it

tnevrlka (approved):

hermeto/interface/cli.py
```diff
@@ -184,13 +184,19 @@ def main(  # noqa: D103 -- docstring becomes part of --help message
         case_sensitive=False,
         help="Set log level.",
     ),
+    log_color: ColorMode = typer.Option(
+        ColorMode.AUTO.value,
+        "--log-color",
```
eskultety: nitpick: I think `--color` as a name might be slightly more consistent with the industry, at least based on the mention in https://force-color.org/ and the list of SW that somehow deals with colors in somewhat uniform manner (the actual list is better viewed at the counterpart website: https://no-color.org/).
Bonus points for making this a boolean option in Typer -> `--color/--no-color`.

tests/unit/test_logging.py
```diff
@@ -0,0 +1,85 @@
...
+    def test_on_mode_forces_color_without_tty(self) -> None:
+        fmt = ColoredFormatter(FMT, stream=FakeNonTTY(), color_mode=ColorMode.ON)
+        result = fmt.format(_make_record(logging.INFO))
+        assert f"\033[34mINFO{RESET}" in result
+
+    def test_off_mode_disables_color_with_tty(self) -> None:
+        fmt = ColoredFormatter(FMT, stream=FakeTTY(), color_mode=ColorMode.OFF)
+        result = fmt.format(_make_record(logging.ERROR))
+        assert "\033[" not in result
+        assert "ERROR" in result
```
eskultety: I believe ^these color testing tests + `test_auto_mode_no_stream` below test the same thing so don't benefit from descriptive test names that much IMO and could be collapsed to a single coloring parametrized test, e.g.:
```
 @pytest.mark.parametrize(
      "color_mode, stream, has_color",
      [
          (ColorMode.AUTO, FakeTTY(), True),
          (ColorMode.OFF, FakeTTY(), False),
          (ColorMode.ON, FakeNonTTY(), True),
          (ColorMode.AUTO, FakeNonTTY(), False),
          (ColorMode.AUTO, None, False),
      ],
  )
  def test_color_mode(self, color_mode, stream, expect_color):
     ...
```

hermeto/interface/logging.py
```diff
@@ -2,12 +2,53 @@
...
+    AUTO = "auto"
+    ON = "on"
+    OFF = "off"
+
+
+class ColoredFormatter(logging.Formatter):
+    """Formatter that colorizes the log level name when the output stream supports color."""
+
+    def __init__(
+        self, fmt: str, stream: Optional[TextIO] = None, color_mode: ColorMode = ColorMode.AUTO
```
eskultety: nitpick: we no longer use `Optional`, since Python 3.10 (our min) these become `TextIO | None`

hermeto/interface/logging.py
```diff
@@ -44,10 +85,20 @@ def error_or_warn(self, msg: str, *args: Any, enforcing_mode: Mode, **kwargs: An
...
+    color_mode: ColorMode = ColorMode.AUTO,
+    additional_modules: Iterable[str] = (),
+) -> None:
+    """Set up logging. By default, enables only the application root logger.
+
+    Args:
+        level: The minimum log level to output.
+        color_mode: Controls whether log output is colorized.
+        additional_modules: Other logger modules to configure.
+    """
```
eskultety: I think the signature is self-explanatory in this case, we don't need these listed again :) .

hermeto/interface/logging.py
```diff
@@ -2,12 +2,53 @@
...
+RESET = "\033[0m"
+
+
+class ColorMode(str, enum.Enum):
+    """Control whether log output uses colors."""
+
+    AUTO = "auto"
+    ON = "on"
+    OFF = "off"
+
```
eskultety: I think Typer would give us None if neither was set which would translate to auto, so I'm not sure we need the color mode enum at all since AUTO is handled in an `else` clause in the formatter?

eskultety (commented):
Since you're adding this, I stumbled upon https://no-color.org/ and https://force-color.org/ would you mind adding support for these as well? Seems trivial enough and we'll this endeavour a mission accomplished :) .

eskultety:
> Since you're adding this, I stumbled upon https://no-color.org/ and https://force-color.org/ would you mind adding support for these as well? Seems trivial enough and we'll this endeavour a mission accomplished :) .

@MartinBasti I retract on ^this comment of mine - it would be nice to have support for ^this, but we can convert it to a "my first issue" for Hermeto to have something for new contributors. `--color/--no-color` therefore remains my only substantial request.

hermeto/interface/logging.py
MartinBasti: IIRC I added it because a linter was yelling at me that options are not documented

hermeto/interface/cli.py
MartinBasti: changed

tests/unit/test_logging.py
MartinBasti: updated

hermeto/interface/logging.py
MartinBasti: fixed

hermeto/interface/logging.py
MartinBasti: updated

MartinBasti:
> > Since you're adding this, I stumbled upon https://no-color.org/ and https://force-color.org/ would you mind adding support for these as well? Seems trivial enough and we'll this endeavour a mission accomplished :) .
> 
> @MartinBasti I retract on ^this comment of mine - it would be nice to have support for ^this, but we can convert it to a "my first issue" for Hermeto to have something for new contributors. `--color/--no-color` therefore remains my only substantial request.

It was implemented in separate commit

hermeto/interface/cli.py
```diff
@@ -181,13 +184,18 @@ def main(  # noqa: D103 -- docstring becomes part of --help message
         case_sensitive=False,
         help="Set log level.",
     ),
+    color: bool | None = typer.Option(
+        None,
+        "--color/--no-color",
+        help="Force colored log output on or off (default: auto-detect terminal).",
```
eskultety: nitpick:
`s/auto-detect terminal/auto - terminal is a TTY`

hermeto/interface/logging.py
eskultety: You sure?

 ```bash
$ git diff
diff --git hermeto/interface/logging.py hermeto/interface/logging.py
index b4702d3b..dbb681ef 100644
--- hermeto/interface/logging.py
+++ hermeto/interface/logging.py
@@ -96,13 +96,7 @@ def setup_logging(
     color: bool | None = None,
     additional_modules: Iterable[str] = (),
 ) -> None:
-    """Set up logging. By default, enables only the application root logger.
-
-    Args:
-        level: The minimum log level to output.
-        color: Controls whether log output is colorized (True=on, False=off, None=auto).
-        additional_modules: Other logger modules to configure.
-    """
+    """Set up logging. By default, enables only the application root logger."""
     handler = logging.StreamHandler()
     handler.setFormatter(ColoredFormatter(LOG_FORMAT, stream=handler.stream, color=color))

$ nox -s lint
nox > Running session lint
nox > Reusing existing virtual environment at .nox/lint.
nox > python -m pip install --no-deps -r requirements-extras.txt
nox > ruff check hermeto tests noxfile.py
nox > ruff format --check --diff hermeto tests noxfile.py
nox > mypy --install-types --non-interactive hermeto tests noxfile.py
nox > Session lint was successful.
```

eskultety (approved):
Just nitpicks, implementation is solid -> ACK.

hermeto/interface/logging.py
MartinBasti: ah right, it was gemini not linter https://github.com/hermetoproject/hermeto/pull/1563#discussion_r3234433352

eskultety: It's hard to capture this in natural language instructions, because it requires judgement... we tend to ignore stylistic comments from Gemini, but keeping the config just in case it catches a real case where we'd benefit from more docs - it is what it is.

hermeto/interface/cli.py
MartinBasti: updated

hermeto/interface/logging.py
MartinBasti: updated

PR #1564, opened 2026-05-14
A design doc for implementing UV support.

harshitbansal184507:
Initial draft for UV support implementation design. Some sections are incomplete and additional details will be added before review.

### DIFF
--- a/docs/design/uv.md
+++ b/docs/design/uv.md
@@ -1,0 +1,1071 @@
+# Adding uv support to Hermeto
+
+## Table of Content
+
+
+- [Adding uv support to Hermeto](#adding-uv-support-to-hermeto)
+  - [Table of Content](#table-of-content)
+  - [Background](#background)
+  - [How dependencies are specified in `pyproject.toml`](#how-dependencies-are-specified-in-pyprojecttoml)
+  - [`uv.toml`](#uvtoml)
+  - [`uv.lock`](#uvlock)
+  - [Other dependency types](#other-dependency-types)
+  - [Workspace](#workspace)
+  - [Binary filters](#binary-filters)
+    - [Cases](#cases)
+  - [Markers](#markers)
+  - [Build process in uv](#build-process-in-uv)
+    - [Build constraints](#build-constraints)
+  - [uv support in Hermeto](#uv-support-in-hermeto)
+    - [`uv sync` and `uv build`: why both are required](#uv-sync-and-uv-build-why-both-are-required)
+    - [Prefetch (fetch-deps)](#prefetch-fetch-deps)
+      - [1) Runtime dependencies (from `uv.lock`)](#1-runtime-dependencies-from-uvlock)
+      - [2) Build-Time dependencies](#2-build-time-dependencies)
+      - [Checksum handling by source kind](#checksum-handling-by-source-kind)
+      - [Lockfile validation](#lockfile-validation)
+    - [Lockfile rewriting (inject-files)](#lockfile-rewriting-inject-files)
+      - [Registry packages](#registry-packages)
+      - [Git packages](#git-packages)
+      - [URL packages](#url-packages)
+      - [Other source kinds](#other-source-kinds)
+      - [Path and URL format conventions](#path-and-url-format-conventions)
+    - [Build environment (generate-env)](#build-environment-generate-env)
+      - [`UV_OFFLINE`](#uv_offline)
+      - [`UV_FIND_LINKS`](#uv_find_links)
+      - [`UV_FROZEN`](#uv_frozen)
+      - [`UV_NO_BINARY`](#uv_no_binary)
+      - [`UV_NO_BINARY_PACKAGE`](#uv_no_binary_package)
+    - [Example](#example)
+    - [SBOM](#sbom)
+      - [Sources](#sources)
+      - [Example purls](#example-purls)
+      - [Component properties](#component-properties)
+      - [Local sources: path, directory, editable, virtual](#local-sources-path-directory-editable-virtual)
+    - [Out of scope](#out-of-scope)
+    - [Appendix: rejected approaches](#appendix-rejected-approaches)
+      - [Using uv natively for fetch](#using-uv-natively-for-fetch)
+      - [Using a uv flat index instead of URL rewriting](#using-a-uv-flat-index-instead-of-url-rewriting)
+    - [References](#references)
+
+## Background
+
+[uv](https://docs.astral.sh/uv/) is a Python package and project manager written
+in Rust. It replaces the traditional Python packaging files (`setup.py`,
+`requirements.txt`, `setup.cfg`, `Pipfile`) with a `pyproject.toml` + `uv.lock` model.
+
+A typical uv-managed project has the following structure:
+
+```
+├── .python-version
+├── pyproject.toml
+├── uv.lock
+└── src/
+    └── main.py
+```
+
+Where `pyproject.toml` declares dependencies and project configuration, and
+`uv.lock` is the universal lockfile that pins resolved dependencies for all
+platforms.
+
+## How dependencies are specified in `pyproject.toml`
+
+uv allows dependencies to come from
+[different sources](https://docs.astral.sh/uv/concepts/projects/dependencies/).
+They're declared in
+[`[project.dependencies]`](https://docs.astral.sh/uv/concepts/projects/dependencies/),
+with
+[`[tool.uv.sources]`](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-sources)
+extending the standard dependency tables with alternative dependency
+sources used during development. The examples below show how each type appears in
+`pyproject.toml`.
+
+<details>
+  <summary>default registry (PyPI)</summary>
+
+  ```toml
+  [project]
+  dependencies = [
+      "requests>=2.28",
+      "flask==3.0.0",
+  ]
+  ```
+</details>
+<details>
+  <summary>git</summary>
+
+  ```toml
+  [project]
+  dependencies = ["flask"]
+
+  [tool.uv.sources]
+   # pin to a commit
+flask = { git = "git+https://github.com/pallets/flask", rev = "954f5684e4841aad84a8eec7ace7b81a0d3f6831" }
+  # pin to a tag
+  flask = { git = "https://github.com/pallets/flask", tag = "3.0.0" }
+  # pin to a branch
+  flask = { git = "https://github.com/pallets/flask", branch = "main" }
+  # subdirectory (if package isn't in repo root)
+  flask = { git = "https://github.com/pallets/flask", subdirectory = "libs/flask" }
+  ```
+  >uv treats any git-clonable URL the same way regardless of forge - GitHub, GitLab, Bitbucket etc. 
+
+  >The git+ scheme prefix from `pyproject.toml` is stripped during resolution; uv.lock stores a plain URL with the ref type as a query parameter and the resolved commit in the fragment (source = { git = "https://.../?<tag|branch|rev>=<value>#<sha>" }).
+</details>
+
+<details>
+  <summary>url</summary>
+
+  ```toml
+  [tool.uv.sources]
+  httpx = { url = "https://files.pythonhosted.org/.../httpx-0.28.1.tar.gz" }
+  ```
+</details>
+
+
+<details>
+  <summary>path / directory / editable (local filesystem sources)</summary>
+
+   All three are declared with `path = ...` in `pyproject.toml`. They differ
+  on two axes: what `path` points at (directory or file artifact), and
+  whether `editable = true` is set.
+ 
+  | `path` points at      | `editable = true` | uv source(in uv.lock) for this case |
+  |-----------------------|-------------------|-----------------------|
+  | A directory           | not set           | `directory`           |
+  | A directory           | set               | `editable`            |
+  | A `.whl` or `.tar.gz` | not set           | `path`                |
+  | A `.whl` or `.tar.gz` | set               | rejected by uv        |
+ 
+  ---
+ 
+  **`directory`** - `path` points at a source tree (a folder containing
+  `pyproject.toml`).
+  ```toml
+  [tool.uv.sources]
+  my-lib = { path = "./libs/my-lib" }
+  ```
+--- 
+**`editable`** - `path` points at a source tree, and `editable = true` is
+set. uv installs the package so that edits to the source directory take
+effect immediately, without reinstall.
+
+```toml
+[tool.uv.sources]
+my-lib = { path = "./libs/my-lib", editable = true }
+```
+>**Only valid for directories** 
+  ---
+  **`path`** - `path` points at a pre-built artifact file (`.whl` or
+  `.tar.gz`). No build step runs for wheels; sdists go through PEP 517.
+ 
+  ```toml
+  [tool.uv.sources]
+  my-lib = { path = "./dist/my_lib-0.1.0-py3-none-any.whl" }
+  ```
+ 
+  ---
+ 
+  Workspace members behave like `editable` by default - see
+  [Workspace](#workspace).
+
+
+</details>
+<details>
+  <summary>Workspace member</summary>
+
+```toml
+  [project]
+  dependencies = ["my-lib"]
+
+  [tool.uv.sources]
+  my-lib = { workspace = true }
+```
+>For more details see - [Workspaces](#workspace) 
+</details>
+
+
+<details>
+  <summary>platform-specific source</summary>
+
+```toml
+  [project]
+  dependencies = ["httpx"]
+
+  [tool.uv.sources]
+  # only use this source on macOS, fall back to PyPI on other platforms
+  httpx = { git = "https://github.com/encode/httpx", tag = "0.27.2", marker = "sys_platform == 'darwin'" }
+```
+</details>
+
+<details>
+  <summary>multiple sources</summary>
+
+```toml
+  [project]
+  dependencies = ["httpx"]
+
+  [tool.uv.sources]
+  # different source per platform
+  httpx = [
+    { git = "https://github.com/encode/httpx", tag = "0.27.2", marker = "sys_platform == 'darwin'" },
+    { git = "https://github.com/encode/httpx", tag = "0.24.1", marker = "sys_platform == 'linux'" },
+  ]
+```
+</details>
+
+## `uv.toml`
+`uv.toml` is a uv-only [configuration file](https://docs.astral.sh/uv/concepts/configuration-files/). It holds the same settings
+that would otherwise go under the `[tool.uv]` section of `pyproject.toml`,
+nothing more. The difference is structural: for example, `pyproject.toml`
+requires the `[tool.uv]` prefix since it's a shared file for multiple
+tools. `uv.toml` is uv-only, so the prefix is dropped.
+
+```toml
+# pyproject.toml          # uv.toml
+[tool.uv]                 # (no prefix)
+resolution = "lowest"     resolution = "lowest"
+
+[[tool.uv.index]]         [[index]]
+url = "https://..."       url = "https://..."
+```
+
+> **Note:** `[tool.uv.sources]` cannot go in `uv.toml`. It is project
+> metadata, not configuration, and requires a `[project]` table to be
+> meaningful. uv enforces this explicitly:
+> *"The `sources` field is not allowed in a `uv.toml` file. `sources` is only
+> applicable in the context of a project, and should be placed in a
+> `pyproject.toml` file instead."*
+>See: https://github.com/astral-sh/uv/issues/11632
+
+## `uv.lock`
+
+The `uv.lock` file follows the TOML format.The top of the file holds resolver-wide settings; below that comes an array of `[[package]]` tables, one per locked package. Within each package, the `source` form determines which sub-keys appear.  The format is versioned — this design targets
+`version = 1` and all its future revisions.
+
+> Each version of uv only understands one lockfile format version, and major
+> version changes are completely incompatible. Changes in Revision are forward and backward compatible. See [uv source][uv-lockfile-version].
+
+[uv-lockfile-version]: https://docs.astral.sh/uv/concepts/resolution/#lockfile-versioning
+
+> `uv.lock` is a fixed file name. At the time of writing uv offers no way to
+> configure it or supply an alternative
+> ([astral-sh/uv#6830](https://github.com/astral-sh/uv/issues/6830)). Hermeto therefore relies on the `uv.lock` name and looks for no other.
+
+<details>
+  <summary><strong>Top-level fields (before any <code>[[package]]</code>)</strong></summary>
+
+  Written in this fixed order. Each is emitted only when its condition holds.
+
+  - **`version =` 1** - *always present*. Lockfile-format major version.
+  - **`revision =` 3** - *present only if `> 0`*. Backwards-compatible sub-version of `version`.
+  - **`requires-python =` ">=3.10"** - *always present*. Supported Python version range as a string.
+  - **`resolution-markers =` ["python_full_version < '3.11'", ...]** - *present only when the resolution forked*. 
+  - **`supported-markers =` ["sys_platform == 'linux'", ...]** - *present only if user-declared in by*`[tool.uv.environments]`(limits the set of environments that uv will consider when resolving dependencies).
+  - **`required-markers =` ["sys_platform == 'linux'", ...]** - *present only if user-declared by* [tool.uv.required-environments]`.
+  - **`conflicts =` [[{ package = "foo", extra = "a" }, { package = "foo", extra = "b" }]]** - *present only if non-empty*. Declares mutually-exclusive dependency groups.
+  - **`[options]`** - *present only if any sub-field differs from its default*. 
+      >Sub-keys: `resolution-mode`, `prerelease-mode`, `fork-strategy`, `exclude-newer`, `exclude-newer-span`, `exclude-newer-package`.
+  - **`[manifest]`** - *present only if any sub-field is non-empty*. 
+      >Sub-keys: `members = ["my-workspace", "pkg-a", "pkg-b"]`, `requirements`, `constraints`, `overrides`, `excludes`, `build-constraints`, `dependency-groups`, `dependency-metadata`.
+</details>
+<details><summary><strong><code>[[package]]</code> table</strong></summary>
+
+```toml
+[[package]]
+name = "..."                       # always present
+version = "..."                    # always present
+source = { ... }                   # always; exactly one of:
+                                     #registry  | git       | url       | path
+                                     #directory | editable  | virtual
+
+resolution-markers = [...]         # only if this package is locked at multiple versions
+dependencies = [...]               # # only if the package has resolved dependencies
+sdist = { ... }                    # varies by source type 
+wheels = [...]                     # varies by source type 
+
+## One sub-table per extra declared in [project.optional-dependencies].
+# <extra-name> is the user-chosen name of the extra (e.g. "plot", "test").
+[package.optional-dependencies]    # only if optional dependencies are present in package
+<extra-name> = [...]
+
+# One sub-table per dependency group.
+# <group-name> is the user-chosen name of the group. uv treats `dev` as the
+# default group (synced unless --no-dev or --no-default-groups is passed).
+[package.dev-dependencies]         # only if non-empty
+<group-name> = [...]
+
+[package.metadata]                 # only for mutable sources (not registry, not git)
+requires-dist = [...]              
+provides-extras = [...]            
+
+[package.metadata.requires-dev]    # # only if the package declares dependency groups
+<group-name> = [...]
+```
+
+**How `pyproject.toml` declarations map to `uv.lock` source kinds** :
+
+
+
+| `pyproject.toml` declaration                                 | `uv.lock` source kind |
+|--------------------------------------------------------------|-----------------------|
+| Plain dependency (e.g. `"flask>=3"`)         | `registry`            |
+| `{ git = "..." }`                                            | `git`                 |
+| `{ url = "..." }`                                            | `url`                 |
+| `{ path = "./dir" }`                                         | `directory`           |
+| `{ path = "./dir", editable = true }`                        | `editable`            |
+|  `{ path = "./<wheel-or-sdist-file>" }`   | `path`                |
+| `{ workspace = true }`                                       | `editable`            |
+| Root project, with `[build-system]`                          | `editable`            |
+| Root project, without `[build-system]`                       | `virtual`             |
+
+*Examples for  source types :*
+
+<details>
+  <summary>registry package</summary>
+
+  ```toml
+  [[package]]
+  name = "anyio"
+  version = "4.13.0"
+  source = { registry = "https://pypi.org/simple" }
+  sdist = { url = "...", hash = "sha256:...", size = 99999, upload-time = "..."}
+  wheels = [
+      { url = "...", hash = "sha256:...", size = 99999, upload-time = "..."},
+  ]
+  ```
+</details>
+
+<details>
+  <summary>git package</summary>
+
+  ```toml
+  [[package]]
+  name = "flask"
+  version = "3.0.0"
+  source = { git = "https://github.com/pallets/flask?rev=7ef2946f#7ef2946f..." }
+  ```
+</details>
+
+<details>
+  <summary>url package</summary>
+
+  ```toml
+  [[package]]
+  name = "httpx"
+  version = "0.28.1"
+  source = { url = "https://files.pythonhosted.org/.../httpx-0.28.1.tar.gz" }
+  sdist = { hash = "sha256:..." }
+  ```
+</details>
+<details>
+  <summary>path</summary>
+
+  ```toml
+  [[package]]
+  name = "example-package"
+  version = "1.0.0"
+  source = { path = "dist/example_package-1.0.0-py3-none-any.whl" }
+  wheels = [
+      { filename = "example_package-1.0.0-py3-none-any.whl", hash = "sha256:..." },
+  ]
+  ```
+</details>
+
+<details>
+  <summary>directory</summary>
+
+  ```toml
+  [[package]]
+  name = "example-package"
+  version = "1.0.0"
+  source = { directory = "packages/example-package" }
+  ```
+</details>
+
+<details>
+  <summary>editable</summary>
+
+  ```toml
+  [[package]]
+  name = "example-package"
+  version = "1.0.0"
+  source = { editable = "packages/example-package" }
+  ```
+</details>
+
+<details>
+  <summary>virtual</summary>
+
+  ```toml
+  [[package]]
+  name = "root-project"
+  version = "0.1.0"
+  source = { virtual = "." }
+  dependencies = [
+      { name = "example-package" },
+  ]
+  ```
+</details>
+
+
+</details>
+
+
+
+## Other dependency types
+uv supports two distinct mechanisms for grouping dependencies beyond the main
+`[project.dependencies]` list. They are **not the same feature** and serve
+different purposes:
+
+- **[Optional dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#optional-dependencies)** ("extras") - defined under
+  `[project.optional-dependencies]`. They are part of
+  the published package metadata (recorded as `Provides-Extra` and
+  `Requires-Dist` in the built distribution) and are installable by
+  downstream consumers via `pip install pkg[extra-name]`.
+
+- **[Dependency groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups)** - defined under the top-level
+  `[dependency-groups]` table per [dependency-groups](https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups). They are **not**
+  included in any built distribution; they exist only in `pyproject.toml`
+  for project-local use (linting, testing, docs, etc.). uv treats the
+  `dev` group as a default group and installs it during `uv sync` unless
+  `--no-dev` or `--no-default-groups` is passed.
+
+```toml
+[project.optional-dependencies]
+plot = ["matplotlib"]
+
+[dependency-groups]
+dev = ["pytest", "ruff"]
+```
+
+Both are resolved into `uv.lock` the same way as regular dependencies, with
+marker / group metadata indicating where they belong.
+
+
+
+## Workspace
+
+ A workspace in uv is a collection of one or more packages, called `workspace members`, that are managed together. 
+ The workspace root is itself a member, and other members are declared explicitly under `[tool.uv.workspace]` using members (required) and exclude (optional) keys, both of which accept glob patterns. All included members share one uv.lock file and common uv commands will affect all included packages within the workspace.
+
+<details>
+  <summary>Sample project with workspaces structure</summary>
+
+```
+my-app/
+├── packages/
+│   ├── pkg-a/
+│   │   ├── pyproject.toml
+│   │   └── src/
+│   │       └── pkg_a/
+│   │           └── __init__.py
+│   └── pkg-b/
+│       ├── pyproject.toml
+│       ├── uv.lock
+│       └── src/
+│           └── pkg_b/
+│               └── __init__.py
+├── pyproject.toml
+├── uv.lock
+└── src/
+    └── my_app/
+        └── __init__.py
+```
+ 
+Here `my-app` is the workspace root and `pkg-a` is an included member and `pkg-b` is an excluded member.
+Whether `pkg-b` is included or excluded isn't visible from the directory
+tree, only the root `pyproject.toml`'s `[tool.uv.workspace].exclude`
+entry decides that. The nested `pyproject.toml` and `uv.lock` inside
+`pkg-b/` just mean it's also a self-contained project on its own.
+ 
+</details>
+<details>
+  <summary>root <code>pyproject.toml</code> for a project with workspaces</summary>
+
+  ```toml
+  [project]
+  name = "my-app"
+  version = "0.1.0"
+  requires-python = ">=3.10"
+  dependencies = ["pkg-a", "tqdm>=4,<5"]
+ 
+  [tool.uv.sources]
+  pkg-a = { workspace = true }
+ 
+  [tool.uv.workspace]
+  members = ["packages/*"]
+  exclude = ["packages/pkg-b"]
+  ```
+ 
+</details>
+
+<details>
+<summary>how workspace members are recorded in <code>uv.lock</code></summary>
+
+Workspaces add two things to the lockfile on top of the format documented above: a top-level `[manifest]` table listing the workspace members, and
+a `[[package]]` entry per member:
+
+**1. `[manifest].members`** - lists every included workspace member, including the root. Excluded packages (e.g. those matched by `[tool.uv.workspace].exclude`) are absent.
+
+```toml
+[manifest]
+members = [
+    "my-app",
+    "pkg-a",
+]
+```
+
+**2. Each member gets a `[[package]]` entry with an editable source**.The workspace root points to `"."`, members point to their subfolder relative to the root.
+
+```toml
+[[package]]
+name = "my-app"
+version = "0.1.0"
+source = { editable = "." }
+dependencies = [
+    { name = "pkg-a" },
+    { name = "tqdm" },
+]
+
+[package.metadata]
+requires-dist = [
+    { name = "pkg-a", editable = "packages/pkg-a" },
+    { name = "tqdm", specifier = ">=4,<5" },
+]
+
+[[package]]
+name = "pkg-a"
+version = "0.1.0"
+source = { editable = "packages/pkg-a" }
+dependencies = [
+    { name = "requests" },
+]
+
+[package.metadata]
+requires-dist = [{ name = "requests" }]
+```
+Other dependencies  appear as `[[package]]` entries, workspaces don't change how those are recorded.
+</details>
+<details>
+  <summary>tool.uv.sources cascading</summary>
+  Any `tool.uv.sources` defined at the root level applies to all members automatically, unless a specific member overrides it in its own `tool.uv.sources`:
+ 
+  ```toml
+  # root pyproject.toml — applies to all members
+  [tool.uv.sources]
+  pkg-a = { workspace = true }
+  requests = { git = "https://github.com/psf/requests", branch = "main" }
+  ```
+ 
+  Now every member gets `requests` from git by default. A member can override this:
+ 
+  ```toml
+  # packages/pkg-a/pyproject.toml — overrides root for this member only
+  [tool.uv.sources]
+  requests = { registry = "https://pypi.org/simple" }
+  ```
+ 
+  That is member's own `tool.uv.sources` takes priority, root's `tool.uv.sources` is the fallback for everything else.
+ 
+</details>
+
+## Binary filters
+
+By default Hermeto fetches sdists only. The `binary` input field opts specific packages into wheels. Field shape matches Hermeto's existing pip binary filter.
+ 
+`packages` selects which packages may install as wheels. The other fields constrain which `wheels[].url` entries in `uv.lock` qualify, matched by filename tags the same way pip does.
+ 
+ ```json
+{
+  "type": "uv",
+  "binary": {
+    "packages": "numpy,scipy",
+    "arch": "x86_64",
+    "os": "linux",
+    "py_version": 312,
+    "py_impl": "cp",
+    "abi": ":all:",
+    "platform": null
+  }
+}
+```
+
+### Cases
+
+Two environment variables control uv's binary/sdist split at build time:
+
+- `UV_NO_BINARY=true` — forces sdist for every package, including any listed in packages.
+- `UV_NO_BINARY_PACKAGE=<pkg1> <pkg2> ...` — forces sdist only for the named packages, leaving the rest free to use wheels. 
+
+| `binary` input                  | Prefetched                                                   | Build-time env                                  |
+|---------------------------------|--------------------------------------------------------------|-------------------------------------------------|
+| absent                          | sdist for every package                                      | `UV_NO_BINARY=true`                             |
+| `packages = :all:`              | sdist + matching wheels for every package                    | (neither set)                                   |
+| `packages = <list>`             | sdist for non-listed, matching wheels for listed             | `UV_NO_BINARY_PACKAGE = <non-listed deps>`      |
+ 
+ 
+- `UV_NO_BINARY=true` forces sdist for *every* package, including listed ones — ignoring the filter.
+- Setting neither lets uv pick wheels for any package with a `wheels[]` entry in `uv.lock`. Hermeto only prefetched sdists for non-listed packages, so the build fails on a missing wheel.
+[`UV_NO_BINARY_PACKAGE`](https://docs.astral.sh/uv/reference/environment/#uv_no_binary_package) forbids wheels per-package. Listing every non-filter package gives the intended split without rewriting `uv.lock`.
+ 
+
+## Markers
+
+uv.lock is universal — it contains entries for all platforms and Python
+versions the project supports. Each dependency edge can carry a [PEP 508
+marker](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers) such as
+`sys_platform == 'win32'`, restricting it to specific environments.
+
+> TODO: confirm final approach for marker evaluation. Current proposal:
+> introduce `evaluate_markers: bool = False` field on `UvPackageInput` to keep
+> the default behaviour fetch everything.
+
+
+
+
+## Build process in uv
+uv builds distributions through the [PEP 517](https://peps.python.org/pep-0517/)
+process, which splits the work between two tools:
+
+`build frontend` - the command the user invokes. For uv that's [`uv build`](https://docs.astral.sh/uv/concepts/projects/build/#using-uv-build). `uv build` invokes the declared backend in an isolated environment per PEP 517, producing a source distribution (.tar.gz) and a binary distribution (.whl) in dist/ by default. The --sdist / --wheel flags limit the output to one format.
+
+`build backend` - the tool that actually turns the project's source tree into a wheel or sdist, declared by the project's pyproject.toml [build-system] table (e.g. hatchling, setuptools, flit-core). **uv works with any PEP 517-compliant backend and also ships its own, [uv_build](https://docs.astral.sh/uv/concepts/build-backend/#using-the-uv-build-backend)**.
+
+The build-time dependencies are
+**not recorded in `uv.lock`**  ([astral-sh/uv#5190](https://github.com/astral-sh/uv/issues/5190)).This is why Hermeto must prefetch build dependencies separately via
+`uv export` + `pybuild-deps` (see [**2) Build dependencies**](#2-build-dependencies)). All build dependencies live in `{output_dir}/deps/uv/` and are exposed to uv via
+`UV_FIND_LINKS` during the build.
+
+
+### Build constraints
+
+Build constraints restrict which versions of build-time dependencies are used
+when uv builds a package from source. They are a guard, including a
+package here does **not** install it; the constraint only applies if that
+package is actually required as a direct or transitive build-time dependency.
+([source](https://docs.astral.sh/uv/pip/compile/#adding-build-constraints))
+
+Supported ways of specifying build constraints are listed here
+
+<details>
+  <summary><strong><code>pyproject.toml</code></strong></summary>
+
+  ```toml
+  [tool.uv]
+  build-constraint-dependencies = ["setuptools>=70", "setuptools!=72.0.0"]
+  ```
+</details>
+
+<details>
+  <summary><strong><code>uv.toml</code></strong> (same setting, no <code>[tool.uv]</code> prefix)</summary>
+
+  ```toml
+  build-constraint-dependencies = ["setuptools>=70", "setuptools!=72.0.0"]
+  ```
+</details>
+
+<details>
+  <summary><strong><code>uv.lock</code></strong> (recorded under <code>[manifest]</code> for lockfile invalidation)</summary>
+
+  ```toml
+  [manifest]
+  build-constraints = [{ name = "setuptools", specifier = ">=70,!=72.0.0" }]
+  ```
+</details>
+
+<details>
+  <summary><strong>Text file</strong>` - <code>requirements.txt</code>like file passed via CLI</summary>
+
+  The filename is user-defined; uv does not look for it automatically.
+
+  ```bash
+  uv lock --build-constraints build-constraints.txt
+  uv sync --build-constraints build-constraints.txt
+  uv build --build-constraint build-constraints.txt   # note: singular flag here
+  ```
+</details>
+
+
+
+The equivalent environment variable is `UV_BUILD_CONSTRAINTS`, which accepts a
+space-separated list of file paths.
+
+## uv support in Hermeto
+
+For uv, Hermeto's three phases do the following:
+
+- **[fetch-deps](#prefetch-fetch-deps)** - download every artifact listed in `uv.lock`, plus the
+ build dependencies `uv.lock` does not record, into
+  `{output_dir}/deps/uv/`.
+- **[inject-files](#lockfile-rewriting-inject-files)** - rewrite every artifact URL in `uv.lock` to point into
+  `{output_dir}/deps/uv/` instead of the package origin
+- **[generate-env](#build-environment-generate-env)** - produce an environment file with the variables the
+  user's build step needs to run offline against `{output_dir}/deps/uv/`.
+
+The rejected alternative  is in the [Appendix](#appendix-rejected-approach-using-uv-natively).
+
+### `uv sync` and `uv build`: why both are required
+
+Hermeto has to provide data sufficient for running two uv commands:
+
+- `uv build` produces the project's own wheel/sdist. It does **not** install
+  the project's runtime dependencies into any environment.
+- `uv sync` installs the runtime dependencies from `uv.lock` into
+  the target environment.
+
+The two commands read from different inputs:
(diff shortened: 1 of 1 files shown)


docs/design/uv.md
```diff
@@ -0,0 +1,658 @@
...
+resolution = "lowest"     resolution = "lowest"
+
+[[tool.uv.index]]         [[index]]
+url = "https://..."       url = "https://..."
+```
+
+> **Note:** `[tool.uv.sources]` cannot go in `uv.toml`. It is project
+> metadata, not configuration, and requires a `[project]` table to be
+> meaningful. uv enforces this explicitly:
+> *"The `sources` field is not allowed in a `uv.toml` file. `sources` is only
```
eskultety: Can you attach a URL to this reference?

harshitbansal184507: Information around uv.toml is quite scattered across the uv docs, so I could not find a single direct reference for this. I found this while testing locally, but this should be helpful: https://github.com/astral-sh/uv/issues/11632

eskultety: That will do, please incorporate and you can resolve this thread.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
+# Draft Proposal: Adding uv support to Hermeto
```
a-ovchinnikov: I think you can safely drop `Draft Proposal:`

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+Where `pyproject.toml` declares dependencies and project configuration, and
+`uv.lock` is the universal lockfile that pins resolved dependencies for all
+platforms.
+
+### Glossary
+- **pyproject.toml**: standard Python tool-agnostic project specification file, introduced by [PEP 518](https://peps.python.org/pep-0518/) and extended by [pyproject.toml specification](https://packaging.python.org/en/latest/specifications/pyproject-toml/#pyproject-toml-spec). Declares project metadata, dependencies, build system, and tool-specific settings under namespaced tables (`[tool.uv]`).
+- **uv.lock**: universal cross-platform lockfile produced by `uv lock`.
+- **uv.toml**: a uv-only configuration file.
+- **source**: where uv resolved a package from — registry, git, url, path,
+  editable, directory, or virtual.
```
a-ovchinnikov: What is `editable` in the context of source? It should be either explained or excluded, excluded is preferable unless it is crucial to the definition.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+  [project]
+  dependencies = [
+      "requests>=2.28",
+      "flask==3.0.0",
+  ]
+  ```
+</details>
+
+<details>
+  <summary>git</summary>
```
a-ovchinnikov, line 94: Can GitLab be used in the same manner?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+  <summary>url</summary>
+
+  ```toml
+  [tool.uv.sources]
+  httpx = { url = "https://files.pythonhosted.org/.../httpx-0.28.1.tar.gz" }
+  ```
+</details>
+
+<details>
+  <summary>path / editable / directory</summary>
```
a-ovchinnikov: Did you mean that path is the same as editable and the same as directory?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+  ```toml
+  [tool.uv.sources]
+  my-lib = { path = "./libs/my-lib", editable = true }
+  ```
+</details>
+<details>
+  <summary>workspace member</summary>
```
a-ovchinnikov: Here you use the concept of workspaces without introducing it. It should be either briefly mentioned before or a link to a section should be provided.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+The `uv.lock` file follows the TOML format. Each resolved dependency is stored
+as a `[[package]]` entry. The format is versioned; this design targets
+`version = 1` and all its revisions.
+
+> Each version of uv only understands one lockfile format version, and major
+> version changes are completely incompatible. Changes in Revision are forward and backward compatible. See [uv source][uv-lockfile-version].
+
+
+[uv-lockfile-version]: https://github.com/astral-sh/uv/blob/main/crates/uv-resolver/src/lock/mod.rs
```
a-ovchinnikov: Please use permalinks when referencing implementations: this way your future readers will have a better chance of seeing the same document as you have cited: https://github.com/astral-sh/uv/blob/5aba1bde10b8a397ea4faff5f83a9504345f4dac/crates/uv-resolver/src/lock/mod.rs

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+  <summary>url package</summary>
+
+  ```toml
+  [[package]]
+  name = "httpx"
+  version = "0.28.1"
+  source = { url = "https://files.pythonhosted.org/.../httpx-0.28.1.tar.gz" }
+  sdist = { hash = "sha256:..." }
+  ```
+</details>
```
a-ovchinnikov, line 359: This list is shorter than the list of package types supported by `pyproject.toml`, does it mean that several package types map into same type of lock-file entries?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+
+
+### Other dependency types
+
+uv supports [optional
+dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#optional-dependencies)
+(extras) and [dependency
```
a-ovchinnikov: Did you mean that optional dependencies are supported _as_ dependency groups? (I.e. that dependency groups are the mechanism for supporting optionals.) Also note, that you mention PEP-735 but link to uv docs.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+dev = ["pytest", "ruff"]
+```
+
+These are resolved into `uv.lock` the same way as regular dependencies, with
+markers/group metadata indicating where they belong.
+
+
+## uv.toml
+
+Both carry identical configuration, the difference is nesting.
```
a-ovchinnikov: Please add some brief introduction for the sake of a reader: "both" implies that you are planning to compare something to something else that has just been introduced, here you effectively loop back to `pyproject.toml`. I'd say mentioning it in some way would make reading the design a much smoother process. Something along the lines of `uv.toml is anoter way of specifying dependencies for uv-basedproject. It has the same capabilities as pyproject.toml, however the format is slightly different: ...`

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+`pyproject.toml` requires the `[tool.uv]` prefix since it can be a shared file for multiple tools;
+`uv.toml` is uv-only so the prefix is dropped.
+
+```toml
+# pyproject.toml          # uv.toml
+[tool.uv]                 # (no prefix)
+resolution = "lowest"     resolution = "lowest"
+
+[[tool.uv.index]]         [[index]]
```
a-ovchinnikov: What is `index` and how does it relate to `sources`?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+
+## Building distributions
+
+uv supports two related build concepts:
+
+- **Build frontend** — `uv build` itself, which invokes the build backend
+  per [PEP 517](https://peps.python.org/pep-0517/).
+- **Build backend** —
+  uv supports all build backends (as specified by PEP 517), but also provides a native build backend (uv_build).
```
a-ovchinnikov: What is a build backend? Is it something that an implementer of this design should be concerned about? Is default implementation good enough to be used for building? Are there any indicators that someone might want alternative build backends? At the very least a short definition of a build backend is necessary.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+> TODO: confirm final approach for marker evaluation. Current proposal:
+> introduce `evaluate_markers: bool = False` field on `UvPackageInput` to keep
+> the default behaviour fetch everything.
+
+
+
+
+## Building distributions
+
+uv supports two related build concepts:
```
a-ovchinnikov: I am sorry, what is a build concept?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+- **Build frontend** — `uv build` itself, which invokes the build backend
+  per [PEP 517](https://peps.python.org/pep-0517/).
```
a-ovchinnikov: A general suggestion: if there is some relevant bit of information please include it in this document, it should be self contained to the point when a reader does not need to follow external links unless they want to learn about extra details. Here, for example, you mention a new concept: a build backend which has not been introduced yet and which is immediately followed by a link to a PEP. Having a link is good, mentioning a new concept is good, but does it mean that you are not going to describe this new concept and a user must follow the link to make sense of what comes next?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+> **TODO**: use uv_build backend to see what it does with the tool.uv.sources table
+
+## Build constraints
+
+Build constraints restrict which versions of build-time dependencies are used
+when uv builds a package from source. They are a guard, including a
+package here does **not** install it; the constraint only applies if that
+package is actually required as a direct or transitive build-time dependency.
+([source](https://docs.astral.sh/uv/pip/compile/#adding-build-constraints))
```
a-ovchinnikov, line 646: Down below you list a number of examples of how build constraints could be applied. Please add some linkage between this paragraph and examples below, for instance `Supported ways of specifying build constraints are listed below:`

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+
+
+## Prefetching
+
+Hermeto's prefetch phase populates `{output_dir}/deps/uv/` as a flat directory
+containing every artifact needed for both the runtime and build-from-source phases.
+Two sub-steps run in sequence.
+
+### Step A - Runtime artifacts from `uv.lock`
```
a-ovchinnikov: I suggest the following substitution:
s/Runtime artifacts from `uv.lock`/Collect runtime artifacts from `uv.lock`/

Also WDYT about calling them just dependencies instead of runtime artifacts?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+For every `[[package]]` entry in `uv.lock`, dispatch by source kind:
+
+| Source kind | Action |
+|-------------|--------|
+| `registry` | Download artifact at `sdist.url` and every `wheels[].url` |
+| `git` | Clone repo at the resolved commit, archive as tarball |
+| `url` | Download artifact at `source.url` |
+| `path`, `editable`, `directory`, `virtual` | Skip — already on local filesystem |
+
+### Why rewrite registry URLs instead of configuring a uv flat index
```
a-ovchinnikov: Have you already mentioned that a rewrite is necessary?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+  hardcoded `https://files.pythonhosted.org/...` URLs in `sdist.url` /
+  `wheels[].url`, and uv fetches those exactly. The flat index does not
+  override lockfile-pinned URLs.
+
+Registry-URL rewriting in the lockfile is therefore required. Once those
+URLs point at local files, the flat index becomes redundant and is not
+configured.
+
+
+### Step B — Build dependencies
```
a-ovchinnikov: This is a bit confusing, are you suggesting that something must be built?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+
+**`uv export` regenerates `uv.lock`.** So the flow should be:
+
+1. `uv export` (regenerates lockfile)
+2. `pybuild-deps compile`
+3. Fetch both runtime and build-time dependencies
+4. Lockfile rewrite
+
+After step 4, no uv command that regenerates the lockfile should run.
```
a-ovchinnikov: The question is, can we somehow not use `uv` at all?

a-ovchinnikov (commented):
Please find some initial comments inline.

docs/design/uv.md
eskultety: Not at the time of writing this design, `pybuild-deps` would need to be adjusted first to understand `uv`. I'd like to see that done, but that would hold up this work. I think `uv export -> requirements.txt -> pybuild-deps` sounds like an acceptable workaround.
Alternatively, we could keep `uv` support in Hermeto experimental due to ^this and wait until `pybuild-deps` catches up so that we can get rid of `uv` interactions before we make this fully supported.

eskultety: @a-ovchinnikov there's also lockfile validation which the design doesn't mention, but following our Yarn backends we have the option to run `uv` to validate `uv.lock` and stop if updates would be needed.

docs/design/uv.md
harshitbansal184507: yes, but will it be used? (never used gitlab) .
I have seen GitLab used majorly for private PyPI registries, and there the usage is usually this way:
```
[project]
dependencies = ["my-pkg"]

[tool.uv.sources]
my-pkg= { index = "gitlab-pypi" }

[[tool.uv.index]]
url = "https://gitlab.com/api/v4/projects/<id>/packages/pypi/simple"
name = "gitlab-pypi"
```
Are private registries in scope?

docs/design/uv.md
harshitbansal184507: I'll add a link to the section

docs/design/uv.md
harshitbansal184507: That's not what I meant. Not at all. I didn't mention the other source types here simply because there was no rewrite for them. I'll include those as well.
Also, I haven't mentioned the dependencies field and the package.metadata sub table (which appears only for mutable source types). Should I add those?

docs/design/uv.md
harshitbansal184507: Sorry for this😅. I just copied it from the docs.

docs/design/uv.md
a-ovchinnikov: My concern was more about not mixing git itself and git forges than about GitLab specifically. I would replaced "git" with "git forge", depending of whether GitHub has any special treatment or not I would have considered adding a short note about that. Private registries will be in scope eventually, but would they require any special handling though?

docs/design/uv.md
a-ovchinnikov: If any types should be excluded for any reason then it is ok to not list them as long as there is some clear explanation why they are not listed. Here if I understood your intent correctly you would want to tell your reader that different package types could be found in a lock file; different package types would require different treatment: some will be left intact, while others would require a rewrite (more on that in implementation section [link]). Here only the types of entries which will be updated are listed.

When writing a document it is important to keep your audience in mind -- you likely already know more about the ecosystem you are describing than most of your future readers, your task is to concisely explain it to them. This requires some practice and almost always several iterations. What I find useful is stopping now and then and asking myself: "Which assumptions about reader's knowledge have I made so far? If I were new to this subject would I be able to figure this section out on my own without digging through additional documentation? Why do I think that something is obvious, do I know something else that makes it obvious but is not obvious without this specific knowledge?" Answering these questions helps me find what is missing from a document. A good design document would be able to help an engineer familiar with Hermeto  and generally familiar with the topic to become sufficiently proficient in it to understand what has to be done to support it in scope of this project.

docs/design/uv.md
a-ovchinnikov: No problem! We'll have to polish the document to be as easy to digest as practical. This usually means that you, the author, will need to lead your reader through the text by providing the right amount of connective sentences and phrases: anything that just appears out of nowhere without any introduction confuses readers, and you do not want them to be confused. Don't worry, it is a process, we'll get there.

docs/design/uv.md
a-ovchinnikov: Fair enough, however do we plan to carry `uv` around for just that or maybe compute a hash or two on our own? Or just trust package maintainers to keep the lock file in sync with `uv.toml` (which, judging from past experiences,  _will_ result in issues now and then).

docs/design/uv.md
harshitbansal184507: got it. Regarding any special handling, I don't think any `uv` specific handling for private registries would be needed during prefetch, though I'd research more about it after I complete more important parts of this document.

docs/design/uv.md
harshitbansal184507: In workspaces, every workspace member appears in lockfile as an editable source like this 

```
[[package]]
name = "pkg-a"
version = "0.1.0"
source = { editable = "pkg-a" }
dependencies = [...]
```

a-ovchinnikov: Thank you, the question remains, whether this belongs right here in the glossary or is not immediately relevant and can be omitted (I think it can be omitted, however you could add something like "(i.e. effectively a workspace member)" right after "editable").

harshitbansal184507: Ok, I would emit it here, that's a better option. Detailed explanation would be there under workspaces section.

docs/design/uv.md
eskultety: IOW this should be renamed to something like `Build-time dependencies`

docs/design/uv.md
eskultety: Instead I'd suggest to make this a Draft PR as long as the TODO sections are being still figured out.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+Where `pyproject.toml` declares dependencies and project configuration, and
+`uv.lock` is the universal lockfile that pins resolved dependencies for all
+platforms.
+
+### Glossary
+- **pyproject.toml**: standard Python tool-agnostic project specification file, introduced by [PEP 518](https://peps.python.org/pep-0518/) and extended by [pyproject.toml specification](https://packaging.python.org/en/latest/specifications/pyproject-toml/#pyproject-toml-spec). Declares project metadata, dependencies, build system, and tool-specific settings under namespaced tables (`[tool.uv]`).
+- **uv.lock**: universal cross-platform lockfile produced by `uv lock`.
+- **uv.toml**: a uv-only configuration file.
+- **source**: where uv resolved a package from — registry, git, url, path,
```
eskultety: same here, I don't think we need a glossary for `source`, you'll take care of it in the section where you discuss dependencies to a more detail.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+  editable, directory, or virtual.
+- **marker**: [PEP 508](https://peps.python.org/pep-0508/#environment-markers)
+  environment marker constraining a dependency to specific platforms / Python
+  versions.
+
```
eskultety: Reading through the rest of the doc I think we can safely drop this whole glossary, you discuss all of it to a good detail in sections below, we don't need this "preamble" I believe.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+    { git = "https://github.com/encode/httpx", tag = "0.24.1", marker = "sys_platform == 'linux'" },
+  ]
+```
+</details>
+
+### uv.lock
+
+The `uv.lock` file follows the TOML format. Each resolved dependency is stored
+as a `[[package]]` entry. The format is versioned; this design targets
+`version = 1` and all its revisions.
```
eskultety: nitpick: maybe worth an em dash?
_The format is versioned -- this design targets only `version = 1` and all its future revisions_

docs/design/uv.md
eskultety: Yes, the word "both" comes a bit out of nowhere here.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+## Binary filters
+
+> TODO:
+
+## Markers
+
+uv.lock is universal — it contains entries for all platforms and Python
+versions the project supports. Each dependency edge can carry a [PEP 508
+marker](https://peps.python.org/pep-0508/#environment-markers) such as
+`sys_platform == 'win32'`, restricting it to specific environments.
```
eskultety, line 617: I'll just place https://github.com/hermetoproject/hermeto/issues/1570 here -- we may need to consult markers when binary filtering is in place to make sure we'd not attempt to fetch something that isn't available on the index and fail with a fatal error (if there's an sdist, we'd fetch that one anyway based on our current logic).

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+- **Build frontend** — `uv build` itself, which invokes the build backend
+  per [PEP 517](https://peps.python.org/pep-0517/).
+- **Build backend** —
+  uv supports all build backends (as specified by PEP 517), but also provides a native build backend (uv_build).
+
+`uv build` produces a source distribution (`.tar.gz`) followed by a binary
+distribution (`.whl`) from that sdist, placing both in `dist/` by default.
+Subcommands `--sdist` / `--wheel` limit the output to one format. See
+[Building distributions](https://docs.astral.sh/uv/concepts/projects/build/#building-distributions).
```
eskultety: To what extent do we care about this information for our purposes and purposes of this doc?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+`uv build` works offline under the same environment Hermeto provides for
+`uv sync`:
+
+```
+UV_OFFLINE=1
+UV_FIND_LINKS={output_dir}/deps/uv
+```
+
+`uv build --no-binary` produces both sdist and
+wheel without network access, using only the prefetched build dependencies.
```
eskultety: This section is confusing, it mentions both `uv build` and `uv sync`. We'd probably need a clearer description or separation in this doc, that is indeed something we care about to a great extent, so rewording this section is desired.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+uv uses build isolation by default. When a package is built from sdist, uv
+creates an isolated environment and installs the packages listed in that
+sdist's `[build-system].requires` (e.g. `setuptools`, `hatchling`,
+`flit-core`) before invoking the backend. These build-time dependencies are
+**not recorded in `uv.lock`**
+
+This is why Hermeto must prefetch build dependencies separately via
+`uv export` + `pybuild-deps` (see [Step B](#step-b--build-dependencies)). All
+build deps live in `{output_dir}/deps/uv/` and are exposed to uv via
+`UV_FIND_LINKS` during the build.
```
eskultety: The content here is valid, but the section name (PEP 517...) is IMO confusing in context and purpose of the doc, so I'd suggest to moving this somewhere under `Prefetching` and rename this to `Build-time dependencies` (you already have a section with that name, consider merging those 2 or reshuffling the content to make sense for the reader.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+> **Note on reproducibility for distributed wheels**: when a project has
+> `[tool.uv.sources]` overrides (git/url), these are stripped from the
+> published wheel's metadata by default. Consumers of the wheel will resolve
+> `Requires-Dist: flask` from PyPI, not from the original git source. This
+> is uv's intended behavior and a property of the PEP 621 / PyPI ecosystem.
+
+> **TODO**: use uv_build backend to see what it does with the tool.uv.sources table
+
+## Build constraints
```
eskultety: Same here, this is valuable content that should be paired with the above within a `Build-time dependencies` section.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+sdist's `[build-system].requires` (e.g. `setuptools`, `hatchling`,
+`flit-core`) before invoking the backend. These build-time dependencies are
+**not recorded in `uv.lock`**
+
+This is why Hermeto must prefetch build dependencies separately via
+`uv export` + `pybuild-deps` (see [Step B](#step-b--build-dependencies)). All
+build deps live in `{output_dir}/deps/uv/` and are exposed to uv via
+`UV_FIND_LINKS` during the build.
+
+### `uv build`
```
eskultety: This section is misplaced in between the one above and the one below. Those 2 should be strictly paired together and this one about building moving somewhere else.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+**Text file** — a `requirements.txt`-like file passed via CLI. The filename is
+user-defined; uv does not look for it automatically.
+```bash
+uv lock --build-constraints build-constraints.txt
+uv sync --build-constraints build-constraints.txt
+uv build --build-constraint build-constraints.txt   # note: singular flag here
+```
+The equivalent environment variable is `UV_BUILD_CONSTRAINTS`, which accepts a
+space-separated list of file paths.
```
eskultety, line 691: Environment variables should IMO be discussed in a dedicated section, otherwise they will be scattered everywhere in the doc and hard to follow, I suggest having a dedicated section dealing with env vars.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+
+
+# uv support in Hermeto
+
+## Approach 1 (rejected): using uv
```
eskultety: Feels weird to start with a "rejected" solution. I think we can briefly mention it after describing the solution we went for.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+  
+2. `uv export` → `requirements.txt` → fetch is lossy: no structured download
+   URLs, no source-type distinctions, flattened markers etc.
+
+
+
+## Approach 2 (preferred): parse uv.lock and fetch directly
+
+
+
```
eskultety, line 696: Why so many blank lines here?

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+## Prefetching
```
eskultety: styling nitpick: This duplicates H2 heading. Maybe the best we can do here is directly work with this approach and after `# uv support in Hermeto` you lead the section with something like (use your own wording): 
_The described approach focuses on parsing uv.lock followed by fetching dependencies directly._
And THEN you follow with the `## Prefetching` section on details. How about that? @a-ovchinnikov, thoughts?

docs/design/uv.md
eskultety: Why Step A instead of `Step 1`?

docs/design/uv.md
eskultety: Yes, leading a section like this is confusing, the reader sees this for the first time, you need to build upon previous information towards this claim (general advice).

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+```toml
+[[index]]
+name = "local"
+url = "/path/to/hermeto-output/deps/uv"
+format = "flat"
+default = true
+```
+
+
+Why this didn't work:
```
eskultety: How about a section called `Limitations` or `Problems` instead of "why this didn't work"

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+4. Lockfile rewrite
+
+After step 4, no uv command that regenerates the lockfile should run.
+
+## Lockfile rewriting (`inject-files`)
+
+`inject-files` modifies `uv.lock` in place to redirect every artifact reference
+to a local path inside `{output_dir}/deps/uv/`. `pyproject.toml` is never
+modified, uv's `--frozen` flag (see [Build environment](#build-environment))
+makes lockfile-only rewriting sufficient.
```
eskultety: We may need a few more details on this. Will `uv.lock` be rewritten from a `ProjectFile` with templating or directly with some logic in `inject-files`? I mean the output from prefetch needs to write metadata on which deps need rewriting so that `inject-files` doesn't need to re-run the whole resolver logic and instead just fills out the template with correct `output_dir` and rewrites `uv.lock`. This is IMO an important technical bit. The sections below explain WHAT you're rewriting, but don't expand on HOW. We may not need any templating, that is also an option, I'd just like to read and understand what you suggest and how you think about it.

docs/design/uv.md
```diff
@@ -0,0 +1,659 @@
...
+The build runs with:
+
+```
+uv sync --frozen [--no-binary]
+```
+
+`--no-binary` is set when binary filters disallow wheels (default mode in
+Hermeto). Omitted when wheels were prefetched.
+
+### Why `--frozen` is required
```
eskultety: IMO this section is either not named well or it just feels misplaced to me, maybe a different section name, or merging with a different section would do the trick for me.

eskultety (commented):
I don't see any mention of whether we plan on verifying uv.lock's integrity with UV tooling or not (and why). Like I mentioned in my comments, maybe a restructured/renamed section (there's `Build environment`) explicitly target at all environment variables Hermeto needs to consider either for its own use for hermetic builds would be nice (unless I missed it somehow).
Overall, there's good technical contents here, this mostly needs restructuring/reshuffling, but we're off to a strong start implementing a PoC based on this.

harshitbansal184507:
`uv sync` doesn't respect `UV_FIND_LINKS` — see [Discord thread](https://discord.com/channels/1039017663004942429/1504036823481974876).

docs/design/uv.md
harshitbansal184507: There are env variables which are available for these flags, included them in the generate-env section with explanation.

docs/design/uv.md
harshitbansal184507: moved this whole section to rejected approaches

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
+# Adding uv support to Hermeto
+
+## Table of Content
+
+- [Adding uv support to Hermeto](#adding-uv-support-to-hermeto)
+  - [Table of Content](#table-of-content)
```
a-ovchinnikov, line 7: Nitpick: I think the top level link and a link to the table of content could be removed from the table of content.

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+├── uv.lock
+└── src/
+    └── main.py
+```
+
+Where `pyproject.toml` declares dependencies and project configuration, and
+`uv.lock` is the universal lockfile that pins resolved dependencies for all
+platforms.
+
+## Specifying dependencies(`pyproject.toml`)
```
a-ovchinnikov: What do you think about
```diff
- ## Specifying dependencies(`pyproject.toml`)
+ ## How dependencies are specified in `pyproject.toml`
```

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+
+uv supports several types of [dependency
```
a-ovchinnikov: WDYT about
```diff
- uv supports several types of [dependency sources](...).
+ uv allows dependencies to come from [different sources](...).
```
?

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+  httpx = [
+    { git = "https://github.com/encode/httpx", tag = "0.27.2", marker = "sys_platform == 'darwin'" },
+    { git = "https://github.com/encode/httpx", tag = "0.24.1", marker = "sys_platform == 'linux'" },
+  ]
+```
+</details>
+
+## `uv.toml`
+
+`pyproject.toml` and `uv.toml` carry identical configuration, the difference is nesting.
```
a-ovchinnikov: This is a bit unclear, maybe something like
```diff
- the difference is nesting.
+ the difference is structural: the former requires extra levels of nesting for sections.
```
could help?

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+`pyproject.toml` requires the `[tool.uv]` prefix since it can be a shared file for multiple tools;
```
a-ovchinnikov: Let's expand this sentence a bit to make it easier on readers.
```diff
- `pyproject.toml` requires the `[tool.uv]` prefix
+ For example `pyproject.toml` requires the `[tool.uv]` prefix
```
Also, why is it a prefix? Is it called this way elsewhere? I would have called it a section header.

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+  [tool.uv.sources]
+  # different source per platform
+  httpx = [
+    { git = "https://github.com/encode/httpx", tag = "0.27.2", marker = "sys_platform == 'darwin'" },
+    { git = "https://github.com/encode/httpx", tag = "0.24.1", marker = "sys_platform == 'linux'" },
+  ]
+```
+</details>
+
+## `uv.toml`
```
a-ovchinnikov, line 216: You'd need to introduce `uv.toml` to the reader in some way. Now you start by saying that it carries the same data as `pyproject.toml` just in a different format, and then immediately add that _some_ fields from `pyproject.toml` cannot be represented. This contradicts the identity statement.

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+source = { ... }                   # always; exactly one of:
+                                     #registry  | git       | url       | path
+                                     #directory | editable  | virtual
+
+resolution-markers = [...]         # only if this package is locked at multiple versions
+dependencies = [...]               # only if non-empty; array of inline tables
+sdist = { ... }                    # varies by source type 
+wheels = [...]                     # varies by source type 
+
+[package.optional-dependencies]    # only if non-empty
```
a-ovchinnikov: Nitpick: do you mean "only if optional dependencies are present"? It is also not exactly clear what `extra-name` and `group-name` are, are those effectively `optional-dependency-name` and `dev-dependency-name` respectively or something else?

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+
+
+
+## Workspace
+
+ A workspace in uv is a collection of one or more packages, called `workspace members`, that are managed together. 
+ The workspace root is itself a member, and other members are declared explicitly under `[tool.uv.workspace]` using members (required) and exclude (optional) keys, both of which accept glob patterns. All included members share one uv.lock file and common uv commands will affect all packages within the workspace(Packages which are included).
+
+<details>
+  <summary>sample structure</summary>
```
a-ovchinnikov: WDYT?
```diff
- sample structure
+ Sample project with workspaces structure
```

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+└── src/
+    └── my_app/
+        └── __init__.py
+```
+ 
+  Here `my-app` is the workspace root. `pkg-a` is an included workspace member. `pkg-b` is excluded, it has its own independent `pyproject.toml` and `uv.lock`, and uv ignores it entirely when operating on the workspace.
+ 
+</details>
+<details>
+  <summary>sample pyproject.toml (root)</summary>
```
a-ovchinnikov: WDYT?
```diff
- sample pyproject.toml (root)
+ root pyproject.toml for a project with workspaces
```

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+ 
+  [tool.uv.workspace]
+  members = ["packages/*"]
+  exclude = ["packages/pkg-b"]
+  ```
+ 
+</details>
+
+<details>
+<summary>how members are recorded in uv.lock</summary>
```
a-ovchinnikov:
```diff
- how members
+ how workspace members
```

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+│           └── pkg_b/
+│               └── __init__.py
+├── pyproject.toml
+├── uv.lock
+└── src/
+    └── my_app/
+        └── __init__.py
+```
+ 
+  Here `my-app` is the workspace root. `pkg-a` is an included workspace member. `pkg-b` is excluded, it has its own independent `pyproject.toml` and `uv.lock`, and uv ignores it entirely when operating on the workspace.
```
a-ovchinnikov: > pkg-b is excluded

Can we see this from the structure alone or should we inspect pyproject.toml to learn that? In other words, is it excluded because it has its own pyproject.toml or because it is listed in `excluded` section in root pyproject.toml?

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+  members = ["packages/*"]
+  exclude = ["packages/pkg-b"]
+  ```
+ 
+</details>
+
+<details>
+<summary>how members are recorded in uv.lock</summary>
+
+Workspaces add two things to the lockfile on top of the format documented above:
```
a-ovchinnikov: It won't hurt to spell out which things are added first and dive into explaining how they are structured next, could you please do that?

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+]
+
+[package.metadata]
+requires-dist = [{ name = "requests" }]
+```
+Other dependencies  appear as `[[package]]` entries, workspaces don't change how those are recorded.
+</details>
+<details>
+  <summary>tool.uv.sources cascading</summary>
+  Any `tool.uv.sources` defined at the root level applies to all members automatically, unless a specific member overrides it in its own `tool.uv.sources`:
```
a-ovchinnikov, line 551: What is `tool.uv.sources`?

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+> TODO: confirm final approach for marker evaluation. Current proposal:
+> introduce `evaluate_markers: bool = False` field on `UvPackageInput` to keep
+> the default behaviour fetch everything.
+
+
+
+
+## Build process in uv
+
+uv uses [PEP 517](https://peps.python.org/pep-0517/) for building distributions by default. When building a Python distribution from source, PEP 517 splits the work between two tools:
```
a-ovchinnikov: Semantics nitpick: uv does not use a PEP, it cannot use a PEP, it can use a process or framework or something else described or introduced by a PEP.

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+- **[inject-files](#lockfile-rewriting-inject-files)** - rewrite every artifact URL in `uv.lock` to point into
+  `{output_dir}/deps/uv/` instead of the network
+- **[generate-env](#build-environment-generate-env)** - produce an environment file with the variables the
+  user's build step needs to run offline against `{output_dir}/deps/uv/`.
+
+The rejected alternative  is in the [Appendix](#appendix-rejected-approach-using-uv-natively).
+
+### `uv sync` and `uv build`: why both are required
+
+Hermeto must support two uv commands, not one:
```
a-ovchinnikov: I am having hard time with this. Why would Hermeto need to support `uv build`? Hermeto by itself does not build anything, it prepares an environment for a build. Or did you mean that Hermeto must provide output sufficient for these two commands to run?

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+
+| Source kind                                | Checksum                       |
+|--------------------------------------------|--------------------------------|
+| `registry` (PyPI)                          | Optional (usually present).    |
+| `url`                                      | Required (uv errors otherwise).|
+| `git`                                      | Must not be present.           |
+| `path` (file artifact) |Required (uv errors if missing) |
+| `editable`, `directory`, `virtual` | None |
+
+Hermeto verifies checksums at download time when the lockfile records them.
```
a-ovchinnikov: Please move this sentence to the beginning of `Checksum handling by source kind`, and also s/verifies/should verify`

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+
+#### Lockfile validation
+
+> **TODO**: 
+
+### Lockfile rewriting (inject-files)
+
+`inject-files` modifies `uv.lock` in place to redirect every artifact
```
a-ovchinnikov: s/modifies/needs to modify

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+reference to a local path inside `{output_dir}/deps/uv/`. `pyproject.toml`
+is never modified — `--frozen` (see [Build environment
+(generate-env)](#build-environment-generate-env)) makes lockfile-only
+rewriting sufficient.
+
```
a-ovchinnikov, line 785: Here again some linkage is missing: something along the lines of "Rewrites necessary for different types of entries are listed below". Remember, the document should guide a reader, not befuddle them!

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+
+#### `UV_NO_BINARY`
+
+[`UV_NO_BINARY=1`](https://docs.astral.sh/uv/reference/environment/#uv_no_binary) tells uv to install from sdists, ignoring the `wheels[]` entries in `uv.lock`.
+This is a conditional variable which depends upon whether user uses binary filters or not.
+
+### Example
+
+A typical user Dockerfile after chaanges:
```
a-ovchinnikov: s/chaanges/changes

docs/design/uv.md
```diff
@@ -0,0 +1,954 @@
...
+```
+export UV_OFFLINE=1
+export UV_FIND_LINKS=/hermeto-output/deps/uv
+export UV_FROZEN=1
+export UV_NO_BINARY=1
+```
+
+### SBOM
+
+> **TODO**: deeper research needed; deprioritized in the proposal.
```
a-ovchinnikov: What do you plan to research here?

a-ovchinnikov (commented):
This document is getting to a better shape, thank you! Please find some more comments inline, quite a few of them are just nitpicks and minor suggestions.

docs/design/uv.md
harshitbansal184507: Sure, I didn't add this manually, I am using a VS extension for this. I just added the TOC for easier navigation or is it ok to have a TOC in the final document?

eskultety: >  is it ok to have a TOC in the final document

We mostly don't use it. That said, your proposal would not be the one setting a precedent, that was already set by `docs/design/bundler.md` . On one hand I feel like we need a template for these designs such that they strictly follow some structure, but we tried that in the past but came nowhere near a consensus as a team on what would such a generic template look like to be usable in any situation and provide value. On the other hand I feel strongly about having a ToC in comprehensive designs like yours because it makes navigation much easier, so I'd say go for it!

docs/design/uv.md
harshitbansal184507: > Is it called this way elsewhere?


https://docs.astral.sh/uv/concepts/configuration-files/, here just above the second note.

docs/design/uv.md
harshitbansal184507: From the structure alone we cannot conclude that a package is included or not, we have to inspect pyproject.toml for that. I'll rephrase it properly.

docs/design/uv.md
harshitbansal184507: > Or did you mean that Hermeto must provide output sufficient for these two commands to run?

This is exactly what I meant. 
```diff
- Hermeto must support two uv commands, not one:
+ Hermeto's output must support two uv commands, not one:
```

This will work right?

docs/design/uv.md
a-ovchinnikov: Thank you for confirming!

docs/design/uv.md
a-ovchinnikov: Thank you!

docs/design/uv.md
a-ovchinnikov: Oh, I see it now, then I suggest something along the lines of "Hermeto has to provide data sufficient for running two `uv` commands:"

harshitbansal184507:
Here is the PoC experiment with the rewritten lockfile as requested:

https://github.com/harshitbansal184507/uv-experiment

Any questions, feedback, or pointers on shortcomings are very welcome, happy to iterate on this further.

docs/design/uv.md
harshitbansal184507: The rewording is acceptable right?

docs/design/uv.md
harshitbansal184507: added info regarding `tool.uv.sources`  in `How dependencies are specified in pyproject.toml` section.

docs/design/uv.md
a-ovchinnikov: Yes, it is much better now, thank you.

docs/design/uv.md
```diff
@@ -0,0 +1,971 @@
...
+name = "httpx"
+version = "0.28.1"
+source = { path = "{output_dir}/httpx-0.28.1.tar.gz" }
+sdist = { hash = "sha256:..." }
+```
+
+#### Other source kinds
+
+Already existing `path`, `editable`, `directory`, `virtual` dependencies have no change.
+#### Path and URL format conventions
```
a-ovchinnikov, line 868: Nitpick: am empty line is missing above this header.

docs/design/uv.md
```diff
@@ -0,0 +1,971 @@
...
+Two fields, two conventions:
+
+| Field                          | Format                    | Example                                      |
+|--------------------------------|---------------------------|----------------------------------------------|
+| `source = { path = "..." }`    | Plain path                | `"{output_dir}/flask-...tar.gz"`             |
+| `sdist.url`, `wheels[].url`    | URL with `file://` scheme | `"file:///{output_dir}/anyio-...tar.gz"`     |
+
+### Build environment (generate-env)
+
+Hermeto uses the following environment variables for the user's build step:
```
a-ovchinnikov: s/uses/should set/

docs/design/uv.md
```diff
@@ -0,0 +1,971 @@
...
+
+[package.metadata]                 # only for mutable sources (not registry, not git)
+requires-dist = [...]              
+provides-extras = [...]            
+
+[package.metadata.requires-dev]    # # only if the package declares dependency groups
+<group-name> = [...]
+```
+
+**How `pyproject.toml` declarations map to `uv.lock` source kinds** :
```
eskultety, line 305: formatting question - Was ^this _How pyproject.toml declarations..._ section deliberately put to the `[[package]]` expandable field? I understand it's related, but it makes the expansion quite long and so it feels like it deserves it's own expandable section, IOW not be nested under `[[package]]` because from a reader's POV it's unexpected information that could be read/not read on demand.

docs/design/uv.md
```diff
@@ -0,0 +1,971 @@
...
+
+Both are resolved into `uv.lock` the same way as regular dependencies, with
+marker / group metadata indicating where they belong.
+
+
+
+## Workspace
+
+ A workspace in uv is a collection of one or more packages, called `workspace members`, that are managed together. 
+ The workspace root is itself a member, and other members are declared explicitly under `[tool.uv.workspace]` using members (required) and exclude (optional) keys, both of which accept glob patterns. All included members share one uv.lock file and common uv commands will affect all packages within the workspace(Packages which are included).
```
eskultety: _Packages which are included_ . Is the capital 'P' deliberate here? Was that supposed to be a hyperlink to a different section?

docs/design/uv.md
```diff
@@ -0,0 +1,971 @@
...
+> TODO:
+
+## Markers
+
+uv.lock is universal — it contains entries for all platforms and Python
+versions the project supports. Each dependency edge can carry a [PEP 508
+marker](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers) such as
+`sys_platform == 'win32'`, restricting it to specific environments.
+
+> TODO: confirm final approach for marker evaluation. Current proposal:
```
eskultety, line 619: See my comment https://github.com/hermetoproject/hermeto/pull/1564/changes#r3267710598. I think that will lead you to what needs to be added to this section and drop the TODO, i.e. markers become relevant for us only when specified via key-value pairs that directly map to our binary filters, e.g. OS, arch, etc. So the point here is that when binary filter cannot be satisfied due to a marker, we should log and skip that dependency instead of failing.
For sdists markers are irrelevant because sdists are agnostic to target envs.

docs/design/uv.md
```diff
@@ -0,0 +1,971 @@
...
+
+
+
+## Build process in uv
+uv builds distributions through the [PEP 517](https://peps.python.org/pep-0517/)
+process, which splits the work between two tools:
+
+`build frontend` - the command the user invokes. For uv that's [`uv build`](https://docs.astral.sh/uv/concepts/projects/build/#using-uv-build). `uv build` invokes the declared backend in an isolated environment per PEP 517, producing a source distribution (.tar.gz) and a binary distribution (.whl) in dist/ by default. The --sdist / --wheel flags limit the output to one format.
+
+`build backend` - the tool that actually turns the project's source tree into a wheel or sdist, declared by the project's pyproject.toml [build-system] table (e.g. hatchling, setuptools, flit-core). **uv works with any PEP 517-compliant backend and also ships its own, [uv_build](https://docs.astral.sh/uv/concepts/build-backend/#using-the-uv-build-backend)**.
```
eskultety, line 632: I vaguely recall us having raised comments on this section, anyway, I'm unsure how much value this content provides in context of hermeto, so I'd easily drop it. NOTE - the critical information here is below - build deps not being recorded!

eskultety: Maybe keep the section name, but drop the intro. @a-ovchinnikov thoughts?

docs/design/uv.md
```diff
@@ -0,0 +1,971 @@
...
+space-separated list of file paths.
+
+## uv support in Hermeto
+
+For uv, Hermeto's three phases do the following:
+
+- **[fetch-deps](#prefetch-fetch-deps)** - download every artifact listed in `uv.lock`, plus the
+ build dependencies `uv.lock` does not record, into
+  `{output_dir}/deps/uv/`.
+- **[inject-files](#lockfile-rewriting-inject-files)** - rewrite every artifact URL in `uv.lock` to point into
```
eskultety, line 700: `s/instead of the network/instead of package origin`

docs/design/uv.md
```diff
@@ -0,0 +1,971 @@
...
+
+`build frontend` - the command the user invokes. For uv that's [`uv build`](https://docs.astral.sh/uv/concepts/projects/build/#using-uv-build). `uv build` invokes the declared backend in an isolated environment per PEP 517, producing a source distribution (.tar.gz) and a binary distribution (.whl) in dist/ by default. The --sdist / --wheel flags limit the output to one format.
+
+`build backend` - the tool that actually turns the project's source tree into a wheel or sdist, declared by the project's pyproject.toml [build-system] table (e.g. hatchling, setuptools, flit-core). **uv works with any PEP 517-compliant backend and also ships its own, [uv_build](https://docs.astral.sh/uv/concepts/build-backend/#using-the-uv-build-backend)**.
+
+The build-time dependencies are
+**not recorded in `uv.lock`**  ([astral-sh/uv#5190](https://github.com/astral-sh/uv/issues/5190)).This is why Hermeto must prefetch build dependencies separately via
+`uv export` + `pybuild-deps` (see [**2) Build dependencies**](#2-build-dependencies)). All build dependencies live in `{output_dir}/deps/uv/` and are exposed to uv via
+`UV_FIND_LINKS` during the build.
```
eskultety, line 637: For the sake of argument I could also imagine ^this hunk completely integrated into `Build-time dependencies` in the Hermeto approach below.

docs/design/uv.md
```diff
@@ -0,0 +1,971 @@
...
+[`pybuild-deps`](https://pybuild-deps.readthedocs.io/) is the tool Hermeto
+uses to enumerate build-time dependencies. It does not support `uv.lock` as an input format - it reads
+`requirements.txt`. The intermediate step is to convert `uv.lock` to
+`requirements.txt` via `uv export`, then run `pybuild-deps compile`:
+
+```
+uv export --format requirements-txt > requirements.txt
+pybuild-deps compile requirements.txt -o build-requirements.txt
+```
+
```
eskultety: I agree this is critical & valuable information, however, the design doc will then be tied to another tool that is planned to add native `uv` support in the future which means once that happens it would invalidate our design doc which should ideally be invalidated by hermeto's own changes/adjustments.

Therefore, I'd suggest extracting this bit, documenting the build deps process such that "Hermeto requires build dependencies to be present in a requirements file" and let to users to do. We will then have to put the ^hunk above about `uv export` and `pybuild-deps` to our user-facing docs for users to navigate.

docs/design/uv.md
```diff
@@ -0,0 +1,971 @@
...
+|--------------------------------------------|--------------------------------|
+| `registry` (PyPI)                          | Optional (usually present).    |
+| `url`                                      | Required (uv errors otherwise).|
+| `git`                                      | Must not be present.           |
+| `path` (file artifact) |Required (uv errors if missing) |
+| `editable`, `directory`, `virtual` | None |
+
+#### Lockfile validation
+
+> **TODO**:
```
eskultety: I think this section should be fairly trivial - inspecting what `uv lock --check` does, whether it fetches anything or modifies anything and then mention the fact that we'd need the uv binary in our image for this functionality to work.

docs/design/uv.md
eskultety: ACK, SBOM handling tends to be pretty straightforward most of the time, because we already have a lot of precedent for components, workspaces, different package types, etc. It's just plumbing work I'd say. We can leave out the section for now and add it at the end of implementation if needed.

eskultety (commented):
As far as I'm concerned, I'm not missing anything critical anymore. The design is comprehensive, full of references. My comments are mostly styling things. For the purposes of the actual work I think we can happily proceed to the implementation part starting with a PoC and then expand that towards proper implementation. I'd keep the doc open for now if we need to make further adjustments further down the road before merging it with the bulk of the implementation work. ACK.

docs/design/uv.md
harshitbansal184507: The section covers the split(not in very detailed way though) and why build-time deps need separate handling. Likely redundant for seasoned engineers, but useful for anyone newer to Python packaging(like me) especially with no pip design doc to point at. Happy to drop it if you want to keep design document tight to hermeto.

docs/design/uv.md
harshitbansal184507: > Is the capital 'P' deliberate here? Was that supposed to be a hyperlink to a different section?

Not really, It's a mistake by me. Do you want to lead it to any other section though?

eskultety: Nah, in fact I'd probably drop those parentheses completely.

docs/design/uv.md
eskultety: > but useful for anyone newer to Python packaging(like me) especially with no pip design doc to point at. Happy to drop it if you want to keep design document tight to hermeto.

I see. Well, design docs are IMO tightly coupled to the projects where they reside, so some fundamental knowledge can be expected (the fact that you may not have had it prior writing the doc is kinda irrelevant :) ). WIth the abundance of AI tooling everywhere, if that thing is good for one thing, it's processing text. I won't block on this one, I only feel that the smaller the wall of text we need the better :)  and it's supposed to be a "quick" reference.

docs/design/uv.md
harshitbansal184507: we should use it with `--no-cache` so that cache is not created. Right?

harshitbansal184507: we should use it with `--no-cache` so that cache is not created. Right?

a-ovchinnikov: If it creates a cache otherwise then yes.

harshitbansal184507: It does. With '--no-cache' also temporarily cache is created but that should not be an issue ig.

docs/design/uv.md
a-ovchinnikov: I would rather keep this (not blocking on it though), it is small and makes transition between sections smoother, the same generally applies to other more verbose parts of the document: I think the point is to make the document sufficiently self-contained to help a reader, to a point of course, but having a few expanded sections won't hurt, the difference will end up being on the scale of ~100 lines of markdown. And the document could always be summarized by an AI after all.

docs/design/uv.md
```diff
@@ -0,0 +1,1024 @@
...
+> meaningful. uv enforces this explicitly:
+> *"The `sources` field is not allowed in a `uv.toml` file. `sources` is only
+> applicable in the context of a project, and should be placed in a
+> `pyproject.toml` file instead."*
+>See: https://github.com/astral-sh/uv/issues/11632
+
+## `uv.lock`
+
+The `uv.lock` file follows the TOML format.The top of the file holds resolver-wide settings; below that comes an array of `[[package]]` tables, one per locked package. Within each package, the `source` form determines which sub-keys appear.  The format is versioned — this design targets
+`version = 1` and all its future revisions.
```
eskultety, line 243: I just found something this section might want to include (or some other better fitting section for that matter) - https://github.com/astral-sh/uv/issues/6830 , i.e. be explicit that `uv.lock` is a fixed file name and at the time of writing there's isn't a way to supply a different name.

harshitbansal184507: Added as a note.

harshitbansal184507:
@coderabbitai review

tnevrlka:
@coderabbitai review

docs/design/uv.md
```diff
@@ -0,0 +1,1074 @@
...
+| `editable`, `directory`, `virtual` | None |
+
+#### Lockfile validation
+
+Before any artifact download, Hermeto verifies that `uv.lock` is in sync with `pyproject.toml`:
+
+```
+uv lock --check --no-cache --no-python-downloads
+```
+
```
eskultety: I'd probably drop this based on: https://github.com/hermetoproject/hermeto/pull/1631#discussion_r3819846542 . Maybe we could add it to the future scope section or remove entirely, either way we'll most likely need to start tracking it as a standalone follow-up issue item.

harshitbansal184507:
Removed the lockfile validation section in the design document. I don't think it has to be mentioned here based on discussion in the implementation PR.

docs/design/uv.md
```diff
@@ -0,0 +1,1071 @@
...
+[`UV_NO_BINARY_PACKAGE=<pkg1> <pkg2> …`](https://docs.astral.sh/uv/reference/environment/#uv_no_binary_package) tells uv to install from sdist only for the listed packages. Set when `binary.packages` is an explicit list; see [Binary filters](#binary-filters) for the conditional logic.
+
+### Example
+
+A typical user Dockerfile after changes:
+
+```dockerfile
+FROM python:3.12-slim
+
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
```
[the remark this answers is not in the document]
harshitbansal184507: This is a Dockerfile in a design document, not code Hermeto ships or runs

docs/design/uv.md
```diff
@@ -0,0 +1,1071 @@
...
+| `git`                                      | Clone repo at the resolved commit, archive as tarball.              |
+| `url`                                      | Download artifact at `source.url`.                                  |
+| `path`, `editable`, `directory`, `virtual` | Skip - already on local filesystem.                                 |
+
+#### 2) Build-Time dependencies
+
+When uv installs a package from an sdist, it builds a wheel using the PEP 517 build backend declared in that sdist's `[build-system].requires`. `uv.lock` does not record these dependencies ([astral-sh/uv#5190](https://github.com/astral-sh/uv/issues/5190)).
+
+Hermeto requires build dependencies to be present in a `requirements-build.txt` file in the project directory. If present, entries are downloaded into {output_dir}/deps/uv/; if absent, no build-time dependencies are fetched.
+``
```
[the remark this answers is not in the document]
harshitbansal184507: Hermeto currently can't reliably tell that build dependencies are required or not, so no validation is required, already matching the pip backend.

docs/design/uv.md
```diff
@@ -0,0 +1,1057 @@
...
+
+**Rewrites necessary for different types of entries are listed below.**
+
+#### Registry packages
+
+Rewrite `sdist.url` and every entry in `wheels[].url` from
+`https://files.pythonhosted.org/...` to `file:///{output_dir}/<filename>`.
+
+Before:
+
```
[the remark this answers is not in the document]
harshitbansal184507: This was left out on purpose: uv lock --check resolves the project, which can run build backends(ACE).

docs/design/uv.md
```diff
@@ -0,0 +1,1057 @@
...
+```toml
+[[package]]
+name = "flask"
+version = "3.2.0.dev0"
+source = { path = "{output_dir}/flask-gitcommit-7ef2946f.tar.gz" }
+```
+
+#### URL packages
+
+Same pattern: `source = { url = "..." }` becomes `source = { path = "..." }`.
```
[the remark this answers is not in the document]
harshitbansal184507: Out of scope currently

docs/design/uv.md
```diff
@@ -0,0 +1,1057 @@
...
+
+  [tool.uv.sources]
+   # pin to a commit
+flask = { git = "git+https://github.com/pallets/flask", rev = "954f5684e4841aad84a8eec7ace7b81a0d3f6831" }
+  # pin to a tag
+  flask = { git = "https://github.com/pallets/flask", tag = "3.0.0" }
+  # pin to a branch
+  flask = { git = "https://github.com/pallets/flask", branch = "main" }
+  # subdirectory (if package isn't in repo root)
+  flask = { git = "https://github.com/pallets/flask", subdirectory = "libs/flask" }
```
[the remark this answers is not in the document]
harshitbansal184507: Intentional, these are alternatives shown side by side, once per comment.

docs/design/uv.md
```diff
@@ -0,0 +1,1071 @@
...
+
+## `uv.lock`
+
+The `uv.lock` file follows the TOML format.The top of the file holds resolver-wide settings; below that comes an array of `[[package]]` tables, one per locked package. Within each package, the `source` form determines which sub-keys appear.  The format is versioned — this design targets
+`version = 1` and all its future revisions.
+
+> Each version of uv only understands one lockfile format version, and major
+> version changes are completely incompatible. Changes in Revision are forward and backward compatible. See [uv source][uv-lockfile-version].
+
+[uv-lockfile-version]: https://docs.astral.sh/uv/concepts/resolution/#lockfile-versioning
```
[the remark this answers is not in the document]
harshitbansal184507: The wording matches uv's docs [lockfile versioning](https://docs.astral.sh/uv/concepts/resolution/#lockfile-versioning).

  Only `version = 1` exists today, and we can't know what future versions will change, so pinning exact versions and revisions doesn't belong in the design document.

docs/design/uv.md
harshitbansal184507: Done

docs/design/uv.md
```diff
@@ -0,0 +1,1058 @@
...
+
+When uv installs a package from an sdist, it builds a wheel using the PEP 517 build backend declared in that sdist's `[build-system].requires`. `uv.lock` does not record these dependencies ([astral-sh/uv#5190](https://github.com/astral-sh/uv/issues/5190)).
+
+Hermeto requires build dependencies to be present in a `requirements-build.txt` file in the project directory. If present, entries are downloaded into `{output_dir}/deps/uv/`; if absent, no build-time dependencies are fetched.
+
+During the build, uv finds them there through [`UV_FIND_LINKS`](#uv_find_links).
+
+#### Checksum handling by source kind
+Hermeto should verify checksums at download time when the lockfile records them.
+Per uv's
```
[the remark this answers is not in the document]
harshitbansal184507: Verifying checksums already implies failing on a mismatch

PR #1565, opened 2026-05-14
docs: Add AGENTS.md for AI agents

derasdf:
Simple AGENTS.md + CLAUDE.md symlink + README reference.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,92 @@
+# Hermeto
+
+Hermeto prefetches dependencies for hermetic builds and
+produces accurate SBOMs.
+
+If unsure about hermeto conventions, read the relevant
+files or ask. Do not assume.
+
+## Principles
+
+1. **No arbitrary code execution** -- never rely on third-party
+   tools that could execute untrusted code during prefetch,
+   otherwise the whole operation including the SBOM is compromised
+2. **Checksum validation** -- hermeto verifies checksums when
+   available; mark dependencies without user-provided checksums
+   in the SBOM. Not all ecosystems supply checksums for every
+   artifact.
+3. **Lockfile-driven** -- hermeto always parses pre-resolved
+   lockfiles, it never resolves dependencies itself. Each backend
+   reads some ecosystem lockfile format.
+4. **SBOM accuracy** -- hermeto downloads only what is explicitly
+   declared in lockfiles and reports it accurately.
+
+## Project Structure
+
+- `hermeto/core/` core logic, `resolver.py` dispatches to
+  backends based on language ecosystems
+  - `package_managers/` backend implementations,
+    experimental ones use `x-` prefix (`"type": "x-foo"`)
+  - `models/` input/output and SBOM models
+  - `errors.py` error classes for all backends
+- `hermeto/interface/cli.py` CLI entry points `hermeto` and
+  `cachi2` (legacy, same app, avoid `cachi2` in new code)
+- `docs/` user-facing documentation
+- `docs/design/` design documents for each backend
+- `tests/unit/` unit tests
+- `tests/integration/` integration tests
+- [pyproject.toml](pyproject.toml) has the minimum Python
+  version (`requires-python`) and all tool config
+
+## Git Workflow
+
+- Work in a new clean git worktree, never commit directly on
+  an existing branch
+- DCO sign-off and AI trailer required:
+  `git commit -s --trailer "Assisted-by: Claude"`
+- Rebase only, no merge commits
+- Every commit must pass `pytest tests/unit/` and
+  `nox -s lint` independently
+- Changes split into standalone commits, not a single blob
+- Commit messages explain WHY, not what
+- `# SPDX-License-Identifier: GPL-3.0-only` header on new files
+- No gitmojis
+
+## Environment
+
+- Use the shared venv at the repository root across worktrees,
+  never work directly in the main checkout. Create only if
+  missing: `python3 -m venv venv && venv/bin/pip install -r
+  requirements-extras.txt -e .`
+- `requirements*.txt` are auto-generated -- add deps to
+  `pyproject.toml`, then `nox -s pip-compile`
+
+## Code Style
+
+- Include type annotations for all new code
+- Error messages must be friendly and actionable. Suggest
+  a solution or where to look for help, link to docs when
+  suitable
+- Comments explain WHY, never repeat HOW the code works
+- If code was inspired by third-party sources, link them
+- Always preserve trailing newlines at end of files
+
+## Testing
+
+- Unit tests must not require network access
+- Aim for near full coverage of new code
+- Add new test cases instead of modifying existing ones
+- Adding params to existing parametrize is OK if the test
+  function stays unchanged
+- Avoid unit tests for backend entry point functions
+  (`fetch_*`) - they require heavy mocking; cover them
+  with integration tests instead
+- Test hermeto logic, not stdlib or library behavior
+
+If you encounter what appears to be a potential security
+vulnerability, do not fix it or include it in a commit. Stop
+and alert the human operator to follow the process in
+[SECURITY.md](SECURITY.md).
+
+New backends need a
+[design document](docs/design/package-manager-template.md).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md


AGENTS.md
```diff
@@ -0,0 +1,28 @@
+# Hermeto
```
derasdf, line 1: Had a patterns section but not sure if it belongs here, in a skill, or if it's not needed at all.
Dumping it here for discussion, in case it's useful.

**Commits:**
- EVERY commit passes nox independently. Prefix: fix(cargo):, docs(pip):
- DCO: git commit -s. Body explains WHY if not obvious
- Don't mix code movement with new functionality
- Amend into earlier commits instead of fixup commits

**Testing:**
- No match= in pytest.raises. No log assertions. Exception type only
- No branching in test bodies. Linear Arrange-Act-Assert
- Short IDs ok for large parametrize, descriptive names for standalone
- Integration tests mandatory for new features
- Separate positive/negative tests

**Code:**
- SPDX header on every new file
- Find the right place by looking at existing structure
- Config in Config, not threaded through signatures
- Docstrings only on public functions
- Don't build for hypothetical futures

README.md
```diff
@@ -62,6 +62,8 @@ achieve network isolation.
 We do not distribute Hermeto as a standalone package as of now.
 
 To install Hermeto for local development, see CONTRIBUTING.md.
+For a short description see [AGENTS.md](AGENTS.md).
+`AGENTS.md` should be reviewed each quarter.
```
tnevrlka: I don't see how this fits in the README. Is `AGENTS.md` supposed to be read by humans?

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+- `docs/design/` design documents for each backend
+- `tests/unit/` mirrors source structure, run with `nox`.
+- `tests/integration/` run with `nox -s integration-tests`
+  - each test case is a branch in [hermetoproject/integration-tests](https://github.com/hermetoproject/integration-tests)
+
+## References
+
+[Contributing guidelines](CONTRIBUTING.md) covers dev setup, testing, PR guidelines
+
+[AI contribution policy](AI_CONTRIBUTION_POLICY.md)
```
tnevrlka: Is the AI contribution policy relevant for agents?

README.md
a-ovchinnikov: Do we need this bit of information in README?

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+    (bundler, cargo, gomod, npm, pip, rpm, yarn)
+  - `models/` Pydantic input/output models, SBOM generation
+- `hermeto/interface/cli.py` Typer CLI, entry points `hermeto` and `cachi2` (legacy)
+- `docs/` per-backend documentation
+- `docs/design/` design documents for each backend
+- `tests/unit/` mirrors source structure, run with `nox`.
+- `tests/integration/` run with `nox -s integration-tests`
+  - each test case is a branch in [hermetoproject/integration-tests](https://github.com/hermetoproject/integration-tests)
+
+## References
```
tnevrlka: Is `References` the right word? 

I would understand it to mean that the linked documents were used for writing the AGENTS.md. Which is not the case as far as I know. 

I think something like `Further reading` could be more appropriate

AGENTS.md
a-ovchinnikov: The list itself looks useful to me, in my naivety I would have just placed it right here, agents are supposed to be able to figure it out on their own after all.

a-ovchinnikov (approved):

AGENTS.md
eskultety: That's what the AI generates itself and I think this is spreading widely everywhere: https://github.com/konflux-ci/konflux-build-cli/pull/110/changes#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9

eskultety: I believe when you ask AI to reference something, that's the template it follows. Per the agent instructions guidelines, it should be tested whether the AI is sensitive to the format and expects something like 'References' to find or can "intelligently" infer from other pointers as well. I'd like us not introduce some boiler plate if not necessary.

AGENTS.md
eskultety: > , agents are supposed to be able to figure it out on their own after all.

They can eventually, but it's all about avoiding wasting context window.

AGENTS.md
tnevrlka: > I believe when you ask AI to reference something, that's the template it follows

I haven't thought of it like that. If AI likes it, who am I to judge?

> whether the AI is sensitive to the format and expects something like 'References' 

Note that I'm not an expert on AI development, but considering I haven't seen it in https://agents.md, I wouldn't think so? But I'm not willing to die on this hill

AGENTS.md
derasdf: >agents are supposed to be able to figure it out on their own after all.

All of the stuff I have here are what agents constantly get wrong in my experience.

eskultety: @derasdf yes, principles are important to include. However, you'd want to start with something high-level which influences how the agent approaches work in the codebase, this is what I'd envision:
```
Hermeto prefetches dependencies for hermetic builds and produces accurate SBOMs.

Principles:

1. **No arbitrary code execution** - never rely on third-party tools that could execute untrusted code during prefetch, otherwise the whole operation including the SBOM is compromised
2. **Checksum validation** - hermeto verifies checksums when available; mark dependencies
   without user-provided checksums in the SBOM. Not all ecosystems supply
   checksums for every artifact.
3. **Lockfile-driven** - hermeto always parses pre-resolved lockfiles, it never resolves dependencies itself. Each backend reads some ecosystem lockfile format.
4. **SBOM accuracy** - hermeto downloads only what is explicitly declared in lockfiles and reports it accurately. Experimental backends (`x-` prefix) build toward this progressively; production backends must uphold this principle strictly.
```

README.md
derasdf: I copied it from here [konflux-ci/konflux-build-cli/pull/110/](https://github.com/konflux-ci/konflux-build-cli/pull/110/changes/2097f1c6c8b938e4e3d9c1835fb81122439fd356#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9) and `should be reviewed` part is a requirement: `Acceptance Criteria - A quarterly review process for the file is scheduled/documented`

derasdf: I copied it from here [konflux-ci/konflux-build-cli/pull/110](https://github.com/konflux-ci/konflux-build-cli/pull/110/changes#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9) (same answer as for [a-ovchinnikov](https://github.com/a-ovchinnikov) down below)

AGENTS.md
derasdf: Yes. Agents need to know about disclosure requirements.

README.md
tnevrlka: Ok, I will resolve this and continue in the linked thread.

AGENTS.md
derasdf: >I believe when you ask AI to reference something, that's the template it follows. 
 
 Yes I believe so too, and yes i took it from that other pull, but also i have alternative and more assertive version that i considered as well (in case ai will like to skip over the references and will not read them):
 
   Read [CONTRIBUTING.md](CONTRIBUTING.md) before writing any code (covers dev setup, testing, PR guidelines).
   Read [AI_CONTRIBUTION_POLICY.md](AI_CONTRIBUTION_POLICY.md) for any AI-assisted work.
  New backends need a [design document](docs/design/package-manager-template.md).
  For any work with security vulnerabilities read [SECURITY.md](SECURITY.md).

AGENTS.md
eskultety: > EVERY commit passes nox independently. Prefix: fix(cargo):, docs(pip):

What is the prefix part? What does it do?

> Don't mix code movement with new functionality

We need to strenghten this: _Changes must be split logically into several standalone commits rather than a single blob_

> Amend into earlier commits instead of fixup commits

This IMO lacks context under what circumstances this applies.

> No match= in pytest.raises. No log assertions. Exception type only
No branching in test bodies. Linear Arrange-Act-Assert
Short IDs ok for large parametrize, descriptive names for standalone
Integration tests mandatory for new features
Separate positive/negative tests

^This is IMO mostly skill material, especially the integration tests part, because that's a complex topic, unit test hints I could accept. Note that everything that goes in here is loaded for every single agent in every session, even subagents...

> Find the right place by looking at existing structure

This is very vague, the agent can interpret this any way it wants, if you omit this I don't think you'd see any difference in the interaction. The general rule of thumb is - be explicit, only include what the agent missed, if the experience hasn't changed -> drop it. Rinse and repeat. btw. I love how we're moving from a deterministic world to a human language equivalent of spamming `/retest` on CI requests.

> Docstrings only on public functions

This is a human actor requirement, for agents it doesn't matter, it won't forget to update the doc string if it needs to update a function signature and a doc string exists, that's the very basic 101 of programming all of them can handle nowadays.

> Don't build for hypothetical futures

Vague, how does the agent interpret this? Even a human can interpret it differently given different contexts.

README.md
tnevrlka: I still don't think the README is the proper place for any of this.

- The README of `konflux-build-cli` is not something we necessarily need to follow if it doesn't make sense to us
 - I really don't think README of a community project is a good place for scheduling reminders dictated by an internal Jira without the input of Hermeto maintainers

AGENTS.md
eskultety: > but considering I haven't seen it in https://agents.md/, I wouldn't think so? But I'm not willing to die on this hill

That's the thing, this is supposed to be manually tested (yay, we're going back from automation to manualization) to verify if an instruction changes anything -> if not ditch it -> save context and tokens.

AGENTS.md
tnevrlka: Fair enough

README.md
eskultety: This should IMO be left out completely, that's not something upstream cares about. It should follow the exact same update policy as we have for any other md file like README, CONTRIBUTING, SECURITY, etc. which is officially "no policy" but naturally it should be updated wherever someone notices an update is required, otherwise you'd have to apply the exact same policy to every MD file and every user-facing doc we have, ain't nobody got the capacity.

AGENTS.md
eskultety: No, as long as we instruct the agent to correctly use commit trailers.

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+
+Dependency pre-fetcher for hermetic builds. Generates SBOMs.
+
+If unsure, don't guess. Read full files, not snippets. Verify before claiming.
+
+## High-Level Project Structure
+
+- `hermeto/core/` core logic, `resolver.py` dispatches to backends
+  - `package_managers/` backend implementations
+    (bundler, cargo, gomod, npm, pip, rpm, yarn)
```
eskultety: this line is bound to change as we add more backends or even more backends to the same ecosystem, I'd leave it out.

README.md
derasdf: Okey, I agree, i'll drop the whole readme.

AGENTS.md
```diff
@@ -0,0 +1,28 @@
+# Hermeto
+
+Dependency pre-fetcher for hermetic builds. Generates SBOMs.
+
+If unsure, don't guess. Read full files, not snippets. Verify before claiming.
+
+## High-Level Project Structure
+
+- `hermeto/core/` core logic, `resolver.py` dispatches to backends
+  - `package_managers/` backend implementations
```
eskultety: Good, maybe I'd append `based on language ecosystems` but up to you really.

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+    (bundler, cargo, gomod, npm, pip, rpm, yarn)
+  - `models/` Pydantic input/output models, SBOM generation
```
eskultety: nitpick: pydantic is an impl. detail :),` input/output and SBOM models` should be enough.

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+- `hermeto/interface/cli.py` Typer CLI, entry points `hermeto` and `cachi2` (legacy)
```
eskultety: nitpick: same here, leave out `Typer`, that is easily inferred from the module, i.e. `CLI` or `CLI entrypoint` should suffice.

AGENTS.md
```diff
@@ -0,0 +1,28 @@
+# Hermeto
+
+Dependency pre-fetcher for hermetic builds. Generates SBOMs.
+
+If unsure, don't guess. Read full files, not snippets. Verify before claiming.
```
eskultety: > Read full files, not snippets.

Does the agent not read full files when it's trying to understand structure? E.g. if you reference a particular section from a particular MD file like (read the development setup), it'll ingest the whole CONTRIBUTING regardless before it can parse the section.

> Verify before claiming.

Good instruction actually on hallucination. Shouldn't we narrow this down to "Do not guess, if unsure, ask the human operator"?

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+
+## High-Level Project Structure
+
+- `hermeto/core/` core logic, `resolver.py` dispatches to backends
+  - `package_managers/` backend implementations
+    (bundler, cargo, gomod, npm, pip, rpm, yarn)
+  - `models/` Pydantic input/output models, SBOM generation
+- `hermeto/interface/cli.py` Typer CLI, entry points `hermeto` and `cachi2` (legacy)
+- `docs/` per-backend documentation
```
eskultety: nitpick: `s/per-backend/user-facing documentation`

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+- `docs/design/` design documents for each backend
+- `tests/unit/` mirrors source structure, run with `nox`.
```
eskultety: > mirrors source structure,

What does this do? Like - a test per each function added? I guess I now need a clarifying instruction :D.

> run with `nox`.

read on below on ^this one.

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+- `tests/integration/` run with `nox -s integration-tests`
```
eskultety: I'd advise to instruct the agent to always run tests with vanilla `pytest` for easier debugging, nox is just syntactic sugar envelope. The agent may (my guess) yield a faster and better reaction when this is condensed to a single tool instead of wrapped by more code.

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+  - each test case is a branch in [hermetoproject/integration-tests](https://github.com/hermetoproject/integration-tests)
```
eskultety: this is a skill material, the agent might IMO need more than this simple pointer to get it right without too much inference, token spend, and context window bloating.

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+
+## References
+
+[Contributing guidelines](CONTRIBUTING.md) covers dev setup, testing, PR guidelines
+
+[AI contribution policy](AI_CONTRIBUTION_POLICY.md)
+
+[Design template for new backends](docs/design/package-manager-template.md)
+
+[Security policy](SECURITY.md)
```
eskultety: What does the agent do with the policy? That's a human contributor instruction. The agent doesn't interact with this in any way. That said, there's one thing I have in my config (backed by a recent incident):

> If you encounter what appears to be a potential security vulnerability, **do
  not** fix it or include it in a commit - stop and alert the human operator to
  follow the process in [SECURITY.md](SECURITY.md)

I am yet to find out if it works, but given this really happened to us recently it's worth a try :) .

AGENTS.md
```diff
@@ -0,0 +1,28 @@
...
+- `tests/integration/` run with `nox -s integration-tests`
+  - each test case is a branch in [hermetoproject/integration-tests](https://github.com/hermetoproject/integration-tests)
+
+## References
+
+[Contributing guidelines](CONTRIBUTING.md) covers dev setup, testing, PR guidelines
+
+[AI contribution policy](AI_CONTRIBUTION_POLICY.md)
+
+[Design template for new backends](docs/design/package-manager-template.md)
```
eskultety: Yep!

AGENTS.md
tnevrlka: Obsolete, resolving.

derasdf: Tested both. Gave two agents the same task (prepare commit and PR for a gomod fix) with different reference styles.

Passive "References" heading: agent never read CONTRIBUTING.md, jumped straight to code.

Assertive "Read CONTRIBUTING.md before writing any code": agent read it at step 3, right before composing the commit.

Switched to assertive format, dropped the heading.

AGENTS.md
derasdf: Updated based on this.

AGENTS.md
```diff
@@ -0,0 +1,53 @@
...
+   artifact.
+3. **Lockfile-driven** -- hermeto always parses pre-resolved
+   lockfiles, it never resolves dependencies itself. Each backend
+   reads some ecosystem lockfile format.
+4. **SBOM accuracy** -- hermeto downloads only what is explicitly
+   declared in lockfiles and reports it accurately. Experimental
+   backends (`x-` prefix) build toward this progressively;
+   production backends must uphold this principle strictly.
+
+## Project Structure
```
eskultety, line 24: You're missing one key Python thing here which the agents might have a hard time to infer immediately unless you task it to analyze the whole repo first before commencing any work - which Python version to target:
```
- Target the minimum Python version from `requires-python` in
  [pyproject.toml](pyproject.toml)
```

eskultety: This also has the benefit that, while reducing context somewhat (the file is ~130 lines), every type of agent that is spun up immediately knows the whole project configuration setup, which tools we configure, what direct deps we have, etc. so it can immediately tell what build toolchain we use, how to use it and where to put additional configuration for our tooling without inference which is valuable not only when producing code but also when performing code reviews, etc.

eskultety (changes requested):
Good for a starting point but I think structure-wise this we can improve quite a bit and add some crucial information on the working environment (confinement if you will) and workflow.

Based on the internet content I've read so far, to make the most out of this we need to be explicit on tasks the agent spends time figuring out how to do. Let me show you some bits that I have in my config that I think we may want to incorporate:

```
 ## Git Workflow
 
 - **Always** work in a new clean git worktree unless instructed otherwise, 
   never commit directly on an existing branch
 - DCO sign-off required: `git commit -s`    <<<<< you already have this bit
 - AI-assisted code requires `Assisted-by:` trailer:
   `git commit -s --trailer "Assisted-by: Claude"`
 - Rebase only, no merge commits       <<<<< pairs with your amend instruction but this is a bit stronger
 - Every commit must pass `nox -s lint` and `pytest -v tests/unit/`
   independently       <<<< you already have this
```

Environment instructions - python development should nowadays primarily happen in isolated environments for convenience and more --> `venv`. The agent itself should never work within the main git repo context, that not only prevents agents working on several things in parallel, but it also protects the main branch and whatever changes you as a human operator may have in a given directory.

```
## Environment

Use the existing venv at the repository root, it is shared across worktrees.
Only create a new venv inside the worktree when adding new dependencies:

python3 -m venv venv
venv/bin/pip install --upgrade pip
venv/bin/pip install -r requirements-extras.txt
venv/bin/pip install -e .

## Commands

Ensure the venv is set up first (see Environment above) - use an existing one or
create one if missing.

nox                              # lint + unit tests (default)
nox -s ruff-fix             # auto-fix formatting/linting
nox -l                           # list all available sessions

For unit and integration tests, prefer native pytest for easier debugging:

pytest tests/unit/               # all unit tests
pytest tests/integration/  # integration tests (builds and uses hermeto                                                     container image)
```
Now, for git worktrees I specifically don't include instructions how to create one because the agent can do that reliably well (and Claude even has some built-in tool and doesn't use Bash for it). One thing missing in my config because I have it in a skill is git worktree placement, so we might need to tell the agents where those worktrees should be created, somewhere it has access to otherwise it'll keep asking for the `cd` permission combined with some other command which will trigger an unsafe operation alert --> permission request.

I also have notes on working with our own dependencies, experimental backend stuff (since you're linking the template; I do too, my instruction if questionable):
```
- `requirements*.txt` are auto-generated - add deps to `pyproject.toml`,
  then `nox -s pip-compile`
- Experimental package managers use `x-` prefix (`"type": "x-foo"`)
```

Lastly, on testing, I constrained the agent to avoid writing tests for `fetch_` entrypoints, and told it to use doctests more (however, with doctests I wanted to experiment more with providing examples, because otherwise it can be interpreted surprisingly vaguely):
```
## Code Quality
- Avoid unit tests for backend entry point functions (`fetch_*`) - they
  require heavy mocking; cover them with integration tests instead
- For trivial single-purpose helpers, use doctests instead of pytest
  parametrized tests
```

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Done but not fully sure about this one, did a bit of testing and without principles this makes agents ask about everything with its better but I'm not sure if fullsend will like this one, don't they want them to be fully autonomous and unstopped ? (even tho yes we would like it done as this)

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Done, split nox into commands that it does.

AGENTS.md
derasdf: Sure, and probably will need separate skills for unit and integration testing

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Done, yes it works better this way but same concern if fullsend will like this or not

derasdf:
To the top level @eskultety comment: Incorporated most of it.
  Tested each addition against the old version to see what actually changes behavior. Cuts:
  - Didn't add venv setup commands, agent reads CONTRIBUTING.md for that
  - Didn't add fetch_* and doctests, too inconsistent in broader tasks, better as a skill
  - Didn't add Commands section, agents discover nox sessions on their own

AGENTS.md
eskultety: > but I'm not sure if fullsend will like this one, don't they want them to be fully autonomous and unstopped ?

What is fullsend? What's the connection? Why should we care? :)

AGENTS.md
```diff
@@ -0,0 +1,69 @@
...
+- Experimental package managers use `x-` prefix
+  (`"type": "x-foo"`)
+
+If you encounter what appears to be a potential security
+vulnerability, do not fix it or include it in a commit -- stop
+and alert the human operator to follow the process in
+[SECURITY.md](SECURITY.md).
+
+Read [CONTRIBUTING.md](CONTRIBUTING.md) before writing any
+code.
```
eskultety: I don't think we're quite there yet that we could do just ^this and expect strong outputs (well, maybe it won't be bad, but context would fill faster and tokens would burn faster). If you think about it, CONTRIBUTING.md can essentially replace this whole document because it holds everything, even the principles and the code instructions are quite explicit in there.

My original idea was pretty much the same - link our CONTRIBUTING.md and call it a day, I don't want to maintain 2 files that serve the same function - being source of truth! That said, reading up online told me otherwise; There are several resources online, so I'll link just a single one which is a good canon:
https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md:

> Keep it concise. For each line, ask: “Would removing this cause Claude to make mistakes?” If not, cut it. Bloated CLAUDE.md files cause Claude to ignore your actual instructions!

If you look at the table below it:

> Exclude:
> - Long explanations or tutorials
> - Self-evident practices like “write clean code”
> - Standard language conventions Claude already knows

When you look at our CONTRIBUTING.md guide, there's quite a lot of human contributor fluff agents don't really need to care about, e.g. our Ethos, release schedule, etc.

So, until we get there that context and token spend don't matter as much, we should follow the recommendation and make peace with maintaining 2 files to make the most out of it.

AGENTS.md
```diff
@@ -0,0 +1,69 @@
...
+- Every commit must pass `pytest tests/unit/`,
+  `ruff check`, `ruff format --check` and `mypy`
+  independently
+- Changes split into standalone commits, not a single blob
+- `# SPDX-License-Identifier: GPL-3.0-only` header on new files
+
+## Environment
+
+- Always work inside a worktree with a venv, never in the
+  main checkout
```
eskultety: Have you tried that the agent can follow this without complaining? This forces the agent to create a venv in every checkout which we explicitly don't want if you look at the examples I shared, it only ever uses a single venv unless totally necessary.

AGENTS.md
```diff
@@ -0,0 +1,69 @@
...
+- `requirements*.txt` are auto-generated -- add deps to
+  `pyproject.toml`, then `nox -s pip-compile`
+- Experimental package managers use `x-` prefix
+  (`"type": "x-foo"`)
+
+If you encounter what appears to be a potential security
```
eskultety, line 86: One thing that just came to mind and that agents get wrong ALL the time - when they create new files OR append to existing ones they always delete the trailing new line which trips linters because they expect a trailing new line at the end of a file, so please add that bit.

eskultety (commented):
I'm wondering if with the current composition of the document the structuring is good and clear, it sounds weird but when it comes to formatting and structuring an AI agent could advise on what the best composition of the content is ONCE you have all instructions ready to work with, so give that a try.

AGENTS.md
derasdf: >What is fullsend?

https://github.com/fullsend-ai/fullsend, this magical tool we may use in the future and for which we're preparing our repo. It may run AI agents on issues and PRs autonomously, so the AGENTS.md is what it reads to understand the project. The "ask human" line might not work if there's no human watching, but for now it's the safer default and what I prefer as well :)

eskultety: I know what it is :) . My intent with the question was to emphasize the connection of our upstream project to some other tools out there we don't currently have any relations with and so, why should we care about some fullsend in Hermeto right now? Is that project mature enough that we'd want to consider it immediately and conform to some policies we haven't discussed publicly anywhere in any GH discussion just yet? :) I hope my message is now clearer.

AGENTS.md
derasdf: Thought about this for a while after which I'm inclined to agree, extracting everything agent-relevant from CONTRIBUTING.md into AGENTS.md directly does make sense and it does reduce context bloat.  Did a first pass at it, probably still missing things. Will iterate as we see what agents fail on.

AGENTS.md
derasdf: Fixed

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Fair, not relevant here.

AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Git Workflow
+
+- Work in a new clean git worktree, never commit directly on
+  an existing branch
+- DCO sign-off and AI trailer required:
+  `git commit -s --trailer "Assisted-by: Claude"`
+- Rebase only, no merge commits
+- Every commit must pass `pytest tests/unit/`,
+  `ruff check`, `ruff format --check` and `mypy`
```
eskultety: nitpick: `nox -s lint` instead of explicit mentions like ruff/mypymay be preferable here.

AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+- `# SPDX-License-Identifier: GPL-3.0-only` header on new files
+- No gitmojis
+
+## Environment
+
+- Use the shared venv at the repository root across worktrees,
+  never work directly in the main checkout. Create only if
+  missing: `python3 -m venv venv && venv/bin/pip install -r
+  requirements-extras.txt -e .`
+- System deps: `golang-bin`, `git`
```
eskultety: Why is golang-bin a system dep? The only system deps we have should be git and python

AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+- `requirements*.txt` are auto-generated -- add deps to
+  `pyproject.toml`, then `nox -s pip-compile`
+- Experimental package managers use `x-` prefix
+  (`"type": "x-foo"`)
```
eskultety: nitpick: does experimental package manager handling belong to Environment? The agent probably doesn't care, but it's confusing to a human.

AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Code Style
+
+- Include type annotations for all new code
+- Error messages must be friendly and actionable -- suggest
+  a solution or where to look for help, link to docs when
+  suitable. Use error classes from
+  `hermeto/core/errors.py` with `solution=`
```
eskultety: mention error classes in project structure above and drop it from here, including the mention of `solution`

AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+- Always preserve trailing newlines at end of files
+
+## Testing
+
+- Aim for near full coverage of new code
+- Add new test cases instead of modifying existing ones
+- Adding params to existing parametrize is OK if the test
+  function stays unchanged
+- Copying large parts of existing tests for new scenarios
+  is OK
```
eskultety: What would the agent resort to otherwise? Creating more efficient tests? I'm confused by the intent of this one :)

AGENTS.md
tnevrlka, line 1: FYI this is more than 60 lines. Personally, I think it's a non-issue and the requirements are unnecessary strict ¯\\\_(ツ)_/¯

eskultety (approved):
A few more nits, but let's say any more comments from my side is just pure bike shedding about something that is a complete blackbox. For the record, I'm being picky because we're trying to confine a "programming" tool with natural language, we'll never know what percentage of this actually works, it's completely non-deterministic.

AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+  an existing branch
+- DCO sign-off and AI trailer required:
+  `git commit -s --trailer "Assisted-by: Claude"`
+- Rebase only, no merge commits
+- Every commit must pass `pytest tests/unit/`,
+  `ruff check`, `ruff format --check` and `mypy`
+  independently
+- Changes split into standalone commits, not a single blob
+- Commit messages explain WHY, not what
+- Amend into earlier commits during review, no fixup commits
```
tnevrlka: I don't get this, why should there be amending during review? Isn't review read-only?

tnevrlka: Why are fixup commits discouraged? In my experience agents have a horrible time with interactive rebases and stopping to amend. I would imagine fixup commits could make sense in many scenarios

AGENTS.md
tnevrlka: nit: does this fit in an `Environment` section?

AGENTS.md
```diff
@@ -0,0 +1,92 @@
...
+## Git Workflow
+
+- Work in a new clean git worktree, never commit directly on
+  an existing branch
+- DCO sign-off and AI trailer required:
+  `git commit -s --trailer "Assisted-by: Claude"`
+- Rebase only, no merge commits
+- Every commit must pass `pytest tests/unit/`,
+  `ruff check`, `ruff format --check` and `mypy`
+  independently
```
tnevrlka: Is there any possibility to have this as a single command? As far as I know, it's encouraged to have the feedback loop as simple as possible so that agents can run it often

AGENTS.md
tnevrlka: duplicate of https://github.com/hermetoproject/hermeto/pull/1565#discussion_r3265865346

AGENTS.md
tnevrlka: duplicate of https://github.com/hermetoproject/hermeto/pull/1565#discussion_r3265844683

AGENTS.md
eskultety: I think the last comment here might be the root of the misunderstanding that led to having ALL nox commands explicitly listed in the file (note I only wanted pytest explicit).

AGENTS.md
eskultety: Where does 60 lines come from? :) AFAIK even Anthropic suggests anything below 200.

AGENTS.md
eskultety: Probably meant as the review step DURING a coding session.

eskultety: I was on the verge of flagging this for the exact same reason as well, but backed off as I've had some successes already (for some token expenditure), but yeah, I agree with ^this.

eskultety (changes requested):
I may have hit approve pre-maturely. To embark on the journey of non-determinism I let an agent review it's own instructions (as absurd as it sounds):

```
 "Use the shared venv at the repository root across worktrees" — how? Symlink? Absolute path? The activation command from a worktree is ambiguous.

 "Amend into earlier commits during review, no fixup commits" — requires interactive rebase (git rebase -i), which AI agents cannot run. Unachievable as written.

 Exact commands block — the single highest-impact gap. Agent must currently discover commands from the noxfile:
  nox -s lint | nox -s ruff-fix | nox -s python-3.10 | nox -s pip-compile (needs podman)

 Backend structure pattern — how backends are wired: the package_managers/<name>/ layout, resolver.py registration, PackageManagerType literal in models/input.py. Critical for adding new backends, completely absent.

 Network access constraints — unit tests must not need network, integration tests do. This boundary isn't stated.

 cachi2 legacy guidance — file mentions it exists but gives no direction. Should the agent avoid the name in new code? Both entry points call the same app — that's not stated.

 Principles section — collapse 8 lines to one: "Hermeto parses lockfiles (never resolves), prefetches with checksum validation, and never executes third-party code."
Add a Backend Pattern section replacing Principles — swap philosophy for the mechanical pattern needed to add/modify backends. 

 Opening description — README tagline, redundant with principles.

 "Do not guess" — generic agent instruction, not project-specific.

 pyproject.toml link in Project Structure — agent reads it anyway, derivable.  (eskultet's note: I doubt it does, one would either have to tell it OR ask about it)
```
Some of ^these are fair.

AGENTS.md
tnevrlka: The beautiful open guidelines by Red Hat that are not accessible by public, haha. I'm not expecting any change from the PR author, just a FYI. Feel free to resolve

AGENTS.md
tnevrlka: Like amending a fix if an issue was flagged during review? I still don't understand why it's specific to the review step

derasdf: Agents do get this part wrong really often during code writing, so i guess the question is do we want to keep it here or do we want to later have a separate code writing skill ?

derasdf: Hm well it is just this line a bit reworded https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md?plain=1#L165, but yes they are bad with it, should i expand on it or really just remove this ?

derasdf:
@eskultety Some fair points, but this is also showing why maintaining a second copy of CONTRIBUTING.md content in AGENTS.md is not the best idea. Half the gaps here are things already covered in CONTRIBUTING.md that agents would understand and some agents would understand it one way and others other way, it is close to impossible to make this effort deterministic and hope that agents won't miss anything that was overexplained in AGENTS.md.
I'd rather bring back a strong link to CONTRIBUTING.md and keep AGENTS.md focused on forcing ai to not be lazy and read all the files and never guess and stop and ask more to be more useful.
The added token cost is ~4400 extra when the agent reads CONTRIBUTING.md. Acceptable tradeoff vs maintaining duplicate content that goes stale and does not improve agents much in my experience (in my testing they seem to behave the same way with CONTRIBUTING.md linked and with it described in AGENTS.md, but ofc i can be wrong...).

AGENTS.md
tnevrlka: > separate code writing skill

Are we expecting to have a skill for code writing? I thought skills are supposed to be mostly used for the automation of some specific project-specific procedures, not for writing code in general which is a large part of what the agents are expected to do
https://code.claude.com/docs/en/best-practices#create-skills

eskultety:
> @eskultety Some fair points, but this is also showing why maintaining a second copy of CONTRIBUTING.md content in AGENTS.md is not the best idea. Half the gaps here are things already covered in CONTRIBUTING.md that agents would understand and some agents would understand it one way and others other way, it is close to impossible to make this effort deterministic and hope that agents won't miss anything that was overexplained in AGENTS.md. I'd rather bring back a strong link to CONTRIBUTING.md and keep AGENTS.md focused on forcing ai to not be lazy and read all the files and never guess and stop and ask more to be more useful. The added token cost is ~4400 extra when the agent reads CONTRIBUTING.md. Acceptable tradeoff vs maintaining duplicate content that goes stale and does not improve agents much in my experience (in my testing they seem to behave the same way with CONTRIBUTING.md linked and with it described in AGENTS.md, but ofc i can be wrong...).

The problem is CONTRIBUTING.md does not contain simple explicit instructions overall (it does for running tests). There's also instructions on setting up venv which is good. There's complete absence on working in an isolated environment which isn't something that you should force upon humans, but it is something for agents. How to run linters is also not mentioned in CONTRIBUTING (`nox -s lint`). It also only suggests running everything through nox which is fine for humans and they can dive into the internals if they need debugging, but you really want the agent to run `pytest` directly instead of spinning in circles and running nox with all installed versions of python which is exactly what would happen by default - have a Fedora? You probably have at least 2 Python versions installed (most devs will have 2 + the one that is the min required version for Hermeto) and the agent would run all of them instead of giving quick feedback on the failure. Those are the little details that I'm not sure we need to be explicit about for humans, but you want agents to get right to ease the frustration. If you can find a good way of putting all that info to CONTRIBUTING.md without looking odd AND iff you can make sure the agent indeed CAN extract all that technical info reliably from there, by all means, go for it :) . I personally think it might be easier to just have a few explicit commands listed in 2 files to make the most out of it. But I hear you, who wants to maintain 2 versions of similar data? That's madness.

AGENTS.md
tnevrlka: @eskultety What was your intention with that line? 
Did you actually mean fixup commits, as in the [actual fixup commits](https://git-scm.com/docs/git-commit/2.32.0#Documentation/git-commit.txt---fixupamendrewordcommit) which should then be autosquashed before force-pushing to the branch, or did you just want to avoid adding new "normal" commits to the branch like `fix: addressed review comments`?

eskultety: > or did you just want to avoid adding new "normal" commits to the branch like fix: addressed review comments?

The latter. A human developer would make sense out of it :)...we'll now going to write in lawyer gibberish to satisfy agents, huh, I call that progress...

eskultety: >  I thought skills are supposed to be mostly used for the automation of some specific procedures, not for writing code

Yes. A skill is a series of instructions for complex tasks that are not obvious at first glance.

derasdf: Dropped

derasdf:
> The problem is CONTRIBUTING.md does not contain simple explicit instructions overall (it does for running tests). There's also instructions on setting up venv which is good. There's complete absence on working in an isolated environment which isn't something that you should force upon humans, but it is something for agents. How to run linters is also not mentioned in CONTRIBUTING (`nox -s lint`). It also only suggests running everything through nox which is fine for humans and they can dive into the internals if they need debugging, but you really want the agent to run `pytest` directly instead of spinning in circles and running nox with all installed versions of python which is exactly what would happen by default - have a Fedora? You probably have at least 2 Python versions installed (most devs will have 2 + the one that is the min required version for Hermeto) and the agent would run all of them instead of giving quick feedback on the failure. Those are the little details that I'm not sure we need to be explicit about for humans, but you want agents to get right to ease the frustration. If you can find a good way of putting all that info to CONTRIBUTING.md without looking odd AND iff you can make sure the agent indeed CAN extract all that technical info reliably from there, by all means, go for it :) . I personally think it might be easier to just have a few explicit commands listed in 2 files to make the most out of it. But I hear you, who wants to maintain 2 versions of similar data? That's madness.

Yep makes sense, keeping it split.

tnevrlka (commented):
Probably lgtm once the existing review comments are addressed

AGENTS.md
eskultety: This comment hasn't been addressed yet.

AGENTS.md
eskultety: This comment hasn't been addressed yet.

AGENTS.md
eskultety: This comment hasn't been addressed yet.

eskultety:
> Probably lgtm once the existing review comments are addressed

Same here, there are still some open threads, but we're close.

AGENTS.md
```diff
@@ -0,0 +1,93 @@
...
+  suitable. Use error classes from
+  `hermeto/core/errors.py` with `solution=`
+- Comments explain WHY, never repeat HOW the code works
+- If code was inspired by third-party sources, link them
+- Always preserve trailing newlines at end of files
+
+## Testing
+
+- Unit tests must not require network access
+- Aim for near full coverage of new code
```
taylormadore, line 77: One consistent review comment for AI-generated unit tests has been to ensure that they are adding real value and not just adding coverage for the sake of coverage (testing stdlib, other libraries, etc.)

taylormadore (commented):
Mostly LGTM once the remaining threads are resolved

AGENTS.md
a-ovchinnikov: I wonder how that would work with LLMs, it requires _a lot_ of insight into the problem space. That and the fact that at least a half of training set was demonstrating precisely this behavior.

eskultety: I believe I tried to mitigate this a little bit in an earlier suggestion above from my own config:
```
- Avoid unit tests for backend entry point functions (`fetch_*`) - they
  require heavy mocking; cover them with integration tests instead
```
but it hasn't been incorporated into the PR it was dismissed for some reason. @taylormadore @a-ovchinnikov would ^this still not be enough? Would that be at least a band-aid?

a-ovchinnikov: This one is good and will probably help, @taylormadore's suggestion though is much broader in scope and I believe that it is not feasible with any LLM-based AI agents. I see no harm in adding it, but am not expecting that an agent will be able to properly follow it.

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Done

AGENTS.md
derasdf: Fair, dropped. It's from [CONTRIBUTING.md L200](https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md?plain=1#L200) but reads weird without context.

AGENTS.md
derasdf: Added

AGENTS.md
slimreaper35: https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md?plain=1#L114

eskultety: > https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md?plain=1#L114

Wow, that must be ancient, by that same logic we should report Yarn, bundler and rust deps. We need to fix that.

AGENTS.md
eskultety: I think the message that was conveyed in proposed sentence is quite different from the original :) .

eskultety (approved):
LGTM. I think this is now good enough to start testing it in practice and adjust as we go. It'll be a collective effort to tweak this such that it consistently yields satisfactory results.

PR #1566, opened 2026-05-14
Unify cargo config.toml file data for pip and cargo backends

### LINKED ISSUE #1097: Hermeto injects .cargo/config.toml for pip and cargo and breaks cargo builds

slimreaper35:
Closes https://github.com/hermetoproject/hermeto/issues/1097

See commit message.

### DIFF
--- a/hermeto/core/package_managers/pip/rust.py
+++ b/hermeto/core/package_managers/pip/rust.py
@@ -119,9 +119,9 @@ def _config_data() -> str:
     return dedent(
         """
         [source.crates-io]
-        replace-with = "local"
+        replace-with = "vendored-sources"
 
-        [source.local]
+        [source.vendored-sources]
         directory = "${output_dir}/deps/cargo"
         """
     )
--- a/tests/integration/test_data/pip_e2e_rust_extensions/.build-config.yaml
+++ b/tests/integration/test_data/pip_e2e_rust_extensions/.build-config.yaml
@@ -10,9 +10,9 @@ project_files:
   template: |2
 
     [source.crates-io]
-    replace-with = "local"
+    replace-with = "vendored-sources"
 
-    [source.local]
+    [source.vendored-sources]
     directory = "${output_dir}/deps/cargo"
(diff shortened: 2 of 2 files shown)


tnevrlka:
I'm not sure I understand this. Is the usage of `vendored-sources` name due to https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/cargo/main.py#L488?

slimreaper35:
> I'm not sure I understand this. Is the usage of vendored-sources name due to https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/cargo/main.py#L488?

Yes. If you run `cargo vendor` in any Rust project, you will see the following lines:

```
...
To use vendored sources, add this to your .cargo/config.toml for this project:

[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "vendor"
```

But we don't have a static string for that in cargo.py (because of git dependencies). We just read from the stdout.

eskultety (approved):

a-ovchinnikov (approved):

PR #1567, opened 2026-05-14
feat(http): make retry count configurable

### LINKED ISSUE #1551: HTTP download retry count is not configurable
The retry count for all HTTP downloads is hardcoded. Both the synchronous path using requests and the asynchronous path using aiohttp read the retry count from `DEFAULT_RETRY_OPTIONS`. There is currently no way to change it via config file or environment variable.

Harsh-2005d:
Fixes #1551

## Description
- Added `max_retries` to `HttpSettings`
  - env var: `HERMETO_HTTP__MAX_RETRIES`
  - config file: `http.max_retries`
  - default is 5.

- removed hardcoded count from `DEFAULT_RETRY_OPTIONS`, both `get_requests_session` and `async_download_files` now read from config at call time

- Before `pkg_requests_session` was created at module import time, so config file overrides were ignored. replaced with `_get_pkg_requests_session` that builds the session on first download call instead.

## Testing

ran manually with `http.max_retries: 10` in config file, confirmed the session picked up the value correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ Any settings specified will override the default values present in the
 - `http.read_timeout` read timeout (seconds) for HTTP requests (default: 300).
   Long-running downloads can take arbitrarily long as long as bytes keep
   flowing.
+- `http.max_retries` maximum number of retries for HTTP requests (default: 5)
 - `http.timeout` (deprecated) automatically migrated to `http.read_timeout`.
 - `runtime.concurrency_limit` max concurrent operations.
 - `runtime.subprocess_timeout` timeout (seconds) for subprocess commands.
--- a/hermeto/core/http_requests.py
+++ b/hermeto/core/http_requests.py
@@ -12,9 +12,6 @@ ALL_REQUEST_METHODS = frozenset(
 # The set includes only methods which don't modify state of the service.
 SAFE_REQUEST_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE"})
 DEFAULT_RETRY_OPTIONS: dict[str, Any] = {
-    "total": 5,
-    "read": 5,
-    "connect": 5,
     "backoff_factor": 1.3,
     "status_forcelist": (500, 502, 503, 504),
 }
--- a/hermeto/core/http_requests.py
+++ b/hermeto/core/http_requests.py
@@ -2,10 +2,6 @@
 import logging
 from typing import Any
 
-from requests import Session
-from requests.adapters import HTTPAdapter
-from urllib3.util.retry import Retry
-
 log = logging.getLogger(__name__)
 
 # The set is extended version of constant Retry.DEFAULT_ALLOWED_METHODS
@@ -22,21 +18,3 @@ DEFAULT_RETRY_OPTIONS: dict[str, Any] = {
     "backoff_factor": 1.3,
     "status_forcelist": (500, 502, 503, 504),
 }
-
-
-def get_requests_session(retry_options: dict | None = None) -> Session:
-    """
-    Create a requests session with retries.
-
-    :param dict retry_options: overwrite options for initialization of Retry instance
-    :return: the configured requests session
-    :rtype: requests.Session
-    """
-    if retry_options is None:
-        retry_options = {}
-    session = Session()
-    retry_options = {**DEFAULT_RETRY_OPTIONS, **retry_options}
-    adapter = HTTPAdapter(max_retries=Retry(**retry_options))
-    session.mount("http://", adapter)
-    session.mount("https://", adapter)
-    return session
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -9,19 +9,42 @@ from urllib.parse import urlparse
 import aiohttp
 import aiohttp_retry
 import requests
+from requests import Session
+from requests.adapters import HTTPAdapter
 from requests.auth import AuthBase
+from urllib3.util.retry import Retry
 
 from hermeto.core.config import get_config
 from hermeto.core.errors import FetchError
 from hermeto.core.http_requests import (
     DEFAULT_RETRY_OPTIONS,
     SAFE_REQUEST_METHODS,
-    get_requests_session,
 )
 from hermeto.core.scm import get_repo_id
 from hermeto.core.type_aliases import StrPath
 
-pkg_requests_session = get_requests_session(retry_options={"allowed_methods": SAFE_REQUEST_METHODS})
+_pkg_requests_session: requests.Session | None = None
+
+
+def _get_pkg_requests_session() -> requests.Session:
+    """
+    A lazy initialised, module-level requests.Session with retry config.
+    """
+    global _pkg_requests_session
+    if _pkg_requests_session is None:
+        max_retries = get_config().http.max_retries
+        retry_options = {
+            **DEFAULT_RETRY_OPTIONS,
+            "allowed_methods": SAFE_REQUEST_METHODS,
+            "total": max_retries,
+        }
+        _pkg_requests_session = Session()
+        adapter = HTTPAdapter(max_retries=Retry(**retry_options))
+        _pkg_requests_session.mount("http://", adapter)
+        _pkg_requests_session.mount("https://", adapter)
+
+    return _pkg_requests_session
+
 
 log = logging.getLogger(__name__)
 
@@ -46,7 +69,7 @@ def download_binary_file(
     config = get_config()
     timeout = (config.http.connect_timeout, config.http.read_timeout)
     try:
-        resp = pkg_requests_session.get(
+        resp = _get_pkg_requests_session().get(
             url, stream=True, verify=not insecure, auth=auth, timeout=timeout
         )
         resp.raise_for_status()
@@ -130,10 +153,12 @@ async def async_download_files(
     :param auth: Optional authorization data for proxies.
     """
     trace_config = aiohttp.TraceConfig()
-    num_attempts: int = int(DEFAULT_RETRY_OPTIONS["total"])
+    max_retries = get_config().http.max_retries
+    # aiohttp uses n calls (1 call, n-1 retries).
+    max_retries = max_retries + 1
     retry_options = aiohttp_retry.JitterRetry(
         start_timeout=DEFAULT_RETRY_OPTIONS["backoff_factor"],
-        attempts=num_attempts,
+        attempts=max_retries,
         statuses=set(DEFAULT_RETRY_OPTIONS["status_forcelist"]),
         exceptions={
             aiohttp.ClientConnectionError,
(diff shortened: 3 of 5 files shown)


hermeto/core/http_requests.py
```diff
@@ -35,8 +36,11 @@ def get_requests_session(retry_options: dict | None = None) -> Session:
     if retry_options is None:
         retry_options = {}
     session = Session()
-    retry_options = {**DEFAULT_RETRY_OPTIONS, **retry_options}
-    adapter = HTTPAdapter(max_retries=Retry(**retry_options))
+    max_retries = get_config().http.max_retries
+    count_options = {"total": max_retries}
+    print(f"[DEBUG] sync path max_retries={max_retries}", flush=True)
```
eskultety: This leaves a debug print in production code bypassing the default log-level.

hermeto/core/http_requests.py
```diff
@@ -35,8 +36,11 @@ def get_requests_session(retry_options: dict | None = None) -> Session:
...
+    merged = {**DEFAULT_RETRY_OPTIONS, **count_options, **retry_options}
```
eskultety: `total` is now user configurable, but `read` and `connect` retry limits remain hardcoded which either needs to be properly documented now that it's user visible OR we should expose the rest as well, but then again, neither `read` nor `connect` retry attributes have a corresponding pairing in aiohttp_retry.

hermeto/core/package_managers/general.py
```diff
@@ -130,10 +140,10 @@ async def async_download_files(
     :param auth: Optional authorization data for proxies.
     """
     trace_config = aiohttp.TraceConfig()
-    num_attempts: int = int(DEFAULT_RETRY_OPTIONS["total"])
+    max_retries = get_config().http.max_retries
     retry_options = aiohttp_retry.JitterRetry(
         start_timeout=DEFAULT_RETRY_OPTIONS["backoff_factor"],
-        attempts=num_attempts,
+        attempts=max_retries,
```
eskultety, line 161: Pre-existing, but now that you're exposing this setting to users we may want to address the misalignment in how many calls are actually made when e.g. `max_retries = 5`:

- urllib3 will use N+1 calls (1 call, N retries), you can verify by looking here:
https://github.com/urllib3/urllib3/blob/4864bd81c81ced5abfff5015f124c156d6080bbd/src/urllib3/connectionpool.py#L945 followed by https://github.com/urllib3/urllib3/blob/4864bd81c81ced5abfff5015f124c156d6080bbd/src/urllib3/util/retry.py#L541
- aiohttp will use exactly N calls:
https://github.com/inyutin/aiohttp_retry/blob/master/aiohttp_retry/client.py#L97-L103

hermeto/core/http_requests.py
```diff
@@ -35,8 +36,11 @@ def get_requests_session(retry_options: dict | None = None) -> Session:
     if retry_options is None:
         retry_options = {}
     session = Session()
-    retry_options = {**DEFAULT_RETRY_OPTIONS, **retry_options}
-    adapter = HTTPAdapter(max_retries=Retry(**retry_options))
+    max_retries = get_config().http.max_retries
```
eskultety: given this function consumes `retry_options` it should remain a simple helper independent of the global config state and it should receive the configurable value from the caller.

eskultety (changes requested):
I'm still not sure what to do with the `Retry` specific args: `read` and `connect` that do not have a counterpart in aiohttp_retry.

hermeto/core/package_managers/general.py
Harsh-2005d: renaming the field to max_attempts and subtracting when passed to urllib3 should work. will also document the workaround.

hermeto/core/http_requests.py
Harsh-2005d: since `total` takes precedence over the other two in [urllib3](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html). I think we can drop those values completely.

eskultety: @Harsh-2005d fair enough, we can replace them with a user configurable `total` with sensible default.

hermeto/core/package_managers/general.py
```diff
@@ -21,7 +21,28 @@
...
+    """
+    A lazy initialised, module-level requests.Session with retry config.
+
+    The total number of retries are set to max_attempts - 1,
+    as urllib3 uses N+1 calls (1 call,N retries).
+    """
+    global _pkg_requests_session
+    if _pkg_requests_session is None:
+        max_attempts = get_config().http.max_attempts
+        _pkg_requests_session = get_requests_session(
```
slimreaper35: The function seems like a simple wrapper around `get_requests_session` which is not used anywhere else. Can we merge them?

slimreaper35 (changes requested):
@Harsh-2005d Please do a proper rebase without merge commits.

hermeto/core/package_managers/general.py
eskultety: @Harsh-2005d I think we should define this in the opposite direction, i.e. not as `max_attempts` (attempts is a bit unexpected for users who would most likely expect setting for retries instead) to tie this with `aiohttp`, but instead `max_retries` to make `urrlib3` the authoritative implementation and adjust the logic for JitterRetry.

hermeto/core/http_requests.py
```diff
@@ -16,9 +16,6 @@
 # The set includes only methods which don't modify state of the service.
 SAFE_REQUEST_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE"})
 DEFAULT_RETRY_OPTIONS: dict[str, Any] = {
-    "total": 5,
-    "read": 5,
```
eskultety, line 20: Do not delete these 2 in the same commit please. Drop them in a standalone commit linking the urllib docs where it says `total` (our `max_retries`) takes precedence.

eskultety (changes requested):

hermeto/core/package_managers/general.py
Harsh-2005d: agreed, will merge them.

hermeto/core/package_managers/general.py
Harsh-2005d: Done!

Harsh-2005d:
@slimreaper35 Rebased.

hermeto/core/http_requests.py
Harsh-2005d: Done.

tests/unit/package_managers/test_general.py
```diff
@@ -19,15 +19,14 @@
...
-    pkg_requests_session,
 )
 from tests.common_utils import GIT_REF
 
 
 @pytest.mark.parametrize("auth", [None, HTTPBasicAuth("user", "password")])
 @pytest.mark.parametrize("insecure", [True, False])
 @pytest.mark.parametrize("chunk_size", [1024, 2048])
-@mock.patch.object(pkg_requests_session, "get")
+@mock.patch("hermeto.core.package_managers.general.requests.Session.get")
```
eskultety: this patches only `get` one the `Session` class globally, not on the instance we create in our helper. Can't we patch `_get_pkg_requests_session` itself? It'll require a few more changes to the existing test, but I feel we'd have more control over what the test actually does and tests. Additionally, we're currently not checking if `max_retries` (the core of this PR) was propagated to session args anywhere, so a small additional test would be needed.

eskultety (commented):
@Harsh-2005d we'll also need an update to our README, see https://github.com/hermetoproject/hermeto/blob/main/README.md#settings to make the config option user-visible. That and the unit test adjustments, then I'll approve.

README.md
```diff
@@ -145,6 +145,7 @@ Any settings specified will override the default values present in the
 - `http.read_timeout` read timeout (seconds) for HTTP requests (default: 300).
   Long-running downloads can take arbitrarily long as long as bytes keep
   flowing.
+- `http.max_retries` maximum number of retries for HTTP requests (default: 5)
```
slimreaper35, line 148: I believe this change belongs to the previous commit.

tests/unit/package_managers/test_general.py
```diff
@@ -10,52 +10,79 @@
 import aiohttp_retry
 import pytest
 import requests
+from requests.adapters import HTTPAdapter
 from requests.auth import AuthBase, HTTPBasicAuth
 
-from hermeto.core.config import get_config
+import hermeto.core.config as config_module
```
slimreaper35: Why?

README.md
eskultety: Yes.

tests/unit/package_managers/test_general.py
Harsh-2005d: just an unnecessary workaround. Fixed it.

tests/unit/package_managers/test_general.py
```diff
@@ -10,52 +10,79 @@
...
+    mock_session = mock.MagicMock()
+    mock_session.get.side_effect = requests.RequestException("Something went wrong")
+
+    with mock.patch(
+        "hermeto.core.package_managers.general._get_pkg_requests_session",
+        return_value=mock_session,
+    ):
+        expected = "Could not download http://example.org/example.tar.gz: Something went wrong"
+        with pytest.raises(FetchError, match=expected):
+            download_binary_file("http://example.org/example.tar.gz", "/example.tar.gz")
```
slimreaper35: We usually have mocks as decorators. I think it's more elegant approach and shorter.

```suggestion
@mock.patch("hermeto.core.package_managers.general._get_pkg_requests_session")
def test_download_binary_file_failed(mock_get_pkg_requests_session: MagicMock) -> None:
    mock_get_pkg_requests_session.return_value.get.side_effect = requests.RequestException()
    with pytest.raises(FetchError, match="Could not download http://example.org/example.tar.gz"):
        download_binary_file("http://example.org/example.tar.gz", "/example.tar.gz")
```

hermeto/core/package_managers/general.py
```diff
@@ -9,19 +9,43 @@
...
 from hermeto.core.type_aliases import StrPath
 
-pkg_requests_session = get_requests_session(retry_options={"allowed_methods": SAFE_REQUEST_METHODS})
+_pkg_requests_session: requests.Session | None = None
+
+
+def _get_pkg_requests_session() -> requests.Session:
+    """
+    A lazy initialised, module-level requests.Session with retry config.
+
```
slimreaper35: nitpick: Extra blank line.

slimreaper35 (approved):

tests/unit/package_managers/test_general.py
eskultety: True...as long as you don't need to patch dicts, vars, and attributes in which case the decorator approach is no longer feasible and even unittest's mock falls short of monkeypatch :) .

eskultety (approved):
@Harsh-2005d thank you for the contribution, appreciated!

PR #1571, opened 2026-05-19
Use a single PackageJson class

slimreaper35:


tests/unit/package_managers/yarn/test_project.py
```diff
@@ -140,7 +140,7 @@ def _prepare_package_json_file(rooted_tmp_path: RootedPath, data: str) -> Packag
     with open(path, "w") as f:
         f.write(data)
 
-    return PackageJson.from_file(path)
+    return PackageJson.from_file(path.path)
```
eskultety: `path.path` huh? I totally respect the cleaner diff, but maybe this is one of the cases where a change a few lines above to rename `path` to something like `rootedpath` is justified since `pathlib.Path` doesn't have a `path` attribute so it reads weird to anyone not familiar with `RootedPath`.
Non-blocking, just saying.

hermeto/core/package_managers/javascript/package_json.py
```diff
@@ -0,0 +1,48 @@
...
+class PackageJson(UserDict):
+    """Class representing package.json files."""
+
+    def __init__(self, path: Path, data: dict[str, Any]) -> None:
+        """Initialize a PackageJson object."""
+        self.path = path
+        super().__init__(data)
+
+    @classmethod
+    def from_file(cls, path: Path) -> "PackageJson":
```
eskultety, line 19: Previously we worked on top of our `RootedPath` abstraction, this change leads to (easily fixable) weirdness - `path.path`. What's the motivation to bypass our rooted path abstraction since our codebase uses it pretty much everywhere except for the elementary helpers which actually need to work with `Path` primitives. I also believe some of the `join_within_root/joinpath` changes in the diff could be dropped making the overall diff smaller.
I'm just curious, what led to this decision.

slimreaper35: > What's the motivation to bypass our rooted path abstraction since our codebase uses it pretty much everywhere except for the elementary helpers which actually need to work with Path primitives. I also believe some of the join_within_root/joinpath changes in the diff could be dropped making the overall diff smaller.
I'm just curious, what led to this decision.

I thought using the regular `Path` would be much more convenient, especially for unit tests, where I see `rooted_tmp_path.path.*` everywhere. If I am not mistaken, we discussed in some old PR. It makes things a bit inconsistent now. That would require a lot more refactoring, which is out of scope for this PR. So I am okay with changing it back to `RootedPath`.

eskultety: > hat would require a lot more refactoring, which is out of scope for this PR. So I am okay with changing it back to RootedPath.

So, the original classes used RootedPath so I'd argue that NOT changing it right now is remaining consistent. That said, if  cleaning it up properly everywhere  would require significant amount of refactoring, I agree that's out of scope, maybe worth tracking it in the refactoring tracker if it's something we see value in?

tests/unit/package_managers/yarn/test_project.py
slimreaper35: So, looking at the original PackageJson class in yarn v2. There is the same pattern, e.g. `with self._path.path.open("w") as f`. So I would prefer to keep the regular Path object in the new PackageJson class. It would be just moving the problem elsewhere.

I can fix the case above with the following:

```
def _prepare_package_json_file(rooted_tmp_path: RootedPath, data: str) -> PackageJson:
    path = rooted_tmp_path.path.joinpath("package.json")

    with open(path, "w") as f:
        f.write(data)

    return PackageJson.from_file(path)
```

hermeto/core/package_managers/javascript/package_json.py
```diff
@@ -0,0 +1,48 @@
...
+        """Create a PackageJson object from a package.json file."""
+        if not path.exists():
+            raise LockfileNotFound(
+                path,
+                solution="Make sure the package.json file exists in the specified directory.",
+            )
+
+        try:
+            with path.open("r") as f:
```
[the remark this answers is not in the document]
eskultety: Fair enough, but this is not a medium priority item. Since encoding is platform dependent by default, it's reasonable, though similar changes to other modules should be done outside this PR.

hermeto/core/package_managers/javascript/package_json.py
eskultety: I don't feel too strong about this, I can live with the current approach just fine.

eskultety (approved):

hermeto/core/package_managers/javascript/package_json.py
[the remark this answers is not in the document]
slimreaper35: added

hermeto/core/package_managers/javascript/package_json.py
a-ovchinnikov, line 1: _Commit message for_ https://github.com/hermetoproject/hermeto/commit/4bb0964c3f8d811f07f413ba26eb380ee103ca4f  : a non-blocking nitpick, could you please mention that this applies to npm-based package managers? (I.e. something like s/across different package managers/ across different npm-based package managers/)

a-ovchinnikov (approved):
LGTM with one non-blocking suggestion for a commit message wording.

hermeto/core/package_managers/javascript/package_json.py
slimreaper35: Ok. Reworded to `across different JavaScript package managers (npm, pnpm, yarn v1, yarn v2+)`.

a-ovchinnikov: Thank you!

PR #1574, opened 2026-05-20
Python and Rust extensions improvements

slimreaper35:


tests/integration/test_pip.py
```diff
@@ -209,6 +209,17 @@ def test_pip_packages(
...
+        pytest.param(
+            utils.TestParameters(
+                repo_url="https://github.com/slimreaper35/integration-tests",
+                # branch="pip/e2e-rust-git-dependency",
+                branch="cryptography-git",
+                packages=(({"type": "pip"}, {"type": "rpm"})),
+            ),
+            [],
+            [],
+            id="pip_e2e_rust_git_dependency",
```
slimreaper35: Maybe this could be part of the existing pip_e2e_rust test 🤔. Both tests take a lot of time and produce huge SBOMs.

tests/integration/test_pip.py
```diff
@@ -209,6 +209,17 @@ def test_pip_packages(
             [],
             id="pip_e2e_rust_extensions",
         ),
+        pytest.param(
+            utils.TestParameters(
+                repo_url="https://github.com/slimreaper35/integration-tests",
+                # branch="pip/e2e-rust-git-dependency",
+                branch="cryptography-git",
```
eskultety: Some development remnant? :)

tests/integration/test_data/pip_e2e_rust_git_dependency/container/Containerfile
```diff
@@ -0,0 +1,18 @@
+FROM registry.access.redhat.com/ubi10/ubi@sha256:eb702361cac064da1e7fb9b09eb030562a832209ed1fb1d6765d617ed6294e61
```
eskultety: Does pinning UBI to a hash in tests make a difference? We need version pinning in some specific ITs but these UBI base images don't need pinning IMO. At this point I'm wondering how consistent we are with these in tests, it's easily missed during reviews. I have a preference, but no strong blocking opinion on this, it's just latest should be safe and fine here.

hermeto/core/package_managers/pip/rust.py
```diff
@@ -155,10 +172,11 @@ def remove_extracted(packages: list[CargoPackageInput]) -> None:
...
 
+        template = _merge_cargo_config_files(result.build_config.project_files)
         # A config pointing to deps/cargo directory and an environment variable
         # poiting to the config are necessary for pip to be able to build the extension.
         ev = [EnvironmentVariable(name="CARGO_HOME", value="${output_dir}/.cargo")]
-        pf = [ProjectFile(abspath=_config_path(request), template=_config_data())]
+        pf = [ProjectFile(abspath=_config_path(request), template=template)]
 
         remove_extracted(packages_containing_rust_code)
         return result + RequestOutput.from_obj_list([], ev, pf)
```
eskultety: shouldn't we clear the original project files templates we've accumulated so far now that we merged their configuration to the main config? Since it's pre-existing it should probably happen in a standalone 
commit preceding this one so that we build off of a clean state.

See also my other comment to the build-config template.

tests/integration/test_data/pip_e2e_rust_git_dependency/.build-config.yaml
```diff
@@ -0,0 +1,47 @@
...
+    [source.crates-io]
+    replace-with = "vendored-sources"
+
+    [source."git+https://github.com/pyca/cryptography.git?tag=45.0.4"]
+    git = "https://github.com/pyca/cryptography.git"
+    tag = "45.0.4"
+    replace-with = "vendored-sources"
+
+    [source.vendored-sources]
+    directory = "${output_dir}/deps/cargo"
```
eskultety: Pre-existing, but IIUC these all point to non-existing locations that have been wiped when we processed all cargo extensions. The `inject-files` command will re-create these empty directory hierarchies with the config files in them, but only the top-level cargo config one is what `cargo` will actually consider during hermetic builds.

tests/integration/test_pip.py
eskultety: Shouldn't this be consistent with the e2e test above? IOW in the above test we test both a build as well as correct install, here we only verified the build which seemingly suffices, but then the question is, shouldn't the test above also drop its `check_cmd`?

hermeto/core/package_managers/pip/rust.py
```diff
@@ -155,10 +159,11 @@ def remove_extracted(packages: list[CargoPackageInput]) -> None:
...
```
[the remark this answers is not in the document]
eskultety: Oh, finally a useful comment from the bot, I didn't notice its response. ACK, we should address this (it's not a high severity item though).

tests/integration/test_pip.py
eskultety: Hmm, yeah, why not, a compound test is what e2e is about!

eskultety:
Something I forgot to mention in my review message - the ramp-up on these changes was a bit slow from a reviewer perspective (subjectively) and it required some tracing and understanding what the exact issue with git deps is, etc. Since there is no issue in our tracker I'd suggest enhacing both the PR description as well as the first commit message with some basic info, i.e. if transitive deps have git dependencies, it's not that we don't support them AT ALL, it's because of our former logic that recorded and "enabled" git deps in the respective per-package config files which the cargo in the offline build environment would never consider!

tests/integration/test_pip.py
slimreaper35: That's the branch I created in my fork, but didn't merge 😬

tests/integration/test_data/pip_e2e_rust_git_dependency/container/Containerfile
slimreaper35: We have to pin the image as we do for other ITs. The rpms lockfile would become outdated, and the test would randomly fail.

tests/integration/test_pip.py
slimreaper35: Looks like, we agree to merge both tests.

tests/integration/test_data/pip_e2e_rust_git_dependency/container/Containerfile
eskultety: > We have to pin the image as we do for other ITs.

I suppose you mean ones that use the UBI base image, correct? Because in general, this statement is not true - the vast majority of test base images are digest-unpinned.

> The rpms lockfile would become outdated, and the test would randomly fail.

the lockfile will become stale anyway, base image pinning doesn't prevent that because the DNF repos aren't connected to digests in any way and UBI doesn't maintain more than a single version for a package.
_Edit: so the next time a package we depend on gets a version update a given lockfile will break our tests._

slimreaper35: > I suppose you mean ones that use the UBI base image, correct?

Yes.

eskultety: Fair enough, in the name of consistency, let's keep it pinned, just keep in mind that pinning in the base images doesn't protect us from what you're describing.

taylormadore (commented):
I have nothing new to add beyond the items that @eskultety already mentioned. After those threads are resolved, LGTM

hermeto/core/package_managers/pip/rust.py
```diff
@@ -8,6 +8,8 @@
 from textwrap import dedent
 from typing import Any
 
+import tomlkit
```
eskultety, line 10: **Commit message:**

> Make sure all config files are merged correctly

I would end the sentence ^here. The rest of that sentence is debatable - you're doing a `dict.update` which does "resolve" duplicate keys, but it's not a lossless operation, the existing key is being overwritten (which is okay for the time being).
 
> Add a new function to merge multiple Cargo config project files into a
> single file for all Rust packages. They can have git dependencies that
> are not handled by a static config data. Those project files are read
> from the stdout directly after vendoring.

Please add some more contenxt and a reference to the previous commit that with fixing the N template files we are now easily able to introduce proper support for git dependencies, see my review comment https://github.com/hermetoproject/hermeto/pull/1574#issuecomment-4574819485 . Since we don't have an upstream issue filed for this, context is everything.

eskultety (approved):
ACK with adding a bit more context to the change.

eskultety:
I'm good with merging once this passes the new integration test we just added.

a-ovchinnikov (approved):

tests/integration/test_pip.py
```diff
@@ -199,13 +199,13 @@ def test_pip_packages(
         # (/etc/yum.repos.d/ubi.repo) and extracted to a local directory.
         pytest.param(
             utils.TestParameters(
-                branch="pip/e2e_rust_extensions",
+                repo_url="https://github.com/slimreaper35/integration-tests",
```
taylormadore: Let's just not forget to switch this to the real test now

taylormadore (approved):

PR #1575, opened 2026-05-20
Artifact registry proxy support in gomod

a-ovchinnikov:
This PR introduces necessary changes to support fetching dependencies via an artifact registry proxies in gomod.

### DIFF
--- a/hermeto/core/package_managers/gomod/main.py
+++ b/hermeto/core/package_managers/gomod/main.py
@@ -55,7 +55,7 @@ from hermeto.core.package_managers.gomod.go import (
 )
 from hermeto.core.package_managers.gomod.utils import _clean_go_modcache, _go_exec_env
 from hermeto.core.rooted_path import RootedPath
-from hermeto.core.scm import GitRepo, get_repo_for_path, get_repo_id
+from hermeto.core.scm import GitRepo, RepoID, get_repo_for_path, get_repo_id
 from hermeto.core.utils import GIT_PRISTINE_ENV, load_json_stream
 from hermeto.interface.logging import EnforcingModeLoggerAdapter
 
@@ -127,6 +127,8 @@ class Module(NamedTuple):
         had a checksum for this module but didn't
     proxy: the list of custom proxy URLs used to fetch this module, or None if the module was
         fetched directly from VCS or only via the canonical Go proxy (proxy.golang.org).
+    repo_id: optional RepoID to carry data necessary to construct vcs_url
+    local_replace: a flag to mark Modules which were obtained via a replace with a local path.
     """
 
     name: str
@@ -136,6 +138,8 @@ class Module(NamedTuple):
     main: bool = False
     missing_hash_in_file: Path | None = None
     proxy: list[str] | None = None
+    repo_id: RepoID | None = None
+    local_replace: bool = False
 
     @property
     def purl(self) -> str:
@@ -144,7 +148,7 @@ class Module(NamedTuple):
             type="golang",
             name=self.real_path,
             version=self.version,
-            qualifiers={"type": "module"},
+            qualifiers={"type": "module"} | _vcs_url_if_needed(self),
         )
         return purl.to_string()
 
@@ -204,13 +208,24 @@ class Package(NamedTuple):
             type="golang",
             name=self.real_path,
             version=self.module.version,
-            qualifiers={"type": "package"},
+            qualifiers={"type": "package"} | _vcs_url_if_needed(self.module),
         )
         return purl.to_string()
 
     def to_component(self) -> Component:
         """Create a SBOM component for this package."""
         return Component(name=self.name, version=self.module.version, purl=self.purl)
+
+
+def _vcs_url_if_needed(module: Module) -> dict:
+    """Determine whether a vcs_url qualifier is necessary and provide one."""
+    if (module.main or module.local_replace) and (module.repo_id is not None):
+        # The main module is always consumed from a local filesystem and it is always
+        # a clone of a repository, thus adding vcs_url to it.
+        # Local replaces need a vcs_url as well and are supposed to be consumed from the file
+        # system too.
+        return {"vcs_url": module.repo_id.as_vcs_url_qualifier()}
+    return dict()
 
 
 class StandardPackage(NamedTuple):
@@ -296,6 +311,7 @@ def _create_modules_from_parsed_data(
         original_name = module.path
         missing_hash_in_file = None
 
+        repo_id, proxy, local_replace = None, None, False
         if not version_or_path.startswith("."):
             version = version_or_path
             real_path = name
@@ -314,7 +330,8 @@ def _create_modules_from_parsed_data(
             resolved_replacement_path = main_module_dir.join_within_root(version_or_path)
             version = version_resolver.get_golang_version(module.path, resolved_replacement_path)
             real_path = _resolve_path_for_local_replacement(module)
-            proxy = None
+            local_replace = True
+            repo_id = get_repo_id(main_module_dir.path)
 
         return Module(
             name=name,
@@ -323,6 +340,8 @@ def _create_modules_from_parsed_data(
             real_path=real_path,
             missing_hash_in_file=missing_hash_in_file,
             proxy=proxy,
+            local_replace=local_replace,
+            repo_id=repo_id,
         )
 
     def _resolve_path_for_local_replacement(module: ParsedModule) -> str:
@@ -561,10 +580,13 @@ def _create_main_module_from_parsed_data(
     if repo_name is None:
         # PERMISSIVE mode without git repo - use the module path as resolved_path
         resolved_path = parsed_main_module.path
+        repo_id = None
     elif str(resolved_subpath) == ".":
         resolved_path = repo_name
+        repo_id = get_repo_id(main_module_dir)
     else:
         resolved_path = f"{repo_name}/{resolved_subpath}"
+        repo_id = get_repo_id(main_module_dir)
 
     if not parsed_main_module.version:
         # Should not happen, since the version is always resolved from the Git repo
@@ -576,6 +598,7 @@ def _create_main_module_from_parsed_data(
         version=parsed_main_module.version,
         real_path=resolved_path,
         main=True,
+        repo_id=repo_id,
     )
 
 
--- a/hermeto/core/package_managers/gomod/main.py
+++ b/hermeto/core/package_managers/gomod/main.py
@@ -227,7 +227,7 @@ class Package(NamedTuple):
         )
 
 
-def _vcs_url_if_needed(module: Module) -> dict:
+def _vcs_url_if_needed(module: Module) -> dict[str, str]:
     """Determine whether a vcs_url qualifier is necessary and provide one."""
     if (module.main or module.local_replace) and (module.repo_id is not None):
         # The main module is always consumed from a local filesystem and it is always
@@ -728,6 +728,82 @@ def _parse_packages(
     return iter(all_packages)
 
 
+def _determine_cache_path(should_vendor: bool, tmp_dir: Path) -> str:
+    if should_vendor:
+        # Even though we do not perform a "go mod download" when vendoring is detected, some
+        # go commands still download dependencies as a side effect. Since we don't want those
+        # copied to the output dir, we need to set the GOMODCACHE to a different directory.
+        return f"{tmp_dir}/vendor-cache"
+    return f"{tmp_dir}/pkg/mod"
+
+
+def _prepare_run_params(
+    request_flags: Iterable[str],
+    should_vendor: bool,
+    tmp_dir: Path,
+    app_dir: RootedPath,
+    temp_netrc_dir: Path,
+) -> dict:
+    config = get_config()
+    gomod_cache = _determine_cache_path(should_vendor, tmp_dir)
+
+    go_vars: dict[str, str] = {
+        "GOPATH": str(tmp_dir),
+        "GO111MODULE": "on",
+        "GOCACHE": str(tmp_dir),
+        "GOMODCACHE": gomod_cache,
+        "GOSUMDB": "sum.golang.org",
+        "GOTOOLCHAIN": "auto",
+    }
+    if config.gomod.proxy_url:
+        go_vars["GOPROXY"] = config.gomod.proxy_url
+
+    if config.gomod.proxy_login is not None:
+        # This is supposed to be happening only in systems utilizing artifact registry proxies.
+        if "," in config.gomod.proxy_url:
+            raise InvalidInput(
+                "proxy_url must contain exactly one URL when login is present. "
+                f"Contains now: {config.gomod.proxy_url}"
+            )
+        path_to_netrc = inject_netrc(prepare_netrc_contents(), temp_netrc_dir)
+        go_vars["NETRC"] = path_to_netrc
+        if os.environ.get("NETRC", ""):
+            log.warning(
+                "Proxy login was specified, at the same time a custom .netrc file path was"
+                " provided via environment. A new .netrc will be generated from proxy"
+                " parameters and will shadow the environment one. Please disable proxy"
+                " support if you intend to use the .netrc from environment."
+            )
+
+    if "cgo-disable" in request_flags:
+        go_vars["CGO_ENABLED"] = "0"
+
+    env = _go_exec_env(**go_vars)
+    run_params = {"env": env, "cwd": app_dir}
+
+    return run_params
+
+
+def _determine_which_modules_go_in_go_sum(
+    app_dir: RootedPath,
+    go_work: GoWork | None,
+) -> frozenset[ModuleID]:
+    if go_work:
+        return _parse_go_sum_from_workspaces(go_work)
+    return _parse_go_sum(app_dir.join_within_root("go.sum"))
+
+
+def _acquire_modules(
+    should_vendor: bool, go: Go, app_dir: RootedPath, go_work: GoWork | None, run_params: dict
+) -> Iterable[ParsedModule]:
+    # Vendor dependencies if the gomod-vendor flag is set
+    if should_vendor:
+        return _vendor_deps(go, app_dir, bool(go_work), run_params)
+    log.info("Downloading the gomod dependencies")
+    json_stream = go(["mod", "download", "-json"], run_params, retry=True)
+    return (ParsedModule.model_validate(obj) for obj in load_json_stream(json_stream))
+
+
 def _resolve_gomod(
     app_dir: RootedPath,
     request: Request,
@@ -750,86 +826,31 @@ def _resolve_gomod(
     :raises PackageManagerError: if fetching dependencies fails
     """
     _protect_against_symlinks(app_dir)
-
-    config = get_config()
-
     should_vendor = app_dir.join_within_root("vendor").path.is_dir()
 
-    if should_vendor:
-        # Even though we do not perform a "go mod download" when vendoring is detected, some
-        # go commands still download dependencies as a side effect. Since we don't want those
-        # copied to the output dir, we need to set the GOMODCACHE to a different directory.
-        gomod_cache = f"{tmp_dir}/vendor-cache"
-    else:
-        gomod_cache = f"{tmp_dir}/pkg/mod"
-
-    go_vars: dict[str, str] = {
-        "GOPATH": str(tmp_dir),
-        "GO111MODULE": "on",
-        "GOCACHE": str(tmp_dir),
-        "GOMODCACHE": gomod_cache,
-        "GOSUMDB": "sum.golang.org",
-        "GOTOOLCHAIN": "auto",
-    }
-    if config.gomod.proxy_url:
-        go_vars["GOPROXY"] = config.gomod.proxy_url
-
-    temp_netrc_dir = TemporaryDirectory()
-    if config.gomod.proxy_login is not None:
-        # This is supposed to be happening only in systems utilizing artifact registry proxies.
-        if "," in config.gomod.proxy_url:
-            raise InvalidInput(
-                "proxy_url must contain exactly one URL when login is present. "
-                f"Contains now: {config.gomod.proxy_url}"
-            )
-        path_to_netrc = inject_netrc(prepare_netrc_contents(), Path(temp_netrc_dir.name))
-        go_vars["NETRC"] = path_to_netrc
-        if os.environ.get("NETRC", ""):
-            log.warning(
-                "Proxy login was specified, at the same time a custom .netrc file path was"
-                " provided via environment. A new .netrc will be generated from proxy"
-                " parameters and will shadow the environment one. Please disable proxy"
-                " support if you intend to use the .netrc from environment."
-            )
-
-    if "cgo-disable" in request.flags:
-        go_vars["CGO_ENABLED"] = "0"
-
-    env = _go_exec_env(**go_vars)
-    run_params = {"env": env, "cwd": app_dir}
-
-    # Explicitly disable toolchain telemetry for go >= 1.23
-    _disable_telemetry(go, run_params)
-
-    if go_work:
-        modules_in_go_sum = _parse_go_sum_from_workspaces(go_work)
-    else:
-        modules_in_go_sum = _parse_go_sum(app_dir.join_within_root("go.sum"))
-
-    # Vendor dependencies if the gomod-vendor flag is set
-    if should_vendor:
-        downloaded_modules = _vendor_deps(go, app_dir, bool(go_work), run_params)
-    else:
-        log.info("Downloading the gomod dependencies")
-        downloaded_modules = (
-            ParsedModule.model_validate(obj)
-            for obj in load_json_stream(go(["mod", "download", "-json"], run_params, retry=True))
+    with TemporaryDirectory() as temp_netrc_dir:
+        run_params = _prepare_run_params(
+            request.flags, should_vendor, tmp_dir, app_dir, Path(temp_netrc_dir)
         )
 
-    main_module, workspace_modules = _parse_local_modules(
-        go_work, go, run_params, app_dir, version_resolver
-    )
+        # Explicitly disable toolchain telemetry for go >= 1.23
+        _disable_telemetry(go, run_params)
+        modules_in_go_sum = _determine_which_modules_go_in_go_sum(app_dir, go_work)
+        downloaded_modules = _acquire_modules(should_vendor, go, app_dir, go_work, run_params)
 
-    deps = _go_list_deps(go, "all", run_params)
-    package_modules = [pkg.module for pkg in deps if pkg.module and not pkg.module.main]
-    package_modules.extend(workspace_modules)
-    all_modules = _deduplicate_resolved_modules(package_modules, downloaded_modules)
-    _validate_local_replacements(all_modules, app_dir)
+        main_module, workspace_modules = _parse_local_modules(
+            go_work, go, run_params, app_dir, version_resolver
+        )
 
-    log.info("Retrieving the list of packages")
-    all_packages = _parse_packages(go_work, go, run_params)
+        deps = _go_list_deps(go, "all", run_params)
+        package_modules = [pkg.module for pkg in deps if pkg.module and not pkg.module.main]
+        package_modules.extend(workspace_modules)
+        all_modules = _deduplicate_resolved_modules(package_modules, downloaded_modules)
+        _validate_local_replacements(all_modules, app_dir)
 
-    temp_netrc_dir.cleanup()
+        log.info("Retrieving the list of packages")
+        all_packages = _parse_packages(go_work, go, run_params)
+
     return ResolvedGoModule(main_module, all_modules, all_packages, modules_in_go_sum)
 
 
--- a/hermeto/core/package_managers/gomod/main.py
+++ b/hermeto/core/package_managers/gomod/main.py
@@ -575,6 +575,7 @@ def _create_main_module_from_parsed_data(
         original_name=parsed_main_module.path,
         version=parsed_main_module.version,
         real_path=resolved_path,
+        main=True,
     )
 
 
--- a/hermeto/core/package_managers/gomod/main.py
+++ b/hermeto/core/package_managers/gomod/main.py
@@ -4,11 +4,13 @@ import os
 import re
 import shutil
 import tempfile
+import urllib
 from collections.abc import Iterable, Iterator
 from datetime import datetime, timezone
 from functools import cached_property
 from itertools import chain
 from pathlib import Path
+from tempfile import TemporaryDirectory
 from types import TracebackType
 from typing import TYPE_CHECKING, Any, Literal, NamedTuple, NoReturn, Optional
 
@@ -26,6 +28,7 @@ from hermeto.core.config import get_config
 from hermeto.core.errors import (
     FetchError,
     GitError,
+    InvalidInput,
     LockfileNotFound,
     NotAGitRepo,
     PackageManagerError,
@@ -152,6 +155,13 @@ class Module(NamedTuple):
         )
         return purl.to_string()
 
+    def get_external_refs(self) -> list[ExternalReference] | None:
+        """Get external references for a module."""
+        ref_rest = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
+        if self.proxy:
+            return [ExternalReference(url=p, **ref_rest) for p in self.proxy]
+        return None
+
     def to_component(self) -> Component:
         """Create a SBOM component for this module."""
         if self.missing_hash_in_file:
@@ -159,12 +169,7 @@ class Module(NamedTuple):
         else:
             missing_hash_in_file = frozenset()
 
-        ref_rest = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
-
-        if self.proxy:
-            refs = [ExternalReference(url=p, **ref_rest) for p in self.proxy]
-        else:
-            refs = None
+        refs = self.get_external_refs()
 
         return Component(
             name=self.name,
@@ -214,7 +219,12 @@ class Package(NamedTuple):
 
     def to_component(self) -> Component:
         """Create a SBOM component for this package."""
-        return Component(name=self.name, version=self.module.version, purl=self.purl)
+        return Component(
+            name=self.name,
+            version=self.module.version,
+            purl=self.purl,
+            external_references=self.module.get_external_refs(),
+        )
 
 
 def _vcs_url_if_needed(module: Module) -> dict:
@@ -244,6 +254,8 @@ class StandardPackage(NamedTuple):
 
     def to_component(self) -> Component:
         """Create a SBOM component for this package."""
+        # Note, these components come from the standard lib, they lack version, they
+        # are not collected from a repository and are not present in the output directory.
         return Component(name=self.name, purl=self.purl)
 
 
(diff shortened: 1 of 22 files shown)


hermeto/core/package_managers/gomod/main.py
```diff
@@ -1417,3 +1475,24 @@
...
+        login {config.gomod.proxy_login}
+        password {config.gomod.proxy_password}"""
+    )
+
+
+def inject_netrc(netrc_stuff: str) -> tuple[str, TemporaryDirectory[str]]:
+    """Inject a temporary .netrc."""
+    tmpdir = TemporaryDirectory()
+    netrc = Path(tmpdir.name) / ".netrc"
+    netrc.write_text(netrc_stuff)
```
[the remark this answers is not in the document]
a-ovchinnikov: I am sorry, I will consider applying rot13 next time.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -204,13 +208,30 @@ def purl(self) -> str:
...
+
+
+def _vcs_url_if_needed(module: Module) -> dict:
+    """Determine whether a vcs_url quialifier is necessary and provide one."""
+    if module.main or module.local_replace:
+        # The main module is always consumed from a local filesystem and it is always
+        # a clone of a repository, thus adding vcs_url to it.
+        # Local replaces need a vcs_url as well and are supposed to be consumed from the file
+        # system too.
+        # Ignore because at this poing repo_id cannot be None, but that is not something mypy can
```
taylormadore: typo: point

hermeto/core/package_managers/gomod/main.py
```diff
@@ -204,13 +208,30 @@ def purl(self) -> str:
...
-        return Component(name=self.name, version=self.module.version, purl=self.purl)
+        return Component(
+            name=self.name,
+            version=self.module.version,
+            purl=self.purl,
+        )
+
+
+def _vcs_url_if_needed(module: Module) -> dict:
+    """Determine whether a vcs_url quialifier is necessary and provide one."""
```
taylormadore: typo: qualifier

hermeto/core/package_managers/gomod/main.py
```diff
@@ -4,11 +4,14 @@
 import re
 import shutil
 import tempfile
+import urllib
 from collections.abc import Iterable, Iterator
 from datetime import datetime, timezone
 from functools import cached_property
 from itertools import chain
 from pathlib import Path
+from tempfile import TemporaryDirectory
```
taylormadore, line 13: tempfile already imported above

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1447,3 +1459,24 @@ def _vendor_changed(context_dir: RootedPath) -> bool:
...
     return False
+
+
+def prepare_netrc_contents() -> str:
+    """Prepare .netrc contents."""
+    config = get_config()
+    # it should be `machine foo.bar.baz` in netrc, but is
+    # set to https://foo.bar.baz/repository/go-proxy/ via a variable.
+    machine = urllib.parse.urlparse(config.gomod.proxy_url).netloc
+    return dedent(
```
taylormadore: Is dedent doing anything here?

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1447,3 +1459,24 @@ def _vendor_changed(context_dir: RootedPath) -> bool:
...
+        f"""machine {machine}
+        login {config.gomod.proxy_login}
+        password {config.gomod.proxy_password}"""
+    )
+
+
+def inject_netrc(netrc_stuff: str) -> tuple[str, TemporaryDirectory[str]]:
```
taylormadore: Do we need to worry about any interaction/conflict between this netrc injection and the one we support specifying via the NETRC env var? As currently implemented, it looks like the proxy netrc injection will replace the other. Not saying that is wrong, just something we should be aware of.

tests/integration/proxy.py
```diff
@@ -138,6 +141,23 @@ def _strip_proxy_refs(
         del component_raw["externalReferences"]
 
 
+def should_skip_this_component(component: Component) -> bool:
```
taylormadore: I think we could try to merge these conditions into a (renamed) `_is_registry_component`. They are essentially trying to accomplish the same thing--except that the reason I didn't just skip registry components before was so that we could also assert that components that shouldn't have proxy URLs don't have them

hermeto/core/package_managers/gomod/main.py
```diff
@@ -773,35 +753,71 @@ def _resolve_gomod(
...
+    :return: a dict containing the Go module itself ("module" key), the list of dictionaries
+        representing the dependencies ("module_deps" key), the top package level dependency
+        ("pkg" key), and a list of dictionaries representing the package level dependencies
+        ("pkg_deps" key)
+    :raises PackageManagerError: if fetching dependencies fails
+    """
+    _protect_against_symlinks(app_dir)
+    should_vendor = app_dir.join_within_root("vendor").path.is_dir()
+
+    run_params, _ = _prepare_run_params(request.flags, should_vendor, tmp_dir, app_dir)
```
taylormadore: Should we be discarding that TemporaryDirectory here?

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Fixed.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Fixed.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Apparently not, it was more useful in the previous iteration when the template started with a newline, here it can be dropped.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Having two .netrcs sounds like an error condition to me, to the best of my understanding one would want to supply one if they have some custom authentication in mind, at the same time using gomod with a proxy would require another authentication mechanism for a likely different server. I think I will add a warning to let a user know that an override took place

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Yes, I wanted to have a name without a path.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Oops, thank you for catching this! For some reason I was absolutely sure that it autocleans when it leaves the scope! I'll have to fix it.

tests/integration/proxy.py
a-ovchinnikov: Good point, I'll update the `_is_registry_component`. I don't think it should be renamed, that's precisely what we will be checking here

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1417,3 +1486,21 @@
...
+    machine = urllib.parse.urlparse(config.gomod.proxy_url).netloc
+    return f"""machine {machine}
+        login {config.gomod.proxy_login}
+        password {config.gomod.proxy_password}"""
+
+
+def inject_netrc(netrc_stuff: str, temp_netrc_dir: Path) -> str:
+    """Inject a temporary .netrc."""
+    netrc = temp_netrc_dir / ".netrc"
+    netrc.write_text(netrc_stuff)
```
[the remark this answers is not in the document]
a-ovchinnikov: This is getting tiring.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1417,3 +1486,21 @@ def _vendor_changed(context_dir: RootedPath) -> bool:
...
+
+def prepare_netrc_contents() -> str:
+    """Prepare .netrc contents."""
+    config = get_config()
+    # it should be `machine foo.bar.baz` in netrc, but is
+    # set to https://foo.bar.baz/repository/go-proxy/ via a variable.
+    machine = urllib.parse.urlparse(config.gomod.proxy_url).netloc
+    return f"""machine {machine}
+        login {config.gomod.proxy_login}
+        password {config.gomod.proxy_password}"""
```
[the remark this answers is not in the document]
a-ovchinnikov: This has to be fixed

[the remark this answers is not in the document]
a-ovchinnikov: Or not, the intended use-case is to have precisely one artifact registry proxy. I'll have to mention this in a comment somewhere.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -204,13 +210,31 @@ def purl(self) -> str:
...
+    """Determine whether a vcs_url qualifier is necessary and provide one."""
+    if module.main or module.local_replace:
+        # The main module is always consumed from a local filesystem and it is always
+        # a clone of a repository, thus adding vcs_url to it.
+        # Local replaces need a vcs_url as well and are supposed to be consumed from the file
+        # system too.
+        # Ignore because at this point repo_id cannot be None, but that is not something mypy can
+        # infer.
+        return {"vcs_url": module.repo_id.as_vcs_url_qualifier()}  # type: ignore
+    return dict()
```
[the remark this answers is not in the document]
a-ovchinnikov: Failure condition should be unreachable in practice (see the purpose of permissive mode).

hermeto/core/package_managers/gomod/main.py
```diff
@@ -4,11 +4,13 @@
 import re
 import shutil
 import tempfile
+import urllib
```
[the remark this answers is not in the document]
a-ovchinnikov: You don't say!
Screenshot from 2026-05-27 11-54-59

tests/integration/proxy.py
```diff
@@ -88,7 +91,18 @@ def _is_registry_component(component: Component) -> bool:
...
+            component.purl.startswith("pkg:golang")
+            and (
+                (component.version is None and component.external_references is None)
+                or "vcs_url" in component.purl
+            )
+        ),
+    )
+
+    if is_bundled or any(is_local):
         return False
```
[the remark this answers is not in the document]
a-ovchinnikov: This is wrong on so many levels that I am not interested in even touching it.

hermeto/core/package_managers/gomod/main.py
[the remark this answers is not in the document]
taylormadore: Is it unreachable though or am I missing something? We started allowing non-git sources.
```
tmadore@neon:/tmp/test-gomod$ hermeto --mode=permissive fetch-deps gomod
2026-05-28 15:34:10,161 INFO Fetching the gomod dependencies at subpath .
2026-05-28 15:34:10,181 INFO Downloading the gomod dependencies
2026-05-28 15:34:11,846 INFO Retrieving the list of packages
2026-05-28 15:34:11,922 ERROR AttributeError: 'NoneType' object has no attribute 'as_vcs_url_qualifier'
```

tests/integration/proxy.py
```diff
@@ -88,7 +91,18 @@ def _is_registry_component(component: Component) -> bool:
         p.name == PropertyEnum.PROP_CDX_NPM_PACKAGE_BUNDLED and p.value == "true"
         for p in component.properties
     )
-    if is_bundled:
+
+    is_local = (
+        (  # local traits for go mod:
+            component.purl.startswith("pkg:golang")
+            and (
+                (component.version is None and component.external_references is None)
```
taylormadore: What is the significance of the `and component.external_references is None`? Doesn't `component.version is None` alone identify StandardPackages?

hermeto/core/package_managers/gomod/main.py
```diff
@@ -250,6 +253,8 @@ def purl(self) -> str:
 
     def to_component(self) -> Component:
         """Create a SBOM component for this package."""
+        # Note, these copmonents come from the standard lib, they lack version, they
```
taylormadore: typo: components

hermeto/core/package_managers/gomod/main.py
```diff
@@ -767,6 +772,21 @@ def _resolve_gomod(
     if config.gomod.proxy_url:
         go_vars["GOPROXY"] = config.gomod.proxy_url
 
+    temp_netrc_dir = TemporaryDirectory()
+    if config.gomod.proxy_login is not None:
+        # This is supposed to be happening only in systems utilizing artifact registry proxies.
+        if "," in config.gomod.proxy_url:
+            raise ValueError("Multiple authentication requiring proxy URLs detected!")
```
taylormadore: Maybe` InvalidInput` instead of ValueError?

hermeto/core/package_managers/gomod/main.py
```diff
@@ -787,30 +770,69 @@ def _resolve_gomod(
...
-    if "cgo-disable" in request.flags:
+    if "cgo-disable" in request_flags:
         go_vars["CGO_ENABLED"] = "0"
 
     env = _go_exec_env(**go_vars)
     run_params = {"env": env, "cwd": app_dir}
 
-    # Explicitly disable toolchain telemetry for go >= 1.23
-    _disable_telemetry(go, run_params)
+    return run_params, temp_netrc_dir
```
taylormadore: Do we need to return temp_netrc_dir anymore given that it is supplied as a parameter by the caller?

hermeto/core/package_managers/gomod/main.py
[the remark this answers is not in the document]
a-ovchinnikov: My understanding is that permissive mode is not supposed to be used anywhere near an artifact registry proxy, won't permissive mode effectively undermine the strictness of it?

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Done.

tests/integration/proxy.py
a-ovchinnikov: You are right, the lack of `version` is enough.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Done (I do not have a strong preference here).

hermeto/core/package_managers/gomod/main.py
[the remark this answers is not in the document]
taylormadore: I think they're independent concerns, so I would lean towards raising UnsupportedFeature or something like that if someone tries to use them together. I'm mostly opposed to the current unhandled AttributeError when two supported features are combined. We should fail explicitly if the combination isn't supported.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: No, we don't, this is a leftover from a merge conflict resolution. (I also somehow cannot see this comment in the code view.)

hermeto/core/package_managers/gomod/main.py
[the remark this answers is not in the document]
a-ovchinnikov: That's fair, I'll update this on the next push (the one where I'll be dealing with merge conflicts).

hermeto/core/package_managers/gomod/main.py
```diff
@@ -768,6 +774,21 @@ def _resolve_gomod(
...
+            raise InvalidInput("Multiple authentication requiring proxy URLs detected!")
+        path_to_netrc = inject_netrc(prepare_netrc_contents(), Path(temp_netrc_dir.name))
+        go_vars["NETRC"] = str(path_to_netrc)
+        if os.environ.get("NETRC", ""):
+            log.warning(
+                "Proxy login was specified, at the same time a custom .netrc file path was"
+                " provided via environment. A new .netrc will be generated from proxy"
+                " parameters and will shadow the environment one. Please disable proxy"
+                " support if you intend to use the .netrc from environment."
+            )
```
slimreaper35, line 776: Does this somehow affect the prefetch task - https://github.com/konflux-ci/build-definitions/blob/main/task/prefetch-dependencies/0.3/prefetch-dependencies.yaml#L114 ?

a-ovchinnikov: Yes, if proxy credentials are defined they override any  external settings, the point of it is to ensure the source of dependencies.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -204,13 +208,30 @@ def purl(self) -> str:
...
         return purl.to_string()
 
     def to_component(self) -> Component:
         """Create a SBOM component for this package."""
-        return Component(name=self.name, version=self.module.version, purl=self.purl)
+        return Component(
+            name=self.name,
+            version=self.module.version,
+            purl=self.purl,
+        )
```
eskultety, line 227: nitpick: Do we need this hunk? The current call still fits on a single line.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -204,13 +208,30 @@ def purl(self) -> str:
...
+    """Determine whether a vcs_url qualifier is necessary and provide one."""
+    if (module.main or module.local_replace) and (module.repo_id is not None):
+        # The main module is always consumed from a local filesystem and it is always
+        # a clone of a repository, thus adding vcs_url to it.
+        # Local replaces need a vcs_url as well and are supposed to be consumed from the file
+        # system too.
+        # Ignore because at this point repo_id cannot be None, but that is not something mypy can
+        # infer.
+        return {"vcs_url": module.repo_id.as_vcs_url_qualifier()}  # type: ignore
+    return dict()
```
eskultety, line 238: nitpick: is dict more idiomatic than `{}`?

hermeto/core/package_managers/gomod/main.py
```diff
@@ -204,13 +211,31 @@ def purl(self) -> str:
...
+def _vcs_url_if_needed(module: Module) -> dict:
+    """Determine whether a vcs_url qualifier is necessary and provide one."""
+    if (module.main or module.local_replace) and (module.repo_id is not None):
+        # The main module is always consumed from a local filesystem and it is always
+        # a clone of a repository, thus adding vcs_url to it.
+        # Local replaces need a vcs_url as well and are supposed to be consumed from the file
+        # system too.
+        # Ignore because at this point repo_id cannot be None, but that is not something mypy can
+        # infer.
+        return {"vcs_url": module.repo_id.as_vcs_url_qualifier()}  # type: ignore
```
slimreaper35: If I remove the `type: ignore`, I don't see any error.

tests/unit/package_managers/gomod/test_main.py
```diff
@@ -1215,6 +1222,7 @@ def test_missing_gomod_file(
 @pytest.mark.parametrize(
     "gomod_input_packages, packages_output_by_path, expect_components",
     (
+        # First scenario
```
slimreaper35, line 1225: nitpick: Seems like an unrelated change. Also, the commit 707bce8e2a28399d59f57f8dfb6e665e05daf2bc says cosmetically refactor gomod resolver, but that happens in the next commit.

slimreaper35 (approved):

hermeto/core/package_managers/gomod/main.py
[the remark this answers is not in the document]
eskultety: Agreed with @taylormadore on this, unhandled exception is kinda unacceptable since we already know about "a codepath".

hermeto/core/package_managers/gomod/main.py
```diff
@@ -218,6 +221,7 @@ def to_component(self) -> Component:
             name=self.name,
             version=self.module.version,
             purl=self.purl,
+            external_references=self.module.to_component().external_references,
```
eskultety: This feels a bit wasteful - to construct an object just to pull a single property out of it, could we introduce a helper on module to query this information instead? I feel we're also coupling Package and Module component creation.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -789,30 +772,69 @@ def _resolve_gomod(
...
+    :return: a dict containing the Go module itself ("module" key), the list of dictionaries
+        representing the dependencies ("module_deps" key), the top package level dependency
+        ("pkg" key), and a list of dictionaries representing the package level dependencies
+        ("pkg_deps" key)
+    :raises PackageManagerError: if fetching dependencies fails
+    """
+    _protect_against_symlinks(app_dir)
+    should_vendor = app_dir.join_within_root("vendor").path.is_dir()
+
+    with TemporaryDirectory() as temp_netrc_dir:
```
eskultety, line 831: This context manager should IMO have been part of commit 4/5 instead.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -768,6 +774,21 @@ def _resolve_gomod(
     if config.gomod.proxy_url:
         go_vars["GOPROXY"] = config.gomod.proxy_url
 
+    temp_netrc_dir = TemporaryDirectory()
+    if config.gomod.proxy_login is not None:
+        # This is supposed to be happening only in systems utilizing artifact registry proxies.
+        if "," in config.gomod.proxy_url:
+            raise InvalidInput("Multiple authentication requiring proxy URLs detected!")
+        path_to_netrc = inject_netrc(prepare_netrc_contents(), Path(temp_netrc_dir.name))
+        go_vars["NETRC"] = str(path_to_netrc)
```
eskultety: nitpick: double `str` invocation, `inject_netrc` already stringifies its return value.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -768,6 +774,21 @@ def _resolve_gomod(
     if config.gomod.proxy_url:
         go_vars["GOPROXY"] = config.gomod.proxy_url
 
+    temp_netrc_dir = TemporaryDirectory()
+    if config.gomod.proxy_login is not None:
+        # This is supposed to be happening only in systems utilizing artifact registry proxies.
+        if "," in config.gomod.proxy_url:
+            raise InvalidInput("Multiple authentication requiring proxy URLs detected!")
```
eskultety: Okay, but I'm not sure whether the error message is very user friendly or actionable, would something like `proxy_url must contain exactly one URL when login is present` and maybe even provide context by concatenating `config.gomod.proxy_url`.

tests/integration/proxy.py
```diff
@@ -95,7 +98,15 @@ def _is_registry_component(component: Component) -> bool:
         p.name == PropertyEnum.PROP_CDX_NPM_PACKAGE_BUNDLED and p.value == "true"
         for p in component.properties
     )
-    if is_bundled:
+
+    is_local = (
```
eskultety, line 104: Why is this a single element tuple? I don't see it extended in the last commit.

hermeto/core/package_managers/gomod/main.py
eskultety, line 2: **commit mesage nitpick:**
`s/_resovle_gomod/_resolve_gomod`

hermeto/core/package_managers/gomod/main.py
```diff
@@ -728,40 +728,24 @@ def _parse_packages(
...
         # copied to the output dir, we need to set the GOMODCACHE to a different directory.
-        gomod_cache = f"{tmp_dir}/vendor-cache"
-    else:
-        gomod_cache = f"{tmp_dir}/pkg/mod"
+        return f"{tmp_dir}/vendor-cache"
+    return f"{tmp_dir}/pkg/mod"
+
+
+def _prepare_run_params(
+    request_flags: Iterable,
```
eskultety: nitpick: `Iterable[str] ` to be precise?

hermeto/core/package_managers/gomod/main.py
```diff
@@ -789,30 +772,69 @@ def _resolve_gomod(
...
+        representing the dependencies ("module_deps" key), the top package level dependency
+        ("pkg" key), and a list of dictionaries representing the package level dependencies
+        ("pkg_deps" key)
+    :raises PackageManagerError: if fetching dependencies fails
+    """
+    _protect_against_symlinks(app_dir)
+    should_vendor = app_dir.join_within_root("vendor").path.is_dir()
+
+    with TemporaryDirectory() as temp_netrc_dir:
+        run_params = _prepare_run_params(
```
eskultety, line 832: You're encoding netrc tempdir to run_params, a structure that lives beyond this `TemporaryDirectory` context, i.e. any go command that receives params will see netrc pointing to a non-existent location.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -4,11 +4,13 @@
 import re
 import shutil
 import tempfile
+import urllib
 from collections.abc import Iterable, Iterator
 from datetime import datetime, timezone
 from functools import cached_property
 from itertools import chain
 from pathlib import Path
+from tempfile import TemporaryDirectory
```
slimreaper35, line 13: Can we have just one import of the `tempfile` module?

hermeto/core/package_managers/gomod/main.py
slimreaper35: same here: `def _vcs_url_if_needed(module: Module) -> dict:` could be -> `dict[str, str]`

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: No, removed.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Indeed, thank you. This turns out to be a leftover from an earlier iteration. Removed.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: This is a personal preference: I noticed that I always pause for a fraction of a second each time I see `{}` in code wondering whether it is a dictionary creation or a set creation, I don't know why I cannot just learn this as an instinct. I understand where the decision to re-use curly braces comes from historically, but find it an unfortunate one, I saw people accidentally creating dictionaries when they intended to create a set and vice versa more than once, that is the other reason why I prefer to be explicit (even at the price of three more bytecode operations).

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Yes, we can, but that would require importing all names directly and that would be an unrelated refactoring which I don't think fits into the scope of this change. I generally tend to prefer importing names directly rather than accessing them through a dot because I don't think that having a module name prefix adds much clarity to the code (i.e. when there is no way for confusion), at the same time it eats up valuable visual space. I'd rather trade visual space in the imports section which I don't usually read as carefully as the code itself for some savings in the latter.

tests/integration/proxy.py
a-ovchinnikov: Oh, this is a tuple to populate with more conditions once more PMs become aware of proxies. I'll add a comment explaining this, sorry, it was very obvious in my head, to the point that I forgot to mention this explicitly.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Fixed.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Yes, this reads better, thank you, updated.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Well, a `Module` is a named tuple, it has already been created by now, a `Package` has been created too, this is a conversion method, however a `Component` is being created here sort of excessively. I guess we can add a helper here, it won't hurt, at least it will make the extraction a bit tidier.

tests/unit/package_managers/gomod/test_main.py
a-ovchinnikov: Thank you, fixed.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Fixed.

hermeto/core/package_managers/gomod/main.py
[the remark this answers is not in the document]
a-ovchinnikov: This was updated in the previous push, resolving.

hermeto/core/package_managers/gomod/main.py
eskultety: > at least it will make the extraction a bit tidier.

Yes, thank you.

tests/integration/proxy.py
eskultety: Can we add the tuple when it becomes actually relevant? It will contextually be a better fit IMO.

a-ovchinnikov: In this case I am pretty sure we will need it soon enough to warrant complicating it right away.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Could be, I am not a fan of `str` as a refinement since it does not provide a good enough refinement: `@slimreaper35`, `@eskultety` and `@a-ovchinnikov` are all valid `str`s after all and they have no place in neither arguments nor values of these functions. A better option is to have precise types defined for all these cases, but that would require a significant rework for the entire code base. I'll add `str`s for consistency though.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: I did it this way to avoid having a state where a long and winding function body sits in a context. It could have been introduced earlier, but I'd rather not do it unless you feel strongly about that.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Yes, to the best of my understanding this is not a problem as long as no additional fetches happen.

hermeto/core/package_managers/gomod/main.py
eskultety: > I did it this way to avoid having a state where a long and winding function body sits in a context. It could have been introduced earlier, but I'd rather not do it unless you feel strongly about that.

The principle we try to follow is that the codebase remains bisectable, I'd say there was a bit of luck that no unit test exercised the netrc path. When it comes to long function sitting in a context, that's a temporary state you're resolving within the same series, so I'd argue maintaining strong correctness and bisectability with every change is more important.

hermeto/core/package_managers/gomod/main.py
eskultety: > the best of my understanding this is not a problem as long as no additional fetches happen.

This remains an assumption, is there something we could do about not threading non-existent locations through function calls?

hermeto/core/package_managers/gomod/main.py
eskultety: ...to a point where one imports so many names that it's actually better to leverage the namespace IMO. This is not the case though -> resolving.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: I am sorry, I don't get the bisectability part -- an explicit exit from a context gets replaced with an automated one during refactoring, how does this affect our ability to bisect the codebase?

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Ok, that's fair, I'll move the rest of the code into the context.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: > one imports so many names

There is always a star import option for that case! Imports section is still just a sort of a crutch to help an interpreter locate necessary names and (thankfully) stays out of focus most of the time. Namespaced names would have made more sense if there was no need to explicitly include every namespace beforehand (since the interpreter already knows where to look for them anyway what's the point in telling it to go look for specific ones explicitly?), but unfortunately this is not the case for Python.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1418,3 +1494,26 @@ def _vendor_changed(context_dir: RootedPath) -> bool:
...
+    machine = urllib.parse.urlparse(config.gomod.proxy_url).netloc
+    return f"""machine {machine}
+        login {config.gomod.proxy_login}
+        password {config.gomod.proxy_password}"""
+
+
+def inject_netrc(netrc_stuff: str, temp_netrc_dir: Path) -> str:
+    """Inject a temporary .netrc."""
+    netrc = temp_netrc_dir / ".netrc"
+    netrc.write_text(netrc_stuff)
```
eskultety: AI finding here on the netrc `write`: this writes the netrc config using default umask (predominantly 644). I think we should default to the same practice as with ssh keys - 0600.

a-ovchinnikov: This is presumably a rather isolated environment, no? I'll add a chmod here anyway.

hermeto/core/package_managers/gomod/main.py
eskultety: My bad. I got tangled up in my own thoughts as I was clicking back and forth between the individual commits and this is the result, apologies for the confusion. It could be introduced earlier, but given the tests are happy, I won't block on this.

hermeto/core/package_managers/gomod/main.py
eskultety: Can we guarantee the deployment environment for Hermeto in general? Maybe you're considering a particular setup where this isolation is in place, but is that true in general? :)

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1418,3 +1494,27 @@ def _vendor_changed(context_dir: RootedPath) -> bool:
...
+    netrc.chmod(0o600)
```
eskultety: This is elegant, true, but it's not atomic. I think this instance warrants an `open` with mode and then write. With a bit of luck that open might even silence CodeQL on its own. :)

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: No, we can't do this in general, however I do believe that an environment leaky enough for this to be a problem will have a host of other, more serious issues.  At least this one won't be anymore (yet I would have expected this to be managed on system level by setting umask rules for the user which runs Hermeto).

eskultety (commented):
One last comment from me, LGTM.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: Good point, updated the code.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1418,3 +1494,30 @@
...
+        password {config.gomod.proxy_password}"""
+
+
+def inject_netrc(netrc_stuff: str, temp_netrc_dir: Path) -> str:
+    """Inject a temporary .netrc."""
+    netrc = temp_netrc_dir / ".netrc"
+    old_mask = os.umask(0)
+    netrc_fd = os.open(path=netrc, flags=(os.O_WRONLY | os.O_CREAT | os.O_TRUNC), mode=0o600)
+    with open(netrc_fd, "w") as f:
+        f.write(netrc_stuff)
```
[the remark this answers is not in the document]
a-ovchinnikov: @eskultety you see, this thing cannot be satisfied unless I rot13 the password. It now also complains about a descriptor not being released which, to the best of my knowledge, happens on exit from `open` context.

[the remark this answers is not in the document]
a-ovchinnikov: The other day one of the robots got worried that I am not importing `os` and placed a commen about not importing `os` and a suggested fix one line below an `import os` statement.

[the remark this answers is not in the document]
eskultety: Sigh, I'm already worried what these false positives will do once we hit a fully agentic age one day, like how many iterations until the agents are happy with their bikeshedding...dismissing.

eskultety (approved):

taylormadore (approved):
LGTM

PR #1576, opened 2026-05-25
feat(http): Add retry logs to sync and async downloads

### LINKED ISSUE #1552: Consider logging HTTP download retries
When Hermeto retries a failed HTTP download, no log output is produced. Both download paths retry on 500/502/503/504 status codes, but the retries are invisible in our logs even with `--log-level=DEBUG`. The underlying libraries do log retries, so it should be possible.

This could be improved from a UX perspective, since without retries being logged, a download that fails after multiple retry attempts looks like a single slow request that errored out. The ~10-second gap suggests retries occurred, but there's no attempt counts, no intermediate status codes, and no backoff durations show.

Harsh-2005d:
Fixes: #1552

## Summary
Both sync and async download paths produced no log output when retrying failed HTTP requests, making a download that fails after multiple retries look like a single slow request that errored out.

## Changes
**Sync path:** Introduces `_LoggingRetry`, a `urllib3.Retry` subclass that overrides `increment()` to log each retry at DEBUG level with retry count, total retries, URL, HTTP status, and backoff duration.

**Async path:** Introduces an `on_request_end` TraceConfig callback that logs each retry at DEBUG level with retry count, total retries, filename, and HTTP status. Backoff duration is unavailable at the TraceConfig callback point as it is computed by `aiohttp_retry` after the callback fires.

The async retry logging is inspired by [this](https://github.com/hermetoproject/hermeto/commit/c7f3de8cf32faacf3c4f4f142b56bf3f29fa53d1) earlier implementation which was [removed ](https://github.com/hermetoproject/hermeto/commit/3215e8de029f04b91e9bc1cda512e0ae43f778a7) due to no explanation about the code. This implementation switches from `on_request_start` to `on_request_end` to make the HTTP status code available in the log output.

## Example output
**Sync**
```
DEBUG hermeto.core.http_requests Retrying request: retry=1/5 url=/deps/pip/package-1.0.tar.gz status=503 backoff=1.3s
```
**Async**
```
DEBUG hermeto.core.package_managers.general Retrying request: retry=1/5 file=lodash-4.17.21.tgz status=503
```

### DIFF
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -2,8 +2,10 @@
 import asyncio
 import logging
 import ssl
+import types
 from collections.abc import Mapping
-from typing import Any
+from types import TracebackType
+from typing import Any, cast
 from urllib.parse import urlparse
 
 import aiohttp
@@ -12,6 +14,9 @@ import requests
 from requests import Session
 from requests.adapters import HTTPAdapter
 from requests.auth import AuthBase
+from typing_extensions import Self
+from urllib3.connectionpool import ConnectionPool
+from urllib3.response import BaseHTTPResponse
 from urllib3.util.retry import Retry
 
 from hermeto.core.config import get_config
@@ -26,6 +31,50 @@ BACKOFF_FACTOR = 1.3
 STATUS_FORCELIST = (500, 502, 503, 504)
 
 
+log = logging.getLogger(__name__)
+
+
+class SyncLoggingRetry(Retry):
+    """
+    Retry subclass that emits a hermeto-style debug log on each retry.
+    """
+
+    def increment(
+        self,
+        method: str | None = None,
+        url: str | None = None,
+        response: BaseHTTPResponse | None = None,
+        error: Exception | None = None,
+        _pool: ConnectionPool | None = None,
+        _stacktrace: TracebackType | None = None,
+    ) -> Self:
+        """
+        Log retry attempt at DEBUG level.
+        """
+        # retry object - Named to match urllib3.Retry internal logic.
+        new_retry = super().increment(
+            method=method,
+            url=url,
+            response=response,
+            error=error,
+            _pool=_pool,
+            _stacktrace=_stacktrace,
+        )
+        retry_number = len(new_retry.history)
+        status = response.status if response else "N/A"
+        backoff = new_retry.get_backoff_time()
+        total = retry_number + cast(int, self.total) - 1
+        log.debug(
+            "Retrying request: retry=%d/%d url=%s status=%s backoff=%.1fs",
+            retry_number,
+            total,
+            url,
+            status,
+            backoff,
+        )
+        return new_retry
+
+
 def _get_pkg_requests_session() -> requests.Session:
     """
     A lazy initialised, module-level requests.Session with retry config.
@@ -35,7 +84,7 @@ def _get_pkg_requests_session() -> requests.Session:
         max_retries = get_config().http.max_retries
         _pkg_requests_session = Session()
         adapter = HTTPAdapter(
-            max_retries=Retry(
+            max_retries=SyncLoggingRetry(
                 backoff_factor=BACKOFF_FACTOR,
                 status_forcelist=STATUS_FORCELIST,
                 allowed_methods=SAFE_REQUEST_METHODS,
@@ -46,9 +95,6 @@ def _get_pkg_requests_session() -> requests.Session:
         _pkg_requests_session.mount("https://", adapter)
 
     return _pkg_requests_session
-
-
-log = logging.getLogger(__name__)
 
 
 def download_binary_file(
@@ -142,6 +188,53 @@ async def _async_download_binary_file(
     log.debug(f"Download completed - {url}")
 
 
+def _aiohttp_create_retry_trace_config(
+    retry_options: aiohttp_retry.JitterRetry,
+) -> aiohttp.TraceConfig:
+    """
+    Creates a TraceConfig that logs retry attempts using the provided retry options.
+    """
+
+    trace_config = aiohttp.TraceConfig()
+
+    async def on_request_end(
+        _session: aiohttp.ClientSession,
+        trace_config_ctx: types.SimpleNamespace,
+        params: aiohttp.TraceRequestEndParams,
+    ) -> None:
+        attempt = trace_config_ctx.trace_request_ctx["current_attempt"]
+        if attempt > 1 and params.response.status in retry_options.statuses:
+            file_name = params.url.name
+            log.debug(
+                "Retrying request: retry=%d/%d file=%s status=%d",
+                attempt - 1,
+                retry_options.attempts - 1,
+                file_name,
+                params.response.status,
+            )
+
+    async def on_request_exception(
+        _session: aiohttp.ClientSession,
+        trace_config_ctx: types.SimpleNamespace,
+        params: aiohttp.TraceRequestExceptionParams,
+    ) -> None:
+        attempt = trace_config_ctx.trace_request_ctx["current_attempt"]
+        exception_valid = any(isinstance(params.exception, exc) for exc in retry_options.exceptions)
+        if attempt > 1 and exception_valid:
+            file_name = params.url.name
+            log.debug(
+                "Retrying request: retry=%d/%d file=%s exception=%s",
+                attempt - 1,
+                retry_options.attempts - 1,
+                file_name,
+                type(params.exception).__name__,
+            )
+
+    trace_config.on_request_exception.append(on_request_exception)
+    trace_config.on_request_end.append(on_request_end)
+    return trace_config
+
+
 async def async_download_files(
     files_to_download: Mapping[str, StrPath],
     concurrency_limit: int,
@@ -155,7 +248,7 @@ async def async_download_files(
     :param ssl_context: Optional SSL context for the requests.
     :param headers: Optional per-URL headers mapping (URL -> headers dict).
     """
-    trace_config = aiohttp.TraceConfig()
+
     max_retries = get_config().http.max_retries
     # aiohttp uses n calls (1 call, n-1 retries).
     max_retries = max_retries + 1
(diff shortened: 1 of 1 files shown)


hermeto/core/http_requests.py
```diff
@@ -15,3 +22,35 @@
     "backoff_factor": 1.3,
     "status_forcelist": (500, 502, 503, 504),
 }
+
+
+class syncLog(Retry):
```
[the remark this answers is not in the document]
a-ovchinnikov: I am sorry, the name `syncLog` makes no sense here: it hides the meaning of the operation (which still means a retry, just with some decoration). Please update the name accordingly.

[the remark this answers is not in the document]
Harsh-2005d: will `SyncLoggingRetry` work or something else?

Harsh-2005d:
@eskultety @a-ovchinnikov am i missing something on the PR ?

eskultety:
> @eskultety @a-ovchinnikov am i missing something on the PR ?

@Harsh-2005d apologies, we haven't had the time to get back to this PR just yet, please stay tuned.

Harsh-2005d:
@eskultety No worries.

hermeto/core/http_requests.py
```diff
@@ -15,3 +20,42 @@
...
+        url: str | None = None,
+        response: Any | None = None,
+        error: Exception | None = None,
+        _pool: ConnectionPool | None = None,
+        _stacktrace: TracebackType | None = None,
+    ) -> Self:
+        """
+        Log retry attempt at DEBUG level and delegate to the parent implementation.
+        """
+        new_retry = super().increment(
```
a-ovchinnikov: Naming nitpick: this is a new attempt, not a new retry, I suggest naming it as such.

hermeto/core/http_requests.py
```diff
@@ -15,3 +20,42 @@
...
+            method=method,
+            url=url,
+            response=response,
+            error=error,
+            _pool=_pool,
+            _stacktrace=_stacktrace,
+        )
+        retry = len(new_retry.history)
```
a-ovchinnikov: Another naming nitpick: this is a retry number, not just a retry.

hermeto/core/package_managers/general.py
```diff
@@ -165,6 +169,43 @@ async def async_download_files(
             aiohttp.ClientPayloadError,
         },
     )
+    trace_config = aiohttp.TraceConfig()
+
+    async def on_request_end(
```
a-ovchinnikov, line 203: Why do you define these two helpers in the function scope?

hermeto/core/http_requests.py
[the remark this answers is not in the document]
a-ovchinnikov: It is a mouthful, but will do.

hermeto/core/package_managers/general.py
Harsh-2005d: Just a [earlier work](https://github.com/hermetoproject/hermeto/commit/c7f3de8cf32faacf3c4f4f142b56bf3f29fa53d1) i restored. Should i move them out in the same file or in `http_request.py`

a-ovchinnikov: Yes, but is there any reason for the methods to be defined in this particular scope? The function is pretty big as it is now, I would have rather kept it compact.

hermeto/core/http_requests.py
Harsh-2005d: I kept it as `new_retry` here to maintain consistency with urllib3's internal terminology (retries vs. attempts). Standardized on using 'retries' for this exact reason in an earlier PR [discussion](https://github.com/hermetoproject/hermeto/pull/1567#discussion_r3272586498).

a-ovchinnikov: Oh, I see, could you please add a short comment telling the reader that this is to keep the names aligned with urllib3's retry implementation?

hermeto/core/package_managers/general.py
Harsh-2005d: > Why do you define these two helpers in the function scope?

aiohttp strictly enforces the trace callback signatures, nesting them allowed the helpers to access the local `retry_options` variable without trying to pass it as an unsupported parameter.



>The function is pretty big as it is now, I would have rather kept it compact.

the best we can do is to extract that setup into a separate helper function.

a-ovchinnikov: > to access the local retry_options variable

You could have returned the function from another function accepting this variable as an argument, however moving away the entire setup sounds like a better option.

a-ovchinnikov (dismissed):
LGTM with one minor request: could you please amend your commit and add "Resolves: `https://github.com/hermetoproject/hermeto/issues/1552`" line to the commit message right above "Signed-off-by:" line?

hermeto/core/package_managers/general.py
```diff
@@ -139,6 +143,50 @@ async def _async_download_binary_file(
...
+        _session: aiohttp.ClientSession,
+        trace_config_ctx: types.SimpleNamespace,
+        params: aiohttp.TraceRequestEndParams,
+    ) -> None:
+        attempt = trace_config_ctx.trace_request_ctx["current_attempt"]
+        if attempt > 1 and params.response.status in retry_options.statuses:
+            file_name = params.url.name
+            log.debug(
+                "Retrying request: retry=%d/%d file=%s status=%d",
+                attempt,
```
eskultety: Since this would run on the first retry, `current_attempt == 2` which means we'd log `Retrying request: retry=2/N` which is a bit misleading.

hermeto/core/package_managers/general.py
```diff
@@ -139,6 +143,50 @@ async def _async_download_binary_file(
...
+                file_name,
+                params.response.status,
+            )
+
+    async def on_request_exception(
+        _session: aiohttp.ClientSession,
+        trace_config_ctx: types.SimpleNamespace,
+        params: aiohttp.TraceRequestExceptionParams,
+    ) -> None:
+        attempt = trace_config_ctx.trace_request_ctx["current_attempt"]
```
eskultety, line 224: We might want to retry only on the defined set of exceptions `retry_options.exceptions` that we set in ` async_download_files`

hermeto/core/package_managers/general.py
```diff
@@ -80,6 +82,8 @@ def download_binary_file(
         for chunk in resp.iter_content(chunk_size=chunk_size):
             f.write(chunk)
 
+    log.debug("Download completed - %s", url)
+
```
eskultety, line 133: These seem like unrelated changes to the retry logic -> standalone commit.

hermeto/core/http_requests.py
```diff
@@ -15,3 +20,58 @@
...
+        url: str | None = None,
+        response: Any | None = None,
+        error: Exception | None = None,
+        _pool: ConnectionPool | None = None,
+        _stacktrace: TracebackType | None = None,
+    ) -> Self:
+        """
+        Log retry attempt at DEBUG level.
+        """
+        # retry object - follows urllib3's internal naming convention for the next network call.
```
eskultety: nitpick: _Named to match urllib3.Retry internal logic_ might read a bit better.

hermeto/core/http_requests.py
```diff
@@ -15,3 +20,58 @@
...
+
+class SyncLoggingRetry(Retry):
+    """
+    Retry subclass that emits a hermeto-style debug log on each retry.
+    """
+
+    def __init__(self, *args: Any, **kw: Any) -> None:
+        """
+        Initialize and store the user configured total retries.
+        """
```
eskultety: Probably redundant.

eskultety (commented):
Minor comments, one more respin will do.

hermeto/core/http_requests.py
Harsh-2005d: do u mean the doc string or the `__init__` ?

hermeto/core/package_managers/general.py
Harsh-2005d: Fixed.

hermeto/core/package_managers/general.py
Harsh-2005d: added a `exception_valid` boolean.

a-ovchinnikov (approved):

hermeto/core/http_requests.py
```diff
@@ -15,3 +20,58 @@
...
+        new_instance = super().new(**kw)
+        new_instance.total_retries = self.total_retries
+        return new_instance
+
+    def increment(
+        self,
+        method: str | None = None,
+        url: str | None = None,
+        response: Any | None = None,
```
ST3V1K: Nitpick: `response` can be of type `BaseHTTPResponse`

hermeto/core/http_requests.py
```diff
@@ -15,3 +20,58 @@
...
+        self.total_retries = kw.get("total")
+        super().__init__(*args, **kw)
+
+    def new(self, **kw: Any) -> Self:
+        """
+        Propagate the original total across retry instances.
+        """
+        new_instance = super().new(**kw)
+        new_instance.total_retries = self.total_retries
+        return new_instance
```
ST3V1K: Non-blocking: Could simplify by computing the original total from `self.total + retry_number - 1` rather than storing and copying it in `new()`.

ST3V1K (commented):
Looks good, just a few small suggestions inline.

hermeto/core/http_requests.py
Harsh-2005d: lol, Thanks.

hermeto/core/package_managers/general.py
```diff
@@ -168,6 +259,8 @@ async def async_download_files(
             aiohttp.ClientPayloadError,
         },
     )
+    trace_config = _create_retry_trace_config(retry_options)
```
eskultety: nitpick: maybe call the function directly in the parameter assignment below, should fit the line :) .

hermeto/core/package_managers/general.py
```diff
@@ -142,6 +188,51 @@ async def _async_download_binary_file(
     log.debug(f"Download completed - {url}")
 
 
+def _create_retry_trace_config(retry_options: aiohttp_retry.JitterRetry) -> aiohttp.TraceConfig:
```
eskultety: naming nitpick: maybe `_aiohttp_create_retry_trace_config` to be more explicit? Non-blocking, but since we're tying this to a particular retry logic, might be worth emphasizing the difference.

eskultety (approved):

PR #1581, opened 2026-05-27
fix(aiohttp): Preserve percent-encoding in redirect URL

### LINKED ISSUE #1578: 403 on CloudFront signed URL for pip artifacts (packages.redhat.com)
## Summary

`hermeto fetch-deps` fails with `403 Forbidden` while downloading a Python sdist from a Red Hat Trusted Libraries artifact URL that redirects to a signed CloudFront URL.

The same artifact downloads successfully with `curl` and `requests` using the exact signed URL.

### 1) Hermeto fails

```bash
podman run --rm -t \
  -v "$PWD:$PWD:z" \
  -v "/home/julen/.netrc:/tmp/user.netrc:ro,z" \
  -e NETRC="/tmp/user.netrc" \
  -w "$PWD" \
  ghcr.io/hermetoproject/hermeto:latest \
  --log-level DEBUG fetch-deps --source . --output .tmp/hermeto-repro/debug-output \
  '{"type":"pip","path":".","requirements_files":["requirements-redhat-trusted-libraries.txt"]}'
```

Relevant log lines:

```text
INFO -- Processing requirement line 'aiofiles==25.1.0'
DEBUG aiohttp.ClientSession.get(url: https://packages.redhat.com/api/pulp-content/public-trusted-libraries/main/aiofiles-25.1.0.tar.gz, ...)
ERROR Unsuccessful download: https://packages.redhat.com/api/pulp-content/public-trusted-libraries/main/aiofiles-25.1.0.tar.gz
ERROR FetchError: ... details: 403 ... url='https://dkvt8etmt0k4u.cloudfront.net/artifact/...?...&ResponseContentType=application/gzip&...'
```

### 2) `curl` with netrc succeeds

```bash
curl -sS -L -f --netrc-file "/home/julen/.netrc" \
  -o /tmp/aiofiles-25.1.0-auth.tar.gz \
  "https://packages.redhat.com/api/pulp-content/public-trusted-libraries/main/aiofiles-25.1.0.tar.gz"
```

This succeeds and the tarball is valid (`tar -tzf /tmp/aiofiles-25.1.0-auth.tar.gz` lists expected files).

## Pinpointed behavior

From the redirect response, the signed CloudFront URL contains:

`ResponseContentType=application%2Fgzip`

In the failing Hermeto/aiohttp path, the final URL shown in logs becomes:

`ResponseContentType=application/gzip`

This encoding change appears to invalidate the CloudFront signature.

## Minimal proof outside Hermeto

The behavior is reproducible with `aiohttp` alone:

- `aiohttp` GET with plain string URL -> `403`
- `aiohttp` GET with `yarl.URL(signed_url, encoded=True)` -> `200`
- `requests.get(signed_url)` -> `200`
- mutating `%2F` to `/` in query parameter -> `403`

Example snippet:

```python
import aiohttp
from yarl import URL

signed_url = "https://dkvt8etmt0k4u.cloudfront.net/artifact/...&ResponseContentType=application%2Fgzip&..."

async with aiohttp.ClientSession(trust_env=True) as s:
    async with s.get(signed_url) as r:
        print("plain:", r.status)  # 403

async with aiohttp.ClientSession(trust_env=True) as s:
    async with s.get(URL(signed_url, encoded=True)) as r:
        print("encoded:", r.status)  # 200
```

## Hypothesis

During download, Hermeto (or the underlying aiohttp request path) is re-quoting/normalizing the redirect URL query string, decoding `%2F` in a signed query parameter. CloudFront then rejects the modified URL with `AccessDenied`.

## Suggested fix direction

- Preserve redirect URL encoding exactly for signed artifact URLs.
- If constructing `yarl.URL` objects, use `URL(raw_url, encoded=True)` for already signed URLs.
- Avoid transformations that alter percent-encoded query parameter values before the request is sent.

## Expected behavior

Given a valid authenticated redirect from `packages.redhat.com` to a CloudFront signed URL, Hermeto should download the artifact successfully, matching `curl -L --netrc-file ...` behavior.

Assisted-By: Codex 5.3

eskultety:
See the commit for more information.
The unit test was entirely written and proposed by AI, but the test utilities by aiohttp seem solid enough.

Closes https://github.com/hermetoproject/hermeto/issues/1578

### DIFF
--- a/tests/unit/package_managers/test_general.py
+++ b/tests/unit/package_managers/test_general.py
@@ -268,3 +268,30 @@ async def test_async_download_files(
     for call in mock_download_file.mock_calls:
         _, file, path = call.args
         assert file, path in files_to_download.items()
+
+
+@pytest.mark.asyncio
+async def test_async_download_preserves_redirect_url_encoding(tmp_path: Path) -> None:
+    """Redirect URLs with percent-encoded query params must not be re-quoted."""
+    from aiohttp import web
+    from aiohttp.test_utils import TestServer
+
+    async def redirect_handler(request: web.Request) -> web.Response:
+        return web.Response(
+            status=302,
+            headers={"Location": "/target?ResponseContentType=text%2Fplain"},
+        )
+
+    # we purposefully return the raw redirect URL in the body to verify what aiohttp did with it
+    async def target_handler(request: web.Request) -> web.Response:
+        return web.Response(status=200, body=request.raw_path.encode())
+
+    app = web.Application()
+    app.router.add_get("/redirect", redirect_handler)
+    app.router.add_get("/target", target_handler)
+
+    async with TestServer(app) as server:
+        url = str(server.make_url("/redirect"))
+        download_path = tmp_path / "artifact"
+        await async_download_files({url: str(download_path)}, concurrency_limit=1)
+        assert b"text%2Fplain" in download_path.read_bytes()
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -170,6 +170,8 @@ async def async_download_files(
         trace_configs=[trace_config],
         # respect proxy settings and .netrc
         trust_env=True,
+        # preserve percent-encoding in redirect URLs (e.g. signed CloudFront URLs)
+        requote_redirect_url=False,
     )
 
     async with retry_client as session:


tests/unit/package_managers/test_general.py
```diff
@@ -268,3 +268,30 @@ async def mock_download_binary_file(
...
+
+    app = web.Application()
+    app.router.add_get("/redirect", redirect_handler)
+    app.router.add_get("/target", target_handler)
+
+    async with TestServer(app) as server:
+        url = str(server.make_url("/redirect"))
+        download_path = tmp_path / "artifact"
+        await async_download_files({url: str(download_path)}, concurrency_limit=1)
+        assert b"text%2Fplain" in download_path.read_bytes()
```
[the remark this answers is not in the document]
eskultety: Yeah, how about 'No' in this case?!

tests/unit/package_managers/test_general.py
```diff
@@ -268,3 +268,30 @@ async def mock_download_binary_file(
     for call in mock_download_file.mock_calls:
         _, file, path = call.args
         assert file, path in files_to_download.items()
+
+
+@pytest.mark.asyncio
+async def test_async_download_preserves_redirect_url_encoding(tmp_path: Path) -> None:
+    """Redirect URLs with percent-encoded query params must not be re-quoted."""
+    from aiohttp import web
+    from aiohttp.test_utils import TestServer
```
a-ovchinnikov, line 277: Do you really need the imports to happen in a functions?

a-ovchinnikov (approved):

slimreaper35 (approved):

tests/unit/package_managers/test_general.py
eskultety: I don't, but this is a really scoped test, I don't expect to need these utils for anything else in unit tests.

eskultety: I can change if you feel strongly about this.

a-ovchinnikov: No, I don't feel too strong about this, I have already approved the change, however the approach that has been prevalent so far was to keep all imports at module level. My only concern is about setting a precedent

eskultety: > My only concern is about setting a precedent

Is that a problem for heavily scoped tests? I think we can always judge on a case by case basis. I would personally not be inclined towards importing from stdlib or our own modules on function level, that would smell IMO.

a-ovchinnikov: Fair enough, let's keep it as is.

PR #1585, opened 2026-05-28
Add Registry Proxy Support for Pip Backend

taylormadore:
This PR adds support to the pip backend for fetching packages via a registry proxy and reporting them


hermeto/core/package_managers/pip/main.py
```diff
@@ -381,19 +386,29 @@ def _resolve_and_download_pypi_packages(
     """Resolve and download all PyPI packages."""
     config = get_config()
     fetch_url = index_url
+    pip_requests_auth = None
+    pip_aiohttp_auth = None
     if config.pip.proxy_url is not None:
         fetch_url = str(config.pip.proxy_url)
+        if config.pip.proxy_login and config.pip.proxy_password:
```
a-ovchinnikov: Nitpick: it should be enough to check for just the login, partially set 
credentials get rejected (https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/config.py#L109)

_Edit_: this was a comment on https://github.com/hermetoproject/hermeto/commit/6c97878ec5962a726c161ce485d7ee0d471033b5 (GitHub helpfully carries the same comment across commits)

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -119,14 +119,25 @@ def _get_project_packages_from(
...
-    with pypi_simple.PyPISimple(index_url) as client:
+    with pypi_simple.PyPISimple(index_url, auth=auth) as client:
         try:
             project_page = client.get_project_page(name, timeout)
         except (requests.RequestException, pypi_simple.NoSuchProjectError) as e:
+            if (
+                config.pip.proxy_url is not None
+                and isinstance(e, requests.HTTPError)
+                and e.response is not None
+                and e.response.status_code in (401, 403)
```
a-ovchinnikov: In Sonatype Nexus 403 could mean a quarantined artifact, would not one see it here in that case as well?

a-ovchinnikov (commented):
A couple of small nitpicks, LGTM otherwise.

hermeto/core/package_managers/pip/packages.py
```diff
@@ -0,0 +1,118 @@
...
+
+@dataclass
+class VCSPackage(PipPackage):
+    """A package fetched from a VCS repository (git)."""
+
+    url: str
+    ref: str
+    host: str
+    namespace: str
+    repo: str
```
eskultety: These 3 don't seem to be used anywhere (not even in the later patches), we're only defining them because we get them from `extract_git_info`. Are they supposed to be used in some follow-up work that hasn't been proposed yet? Do we need to add them now?

hermeto/core/package_managers/pip/packages.py
```diff
@@ -0,0 +1,118 @@
...
+from hermeto.core.models.property_semantics import PropertySet
+from hermeto.core.models.sbom import Component
+from hermeto.core.rooted_path import RootedPath
+
+
+@dataclass
+class PipPackage(ABC):
+    """Base class for a fetched pip package."""
+
+    package: str
```
eskultety: nitpick: Could we use the opportunity of adding new classes to rename this to `name` and be more consistent with `PipPackageInfo` that declares a `name`?

eskultety (commented):
Since this is still a draft, I suggest proposing the refactoring patches in a standalone PR. I had a couple of nitpicks but patches 1-4 are very clean and ready to go in immediately.
I haven't looked at the proxy stuff just yet.

hermeto/core/package_managers/pip/package_distributions.py
taylormadore: Yeah, this is the part that isn't final. Waiting on a test-case for the blocked package scenario

hermeto/core/package_managers/pip/packages.py
taylormadore: This makes sense and I will make that change :+1:

hermeto/core/package_managers/pip/packages.py
taylormadore: Good point, I agree

hermeto/core/package_managers/pip/package_distributions.py
taylormadore: Dropped the 403

hermeto/core/package_managers/pip/main.py
taylormadore: It was to appease mypy. Either this or an ignore :man_shrugging:

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -119,14 +119,25 @@ def _get_project_packages_from(
...
+    with pypi_simple.PyPISimple(index_url, auth=auth) as client:
         try:
             project_page = client.get_project_page(name, timeout)
         except (requests.RequestException, pypi_simple.NoSuchProjectError) as e:
+            if (
+                config.pip.proxy_url is not None
+                and isinstance(e, requests.HTTPError)
+                and e.response is not None
+                and e.response.status_code == 401
+            ):
```
a-ovchinnikov, line 136: WDYT about adding a comment stating that some artifact registry proxies effectively respond with 404 when some of their subsystems block a package from an index? It might help someone debugging a seemingly missing package some day in the future.

a-ovchinnikov (approved):

hermeto/core/package_managers/pip/package_distributions.py
taylormadore: We discussed this a bit offline, but unfortunately there is no error response from the server when a block occurs--the package+version simply appears as though it doesn't exist in the index, which is indistinguishable from when it truly doesn't exist. :cry:

taylormadore: I can add a comment somewhere explaining this, but I think it would have to be on the "No distributions found for package" errors, which is what will be raised if this happens. WDYT?

a-ovchinnikov: Yes, that would work too, just anything to leave a hint for the future maintainers will do -- someone is bound to run into it too at some point.

slimreaper35 (approved):

hermeto/core/package_managers/pip/package_distributions.py
```diff
@@ -120,14 +119,25 @@ def _get_project_packages_from(
     index_url: str,
     name: str,
     version: str,
+    auth: requests.auth.HTTPBasicAuth | None = None,
 ) -> Iterable[pypi_simple.DistributionPackage]:
     """Get all the project packages from the given index URL."""
     config = get_config()
     timeout = (config.http.connect_timeout, config.http.read_timeout)
-    with pypi_simple.PyPISimple(index_url) as client:
+    with pypi_simple.PyPISimple(index_url, auth=auth) as client:
```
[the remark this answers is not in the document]
taylormadore: False, it does

hermeto/core/package_managers/pip/package_distributions.py
taylormadore: Added comments to both of the spots

hermeto/core/package_managers/pip/packages.py
```diff
@@ -51,6 +51,16 @@ class PyPIPackage(PipPackage):
...
     index_url: str
+    proxy_url: str | None = None
+
+    def to_component(self, build_dependency: bool) -> Component:
+        """Build an SBOM Component, attaching proxy ExternalReference if applicable."""
+        component = super().to_component(build_dependency)
+        if self.proxy_url is not None:
+            component.external_references = [
+                ExternalReference(url=self.proxy_url, type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
+            ]
```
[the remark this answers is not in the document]
taylormadore: Fine, this is reasonable. I'll do it differently though

PR #1586, opened 2026-05-28
docs: Added a Yarn Berry package manager design documentation

mohamedbouchtout:
### Summary

This PR introduces **docs/design/yarnberry.md** covering Yarn Berry's architecture, developer workflow, dependency resolution, lockfile format, plugin system, and how Hermeto processes Yarn Berry projects. The document follows the standard design doc template and has been updated to match the current codebase, removing references to features that are no longer supported and adding documentation for features introduced since the original doc was written.

###   Details

- Add a comprehensive design document for the Yarn Berry package manager, converted from a existing Google Doc into the standard design doc template format.
- Remove the outdated **Scope** section (all Yarn versions are now supported together).
- Update **Zero Installs** section to reflect it is no longer supported.
- Update **Git Dependencies** section to reflect they raise `UnsupportedFeature` error.
- Remove all legacy Cachito references and replace with Hermeto equivalents for examples.
- Add Yarn v4 workspace support documentation.
- Add proxy support documentation.
- Document permissive mode for non-git sources (NotAGitRepo handling).
- Document NODE_USE_SYSTEM_CA environment variable for TLS verification.


eskultety (commented):
I am yet to do a deep-dive review on this, but for the purposes of reviewing an old doc (that partially predates my involvement with the project) seeing this PR split into distinct chunks starting with the original and then gradually updating it until it actually reflects our logic is a nice take that I as a reviewer appreciate in this particular case! However, before we merge this all commits should be squashed to a single one because we're adding **a single (presumably) complete document**.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,701 @@
...
+##### Override for Prefetch
+
+Options useful for implementing Hermeto's functionality or required for security:
+
+| Option | Value | Purpose |
+|--------|-------|---------|
+| [`checksumBehavior`][checksumBehavior] | `"throw"` | Strict checksum validation |
+| [`enableImmutableInstalls`](https://yarnpkg.com/configuration/yarnrc#enableImmutableInstalls) | `true` | Fail if yarn.lock needs an update |
+| [`globalFolder`][globalFolder] | `<output_dir>` | Where to download dependencies |
+| [`enableMirror`][enableMirror] | `true` | false would break globalFolder |
```
eskultety: `enableMirror` override description doesn't seem to reflect the reality:
https://github.com/hermetoproject/hermeto/blob/20ff8bb0d09b249e14e720865b118ddbc1625867/hermeto/core/package_managers/yarn/main.py#L245

eskultety: Additionally, since we have a detailed table like ^this one, then the following options are not covered, but we do set them for prefetch:

- https://github.com/hermetoproject/hermeto/blob/20ff8bb0d09b249e14e720865b118ddbc1625867/hermeto/core/package_managers/yarn/main.py#L240
- https://github.com/hermetoproject/hermeto/blob/20ff8bb0d09b249e14e720865b118ddbc1625867/hermeto/core/package_managers/yarn/main.py#L247

- https://github.com/hermetoproject/hermeto/blob/20ff8bb0d09b249e14e720865b118ddbc1625867/hermeto/core/package_managers/yarn/main.py#L262 (This is for Yarn V4 only!)

docs/design/yarnberry.md
```diff
@@ -0,0 +1,701 @@
...
+- [`enableGlobalCache`][enableGlobalCache] (could solve
+  cacheFolder for non-zero installs)
+- [`installStatePath`](https://yarnpkg.com/configuration/yarnrc#installStatePath)
+- [`pnpDataPath`](https://yarnpkg.com/configuration/yarnrc#pnpDataPath) when
+  [`pnpEnableInlining`](https://yarnpkg.com/configuration/yarnrc#pnpEnableInlining) is false
+- [`pnpEnableEsmLoader`](https://yarnpkg.com/configuration/yarnrc#pnpEnableEsmLoader)
+- [`pnpUnpluggedFolder`](https://yarnpkg.com/configuration/yarnrc#pnpUnpluggedFolder)
+- [`virtualFolder`](https://yarnpkg.com/configuration/yarnrc#virtualFolder)
+
+#### Environment Variables
```
eskultety, line 555: Since we produce build-time environment variables we should be extra clear here that these are variables set during prefetch, not during build.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,701 @@
...
+```json
+{
+  "type": "unofficial",
+  "diff": {
+    "text": {
+      "encoding": "base64",
+      "content": "<base64 encoded content of the patch file>"
+    }
+  }
+}
```
eskultety: This doesn't seem to be accurate anymore, IIUC we do it using URL references passed to `PatchDiff`:
https://github.com/hermetoproject/hermeto/blob/20ff8bb0d09b249e14e720865b118ddbc1625867/hermeto/core/package_managers/yarn/resolver.py#L296

docs/design/yarnberry.md
```diff
@@ -0,0 +1,701 @@
...
+**Option 2: Allowlist of plugins to ignore** — Maintain an allowlist of plugins by
+[spec](https://github.com/chmeliik/berryscary/blob/c424d96e1e36542e52985aee716e1b12881c24fb/.yarnrc.yml#L3).
+Verify the user-configured set of plugins is a subset of the allowlist. Then ignore all plugins as
+in Option 1.
+
+- Pros: Definitely safe; probably does not affect the prefetched content
+- Cons: Requires maintaining an allowlist; plugins are unsupported by default until added to the
+  allowlist
+
+**Option 3: Allowlist of plugins to execute** — Maintain an allowlist of plugins by spec + set of
```
eskultety, line 500: We should also explicitly document which approach solution we went with in the end.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,701 @@
...
+**Option 1: Ignore all plugins** — Before running `yarn install` during prefetch, set plugins in
+`.yarnrc.yml` to an empty array. Restore the original content afterwards.
+
+- Pros: Definitely safe
+- Cons: Disabling plugins may affect the prefetched content (but if all the protocols in
+  `yarn.lock` are known, the chance should be small — plugins can *add* resolvers and fetchers but
+  not necessarily *modify* existing ones)
+
+**Option 2: Allowlist of plugins to ignore** — Maintain an allowlist of plugins by
+[spec](https://github.com/chmeliik/berryscary/blob/c424d96e1e36542e52985aee716e1b12881c24fb/.yarnrc.yml#L3).
```
eskultety, line 492: Since our design depended on this experiment we might want to host it under our org to make sure it won't be deleted by accident. Non-actionable, just something to keep in mind (and maybe file an issue for).

docs/design/yarnberry.md
```diff
@@ -0,0 +1,701 @@
...
+[patch:](https://v3.yarnpkg.com/features/protocols#patch) protocol. Yarnberry will patch some
+dependencies automatically (e.g. typescript as explained in [Plug'n'Play][pnp]). For any
+patched dependency, Yarnberry creates two entries in both `yarn.lock` and `.yarn/cache`:
+- [typescript built-in patch](https://github.com/chmeliik/berryscary/commit/7d1727907e28759c9324f33289e841f2fe05e192)
+- [left-pad custom patch](https://github.com/chmeliik/berryscary/commit/cf1af13718236ee06635928a153662ed94b29490)
+
+**[Exec][protocols]**: The exec plugin allows running arbitrary code to generate a package. Either we ban
+exec altogether, or we accept arbitrary code execution and figure out SBOM reporting. Banning exec
+means parsing `yarn.lock` before running `yarn install` and failing if there's anything with an
+`exec:` protocol.
```
eskultety: The mechanics is a tiny bit different, because: 
- we detect that with `yarn info` rather than by parsing the `yarn.lock` file: https://github.com/hermetoproject/hermeto/blob/20ff8bb0d09b249e14e720865b118ddbc1625867/hermeto/core/package_managers/yarn/resolver.py#L202
- we parse the locators: https://github.com/hermetoproject/hermeto/blob/20ff8bb0d09b249e14e720865b118ddbc1625867/hermeto/core/package_managers/yarn/resolver.py#L167
- we reject the exec locators: https://github.com/hermetoproject/hermeto/blob/20ff8bb0d09b249e14e720865b118ddbc1625867/hermeto/core/package_managers/yarn/locators.py#L178

docs/design/yarnberry.md
```diff
@@ -0,0 +1,701 @@
...
+
+1. Should we trust the Yarnberry binary committed into the repo? This goes back to the same issue as plugins and the exec protocol.
+2. Can we trust Corepack to download the real Yarnberry?
+    - Corepack does validate that the packageManager field is <supported_pkg_manager>@<semver>. Hopefully that’s good enough.
+    - Corepack allows the user to specify the expected checksum. That protects the user (so we should recommend it), but does not protect us from a malicious user.
+    - And even if the answer is yes, is it OK to do the prefetch using a dynamically fetched version of Yarnberry? For what it’s worth, it shouldn’t hurt reproducibility (quite the opposite, given that the version is pinned in package.json).
+
+For now, I will assume a **1-No, 2-Yes situation**:
+
+1. Install Corepack in the cachi2 container, enable the `yarn` shim
```
eskultety: It's no longer `cachi2` but Hermeto :) .

docs/design/yarnberry.md
mohamedbouchtout: @eskultety Quick questions for "enableConstraintsChecks", here in the [yarn doc](https://yarnpkg.com/configuration/yarnrc#enableConstraintsChecks) it sets it to true, is this something we are aware of?

docs/design/yarnberry.md
mohamedbouchtout: I agree as that repo is referenced in different places in the doc as well.

docs/design/yarnberry.md
mohamedbouchtout: Just to clarify the approach that was taken is allowing a narrow exception list of plugins that don't execute code during yarn info, and failing everything else?

taylormadore: Yes, the docstring for the related function has some good info: https://github.com/hermetoproject/hermeto/blob/78c2c6b73f1614b395a4a8c09e32be0b5b847cc3/hermeto/core/package_managers/yarn/main.py#L201

docs/design/yarnberry.md
taylormadore: Yes, we use this function to specifically override any settings we need to control during the prefetch: https://github.com/hermetoproject/hermeto/blob/78c2c6b73f1614b395a4a8c09e32be0b5b847cc3/hermeto/core/package_managers/yarn/main.py#L225

docs/design/yarnberry.md
mohamedbouchtout: I'll file a issue for it once this is merged so I can link it in the issue.

docs/design/yarnberry.md
```diff
@@ -342,20 +341,8 @@ This workflow isn't supported and Hermeto will throw a `PackageRejected` error.
...
-
-Cachito deals with this by
-[banning git dependencies](https://github.com/containerbuildsystem/cachito/blob/6d7f809b0ab3ed34b426263d95bf0ae10213b436/cachito/workers/pkg_managers/general_js.py#L738-L741)
-that have
-[prepack or prepare scripts](https://github.com/containerbuildsystem/cachito/blob/6d7f809b0ab3ed34b426263d95bf0ae10213b436/cachito/workers/pkg_managers/general_js.py#L534-L535).
-We can do the same, but we would have to ban every script relevant to the Yarn 1, Yarnberry and npm
-`install` and `pack` commands (see
-[prepareExternalProject][prepareExternalProject]).
-And pnpm as well for good measure (in case Yarnberry adds support for it).
+Git deps are unsupported outright, so arbitrary code execution from git deps isn't a concern. Hermeto
```
taylormadore, line 388: For this entire commit about git dependencies, I wonder if instead of removing the information, we should just add notes that we don't support git dependencies for these reasons? We could conceivably add support in the future and it might be useful to have this data from the initial investigation.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,712 @@
...
+#### Project Structure
+
+```
+${output_dir}/deps/yarn
+├── cache
+│    ├── ccto-wo-deps-git@github.com-e0fce8c89c-8.zip
+└──  └── ccto-wo-deps-patch-c3567b709f-8.zip
+```
+
+The symlink structure is intended to give `deps/yarn` a familiar structure consistent with `deps/pip` and `deps/npm`.
```
taylormadore: I don't think we ever implemented a symlink structure, so we can probably strike all mention of that and denylists, etc. from this section

docs/design/yarnberry.md
```diff
@@ -117,6 +117,25 @@ Pretty much the same workspaces you already know from npm and Yarn 1. Could be w
...
+
+When workspace focus is used:
+
+- **Dependency resolution** runs `yarn workspace <name> info --recursive --cache --json` per
+  workspace instead of `yarn info --all`, deduplicating overlapping dependencies across workspaces.
+- **Fetching** uses `yarn workspaces focus <name> [<name>...]` instead of
+  `yarn install --mode skip-build`.
+- **Lifecycle scripts**: Because `yarn workspaces focus` does not support `--mode skip-build` and
+  `enableScripts: false` does not apply to workspace scripts
+  ([yarnpkg/berry#4781](https://github.com/yarnpkg/berry/pull/4781)), Hermeto strips the `scripts`
```
taylormadore: We should probably permalink to the code here rather than the PR

docs/design/yarnberry.md
```diff
@@ -308,6 +308,11 @@ For builtin patches, such as the one
   }
 }
 ```
+##### Proxy External References
+
+When a proxy URL is configured (see [Proxy Support](#proxy-support)), npm-registry components
```
taylormadore: nit: "Components with PURLs of type npm"

docs/design/yarnberry.md
```diff
@@ -0,0 +1,712 @@
...
+4. Prepare the configuration options relevant for prefetch (including proxy settings if configured)
+5. Disable plugins
+6. Run `yarn info …` to get the necessary data (or per-workspace info queries if using workspace
+   focus)
+7. Validate that we can parse needed locator in the output
+8. Reject unsupported dependency types (git, exec)
+9. If using workspace focus, strip `scripts` from workspace `package.json` files
+10. Run `yarn install …` or `yarn workspaces focus …` to fetch the dependencies (or check the
+    existing cache, if zero-installs)
+11. Restore the project to the original state (e.g. re-enable plugins)
```
taylormadore: I think we can drop this one. We don't restore plugins, we just do this work in an ephemeral copy of the repo and it gets thrown away as soon as the request is completed.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,712 @@
...
+    }
+  ]
+}
+```
+
+The path to the patch file can be parsed from the
+[locator](https://github.com/chmeliik/berryscary/blob/c424d96e1e36542e52985aee716e1b12881c24fb/yarn.lock#L740).
+We may want to set a reasonably large upper limit for the size of the patch file.
+
+For builtin patches, such as the one
```
taylormadore: I think we can strike this mention of separate handling for builtin patches. We didn't ever implement this and aren't planning to

docs/design/yarnberry.md
```diff
@@ -0,0 +1,712 @@
...
+
+| Variable Name | Purpose | Example Value | Required |
+|---------------|---------|---------------|----------|
+| `YARN_GLOBAL_FOLDER` | Points to dependency cache for non-zero-installs | `{hermeto_output}/deps/yarn` | Yes |
+| `YARN_IGNORE_PATH` | Use global `yarn` instead of local binary | `true` | Yes |
+| `NODE_USE_SYSTEM_CA` | Allow Node.js to use the OS trust store (certificates added via `update-ca-trust`) | `1` | Set automatically by Hermeto on every `yarn` invocation |
+
+#### Configuration Files
+
+##### Override for Build
```
taylormadore, line 579: We also set enableImmutableCache for the build: https://github.com/hermetoproject/hermeto/blob/c385a7a76d12a1ae6cd116c40962da18c4c95150/hermeto/core/package_managers/yarn/main.py#L323

docs/design/yarnberry.md
```diff
@@ -0,0 +1,712 @@
...
+always require git repository context regardless of mode and will raise `PackageRejected` if the
+project is not a git repo.
+
+##### Proxy External References
+
+When a proxy URL is configured (see [Proxy Support](#proxy-support)), npm-registry components
+include an `externalReference` in the SBOM with the proxy URL, annotated with the standard proxy
+reference type and comment.
+
+##### Alternative Registries
```
taylormadore, line 314: I don't believe we ever implemented/handled this. You can leave it here--IMO it's a To-Do

docs/design/yarnberry.md
```diff
@@ -0,0 +1,712 @@
...
+**For a regular workflow:**
+
+1. Set `$YARN_GLOBAL_FOLDER` to the directory where you want to prefetch dependencies
+   (i.e. `{hermeto_output}/deps/yarn`)
+2. `yarn install --mode=skip-build`
+   - `skip-build` makes sure that Yarnberry won't try to compile any node-gyp C(++) libraries and
+     will instead leave that to the build
+   - `skip-build` also ensures that the user's preinstall, install and postinstall lifecycle scripts
+     will not run (again, leaving them for the build)
+3. Undo accidental changes to the user's repo
```
taylormadore: nit: you can drop this--we don't have to do any special "undo" handling . Same comment for the section below

docs/design/yarnberry.md
```diff
@@ -0,0 +1,712 @@
...
+##### Respect for Prefetch
+
+| Option | Purpose |
+|--------|---------|
+| [`cacheFolder`][cacheFolder] | To find out and reject if the user is using [zero-installs][zero-installs] |
+| [`lockfileFilename`](https://yarnpkg.com/configuration/yarnrc#lockfileFilename) | Parse the lockfile specified here (default yarn.lock); probably not needed if we base SBOM generation on `yarn info` output instead |
+| [`npmRegistryServer`][v4-npmRegistryServer] and [`npmScopes`](https://yarnpkg.com/configuration/yarnrc#npmScopes) | The user can configure multiple different registries; if we don't respect them we cause Dependency Confusion. `yarn install` will respect them automatically |
+| [`yarnPath`](https://yarnpkg.com/configuration/yarnrc#yarnPath) | Depending on how we [handle Yarnberry installs][yarn-install-guide] |
+
+##### Don't Pollute the User's Repo
```
taylormadore: We can drop this section. There's no yarn-module-specific handling that we need to do to avoid changes to the user repo

docs/design/yarnberry.md
mohamedbouchtout: I'll leave it as is in that case.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+3. Check and reject if the project uses zero-installs
+4. Prepare the configuration options relevant for prefetch (including proxy settings if configured)
+5. Disable plugins
+6. Run `yarn info …` to get the necessary data (or per-workspace info queries if using workspace
+   focus)
+7. Validate that we can parse needed locator in the output
+8. Reject unsupported dependency types (git, exec)
+9. If using workspace focus, strip `scripts` from workspace `package.json` files
+10. Run `yarn install …` or `yarn workspaces focus …` to fetch the dependencies (or check the
+    existing cache, if zero-installs)
```
eskultety: > if zero-installs)

You concluded in step 3. in this section and multiple times across the doc that we'd reject zero-installs pretty early, it should be removed from here.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+include an `externalReference` in the SBOM with the proxy URL, annotated with the standard proxy
+reference type and comment.
+
+##### Alternative Registries
+
+See the [npmRegistryServer][v3-npmRegistryServer] and
+[npmScopes][v3-npmScopes] options. When these are present,
+add the `repository_url`
+[qualifier](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairs)
+to the purls for registry dependencies.
```
eskultety: looking at our Yarn Berry backend, it doesn't implement this logic with repository url qualifier.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+| Option | Purpose |
+|--------|---------|
+| [`cacheFolder`][cacheFolder] | To find out and reject if the user is using [zero-installs][zero-installs] |
+| [`lockfileFilename`](https://yarnpkg.com/configuration/yarnrc#lockfileFilename) | Parse the lockfile specified here (default yarn.lock); probably not needed if we base SBOM generation on `yarn info` output instead |
+| [`npmRegistryServer`][v4-npmRegistryServer] and [`npmScopes`](https://yarnpkg.com/configuration/yarnrc#npmScopes) | The user can configure multiple different registries; if we don't respect them we cause Dependency Confusion. `yarn install` will respect them automatically |
+| [`yarnPath`](https://yarnpkg.com/configuration/yarnrc#yarnPath) | Depending on how we [handle Yarnberry installs][yarn-install-guide] |
+
+#### Environment Variables
+
+**NOTE:** These are build-time environment variables that are set during prefetch, not during build.
```
eskultety: You're contradicting yourself here:

> These are build-time environment variables...set during prefetch

So these are either variables we set during prefetch for Hermeto's own purposes, i.e. when hermeto downloads stuff from the internet by driving Yarn vs **build-time variables that Hermeto dumps for the users** to adopt in their build env which does not include Hermeto in any way. So this is either-or, can't be both ;) .

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+
+| Variable Name | Purpose | Example Value | Required |
+|---------------|---------|---------------|----------|
+| `YARN_GLOBAL_FOLDER` | Points to dependency cache for non-zero-installs | `{hermeto_output}/deps/yarn` | Yes |
+| `YARN_IGNORE_PATH` | Use global `yarn` instead of local binary | `true` | Yes |
+| `NODE_USE_SYSTEM_CA` | Allow Node.js to use the OS trust store (certificates added via `update-ca-trust`) | `1` | Set automatically by Hermeto on every `yarn` invocation |
```
eskultety: These are prefetch variables.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+
+| Option | Value | Purpose |
+|--------|-------|---------|
+| [`checksumBehavior`][checksumBehavior] | `"throw"` | Strict checksum validation |
+| [`enableImmutableInstalls`](https://yarnpkg.com/configuration/yarnrc#enableImmutableInstalls) | `true` | Fail if yarn.lock needs an update |
+| [`globalFolder`][globalFolder] | `<output_dir>` | Where to download dependencies |
+| [`pnpMode`](https://yarnpkg.com/configuration/yarnrc#pnpMode) | `strict` | Modules won't be allowed to require packages they didn't list  |
+| [`enableMirror`][enableMirror] | `false` | Define whether to mirror local cache entries into the global cache or not |
+| [`enableGlobalCache`](https://yarnpkg.com/configuration/yarnrc#enableGlobalCache) | `true` | Define whether the cache should be shared between all local projects |
+| [`enableConstraintsChecks`](https://yarnpkg.com/configuration/yarnrc#enableConstraintsChecks) | `false` | Define whether constraints should run on every install |
```
eskultety: This last one is documented as unconditionally set always, but it's only relevant for Yarn v4 and hence set unconditionally only for a particular version.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+#### Configuration Files
+
+##### Override for Build
+
+| Option | Value | Purpose |
+|--------|-------|---------|
+| [`globalFolder`][globalFolder] | `<output_dir>` | Make offline builds work |
+| [`enableMirror`][enableMirror] | `true` | false would break globalFolder |
+| [`enableGlobalCache`][enableGlobalCache] | `false` | true would cause the same issue as [mounting the local cache][global-cache-commit] |
+| [`enableImmutableCache`](https://yarnpkg.com/configuration/yarnrc#enableImmutableCache) | `false` | Define whether to allow adding/removing entries from the lockfile or not |
```
eskultety: So you documented prefetch config in form of env variables, but you document the build override as Yarn config options verbatim. We expose these as variables too,have a look at `_generate_environment_variables`

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+The `yarn install` command will execute the lifecycle scripts (prepack, postpack etc.) of any
+git/github dependency that happens to have them. And yes, it's only git/github, no other type —
+see the references to the
+[prepareExternalProject][prepareExternalProject]
+method.
+
+Hermeto may deal with this by
+[banning git dependencies](https://github.com/containerbuildsystem/cachito/blob/6d7f809b0ab3ed34b426263d95bf0ae10213b436/cachito/workers/pkg_managers/general_js.py#L738-L741)
+that have
+[prepack or prepare scripts](https://github.com/containerbuildsystem/cachito/blob/6d7f809b0ab3ed34b426263d95bf0ae10213b436/cachito/workers/pkg_managers/general_js.py#L534-L535).
```
eskultety: In commit 5/10 you say 

> Removed all Cachito reference

Clearly some have survived. I think we can safely drop the links completely. The reason for that is that we have an open PR on adding git deps support to the yarn backend. So eventually this whole section will be either gone or heavily adjusted.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+[prepareExternalProject][prepareExternalProject]).
+And pnpm as well for good measure (in case Yarnberry adds support for it).
+
+#### Output Structure
+
+```
+${output_dir}/deps/yarn
+├── cache
+│    ├── ccto-wo-deps-git@github.com-e0fce8c89c-8.zip
+└──  └── ccto-wo-deps-patch-c3567b709f-8.zip
```
eskultety: Aren't these an example of git deps which we don't support?

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+| `NODE_USE_SYSTEM_CA` | Allow Node.js to use the OS trust store (certificates added via `update-ca-trust`) | `1` | Set automatically by Hermeto on every `yarn` invocation |
+
+#### Configuration Files
+
+##### Override for Build
+
+| Option | Value | Purpose |
+|--------|-------|---------|
+| [`globalFolder`][globalFolder] | `<output_dir>` | Make offline builds work |
+| [`enableMirror`][enableMirror] | `true` | false would break globalFolder |
```
eskultety: the description here seems a bit vague IMO.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,701 @@
...
+```
+
+When you have this shim installed and invoke `yarn` in a project that sets packageManager in package.json, Corepack will automatically download the right version of Yarn[berry] and use it to execute your command.
+
+Failing all the above (no `yarnPath`, no `packageManager` or no Corepack), the globally installed
+`yarn` will be used (if there is one).
+
+**So, what does this mean for us?**
+
+We need to make some decisions.
```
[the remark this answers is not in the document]
eskultety: Valid.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+#!/usr/bin/env node
+require('./lib/corepack.cjs').runMain(['yarn', ...process.argv.slice(2)]);
+```
+
+When you have this shim installed and invoke `yarn` in a project that sets packageManager in package.json, Corepack will automatically download the right version of Yarn[berry] and use it to execute your command.
+
+Failing all the above (no `yarnPath`, no `packageManager` or no Corepack), the globally installed
+`yarn` will be used (if there is one).
+
+**So, what does this mean for us?**
```
eskultety: This is conversational style, but we're adding a technical doc, so this needs to be reworded properly.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+For now, I will assume a **1-No, 2-Yes situation**:
+
+1. Install Corepack in the Hermeto container, enable the `yarn` shim
+2. Before processing a Yarnberry project, validate that it sets packageManager
+    - If the project doesn’t set packageManager but does set yarnPath, we can try to parse the version from the filename in yarnPath (and use Corepack to get that version).
+    - If the project sets both and yarnPath includes a version, validate that the versions match
+3. When running any `yarn` commands, make sure to set the ignorePath option (or the YARN_IGNORE_PATH variable) to use the global `yarn` rather than the local one
+4. And that’s it.
+
+**And what does it mean for the user?**
```
eskultety: ...same here, this is conversational style, not very appropriate for the doc.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+**For a zero-installs workflow:**
+
+This workflow isn't supported and Hermeto will throw a `PackageRejected` error.
+
+##### Arbitrary Code Execution During Prefetch
+
+Git deps are unsupported outright, so arbitrary code execution from git deps isn't a concern. Hermeto 
+will raise a `UnsupportedFeature` error when it encounters them.
+
+**Useful info:** Althrough git dependencies aren't supported, here is some useful information for it
```
eskultety: s/Althrough/Although
s/incase/in case

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+When workspace focus is used:
+
+- **Dependency resolution** runs `yarn workspace <name> info --recursive --cache --json` per
+  workspace instead of `yarn info --all`, deduplicating overlapping dependencies across workspaces.
+- **Fetching** uses `yarn workspaces focus <name> [<name>...]` instead of
+  `yarn install --mode skip-build`.
+- **Lifecycle scripts**: Because `yarn workspaces focus` does not support `--mode skip-build` and
+  `enableScripts: false` does not apply to workspace scripts. Hermeto strips the `scripts`
+  field from each workspace's `package.json` before running focus
+  ([_strip_workspace_scripts()](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/yarn/main.py#L267)).
```
eskultety: This points to a particular line in the code base which will invalidate the link the moment the function changes, we don't even do this for commits (you lock the reference to a particular past revision, not main).

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+| `npmAlwaysAuth` | `true` | Force authentication on every request (only if credentials are provided) |
+| `npmAuthIdent` | `<login>:<password>` | Proxy credentials (only if both login and password are provided) |
+
+If the proxy requires authentication and invalid or no credentials are provided, `yarn install` (or
+`yarn workspaces focus`) will fail with an "Invalid authentication" error. Hermeto catches this and
+raises a `PackageManagerError` with guidance to verify proxy settings.
+
+**For a zero-installs workflow:**
+
+This workflow isn't supported and Hermeto will throw a `PackageRejected` error.
```
eskultety: I think we already stated twice in an earlier section that this use case is unsupported.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,692 @@
...
+
+If `.yarn/cache` is checked into the repo, Hermeto doesn't even have to do anything. If it's not,
+Hermeto can populate the *global* cache (not the local one,
+[here's why][global-cache-commit]).
+The container build can mount the cache, set `YARN_GLOBAL_FOLDER` to the right path and offline
+installs work.
+
+##### [Zero-Installs][zero-installs]
+
+The concept of [zero-Installs][zero-installs], i.e. no install needed (git clone is
```
eskultety: s/-Installs/-installs

slimreaper35:
Maybe we could merge yarn v4 [ADR](https://github.com/hermetoproject/hermeto/blob/main/docs/adr/0002-yarn-v4.md) to this document. The file is quite short and it looks like ADRs are obsolete in the repo.

eskultety:
> Maybe we could merge yarn v4 [ADR](https://github.com/hermetoproject/hermeto/blob/main/docs/adr/0002-yarn-v4.md) to this document. The file is quite short and it looks like ADRs are obsolete in the repo.

Yes, this is a good opportunity to do that.

docs/design/yarnberry.md
mohamedbouchtout: I can expand on it, at first I didn't want to have really long rows in the table.

docs/design/yarnberry.md
mohamedbouchtout: A lot of it was copied from the old doc and for some reason it was written in this tone, I'll rewrite it.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,899 @@
...
+[in the lockfile](https://github.com/chmeliik/berryscary/blob/c424d96e1e36542e52985aee716e1b12881c24fb/yarn.lock#L1320-L1338)
+and in the cache — for both the unpatched and patched versions.
+
+##### [Offline Cache](https://v3.yarnpkg.com/features/offline-cache)
+
+Yarnberry is [designed to work offline by default](https://v3.yarnpkg.com/features/offline-cache).
+There is a local cache (`.yarn/cache`) and a global mirror (`~/.local/share/yarn/berry/cache`).
+
+If `.yarn/cache` is checked into the repo, Hermeto doesn't even have to do anything. If it's not,
+Hermeto can populate the *global* cache (not the local one,
```
eskultety: > Hermeto doesn't even have to do anything

This particular wording may be a bit unfortunate because in this instance "doesn't have to do anything" literally means Hermeto [will reject processing such a project](https://github.com/hermetoproject/hermeto/blob/4fc76df98d71df67c740bb1c85029d39124626fb/hermeto/core/package_managers/javascript/yarn/main.py#L86) because the cache presence [effectively indicates zero-installs](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/javascript/yarn/project.py#L97-L99), so we should IMO be more accurate here, i.e. offline-cache/zero-installs -> no need for Hermeto (should theoritically work with just Yarn), otherwise Hermeto handles deps via Yarn global cache...

docs/design/yarnberry.md
```diff
@@ -0,0 +1,899 @@
...
+build. If there are unknown plugins, either disable them or fail the build.
+
+- Pros: Probably safe; the allowlist can be more extensive (a plugin that affects the prefetched
+  content can be safe to execute)
+- Cons: Requires tediously maintaining a very precise allowlist and personally verifying the safety
+  of each addition
+
+**The approach taken: Have a default list and reject everything else** - The solution approach that was taken is a mixture of 
+both option 1 and 3. There is a list of only official plugins that add new protocols and that also do not implement the 
+`fetchPackageInfo` hook, since it would allow arbitrary code execution. Everything else not in the list will cause a failure.
```
eskultety: > Everything else not in the list will cause a failure.

it won't - we actually silently remove everything not in our plugin allow list: https://github.com/hermetoproject/hermeto/blob/4fc76df98d71df67c740bb1c85029d39124626fb/hermeto/core/package_managers/javascript/yarn/main.py#L237. The actual failure is more delicate - it only happens if a  plugin we dropped from the list appears in the package's locator IIUC.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,899 @@
...
+  }
+}
+```
+
+Whether the reported Path is the local or global one depends on the
+[enableGlobalCache](https://v3.yarnpkg.com/configuration/yarnrc#enableGlobalCache) setting.
+The command works even if the cache is empty — it reports the path where `yarn install` would place
+the dependency.
+
+Example (from "https://github.com/hermetoproject/hermeto/blob/main/tests/unit/package_managers/yarn/test_resolver.py#L52"):
```
eskultety: 2 issues with the URL:

1. we have since adopted an additional directory level `javascript` so it's a dead link
2. It's tied to the main branch instead of a particular commit hash where this design is most relevant for, so it may drift in the future.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,899 @@
...
+
+##### pnpDataPath is no longer configurable
+
+The `pnpDataPath` config option was removed; the default path `./.pnp.data.json` is now hard-coded.
+The only mention of `pnpDataPath` in Hermeto is the check for paths pointing outside of the repo.
+The check is kept for Yarn v3 compatibility and is simply skipped in v4.
+
+##### Yarn now caches npm version metadata
+
+Although Yarn only seems to create the metadata cache folder (`{globalFolder}/metadata/npm`) when
```
eskultety: nitpick: the sentence reads weird starting with "Although" and ending with "is enabled". One would expect some more information justifying the "Although".

docs/design/yarnberry.md
```diff
@@ -0,0 +1,899 @@
...
+Yarn v4 introduced the subdirectory and the hash to the previously existing locator. These parts
+are already handled by the v3 implementation.
+
+**Changes to some instances of the patch locator:**
+```
+-  resolution: "typescript@patch:typescript@npm%3A5.1.6#~builtin<compat/typescript>::version=5.1.6&hash=5da071"
++  resolution: "typescript@patch:typescript@npm%3A5.1.6#optional!builtin<compat/typescript>::version=5.1.6&hash=5da071"
+```
+
+Instances of the patch locator are currently being ignored in the v3 implementation.
```
eskultety: This doesn't seem to be right - we do process these locators and produce pedigree records in the CycloneDX SBOM.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,909 @@
...
+Yarnberry is [designed to work offline by default](https://v3.yarnpkg.com/features/offline-cache).
+There is a local cache (`.yarn/cache`) and a global mirror (`~/.local/share/yarn/berry/cache`).
+
+If `.yarn/cache` is checked into the repo, the project is using
+[zero-installs][zero-installs] and does not need Hermeto — Yarn can resolve dependencies offline
+on its own. Hermeto rejects such projects (see [Zero-Installs](#zero-installs)). Otherwise,
+Hermeto populates the *global* cache (not the local one,
+[here's why][global-cache-commit]).
+The container build can mount the cache, set `YARN_GLOBAL_FOLDER` to the right path and offline
+installs work.
```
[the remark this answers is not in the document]
mohamedbouchtout: Fixed — updated the offline cache section to document both the `.yarn/cache` (PnP mode) and `node_modules` (`nodeLinker: node-modules` or `pnpm`) zero-installs signals.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,909 @@
...
+
+##### Alternative Registries
+
+See the [npmRegistryServer][v3-npmRegistryServer] and
+[npmScopes][v3-npmScopes] options. When these are present,
+add the `repository_url`
+[qualifier](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairs)
+to the purls for registry dependencies.
+
+> **Note:** The `repository_url` qualifier is not yet implemented in the Yarn Berry backend.
```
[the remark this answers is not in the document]
mohamedbouchtout: Fixed — reworded to "should be added" instead of the imperative, making it clear this is planned behavior. The note about it not being implemented is preserved.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,909 @@
...
+Hermeto will rely on Yarnberry itself to populate the offline cache during prefetch.
+
+#### Prefetch Implementation
+
+Once you've [installed Yarnberry][yarn-install-guide],
+[dealt with plugins](#dealing-with-plugins) and the
+[user configuration](#dealing-with-user-configuration), prefetching is simple.
+
+We currently support only regular installs workflow. [Zero-installs][zero-installs] workflow will
+be rejected with a `PackageRejected` error.
```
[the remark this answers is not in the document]
mohamedbouchtout: Fixed — removed the stale "We currently support only regular installs workflow" sentence. The zero-installs rejection is still documented.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,909 @@
...
+**The approach taken: Have a default list and reject everything else** — A mixture of Option 1 and                                        
+Option 3. Hermeto maintains an                                                                                                            
+[allowlist](https://github.com/hermetoproject/hermeto/blob/4fc76df98d71df67c740bb1c85029d39124626fb/hermeto/core/package_managers/javascript/yarn/mainpy)
+of official plugins that add new protocols but do not implement the `fetchPackageInfo` hook (which 
+would allow arbitrary code execution). Currently, only the exec plugin is on this list — it is kept 
+so that `yarn info` can recognize exec locators, which are then rejected. All other plugins are                                           
+silently removed from `.yarnrc.yml` before processing. If a removed plugin's protocol appears in 
+the lockfile, the build fails with an unsupported locator error. Note that starting from v4, the 
+official plugins are enabled by default and can't be disabled. Since they're not present in the 
+[.yarnrc.yml][yarnrc-ref] file anymore, this function has no effect on v4 projects.
```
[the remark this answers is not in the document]
mohamedbouchtout: Fixed — corrected `mainpy` to `main.py` in the permalink.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,909 @@
...
+|--------|-------|---------|
+| [`enableScripts`](https://yarnpkg.com/configuration/yarnrc#enableScripts) | `false` | Extra safety (also achieved by --mode=skip-build) |
+| [`enableStrictSsl`](https://yarnpkg.com/configuration/yarnrc#enableStrictSsl) | `true` | Enforce strict SSL |
+| [`enableTelemetry`](https://yarnpkg.com/configuration/yarnrc#enableTelemetry) | `false` | Disable telemetry |
+| [`npmRegistryServer`][v4-npmRegistryServer] | proxy URL | Redirect registry traffic through proxy (only when proxy is configured) |
+| [`npmAlwaysAuth`](https://yarnpkg.com/configuration/yarnrc#npmAlwaysAuth) | `true` | Force authentication (only when proxy credentials are provided) |
+| [`npmAuthIdent`](https://yarnpkg.com/configuration/yarnrc#npmAuthIdent) | `<login>:<password>` | Proxy credentials (only when proxy credentials are provided) |
+| [`ignorePath`](https://yarnpkg.com/configuration/yarnrc#ignorePath) | `true` | Use global `yarn` instead of local |
+| [`unsafeHttpWhitelist`](https://yarnpkg.com/configuration/yarnrc#unsafeHttpWhitelist) | `[]` | Disallow HTTP |
+
```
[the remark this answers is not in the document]
mohamedbouchtout: Fixed — merged the two tables into a single table. All options are set unconditionally by Hermeto during prefetch.

docs/design/yarnberry.md
```diff
@@ -0,0 +1,903 @@
...
+git/github dependency that happens to have them. And yes, it's only git/github, no other type —
+see the references to the
+[prepareExternalProject][prepareExternalProject]
+method.
+
+Cachito deals with this by
+[banning git dependencies](https://github.com/containerbuildsystem/cachito/blob/6d7f809b0ab3ed34b426263d95bf0ae10213b436/cachito/workers/pkg_managers/general_js.py#L738-L741)
+that have
+[prepack or prepare scripts](https://github.com/containerbuildsystem/cachito/blob/6d7f809b0ab3ed34b426263d95bf0ae10213b436/cachito/workers/pkg_managers/general_js.py#L534-L535).
+We can do the same, but we would have to ban every script relevant to the Yarn 1, Yarnberry and npm
```
eskultety: This bit containing cachito appeared in one of the last revisions. What was the motivation? We have taken a few attempts to drop all cachito mentions in our code base as it's now very different. I don't think we should not mention it in any design anymore.

eskultety (commented):
LGTM with the cachito mention dropped. I think we're now in the shape where commits 1-10 can be squashed together, leaving the v4 info migration in a standalone commit.

docs/design/yarnberry.md
mohamedbouchtout: Must of somehow missed this reference with how many cachito references there were. I'll update it.

taylormadore (approved):

eskultety (approved):

tnevrlka (approved):

PR #1587, opened 2026-05-29
pnpm development dependencies

slimreaper35:


hermeto/core/package_managers/javascript/pnpm/resolver.py
```diff
@@ -104,3 +112,27 @@ def _create_root_component(project_dir: RootedPath, vcs_qualifiers: dict[str, st
...
+    """
+    snapshots = lockfile.snapshots
+    queue = deque(lockfile.root_dependencies)
+
+    seen = set()
+    while queue:
+        current_id = queue.popleft()
+        seen.add(current_id)
+
+        transitive_deps: dict[str, str] = snapshots.get(current_id, {}).get("dependencies", {})
```
taylormadore: Do we need to handle `optionalDependencies` as well? Using [argocd](https://github.com/argoproj/argo-cd/blob/f0d061349aa142e79fdb779560224080fd89783c/ui/pnpm-lock.yaml#L11759) as an example, `tsx@3.14.0` is a direct dependency, which has `fsevents@2.3.3` as an `optionalDependency`. Currently fsevents gets marked as dev, which I don't believe is correct

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -50,6 +55,41 @@ def packages(self) -> dict[str, dict[str, Any]]:
...
+        # Peer dependencies are stored under development dependencies in the pnpm-lock.yaml file.
+
+        ids = []
+        for name, data in {**dependencies, **optional}.items():
+            version = data["version"]
+            # For JSR dependencies, version is not a semver, it's the full lockfile package ID.
+            if version.startswith(JSR_REGISTRY_PREFIX):
+                ids.append(version)
+            else:
+                ids.append(f"{name}@{_clean_version_from_peer_dependencies(version)}")
```
taylormadore: I'm trying to understand why we do this before traversing snapshots, because the snapshot keys seem to have what is being removed: https://github.com/argoproj/argo-cd/blob/f0d061349aa142e79fdb779560224080fd89783c/ui/pnpm-lock.yaml#L7869

Don't we need to do this later when comparing the versions in the packages section rather than in snapshots?

hermeto/core/package_managers/javascript/pnpm/resolver.py
slimreaper35: > Do we need to handle optionalDependencies as well?

Yes, good catch 🌮 (I only processed root optional dependencies)

hermeto/core/package_managers/javascript/pnpm/project.py
slimreaper35: Apparently, I was looking at the packages section 😬

tests/unit/package_managers/javascript/pnpm/test_resolver.py
```diff
@@ -152,3 +162,38 @@ def test_create_root_component(rooted_tmp_path: RootedPath, name: str, expected_
...
+                        "d": {"version": "1"},
+                        "f": {"version": "1"},
+                    },
+                    "optionalDependencies": {
+                        "g": {"version": "1"},
+                    },
+                },
+            },
+            "snapshots": {
+                "a@1": {"dependencies": {"b": "1"}, "c": "1"},
```
eskultety: Is `c:1` as a top-level key instead of being nested under `dependencies` actually correct?

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -50,6 +52,30 @@ def packages(self) -> dict[str, dict[str, Any]]:
...
+        """Return the 'snapshots' key from the pnpm-lock.yaml file."""
+        return self.get("snapshots", {})
+
+    @property
+    def root_dependencies(self) -> list[str]:
+        """Return the root dependencies from the pnpm-lock.yaml file."""
+        importers: dict[str, dict[str, Any]] = self["importers"]["."]
+        dependencies = importers.get("dependencies", {})
+        optional = importers.get("optionalDependencies", {})
+        # Peer dependencies are stored under development dependencies in the pnpm-lock.yaml file.
```
eskultety: So what does this commentary actually tell us? Peer deps are not handled in this function, nor are dev deps.

tests/unit/package_managers/javascript/pnpm/test_resolver.py
```diff
@@ -152,3 +162,38 @@ def test_create_root_component(rooted_tmp_path: RootedPath, name: str, expected_
     component = _create_root_component(subproject, vcs_qualifiers)
 
     assert component.purl == expected_purl
+
+
+def test_find_non_dev_dependencies(tmp_path: Path) -> None:
+    lockfile = PnpmLock(
+        path=tmp_path,
+        data={
+            "lockfileVersion": "9.0",
```
eskultety, line 171: Aren't we targeting PNPM v10?

tests/unit/package_managers/javascript/pnpm/test_resolver.py
```diff
@@ -152,3 +162,38 @@ def test_create_root_component(rooted_tmp_path: RootedPath, name: str, expected_
...
+            "snapshots": {
+                "a@1": {"dependencies": {"b": "1"}, "c": "1"},
+                "b@1": {"dependencies": {"c": "1"}},
+                "c@1": {},
+                "d@1": {"dependencies": {"e": "1"}},
+                "e@1": {},
+                "f@1": {"optionalDependencies": {"h": "1"}},
+                "g@1": {},
+                "h@1": {},
+            },
```
eskultety: These seem quite artificial to me to the point I'm questioning this test's usefulness. See my comment where you do a membership test of package.id in the dev dependency set. I think we either need a tiny artificial lockfile that we'll parse and resolve to test this function OR you'll need to fill in the important bits in the dict object you intend to use.

hermeto/core/package_managers/javascript/pnpm/resolver.py
```diff
@@ -47,11 +52,14 @@ def _create_dependency_components(
             external_references = None
 
         purl = _generate_purl_for(package, vcs_qualifiers)
+        property_set = PropertySet(npm_development=package.id not in non_dev_dependencies)
```
eskultety, line 55: Is this membership verification correct?
Based on

1.  https://pnpm.io/10.x/how-peers-are-resolved, AND
2. https://github.com/pnpm/pnpm/issues/7685
3. https://github.com/hermetoproject/hermeto/blob/bb68e9747317de6f6870de86b5f5ef9e59b6e633/docs/design/pnpm.md?plain=1#L275

the way peer deps are recorded in snapshots is as follows:
```
snapshots:

  /ts-api-utils@1.2.1(typescript@4.9.5):      <<<<<<
    dependencies:
      typescript: 4.9.5
    dev: false
```

which means that snapshot IDs including peer deps will be keyed as `foo@1.0.0(peer-dep@version)` which `package.id` will never match.
Might be worth updating the design doc as well with a tiny note that snapshot IDs can carry peer dependency suffixes as well.

tests/unit/package_managers/javascript/pnpm/test_resolver.py
slimreaper35: no

tests/unit/package_managers/javascript/pnpm/test_resolver.py
slimreaper35: I don't think the lockfileVersion has changed. https://github.com/pnpm/spec/tree/master/lockfile

hermeto/core/package_managers/javascript/pnpm/resolver.py
slimreaper35: I love how pnpm randomly constructs those package IDs. So consistent!

tests/unit/package_managers/javascript/pnpm/test_resolver.py
eskultety: Oh,  well, I must have gotten those 2 mixed up.

tests/unit/package_managers/javascript/pnpm/test_resolver.py
slimreaper35: I don't know if an artificial lockfile would improve the test. Seems like converting JSON to YAML.

We could have more dev/optional deps in the e2e test, but that will significantly increase the total number of all deps, and it could be hard to identify which components should have the property.

eskultety: > We could have more dev/optional deps in the e2e test,

I don't think that's strictly necessary for what I'm trying to achieve - had the test used real package identifiers, versions, peer dep suffixes, it would have uncovered the problem in the check that is being proposed in the code. Like I said, replacing vague placeholder values like "a/b/c/" such that they cover realistic scenarios is fine with me.
My suggestion with an artificial lockfile was only meant as an option, I am aware of the conversion, I was mostly trying to point out that we should select whichever feels more readable in context of test definition.

hermeto/core/package_managers/javascript/pnpm/resolver.py
```diff
@@ -104,3 +112,27 @@ def _create_root_component(project_dir: RootedPath, vcs_qualifiers: dict[str, st
...
+        current_id = queue.popleft()
+        seen.add(current_id)
+
+        transitive_deps: dict[str, str] = snapshots.get(current_id, {}).get("dependencies", {})
+        for name, version in transitive_deps.items():
+            snapshot_id = f"{name}@{version}"
+            if snapshot_id not in seen:
+                queue.append(snapshot_id)
+
+    return seen
```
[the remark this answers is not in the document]
eskultety: Seems trivial to address, the AI comment makes sense.

hermeto/core/package_managers/javascript/pnpm/resolver.py
```diff
@@ -104,3 +112,42 @@ def _create_root_component(project_dir: RootedPath, vcs_qualifiers: dict[str, st
...
+        # https://github.com/argoproj/argo-cd/blob/bf1591de63e39b7c3be5f5ba54abe8763de1a48c/ui/pnpm-lock.yaml#L2164
+        # https://github.com/argoproj/argo-cd/blob/bf1591de63e39b7c3be5f5ba54abe8763de1a48c/ui/pnpm-lock.yaml#L7869
+        seen.add(_strip_peer_suffix(current_id))
+
+        current_snapshot = snapshots.get(current_id, {})
+        transitive_dependencies = current_snapshot.get("dependencies", {})
+        transitive_optional = current_snapshot.get("optionalDependencies", {})
+
+        for name, version in {**transitive_dependencies, **transitive_optional}.items():
+            new_id = f"{name}@{version}"
```
eskultety, line 135: So, we stripped peer suffix a few lines above for `current_id` but the way `new_id` could introduce one back if you have the same dependency, but that is pulled in by a different peer dependency. We'd add it to the set and technically loop twice through the same thing, am I correct in my thinking?
A simple `new_id = _strip_peer_suffix(f"{name}@{version}"` should be all we need to do IMO.

tests/unit/package_managers/javascript/pnpm/test_resolver.py
eskultety: I don't see this one addressed, the code still references arbitrary symbols that would have never exercised the peer suffix path that was flagged so I'd prefer some improvement here, shouldn't be much work you don't need to think too hard about it, just something that doesn't require reliance on an integration test because we're dealing with a simple internal helper function, that should be primarily covered with unit tests.

eskultety (commented):
Almost there, one more iteration and I'm good.

hermeto/core/package_managers/javascript/pnpm/resolver.py
```diff
@@ -104,3 +112,45 @@ def _create_root_component(project_dir: RootedPath, vcs_qualifiers: dict[str, st
...
+    queue = deque(seen)
+
+    while queue:
+        current_id = queue.popleft()
+        current_snapshot = lockfile.snapshots.get(current_id, {})
+
+        transitive_dependencies = current_snapshot.get("dependencies", {})
+        transitive_optional = current_snapshot.get("optionalDependencies", {})
+
+        for name, version in {**transitive_dependencies, **transitive_optional}.items():
```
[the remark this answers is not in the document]
slimreaper35: /gemini Having an empty dictionary as a fallback for `get` method is not enough? 

lockfile.snapshots.get(current_id, {})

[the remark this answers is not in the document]
eskultety: > Having an empty dictionary as a fallback for get method is not enough?

Maybe it's good that Gemini is being discontinued...

tests/unit/package_managers/javascript/pnpm/test_resolver.py
```diff
@@ -170,3 +162,64 @@ def test_create_root_component(rooted_tmp_path: RootedPath, name: str, expected_
...
+                "fresh-mint@2.3.3": {},
+            },
+        },
+    )
+
+    expected = {
+        "ground-beef@1.7.9",
+        "yellow-onion@4.4.3",
+        "russet-potato@1.15.9",
+        "smoked-paprika@1.1.0",
```
taylormadore: Is this one redundant?

tests/unit/package_managers/javascript/pnpm/test_resolver.py
```diff
@@ -170,3 +162,64 @@ def test_create_root_component(rooted_tmp_path: RootedPath, name: str, expected_
...
+
+
+def test_find_non_dev_dependencies(tmp_path: Path) -> None:
+    lockfile = PnpmLock(
+        path=tmp_path,
+        data={
+            "lockfileVersion": "9.0",
+            "importers": {
+                ".": {
+                    "dependencies": {
```
taylormadore, line 174: Can we add a peer dep to this section to test the issue we fixed?

tests/unit/package_managers/javascript/pnpm/test_resolver.py
```diff
@@ -170,3 +162,64 @@ def test_create_root_component(rooted_tmp_path: RootedPath, name: str, expected_
...
+                "active-dry-yeast@3.0.5": {
+                    "dependencies": {"red-food-coloring": "1.1.1"},
+                },
+                "red-food-coloring@1.1.1": {},
+                "fresh-mint@2.3.3": {},
+            },
+        },
+    )
+
+    expected = {
```
taylormadore, line 216: It might be helpful to have the package names describe what case is being tested. Some examples: runtime, dev, runtime-and-dev, transitive-optional, runtime-with-peer, etc.

taylormadore (commented):
Just test nits remaining. Otherwise LGTM

tests/unit/package_managers/javascript/pnpm/test_resolver.py
slimreaper35: Rewrote the whole unit test. No need to have a theme for that. I've probably burnt too many tokens.

eskultety: The test now reads like a real-world scenario, much less artificial!

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -50,6 +52,29 @@ def packages(self) -> dict[str, dict[str, Any]]:
...
+
+        ids = []
+        for name, data in {**dependencies, **optional}.items():
+            version = data["version"]
+            # For JSR dependencies, version is not a semver, it's the full lockfile package ID.
+            if version.startswith(JSR_REGISTRY_PREFIX):
+                ids.append(version)
+            else:
+                ids.append(f"{name}@{version}")
+
```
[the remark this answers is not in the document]
eskultety: This is an easy fix, doesn't cost us anything and promotes defensiveness, fair suggestion.

hermeto/core/package_managers/javascript/pnpm/resolver.py
eskultety: I believe this is the last open comment and we're good to go.

slimreaper35: I don't think I understand. I moved the stripping of the peer suffix at the end of the function for simplicity.

hermeto/core/package_managers/javascript/pnpm/project.py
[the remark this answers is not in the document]
slimreaper35: ok

hermeto/core/package_managers/javascript/pnpm/resolver.py
eskultety: my bad, I didn't expand the context of the hunk to see the original strip is now gone.

eskultety (approved):

taylormadore (approved):

PR #1588, opened 2026-05-29
refactor(tests): consolidate and remove redundant unit tests vol. 2

### LINKED ISSUE #1580: [tests] Revisit unit tests for redundancy and consolidation (volume 2)
Follow-up to #1506 (closed by PR #1526).

  Volume 1 consolidated duplicates and removed IT-covered
  wiring tests. There might be more to cull in the same
  categories, plus stale test infrastructure and mock data
  generation scripts worth looking into.

  Current unit test counts (main, 1401 total)

  - pip: 321
  - yarn: 186
  - gomod: 157
  - models: 134
  - npm: 105
  - cli: 100
  - yarn_classic: 103
  - rpm: 47
  - bundler: 24
  - cargo: 22
  - pnpm: 11
  - other: 191

derasdf:
Follow-up to https://github.com/hermetoproject/hermeto/pull/1526.

  - Delete ~80 wiring and framework test functions, trim parametrize params that removed ~30 more (1414 -> 1303)
  - Fix 3 broken test assertions (dead code that silently passed for a while)
  - Remove hack/mock-unittest-data/ scripts and orphaned test data files
  - Drop fragile string matching in assertions (per previous PR review feedback)

  Each deleted test was checked against the code it tests and the IT that covers it.

  Closes https://github.com/hermetoproject/hermeto/issues/1580


tests/unit/package_managers/javascript/test_metayarn.py
```diff
@@ -14,88 +14,72 @@
...
+    expect_yarnberry_called: bool,
 ) -> None:
-    mock_yarnclassic_fetch_source.side_effect = NotV1Lockfile("/some/path")
+    mock_yarnclassic_fetch_source.side_effect = classic_side_effect
 
     _ = fetch_yarn_source(input_request)
 
     mock_yarnclassic_fetch_source.assert_called_once()
-    mock_yarnberry_fetch_source.assert_called_once()
+    assert mock_yarnberry_fetch_source.called == expect_yarnberry_called
```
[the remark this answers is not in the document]
derasdf: Introduces branching, keeping as-is.

tests/unit/package_managers/javascript/test_metayarn.py
```diff
@@ -14,88 +14,72 @@
...
-    )
+    mock_yarnclassic_fetch_source.side_effect = classic_side_effect
+    mock_yarnberry_fetch_source.side_effect = berry_side_effect
 
     with pytest.raises(PackageRejected):
         _ = fetch_yarn_source(input_request)
 
     mock_yarnclassic_fetch_source.assert_called_once()
-    mock_yarnberry_fetch_source.assert_called_once()
+    assert mock_yarnberry_fetch_source.called == expect_yarnberry_called
```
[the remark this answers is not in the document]
derasdf: Introduces branching, keeping as-is.

a-ovchinnikov (approved):
This is a huge cleanup, thank you!

tests/unit/package_managers/javascript/test_metayarn.py
```diff
@@ -14,84 +14,72 @@
...
+    expect_yarnberry_called: bool,
 ) -> None:
-    mock_yarnclassic_fetch_source.side_effect = NotV1Lockfile("/some/path")
+    mock_yarnclassic_fetch_source.side_effect = classic_side_effect
 
     _ = fetch_yarn_source(input_request)
 
     mock_yarnclassic_fetch_source.assert_called_once()
-    mock_yarnberry_fetch_source.assert_called_once()
+    assert mock_yarnberry_fetch_source.called == expect_yarnberry_called
```
eskultety: This feels odd. Same as above - we should infer the bolean. We should also stick to using the common and recommended method interfaces that also provide call counting rather than simple boolean attributes like 'called'.

tests/unit/package_managers/javascript/test_metayarn.py
```diff
@@ -14,84 +14,72 @@
...
 @mock.patch("hermeto.core.package_managers.javascript.metayarn.fetch_yarnberry_source")
 @mock.patch("hermeto.core.package_managers.javascript.metayarn.fetch_yarn_classic_source")
-def test_fetch_yarn_source_detects_yarnberry(
+def test_fetch_yarn_source_detects_yarn_version(
     mock_yarnclassic_fetch_source: mock.Mock,
     mock_yarnberry_fetch_source: mock.Mock,
     mock_requestoutput_add_: mock.Mock,
     input_request: Request,
+    classic_side_effect: Exception | None,
+    expect_yarnberry_called: bool,
```
eskultety: I'm not convinced this is correct wrt parametrization. We're not parametrizing the test itself, we're passing values to mod the behaviour. The test is simple enough that I'm okay with the consolidation but we should not pass these values. You should infer the boolean based on the nodespec ID.

tests/unit/package_managers/javascript/test_metayarn.py
```diff
@@ -14,84 +14,72 @@
...
+            False,
+            id="yarn_classic_error",
+        ),
+        pytest.param(
+            NotV1Lockfile("/some/path"),
+            PackageRejected("this is a very bad package!"),
+            True,
+            id="yarnberry_error",
+        ),
+    ],
```
eskultety: we don't need as many attributes if you infer the value properly during runtime, this way of test spec is incorrect.

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -146,77 +146,8 @@ def test_retry_failure(
         assert mock_run.call_count == 5
         assert mock_sleep.call_count == 4
 
-    @pytest.mark.parametrize("release", ["go1.20", "go1.21.1"])
-    @mock.patch.object(Go, "__post_init__", lambda self: None)
-    @mock.patch("hermeto.core.package_managers.gomod.go.tempfile.TemporaryDirectory")
-    @mock.patch("pathlib.Path.home")
-    @mock.patch("hermeto.core.package_managers.gomod.go.Go._retry")
-    @mock.patch("hermeto.core.package_managers.gomod.go.get_cache_dir")
-    def test_from_missing_toolchain(
```
eskultety, line 155: Based on the commit message, this test is removed because:

_test_from_missing_toolchain (5 mocks, fragile call_args_list
  inspection, covered by gomod/e2e-1.21 which downloads a real
  toolchain)_

I don't think that's the case. Can you point me to the exact line where we download a real toolchain in the e2e-1.21 using this helper?

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -146,77 +146,8 @@ def test_retry_failure(
...
-        assert mock_go_retry.call_args_list[1][0][0][1] == "download"
-
-        target_binary = dest_cache_dir / f"go/{release}/bin/go"
-        assert not sdk_source_dir.exists()
-        assert target_binary.exists()
-        assert result_go.binary == str(target_binary)
-
 
 class TestGoWork:
-    def test_init(
```
eskultety, line 206: I don't think constructor smoke tests violate any good unit testing practices, this should IMO stay.

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -234,16 +165,6 @@ def test_from_file(
...
         assert go_work.data == go_work_data
 
-    @pytest.mark.parametrize(
-        "go_work_data, expected",
-        [
-            pytest.param({}, False, id="empty"),
-            pytest.param({"foo": "bar"}, True, id="with_data"),
-        ],
-    )
-    def test_bool(self, rooted_tmp_path: RootedPath, go_work_data: dict, expected: bool) -> None:
```
eskultety, line 244: What if `bool` becomes more complex? Given it's trivial now, maybe we'd want a doctest instead which is even lighter.

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -283,15 +204,6 @@ def test_workspace_paths(
...
 
-    def test_get_go_work(self) -> None:
-        mock_go = mock.Mock(spec=Go)
-        mock_go.return_value = None
-
-        GoWork._get_go_work(mock_go, {})
-
-        mock_go.assert_called_once()
-        assert mock_go.call_args[0][0] == ["work", "edit", "-json"]
-
```
eskultety, line 294: Where is the circular mock as reported exactly? I'd lean towards keeping this test.

slimreaper35 (approved):
LGTM, but I would wait for more people to take a look since the PR removes a lot of stuff.

slimreaper35:
Maybe a small nitpick would be that now have `hack/script` nested directory for just one file. IMO `hack/update_artifacts_lockfile` is enough.

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -551,95 +463,6 @@ def test_ignore_toolchain_files(dir_path: str, files: list[str], expected: list[
...
-                ],
-            }
-            """,
-            id="missing_mandatory_attribute",
-        ),
-    ],
-)
-def test_go_work_model_fail(input_json: str) -> None:
-    with pytest.raises(ValueError):
-        ParsedGoWork.model_validate_json(input_json)
```
eskultety, line 640: These don't test the framework as the commit message hints. These test that what we correctly process what we parsed out of the `go.mod` JSON representation. Input representation model tests are  important for us to keep.

tests/unit/package_managers/javascript/yarn/test_main.py
eskultety, line 150: >  Delete test_corepack_installed_correct_yarn_version (happy path,
>   every yarn IT runs corepack version check)

Based on this claim in the commit message I don't think ^this existing test provides enough value - what if we update the code and we suddenly make this branch a dead code? The test would still pass, but would not be in production. I would suggest converting this test to a scoped negative IT test and get rid of this UT as well. That way we guarantee we always perform the version check in production.

eskultety: Same for the Yarn v1 counterpart.

tests/unit/package_managers/javascript/test_metayarn.py
eskultety: nitpick: maybe a short test description along the lines that we always start by considering v1 and only then moving onto Berry would be helpful IMO after this consolidation.
Same comment on the negative variant.

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -155,17 +155,6 @@ def test_prepare_for_hermetic_build_injects_necessary_variable_into_existing_alt
     assert result.template == existing_preamble + expected_alternate_config_contents
 
 
-@mock.patch("hermeto.core.package_managers.bundler.main.get_repo_id")
```
eskultety, line 199: You seem to follow a per-backend commit structure -> split to bundler and generic standalone commits.

tests/unit/package_managers/javascript/test_metayarn.py
```diff
@@ -14,84 +14,72 @@
     (pytest.param([{"type": "yarn", "path": "."}], id="no_input_packages"),),
     indirect=["input_request"],
 )
-@mock.patch("hermeto.core.package_managers.javascript.metayarn.RequestOutput.__add__")
```
eskultety, line 17: nitpick: I believe anything Yarn should go to the same commit, including metayarn logic.

tests/unit/test_utils.py
```diff
@@ -111,30 +111,28 @@ def test_fast_copy(mock_blocksize: mock.Mock, tmp_path: Path) -> None:
...
+    mock_copy_range: mock.Mock, tmp_path: Path, errno_: int, expected_exc: type[Exception]
+) -> None:
     src = tmp_path / "src/test"
     dest = tmp_path / "dest/test"
     for dir_ in [src.parent, dest.parent]:
         dir_.mkdir()
+    src.write_text("data")
 
-    mock_copy_range.side_effect = OSError(errno_)
+    mock_copy_range.side_effect = OSError(errno_, "mock error")
```
eskultety, line 131: Good catch on this one! Easily missed in the docs: https://docs.python.org/3/library/exceptions.html#OSError

tests/unit/test_cli.py
```diff
@@ -217,18 +205,11 @@ def side_effect(request: Any) -> RequestOutput:
...
-        ],
-    )
-    def test_mode_option_is_not_valid(self, mode: str) -> None:
-        args = ["--mode", mode, "fetch-deps", "gomod"]
+    def test_mode_option_is_not_valid(self) -> None:
+        args = ["--mode", "invalid", "fetch-deps", "gomod"]
         with mock_fetch_deps():
             result = invoke_expecting_invalid_usage(app, args, ExitError.ERR_USAGE)
-            assert f"Invalid value for '--mode': '{mode}' is not one of" in result.output
+            assert "is not one of" in result.output
```
eskultety: nitpick: would `'--mode': 'invalid'` be an equally prone assertion providing a slightly better visibility into what's being verified? I'd also like to presume that `ERR_USAGE` should provide enough coverage for such a basic test we could therefore drop string comparison.

hack/mock-unittest-data/gomod.py
```diff
@@ -1,161 +0,0 @@
-#!/usr/bin/env python3
```
eskultety, line 1: **Commit message:**

> became redundant when eskultety added
>   test_go_list_deps (Oct 2024)

No need for personal references :) - I believe one can change their GH username

eskultety (commented):
I have some comments when it comes to removing UTs due to IT coverage, but the series keeps getting better as one goes through the patches providing fixes and improviements to some nasty stuff towards the end of the series :) .

tests/unit/package_managers/javascript/test_metayarn.py
derasdf: Reverted `test_metayarn.py`

tests/unit/package_managers/javascript/test_metayarn.py
derasdf: Reverted `test_metayarn.py`

tests/unit/package_managers/javascript/test_metayarn.py
derasdf: Reverted `test_metayarn.py`

tests/unit/package_managers/gomod/test_go.py
derasdf: AI thing, reverted

tests/unit/package_managers/gomod/test_go.py
derasdf: Reverted

tests/unit/package_managers/gomod/test_go.py
derasdf: Hm fair, I just reverted

tests/unit/package_managers/gomod/test_go.py
derasdf: Reverted

tests/unit/package_managers/gomod/test_go.py
derasdf: Reverted and at this point i just reverted the whole `test_go.py`

tests/unit/package_managers/javascript/yarn/test_main.py
derasdf: Makes sense, should we just file an issue in IT about it to not forget to do it later?

tests/unit/package_managers/javascript/test_metayarn.py
derasdf: Reverted `test_metayarn.py`

tests/unit/package_managers/bundler/test_main.py
derasdf: I'd rather keep it as is already

tests/unit/package_managers/javascript/test_metayarn.py
derasdf: Reverted `test_metayarn.py`

tests/unit/test_cli.py
derasdf: Done

hack/mock-unittest-data/gomod.py
derasdf: Done

tests/unit/package_managers/bundler/test_main.py
eskultety: It's integration tests which are currently not required for every commit (yet), so I don't care much.

tests/unit/package_managers/javascript/yarn/test_main.py
eskultety: Sure.

tests/unit/package_managers/gomod/test_main.py
```diff
@@ -1591,34 +1056,6 @@ def test_fetch_tags_fail(repo_remote_with_tag: tuple[RootedPath, RootedPath]) ->
...
-    params = {"env": {"GOTOOLCHAIN": "auto"}}
-    _disable_telemetry(go, params)
-
-    if not telemetry_disable:
-        assert mock_run_cmd.call_count == 1
-    else:
-        assert mock_run_cmd.call_count == 2
-        mock_run_cmd.assert_called_with(cmd, params)
-
-
```
eskultety, line 1621: This is a correct test I'd say. It's a small function exactly for the reason to be easily unit tested. It checks whether we correctly attempt to disable telemetry. I don't think we'd change the logic, but there's no heavy mocking going on, so I'd keep it.

> test_disable_telemetry (3-line function, mocks run_cmd,
>   branching in test body, every gomod IT exercises via
>   _resolve_gomod)
> 

Note that the fact the ITs exercise it doesn't mean it's working, the code probably flows through it, but we don't verify in ITs we really disabled it, right? So I think the reasoning is a bit inaccurate. I'm ambivalent on whether I'd want us to be hardcore and verify logs on whether we actually even attempted to disable it, most likely we wouldn't - it's not a deal-breaker feature. My point here being - to properly test we actually do disable telemetry it we'd need to re-introduce the `test_resolve_gomod` and we already concluded the level of mocking going on in there was unmanageable.

hack/mock-unittest-data/gomod.py
eskultety, line 1: > - test_parse_packages: became redundant when test_go_list_deps
  was added (Oct 2024) testing the same parsing with inline data.
...
>The hack/mock-unittest-data/ scripts cloned integration-tests
...
>Inlined go_work.json into TestGoWork._GO_WORK_JSON class constant
...

There's quite a few deletions happening here. The other changes may easily get lost. Aren't ^these a proof this commit should be split further for readability?

eskultety (commented):
I think I'm ready to part with `mock-unittest-data` :) . When tests are introduced the right way a strict unit/integration separation is all we should need.
A few comments left, but LGTM already, ready to be merged soon.

tests/unit/package_managers/gomod/test_main.py
derasdf: Restored, fair point

hack/mock-unittest-data/gomod.py
derasdf: Split into 3 commits

eskultety (approved):

PR #1589, opened 2026-06-01
feat(pip): add pylock.toml lockfile support

ST3V1K:
## Summary

PEP 751 introduced pylock.toml as a standardized Python lockfile format. This PR adds support for it in the pip backend, letting users provide pylock lockfiles instead of requirements.txt. The new parser extracts dependencies directly from the TOML structure and feeds them into the existing download and SBOM pipeline.

## Details

- New `lockfile`, `lockfile_extras`, and `packaging_tool` input fields for specifying pylock lockfiles alongside the existing requirements fields
- Dedicated pylock.toml parser handling archive, VCS, and PyPI sources with per-artifact hash extraction
- Integration into the pip backend for download, checksum verification, and SBOM generation 
- Refactored `_resolve_pip` to dispatch through `PipPackagingTool`, eliminating duplicated orchestration logic between lockfile formats
- Unit and integration tests covering parser validation, dependency downloading, and consolidated pylock scenarios (index packages, VCS, multiple packages)


hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,314 @@
...
+        except pydantic.ValidationError as e:
+            raise InvalidLockfileFormat(
+                lockfile_path=self.file_path.path,
+                err_details=str(e),
+                solution="Regenerate the lockfile with the correct version.",
+            ) from e
+
+    def _load_data(self) -> dict[str, Any]:
+        """Load the data from the lockfile."""
+        with open(self.file_path) as f:
```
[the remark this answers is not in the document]
ST3V1K: `.path` is unnecessary. Both will work

tests/integration/test_pip.py
```diff
@@ -133,6 +133,114 @@
             ),
             id="pip_rust_dependency_unusual_cargo_toml_location",
         ),
+        pytest.param(
+            utils.TestParameters(
+                branch="pip/pylock-without-deps",
+                packages=(
```
slimreaper35, line 139: I hope we can reduce the number of test scenarios here. This would lead to a total of 21 integration/e2e tests for pip 🙏

tests/unit/package_managers/python/pip/test_main.py
```diff
@@ -975,3 +987,133 @@ def test_process_url_req(
     result = pip._process_url_req(req, pip_deps_dir=mock.Mock(), trusted_hosts=set())
     assert result is not None
     assert result.get("package_type") == expected_type
+
+
+class TestDownloadPylock:
+    """Tests for dependency downloading from pylock.toml lockfiles."""
```
slimreaper35, line 964: nitpick: It's more common to introduce unit tests along with the code change than separately.

slimreaper35 (commented):
Only high-level notes. Rebase needed.

tests/integration/test_pip.py
ST3V1K: I don't think that this would result in 21 integration/e2e tests for pip. This "only" adds 7 new integration tests for pylock-specific features.
I'm sorry, but I'm not sure how you got to 21.

tests/unit/package_managers/python/pip/test_main.py
ST3V1K: I was doing it this way, and nobody said anything, so I thought that it was okay. I'll start doing it along with the code change.

eskultety: > I was doing it this way, and nobody said anything, so I thought that it was okay

It's fine, I've done it on many occasions myself. Separation really helps with complex changes to reduce the overall diff. Test separation in commits with new stuff is totally fine with me, I'd say it's a matter of taste. What I will say though is that when I'd block on @slimreaper35's comment is when a change requires test updates -> breaking tests in between the series.

slimreaper35: Okay, it was just a nitpick/note.

hermeto/core/models/input.py
```diff
@@ -25,6 +26,25 @@
 ModelT = TypeVar("ModelT", bound=pydantic.BaseModel)
 
 
+class PipPackagingTool(Enum):
+    """A packaging tool used to generate a lockfile."""
```
a-ovchinnikov: Could you please add a few words describing how this class should be used? Maybe a doctest to help its users? Right now it looks like a customized enum with some extra fields bolted on, the "why" is not very obvious.

ST3V1K: Done

a-ovchinnikov: Neat, thank you!

hermeto/core/models/input.py
```diff
@@ -315,8 +335,13 @@ class PipPackageInput(_PackageInputBase):
...
     allow_binary: bool = False
     binary: PipBinaryFilters | None = None
+    lockfile: Path | None = None
+    lockfile_extras: list[Path] | None = None
+    packaging_tool: PipPackagingTool | None = None
 
-    @pydantic.field_validator("requirements_files", "requirements_build_files")
+    @pydantic.field_validator(
+        "requirements_files", "requirements_build_files", "lockfile", "lockfile_extras"
+    )
```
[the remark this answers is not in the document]
eskultety: Valid concern, the current type checking is slightly confusing.

hermeto/core/models/input.py
```diff
@@ -25,6 +26,47 @@
...
+    """
+
+    REQUIREMENTS = "requirements", "requirements.txt", "requirements-build.txt"
+    PYLOCK = "pylock", "pylock.toml", "pylock.build.toml"
+
+    def __new__(cls, *args: Any) -> Self:
+        """Create a new PackagingTool instance."""
+        obj = object.__new__(cls)
+        obj._value_ = args[0]
+        obj._name_ = args[0]
```
eskultety, line 61: What's the point of assigning `_name_`, the Enum metaclass will override it.

hermeto/core/models/input.py
```diff
@@ -25,6 +26,47 @@
...
+    >>> PipPackagingTool.REQUIREMENTS.file
+    'requirements.txt'
+    >>> PipPackagingTool.PYLOCK.file
+    'pylock.toml'
+    >>> PipPackagingTool.PYLOCK.build_file
+    'pylock.build.toml'
+    """
+
+    REQUIREMENTS = "requirements", "requirements.txt", "requirements-build.txt"
+    PYLOCK = "pylock", "pylock.toml", "pylock.build.toml"
```
eskultety, line 55: I have to say this is quite some hacky design - specifying multiple Enum options and then overriding `__new__` to only accept the first one and defer the rest to `__init__`. It took me quite a while to wrap my head around this. The docstring provides examples, but there isn't a commentary on why it's designed this way. Is this by any chance some common pattern in Python with Enums? I'm wondering if we should model this as a plain string enum and provide properties to pull the type of file one needs.

tests/integration/test_pip.py
slimreaper35: Existing 14 (or 15 I don't know) + 7.  I believe the approach follows our current pip integration test suite.

But it depends what the team prefers. I think having a separate test for git dependency, another one for local dependency etc., is overkill and adds SBOM noise. They could be easily merged together to a real-world project and hermetic build test. Related to https://github.com/hermetoproject/hermeto/issues/803.

ST3V1K: Oh, sorry, you were counting the existing ones. I'll try to reduce the number of scenarios.

hermeto/core/models/input.py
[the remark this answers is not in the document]
ST3V1K: Done

hermeto/core/models/input.py
```diff
@@ -315,26 +357,42 @@ class PipPackageInput(_PackageInputBase):
...
+
     @pydantic.model_validator(mode="after")
     def _handle_legacy_allow_binary_field(self) -> Self:
         """Handle backward compatibility for allow_binary field."""
+        if (self.lockfile is not None or self.lockfile_extras is not None) and (
+            self.requirements_files is not None or self.requirements_build_files is not None
+        ):
+            raise ValueError("cannot specify both requirements files and lockfile")
+
         _handle_legacy_allow_binary(self, PipBinaryFilters)
```
eskultety, line 396: I don't think this is the right place for this mutual exclusion check, perhaps a new validator?

hermeto/core/package_managers/python/pip/main.py
```diff
@@ -559,28 +594,66 @@ def resolve_req_files(req_files: list[Path] | None, devel: bool) -> list[RootedP
...
-        output_dir, resolved_build_req_files, binary_filters
-    )
+        resolved_build_lockfiles = resolve_lockfiles(lockfile_extra_paths, True)
+        if not resolved_build_lockfiles:
+            log.info("No build lockfiles found")
+        else:
+            log.info(
+                "Using build lockfiles: %s",
+                ", ".join(str(f.subpath_from_root) for f in resolved_build_lockfiles),
+            )
```
eskultety, line 649: Hmm, going back to the design: https://github.com/hermetoproject/hermeto/pull/1458

Looking at
```json
{
  "type": "pip",
  "lockfile": "pylock.toml",
  "lockfile_extras": ["pylock.build.toml"],
  "packaging_tool": "pylock"
}
```

The code here in this commit has made me think of something I didn't think about back then when we were reviewing the design - `lockfile_extras` might as well be `requirements-build.txt` because it's still pip, so I can kinda see various setups where projects mix and match based on tooling availability, particular downstream needs, etc.

However, as the logic is written if `lockfile` is specified it determines the parser for extra files as well, if it's not specified then we'll try to infer the packaging tool and use that information uniformly for both the main as well as auxiliary files.
Reading the design I don't think this bit is emphasized enough in the design, so we might need to make it more obvious if we want that coupling.
 
On the other hand, should we want more flexibility (e.g. `pylock.toml` + `requirements-build.txt` + `requirements-foo.txt` ) we might need to invent a custom field formatting to carry the information properly, e.g.
`filename@packaging_tool` in which case we no longer need 3 fields in the input JSON to carry the information, just a single one, e.g.
```json
{
  "type": "pip",
  "lockfiles": ["pylock.toml", "requirements-downstream.txt@pip"]
}
```

Also note, naming the auxiliary files with "build" might be misleading because it's not always just build deps. In case of `pylock.<name>.toml` it most likely will be build deps due to it being fully resolved, but e.g. with requirements files the number of files present in a repo is determined by how many indices a given project needs to work with and download content from (again, not something that would need to be covered with pylock).

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,312 @@
...
+    @pydantic.field_validator("lock_version")
+    @classmethod
+    def _validate_version(cls, v: str) -> str:
+        if Version(v) not in Specifier("~=1.0"):
+            raise ValueError(f"unsupported lockfile version: {v}")
+        return v
+
+
+class PipLockfile(ABC):
+    """Base class for a pip lock-style files."""
```
eskultety, line 87: We're declaring an abstract class but not enforcing any abstract methods nor providing common implementation, does such a base class provide real value at this point?

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,312 @@
...
+            )
+
+        return cls(
+            url=source.get("url"),
+            hashes=[f"{alg}:{dig}" for alg, dig in source.get("hashes", {}).items()],
+            subdirectory=source.get("subdirectory"),
+        )
+
+
+class PyLockPackage(PipRequirement, ABC):
```
eskultety, line 179: Hmm, I understand the motivation, but quite a few of the `PipRequirement` fields will not be used so I'm not sure from architectural POV it's the right thing to use inheritance instead of some common interfacing to make this cleaner. The fact that this class directly doesn't use fields/methods from PipRequirment doesn't mean a caller somewhere else in the code could not make it and even get seemingly okay results (still wrong), so I guess I'd probably prefer more encapsulation and separation of concerns.

hermeto/core/package_managers/python/pip/main.py
```diff
@@ -716,3 +790,57 @@ def maybe_replace(requirement: PipRequirement) -> PipRequirement | None:
...
     )
+
+
+def _replace_pylock_requirements(lockfile_path: RootedPath) -> ProjectFile | None:
+    """Generate a modified pylock.toml with local file paths for external dependencies.
+
+    Rewrite URL/VCS dependencies to point to local pre-fetched files (templated).
+    Index (PyPI) dependencies are left unchanged.
+    """
+    lockfile = PyLockfileV1(lockfile_path)
```
eskultety, line 801: Looks like we're re-parsing the lockfile again, the first time it happened in `_resolve_pip` in `_download_from_lockfiles`. Was this approach perhaps taken from `_replace_external_requirements` for consistency reasons?

hermeto/core/package_managers/python/pip/main.py
```diff
@@ -430,25 +439,33 @@ def _resolve_and_download_pypi_packages(
...
+    :param lockfile: a requirements.txt file or pylock lockfile
     :param binary_filters: process wheels?
     :return: list of PipPackage instances for each downloaded package
     """
-    options: dict[str, Any] = process_requirements_options(requirements_file.options)
-    trusted_hosts = set(options["trusted_hosts"])
     processed: list[PipPackage] = []
 
+    if isinstance(lockfile, PyLockfileV1):
+        options: dict[str, Any] = defaultdict(lambda: None)
```
eskultety, line 456: My personal feeling with the `defaultdict` used here is that it's quite opaque, because you later on we do `if options["require_hashes"]:` which returns `None` on any missing keys. Was the choice of defaultdict deliberate to satisfy the existing checks to keep them working (maybe kinda by accident) or is this an unintentional side-effect?

eskultety (commented):
Not digging into tests or going even deeper with some of the logic since the existing comments may affect the overall implementation.

hermeto/core/package_managers/python/pip/main.py
ST3V1K: I agree that the design deserves a more detailed description of this. We don't need to invent the custom field format. I believe that this is mostly just an implementation problem rather than really some design choices. I think we can get all the information from just using the existing fields (unless merging the 3 new fields into 1 is really wanted), and theoretically even drop `packaging_tool` and default to `{"type": "pip"}`. 
We could allow the users to do e.g. this: 
```json
{
    "type": "pip",
    "requirements_files": ["requirements-downstream.txt"],
    "requirements_build_files": ["requirements-build.txt"],
    "lockfile": "pylock.toml"
}
```

And it would treat the `lockfile` and `lockfile_extras` as pylock only files.

We would need to make sure to merge the found dependencies correctly, since it could technically resolve two of the same dependencies just from different sources (e.g. sdist and wheels) and it would count as two different ones.

hermeto/core/models/input.py
ST3V1K: Sorry, it's my oversight. Since I was setting `_value_` I even set `_name_` because it did no harm. But after testing, it did almost nothing and is unnecessary.

hermeto/core/models/input.py
ST3V1K: I'd say it's a semi-Pythonic way of doing it. The other way of doing something similar, I could think of, was with constant class instances instead of enums. But I think I would have to do something more to make it work with the CLI parameters in the input JSON.

hermeto/core/package_managers/python/pip/pylock.py
ST3V1K: I was thinking of it as a later refactoring, but it would indeed make more sense to do it right away.

hermeto/core/package_managers/python/pip/main.py
ST3V1K: It was intentional; I could have set it to default values, but I liked `defaultdict` more. Both used options by this code path (`require_hashes` and `index_url`) resolve to `False` as boolean, so it should work as intended.
```py
{
    "require_hashes": False,
    "trusted_hosts": [],
    "index_url": None,
}
```

hermeto/core/package_managers/python/pip/main.py
ST3V1K: Yes, most of it was taken from the `_replace_external_requirements`. The choice was made to make the integration simple.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,312 @@
...
+
+class _SourceModel(pydantic.BaseModel, extra="ignore"):
+    url: str | None = None
+    hashes: dict[str, str] = {}
+    subdirectory: str | None = None
+
+    @pydantic.model_validator(mode="before")
+    @classmethod
+    def _reject_local_path(cls, data: Any) -> Any:
+        if isinstance(data, dict) and "path" in data:
```
a-ovchinnikov, line 33: https://github.com/hermetoproject/hermeto/commit/2fa6e84d765e7b5a109b23f3c1ed33ee8cf75984:

Do you expect `path` to be an extra field?

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+class _PackageModel(pydantic.BaseModel, extra="ignore"):
+    name: str
+    version: str | None = None
+    dependencies: list[str] | None = None
+    index: str | None = None
+    vcs: _VCSSourceModel | None = None
+    archive: _ArchiveSourceModel | None = None
+    directory: Any = None
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/2fa6e84d765e7b5a109b23f3c1ed33ee8cf75984:

`Any` is as broad as a type could be. Is `lambda: 42` a valid value for `directory`?

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,312 @@
...
+from packaging.version import Version
+from tomlkit.exceptions import ParseError
+from typing_extensions import Self
+
+from hermeto.core.errors import InvalidLockfileFormat, PackageRejected
+from hermeto.core.package_managers.python.pip.requirements import PipRequirement
+from hermeto.core.rooted_path import RootedPath
+
+
+class _SourceModel(pydantic.BaseModel, extra="ignore"):
```
a-ovchinnikov, line 23: https://github.com/hermetoproject/hermeto/commit/2fa6e84d765e7b5a109b23f3c1ed33ee8cf75984:

Out of curiosity, why did you start every class name with an underscore?

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,312 @@
...
+class _PackageModel(pydantic.BaseModel, extra="ignore"):
+    name: str
+    version: str | None = None
+    dependencies: list[str] | None = None
+    index: str | None = None
+    vcs: _VCSSourceModel | None = None
+    archive: _ArchiveSourceModel | None = None
+    directory: Any = None
+    sdist: _SourceModel | None = None
+    wheels: list[_SourceModel] | None = None
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/2fa6e84d765e7b5a109b23f3c1ed33ee8cf75984:

These type annotations tells a reader that there is no difference between a wheel and a sdist, but if there is no difference what's the point of having two identical entities? A similar problem was discussed and resolved in https://github.com/hermetoproject/hermeto/pull/1631/changes/448badd1a015680ad66b59bc11f2b28a29b7b219#diff-2385edf10ca0a4b8dbcdb30faded6eae06e7b740b430215e9d7d40e078b56ac2R225-R226, PTAL.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,312 @@
...
+
+    @pydantic.model_validator(mode="after")
+    def _reject_directory(self) -> Self:
+        if self.directory is not None:
+            raise PackageRejected(
+                f"Package '{self.name}' uses a local directory source which is not supported.",
+                solution="Replace the directory dependency.",
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/2fa6e84d765e7b5a109b23f3c1ed33ee8cf75984:

Is this really in user's hands? Most our users build third-party packages which they usually don't have any influence on, or am I missing something?

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,312 @@
...
+        for package in self.packages:
+            if isinstance(package, PyLockIndexPackage):
+                package.hashes.extend(package.wheel_hashes)
+
+    @cached_property
+    def packages(self) -> list["PyLockPackage"]:
+        """Get the packages from the lockfile."""
+        return [
+            PyLockPackage.from_package_data(package_data, self.file_path.path)
+            for package_data in self._data.get("packages", [])
```
a-ovchinnikov, line 155: https://github.com/hermetoproject/hermeto/commit/2fa6e84d765e7b5a109b23f3c1ed33ee8cf75984:

Nitpick, won't block on this, however: list comprehensions are meant to make code easier to read, a multiline list comprehension fails at that (especially when `black` rules are enforced with no questions asked). Please consider either adding enough helpers/aliases to fit it on a single line or just use regular `for` loop.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,312 @@
...
+    def __init__(self) -> None:
+        """Initialize a PyLockPackage object."""
+        super().__init__()
+        self.name: str = ""
+        self.version: str | None = None
+        self.dependencies: list[str] | None = None
+
+    @classmethod
+    def from_package_data(cls, data: dict[str, Any], path: Path) -> "PyLockPackage":
+        """Parse a package from pylock data."""
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/2fa6e84d765e7b5a109b23f3c1ed33ee8cf75984:

A few lines below:

```python3
    @classmethod
    @abstractmethod
    def from_data(cls, data: dict[str, Any], path: Path) -> Self:
        """Parse a package from pylock data."""
```
Same docstring, same arity, mostly same signature. Could you please elaborate what is the difference?

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,312 @@
...
+        return package
+
+    @classmethod
+    @abstractmethod
+    def from_data(cls, data: dict[str, Any], path: Path) -> Self:
+        """Parse a package from pylock data."""
+
+
+class PyLockVCSPackage(PyLockPackage):
+    """A VCS package from pylock data."""
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/2fa6e84d765e7b5a109b23f3c1ed33ee8cf75984:

WDYT about "A VCS package as defined in pylock."?

a-ovchinnikov (commented):
Please find some comments and questions inline.

hermeto/core/package_managers/python/pip/pylock.py
ST3V1K: No, `path` is a spec field, not an extra. The validator runs on the nested source dict (e.g. `[packages.vcs]`) during pydantic validation; if it contains `path`, we reject. See [packages.vcs.path](https://packaging.python.org/en/latest/specifications/pylock-toml/#packages-vcs-path).

hermeto/core/package_managers/python/pip/pylock.py
ST3V1K: Changed from `Any` to `dict[str, Any] | None`. It only checks presence and rejects. It doesn’t read fields inside the table.

hermeto/core/package_managers/python/pip/pylock.py
ST3V1K: `_` was mainly to avoid docstrings on internal pydantic models. But since I've added docstrings on `_SdistModel` and `_WheelModel`, because they needed them (or `pass`), I also matched the same style on the rest.

hermeto/core/package_managers/python/pip/pylock.py
ST3V1K: Added `_SdistModel` and `_WheelModel` which have no custom field

hermeto/core/package_managers/python/pip/pylock.py
ST3V1K: It can be in the user's hands if he included `directory` package in the project's lockfile. But since it's really uncommon, to my knowledge, to see `directory` packages, I just updated the solution a little bit

hermeto/core/package_managers/python/pip/pylock.py
ST3V1K: I’d keep it. It’s a simple map and the line length is mostly `from_package_data`, not the comprehension.
Maybe adding a helper to extract `package_data` instead of `self._data.get("packages", [])` could make it a little easier. But that still did not make it one line.

hermeto/core/package_managers/python/pip/pylock.py
ST3V1K: Renamed `from_data` to `_parse_source_fields` and split the docstrings: `from_package_data` validates + common fields, `_parse_source_fields` is kind-specific only.

hermeto/core/package_managers/python/pip/pylock.py
ST3V1K: Done for all 4 of the classes

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+from packaging.version import Version
+from tomlkit.exceptions import ParseError
+from typing_extensions import Self
+
+from hermeto.core.errors import InvalidLockfileFormat, PackageRejected
+from hermeto.core.package_managers.python.pip.requirements import PipRequirement
+from hermeto.core.rooted_path import RootedPath
+
+
+class _SourceModel(pydantic.BaseModel, extra="ignore"):
```
eskultety, line 23: I don't think it's we should host all these pylock-related models inside the `pip` backend since pylock is tool-agnostic and so the abstract model representation of `pylock.toml` should be common to any python backed we'll adopt in the future.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+    def _reject_local_path(cls, data: Any) -> Any:
+        if isinstance(data, dict) and "path" in data:
+            raise PackageRejected(
+                "Package uses a local path source which is not supported.",
+                solution="If you maintain this lockfile, use a remote URL instead of a local path.",
+            )
+        return data
+
+
+class _VCSSourceModel(_SourceModel):
```
eskultety, line 41: https://github.com/hermetoproject/hermeto/pull/1589/changes/b011ade68c8925cb5e4b575cca435e57a5df1e30

There's IMO too much going on in this commit - too many models introduced at the same time, each one providing both spec and validators that all comes together in `PyLockPackage` - that's too much for a reviewer to consume in one go I believe to draw a reasonable picture in their head. Please introduce the models gradually, only what's needed to make the change logically enclosed in order not to divert focus to other parts/models.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+        return self
+
+
+class _LockfileModel(pydantic.BaseModel):
+    """The top-level structure of a pylock lockfile."""
+
+    model_config = pydantic.ConfigDict(populate_by_name=True, extra="ignore")
+
+    lock_version: str = pydantic.Field(alias="lock-version")
+    packages: list[Any]
```
eskultety, line 100: Why isn't this modeled as `packages: list[PyLockPackage]` ? Using `Any` as a type is  IMO as good as omitting the type altogether IMO and should be limited to particular cases like `**kwargs: Any` where it makes sense.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+                f"Package '{self.name}' uses a local directory source which is not supported.",
+                solution="If you maintain this lockfile, use a fetchable source instead of a directory dependency.",
+            )
+        return self
+
+
+class _LockfileModel(pydantic.BaseModel):
+    """The top-level structure of a pylock lockfile."""
+
+    model_config = pydantic.ConfigDict(populate_by_name=True, extra="ignore")
```
eskultety, line 97: I'm not sure `extra="ignore"` is a safe choice given the validator which checks the version based on a specifier - `~=1.0` which means any potential new fields (which is a backwards compatible change) would be immediately rejected by our model.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+
+    lock_version: str = pydantic.Field(alias="lock-version")
+    packages: list[Any]
+
+    @pydantic.field_validator("lock_version")
+    @classmethod
+    def _validate_version(cls, v: str) -> str:
+        if Version(v) not in Specifier("~=1.0"):
```
eskultety, line 105: Does pylock.toml follow strict semver schema? I don't think so. We don't have a guarantee that 1.2 will be compatible with 1.0, although it would be kinda stupid if it weren't, so I think it's a reasonable assumption. Given our impl. is based on 1.0 can we at least emit a warning if we're parsing a version >1.0 so at least the user knows we're using an older impl. and they're not getting the latest & greatest?

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+            raise ValueError(f"unsupported lockfile version: {v}")
+        return v
+
+
+class PipLockfile(ABC):
+    """Base class for a pip lock-style files."""
+
```
eskultety, line 112: What's the value of this? It's an empty container that neither enforces any interface nor does it provide common implementation.
It's also not adopted by the pip backend in any way. Could we turn it into protocol that both Pylockfile and `PipRequirementsFile` satisfy? That in turn will mean that the shared downloader logic `_download_dependencies` might stop enforcing a particular type for the input? Just an idea.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+        with open(self.file_path) as f:
+            try:
+                data = tomlkit.load(f)
+            except ParseError:
+                raise InvalidLockfileFormat(
+                    lockfile_path=self.file_path.path,
+                    err_details="Invalid TOML syntax.",
+                    solution="Regenerate the lockfile.",
+                )
+            return data
```
eskultety, line 142: Loading could happen outside this class and we could just instantiate & populate the model with it.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+
+    def merge_wheel_hashes(self) -> None:
+        """Merge the wheel hashes into the sdist hashes."""
+        for package in self.packages:
+            if isinstance(package, PyLockIndexPackage):
+                package.hashes.extend(package.wheel_hashes)
```
eskultety, line 148: Why do we need this merging? It's not obvious to me - a dedicated commit explaining why would IMO be justified.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+class PyLockfileV1(PipLockfile):
+    """Parser for pylock lockfile version 1.x."""
+
+    def __init__(self, path: RootedPath) -> None:
+        """Initialize a PipLockfile object."""
+        self.file_path = path
+        self._data = self._load_data()
+
+        try:
+            _LockfileModel.model_validate(self._data)
```
eskultety, line 123: I feel like this kind of interfacing is suspicious - you defined this model a few lines above just the validate top-level metadata and version only to discard the data afterwards which also means we parsed the data twice (once in `_load_data` and once through the model validation).

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+            )
+
+        return cls(
+            url=source.get("url"),
+            hashes=[f"{alg}:{dig}" for alg, dig in source.get("hashes", {}).items()],
+            subdirectory=source.get("subdirectory"),
+        )
+
+
+class PyLockPackage(PipRequirement, ABC):
```
eskultety, line 203: Isn't inheriting from PipRequirement here a bit too much? We're getting quite a few fields that are irrelevant for us here.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+    @classmethod
+    def from_package_data(cls, data: dict[str, Any], path: Path) -> "PyLockPackage":
+        """Build a ``PyLockPackage`` from a raw ``[[packages]]`` entry.
+
+        Validates ``data``, dispatches to the concrete package class, and sets
+        fields shared by every package kind. Kind-specific fields are parsed by
+        ``_parse_source_fields`` on the concrete class.
+        """
+        try:
+            _PackageModel.model_validate(data)
```
eskultety, line 225: Similarly to the lockfile model and its validator a few lines above - you went the lenght of defining the model's fields accurately only to discard the model instance completely here. I don't think that's a desirable logic flow.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+        """Parse ``[packages.vcs]`` fields."""
+        vcs = data["vcs"]
+        artifact = PyLockArtifact.from_source(vcs, data["name"], path)
+
+        package = cls()
+        package.vcs_type = vcs["type"]
+        package.commit_id = vcs["commit-id"]
+        package._url = f"{package.vcs_type}+{artifact.url}@{package.commit_id}"
+        package.subdirectory = artifact.subdirectory
+        package.download_line = f"{data['name']} @ {package._url}"
```
eskultety, line 281: https://github.com/hermetoproject/hermeto/pull/1589/changes/b011ade68c8925cb5e4b575cca435e57a5df1e30
Instead of this raw parsing (repeating pattern in this commit), could we offload this to pydantic instead and if you need some computed fields, do it in a `__post_init__` constructor ? IOW I'd envision that calling a `from_file` on the top-level lockfile would return all abstract representations parsed in one go by pydantic.

hermeto/core/package_managers/python/pip/pylock.py
```diff
@@ -0,0 +1,343 @@
...
+        hashes: list[str] | None = None,
+        subdirectory: str | None = None,
+    ) -> None:
+        """Initialize a PyLockArtifact."""
+        self.url = url
+        self.hashes = hashes or []
+        self.subdirectory = subdirectory
+
+    @classmethod
+    def from_source(cls, source: dict[str, Any], pkg_name: str, lockfile_path: Path) -> Self:
```
eskultety, line 184: This seems to be a double validation of what `_SourceModel` was supposed to have already done, am I mistaken?

hermeto/core/package_managers/python/pip/pylock.py
eskultety: That said, since I'm not particularly convinced about the modeling being done here I think pylock-related models should not inherit anything from Pip. The common layer with the pip backend IIUC is the download plumbing where you need a `PipRequirement` - can we have an adapter where necessary to just do `to_requirement` and call it a day?

hermeto/core/models/input.py
eskultety: What if we did something like this instead only keeping the enum properties as you have above:
```python
# models/input.py
@dataclass(frozen=True)
class _PackagingFormat:
    default_file: str
    default_build_file: str
    parse: Callable[[RootedPath, bool], ParsedLockfile]

# pip/main.py OR pip/lockfile_formats.py
_FORMATS: dict[PipPackagingTool, _PackagingFormat] = {
    PipPackagingTool.REQUIREMENTS: _PackagingFormat(
        "requirements.txt", "requirements-build.txt",
        lambda p: PipRequirementsFile(p),
    ),
    PipPackagingTool.PYLOCK: _PackagingFormat(
        "pylock.toml", "pylock.build.toml",
        lambda p: Pylock.from_file(p),
    ),
}
```

hermeto/core/package_managers/python/pip/main.py
```diff
@@ -875,16 +813,25 @@ def _download_from_lockfiles(
...
-        if not lockfile.path.exists():
+    for lockfile_path in lockfiles:
+        if not lockfile_path.path.exists():
             raise LockfileNotFound(
-                files=lockfile.path,
+                files=lockfile_path.path,
                 solution="Please check that you have specified correct lockfile paths",
             )
+
+        lockfile: PipRequirementsFile | PyLockfileV1
```
eskultety, line 823: This here justifies a `Protocol` or some common interface as I suggested in another comment.

eskultety (commented):
This round of review is primarily about modelling this as that will have some further implications I think. Once we find common ground on the modelling aspect then it makes sense to look deeply at the actual implementation.

hermeto/core/package_managers/python/pip/pylock.py
a-ovchinnikov: My comment was more about whether a user who builds packages, but does not maintain them can actually do anything about it. I generally doubt the idea of providing solutions this way since for non-maintainers most of these solutions are not applicable. However this discussion is out of scope for this PR, so resolving.

hermeto/core/package_managers/python/pip/pylock.py
a-ovchinnikov: I was thinking about something along these lines: 

```python3
plp = partial(PyLockPackage.from_package_data, path=self.file_path.path)
return [plp(package_data) for package_data in self._data.get("packages", [])]
```
but that is not a blocker, resolving.

PR #1602, opened 2026-06-04
fix(pip): reject VCS deps in hash-locked requirements with clear error

### LINKED ISSUE #1427: fix: misleading error message when git dependency is used with hashed requirements
## Summary

While going through `requirements.py` to understand the pip module, I noticed that when a git dependency is used alongsidehashed PyPI dependencies, hermeto raises a misleading error telling the user to add a `--hash` to a git dependency — which is impossible since git dependencies do not support pip's `--hash` option according to the docs.

## Reproducing
```
# requirements.txt
requests==2.31.0 --hash=sha256:942c5a758f98d790eaed1a29cb6eefc7ffb0d1cf7af05c3d2791656dbd6ad1e1
dockerfile-parse @ git+https://github.com/containerbuildsystem/dockerfile-parse@b6230230987950cfb16d8858c6f9a9642f4d0952
```
```bash
hermeto fetch-deps pip
```

Current error:
```
Error: MissingChecksum: Checksum is missing
  Hash is required, dependency does not specify any: dockerfile-parse @ git+https://github.com/containerbuildsystem/dockerfile-parse@b6230230987950cfb16d8858c6f9a9642f4d0952Please specify the expected hashes for all dependencies
```

## Problem

The error tells the user to specify a hash for the git dependency which is impossible. The docs clearly state .

## Expected
```
Error: MissingChecksum: 'dockerfile-parse @ git+https://github.com/containerbuildsystem/dockerfile-parse@b6230230987950cfb16d8858c6f9a9642f4d0952' is missing mandatory integrity checksum.
  Git dependencies do not support pip's --hash option.
  Please use an https URL instead, for example:
    dockerfile-parse @ https://github.com/.../2.0.0.tar.gz \
        --hash=sha256:...
```

## Fix

Detect VCS requirements specifically in `validate_requirements_hashes()` 
and raise a clear error pointing to the https URL alternative.

Also fixed two missing `\n` separators in `InvalidChecksum` and 
`MissingChecksum` solution strings in the same file.

ST3V1K:
Hashed requirements.txt files trigger mandatory checksum validation for every line. Git and other VCS dependencies cannot carry pip --hash entries, so users who mix the two hit MissingChecksum with advice to add hashes they can never satisfy. That sends people down the wrong fix path instead of switching to a hashable archive URL.

Fixes https://github.com/hermetoproject/hermeto/issues/1427

### DIFF
--- a/hermeto/core/package_managers/pip/requirements.py
+++ b/hermeto/core/package_managers/pip/requirements.py
@@ -658,24 +658,35 @@ def validate_requirements_hashes(requirements: list[PipRequirement], require_has
 
     :param list[PipRequirement] requirements: All requirements from a file
     :param bool require_hashes: True if hashes are required for all requirements
+    :raise UnsupportedFeature: If any VCS requirement is used in a hash-locked requirements file
     :raise MissingChecksum: If any hashes are missing
     :raise InvalidChecksum: If hashes have invalid format
     """
     for req in requirements:
         hashes = req.hashes
 
-        if require_hashes and not hashes:
-            # We shouldn't get here, but it's a definite error if we do.
+        if require_hashes:
             # VCS reqs *cannot* be hashed, so we'll always hit
             # this for any VCS req in a 'requirements.txt' which has *any* hash.
             # For URL requirements, having a hash is required to pass *basic* validation.
-            raise MissingChecksum(
-                None,
-                solution=(
-                    f"Hash is required, dependency does not specify any: {req.download_line}.\n"
-                    "Please specify the expected hashes for all dependencies"
-                ),
-            )
+            if req.kind == "vcs":
+                raise UnsupportedFeature(
+                    f"VCS requirements are not supported in hash-locked requirements files: {req.download_line}",
+                    solution=(
+                        "VCS requirements do not support pip's --hash option.\n"
+                        "Consider using a hashable URL instead, for example:\n"
+                        "  <name> @ https://github.com/.../1.0.0.tar.gz \\\n"
+                        "    --hash=sha256:..."
+                    ),
+                )
+            elif not hashes:
+                raise MissingChecksum(
+                    None,
+                    solution=(
+                        f"Hash is required, dependency does not specify any: {req.download_line}.\n"
+                        "Please specify the expected hashes for all dependencies"
+                    ),
+                )
 
         for hash_spec in hashes:
             _, _, digest = hash_spec.partition(":")
--- a/tests/unit/package_managers/pip/test_main.py
+++ b/tests/unit/package_managers/pip/test_main.py
@@ -509,8 +509,8 @@ class TestDownload:
                     ("bar", "vcs", {}),
                 ],
                 [],
-                MissingChecksum,
-                id="vcs_local_hash_triggers_missing",
+                UnsupportedFeature,
+                id="vcs_local_hash_triggers_unsupported",
             ),
             pytest.param(
                 [
@@ -527,7 +527,7 @@ class TestDownload:
                     ("bar", "vcs", {}),
                 ],
                 ["--require-hashes"],
-                MissingChecksum,
+                UnsupportedFeature,
                 id="vcs_global_require_hashes",
             ),
             pytest.param(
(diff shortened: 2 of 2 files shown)


hermeto/core/package_managers/pip/requirements.py
```diff
@@ -664,18 +664,29 @@ def validate_requirements_hashes(requirements: list[PipRequirement], require_has
...
+            if req.kind == "vcs":
+                raise MissingChecksum(
+                    req.download_line,
+                    solution=(
+                        "VCS requirements do not support pip's --hash option.\n"
+                        "Consider using a hashable URL instead, for example:\n"
+                        "  <name> @ https://github.com/.../1.0.0.tar.gz \\\n"
+                        "    --hash=sha256:..."
+                    ),
+                )
```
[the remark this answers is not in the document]
ST3V1K: Fair.

hermeto/core/package_managers/pip/requirements.py
```diff
@@ -658,24 +658,36 @@ def validate_requirements_hashes(requirements: list[PipRequirement], require_has
 
     :param list[PipRequirement] requirements: All requirements from a file
```
eskultety, line 659: **Commit message:**
_Assisted-by: Harshit Bansal_

Probably `Reported-by` since the original PR was closed, but the issue remained.

hermeto/core/package_managers/pip/requirements.py
```diff
@@ -658,24 +658,36 @@ def validate_requirements_hashes(requirements: list[PipRequirement], require_has
...
+    :raise UnsupportedFeature: If any VCS requirement is used in a hash-locked requirements file
     :raise MissingChecksum: If any hashes are missing
     :raise InvalidChecksum: If hashes have invalid format
     """
     for req in requirements:
         hashes = req.hashes
 
-        if require_hashes and not hashes:
+        if require_hashes:
             # We shouldn't get here, but it's a definite error if we do.
```
eskultety: This commentary no longer applies, because we DO get here now, quite intentionally since this is now an exploded check.

tests/unit/package_managers/pip/test_main.py
```diff
@@ -509,7 +509,7 @@ def test_download_rejects_invalid_dependency(
                     ("bar", "vcs", {}),
                 ],
                 [],
-                MissingChecksum,
+                UnsupportedFeature,
```
eskultety, line 512: `vcs_local_hash_triggers_missing` 
Is ^this test ID still accurate given a different code branch is now being exercised?

hermeto/core/package_managers/pip/requirements.py
ST3V1K: I used part of the code from the closed PR, and it helped me to identify where I should apply the fix. I would keep *Assisted-by*

eskultety (approved):
A couple of nitpicks, easy to address. The PR is straightforward, ACK.

hermeto/core/package_managers/pip/requirements.py
eskultety: In that case ---> `Signed-off-by` that's a standardized trailer. `Assisted-by` is an artificial construct that only came to life with AI.

ST3V1K: Done

hermeto/core/package_managers/pip/requirements.py
ST3V1K: Removed the first comment line -- We shouldn't get here, but it's a definite error if we do.

tests/unit/package_managers/pip/test_main.py
ST3V1K: Changed 
vcs_local_hash_triggers_missing -> vcs_local_hash_triggers_unsupported
vcs_malformed_hash -> vcs_malformed_hash_triggers_unsupported

hermeto/core/package_managers/pip/requirements.py
eskultety: ACK.

slimreaper35 (approved):

PR #1422, opened 2026-03-17
feat: indicate experimental features usage globally and per-component

Gitjay11:
Resolves #614 

### Description
This PR addresses the remaining requirement from #614 (reopened) to definitively state when experimental features were used during generation, and exactly which components were produced by them.

Previously, experimental backend usage mapped to document annotations but lacked an explicit global boolean flag and didn't directly tag the origin of individual components directly via properties.

### Changes
1.  Added `PROP_EXPERIMENTAL` (`hermeto:experimental`) to the [PropertyEnum](cci:2://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/hermeto/core/models/property_semantics.py:14:0-29:25) validation registry in [property_semantics.py](cci:7://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/hermeto/core/models/property_semantics.py:0:0-0:0).
2.  Added support for [properties](cci:1://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/hermeto/core/models/property_semantics.py:100:4-130:61) on the [Metadata](cci:2://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/hermeto/core/models/sbom.py:148:0-152:69) model in [sbom.py](cci:7://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/hermeto/core/models/sbom.py:0:0-0:0), including proper merging logic in `Sbom.__add__`.
3.  Updated [SPDX](cci:2://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/hermeto/core/models/sbom.py:586:0-804:9) conversion logic in [sbom.py](cci:7://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/hermeto/core/models/sbom.py:0:0-0:0) so that global `metadata.properties` correctly propagate to and parse from `SPDXPackageAnnotations` on the DocumentRoot package.
4.  Updated `RequestOutput.generate_sbom` in [output.py](cci:7://file:///c:/Users/Ajay%20Rajera/Desktop/hermeto/hermeto/core/models/output.py:0:0-0:0) to append `PROP_EXPERIMENTAL = "true"` to the global `Sbom.metadata.properties` when experimental backend annotations are present.
[...]

PR #1611, opened 2026-06-09
feat(sbom): flag experimental backend usage in SBOM output

### LINKED ISSUE #614: Define a clear process for experimental features
The current process to start the development of a new package manager starts with the proposal and acceptance of a thorough design document. The actual implementation can only begin after the design document is reviewed and accepted.

This brings a few problems:
- It is very demanding for contributors that are external to the core Cachi2 team
- It forces us to make decisions ahead of time, without the possibility of testing in a real enviroment how the new feature will behave
- When a PoC is written to accompany the design, most of its code is discarded once the real implementation starts 
- It is arguably not the way most people tend to approach writing a technical solution, as it hinders the freedom to experiment and iteratively build upon the code that is written
- The feature is usually only tested in real scenarios once it is completely delivered

There needs to be a formalized process in which a feature can be iteratively added to Cachi2 under an experimental flag. This would allow contributors to start contributing small pieces of working code, and to test them in a real scenario. It would lower the load on maintainer to fully understand a package manager's ecosystem, and to properly review the design doc. It can potentially also bring a better sense of accomplishment and motivate contributors and uses to keep improving on the new feature.

We still want to have a design doc that clearly states what decisions were taken, what options were discarded and what were the reasons for it. But this design does not need to be ready nor approved before the feature is marked as production-ready.

**Acceptance Criteria**
- This process is agreed upon and formalized in a contributors documentation
- There is a clear interface to enable experimental features, so that users are aware of what they are using
- The resulting SBOM clearly states that experimental features were used, and which components (if any) were produced by them

Harsh-2005d:
Resolves #614

## Description
Rework on #1422
This PR addresses the remaining requirement from #614  (reopened) to definitively state when experimental features were used during generation, and exactly which components were produced by them.

## Changes

- `property_semantics.py`: add `PROP_EXPERIMENTAL`  to `PropertyEnum` and pass it through `PropertySet` 
- `sbom.py`: add `properties` field to `Metadata`; fix `Sbom.__add__` to merge metadata properties via `PropertySet.merge`; encode metadata properties onto the SPDX DocumentRoot package for sbom conversions.
- `output.py`: update `generate_sbom` to detect experimental backend annotations and tag both `sbom.metadata.properties` and the affected individual components.
- `test_output.py`: add `TestRequestOutputExperimental`, covering experimental and non-experimental annotation text.

### DIFF
--- a/hermeto/core/models/output.py
+++ b/hermeto/core/models/output.py
@@ -8,7 +8,14 @@ from typing import Any
 import pydantic
 
 from hermeto.core.errors import BaseError
-from hermeto.core.models.sbom import Annotation, Component, Sbom, merge_component_properties
+from hermeto.core.models.property_semantics import Property, PropertyEnum
+from hermeto.core.models.sbom import (
+    BACKEND_ANNOTATION_PREFIX,
+    Annotation,
+    Component,
+    Sbom,
+    merge_component_properties,
+)
 from hermeto.core.models.validators import unique_sorted
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


slimreaper35:
I don't understand why we need (another) experimental property if we use annotations for the same purpose.

Harsh-2005d:
> I don't understand why we need (another) experimental property if we use annotations for the same purpose.

@slimreaper35 I thought having a separate boolean property for experimental backend would be useful, but I think a well-documented annotation format serves the purpose, while keeping the implementation simple. I can drop it if u want.

slimreaper35:
> I can drop it if u want.

It's not about what I want 😄. I just don't see the value of another property. The issue has been created a long time ago (in the meantime closed and reopened again) when the project was in a completely different state.

Also, the current approach with annotations is already used internally to check experimental package managers.

Harsh-2005d:
Makes sense, the existing approach works fine with annotation and it is documented. I'll close this PR as there is no gap left as far as i see. we can also close the #614 if @eskultety agrees.

PR #1612, opened 2026-06-09
fix(generate-env): resolve path vars before dependent expansion

### LINKED ISSUE #1424: generate-env resolves dependent variables using pre-resolved values, leading to incorrect paths

ST3V1K:
Dependent variables like GOPROXY referenced stale relative paths when kind="path" targets were expanded after mappings were built, producing invalid file:// URLs at build time.

Pre-resolve legacy path variables into the substitution mappings before expanding dependents.

Fixes https://github.com/hermetoproject/hermeto/issues/1424

### DIFF
--- a/hermeto/core/extras/envfile.py
+++ b/hermeto/core/extras/envfile.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-only
 import json
 import shlex
+from copy import deepcopy
 from enum import Enum
 from pathlib import Path
 
@@ -54,13 +55,7 @@ def generate_envfile(build_config: BuildConfig, fmt: EnvFormat, relative_to_path
     - env: export GOCACHE=/path/to/output-dir/deps/gomod
            export ...
     """
-    # pass all variables as placeholder mappings to env var template value resolution
-    mappings = {var.name: var.value for var in build_config.environment_variables}
-    mappings["output_dir"] = relative_to_path.as_posix()
-    env_vars = [
-        (env_var.name, env_var.resolve_value(mappings))
-        for env_var in build_config.environment_variables
-    ]
+    env_vars = _resolve_env_vars(build_config, relative_to_path)
     if fmt == EnvFormat.json:
         content = json.dumps([{"name": name, "value": value} for name, value in env_vars])
     else:
@@ -68,3 +63,22 @@ def generate_envfile(build_config: BuildConfig, fmt: EnvFormat, relative_to_path
             f"export {shlex.quote(name)}={shlex.quote(value)}" for name, value in env_vars
         )
     return content
+
+
+def _resolve_env_vars(build_config: BuildConfig, relative_to_path: Path) -> list[tuple[str, str]]:
+    """Resolve the environment variables for the given build config and path."""
+    # pass all variables as placeholder mappings to env var template value resolution
+    mappings = {var.name: var.value for var in build_config.environment_variables}
+    mappings["output_dir"] = relative_to_path.as_posix()
+
+    # Legacy kind="path" variables must be resolved into mappings before dependents
+    # (e.g. GOPROXY's file://${GOMODCACHE}/...) are expanded; otherwise substitution
+    # uses the unresolved relative path from the initial mappings dict.
+    for env_var in build_config.environment_variables:
+        if env_var.kind == "path":
+            mappings[env_var.name] = deepcopy(env_var).resolve_value(mappings)
+
+    return [
+        (env_var.name, env_var.resolve_value(mappings))
+        for env_var in build_config.environment_variables
+    ]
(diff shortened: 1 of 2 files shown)


hermeto/core/extras/envfile.py
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: GPL-3.0-only
 import json
 import shlex
+from copy import deepcopy
 from enum import Enum
```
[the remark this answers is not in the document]
a-ovchinnikov: It looks like our AI overlords have a really hard time figuring out imports.

hermeto/core/extras/envfile.py
```diff
@@ -54,17 +55,30 @@ def generate_envfile(build_config: BuildConfig, fmt: EnvFormat, relative_to_path
...
         content = json.dumps([{"name": name, "value": value} for name, value in env_vars])
     else:
         content = "\n".join(
             f"export {shlex.quote(name)}={shlex.quote(value)}" for name, value in env_vars
         )
     return content
+
+
+def _resolve_env_vars(build_config: BuildConfig, relative_to_path: Path) -> list[tuple[str, str]]:
+    """Resolve the environment variables for the given build config and path."""
```
a-ovchinnikov: It would have been nice to have some form of tests to accompany this change, maybe some doctests?

ST3V1K: Done

hermeto/core/extras/envfile.py
```diff
@@ -54,17 +55,43 @@ def generate_envfile(build_config: BuildConfig, fmt: EnvFormat, relative_to_path
...
+    placeholders such as ``${GOMODCACHE}`` are expanded against resolved mappings::
+
+    >>> config = BuildConfig(
+    ...     environment_variables=[
+    ...         {"name": "GOMODCACHE", "value": "deps/gomod/pkg/mod", "kind": "path"},
+    ...         {"name": "GOPROXY", "value": "file://${GOMODCACHE}/cache/download"},
+    ...     ],
+    ... )
+    >>> _resolve_env_vars(config, Path("/tmp/output"))
+    [('GOMODCACHE', '/tmp/output/deps/gomod/pkg/mod'), ('GOPROXY', 'file:///tmp/output/deps/gomod/pkg/mod/cache/download')]
```
slimreaper35: If we have the same unit test in the `test_envfile.py`, this seems unnecessary.

slimreaper35 (approved):

a-ovchinnikov (approved):

hermeto/core/extras/envfile.py
ST3V1K: Removed

hermeto/core/extras/envfile.py
```diff
@@ -54,17 +55,30 @@ def generate_envfile(build_config: BuildConfig, fmt: EnvFormat, relative_to_path
...
+    # (e.g. GOPROXY's file://${GOMODCACHE}/...) are expanded; otherwise substitution
+    # uses the unresolved relative path from the initial mappings dict.
+    for env_var in build_config.environment_variables:
+        if env_var.kind == "path":
+            mappings[env_var.name] = deepcopy(env_var).resolve_value(mappings)
+
+    return [
+        (env_var.name, env_var.resolve_value(mappings))
+        for env_var in build_config.environment_variables
+    ]
```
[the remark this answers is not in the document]
eskultety: This was a valid comment btw.

PR #1615, opened 2026-06-11
[release-0.28] Update the max supported Go version to 1.26.

baldimir:
Upgrades the max supported version of Go 1.26. There shouldn't be any breaking changes with Hermeto with Go 1.26. However I am new to the Hermeto project contributions, so I apologize if I missed anything relevant and will update it when needed. 

Our build environment uses the Hermeto release stream 0.28.x. I noticed the 1.26 support is already on main branch, but as I am new to this project, I am not sure how much effort it would be to upgrade to a release done from the main branch, therefore adding the support to this branch.


mareknovotny:
can you please look at this ? @eskultety

slimreaper35:
It won't be that simple, 25 integration tests failed :)

eskultety:
Yes, there will be a lot of backporting, there were many changes to the integrations test part before we can release this.

baldimir:
Hi, thank you for checking, taking a look at the failed tests.

baldimir:
With the help of Claude Code, I updated the test data and fixed the references to some repositories based on the changes on the main branch. Let's see how many tests will be remaining to be fixed. @slimreaper35 @eskultety could you please reapprove the workflow? I ran the tests locally so want to make sure it corresponds to the results here in PR checks.

eskultety:
@baldimir this is not how LTS branches work. They are not supposed to deviate completely from upstream. Downstream is done by backporting the necessary stuff from the main branch. We'll also need to backport all dependency updates to make sure the 0.28.N image isn't vulnerable.

eskultety:
> Our build environment uses the Hermeto release stream 0.28.x. I noticed the 1.26 support is already on main branch,

@baldimir yeah, we kept around this old branch here in upstream (as an exception) due to some particular downstream needs. We appreciate the effort, but we'll need to prioritize this work from the maintenance POV and do it the right way. We'll discuss as a team and take care of this eventually, thanks.

baldimir:
Thank you very much. If needed, I can of course update the dependencies too.

baldimir:
Closing this for now after the discussions.

PR #1618, opened 2026-06-15
pnpm workspaces

slimreaper35:


hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -93,6 +93,14 @@ class PnpmPackage:
     url: str
     integrity: str | None = None  # git and URL dependencies don't have an integrity field
 
+    @property
```
eskultety, line 97: **Commit message:**
A brief one describing the need for the property would be nice.

hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -82,6 +72,17 @@ def root_dependencies(self) -> list[str]:
         return ids
 
 
+def load_yaml_file(path: Path) -> dict[str, Any]:
+    """Load a YAML file in the pnpm project."""
+    try:
+        with path.open() as f:
+            data = yaml.safe_load(f) or {}
+    except yaml.YAMLError as e:
+        raise InvalidLockfileFormat(path, err_details=str(e))
```
eskultety: We lost `solution` here, was that intentional because the YAML error is clear enough such that our solution is effectively useless?

hermeto/core/package_managers/javascript/pnpm/resolver.py
```diff
@@ -145,6 +159,48 @@ def _create_root_component(project_dir: RootedPath, vcs_qualifiers: dict[str, st
...
+
+    return components
+
+
+def _read_workspace_globs(project_dir: RootedPath) -> list[str]:
+    pnpm_workspace_path = project_dir.path / "pnpm-workspace.yaml"
+    if not pnpm_workspace_path.exists():
+        return []
+
+    pnpm_workspace = load_yaml_file(pnpm_workspace_path)
```
eskultety: In case of an error ^this returns `InvalidLockfileFormat` while `pnpm-workspace.yaml` isn't a lockfile, it's a workspace configuration file.

tests/unit/package_managers/javascript/pnpm/test_resolver.py
```diff
@@ -284,6 +285,39 @@ def test_create_workspace_components_without_package_json(
...
 
 
+@pytest.mark.parametrize(
+    "asterisks",
+    [
+        pytest.param("*", id="single_asterisk"),
+        pytest.param("**", id="double_asterisk"),
+    ],
+)
+@mock.patch("hermeto.core.package_managers.javascript.pnpm.resolver._read_workspace_globs")
```
eskultety: `_read_workspace_globs` doesn't seem to be exercised via `get_workspace_paths`

hermeto/core/package_managers/javascript/pnpm/resolver.py
```diff
@@ -168,10 +171,16 @@ def _create_workspace_components(
     workspace_paths = get_workspace_paths(workspaces_globs, project_dir)
     ensure_no_path_leads_out(workspace_paths, project_dir)
 
+    log.debug("Found %d workspace directories", len(workspace_paths))
+
     for workspace_path in workspace_paths:
         try:
             package_json = PackageJson.from_dir(workspace_path)
         except LockfileNotFound:
+            log.warning(
```
eskultety, line 177: nitpick: Hmm, you should IMO put this commit somewhere to the beginning of the series and then squash these `resolver.py` changes to commit 4/6 which introduces `_create_workspace_components`

hermeto/core/package_managers/javascript/pnpm/resolver.py
slimreaper35: yeah, that's right, should I just drop it or do you have any better solution?

hermeto/core/package_managers/javascript/pnpm/resolver.py
slimreaper35: ok

tests/unit/package_managers/javascript/pnpm/test_resolver.py
slimreaper35: well no because `get_workspace_paths` is from yarn classic module

eskultety: So why are we mocking it if `_read_workspace_globs` isn't reachable through this test?

hermeto/core/package_managers/javascript/pnpm/resolver.py
eskultety: Well, the biggest problem here is that returning `InvalidLockfileFormat` from that plain YAML parsing helper was wrong all along, `InvalidLockfileFormat` would only ever apply if the parsed data would violate the lockfile schema (not sure if we verify that). If you want to keep the helper, then make it a general util helper that can be used from anywhere and return a different, more generic error to cover the YAML parsing failure.

hermeto/core/package_managers/javascript/pnpm/resolver.py
```diff
@@ -150,6 +159,54 @@ def _create_root_component(project_dir: RootedPath, vcs_qualifiers: dict[str, st
...
+    return components
+
+
+def _read_workspace_globs(project_dir: RootedPath) -> list[str]:
+    pnpm_workspace_path = project_dir.path / "pnpm-workspace.yaml"
+    if not pnpm_workspace_path.exists():
+        return []
+
+    with pnpm_workspace_path.open() as f:
+        pnpm_workspace = yaml.safe_load(f)
```
eskultety: We shouldn't allow any unhandled exceptions to occur.

tests/unit/package_managers/javascript/pnpm/test_resolver.py
```diff
@@ -19,6 +19,7 @@
     _generate_pedigree_for,
     _generate_purl_for,
     generate_sbom_components,
+    get_workspace_paths,
```
eskultety: `get_workspace_paths` lives in the yarn classic module. You sure this is the right module for the test?

slimreaper35: fixed

hermeto/core/package_managers/javascript/pnpm/resolver.py
slimreaper35: added `PackageRejected` that is also raised for invalid .yarnc file

eskultety (approved):

slimreaper35:
Rebased

slimreaper35:
Added two more commits to incorporate integration test changes

hermeto/core/package_managers/javascript/pnpm/resolver.py
```diff
@@ -145,6 +159,59 @@ def _create_root_component(project_dir: RootedPath, vcs_qualifiers: dict[str, st
...
+            log.warning(
+                "The workspace %s does not contain a package.json file and will be ignored",
+                workspace_path,
+            )
+            continue
+
+        name = package_json.get("name")
+        version = package_json.get("version")
+        if name is None:
+            raise PackageRejected(f"Missing 'name' field in the {package_json.path}")
```
a-ovchinnikov, line 186: Non-blocking: this feels like a rather late stage for rejecting a package, would have been nice if the check could happen earlier.

a-ovchinnikov (approved):
LGTM, the observation in the code is non-blocking.

hermeto/core/package_managers/javascript/pnpm/resolver.py
slimreaper35: I don't really know where...

PR #1620, opened 2026-06-15
pnpm: build an offline store instead of rewriting lockfile resolutions

inspired-geek:
## Problem

pnpm 10.34.2 tightened resolution-shape validation: a registry-style lockfile key (`name@version`) may no longer carry a non-registry `resolution.tarball: file://...`. The hermeto pnpm backend rewrote **every** registry dependency that way in `_prepare_lockfile_for_hermetic_build`, so any hermetic build that pins pnpm >= 10.34.2 (via `packageManager` / corepack, or a Renovate bump of the pinned version) now fails at install time with:

```
ERR_PNPM_RESOLUTION_SHAPE_MISMATCH  Lockfile resolution for <pkg>@<ver> is broken
```

Tracked in https://github.com/hermetoproject/hermeto/issues/1619.

## Fix

Stop rewriting the lockfile. Instead, pre-populate a pnpm content-addressable store ("v10") from the downloaded tarballs and point pnpm at it via `NPM_CONFIG_STORE_DIR`, so `pnpm install --offline` resolves every dependency by its `integrity` hash:

- `store/v10/files/<fileSha512hex[0:2]>/<fileSha512hex[2:]>` — raw file content (CAS)
- `store/v10/index/<intHex[0:2]>/<intHex[2:64]>-<pkg>@<ver>.json` — per-package metadata, keyed by the package integrity SRI

The lockfile keeps its registry-style resolutions untouched, so it stays valid for any pnpm version, and the build remains fully offline.

### Store layout subtleties

Matching what `pnpm install --store-dir` writes is load-bearing — pnpm hardlinks the stored files into `node_modules` and reads its own index, so each deviation is a distinct offline failure we hit while validating:

- **Executable files** (`mode & 0o111`) use a `-exec` CAS suffix and are stored `0o750`; plain files use the bare hash at `0o640`. pnpm does not re-apply the index `mode` to the hardlink, so the stored file's mode is the installed file's mode. Wrong suffix → package treated as absent (`ERR_PNPM_NO_OFFLINE_TARBALL`); wrong mode → `spawn ... EACCES` for binaries (e.g. esbuild).
- **`requiresBuild`** is derived from `preinstall`/`install`/`postinstall` scripts or a `*.gyp` file, otherwise pnpm skips building allow-listed native dependencies (e.g. re2).

## Verification

- `pytest --doctest-modules hermeto tests/unit` (1444 passed) and the lint suite (`ruff check`, `ruff format --check`, `mypy` — 171 files, no issues) pass.
- Validated end-to-end in a hermetic build (1453-package project, `--network=none`): `pnpm install --offline` → `reused 1453, downloaded 0`, native modules (re2, better-sqlite3) built from source, full `pnpm build` succeeds.
- Negative control (same lockfile with `tarball: file://`) reproduces `ERR_PNPM_RESOLUTION_SHAPE_MISMATCH`.

The commit is DCO signed-off and carries the `Assisted-by: Claude` trailer.

### DIFF
--- a/hermeto/core/package_managers/javascript/pnpm/main.py
+++ b/hermeto/core/package_managers/javascript/pnpm/main.py
@@ -1,6 +1,10 @@
 # SPDX-License-Identifier: GPL-3.0-only
 import asyncio
+import base64
 import copy
+import hashlib
+import json
+import tarfile
 from pathlib import Path
 
 import aiohttp
@@ -9,7 +13,13 @@ import yaml
 from hermeto.core.checksum import ChecksumInfo, must_match_any_checksum
 from hermeto.core.config import get_config
 from hermeto.core.models.input import Request
-from hermeto.core.models.output import Annotation, Component, ProjectFile, RequestOutput
+from hermeto.core.models.output import (
+    Annotation,
+    Component,
+    EnvironmentVariable,
+    ProjectFile,
+    RequestOutput,
+)
 from hermeto.core.models.sbom import create_backend_annotation
 from hermeto.core.package_managers.javascript.npm import (
     NPM_REGISTRY_URL,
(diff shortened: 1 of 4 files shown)


slimreaper35:
Fix implemented in another PR.

PR #1622, opened 2026-06-17
fix(rpm): accept lockfiles with empty packages and source lists with a warning

Axuba:
## Problem
When `rpm-lockfile-prototype` determines that all required packages are already present, it produces a lockfile with empty `packages: []` and `source: []` lists for each architecture. Hermeto rejects this with `InvalidLockfileFormat`, treating "nothing to fetch" the same as "malformed structure." Notably, the top-level `arches: []` (empty arch list) is already accepted by the `RedhatRpmsLock` model — the restriction was inconsistently applied only at the per-arch level.

## Fix
- **`redhat.py`**: Remove the `LockfileArch._arch_empty` validator entirely. An arch entry with empty `packages` and `source` is a valid state — the downstream `_download` function already handles it gracefully via empty iterators that produce zero download work, verification passes with nothing to verify, and the SBOM output correctly contains zero components.
- **`test_main.py`**: Replace `test_resolve_rpm_project_rejects_empty_arch` with `test_resolve_rpm_project_accepts_empty_arch`. Verify that empty arches produce zero SBOM components and that the download loop is invoked with an empty file map. Add `all_empty` and `mixed_empty_and_valid` parametrized cases covering both fully-empty and mixed lockfiles.

## Validation
Tested locally with an actual empty lockfile produced by `rpm-lockfile-prototype`:
```yaml
lockfileVersion: 1
lockfileVendor: redhat
arches:
- arch: aarch64
  packages: []
  source: []
- arch: x86_64
  packages: []
  source: []
```

All three hermeto commands (fetch-deps, generate-env, inject-files) complete with exit code 0. The output contains valid empty structures (.build-config.json with empty environment_variables and project_files, bom.json with empty components).

### DIFF
--- a/hermeto/core/package_managers/rpm/redhat.py
+++ b/hermeto/core/package_managers/rpm/redhat.py
@@ -3,7 +3,7 @@ import logging
 import uuid
 from functools import cached_property
 
-from pydantic import BaseModel, PositiveInt, field_validator, model_validator
+from pydantic import BaseModel, PositiveInt, field_validator
 
 from hermeto import APP_NAME
 
@@ -32,13 +32,6 @@ class LockfileArch(BaseModel):
     packages: list[LockfilePackage] = []
     source: list[LockfilePackage] = []
     module_metadata: list[LockfileModuleMetadata] = []
-
-    @model_validator(mode="after")
-    def _arch_empty(self) -> "LockfileArch":
-        """Validate arch."""
-        if self.packages == [] and self.source == []:
-            raise ValueError("At least one field ('packages', 'source') must be set in every arch.")
-        return self
 
 
 class RedhatRpmsLock(BaseModel):
--- a/tests/unit/package_managers/rpm/test_main.py
+++ b/tests/unit/package_managers/rpm/test_main.py
@@ -216,10 +216,11 @@ def test_resolve_rpm_project_invalid_lockfile_format(
 
 
 @pytest.mark.parametrize(
-    "lockfile_data",
+    "lockfile_data, expected_components",
     [
         pytest.param(
             {"lockfileVendor": "redhat", "lockfileVersion": 1, "arches": [{"arch": "x86_64"}]},
+            0,
             id="no_packages_or_source",
         ),
         pytest.param(
@@ -228,6 +229,7 @@ def test_resolve_rpm_project_invalid_lockfile_format(
                 "lockfileVersion": 1,
                 "arches": [{"arch": "aarch64", "packages": []}],
             },
+            0,
             id="empty_packages",
         ),
         pytest.param(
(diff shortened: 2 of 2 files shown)


hermeto/core/package_managers/rpm/redhat.py
```diff
@@ -37,7 +37,10 @@ class LockfileArch(BaseModel):
     def _arch_empty(self) -> "LockfileArch":
         """Validate arch."""
         if self.packages == [] and self.source == []:
```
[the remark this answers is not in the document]
Axuba: This just adds more noise to the MR from code that is already there unchanged

eskultety:
> When rpm-lockfile-prototype determines that all required packages are already present, it produces a lockfile with empty packages: [] and source: [] lists for each architecture. Hermeto currently rejects this with InvalidLockfileFormat, causing our prefetch-dependencies task to fail fatally, even though there is genuinely nothing to prefetch.

Makes sense, relaxing a condition is always easier than vice versa.

hermeto/core/package_managers/rpm/redhat.py
```diff
@@ -37,7 +37,10 @@ class LockfileArch(BaseModel):
...
-            raise ValueError("At least one field ('packages', 'source') must be set in every arch.")
+            log.warning(
+                "Arch '%s' has no packages or source RPMs to prefetch.",
+                self.arch,
+            )
```
eskultety: There's no need for a warning - a warning suggests something's off with the input. Since we concluded that there indeed could be a use case for all resource sets to be empty, there's no need for this validator at all.

tests/unit/package_managers/rpm/test_main.py
```diff
@@ -236,20 +236,50 @@ def test_resolve_rpm_project_invalid_lockfile_format(
...
+    components = _resolve_rpm_project(rooted_tmp_path, rooted_tmp_path)
+    assert components == []
+    assert mock_async_download_files.call_count == len(lockfile_data["arches"])
+    for call in mock_async_download_files.call_args_list:
+        assert call[0][0] == {}
+
+
+@mock.patch("hermeto.core.package_managers.rpm.main.run_cmd")
+@mock.patch("hermeto.core.package_managers.rpm.main.async_download_files")
+def test_resolve_rpm_project_mixed_empty_and_valid(
```
eskultety: I think we can consolidate these 2 tests into a single parametrized one without losing much readability.

eskultety (changes requested):

hermeto/core/package_managers/rpm/redhat.py
```diff
@@ -37,7 +37,10 @@ class LockfileArch(BaseModel):
     def _arch_empty(self) -> "LockfileArch":
         """Validate arch."""
         if self.packages == [] and self.source == []:
-            raise ValueError("At least one field ('packages', 'source') must be set in every arch.")
```
eskultety, line 40: **commit message:**
The commit message should not be empty and should provide reasoning behind the change (see CONTRIBUTING.md)

Information in the PR description:

> When rpm-lockfile-prototype determines that all required packages are already present, it produces a lockfile with empty packages: [] and source: [] lists for each architecture. Hermeto currently rejects this with InvalidLockfileFormat, causing our prefetch-dependencies task to fail fatally, even though there is genuinely nothing to prefetch.

^this could be reused if reworded properly, i.e. drop any references to `prefetch-dependencies` task, this project has directly nothing to do with that.

Axuba:
Thanks for the feedback! Implemented the requested changes.

tests/unit/package_managers/rpm/test_main.py
```diff
@@ -236,20 +238,43 @@ def test_resolve_rpm_project_invalid_lockfile_format(
...
+@mock.patch("hermeto.core.package_managers.rpm.main.run_cmd")
+@mock.patch("hermeto.core.package_managers.rpm.main.async_download_files")
+def test_resolve_rpm_project_accepts_empty_arch(
+    mock_async_download_files: mock.Mock,
+    mock_run_cmd: mock.Mock,
+    rooted_tmp_path: RootedPath,
+    lockfile_data: dict,
+    expected_components: int,
 ) -> None:
+    mock_run_cmd.return_value = "name=foo\nversion=1.0\nrelease=1.fc39\narch=x86_64"
```
eskultety: mocking `run_cmd` in this case is IMO overkill, it's like using a canon, especially in this type of parametrized test which is technically error prone if we add some other system command in between in the future. IOW mocking one level above - `Package` -  is going to be cleaner and is agnostic to system invocation changes, the following should do:
```diff
diff --git tests/unit/package_managers/rpm/test_main.py tests/unit/package_managers/rpm/test_main.py
index 45d1311b..a1190e21 100644
--- tests/unit/package_managers/rpm/test_main.py
+++ tests/unit/package_managers/rpm/test_main.py
@@ -260,16 +260,20 @@ def test_resolve_rpm_project_invalid_lockfile_format(
         ),
     ],
 )
-@mock.patch("hermeto.core.package_managers.rpm.main.run_cmd")
+@mock.patch("hermeto.core.package_managers.rpm.main.Package.from_filepath")
 @mock.patch("hermeto.core.package_managers.rpm.main.async_download_files")
 def test_resolve_rpm_project_accepts_empty_arch(
     mock_async_download_files: mock.Mock,
-    mock_run_cmd: mock.Mock,
+    mock_from_filepath: mock.Mock,
     rooted_tmp_path: RootedPath,
     lockfile_data: dict,
     expected_components: int,
 ) -> None:
-    mock_run_cmd.return_value = "name=foo\nversion=1.0\nrelease=1.fc39\narch=x86_64"
+    mock_from_filepath.return_value = mock.Mock(
+        to_component=mock.Mock(
+            return_value=Component(name="foo", version="1.0", purl="pkg:rpm/foo@1.0"),
+        ),
+    )
     with open(rooted_tmp_path.join_within_root("rpms.lock.yaml"), "w") as f:
         yaml.safe_dump(lockfile_data, f)
     components = _resolve_rpm_project(rooted_tmp_path, rooted_tmp_path)
```

Axuba: Applied

eskultety (approved):

slimreaper35 (approved):

PR #1624, opened 2026-06-18
Changed bundler package manager backend to async download HTTP gems

mohamedbouchtout:
### Summery
Bundler now downloads all HTTP dependencies with async instead of sync, similar with the other package managers. Updated the unit tests to test the new changes instead.

### Details
1. Changed the download_to() methods to return the file system location of the gem instead of calling download_binary_file() method; also renamed the method to download_location()
2. Changed _resolve_bundler_package() to download all HTTP gems asynchronously
3. Removed the unit tests that were testing the download_to() method as it's not needed anymore.

Signed-off-by: Mohamed Bouchtout <mbouchto@redhat.com>

### DIFF
--- a/hermeto/core/package_managers/bundler/gem_models.py
+++ b/hermeto/core/package_managers/bundler/gem_models.py
@@ -12,7 +12,7 @@ from typing_extensions import Self
 from hermeto.core.config import get_config
 from hermeto.core.constants import Mode
 from hermeto.core.errors import NotAGitRepo
-from hermeto.core.package_managers.general import download_binary_file, get_vcs_qualifiers
+from hermeto.core.package_managers.general import get_vcs_qualifiers
 from hermeto.core.rooted_path import PathOutsideRoot, RootedPath
 from hermeto.core.scm import GitRepo
 
@@ -41,10 +41,6 @@ class _GemMetadata(pydantic.BaseModel):
     name: str
     version: str
 
-    def download_to(self, deps_dir: RootedPath) -> None:  # noqa: ARG002
-        """Download gem to the specified directory."""
-        return None
-
 
 class GemDependency(_GemMetadata):
     """
@@ -69,11 +65,9 @@ class GemDependency(_GemMetadata):
         """Return remote location to download this gem from."""
         return urljoin(self.source, f"downloads/{self.name}-{self.version}.gem")
 
-    def download_to(self, deps_dir: RootedPath) -> None:
-        """Download represented gem to specified file system location."""
-        fs_location = deps_dir.join_within_root(Path(f"{self.name}-{self.version}.gem"))
-        log.info("Downloading gem %s", self.name)
-        download_binary_file(self.remote_location, fs_location)
+    def download_location(self, deps_dir: RootedPath) -> RootedPath:
+        """Get the file system location of the gem."""
+        return deps_dir.join_within_root(Path(f"{self.name}-{self.version}.gem"))
 
 
 class GemPlatformSpecificDependency(GemDependency):
@@ -89,21 +83,15 @@ class GemPlatformSpecificDependency(GemDependency):
     @property
     def remote_location(self) -> str:
         """Return remote location to download this gem from."""
-        return urljoin(self.source, f"downloads/{self.name}-{self.version}-{self.platform}.gem")
-
-    def download_to(self, deps_dir: RootedPath) -> None:
-        """Download represented gem to specified file system location."""
-        fs_location = deps_dir.join_within_root(
-            Path(f"{self.name}-{self.version}-{self.platform}.gem")
-        )
-        log.info(
-            "Downloading platform-specific gem %s-%s-%s", self.name, self.version, self.platform
-        )
         # A combination of Ruby v.3.0.7 and some Bundler dependencies results in
         # -gnu suffix being dropped from some platforms. This was observed on
         # sqlite3-aarch-linux-gnu. We discourage using outdated platforms
         # for building dependencies and cnsider this to be a limitation of Ruby.
-        download_binary_file(self.remote_location, fs_location)
+        return urljoin(self.source, f"downloads/{self.name}-{self.version}-{self.platform}.gem")
+
+    def download_location(self, deps_dir: RootedPath) -> RootedPath:
+        """Get the file system location of the gem."""
+        return deps_dir.join_within_root(Path(f"{self.name}-{self.version}-{self.platform}.gem"))
 
 
 class GitDependency(_GemMetadata):
--- a/hermeto/core/package_managers/bundler/main.py
+++ b/hermeto/core/package_managers/bundler/main.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: GPL-3.0-only
+import asyncio
 import logging
 import os
 from pathlib import Path
@@ -12,16 +13,17 @@ from hermeto.core.constants import Mode
 from hermeto.core.errors import NotAGitRepo, PackageRejected, UnsupportedFeature
 from hermeto.core.models.input import BundlerBinaryFilters, Request
 from hermeto.core.models.output import EnvironmentVariable, ProjectFile, RequestOutput
-from hermeto.core.models.property_semantics import PropertySet
+from hermeto.core.models.property_semantics import Property, PropertySet
 from hermeto.core.models.sbom import Component, create_backend_annotation
 from hermeto.core.package_managers.bundler.parser import (
+    GemDependency,
     GemPlatformSpecificDependency,
     GitDependency,
     ParseResult,
     PathDependency,
     parse_lockfile,
 )
-from hermeto.core.package_managers.general import get_vcs_qualifiers
+from hermeto.core.package_managers.general import async_download_files, get_vcs_qualifiers
 from hermeto.core.rooted_path import RootedPath
 from hermeto.core.scm import get_repo_id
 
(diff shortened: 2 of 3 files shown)


hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -91,8 +92,8 @@ def remote_location(self) -> str:
         """Return remote location to download this gem from."""
         return urljoin(self.source, f"downloads/{self.name}-{self.version}-{self.platform}.gem")
 
-    def download_to(self, deps_dir: RootedPath) -> None:
-        """Download represented gem to specified file system location."""
+    def download_to(self, deps_dir: RootedPath) -> tuple[str, StrPath] | None:
+        """Get the remote location and the file system location of the gem."""
```
eskultety: Given how this changes the nature of as well as the intent of the method I don't think we can continue using the name given we're returning a tuple. I think something like `download_location` is now more fitting and it should be a cached property instead of a method IMO.

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -103,7 +104,7 @@ def download_to(self, deps_dir: RootedPath) -> None:
         # -gnu suffix being dropped from some platforms. This was observed on
         # sqlite3-aarch-linux-gnu. We discourage using outdated platforms
         # for building dependencies and cnsider this to be a limitation of Ruby.
-        download_binary_file(self.remote_location, fs_location)
+        return self.remote_location, fs_location
```
eskultety: Do we need to return a tuple though? Since `remote_location` is a cached property, we can query it directly  in `_resolve_bundler_package` few lines below.

hermeto/core/package_managers/bundler/main.py
```diff
@@ -97,8 +99,12 @@ def _resolve_bundler_package(
 
     components = [Component(name=name, version=version, purl=main_package_purl.to_string())]
     git_paths = []
+    files_to_download: dict[str, StrPath] = {}
     for dep in dependencies:
-        dep.download_to(deps_dir)
+        result = dep.download_to(deps_dir)
+        if result is not None:
```
eskultety: nitpick: You could wrap this with a walrus operator :) .

hermeto/core/package_managers/bundler/main.py
```diff
@@ -97,8 +99,12 @@ def _resolve_bundler_package(
...
+            url, fs_location = result
```
eskultety: Since both `remote_location` (current) and `download_location` (future) are properties, this expansion feels redundant.

eskultety (changes requested):

hermeto/core/package_managers/bundler/gem_models.py
mohamedbouchtout: I'm not able to cache it as a property since download_location(self, deps_dir: RootedPath) takes a parameter.

eskultety: Well, that error you're getting makes sense. Doesn't matter it MUST be a property though. The linter would accept `@cache` although then you'd get `Unhasheable type`. You could also just have an attribute, but yeah, it's a pydantic model, a function is alright.

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -69,11 +70,11 @@ def remote_location(self) -> str:
         """Return remote location to download this gem from."""
         return urljoin(self.source, f"downloads/{self.name}-{self.version}.gem")
 
-    def download_to(self, deps_dir: RootedPath) -> None:
-        """Download represented gem to specified file system location."""
+    def download_location(self, deps_dir: RootedPath) -> StrPath:
+        """Get the file system location of the gem."""
         fs_location = deps_dir.join_within_root(Path(f"{self.name}-{self.version}.gem"))
         log.info("Downloading gem %s", self.name)
```
eskultety: We're not downloading anything just yet, this is just a helper getter, this logging message should go elsewhere.

hermeto/core/package_managers/bundler/main.py
```diff
@@ -97,18 +100,29 @@ def _resolve_bundler_package(
 
     components = [Component(name=name, version=version, purl=main_package_purl.to_string())]
     git_paths = []
+    files_to_download: dict[str, StrPath] = {}
     for dep in dependencies:
-        dep.download_to(deps_dir)
-        if isinstance(dep, GemPlatformSpecificDependency):
+        properties: list[Property] = []
+        if isinstance(dep, GemDependency):
```
taylormadore: This is going to catch all GemPlatformSpecificDependency objects as well, since it is a subclass of GemDependency

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -41,9 +42,9 @@ class _GemMetadata(pydantic.BaseModel):
     name: str
     version: str
 
-    def download_to(self, deps_dir: RootedPath) -> None:  # noqa: ARG002
-        """Download gem to the specified directory."""
-        return None
+    def download_location(self, deps_dir: RootedPath) -> StrPath:  # noqa: ARG002
```
taylormadore: Should this be in the base class anymore given that it doesn't apply to all of the subclasses?

hermeto/core/package_managers/bundler/main.py
```diff
@@ -97,18 +100,29 @@ def _resolve_bundler_package(
...
-            properties = []
-        if isinstance(dep, GitDependency):
+        elif isinstance(dep, GitDependency):
+            dep.download_to(deps_dir)
             git_paths.append((dep.name, dep.repo_name + "-" + dep.ref[:12]))
 
         c = Component(name=dep.name, version=dep.version, purl=dep.purl, properties=properties)
         components.append(c)
 
+    log.info("Downloading gem(s)")
```
taylormadore: nit: I'm not sure we need this log message. Also, if files_to_download is empty, it will still be emitted

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -202,26 +202,18 @@ def test_parse_gemlock_empty(
         "https://dedicatedprivategemrepo.com",
     ],
 )
-@mock.patch("hermeto.core.package_managers.bundler.gem_models.download_binary_file")
-def test_source_gem_dependencies_could_be_downloaded(
-    mock_downloader: mock.MagicMock,
+def test_source_gem_dependency_download_metadata(
```
taylormadore: We can probably drop these two tests as well. I'm not sure they are adding value. If we feel we want to test the download_location, a doctest will probably suffice

hermeto/core/package_managers/bundler/main.py
```diff
@@ -97,18 +100,28 @@ def _resolve_bundler_package(
 
     components = [Component(name=name, version=version, purl=main_package_purl.to_string())]
     git_paths = []
+    files_to_download: dict[str, StrPath] = {}
     for dep in dependencies:
-        dep.download_to(deps_dir)
+        properties: list[Property] = []
         if isinstance(dep, GemPlatformSpecificDependency):
+            files_to_download[dep.remote_location] = dep.download_location(deps_dir)
```
a-ovchinnikov: A non-blocking, optional suggestion: since you are updating this code WDYT about converting it to match...case?
```python3
match dep:
    case GemPlatformSpecificDependency():
        files_to_download[dep.remote_location] = dep.download_location(deps_dir)
        properties = PropertySet(bundler_package_binary=True).to_properties()
    case GemDependency():
        files_to_download[dep.remote_location] = dep.download_location(deps_dir)
    case GitDependency():
        ...
```

a-ovchinnikov (approved):

taylormadore (approved):

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -69,11 +66,9 @@ def remote_location(self) -> str:
         """Return remote location to download this gem from."""
         return urljoin(self.source, f"downloads/{self.name}-{self.version}.gem")
 
-    def download_to(self, deps_dir: RootedPath) -> None:
-        """Download represented gem to specified file system location."""
-        fs_location = deps_dir.join_within_root(Path(f"{self.name}-{self.version}.gem"))
-        log.info("Downloading gem %s", self.name)
-        download_binary_file(self.remote_location, fs_location)
+    def download_location(self, deps_dir: RootedPath) -> StrPath:
```
eskultety: The return type doesn't sound right - it should be precise - `RootedPath`. Same occurrence below.

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -91,19 +86,13 @@ def remote_location(self) -> str:
...
-        )
-        log.info(
-            "Downloading platform-specific gem %s-%s-%s", self.name, self.version, self.platform
-        )
+    def download_location(self, deps_dir: RootedPath) -> StrPath:
+        """Get the file system location of the gem."""
         # A combination of Ruby v.3.0.7 and some Bundler dependencies results in
         # -gnu suffix being dropped from some platforms. This was observed on
         # sqlite3-aarch-linux-gnu. We discourage using outdated platforms
         # for building dependencies and cnsider this to be a limitation of Ruby.
```
eskultety, line 89: This helper only constructs a path and so the commentary feels misplaced in this context IMO. Consider moving it where `remote_location` is handled.

eskultety (commented):
Minor comments, but looking good. One more respin and we're good to go.

hermeto/core/package_managers/bundler/gem_models.py
mohamedbouchtout: Yeah that seems to used more in the project.

eskultety (approved):

eskultety:
@mohamedbouchtout I just noticed the commit message. Please wrap the long commit msg lines to <52 chars for subject lines and <72 chars for the msg body next time :) (I don't think we have that documented though yet). I fixed that and we can now merge.

PR #1629, opened 2026-06-25
fix(bundler): support git gems

### LINKED ISSUE #1630: bundler: git gems missing from runtime image in multi-stage container builds
### Describe the issue

When a Gemfile declares a git dependency (via `github:`, `git:`, or `branch:`/`ref:` options), Hermeto configures bundler to use the pre-fetched local clone via `BUNDLE_LOCAL.*` config entries. 

This causes bundler to use the gem in-place from Hermeto's output directory rather than installing it into `BUNDLE_PATH`. 
In a multi-stage container build, the runtime image ends up missing those gems, causing load errors at startup.      
                                                                                                                                                                                                                    
This is a problem for repositories depending on a fork or an unreleased branch:                                                                                                                                        

```ruby                                                                                                                                                                                                                    
# Gemfile                                                                                                                                                                                                         
source 'https://rubygems.org'                                                                                                                                                                                     
gem 'json-schema', github: '3scale/json-schema'                                                                                                                                                                   
```
                                                                                                                                                                                                                
After `hermeto fetch-deps`, the generated bundler config contains:                                                                                                                                                  
```ruby                                                                                                                                                                                                                    
BUNDLE_DISABLE_LOCAL_BRANCH_CHECK: "true"                                                                                                                                                                         
BUNDLE_DISABLE_LOCAL_REVISION_CHECK: "true"                                                                                                                                                                       
BUNDLE_LOCAL.JSON___SCHEMA: "${output_dir}/deps/bundler/json-schema-26487618a684" 
#                              ^^^ points into Hermeto's output directory                                                                                                                                
```
                                                                                                                                                                                                                    
With `BUNDLE_LOCAL.*` active, bundler reads the gem in-place from Hermeto's output directory and never copies it into `BUNDLE_PATH`:

```ruby
def requires_checkout?
    allow_git_ops? && !local? && !locked_revision_checked_out?
    #                 ^^^
    #                 false when BUNDLE_LOCAL.* is set -> checkout never runs
end
```

In a multi-stage build the build stage can reach the output directory, but the runtime stage copies only `.bundle/` — so the gem disappears:

```Dockerfile
FROM ruby:3.3 AS build
COPY . .
RUN bundle install   # hermeto supplies BUNDLE_APP_CONFIG

FROM ruby:3.3
COPY --from=build .bundle/ .bundle/
# json-schema is missing here → LoadError at startup
```

### Version

- [x] I have reproduced this issue on the latest version from the main branch

### Steps to reproduce

```ruby
# /tmp/myapp/Gemfile
source 'https://rubygems.org'
gem 'json-schema', github: '3scale/json-schema'
```

```bash
bundle lock # generates Gemfile.lock with a GIT section
hermeto fetch-deps --source /tmp/myapp --output /tmp/hermeto-output bundler
```

```bash
# Inspect the generated config
grep -A1 BUNDLE_LOCAL /tmp/hermeto-output/bundler/config_override/config
# Simulate hermetic bundle install
BUNDLE_APP_CONFIG=/tmp/hermeto-output/bundler/config_override BUNDLE_PATH=/tmp/mybundle bundle install --local
```

```bash
# Gem is not in BUNDLE_PATH
ls /tmp/mybundle/ruby/*/bundler/gems/
```



### Environment

- OS: Fedora Linux 43
- Deployment: localhost
- Ruby: 3.3.10
- Bundler: 2.5.22
- git: 2.54.0


### How does this issue affect you, or how did you find it?

Since merging https://github.com/RedHatInsights/compliance-backend/pull/2753, our service is not starting properly.

We're using gem from a forked repository: https://github.com/RedHatInsights/compliance-backend/blob/a5cf3f037fb25772aa47bf7bb447d27b34ceeeae/Gemfile#L71

romanblanco:
Git gems (declared via `github:` or `git:` in Gemfile) are absent from runtime images in hermetic builds that use multi-stage Dockerfiles.         

## Root cause                                                               
                                                    
`BUNDLE_LOCAL.<gem>` triggers bundler's local override path in `git.rb`:

```ruby
# source/git.rb
def local_override!(path)
  set_install_path!(path)  # @install_path = local clone, not BUNDLE_PATH
  allow_git_ops? && !local? && !locked_revision_checked_out?
  #                 ^^^^^^^^ false when BUNDLE_LOCAL is enabled → skips checkout
end
```

Bundler uses the gem in-place from the clone and never copies it into `BUNDLE_PATH`.
The runtime stage copies only `.bundle/` and the clone is lost.

## Fix

Replace `BUNDLE_LOCAL.*` with `GIT_CONFIG_GLOBAL` pointing to a generated git config with `url.insteadOf` redirects for each git dependency.
Bundler clones from the local `file://` path, `local?` stays `false` and checkout runs which should get the git gem in `BUNDLE_PATH`.

Fixes: https://github.com/hermetoproject/hermeto/issues/1630

Documentation:
https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlbaseinsteadOf
https://git-scm.com/docs/git-config#Documentation/git-config.txt-GITCONFIGCOUNT


eskultety:
> Replace BUNDLE_LOCAL.* with GIT_CONFIG_GLOBAL pointing to a generated git config with url.insteadOf redirects for each git dependency.

I feel like the blast radius of `GIT_CONFIG_GLOBAL` is quite big and quite risky. From Hermeto's POV everything seems fine because we "did our job" and if the user edits their Dockerfile manually it will work. However, one needs to consider Hermeto being used in CI/CD deployments where the context is too wide for the CI/CD system to always do the right thing and (as a matter of fact) `. <dir>/hermeto.env` may be prepended to **every** RUN instruction in the Dockerfile and so every command that you run inside your build (which can be anything really and even use git underneath) could now be indirectly affected by this variable. Note, it's not about the specific options in the config, it's about the fact that we replace the whole global config.

We may be safer with using dedicated key-value pairs: https://git-scm.com/docs/git-config#Documentation/git-config.txt-GITCONFIGCOUNT although that will be more fiddly to correctly generate and still won't be 100% bulletproof in the end, but definitely a lot less likely to cause issues during container build, if this git override works correctly that is (I haven't verified your claim).

romanblanco:
> > Replace BUNDLE_LOCAL.* with GIT_CONFIG_GLOBAL pointing to a generated git config with url.insteadOf redirects for each git dependency.
> 
> We may be safer with using dedicated key-value pairs: https://git-scm.com/docs/git-config#Documentation/git-config.txt-GITCONFIGCOUNT

Thanks for the suggestion!
Updated the PR.

docs/design/bundler.md
```diff
@@ -271,17 +271,17 @@ nokogiri-1.16.6.gem
 tzinfo-2.0.6.gem
 ```
 
-Notice that all the `.gem` dependencies are kept in their original format, and Git dependencies are just plain clones
+Notice that all the `.gem` dependencies are kept in their original format, and Git dependencies are bare git clones
```
eskultety, line 274: @romanblanco so it seems like we have to switch to bare repos anyway. Can you verify this approach works uniformly for both bundler <=2.5 and >=2.6? We must ensure that regardless of the vendor cache format used for git dependencies bundler clones from the local override the same way across versions.
You can play with our integration test here: https://github.com/hermetoproject/integration-tests/tree/bundler/e2e and the corresponding test Dockerfile in this repo to make sure we can cover a list of bundler versions with this approach just fine.

docs/design/bundler.md
```diff
@@ -271,17 +271,17 @@ nokogiri-1.16.6.gem
...
 ```ruby
 "#{base_name}-#{shortref_for_path(revision)}"
 ```
 
-The name of the directory **must come from the Git URL**, not the actual name of the gem, and the cloned folder must
-contain unpacked source code. Any other format will cause bundler to try to re-download the repository, causing the
-build to fail.
+The name of the directory **must come from the Git URL**, not the actual name of the gem. The clones are bare
+repositories used as local remotes via git's `url.insteadOf` mechanism (see
+[Offline installs involving git dependencies](#offline-installs-involving-git-dependencies)).
```
eskultety, line 284: From git config man page:

>  Note that any protocol restrictions will be applied to the rewritten URL. If the rewrite changes the URL to use a custom protocol or remote helper, you may need to adjust the protocol.*.allow config to permit the request. In particular, protocols you expect to use for
           submodules must be set to always rather than the default of user. See the description of protocol.allow above.

We're probably be fine, but to be extra sure we should probably enforce allowing the file protocol just to be safe and also document it for users.

docs/design/bundler.md
```diff
@@ -465,30 +453,15 @@ flag](https://github.com/rubygems/rubygems/issues/8265) via configuration option
...
-
-Therefore, we additionally need to enforce both `BUNDLE_DISABLE_LOCAL_BRANCH_CHECK` and
-`BUNDLE_DISABLE_LOCAL_REVISION_CHECK`.
+Bundler always tries to fetch git dependencies from the remote, so `BUNDLE_DEPLOYMENT` alone
+doesn't prevent network access.
+
+To solve this without coupling to bundler's internal cache format
+(which has [changed across versions](https://github.com/ruby/rubygems/commit/7d6b6316)), we use git's
+[`url.insteadOf`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf)
+via `GIT_CONFIG_COUNT`/`KEY`/`VALUE` environment variables to redirect each remote URL to the
```
eskultety, line 462: It'd be best to have a brief example on this so it's easier for imagination.

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -146,17 +146,13 @@ def download_to(self, deps_dir: RootedPath) -> None:
         git_repo_path.path.mkdir(parents=True)
 
         log.info("Cloning git repository %s", self.url)
-        repo = GitRepo.clone_from(
+        GitRepo.clone_from(
             url=str(self.url),
             to_path=git_repo_path.path,
+            bare=True,
             env={"GIT_TERMINAL_PROMPT": "0"},
         )
```
eskultety, line 140: This commit feels a bit convoluted IMO. I think the changes should be split in 2 commits:

1. enables bare git clones, update env VARS, make sure CI passes (including the integ. tests in this case)
2. modify the logic to start configuring git to support git dependencies the right way.

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -146,17 +146,13 @@ def download_to(self, deps_dir: RootedPath) -> None:
...
 
-        if self.branch is not None:
-            repo.git.checkout(self.branch)
-
-        repo.git.reset("--hard", self.ref)
-
```
eskultety, line 147: Should we test the ref on the bare git clone before bundler tries to clone from it again in the offline environment? I'd be fine with both actually, but maybe users would prefer an early failure rather than build-time failure. I'm ambivalent, any more thoughts on this @hermetoproject/maintainers ?

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -155,6 +157,59 @@ def test_prepare_for_hermetic_build_injects_necessary_variable_into_existing_alt
     assert result.template == existing_preamble + expected_alternate_config_contents
 
 
+def test_prepare_environment_variables_generates_git_config_entries_for_git_deps() -> None:
+    git_paths = [
+        (
+            "yabeda-activejob",
+            "yabeda-activejob-1894d97a925d",
+            "https://github.com/org/yabeda-activejob.git",
```
eskultety: Can we come up with shorter example to fit on a single line?

hermeto/core/package_managers/bundler/gem_models.py
eskultety: I think removing this also makes the `branch` attribute unused now.

tests/unit/package_managers/bundler/test_main.py
```diff
@@ -155,6 +157,59 @@ def test_prepare_for_hermetic_build_injects_necessary_variable_into_existing_alt
...
+        )
+        in result
+    )
+    assert (
+        EnvironmentVariable(
+            name="GIT_CONFIG_VALUE_1",
+            value="https://github.com/org/my-gem.git",
+        )
+        in result
+    )
```
eskultety: This feels quite manual and brittle. Could we instead create a set with these expected values in a loop and then compare `result` with our expected set? It naturally checks ordering based on the exact `GIT_CONFIG_VALUE_N` set which will work for the time being, but one day we should just compare the corresponding _KEY-values with _VALUE-values (that's future though, for now, sets will do).

eskultety (changes requested):

tests/unit/package_managers/bundler/test_main.py
romanblanco: EnvironmentVariable isn't hashable, so used list instead.

docs/design/bundler.md
```diff
@@ -389,8 +389,6 @@ BUNDLE_DEPLOYMENT=true
 BUNDLE_NO_PRUNE=true
 BUNDLE_ALLOW_OFFLINE_INSTALL=true
 BUNDLE_DISABLE_VERSION_CHECK=true
-BUNDLE_DISABLE_LOCAL_BRANCH_CHECK=true
```
eskultety, line 392: **Commit message:**

Unless absolutely trivial we require a descriptive commit message for every change. Here for example it is not obvious at first glance why you removed the branch field from the pudantic model representing a git dependency (my earlier review comment). So explaining the WHY behind this commit is highly desirable. The fact that you removed the `branch` field is only a side effect, mainly related to the fact that our lockfile parser Ruby script never populated that field, so that should be mentioned as well. You might even want to strenghten your statement by referencing https://github.com/ruby/rubygems/blob/v4.0.15/bundler/lib/bundler/source/git.rb#L53-L62 AND https://github.com/ruby/rubygems/blob/e7167b9a42e9804a52c04a2c16aab31949471d69/bundler/lib/bundler/source/git.rb#L38 which both prove that the historical (probably the one we based our impl. on) and current rubygems bundler releases use the same lockfile logic, i.e. only `remote` and `revision` are mandatory fields.

docs/design/bundler.md
```diff
@@ -279,9 +279,9 @@ of the repository placed in a folder. For Git dependencies, the folder name must
 "#{base_name}-#{shortref_for_path(revision)}"
 ```
 
-The name of the directory **must come from the Git URL**, not the actual name of the gem, and the cloned folder must
```
eskultety, line 282: **Commit message:**
Same here, we will need a descriptive commit message (don't forget to add `Closes <issue URL>` in the commit message as well.

eskultety (changes requested):
@romanblanco https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md#pull-request-guidelines

> - Run linters & unit tests before opening a PR
> - Ensure every single commit passes CI

eskultety:
@romanblanco you will need to squash the following to the 2nd commit:
```diff
diff --git tests/integration/test_data/bundler_e2e/.build-config.yaml tests/integration/test_data/bundler_e2e/.build-config.yaml
index b23f60ab..34ab6c93 100644
--- tests/integration/test_data/bundler_e2e/.build-config.yaml
+++ tests/integration/test_data/bundler_e2e/.build-config.yaml
@@ -2,13 +2,13 @@ environment_variables:
 - name: BUNDLE_APP_CONFIG
   value: ${output_dir}/bundler/config_override
 - name: GIT_CONFIG_COUNT
-  value: "2"
+  value: '2'
 - name: GIT_CONFIG_KEY_0
   value: url.file://${output_dir}/deps/bundler/json-schema-26487618a684/.insteadOf
-- name: GIT_CONFIG_VALUE_0
-  value: https://github.com/3scale/json-schema
 - name: GIT_CONFIG_KEY_1
   value: protocol.file.allow
+- name: GIT_CONFIG_VALUE_0
+  value: https://github.com/3scale/json-schema
 - name: GIT_CONFIG_VALUE_1
   value: always
 project_files:
diff --git tests/integration/test_data/bundler_e2e_missing_gemspec/.build-config.yaml tests/integration/test_data/bundler_e2e_missing_gemspec/.build-config.yaml
index b23f60ab..34ab6c93 100644
--- tests/integration/test_data/bundler_e2e_missing_gemspec/.build-config.yaml
+++ tests/integration/test_data/bundler_e2e_missing_gemspec/.build-config.yaml
@@ -2,13 +2,13 @@ environment_variables:
 - name: BUNDLE_APP_CONFIG
   value: ${output_dir}/bundler/config_override
 - name: GIT_CONFIG_COUNT
-  value: "2"
+  value: '2'
 - name: GIT_CONFIG_KEY_0
   value: url.file://${output_dir}/deps/bundler/json-schema-26487618a684/.insteadOf
-- name: GIT_CONFIG_VALUE_0
-  value: https://github.com/3scale/json-schema
 - name: GIT_CONFIG_KEY_1
   value: protocol.file.allow
+- name: GIT_CONFIG_VALUE_0
+  value: https://github.com/3scale/json-schema
 - name: GIT_CONFIG_VALUE_1
   value: always
 project_files:
```

docs/design/bundler.md
eskultety: I tried to  verify simply with the following naive diff on our integration test:
```diff
diff --git tests/integration/test_data/bundler_e2e/container/Containerfile tests/integration/test_data/bundler_e2e/container/Containerfile
index 35e753af..a560f158 100644
--- tests/integration/test_data/bundler_e2e/container/Containerfile
+++ tests/integration/test_data/bundler_e2e/container/Containerfile
@@ -1,4 +1,4 @@
-FROM mirror.gcr.io/ruby:3.3
+FROM mirror.gcr.io/ruby:4.0
```
which has
```bash
$ podman run --it ruby:4.0-alpine bundler --version
4.0.10
```
and the tests have passed. What do you think @romanblanco , can you confirm my request from above?

eskultety (commented):
@romanblanco I did some further experiments on our side in context of this request of mine: https://github.com/hermetoproject/hermeto/pull/1629/changes#r3490366370

```diff
diff --git tests/integration/test_bundler.py tests/integration/test_bundler.py
index 55438345..35966c93 100644
--- tests/integration/test_bundler.py
+++ tests/integration/test_bundler.py
@@ -68,9 +68,19 @@ def test_bundler_packages(
                 packages=({"path": ".", "type": "bundler", "binary": {}},),
                 check_output=True,
             ),
-            [],  # No additional commands are run to verify the build
             [],
-            id="bundler_e2e",
+            [],
+            id="bundler_e2e_ruby33",
+        ),
+        pytest.param(
+            utils.TestParameters(
+                branch="bundler/e2e",
+                packages=({"path": ".", "type": "bundler", "binary": {}},),
+                check_output=True,
+            ),
+            [],
+            [],
+            id="bundler_e2e_ruby40",
         ),
         pytest.param(
             utils.TestParameters(
```
Then I migrated/introduced test case/s to cover both old and new bundler with multi-stage builds, e.g.:
```diff
diff --git tests/integration/test_data/bundler_e2e_ruby40/container/Containerfile tests/integration/test_data/bundler_e2e_ruby40/container/Containerfile
new file mode 100644
index 00000000..0507284e
--- /dev/null
+++ tests/integration/test_data/bundler_e2e_ruby40/container/Containerfile
@@ -0,0 +1,15 @@
+FROM mirror.gcr.io/ruby:4.0 AS builder
+
+# Test disabled network access
+RUN if getent hosts www.google.com; then echo "Has network access!"; exit 1; fi
+
+WORKDIR /src
+
+ENV BUNDLE_PATH=/gems
+RUN . /tmp/hermeto.env && bundle install
+
+FROM mirror.gcr.io/ruby:4.0
+
+COPY --from=builder /gems /gems
+RUN test -d /gems/ruby/*/bundler/gems/json-schema-*
```

and it worked. I'm not encouraging to copy this verbatim, this is something I botched quickly locally. However, I think the PR could go like this:

1. Rename the existing integration test `bundler_e2e` test data to `bundler_e2e_ruby33`
2. Convert to cloning bare repos
3. Introducing the GIT_CONFIG configuration
4. Changing the renamed e2e_ruby33 test to a multi-stage docker build and introducing a new ruby 4.0 test pretty much identical to the e2e_ruby33 one so that we have coverage for both old and new bundler going forward.

How does ^that sound?

eskultety:
> feat(bundler): support git gems

btw. This is not a feature, it is a bugfix ;) .

tests/unit/package_managers/bundler/test_main.py
eskultety: Oh, good point. Here's a fix for that https://github.com/hermetoproject/hermeto/pull/1651 . I believe it will be merged by the time you get to this :) so please go with sets.

docs/design/bundler.md
romanblanco: There was a minor problem in ordering of key/values (is expected to be alphabetical). Fixed that in the spec's expectation.

Otherwise, the support seems to be working.

romanblanco: > There was a minor problem in ordering of key/values (is expected to be alphabetical).

...as you've also noted [below](https://github.com/hermetoproject/hermeto/pull/1629#issuecomment-4852883035)

eskultety: Yeah, test data. We haven't had time to improve the matching logic of that data sadly.

docs/design/bundler.md
```diff
@@ -271,7 +271,7 @@ nokogiri-1.16.6.gem
 tzinfo-2.0.6.gem
 ```
 
-Notice that all the `.gem` dependencies are kept in their original format, and Git dependencies are just plain clones
```
eskultety, line 274: **Commit message:**

> Use bare git clones (`git clone --bare`) instead of full clones for
> git dependencies. Bare clones contain only git objects without a
> working tree, which is all that bundler needs to resolve the source.

This is all true, undisputed, but I'd like to see more context in the message for bookkeeping reasons, i.e. this commit deletes the BUNDLE_LOCAL_* variables and converts to bare git clones but doesn't discuss why that is okay to do, especially since newer bundler would likely complain. Therefore, I'd like to see context expansion saying something along the lines of the following:

- the design was originally based on newer bundler than we tested for which differs in the git gem cache format
- adjust the design to favour the older bundler approach (bare git clones) as tested, the next patch will expand this with proper support for newer bundler and provide version agnostic support

docs/design/bundler.md
```diff
@@ -279,9 +279,9 @@ of the repository placed in a folder. For Git dependencies, the folder name must
 "#{base_name}-#{shortref_for_path(revision)}"
 ```
 
-The name of the directory **must come from the Git URL**, not the actual name of the gem, and the cloned folder must
-contain unpacked source code. Any other format will cause bundler to try to re-download the repository, causing the
-build to fail.
+The name of the directory **must come from the Git URL**, not the actual name of the gem. The clones are bare
+repositories used as local remotes via git's `url.insteadOf` mechanism (see
+[Offline installs involving git dependencies](#offline-installs-involving-git-dependencies)).
```
eskultety, line 284: nitpick: This hunk would deserve to be split to individual adjustments and introduced in the past (bare clones) commit and also this one, but hey, it's a design doc, I won't insist on that surgical work ;) , just an FYI.

docs/design/bundler.md
```diff
@@ -453,30 +453,29 @@ flag](https://github.com/rubygems/rubygems/issues/8265) via configuration option
...
-
->Bundler does many checks to ensure a developer won't work with invalid references. Particularly,
->we force a developer to specify a branch in the Gemfile in order to use this feature. If the
->branch specified in the Gemfile and the current branch in the local git repository do not
->match, Bundler will abort.
-
-Therefore, we additionally need to enforce both `BUNDLE_DISABLE_LOCAL_BRANCH_CHECK` and
-`BUNDLE_DISABLE_LOCAL_REVISION_CHECK`.
+Bundler always tries to fetch git dependencies from the remote, so `BUNDLE_DEPLOYMENT` alone
+doesn't prevent network access.
```
eskultety, line 457: With this hunk here I'd like to be more thorough though - it should go to the earlier commit switching to bare git clones. It's fine to keep the section in that commit brief, we'll add the rest (below) in this commit.

docs/design/bundler.md
```diff
@@ -453,30 +453,29 @@ flag](https://github.com/rubygems/rubygems/issues/8265) via configuration option
...
+GIT_CONFIG_COUNT=2
+GIT_CONFIG_KEY_0=url.file:///output/deps/bundler/json-schema-26487618a684/.insteadOf
+GIT_CONFIG_VALUE_0=https://github.com/3scale/json-schema
+GIT_CONFIG_KEY_1=protocol.file.allow
+GIT_CONFIG_VALUE_1=always
+```
+
+The `protocol.file.allow=always` entry ensures git permits the `file://` protocol after the URL
+rewrite. See the [git documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt-protocolallow)
+for details.
```
eskultety, line 478: Yes, this hunk discussing the actual solution we've implemented should remain here as is. :)

hermeto/core/package_managers/bundler/main.py
```diff
@@ -194,11 +195,48 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+) -> None:
+    """Raise if multiple git deps share the same URL with different revisions."""
+    url_to_dirs: dict[str, set[str]] = {}
+    for _packname, dirname, url in git_paths:
+        url_to_dirs.setdefault(url, set()).add(dirname)
+    for url, dirs in url_to_dirs.items():
+        if len(dirs) > 1:
+            raise UnsupportedFeature(
+                f"Multiple git dependencies point to the same repository ({url}) "
+                f"but use different revisions: {', '.join(sorted(dirs))}. "
```
eskultety: Use populate `revisions` in the error message with `dirs` which are filesystem dir names (e.g. `name-abc123`) rather than bare refs.

eskultety (commented):
This also needs to be documented in the user-facing docs, i.e. `docs/bundler.md` to tell users about the build-time env variables we expose. This is because it's still not a bulletproof solution, IOW the GIT_CONFIG_N numbered variables we expose might potentially conflict with other identically named variables populated from outside the container image during build in which case our variables would overwrite those. That said, what we've come up with seems like the only way for now and is definitely a smaller blast radius than setting global git config. Please make sure to warn users about this, so they can account for that in their builds.

Other than that the logic is IMO sound, I verified each commit in this revision with integration tests to make sure we don't magically break in between commits (it's important here) and everything is green.

Minor comments left since there will have to be 1 more revision and we should be good to go.

hermeto/core/package_managers/bundler/main.py
```diff
@@ -194,11 +195,48 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
+    return env_vars
+
+
+def _check_for_duplicate_git_urls(
+    git_paths: list[tuple[DepName, FSDepName, DepURL]],
+) -> None:
+    """Raise if multiple git deps share the same URL with different revisions."""
+    url_to_dirs: dict[str, set[str]] = {}
+    for _packname, dirname, url in git_paths:
+        url_to_dirs.setdefault(url, set()).add(dirname)
```
a-ovchinnikov: Non-blocking nitpicks: `_packname` is not used in this function thus `_` should be enough (that would also tell the reader that the element is not used); since you are not using the return value of `setdefault` it might be better to use defaultdict:
```python3
from collections improt defaultdict
...
...
...
urls_to_dirs = defaultfict(set)
for _, dirname, url in git_paths:
    urls_to_dirs[url].add(dirname)
```
It would also be nice to find all offending url:dirs pairs in one go (there could be more than one and users would get really upset if they were to fish for them one at a time):
```python3
offenders = [(url, dirs) for url, dirs in urls_to_dirs.items() if len(dirs) > 1]
if offenders:
    raise ...
```

hermeto/core/package_managers/bundler/main.py
```diff
@@ -194,11 +195,48 @@ def _get_repo_name_from_origin_remote(package_dir: RootedPath) -> str:
...
-    return [
+def _prepare_environment_variables_for_hermetic_build(
+    git_paths: list[tuple[DepName, FSDepName, DepURL]] | None = None,
+) -> list[EnvironmentVariable]:
+    env_vars = [
         # Contains path to a directory where a new config could be found.
         EnvironmentVariable(name="BUNDLE_APP_CONFIG", value="${output_dir}/" + CONFIG_OVERRIDE),
     ]
+    if git_paths:
+        # Redirect git remote URLs to pre-fetched local clones via git's
```
a-ovchinnikov: A non-blocking request: could you please move lines 206 to 221 into a separate helper, maybe `_produce_git_redirection_variables(git_paths: list[tuple[DepName, FSDepName, DepURL]] | None) -> list[EnvironmentVariable]` to get something like this:
```python3
env_vars = [
        # Contains path to a directory where a new config could be found.
        EnvironmentVariable(name="BUNDLE_APP_CONFIG", value="${output_dir}/" + CONFIG_OVERRIDE),
]
env_vars.extend(_produce_git_redirection_variables(git_paths))
return env_vars
```

a-ovchinnikov (commented):
A couple of minor nitpicks, LGTM otherwise.

hermeto/core/package_managers/bundler/main.py
romanblanco: Not sure I understood your comment right, but I'm changing the message to `...but need different clones: ...`.

romanblanco:
Thanks for the review, @a-ovchinnikov @eskultety! Updated the PR according to your comments.

a-ovchinnikov (approved):

eskultety (approved):

PR #1632, opened 2026-06-26
pnpm documentation

### LINKED ISSUE #1251: Support for pnpm package manager

slimreaper35:
Closes https://github.com/hermetoproject/hermeto/issues/1251


hermeto/core/package_managers/javascript/pnpm/project.py
```diff
@@ -92,7 +92,14 @@ class PnpmPackage:
...
     version: str
     url: str
-    integrity: str | None = None  # git and URL dependencies don't have an integrity field
+    # Non-registry dependencies may lack the integrity field in lockfiles generated by pnpm v10.
+    # By running various commands, pnpm can "magically" drop it:
+    # - https://github.com/pnpm/pnpm/issues/12067
+    # - https://github.com/pnpm/pnpm/issues/12001
+    #
+    # This broken logic has been fixed in pnpm 11.5.1: https://github.com/pnpm/pnpm/releases/tag/v11.5.1
+    # Once pnpm v10 is EOL and unsupported by Hermeto, integrity should be required.
```
eskultety, line 102: nitpick: Maybe a FIXME note?

docs/pnpm.md
```diff
@@ -0,0 +1,132 @@
+# [pnpm][]
```
eskultety, line 1: > Assisted-by: SoltisGPT

Wut?

eskultety, line 1: This doc isn't linked from the docs index.

docs/pnpm.md
```diff
@@ -0,0 +1,132 @@
...
+```shell
+hermeto generate-env ./hermeto-output --for-output-dir /tmp/hermeto-output --output hermeto.env
+```
+
+#### Inject project files
+
+Modify `pnpm-lock.yaml` and `.npmrc` for the build.
+
+```shell
+hermeto inject-files ./hermeto-output --for-output-dir /tmp/hermeto-output
```
eskultety, line 123: nitpick: this seems inconsistent with the examples in other backends. Our examples should mostly read identically.

eskultety: Additionally, https://github.com/hermetoproject/hermeto/blob/main/docs/npm.md#generate-environment-variables also explicitly states that we currently don't generate any vars.

docs/pnpm.md
```diff
@@ -0,0 +1,132 @@
...
+
+WORKDIR /app
+COPY . .
+
+RUN source /tmp/hermeto.env && pnpm install
+...
+```
+
+```shell
+docker build . \
```
eskultety: nitpick: we use `podman build` consistently across our usage docs.

docs/pnpm.md
```diff
@@ -0,0 +1,132 @@
...
+
+Dockerfile:
+
+```dockerfile
+FROM docker.io/library/node:latest
+
+WORKDIR /app
+COPY . .
+
+RUN source /tmp/hermeto.env && pnpm install
```
eskultety: This won't work universally. The latest node image is 25+ which no longer ships with corepack. You don't mention the need for corepack to use PNPM, here you expect users to know to already have it pre-installed and PNPM enabled.

docs/pnpm.md
```diff
@@ -0,0 +1,132 @@
...
+
+Alternatively, via environment variables:
+
+```shell
+export HERMETO_PNPM__PROXY_URL=https://my-npm-proxy.example.com
+export HERMETO_PNPM__PROXY_LOGIN=user
+export HERMETO_PNPM__PROXY_PASSWORD=pass
+```
+
+The proxy is used **only** for packages from the official npm registry.
```
eskultety: this should IMO appear in the section heading instead, e.g. 'Proxy (npm registry packages only)'

docs/pnpm.md
```diff
@@ -0,0 +1,132 @@
...
+├── npm/
+│   ├── react-18.3.1.tgz
+│   └── ...
+├── jsr/
+│   ├── ...
+```
+
+Hermeto modifies `.npmrc` and `pnpm-lock.yaml` to point registries and tarball URLs
+to local files in the output directory. Other settings in `.npmrc` are preserved.
+If the file does not exist, it is created.
```
eskultety, line 35: nitpick: maybe add a hint that this happens in `inject-files` like NPM mentions : https://github.com/hermetoproject/hermeto/blob/main/docs/npm.md#using-fetched-dependencies

eskultety (changes requested):

hermeto/core/package_managers/javascript/pnpm/project.py
slimreaper35: added

docs/pnpm.md
slimreaper35: fixed

docs/pnpm.md
slimreaper35: fixed

docs/pnpm.md
slimreaper35: added

docs/pnpm.md
eskultety: > nitpick: this seems inconsistent with the examples in other backends. Our examples should mostly read identically.

I realized I attached this to the wrong docs hunk, sorry, it should have been about `generate-env`.  For reference, seems like we consistently write the command as https://github.com/hermetoproject/hermeto/blob/main/docs/npm.md#generate-environment-variables. I would normally not block on such a stupid thing, but when it comes to docs I really value uniform, consistent experience across all docs articles :) .

docs/pnpm.md
```diff
@@ -0,0 +1,171 @@
...
+export HERMETO_PNPM__PROXY_LOGIN=user
+export HERMETO_PNPM__PROXY_PASSWORD=pass
+```
+
+> **NOTE**
+>
+> The proxy is used only for packages from the official npm registry. Other package
+> sources do not work with the proxy.
+
+## Example
```
eskultety, line 88: nitpick: Looks like we consistently (except for rpm.md) use heading H3 for examples.

docs/pnpm.md
```diff
@@ -0,0 +1,171 @@
...
+
+RUN source /tmp/hermeto.env && pnpm install
+
+# ...
+```
+
+> **NOTE**
+>
+> Alternatively, you can try to use the [corepack][] to install a specific version
+> of pnpm. However, corepack only ships with node images up to version 24 (inclusive).
```
eskultety: Hold on, from a user perspective - how does this relate to the example provided exactly? IOW how would a user use corepack in their build env? I am not dismissive of the `corepack` scenario at all, but this is a vague suggestion, do we know it works reliably? If so, we should include just a little snippet like:

>Put this instead of `npm install` in your Dockerfile:
>`RUN corepack enable pnpm@<version>`

docs/pnpm.md
```diff
@@ -0,0 +1,171 @@
...
+
+```shell
+podman build . \
+  --volume "$(realpath ./hermeto-output)":/tmp/hermeto-output:Z \
+  --volume "$(realpath ./hermeto.env)":/tmp/hermeto.env:Z \
+  --network none \
+  --tag pnpm-app
+```
+
+When using pnpm v10, you can omit sourcing hermeto.env and drop the volume mount.
```
eskultety: This should also be a `NOTE` and come just before the podman comman example so it reads nicely e.g.
_When using pnpm v10 you can omit sourccing and volume mount in the command below_

docs/pnpm.md
slimreaper35: > do we know it works reliably? If so, we should include just a little snippet like:

I don't know. I put this as an alternative option. Users would probably have to use something like `corepack pack pnpm@...` and then `corepack enable -g pnpm@/...`

docs/pnpm.md
slimreaper35: I prefer current form of the commands because it aligns with inject-files below minus the `--output` option. 

Nobody will notice any difference. There any many inconsistencies across all package managers docs + usage.md + README.md + design/ ...

eskultety: Fair enough.

docs/pnpm.md
eskultety: If we're not 100% confident, then just leave it out completely. If someone in the JS ecosystem feels strongly about it, we'll happily merge an improvement.

slimreaper35: ok

docs/pnpm.md
slimreaper35: done

eskultety (approved):

derasdf (approved):

a-ovchinnikov (approved):

PR #1633, opened 2026-06-26
skills (vol.1): Introduce an agentic review skill

eskultety:
This introduces a review skill somewhat tailored to the current Hermeto review process. It can handle PR reviews, but also local non-published/uncommitted work. It consists of 2 parts:

1. a prep Python script that does the heavy lifting of figuring out what the correct merge base
2. the actual skill controlling the prep script and prompting the a review subagent injecting the changes to its context (look at the skill header to see how it's used)

For best experience this was modeled around the `gh` tool (not strictly needed) because that's what today's LLMs are inclined to use automatically when dealing with github.


eskultety:
I tried to capture everything that I as a reviewer look at, what I tend to consider and what we have agreed to follow as a project. It somehow translated to agentic instructions, but honestly, it's like writing a poem. What reads like a good poem to me isn't interested to other people...
The agentic apocalypse is upon us and while I definitely do not dispute some of the outputs it provides even in review context and it can provide good value in the ideal world, however... that's not the world we live in, from security perspective all of this is absolutely horrible, it's like casting Harry Potter spells trying to fight prompt injection attacks holding a blackbox instead of a wand not knowing anything about the embedded logic in it. It is not exact language, it is not deterministic, but here we are.

I am open to a discussion on just about any change to the skill, I tried my best and spent many tokens to slop on this to somewhat capture the process in a single markdown file. 

Lastly, ^this all said I still think that if used correctly in local environment responsibly it could decrease the time we spend on reviews quite a bit if this is used to iterate on during development before a PR is opened.

eskultety:
FWIW initially I was playing with the idea of quick nox check in an interactive rebase as well as checking the actual permissions of the github token used with `gh` and I simply abandoned those ideas eventually due to security problems.

skills/scripts/review-prep
```diff
@@ -0,0 +1,228 @@
...
+    log.info("Fetching PR description")
+    pr_desc = _gh("pr", "view", pr_number, "--json", "body", "--jq", ".body // empty")
+
+    log.info("Fetching review comments")
+    comments = _gh(
+        "api",
+        "--paginate",
+        f"repos/{slug}/pulls/{pr_number}/comments",
+        "--jq",
+        ".[] | select(.in_reply_to_id == null) | {path, line, body}",
```
derasdf, line 138: Missing `issues/N/comments` endpoint. You fetch `pulls/N/comments` (inline) and `pulls/N/reviews` (verdicts), but not `issues/N/comments` (conversation discussion). Your own comments on this PR live in `issues/1633/comments` and wouldn't be visible to the reviewer. I ran the prep-script against this PR and confirmed the JSON output has no trace of them.
Adding it raw is a prompt injection risk though. I would filter by repo collaborators and only include content from some whitelist of trusted users. I feel like it should be done for all these endpoints for extra safety.

Also: `select(.in_reply_to_id == null)` drops reply comments. The skill prompt says to check "whether open threads indicate unresolved concerns" but replies that resolve them are filtered out. In testing against PR #1632, the reviewer re-raised concerns that were already resolved.

skills/scripts/review-prep
```diff
@@ -0,0 +1,228 @@
...
+    deletions = re.search(r"(\d+) deletion", shortstat)
+    return (int(insertions.group(1)) if insertions else 0) + (
+        int(deletions.group(1)) if deletions else 0
+    )
+
+
+def _prepare_pr(pr_number: str) -> dict[str, str | None]:
+    """Fetch PR ref and resolve to a concrete SHA."""
+    log.info("Fetching PR %s ref and updating origin/main", pr_number)
+    result = _git("fetch", "origin", f"pull/{pr_number}/head", "main", check=False)
```
derasdf, line 52: `git fetch origin pull/N/head main` writes two entries to FETCH_HEAD. `rev-parse FETCH_HEAD` returns the first one, which is the PR ref because it's the first CLI argument. But if someone refactors the argument order to `git fetch origin main pull/N/head`, rev-parse silently returns main's SHA instead and the script reviews the wrong code. Splitting into two fetch calls or fetching into a named ref avoids this.

skills/scripts/review-prep
```diff
@@ -0,0 +1,228 @@
...
+    return {
+        "diff_args": f"origin/main...{pr_sha}",
+        "log_range": f"origin/main..{pr_sha}",
+        "pr_sha": pr_sha,
+    }
+
+
+def _prepare_ref(ref: str) -> dict[str, str | None]:
+    """Compute diff and log arguments for a git ref or range."""
+    if ".." in ref:
```
derasdf, line 75: nitpick: if someone passes a three-dot range, `git log` will show commits from both branches in the subagent's orientation step, which could confuse it into reviewing unrelated commits. s/`"log_range": ref`/`"log_range": ref.replace("...", "..")`/

skills/scripts/review-prep
```diff
@@ -0,0 +1,228 @@
...
+        "--paginate",
+        f"repos/{slug}/pulls/{pr_number}/reviews",
+        "--jq",
+        ".[] | {user: .user.login, state: .state, body: .body}",
+    )
+
+    return {
+        "available": True,
+        "slug": slug,
+        "pr_description": pr_desc.stdout if pr_desc.returncode == 0 else "",
```
derasdf, line 153: nitpick: when `pr_desc`, `comments`, or `reviews` fail (nonzero returncode), empty strings are silently returned. The reviewer won't know that review comments failed to fetch vs there being none. `log.warning` per failed sub-call would surface this.

skills/scripts/review-prep
```diff
@@ -0,0 +1,228 @@
...
+    )
+
+
+def _prepare_pr(pr_number: str) -> dict[str, str | None]:
+    """Fetch PR ref and resolve to a concrete SHA."""
+    log.info("Fetching PR %s ref and updating origin/main", pr_number)
+    result = _git("fetch", "origin", f"pull/{pr_number}/head", "main", check=False)
+    if result.returncode != 0:
+        raise ReviewPrepError(
+            f"failed to fetch PR {pr_number} — check that the PR exists "
```
derasdf, line 55: nitpick: em dash in error message, they still trigger me.

derasdf (commented):
Solid foundation, the prep script approach is definitely better than raw LLM git operations. 
When we were working on AGENTS.md you [pointed out](https://github.com/hermetoproject/hermeto/pull/1565#discussion_r3241927396) that testing patterns and [design doc requirements](https://github.com/hermetoproject/hermeto/pull/1565#discussion_r3242095594) are "skill material" not AGENTS.md material. Is vol.2 where those go as more specialised skills or is there a plan to expand the review criteria here?
Overall suggestions below are made from my experience with my own local review skill, most of nitpicks with review-prep was found using it.

PR #1638, opened 2026-06-29
fix(pip): Make custom indices override proxy setting

a-ovchinnikov:
Prior to this change a proxy URL would override all sources for pip including custom indices which is not correct -- if a custom index is present it must be used since a proxy is not guaranteed to mirror it (unlike the standard source). This change implements this rule and also ensures that proxy authentication headers will be added only to requests issued to proxies and not to custom indices.

The change is tested with `pip/custom_index` scenario.

### DIFF
--- a/hermeto/core/package_managers/pip/main.py
+++ b/hermeto/core/package_managers/pip/main.py
@@ -386,10 +386,15 @@ def _resolve_and_download_pypi_packages(
     """Resolve and download all PyPI packages."""
     config = get_config()
     proxy_url = str(config.pip.proxy_url) if config.pip.proxy_url is not None else None
-    query_url = proxy_url if proxy_url is not None else index_url
+    # A proxy must be used only if an index is the standard one, custom indices must
+    # be preserved. A custom index would always override proxy.
+    is_standard = lambda index_url: index_url and index_url == pypi_simple.PYPI_SIMPLE_ENDPOINT
+    query_url = proxy_url if (proxy_url is not None and is_standard(index_url)) else index_url
     requests_auth = None
     aiohttp_auth = None
-    if config.pip.proxy_login and config.pip.proxy_password:
+    # No attempt should be made at logging into custom indices, if query URL ends up pointing
+    # to a custom index then authorization must be skipped.
+    if config.pip.proxy_login and config.pip.proxy_password and (query_url == proxy_url):
         requests_auth = requests.auth.HTTPBasicAuth(
             config.pip.proxy_login, config.pip.proxy_password
         )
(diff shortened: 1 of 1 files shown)


eskultety:
> Prior to this change a proxy URL would override all sources for pip including custom indices which is not correct -- if a custom index is present it must be used since a proxy is not guaranteed to mirror it (unlike the standard source). 

Yes, custom indices require careful setup of proxies and from Hermeto POV it isn't obvious at all which index source should be wrapped by a proxy (which proxy?) and which ones should be left intact. That is a non-trivial use case to resolve we might not even have all the necessary bits to control, but should the need arise, we'll need to investigate it deeper and consider our options.

hermeto/core/package_managers/pip/main.py
slimreaper35, line 420: Should we pass here `query_url` too?

hermeto/core/package_managers/pip/main.py
```diff
@@ -386,10 +386,15 @@ def _resolve_and_download_pypi_packages(
     """Resolve and download all PyPI packages."""
     config = get_config()
     proxy_url = str(config.pip.proxy_url) if config.pip.proxy_url is not None else None
-    query_url = proxy_url if proxy_url is not None else index_url
+    # A proxy must be used only if an index is the standard one, custom indices must
+    # be preserved. A custom index would always override proxy.
+    is_standard = lambda index_url: index_url and index_url == pypi_simple.PYPI_SIMPLE_ENDPOINT
+    query_url = proxy_url if (proxy_url is not None and is_standard(index_url)) else index_url
```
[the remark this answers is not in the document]
slimreaper35:
```suggestion
    use_proxy = proxy_url is not None and index_url == pypi_simple.PYPI_SIMPLE_ENDPOINT
    query_url = proxy_url if use_proxy else index_url
```

hermeto/core/package_managers/pip/main.py
eskultety: While the logic above is sound, 2 things on this call:

1. I find the interface now quite confusing, because the downloader only consults auth when actually downloading, index and proxy urls are only for the SBOM component generator. Maybe we should extract that logic for separation of concerns OR pass index and proxy as opaque kwargs to signal that the downloader doesn't touch them and only passes them through. That is non-blocking for this PR, I'm just mentioning it as something I'd probably want to track and improve from maintenance POV in the future.
2. **`proxy_url` will still appear in the SBOM as an extended ref for packages with custom index URL which never consulted the proxy - I believe this is something we should fix in order to adhere to our strictly correct SBOM mission**

eskultety: Do we need to? What information does it hold that we wouldn't already have inside the downloader?

a-ovchinnikov: On 1.: I agree, this has to be reworked, the change as it is now is essentially a hot fix. On 2.: Yes, that a good point, I'll push a fix shortly.

slimreaper35 (approved):

hermeto/core/package_managers/pip/main.py
```diff
@@ -410,7 +415,9 @@ def _resolve_and_download_pypi_packages(
         pip_deps_dir,
         pypi_artifacts,
         index_url=index_url,
-        proxy_url=proxy_url,
+        # If a standard PyPI index is used with proxy URL then proxy URL must be reported,
+        # if a custom index is used then proxy URL must not be reported even if set.
+        proxy_url=proxy_url if (proxy_url is not None and (proxy_url != index_url)) else None,
```
eskultety: ACK to the change, but adding commentaries to function call parameters IMO breaks the natural visual flow, can you please extract it in front of the call? A simple helper named `report_proxy_url_in_sbom` or lambda (you like named lambas, right? :) ) will do.

eskultety (approved):
@a-ovchinnikov we also have https://github.com/hermetoproject/hermeto/pull/1525 in the merge queue which would cause a tiny merge conflict on the `auth` attribute which will be gone after that PR.

hermeto/core/package_managers/pip/main.py
a-ovchinnikov: Done.

eskultety: ACK.

PR #1639, opened 2026-06-29
A collection of small refactors vol. 1

a-ovchinnikov:
This PR is a more or less organized part of a collection of minor changes I accumulated over time which I think make code easier to follow. Most of them are related to a wider effort of making our type system more precise (at least at the naming level) and thus easier to digest and less error prone. Several commits refactor the code because at some point I realized that just declaring a more specific type was not enough. The PR is limited to changes in core/models only and is intended as a demo of a direction we might want to take the codebase.

### DIFF
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -110,6 +110,7 @@ PackageManagerType = Literal[
     "pip",
     "rpm",
     "yarn",
+    "yarn-classic",
     # Add experimental package managers (or package managers whose implementation is in progress)
     # here with an x- prefix (e.g. "x-foo"):
     "x-maven",
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -217,15 +217,24 @@ class BinaryModeOptions(pydantic.BaseModel, extra="forbid"):
     packages: BinaryFilterStr = BINARY_FILTER_ALL
 
 
+PlatformRegexStr = str
+# *(list(range(310, 400))) cannot be used here as long as 3.10 is supported.
+# Once added it would support all Python version for the next fifty years or
+# so.  This could be tightened up further, (like it is implemented now) but
+# that would require constant care -- once a new version gets out this list
+# would need an expansion.
+SupportedPyVersions = Literal[39, 310, 311, 312, 313, 314, 315, 316, 317]
+
+
 class PipBinaryFilters(BinaryModeOptions):
     """Binary filters specific to pip packages."""
 
     arch: BinaryFilterStr = "x86_64"
     os: BinaryFilterStr = "linux"
-    py_version: int | None = None
+    py_version: SupportedPyVersions | None = None
     py_impl: BinaryFilterStr = "cp"
     abi: BinaryFilterStr = BINARY_FILTER_ALL
-    platform: str | None = None
+    platform: PlatformRegexStr | None = None
 
     @pydantic.model_validator(mode="after")
     def _validate_platform_exclusivity(self) -> Self:
@@ -242,7 +251,7 @@ class PipBinaryFilters(BinaryModeOptions):
 
     @pydantic.field_validator("platform")
     @classmethod
-    def _validate_platform(cls, value: str | None) -> str | None:
+    def _validate_platform(cls, value: PlatformRegexStr | None) -> PlatformRegexStr | None:
         if value is None:
             return value
 
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -138,16 +138,34 @@ class SSLOptions(pydantic.BaseModel, extra="forbid"):
     """SSL options model.
 
     Defines extra options fields for client TLS authentication.
+
+    Mandates that if defined then both client_cert and client_key must be files on the
+    file system. When the files are present an object could be instantiated like this:
+
+    >>> SSLOptions(client_cert=Path("/tmp/foo"), client_key=Path("/tmp/bar"))  # doctest: +SKIP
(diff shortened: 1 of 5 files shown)


hermeto/core/models/input.py
```diff
@@ -138,16 +139,34 @@ class SSLOptions(pydantic.BaseModel, extra="forbid"):
...
+    client_key: Path | None = None
+    ca_bundle: Path | None = None
     ssl_verify: bool = True
 
     @pydantic.field_validator("client_key", "client_cert", "ca_bundle")
     @classmethod
-    def _validate_auth_file_paths(cls, val: str, info: pydantic.ValidationInfo) -> str | None:
+    def _validate_auth_file_paths(cls, val: Path, info: pydantic.ValidationInfo) -> Path | None:
         if val is None:
             return val
```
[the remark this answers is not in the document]
eskultety: Valid.

hermeto/core/models/input.py
```diff
@@ -356,12 +390,12 @@ class ExtraOptions(pydantic.BaseModel, extra="forbid"):
...
     """
 
-    dnf: dict[Literal["main"] | str, dict[str, Any]] | None = None
+    dnf: dict[Literal["main"] | str, dict[str, ConfigEntryValue]] | None = None
     ssl: SSLOptions | None = None
 
     @pydantic.model_validator(mode="before")
     @classmethod
-    def _validate_dnf_options(cls, data: Any) -> Any:
+    def _validate_dnf_options(cls, data: dict) -> dict:
```
mohamedbouchtout, line 398: Would it make sense here to specify the dict structure here even more? Like dict[str, Any]

hermeto/core/models/sbom.py
```diff
@@ -39,6 +40,9 @@ def datetime_to_iso_8601(value: datetime) -> str:
 ISODatetime = Annotated[datetime, pydantic.PlainSerializer(datetime_to_iso_8601)]
 SortedSet = Annotated[set[str], pydantic.PlainSerializer(sorted, return_type=list[str])]
 
+# Temporary alias until a purl string could be properly represented
+PURLString = str
```
mohamedbouchtout, line 44: It might be worth considering declaring this in the type_aliases.py and importing it sence PURLs are used across multiple backends and could benefit code readability

hermeto/core/models/input.py
a-ovchinnikov: Yes, this is the first step, in the long run there should be no `Any`s (or `str` for that matter, because `str` tells us not too much, `a-ovchinnikov` is a valid `str` and would pass the type check, but not too many users actually want to see this string in their programs). I do not think that `dict[str, Any]` is fundamentally more expressive than just `dict`, both are rather vague and the former also bring in another `Any` which I personally don't like, however they are still better than just `Any`.

hermeto/core/models/sbom.py
a-ovchinnikov: Yes, this is a correct observation. In fact there are several aliases and helper types which could be moved out for reuse. One day when I get back to this PR I'll eventually do that.

hermeto/core/models/input.py
```diff
@@ -110,6 +110,7 @@ def show_error(error: "ErrorDict") -> str:
     "pip",
     "rpm",
     "yarn",
+    "yarn-classic",
```
derasdf, line 113: "yarn-classic" is only used as a backend name in `create_backend_annotation` (yarn_classic/main.py:73), not as a user-facing input type. Adding it to `PackageManagerType` conflates backend names with input types. Would a separate `BackendName` type be cleaner?

hermeto/core/models/sbom.py
```diff
@@ -80,16 +84,12 @@ def to_annotations(
...
 
 
-a_backend_indicator = object()
-
-
 def convert_annotation_to_property(
     annotation: "SPDXPackageAnnotation",
-) -> Property | object:
+) -> Property:
     """Convert a SPDX annotation to a CycloneDX property or a backend indicator.
```
derasdf, line 90: nitpick: the docstring mentions "or a backend indicator" but that code path was removed in this PR.

hermeto/core/models/sbom.py
```diff
@@ -837,34 +837,32 @@ def to_cyclonedx(self) -> Sbom:
...
+
+
+def separate_backend_annotation_from_others(
+    annotations: Iterable[SPDXPackageAnnotation],
+) -> tuple[SPDXPackageAnnotation, list[SPDXPackageAnnotation]]:
+    """Separate backend annotation from properties annotations."""
+    is_backend = lambda ann: ann.annotator.endswith(ANNOTATOR_BACKEND)
+    backends, others = partition(is_backend, annotations)
+    # There will always be just one backend, so it has to be extracted here:
+    return backends[0], others
```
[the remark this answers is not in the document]
derasdf: May be valid

hermeto/core/models/input.py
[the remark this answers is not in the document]
derasdf: +1

hermeto/core/models/input.py
```diff
@@ -345,6 +373,12 @@ class PnpmPackageInput(_PackageInputBase):
     type: Literal["x-pnpm"]
 
 
+ConfigEntryValue = TypeAliasType(
+    "ConfigEntryValue",
+    'str | int | float | bool | None | list["ConfigEntryValue"] | dict[str, "ConfigEntryValue"]',
+)
```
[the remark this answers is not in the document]
derasdf: +1

derasdf (commented):
Mostly LGTM with 2 comments (and concerns flagged by gemini).

a-ovchinnikov:
This has to be adjusted to current state of repository, closing.

PR #1642, opened 2026-06-30
enhancement(python): Moved pip backend and it's unit tests to a /python directory

### LINKED ISSUE #1548: Group Python backends in a shared directory
This is a tracker issue to group all Python backends in a shared directory so they can share common functionality. Currently, there is only `pip`, but we plan to support `uv` in the future.

--
https://github.com/hermetoproject/hermeto/pull/1431#discussion_r2979917187

mohamedbouchtout:
resolves #1548
### Summery
Created a /python directory for the `pip` backend and `pip` unit tests as requested in this [issue](https://github.com/hermetoproject/hermeto/issues/1548), similar to what was done with the [javascript](https://github.com/hermetoproject/hermeto/pull/1591) package managers. Updated import paths to match the relocation.

### Details
1. Moved the `pip` backend to a /python directory
2. Moved the `pip` unit tests to a /python directory to match the backend structure
3. Future python package managers will be placed in the /python directory (e.g, `uv`)


slimreaper35 (approved):

hermeto/core/package_managers/python/pip/__init__.py
```diff
@@ -0,0 +1,4 @@
+# SPDX-License-Identifier: GPL-3.0-only
```
eskultety, line 1: **Commit message:**
try to keep the commit subject and body under 72 chars ;)

> ; updated import paths accordingly

No need to have this in the commit subject ;).

> as requested in issue https://github.com/hermetoproject/hermeto/issues/1548

The way you normally do this in a commit is:
```
Closes <issue URL>
```
and that's it.

eskultety (commented):
One "mentoring" comment left as an improvement item ;).  ACK to the change.

hermeto/core/package_managers/python/pip/__init__.py
slimreaper35: Also, I think using imperative is more common as well. Moved -> Move ...

mohamedbouchtout: Thanks for the tips!

tnevrlka (commented):
Code changes LGTM, few style notes that I would like to see addressed before merging:
- Consider using the 50/72 rule -- try to fit the title under 50 chars and keep lines in a body under 72 chars, then wrap
- it's -> its 
- Use imperative, e.g. Moved -> Move

eskultety:
> Code changes LGTM, few style notes that I would like to see addressed before merging:
> 
>     * Consider using the 50/72 rule -- try to fit the title under 50 chars and keep lines in a body under 72 chars, then wrap 

While I acknowledge ^this to be the official recommendation (even by the git book itself?) even I don't follow the 50c rule on the subject line for 2 reasons:

1. 50c is quite limiting in delivering good descriptive subjects (let alone using conventional commits syntax..)
2. I didn't find it worth my time figuring out whether I could have a floating wrap column in my vim based setup on the git file type and so I only set it to 72 globally for git related stuff.  I would not be fond of enforcing 50c which would make my vim life harder, but should it become a rule, I will respect it.
 
That said, even 80c is nowadays deemed restrictive, but that's because many people fancy an ultrawide screen and working in an IDE. For those who still rock terminal setup in 24" fullHD screens 80-100c is still quite vital :), my 2 cents.

eskultety (approved):
Updated the commit message according to the review comments. ACK.

PR #1653, opened 2026-07-07
Deprecate the http_requests.py module

### LINKED ISSUE #1609: Drop or populate http_requests.py module
After some recent refactoring, the [module](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/http_requests.py) is almost empty. It contains a leftover `logging` import. Dead `ALL_REQUEST_METHODS` constant, `SAFE_REQUEST_METHODS,` and `DEFAULT_RETRY_OPTIONS` are imported only to [general.py](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/general.py#L19) module.

mohamedbouchtout:
Closes #1609 

### Summery
The [http_requests.py module](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/http_requests.py) is almost empty after some recent refactoring. The module is only being imported in [general.py module](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/general.py#L19) and being used in the `async_download_files()` and `download_binary_file()` methods (which explains why all HTTP request calls are being made here). Due to this, it's better to delete `http_requests.py` and moving what's inside to `general.py`.

### Details
1. Delete the `http_requests.py` module
2. Move only the imported variables to `general.py` (`SAFE_REQUEST_METHODS` and `DEFAULT_RETRY_OPTIONS`)


hermeto/core/package_managers/general.py
```diff
@@ -16,15 +16,17 @@
...
 from hermeto.core.type_aliases import StrPath
 
 _pkg_requests_session: requests.Session | None = None
 
+SAFE_REQUEST_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE"})
+DEFAULT_RETRY_OPTIONS: dict[str, Any] = {
+    "backoff_factor": 1.3,
+    "status_forcelist": (500, 502, 503, 504),
+}
+
```
slimreaper35, line 27: I would go a bit further and make these constants more granular (e.g. `BACKOFF_FACTOR = 1.3`) and pass them as keyword arguments to the functions in the module instead of dict unpacking.

a-ovchinnikov: That would increase the number arguments for functions without a clear reason: there is no plan to use different values anytime soon. Moreover, there are just four places within this module where this name is used, I suggest to not overcomplicate the code.

slimreaper35: I think it simplifies the code. For example here is one constant used directly and other are unpacked:

```python
retry_options = {
    **DEFAULT_RETRY_OPTIONS,
    "allowed_methods": SAFE_REQUEST_METHODS,
```

Yes, it adds more number of arguments, but also `retry_option` variable can be deleted so the function length does not really grow.

a-ovchinnikov (approved):

hermeto/core/package_managers/general.py
a-ovchinnikov: Ok, I am fine with the change both ways.

mohamedbouchtout: I agree with breaking it down to smaller pieces to avoid creating and unpacking unnecessary dicts.

slimreaper35 (approved):

PR #1655, opened 2026-07-07
feat(bundler): checksum validation for bundler

### LINKED ISSUE #1652: Validate bundler checksums from Gemfile.lock
Hermeto bundler backend downloads gems from registries but does not yet verify them against checksums declared in Gemfile.lock.

The [checksum](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/package_managers/bundler/gem_models.py#L55) attribute is set to `None` and is not overwritten anywhere, basically a dead code.

Bundler has supported an optional `CHECKSUMS` section in Gemfile.lock since v2.6 [1] [2].

We originally deferred the implementation because the feature was still opt-in ([design](https://github.com/hermetoproject/hermeto/blob/main/docs/design/bundler.md#dependency-checksums)), but it is now stable enough to support it in Hermeto. It is possible to regenerate Gemfile.lock by running `bundle lock --add-checksums`.

Our integration tests still use old version so it will be necessary to update the `BUNDLED WITH` section as well.

[1]: https://blog.rubygems.org/2024/12/19/bundler-v2-6.html
[2]: https://rubygems.org/gems/bundler/versions

Harsh-2005d:
**Description:**

Closes #1652

## Summary

The `checksum` field on `GemDependency` was always `None` — the lockfile parser never populated it and it was never used for verification. This PR wires up the full checksum pipeline for RubyGems dependencies.

## Changes

**`lockfile_parser.rb`**
- Extract checksums from the `CHECKSUMS` section via `spec.source.checksum_store.to_lock(spec)`
- Emit them in `sha256:<hexdigest>` format in the JSON output

**`gem_models.py`**
- Add checksum qualifier to `GemDependency.purl` for SBOM output

**`main.py`**
- After downloading `.gem` files, verify each against its lockfile checksum using `must_match_any_checksum`
- If no checksum is present in the lockfile, log a warning and skip verification

### DIFF
--- a/hermeto/core/package_managers/bundler/main.py
+++ b/hermeto/core/package_managers/bundler/main.py
@@ -10,6 +10,7 @@ import aiohttp
 from packageurl import PackageURL
 
 from hermeto import APP_NAME
+from hermeto.core.checksum import ChecksumInfo, must_match_any_checksum
 from hermeto.core.config import get_config
 from hermeto.core.constants import Mode
 from hermeto.core.errors import NotAGitRepo, PackageRejected, UnsupportedFeature
@@ -142,6 +143,24 @@ def _download_gems(
             headers=headers,
         )
     )
+
+    verify_checksums(gem_deps, deps_dir)
+
+
+def verify_checksums(
+    gem_deps: list[GemDependency],
+    deps_dir: RootedPath,
+) -> None:
+    """Verify checksums for downloaded gems."""
+    for dep in gem_deps:
+        if dep.checksum is None:
+            log.warning("No checksum found for %s-%s, skipping verification", dep.name, dep.version)
+            continue
+
+        must_match_any_checksum(
+            dep.download_location(deps_dir).path,
+            [ChecksumInfo.from_hash(dep.checksum)],
+        )
 
 
 def _clone_git_deps(
--- a/hermeto/core/package_managers/bundler/main.py
+++ b/hermeto/core/package_managers/bundler/main.py
@@ -10,6 +10,7 @@ import aiohttp
 from packageurl import PackageURL
 
 from hermeto import APP_NAME
+from hermeto.core.checksum import ChecksumInfo, must_match_any_checksum
 from hermeto.core.config import get_config
 from hermeto.core.constants import Mode
 from hermeto.core.errors import NotAGitRepo, PackageRejected, UnsupportedFeature
@@ -143,6 +144,23 @@ def _download_gems(
             headers=headers,
         )
     )
+
+    _verify_checksums(gem_deps, deps_dir)
+
+
+def _verify_checksums(
+    gem_deps: list[GemDependency],
+    deps_dir: RootedPath,
+) -> None:
+    for dep in gem_deps:
+        if dep.checksum is None:
+            log.warning("No checksum found for %s-%s, skipping verification", dep.name, dep.version)
+            continue
+
+        must_match_any_checksum(
+            dep.download_location(deps_dir).path,
+            [ChecksumInfo.from_hash(checksum) for checksum in dep.checksum.split()],
+        )
 
 
 def _clone_git_deps(
--- a/hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
+++ b/hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
@@ -11,12 +11,29 @@ parsed_specs = []
 lockfile_parser.specs.each do |spec|
     case spec.source
     when Bundler::Source::Rubygems
+      # Requires Bundler 2.5.0+ to resolve field.
+      # e.g. input:  gem_lock_name sha256=hash,...
+      #      output: sha256:hash sha....
+      lock_content = spec.source.checksum_store.to_lock(spec)
+      pattern = /^#{Regexp.escape(spec.lock_name)} (.*)/
+
+      checksums_raw = lock_content[pattern, 1]
+      if checksums_raw == nil
+        checksum = nil
+      else
+        checksum = checksums_raw
+          .split(",")
+          .map { |c| c.sub("=", ":") }
+          .join(" ")                         
+      end
+
       parsed_spec = {
         name: spec.name,
         version: spec.version,
         type: 'rubygems',
         source: spec.source.remotes.first,
-        platforms: [spec.platform]
+        platforms: [spec.platform],
+        checksums: checksum.nil? ? {} : { spec.platform => checksum }
       }
 
       existing_spec = parsed_specs.find { |s|
@@ -27,8 +44,9 @@ lockfile_parser.specs.each do |spec|
       }
 
       if existing_spec
-        # extend the platforms array
-        existing_spec[:platforms] << parsed_spec[:platforms].first
+        # extend the platforms array and checksums dictionary
+        existing_spec[:platforms] |= parsed_spec[:platforms]
+        existing_spec[:checksums].merge!(parsed_spec[:checksums])
       else
         parsed_specs << parsed_spec
       end
@@ -57,4 +75,4 @@ lockfile_parser.specs.each do |spec|
 puts JSON.pretty_generate({ bundler_version: lockfile_parser.bundler_version, dependencies: parsed_specs })
 
 # References:
-# https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb
+# https://github.com/ruby/rubygems/blob/master/lib/bundler/lockfile_parser.rb
--- a/hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
+++ b/hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
@@ -11,12 +11,14 @@ parsed_specs = []
 lockfile_parser.specs.each do |spec|
     case spec.source
     when Bundler::Source::Rubygems
+      checksum = spec.source.checksum_store.to_lock(spec)[/sha[^, ]*/]&.sub("=",":")
       parsed_spec = {
         name: spec.name,
         version: spec.version,
         type: 'rubygems',
         source: spec.source.remotes.first,
-        platforms: [spec.platform]
+        platforms: [spec.platform],
+        checksums: checksum.nil? ? {} : { spec.platform => checksum }
       }
 
       existing_spec = parsed_specs.find { |s|
@@ -27,8 +29,9 @@ lockfile_parser.specs.each do |spec|
       }
 
       if existing_spec
-        # extend the platforms array
-        existing_spec[:platforms] << parsed_spec[:platforms].first
+        # extend the platforms array and checksums dictionary
+        existing_spec[:platforms] |= parsed_spec[:platforms]
+        existing_spec[:checksums].merge!(parsed_spec[:checksums])
       else
         parsed_specs << parsed_spec
       end
@@ -57,4 +60,4 @@ lockfile_parser.specs.each do |spec|
 puts JSON.pretty_generate({ bundler_version: lockfile_parser.bundler_version, dependencies: parsed_specs })
 
 # References:
-# https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb
+# https://github.com/ruby/rubygems/blob/master/lib/bundler/lockfile_parser.rb
--- a/tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/Containerfile
+++ b/tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/Containerfile
@@ -1,0 +1,12 @@
+FROM mirror.gcr.io/ruby:3.3
+
+RUN if curl -IsS www.google.com; then echo "Has network access!"; exit 1; fi
+
+WORKDIR /app
+
+COPY Gemfile .
+COPY Gemfile.lock .
+
+RUN bundle install --local
+
+CMD ["rails", "-v"]
--- a/tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/README.md
+++ b/tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/README.md
@@ -1,0 +1,2 @@
+# hermeto-testing
+Tests bundler checksum mismatch and expects ERR_CHECKSUM_VERIFICATION_FAILED
--- a/tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/README.md
+++ b/tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/README.md
@@ -1,0 +1,3 @@
+# hermeto-testing
+Tests bundler checksum mismatch and expects ERR_CHECKSUM_VERIFICATION_FAILED
+incorrect checksum in racc gem
--- a/tests/integration/bundler/scenarios/bundler_e2e_ruby33/in/Gemfile.lock
+++ b/tests/integration/bundler/scenarios/bundler_e2e_ruby33/in/Gemfile.lock
@@ -88,6 +88,7 @@ GEM
     concurrent-ruby (1.3.4)
     crass (1.0.6)
     date (3.3.4)
+    drb (2.2.3)
     erubi (1.13.0)
     globalid (1.2.1)
       activesupport (>= 6.1)
@@ -104,7 +105,10 @@ GEM
     marcel (1.0.4)
     method_source (1.1.0)
     mini_mime (1.1.5)
-    minitest (5.25.1)
+    mini_portile2 (2.8.9)
+    minitest (6.0.6)
+      drb (~> 2.0)
+      prism (~> 1.5)
     net-imap (0.4.16)
       date
       net-protocol
@@ -115,6 +119,9 @@ GEM
     net-smtp (0.5.0)
       net-protocol
     nio4r (2.7.3)
+    nokogiri (1.16.7)
+      mini_portile2 (~> 2.8.2)
+      racc (~> 1.4)
     nokogiri (1.16.7-aarch64-linux)
       racc (~> 1.4)
     nokogiri (1.16.7-arm-linux)
@@ -127,6 +134,7 @@ GEM
       racc (~> 1.4)
     nokogiri (1.16.7-x86_64-linux)
       racc (~> 1.4)
+    prism (1.9.0)
     public_suffix (6.0.1)
     racc (1.8.1)
     rack (2.2.9)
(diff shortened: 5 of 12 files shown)


hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -57,7 +57,10 @@ class GemDependency(_GemMetadata):
     @cached_property
     def purl(self) -> str:
         """Get PURL for this dependency."""
-        purl = PackageURL(type="gem", name=self.name, version=self.version)
+        qualifiers = {}
+        if self.checksum is not None:
+            qualifiers["checksum"] = self.checksum
+        purl = PackageURL(type="gem", name=self.name, version=self.version, qualifiers=qualifiers)
```
eskultety: It would probably be for the best if we skipped this PURL handling because of https://github.com/hermetoproject/hermeto/issues/852, more specifically https://github.com/hermetoproject/hermeto/issues/852#issuecomment-2922124956. Sadly, it seems like the discussion there hasn't been closed yet and a consensus hence reached.

hermeto/core/package_managers/bundler/main.py
```diff
@@ -122,6 +123,17 @@ def _resolve_bundler_package(
...
+            if not isinstance(dep, (GemDependency, GemPlatformSpecificDependency)):
+                continue
+            if dep.checksum is None:
+                log.warning(
+                    "No checksum found for %s-%s, skipping verification", dep.name, dep.version
+                )
+                continue
+            must_match_any_checksum(
+                dep.download_location(deps_dir).path, [ChecksumInfo.from_hash(dep.checksum)]
+            )
```
eskultety: I think this would warrant a test, a unit test or a scoped integration test (depending on how much mocking would be necessary for the unit test), we most certainly have a precedent also in the integration test suite covering simple cases like this one.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -11,12 +11,14 @@
 lockfile_parser.specs.each do |spec|
     case spec.source
     when Bundler::Source::Rubygems
+      checksum = spec.source.checksum_store.to_lock(spec)[/sha256=\S+/]&.sub("=", ":")
```
eskultety: `checksum_store` is what bundler 2.5.0+ or even 2.6.0+? I'm okay if this breaks with old bundlers because our image should contain the necessary version, but I'd add a commentary on the minimum version required, OR maybe we can guard this and recover from an error on old bundlers and set checksum to nil that is also an option that shouldn't be expensive to add.

hermeto/core/package_managers/bundler/main.py
```diff
@@ -122,6 +123,17 @@ def _resolve_bundler_package(
                 concurrency_limit=get_config().runtime.concurrency_limit,
             )
         )
+        for dep in dependencies:
+            if not isinstance(dep, (GemDependency, GemPlatformSpecificDependency)):
```
eskultety: `GemPlatformSpecificDependency` is a subclass of `GemDependency` so this explicit tuple is unnecessary.
Additionally, you could probably encode ^this check into a generator expression right in the for loop as long as it fits a single line without line wrapping.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
eskultety: Seems like for platform package variants, e.g. `nokogiri`, we'd be only recording the first checksum and produce a single parsed_spec with multiple platforms, but a single checksum, discarding the rest IIUC. That would be a deal breaker - emphasized my request for test coverage and this very hunk justifies an integration test rather than a unit test.

eskultety: Can more than a single checksum be recorded in Gemfile.lock? Because the regex seems to assume that only a single sha256 checksum would be present, but doesn't back that decision with docs or evidence, so seems a little fragile to me.

eskultety (changes requested):

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
Harsh-2005d: > `checksum_store` is what bundler 2.5.0+ or even 2.6.0+?

this was introduced [here](https://github.com/ruby/rubygems/commit/86c7084e1ca09c59f99afbea4dbb96b9c9f8149d) which i beleive is 2.5.0+.

>  for platform package variants

the packages are marked with platform hence a single unique checksum . [lock_name](https://ruby-doc.org/3.3.6/stdlibs/bundler/Gem/NameTuple.html) which is used in [to_lock](https://ruby-doc.org/3.3.6/stdlibs/bundler/Bundler/Checksum/Store.html#method-i-store_checksum)

> Can more than a single checksum be recorded in Gemfile.lock?

Yes multiple algorithms can be recorded which are comma separated for a gem.(e.g. sha256=..,sha512=.....). but as of now the lockfile generator defaults to sha256 with no way of configuring multiple algos. [lockfile_generator](https://github.com/ruby/rubygems/blob/08548126d5b58c8c8accdd96063635cd01760115/lib/bundler/lockfile_generator.rb#L114) and [checksum](https://github.com/ruby/rubygems/blob/08548126d5b58c8c8accdd96063635cd01760115/lib/bundler/checksum.rb#L12)

slimreaper35:
Let's also verify this change with updating our bundler/e2e test.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
eskultety: > Yes multiple algorithms can be recorded which are comma separated for a gem.(e.g. sha256=..,sha512=.....). but as of now the lockfile generator defaults to sha256 with no way of configuring multiple algos

ACK, but looking at https://github.com/ruby/rubygems/blob/a8e3c298bf9301a7b17b2c701d6799131301225f/lib/bundler/checksum.rb#L230 can we future proof this proposal just a tiny bit and split the spec by `,` first at least before taking the first SHA?

Harsh-2005d: @eskultety existing merge logic for parsed spec is insuffecient with introduction of checksums. I am thinking we can flatten the platforms array, then every package will have a separate identity for its platform which will integrate well with existing structure.

Harsh-2005d: But I think this will lead to redundancy in the output. Another option can be to instead make checksums a dictionary with platform as its key.

slimreaper35:
@Harsh-2005d Could you verify the integration test is passing? By temporary changing the repo-url for bundler/e2e.

Harsh-2005d:
> @Harsh-2005d Could you verify the integration test is passing? By temporary changing the repo-url for bundler/e2e.

there are assertion errors in test data for some packages, and if i regenerate test data from my repo url it introduces some unintended changes .
```sh
"pkg:gem/quux@0.0.1?vcs_url=git%2Bhttps://github.com/Harsh-2005d/integration-tests.git%40515cd9a5f1a688fdaeda58b50822ce266fe19025#quux",
```

slimreaper35:
> there are assertion errors in test data for some packages, and if i regenerate test data from my repo url it introduces some unintended changes .

That's fine. As I said, it's only temporary. After merging the PR in the integration tests you revert it back.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
eskultety: > Another option can be to instead make checksums a dictionary with platform as its key.

Sounds reasonable to me.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -57,4 +60,4 @@
 puts JSON.pretty_generate({ bundler_version: lockfile_parser.bundler_version, dependencies: parsed_specs })
 
 # References:
-# https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb
+# https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb
```
slimreaper35:
```suggestion
https://github.com/ruby/rubygems/blob/master/lib/bundler/lockfile_parser.rb
```

hermeto/core/package_managers/bundler/main.py
```diff
@@ -142,6 +143,20 @@ def _download_gems(
...
+                "No checksum found for %s-%s, skipping verification",
+                dep.name,
+                dep.version,
+            )
+            continue
+
+        must_match_any_checksum(
+            dep.download_location(deps_dir).path,
+            [ChecksumInfo.from_hash(dep.checksum)],
+        )
```
slimreaper35, line 163: Can we make this a separate function?
Also, the log.warning could fit on one line.

hermeto/core/package_managers/bundler/main.py
```diff
@@ -143,7 +143,7 @@ def _download_gems(
             headers=headers,
         )
     )
-    
+
```
slimreaper35, line 146: Irrelevant change.

tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/tmp.gemspec
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Gem::Specification.new do |spec|
+  spec.name = "tmp"
+  spec.version = "0.1.2"
+  spec.authors = ["Michal Šoltis"]
+  spec.email = ["msoltis@redhat.com"]
```
slimreaper35: Was this required 😄 by Ruby ?

tests/integration/bundler/scenarios/bundler_e2e_ruby40/out/bom.json
```diff
@@ -23,6 +23,7 @@
         "pkg:gem/concurrent-ruby@1.3.4",
         "pkg:gem/crass@1.0.6",
         "pkg:gem/date@3.3.4",
+        "pkg:gem/drb@2.2.3",
```
slimreaper35, line 26: These update should be a separate commit. Not mixing with checksum mismatch.

tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/README.md
```diff
@@ -0,0 +1,2 @@
+# cachi2-testing
+Various correct and malformed structures to be processed by cachi2
```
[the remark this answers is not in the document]
slimreaper35: Yes please. No new cachi2 references.

tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/tmp.gemspec
Harsh-2005d: the tmp gem is not necessary since the test fails regardless. Should i drop it?

slimreaper35: yes

tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/Containerfile
```diff
@@ -0,0 +1,13 @@
+FROM docker.io/library/ruby:latest
```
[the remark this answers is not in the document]
slimreaper35: Tracked in https://github.com/hermetoproject/hermeto/issues/1682

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -152,6 +152,7 @@ def test_parse_gemlock(
             "type": "rubygems",
             "source": "https://rubygems.org/",
             "platforms": ["ruby"],
+            "checksums": {},
             **base_dep,
```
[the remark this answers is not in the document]
Harsh-2005d: @slimreaper35 do we need the unit tests since integration tests already verify the behavior?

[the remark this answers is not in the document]
slimreaper35: @Harsh-2005d I don't think so, maybe you can add checksum field to existing unit test, e.g. `test_parse_gemlock`.

Harsh-2005d:
@slimreaper35 @eskultety ready for review. Thanks.

slimreaper35:
@Harsh-2005d Functionally, it looks good to me, but I am sure @eskultety will point out that commit 2/3 does not pass integration tests because of SBOM mismatch that is fixed in commit 3/3.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -57,4 +60,4 @@
 puts JSON.pretty_generate({ bundler_version: lockfile_parser.bundler_version, dependencies: parsed_specs })
 
 # References:
-# https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/lockfile_parser.rb
+# https://github.com/ruby/rubygems/blob/master/lib/bundler/lockfile_parser.rb
```
eskultety: new-line at the end of file was removed - was this change generated by an AI tool in by any chance? Editors normally don't do that, but AI tools notoriously do this.

tests/integration/bundler/scenarios/bundler_e2e_ruby40/out/bom.json
```diff
@@ -757,4 +799,4 @@
   },
   "specVersion": "1.6",
   "version": 1
-}
+}
```
eskultety: EOF new line removed.

tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/Containerfile
```diff
@@ -0,0 +1,12 @@
+FROM mirror.gcr.io/ruby:3.3
```
eskultety: **Commit message:**

> incorrect checksum in racc gem

nitpick: I think for a commit message this should be formatted a bit more nicely. Then again, I also think this particular bit of info should go to the test README.

eskultety: Do we need a Containerfile if this is not an e2e scenario which is supposed to fail during fetch?

tests/integration/bundler/scenarios/bundler_e2e_ruby33/in/Gemfile.lock
```diff
@@ -193,5 +201,68 @@ DEPENDENCIES
   rails (= 6.1.7)
   tmp!
 
+CHECKSUMS
```
eskultety, line 204: nitpick: this existing e2e update should IMO be in a separate commit to isolate the changes for a new test and the existing one. Structuring like that really speeds up the review process.

eskultety:
> @Harsh-2005d Functionally, it looks good to me, but I am sure @eskultety will point out that commit 2/3 does not pass integration tests because of SBOM mismatch that is fixed in commit 3/3.

Yes, I will :) .

tests/integration/bundler/scenarios/bundler_e2e_ruby33/out/bom.json
```diff
@@ -757,4 +799,4 @@
   },
   "specVersion": "1.6",
   "version": 1
-}
+}
```
eskultety: EOF trailing new line missing.

hermeto/core/package_managers/bundler/main.py
```diff
@@ -143,6 +144,24 @@ def _download_gems(
         )
     )
 
+    verify_checksums(gem_deps, deps_dir)
+
+
+def verify_checksums(
```
eskultety: nitpick: seems to be just a module level helper -> adding a `_` prefix seems appropriate.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -11,12 +11,14 @@
 lockfile_parser.specs.each do |spec|
     case spec.source
     when Bundler::Source::Rubygems
+      checksum = spec.source.checksum_store.to_lock(spec)[/sha[^, ]*/]&.sub("=",":")
```
eskultety: https://github.com/hermetoproject/hermeto/pull/1655#discussion_r3552010743
>...
> but I'd add a commentary on the minimum version required

I don't see any mention of this breaking with old bundlers which acknowledge is fine for our use case. Speaking of commentaries, we could also mention what output (i.e. source string) this matches against so that our future selves can quickly refer to this if Bundler changes the output making the regex return `nil`

eskultety (commented):
Mostly LGTM. One more respin and I think we're good to go.

tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/Containerfile
Harsh-2005d: i tested and it works fine without it. will drop that.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
Harsh-2005d: It was when I copy pasted the URL :)

slimreaper35 (approved):

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -11,12 +11,17 @@
 lockfile_parser.specs.each do |spec|
     case spec.source
     when Bundler::Source::Rubygems
+      # Requires Bundler 2.5.0+ to resolve field.
+      # e.g. input:  sha256=hash
+      #      output: sha256:hash
+      checksum = spec.source.checksum_store.to_lock(spec)[/sha[^, ]*/]&.sub("=", ":")
```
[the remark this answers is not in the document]
Harsh-2005d: as per bundler source [to_lock](https://github.com/ruby/rubygems/blob/master/lib/bundler/checksum.rb#L226) never returns nil.

taylormadore: What happens when `sha` is in the name of the Gem?

Harsh-2005d: interesting edge case. The pattern is
```ruby
#{lock_name} #{checksums.values.map(&:to_lock).sort.join(",")}
```
maybe a more hardened regex would be to check for preceding space and digit after "sha". Does this sound good?

taylormadore: Since we know the string returned from to_lock will have a prefix of spec.lock_name, I'd suggest stripping that prefix off and then dealing with the remainder (if any) as a comma-separated string of algo=checksum (s).

I'm also now wondering now if just taking the first checksum is OK or if we should handle the multiple checksums case. I'd be fine with just taking the first but adding a comment noting that multiple is possible.

Harsh-2005d:
New changes update the regex matching and add support for multiple checksums for Bundler hash verification. Although Bundler currently does not configure multiple checksum algorithms, this keeps the implementation future-proof if support is added later.

tests/integration/bundler/scenarios/bundler_checksum_mismatch/in/Containerfile
slimreaper35: @Harsh-2005d The Containerfile still exists

taylormadore:
@Harsh-2005d Can you rebase on main to pick-up a change to the CI?

tests/unit/package_managers/bundler/test_parser.py
```diff
@@ -162,6 +162,7 @@ def test_parse_gemlock(
     caplog: pytest.LogCaptureFixture,
 ) -> None:
     base_dep: dict[str, str] = sample_parser_output["dependencies"][0]
+    mocked_checksum: str = "sha256:bd2d213996ff7b3b364cd342a585fbee9797dbc1c0c6d868dc4150cc75739781"
```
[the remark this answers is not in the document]
slimreaper35: @coderabbitai Don't suggest type annotations for such trivial variables. It's Python, not Java. Learn this lesson please.

slimreaper35 (approved):

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -11,12 +11,18 @@
 lockfile_parser.specs.each do |spec|
     case spec.source
     when Bundler::Source::Rubygems
+      # Requires Bundler 2.5.0+ to resolve field.
+      # e.g. input:  gem_lock_name sha256=hash,...
+      #      output: sha256:hash sha....
+      checksums = spec.source.checksum_store.to_lock(spec)[/^#{Regexp.escape(spec.lock_name)} (.*)/,1]
+      checksum = checksums&.split(",")&.map { |checksum| checksum.sub("=", ":") }&.join(" ")
```
eskultety: This is quite cryptic. I understand it may be idiomatic in Ruby, but we don't write Ruby on a daily basis and ^this just reads as noise. Please open code it using intermediate variables for better readability.
Out of curiosity - did you write the idiomatic Ruby code yourself or did an AI agent help out? :) If so, please disclose in the commit.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
eskultety: >I'd be fine with just taking the first but adding a comment noting that multiple is possible.

@Harsh-2005d ^this still hasn't been addressed.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
Harsh-2005d: > did you write the idiomatic Ruby code

It was assisted by claude. will improve readability and disclose it in the commit too.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
Harsh-2005d: @eskultety We no longer take just the first hash. the updated regex takes in all comma separated hashes. For eg. 
```python
sha512=xyz,sha256=abc
```
the scripts reads it as 
```python
sha512:xyz sha256:abc
```
which is then mapped against platform and all and this format is a valid input to `must_match_any_checksum()`

Worth to note that bundler does not do multiple hashes right now (IIUC). but there code do seems to support it in [here](https://github.com/ruby/rubygems/blob/master/lib/bundler/checksum.rb#L230)

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -11,12 +11,29 @@
 lockfile_parser.specs.each do |spec|
     case spec.source
     when Bundler::Source::Rubygems
+      # Requires Bundler 2.5.0+ to resolve field.
```
eskultety, line 14: > Assisted by Claude.

How about this :)
https://github.com/hermetoproject/hermeto/blob/main/AI_CONTRIBUTION_POLICY.md#disclosure-requirements

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
```diff
@@ -11,12 +11,29 @@
...
+
+      checksums_raw = lock_content[pattern, 1]
+      if checksums_raw
+        checksum = checksums_raw
+          .split(",")
+          .map { |c| c.sub("=", ":") }
+          .join(" ")
+      else
+        checksum=nil                         
+      end
```
eskultety: please invert the conditional to make the else clause larger and hence easier to read.

eskultety (approved):
A couple more nitpicks, but ACK.

hermeto/core/package_managers/bundler/scripts/lockfile_parser.rb
Harsh-2005d: Sure. Sorry for that.

taylormadore (approved):

tnevrlka (approved):

PR #1657, opened 2026-07-08
fix(sbom): preserve pedigree patches in SPDX conversion

### LINKED ISSUE #1583: SPDX SBOM does not include pedigree patches
Hermeto records patch provenance for patched dependencies in CycloneDX via pedigree patches, but that metadata is not carried over when the SBOM is converted to SPDX format, and the patch information is effectively dropped.

SPDX v2 specification defines a similar way to report patches as the `externalRefs` object. See [spdx-spec/v2.3](https://spdx.github.io/spdx-spec/v2.3/how-to-use/#k17-linking-to-a-code-fix-for-a-security-issue).

--
https://github.com/hermetoproject/integration-tests/tree/yarn/v4

ST3V1K:
Patch provenance recorded in CycloneDX pedigree was silently lost when converting to SPDX, making the SPDX SBOM incomplete for patched dependencies.

### Summary

- Add SPDXPackageExternalRefSecurityFix model for SPDX v2.3 fix external refs
- Emit pedigree patches as SECURITY/fix external refs in to_spdx()
- Reconstruct Pedigree from fix external refs in to_cyclonedx()

Fixes https://github.com/hermetoproject/hermeto/issues/1583

### DIFF
--- a/hermeto/core/models/sbom.py
+++ b/hermeto/core/models/sbom.py
@@ -403,11 +403,30 @@ class Sbom(pydantic.BaseModel):
                 is_proxy = lambda ref: ref.type == PROXY_REF_TYPE and ref.comment == PROXY_COMMENT
                 return sorted(ref.url for ref in component.external_references if is_proxy(ref))
 
+            def external_references(component: Component) -> list[SPDXPackageExternalRefType]:
+                purl_ref = SPDXPackageExternalRefPackageManagerPURL(
+                    referenceLocator=component.purl,
+                    referenceCategory="PACKAGE-MANAGER",
+                    referenceType="purl",
+                )
+
+                if component.pedigree is None:
+                    return [purl_ref]
+
+                patches_refs = [
+                    SPDXPackageExternalRefOtherPatch(
+                        referenceLocator=patch.diff.url,
+                        referenceCategory="OTHER",
+                        referenceType="patch",
+                    )
+                    for patch in component.pedigree.patches
+                ]
+
+                return [purl_ref, *patches_refs]
+
             packages = []
 
             hashdict = lambda c: dict(name=c.name, version=c.version, purl=c.purl)
-            erefbase = dict(referenceCategory="PACKAGE-MANAGER", referenceType="purl")
-            erefdict = lambda c: dict(referenceLocator=c.purl, **erefbase)
 
             for component in libraries:
                 package_hash = SPDXPackage._calculate_package_hash_from_dict(hashdict(component))
@@ -425,7 +444,7 @@ class Sbom(pydantic.BaseModel):
                         ),
                         name=component.name,
                         versionInfo=component.version,
-                        externalRefs=[erefdict(component)],
+                        externalRefs=external_references(component),
                         annotations=generate_package_annotations(
                             properties=component.properties, bom_ref=component.bom_ref
                         ),
@@ -526,6 +545,29 @@ class SPDXPackageExternalRefSecurityPURL(
     referenceType: Literal["cpe23Type"]
 
 
+class SPDXPackageExternalRefOther(SPDXPackageExternalRef):
+    """SPDX Package External Reference for category OTHER.
+
+    Compliant to the SPDX specification:
+    https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field
+    """
+
+    referenceCategory: Literal["OTHER"]
+
+
+class SPDXPackageExternalRefOtherPatch(
+    SPDXPackageExternalRefOther, SPDXPackageExternalRefReferenceLocatorURI
+):
+    """SPDX Package External Reference for category OTHER and custom type patch.
+
+    The "patch" referenceType is a hermeto-defined custom type, not part of the
+    SPDX specification. The OTHER category allows custom types per:
+    https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field
+    """
+
+    referenceType: Literal["patch"]
+
+
 SPDXPackageExternalRefPackageManagerType = Annotated[
     SPDXPackageExternalRefPackageManagerPURL,
     pydantic.Field(discriminator="referenceType"),
@@ -536,9 +578,16 @@ SPDXPackageExternalRefSecurityType = Annotated[
     pydantic.Field(discriminator="referenceType"),
 ]
 
+SPDXPackageExternalRefOtherType = Annotated[
+    SPDXPackageExternalRefOtherPatch,
+    pydantic.Field(discriminator="referenceType"),
+]
+
 
 SPDXPackageExternalRefType = Annotated[
-    SPDXPackageExternalRefPackageManagerType | SPDXPackageExternalRefSecurityType,
+    SPDXPackageExternalRefPackageManagerType
+    | SPDXPackageExternalRefSecurityType
+    | SPDXPackageExternalRefOtherType,
     pydantic.Field(discriminator="referenceCategory"),
 ]
 
@@ -563,6 +612,15 @@ class SPDXPackageAnnotation(pydantic.BaseModel):
 
 def _extract_purls(from_refs: list[SPDXPackageExternalRefType]) -> list[str]:
     return [ref.referenceLocator for ref in from_refs if ref.referenceType == "purl"]
+
+
+def _extract_pedigree(from_refs: list[SPDXPackageExternalRefType]) -> Pedigree | None:
+    patches = [
+        Patch(diff=PatchDiff(url=ref.referenceLocator))
+        for ref in from_refs
+        if ref.referenceCategory == "OTHER" and ref.referenceType == "patch"
+    ]
+    return Pedigree(patches=patches) if patches else None
 
 
 def _parse_purls(purls: list[str]) -> list[PackageURL]:
@@ -904,6 +962,7 @@ def _partial_component(
         version=package.versionInfo,
         properties=properties,
         external_references=external_references,
+        pedigree=_extract_pedigree(package.externalRefs),
     )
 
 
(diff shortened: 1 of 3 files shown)


ST3V1K:
The SPDX v2.3 externalRef does not have a field to carry the CycloneDX patch type (unofficial, backport, cherry-pick, monkey), so the conversion defaults to unofficial on round-trip. This is lossless today since both yarn and pnpm resolvers only produce unofficial patches. 
A workaround would be to encode the type as a URL fragment (e.g. `https://example.com/patch.diff#type=backport`), but that's non-standard.

hermeto/core/models/sbom.py
```diff
@@ -403,11 +403,30 @@ def source_infos(component: Component) -> list[str]:
...
+
+                if component.pedigree is None:
+                    return [purl_ref]
+
+                patches_refs = [
+                    SPDXPackageExternalRefSecurityFix(
+                        referenceLocator=patch.diff.url,
+                        referenceCategory="SECURITY",
+                        referenceType="fix",
+                    )
```
eskultety: Why are we classifying every pedigree patch as security? There was a discussion on this topic in the original PR: https://github.com/hermetoproject/hermeto/pull/1600#issuecomment-4622577942

Looking at the spec and more specifically at Annex F https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f2-security I guess I understand the motivation for recording everything as SECURITY because it has granular and structured types, but if you look closely then all `<type>`s talks specifically about security related metadata. My worry is that if we classify every patch as security then any advanced SBOM parser/analyzer will start flagging even non-security related patches as potentially fixing issues except that no CVE would be attached and hence I would prefer being more vague - `OTHER` - rather than boldly claiming something is a security fix when it is not. Note we do and will never have enough information to decide this classification by ourselves unless we're explicitly told.

hermeto/core/models/sbom.py
```diff
@@ -526,13 +545,26 @@ class SPDXPackageExternalRefSecurityPURL(
...
+    SPDXPackageExternalRefSecurity, SPDXPackageExternalRefReferenceLocatorURI
+):
+    """SPDX Package External Reference for category SECURITY and type fix.
+
+    Compliant to the SPDX specification:
+    https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field
+    """
+
+    referenceCategory: Literal["SECURITY"]
+    referenceType: Literal["fix"]
```
eskultety: The need for reporting pedigree info (original work in CycloneDX we had) was due to Yarn's built in downstream patches for typescript IIRC so the main use case we needed to cover had actually nothing to do with security.

hermeto/core/models/sbom.py
eskultety: To take this even further, you correctly discuss the mapping disparity between CycloneDX and SPDX, none of CycloneDX pedigree types hints at security, so it is definitely incorrect from us to assume something that will most likely be (we generated that CycloneDX ourselves) "unofficial" can suddenly become "SECURITY".

hermeto/core/models/sbom.py
```diff
@@ -403,11 +403,30 @@ def source_infos(component: Component) -> list[str]:
...
+                    return [purl_ref]
+
+                patches_refs = [
+                    SPDXPackageExternalRefSecurityFix(
+                        referenceLocator=patch.diff.url,
+                        referenceCategory="SECURITY",
+                        referenceType="fix",
+                    )
+                    for patch in component.pedigree.patches
+                ]
```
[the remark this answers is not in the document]
eskultety: > However, in CycloneDX, a patch's diff.url can be any string, including relative paths (e.g., "patches/fix-bug.patch") or local file paths. If a component contains a patch with a non-URI or relative URL, calling to_spdx() will raise a pydantic.ValidationError and crash the entire SBOM conversion process.

Technically, this is true in general, except that we generated that CycloneDX ourselves and the way we generated that URL is:
```
 def get_patch_url(key: str) -> str:
        try:
            subpath = patches[key]["path"]
        except (KeyError, TypeError):
            raise InvalidLockfileFormat(lockfile.path, f"Missing path for patched dependency {key}")

        return vcs_url + "#" + subpath
```

So, we could either just ignore this AI comment OR make our `PatchDiff` model stricter when it comes to the type for a URL since we only ever consume our own SBOMs (might be fragile in the future if we ever expand our scope).

eskultety (changes requested):
I don't think SPDX's `SECURITY` external reference field category is the best choice here.

hermeto/core/models/sbom.py
ST3V1K: I haven't looked at the original PR, and just went with what was in the issue. Now I have changed it OTHER/patch.

hermeto/core/models/sbom.py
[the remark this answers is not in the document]
ST3V1K: I'd just ignore this comment.

hermeto/core/models/sbom.py
```diff
@@ -526,6 +545,29 @@ class SPDXPackageExternalRefSecurityPURL(
...
+class SPDXPackageExternalRefOtherPatch(
+    SPDXPackageExternalRefOther, SPDXPackageExternalRefReferenceLocatorURI
+):
+    """SPDX Package External Reference for category OTHER and type patch.
+
+    Compliant to the SPDX specification:
+    https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field
+    """
+
+    referenceCategory: Literal["OTHER"]
```
slimreaper35: If the class inherits `SPDXPackageExternalRefOther`, we don't need to add again `referenceCategory: Literal["OTHER"]`.

ST3V1K: I did it to match the rest. e.g. 
```
class SPDXPackageExternalRefPackageManager(SPDXPackageExternalRef):
    """SPDX Package External Reference for category package-manager.

    Compliant to the SPDX specification:
    https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field
    """

    referenceCategory: Literal["PACKAGE-MANAGER"]


class SPDXPackageExternalRefPackageManagerPURL(
    SPDXPackageExternalRefPackageManager, SPDXPackageExternalRefReferenceLocatorURI
):
    """SPDX Package External Reference for category package-manager and type purl.

    Compliant to the SPDX specification:
    https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field
    """

    referenceCategory: Literal["PACKAGE-MANAGER"]
    referenceType: Literal["purl"]
 ```
 
 If you'd like, I can remove it.

Comment (slimreaper35) [reply hermeto/core/models/sbom.py]: 🤔 I don't know what was the intended approach for the rest, have `referenceCategory` everywhere or just inherit from the base class.

Comment (slimreaper35) [reply hermeto/core/models/sbom.py]: Anyway, I would drop it from `class SPDXPackageExternalRefOtherPatch`.

Comment (ST3V1K) [reply hermeto/core/models/sbom.py]: Done

File: hermeto/core/models/sbom.py [hunk 5]
```diff
@@ -526,6 +545,28 @@ class SPDXPackageExternalRefSecurityPURL(
...
+):
+    """SPDX Package External Reference for category OTHER and type patch.
+
+    Compliant to the SPDX specification:
+    https://spdx.github.io/spdx-spec/v2.3/package-information/#721-external-reference-field
+    """
+
+    referenceType: Literal["patch"]
+
+
```
eskultety, line 570: Hmm, we're defining our type, aren't we? Does that still count as being compliant with the spec? :) 
The doc string might be a bit misleading without further context if you just look at Annex F where the types should be discussed: https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f5-other - I don't see `patch` in there.
That said, when I look at examples then `OTHER` isn't really well described in terms of usage: https://spdx.github.io/spdx-spec/v2.3/package-information/#7213-examples which gives us the flexibility to do this and put just about anything in the metadata.
Since this is consistent with the other categories modeling, this change is fine from my POV, however, please make it explicit and visible that this type is in fact custom and not defined by the standard.

eskultety (approved):
ACK with 1 minor change.

hermeto/core/models/sbom.py
slimreaper35: I believe we should update https://github.com/hermetoproject/hermeto/blob/main/docs/sbom.md#cyclonedx-to-spdx-mapping

hermeto/core/models/sbom.py
```diff
@@ -536,9 +578,16 @@ class SPDXPackageExternalRefSecurityPURL(
...
+    SPDXPackageExternalRefOtherPatch,
+    pydantic.Field(discriminator="referenceType"),
+]
+
 
 SPDXPackageExternalRefType = Annotated[
-    SPDXPackageExternalRefPackageManagerType | SPDXPackageExternalRefSecurityType,
+    SPDXPackageExternalRefPackageManagerType
+    | SPDXPackageExternalRefSecurityType
+    | SPDXPackageExternalRefOtherType,
```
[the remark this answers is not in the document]
ST3V1K: I don't think so

```bash
$ python -c '
from pydantic import TypeAdapter

from hermeto.core.models.sbom import SPDXPackageExternalRefOtherType, SPDXPackageExternalRefType

data = {"referenceCategory": "OTHER", "referenceType": "patch", "referenceLocator": "https://example.com/p.diff"}

print(TypeAdapter(SPDXPackageExternalRefOtherType).validate_python(data))
print(TypeAdapter(SPDXPackageExternalRefType).validate_python(data))
'

referenceLocator='https://example.com/p.diff' referenceType='patch' referenceCategory='OTHER'
referenceLocator='https://example.com/p.diff' referenceType='patch' referenceCategory='OTHER'
```

hermeto/core/models/sbom.py
ST3V1K: Both done

slimreaper35 (approved):

PR #1658, opened 2026-07-08
Pip index url env

### LINKED ISSUE #1656: feat(pip): respect PIP_INDEX_URL environment variable
## Problem

Hermeto's pip backend only reads the index URL from `--index-url` inside `requirements.txt`. It does not honor `PIP_INDEX_URL`, which is pip's standard environment variable for configuring the default index.

This creates friction for organizations that use private PyPI indexes configured at the environment level (via `pip.conf` or `PIP_INDEX_URL`) rather than embedded in every requirements file.

### Example: hermetic builds in Konflux

A common pattern for container builds:

1. A **base image** ships with `pip.conf` pointing to a private index
2. **Component teams** build on that base image — `pip install -r requirements.txt` works because `pip.conf` provides the index
3. When **hermetic builds** are enabled, Hermeto runs *before* the container build in a separate pipeline step — it cannot see the base image's `pip.conf`

Today, teams must add `--index-url` to every `requirements.txt` to tell Hermeto where to download from. This duplicates the index URL across two places (base image config + every requirements file), and when the index URL changes (e.g., a version bump), every requirements file must be updated.

## Proposal

Respect `PIP_INDEX_URL` as a fallback when the requirements file has no `--index-url`:

```
requirements.txt --index-url  >  PIP_INDEX_URL env var  >  PyPI default
```

This matches pip's own precedence. Files that already specify `--index-url` are unaffected.

## How this helps

With `PIP_INDEX_URL` support, the Konflux pipeline can set the index URL once as an environment variable on the `prefetch-dependencies` step:

- **No duplication** — the index URL lives in the pipeline config, not copied into every requirements file. Component teams' `requirements.txt` files contain only dependency declarations.
- **One-line version bumps** — when the index URL changes (e.g., `v3.5` → `v3.6`), update one pipeline parameter instead of every requirements file across every component repo.
- **Portable requirements files** — the same `requirements.txt` works in any environment (upstream with PyPI, downstream with a private index). Only the environment config differs.
- **Gradual adoption** — existing requirements files with `--index-url` keep working unchanged. Teams can migrate at their own pace by removing `--index-url` from files and relying on the env var instead.

## Relation to #1634

Issue #1634 proposed a Hermeto-specific `index_url` JSON field and was [closed](https://github.com/hermetoproject/hermeto/issues/1634) over concerns about vendor lock-in and adding abstractions outside pip's ecosystem.

This proposal takes a different approach: `PIP_INDEX_URL` is pip's own documented environment variable, already supported by pip, uv, and other pip-compatible tools. It's fully backward compatible — if the env var is not set, behavior is identical to today.

LalatenduMohanty:
Hermeto now reads PIP_INDEX_URL as a fallback when a requirements file does not specify --index-url, matching pip's own documented precedence:
    requirements.txt --index-url > PIP_INDEX_URL > PyPI default.
    
This lets CI/CD pipelines parameterize the index URL without embedding it in every requirements file.
    
Closes https://github.com/hermetoproject/hermeto/issues/1656

### DIFF
--- a/docs/pip.md
+++ b/docs/pip.md
@@ -306,8 +306,16 @@ Make Hermeto download packages from the specified Python Package Index server.
 
 Applies to all the packages (and only the packages) from the file which
 contains the `--index-url` option. If file A contains `--index-url` and file B
-does not, Hermeto will download the packages declared in B from the default
-index server (`https://pypi.org/simple`).
+does not, Hermeto will download the packages declared in B from the
+`PIP_INDEX_URL` environment variable (if set) or the default index server
+(`https://pypi.org/simple`).
+
+The precedence follows
+[pip's own configuration precedence](https://pip.pypa.io/en/stable/topics/configuration/#environment-variables):
+
+```text
+requirements.txt --index-url  >  PIP_INDEX_URL env var  >  PyPI default
+```
 
 > [!WARNING]
 > Do not include credentials in the index url. If needed, provide authentication
--- a/hermeto/core/package_managers/python/pip/main.py
+++ b/hermeto/core/package_managers/python/pip/main.py
@@ -18,7 +18,13 @@ from packaging.utils import canonicalize_name
 from hermeto.core.checksum import ChecksumInfo, must_match_any_checksum
 from hermeto.core.config import get_config
 from hermeto.core.constants import Mode
-from hermeto.core.errors import LockfileNotFound, NotAGitRepo, PackageRejected, UnsupportedFeature
+from hermeto.core.errors import (
+    InvalidInput,
+    LockfileNotFound,
+    NotAGitRepo,
+    PackageRejected,
+    UnsupportedFeature,
+)
 from hermeto.core.models.input import PipBinaryFilters, Request
 from hermeto.core.models.output import EnvironmentVariable, ProjectFile, RequestOutput
 from hermeto.core.models.sbom import (
@@ -64,6 +70,32 @@ log = logging.getLogger(__name__)
 
 DEFAULT_BUILD_REQUIREMENTS_FILE = "requirements-build.txt"
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
+
+
+def _validate_index_url(url: str, source: str) -> None:
+    """Validate a PyPI index URL regardless of where it was configured.
+
+    :param url: the index URL to validate
+    :param source: human-readable origin for error messages (e.g. "--index-url", "PIP_INDEX_URL")
+    :raises PackageRejected: if the URL is invalid
+    """
+    parsed = urlparse.urlparse(url)
+    if parsed.scheme not in ("http", "https"):
+        raise InvalidInput(
+            f"{source} must use http or https scheme, got: {parsed.scheme or 'none'}",
+            solution=f"Set {source} to a valid HTTP(S) URL (e.g., https://pypi.org/simple/)",
+        )
+    if not parsed.netloc:
+        raise InvalidInput(
+            f"{source} must include a host",
+            solution=f"Set {source} to a valid HTTP(S) URL (e.g., https://pypi.org/simple/)",
+        )
+    if parsed.username or parsed.password:
+        raise InvalidInput(
+            f"{source} must not contain embedded credentials",
+            solution="Use a .netrc file for authentication instead of embedding "
+            "credentials in the URL",
+        )
 
 
 class _PyPIArtifact(NamedTuple):
@@ -484,7 +516,11 @@ def _download_dependencies(
         log.info("-- Finished processing requirement line '%s'", req.download_line)
 
     if pypi_reqs:
-        index_url = options["index_url"] or pypi_simple.PYPI_SIMPLE_ENDPOINT
+        if options["index_url"]:
+            _validate_index_url(options["index_url"], "--index-url")
+            index_url = options["index_url"]
+        else:
+            index_url = pypi_simple.PYPI_SIMPLE_ENDPOINT
         processed.extend(
             _resolve_and_download_pypi_packages(
                 pypi_reqs, requirements_file, pip_deps_dir, binary_filters, index_url
--- a/hermeto/core/package_managers/python/pip/main.py
+++ b/hermeto/core/package_managers/python/pip/main.py
@@ -2,6 +2,7 @@
 import asyncio
 import functools
 import logging
+import os
 import tarfile
 import zipfile
 from collections.abc import Callable, Iterable, Iterator
@@ -519,6 +520,13 @@ def _download_dependencies(
         if options["index_url"]:
             _validate_index_url(options["index_url"], "--index-url")
             index_url = options["index_url"]
+        elif pip_index_url := os.environ.get("PIP_INDEX_URL", "").strip():
+            _validate_index_url(pip_index_url, "PIP_INDEX_URL")
+            index_url = pip_index_url
+            log.info(
+                "Using PIP_INDEX_URL='%s' (no --index-url in requirements file)",
+                pip_index_url,
+            )
         else:
             index_url = pypi_simple.PYPI_SIMPLE_ENDPOINT
         processed.extend(
--- a/tests/unit/package_managers/python/pip/test_main.py
+++ b/tests/unit/package_managers/python/pip/test_main.py
@@ -998,3 +998,68 @@ class TestValidateIndexUrl:
         )
         with pytest.raises(InvalidInput):
             pip._download_dependencies(rooted_tmp_path, req_file)
+
+
+class TestPipIndexUrlEnv:
+    """Tests for PIP_INDEX_URL environment variable support."""
+
+    @pytest.mark.parametrize(
+        "env_url, file_options, expected_url",
+        [
+            pytest.param(
+                "https://env-index.example.com/simple/",
+                ["-i", "https://file-index.example.com/simple/"],
+                "https://file-index.example.com/simple/",
+                id="file_wins_over_env",
+            ),
+            pytest.param(
+                "https://env-index.example.com/simple/",
+                [],
+                "https://env-index.example.com/simple/",
+                id="env_used_when_no_file_index",
+            ),
+            pytest.param(
+                None,
+                [],
+                "https://pypi.org/simple/",
+                id="pypi_default_when_neither_set",
+            ),
+        ],
+    )
+    @mock.patch("hermeto.core.package_managers.python.pip.main._resolve_and_download_pypi_packages")
+    def test_precedence(
+        self,
+        mock_resolve: Any,
+        env_url: str | None,
+        file_options: list[str],
+        expected_url: str,
+        monkeypatch: pytest.MonkeyPatch,
+        rooted_tmp_path: RootedPath,
+    ) -> None:
+        """Test precedence of index URL configuration sources."""
+        if env_url is not None:
+            monkeypatch.setenv("PIP_INDEX_URL", env_url)
+        else:
+            monkeypatch.delenv("PIP_INDEX_URL", raising=False)
+
+        mock_resolve.return_value = []
+        req = mock_requirement("foo", "pypi", version_specs=[("==", "1.0")])
+        req_file = mock_requirements_file(requirements=[req], options=file_options)
+
+        pip._download_dependencies(rooted_tmp_path, req_file)
+
+        call_args = mock_resolve.call_args
+        assert call_args[0][4] == expected_url
+
+    def test_invalid_env_url_rejected(
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+        rooted_tmp_path: RootedPath,
+    ) -> None:
+        """An invalid PIP_INDEX_URL is rejected when no --index-url is in the file."""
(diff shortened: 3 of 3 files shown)


hermeto/core/package_managers/python/pip/main.py
```diff
@@ -67,33 +67,43 @@
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
 
 
-def _get_pip_index_url_env() -> str | None:
-    """Read and validate PIP_INDEX_URL from the environment.
+def _validate_index_url(url: str, source: str) -> None:
```
eskultety, line 76: `_get_pip_index_url_env` was a new function introduced in the previous commit so this commit (2/2) change should be hence squashed to 1/2. There's no need for refactoring a newly introduced change, introduce it as complete right away :).

hermeto/core/package_managers/python/pip/main.py
```diff
@@ -67,33 +67,43 @@
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
 
 
-def _get_pip_index_url_env() -> str | None:
```
eskultety: **Commit message:**
nitpick: https://github.com/hermetoproject/hermeto/issues/1656 - prefix with `Closes/Resolves`

hermeto/core/package_managers/python/pip/main.py
eskultety, line 76: I think the correct order of commits should be:

1. index URL validator (useful on its own regardless of PIP_INDEX_URL
2. PIP_INDEX_URL enhancement

hermeto/core/package_managers/python/pip/main.py
```diff
@@ -67,33 +67,43 @@
...
+    """Read and validate PIP_INDEX_URL from the environment.
+
+    :return: the validated URL, or None if the variable is not set or empty
+    :raises PackageRejected: if the URL is invalid
+    """
+    url = os.environ.get("PIP_INDEX_URL", "").strip()
+    if not url:
+        return None
+
+    _validate_index_url(url, "PIP_INDEX_URL")
```
eskultety: nitpick: Do we need a helper to encode 4 lines of code in the if-elif-else case? I think the code is trivial enough to be open coded where this is being called from.

tests/unit/package_managers/python/pip/test_main.py
```diff
@@ -958,6 +958,57 @@ def test_fetch_pip_source_correctly_reraises_when_there_is_a_dependency_cargo_lo
...
+class TestValidateIndexUrl:
+    """Tests for index URL validation shared by --index-url and PIP_INDEX_URL."""
+
+    def test_validate_index_url_valid_https(self) -> None:
+        """Accept a valid HTTPS index URL."""
+        pip._validate_index_url(CUSTOM_PYPI_ENDPOINT, "--index-url")
+
+    def test_validate_index_url_valid_http(self) -> None:
+        """Accept a valid HTTP index URL."""
+        pip._validate_index_url("http://internal.example.com/simple/", "--index-url")
```
eskultety: These 2 can be consolidated to a single parametrized test.

tests/unit/package_managers/python/pip/test_main.py
```diff
@@ -958,6 +958,57 @@ def test_fetch_pip_source_correctly_reraises_when_there_is_a_dependency_cargo_lo
...
+                "https://user@example.com/simple/",
+                "must not contain embedded credentials",
+                id="user_only",
+            ),
+        ],
+    )
+    def test_validate_index_url_rejects_invalid(self, url: str, match: str) -> None:
+        """Reject index URLs with bad scheme, missing host, or embedded credentials."""
+        with pytest.raises(PackageRejected, match=match):
+            pip._validate_index_url(url, "--index-url")
```
eskultety: No string matching in unit tests please. For the purposes of the helper we don't care about the exact message, just that the validation failed, rightfully so.

tests/unit/package_managers/python/pip/test_main.py
```diff
@@ -958,6 +958,57 @@ def test_fetch_pip_source_correctly_reraises_when_there_is_a_dependency_cargo_lo
...
+
+    def test_validate_index_url_includes_source_in_message(self) -> None:
+        """Error message references the source that provided the invalid URL."""
+        with pytest.raises(PackageRejected, match="PIP_INDEX_URL must use http or https"):
```
eskultety: If you drop string matching, you can further consolidate by making this an additional parametrized case in the test above.

eskultety (changes requested):

hermeto/core/package_managers/python/pip/main.py
LalatenduMohanty: Sounds good to me. Let me fix that.

hermeto/core/package_managers/python/pip/main.py
LalatenduMohanty: Works for me. Will fix it.

hermeto/core/package_managers/python/pip/main.py
LalatenduMohanty: Fixed it. PTAL

tests/unit/package_managers/python/pip/test_main.py
LalatenduMohanty: Done. Resolving the comment.

hermeto/core/package_managers/python/pip/main.py
LalatenduMohanty: marking this as resolved. PTAL

hermeto/core/package_managers/python/pip/main.py
LalatenduMohanty: This is fixed in the commit message as well as the PR description. Marking it as resolved.

tests/unit/package_managers/python/pip/test_main.py
LalatenduMohanty: Agreed, dropped all match= from the validation tests. We only care that PackageRejected is raised, not the exact wording

tests/unit/package_managers/python/pip/test_main.py
LalatenduMohanty: Removed it entirely rather than adding another parametrized case, since the source-in-message behavior is already covered by test_validate_index_url_rejects_invalid.

hermeto/core/package_managers/python/pip/main.py
```diff
@@ -66,6 +66,32 @@
...
+            f"{source} must use http or https scheme, got: {parsed.scheme or 'none'}",
+            solution=f"Set {source} to a valid HTTP(S) URL (e.g., https://pypi.org/simple/)",
+        )
+    if not parsed.netloc:
+        raise PackageRejected(
+            f"{source} must include a host",
+            solution=f"Set {source} to a valid HTTP(S) URL (e.g., https://pypi.org/simple/)",
+        )
+    if parsed.username or parsed.password:
+        raise PackageRejected(
```
eskultety: nitpick: I failed to raise this during the first round of review - with `PIP_INDEX_URL` validation I think this falls better into `InvalidInput` error rather than `PackageRejected`.

hermeto/core/package_managers/python/pip/main.py
```diff
@@ -66,6 +66,32 @@
 DEFAULT_REQUIREMENTS_FILE = "requirements.txt"
 
 
+def _validate_index_url(url: str, source: str) -> None:
```
eskultety, line 76: **commit message**

> feat(pip): validate --index-url ...

nitpick: ^This is not a feature, it's either nothing or maybe refactoring :).

\<rant> this is btw. why I have my personal reservations towards conventional commits, because having work split into several logical commits does not play well with the limited number of topics available and one always needs to think which one is the most fitting for a given commit, literally inviting for this kind of bike shedding :) \</rant>.

docs/pip.md
```diff
@@ -306,8 +306,16 @@ Make Hermeto download packages from the specified Python Package Index server.
...
 Applies to all the packages (and only the packages) from the file which
 contains the `--index-url` option. If file A contains `--index-url` and file B
-does not, Hermeto will download the packages declared in B from the default
-index server (`https://pypi.org/simple`).
+does not, Hermeto will download the packages declared in B from the
+`PIP_INDEX_URL` environment variable (if set) or the default index server
+(`https://pypi.org/simple`).
+
+The precedence follows
+[pip's own configuration precedence](https://pip.pypa.io/en/stable/topics/configuration/#environment-variables):
```
eskultety, line 314: **Commit message:**

Please record this pypa docs URL in the commit message as well since you're mentioning the order of precedence there as well.

eskultety (approved):
@LalatenduMohanty please fix the failing markdown CI check.

docs/pip.md
LalatenduMohanty: Done.

hermeto/core/package_managers/python/pip/main.py
LalatenduMohanty: true, I could have merged everything in to one commit. But I was just trying to be nice for easy code review.

LalatenduMohanty: fixed it.

hermeto/core/package_managers/python/pip/main.py
LalatenduMohanty: Fixed it.

eskultety (approved):

tests/unit/package_managers/python/pip/test_main.py
```diff
@@ -995,5 +996,79 @@ def test_download_dependencies_rejects_invalid_file_index_url(
...
+
+        mock_resolve.return_value = []
+        req = mock_requirement("foo", "pypi", version_specs=[("==", "1.0")])
+        req_file = mock_requirements_file(requirements=[req], options=file_options)
+
+        pip._download_dependencies(rooted_tmp_path, req_file)
+
+        call_args = mock_resolve.call_args
+        assert call_args[0][4] == expected_url
+        if expect_log:
```
a-ovchinnikov: Could you please remove the logic that checks logs? It essentially validates Python standard library which must be tested elsewhere. We are trying to get trid of such validations in the codebase.

tests/unit/package_managers/python/pip/test_main.py
```diff
@@ -956,3 +956,165 @@ def test_fetch_pip_source_correctly_reraises_when_there_is_a_dependency_cargo_lo
...
+        monkeypatch.delenv("PIP_INDEX_URL", raising=False)
+        mock_resolve.return_value = []
+
+        req = mock_requirement("foo", "pypi", version_specs=[("==", "1.0")])
+        req_file = mock_requirements_file(requirements=[req])
+
+        pip._download_dependencies(rooted_tmp_path, req_file)
+
+        call_args = mock_resolve.call_args
+        assert call_args[0][4] == "https://pypi.org/simple/"
```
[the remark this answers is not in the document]
a-ovchinnikov: This is a pretty bad advice: not only it makes new assertions on caplog, it also promotes branching in tests.

a-ovchinnikov (commented):
LGTM with one minor request, will approve once resolved.

tests/unit/package_managers/python/pip/test_main.py
LalatenduMohanty: Fixed it. PTAL. Resolving this conversation.

a-ovchinnikov (dismissed):

tests/unit/package_managers/python/pip/test_main.py
```diff
@@ -956,3 +956,44 @@ def test_fetch_pip_source_correctly_reraises_when_there_is_a_dependency_cargo_lo
...
+        [
+            pytest.param("ftp://example.com/simple/", id="ftp_scheme"),
+            pytest.param("https://", id="no_host"),
+            pytest.param("https://user:pass@example.com/simple/", id="user_pass"),
+            pytest.param("https://user@example.com/simple/", id="user_only"),
+        ],
+    )
+    def test_validate_index_url_rejects_invalid(self, url: str) -> None:
+        """Reject index URLs with bad scheme, missing host, or embedded credentials."""
+        with pytest.raises(PackageRejected):
```
a-ovchinnikov: The code raises `InvalidInput` now, please update the new tests accordingly in the same commit as the change is introduced.

eskultety:
> UTs fail for the first commit

Yeah, this is a dealbreaker, great catch @a-ovchinnikov !

a-ovchinnikov (commented):
I found one minor omisson, everything else is fine.

tests/unit/package_managers/python/pip/test_main.py
LalatenduMohanty: Done,  the validator commit now introduces InvalidInput in both the production code and the tests together

LalatenduMohanty:
@a-ovchinnikov Does the UTs fail for the first commit? can you point me the failure please, I could not find it.

a-ovchinnikov:
> Does the UTs fail for the first commit?

Not anymore, the most recent push fixed them

a-ovchinnikov (approved):

PR #1661, opened 2026-07-14
Replace Gemini Code Assist with CodeRabbit

tnevrlka:
Gemini Assist Code is being deprecated and will stop working on 2026-07-17. Get rid of its leftover config files.

Hermeto has been previously using CodeRabbit, let's try using it again.

Write a .coderabbit.yaml config file that contains CodeRabbit specific settings that were already used before.
It disables pretty much all the additional "fun" fluff, keeping only the actual reviews.


slimreaper35:
@coderabbitai What's up man. Welcome back!

.coderabbit.yaml
```diff
@@ -0,0 +1,24 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+language: "en-US"
+reviews:
+  profile: "chill"
+  high_level_summary: true
```
eskultety: I'd personally disable this one. I don't want to rely too heavily on some hallucinated summary of changes, as a maintainer I'd expect the intent of a PR to have been captured in the PR description, so preferably, let's reduce the noise.

.coderabbit.yaml
```diff
@@ -0,0 +1,24 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+language: "en-US"
+reviews:
+  profile: "chill"
```
eskultety: These configuration names "chill" vs "assertive" are quite vague without having a good idea on what they actually do. Note that every change to the config requires the review process which both good and a bad thing because in order to experiment an org admin can just flip the switch on these settings in the code rabbit dashboard, but that's not transparent.
I personally would configure only settings which could be classified as  "general truth" this way via a config file on things that we know for sure we don't want because they add to the noise. Because the profile is non-deterministic, whatever is chill today may be completely unusable tomorrow or vice versa, so having the option to just flip a switch on these doesn't sound too bad.

eskultety (approved):
This is IMO good for starters, 2 tiny comments. We can always adopt more "disablements" as we go depending on how bad the default experience is.

.coderabbit.yaml
tnevrlka: That's a really good catch, it might actually be even worse than that and edit the existing PR description, see https://docs.coderabbit.ai/pr-reviews/summaries

.coderabbit.yaml
tnevrlka: > These configuration names "chill" vs "assertive" are quite vague without having a good idea on what they actually do. Note that every change to the config requires the review process which both good and a bad thing because in order to experiment an org admin can just flip the switch on these settings in the code rabbit dashboard, but that's not transparent.

Agreed. I wouldn't mind a transition period when only settings in the UI were to be used and changed often until a configuration Hermeto maintainers are happy with would be set in stone in the YAML.

On the other hand, the CodeRabbit UI is imo very clunky and I'm not sure how the change in behavior could be tracked. I could imagine that having the setting change recorded in Git could then make it easier to lookup which review comments were created based on which setting. 
I guess you could work around that with some notekeeping, but that's up to Hermeto org admins. 

> I personally would configure only settings which could be classified as "general truth" this way via a config file on things that we know for sure we don't want because they add to the noise. Because the profile is non-deterministic, whatever is chill today may be completely unusable tomorrow or vice versa, so having the option to just flip a switch on these doesn't sound too bad.

I'm afraid it's not possible to combine the UI settings with the YAML. It's one or the other. IIUC, if it wasn't explicitly mentioned in the YAML, it would default to `chill`, see https://docs.coderabbit.ai/reference/configuration#param-profile

taylormadore (approved):

.coderabbit.yaml
tnevrlka: Updated to false

.coderabbit.yaml
eskultety: > I'm afraid it's not possible to combine the UI settings with the YAML

Oh, that changes things.

> I wouldn't mind a transition period when only settings in the UI 

Can we do this for a brief period? Say 2-4 weeks? I fully support the transparency here wrt/ a checked in config, it's just AI is such a black box that one needs to set something first to really know the outcome (sadly).

tnevrlka: > Can we do this for a brief period? Say 2-4 weeks?

¯\\\_(ツ)_/¯ Be my guest, but you have to find a Hermeto org admin willing to play with it

eskultety: > but you have to find a Hermeto org admin willing to play with it

No worries, it is not my intent to spend too much time on this, rather trying to find out what the output is going to look like when we disable most of the features and read some more on the ones we're not sure about as we're testing. The good thing is that the web UI shows you the whole YAML (except the default values).
This is what I ended up with and applied. If it turns out it's satisfactory, we can commit the config below. Feel free to continue with the discussion in this PR. I expect we'll merge it at the end. As one of the org admins I can toggle any switch anyone wishes to try out.

```
reviews:
  profile: assertive
  high_level_summary: false
  changed_files_summary: false
  sequence_diagrams: false
  estimate_code_review_effort: false
  assess_linked_issues: false
  suggested_labels: false
  suggested_reviewers: false
  in_progress_fortune: false
  enable_prompt_for_ai_agents: false
  path_filters:
    - "!requirements*.txt"
  auto_review:
    auto_pause_after_reviewed_commits: 20
    ignore_title_keywords:
      - WIP
      - DO NOT MERGE
      - REPRODUCER
      - DNM
    ignore_usernames:
      - dependabot
  finishing_touches:
    docstrings:
      enabled: false
    unit_tests:
      enabled: false
    autofix:
      enabled: false

chat:
  art: false
  allow_non_org_members: false
  auto_reply: false
  integrations:
    jira:
      usage: disabled

knowledge_base:
  jira:
    usage: disabled

issue_enrichment:
  planning:
    enabled: false
    auto_planning:
      enabled: false
```

tnevrlka: > ```
> knowledge_base:
>   jira:
>     usage: disabled
> ```

I am not sure about this. I believe `code_guidelines` should be enabled so that CodeRabbit uses Hermeto's AGENTS.md etc. and I'm not sure they're enabled by default. https://docs.coderabbit.ai/reference/configuration#param-code-guidelines

eskultety: ~~Good point, my bad, the webUI provides less information than the docs. However, we'll need to explicitly point it to AGENTS.md (I assumed it reads either that or CLAUDE.md automatically).~~ I'll enable it.

Edit: scratch the above, the docs says

> Defaults include: (**/.cursorrules, .github/copilot-instructions.md, **/CLAUDE.md, **/GEMINI.md, **/.cursor/rules/, **/.windsurfrules, **/.clinerules/, **/.rules/*, **/AGENT.md, **/AGENTS.md, **/REVIEW.md).

so `[]` equals the above. I'll just enable it.

eskultety: Hmm,

Image

Image

It should be enabled already, defaults are not dumped to the YAML from within the webUI.

tnevrlka: Oh, are the empty curly brackets supposed to represent "all suboptions have their default values"? I thought it means it's not enabled.
```
Use your coding guideline documents (see File Patterns) as review criteria.
Defaults to {}.
```

I guess that makes sense, otherwise `enabled` default to `true` wouldn't make sense. My bad

tnevrlka: Ok, I went with your proposed config.

.coderabbit.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - "!requirements*.txt"
+  auto_review:
+    auto_pause_after_reviewed_commits: 20
+    ignore_title_keywords:
+      - WIP
+      - DO NOT MERGE
+      - REPRODUCER
+      - DNM
+    ignore_usernames:
+      - dependabot
```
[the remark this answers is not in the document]
tnevrlka: This seems valid https://docs.coderabbit.ai/configuration/auto-review#ignore_usernames

.coderabbit.yaml
```diff
@@ -0,0 +1,48 @@
...
+  profile: assertive
+  high_level_summary: false
+  changed_files_summary: false
+  sequence_diagrams: false
+  estimate_code_review_effort: false
+  assess_linked_issues: false
+  suggested_labels: false
+  suggested_reviewers: false
+  in_progress_fortune: false
+  enable_prompt_for_ai_agents: false
```
eskultety, line 13: I also found the following which is True by default, but better be explicit if the default silently changes.
```
  slop_detection:
    enabled: true
```

eskultety (approved):
Looking at the webUI it seems like we have disabled most of what can be disabled in the free edition, I can live with this for the time being, but I encourage anyone to propose further setting updates as we go and fine-tune the free-tier experience if possible.

.coderabbit.yaml
tnevrlka: Sure, added.

.coderabbit.yaml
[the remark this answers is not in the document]
tnevrlka: Changed to `dependabot[bot]`

PR #1662, opened 2026-07-15
yarn berry dev dependencies

### LINKED ISSUE #1590: Unify node image for integration tests
We use various versions of node images for npm, yarn classic, yarn berry, and pnpm integration tests. That is unnecessary, and we can reduce the number of images that are being pulled.

Make sure `corepack` is included for yarn berry.

slimreaper35:


hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -264,6 +272,36 @@ def _set_yarnrc_configuration(
...
+    is executed in a temporary directory.
+    """
+    with tempfile.TemporaryDirectory() as temp_dir:
+        # Changing the global folder will prevent file dependencies from being re-packaged after
+        # modifying package.json files.
+        project.yarn_rc["globalFolder"] = temp_dir
+        project.yarn_rc["enableImmutableInstalls"] = False
+        project.yarn_rc.write()
+
+        cmd = ["install", "--mode", "update-lockfile", "--mode", "skip-build"]
```
taylormadore: Can `--mode` be used twice?

taylormadore: Have we confirmed that we are not re-downloading dependencies?

slimreaper35: No, `--mode=update-lockfile` is enough.

slimreaper35: I tested it with [yarn/e2e](https://github.com/hermetoproject/integration-tests/tree/yarn/e2e) and after replacing `temp_dir` with some hardcoded path, the `"globalFolder"` contains only one local package that yarn re-packed because its content has changed.

```
> ls hermeto-output/deps/yarn/cache | wc -l
     151

> ls here/cache                                                  
supports-hyperlinks-folder-file-a1f4a44e40-8.zip
```

Also, I checked these yarn berry source code files:
- https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-core/sources/Project.ts#L1884
- https://github.com/yarnpkg/berry/pull/2913/changes

hermeto/core/package_managers/javascript/yarn/resolver.py
```diff
@@ -229,12 +230,17 @@ def _resolve_workspace_packages(source_dir: RootedPath, workspaces: list[str]) -
...
+    packages: list[Package], prod_packages: list[Package], project: Project, output_dir: RootedPath
 ) -> list[Component]:
     """Create SBOM components for all the packages parsed from the 'yarn info' output."""
     package_mapping: dict[Locator, Package] = {}
     patch_locators: list[PatchLocator] = []
 
+    prod_locators = {pkg.raw_locator for pkg in prod_packages}
+    # The number of components with the cdx:npm:development property might be lower than the number
+    # of dev locators because patches are not components themselves.
+    dev_locators = {pkg.raw_locator for pkg in packages if pkg.raw_locator not in prod_locators}
```
a-ovchinnikov: Nitpick (a strictly optional suggestion): since these are sets you can do set arithmetic on them:
```python
raw_locators = lambda pkgs: {pkg.raw_locator for pkg in pkgs}
prod_locators = raw_locators(prod_packages)
dev_locators = raw_locators(packages) - prod_locators
```

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -134,13 +135,20 @@ def _resolve_yarn_project(
...
         _strip_workspace_scripts(project.source_dir, packages)
 
     _fetch_dependencies(project.source_dir, workspaces)
 
-    return create_components(packages, project, output_dir)
+    _remove_dev_dependencies(project.source_dir)
+    _update_lockfile(project)
+
+    prod_packages = resolve_packages(project.source_dir, workspaces)
+    log.debug("Resolved %d development locators", len(packages) - len(prod_packages))
```
a-ovchinnikov: Nitpick: since these four lines form a distinct group (i.e. make sense only together) I would suggest moving them out into a separate function to highlight that, something along the lines of:
```python3
prod_packages = resolve_only_prod_packages(project.source_dir, workspaces)
...
def resolve_only_prod_packages(:Path, :list[Path]) -> list[Package]:
    _remove_dev_dependencies(project.source_dir)
    _update_lockfile(project)
    result = resolve_packages(project.source_dir, workspaces)
    log.debug("Resolved %d production locators", len(result))
    return result
```

One problem with this approach is that the order of `resolve_packages()` and `resolve_only_prod_packages()` matters (because the second function modifies external state) and it is not immediately obvious. An even better approach would be to ensure that the state is not changed, something like below:
```python3
def resolve_only_prod_packages(:Path, :list[Path]) -> list[Package]:
    with _temporary_hide_dev_deps():
        result = resolve_packages(project.source_dir, workspaces)
    log.debug("Resolved %d production locators", len(result))
    return result

@contextmanager
def _temporary_hide_dev_deps():
    dev_deps = _remove_dev_dependencies*(project.source_dir)
    _update_lockfile(project)
    try:
        yield
    finally:
        _insert_dev_deps_back(dev_deps)
        _update_lockfile(project)
```

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -264,6 +272,36 @@ def _set_yarnrc_configuration(
...
+        log.debug("Found devDependencies in %s", path)
+        del package_json["devDependencies"]
+        package_json.write()
+
+
+def _update_lockfile(project: Project) -> None:
+    """
+    Update yarn.lock to synchronize with package.json files without devDependencies.
+
+    Note that this operation does not modify the actual source directory, since the prefetching
```
a-ovchinnikov: I am sorry, this gets pretty confusing: `_remove_dev_dependencies()` modifies the source directory, `_update_lockfile()` does not, but sets a different global directory for Yarn, then Yarn updates lock file. Where does this update happen, in the new `globalFolder`? If so then by the time the function completes the directory gets cleaned (the context manager is supposed to take care of that), and thus `globalFoler` ends up pointing to nowhere. What am I missing here? _Edit_: typo

a-ovchinnikov (commented):
A few minor questions and nitpicks.

hermeto/core/package_managers/javascript/yarn/main.py
slimreaper35: `_update_lockfile()` and `_remove_dev_dependencies()` functions modify the lockfile inside `tmp*.hermeto-source-copy`. Same as other files like `.yarnrc.yaml`

If you apply this diff in [resolver.py](https://github.com/hermetoproject/hermeto/blob/main/hermeto/core/resolver.py)

```diff
- with TemporaryDirectory(f".{APP_NAME}-source-copy", dir=".") as temp_dir:
+ with TemporaryDirectory(f".{APP_NAME}-source-copy", dir=".", delete=False) as temp_dir:
```

you can cd into the temporary directory and see all changes we made. But the actual source directory is clean. That's what I meant. If the note is confusing I can drop it or rephrase it.

> Where does this update happen, in the new globalFolder?

So we don't modify the Hermeto output directory at all.

a-ovchinnikov: Yes, everything is supposed to be happening in a copy in a temporary directory, now this appears to be yet another temporary directory that get pointed to. I read the comment as if this time around the _copy_ does not get modified

hermeto/core/package_managers/javascript/yarn/resolver.py
slimreaper35:
```python
dev_locators = {pp.raw_locator for pp in prod_packages} - {p.raw_locator for p in packages}
```
?

prod_locators variable isn't used anywhere

a-ovchinnikov: Yes, this is fine too.

hermeto/core/package_managers/javascript/yarn/main.py
slimreaper35: > Nitpick: since these four lines form a distinct group (i.e. make sense only together) I would suggest moving them out into a separate function to highlight that, something along the lines of:

Added. Still thinking about `contextmanager`. But it's probably a good idea to avoid future regressions and suprises.

taylormadore: I too am worried about leaving footguns laying around

slimreaper35: Done.

I did some testing yesterday and running `_update_lockfile` again, yarn would literally try to update devDependencies to newer versions available. So the lockfile content is restored simply by read -> save -> write it back.

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -264,6 +271,56 @@ def _set_yarnrc_configuration(
     yarn_rc.write()
 
 
+def _resolve_only_prod_packages(project: Project, workspaces: list[str] | None) -> list[Package]:
+    """Resolve only the production packages."""
+    deps_dir = project.yarn_rc.get("globalFolder")
+    with tempfile.TemporaryDirectory() as temp_dir:
+        # Prevent any changes to the hermeto-output/deps directory.
+        # Otherwise, yarn will re-package file dependencies after modifying their package.json files.
+        project.yarn_rc["globalFolder"] = temp_dir
```
taylormadore, line 282: Can the yarnrc handling and tempdir also be moved into the context manager? Then this function just becomes a call to resolve_packages wrapped in _hide_dev_dependencies

tnevrlka (approved):

slimreaper35:
New revision:
- refactored context manager based on https://github.com/hermetoproject/hermeto/pull/1662#discussion_r3646735945
- dropped `project.yarn_rc["enableImmutableInstalls"] = False` since it has no effect (https://github.com/yarnpkg/berry/pull/3933)
- Cursor updated the yarn berry design document

taylormadore (approved):

docs/design/yarnberry.md
```diff
@@ -260,6 +260,49 @@ Example (from "https://github.com/hermetoproject/hermeto/blob/4fc76df98d71df67c7
...
+locator appears as a regular package.
+
+Other JavaScript backends (npm, pnpm, Yarn classic) already report development
+dependencies via the CycloneDX `cdx:npm:package:development`
+[property](https://github.com/CycloneDX/cyclonedx-property-taxonomy/blob/main/cdx/npm.md).
+
+**Approach:** resolve the full dependency tree once, then temporarily hide all `devDependencies`,
+re-resolve, and treat the set difference as development.
+
+1. Resolve all locators with `yarn info` (the normal dependency-list step).
```
a-ovchinnikov: Nitpick: the list essentially repeats the code (which is rather straightforward), I think it could be dropped from the document, "Note" as well.

a-ovchinnikov (approved):

docs/design/yarnberry.md
slimreaper35: Ok, dropped.

hermeto/core/package_managers/javascript/yarn/main.py
```diff
@@ -264,6 +272,46 @@ def _set_yarnrc_configuration(
...
+        content = lockfile.read_text()
+        try:
+            run_yarn_cmd(["install", "--mode", "update-lockfile"], project.source_dir)
+            yield
+        finally:
+            for package_json, deps in saved_deps:
+                package_json["devDependencies"] = deps
+                package_json.write()
+
+            lockfile.write_text(content)
```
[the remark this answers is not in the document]
slimreaper35: Changed for `join_within_root`

PR #1663, opened 2026-07-16
doc(rpm): Add RPM design doc

### LINKED ISSUE #1536: [docs][rpm] Add design retroactively
Publish the design in a markdown form with other designs. The original design happened as a GH discussion here: https://github.com/orgs/hermetoproject/discussions/465.

mohamedbouchtout:
Closes #1536 

### Summery
The original design doc for RPM was created in a discussion post (https://github.com/orgs/hermetoproject/discussions/465). Publish the design doc as a markdown with the rest of the design docs.

### Details
1. Copy the design doc as a markdown file
2. Update the doc to match current codebase per commit

### DIFF
--- a/docs/design/rpm.md
+++ b/docs/design/rpm.md
@@ -1,0 +1,302 @@
+# Hermeto support for prefetching RPMs - design
+
+## Introduction
+
+Containers download RPMs at build time. With hermetic builds, this is not possible
+because the build doesn't have access to the network. This feature solves this problem by pre-fetching
+RPMs before the build process to allow for hermetic builds.
+
+The general idea is that the developer specifies ahead of time what RPMs they want to have
+available. In a step before the actual build, hermeto takes this information, downloads all of the
+packages and prepares a repository that will be available during build.
+
+## Requirements
+
+1. The solution must work for processing all required architectures in one process.
+2. The solution must work without changes to RPM or package management tooling.
+3. The solution should work for all RPM based distributions.
+4. The solution shouldn't require the maintainer to make significant changes to their Dockerfiles.
+5. The solution should be able to propagate repoids for prefetched RPMs and modules to the container
+   build.
+    1. This will make sure the yum/dnf history command shows origin repositories correctly. This may
+       be leveraged by user tools (e.g. Clair) to identify sources of the RPM content.
+6. The solution must collect SRPMs for source container generation.
+    1. Only applicable if the SRPMs were listed in the *rpms.lock.yaml* file.
+
+## Block schema
+
+The diagram shows the main function blocks, inputs and outputs. The fetch-deps command processes the
+lock file and downloads all required items (RPMs, source RPMs, and module metadata) into a directory
+structure that serves as input for the container build process. It also produces an SBOM with metadata
+of the fetched RPMs.
+
+As a next (optional) step, the other hermeto command (`inject-files`) will process the structure
+created in the previous step and will generate repositories and .repo file.
+
+```mermaid
+flowchart TB
+    A[/rpms.in.yaml/] --> B[resolve dependencies]
+
+    subgraph Phase1 ["1. hermeto 'fetch-deps'"]
+        direction LR
+        C[/rpms.lock.yaml/] --> D[fetching RPMs and sources]
+        D --> E[/fetched packages/]
+        E --> F[generate SBOM]
+        F --> H[/output directory structure/]
+    end
+
+    subgraph Phase2 ["2. hermeto 'inject-files'"]
+        G[generate repositories and repofile]
+    end
+
+    B --> C
+    E -.-> G
+    G -.-> H
+    H --> I[container build process]
+
+    style Phase1 fill:#e0e0e0,stroke:#999,color:#000
+    style Phase2 fill:#e0f2fe,stroke:#999,color:#000
+
+    classDef white fill:#fff,stroke:#000,color:#000
+    class A,B,C,D,E,F,G,H,I white
+
+    linkStyle default stroke:#000
+```
+
+## rpms.lock.yaml
+
+Works as a configuration input file for prefetching. The format allows multiple architectures to be
+listed.
+
+```yaml
+lockfileVersion: <version>
+lockfileVendor: <vendor>
+arches:
+  - arch: <arch>
+    packages:
+      - repoid: <repoid>
+        url: <url>
+        checksum: <method>:<digest>
+        size: <size>
+    source:
+      - repoid: <repoid-source>
+        url: <url>
+        checksum: <method>:<digest>
+        size: <size>
+    module_metadata:
+      - repoid: <repoid>
+        url: <url>
+        checksum: <method>:<digest>
+        size: <size>
+```
+
+`module_metadata` entries are needed when installing modular packages (e.g. from RHEL 8 module
+streams). Unlike `packages` and `source`, the `repoid` field is **mandatory** for module metadata.
+
+`<repoid>` or `<repoid-source>` values are not mandatory - when not specified, a random repoid name
+is generated.
+`checksum` key and `<size>` are not mandatory either - when not specified, (S)RPM file verification
+won't be performed after download.
+
+An example of the format follows.
+
+```yaml
+lockfileVersion: 1
+lockfileVendor: redhat
+arches:
+  - arch: x86_64
+    packages:
+      - repoid: fedora
+        url: https://kojipkgs.fedoraproject.org/compose/rawhide/latest-Fedora-Rawhide/compose/Everything/x86_64/os/Packages/l/libsodium-1.0.19-4.fc40.x86_64.rpm
+        checksum: sha256:e8ec4a0e9b5e9246c661d5562fc692c697f6769846de718d3c8e0929dd5ae9de
+        size: 178718
+      - repoid: fedora
+        url: https://kojipkgs.fedoraproject.org/compose/rawhide/latest-Fedora-Rawhide/compose/Everything/x86_64/os/Packages/v/vim-enhanced-9.1.113-1.fc41.x86_64.rpm
+        checksum: sha256:dca5053a4ce789b8f70e063a45e9a7b512d8d8bf5baa8e265a7f5b9ce0c90412
+        size: 1954810
+      - repoid: fedora
+        url: https://kojipkgs.fedoraproject.org/compose/rawhide/latest-Fedora-Rawhide/compose/Everything/x86_64/os/Packages/v/vim-data-9.1.113-1.fc41.noarch.rpm
+        checksum: sha256:4d3169872508b93e8ce4d6d9b7fa5ab4177cd571258248138dc1b3c8477d5c36
+        size: 23475
+      - repoid: fedora
+        url: https://kojipkgs.fedoraproject.org/compose/rawhide/latest-Fedora-Rawhide/compose/Everything/x86_64/os/Packages/x/xxd-9.1.113-1.fc41.x86_64.rpm
+        checksum: sha256:81fa409c8be8c125964787f93f13a4dc9d8f1667d8630cf8ba425ec1a266beca
+        size: 37701
+      - repoid: fedora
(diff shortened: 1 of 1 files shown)


docs/design/rpm.md
```diff
@@ -0,0 +1,276 @@
+# Hermeto support for prefetching RPMs - design
+
+## Introduction
+
+Containers want to install RPMs at build time. This is not possible when the build doesn't have
+access to the network (hermetic build). This feature is meant to enable such workflow.
```
tnevrlka: > Containers want to install RPMs at build time. This is not possible when the build doesn't have access to the network

While I get the idea, I think "download" or "prefetch" would be more accurate 

I'm not sure about the flow of the paragraph. Note I'm not native English speaker, but from the wording it feels to me like the "build doesn't have access to the network" just happened and wasn't a conscious decision.

docs/design/rpm.md
```diff
@@ -0,0 +1,276 @@
...
+
+## Block schema
+
+The image shows the main function blocks, input and outputs. In the first phase (MVP) only a subset
+of the whole functionality will be delivered. That means processing the lock file, and gathering
+(downloading) all required items (RPMs, source RPMs) into a directory structure that will represent
+input for a container build process. The basic fetch command will produce SBOM with metadata of
+fetched RPMs too.
+As a next (optional) step, the other hermeto command (`generate-env` or `inject-files`) will
+process the structure created in the previous step and will generate repositories and .repo file.
```
tnevrlka: Does `generate-env` actually do anything for RPMs?

docs/design/rpm.md
```diff
@@ -0,0 +1,276 @@
...
+
+![block_schema6](https://github.com/user-attachments/assets/c3da6cb0-47a0-4202-bd76-a59d27372e79)
```
tnevrlka: What does this link to? Is it possible it will ever break?

docs/design/rpm.md
```diff
@@ -0,0 +1,276 @@
...
+  list of URLs where packages are located and repoids under which the packages should appear in the
+  container. This means that RPMs can be taken directly from artifact storage solutions without the
+  need to create a repodata for them first. Hermeto will take care of creating repodata for the
+  build process.
+
+## Important things to keep in mind
+
+* For RPMs hermeto doesn't work with repodata, it only downloads a list of URLs as listed in the
+  *rpms.lock.yaml* file. Because of that, hermeto cannot validate repodata signatures.
+* For Modules, hermeto will likely need to access and understand the repodata. Modules are not just
```
tnevrlka: Isn't this outdated?

tnevrlka (commented):
Before continuing with review, I think it needs to be decided whether this is supposed to really be just a carbon copy of whatever document came from the discussion or whether it should be changed to reflect the current situation.
I believe there are some outdated things in the document

docs/design/rpm.md
eskultety: No.

docs/design/rpm.md
eskultety: Yep. We generate repodata for both the usual scenario as well for modularity for which we need to download modules metadata first which we `createrepo-c` then implicitly reads and uses.

docs/design/rpm.md
mohamedbouchtout: It links to a image diagram that was in the discussion post, I can change it to a markdown schematic diagram to avoid the link breaking at some point.

mohamedbouchtout:
> Before continuing with review, I think it needs to be decided whether this is supposed to really be just a carbon copy of whatever document came from the discussion or whether it should be changed to reflect the current situation. I believe there are some outdated things in the document

I think it makes more sense to update the doc to match the codebase to keep the information accurate, even if the issue meant to just copy it over.

docs/design/rpm.md
```diff
@@ -0,0 +1,302 @@
+# Hermeto support for prefetching RPMs - design
+
+## Introduction
+
+Containers require to download RPMs at build time. With hermetic builds, this is not possible
```
tnevrlka: I don't think this is grammatically correct, two options off the top of my head:
a) require -> need
b) to download -> downloading

docs/design/rpm.md
tnevrlka, line 1: Check for trailing whitespaces

docs/design/rpm.md
```diff
@@ -0,0 +1,302 @@
...
+5. The solution SHOULD be able to propagate repoids for prefetched RPMs and modules to the container
+   build.
+    1. This will make sure the yum/dnf history command shows origin repositories correctly. This may
+       be leveraged by user tools (e.g. Clair) to identify sources of the RPM content.
+6. The solution MUST collect SRPMs for source container generation.
+    1. Only applicable if the SRPMs were listed in the *rpms.lock.yaml* file.
+
+## Block schema
+
+The image shows the main function blocks, input and outputs. In the first phase (MVP) only a subset
```
tnevrlka: Is this up-to-date? Which "phase" are we currently in?

docs/design/rpm.md
```diff
@@ -0,0 +1,302 @@
...
+[<repoid>]
+baseurl = file://<for-output-dir>/deps/rpm/<arch>/<repoid>
+```
+
+Note that the repo identifiers in the filesystem structure match the identifiers in the repo file,
+and come from the lockfile. The `<for-output-dir>` is the path where the output will be mounted
+in the build container, provided via the `--for-output-dir` CLI argument.
+If `<repoid>` wasn't specified, a random value is generated. And the section will additionally
+contain the key `name` with an additional comment.
+
```
tnevrlka: This phrasing makes it seem to me like there's the key `name` AND an additional comment.
Maybe something like: _"...will additionally contain the key `name` describing the repository."_?

docs/design/rpm.md
mohamedbouchtout: I think dropping the require to/need to would sounds much better.

docs/design/rpm.md
```diff
@@ -0,0 +1,269 @@
...
+    H --> I[container build process]
+
+    style Phase1 fill:#e0e0e0,stroke:#999,color:#000
+    style Phase2 fill:#e0f2fe,stroke:#999,color:#000
+
+    classDef white fill:#fff,stroke:#000,color:#000
+    class A,B,C,D,E,F,G,H,I white
+
+    linkStyle default stroke:#000
+```
```
a-ovchinnikov, line 64: The diagram is rendered as a linear sequence of blocks ordered from left to right, WDYT about arranging it top to bottom with "hermeto fetch-deps" subelement keeping LR ordering?

mohamedbouchtout: That also works, in fact that sounds better for a markdown file IMO.

docs/design/rpm.md
```diff
@@ -0,0 +1,269 @@
...
+because the build doesn't have access to the network. This feature solves this problem by pre-fetching
+RPMs before the build process to allow for hermetic builds.
+
+The general idea is that the developer specifies ahead of time what RPMs they want to have
+available. In a step before the actual build, hermeto takes this information, downloads all of the
+packages and prepares a repository that will be available during build.
+
+## Requirements
+
+1. The solution MUST work for processing all required architectures in one process.
```
a-ovchinnikov: A small nitpick/question: do we plan to preserve the style as close to the original document as possible or are we ok with some minor style tweaks? If the latter then could you please remove upper-case highlights? This is a list of requirements anyway, it is expected that each requirement must be satisfied.

a-ovchinnikov (commented):
The document looks good to me under the assumption that it was planned to consolidate the last remaining non-standard design into the documentation section of the repository mostly as is, with minimal changes (it could be improved otherwise). Do you plan to merge all fix commits into the first one? (Prior to the first one there is nothing to fix in the codebase, so I question the necessity of keeping a chain of fix commits in this particular case).

mohamedbouchtout:
> The document looks good to me under the assumption that it was planned to consolidate the last remaining non-standard design into the documentation section of the repository mostly as is, with minimal changes (it could be improved otherwise). Do you plan to merge all fix commits into the first one? (Prior to the first one there is nothing to fix in the codebase, so I question the necessity of keeping a chain of fix commits in this particular case).

Yes all commits will get squashed to the first one. The only reason for the chain commits is to make the code review easier.

docs/design/rpm.md
mohamedbouchtout: I would say minor tweaks and style changes is fine if it makes the doc better. I'm ok with changing the upper-case highlights here.

docs/design/rpm.md
```diff
@@ -0,0 +1,302 @@
...
+<output>/deps/rpm/<arch>/<repoid>/*.rpm
+<output>/deps/rpm/<arch>/<repoid-source>/*.src.rpm
+<output>/bom.json
+```
+
+`deps/rpm` part - this part of the path is for compatibility with other package manager modules
+hermeto works with (for example `deps/npm`).
+
+The SBOM file format is CycloneDX and PURLs are generated as per
+[specification](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm).
```
eskultety: This is now an incorrect URL, the contents has moved. This wouldl be more stable for our purposes: https://github.com/package-url/purl-spec/blob/v1.0.1/docs/types/types-overview.md.
However, the RPM type spec  ^here isn't the one we implement, that's https://redhatproductsecurity.github.io/security-data-guidelines/purl/ , the official PURL spec on RPMs from that github repo was terrible unfortunately and we had to change it.

docs/design/rpm.md
```diff
@@ -0,0 +1,302 @@
...
+The prefetching step doesn't really care about dependencies. It just downloads and prepares what
+it's told to.
+
+If there is a missing dependency, there will be a failure at build time.
+
+If the lockfile lists additional packages, the build will not use them, so it's just wasting time
+and bandwidth at the prefetching step (and possibly including irrelevant sources in source
+containers).
+
+## Authors
```
eskultety: We don't usually do this. To give credit, you can mention in the commit message that this design builds on some original work and link the discussions page, that one isn't going anywhere ;) .

eskultety (commented):
Following up on

> The document looks good to me under the assumption that it was planned to consolidate the last remaining non-standard design into the documentation section of the repository mostly as is, with minimal changes (it could be improved otherwise)

Agreed 100%. The doc definitely needs more love to be more aligned with other designs we have, but for the purposes of the linked issue (just to migrate), this will do IMO for now. Still, I found 1 problem with information correctness and 1 suggestion to remove a section. Otherwise LGTM.

docs/design/rpm.md
```diff
@@ -0,0 +1,297 @@
...
+## Block schema
+
+The diagram shows the main function blocks, inputs and outputs. The fetch-deps command processes the
+lock file and downloads all required items (RPMs, source RPMs, and module metadata) into a directory
+structure that serves as input for the container build process. It also produces an SBOM with metadata
+of the fetched RPMs.
+
+As a next (optional) step, the other hermeto command (`inject-files`) will process the structure
+created in the previous step and will generate repositories and .repo file.
+
```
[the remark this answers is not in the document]
mohamedbouchtout: Seems valid.

docs/design/rpm.md
```diff
@@ -0,0 +1,302 @@
...
+        size: <size>
+```
+
+`module_metadata` entries are needed when installing modular packages (e.g. from RHEL 8 module
+streams). Unlike `packages` and `source`, the `repoid` field is **mandatory** for module metadata.
+
+`<repoid>` or `<repoid-source>` values are not mandatory - when not specified, a random repoid name
+is generated.
+`checksum` key and `<size>` are not mandatory either - when not specified, (S)RPM file verification
+won't be performed after download.
```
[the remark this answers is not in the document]
eskultety: 2 scenarios:

1. someone managed to comprimise the repo data & metadata -> checksum & size would match -> supply chain attack -> out of our control
2. someone managed to sneak in malicious URL to the lockfile -> they can sneak in checksum & size the same way

This is a weak argument, meaning that whatever we'd do the same risk would still be there.

eskultety (approved):

tnevrlka (approved):

PR #1669, opened 2026-07-23
[tests] [integration] Migrate test scenarios from the integration test repo to the main git repo vol.1

eskultety:
This is a PoC to https://github.com/hermetoproject/hermeto/issues/1668.

This PR co-locates the integration tests to the main hermeto repository under `tests/integration/<backend>/scenarios`. Each scenario has corresponding `in/` and `out/` directories for test data. Note some scenarios may share the `in/` directory in which case these are symlinked such that one scenario acts as a source donor to other similar variants (doesn't matter which one).

The test harness is modified to create a clean deterministic synthetic git repository (named `synthetic_repo` in pytest tmp worker directory) for each test scenario. The scenarios themselves reside in the respective scenario `in/` directory. It uses fixed git metadata (author, date, commit message) in order to produce reproducible SHAs (see commit 1 for the plumbing).

This PR implements the concept and performs a phase 1 migration of tests, i.e. everything **except for gomod** tests are moved over because gomod requires more effort (consumes git tag and works with git submodules) that will be taken care of in phase 2 provided this approach is agreed upon in https://github.com/hermetoproject/hermeto/issues/1668.

The obvious disadvantage of this migration is that we'd start tracking binary blobs in our main repo (e.g. tarballs, zips, etc.). We can look at optimizing for size in the future.


eskultety:
Since last revision:
- updated output test data
- harden the test harness synthetic repo logic: copy main repo's gitignore only to `.git/info/exclude` inside the synthetic git repo to prevent main repo changes to `.gitignore` invalidating many test data

eskultety:
The CI failure is quite amusing:
```
FileNotFoundError: [Errno 2] No such file or directory: 
  '/tmpuvpgfb_w.hermeto-source-copy/eggs-packages/eggs/package.json'
```
This is due to the Hermeto's `.gitignore` which contains 
```
eggs/
.eggs/
```
Fixed.

eskultety:
Ready to review.

tests/integration/utils.py
```diff
@@ -501,12 +554,15 @@ def build_image_and_check_cmd(
     assert exit_code == 0, f"Injecting project files failed. output-cmd: {output}"
 
     log.info("Build container image with all prerequisites retrieved in previous steps")
-    container_folder = test_data_dir.joinpath(test_case, "container")
+    if test_params is not None and test_params.branch is None:
+        containerfile_path = test_data_dir / test_case / "in" / test_params.containerfile
```
tnevrlka: TIL that it's possible to use slashes like this.

Is there a reason why it's inconsistent with [lines 485 and 487](https://github.com/eskultety/hermeto/blob/9936b356b7d9a896a215bf8ef6d7f9931245d0ba/tests/integration/utils.py#L484-L487)?  
`deps_content_file = Path(test_data_dir, test_case, "out", "fetch_deps_file_contents.yaml")`

And [lines 454-560](https://github.com/eskultety/hermeto/blob/9936b356b7d9a896a215bf8ef6d7f9931245d0ba/tests/integration/utils.py#L454-L460)
```python
expected_build_config_path = test_data_dir.joinpath(
    test_case, "out", ".build-config.yaml"
)
expected_sbom_path = test_data_dir.joinpath(test_case, "out", "bom.json")
```

tests/integration/utils.py
```diff
@@ -298,6 +300,47 @@ def _clone_custom_test_repo(tmp_path: Path, repo_url: str, branch: str) -> Path:
...
+    with synthetic_repo.git.custom_environment(
+        GIT_AUTHOR_NAME=SYNTHETIC_AUTHOR,
+        GIT_AUTHOR_EMAIL=SYNTHETIC_EMAIL,
+        GIT_COMMITTER_NAME=SYNTHETIC_AUTHOR,
+        GIT_COMMITTER_EMAIL=SYNTHETIC_EMAIL,
+        GIT_AUTHOR_DATE=SYNTHETIC_DATE,
+        GIT_COMMITTER_DATE=SYNTHETIC_DATE,
+    ):
+        synthetic_repo.git.add(".")
+        synthetic_repo.git.commit(m=SYNTHETIC_MESSAGE)
```
slimreaper35:
```suggestion
        synthetic_repo.git.commit("--no-gpg-sign", m=SYNTHETIC_MESSAGE)
```

slimreaper35 (commented):
Is this https://github.com/hermetoproject/integration-tests/issues/31 still relevant btw?

eskultety:
> Is this [hermetoproject/integration-tests#31](https://github.com/hermetoproject/integration-tests/issues/31) still relevant btw?

not after this migration if it's accepted (can be rejected).

tests/integration/utils.py
eskultety: I think this comes from your git config so what I think we need instead is a null git config when creating this synthetic repo instead of addopting particular options here.

eskultety:
> The obvious disadvantage of this migration is that we'd start tracking binary blobs in our main repo (e.g. tarballs, zips, etc.). We can look at optimizing for size in the future.

To provide more data on this, this PR adds additional:
- ~33MB of extra text data (test scenarios)
- ~21MB of extra binary data (in test scenarios)

Personally I'm not fond of ^this, so I'll take this back to a draft state and try employing LFS to aleviate the situation somewhat.

tests/integration/utils.py
slimreaper35: Yes, `GIT_PRISTINE_ENV` works too.

tests/integration/utils.py
```diff
@@ -450,6 +502,7 @@ def build_image_and_check_cmd(
     hermeto_image_entrypoint: str | None = None,
     fetch_output_dirname: str = DEFAULT_OUTPUT,
     env_vars_filename: str = f"{APP_NAME}.env",
+    test_params: TestParameters | None = None,
```
tnevrlka, line 510: nit: missing in docstring

tests/integration/utils.py
```diff
@@ -298,6 +300,47 @@ def _clone_custom_test_repo(tmp_path: Path, repo_url: str, branch: str) -> Path:
...
+    synthetic_repo_path = tmp_path / "repo"
+    shutil.copytree(source_dir, synthetic_repo_path)
+
+    synthetic_repo = GitRepo.init(synthetic_repo_path)
+    project_repo_root = GitRepo(Path(__file__), search_parent_directories=True).working_dir
+    gitignore = Path(project_repo_root) / ".gitignore"
+    if gitignore.is_file():
+        exclude_file = synthetic_repo_path / ".git" / "info" / "exclude"
+        exclude_file.parent.mkdir(parents=True, exist_ok=True)
+        shutil.copy2(gitignore, exclude_file)
```
tnevrlka, line 331: This was not easily readable to me on first glance.
As far as I know, `project_repo_root` seems to be only used for the `.gitignore`.
Would you consider separating it by whitespace and maybe adding a comment?

e.g.
```python
...

# Copy hermeto's .gitignore into synthetic repo 
project_repo_root = GitRepo(Path(__file__), search_parent_directories=True).working_dir
gitignore = Path(project_repo_root) / ".gitignore"
if gitignore.is_file():
    exclude_file = synthetic_repo_path / ".git" / "info" / "exclude"
    exclude_file.parent.mkdir(parents=True, exist_ok=True)
    shutil.copy2(gitignore, exclude_file)

synthetic_repo = GitRepo.init(synthetic_repo_path)
with synthetic_repo.git.custom_environment(
...
```

tnevrlka (commented):
Approach is okay in my opinion, after some research monorepo makes sense to me.

There's an open question on what to do with the migrated tests. There are quite a few useless files adding quite a few MBs that should maybe be trimmed or handled somehow. Otherwise, renaming leftover cachi2 to hermeto might be considered. 

For other test cleanup, I think it's fine to leave for another PR

tests/integration/conftest.py
```diff
@@ -35,6 +35,10 @@
     ("HERMETO_TEST_LOCAL_NEXUS_NO_CLEANUP", "--hermeto-local-nexus-no-cleanup"),
 ]
 
+# Scenario in/ dirs contain real project sources (e.g. Python packages for pip) that pytest
+# would otherwise try to collect and import, causing ModuleNotFoundError during collection.
+collect_ignore_glob = ["*/scenarios/*/in/*"]
```
tnevrlka: `["*/scenarios/*/in/*"]` -> `["*/scenarios/*/in/**"]`

tests/integration/utils.py
eskultety: I fixed the 
`containerfile_path = test_data_dir / test_case / "in" / test_params.containerfile` line locally, that's definitely an oversight on my side. However, when it comes to 

> inconsistent with [lines 485 and 487](https://github.com/eskultety/hermeto/blob/9936b356b7d9a896a215bf8ef6d7f9931245d0ba/tests/integration/utils.py#L484-L487)

that's pre-existing inconsistency, the diff is actually consistent in that `deps_content_file` is constructed using the original inconsistent pattern, it's just branching was added. It's one of those things where I would prefer keeping the existing inconsistency to keep the diff predictable and simple and open a different PR addressing these inconsistencies on module or even repo level. Does that sound reasonable?

tnevrlka: Okay, resolving

tnevrlka (commented):
Draft looks good to me, I think it can be moved to ready for review

tests/integration/utils.py
```diff
@@ -85,8 +87,8 @@ def _env_to_engine_flags(env: Mapping[str, str] | None) -> list[str]:
 
 @dataclass
 class TestParameters:
-    branch: str
     packages: tuple[dict[str, Any], ...]
+    branch: str | None = None
```
a-ovchinnikov, line 91: This is kept around for backwards compatibility, right?

tests/integration/rpm/scenarios/rpm_dnf_tls_client_auth/in/README.md
```diff
@@ -0,0 +1,9 @@
+# Cachi2 DNF server TLS client authentication test
```
a-ovchinnikov, line 1: Should Cachi2 be replaced?

a-ovchinnikov (commented):
When I was thinking about this change I envisioned a directory with tarballs, each representing an integration test. I like this approach more, LGTM.

tests/integration/utils.py
eskultety: yes, branch will be deleted once all tests are migrated.

tests/integration/rpm/scenarios/rpm_dnf_tls_client_auth/in/README.md
eskultety: Originally I wanted to introduce small fixes exactly like this one in a follow-up PR, but since I'll be introducing other changes (to keep the binary blob count at bay) I can squeeze this one in, no problem.

eskultety:
Since last revision:
 
>     * ~21MB of extra binary data (in test scenarios)
> 
> 
> Personally I'm not fond of ^this, so I'll take this back to a draft state and try employing LFS to aleviate the situation somewhat.

- experimented with LFS, but then remembered there will be some incurred [quota](https://docs.github.com/en/billing/concepts/product-billing/git-lfs#free-use-of-git-lfs) from GH which means once depleted, our CI would stop working (more so due to the bandwidth limit rather than storage). I did work with old data I had in the back of my head though - 1GB for both - only to later find out it's 10GB for the free tier, so I ditched the idea whilst working on this (we can still push whatever is left there :) , it's a few KB)
- extracted some of the changes I had originally made to the migration commits such that the migration would truly be 1:1
  - it's cannot be 1:1 really, because the output data + Containerfiles lived in the main repo and without migrating in the same commit, individual commits would not be testable, so that had to go in
- dropped a whole batch of tarballs for the Yarn zero install test which is a negative test that just checks if there's any ZIP in the yarn cache yet it cumulatively carried 19MB out of those ~21MB measured initially
- inspected the blobs tests needed and replaced the bigger blobs with tiny tarballs (few KBs), you can see that in the commit log, let me know if I need to drop some of them :)
- symlinked identical blobs so that only a single copy is stored
- dropped some useless PNG file (thanks @tnevrlka for spotting)

Commits marked as SQUASH purposefully don't have DCO to fail the check since they're supposed to be squashed to respective migration commits before merging and it's out of courtesy to the reviewers. Every commit should pass integration CI, please double check. Let me know if some test changes wrt/ blobs should be left out if you think they're pointless.

Total tarball blob storage in this revision + dedup - ~13KB.
Ready to review again.

tests/integration/utils.py
```diff
@@ -432,7 +486,10 @@ def _fmt(code: int) -> str:
         schema = _fetch_cyclone_dx_schema()
         jsonschema.validate(instance=sbom, schema=schema)
 
-    deps_content_file = Path(test_data_dir, test_case, "fetch_deps_file_contents.yaml")
+    if test_params.branch is None:
+        deps_content_file = Path(test_data_dir, test_case, "out", "fetch_deps_file_contents.yaml")
+    else:
+        deps_content_file = Path(test_data_dir, test_case, "fetch_deps_file_contents.yaml")
```
slimreaper35, line 492: I didn't even know we had such a file in the repo. I see only one though, which makes me feel like it is some kind of leftover.

tests/integration/cargo/scenarios/cargo_e2e/in/heck/README.md
```diff
@@ -1,6 +1,5 @@
 # **heck** is a case conversion library
 
-!["I specifically requested the opposite of this."](./no_step_on_snek.png)
```
slimreaper35: xD

slimreaper35 (commented):
nitpick: For discovery reasons, maybe those symlinked tarballs in could be in higher-level directory e.g `hermeto/tests/integration/npm/shared` ?

eskultety:
> nitpick: For discovery reasons, maybe those symlinked tarballs in could be in higher-level directory e.g `hermeto/tests/integration/npm/shared` ?

I tried a similar thing in my first draft before posting (or better said, that's what AI came up with on its own without rigorous prompting) and it turned out to be a mess IMO when I opened the scenarios directory. That's where I stepped in and did it myself as I envisioned it. We could do that, but how important is it to know a file is symlinked at to what? Probably not that much and once you do `ls -l` would show you exactly. The git history would also be transparent about it.

In any case, we can do that in a follow-up, can't we? I'd rather not touch the files more than absolutely necessary in this series.

tests/integration/yarn/scenarios/yarn_missing_lockfile/in/.yarn/releases/yarn-3.6.1.cjs
tnevrlka, line 1: Is it viable to symlink the yarn `.cjs` files as well?

tests/integration/utils.py
eskultety: We have https://github.com/hermetoproject/hermeto/issues/803 so anything suspicious can be done as part of that work.

tnevrlka (commented):
The snek has been stepped on. LGTM

tests/integration/yarn/scenarios/yarn_missing_lockfile/in/.yarn/releases/yarn-3.6.1.cjs
eskultety: Sure, that's a good point, will do for the identical ones.

eskultety: Done.

tnevrlka (approved):
Squash and send it 🚀

a-ovchinnikov:
> Total tarball blob storage in this revision + dedup - ~13KB.

Wow, that's about 1.5k times less than before

tnevrlka:
image

Well that's not ideal  @eskultety https://github.com/hermetoproject/hermeto/actions/runs/30537694959/job/90854838665?pr=1669

eskultety:
> Well that's not ideal @eskultety https://github.com/hermetoproject/hermeto/actions/runs/30537694959/job/90854838665?pr=1669

The CI crashed at spawning Nexus, I didn't touch that part, I re-run the job, it's currently past Nexus stage and running tests: https://github.com/hermetoproject/hermeto/actions/runs/30537694959/job/90920083048

a-ovchinnikov (approved):
LGTM!

slimreaper35 (approved):

eskultety:
Squashed, will wait for CI and merge.

PR #1671, opened 2026-07-24
enhancement(workflow): add a github workflow that removes a stale pr/issue

mohamedbouchtout:
### Summery
This workflow will mark a PR or issue stale if it has been inactive for 30 days (no comments or pushes), then closes it after 7 days. This will help keep the PR and issue backlog up-to-date.

### Discussion
It may be worth adding exemptions to the workflow to not close a PR or issue if it's assigned to a specific person or it has a specific tag/milestone.

### Source
[Here](https://github.com/actions/stale) is the source of the github action.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,23 @@
+name: Mark stale issues and PRs and close them
+
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v10
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 7
+          days-before-pr-stale: 30
+          days-before-pr-close: 7
+          stale-issue-label: 'no-issue-activity'
+          stale-pr-label: 'no-pr-activity'
+          stale-issue-message: 'This issue is stale because it has been inactive for 30 days. Please reply or it will be closed in 7 days.'
+          stale-pr-message: 'This PR is stale because it has been inactive for 30 days. Please reply or it will be closed in 7 days.'


a-ovchinnikov:
Please don't auto-close issues! Quite a few of them remain dormant for month, especially the lower priority ones. Just posting something now and then to keep them from being harvested feels counterproductive to me.

mohamedbouchtout:
> Please don't auto-close issues! Quite a few of them remain dormant for month, especially the lower priority ones. Just posting something now and then to keep them from being harvested feels counterproductive to me.

Oh yes that makes sense, just wanted to draft an idea I had for the PRs and issues that have been stale for long. It's possible to add tags to issues and PRs so they don't get closed regardless.

Harsh-2005d:
I'm not sure the repository is large enough to benefit from automatically closing stale issues or PRs. If an issue is no longer relevant, maintainers can close it once they determine that's the case. If it hasn't been resolved, inactivity alone doesn't make it any less valuable. The same applies to PRs. If reviewers decide a PR isn't something the project wants to merge, it can be closed with that decision. Otherwise, keeping it open preserves the discussion and allows work to continue later without losing context. I think a workflow like this is more beneficial for repositories that receive a much larger volume of issues and PRs than Hermeto currently does.

slimreaper35:
Unfortunately, this is not a way to go.

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - uses: actions/stale@v10
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 7
+          days-before-pr-stale: 30
+          days-before-pr-close: 7
+          stale-issue-label: 'no-issue-activity'
+          stale-pr-label: 'no-pr-activity'
+          stale-issue-message: 'This issue is stale because it has been inactive for 30 days. Please reply or it will be closed in 7 days.'
+          stale-pr-message: 'This PR is stale because it has been inactive for 30 days. Please reply or it will be closed in 7 days.'
```
tnevrlka, line 23: I'm curious, what would happen if closing was disabled, but the stale label was kept? 
Is it possible to remove the label manually or mark the PR/issue as never stale? I'm afraid that if the label was removed manually, the action would add it back during every run

mohamedbouchtout: Yes you can set 'days-before-pr-stale' and 'days-before-issue-stale' to -1 and it will never close them. You can remove the stale label and the workflow will put it back after the 30 day stale timer ends, since removing the label resets its inactivity.

tnevrlka: > since removing the label resets its inactivity

That's pretty good. Any chance it's possible to override an issue/PR to never get stale?

mohamedbouchtout: The ways you can do that is to create an exempt label, exempt assignee, milestones, or draft PRs.

PR #1674, opened 2026-07-27
Artifact registry proxy support in cargo

a-ovchinnikov:
This PR introduces support for artifact registry proxies in cargo backend.

### DIFF
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -133,7 +133,7 @@ class LocalCargoPackage:
         return Component(name=self.name, version=self.version, purl=self.purl.to_string())
 
 
-def fetch_cargo_source(request: Request) -> RequestOutput:
+def fetch_cargo_source(request: Request, invoked_through_pip: bool = False) -> RequestOutput:
     """Fetch the source code for all cargo packages specified in a request."""
     components: list[Component] = []
     project_files: list[ProjectFile] = []
@@ -151,7 +151,7 @@ def fetch_cargo_source(request: Request) -> RequestOutput:
                 vendor_result.config_template
             )
             project_files.append(_use_vendored_sources(package_dir, config_template))
-        package_components = _generate_sbom_components(package_dir, request)
+        package_components = _generate_sbom_components(package_dir, request, invoked_through_pip)
 
         if vendor_result.lockfile_was_generated:
             _update_permissive_mode_annotation(annotations, package_components)
@@ -419,7 +419,11 @@ def _find_local_packages(package_dir: RootedPath) -> dict[str, str]:
     return result
 
 
-def _generate_sbom_components(package_dir: RootedPath, request: Request) -> list[Component]:
+def _generate_sbom_components(
+    package_dir: RootedPath,
+    request: Request,
+    invoked_through_pip: bool = False,
+) -> list[Component]:
     """Generate SBOM components from Cargo.lock and for the main package."""
     parsed_lockfile = _parse_toml_project_file(package_dir.path / "Cargo.lock")
 
@@ -451,14 +455,21 @@ def _generate_sbom_components(package_dir: RootedPath, request: Request) -> list
         pkg_version = pkg.get("version")
 
         if pkg_name == main_package_name:
-            components.append(
-                LocalCargoPackage(
-                    name=main_package_name,
-                    version=main_package_version,
-                    vcs_url=vcs_url,
-                    subpath=str(package_dir.path.relative_to(package_dir.root)),
-                ).to_component()
-            )
+            if invoked_through_pip:
+                # The package was collected as a part of processing a python dependency,
+                # it has been already reported when collected with pip, so here it must
+                # be ignored.
+                pass
+            else:
+                components.append(
+                    LocalCargoPackage(
+                        name=main_package_name,
+                        version=main_package_version,
+                        vcs_url=vcs_url,
+                        subpath=str(package_dir.path.relative_to(package_dir.root)),
+                    ).to_component()
+                )
+
         elif pkg_name in local_packages:
             # Local packages have no other fields in the Cargo.lock file besides the name and version.
             components.append(
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -1,11 +1,12 @@
 # SPDX-License-Identifier: GPL-3.0-only
+import base64
 import logging
 import os
 import subprocess
 from collections.abc import Generator
 from contextlib import contextmanager
 from dataclasses import dataclass
-from functools import cached_property
+from functools import cache, cached_property
 from pathlib import Path
 from typing import Any, ClassVar, NamedTuple
 from urllib.parse import urlparse, urlunparse
@@ -13,9 +14,10 @@ from urllib.parse import urlparse, urlunparse
 import tomlkit
 import tomlkit.exceptions
 from packageurl import PackageURL
+from pydantic import HttpUrl
 
 from hermeto import APP_NAME
-from hermeto.core.config import get_config
+from hermeto.core.config import CargoSettings, get_config
 from hermeto.core.constants import Mode
 from hermeto.core.errors import (
     ExitError,
@@ -27,7 +29,13 @@ from hermeto.core.errors import (
 )
 from hermeto.core.models.input import Request
 from hermeto.core.models.output import Annotation, Component, ProjectFile, RequestOutput
-from hermeto.core.models.sbom import create_backend_annotation, spdx_now
+from hermeto.core.models.sbom import (
+    PROXY_COMMENT,
+    PROXY_REF_TYPE,
+    ExternalReference,
+    create_backend_annotation,
+    spdx_now,
+)
 from hermeto.core.rooted_path import RootedPath
 from hermeto.core.scm import get_repo_id
 from hermeto.core.utils import run_cmd
@@ -70,6 +78,7 @@ class CargoPackage:
     version: str
     source: str | None = None  # [git|registry]+https://github.com/<org>/<package>#[|<sha>]
     checksum: str | None = None
+    proxy: HttpUrl | None = None
 
     @cached_property
     def purl(self) -> PackageURL:
@@ -95,9 +104,32 @@ class CargoPackage:
 
         return PackageURL(type="cargo", name=self.name, version=self.version, qualifiers=qualifiers)
 
+    @property
+    def _is_proxied(self) -> bool:
+        # Custom registries are not proxied, so only those which actually use
+        # proxy_url should be reported, vcs_urls are not proxied.
+        # Note, that crates.io gets replaced with proxy_url on .cargo/config.toml level
+        if self.proxy is None:
+            return False
+        if self.source is None:
+            # This can happen to some Rust dependencies for Python project,
+            # e.g. cryptography-cffi@0.1.0. This is not a local package, those
+            # are handled elsewhere.
+            return True
+        if self.source.startswith("git+"):
+            return False
+        return "crates.io" in self.source
+
     def to_component(self) -> Component:
         """Convert CargoPackage into SBOM component."""
-        return Component(name=self.name, version=self.version, purl=self.purl.to_string())
+        ref_rest = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
+        proxy = [ExternalReference(url=str(self.proxy), **ref_rest)] if self._is_proxied else None
+        return Component(
+            name=self.name,
+            version=self.version,
+            purl=self.purl.to_string(),
+            external_references=proxy,
+        )
 
 
 @dataclass
@@ -198,9 +230,25 @@ def _fetch_dependencies(package_dir: RootedPath, request: Request) -> CargoVendo
     #                    It is necessary to make Cargo keep dependencies that are already
     #                    present in the vendored directory. This flag has no effect on standalone
     #                    cargo operations however is crucial when it is invoked from pip.
-    cmd = ["cargo", "vendor", "--locked", "--versioned-dirs", "--no-delete", str(vendor_dir)]
+    # --respect-source-config tells cargo to respect config in .cargo/config.toml in the repository.
+    #                         Is necessary when working through a proxy or when custom registries
+    #                         must be used.
+    cmd = [
+        "cargo",
+        "vendor",
+        "--locked",
+        "--versioned-dirs",
+        "--no-delete",
+        "--respect-source-config",
+        str(vendor_dir),
+    ]
     log.info("Fetching cargo dependencies at %s", package_dir)
-    with _sanitized_cargo_config_file(package_dir):
+    # NOTE: ordering is important here, a config must be sanitized first, extended to use
+    # a proxy after that, otherwise proxy data will be scrubbed by the sanitizer.
+    with (
+        _sanitized_cargo_config_file(package_dir),
+        _inject_proxy_configuration_into_config_if_needed(package_dir),
+    ):
         # Prevent Cargo from invoking rustc
         env = {"CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS": "allow"}
         # The necessary configuration to use the vendored sources will be printed to STDOUT.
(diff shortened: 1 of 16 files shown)


hermeto/core/package_managers/cargo/main.py
```diff
@@ -253,6 +296,64 @@ def _verify_lockfile_is_present(package_dir: RootedPath) -> None:
...
+    return package_dir.join_within_root(Path(cfn)).path
+
+
+@contextmanager
+def _inject_proxy_configuration_into_config_if_needed(
+    package_dir: RootedPath,
+) -> Generator[None, None, None]:
+    hermeto_configuration = get_config().cargo
+    if hermeto_configuration.proxy_url is None:
+        return
```
tnevrlka, line 338: I'm getting several `RuntimeError: generator didn't yield` errors when running the integration tests without setting up proxy. Main branch passes fine.

```
❯ nox -s integration-tests -- -k cargo
[...]
platform darwin -- Python 3.14.6, pytest-9.1.1, pluggy-1.6.0
Effective Hermeto test environment:
  HERMETO_TEST_INTEGRATION_TESTS_REPO=https://github.com/hermetoproject/integration-tests.git
  HERMETO_TEST_IMAGE=
  HERMETO_TEST_GENERATE_DATA=0
  HERMETO_TEST_CONTAINER_ENGINE=podman
  HERMETO_TEST_LOCAL_NEXUS=0
  HERMETO_TEST_LOCAL_NEXUS_PROXY=0
  HERMETO_TEST_LOCAL_NEXUS_NO_CLEANUP=0
[...]
```

a-ovchinnikov: Good catch, thank you! This is also a clear typing error, I wonder how come mypy did not point it out?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -253,6 +296,64 @@ def _verify_lockfile_is_present(package_dir: RootedPath) -> None:
...
+    if p.exists():
+        return False
+    p.parent.mkdir(parents=True, exist_ok=True)
+    p.write_text("")
+    return True
+
+
+def _write_to_cargo_config(package_dir: RootedPath, config_value: str) -> None:
+    p = package_dir.join_within_root(Path(".cargo/config.toml")).path
+    p.write_text(config_value)
```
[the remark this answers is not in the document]
a-ovchinnikov: Not anymore

tests/unit/package_managers/cargo/test_main.py
```diff
@@ -143,6 +143,7 @@ def test_cargo_package_purl_generation(pkg: dict[str, Any], expected_purl: str)
             """,
             textwrap.dedent(
                 """
+                [registries]
```
tnevrlka, line 146: Curious, what does this do?

hermeto/core/package_managers/cargo/main.py
tnevrlka: > I wonder how come mypy did not point it out?

¯\\\_(ツ)_/¯ 

Tests are passing now 🚀

tests/integration/test_cargo.py
```diff
@@ -61,6 +62,10 @@
                 check_output=False,
             ),
             id="cargo_missing_lockfile_permissive_mode",
+            marks=pytest.mark.skipif(
+                os.environ.get("HERMETO_TEST_LOCAL_NEXUS", "0") == "1",
```
tnevrlka: nit: `os.getenv()` seems to be used more often with `HERMETO_TEST_LOCAL_NEXUS`

hermeto/core/package_managers/cargo/main.py
```diff
@@ -253,6 +296,65 @@ def _verify_lockfile_is_present(package_dir: RootedPath) -> None:
...
+
+
+@contextmanager
+def _inject_proxy_configuration_into_config_if_needed(
+    package_dir: RootedPath,
+) -> Generator[None, None, None]:
+    hermeto_configuration = get_config().cargo
+    if hermeto_configuration.proxy_url is None:
+        yield
+    else:
```
tnevrlka: nit: Would early return work? I don't like nesting, but I guess it's a matter of taste
```python
[...]
if hermeto_configuration.proxy_url is None:
    yield
    return
    
config_did_not_exist_before_injection = _create_cargo_config_if_missing_in(package_dir)
[...]
```

hermeto/core/package_managers/cargo/main.py
```diff
@@ -97,7 +106,27 @@ def purl(self) -> PackageURL:
...
-        return Component(name=self.name, version=self.version, purl=self.purl.to_string())
+        ref_rest = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
+        proxy = None
+        # vcs_urls are not proxied
+        if self.proxy:
+            if self.source is not None and not self.source.startswith("git+"):
+                # Custom registries are not replaced, so only those which actually use
+                # proxy_url should be reported.
+                # Note, that crates.io gets replaced with proxy_url on .cargo/config.toml level
+                if str("crates.io") in str(self.source):
```
tnevrlka: nit: does `str("crates.io")` do anything?

tests/unit/package_managers/cargo/test_main.py
a-ovchinnikov: This is an apparently optional header which, judging from my experiments with cargo can be either omitted or present. It was missing previously, but I explicitly added it at some point because I found that it makes it slightly easier for me to navigate in generated configs.

tests/integration/test_cargo.py
a-ovchinnikov: Done.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: It will, I just don't like the `yield..return` sandwich.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Nope, thanks.

tests/unit/package_managers/cargo/test_main.py
tnevrlka: Thanks

hermeto/core/package_managers/cargo/main.py
tnevrlka: A deadlock... I'll stand down

tests/nexusserver/nginx.conf
```diff
@@ -11,7 +11,7 @@ server {
     location / {
         proxy_pass http://nexus:8081;
         proxy_set_header Authorization "";
-        proxy_set_header Host $host;
+        proxy_set_header Host $http_host;
```
taylormadore, line 14: nit: This change might be worth a separate commit explaining why this was necessary

tests/integration/test_cargo.py
```diff
@@ -61,6 +62,10 @@
                 check_output=False,
             ),
             id="cargo_missing_lockfile_permissive_mode",
+            marks=pytest.mark.skipif(
```
taylormadore: Can you remind me why?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -253,6 +296,65 @@ def _verify_lockfile_is_present(package_dir: RootedPath) -> None:
...
+@contextmanager
+def _inject_proxy_configuration_into_config_if_needed(
+    package_dir: RootedPath,
+) -> Generator[None, None, None]:
+    hermeto_configuration = get_config().cargo
+    if hermeto_configuration.proxy_url is None:
+        yield
+    else:
+        config_did_not_exist_before_injection = _create_cargo_config_if_missing_in(package_dir)
+        package_config = _path_to_package_config(package_dir)
```
taylormadore: Would it be simpler to compute this once early and then pass it to all of the helpers rather than calling it repeatedly? _write_to_cargo_config would then become unnecessary

tnevrlka (approved):

tests/integration/pip/scenarios/pip_e2e_rust_extensions/out/bom.json
```diff
@@ -204,7 +204,6 @@
         "pkg:cargo/lzxd@0.2.5?checksum=5de7336a183103429ad66d11d56d8bdc9c4a2916f6b85a8f11e5b127bde12001",
         "pkg:cargo/mailparse@0.15.0?checksum=3da03d5980411a724e8aaf7b61a7b5e386ec55a7fb49ee3d0ff79efc7e5e7c7e",
         "pkg:cargo/matchers@0.1.0?checksum=8263075bb86c5a1b1427b5ae862e8889656f126e9f77c484496e8b47cf5c5558",
-        "pkg:cargo/maturin@1.8.3",
```
slimreaper35, line 207: What happened here ?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -316,7 +436,7 @@ def _sanitize_cargo_config(config_content: str) -> str:
         raise UnexpectedFormat("Cargo config file contains invalid data and cannot be parsed")
 
     allowed_fields = {"index", "token", "credential-provider"}
-    filtered_registries = tomlkit.table()
+    filtered_registries = tomlkit.table(False)
```
slimreaper35:
```suggestion
    filtered_registries = tomlkit.table(False)
    filtered_registries = tomlkit.table(is_super_table=False)
```

tests/integration/pip/scenarios/pip_e2e_rust_extensions/out/bom.json
a-ovchinnikov: A duplicate entry happened here quite some time ago: maturin gets counted twice, as a pip package and then also as a cargo package.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Ok, this does look better.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -253,6 +296,65 @@ def _verify_lockfile_is_present(package_dir: RootedPath) -> None:
...
+    # Refer to https://doc.rust-lang.org/cargo/reference/config.html for further details.
+    # Since we could potentially end up building a somewhat stale Rust-based
+    # Python extension it is better to check if there is an old-style config present and
+    # process it if found.
+    cfn = ".cargo/config" if _old_style_config_is_present_in(package_dir) else ".cargo/config.toml"
+    return package_dir.join_within_root(Path(cfn)).path
+
+
+@contextmanager
+def _inject_proxy_configuration_into_config_if_needed(
```
taylormadore, line 332: This context manager is very dependent on the _sanitized_cargo_config_file one from an ordering perspective. We should probably make that more explicit with at least a comment somewhere

hermeto/core/package_managers/cargo/main.py
```diff
@@ -253,6 +296,65 @@ def _verify_lockfile_is_present(package_dir: RootedPath) -> None:
...
+    else:
+        config_did_not_exist_before_injection = _create_cargo_config_if_missing_in(package_dir)
+        package_config = _path_to_package_config(package_dir)
+        parsed = tomlkit.parse(package_config.read_text())
+        modified_registries_section = _inject_cargo_proxy_registry(parsed.get("registries", {}))
+        updated_config = tomlkit.document()
+        updated_config["registries"] = modified_registries_section
+        updated_config.add("source", {"crates-io": {"replace-with": "cargo-proxy"}})
+        # cargo must be told explicitly to use tokens for authentication.
+        # Note, that the token itself should be provided elsewhere, the most
```
taylormadore: nit: Should we comment on the environment given that we didn't take that approach in the implementation?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -97,7 +106,27 @@ def purl(self) -> PackageURL:
 
     def to_component(self) -> Component:
         """Convert CargoPackage into SBOM component."""
-        return Component(name=self.name, version=self.version, purl=self.purl.to_string())
+        ref_rest = dict(type=PROXY_REF_TYPE, comment=PROXY_COMMENT)
+        proxy = None
+        # vcs_urls are not proxied
+        if self.proxy:
```
taylormadore: Optional suggestion. A short helper might make this a little easier to read:
```
def _is_proxied(self):
    if not self.proxy:
        return False
    if self.source is None:
        return True   
    if self.source.startswith("git+"):
        return False  
    return "crates.io" in self.source
```

hermeto/core/package_managers/cargo/main.py
```diff
@@ -253,6 +296,65 @@ def _verify_lockfile_is_present(package_dir: RootedPath) -> None:
...
+    return package_dir.join_within_root(Path(cfn)).path
+
+
+@contextmanager
+def _inject_proxy_configuration_into_config_if_needed(
+    package_dir: RootedPath,
+) -> Generator[None, None, None]:
+    hermeto_configuration = get_config().cargo
+    if hermeto_configuration.proxy_url is None:
+        yield
```
taylormadore, line 337: nit: Can we return after yield to drop a level of nesting?

hermeto/core/package_managers/cargo/main.py
```diff
@@ -253,6 +296,65 @@ def _verify_lockfile_is_present(package_dir: RootedPath) -> None:
...
+    else:
+        config_did_not_exist_before_injection = _create_cargo_config_if_missing_in(package_dir)
```
taylormadore: optional naming suggestion: `config_was_absent = _touch_cargo_config()`

hermeto/core/package_managers/cargo/main.py
```diff
@@ -300,6 +402,24 @@ def _sanitized_cargo_config_file(package_dir: RootedPath) -> Generator[None, Non
                 config.path.write_text(data)
 
 
+def _make_basic_token_from_proxy_credential(cargo_config: CargoSettings) -> str:
+    credentials = f"{cargo_config.proxy_login}:{cargo_config.proxy_password}"
+    token = base64.b64encode(credentials.encode("utf-8")).decode("utf-8")
+    return f"Basic {token}"
+
+
+def _inject_cargo_proxy_registry(partial_cargo_config: dict) -> dict:
```
taylormadore, line 410: Optional suggestion:
```
def _inject_cargo_proxy_registry(existing_registries, cargo_config):
    proxy_entry = {"index": f"sparse+{cargo_config.proxy_url}"}
    if cargo_config.proxy_login is not None:
        proxy_entry["token"] = _make_basic_token_from_proxy_credential(cargo_config)
    return {**existing_registries, "cargo-proxy": proxy_entry}
```

taylormadore (commented):
LGTM from a functional perspective. Is there anything we need to add tests for?

hermeto/core/package_managers/cargo/main.py
tnevrlka: https://github.com/hermetoproject/hermeto/pull/1674#discussion_r3666931229

tests/integration/pip/scenarios/pip_e2e_rust_extensions/out/bom.json
taylormadore: Yeah, we were discussing this offline yesterday. Maturin is sourced as a pip package, so we probably shouldn't also be reporting it as a cargo registry package

slimreaper35: Okay, the `invoked_through_pip` fix would deserve a separate commit.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: I personally don't like sandwiching `yield..return`, however now I have two voices in favor of it, I will consider adding it.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Same meaning, but shorter, thank you!

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: That's a leftover from a previous iteration, I will remove it.

tests/integration/test_cargo.py
a-ovchinnikov: This is a left-over from the unpinned Nexus era: back then Nexus could not return an index properly which resulted in a need to inject the index; a lock update would overwrite it and break the test. Thank you for spotting this, I will remove this condition since the pinned version seems to work fine.

tests/nexusserver/nginx.conf
a-ovchinnikov: Fair enough.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -70,6 +78,7 @@ class CargoPackage:
     version: str
     source: str | None = None  # [git|registry]+https://github.com/<org>/<package>#[|<sha>]
     checksum: str | None = None
+    proxy: HttpUrl | None = None  # proxy URL (if present)
```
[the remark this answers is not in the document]
a-ovchinnikov: You are absolutely right! Let me re-phrase this line.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Done.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: It should not be dependent, I tried formulating it in such a way that they could be invoked in any order. I gave it a try and reordering CMs does not break any IT.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: `_write_to_cargo_config()` is not needed indeed, I added a cache to avoid re-computing the path on each call.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Injection is unconditional so it must check if anything has to be done to the registries, i.e. check if proxy_url is set. Config is a global singleton object, I don't think it is worth it passing it around.

tests/integration/pip/scenarios/pip_e2e_rust_extensions/out/bom.json
a-ovchinnikov: Ok, fair enough.

hermeto/core/package_managers/cargo/main.py
taylormadore: If I use `_inject_proxy_configuration_into_config_if_needed` first, then won't `_sanitized_cargo_config_file` strip the source block we added in `_inject_proxy_configuration_into_config_if_needed`? Then the modified cargo config produced by `_inject_proxy_configuration_into_config_if_needed` gets restored. It falls-back silently to no-proxy

a-ovchinnikov: Right, and the tests pass because there appears to be no good way of actually querying cargo where it got the crates from. Great point, thank you, will update the code.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: Since there were two voices in favor of the sandwich the sandwich wins.

taylormadore (approved):

Issue #1641, opened 2026-06-30
Add OCI artifact support to generic prefetcher


cfergeau:
### Requested new feature or change

We currently uses `oras` and OCI artifacts to build artifacts in a konflux pipeline, and reuse these artifacts in another pipeline. If we switch to hermetic builds, there is no easy way of prefetching these artifacts, having support for OCI artifacts in hermeto would help. It might even be possible to fetch SBOM details at the same time if these are available.

NB: I’m mentioning the generic prefetcher here, but a dedicated OCI prefetcher would also be good.

eskultety:
@cfergeau thanks for reaching out. Yeah, I can see how OCI artifacts can become increasingly popular in various scenarios. I don't know when we'll get to this, but the least I can say is that this makes sense for us to investigate and consider the options at some point.

eskultety:
@cfergeau hmm, out of curiosity, can you enumerate a list of limitations where the generic backend falls short for your needs?

cfergeau:
> [@cfergeau](https://github.com/cfergeau) hmm, out of curiosity, can you enumerate a list of limitations where the generic backend falls short for your needs?

During konflux builds, we have easy access to a container registry for pushing the artifacts we built, so it makes sense to use this. The generic prefetcher only support https URLs (I think?), and I don’t know if we can easily get a https:// URL to fetch artifacts we pushed to a registry? I also don’t know if as an alternative, konflux gives us a `https` location where we can easily publish our artifacts.

vyasgun:
I found a related issue about adding support for `oci://` URLs in rpm lockfile: https://github.com/hermetoproject/hermeto/issues/1080. I think the OCI pull/verify logic could be shared between the RPM and generic fetchers.

eskultety:
> I found a related issue about adding support for `oci://` URLs in rpm lockfile: [#1080](https://github.com/hermetoproject/hermeto/issues/1080). I think the OCI pull/verify logic could be shared between the RPM and generic fetchers.

Adding OCI capabilities to RPMs seems unlikely at least at the moment. That backend will be replaced with a DNF one instead.

vyasgun:
Noted - adding OCI support in rpm lockfiles is not a priority. However, for our specific Konflux use case and security goals, it is important to have a way to prefetch these verified konflux built OCI artifacts instead of referencing GitHub release URLs.

eskultety:
> However, for our specific Konflux use case

[...]

PR #1678, opened 2026-07-28
docs: Add design document for x-oci package manager

jangel97:
## Summary

Design document for a new experimental `x-oci` backend that prefetches OCI container images from registries during the dependency prefetch step, so hermetic builds can consume them without network access.

This PR contains only the design document. Implementation will follow in small PRs once the design is agreed upon.

Ref: #1641, #1677

## What the design covers

- Lockfile format (`oci-images.lock.yaml`) with digest-pinned image entries
- OCI Distribution API client (token auth, manifest fetch, image index handling, blob download)
- Output as OCI Image Layout directories under `deps/x-oci/`
- SBOM representation using `pkg:oci` PURLs
- Scope boundaries (what's in, what's out)

## Motivating use case

Build pipelines that consume base images or pre-built container images from registries (e.g. Red Hat UBI, NVIDIA CUDA). In a hermetic build environment, these images must be fetched in advance and made available on disk.

## Test plan

- [ ] Design review by maintainers
- [ ] Iterate based on feedback

### DIFF
--- a/docs/design/oci.md
+++ b/docs/design/oci.md
@@ -1,0 +1,247 @@
+# OCI artifacts design document
+
+## Overview
+
+The [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md)
+defines an HTTP API for pushing and pulling content from registries. The same API serves container
+images and [arbitrary artifacts](https://github.com/opencontainers/image-spec/blob/main/manifest.md)
+(signatures, SBOMs, Helm charts, Wasm modules, ML models, etc.). This backend allows Hermeto to
+prefetch any of these during the dependency prefetch step, so that hermetic builds can consume them
+without network access.
+
+**Motivating use case**: build pipelines that need OCI artifacts at build time. A build might need
+a Helm chart to template, a Wasm plugin to embed, a base image to layer on top of, or any other
+content stored in an OCI registry. In a hermetic build environment, none of these are reachable
+over the network, so they must be prefetched and made available on disk.
+
+### Developer workflow
+
+1. The developer identifies the OCI images or artifacts their build needs.
+2. The developer resolves each reference to a digest using standard tooling:
+   ```sh
+   skopeo inspect --raw docker://registry.example.com/my-image:1.0 | sha256sum
+   # or
+   crane digest registry.example.com/my-image:1.0
+   ```
+3. The developer declares each dependency in `oci-images.lock.yaml` with its digest
+   and repository.
+4. Hermeto prefetches the declared images, producing an [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)
+   directory for each.
+5. The build environment consumes the layouts via `skopeo copy oci:...`,
+   `buildah from oci:...`, or similar OCI-aware tooling.
+
+### How OCI registries work
+
+The OCI Distribution API exposes two key endpoints:
+
+- `GET /v2/<name>/manifests/<reference>`: fetches a manifest (image configuration, list of layers,
+   or multi-arch index).
+- `GET /v2/<name>/blobs/<digest>`: fetches a content-addressable blob (layer tarball, config JSON).
+
+All content is addressed by cryptographic digest (`sha256:<hex>`). A manifest lists the digests of
+its blobs. Verifying the manifest digest confirms the manifest is authentic, and the blob digests
+listed inside it become the expected checksums. Each blob must still be independently verified
+against its descriptor digest after download (see [Checksum generation](#checksum-generation)).
+
+Registries typically require authentication via the [Docker Token Authentication](https://distribution.github.io/distribution/spec/auth/token/)
+flow: the client receives a `401` with a `WWW-Authenticate` challenge, exchanges credentials at a
+token endpoint, and uses the resulting bearer token for subsequent requests.
+
+## Design
+
+### Scope
+
+**In scope:**
+
+- Fetching OCI images (single-arch and multi-arch) and arbitrary OCI artifacts from any
+  OCI-compliant registry
+- OCI Image Manifests (`application/vnd.oci.image.manifest.v1+json`)
+- Docker V2 Manifests (`application/vnd.docker.distribution.manifest.v2+json`)
+- OCI Image Indexes / Docker Manifest Lists (selecting a single platform)
+- Arbitrary media types (the client accepts whatever the registry returns)
+- Docker Token Authentication (anonymous and authenticated)
+- HTTP Basic Authentication for registries that do not use token auth
+- Output as OCI Image Layout directories
+
+### Dependency list generation
+
+#### Dependency list toolchain
+
+Hermeto does not resolve image references to digests. The developer must provide a lockfile with
+fully resolved digests. How to obtain digests is documented in user-facing docs, not here.
+
+#### Dependency list format
+
+The lockfile is YAML, named `oci-images.lock.yaml`.
+
+> **Note**: The team is discussing whether non-packaging backends (generic, OCI, etc.) should
+> share a single lockfile (`artifacts.lock.yaml`) with a type discriminator. This design proposes
+> a separate file for now; the lockfile format may change based on that discussion.
+
+```yaml
+metadata:
+  version: "1.0"
+images:
+  - repository: docker.io/library/alpine
+    digest: sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
+    tag: "3.21"
+    media_type: application/vnd.oci.image.index.v1+json
+    platform:
+      os: linux
+      architecture: arm64
+      variant: v8
+    auth:
+      basic:
+        username: "$REGISTRY_USER"
+        password: "$REGISTRY_PASS"
+
+  - repository: ghcr.io/org/my-artifact
+    digest: sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef
+    platform:
+      os: linux
+      architecture: amd64
+```
+
+**Field reference:**
+
+| Field | Required | Description |
+|---|---|---|
+| `repository` | yes | Full registry/repository path (e.g. `docker.io/library/alpine`) |
+| `digest` | yes | Content-addressable SHA-256 digest (`sha256:<hex>`). Immutable identifier used for fetching. Only SHA-256 is supported. |
+| `media_type` | no | Media type hint for content negotiation. When provided, used as the `Accept` header. When omitted, the client sends all supported manifest types (`application/vnd.oci.image.manifest.v1+json`, `application/vnd.oci.image.index.v1+json`, `application/vnd.docker.distribution.manifest.v2+json`, `application/vnd.docker.distribution.manifest.list.v2+json`) and determines the actual type from the response `Content-Type`. |
+| `platform` | no | Platform selector for multi-arch image indexes. Required when `digest` points to an image index. |
+| `platform.os` | no | Operating system (default: `linux`) |
+| `platform.architecture` | yes | CPU architecture (e.g. `amd64`, `arm64`) |
+| `platform.variant` | no | CPU variant (e.g. `v8` for arm64). Used to disambiguate image index entries that share the same os/architecture but differ by variant. |
+| `tag` | no | Human-readable tag. Informational only, used in SBOM output. Never used for fetching. |
+| `auth` | no | Authentication credentials. Mutually exclusive `basic` or `bearer` blocks. `basic` credentials are sent on the token exchange request (step 1 of the fetch procedure). `bearer` provides a pre-obtained token used directly as `Authorization: Bearer <token>`, skipping the token exchange. Supports environment variable expansion (e.g. `$REGISTRY_PASS`). Reuses the existing `LockfileArtifactAuth` model. |
+
+#### Checksum generation
+
+OCI images are content-addressable by design. The `digest` field in the lockfile serves as both
+the identifier and the checksum. Hermeto verifies:
+
+1. The manifest bytes hash to the lockfile `digest`.
+2. Each blob (config + layers) hashes to the digest declared in the manifest.
+
+This provides a complete chain of trust: the lockfile pins the manifest, and the manifest pins every
+blob. No separate checksum generation step is needed.
+
+### Fetching content
+
+#### Fetch implementation
+
+Hermeto fetches directly via the OCI Distribution HTTP API. No external tools (skopeo, crane,
+podman) are invoked. The existing `aiohttp` infrastructure in
+`hermeto/core/package_managers/general.py` (`async_download_files`) is reused for blob downloads.
+
+#### Fetch procedure
+
+For each image entry in the lockfile:
+
+1. **Authenticate**: Send `GET /v2/` to the registry. If a `401` response includes a
+   `WWW-Authenticate: Bearer realm="...",service="...",scope="..."` header, exchange credentials
+   at the realm URL. If the lockfile provides `basic` auth, include it on the token request. If
+   `bearer` auth is configured, use the provided token directly, skipping the exchange. If no
+   auth is configured, request an anonymous token (works for public registries). If the registry
+   returns `200` with no challenge (e.g. a private registry behind a VPN), skip token exchange
+   entirely and proceed without an `Authorization` header.
+
+2. **Fetch manifest**: `GET /v2/<name>/manifests/<digest>` with the bearer token. If the lockfile
+   provides `media_type`, use it as the `Accept` header; otherwise send all supported manifest
+   types. Compute `sha256` of the response body and verify it matches `digest` from the lockfile.
+   Determine the manifest type from the response `Content-Type` header. Parse the manifest JSON.
+
+3. **Handle image indexes**: If the manifest is an Image Index or Docker Manifest List, find the
+   entry matching the specified `platform`. If no `platform` was specified, raise an error asking
+   the user to set it. Fetch the platform-specific manifest by its digest, using the child
+   descriptor's `mediaType` from the index entry as the `Accept` header (not the parent index type).
+
+4. **Download blobs**: Extract the config descriptor and layer descriptors from the manifest.
+   Deduplicate by digest (two descriptors can reference the same blob). For each unique descriptor,
+   build the download URL (`/v2/<name>/blobs/<digest>`) and the target path
+   (`blobs/<algorithm>/<hex>`). Download to a temporary file, verify its hash against the
+   descriptor digest (`sha256`), and
+   atomically rename to the final path. Downloads run concurrently, limited by a single semaphore
+   that caps the total number of in-flight blob downloads across all images being fetched. The
+   limit is configurable via Hermeto's `concurrency_limit` runtime setting.
+
+5. **Write OCI Image Layout**: Write the manifest bytes to `blobs/<algorithm>/<hex>` (the same
+   digest verified in step 2). Create `index.json` referencing this manifest blob. Write the
+   `oci-layout` file (`{"imageLayoutVersion": "1.0.0"}`). The config and layer blobs are already
+   in place from step 4.
+
+#### Project structure (output)
+
+For each image, Hermeto produces an [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)
+directory:
+
+```
+deps/x-oci/<sanitized-repository>/
+  oci-layout                    # {"imageLayoutVersion": "1.0.0"}
+  index.json                    # references the fetched manifest
+  blobs/
+    sha256/
+      <manifest-digest-hex>     # image manifest JSON
+      <config-digest-hex>       # image config JSON
+      <layer1-digest-hex>       # layer tarball
+      <layer2-digest-hex>       # layer tarball
+      ...
+```
+
+The directory name is derived from the repository path by replacing `/` with `_` and appending a
+12-char digest hex prefix (e.g. `docker.io_library_alpine_a8560b36e8b8`). Since this
+replacement is not injective (e.g. `a/b_c` and `a_b/c` produce the same string), the lockfile
+parser validates that all entries produce unique directory names and rejects the lockfile with a
+clear error if a collision is detected.
+
+#### Network requirements
+
+- **Registry endpoints**: Any OCI-compliant registry (Docker Hub, Quay, GHCR, etc.)
+- **Authentication**: Docker Token Authentication (automatic challenge-response). Credentials come
+  from the lockfile's `auth` block with environment variable expansion.
+- **Redirects**: Some registries (notably Docker Hub) redirect blob downloads to CDN URLs. The
+  client follows redirects with `requote_redirect_url=False` to preserve signed URLs. Per the
+  OCI Distribution Spec, `Authorization` headers are stripped on cross-host redirects to prevent
+  credential leakage to CDN endpoints.
+- **Rate limiting**: Docker Hub imposes pull rate limits. The retry infrastructure handles 429
+  responses.
+
+### SBOM representation
+
+Each fetched image is reported as an SBOM component using the
+[`pkg:oci` PURL type](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci).
+
+**PURL mapping:**
+
+| PURL field | Source | Example |
+|---|---|---|
+| `type` | fixed | `oci` |
+| `name` | last segment of `repo_path` | `alpine` |
+| `version` | digest of the fetched manifest (child digest for multi-arch, lockfile digest for single-arch) | `sha256:a8560b36e8b8` |
+| `qualifiers.repository_url` | `{registry}/{repo_path}` | `docker.io/library/alpine` |
+| `qualifiers.arch` | platform architecture (if platform was specified) | `amd64` |
+| `qualifiers.tag` | lockfile `tag` (if present) | `3.21` |
+
+The `pkg:oci` type spec forbids a `namespace` component. The full repository path (including
+registry) goes into the `repository_url` qualifier. For multi-arch images, the `version` field
+uses the selected child manifest digest (not the parent index digest), so that different platform
+fetches produce distinct SBOM components.
+
+Example PURL (percent-encoded per the PURL spec):
+```
+pkg:oci/alpine@sha256%3Aa8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c?repository_url=docker.io%2Flibrary%2Falpine&arch=amd64&tag=3.21
+```
+
+The SBOM is emitted as part of Hermeto's standard request output. Because `x-oci` is an
+experimental backend, all components are annotated with `hermeto:backend:experimental:x-oci`.
+
+## References
+
+- [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md)
+- [OCI Image Specification](https://github.com/opencontainers/image-spec/blob/main/spec.md)
+- [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)
+- [Docker Token Authentication](https://distribution.github.io/distribution/spec/auth/token/)
+- [Package URL `pkg:oci` type](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci)
+- [Hermeto issue #1641](https://github.com/hermetoproject/hermeto/issues/1641)
+- [Hermeto issue #1080 (related, OCI URLs for RPMs)](https://github.com/hermetoproject/hermeto/issues/1080)
--- a/docs/design/oci.md
+++ b/docs/design/oci.md
@@ -1,0 +1,247 @@
+# OCI artifacts design document
+
+## Overview
+
+The [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md)
+defines an HTTP API for pushing and pulling content from registries. The same API serves container
+images and [arbitrary artifacts](https://github.com/opencontainers/image-spec/blob/main/manifest.md)
+(signatures, SBOMs, Helm charts, Wasm modules, ML models, etc.). This backend allows Hermeto to
+prefetch any of these during the dependency prefetch step, so that hermetic builds can consume them
+without network access.
+
+**Motivating use case**: build pipelines that need OCI artifacts at build time. A build might need
+a Helm chart to template, a Wasm plugin to embed, a base image to layer on top of, or any other
+content stored in an OCI registry. In a hermetic build environment, none of these are reachable
+over the network, so they must be prefetched and made available on disk.
+
+### Developer workflow
+
+1. The developer identifies the OCI images or artifacts their build needs.
+2. The developer resolves each reference to a digest using standard tooling:
+   ```sh
+   skopeo inspect --raw docker://registry.example.com/my-image:1.0 | sha256sum
+   # or
+   crane digest registry.example.com/my-image:1.0
+   ```
+3. The developer declares each dependency in `oci-images.lock.yaml` with its digest
+   and repository.
+4. Hermeto prefetches the declared images, producing an [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)
+   directory for each.
+5. The build environment consumes the layouts via `skopeo copy oci:...`,
+   `buildah from oci:...`, or similar OCI-aware tooling.
+
+### How OCI registries work
+
+The OCI Distribution API exposes two key endpoints:
+
+- `GET /v2/<name>/manifests/<reference>`: fetches a manifest (image configuration, list of layers,
+   or multi-arch index).
+- `GET /v2/<name>/blobs/<digest>`: fetches a content-addressable blob (layer tarball, config JSON).
+
+All content is addressed by cryptographic digest (`sha256:<hex>`). A manifest lists the digests of
+its blobs. Verifying the manifest digest confirms the manifest is authentic, and the blob digests
+listed inside it become the expected checksums. Each blob must still be independently verified
+against its descriptor digest after download (see [Checksum generation](#checksum-generation)).
+
+Registries typically require authentication via the [Docker Token Authentication](https://distribution.github.io/distribution/spec/auth/token/)
+flow: the client receives a `401` with a `WWW-Authenticate` challenge, exchanges credentials at a
+token endpoint, and uses the resulting bearer token for subsequent requests.
+
+## Design
+
+### Scope
+
+**In scope:**
+
+- Fetching OCI images (single-arch and multi-arch) and arbitrary OCI artifacts from any
+  OCI-compliant registry
+- OCI Image Manifests (`application/vnd.oci.image.manifest.v1+json`)
+- Docker V2 Manifests (`application/vnd.docker.distribution.manifest.v2+json`)
+- OCI Image Indexes / Docker Manifest Lists (selecting a single platform)
+- Arbitrary media types (the client accepts whatever the registry returns)
+- Docker Token Authentication (anonymous and authenticated)
+- HTTP Basic Authentication for registries that do not use token auth
+- Output as OCI Image Layout directories
+
+### Dependency list generation
+
+#### Dependency list toolchain
+
+Hermeto does not resolve image references to digests. The developer must provide a lockfile with
+fully resolved digests. How to obtain digests is documented in user-facing docs, not here.
+
+#### Dependency list format
+
+The lockfile is YAML, named `oci-images.lock.yaml`.
+
+> **Note**: The team is discussing whether non-packaging backends (generic, OCI, etc.) should
+> share a single lockfile (`artifacts.lock.yaml`) with a type discriminator. This design proposes
+> a separate file for now; the lockfile format may change based on that discussion.
+
+```yaml
+metadata:
+  version: "1.0"
+images:
+  - repository: docker.io/library/alpine
+    digest: sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
+    tag: "3.21"
+    media_type: application/vnd.oci.image.index.v1+json
+    platform:
+      os: linux
+      architecture: amd64
+      variant: v8
+    auth:
+      basic:
+        username: "$REGISTRY_USER"
+        password: "$REGISTRY_PASS"
+
+  - repository: ghcr.io/org/my-artifact
+    digest: sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef
+    platform:
+      os: linux
+      architecture: amd64
+```
+
+**Field reference:**
+
+| Field | Required | Description |
+|---|---|---|
+| `repository` | yes | Full registry/repository path (e.g. `docker.io/library/alpine`) |
+| `digest` | yes | Content-addressable SHA-256 digest (`sha256:<hex>`). Immutable identifier used for fetching. Only SHA-256 is supported. |
+| `media_type` | no | Media type hint for content negotiation. When provided, used as the `Accept` header. When omitted, the client sends all supported manifest types (`application/vnd.oci.image.manifest.v1+json`, `application/vnd.oci.image.index.v1+json`, `application/vnd.docker.distribution.manifest.v2+json`, `application/vnd.docker.distribution.manifest.list.v2+json`) and determines the actual type from the response `Content-Type`. |
+| `platform` | no | Platform selector for multi-arch image indexes. Required when `digest` points to an image index. |
+| `platform.os` | no | Operating system (default: `linux`) |
+| `platform.architecture` | yes | CPU architecture (e.g. `amd64`, `arm64`) |
+| `platform.variant` | no | CPU variant (e.g. `v8` for arm64). Used to disambiguate image index entries that share the same os/architecture but differ by variant. |
+| `tag` | no | Human-readable tag. Informational only, used in SBOM output. Never used for fetching. |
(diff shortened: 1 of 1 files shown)


docs/design/oci.md
```diff
@@ -0,0 +1,244 @@
...
+### How OCI registries work
+
+The OCI Distribution API exposes two key endpoints:
+
+- `GET /v2/<name>/manifests/<reference>`: fetches a manifest (image configuration, list of layers,
+   or multi-arch index).
+- `GET /v2/<name>/blobs/<digest>`: fetches a content-addressable blob (layer tarball, config JSON).
+
+All content is addressed by cryptographic digest (`sha256:<hex>`). A manifest lists the digests of
+its blobs, so verifying the manifest digest implicitly verifies the integrity of the entire image.
```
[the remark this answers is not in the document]
jangel97: Yeah, the wording was wrong. Rewrote to separate the two steps: manifest digest verifies the manifest, blobs are verified individually against their descriptor digests after download. Added a cross-reference to the Checksum generation section.

docs/design/oci.md
```diff
@@ -0,0 +1,244 @@
...
+- OCI Image Manifests (`application/vnd.oci.image.manifest.v1+json`)
+- Docker V2 Manifests (`application/vnd.docker.distribution.manifest.v2+json`)
+- OCI Image Indexes / Docker Manifest Lists (selecting a single platform)
+- Docker Token Authentication (anonymous and authenticated)
+- Output as OCI Image Layout directories
+
+**Out of scope (may be addressed in future work):**
+
+- Basic auth for registries that do not use token auth
+- OCI artifacts with arbitrary/custom media types (ORAS-style)
```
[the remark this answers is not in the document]
jangel97: Fair point. Expanded the out-of-scope bullet to clarify that "Basic auth" there means registries that *only* do HTTP Basic (no token exchange). The `basic` credentials used within the Docker Token Auth flow are in scope. Also documented the `auth.bearer` path in the field reference table and the fetch procedure.

docs/design/oci.md
```diff
@@ -0,0 +1,244 @@
...
+| Field | Required | Description |
+|---|---|---|
+| `repository` | yes | Full registry/repository path (e.g. `docker.io/library/alpine`) |
+| `digest` | yes | Content-addressable digest (e.g. `sha256:abc...`). Immutable identifier used for fetching. |
+| `media_type` | no | Media type of the manifest at the given digest. Defaults to `application/vnd.oci.image.manifest.v1+json`. |
+| `platform` | no | Platform selector for multi-arch image indexes. Required when `digest` points to an image index. |
+| `platform.os` | no | Operating system (default: `linux`) |
+| `platform.architecture` | yes | CPU architecture (e.g. `amd64`, `arm64`) |
+| `tag` | no | Human-readable tag. Informational only, used in SBOM output. Never used for fetching. |
+| `auth` | no | Authentication credentials. Mutually exclusive `basic` or `bearer` blocks. Supports environment variable expansion (e.g. `$REGISTRY_PASS`). Same model as the generic fetcher's `LockfileArtifactAuth`. |
```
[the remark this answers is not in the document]
jangel97: Fixed. Step 3 now says the child manifest is fetched using the child descriptor's `mediaType` from the index entry, not the parent index type.

docs/design/oci.md
```diff
@@ -0,0 +1,244 @@
...
+   lockfile. Parse the manifest JSON.
+
+3. **Handle image indexes**: If the manifest is an Image Index or Docker Manifest List, find the
+   entry matching the specified `platform`. If no `platform` was specified, raise an error asking
+   the user to set it. Fetch the platform-specific manifest by its digest (repeating step 2).
+
+4. **Download blobs**: Extract the config descriptor and layer descriptors from the manifest. For
+   each descriptor, build the download URL (`/v2/<name>/blobs/<digest>`) and the target path
+   (`blobs/sha256/<hex>`). Download concurrently. After download, verify each file's sha256
+   against its descriptor digest.
```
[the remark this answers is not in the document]
jangel97: Right, two descriptors can share a digest. Step 4 now deduplicates by digest before fetching, and describes the flow as temp file, verify hash, atomic rename.

docs/design/oci.md
```diff
@@ -0,0 +1,244 @@
...
+      <config-digest-hex>       # image config JSON
+      <layer1-digest-hex>       # layer tarball
+      <layer2-digest-hex>       # layer tarball
+      ...
+```
+
+The directory name is derived from the repository (replacing `/` and `:` with `_`) plus a 12-char
+digest prefix to avoid collisions when the same repository appears at different digests
+(e.g. `registry.redhat.io/rhel9/rhel-bootc@sha256:7c4e86...` becomes
+`registry.redhat.io_rhel9_rhel-bootc_7c4e86de0c80`).
```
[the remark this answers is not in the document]
jangel97: True, the underscore replacement isn't injective. Instead of switching to hashed names (which hurts readability), the lockfile parser now validates that all entries produce unique directory names and rejects with a clear error on collision. In practice this is extremely unlikely given the 12-char digest hex suffix, but worth catching.

docs/design/oci.md
```diff
@@ -0,0 +1,244 @@
...
+
+#### Network requirements
+
+- **Registry endpoints**: Any OCI-compliant registry (Docker Hub, Quay, GHCR, registry.redhat.io, etc.)
+- **Authentication**: Docker Token Authentication (automatic challenge-response). Credentials come
+  from the lockfile's `auth` block with environment variable expansion.
+- **Redirects**: Some registries (notably Docker Hub) redirect blob downloads to CDN URLs. The
+  `aiohttp` client handles redirects with `requote_redirect_url=False` to preserve signed URLs.
+- **Rate limiting**: Docker Hub imposes pull rate limits. The retry infrastructure handles 429
+  responses.
```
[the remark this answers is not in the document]
jangel97: Added that `Authorization` headers are stripped on cross-host redirects per the OCI Distribution Spec, so tokens don't leak to CDN endpoints.

docs/design/oci.md
```diff
@@ -0,0 +1,244 @@
...
+| `HERMETO_OCI_LAYOUT_DIR` | Root directory containing all OCI Image Layout directories | `deps/x-oci` | Yes |
+
+#### Configuration files
+
+No additional configuration files are injected. The OCI Image Layout is self-contained and can be
+consumed directly by container tools:
+
+```sh
+# Load into podman's local storage
+podman load --oci-dir deps/x-oci/registry.redhat.io_rhel9_rhel-bootc
```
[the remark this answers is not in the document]
jangel97: Right, `podman load` only takes archives. Replaced with `skopeo copy oci:... containers-storage:...`.

docs/design/oci.md
```diff
@@ -0,0 +1,244 @@
...
+- **Single lockfile per package input**: Each `x-oci` package input processes one lockfile. Multiple
+  lockfiles require multiple package input entries.
+
+## References
+
+- [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md)
+- [OCI Image Specification](https://github.com/opencontainers/image-spec/blob/main/spec.md)
+- [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)
+- [Docker Token Authentication](https://distribution.github.io/distribution/spec/auth/token/)
+- [Package URL `pkg:oci` type](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci)
```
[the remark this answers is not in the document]
jangel97: Added an "SBOM representation" section with a field mapping table, example PURL, and a note on the experimental backend annotation. The `pkg:oci` type forbids a namespace component, so the full repository path goes into the `repository_url` qualifier.

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+**Out of scope (may be addressed in future work):**
+
+- Registries that use only HTTP Basic authentication (no token exchange). Note: Basic credentials
+  *used within* the Docker Token Authentication flow (to authenticate the token request) are
+  supported via the lockfile `auth.basic` block.
+- OCI artifacts with arbitrary/custom media types (ORAS-style)
+- Signature verification (cosign, Notary)
+- Image attestation fetching
+- Pushing images or artifacts to registries
+- Building or modifying images
```
tnevrlka: Why would Hermeto ever "build or modify" images? There are other tools for that

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+
+**Out of scope (may be addressed in future work):**
+
+- Registries that use only HTTP Basic authentication (no token exchange). Note: Basic credentials
+  *used within* the Docker Token Authentication flow (to authenticate the token request) are
+  supported via the lockfile `auth.basic` block.
+- OCI artifacts with arbitrary/custom media types (ORAS-style)
+- Signature verification (cosign, Notary)
+- Image attestation fetching
+- Pushing images or artifacts to registries
```
tnevrlka: I can't imagine the workflow and usecase. 

Hermeto's purpose is basically to prefetch dependencies for hermetic builds. What and why would ever get pushed to a registry?

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+- Docker Token Authentication (anonymous and authenticated)
+- Output as OCI Image Layout directories
+
+**Out of scope (may be addressed in future work):**
+
+- Registries that use only HTTP Basic authentication (no token exchange). Note: Basic credentials
+  *used within* the Docker Token Authentication flow (to authenticate the token request) are
+  supported via the lockfile `auth.basic` block.
+- OCI artifacts with arbitrary/custom media types (ORAS-style)
+- Signature verification (cosign, Notary)
```
tnevrlka: I can't imagine that ever being supported, for that purpose there are the tools that are listed

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+- Image attestation fetching
```
tnevrlka: Image attestation is technically a "OCI artifact with arbitrary/custom media types", isn't it?

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+- OCI Image Indexes / Docker Manifest Lists (selecting a single platform)
+- Docker Token Authentication (anonymous and authenticated)
+- Output as OCI Image Layout directories
+
+**Out of scope (may be addressed in future work):**
+
+- Registries that use only HTTP Basic authentication (no token exchange). Note: Basic credentials
+  *used within* the Docker Token Authentication flow (to authenticate the token request) are
+  supported via the lockfile `auth.basic` block.
+- OCI artifacts with arbitrary/custom media types (ORAS-style)
```
tnevrlka: Is there any difference in fetching OCI artifacts due to their type? It's even mentioned at the beginning of the doc
> Container images are the primary content type, but the same API is used to store
[arbitrary artifacts](https://github.com/opencontainers/image-spec/blob/main/manifest.md) (signatures, SBOMs, Helm charts, Wasm modules, etc.).

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+- Fetching OCI images (single-arch and multi-arch) from any OCI-compliant registry
+- OCI Image Manifests (`application/vnd.oci.image.manifest.v1+json`)
+- Docker V2 Manifests (`application/vnd.docker.distribution.manifest.v2+json`)
+- OCI Image Indexes / Docker Manifest Lists (selecting a single platform)
+- Docker Token Authentication (anonymous and authenticated)
+- Output as OCI Image Layout directories
+
+**Out of scope (may be addressed in future work):**
+
+- Registries that use only HTTP Basic authentication (no token exchange). Note: Basic credentials
```
tnevrlka: Is it worth putting out of scope? I would imagine it'd be easy to implement

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+
+The [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md)
+defines an HTTP API for pushing and pulling container images and other content from registries.
+Container images are the primary content type, but the same API is used to store
+[arbitrary artifacts](https://github.com/opencontainers/image-spec/blob/main/manifest.md)
+(signatures, SBOMs, Helm charts, Wasm modules, etc.).
+
+This backend allows Hermeto to prefetch OCI container images from registries during the dependency
+prefetch step, so that hermetic builds can consume them without network access. Support for arbitrary
+OCI artifacts (ORAS-style custom media types) may be added in future work.
```
tnevrlka: > Support for arbitrary OCI artifacts (ORAS-style custom media types) may be added in future work.

Why not now if it's the same API?

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+
+**Motivating use case**: build pipelines that consume base images or pre-built container images from
+registries (e.g. Red Hat UBI images, NVIDIA CUDA images).
```
tnevrlka: I'm not sure the examples are needed, Hermeto is a community project that does not have any preference for Red Hat or Nvidia

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+## Implementation notes
+
+### Current limitations
+
+- **No signature verification**: cosign and Notary signatures are not fetched or verified.
+- **No attestation support**: SLSA provenance and other attestation artifacts are not fetched.
+- **No custom media types**: Only standard OCI and Docker V2 image manifests are supported.
+  ORAS-style artifacts with arbitrary media types are not yet handled.
+- **Single lockfile per package input**: Each `x-oci` package input processes one lockfile. Multiple
+  lockfiles require multiple package input entries.
```
tnevrlka: What's the difference between current limitations and the out of scope section?

tnevrlka: Why is this a limitation?

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+
+```sh
+# Copy into podman's local storage
+skopeo copy oci:deps/x-oci/registry.redhat.io_rhel9_rhel-bootc containers-storage:my-image:latest
+
+# Copy to a Docker daemon
+skopeo copy oci:deps/x-oci/registry.redhat.io_rhel9_rhel-bootc docker-daemon:my-image:latest
+
+# Use as a buildah source
+buildah from oci:deps/x-oci/registry.redhat.io_rhel9_rhel-bootc
```
tnevrlka: It'd be nice to avoid mentioning a specific company

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+### Dependency list generation
+
+#### Dependency list toolchain
+
+Hermeto does not resolve image references to digests. The developer must provide a lockfile with
+fully resolved digests. Standard tools to obtain digests:
+
+- `skopeo inspect --raw docker://<image>:<tag>` (pipe through `sha256sum`)
+- `crane digest <image>:<tag>`
+- `podman inspect --format '{{.Digest}}' <image>:<tag>`
```
tnevrlka: I'm not sure that `podman inspect` is the right tool for this since it as far as I know doesn't work with remote registries

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+
+#### Native vs. Hermeto fetch
+
+Hermeto handles all fetching directly via the OCI Distribution HTTP API. No external tools
+(skopeo, crane, podman) are invoked, consistent with the principle of avoiding arbitrary code
+execution.
+
+The OCI Distribution API is HTTP-based, so the existing `aiohttp` infrastructure in
+`hermeto/core/package_managers/general.py` (`async_download_files`) can be reused for blob
+downloads.
```
tnevrlka: I'm not sure I see the comparison between "native" and "Hermeto" fetch

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+   entry matching the specified `platform`. If no `platform` was specified, raise an error asking
+   the user to set it. Fetch the platform-specific manifest by its digest, using the child
+   descriptor's `mediaType` from the index entry as the `Accept` header (not the parent index type).
+
+4. **Download blobs**: Extract the config descriptor and layer descriptors from the manifest.
+   Deduplicate by digest (two descriptors can reference the same blob). For each unique descriptor,
+   build the download URL (`/v2/<name>/blobs/<digest>`) and the target path
+   (`blobs/<algorithm>/<hex>`). Download to a temporary file, verify its hash against the
+   descriptor digest using the algorithm from the digest string (e.g. `sha256`, `sha512`), and
+   atomically rename to the final path. Downloads run concurrently with a shared concurrency limit.
```
tnevrlka: What does the "shared concurrency limit" mean in this context?

tnevrlka (commented):
First wave of comments

docs/design/oci.md
eskultety: Note I'm not even sure it's strictly necessary for a design doc since resolution will happen outside Hermeto. The better place to document how to obtain a fully resolved file (since there's no standardized ecosystem way of doing it) would be user-facing docs. As far as hermeto impl. is concerned IMO, we can simply work with the premise we have all the information we require passed via configuration.

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+        username: "$REGISTRY_USER"
+        password: "$REGISTRY_PASS"
+
+  - repository: ghcr.io/org/my-artifact
+    digest: sha256:deadbeef...
+    media_type: application/vnd.oci.image.index.v1+json
+    platform:
+      os: linux
+      architecture: amd64
+```
```
eskultety, line 103: This is effectively a 3rd custom input file format we'd be maintaining going forward. The same initial expectation was for pulling hugging face models (that would be 4th file format). Adding more custom files for users to add to their repos only adds to the frustration (yes, we've received such a ticket not so long ago). Given the context we have on these non-standardized (in terms of no standardized ecosystem tooling) fetch backends I've been wondering for a while if we shouldn't rethink our approach towards `artifacts.lock.yaml` not serving just the generic backend. I think going forward we should sit down and come up with a robust solution where non-packaging backends could be described by a single file and then having a dispatching module calling into specialized backends.
I'm very open to the discussion on "N input files vs a single input file" to hear good arguments in favour of the former since at the moment I'm quite dismissive of the possibility of us just adding a new file format to the pool, potentially re-architecting this in the future, forcing our users to migrate to a different setup using a new major release of Hermeto.

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+
+#### Fetch procedure
+
+For each image entry in the lockfile:
+
+1. **Authenticate**: Send `GET /v2/` to the registry. If a `401` response includes a
+   `WWW-Authenticate: Bearer realm="...",service="...",scope="..."` header, exchange credentials
+   at the realm URL. If the lockfile provides `basic` auth, include it on the token request. If
+   `bearer` auth is configured, use the provided token directly, skipping the exchange. If no
+   auth is configured, request an anonymous token (works for public registries).
```
eskultety: can 'no auth configured' also mean unauthenticated tokenless access for any registry or token auth (be it anonymous) is mandatory with all registries?

docs/design/oci.md
eskultety: Probably that our concurrency config option is respected.

docs/design/oci.md
```diff
@@ -0,0 +1,281 @@
...
+  responses.
+
+### Build environment config
+
+#### Environment variables
+
+| Variable Name | Purpose | Example Value | Required |
+|---|---|---|---|
+| `HERMETO_OCI_LAYOUT_DIR` | Root directory containing all OCI Image Layout directories | `deps/x-oci` | Yes |
+
```
eskultety: Who is this variable for? Hermeto users or build environments? I presume the intent is for the former?

docs/design/oci.md
eskultety: Yes, if we're to add an OCI backend, then the functionality should cover many/most use cases out of the box, not limiting itself to one particular subset (e.g. just images).

docs/design/oci.md
eskultety: I understand the motivation, but I don't think this one provides the right justification for the work. IOW to me the main motivating driver for anything OCI related is the need for arbitrary artifacts to be present at build time, not consuming pre-cached container images. On the topic of consuming base images stored on disk - could you do e.g. `podman save` to your controlled cache and then use it in your pipeline? I suppose not, but these are the type of questions to ask when adding support for a new backend, i.e. "Does it solve just my tailored use case or does the wider user audience benefit?", "Do other platform solutions suffer from the same problem as I have or is there something other platforms allow?"

docs/design/oci.md
jangel97: Removed. Hermeto is a prefetcher, that doesn't belong here.

docs/design/oci.md
jangel97: Same, removed.

docs/design/oci.md
jangel97: Removed, that's cosign/Notary territory.

docs/design/oci.md
jangel97: Agreed, attestations are just OCI artifacts with specific media types. Removed as a separate item.

docs/design/oci.md
jangel97: Makes sense, the API is the same regardless of media type. Moved arbitrary OCI artifacts into scope. The client will accept whatever the registry returns.

docs/design/oci.md
jangel97: Added to in-scope. Straightforward to support alongside token auth.

docs/design/oci.md
jangel97: Dropped the "future work" hedging. The overview now covers both container images and arbitrary artifacts as one thing.

docs/design/oci.md
jangel97: Swapped all examples to docker.io/library/alpine. The motivating use case description is generic now too.

docs/design/oci.md
jangel97: Removed the whole section. Everything there either moved into scope or doesn't apply to Hermeto. The out-of-scope section is gone too for the same reason.

jangel97: Same fix as above, the whole section is gone.

docs/design/oci.md
jangel97: Fixed, all examples use docker.io/library/alpine.

docs/design/oci.md
jangel97: Removed. Only skopeo and crane left, both work against remote registries.

docs/design/oci.md
jangel97: Renamed to "Fetch implementation" and dropped the comparison framing. Just says Hermeto uses the OCI Distribution HTTP API directly, no external tools.

docs/design/oci.md
jangel97: Clarified: a single semaphore caps total in-flight blob downloads across all images. Configurable via Hermeto's concurrency_limit runtime setting.

docs/design/oci.md
jangel97: This is a fair concern and I think it deserves a proper discussion. I see the appeal of a single artifacts.lock.yaml with a type discriminator that dispatches to specialized backends, rather than each backend inventing its own file. Happy to go either direction here. If the consensus is to extend artifacts.lock.yaml, I can rework the design around that. Worth noting that the OCI entries need fields the generic backend doesn't have (platform selector, media type, multi-blob output), so we'd need to figure out how the schema accommodates that without making the generic case more complicated.

docs/design/oci.md
jangel97: Yes, some registries (typically private ones behind a VPN or firewall) serve content with no auth challenge at all. Updated the auth procedure: if GET /v2/ returns 200 with no WWW-Authenticate header, the client skips token exchange and proceeds without an Authorization header.

docs/design/oci.md
jangel97: Clarified in the doc. The variable is for build environments (Containerfile, build scripts) to locate the prefetched OCI layouts. Hermeto determines the output path, the variable just tells downstream consumers where to find it.

docs/design/oci.md
```diff
@@ -0,0 +1,263 @@
...
+   skopeo inspect --raw docker://registry.example.com/my-image:1.0 | sha256sum
+   # or
+   crane digest registry.example.com/my-image:1.0
+   ```
+3. The developer declares each dependency in `oci-images.lock.yaml` with its digest,
+   repository, and media type.
+4. Hermeto prefetches the declared images, producing an [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)
+   directory for each.
+5. The build environment consumes the layouts via `podman load`, `skopeo copy oci:...`, or
+   `buildah from oci:...`.
```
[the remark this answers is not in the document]
jangel97: Already fixed in a previous round. Removed the remaining podman load reference from the developer workflow section too.

docs/design/oci.md
```diff
@@ -0,0 +1,263 @@
...
+      basic:
+        username: "$REGISTRY_USER"
+        password: "$REGISTRY_PASS"
+
+  - repository: ghcr.io/org/my-artifact
+    digest: sha256:deadbeef...
+    media_type: application/vnd.oci.image.index.v1+json
+    platform:
+      os: linux
+      architecture: amd64
```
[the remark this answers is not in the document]
jangel97: Good catch. Added platform.variant to the field reference and the lockfile example. Some arm64 images have multiple index entries that only differ by variant (e.g. v8), so this is needed to disambiguate.

docs/design/oci.md
```diff
@@ -0,0 +1,263 @@
...
+deps/x-oci/<sanitized-repository>/
+  oci-layout                    # {"imageLayoutVersion": "1.0.0"}
+  index.json                    # references the fetched manifest
+  blobs/
+    sha256/
+      <manifest-digest-hex>     # image manifest JSON
+      <config-digest-hex>       # image config JSON
+      <layer1-digest-hex>       # layer tarball
+      <layer2-digest-hex>       # layer tarball
+      ...
```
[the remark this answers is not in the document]
jangel97: Right, that was implicit. Step 5 now explicitly says: write the manifest bytes to blobs/<algorithm>/<hex>, create index.json referencing it, and write the oci-layout file.

docs/design/oci.md
```diff
@@ -0,0 +1,262 @@
...
+**Field reference:**
+
+| Field | Required | Description |
+|---|---|---|
+| `repository` | yes | Full registry/repository path (e.g. `docker.io/library/alpine`) |
+| `digest` | yes | Content-addressable digest (e.g. `sha256:abc...`). Immutable identifier used for fetching. |
+| `media_type` | no | Media type of the manifest at the given digest. Defaults to `application/vnd.oci.image.manifest.v1+json`. |
+| `platform` | no | Platform selector for multi-arch image indexes. Required when `digest` points to an image index. |
+| `platform.os` | no | Operating system (default: `linux`) |
+| `platform.architecture` | yes | CPU architecture (e.g. `amd64`, `arm64`) |
```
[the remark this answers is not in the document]
jangel97: On second thought, media_type is now optional. Since we fetch by digest, the content is already pinned. When provided, it's used as the Accept header (content-negotiation hint). When omitted, the client sends all supported manifest types and determines the actual type from the response Content-Type. No wrong default, no unnecessary burden on the user.

docs/design/oci.md
```diff
@@ -0,0 +1,262 @@
...
+
+**PURL mapping:**
+
+| PURL field | Source | Example |
+|---|---|---|
+| `type` | fixed | `oci` |
+| `name` | last segment of `repo_path` | `alpine` |
+| `version` | lockfile `digest` | `sha256:a8560b36...` |
+| `qualifiers.repository_url` | `{registry}/{repo_path}` | `docker.io/library/alpine` |
+| `qualifiers.tag` | lockfile `tag` (if present) | `3.21` |
```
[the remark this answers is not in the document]
jangel97: Good catch. The PURL version now uses the child manifest digest for multi-arch images (not the parent index digest), and arch is included as a qualifier. This way different platform fetches produce distinct SBOM components.

docs/design/oci.md
```diff
@@ -0,0 +1,262 @@
...
+- Fetching OCI images (single-arch and multi-arch) and arbitrary OCI artifacts from any
+  OCI-compliant registry
+- OCI Image Manifests (`application/vnd.oci.image.manifest.v1+json`)
+- Docker V2 Manifests (`application/vnd.docker.distribution.manifest.v2+json`)
+- OCI Image Indexes / Docker Manifest Lists (selecting a single platform)
+- Arbitrary media types (the client accepts whatever the registry returns)
+- Docker Token Authentication (anonymous and authenticated)
+- HTTP Basic Authentication for registries that do not use token auth
+- Output as OCI Image Layout directories
+
```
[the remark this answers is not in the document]
jangel97: The scope says arbitrary media types for what the client accepts from the registry. The fetch procedure handles the manifest structures we support (image manifests, image indexes, Docker V2). For artifacts that use the standard OCI manifest format with custom layer media types, the same flow works. We don't need to parse layers, just download them by digest.

docs/design/oci.md
```diff
@@ -0,0 +1,262 @@
...
+2. **Fetch manifest**: `GET /v2/<name>/manifests/<digest>` with `Accept: <media_type>` and the
+   bearer token. Compute `sha256` of the response body and verify it matches `digest` from the
+   lockfile. Parse the manifest JSON.
+
+3. **Handle image indexes**: If the manifest is an Image Index or Docker Manifest List, find the
+   entry matching the specified `platform`. If no `platform` was specified, raise an error asking
+   the user to set it. Fetch the platform-specific manifest by its digest, using the child
+   descriptor's `mediaType` from the index entry as the `Accept` header (not the parent index type).
+
+4. **Download blobs**: Extract the config descriptor and layer descriptors from the manifest.
```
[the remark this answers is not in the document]
jangel97: Already covered. Step 3 fetches the child manifest by its digest from the index entry, and step 2 verifies any fetched manifest against its expected digest. The child manifest is verified independently before its descriptors are used.

docs/design/oci.md
```diff
@@ -0,0 +1,272 @@
...
+        username: "$REGISTRY_USER"
+        password: "$REGISTRY_PASS"
+
+  - repository: ghcr.io/org/my-artifact
+    digest: sha256:deadbeef...
+    media_type: application/vnd.oci.image.index.v1+json
+    platform:
+      os: linux
+      architecture: amd64
+```
```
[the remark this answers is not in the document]
jangel97: Already fixed. Both lockfile examples have platform selectors.

docs/design/oci.md
```diff
@@ -0,0 +1,272 @@
...
+`hermeto/core/package_managers/general.py` (`async_download_files`) is reused for blob downloads.
+
+#### Fetch procedure
+
+For each image entry in the lockfile:
+
+1. **Authenticate**: Send `GET /v2/` to the registry. If a `401` response includes a
+   `WWW-Authenticate: Bearer realm="...",service="...",scope="..."` header, exchange credentials
+   at the realm URL. If the lockfile provides `basic` auth, include it on the token request. If
+   `bearer` auth is configured, use the provided token directly, skipping the exchange. If no
```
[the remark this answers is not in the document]
jangel97: Valid concern. The registry controls the realm URL, so a malicious registry could redirect credentials to an attacker-controlled endpoint. Worth documenting as a security consideration before implementation. Not something to fix in this design doc PR.

docs/design/oci.md
```diff
@@ -0,0 +1,272 @@
...
+   auth is configured, request an anonymous token (works for public registries). If the registry
+   returns `200` with no challenge (e.g. a private registry behind a VPN), skip token exchange
+   entirely and proceed without an `Authorization` header.
+
+2. **Fetch manifest**: `GET /v2/<name>/manifests/<digest>` with the bearer token. If the lockfile
+   provides `media_type`, use it as the `Accept` header; otherwise send all supported manifest
+   types. Compute `sha256` of the response body and verify it matches `digest` from the lockfile.
+   Determine the manifest type from the response `Content-Type` header. Parse the manifest JSON.
```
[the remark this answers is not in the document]
jangel97: Agreed. Restricted lockfile digests to SHA-256 only and removed the sha512 mention from the blob verification step. OCI registries use SHA-256 universally, no reason to complicate things.

docs/design/oci.md
```diff
@@ -0,0 +1,272 @@
...
+
+The `pkg:oci` type spec forbids a `namespace` component. The full repository path (including
+registry) goes into the `repository_url` qualifier. For multi-arch images, the `version` field
+uses the selected child manifest digest (not the parent index digest), so that different platform
+fetches produce distinct SBOM components.
+
+Example PURL:
+```
+pkg:oci/alpine@sha256:a8560b36e8b8...?repository_url=docker.io/library/alpine&arch=amd64&tag=3.21
+```
```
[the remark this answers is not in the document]
jangel97: Fixed. The example PURL now uses percent-encoded values per the PURL spec: sha256%3A for the version separator and docker.io%2Flibrary%2Falpine for the repository_url qualifier.

taylormadore:
@coderabbitai pause

docs/design/oci.md
```diff
@@ -0,0 +1,272 @@
...
+The [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec/blob/main/spec.md)
+defines an HTTP API for pushing and pulling container images and other content from registries.
+Container images are the primary content type, but the same API is used to store
+[arbitrary artifacts](https://github.com/opencontainers/image-spec/blob/main/manifest.md)
+(signatures, SBOMs, Helm charts, Wasm modules, etc.).
+
+This backend allows Hermeto to prefetch OCI content from registries during the dependency prefetch
+step, so that hermetic builds can consume them without network access. OCI registries store
+container images, but also arbitrary artifacts (signatures, SBOMs, Helm charts, Wasm modules, ML
+models). The same Distribution API serves all of them, and this backend handles all of them.
```
a-ovchinnikov: This paragraph mostly repeats the previous one.

docs/design/oci.md
```diff
@@ -0,0 +1,272 @@
...
+### Developer workflow
+
+1. The developer identifies the OCI images or artifacts their build needs.
+2. The developer resolves each reference to a digest using standard tooling:
+   ```sh
+   skopeo inspect --raw docker://registry.example.com/my-image:1.0 | sha256sum
+   # or
+   crane digest registry.example.com/my-image:1.0
+   ```
+3. The developer declares each dependency in `oci-images.lock.yaml` with its digest,
```
a-ovchinnikov: This effectively means that a yet another lock file is proposed for the project. WDYT about extending the one we already have?

docs/design/oci.md
a-ovchinnikov: I think we should agree on a way of describing multiple entries in `artifacts.lock.yaml` with the current behavior preserved for backwards compatibility and to not create unnecessary disruption to users' workflows, and the new version providing a way to dispatch sections to a specific backend. Maybe we could add a level in `artifacts` or require a specific field in each artifact to be present, e.g. `back-end: generic` or `back-end: oci`?

docs/design/oci.md
```diff
@@ -0,0 +1,272 @@
...
+      <layer1-digest-hex>       # layer tarball
+      <layer2-digest-hex>       # layer tarball
+      ...
+```
+
+The directory name is derived from the repository path by replacing `/` with `_` and appending a
+12-char digest hex prefix (e.g. `docker.io_library_alpine_a8560b36e8b8`). Since this
+replacement is not injective (e.g. `a/b_c` and `a_b/c` produce the same string), the lockfile
+parser validates that all entries produce unique directory names and rejects the lockfile with a
+clear error if a collision is detected.
```
a-ovchinnikov, line 196: Is there anything a user could do if their lock file were rejected?

docs/design/oci.md
```diff
@@ -0,0 +1,272 @@
...
+| `qualifiers.tag` | lockfile `tag` (if present) | `3.21` |
+
+The `pkg:oci` type spec forbids a `namespace` component. The full repository path (including
+registry) goes into the `repository_url` qualifier. For multi-arch images, the `version` field
+uses the selected child manifest digest (not the parent index digest), so that different platform
+fetches produce distinct SBOM components.
+
+Example PURL (percent-encoded per the PURL spec):
+```
+pkg:oci/alpine@sha256%3Aa8560b36e8b8...?repository_url=docker.io%2Flibrary%2Falpine&arch=amd64&tag=3.21
```
a-ovchinnikov: Does ellipsis belong in PURL?

docs/design/oci.md
eskultety: > Maybe we could ... require a specific field in each artifact to be present, e.g. back-end: generic or back-end: oci?

Yes, that would have been the first random thought from the top of my head on the matter. We already get that information in the input JSON as type so those would strictly have to match (skipping/rejecting the common ecosystem ones):
```
 hermeto fetch-deps '[
     {"type": "gomod"},
     {"type": "pip", "path": "other-path"},
]'
```
so that could be the discriminator (pydantic unions also revolve around discriminators) for the new abstract representation of `artifacts.lock.yaml`.

docs/design/oci.md
jangel97: You're right, merged the two paragraphs into one.

docs/design/oci.md
jangel97: Agreed, see the discussion on the other thread. The consensus is moving toward extending artifacts.lock.yaml with a type discriminator. I'll rework the design around that.

docs/design/oci.md
jangel97: Makes sense. Using the type field as the discriminator (matching the CLI input JSON) keeps things consistent. Reworked the lockfile section: OCI entries now live in artifacts.lock.yaml with `type: oci`. Entries without a type field default to generic, so existing lockfiles keep working. The example shows both generic and OCI entries side by side.

docs/design/oci.md
jangel97: In practice, collisions are near-impossible. They require two repositories whose paths differ only in `/` vs `_` placement AND whose 12-char digest prefixes also match. But if it did happen, the error message would show which entries collide so the user knows what to fix. The only real fix would be to use a longer digest suffix or a hash-based name, but that trades readability for a problem that won't occur in practice.

docs/design/oci.md
jangel97: No, it was just showing the digest is truncated in the example. Replaced with the full digest to avoid confusion.

docs/design/oci.md
eskultety: >  The consensus is moving toward extending artifacts.lock.yaml with a type discriminator. I'll rework the design around that.

@jangel97 I'm going to have to stop you right there. Are you using some kind of AI agent in a loop responding automatically to these threads?
In any case, so far you have received comments from 2 engineers on the "common artifacts file" idea. Please, let us sit down as a team to reach a wider consensus, let us track that work separately from this oci backend work. We'll likely want to spend time and resources on proper modeling and laying down the foundation before we're ready to ingest and accept the work in this PR.

jangel97: @eskultety I'm using Claude Code to help draft responses, but I review everything before submitting. 
I'll revert to the separate `oci-images.lock.yaml` format and wait for the team to reach consensus on the unified lockfile question before making changes there.

For context, this feature would help my team avoid some workarounds we have in place for hermetic builds, which is why I've been eager to move it forward.

Thank you for looking into it!

docs/design/oci.md
eskultety: > The variable is for build environments

Okay, thanks for clarifying. In that case, this would be its first instance of use - we haven't needed one for any generic artifacts so far, IOW that's exactly what downstream consumers (e.g. pipelines) are supposed to take care of and in fact, the prefetch task you're likely using/consuming already handles this for you transparently, so I don't think this is the right place to consider nor add such a variable, it's also quite specific and tailored only to a particular backend, what about other backends then which might be part of the same fetch request?
I think our output dir layout is already deterministic - `<output_dir>/deps/<backend>`, so I don't think one needs to point to a different oci (backend) dir. Note the top level output dir which can be pointed elsewhere with CLI. 

@jangel97 can you share what the exact use case be instead of just pointing hermeto to a different output dir?

Having said all of ^this, I really do see the value and convenience of using env vars in the container world in general. In fact we have overhauled our config such that each existing config option is paired with a corresponding env variable not that long ago and so I think we might want to support env variables for our CLI options that are not part of the hermeto YAML config as well, e.g. `HERMETO_OUTPUT_DIR`, `HERMETO_FOR_OUTPUT_DIR`, etc. which, however, is an unrelated piece of work to this PR.

jangel97: hm so perhaps we can just drop this env var for now, sounds good? @eskultety

eskultety: Yes.

docs/design/oci.md
a-ovchinnikov: > But if it did happen, the error message would show which entries collide so the user knows what to fix

Can a user realistically fix this? To the best of my knowledge most users of Hermeto just consume repositories containing various lock file, they are rarely maintainers and often have no say in what is consumed. Will it be different for OCI backend users?

jangel97: Hm, so what if instead of deriving the directory name by replacing / with _ and appending a 12-char digest hex prefix , and erroring out in case of error, maybe we can eliminate the collision entirely by using percent-encoding (`/` -> `%2F`) 

With _ replacement:     
```                                                                                                                                                                                                          
  a/b_c  ->  a_b_c                                                                                                                                                                                                                  
  a_b/c  ->  a_b_c 
```

Both repos would have same derived directory name.

With percent-encoding (/ -> %2F):
```
  a/b_c  ->  a%2Fb_c                                                                                                                                                                                                                
  a_b/c  ->  a_b%2Fc
```
The repos now have different derived directory name (safe and no need to error out).

this should be safe because `%` is not allowed in repository names as per https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests, so `%2F` can never appear in a container image name. This way we avoid erroring out in case of collision because collision cannot happen

Does it make sense to you @a-ovchinnikov if I update the proposal to use percent-encoding instead?

docs/design/oci.md
```diff
@@ -0,0 +1,247 @@
...
+  version: "1.0"
+images:
+  - repository: docker.io/library/alpine
+    digest: sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c
+    tag: "3.21"
+    media_type: application/vnd.oci.image.index.v1+json
+    platform:
+      os: linux
+      architecture: amd64
+      variant: v8
```
cfergeau, line 92: Nit: might be better to use `arm64` instead of `amd64` as the `amd64/v8` combination is invalid.

jangel97: Thanks for pointing this out, I just edited it!

PR #1679, opened 2026-07-30
refactor(pip): make PipRequirement a subclass of packaging.Requirement

### LINKED ISSUE #486: Make PipRequirement a subclass of packaging.Requirement

filipovskyj02:
PipRequirement now inherits name, extras, specifier, marker and url directly from packaging.Requirement. Backward-compatible properties are kept for package, raw_package, version_specs and environment_marker so callers outside this file need no changes.

Closes https://github.com/hermetoproject/hermeto/issues/486

Assisted-by: Claude

### DIFF
--- a/hermeto/core/package_managers/python/pip/main.py
+++ b/hermeto/core/package_managers/python/pip/main.py
@@ -324,7 +324,7 @@ def _download_vcs_package(
     pip_deps_dir: RootedPath,
 ) -> VCSPackage:
     """Fetch a Python package from VCS (only git is supported)."""
-    git_info = extract_git_info(req.url)
+    git_info = extract_git_info(req.direct_access_url)
 
     download_to = pip_deps_dir.join_within_root(_get_external_requirement_filepath(req))
     download_to.path.parent.mkdir(exist_ok=True, parents=True)
@@ -358,7 +358,7 @@ def _download_url_package(
 
     :param trusted_hosts: if host (or host:port) is trusted, do not verify SSL
     """
-    parsed_url = urlparse.urlparse(req.url)
+    parsed_url = urlparse.urlparse(req.direct_access_url)
 
     download_to = pip_deps_dir.join_within_root(_get_external_requirement_filepath(req))
     download_to.path.parent.mkdir(exist_ok=True, parents=True)
@@ -376,7 +376,7 @@ def _download_url_package(
     else:
         insecure = False
 
-    download_binary_file(req.url, download_to.path, insecure=insecure)
+    download_binary_file(req.direct_access_url, download_to.path, insecure=insecure)
 
     hashes = req.hashes
     missing_req_file_checksum = True
@@ -393,7 +393,7 @@ def _download_url_package(
         requirement_file=str(requirements_file.file_path.subpath_from_root),
         missing_req_file_checksum=missing_req_file_checksum,
         package_type="wheel" if parsed_url.path.endswith(WHEEL_FILE_EXTENSION) else "",
-        original_url=req.url,
+        original_url=req.direct_access_url,
         checksum=req.hashes[0],
     )
     log.debug(
@@ -648,7 +648,7 @@ def _get_external_requirement_filepath(requirement: PipRequirement) -> Path:
         package = requirement.package
         hash_spec = requirement.hashes[0]
         _, _, digest = hash_spec.partition(":")
-        orig_url = urlparse.urlparse(requirement.url)
+        orig_url = urlparse.urlparse(requirement.direct_access_url)
         file_ext = ""
         for ext in ALL_FILE_EXTENSIONS:
             if orig_url.path.endswith(ext):
@@ -663,7 +663,7 @@ def _get_external_requirement_filepath(requirement: PipRequirement) -> Path:
             filepath = Path(f"{package}-{digest}{file_ext}")
 
     elif requirement.kind == "vcs":
-        git_info = extract_git_info(requirement.url)
+        git_info = extract_git_info(requirement.direct_access_url)
         repo = git_info["repo"]
         ref = git_info["ref"]
         filepath = Path(f"{repo}-gitcommit-{ref}.tar.gz")
@@ -741,7 +741,7 @@ def _replace_external_requirements(requirements_file_path: RootedPath) -> Projec
         if requirement.kind in ("url", "vcs"):
             path = _get_external_requirement_filepath(requirement)
             templated_abspath = Path("${output_dir}", "deps", "pip", path)
-            return requirement.copy(url=f"file://{templated_abspath}")
+            return requirement.update(url=f"file://{templated_abspath}")
         return None
 
     replaced = [maybe_replace(requirement) for requirement in requirements_file.requirements]
--- a/hermeto/core/package_managers/python/pip/requirements.py
+++ b/hermeto/core/package_managers/python/pip/requirements.py
@@ -231,44 +231,48 @@ class PipRequirementsFile:
         return global_options, requirement_options, " ".join(requirement)
 
 
-class PipRequirement:
+class PipRequirement(Requirement):
     """Parse a requirement and its options from a requirement line."""
 
     # Regex used to determine if a direct access requirement specifies a
     # package name, e.g. "name @ https://..."
     HAS_NAME_IN_DIRECT_ACCESS_REQUIREMENT = re.compile(r"@.+://")
 
-    def __init__(self) -> None:
+    def __init__(self, requirement_string: str) -> None:
         """Initialize a PipRequirement."""
-        # The package name after it has been processed by setuptools, e.g. "_" are replaced
-        # with "-"
-        self.package: str = ""
-        # The package name as defined in the requirement line
-        self.raw_package: str = ""
-        self.extras: set[str] = set()
-        self.version_specs: list[tuple[str, str]] = []
-        self.environment_marker: str | None = None
+        super().__init__(requirement_string)
         self.hashes: list[str] = []
         self.qualifiers: dict[str, str] = {}
-
-        self.kind: Literal["pypi", "url", "vcs"]
-        self.download_line: str = ""
-
+        self.kind: Literal["pypi", "url", "vcs"] = "pypi"
+        self.download_line: str = requirement_string
         self.options: list[str] = []
 
-        self._url: Any = None
+    @property
+    def package(self) -> str:
+        """Return the canonicalized package name."""
+        return canonicalize_name(self.name)
 
     @property
-    def url(self) -> str:
-        """Extract the URL from the download line of a VCS or URL requirement."""
-        if self._url is None:
-            if self.kind not in ("url", "vcs"):
-                raise ValueError(f"Cannot extract URL from {self.kind} requirement")
-            # package @ url ; environment_marker
-            parts = self.download_line.split()
-            self._url = parts[2]
+    def raw_package(self) -> str:
+        """Return the package name as defined in the requirement line."""
+        return self.name
 
-        return self._url
+    @property
+    def direct_access_url(self) -> str:
+        """URL for url/vcs requirements. Raises ValueError if kind is pypi."""
+        if self.url is None:
+            raise ValueError(f"Cannot extract URL from {self.kind} requirement")
+        return self.url
+
+    @property
+    def version_specs(self) -> list[tuple[str, str]]:
+        """Return version specifiers as a list of (operator, version) tuples."""
+        return [(spec.operator, spec.version) for spec in self.specifier]
+
+    @property
+    def environment_marker(self) -> str | None:
+        """Return the environment marker as a string, or None."""
+        return str(self.marker) if self.marker else None
 
     def __str__(self) -> str:
         """Return the string representation of the PipRequirement."""
@@ -278,8 +282,20 @@ class PipRequirement:
         line.extend(f"--hash={h}" for h in self.hashes)
         return " ".join(line)
 
-    def copy(self, url: str | None = None, hashes: list[str] | None = None) -> "PipRequirement":
-        """Duplicate this instance of PipRequirement.
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, PipRequirement):
+            return NotImplemented
+        return super().__eq__(other) and self.kind == other.kind
+
+    def __hash__(self) -> int:
+        return hash((super().__hash__(), self.kind))
+
+    def copy(self) -> "PipRequirement":
+        """Return an unmodified clone of this requirement."""
+        return self.update(None, None)
+
+    def update(self, url: str | None = None, hashes: list[str] | None = None) -> "PipRequirement":
+        """Return a clone of this requirement, optionally changing the URL and/or hashes.
 
         :param str url: set a new direct access URL for the requirement. If provided, the
             new requirement is always of ``url`` kind.
@@ -313,20 +329,10 @@ class PipRequirement:
                     "Removed editable option when copying the requirement %r", self.raw_package
                 )
 
-        requirement = self.__class__()
-        # Extras are incorrectly treated as part of the URL itself. If we're setting
-        # the URL, they must be skipped.
-        if not url:
-            requirement.extras = set(self.extras)
-        requirement.package = self.package
-        requirement.raw_package = self.raw_package
-        # Version specs are ignored by pip when applied to a URL, let's do the same.
-        requirement.version_specs = [] if url else list(self.version_specs)
-        requirement.environment_marker = self.environment_marker
-        requirement.hashes = list(hashes or self.hashes)
+        requirement = self.__class__(download_line)
+        requirement.hashes = list(self.hashes if hashes is None else hashes)
         requirement.qualifiers = dict(self.qualifiers)
         requirement.kind = "url" if url else self.kind
-        requirement.download_line = download_line
         requirement.options = options
 
         return requirement
@@ -343,31 +349,27 @@ class PipRequirement:
         """
         to_be_parsed = line
         qualifiers: dict[str, str] = {}
-        requirement = cls()
 
-        if not (direct_access_kind := cls._assess_direct_access_requirement(line)):
-            requirement.kind = "pypi"
-        else:
-            requirement.kind = direct_access_kind
+        direct_access_kind = cls._assess_direct_access_requirement(line)
+        if direct_access_kind:
             to_be_parsed, qualifiers = cls._adjust_direct_access_requirement(
                 to_be_parsed, cls.HAS_NAME_IN_DIRECT_ACCESS_REQUIREMENT
             )
 
         try:
-            req = Requirement(to_be_parsed)
+            requirement = cls(to_be_parsed)
         except InvalidRequirement as exc:
             # see https://github.com/pypa/setuptools/pull/2137
-            raise UnexpectedFormat(f"Unable to parse the requirement {to_be_parsed!r}: {exc}")
+            raise UnexpectedFormat(
+                f"Unable to parse the requirement {to_be_parsed!r}: {exc}"
+            ) from exc
 
         hashes, options = cls._split_hashes_from_options(options)
 
-        requirement.download_line = to_be_parsed
+        requirement.kind = direct_access_kind or "pypi"
+        if direct_access_kind and requirement.url is None:
+            raise UnexpectedFormat(f"Direct access requirement {to_be_parsed!r} has no URL")
         requirement.options = options
-        requirement.package = canonicalize_name(req.name)
-        requirement.raw_package = req.name
-        requirement.version_specs = [(spec.operator, spec.version) for spec in req.specifier]
-        requirement.extras = req.extras
-        requirement.environment_marker = str(req.marker) if req.marker else None
         requirement.hashes = hashes
         requirement.qualifiers = qualifiers
 
@@ -613,7 +615,7 @@ def validate_requirements(requirements: list[PipRequirement]) -> None:
 
         # Fail if VCS requirement uses any VCS other than git or does not have a valid ref
         elif req.kind == "vcs":
-            url = urlparse.urlparse(req.url)
+            url = urlparse.urlparse(req.direct_access_url)
 
             if not url.scheme.startswith("git"):
                 raise UnsupportedFeature(
@@ -643,7 +645,7 @@ def validate_requirements(requirements: list[PipRequirement]) -> None:
                     ),
                 )
 
-            url = urlparse.urlparse(req.url)
+            url = urlparse.urlparse(req.direct_access_url)
             if not any(url.path.endswith(ext) for ext in ALL_FILE_EXTENSIONS):
                 msg = (
                     "URL for requirement does not contain any recognized file extension: "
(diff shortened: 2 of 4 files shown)


hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -231,45 +231,53 @@ def _split_options_and_requirement(self, line: str) -> tuple[list[str], list[str
...
+
+    @property
+    def raw_package(self) -> str:
+        """Return the package name as defined in the requirement line."""
+        return self.name
 
     @property
-    def url(self) -> str:
-        """Extract the URL from the download line of a VCS or URL requirement."""
+    def url(self) -> str:  # type: ignore[override]
```
a-ovchinnikov: Did you need to override because it is typed differently in the parent class?

hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -231,45 +231,53 @@ def _split_options_and_requirement(self, line: str) -> tuple[list[str], list[str
...
+        """Return the URL of a VCS or URL requirement."""
```
a-ovchinnikov: Technically a property returns nothing, a property is a value

hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -231,45 +231,53 @@ def _split_options_and_requirement(self, line: str) -> tuple[list[str], list[str
...
         if self._url is None:
-            if self.kind not in ("url", "vcs"):
-                raise ValueError(f"Cannot extract URL from {self.kind} requirement")
-            # package @ url ; environment_marker
-            parts = self.download_line.split()
-            self._url = parts[2]
-
+            raise ValueError(f"Cannot extract URL from {self.kind} requirement")
```
a-ovchinnikov: I have some trouble with deferring input validation to the moment when someone actually uses this class. I wonder if checking at instantiation time would break anything?

hermeto/core/package_managers/python/pip/requirements.py
a-ovchinnikov: Pre-existing, but I cannot not point out that this is effectively an `update(self, url, hashes)`, `copy` should be just a `self.update(None, None)` call.

hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -231,45 +231,53 @@ def _split_options_and_requirement(self, line: str) -> tuple[list[str], list[str
...
         return self._url
 
+    @url.setter
+    def url(self, value: str | None) -> None:
```
a-ovchinnikov: Is this setter used anywhere?

a-ovchinnikov (commented):
The code as submitted looks good to me in the sense that it addresses the issue as it is stated, however I wonder if this opportunity could be used to clean the class up a bit?

hermeto/core/package_managers/python/pip/requirements.py
filipovskyj02: The parent has str | None. This just str. However after testing, it is indeed not neccessary. Removed

hermeto/core/package_managers/python/pip/requirements.py
slimreaper35: Looks like the "getter" and "setter" can be deleted. The parent class already has the `url` attribute.

hermeto/core/package_managers/python/pip/requirements.py
filipovskyj02: Good point. I made it this way as the previous code also raised the ValueError here, so to have as few changes as possible. This design choice is the one i would like to discuss the most.

hermeto/core/package_managers/python/pip/requirements.py
filipovskyj02: Yes, (by the url getter/property) but it is only needed because of the design decision to keep the value error the same. An alternative would be using parent's url, with which there would be no need to even have the url property, however that would require checks/asserts on every call site as the parents url can be None and I figured that an error would be better than a silent None. Also as mentioned it keeps the same error in the same place as the previous code. 
I dont find it ideal so am definitely open to discussion and exploring alternatives

filipovskyj02: If we wanna use the parents url (which i agree is much more elegant). We have to assert that the url's value is not None, as multiple methods like download_binary_file, or extract_git_info expect str, not str | None (which the parent has). So we can either assert on callsites, or modify these methods to check the value themselves and throw if none. What do you think ?

hermeto/core/package_managers/python/pip/requirements.py
a-ovchinnikov: Superclass's `__init__()` gets called on init time, once it returns we could check for url being `None` and raise there. One open question here is whether any logic depends on this deferred validation or not (I hope that nothing depends on it, but the best way of checking that would be to implement the change and then see if any UTs or ITs fail).

hermeto/core/package_managers/python/pip/requirements.py
a-ovchinnikov: If this class clients expect `str` then we should give them a string. Not having a string would mean that the input is broken and failure at instantiation time. This could upset mypy though, in that case I would suggest providing an override which would return super's url and andjust types accordingly.

taylormadore: Keep in mind that several methods in the parent class access url and don't expect an exception to be raised if url is None

filipovskyj02: Great catch @taylormadore. The parents iter_parts and hash do indeed use our getter for accessing it, which throws instead of returning None. So this would basically mean any PipRequirement without an url would be unhashable. Tried with a simple test.
```
def test_pypi_requirement_is_hashable(self) -> None:
        
        req = PipRequirement.from_line("requests==2.28.0", [])
        assert hash(req) is not None
        assert req == PipRequirement.from_line("requests==2.28.0", [])
        assert req in {req}
```


```
packaging/requirements.py:117: in __hash__
      return hash(tuple(self._iter_parts(canonicalize_name(self.name))))

  packaging/requirements.py:78: in _iter_parts
      if self.url:                         
  requirements.py:264: in url (getter)
      raise ValueError(f"Cannot extract URL from {self.kind} requirement")
 ValueError: Cannot extract URL from pypi requirement

```

hermeto/core/package_managers/python/pip/requirements.py
filipovskyj02: Sadly we do not have enough information about the requirement to throw in our init (after calling the parents init), as we do not know the kind, which is set later in the from_line method. Eitherway the deferred validation is going away as mentioned in a different comment as it breaks hashing for reqs without a url.

a-ovchinnikov: Is the `__init__()` used somewhere standalone? The next good candidate for raising this exception is `from_line()` -- it is used for processing individual requirements lines anyway and should fail once something unexpected is fed to it.

hermeto/core/package_managers/python/pip/requirements.py
filipovskyj02: Renamed copy to update, and added copy which just calls update(None, None).

filipovskyj02:
Addressed the review notes. Now we use the parents url. The url method call sites use casts (as the methods except just Str) and are only called in the url/vcs path so should always be populated with a Str. Asserts are forbidden by the coding standart https://docs.astral.sh/ruff/rules/assert/ as i understood.

An alternative of  casting for calling the url methods like extract_git_info would be accepting Str | None as a parameter instead of just Str and checking for None in the methods themselves. If preffered over casting i will change it as so.

mohamedbouchtout:
A feedback I recently got was to consider using the 50/72 rule; try to fit the title under 50 chars and keep lines in a body under 72 chars for the commit message.

hermeto/core/package_managers/python/pip/requirements.py
taylormadore: Consider adding a new property named something like `direct_access_url` in the `PipRequirement` class that does what the url property used to do: raise if url is None and otherwise return the url string. That wouldn't interfere with the behavior of the parent class and wouldn't require any type-checking workarounds.

hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -604,7 +599,7 @@ def validate_requirements(requirements: list[PipRequirement]) -> None:
 
         # Fail if VCS requirement uses any VCS other than git or does not have a valid ref
         elif req.kind == "vcs":
-            url = urlparse.urlparse(req.url)
+            url = urlparse.urlparse(cast(str, req.url))
```
a-ovchinnikov: Not really actionable, matter of personal preference: I find `cast` in Python code to be rather off-putting, if I am to pollute the code with kludges to substitute for mypy's lack of power I would rather choose a `type: ignore` comment with an explanation of why it is safe to ignore the error in this particular case.

tests/unit/package_managers/python/pip/test_requirements.py
```diff
@@ -987,29 +987,39 @@ def test_pip_requirement_to_str(
...
         original_requirement = PipRequirement.from_line(requirement_line, requirement_options)
-        new_requirement = original_requirement.copy(**new_values)
+        new_requirement = original_requirement.update(**new_values)
 
         for attr in self.PIP_REQUIREMENT_ATTRS:
             expected_changes.setdefault(attr, getattr(original_requirement, attr))
 
         self._assert_pip_requirement(new_requirement, expected_changes)
 
+    def test_pip_requirement_copy(self) -> None:
```
a-ovchinnikov: `copy()` is a thin wrapper around `update()`, a shorthand for calling it, it does not make sense to test it when `update()` is properly covered.

a-ovchinnikov (commented):
Looks good to me in general, will approve once the PR is ready for review, but please at the very least squash the commits -- the second commit is not logically separate from the first one, but amends what was done there instead. For the perfect score you can split the resulting commit into separate commits by grouping related changes together, e.g. by splitting out `copy()` changes.

tests/unit/package_managers/python/pip/test_requirements.py
filipovskyj02: Removed.

hermeto/core/package_managers/python/pip/requirements.py
filipovskyj02: Good suggestion, i did not like the casts nor type ignores, this feels much better.

hermeto/core/package_managers/python/pip/requirements.py
filipovskyj02: I tried to rework it to use the type ignores, then came across Taylors suggestion of a direct_access_url property and did it that way, as i personally did not like casts nor the type ignores as they are just workarounds.

filipovskyj02:
Squashed commits into one. Went with the suggestion to introduce a new property  direct_access_url so that we dont have to do casting or ignore type check workourounds. Adhered to the 50/72 git rule

hermeto/core/package_managers/python/pip/requirements.py
a-ovchinnikov: > the deferred validation is going away

Did it go away after all? I tried it out and from the looks of it just failing validation at instantiation did not really work, quite a few tests fell apart. I have not dug deeper into it, but now I suspect that `PipRequriement` models three different classes instead of one. I think it is a good candidate for becoming a union type of several individual subtypes, but that is out of scope for this PR.

a-ovchinnikov (commented):
Do you plan to work more on this PR? If no then please rebase and mark it as ready for review.

tnevrlka:
@filipovskyj02 How's this progressing?

filipovskyj02:
Rebased the main repo (hopefully correctly). Changed from draft to ready. I plan no further additions, so ready to merge. Lets see what the pipeline says as most checks did not run on a draft

a-ovchinnikov (approved):

hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -343,31 +341,26 @@ def from_line(cls, line: str, options: list[str]) -> "PipRequirement":
...
         except InvalidRequirement as exc:
             # see https://github.com/pypa/setuptools/pull/2137
             raise UnexpectedFormat(f"Unable to parse the requirement {to_be_parsed!r}: {exc}")
 
         hashes, options = cls._split_hashes_from_options(options)
 
+        requirement.kind = direct_access_kind or "pypi"
+        if direct_access_kind and requirement.url is None:
+            raise UnexpectedFormat(f"Direct access requirement {to_be_parsed!r} has no URL")
         requirement.download_line = to_be_parsed
```
tnevrlka: Is this necessary? I'd say it's set in `__init__` [link](https://github.com/filipovskyj02/hermeto/blob/f9003e76b63ea73a418aff292d7e5783722c114f/hermeto/core/package_managers/python/pip/requirements.py#L247)

tnevrlka (approved):
ACK

hermeto/core/package_managers/python/pip/main.py
tnevrlka, line 1: ad commit message:

`Add backward-compat` is cutoff

hermeto/core/package_managers/python/pip/requirements.py
filipovskyj02: Redundant download line removed

tests/unit/package_managers/python/pip/test_requirements.py
```diff
@@ -987,30 +987,27 @@ def test_pip_requirement_to_str(
...
-        requirement.download_line = "aiowsgi==0.7"
-        requirement.kind = "pypi"
-
-        with pytest.raises(ValueError):
-            _ = requirement.url
+    def test_pypi_requirement_has_no_url(self) -> None:
+        """Test that a PyPI requirement has no URL."""
+        requirement = PipRequirement.from_line("aiowsgi==0.7", [])
+        assert requirement.kind == "pypi"
+        assert requirement.url is None
```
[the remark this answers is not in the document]
filipovskyj02: Added.

hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -313,20 +321,10 @@ def copy(self, url: str | None = None, hashes: list[str] | None = None) -> "PipR
...
-            requirement.extras = set(self.extras)
-        requirement.package = self.package
-        requirement.raw_package = self.raw_package
-        # Version specs are ignored by pip when applied to a URL, let's do the same.
-        requirement.version_specs = [] if url else list(self.version_specs)
-        requirement.environment_marker = self.environment_marker
+        requirement = self.__class__(download_line)
         requirement.hashes = list(hashes or self.hashes)
         requirement.qualifiers = dict(self.qualifiers)
         requirement.kind = "url" if url else self.kind
```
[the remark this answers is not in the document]
filipovskyj02: Adde

taylormadore (approved):

tests/unit/package_managers/python/pip/test_requirements.py
```diff
@@ -987,30 +987,33 @@ def test_pip_requirement_to_str(
...
-        """Test PipRequirement.copy method."""
+        """Test PipRequirement.update method."""
         original_requirement = PipRequirement.from_line(requirement_line, requirement_options)
-        new_requirement = original_requirement.copy(**new_values)
+        new_requirement = original_requirement.update(**new_values)
 
         for attr in self.PIP_REQUIREMENT_ATTRS:
             expected_changes.setdefault(attr, getattr(original_requirement, attr))
 
         self._assert_pip_requirement(new_requirement, expected_changes)
```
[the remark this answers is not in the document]
filipovskyj02: Added dedicated update covering hashes, did not add copy test back in as previous discussions deemed it unneccessary

tests/unit/package_managers/python/pip/test_requirements.py
```diff
@@ -987,30 +987,33 @@ def test_pip_requirement_to_str(
...
+        assert requirement.kind == "pypi"
+        assert requirement.url is None
 
-        with pytest.raises(ValueError):
-            _ = requirement.url
+    def test_direct_access_url_raises_for_pypi(self) -> None:
+        """Test that direct_access_url raises ValueError for PyPI requirements."""
+        requirement = PipRequirement.from_line("aiowsgi==0.7", [])
+        with pytest.raises(ValueError, match="Cannot extract URL from pypi requirement"):
+            requirement.direct_access_url
```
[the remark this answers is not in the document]
filipovskyj02: Done.

tests/unit/package_managers/python/pip/test_requirements.py
```diff
@@ -987,30 +987,43 @@ def test_pip_requirement_to_str(
...
+        assert requirement.kind == "pypi"
+        assert requirement.url is None
 
-        with pytest.raises(ValueError):
-            _ = requirement.url
+    def test_direct_access_url_raises_for_pypi(self) -> None:
+        """Test that direct_access_url raises ValueError for PyPI requirements."""
+        requirement = PipRequirement.from_line("aiowsgi==0.7", [])
+        with pytest.raises(ValueError, match="Cannot extract URL from pypi requirement"):
+            _ = requirement.direct_access_url
```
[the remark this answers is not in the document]
filipovskyj02: Added.

tests/unit/package_managers/python/pip/test_requirements.py
```diff
@@ -987,30 +987,43 @@ def test_pip_requirement_to_str(
...
         ),
     )
-    def test_pip_requirement_copy(
+    def test_pip_requirement_update(
         self,
         requirement_line: str,
         requirement_options: list[str],
         new_values: dict[str, str] | dict[str, list[str]],
         expected_changes: dict[str, str],
     ) -> None:
```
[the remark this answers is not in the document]
filipovskyj02: Changed to str, Any

tnevrlka (approved):
Still ACK

hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -231,44 +231,48 @@ def _split_options_and_requirement(self, line: str) -> tuple[list[str], list[str
         return global_options, requirement_options, " ".join(requirement)
 
 
-class PipRequirement:
+class PipRequirement(Requirement):
```
eskultety, line 234: Note by inheriting from Requirement we also inherit the `__eq__` and `__hash__` : https://github.com/pypa/packaging/blob/55cbf1b9426f44455fa1a9e0836f1fc082cc8452/src/packaging/requirements.py#L163-L179

^this has a significant implication - any equality check (either directly or indirectly via hashing) is completely based off of the inherited fields rather than our own. Looking at our fields though only `kind` looks like it might be worth tossing into the hashing mix. We should override those magic methods and be explicit rather than implicit and face some surprising errors in the future.

filipovskyj02: Great point, gonna override

filipovskyj02: Added __eq__ and __hash_override with a simple test. Also added a default Kind value (that gets overwritten in the from_line and update with the correct one) in case of a construction without using these 2 methods

tests/unit/package_managers/python/pip/test_requirements.py
```diff
@@ -987,30 +987,79 @@
...
+        req_a = PipRequirement.from_line(
+            "cnr_server @ https://github.com/quay/appr/archive/58c88e49.tar.gz", []
+        )
+        req_b = PipRequirement.from_line(
+            "cnr_server @ https://github.com/quay/appr/archive/58c88e49.tar.gz", []
+        )
+        assert req_a == req_b
+
+        req_b.kind = "vcs"
+        assert req_a != req_b
```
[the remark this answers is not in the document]
tnevrlka: `req_b` gets changed by `req_b.kind = "vcs"` so the asserts should have the chance to be different

hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -278,8 +282,27 @@ def __str__(self) -> str:
...
-        """Duplicate this instance of PipRequirement.
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, PipRequirement):
+            return NotImplemented
+        return (
+            canonicalize_name(self.name) == canonicalize_name(other.name)
+            and self.extras == other.extras
+            and self.specifier == other.specifier
+            and self.url == other.url
+            and self.marker == other.marker
```
eskultety: Can we re-use the parent implementation here (instead of re-implementing it ourselves) and just add `self.kind == other.kind` to it?

hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -278,8 +282,27 @@ def __str__(self) -> str:
...
+            and self.kind == other.kind
+        )
+
+    def __hash__(self) -> int:
+        return hash((*self._iter_parts(canonicalize_name(self.name)), self.kind))
```
eskultety: How about more explicitly:
`return hash((super().__hash__(), self.kind))`?

eskultety (changes requested):

hermeto/core/package_managers/python/pip/requirements.py
filipovskyj02: We could, but i wanted do be explicit, so that when someone want to see the implementation, they dont have to check the parents one as well. And to have full control over the fields, but either seems valid so i can change if needed.

eskultety: Calling into the parent is **explicit** because you're saying, whatever the parent is doing is fine with us, we don't care, we just need to tweak it slightly... :)

filipovskyj02: I see its a requested change, so gonna reuse as stated.

hermeto/core/package_managers/python/pip/requirements.py
```diff
@@ -231,44 +231,48 @@ def _split_options_and_requirement(self, line: str) -> tuple[list[str], list[str
         return global_options, requirement_options, " ".join(requirement)
 
 
-class PipRequirement:
+class PipRequirement(Requirement):
     """Parse a requirement and its options from a requirement line."""
 
     # Regex used to determine if a direct access requirement specifies a
     # package name, e.g. "name @ https://..."
     HAS_NAME_IN_DIRECT_ACCESS_REQUIREMENT = re.compile(r"@.+://")
```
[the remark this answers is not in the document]
filipovskyj02: I feel like this is unnecessary and such a long messy type annotation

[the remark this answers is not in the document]
a-ovchinnikov: > I feel like this is unnecessary

It is unnecessary for pretty much any code _within_ a function, our beloved code review agent applies guidelines without a hint of understanding context.

hermeto/core/package_managers/python/pip/requirements.py
filipovskyj02: Probably need a confirmation

eskultety (approved):

PR #1702, opened 2026-08-05
fix(pip): skip requirements excluded by environment markers under binary filters

### LINKED ISSUE #1570: pip: binary filter ignores environment markers, Hermeto could fails on excluded-platform packages
## Description

When using the `binary` filter with `arch: ppc64le` (or `s390x`), hermeto attempts to download packages that have environment markers explicitly excluding those platforms (e.g. `platform_machine != "ppc64le"`). ~~Hermeto should skip these requirements entirely based on the markers, but instead tries to find arch-specific wheels and fails with `PackageRejected`.~~

_Edit : Hermeto should skip these requirements entirely based on the markers. 
**Hermeto falls back to fetching an sdist if available, skipping the missing wheel.** However, Hermeto could potentially fail if the `requirements.txt` file doesn't record the hash for the respective sdist (via manual requirements.txt edits) and only records wheel hashes._

## Reproducer (synthetic)

```bash
mkdir /tmp/hermeto-repro && cd /tmp/hermeto-repro

#   NOTE THE EXPLICIT HASH SPECIFIER IT WILL NOT MATCH AN SDIST
cat > requirements.txt << 'EOF'
bcrypt==5.0.0 ; implementation_name == "cpython" and platform_machine != "ppc64le" and platform_machine != "s390x" and sys_platform == "linux" \
    --hash=sha256:81e5dab254e44e82bee49bca4e593b21b7e70a48bcf15725d6588df387098e54
EOF

cat > pyproject.toml << 'EOF'
[project]
name = "repro"
version = "0.1.0"
EOF

mkdir -p /tmp/hermeto-out
podman run --rm \
  -v "$(pwd):/source:z" \
  -v /tmp/hermeto-out:/output:z \
  ghcr.io/hermetoproject/hermeto:0.51.0 \
  fetch-deps --source /source --output /output \
  '[{"type":"pip","path":".","requirements_files":["requirements.txt"],"binary":{"arch":"ppc64le","os":"linux"}}]'
```

## Expected behavior

Hermeto should recognize that `bcrypt==5.0.0` has `platform_machine != "ppc64le"` in its environment marker and skip it entirely when the binary filter specifies `arch: ppc64le`.

## Actual behavior

```
INFO -- Processing requirement line 'bcrypt==5.0.0 ; implementation_name == "cpython" and platform_machine != "ppc64le" ...'
INFO Filtering out bcrypt-5.0.0.tar.gz due to checksum mismatch
ERROR PackageRejected: No distributions found for package bcrypt==5.0.0
```

Hermeto processes the requirement, tries to find a ppc64le wheel, can't find one, and fails.

ide-developer:
Closes https://github.com/hermetoproject/hermeto/issues/1570

## Problem

When a pip `binary` filter declares a target platform (`arch`/`os`/`py_impl`/`py_version`), Hermeto never consults a requirement's PEP 508 environment marker. A requirement whose marker excludes the target — e.g. `platform_machine != "ppc64le"` — is resolved anyway, even though `pip install` would simply skip it.

For a **wheel-only lock with no sdist hash** this is fatal: `WheelsFilter` finds no matching wheel, the sdist fallback is filtered out by checksum (or no sdist exists at all), and Hermeto fails with `PackageRejected: No distributions found`.

This is the real-world AIPCC case from the issue thread: the binary index [deliberately ships no sdists](https://packages.redhat.com/domains/public-rhai/distributions), and the lock uses markers to declare per-arch availability. As the thread notes, a `pip-compile --generate-hashes` lock that also records the sdist hash already succeeds via the existing sdist fallback — **this PR does not change that path**. It fixes only the wheel-only case, where there is nothing valid to fall back to and the marker already says "not for this platform."

## Fix

Add `requirement_matches_binary_filters` (`pip/markers.py`) and consult it in `_download_dependencies` before any wheel/sdist resolution, for pypi, url and vcs kinds alike. When a binary filter is set and a requirement's marker is false for every selected filter combination, the requirement is skipped. Source-only mode (no `binary`) is unchanged.

The marker is evaluated against an environment **synthesized only from the user's filters, never the host** (honoring *No Runtime Platform Detection*). Each filter pins the marker fields it fully determines:

- `arch` → `platform_machine`
- `os` → `sys_platform`, `os_name`, `platform_system` (tokens are translated to PEP 508 values — the documented `macosx` token maps to `sys_platform` `darwin`, not the literal token)
- `py_impl` → `implementation_name`, `platform_python_implementation`
- `py_version` → `python_version` (minor only; the `python_full_version` patch level stays unknown)

**Referenced-key gate + fail-open.** A marker is evaluated only when every variable it reads is pinned by the filters; if it reads any unpinned dimension (`:all:`, an unset `py_version`, `platform` regex mode, or an unmodeled key such as `platform_release`), the requirement is kept. This gate is used rather than relying on evaluation to fail on unpinned values, because `packaging` routes operators through different code paths (`x in y` runs `str.__contains__` on the literal; `<`/`>` on non-version keys are a constant `False`), so no sentinel value can signal "unknown" for every operator.

Fail-open is deliberate: a false *skip* silently omits a needed dependency (the worst outcome for a hermetic build — a late `ModuleNotFoundError` at image-build time), whereas a false *keep* is wasted bandwidth or a loud, diagnosable `PackageRejected`.

## Scope / limitations

- Active only when a `binary` filter is set; source-only mode never skips on markers.
- `platform` (regex) mode pins no marker fields (it matches wheel tags directly) → always fail-open.
- Hash and requirement validation still run over the whole file **before** the skip, so a skipped line with a missing/invalid hash is still rejected — deliberate, since validation is a security control.

## Testing

- New `tests/unit/package_managers/python/pip/test_markers.py` drives the real `packaging.markers.Marker.evaluate` through real `PipRequirement.from_line` / `PipBinaryFilters` (flagship: the issue's bcrypt marker under `arch=ppc64le` is skipped). `markers.py` is at 100% line coverage.
- Download-loop regression in `test_main.py` proves a marker-excluded requirement never reaches the resolver/downloader — keep **and** skip paths, for pypi/url/vcs.
- Integration scenario `pip_binary_marker_skip_excluded_arch` mirrors `pip_no_sdists` but asserts success + a skip log (runs under `nox -s integration-tests`).
- Docs: `docs/pip.md` (user-facing) and `docs/design/binary-platform-filtering.md` (rationale).
- Locally: full unit suite (`1463 passed, 7 skipped`) and `ruff check` / `ruff format --check` / `mypy` all green.

## Ethos fit

- **Accurate prefetch/SBOM** — does not fetch (or fail fetching) artifacts that would never be installed on the target platform.
- **No arbitrary code execution** — pure `packaging.markers` evaluation; no resolver invocation.
- **Checksum validation** — unchanged for fetched packages; skipped lines never enter download.
- **Reproducibility** — the decision depends only on the committed markers and the request's `binary` filters, never on the host.

## AI assistance

This contribution is AI-assisted, disclosed via `Assisted-by:` trailers on every commit alongside the DCO `Signed-off-by`, per the AI Contribution Policy. The human contributor has reviewed and understands every line and is accountable for the submission.

### DIFF
--- a/docs/design/binary-platform-filtering.md
+++ b/docs/design/binary-platform-filtering.md
@@ -165,6 +165,12 @@ The pip backend will be updated to filter Python wheels based on user-provided p
   - For OS, linux
   - For architecture, x86_64
 
+**Environment marker evaluation**:
+- Before artifact selection, the pip backend evaluates each requirement's [PEP 508 environment marker][] against an environment synthesized *only* from the user-provided binary filters (`arch`, `os`, `py_impl`, `py_version`), never from the host — consistent with **No Runtime Platform Detection**. This is separate from, and precedes, the wheel-tag filtering described above.
+- A requirement whose marker is false for every selected filter combination is skipped entirely (no wheel and no sdist). This restores pip's own behavior for wheel-only locks (e.g. private indexes without sdists) that mark a package unavailable on an arch: instead of failing with `PackageRejected` while searching for a nonexistent wheel, Hermeto skips it. See [#1570](https://github.com/hermetoproject/hermeto/issues/1570).
+- A marker is evaluated only when every environment variable it reads is pinned by the filters; if it reads any unpinned dimension the requirement is kept (fail-open). This referenced-variable gate is used rather than relying on evaluation to fail on unpinned values, because packaging routes operators through different code paths (`x in y` runs `str.__contains__` on the literal; `<`/`>` on non-version keys are a constant `False`), so no sentinel value can signal "unknown" for every operator. Unpinnable dimensions include `platform_release`, `extra`, and `python_full_version`'s patch level (`py_version` pins only the minor). A false *keep* wastes bandwidth or surfaces as a diagnosable error; a false *skip* would silently omit a needed dependency, which is worse for a hermetic build.
+- `os` tokens live in a different string space than PEP 508 `sys_platform`: they are substring-matched against wheel platform *tags* (`macosx_*`, `manylinux*`, `win*`), so the canonical macOS token is `macosx` (not `darwin`). The marker env therefore translates each `os` token to PEP 508 values (`macosx`/`macos`/`osx`/`darwin` → `sys_platform="darwin"`, `win`/`win32`/`windows` → `win32`); an unrecognized token leaves the OS marker fields unknown (fail-open). The `platform` (regex) filter pins no marker fields at all — it matches tags directly — so markers are always kept in that mode.
+
 **Platform Matching Examples**:
 - Example 1: Prefetching for Python 3.11 on Linux
   ```json
@@ -269,4 +275,5 @@ The rpm backend will filter packages based on a user-provided architecture list,
 [prebuildify]: https://github.com/prebuild/prebuildify
 [wheelios]: https://github.com/chmeliik/wheelios
 [packaging platform tags]: https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/
+[PEP 508 environment marker]: https://peps.python.org/pep-0508/#environment-markers
 [Gem:Platform]: https://github.com/rubygems/rubygems/blob/8ad4509f95c90cf9523a82ca917b6b842fd37132/lib/rubygems/platform.rb#L10
--- a/hermeto/core/package_managers/python/pip/markers.py
+++ b/hermeto/core/package_managers/python/pip/markers.py
@@ -1,0 +1,299 @@
+# SPDX-License-Identifier: GPL-3.0-only
+"""Evaluate PEP 508 environment markers against user-supplied binary filters.
+
+When a ``binary`` filter declares a target platform (arch/os/python), a
+requirement whose environment marker excludes that platform (for example
+``platform_machine != "ppc64le"``) would never be installed by pip. Hermeto must
+therefore skip it instead of trying to resolve an artifact that does not exist,
+which otherwise fails with ``PackageRejected`` for wheel-only locks that carry no
+matching sdist hash.
+
+See https://github.com/hermetoproject/hermeto/issues/1570
+
+Approach: a marker is evaluated only when every environment variable it reads is
+pinned by the filters; if it reads any dimension the filters do not pin, the
+requirement is kept (fail-open). This "referenced-key" gate is deliberate rather
+than relying on marker evaluation to fail on unpinned values: packaging routes
+different operators through different code paths (e.g. ``x in y`` runs
+``str.__contains__`` on the literal, and ``<`` / ``>`` on non-version keys are a
+constant ``False``), so a sentinel value cannot reliably signal "unknown" for
+every operator.
+
+Design constraints:
+
+* No runtime platform detection. The evaluation environment is built solely from
+  the user's filters. ``packaging.markers.Marker.evaluate`` merges the supplied
+  environment *over* ``default_environment()`` (the real host), so every
+  ``default_environment()`` key must be present; unpinned keys use a poison value
+  (never a host value).
+
+* Fail open. For a hermetic build tool a false *skip* is the worst outcome (a
+  silent ``ModuleNotFoundError`` at image build time), whereas a false *keep* is
+  either wasted bandwidth or a loud, diagnosable ``PackageRejected``.
+"""
+
+import logging
+from collections.abc import Iterator
+from itertools import product
+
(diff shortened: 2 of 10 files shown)


tnevrlka:
@jiridanek I'm curious, this PR is opened by a bot 
> This is a bot account helping on automated tasks

but the commits are signed-off by you and apparently reviewed by you
> This contribution is AI-assisted, disclosed via Assisted-by: trailers on every commit alongside the DCO Signed-off-by, per the AI Contribution Policy. The human contributor has reviewed and understands every line and is accountable for the submission.

Why not just create the PR yourself then?

jiridanek:
don't know, no reason, really

docs/design/binary-platform-filtering.md
jiridanek, line 1: @coderabbitai I can't stop questioning myself whether this PR is really the correct solution for the linked issue.

Currently, the opendatahub-io/notebooks repo solves python downloads by an ad-hoc script that reads all file references in a requirements.txt and downloads that. If I wanted to make that approach platform aware, I'd add filtering like this:

> If an artifact has a platform suffix, and that platform is not the desired platform, then skip it.

This PR tries something much more ambitious. It extends the hermeto's parsing of the per dependency predicates, so that hermeto can check that skipping package A on PowerPC is not just because no wheel was locked for platform, but because package A is excluded for platform. I do not really need that check, do I.

I'd be just as happy with my simple downloader script.

tnevrlka: @coderabbitai I can't stop questioning myself whether this PR is really the correct solution for the linked issue.

Currently, the opendatahub-io/notebooks repo solves python downloads by an ad-hoc script that reads all file references in a requirements.txt and downloads that. If I wanted to make that approach platform aware, I'd add filtering like this:

> If an artifact has a platform suffix, and that platform is not the desired platform, then skip it.

This PR tries something much more ambitious. It extends the hermeto's parsing of the per dependency predicates, so that hermeto can check that skipping package A on PowerPC is not just because no wheel was locked for platform, but because package A is excluded for platform. I do not really need that check, do I.

I'd be just as happy with my simple downloader script.

tnevrlka: @coderabbitai
> Learning: For Hermeto pip dependency downloads, tnevrlka prefers a simple explicit-artifact workflow when `requirements.txt` lists concrete file references: skip an artifact when its platform suffix does not match the target platform. PEP 508 requirement-marker evaluation is justified only when Hermeto must distinguish a requirement intentionally excluded for a target platform from an incomplete lock with no matching artifact.

I don't XD

hermeto/core/package_managers/python/pip/markers.py
```diff
@@ -0,0 +1,299 @@
...
+            ", ".join(repr(key) for key in sorted(uncontrolled)),
+        )
+        return True
+
+    for env in _iter_filter_envs(filters):
+        try:
+            # context="requirement" avoids injecting extra="" the way the default
+            # metadata context does; combined with the gate above, evaluation only
+            # touches pinned, concrete values.
+            if marker.evaluate(env, context="requirement"):
```
derasdf: context="requirement" only exists in recent packaging, but pyproject.toml pins no lower bound (only the generated requirements*.txt do). Older packaging -> TypeError, which the except below swallows into a fail-open keep, so the whole feature silently no-ops.

And I don't think it's needed: the referenced-key gate already drops any marker reading `extra` before evaluate() runs, so no extra="" ever reaches here. s/marker.evaluate(env, context="requirement")/marker.evaluate(env)/ and the version floor goes away with it (supersedes coderabbit's add-a-lower-bound note).

docs/design/binary-platform-filtering.md
```diff
@@ -269,4 +275,5 @@ The rpm backend will filter packages based on a user-provided architecture list,
 [prebuildify]: https://github.com/prebuild/prebuildify
 [wheelios]: https://github.com/chmeliik/wheelios
 [packaging platform tags]: https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/
+[PEP 508 environment marker]: https://peps.python.org/pep-0508/#environment-markers
 [Gem:Platform]: https://github.com/rubygems/rubygems/blob/8ad4509f95c90cf9523a82ca917b6b842fd37132/lib/rubygems/platform.rb#L10
```
derasdf: nitpick: missing trailing newline at EOF.

PR #1703, opened 2026-08-06
Fix the language stats on our main repo page

eskultety:
Ever since we migrated most of our integration tests from the dedicated repo GH recalculated our language stats downing our main language - Python in favour of JS due to all the test scenarios we have for npm/yarn.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,7 @@
 .git_archival.txt  export-subst
+
+# Integration test fixtures contain real lockfiles and source trees
+# in many languages; exclude them from GitHub's language stats.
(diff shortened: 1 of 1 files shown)


eskultety:
Let me know if you prefer excluding the whole `tests/integration` dir. The way it's proposed is a bit error prone, but excluding the whole tests dir skips our harness configuration files and actual test logic, which could also be fine, I'm open to either.

tnevrlka (approved):
Good find, I didn't even realize this is possible 👍

a-ovchinnikov (approved):

PR #1712, opened 2026-08-10
.github: gating: Add the 'merge_group' even trigger to the workflow

eskultety:
Apparently the following [1]:

    'You must use the merge_group event to trigger your GitHub Actions
    workflow when a pull request is added to a merge queue.'

means that if you use "Require status checks to pass" as well as "Require merge queues" on your github project the CI will be stuck on waiting for required checks in the merge queue unless `merge_group` trigger is configured for a workflow. IOW checks are not skipped in the merge queue, they are again required to run (even if they just had passed, allowing one to actually put a PR to the queue), thanks GitHub...

[1] https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#configuring-continuous-integration-ci-workflows-for-merge-queues

### DIFF
--- a/.github/workflows/gating.yaml
+++ b/.github/workflows/gating.yaml
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - main
+  merge_group:
+    types: [checks_requested]
   workflow_dispatch:
     inputs: {}
 


tnevrlka (approved):

tnevrlka (approved):
nit:
```
  merge_group:
    types: [checks_requested]
```
might be slightly better

> More than one activity type triggers this event. Although only the checks_requested activity type is supported, specifying the activity type will keep your workflow specific if more activity types are added in the future. For information about each activity type, see [Webhook events and payloads](https://docs.github.com/en/webhooks/webhook-events-and-payloads#merge_group). By default, all activity types trigger workflows that run on this event. You can limit your workflow runs to specific activity types using the types keyword. For more information, see [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onevent_nametypes).
> If your repository uses GitHub Actions to perform required checks on pull requests in your repository, you need to update the workflows to include the merge_group event as an additional trigger. Otherwise, status checks will not be triggered when you add a pull request to a merge queue. The merge will fail as the required status check will not be reported. The merge_group event is separate from the pull_request and push events.

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#merge_group

derasdf (approved):

PR #1716, opened 2026-08-11
fix(bundler): add repository_url PURL qualifier

### LINKED ISSUE #1705: bundler backend omits repository_url PURL qualifier for gems from non-default registries
### Describe the issue

When a Gemfile.lock declares gems from a source other than https://rubygems.org, the bundler backend produces PURLs without a `repository_url` qualifier. This makes gems from private or alternative registries indistinguishable from rubygems.org gems in the SBOM. Other backends like pip handle this correctly.

### Version

- [x] I have reproduced this issue on the latest version from the main branch

### Steps to reproduce

Using Hermeto, fetch-deps for a bundler project that uses something other than https://rubygems.org in Gemfile/Gemfile.lock. Observe that the gems listed in the SBOM do not have `repository_url` qualifiers.

### Environment

Linux, containerized

### How does this issue affect you, or how did you find it?

This bug reduces SBOM accuracy

Nikita-003002:
When a Gemfile.lock specifies gems from a source other than rubygems.org, the bundler backend now attaches the repository_url qualifier to the PURL. This ensures custom registries are properly distinguished in the SBOM.

Fixes #1705

Assisted-by: Gemini

### DIFF
--- a/hermeto/core/package_managers/bundler/gem_models.py
+++ b/hermeto/core/package_managers/bundler/gem_models.py
@@ -3,7 +3,7 @@ import logging
 from functools import cached_property
 from pathlib import Path
 from typing import Annotated
-from urllib.parse import urljoin, urlparse
+from urllib.parse import ParseResult, urljoin, urlparse
 
 import pydantic
 from packageurl import PackageURL
@@ -18,6 +18,8 @@ from hermeto.core.models.sbom import PROXY_COMMENT, Component, ExternalReference
 from hermeto.core.package_managers.general import get_vcs_qualifiers
 from hermeto.core.rooted_path import PathOutsideRoot, RootedPath
 from hermeto.core.scm import GitRepo
+
+RUBYGEMS_URLS: tuple[str, str] = ("https://rubygems.org", "http://rubygems.org")
 
 AcceptedUrl = Annotated[
     pydantic.HttpUrl,
@@ -75,19 +77,35 @@ class GemDependency(_GemMetadata):
         checksum:   The checksum of the gem.
     """
 
-    source: str
+    source: HttpUrl
     checksum: str | None = None
 
     @cached_property
     def purl(self) -> str:
         """Get PURL for this dependency."""
-        purl = PackageURL(type="gem", name=self.name, version=self.version)
+        qualifiers: dict[str, str] | None = None
+        normalized_source: str = str(self.source).rstrip("/")
+
+        parsed: ParseResult = urlparse(normalized_source)
+
+        host: str = parsed.hostname or ""
+        if ":" in host:
+            host = f"[{host}]"
+        if parsed.port is not None:
+            host += f":{parsed.port}"
+
+        normalized_source = parsed._replace(netloc=host, query="", fragment="").geturl()
+
+        if normalized_source not in RUBYGEMS_URLS:
+            qualifiers = {"repository_url": normalized_source}
+
+        purl = PackageURL(type="gem", name=self.name, version=self.version, qualifiers=qualifiers)
         return purl.to_string()
 
     @cached_property
     def remote_location(self) -> str:
         """Return remote location to download this gem from."""
-        return urljoin(self.source, f"downloads/{self.name}-{self.version}.gem")
+        return urljoin(str(self.source), f"downloads/{self.name}-{self.version}.gem")
 
     def download_location(self, deps_dir: RootedPath) -> RootedPath:
         """Get the file system location of the gem."""
@@ -116,7 +134,9 @@ class GemPlatformSpecificDependency(GemDependency):
         # -gnu suffix being dropped from some platforms. This was observed on
         # sqlite3-aarch-linux-gnu. We discourage using outdated platforms
         # for building dependencies and cnsider this to be a limitation of Ruby.
-        return urljoin(self.source, f"downloads/{self.name}-{self.version}-{self.platform}.gem")
+        return urljoin(
+            str(self.source), f"downloads/{self.name}-{self.version}-{self.platform}.gem"
+        )
 
     def download_location(self, deps_dir: RootedPath) -> RootedPath:
         """Get the file system location of the gem."""
--- a/tests/unit/package_managers/bundler/test_gem_models.py
+++ b/tests/unit/package_managers/bundler/test_gem_models.py
@@ -74,3 +74,43 @@ def test_to_component_non_registry_deps_ignore_proxy(rooted_tmp_path_repo: Roote
     for dep in deps:
         component = dep.to_component(proxy_url=HttpUrl(FAKE_PROXY_URL))
         assert component.external_references is None
+
+
+@pytest.mark.parametrize(
+    ("dep", "expected_purl"),
+    [
+        pytest.param(
+            GemDependency(name="rails", version="7.0.0", source="https://rubygems.org/"),
+            "pkg:gem/rails@7.0.0",
+            id="default-rubygems-source",
+        ),
+        pytest.param(
+            GemDependency(name="rails", version="7.0.0", source="http://rubygems.org/"),
+            "pkg:gem/rails@7.0.0",
+            id="default-rubygems-source-http",
+        ),
+        pytest.param(
+            GemDependency(name="custom_gem", version="1.0.0", source="https://gems.example.com"),
+            "pkg:gem/custom_gem@1.0.0?repository_url=https://gems.example.com",
+            id="custom-registry-source",
+        ),
+        pytest.param(
+            GemDependency(
+                name="trailing_slash_gem", version="1.0.0", source="https://gems.example.com/"
+            ),
+            "pkg:gem/trailing_slash_gem@1.0.0?repository_url=https://gems.example.com",
+            id="custom-registry-source-trailing-slash",
+        ),
+    ],
+)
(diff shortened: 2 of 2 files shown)


tnevrlka (commented):
Thanks for the contribution. Have you tried running integration tests? I'm afraid they're failing on the changes

hermeto/core/package_managers/bundler/gem_models.py
a-ovchinnikov, line 1: https://github.com/hermetoproject/hermeto/commit/22c775fea7bed4b3230d47dbb05524bde2706f3d _Commit_: please incorporate suggested fixes to commits that you introduce into these commits, there is no point in tracking them separately.

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -81,7 +81,12 @@ class GemDependency(_GemMetadata):
     @cached_property
     def purl(self) -> str:
         """Get PURL for this dependency."""
-        purl = PackageURL(type="gem", name=self.name, version=self.version)
+        qualifiers = None
+        normalized_source = self.source.rstrip("/")
+        if normalized_source and normalized_source != "https://rubygems.org":
```
a-ovchinnikov: Nitpick: the first check is redundant, `source` must be passed to the object during initialization. Now the way the code is currently written one could technically pass "/" to the initializer and that will work, but that problem must be solved by re-annotating `source` with `HttpUrl`

tests/integration/bundler/scenarios/bundler_e2e_missing_gemspec/out/bom.json
a-ovchinnikov, line 1: https://github.com/hermetoproject/hermeto/commit/a8a949fd39645dfae54ee851b2c6689262a2f854:

Please squash this commit into previous one. Without changes introduced in this commit there will be `main` state in which ITs fail and that should not happen for several reasons.

tests/integration/bundler/scenarios/bundler_e2e_missing_gemspec/out/bom.json
```diff
@@ -34,7 +34,7 @@
         "pkg:gem/method_source@1.1.0",
         "pkg:gem/mini_mime@1.1.5",
         "pkg:gem/minitest@5.24.1",
-        "pkg:gem/mygem@0.0.1",
+        "pkg:gem/mygem@0.0.1?repository_url=http://rubygems.org",
```
slimreaper35: ou, that doesn't seem right

slimreaper35: I don't remember why each Gemfile has:

```ruby
source "http://rubygems.org" do
  gem "mygem"
end
```

Nikita-003002: Which is why it was causing the old logic to treat it as a custom registry. Now that http and https are both recognized as default registries, it bypasses the qualifier properly.

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -81,7 +81,12 @@ class GemDependency(_GemMetadata):
     @cached_property
     def purl(self) -> str:
         """Get PURL for this dependency."""
-        purl = PackageURL(type="gem", name=self.name, version=self.version)
```
eskultety, line 84: **Commit message:**

> 
> Integration test expected outputs have been regenerated to account for
> this new qualifier

I don't think this is actually true as I don't see any IT suite changes in this patch.

> Fixes 1705

This is a short ID reference which only renders correctly in GH, we always mandate using full URLs when referencing issues which renders correctly both in local environments as well as in GH web UI.

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -81,7 +81,12 @@ class GemDependency(_GemMetadata):
...
+        qualifiers: dict[str, str] | None = None
+        normalized_source: str = self.source.rstrip("/")
+        if normalized_source not in ("https://rubygems.org", "http://rubygems.org"):
```
eskultety: nitpick: these hardcoded values don't seem to be consistent with what we do in other backends, e.g. defining `NPM_REGISTRY_URL` , `YARN_REGISTRY_URL`, etc.

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -81,7 +81,12 @@ class GemDependency(_GemMetadata):
     @cached_property
     def purl(self) -> str:
         """Get PURL for this dependency."""
-        purl = PackageURL(type="gem", name=self.name, version=self.version)
+        qualifiers: dict[str, str] | None = None
+        normalized_source: str = self.source.rstrip("/")
```
eskultety: nitpick: I probably wouldn't call this normalization, because normalization would need to conform to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), more specifically section https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2 and onwards.
Looking at the rest of the backends, we're heavily inconsistent in source comparisons, so proper URL normalization is out of scope of this PR and should be filed as follow-up work. So, let's keep this as is.

eskultety (commented):
LGTM. Will approve once the minor comments are fixed.

hermeto/core/package_managers/bundler/gem_models.py
eskultety: Resolving.

tnevrlka (approved):
ACK

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -19,6 +19,9 @@
 from hermeto.core.rooted_path import PathOutsideRoot, RootedPath
 from hermeto.core.scm import GitRepo
 
+RUBYGEMS_URL = "https://rubygems.org"
+RUBYGEMS_HTTP_URL = "http://rubygems.org"
+
```
eskultety: Why can't these be a tuple directly?

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -19,6 +19,9 @@
 from hermeto.core.rooted_path import PathOutsideRoot, RootedPath
 from hermeto.core.scm import GitRepo
 
+RUBYGEMS_URL = "https://rubygems.org"
```
eskultety: Commit message:
Oh, I just noticed 

> Signed-off-by: Nikita Nikita <nikita003002@gmail.com>

Is your real name Nikita Nikita or was this just an attempt to pass the DCO check? In any case please disclose your real name, ^this seems quite artificial and doesn't comply with https://github.com/hermetoproject/hermeto/blob/main/CONTRIBUTING.md#pull-request-guidelines, thank you

hermeto/core/package_managers/bundler/gem_models.py
```diff
@@ -81,7 +84,12 @@ class GemDependency(_GemMetadata):
     @cached_property
     def purl(self) -> str:
         """Get PURL for this dependency."""
-        purl = PackageURL(type="gem", name=self.name, version=self.version)
+        qualifiers: dict[str, str] | None = None
+        normalized_source: str = self.source.rstrip("/")
+        if normalized_source not in (RUBYGEMS_URL, RUBYGEMS_HTTP_URL):
+            qualifiers = {"repository_url": normalized_source}
```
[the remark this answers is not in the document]
eskultety: Well, 

> When a Gemfile.lock specifies gems from a source other than rubygems.org

in most cases ^this will imply that the credentials have already been leaked by the time they'd appear in the SBOM. :) It's also quite unlikely to happen. That said - it's a valid finding from coderabbit and we should `urlparse` the source URL and strip basic auth just in case.

hermeto/core/package_managers/bundler/gem_models.py
Nikita-003002: My actual real name is just Nikita, I don't have a last name even on papers. I formatted it as 'Nikita Nikita' because the DCO check requires two words. What is the preferred format for single-name contributors in this situation?

hermeto/core/package_managers/bundler/gem_models.py
Nikita-003002: Done. I've consolidated these into a single tuple with the proper type annotations.

hermeto/core/package_managers/bundler/gem_models.py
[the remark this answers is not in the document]
Nikita-003002: I've added the urlparse logic to completely strip any basic auth credentials before the URL is passed to the PURL qualifiers.

hermeto/core/package_managers/bundler/gem_models.py
eskultety: >  I formatted it as 'Nikita Nikita' because the DCO check requires two words

It shouldn't. The DCO check is verifying the signoff against the git author field, so that's what was probably happening if it failed.

> My actual real name is just Nikita, I don't have a last name even on papers.

If that is really the case and you're from one of the mononym regions then I'd advise updating your git config author field locally so that you can use just your first name in the DCO signoff.

eskultety (approved):

PR #1717, opened 2026-08-11
[tests] [integration] Migrate test scenarios from the integration test repo to the main git repo vol.2

### LINKED ISSUE #806: Host test Dockerfiles directly with the test cases
- Move each `Containerfile` from `tests/integration/test_data/<test-name>/container/` to the corresponding test branch in https://github.com/cachito-testing/integration-tests
- Update utils for integration tests to use `Containerfile` from the repository for building the image

eskultety:
This is a continuation of https://github.com/hermetoproject/hermeto/pull/1669 which migrated everything except for gomod tests.
This series migrates the gomod backend tests. The main logical change here is adding git submodule support to the synthetic repo creation logic. The reason for that is because the production gomod logic makes heavy use of git submodule metadata to derive version tag info for SBOM and so we need to make sure we can seamlessly test with it.

Note for review purposes this PR only focuses on the gomod migration. The branch-scenario plumbing in the infra logic will be removed in vol.3 of changes where some newly introduced cachito-testing references will be obliterated (as well as 2 gomod packages referenced by release tags hosted in the the same testing organization).

As with the previous volume of changes, every commit must pass the integration test suite (try for yourself).


tests/integration/utils.py
```diff
@@ -40,6 +40,42 @@
...
+        "GIT_AUTHOR_EMAIL": "test@example.com",
+        "GIT_COMMITTER_NAME": "Test Author",
+        "GIT_COMMITTER_EMAIL": "test@example.com",
+        "GIT_AUTHOR_DATE": "1970-01-01T00:00:00+00:00",
+        "GIT_COMMITTER_DATE": "1970-01-01T00:00:00+00:00",
+    }
+
+    def __init__(self, repo_path: Path, origin_url: str) -> None:
+        """Create a temporary synthetic git repository to host our test scenarios."""
+        # copy hermeto's root .gitignore into the synthetic repo
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/ffda8df2cc9913309c9c0a7d148bbd7d638083be:
 
Nitpick, non-blocking: this comment tells what the code does, not why it does what it does, so could be either removed or amended.

a-ovchinnikov (approved):

tests/integration/utils.py
eskultety: Updated.

tnevrlka (approved):

PR #1724, opened 2026-08-13
Remove dead code, fix subtle bugs, and clean up stale leftovers

### LINKED ISSUE #1723: Dead code and stale leftovers across the codebase
The codebase has accumulated dead production code and stale patterns
that could use a cleanup pass.

  - dead code with zero callers (functions, parameters, fields, type aliases)
  - stale patterns from older versions of dependencies
  - subtle bugs hiding in error handling and type annotations
  - redundant guards and expressions

Similar to https://github.com/hermetoproject/hermeto/issues/1506 and https://github.com/hermetoproject/hermeto/issues/1580 but for production code rather than tests.

derasdf:
Remove dead production code and fix subtle bugs found during a codebase sweep.

- Fix pydantic v1 leftovers, auth validation, missing try/finally guards,
  dropped exception chains, wrong type annotations, and other silent bugs
- Remove unused functions, parameters, fields, type aliases, and constants
- Remove redundant guards that became dead after later refactors
- Clean up unused loggers, redundant expressions, and stale test infrastructure

Fixes come first (commits 1-3), cleanup after (commits 4-9).

Closes https://github.com/hermetoproject/hermeto/issues/1723

### DIFF
--- a/hermeto/core/checksum.py
+++ b/hermeto/core/checksum.py
@@ -23,18 +23,6 @@ class ChecksumInfo(NamedTuple):
     algorithm: str
     hexdigest: str
 
-    def to_sri(self) -> str:
-        """Return the Subresource Integrity representation of this ChecksumInfo.
-
-        https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
-
-        Note: npm and Yarn classic (v1) use this format in their lockfiles; newer Yarn
-        versions use a different integrity representation that does not go through this helper.
-        """
-        bytes_sha = bytes.fromhex(self.hexdigest)
-        base64_sha = base64.b64encode(bytes_sha).decode("utf-8")
-        return f"{self.algorithm}-{base64_sha}"
-
     def __str__(self) -> str:
         return f"{self.algorithm}:{self.hexdigest}"
 
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -16,7 +16,7 @@ from hermeto.core.rooted_path import PathOutsideRoot, RootedPath
 BINARY_FILTER_ALL = ":all:"
 
 if TYPE_CHECKING:
-    from pydantic.error_wrappers import ErrorDict
+    from pydantic_core import ErrorDetails
 
 
 log = logging.getLogger(__name__)
@@ -77,7 +77,7 @@ def _present_user_input_error(validation_error: pydantic.ValidationError) -> str
     errors = validation_error.errors()
     n_errors = len(errors)
 
-    def show_error(error: "ErrorDict") -> str:
+    def show_error(error: "ErrorDetails") -> str:
         location = " -> ".join(map(str, error["loc"]))
         if error.get("type") != "union_tag_invalid":
             message = error["msg"]
--- a/hermeto/core/models/input.py
+++ b/hermeto/core/models/input.py
@@ -91,7 +91,7 @@ def _present_user_input_error(validation_error: pydantic.ValidationError) -> str
             quoted = ", ".join(f"'{t}'" for t in sorted(expected))
             message = f"Requested backend type '{ctx.get('tag', '<unknown>')}' doesn't match expected ones: {quoted}"
 
-        if location != "__root__":
+        if location:
             message = f"{location}\n  {message}"
 
         return message
@@ -153,10 +153,8 @@ class SSLOptions(pydantic.BaseModel, extra="forbid"):
 
         if not Path(val).is_file():
             raise ValueError(
-                (
-                    f"Specified ssl auth file '{info.field_name}':'{val}' is not a regular file.",
-                    "Make sure the file exists and that it has correct permissions.",
-                )
+                f"Specified ssl auth file '{info.field_name}':'{val}' is not a regular file."
+                " Make sure the file exists and that it has correct permissions."
             )
 
         return val
--- a/hermeto/core/models/sbom.py
+++ b/hermeto/core/models/sbom.py
@@ -162,7 +162,7 @@ class Pedigree(pydantic.BaseModel):
     patches: list[Patch]
 
 
-FOUND_BY_APP_PROPERTY: Property = Property(name=PropertyEnum.PROP_FOUND_BY, value=f"{APP_NAME}")
+FOUND_BY_APP_PROPERTY = Property(name=PropertyEnum.PROP_FOUND_BY, value=APP_NAME)
 
 
 class Component(pydantic.BaseModel):
@@ -226,7 +226,7 @@ class Tool(pydantic.BaseModel):
 class Metadata(pydantic.BaseModel):
     """Metadata field in a SBOM."""
 
-    tools: list[Tool] = [Tool(vendor="red hat", name=f"{APP_NAME}")]
+    tools: list[Tool] = [Tool(vendor="red hat", name=APP_NAME)]
 
 
 def spdx_now() -> datetime:
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -532,8 +532,10 @@ def _sanitize_cargo_config(config_content: str) -> str:
 def _temporary_cwd(path_to_new_cwd: Path) -> Generator[None, None, None]:
     oldcwd = os.getcwd()
     os.chdir(path_to_new_cwd)
-    yield
-    os.chdir(oldcwd)
+    try:
+        yield
+    finally:
+        os.chdir(oldcwd)
 
 
 def _run_cmd_watching_out_for_lock_mismatch(
--- a/hermeto/core/package_managers/cargo/main.py
+++ b/hermeto/core/package_managers/cargo/main.py
@@ -442,8 +442,7 @@ def _sanitized_cargo_config_file(package_dir: RootedPath) -> Generator[None, Non
         yield
     finally:
         for config, data in configs_contents:
-            if data is not None:
-                config.path.write_text(data)
+            config.path.write_text(data)
 
 
 def _make_basic_token_from_proxy_credential(cargo_config: CargoSettings) -> str:
@@ -573,8 +572,7 @@ def _run_cmd_watching_out_for_lock_mismatch(
                 log.warning(warn_about_imminent_update_to_cargo_lock)
                 with _temporary_cwd(package_dir):
                     # Extract env from params if present to pass to cargo generate-lockfile
-                    env = params.get("env", {})
-                    update_cmd_params = {"env": env} if env else {}
+                    update_cmd_params = {"env": params.get("env", {})}
                     run_cmd(cmd=update_cargo_lock_cmd, params=update_cmd_params)
                 # If it fails here then something else is horribly broken.
                 # No more attempts to salvage the situation will be made.
--- a/hermeto/core/package_managers/generic/models.py
+++ b/hermeto/core/package_managers/generic/models.py
@@ -121,14 +121,11 @@ class LockfileArtifactAuth(BaseModel):
     bearer: BearerAuth | None = None
     model_config = ConfigDict(extra="forbid")
 
-    @model_validator(mode="before")
-    @classmethod
-    def _check_mutually_exclusive(cls, values: dict) -> dict:
-        if ("basic" not in values and "bearer" not in values) or (
-            "basic" in values and "bearer" in values
-        ):
+    @model_validator(mode="after")
+    def _check_mutually_exclusive(self) -> "LockfileArtifactAuth":
+        if (self.basic is None) == (self.bearer is None):
             raise ValueError("Exactly one of the auth types must be set")
-        return values
+        return self
 
     def get_headers(self) -> dict[str, str]:
         """Return the headers for the artifact."""
--- a/hermeto/core/package_managers/gomod/go.py
+++ b/hermeto/core/package_managers/gomod/go.py
@@ -7,7 +7,7 @@ import shutil
 import subprocess
 import tempfile
 from collections import UserDict
-from collections.abc import Iterable
+from collections.abc import Collection
 from functools import cache, cached_property, total_ordering
 from pathlib import Path
 from typing import Any, Sequence
@@ -202,7 +202,7 @@ class Go:
         # references below:
         # [1] https://github.com/golang/go/issues/26520
         # [2] https://golang.org/cl/34385
-        with tempfile.TemporaryDirectory(prefix=f"{APP_NAME}", suffix="go-download") as td:
+        with tempfile.TemporaryDirectory(prefix=APP_NAME, suffix="go-download") as td:
             log.debug("Installing Go %s toolchain shim from '%s'", release, url)
             env = {
                 "PATH": os.environ.get("PATH", ""),
@@ -271,6 +271,8 @@ class Go:
                 f"Go execution failed: {APP_NAME} re-tried running `{' '.join(cmd)}` command "
                 f"{n_tries} times."
             )
+            # run_cmd already logs stderr; the caught error only summarizes it,
+            # so `from None` avoids a redundant chained duplicate.
             raise PackageManagerError(err_msg) from None
 
     @staticmethod
@@ -396,12 +398,12 @@ def _get_gomod_version(go_mod_file: RootedPath) -> tuple[str | None, str | None]
     return (go_version, toolchain_version)
 
 
-def _select_toolchain(go_mod_file: RootedPath, installed_toolchains: Iterable[Go]) -> Go | None:
+def _select_toolchain(go_mod_file: RootedPath, installed_toolchains: Collection[Go]) -> Go | None:
     """
     Pick the closest matching installed toolchain give a go.mod file.
 
     :param go_mod_file: path to an application go.mod file as RootedPath
-    :param installed_toolchains: an iterable of Go instances pointing to actual Go binaries
+    :param installed_toolchains: a collection of Go instances pointing to actual Go binaries
     :return: a Go instance which matches the go.mod version constraints or None if we could not find
              a satisfying toolchain version installed
     """
@@ -468,7 +470,12 @@ def _select_toolchain(go_mod_file: RootedPath, installed_toolchains: Iterable[Go
                 work_toolchain = first(installed_toolchains)
                 go = Go.from_missing_toolchain(release_str, work_toolchain.binary)
                 log.debug("Using Go toolchain version '%s'", go.version)
-            except Exception as ex:
+            except (
+                PackageManagerError,
+                OSError,
+                subprocess.CalledProcessError,
+                subprocess.TimeoutExpired,
+            ) as ex:
                 log.error("Failed to download a Go toolchain version '%s': '%s'", release_str, ex)
                 return None
     return go
@@ -507,9 +514,7 @@ def _list_installed_toolchains() -> set[Go]:
         try:
             log.debug("Probing %s toolchain...", bin_path)
             ret.add(Go(binary=bin_path.as_posix()))
-        except Exception as e:
-            # Logging toolchain probing failures due to [1].
-            # [1] https://bandit.readthedocs.io/en/1.8.3/plugins/b112_try_except_continue.html
+        except (PackageManagerError, OSError, subprocess.CalledProcessError) as e:
             log.debug("Toolchain %s failed probing: %s, skipping...", bin_path, e)
 
     log.debug(
--- a/hermeto/core/package_managers/gomod/main.py
+++ b/hermeto/core/package_managers/gomod/main.py
@@ -78,8 +78,6 @@ class ParsedOrigin(_ParsedModel):
     vcs: str
     url: str
     hash: str
-    tag_sum: str | None = None
-    ref: str | None = None
 
 
 class ParsedModule(_ParsedModel):
@@ -387,7 +385,7 @@ def _create_packages_from_parsed_data(
 
         relative_path = _resolve_package_relative_path(package, module)
 
-        return Package(relative_path=str(relative_path), module=module)
+        return Package(relative_path=relative_path, module=module)
 
     def _find_parent_module_by_name(package: ParsedPackage) -> Module:
         """Return the longest module name that is contained in package's import_path."""
@@ -682,7 +680,7 @@ def _disable_telemetry(go: Go, run_params: dict[str, Any]) -> None:
 
 
 def _go_list_deps(
-    go: Go, pattern: Literal["./...", "all"], run_params: dict[str, Any] | None = None
+    go: Go, pattern: Literal["./...", "all"], run_params: dict[str, Any]
 ) -> Iterator[ParsedPackage]:
     """Run go list -deps -json and return the parsed list of packages.
 
@@ -691,7 +689,7 @@ def _go_list_deps(
     The "all" pattern includes dependencies needed only for tests. Use it to get a more
     complete module list (roughly matching the list of downloaded modules).
     """
-    cmd = ["list", "-e", "-deps", "-json=ImportPath,Module,Standard,Deps", pattern]
+    cmd = ["list", "-e", "-deps", "-json=ImportPath,Module,Standard", pattern]
     return map(
         ParsedPackage.model_validate,
         load_json_stream(go(cmd, run_params)),
@@ -1360,14 +1358,10 @@ class ModuleVersionResolver:
 
 
 def _validate_local_replacements(modules: Iterable[ParsedModule], app_path: RootedPath) -> None:
-    replaced_paths = [
-        (module.path, module.replace.path)
-        for module in modules
-        if module.replace and module.replace.path.startswith(".")
-    ]
-
-    for _, path in replaced_paths:
-        app_path.join_within_root(path)
+    """Check that local replacements point to paths within app_path."""
+    for module in modules:
+        if module.replace and module.replace.path.startswith("."):
+            app_path.join_within_root(module.replace.path)
 
 
 def _parse_vendor(context_dir: RootedPath) -> Iterable[ParsedModule]:
--- a/hermeto/core/package_managers/gomod/main.py
+++ b/hermeto/core/package_managers/gomod/main.py
@@ -1164,7 +1164,9 @@ class ModuleVersionResolver:
         module_major_version = int(match.group("major_version")) if match else None
 
         # If no match, prefer v1.x.x tags but fallback to v0.x.x tags if both are present
-        major_versions_to_try = (module_major_version,) if module_major_version else (1, 0)
+        major_versions_to_try = (
+            (module_major_version,) if module_major_version is not None else (1, 0)
+        )
 
         if app_dir.path == app_dir.root:
             subpath = None
@@ -1314,7 +1316,8 @@ class ModuleVersionResolver:
         if tag is None:
             # If the major version isn't in the import path and there is not a versioned commit with the
             # version of 1, the major version defaults to 0.
-            return f"v{module_major_version or '0'}.0.0-{commit_timestamp}-{commit_hash}"
+            major = module_major_version if module_major_version is not None else 0
+            return f"v{major}.0.0-{commit_timestamp}-{commit_hash}"
 
         tag_semantic_version = self._get_semantic_version_from_tag(tag.name, subpath)
 
@@ -1514,12 +1517,19 @@ def prepare_netrc_contents() -> str:
         password {secret}"""
 
 
+def _write_file_with_mode(path: Path, content: str, mode: int) -> None:
+    """Write content to a file with explicit permissions, bypassing umask."""
+    old_mask = os.umask(0)
+    try:
+        fd = os.open(path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, mode)
+        with open(fd, "w") as f:
+            f.write(content)
+    finally:
+        os.umask(old_mask)
+
+
 def inject_netrc(netrc_stuff: str, temp_netrc_dir: Path) -> str:
     """Inject a temporary .netrc."""
     netrc = temp_netrc_dir / ".netrc"
-    old_mask = os.umask(0)
-    netrc_fd = os.open(path=netrc, flags=(os.O_WRONLY | os.O_CREAT | os.O_TRUNC), mode=0o600)
-    with open(netrc_fd, "w") as f:
-        f.write(netrc_stuff)
-    os.umask(old_mask)
+    _write_file_with_mode(netrc, netrc_stuff, 0o600)
     return str(netrc)
--- a/hermeto/core/package_managers/javascript/npm/resolver.py
+++ b/hermeto/core/package_managers/javascript/npm/resolver.py
@@ -96,9 +96,9 @@ def _get_npm_dependencies(
         dep_type = classify_resolved_url(url)
         proxy_auth = None
 
-        if dep_type == "file":
-            continue
-        elif dep_type == "git":
+        # dep_type is "git", "registry", or "https". "file" deps are filtered
+        # upstream by get_dependencies_to_download() and never reach here.
+        if dep_type == "git":
             download_paths[url] = clone_repo_pack_archive(parse_git_clone_url(url), download_dir)
         else:
             if dep_type == "registry":
--- a/hermeto/core/package_managers/python/pip/package_distributions.py
+++ b/hermeto/core/package_managers/python/pip/package_distributions.py
@@ -212,9 +212,6 @@ def process_package_distributions(
     filtered_wheels: list[DistributionPackageInfo] = []
 
     for package in to_process:
-        if package.package_type is None or package.package_type not in allowed_distros:
-            continue
-
         pypi_checksums: set[ChecksumInfo] = {
             ChecksumInfo(algorithm, digest) for algorithm, digest in package.digests.items()
         }
@@ -367,11 +364,14 @@ class WheelsFilter(BinaryPackageFilter):
         return self.abi is None or abi in ("abi3", "none") or any(a in abi for a in self.abi)
 
     def _compatible_tag_interpreter(self, interpreter: str, abi: str) -> bool:
-        compatible_py_impl = (
+        is_compatible_py_impl = (
             self.py_impl is None
             or "py3" in interpreter
             or any(impl in interpreter for impl in self.py_impl)
         )
+
+        if self.py_version is None:
+            return is_compatible_py_impl
 
         wheel_py_versions = _parse_py_versions(interpreter)
 
@@ -390,9 +390,7 @@ class WheelsFilter(BinaryPackageFilter):
             is_compatible_candidate(v, this) for v in wheel_py_versions
         )
 
-        compatible_py_version = self.py_version is None or some_candidate_is_compatible
-
-        return compatible_py_impl and compatible_py_version
+        return is_compatible_py_impl and some_candidate_is_compatible
 
 
 def _parse_py_versions(interpreter: str) -> list[int]:
--- a/hermeto/core/package_managers/rpm/main.py
+++ b/hermeto/core/package_managers/rpm/main.py
@@ -400,7 +400,7 @@ def _verify_downloaded(metadata: dict[Path, Any]) -> None:
 
         # checksum is optional
         if file_metadata["checksum"] is not None:
-            alg, digest = file_metadata["checksum"].split(":")
+            alg, digest = file_metadata["checksum"].split(":", 1)
             method = getattr(hashlib, alg.lower(), None)
             if method is not None:
                 h = method(usedforsecurity=False)
@@ -431,8 +431,8 @@ def _generate_sbom_components(
             continue
         package = Package.from_filepath(file_path, file_metadata)
         component = package.to_component(lockfile_path)
-        if include_summary_in_sbom:
-            summary = Property(name=f"{APP_NAME}:rpm_summary", value=str(package.summary))
+        if include_summary_in_sbom and package.summary:
+            summary = Property(name=f"{APP_NAME}:rpm_summary", value=package.summary)
             component.properties.append(summary)
         components.append(component)
     return components
--- a/hermeto/core/package_managers/rpm/main.py
+++ b/hermeto/core/package_managers/rpm/main.py
@@ -70,9 +70,7 @@ class Package:
             # Red Hat PURL RPM guideline suggests injecting 'src' into the arch qualifier for SRPMS
             kwargs["arch"] = "src"
 
-        kwargs["repository_id"] = (
-            repoid if repoid and not repoid.startswith(f"{APP_NAME}") else None
-        )
+        kwargs["repository_id"] = repoid if repoid and not repoid.startswith(APP_NAME) else None
         kwargs["download_url"] = rpm_download_metadata["url"]
         kwargs["checksum"] = rpm_download_metadata.get("checksum")
 
@@ -169,7 +167,7 @@ class Package:
 
         if self.modularity_label:
             properties.append(
-                Property(name=f"{APP_NAME}:rpm_modularity_label", value=str(self.modularity_label))
+                Property(name=f"{APP_NAME}:rpm_modularity_label", value=self.modularity_label)
             )
 
         return Component(
@@ -481,11 +479,9 @@ def _generate_repofiles(
     in the project file.
     """
     dnf_options = None
-    dnf_options_repos = None
 
     if options:
         dnf_options = options.get("rpm", {}).get("dnf", {})
-        dnf_options_repos = dnf_options.keys()
 
     package_dir = from_output_dir.joinpath(DEFAULT_PACKAGE_DIR)
     for arch in package_dir.iterdir():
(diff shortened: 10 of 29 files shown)


hermeto/core/package_managers/generic/models.py
```diff
@@ -124,9 +124,9 @@ class LockfileArtifactAuth(BaseModel):
     @model_validator(mode="before")
     @classmethod
     def _check_mutually_exclusive(cls, values: dict) -> dict:
-        if ("basic" not in values and "bearer" not in values) or (
-            "basic" in values and "bearer" in values
-        ):
+        has_basic = values.get("basic") is not None
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/0ec593d7011e757fa456edba6b70387699dd37f3:

This one is neat, however I wonder how one could actually run into it?

hermeto/core/models/input.py
a-ovchinnikov, line 1: https://github.com/hermetoproject/hermeto/commit/6ab03ae741aaef03efb9628309cf99595a253a68:

> making the guard always True and inserting a
spurious blank line in error output

I am curious to see an example of an input that triggers that.

hermeto/core/package_managers/generic/models.py
```diff
@@ -124,9 +124,9 @@ class LockfileArtifactAuth(BaseModel):
...
+        has_bearer = values.get("bearer") is not None
+        if has_basic == has_bearer:
```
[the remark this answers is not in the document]
a-ovchinnikov: Dear coderabbit, this is counterproductive.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -530,8 +530,10 @@ def _sanitize_cargo_config(config_content: str) -> str:
 def _temporary_cwd(path_to_new_cwd: Path) -> Generator[None, None, None]:
     oldcwd = os.getcwd()
     os.chdir(path_to_new_cwd)
-    yield
-    os.chdir(oldcwd)
+    try:
+        yield
+    finally:
+        os.chdir(oldcwd)
```
a-ovchinnikov, line 537: https://github.com/hermetoproject/hermeto/commit/f3ee56f464110ab9e97b006266b8e8a6295a9d5f:

Is any of exceptions raised during yield recoverable? It looks like not to me, so I don't see the value of setting cwd when unwinding the stack, but I could be missing something here.

hermeto/core/package_managers/gomod/go.py
```diff
@@ -265,12 +265,12 @@ def run_go(_cmd: list[str], **kwargs: Any) -> str:
...
         try:
             return run_go(cmd, **kwargs)
-        except PackageManagerError:
+        except PackageManagerError as e:
             err_msg = (
                 f"Go execution failed: {APP_NAME} re-tried running `{' '.join(cmd)}` command "
                 f"{n_tries} times."
             )
-            raise PackageManagerError(err_msg) from None
+            raise PackageManagerError(err_msg) from e
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/f3ee56f464110ab9e97b006266b8e8a6295a9d5f:

From the looks of it this could have been done on purpose, it's a pity there is no comment explaining author's intent here.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1071,7 +1071,7 @@ def from_repo_path(cls, repo_path: RootedPath) -> "Self":
                 f"Failed to fetch the tags on the Git repository ({type(ex).__name__}) "
                 f"for {repo.working_tree_dir}: "
                 f"{str(ex)}"
-            )
+            ) from ex
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/f3ee56f464110ab9e97b006266b8e8a6295a9d5f:

I am having trouble understanding the benefit of raising from here, isn't the FetchError trapped downstream anyway? _Edit_: the same question applies to all additions of `from ex`.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1314,7 +1316,7 @@ def _get_golang_pseudo_version(
         if tag is None:
             # If the major version isn't in the import path and there is not a versioned commit with the
             # version of 1, the major version defaults to 0.
-            return f"v{module_major_version or '0'}.0.0-{commit_timestamp}-{commit_hash}"
+            return f"v{module_major_version if module_major_version is not None else '0'}.0.0-{commit_timestamp}-{commit_hash}"
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/f3ee56f464110ab9e97b006266b8e8a6295a9d5f:

Please lift `module_major_version` computation up from this string. Also why `or` is bad here?

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1516,8 +1518,10 @@ def inject_netrc(netrc_stuff: str, temp_netrc_dir: Path) -> str:
...
-    netrc_fd = os.open(path=netrc, flags=(os.O_WRONLY | os.O_CREAT | os.O_TRUNC), mode=0o600)
-    with open(netrc_fd, "w") as f:
-        f.write(netrc_stuff)
-    os.umask(old_mask)
+    try:
+        netrc_fd = os.open(path=netrc, flags=(os.O_WRONLY | os.O_CREAT | os.O_TRUNC), mode=0o600)
+        with open(netrc_fd, "w") as f:
+            f.write(netrc_stuff)
+    finally:
+        os.umask(old_mask)
```
a-ovchinnikov, line 1522: https://github.com/hermetoproject/hermeto/commit/f3ee56f464110ab9e97b006266b8e8a6295a9d5f:

If you are reworking this could you please make it into a porper context manager then? No use in exposing low-level details in high-level code. _Edit_: typo

hermeto/core/package_managers/rpm/main.py
```diff
@@ -400,7 +400,7 @@ def raise_exception(filename: Path, message: str) -> None:
 
         # checksum is optional
         if file_metadata["checksum"] is not None:
-            alg, digest = file_metadata["checksum"].split(":")
+            alg, digest = file_metadata["checksum"].split(":", 1)
```
a-ovchinnikov, line 401: https://github.com/hermetoproject/hermeto/commit/f3ee56f464110ab9e97b006266b8e8a6295a9d5f:

Can this actually happen? We are not ingesting random fields from the internet after all, the set of files we work with is much narrower. This is also an indicator that wee might want to add a proper model for that and let _the model_ take care about validation.

hermeto/core/package_managers/rpm/main.py
```diff
@@ -431,8 +431,8 @@ def _generate_sbom_components(
             continue
         package = Package.from_filepath(file_path, file_metadata)
         component = package.to_component(lockfile_path)
-        if include_summary_in_sbom:
-            summary = Property(name=f"{APP_NAME}:rpm_summary", value=str(package.summary))
+        if include_summary_in_sbom and package.summary:
```
a-ovchinnikov, line 432: https://github.com/hermetoproject/hermeto/commit/f3ee56f464110ab9e97b006266b8e8a6295a9d5f:

Is it bad to be explicit about stating that a package has no summary?

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1359,14 +1357,9 @@ def _get_semantic_version_from_tag(
...
 def _validate_local_replacements(modules: Iterable[ParsedModule], app_path: RootedPath) -> None:
-    replaced_paths = [
-        (module.path, module.replace.path)
-        for module in modules
-        if module.replace and module.replace.path.startswith(".")
-    ]
-
-    for _, path in replaced_paths:
-        app_path.join_within_root(path)
+    for module in modules:
```
a-ovchinnikov, line 1362: https://github.com/hermetoproject/hermeto/commit/4aea6ca3d2262ffa06f1a1b3ed272a7671ed7f64:

Since you are touching this part could you please add a comment describing validation: something along the lines of "Checks that local replacements are within app_path"?

hermeto/core/package_managers/javascript/npm/resolver.py
```diff
@@ -119,9 +119,7 @@ def _get_npm_dependencies(
         dep_type = classify_resolved_url(url)
         proxy_auth = None
 
-        if dep_type == "file":
-            continue
-        elif dep_type == "git":
+        if dep_type == "git":
```
a-ovchinnikov, line 101: https://github.com/hermetoproject/hermeto/commit/c8a5265a9ab849acbb14d1c0adc4f0fd1c5837c4:

This code is effectively a match on literal values, the way it was written previously it was total, now it is partial. I suggest at least acknowledging the fact that "file" cannot reach this match with a comment.

hermeto/core/package_managers/python/pip/package_distributions.py
```diff
@@ -373,9 +370,11 @@ def _compatible_tag_interpreter(self, interpreter: str, abi: str) -> bool:
             or any(impl in interpreter for impl in self.py_impl)
         )
 
+        if self.py_version is None:
+            return compatible_py_impl
+
         wheel_py_versions = _parse_py_versions(interpreter)
 
-        # we compare strings because e.g. '312' is all we get from parsing wheel interpreter tag
```
a-ovchinnikov, line 378: https://github.com/hermetoproject/hermeto/commit/a640a37736a65819bc362fa30be647aa44fbf6e8:

Why did you remove this comment, is it not true anymore?

hermeto/core/package_managers/python/pip/package_distributions.py
```diff
@@ -373,9 +370,11 @@ def _compatible_tag_interpreter(self, interpreter: str, abi: str) -> bool:
             or any(impl in interpreter for impl in self.py_impl)
         )
 
+        if self.py_version is None:
+            return compatible_py_impl
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/a640a37736a65819bc362fa30be647aa44fbf6e8:

While you are here, could you please s/compatible_py_impl/is_compatible_py_impl/? It reads like a non-boolean the way it is named now.

hermeto/core/resolver.py
```diff
@@ -85,7 +85,4 @@ def _resolve_packages(request: Request) -> RequestOutput:
 
 def inject_files_post(from_output_dir: Path, for_output_dir: Path, **kwargs: Any) -> None:
     """Do extra steps for package manager."""
-    # if there is a callback method defined within the particular package manager, run it
```
a-ovchinnikov, line 88: https://github.com/hermetoproject/hermeto/commit/f3ee56f464110ab9e97b006266b8e8a6295a9d5f:

This looks like something that could have been done for a reason, maybe @eskultety remembers why this was done this way?

hermeto/core/models/sbom.py
```diff
@@ -162,7 +162,7 @@ class Pedigree(pydantic.BaseModel):
     patches: list[Patch]
 
 
-FOUND_BY_APP_PROPERTY: Property = Property(name=PropertyEnum.PROP_FOUND_BY, value=f"{APP_NAME}")
+FOUND_BY_APP_PROPERTY: Property = Property(name=PropertyEnum.PROP_FOUND_BY, value=APP_NAME)
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/550f195b050ca7cb830cf58cabf49ca46ff94072:

Since you are cleaning this up, could you please also drop `: Property` annotation? There is a `Property` call right to the right of `=` sign, it's hard to misinterpret this assignment.

hermeto/core/models/sbom.py
```diff
@@ -645,9 +645,6 @@ class SPDXPackage(pydantic.BaseModel):
     # should be separated with semicolons.
     sourceInfo: str | None = None
 
-    def __lt__(self, other: "SPDXPackage") -> bool:
```
a-ovchinnikov, line 646: https://github.com/hermetoproject/hermeto/commit/550f195b050ca7cb830cf58cabf49ca46ff94072:

How did you verify that less than was never invoked?

hermeto/core/checksum.py
```diff
@@ -14,7 +14,7 @@
 log = logging.getLogger(__name__)
 
 
-SUPPORTED_ALGORITHMS = hashlib.algorithms_guaranteed
+SUPPORTED_ALGORITHMS = hashlib.algorithms_guaranteed - {"shake_128", "shake_256"}
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/550f195b050ca7cb830cf58cabf49ca46ff94072:

This does not sound right, if it requires additional arguments then we should provide those arguments, not drop algorithms.

hermeto/core/utils.py
```diff
@@ -152,7 +151,7 @@ def _fast_copy(src: Path, dest: Path) -> int:
             ex.filename2 = fdest.name
 
             if ex.errno == errno.ENOSYS or ex.errno == errno.EXDEV:
-                raise _FastCopyFailedFallback
+                raise _FastCopyFailedFallback()
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/550f195b050ca7cb830cf58cabf49ca46ff94072:

In the commit message you say that you are fixing the raise, could you please elaborate what exactly is being fixed? [[link](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement)]

hermeto/core/utils.py
```diff
@@ -206,15 +205,9 @@ def get_cache_dir() -> Path:
...
-    return cache_dir.joinpath(f"{APP_NAME}")
+    return cache_dir.joinpath(APP_NAME)
 
 
 def first_for(predicate: Callable, iterable: Iterable, fallback: Any) -> Any:
     """Return the first match of predicate in iterable or fallback value."""
     return next((x for x in iterable if predicate(x)), fallback)
-
-
-def partition_by(predicate: Callable, iterable: Iterable) -> tuple[Iterable, Iterable]:
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/550f195b050ca7cb830cf58cabf49ca46ff94072:

I suggest not touching utilities, at least until Hermeto embraces `more_itertools`. The fact that the function is not used at the moment is probably the reason to refactor places which would benefit from it rather than dropping it.

hermeto/interface/logging.py
```diff
@@ -94,15 +93,13 @@ def error_or_warn(self, msg: str, *args: Any, enforcing_mode: Mode, **kwargs: An
 def setup_logging(
     level: LogLevel,
     color: bool | None = None,
-    additional_modules: Iterable[str] = (),
```
a-ovchinnikov, line 97: https://github.com/hermetoproject/hermeto/commit/550f195b050ca7cb830cf58cabf49ca46ff94072:

I suggest not touching logging unless it is obviously misconfigured.

hermeto/core/package_managers/bundler/main.py
```diff
@@ -30,8 +29,6 @@
 from hermeto.core.rooted_path import RootedPath
 from hermeto.core.scm import get_repo_id
 
-log = logging.getLogger(__name__)
```
a-ovchinnikov, line 34: https://github.com/hermetoproject/hermeto/commit/525c26375b545b61c34f49c5ac2d17454dbcd3a0:

Let's keep loggers for now (https://github.com/hermetoproject/hermeto/issues/1699). Logging should be made more uniform anyway.

hermeto/core/package_managers/generic/models.py
derasdf: YAML `basic: null`, yaml.safe_load turns null into None, pydantic before-validator sees `{"basic": None}`. Unlikely but the logic was wrong.

hermeto/core/models/input.py
derasdf: `SSLOptions(client_cert="/etc/hostname")` without client_key. model_validator raises with `loc=()`, location becomes `""`, the old guard is True, output gets a blank line + indent.

hermeto/core/package_managers/cargo/main.py
derasdf: No, but os.chdir is global state. The adjacent _sanitized_cargo_config_file uses try/finally for the same pattern.

hermeto/core/package_managers/gomod/go.py
derasdf: Added

hermeto/core/package_managers/gomod/main.py
derasdf: It is, but the traceback is logged. Without `from ex` you see "Failed to fetch tags." With it you also see what git actually said.

hermeto/core/package_managers/gomod/main.py
derasdf: Done. `or` treats int(0) as falsy. Works by accident here because str(0) == '0'. Not triggerable (Go forbids /v0).

hermeto/core/package_managers/gomod/main.py
derasdf: Done, extracted `_write_file_with_mode` helper.

hermeto/core/package_managers/rpm/main.py
derasdf: Probably not. Consistency with the generic backend which uses split(":", 1). Can drop.

hermeto/core/package_managers/rpm/main.py
derasdf: str(None) writes literal "None" into the SBOM, not an absent value.

hermeto/core/package_managers/gomod/main.py
derasdf: Done.

hermeto/core/package_managers/javascript/npm/resolver.py
derasdf: Done.

hermeto/core/package_managers/python/pip/package_distributions.py
derasdf: Unintentional, restored.

hermeto/core/package_managers/python/pip/package_distributions.py
derasdf: Done.

hermeto/core/resolver.py
derasdf: From the RPM PoC PR (#502): the hasattr/getattr was written as a soft dependency so other backends could add their own `inject_files_post` without changing resolver.py. That extension never happened, rpm is still the only one, and the module is imported unconditionally. But if the intent is to keep the extension point open, can revert.

hermeto/core/models/sbom.py
derasdf: Done.

hermeto/core/models/sbom.py
derasdf: Grepped for __lt__, sorted(), .sort(). All sorted() calls use explicit key=.

hermeto/core/checksum.py
derasdf: SHAKE needs a length param that no other algorithm needs. No lockfile format uses it. Can drop this change if you prefer.

hermeto/core/utils.py
derasdf: Nothing behavioral. Consistency with the 3 other raises in the same function. Can drop.

hermeto/core/utils.py
derasdf: Reverted.

hermeto/interface/logging.py
derasdf: Reverted.

hermeto/core/package_managers/bundler/main.py
derasdf: Reverted.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -569,8 +570,7 @@ def _run_cmd_watching_out_for_lock_mismatch(
                 log.warning(warn_about_imminent_update_to_cargo_lock)
                 with _temporary_cwd(package_dir):
                     # Extract env from params if present to pass to cargo generate-lockfile
-                    env = params.get("env", {})
-                    update_cmd_params = {"env": env} if env else {}
+                    update_cmd_params = {"env": params.get("env", {})}
```
derasdf, line 575: Intentional.

hermeto/core/utils.py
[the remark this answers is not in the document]
derasdf: is proper fallback

a-ovchinnikov: In this particular case this is no-op, I suggest dropping it.

hermeto/core/package_managers/cargo/main.py
a-ovchinnikov: My point is: if an exception is not recoverable (i.e. there is no code that catches and processes it), then it does not really matter whether there is a `finally` clause or not, an interpreter is going to exit any way. A better way of justifying a `finally` clause would be to mention that decorators could be stacked and in _that_ case one would like to roll back cwd changes.

hermeto/core/package_managers/rpm/main.py
a-ovchinnikov: I am ok with keeping it, it just looks a bit misplaced.

hermeto/core/package_managers/gomod/main.py
a-ovchinnikov: TY. Yes, `int(0)` is considered `False` and that's precisely the problem with broad or partially fitting types. A variable `go_version` should be of type `GoVersion` which is at least an alias for `conint(ge=1)`.

hermeto/core/package_managers/rpm/main.py
a-ovchinnikov: This looks like a very explicit way of telling a user that there is no summary available, is there any requirement that states that a missing summary must be not present at all?

hermeto/core/package_managers/gomod/go.py
a-ovchinnikov: I am sorry, I was not clear enough, I meant the original `from None` part -- it was likely done for some purpose, so I wondered what that purpose was.

hermeto/core/package_managers/generic/models.py
a-ovchinnikov: Neat, thank you!

hermeto/core/models/input.py
a-ovchinnikov: Thanks.

hermeto/core/checksum.py
a-ovchinnikov: I'd quote myself:

> if it requires additional arguments then we should provide those arguments

How prevalent are those? Judging from the fact that no-one has complained so far probably not very much. If they cannot appear in practice then they should be removed with a comment explaining why they are removed. If they can then they must be handled properly.

tests/unit/package_managers/test_generic.py
```diff
@@ -582,6 +582,18 @@ def test_raise_error_when_no_auth_type_set(self) -> None:
         with pytest.raises(ValidationError):
             LockfileArtifactAuth()
 
+    @pytest.mark.parametrize(
+        "kwargs",
+        [
+            pytest.param({"basic": None}, id="basic_none"),
```
a-ovchinnikov: WDYT: s/basic_none/basic_explicitly_set_to_none/ (`None` is the default value, it is a bit confusing as it is named now)

hermeto/core/package_managers/gomod/go.py
```diff
@@ -265,12 +265,13 @@ def run_go(_cmd: list[str], **kwargs: Any) -> str:
...
         try:
             return run_go(cmd, **kwargs)
-        except PackageManagerError:
+        except PackageManagerError as e:
             err_msg = (
                 f"Go execution failed: {APP_NAME} re-tried running `{' '.join(cmd)}` command "
                 f"{n_tries} times."
             )
-            raise PackageManagerError(err_msg) from None
+            # Preserve the original error chain so subprocess stderr is visible
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/efb3803e62fc0d71afe8b81152dac3552f5c102b:

My original question was about the original explicit `from None`, I believe it was done for a reason.

hermeto/core/scm.py
```diff
@@ -357,4 +357,4 @@ def _reset_git_head(repo: GitRepo, ref: str) -> None:
         raise FetchError(
             "Failed on checking out the Git repository. Please verify the supplied reference "
             f'of "{ref}" is valid.'
-        )
+        ) from ex
```
a-ovchinnikov: https://github.com/hermetoproject/hermeto/commit/efb3803e62fc0d71afe8b81152dac3552f5c102b:

Do I get it right and the idea of addiing `from ex` everywhere is to have underlying error visible in the logs, right?

hermeto/core/models/sbom.py
a-ovchinnikov, line 646: https://github.com/hermetoproject/hermeto/commit/b1bd2fb56a9d033ec57d6414d526e12527d5d93f:

I am not comfortable with deleting `__lt__()`  for `SPDXPackage`, it is a part of type definition.

a-ovchinnikov (commented):
Mostly LGTM with some minor comments inline.

hermeto/core/models/sbom.py
derasdf: Reverted

hermeto/core/scm.py
derasdf: Pretty much. It's mostly for consistency here since `log.exception` already dumps the cause. The place it actually matters is tag fetching, where nothing logs the `GitError`, so `from ex` is the only thing keeping it. Rest of `scm.py` chains the same way anyway.

hermeto/core/package_managers/gomod/go.py
derasdf: Reverted

tests/unit/package_managers/test_generic.py
derasdf: Renamed

hermeto/core/resolver.py
a-ovchinnikov: Well, since there are no objections from anyone else let's remove this mechanism then.

hermeto/core/scm.py
a-ovchinnikov: Why would _a user_ want to see all these details in the logs? Those are an instrument for a user after all, once a user notices that some package does not work well with Hermeto they should report a bug from where we could always recreate their situation.

hermeto/core/resolver.py
eskultety: >> This looks like something that could have been done for a reason, maybe @eskultety remembers why this was done this way?

> Well, since there are no objections from anyone else let's remove this mechanism then.

Coming late to the party, sorry, thanks for the ping!

> From the RPM PoC PR (#502): the hasattr/getattr was written as a soft dependency so other backends could add their own `inject_files_post` without changing resolver.py. That extension never happened, rpm is still the only one, and the module is imported unconditionally. But if the intent is to keep the extension point open, can revert.

Oh, this is fine to do. ^This assumption is correct. The moment we need other backends to run any inject callbacks we'll need to add more plumbing and hence, refactoring. Resolving the thread.

hermeto/core/checksum.py
eskultety: Agreed with @a-ovchinnikov , the change as proposed is suspicious and must be reworked such that it is obvious at first glance why we're doing something, i.e. use another constant for rejected algos in a standalone commit.
btw. Have we ever encountered these algos? I'd might event prefer staying reactive rather than proactive here, IOW let Hermeto fail for someone and let them file an issue. That said, if you rework this accordingly, I'm okay, seems harmless.

hermeto/core/scm.py
derasdf: Hm fair, removed.

hermeto/core/checksum.py
derasdf: Dropped.

tests/unit/models/test_input.py
```diff
@@ -271,7 +271,7 @@ def test_valid_packages(self, input_data: dict[str, Any], expect_data: dict[str,
             pytest.param(
                 {"type": "pip", "requirements_build_files": ["weird/../subpath"]},
                 r"pip.requirements_build_files\n  Value error, path contains ..: weird/../subpath",
-                id="pip_path_references_parent_directory",
+                id="pip_build_path_references_parent_directory",
```
slimreaper35, line 274: A good way of preventing such test misconfiguration is to enable pytest's [strict](https://docs.pytest.org/en/stable/explanation/goodpractices.html#using-pytest-s-strict-mode) mode.

hermeto/core/package_managers/python/pip/package_distributions.py
```diff
@@ -390,9 +390,9 @@ def _compatible_tag_interpreter(self, interpreter: str, abi: str) -> bool:
             is_compatible_candidate(v, this) for v in wheel_py_versions
         )
 
-        compatible_py_version = self.py_version is None or some_candidate_is_compatible
+        compatible_py_version = some_candidate_is_compatible
```
a-ovchinnikov: You don't need this assignment anymore. Without it the return statement would read even better:
```python3
return is_compatible_py_impl and some_candidate_is_compatible
```

hermeto/core/utils.py
```diff
@@ -152,7 +152,7 @@ def _fast_copy(src: Path, dest: Path) -> int:
             ex.filename2 = fdest.name
 
             if ex.errno == errno.ENOSYS or ex.errno == errno.EXDEV:
-                raise _FastCopyFailedFallback
+                raise _FastCopyFailedFallback()
```
a-ovchinnikov: It is still not clear to me why this is necessary, as I have mentioned in a previous review Python allows short-cutting parentheses in this particular case. Won't block on this.

hermeto/core/utils.py
```diff
@@ -206,15 +206,15 @@ def get_cache_dir() -> Path:
...
+    return cache_dir.joinpath(APP_NAME)
 
 
 def partition_by(predicate: Callable, iterable: Iterable) -> tuple[Iterable, Iterable]:
     """Partition iterable in two by predicate."""
     i1, i2 = tee(iterable)
     return filterfalse(predicate, i1), filter(predicate, i2)
+
+
+def first_for(predicate: Callable, iterable: Iterable, fallback: Any) -> Any:
```
a-ovchinnikov: Nitpick: this just moves the code around without a clear reason.

a-ovchinnikov (approved):

tests/unit/models/test_input.py
derasdf: Good suggestion but that's a repo-wide config change though so better do it as a separate PR.

hermeto/core/package_managers/python/pip/package_distributions.py
derasdf: Done

hermeto/core/utils.py
derasdf: Sure, reverted

hermeto/core/utils.py
derasdf: Fixed

tests/unit/models/test_input.py
slimreaper35: I can do it, it should be a simple change.

hermeto/core/package_managers/generic/models.py
taylormadore: If this is a before validator, what happens if auth is not a mapping?

hermeto/core/package_managers/gomod/go.py
taylormadore: nit: docstring still says iterable

hermeto/core/package_managers/gomod/go.py
```diff
@@ -467,7 +469,7 @@ def _select_toolchain(go_mod_file: RootedPath, installed_toolchains: Iterable[Go
                 work_toolchain = next(iter(installed_toolchains))
                 go = Go.from_missing_toolchain(release_str, work_toolchain.binary)
                 log.debug("Using Go toolchain version '%s'", go.version)
-            except Exception as ex:
+            except (PackageManagerError, OSError, subprocess.CalledProcessError) as ex:
```
taylormadore: Isn't TimeoutExpired something else that could be raised?

hermeto/core/package_managers/javascript/npm/resolver.py
```diff
@@ -120,9 +120,9 @@ def _get_npm_dependencies(
         dep_type = classify_resolved_url(url)
         proxy_auth = None
 
-        if dep_type == "file":
-            continue
-        elif dep_type == "git":
+        # dep_type is "git", "registry", or "url". "file" deps are filtered
```
taylormadore: nit: dep type is https not url

derasdf: Fixed

hermeto/core/package_managers/gomod/go.py
derasdf: Added

hermeto/core/package_managers/gomod/go.py
derasdf: Fixed

hermeto/core/package_managers/generic/models.py
derasdf: Added check for this.

taylormadore: I'm wondering whether this should be an after validator instead. I don't really see any benefits of doing the mutual exclusivity check on the raw keys before the basic validation occurs. An after validator seems cleaner and simpler to me

derasdf: Sure, switched to an after validator.

hermeto/core/package_managers/generic/models.py
```diff
@@ -121,14 +121,11 @@ class LockfileArtifactAuth(BaseModel):
...
 
-    @model_validator(mode="before")
-    @classmethod
-    def _check_mutually_exclusive(cls, values: dict) -> dict:
-        if ("basic" not in values and "bearer" not in values) or (
-            "basic" in values and "bearer" in values
-        ):
+    @model_validator(mode="after")
+    def _check_mutually_exclusive(self) -> "LockfileArtifactAuth":
+        if (self.basic is None) == (self.bearer is None):
```
eskultety, line 126: Why the `==` operator instead of  `and` since we're doing a compound boolean check?

hermeto/core/package_managers/gomod/go.py
```diff
@@ -507,9 +514,7 @@ def _list_installed_toolchains() -> set[Go]:
         try:
             log.debug("Probing %s toolchain...", bin_path)
             ret.add(Go(binary=bin_path.as_posix()))
-        except Exception as e:
-            # Logging toolchain probing failures due to [1].
-            # [1] https://bandit.readthedocs.io/en/1.8.3/plugins/b112_try_except_continue.html
```
eskultety, line 512: You somehow lost this commentary which explained why we're not swallowing failures here silently.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1514,12 +1517,19 @@ def prepare_netrc_contents() -> str:
...
+    """Write content to a file with explicit permissions, bypassing umask."""
+    old_mask = os.umask(0)
+    try:
+        fd = os.open(path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, mode)
+        with open(fd, "w") as f:
+            f.write(content)
+    finally:
+        os.umask(old_mask)
+
+
```
eskultety, line 1530: nitpick: this is not a bug, but a cleanup extraction --> standalone commit.

tests/unit/package_managers/gomod/test_go.py
```diff
@@ -384,6 +384,30 @@ def test_select_toolchain(
         assert str(result.version) == expected_result
 
 
+@mock.patch("hermeto.core.package_managers.gomod.go.Go.from_missing_toolchain")
+@mock.patch("hermeto.core.package_managers.gomod.go.Go._get_release")
+@mock.patch("hermeto.core.package_managers.gomod.go._get_gomod_version")
+def test_select_toolchain_returns_none_on_download_timeout(
```
eskultety, line 390: This test is not mentioned in the commit message. Is it related to any of the issues you noted in there? To me it sounds like a new test case.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -78,8 +78,6 @@ class ParsedOrigin(_ParsedModel):
     vcs: str
     url: str
```
eskultety, line 79: > but the CLI   flag was left behind, causing Go to serialize unused data.

Do you have an example of this to share?

> The Deps field was removed
>   from ParsedPackage during the pydantic v2 migration but the CLI
>   flag was left behind, causing Go to serialize unused data.

Commit blame would also be nice with these types of claims.

hermeto/core/package_managers/javascript/npm/resolver.py
```diff
@@ -96,9 +96,9 @@ def _get_npm_dependencies(
         dep_type = classify_resolved_url(url)
         proxy_auth = None
```
eskultety, line 97: Commit message:

> Orphaned when YarnRc was  refactored from typed properties to UserDict.

A commit blame would be nice.

hermeto/core/utils.py
```diff
@@ -204,4 +206,15 @@ def get_cache_dir() -> Path:
...
+    """Return the first match of predicate in iterable or fallback value."""
+    return next((x for x in iterable if predicate(x)), fallback)
+
+
+def partition_by(predicate: Callable, iterable: Iterable) -> tuple[Iterable, Iterable]:
+    """Partition iterable in two by predicate."""
+    i1, i2 = tee(iterable)
+    return filterfalse(predicate, i1), filter(predicate, i2)
```
eskultety, line 220: https://github.com/hermetoproject/hermeto/pull/1724/changes/73ff21951ca1494630f6bada06386a1f1a6cb363

Isn't this commit supposed to remove dead utilities rather than adding them? :)

tests/integration/utils.py
```diff
@@ -28,9 +28,6 @@
     validate_and_strip_proxy_refs,
 )
```
eskultety, line 30: Commit message:

> Dead copy, consumer removed in test consolidation.

Commit blame would be nice for easier tracking.

eskultety (commented):
Overall LGTM, left a few comments since you need to respin one more time anyway.

hermeto/core/package_managers/generic/models.py
derasdf: Added an explanation `reject when both are set or both are unset; exactly one must be present`.

hermeto/core/package_managers/gomod/go.py
derasdf: Restored.

hermeto/core/package_managers/gomod/main.py
derasdf: Split.

tests/unit/package_managers/gomod/test_go.py
derasdf: Added a message bullet, it's the regression guard for this commit's `_select_toolchain` narrowing.

hermeto/core/package_managers/gomod/main.py
derasdf: Fixed.

hermeto/core/package_managers/javascript/npm/resolver.py
derasdf: Updated.

hermeto/core/utils.py
derasdf: Oops, bad rebase i missed.

tests/integration/utils.py
derasdf: Added.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1518,14 +1518,19 @@ def prepare_netrc_contents() -> str:
         password {secret}"""
 
 
-def inject_netrc(netrc_stuff: str, temp_netrc_dir: Path) -> str:
```
eskultety, line 1521: Commit message:

> kept separate from the try/finally bug fix it was previously bundled with.

I don't think this is commit message material, that's a PR comment material, but a commit always describes the change being proposed only.

hermeto/core/package_managers/cargo/main.py
```diff
@@ -444,8 +444,7 @@ def _sanitized_cargo_config_file(package_dir: RootedPath) -> Generator[None, Non
         yield
     finally:
         for config, data in configs_contents:
-            if data is not None:
-                config.path.write_text(data)
+            config.path.write_text(data)
```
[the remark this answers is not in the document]
eskultety: IIUC this is already a problem on main, this PR doesn't change that. Better address it in a follow-up.

eskultety (approved):

hermeto/core/package_managers/rpm/main.py
eskultety: > is there any requirement that states that a missing summary must be not present at all?

There is not. That said, some of our pydantic models declare that `None` fields should not be formatted. Dumping empty fields is not wrong per-se, it's just more explicit and I'd argue it's even simpler to downstream parsers which don't need to handle the "missing" case as "not being set", but if we want to adopt that practice, then we should do it globally as a policy, i.e. we're deliberately formatting empty values in all scenarios.

hermeto/core/package_managers/gomod/main.py
derasdf: Removed

PR #1727, opened 2026-08-15
fix(cargo): Report workspace members as local packages

### LINKED ISSUE #1455: bug(cargo) :Cargo workspace members are not recognized as local dependencies in SBOM (missing VCS metadata)

NoiceHax:
Cargo workspace members do not have to be repeated as path dependencies. Hermeto only read `[dependencies]` and `[patch.crates-io]` when looking for local crates, so a member declared only in `[workspace] members` landed in the SBOM as an external package with no `vcs_url` and no `subpath`. A crate that lives in the repository then looks just like one downloaded from crates.io.

`_find_local_packages` now resolves `[workspace] members` too. Entries can be globs, so each pattern is expanded, and every candidate manifest is read to get `[package].name`, since the crate name does not have to match the name of its directory. Paths under `[workspace] exclude` and directories that are workspace roots of their own are skipped, and an explicit path dependency still wins, so behaviour for path deps is unchanged.

Tests cover a virtual workspace with a globbed member whose directory name differs from the crate name, and that registry crates are still reported as external. Claude Code helped with this change, disclosed with an Assisted-by trailer as the AI policy requires.

Fixes https://github.com/hermetoproject/hermeto/issues/1455


tests/unit/package_managers/cargo/test_main.py
```diff
@@ -508,3 +510,117 @@ def test_generate_sbom_components_strict_mode_raises_without_git_repo(
 
     with pytest.raises(NotAGitRepo):
         _generate_sbom_components(rooted_tmp_path, request)
+
+
+def _write_member(package_dir: RootedPath, subpath: str, content: str) -> None:
```
a-ovchinnikov, line 515: Please avoid interacting with file system when implementing unit tests, if you need FS interactions then it is likely not a UT.

PR #1728, opened 2026-08-15
config: Name the offending settings in proxy validation errors

NoiceHax:
When proxy settings are only half filled in, Hermeto says what is wrong but not where. "Proxy URL must be set when proxy credentials are set" is the same sentence for every backend, since they all share ProxyMixin, so anyone running proxies for a few package managers has to go hunting for the one at fault.

The mixin can answer that itself. Each subclass is named after the Config field holding it, so PipSettings belongs to the pip section, and the message can name pip.proxy_url and HERMETO_PIP__PROXY_URL. Printing both spellings is deliberate: config can come from a YAML file as well as from the environment, so naming only the variable would be misleading for anyone using --config-file or ~/.config/hermeto/config.yaml.

The error type stays InvalidInput. https://github.com/hermetoproject/hermeto/issues/1714 covers reporting several problems in one go, and changing the type here would collide with it.

Covered by three tests in tests/unit/test_config.py plus a doctest.


hermeto/core/config.py
```diff
@@ -114,18 +114,59 @@ class ProxyMixin(BaseModel):
...
+        name the offending backend. Both spellings are given because the value could
+        have come from a config file just as well as from the environment.
+
+        >>> PipSettings()._setting_name("proxy_login")
+        'pip.proxy_login (HERMETO_PIP__PROXY_LOGIN)'
+        """
+        # Every ProxyMixin subclass is named after the Config field holding it,
+        # which is also the config file key and the environment variable section.
+        section = type(self).__name__.removesuffix("Settings").lower()
+        env_var = f"{APP_NAME.upper()}_{section.upper()}__{field.upper()}"
```
a-ovchinnikov, line 130: This looks brittle, both the prefix and the separator are configurable. Have you considered other ways of extracting the values?

hermeto/core/config.py
```diff
@@ -114,18 +114,59 @@ class ProxyMixin(BaseModel):
...
 
+    def _setting_name(self, field: str) -> str:
+        """Spell out one of this section's settings the way a user can set it.
+
+        Several backends can be misconfigured the same way, so an error message has to
+        name the offending backend. Both spellings are given because the value could
+        have come from a config file just as well as from the environment.
+
+        >>> PipSettings()._setting_name("proxy_login")
+        'pip.proxy_login (HERMETO_PIP__PROXY_LOGIN)'
```
a-ovchinnikov, line 125: WDYT about the other issue which deals with sources?

a-ovchinnikov (commented):
Hi @NoiceHax, thank you for your contribution! Please note, that Hermeto has a strict policy on work attribution, make sure your commits adhere to it.

a-ovchinnikov:
When attributing commits please follow the precedent set in repository history.

PR #1729, opened 2026-08-18
chore(tests): remove redundant pip_full_hashes integration test

### LINKED ISSUE #1065: Merge pip/full-hashes and pip/e2e
As our pip backend features continue to expand, so do our integration tests for pip. Part of #803 could be merging two of them, namely:

- https://github.com/hermetoproject/integration-tests/tree/pip/full-hashes
- https://github.com/hermetoproject/integration-tests/tree/pip/e2e

The former is basically a subset of the latter.

**A/C**

The e2e test should contain all dependency types, with generated hashes, without any redundant files, and use a command to verify a successful build.

derasdf:
pip_e2e covers the same dep types and additionally builds an image.
pip_full_hashes was needed when pip_e2e had VCS deps that couldn't
have hashes. pip_e2e later switched to URL deps, making it redundant.

The remaining clean up of test files (cachi2 refs, dead files)
will be done together with other scenarios in a follow-up.

Closes: https://github.com/hermetoproject/hermeto/issues/1065


eskultety (commented):
> Also remove dead Dockerfile and devfile.yaml from pip_e2e. The
> scenario uses Containerfile, the other two are leftovers from the
> old branch-based layout.

This belongs to a dedicated commit, not the main one. Since you're touching the e2e test (by removing stuff), you might also want to simplify the test itself given it references cachi2 sources, otherwise that particular "cleanup" feels largely incomplete. ACK to dropping the test, but if you want to touch the e2e test, then it should be done properly and not partially.

eskultety (approved):

tnevrlka (approved):

PR #1743, opened 2026-08-24
.github: renovate: replace Dependabot with Renovate

ST3V1K:
## Summary

Replace Dependabot with Renovate to cut down on dependency-update noise from integration-test fixtures while keeping automated updates for GitHub Actions, the production `Dockerfile`, root `requirements*.txt`, and test Containerfiles.

- Remove `.github/dependabot.yml` and add `.github/renovate.jsonc` with rules that mirror the old Dependabot policy (weekly Actions, monthly Docker/pip, manual Go/Rust major+minor, Node `<25`, grouped production image and Python minor/patch updates).
- Disable updates and vulnerability alerts for most test fixtures; allow grouped **patch-only** bumps for `tests/**` Containerfiles.
- Map `mirror.gcr.io` to Docker Hub for Renovate tag lookups.
- Rename `dependabot-pipcompile.yml` → `renovate-pipcompile.yml`.
- Update `docs/gomod.md` to reference Renovate.
- Disable the Renovate dependency dashboard (`:disableDependencyDashboard`) to avoid keeping an open tracking issue in the repo.

## Deployment note

This only takes effect once the [Mend Renovate app](https://github.com/apps/renovate) is [installed on the org or repository](https://docs.renovatebot.com/getting-started/installing-onboarding/#repository-installation) and this repo is included in the selected repositories. Merging the config alone does not enable Renovate.

**Edit:** I tested this Renovate config on my fork [ST3V1K/hermeto2](https://github.com/ST3V1K/hermeto2) with the Mend app installed. The open Renovate PRs there show what the config produces in practice (grouped updates, scoped managers, test fixture exclusions, etc.).


.github/renovate.jsonc
```diff
@@ -0,0 +1,151 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
```
eskultety, line 2: I think it would be beneficial to split this work into a few standalone commits (even the config changes themselves for easier reviewability), e.g.:

1. Introduce basic renovate config (basic settings)
2. Scoped changes: e.g. ecosystem settings, production Dockerfile updates, test Dockerfile updates (each can be a standalone commit or some of ^these could be paired together, but 1 big lump hinders the review experience)
3. rename the workflow
4. drop dependabot (we'll disable it right after in the repo)

ST3V1K: Done

.github/renovate.jsonc
```diff
@@ -37,6 +37,41 @@
         "python"
       ],
       "enabled": false
+    },
+    // Go: handle major.minor releases manually due to direct feature support
+    // Rust: handle major.minor releases manually since they might break tests
```
eskultety: > / Rust: handle major.minor releases manually since they might break tests

nitpick: maybe be more explicit on what the actual problem we faced was - newer cargo breaking tests (and IIRC even production)

.github/renovate.jsonc
```diff
@@ -72,6 +72,56 @@
...
+    },
+    {
+      "description": "Disable updates on pydantic-core",
+      "matchManagers": [
+        "pip_requirements"
+      ],
+      "matchPackageNames": [
+        "pydantic-core"
+      ],
+      "enabled": false
```
eskultety, line 136: Note, IIRC pydantic has changed their release schedule where `pydantic` should be finally in sync with `pydantic-core` (the desync originally motivated this rule). A nice follow-up to this would be double checking my memory and potentially dropping the rule if we no longer need to worry about dep updates that could not be merged because a pydantic version bump required a different pydantic-core version than the one also being updated by dependabot within the same PR (since the bots don't understand dependency relationships).

.github/renovate.jsonc
```diff
@@ -72,6 +72,56 @@
         "Dockerfile"
       ],
       "groupName": "production Dockerfile"
+    },
+    // Disable non-minor/patch update types instead of the whole manager.
+    // A blanket enabled:false leaves skipReason=disabled and blocks
+    // vulnerability-alert PRs (and can hide deps on the dashboard).
```
eskultety: > A blanket enabled:false leaves skipReason=disabled and blocks vulnerability-alert PRs

Can you translate this to me please? What does it mean in practice?

.github/renovate.jsonc
```diff
@@ -72,6 +72,56 @@
...
+    {
+      "description": "Group Python minor and patch updates",
+      "matchManagers": [
+        "pip_requirements"
+      ],
+      "matchUpdateTypes": [
+        "minor",
+        "patch"
+      ],
+      "groupName": "minor and patch Python dependencies"
```
eskultety, line 126: nitpick: shouldn't the group names follow some programming conventions like dashes or underscores?

.github/renovate.jsonc
```diff
@@ -122,6 +122,30 @@
         "createrepo-c"
       ],
       "allowedVersions": "<1.2.2"
+    },
+    {
+      "description": "Disable updates on test dependencies",
+      "matchFileNames": [
+        "tests/**"
```
eskultety: If we need to do this, how exactly is
```
ignorePresets": [
    ":ignoreModulesAndTests"
```
at the top of the config insufficient? Looking at the [docs](https://docs.renovatebot.com/presets-default/#ignoremodulesandtests) it already ignores various test directories.
Is it due to the vulnerability alerts that the vulnerability setting would not take effect until we disabled test directories explicitly? Please explain.

.github/renovate.jsonc
```diff
@@ -122,6 +122,30 @@
...
+      ],
+      "vulnerabilityAlerts": {
+        "enabled": false
+      },
+      "enabled": false
+    },
+    {
+      "description": "Group all test Containerfile updates",
```
eskultety: For better visibility this one should be moved after the production Dockerfile updates.

.github/renovate.jsonc
```diff
@@ -122,6 +122,30 @@
...
+      "matchManagers": [
+        "dockerfile"
+      ],
+      "matchFileNames": [
+        "tests/**"
+      ],
+      "matchUpdateTypes": [
+        "patch"
```
eskultety: For initial Renovate config I don't care much about this rule, but note that once we unify the test images' base image deps such that they are (ideally) in sync with the production image we'll most likely want the same rules to apply IMO to be aligned wrt/ base image deps across the repo.

.github/dependabot.yml
```diff
@@ -1,85 +0,0 @@
-# Please see the documentation for all configuration options:
```
eskultety, line 1: **Commit message:**

>  once the app is installed on the org.

It already is, so ^this can be dropped from the commit.

eskultety (commented):
A few minor comments. With PRs like this one regarding Github actions I think production testing is probably the easiest and time efficient way of going about it rather than deploying on the side and perform rigorous testing. The config looks sound (without digging much in the documentation).
@ST3V1K I trust you have deployed  it in your fork (you forgot to link it in the PR description as a proof ;) ).
ACK with the minor comments addressed. Let's see what Renovate has to offer.

.github/renovate.jsonc
ST3V1K: I first tried disabling all `pip_requirements` updates and then re-enabling minor/patch in a later rule. That didn’t work: Renovate marked everything as fully disabled, a later rule couldn’t override that for specific update types, and security-alert PRs didn’t show up.

So we disable only the update types we don’t want (major, pin, digest, etc.) and leave minor/patch enabled. Routine updates stay limited, but security fixes can still come through.

.github/renovate.jsonc
ST3V1K: They don't have to, and IMO this is better since they appear in the name of PRs. The name follows "update {groupName}" and "update minor and patch Python dependencies" looks better than "update minor-patch-python-deps".

.github/renovate.jsonc
ST3V1K: No, we reenable the scanning of `tests/**` with 
```
ignorePresets": [
    ":ignoreModulesAndTests"
```
The `config:recommended` disables it by default and we need Renovate to scan those for both normal and security test Dockerfile updates.
I found this the easiest way to do this since [`config:recommended`](https://docs.renovatebot.com/presets-config/#configrecommended) still brings some value

.github/renovate.jsonc
ST3V1K: They merged `pydantic-core` into `pydantic` and releases are meant to stay in sync, but bots still treat both as independent pins in `requirements.txt`. A newer `pydantic-core` can land on PyPI before any `pydantic` release accepts it, which this rule avoids.

We already have a safety net: the `pip-compile` workflow runs on requirements PRs, recompiles from `pyproject.toml`, and comments with a patch if transitive pins (including `pydantic-core`) need updating. It doesn't auto-commit, but that's the intended path when `pydantic` moves. I'd keep the rule until Pydantic v3 drops the standalone package.

.github/renovate.jsonc
ST3V1K: Agreed - once test images share the production base deps, we should mirror the same rules here

.github/dependabot.yml
ST3V1K: Done

.github/renovate.jsonc
ST3V1K: Done

.github/renovate.jsonc
ST3V1K: Done

.github/renovate.jsonc
eskultety: ACK.

.github/renovate.jsonc
eskultety: > We already have a safety net: the pip-compile workflow runs on requirements PRs, recompiles from pyproject.toml, and comments with a patch if transitive pins (including pydantic-core) need updating

I know we have it, I added it ;) . My question simply was whether it's still applicable? I mean if releases of both happen at the same time and the versions are locked, i.e. a newly released `pydantic` is pinned against a version of `pydantic-core` which was released along with it **and not to an older release**, then

> A newer pydantic-core can land on PyPI before any pydantic release accepts it

is practically irrelevant, isn't it?

ST3V1K: Technically, yes. But nothing enforces that releases will happen at the same time. And since Pydantic V3 is supposed to merge `pydantic-core`as an internal sub-module, I'd wait with removing this.
https://pydantic.dev/docs/validation/dev/get-started/version-policy/#pydantic-v3-and-beyond

eskultety: Well, waiting for v3 might - firstly, take a while, and secondly, be incompatible with v2 so tough luck for us. Anyway, I was just curious about the status quo of their release situation, I'm fine keeping our working hack. Resolving.

.github/renovate.jsonc
eskultety: Resolving, @ST3V1K feel free to update the tracking issue for the test images so that we don't forget unless you plan doing it right after :).

.github/renovate.jsonc
```diff
@@ -0,0 +1,151 @@
...
+    ":disableDependencyDashboard"
+  ],
+  "ignorePresets": [
+    ":ignoreModulesAndTests"
+  ],
+  "enabledManagers": [
+    "github-actions",
+    "dockerfile",
+    "pip_requirements"
+  ],
```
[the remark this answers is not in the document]
eskultety: > Disable standalone pip_requirements updates, and move its existing package rules to pip-compile

@ST3V1K can you try if using the `pip-compile` mgr would render the need for a pip-compile workflow unnecessary? IOW if renovate can run pip-compile on requirements then theoretically the `pydantic-core` issue would disappear.

.github/renovate.jsonc
eskultety: > So we disable only the update types we don’t want (major, pin, digest, etc.) and leave minor/patch enabled. Routine updates stay limited, but security fixes can still come through.

Cool, now I get it. Can you please rephrase the original comment in a more humanly way? :)

.github/renovate.jsonc
eskultety: Ah, I see it now, thanks. Resolving.

.github/renovate.jsonc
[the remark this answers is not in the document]
ST3V1K: I tried using the `pip-compile` manager first, but it's still pretty limited. A lot of CLI flags aren't supported yet, including `--python=3.10` and `--no-strip-markers`, which we need. I don't think there's a way around that unless Renovate adds support for them.

.github/renovate.jsonc
ST3V1K: Done

.github/renovate.jsonc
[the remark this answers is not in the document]
eskultety: Thanks for double-checking. Please update commit message 8/9 with this finding, i.e. the reason we're still keeping our workflow despite Renovate having "something" in place. Make sure to link the relevant docs as well.

eskultety (approved):

.github/renovate.jsonc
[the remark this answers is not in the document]
ST3V1K: I added it to 61ffaa59 and referenced it in 8d12507a. 8th is 699b5299 and that doesn't seem right.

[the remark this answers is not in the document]
eskultety: Apparently I forgot how to count :).

derasdf (approved):

PR #1752, opened 2026-08-30
fix(gomod): resolve module versions for go submodules

### LINKED ISSUE #1736: [gomod] main modules inside git submodules inherit the parent repo version
When a main module lives in a git submodule, the version is derived from the parent repo instead of the submodule repo. This reported version does not accurately identify the package. It also means that the reported version of the module will change on each new commit in the parent repo even if the submodule ref itself has not changed.

Harsh-2005d:
Fixes: https://github.com/hermetoproject/hermeto/issues/1736

## Changes
- Initialize and cache `ModuleVersionResolver` per git repo, keyed by `working_tree_dir`, instead of a single resolver
- Use `get_repo_for_path` to determine the correct repo before resolver lookup
- Fix subpath resolution in `get_golang_version`

### DIFF
--- a/hermeto/core/package_managers/gomod/main.py
+++ b/hermeto/core/package_managers/gomod/main.py
@@ -6,6 +6,7 @@ import shutil
 import tempfile
 import urllib
 from collections.abc import Iterable, Iterator
+from dataclasses import dataclass
 from datetime import datetime, timezone
 from functools import cached_property
 from itertools import chain
@@ -259,6 +260,47 @@ class StandardPackage(NamedTuple):
         return Component(name=self.name, purl=self.purl)
 
 
+@dataclass
+class RepoContext:
+    """Resolved repository context used for Go module processing.
+
+    repo_dir: root directory of the source repository
+    repo_name: repository name, or None for non-Git sources
+    repo_id: repository identity, or None for non-Git sources
+    version_resolver: resolver used to determine module versions
+    """
+
+    repo_dir: RootedPath
+    repo_name: str | None
+    repo_id: RepoID | None
+    version_resolver: "ModuleVersionResolver"
+
+    @classmethod
+    def from_repo_path(cls, repo_dir: RootedPath) -> "RepoContext":
+        """Create a repository context from a Git repository path."""
+        return cls(
+            repo_dir=repo_dir,
+            repo_name=_get_repository_name(repo_dir),
+            repo_id=get_repo_id(repo_dir.path),
+            version_resolver=ModuleVersionResolver.from_repo_path(repo_dir),
+        )
+
+    @classmethod
+    def from_non_git_source(cls, source_dir: RootedPath) -> "RepoContext":
+        """Create a repository context for a source without Git metadata."""
+        return cls(
+            repo_dir=source_dir,
+            repo_name=None,
+            repo_id=None,
+            version_resolver=ModuleVersionResolver.from_non_git_source(),
+        )
+
+    def subpath_for(self, module_dir: RootedPath) -> str | None:
+        """Return the module path relative to the repository root."""
+        relative = module_dir.path.relative_to(self.repo_dir.path)
+        return None if str(relative) == "." else str(relative)
+
+
 ModuleID = tuple[str, str]
 
 
@@ -467,14 +509,7 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
     components: list[Component] = []
     annotations: list[Annotation] = []
 
-    repo_name = _get_repository_name(request.source_dir)
-    try:
-        version_resolver = ModuleVersionResolver.from_repo_path(request.source_dir)
-    except NotAGitRepo:
-        if get_config().mode == Mode.PERMISSIVE:
-            version_resolver = ModuleVersionResolver.from_non_git_source()
-        else:
-            raise
+    repo_contexts: dict[RootedPath, RepoContext] = {}
 
     gomod_download_dir = request.output_dir.join_within_root("deps/gomod/pkg/mod/cache/download")
     gomod_download_dir.path.mkdir(exist_ok=True, parents=True)
@@ -485,6 +520,31 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
             go_work: GoWork | None = None
 
             main_module_dir = request.source_dir.join_within_root(subpath)
+            try:
+                repo, _ = get_repo_for_path(request.source_dir.root, main_module_dir.path)
+                if repo.working_tree_dir is None:
+                    raise NotAGitRepo(
+                        f"Git repository at {main_module_dir} has no working tree (bare clone).",
+                        solution="Use a non-bare git clone when running hermeto.",
+                    )
+                repo_dir = RootedPath(repo.working_tree_dir)
+            except NotAGitRepo:
+                log.warning(
+                    "Could not find a git repository for %s, falling back to source directory",
+                    main_module_dir.path,
+                )
+                repo_dir = request.source_dir
+
+            if repo_dir not in repo_contexts:
+                try:
+                    repo_contexts[repo_dir] = RepoContext.from_repo_path(repo_dir)
+                except NotAGitRepo:
+                    if get_config().mode == Mode.PERMISSIVE:
+                        repo_contexts[repo_dir] = RepoContext.from_non_git_source(repo_dir)
+                    else:
+                        raise
+
+            ctx = repo_contexts[repo_dir]
             go = _select_toolchain(main_module_dir.join_within_root("go.mod"), installed_toolchains)
             if go is None:
                 raise FetchError(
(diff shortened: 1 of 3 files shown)


hermeto/core/package_managers/gomod/main.py
```diff
@@ -485,6 +479,22 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
             go_work: GoWork | None = None
 
             main_module_dir = request.source_dir.join_within_root(subpath)
+            try:
+                repo, _ = get_repo_for_path(request.source_dir.root, main_module_dir.path)
+                repo_dir = RootedPath(repo.working_tree_dir or request.source_dir.root)
+            except NotAGitRepo:
+                repo_dir = request.source_dir
```
derasdf: NotAGitRepo here (uninitialized submodule) falls back to the parent repo silently, the exact #1736 behavior. At least a log.warning, and IMO worth deciding whether the parent fallback is even right.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1166,10 +1176,16 @@ def get_golang_version(
         # If no match, prefer v1.x.x tags but fallback to v0.x.x tags if both are present
         major_versions_to_try = (module_major_version,) if module_major_version else (1, 0)
 
+        repo_dir = (
+            Path(self._repo.working_tree_dir)
+            if self._repo and self._repo.working_tree_dir
+            else app_dir.root
+        )
+
         if app_dir.path == app_dir.root:
```
derasdf: For a main module at a submodule root, app_dir.path == repo_dir, so this hits the else and subpath becomes "." (relative_to of equal paths), so prefix "./v" matches no tags and a tagged submodule HEAD gets a pseudo-version instead. Guard should key off repo_dir:

```python
if app_dir.path == repo_dir:
    subpath = None
```

That is the common submodule shape. How does this match `go list`?

hermeto/core/package_managers/gomod/main.py
```diff
@@ -1166,10 +1176,16 @@ def get_golang_version(
...
             subpath = None
         else:
-            subpath = app_dir.path.relative_to(app_dir.root).as_posix()
+            subpath = app_dir.path.relative_to(repo_dir).as_posix()
```
derasdf: nitpick: relative_to(repo_dir) throws for a local `replace => ../x` above the submodule. Niche, can guard with is_relative_to.

hermeto/core/package_managers/gomod/main.py
```diff
@@ -485,6 +479,22 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
             go_work: GoWork | None = None
 
             main_module_dir = request.source_dir.join_within_root(subpath)
+            try:
```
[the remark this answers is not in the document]
derasdf: Also please rebase, branch is stale.

hermeto/core/package_managers/gomod/main.py
derasdf: For a main module at a submodule root, `app_dir.path == repo_dir`, so this hits the else and subpath becomes `.` (relative_to of equal paths), so prefix `./v` matches no tags and a tagged submodule HEAD gets a pseudo-version instead. Guard should key off `repo_dir`:

```python
if app_dir.path == repo_dir:
    subpath = None
```

That is the common submodule shape. How does this match `go list`?

hermeto/core/package_managers/gomod/main.py
```diff
@@ -485,6 +479,22 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
...
+                repo, _ = get_repo_for_path(request.source_dir.root, main_module_dir.path)
```
derasdf, line 524: This warrants a unit test, `get_golang_version` isn't a `fetch_*` entrypoint. Note `test_get_golang_version` won't catch a regression here: it roots `RootedPath` at `working_tree_dir`, so `root == repo_dir`. Also the failing IT, can you trace why before this lands, and regenerate the bom only if the new version is actually correct?

hermeto/core/package_managers/gomod/main.py
```diff
@@ -485,6 +479,22 @@ def fetch_gomod_source(request: Request) -> RequestOutput:
...
+                repo_dir = RootedPath(repo.working_tree_dir or request.source_dir.root)
```
derasdf: Design question: this walks down from `source_dir.root` to find the owning repo (feeds version), but [`get_repo_id`](https://github.com/hermetoproject/hermeto/blob/dd5ac7c244b8dca2b0b7017e1baa576641c75342/hermeto/core/package_managers/gomod/main.py#L608) walks up to the nearest `.git` (feeds `vcs_url`). Two ways to answer the same question, feeding two fields, that can disagree for nested submodules. Worth resolving the owning repo once and deriving `name`/`vcs_url`/`version`/`subpath` from it? Would also make `get_golang_version`'s root handling consistent and kill the `.` case above.

hermeto/core/package_managers/gomod/main.py
derasdf, line 536: `NotAGitRepo` here (uninitialized submodule) falls back to the parent repo silently, the exact #1736 behavior. At least a `log.warning`, and IMO worth deciding whether the parent fallback is even right.

hermeto/core/package_managers/gomod/main.py
derasdf, line 523: nitpick: reference the issue in the commit body, full URL:

`Fixes: https://github.com/hermetoproject/hermeto/issues/1736`

hermeto/core/package_managers/gomod/main.py
derasdf, line 1236: nitpick: `relative_to(repo_dir)` throws for a local `replace => ../x` above the submodule. Niche, can guard with `is_relative_to`.

hermeto/core/package_managers/gomod/main.py
derasdf: Also branch is stale, rebase main

derasdf (changes requested):
Blocking: the subpath change breaks the module-at-submodule-root case (falls back to a pseudo-version instead of the tag), and there are no tests plus a self-declared failing IT. Also a design question on repo discovery. Details inline.

hermeto/core/package_managers/gomod/main.py
Harsh-2005d: I have added a warning for this meanwhile. I think the parent fallback is fine, but I’ll leave the decision to the maintainers.

hermeto/core/package_managers/gomod/main.py
Harsh-2005d: The failing IT was caused by the module's vcs_url being resolved from the submodule, while its version was resolved from the parent repository. So the two sources of information disagreed. I think a integration test is more suitable here.

hermeto/core/package_managers/gomod/main.py
Harsh-2005d:
```sh
wolverine@fedora:/tmp/go-submodule-test$ go list -m -json
{
	"Path": "example.com/module",
	"Main": true,
	"Dir": "/tmp/go-submodule-test/parent/submodule",
	"GoMod": "/tmp/go-submodule-test/parent/submodule/go.mod",
	"GoVersion": "1.22"
}
```
This behavior did match the go list output. For a module at the root of the submodule repository, go list reports the submodule path itself as Dir

hermeto/core/package_managers/gomod/main.py
Harsh-2005d: > This warrants a unit test

I think a integration would be more suitable here, it will verify that when a Go module is located inside a Git submodule, hermeto resolves and uses the submodule's repository, rather than the parent repository, for version resolution.

hermeto/core/package_managers/gomod/main.py
derasdf: This one's still unanswered, and it's the one that matters before the rest.

The new bom is the symptom: the submodule module still shows the parent's commit for both `version` and `vcs_url`.

Harsh-2005d: Sorry i missed that, I think we should resolve the repo once via a ctx dataclass keyed by the repo path. 
```python
@dataclass
class RepoContext:
    """Resolved git repository context for a main module."""

    repo_dir: RootedPath
    repo_name: str
    repo_id: RepoID

    @classmethod
    def from_repo_path(cls, repo_dir: RootedPath) -> "RepoContext":

```
I’ll push the changes if you agree with this approach.

derasdf: Yep, go with that. Two things:

- handle the permissive non-git case (`repo_name`/`repo_id` are `None` there)
- make sure the submodule IT bom shows the submodule's own commit and origin, not the parent's. It still shows the parent now, so check `get_repo_for_path` actually descends.

I'll review the rest once that's in.

Harsh-2005d: @derasdf
cc: #1767. A plausible fix is to allow `SyntheticSubmoduleSpec` to provide its own `origin_url`, so the generated submodule can use its own repository URL instead of the parent's.

I'm not completely sure whether the current behavior is intentional for the synthetic repo helper, though, so it would be good to clarify whether submodules are expected to inherit the parent origin in these tests.

PR #1755, opened 2026-09-02
Revisit older integration tests (vol. 1)

### LINKED ISSUE #803: Revisit older integration tests

derasdf:
First pass at trimming the suite migrated in #1669. Drops scenarios that
don't exercise any code path a kept scenario already does, removes input
files hermeto never reads, and refreshes some stale infra. No production
code touched.

Removed as redundant: one cargo git-crate test (cargo_e2e covers it),
three gomod (1.18 is EOL, an order-only vendor dupe, and the empty-vendor
pair that hits the same reject branch as the kept wrong-vendor tests),
five npm and three yarn/yarn-classic that all reduce to plain lockfile
entries through the single-package path, pip_yanked, and rpm_missing_checksum.
Rationale for each is in its commit message.

Also: trimmed dead files from the yarn error scenarios, removed unused
build/tool artifacts, pinned the CycloneDX schema to the 1.6.2 tag, and
bumped EOL base images.

Closes: https://github.com/hermetoproject/hermeto/issues/803


tests/integration/cargo/test_cargo.py
```diff
@@ -28,13 +28,6 @@
             ),
             id="cargo_just_a_git_dependency",
         ),
-        pytest.param(
-            utils.TestParameters(
-                packages=({"path": ".", "type": "cargo"},),
-                check_output=False,
-            ),
-            id="cargo_mixed_git_crate_dependency",
```
a-ovchinnikov, line 36: What is the plan for in/out files for these two tests?

tests/integration/gomod/test_gomod.py
```diff
@@ -156,14 +156,6 @@ def test_gomod_packages(
 @pytest.mark.parametrize(
     "test_params,check_cmd,expected_cmd_output",
     [
-        pytest.param(
-            utils.TestParameters(
-                packages=({"path": ".", "type": "gomod"},),
-            ),
-            ["retrodep", "--help"],
-            ["retrodep: help requested"],
-            id="gomod_e2e_1_18",
```
a-ovchinnikov, line 165: If this version is still supported by Hermeto then it probably should not be removed, AFAIK there could be significant changes in behavior between versions. These tests help ensure that there are no regressions when changes are introduced to Hermeto.

tests/integration/rpm/test_rpm.py
```diff
@@ -16,13 +16,6 @@
 @pytest.mark.parametrize(
     "test_params",
     [
-        pytest.param(
-            utils.TestParameters(
-                packages=({"path": ".", "type": "rpm"},),
-                check_output=True,
-            ),
-            id="rpm_missing_checksum",
```
a-ovchinnikov, line 24: I think a comment right next to `rpm_multiple_packages` case would help with maintenance: right now there is no good way to tell that it also checks missing checksum case.

tests/integration/gomod/test_gomod.py
```diff
@@ -195,17 +195,6 @@ def test_gomod_packages(
...
-        pytest.param(
-            utils.TestParameters(
-                packages=(
-                    {"path": "non-vendored-module", "type": "gomod"},
-                    {"path": "vendored-module", "type": "gomod"},
-                ),
-            ),
-            [],
-            [""],
-            id="gomod_e2e_vendor_nonvendor_module_mix_ordering_2",
```
a-ovchinnikov, line 215: > This is a leftover from cachito's global-state assumptions that do
not hold for a per-package, lockfile-driven tool.

Could you please point to source of this statement?

tests/integration/gomod/test_gomod.py
```diff
@@ -89,25 +89,6 @@ def _create_repo(tmp_path: Path, test_case: str) -> Path:
...
             id="gomod_wrong_vendor_passes_vendor_check_in_permissive_mode",
         ),
-        pytest.param(
-            utils.TestParameters(
-                packages=({"path": ".", "type": "gomod"},),
-                check_output=False,
-                expected_error=ExitError.ERR_PACKAGE_REJECTED,
-                expected_output=(
-                    "PackageRejected: The content of the vendor directory is not "
-                    "consistent with go.mod. Please check the logs for more details"
```
a-ovchinnikov, line 99: > gomod_empty_vendor_fails_vendor_check raises the same
ERR_PACKAGE_REJECTED at gomod/main.py as the wrong-vendor scenarios that
stay

The error is the same, the problem that triggers it is different. I think the point of this test is to verify that path. This test looks to be rather fast. I am not sure it is a good idea to remove it.

tests/integration/npm/test_npm.py
```diff
@@ -59,14 +59,6 @@ def test_npm_packages(
             [],
             id="npm_smoketest_lockfile3",
         ),
-        pytest.param(
-            utils.TestParameters(
-                packages=({"path": ".", "type": "npm"},),
-            ),
-            [],
-            [],
-            id="npm_multiple_dep_versions",
```
a-ovchinnikov, line 74: > Multiple versions of the same dependency are just multiple lockfile
entries; hermeto processes them through the same path as any other
dependency, with no version-collision branch. Covered by the kept
npm_smoketest_lockfile3 e2e.

This has to be repeated in a comment in `npm_smoketest_lockfile3`, this will get easily overlooked otherwise

tests/integration/npm/test_npm.py
```diff
@@ -59,18 +59,6 @@ def test_npm_packages(
...
-            utils.TestParameters(
-                packages=(
-                    {"path": "first_pkg", "type": "npm"},
-                    {"path": "second_pkg", "type": "npm"},
-                    {"path": "third_pkg", "type": "npm"},
-                ),
-            ),
-            [],
-            [],
-            id="npm_multiple_packages",
```
a-ovchinnikov, line 102: It is generally considered a good practice to exercise list processor over one, two and possibly three-element lists. Here we deal with state external to our code, I would have kept this.

tests/integration/yarn/test_yarn.py
```diff
@@ -106,17 +106,6 @@ def test_yarn_packages(
...
-        pytest.param(
-            utils.TestParameters(
-                packages=(
-                    {"path": "first-pkg", "type": "yarn"},
-                    {"path": "second-pkg", "type": "yarn"},
-                ),
-            ),
-            ["yarn", "node", "index.js"],
-            "Hello from first package!",
-            id="yarn_e2e_multiple_packages",
```
a-ovchinnikov, line 144: Same concern as in npm.

a-ovchinnikov (commented):
Please find the first batch of comments inline.

tests/integration/cargo/test_cargo.py
derasdf: Removed out/ file, kept in/ as legacy test uses it https://github.com/hermetoproject/hermeto/blob/e2b6faca96756a651dd624a5f5249c976bf4de67/tests/integration/legacy/test_legacy.py#L61

tests/integration/gomod/test_gomod.py
derasdf: Restored

tests/integration/rpm/test_rpm.py
derasdf: Added comment

tests/integration/gomod/test_gomod.py
derasdf: Dropped cachito line

tests/integration/gomod/test_gomod.py
derasdf: Restored

tests/integration/npm/test_npm.py
derasdf: Added comment

tests/integration/npm/test_npm.py
derasdf: Restored

tests/integration/yarn/test_yarn.py
derasdf: Restored

tests/integration/rpm/test_rpm.py
```diff
@@ -42,6 +35,9 @@
             id="rpm_unexpected_size",
         ),
         pytest.param(
+            # Also covers the missing-checksum path: this-project pulls
+            # centos-stream-repos without a checksum, another-project pulls the
+            # same package with one, and the expected SBOM records both forms.
```
eskultety: I think this note belongs to the particular test's README instead of here.

PR #1757, opened 2026-09-02
fix(gomod): validate GOPROXY entries the same way as other proxy URLs

### LINKED ISSUE #1756: GomodSettings does not run the same proxy_url validators as other settings classes
Due to the nature of gomod proxy_url it is typed as a [string](https://github.com/hermetoproject/hermeto/blob/762ada2e8a320162d075d3426eff5f328241f395/hermeto/core/config.py#L161). This bypasses validations applied to proxy URLs defined in other settings classes. To resolve this issue either a proper type must be defined to deal with gomod convention of describing proxies or an after validator should be added to run the checks. The first option is preferred, the second option should be treated as a back-up plan in case this proves to be too complicated with the tools at hand.

Shriprasad-P:
## Summary
- GomodSettings.proxy_url was a plain str (GOPROXY comma-list), so it skipped the no-embedded-credentials check used by other ProxyMixin settings.
- Introduce a GOPROXY-aware type: split on commas, allow direct/off, run the existing credential policy on URL entries, keep the default https://proxy.golang.org,direct.

## Test plan
- [x] new unit tests for valid GOPROXY lists, special tokens, and rejected credential URLs
- [x] existing tests/unit/test_config.py still pass
- [x] pytest tests/unit
- [x] nox -s lint

Closes https://github.com/hermetoproject/hermeto/issues/1756

### DIFF
--- a/hermeto/core/config.py
+++ b/hermeto/core/config.py
@@ -10,6 +10,7 @@ from pydantic import (
     FieldSerializationInfo,
     HttpUrl,
     SecretStr,
+    TypeAdapter,
     ValidationError,
     field_serializer,
     model_validator,
@@ -103,6 +104,41 @@ ProxyUrl = Annotated[
     AfterValidator(_proxy_url_must_not_contain_credentials),
 ]
 
+# Go's GOPROXY tokens are case-sensitive; see https://go.dev/ref/mod#environment-variables
+_GOPROXY_SPECIAL_TOKENS = frozenset({"direct", "off"})
+_HTTP_URL_ADAPTER: TypeAdapter[HttpUrl] = TypeAdapter(HttpUrl)
+_GOPROXY_SYNTAX_HINT = (
+    "Set gomod.proxy_url to a comma-separated list of HTTP(S) URLs and the tokens "
+    "'direct' or 'off' (for example https://proxy.golang.org,direct) using "
+    "HERMETO_GOMOD__PROXY_URL or a configuration file."
+)
+
+
+def _validate_goproxy(value: str) -> str:
+    if not value:
+        raise ValueError(f"gomod.proxy_url must not be empty. {_GOPROXY_SYNTAX_HINT}")
+
+    for entry in value.split(","):
+        if entry in _GOPROXY_SPECIAL_TOKENS:
+            continue
+        if not entry:
+            raise ValueError(
+                f"gomod.proxy_url must not contain empty GOPROXY entries. {_GOPROXY_SYNTAX_HINT}"
+            )
+        try:
+            parsed_url = _HTTP_URL_ADAPTER.validate_python(entry)
+        except ValidationError:
+            raise ValueError(
+                "Each GOPROXY entry must be an HTTP(S) URL or the special token "
+                f"'direct' or 'off' (case-sensitive). {_GOPROXY_SYNTAX_HINT}"
+            ) from None
+        _proxy_url_must_not_contain_credentials(parsed_url)
+
+    return value
+
+
+GoProxy = Annotated[str, AfterValidator(_validate_goproxy)]
+
 
 class ProxyMixin(BaseModel):
     """Proxy URL and authorization mixin.
@@ -156,9 +192,8 @@ class YarnSettings(ProxyMixin, extra="forbid"):
 class GomodSettings(ProxyMixin, extra="forbid"):
     """Settings for Go modules."""
 
-    # TODO: refactor typesystem to trivially include comma-separated lists of URLs.
-    # Ignore type error until that happens.
-    proxy_url: str = "https://proxy.golang.org,direct"  # type: ignore
+    # GOPROXY is a comma-separated list, not a single HttpUrl like other backends.
+    proxy_url: GoProxy = "https://proxy.golang.org,direct"  # type: ignore[assignment]
     download_max_tries: int = 5
     environment_variables: dict[str, str] = {}
 
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -4,8 +4,10 @@ from typing import Any, Generator
 
 import pytest
 import yaml
+from pydantic import ValidationError
 
 import hermeto.core.config as config_module
+from hermeto.core.errors import InvalidInput
 
 DEFAULT_CONCURRENCY = config_module.RuntimeSettings.model_fields["concurrency_limit"].default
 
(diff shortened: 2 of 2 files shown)


tests/unit/test_config.py
```diff
@@ -118,3 +120,82 @@ def test_cli_config_file_overrides_defaults(tmp_home_cwd: Path) -> None:
...
+def test_gomod_proxy_url_rejects_empty_or_invalid_entries(proxy_url: str) -> None:
+    with pytest.raises(ValidationError):
+        config_module.GomodSettings(proxy_url=proxy_url)
+
+
+def test_gomod_proxy_login_and_password_still_pair() -> None:
+    settings = config_module.GomodSettings(
+        proxy_url="https://goproxy.example.com",
+        proxy_login="user",
+        proxy_password="secret",  # noqa: S106
```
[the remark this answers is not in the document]
a-ovchinnikov: @coderabbitai these are tests ergo the comment is noise

hermeto/core/config.py
```diff
@@ -103,6 +104,41 @@ def _proxy_url_must_not_contain_credentials(url: HttpUrl | None) -> HttpUrl | No
     AfterValidator(_proxy_url_must_not_contain_credentials),
 ]
 
+# Go's GOPROXY tokens are case-sensitive; see https://go.dev/ref/mod#environment-variables
+_GOPROXY_SPECIAL_TOKENS = frozenset({"direct", "off"})
+_HTTP_URL_ADAPTER: TypeAdapter[HttpUrl] = TypeAdapter(HttpUrl)
```
a-ovchinnikov, line 109: This is an interesting maneuver, could you please tell me why a need for adapter arises?

hermeto/core/config.py
```diff
@@ -103,6 +104,41 @@ def _proxy_url_must_not_contain_credentials(url: HttpUrl | None) -> HttpUrl | No
...
+_GOPROXY_SYNTAX_HINT = (
+    "Set gomod.proxy_url to a comma-separated list of HTTP(S) URLs and the tokens "
+    "'direct' or 'off' (for example https://proxy.golang.org,direct) using "
+    "HERMETO_GOMOD__PROXY_URL or a configuration file."
+)
+
+
+def _validate_goproxy(value: str) -> str:
+    if not value:
```
a-ovchinnikov, line 118: Won't an empty value result in a fallback?

hermeto/core/config.py
```diff
@@ -103,6 +104,41 @@ def _proxy_url_must_not_contain_credentials(url: HttpUrl | None) -> HttpUrl | No
...
+        raise ValueError(f"gomod.proxy_url must not be empty. {_GOPROXY_SYNTAX_HINT}")
+
+    for entry in value.split(","):
+        if entry in _GOPROXY_SPECIAL_TOKENS:
+            continue
+        if not entry:
```
a-ovchinnikov, line 124: Why not?

tests/unit/test_config.py
```diff
@@ -118,3 +120,82 @@ def test_cli_config_file_overrides_defaults(tmp_home_cwd: Path) -> None:
...
+        pytest.param("off", id="off"),
+        pytest.param("direct", id="direct"),
+    ],
+)
+def test_gomod_proxy_url_accepts_goproxy_lists(proxy_url: str) -> None:
+    settings = config_module.GomodSettings(proxy_url=proxy_url)
+    assert settings.proxy_url == proxy_url
+
+
+def test_gomod_proxy_url_default() -> None:
```
a-ovchinnikov, line 144: Could you please tell me what does this test verify?

tests/unit/test_config.py
```diff
@@ -118,3 +120,82 @@ def test_cli_config_file_overrides_defaults(tmp_home_cwd: Path) -> None:
...
+        pytest.param("OFF", id="off_wrong_case"),
+        pytest.param("ftp://goproxy.example.com", id="non_http_scheme"),
+    ],
+)
+def test_gomod_proxy_url_rejects_empty_or_invalid_entries(proxy_url: str) -> None:
+    with pytest.raises(ValidationError):
+        config_module.GomodSettings(proxy_url=proxy_url)
+
+
+def test_gomod_proxy_login_and_password_still_pair() -> None:
```
a-ovchinnikov, line 187: Could you please explain how this test relates to the proposed change?

hermeto/core/config.py
Shriprasad-P: `proxy_url` still needs to stay a plain `str` (comma-separated GOPROXY list + `direct`/`off`), so it cannot be typed as `HttpUrl` the way the other package backends are. `TypeAdapter(HttpUrl)` lets each list entry reuse the same pydantic `HttpUrl` parsing/rules without changing the field type to a single URL.

hermeto/core/config.py
Shriprasad-P: No — the model default only applies when the field is omitted. Passing an explicit empty string (`""` or via env) is treated as a provided value, so without this check it would be accepted as a valid `str` and later produce a bad `GOPROXY`. Rejecting empty keeps misconfiguration fail-closed instead of silently falling back.

hermeto/core/config.py
Shriprasad-P: Assuming this is about `from None` on the `ValidationError` → `ValueError` rewrite: chaining the pydantic error would surface internal adapter details in the outer message. Suppressing the cause keeps the user-facing error limited to the GOPROXY syntax hint (and matches how we want config errors to read). Happy to chain it if you prefer the full context.

tests/unit/test_config.py
Shriprasad-P: It checks that valid GOPROXY forms still round-trip through `GomodSettings` after the new validator: default-style lists, a single URL, URL+`direct`, multi-URL lists, and the bare `direct`/`off` tokens. That guards against the validator being too strict and rejecting legitimate Go proxy values.

tests/unit/test_config.py
Shriprasad-P: `GomodSettings` still inherits `ProxyMixin`, so login/password pairing must keep working when `proxy_url` becomes `GoProxy` instead of a bare `str`. This test locks that: credentials still pair with a plain URL, and login/password-only configs still raise `InvalidInput` as before. It is regression coverage for the type change, not for GOPROXY list parsing itself.

hermeto/core/config.py
a-ovchinnikov: This deserves a comment clarifying that this deals with a corner case of variable being set to an empty string.

hermeto/core/config.py
a-ovchinnikov: I am sorry, I did not finish my question, I am surprised you managed to make any sense of it at all! This `if` statement apparently guards for possible `,,` substring in the environment variable, my question is -- do we really want to raise on such misconfiguration or just skip it? I would also recommend expressing candidates for verification in a more declarative way e.g. like this:

```python
proxies = [p for p in value.split(",") if p]
for proxy in proxies:
    _entry_must_be_a_correct_url(proxy)  # it is either a default or HttpUrl
    _proxy_url_must_not_contain_credentials(parsed_url)
return ",".join(proxies)
```

tests/unit/test_config.py
a-ovchinnikov: Thank you. Nothing in the test betrays your intent. My usual recommendation is to  embed test's purpose into test name, e.g. `test_default_proxy_values_are_not_rejected_by_validation`.

tests/unit/test_config.py
a-ovchinnikov: I am sorry, I don't get it. How functionality in a mixin could be affected by a class you base off it? _Edit_: spelling

hermeto/core/config.py
```diff
@@ -156,9 +192,8 @@ class YarnSettings(ProxyMixin, extra="forbid"):
 class GomodSettings(ProxyMixin, extra="forbid"):
     """Settings for Go modules."""
 
-    # TODO: refactor typesystem to trivially include comma-separated lists of URLs.
-    # Ignore type error until that happens.
-    proxy_url: str = "https://proxy.golang.org,direct"  # type: ignore
+    # GOPROXY is a comma-separated list, not a single HttpUrl like other backends.
+    proxy_url: GoProxy = "https://proxy.golang.org,direct"  # type: ignore[assignment]
```
a-ovchinnikov, line 196: The point of the change was (among other things) to remove `#type: ignore`, why is it still necessary?

hermeto/core/config.py
a-ovchinnikov: That's true, however you do not use lists of HttpUrls anywhere, you directly construct values. Please note all instances in your commit where you use this adapter. Have you tried using a constructor instead?

PR #1761, opened 2026-09-04
fix: send a descriptive User-Agent on all HTTP requests

algomaster99:
`hermeto` never set a User-Agent, so requests/aiohttp fell back to their library defaults (`python-requests/2.34.2`, `Python/3.12 aiohttp/3.14.1`). These defaults can be treated as bot/scraper signature by Maven Central and block it. Moreover, Sonatype Central also blocks the IP corresponding to that user agent for a day once this is triggered. This is also documented here: https://central.sonatype.org/faq/429-tooling-provider/#identify-your-tool-in-the-user-agent

### DIFF
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import asyncio
+import importlib.metadata
 import logging
 import ssl
 import types
@@ -20,6 +21,7 @@ from urllib3.connectionpool import ConnectionPool
 from urllib3.response import BaseHTTPResponse
 from urllib3.util.retry import Retry
 
+from hermeto import APP_NAME
 from hermeto.core.config import ProxyUrl, get_config
 from hermeto.core.errors import FetchError
 from hermeto.core.scm import get_repo_id
@@ -31,6 +33,27 @@ SAFE_REQUEST_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE"})
 BACKOFF_FACTOR = 1.3
 STATUS_FORCELIST = (500, 502, 503, 504)
 DEFAULT_CHUNK_SIZE = 65536  # 64KB
+
+
+def _get_user_agent() -> str:
+    """
+    Build a descriptive User-Agent string identifying hermeto to remote servers.
+
+    The default User-Agent sent by requests/aiohttp (e.g. "python-requests/2.34.2",
+    "Python/3.12 aiohttp/3.14.1") is indistinguishable from generic scraping traffic
+    and gets blocked by some package registries' WAFs (see
+    https://central.sonatype.org/faq/429-tooling-provider/). This affects every
+    hermeto HTTP request, including ones unrelated to the artifact that tripped the
+    block, since some WAFs blackhole the whole source IP once triggered.
+    """
+    try:
+        version = importlib.metadata.version("hermeto")
+    except importlib.metadata.PackageNotFoundError:
+        version = "unknown"
+    return f"{APP_NAME}/{version} (+https://github.com/hermetoproject/hermeto)"
+
+
+USER_AGENT: str = _get_user_agent()
 
 log = logging.getLogger(__name__)
 
@@ -84,6 +107,7 @@ def _get_pkg_requests_session() -> requests.Session:
     if _pkg_requests_session is None:
         max_retries = get_config().http.max_retries
         _pkg_requests_session = Session()
+        _pkg_requests_session.headers["User-Agent"] = USER_AGENT
         adapter = HTTPAdapter(
             max_retries=SyncLoggingRetry(
                 backoff_factor=BACKOFF_FACTOR,
@@ -256,6 +280,7 @@ async def async_download_files(
         trust_env=True,
         # preserve percent-encoding in redirect URLs (e.g. signed CloudFront URLs)
         requote_redirect_url=False,
+        headers={"User-Agent": USER_AGENT},
     )
 
     async with retry_client as session:
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -189,7 +189,11 @@ async def _async_download_binary_file(
 
     try:
         async with session.get(
-            url, timeout=timeout, raise_for_status=True, ssl=ssl_context, headers=headers
+            url,
+            timeout=timeout,
+            raise_for_status=True,
+            ssl=ssl_context,
+            headers={**(headers or {}), "User-Agent": _get_user_agent()},
         ) as response:
             async with aiofiles.open(download_path, "wb") as f:
                 async for chunk in response.content.iter_chunked(chunk_size):
--- a/hermeto/core/package_managers/general.py
+++ b/hermeto/core/package_managers/general.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 import asyncio
+import functools
 import importlib.metadata
 import logging
 import ssl
@@ -35,6 +36,7 @@ STATUS_FORCELIST = (500, 502, 503, 504)
 DEFAULT_CHUNK_SIZE = 65536  # 64KB
 
 
+@functools.cache
 def _get_user_agent() -> str:
     """
     Build a descriptive User-Agent string identifying hermeto to remote servers.
@@ -52,8 +54,6 @@ def _get_user_agent() -> str:
         version = "unknown"
     return f"{APP_NAME}/{version} (+https://github.com/hermetoproject/hermeto)"
 
-
-USER_AGENT: str = _get_user_agent()
 
 log = logging.getLogger(__name__)
 
@@ -107,7 +107,7 @@ def _get_pkg_requests_session() -> requests.Session:
     if _pkg_requests_session is None:
         max_retries = get_config().http.max_retries
         _pkg_requests_session = Session()
-        _pkg_requests_session.headers["User-Agent"] = USER_AGENT
+        _pkg_requests_session.headers["User-Agent"] = _get_user_agent()
         adapter = HTTPAdapter(
             max_retries=SyncLoggingRetry(
                 backoff_factor=BACKOFF_FACTOR,
@@ -280,7 +280,7 @@ async def async_download_files(
         trust_env=True,
         # preserve percent-encoding in redirect URLs (e.g. signed CloudFront URLs)
         requote_redirect_url=False,
-        headers={"User-Agent": USER_AGENT},
+        headers={"User-Agent": _get_user_agent()},
     )
 
     async with retry_client as session:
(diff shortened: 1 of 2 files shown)


hermeto/core/package_managers/general.py
```diff
@@ -32,6 +34,27 @@
...
+
+def _get_user_agent() -> str:
+    """
+    Build a descriptive User-Agent string identifying hermeto to remote servers.
+
+    The default User-Agent sent by requests/aiohttp (e.g. "python-requests/2.34.2",
+    "Python/3.12 aiohttp/3.14.1") is indistinguishable from generic scraping traffic
+    and gets blocked by some package registries' WAFs (see
+    https://central.sonatype.org/faq/429-tooling-provider/). This affects every
+    hermeto HTTP request, including ones unrelated to the artifact that tripped the
```
a-ovchinnikov, line 48: Have you observed this behavior or is this a theoretical possibility?

tests/unit/package_managers/test_general.py
```diff
@@ -80,6 +82,63 @@ def test_max_retries_propagated_to_session(monkeypatch: pytest.MonkeyPatch) -> N
...
+
+
+def test_pkg_requests_session_sends_custom_user_agent(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Requests must identify themselves as hermeto, not as the default python-requests UA.
+
+    Some package registries block the generic library-default User-Agent as a suspected
+    scraper/bot signature (see https://central.sonatype.org/faq/429-tooling-provider/).
+    """
+    monkeypatch.setattr("hermeto.core.package_managers.general._pkg_requests_session", None)
+    monkeypatch.setattr("hermeto.core.package_managers.general.USER_AGENT", "hermeto/test-ua")
```
a-ovchinnikov: A minor consistency nitpick: could you please patch these objects in tests by decorating the test, like in https://github.com/hermetoproject/hermeto/blob/e2b6faca96756a651dd624a5f5249c976bf4de67/tests/unit/package_managers/test_general.py#L61 ?

tests/unit/package_managers/test_general.py
```diff
@@ -80,6 +82,63 @@ def test_max_retries_propagated_to_session(monkeypatch: pytest.MonkeyPatch) -> N
...
+
+    session = _get_pkg_requests_session()
+
+    assert session.headers["User-Agent"] == "hermeto/test-ua"
+
+
+@pytest.mark.asyncio
+@mock.patch("hermeto.core.package_managers.general.aiohttp_retry.RetryClient")
+async def test_async_download_files_sends_custom_user_agent(
```
a-ovchinnikov, line 98: Does this test check Hermeto code or aiohttp code? As long as aiohttp respects its interface contract we, as its users, should be confident that it does as instructed, i.e. it includes the `User-Agent` header, it is [tested there](https://github.com/aio-libs/aiohttp/blob/25522ad29937a6b586cb4dbe906c96db45fdc672/tests/test_client_request.py#L405) after all. Checking behavior of interpreters and third party tools should never be a concern of unit tests.

hermeto/core/package_managers/general.py
```diff
@@ -32,6 +34,27 @@
...
+    block, since some WAFs blackhole the whole source IP once triggered.
+    """
+    try:
+        version = importlib.metadata.version("hermeto")
+    except importlib.metadata.PackageNotFoundError:
+        version = "unknown"
+    return f"{APP_NAME}/{version} (+https://github.com/hermetoproject/hermeto)"
+
+
+USER_AGENT: str = _get_user_agent()
```
a-ovchinnikov: Why did you choose to add a module-level constant for this value? _Edit_: to give some additional context to my question: this particular call is cheap, and could be made even cheaper by caching the return value instead of introducing a new constant.

hermeto/core/package_managers/general.py
```diff
@@ -32,6 +34,27 @@
...
+    The default User-Agent sent by requests/aiohttp (e.g. "python-requests/2.34.2",
+    "Python/3.12 aiohttp/3.14.1") is indistinguishable from generic scraping traffic
+    and gets blocked by some package registries' WAFs (see
+    https://central.sonatype.org/faq/429-tooling-provider/). This affects every
+    hermeto HTTP request, including ones unrelated to the artifact that tripped the
+    block, since some WAFs blackhole the whole source IP once triggered.
+    """
+    try:
+        version = importlib.metadata.version("hermeto")
+    except importlib.metadata.PackageNotFoundError:
```
a-ovchinnikov: When do you expect to see this error?

tests/unit/package_managers/test_general.py
```diff
@@ -80,6 +82,63 @@ def test_max_retries_propagated_to_session(monkeypatch: pytest.MonkeyPatch) -> N
     assert adapter.max_retries.total == 7
 
 
+def test_get_user_agent() -> None:
+    """The User-Agent must identify hermeto by name, version, and project URL."""
+    with mock.patch("importlib.metadata.version", return_value="1.2.3"):
+        user_agent = _get_user_agent()
+
+    assert user_agent == f"{general.APP_NAME}/1.2.3 (+https://github.com/hermetoproject/hermeto)"
```
a-ovchinnikov: This test verifies the function as it is written, `f"{general.APP_NAME}/1.2.3 (+https://github.com/hermetoproject/hermeto)"` is an almost verbatim copy of the return value (which, arguably, is constructed by the only non-trivial statement in the body of the function), it effectively tests whether substitutions in Python work as expected. I won't block the PR due to this, but I am not sure this test actually protects against any regressions.

tests/unit/package_managers/test_general.py
```diff
@@ -80,6 +82,63 @@ def test_max_retries_propagated_to_session(monkeypatch: pytest.MonkeyPatch) -> N
...
+
+
+def test_get_user_agent_falls_back_when_package_not_found() -> None:
+    """Building the User-Agent must not fail if hermeto's own package metadata is missing.
+
+    This can happen when hermeto is run from source without being installed (e.g. `python -m`
```
a-ovchinnikov: How realistic is this scenario? Wouldn't this require to have all Hermeto dependencies to be available in such environment?

a-ovchinnikov (commented):
Hi @algomaster99, thank you for your contribution! Please find some comments and questions inline.

hermeto/core/package_managers/general.py
algomaster99: Yes, this happened to me today, and I had a conversation with @brianf, who suggested that changing the `User-Agent` to something related to the tool should prevent requests from `hermeto` to classified as spam.

tests/unit/package_managers/test_general.py
algomaster99: >  it effectively tests whether substitutions in 

That's a good point. I will remove this test entirely.

tests/unit/package_managers/test_general.py
algomaster99: I haven't tried this out. Need to confirm.

tests/unit/package_managers/test_general.py
algomaster99: `  monkeypatch.setattr("hermeto.core.package_managers.general.USER_AGENT", "hermeto/test-ua")` Based on this line, it seems it is indeed testing the header value of `RetryClient`.

hermeto/core/package_managers/general.py
a-ovchinnikov: Thank you for confirming! If possible could you please provide some additional details? Was it some custom registry? Was this your first attempt to work with that registry or did it work before and stopped working at some point?

algomaster99: So I was running this command
```
./hermeto fetch-deps --source ../01-project --output ./output '{"type": "x-maven", "path": "."}'  
```
on [this lockfile](https://github.com/algomaster99/maven-hermeto-demo/blob/1f5d395305ff23c7cc6de3fb795b20de0a16effb/01-project/lockfile.json).

I think all the URLs in the lockfile are https://repo.maven.apache.org/maven2/ so it was the default maven central registry.

tests/unit/package_managers/test_general.py
algomaster99: The agent thought about this because I was running hermeto from source, but this should not be a problem in docker image of hermeto.

hermeto/core/package_managers/general.py
a-ovchinnikov: Thank you!

hermeto/core/package_managers/general.py
```diff
@@ -189,7 +189,11 @@ async def _async_download_binary_file(
 
     try:
         async with session.get(
-            url, timeout=timeout, raise_for_status=True, ssl=ssl_context, headers=headers
+            url,
+            timeout=timeout,
+            raise_for_status=True,
+            ssl=ssl_context,
+            headers={**(headers or {}), "User-Agent": _get_user_agent()},
```
a-ovchinnikov: I am sorry, I don't quite follow this change. Why would pip or npm silently clobber User-Agent header on retry?

a-ovchinnikov (commented):
@algomaster99 please squash commits addressing requested changes into commits introducing bits that were requested to be changed.

tests/unit/package_managers/test_general.py
a-ovchinnikov: Did you run it in a virtual environment or in a shared one?

algomaster99: Shared one

hermeto/core/package_managers/general.py
algomaster99: I think the intent here that the user should not be able to control the header values (It should always be `hermeto` specific). In retrospect, is the header arg, user controlled by any chance? If it is not, then yes, this test is not necessary.

algomaster99: The user can probably provide headers using [this arg](https://github.com/hermetoproject/hermeto/blob/e2b6faca96756a651dd624a5f5249c976bf4de67/hermeto/core/package_managers/generic/models.py#L103)?

a-ovchinnikov: Oh, got it, thank you!

hermeto/core/package_managers/general.py
```diff
@@ -165,7 +189,11 @@ async def _async_download_binary_file(
...
```
a-ovchinnikov: Optional suggestion:
```python
headers = (headers or {}} | {"User-Agent": _get_user_agent()}
```

a-ovchinnikov (commented):
Yes, squashing all commits in this case looks right.

hermeto/core/package_managers/general.py
algomaster99: Incorporating it.

tests/unit/package_managers/test_general.py
a-ovchinnikov: If it is not testing Hermeto code then it is not needed in the codebase, it creates additional noise that way.

a-ovchinnikov (commented):
@algomaster99 the change looks reasonably good, please address the remaining outstanding comment and squash commits that need squashing.

PR #1762, opened 2026-09-06
Refactor/rooted path pathlib interface

filipovskyj02:
## Summary
- Add `__truediv__` (`/` operator), read-only properties (`name`, `stem`,
  `suffix`) and query methods (`exists`, `is_dir`, `is_file`,
  `is_relative_to`, `relative_to`, `as_posix`) to `RootedPath`
- Migrate existing callsites to use the new interface

### DIFF
--- a/hermeto/core/resolver.py
+++ b/hermeto/core/resolver.py
@@ -66,7 +66,7 @@ def resolve_packages(request: Request) -> RequestOutput:
         for project_file in output.build_config.project_files:
             try:
                 subpath = project_file.abspath.relative_to(source_backup)
-                project_file.abspath = original_source_dir / subpath
+                project_file.abspath = (original_source_dir / subpath).path
             except ValueError:
                 # '<project_file.abspath> is not in the subpath of <source_backup>', i.e the file
                 # is referenced directly from the output directory and doesn't need replacing
--- a/hermeto/core/rooted_path.py
+++ b/hermeto/core/rooted_path.py
@@ -34,6 +34,9 @@ class RootedPath(os.PathLike[str]):
     The join_within_root method remembers the original root. See the join_within_root
     and re_root docstrings for more details.
 
+    Exposes common Path queries (exists, is_dir, name, stem, etc.) directly so
+    callers need not reach through the .path accessor for everyday operations.
+
     Implements the PathLike interface -> most stdlib methods that accept paths will work
     with a RootedPath as well.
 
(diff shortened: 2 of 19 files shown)


hermeto/core/resolver.py
```diff
@@ -66,7 +66,7 @@ def resolve_packages(request: Request) -> RequestOutput:
         for project_file in output.build_config.project_files:
             try:
                 subpath = project_file.abspath.relative_to(source_backup)
-                project_file.abspath = original_source_dir / subpath
+                project_file.abspath = (original_source_dir / subpath).path
```
tnevrlka, line 69: It looks like this is not related to the exposure of `pathlib.Path` operations.
Should this be in another commit?

hermeto/core/rooted_path.py
```diff
@@ -88,6 +91,49 @@ def __repr__(self) -> str:
...
+        return self._path.stem
+
+    @property
+    def suffix(self) -> str:
+        """Return the file extension of the final component."""
+        return self._path.suffix
+
+    def exists(self) -> bool:
+        """Return whether the path points to an existing filesystem entry."""
+        return self._path.exists()
```
tnevrlka, line 119: How were the exposed operations chosen? I assume read-only that cannot escape root? Might be worth describing.

hermeto/core/rooted_path.py
```diff
@@ -88,6 +91,49 @@ def __repr__(self) -> str:
     def __hash__(self) -> int:
         return hash((self._path, self._root))
 
+    def __truediv__(self: RootedPathT, other: StrPath) -> RootedPathT:
+        """Join via ``join_within_root``; raises PathOutsideRoot on escape."""
+        return self.join_within_root(other)
```
tnevrlka, line 96: Since `__truediv__` is the only operation behaving differently from `pathlib.Path`, would it make sense to have it in its own commit?

tests/unit/test_rooted_path.py
```diff
@@ -123,6 +123,85 @@ def test_rooted_path_eq() -> None:
...
+    rp = RootedPath(test_path).join_within_root(path_suffix)
+    assert getattr(rp, prop) == expected
+
+
+def test_name_at_root(test_path: Path) -> None:
+    rp = RootedPath(test_path)
+    assert rp.name == test_path.name
+
+
+
```
tnevrlka: nit: looks like three newlines

tests/unit/test_rooted_path.py
```diff
@@ -123,6 +123,85 @@ def test_rooted_path_eq() -> None:
...
+@pytest.mark.parametrize(
+    "prop, path_suffix, expected",
+    [
+        pytest.param("name", "subpath", "subpath", id="name"),
+        pytest.param("name", "archive.tar.gz", "archive.tar.gz", id="name-with-dots"),
+        pytest.param("stem", "archive.tar", "archive", id="stem"),
+        pytest.param("suffix", "archive.tar", ".tar", id="suffix"),
+        pytest.param("suffix", "no_extension", "", id="suffix-empty"),
+    ],
+)
```
tnevrlka, line 163: Is it possible to simplify the tests? 

Exposed operations (besides `__truediv__`) are supposed to call `pathlib.Path`'s operations and should thus behave 1:1.

I don't think it's necessary to try to test edge cases, I would assume that just checking the happy case is behaving same as `pathlib.Path` should be enough

tnevrlka (commented):
Overall looks good to me, few comments for thought

hermeto/core/rooted_path.py
filipovskyj02: Yes, as the minimum issue desc stated, wanted to open the pr for feedback early, so open to expanding on it.

filipovskyj02: Added a comment for it for now

hermeto/core/resolver.py
filipovskyj02: Moved to the migration commit

hermeto/core/rooted_path.py
filipovskyj02: Added a third commit for it

tests/unit/test_rooted_path.py
a-ovchinnikov: Agreed, these tests are an overkill for all cases when the underlying implementation is just passed through.

filipovskyj02: Simplified

hermeto/core/rooted_path.py
taylormadore: What about I/O methods? Things like open, read_text, write_text, mkdir, etc.

PR #1763, opened 2026-09-07
fix(core): retry source tree copy racing git gc

derasdf:
Before running a backend hermeto copies the source tree, including its `.git`, into a working copy. That copy isn't atomic, so a `git gc` running on the source repo can pull objects out from under it mid-copy. Usually the copy just crashes with a `shutil.Error`, but in the nasty case it finishes with a broken `.git` and no error at all, and an unresolvable HEAD then makes gomod quietly resolve the wrong module version.

Rather than skip whatever vanished (which would turn a loud crash into silent SBOM corruption), we retry the copy until git has settled and check that the copied repo still resolves HEAD. Our own integration tests tripped the same race when git's auto-gc fired on their test repo, so auto-gc is now disabled there.

See the commit messages for the details.

CI failure this fixes:
https://github.com/hermetoproject/hermeto/actions/runs/33869701229/job/101012505958

### DIFF
--- a/hermeto/core/resolver.py
+++ b/hermeto/core/resolver.py
@@ -18,6 +18,7 @@ from hermeto.core.package_managers import (
 from hermeto.core.package_managers.javascript import metayarn, npm, pnpm
 from hermeto.core.package_managers.python import pip
 from hermeto.core.rooted_path import RootedPath
+from hermeto.core.scm import repo_head_resolves
 from hermeto.core.utils import copy_directory
 
 Handler = Callable[[Request], RequestOutput]
@@ -46,6 +47,21 @@ def get_experimental_backends() -> list[PackageManagerType | str]:
     return sorted(x for x in _package_managers if x.startswith("x-"))
 
 
+def _source_copy_validator(original: Path) -> Callable[[Path], bool]:
+    """Build a validator that rejects a source copy which lost a resolvable HEAD to the copy race.
+
+    A copy racing a concurrent 'git gc' on the source can end up with an inconsistent .git whose
+    HEAD no longer resolves, in which case copy_directory retries. The validator only checks HEAD,
+    not the whole object graph: a full 'git fsck' would run O(repo) on every request with a .git,
+    including ecosystems that never read git history.
+
+    The check is only enforced when the original HEAD resolves, so a non-git or commit-less source
+    is not rejected for an unresolvable HEAD it never had.
+    """
+    original_resolves = repo_head_resolves(original)
+    return lambda copy: not original_resolves or repo_head_resolves(copy)
+
+
 def resolve_packages(request: Request) -> RequestOutput:
     """
     Resolve all packages specified in a request.
@@ -56,7 +72,11 @@ def resolve_packages(request: Request) -> RequestOutput:
     original_source_dir = request.source_dir
 
     with TemporaryDirectory(f".{APP_NAME}-source-copy", dir=original_source_dir.path) as temp_dir:
-        source_backup = copy_directory(original_source_dir.path, Path(temp_dir).resolve())
+        source_backup = copy_directory(
+            original_source_dir.path,
+            Path(temp_dir).resolve(),
+            is_valid=_source_copy_validator(original_source_dir.path),
+        )
 
         request.source_dir = RootedPath(source_backup)
         output = _resolve_packages(request)
--- a/hermeto/core/utils.py
+++ b/hermeto/core/utils.py
@@ -7,6 +7,7 @@ import re
 import shutil
 import subprocess
 import sys
+import time
 from collections.abc import Callable, Iterator, Sequence
 from functools import cache
 from pathlib import Path
@@ -166,14 +167,34 @@ def _fast_copy(src: Path, dest: Path) -> int:
     return total
 
 
-def copy_directory(origin: Path, destination: Path) -> Path:
+# A concurrent 'git gc' on the source repo can make a copy fail or copy an inconsistent .git.
+# Housekeeping is short-lived, so a few retries are enough to land on a clean tree.
+_COPY_MAX_ATTEMPTS = 5
+_COPY_RETRY_BACKOFF = 0.1
+
+
+def copy_directory(
+    origin: Path,
+    destination: Path,
+    is_valid: Callable[[Path], bool] = lambda _: True,
+) -> Path:
     """
     Recursively copy directory to another path.
 
     Use fast in-kernel copying (including reflink file system optimization) and fall back to
     regular copy if the former fails for some reason.
 
-    :raise FileExistsError: if the destination path already exists.
+    copytree first scans a directory and then copies each entry, so a process that mutates the tree
+    between those steps can make an entry disappear mid-copy. The usual cause is a concurrent
+    'git gc' on a source repo, which packs loose objects and refs and then removes the loose
+    originals. Such a race either raises shutil.Error or produces an inconsistent .git, so the copy
+    is retried until it succeeds.
+
+    :param is_valid: predicate run on a successful copy; returning False rejects the copy and
+        triggers another attempt. Callers copying a source repo use it to reject an inconsistent
+        .git that copied without raising. Defaults to accepting every copy.
+    :raise shutil.Error: if the copy keeps failing, or keeps validating as inconsistent, across
+        all retries.
     :raise FileNotFoundError: if the origin directory does not exist.
     """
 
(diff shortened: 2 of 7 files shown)


hermeto/core/utils.py
```diff
@@ -187,14 +208,37 @@ def _copy_using(copy_function: Callable) -> None:
...
+            log.debug("Copying %s to %s using fast in-kernel copy.", origin, destination)
+            _copy_using(_fast_copy)
+        except _FastCopyFailedFallback:
+            log.debug("Fast copying failed, falling back to standard copy.")
+            shutil.rmtree(destination)
+            _copy_using(shutil.copy2)
+
+    # Retry the copy while it keeps losing the race. The final attempt runs outside the loop so a
+    # persistent failure propagates instead of silently returning a broken copy.
+    for attempt in range(_COPY_MAX_ATTEMPTS - 1):
```
a-ovchinnikov: Why did you choose implementing a retry locally when there is `tenacity.retry` available?

hermeto/core/utils.py
```diff
@@ -187,14 +208,37 @@ def _copy_using(copy_function: Callable) -> None:
...
+        except shutil.Error:
+            pass  # lost the race, retry below
+        else:
+            if validate is None or validate(destination):
+                return destination
+        log.debug("Source copy lost a race with git gc, retrying (attempt %d)", attempt + 1)
+        time.sleep(_COPY_RETRY_BACKOFF * (attempt + 1))
+
+    _attempt()
+    if validate is not None and not validate(destination):
```
a-ovchinnikov: Naming nitpick: validate vs is_valid, the latter reads as something that is either True or False, it is a bit hard to tell what the former is.

hermeto/core/utils.py
```diff
@@ -166,14 +167,34 @@ def _fast_copy(src: Path, dest: Path) -> int:
...
+# A concurrent 'git gc' on the source repo can make a copy fail or copy an inconsistent .git.
+# Housekeeping is short-lived, so a few retries are enough to land on a clean tree.
+_COPY_MAX_ATTEMPTS = 5
+_COPY_RETRY_BACKOFF = 0.1
+
+
+def copy_directory(
+    origin: Path,
+    destination: Path,
+    validate: Callable[[Path], bool] | None = None,
```
a-ovchinnikov: `Callable[[Path], bool] | None` feels like an overkill: who and when will call it without specifying `validate`? Why not to short-circuit it like this:
```python
validate: Callable[[Path], bool] = λ _: True
```
?

hermeto/core/resolver.py
```diff
@@ -46,6 +47,21 @@ def get_experimental_backends() -> list[PackageManagerType | str]:
...
+    A copy racing a concurrent 'git gc' on the source can end up with an inconsistent .git whose
+    HEAD no longer resolves, in which case copy_directory retries. The validator only checks HEAD,
+    not the whole object graph: a full 'git fsck' would run O(repo) on every request with a .git,
+    including ecosystems that never read git history.
+
+    The check is only enforced when the original HEAD resolves, so a non-git or commit-less source
+    is not rejected for an unresolvable HEAD it never had.
+    """
+    original_resolves = repo_head_resolves(original)
+    return lambda copy: not original_resolves or repo_head_resolves(copy)
```
a-ovchinnikov, line 62: This is a neat closure trick!

hermeto/core/resolver.py
```diff
@@ -46,6 +47,21 @@ def get_experimental_backends() -> list[PackageManagerType | str]:
     return sorted(x for x in _package_managers if x.startswith("x-"))
 
 
+def _source_copy_validator(original: Path) -> Callable[[Path], bool]:
+    """Build a validator that rejects a source copy which lost a resolvable HEAD to the copy race.
+
+    A copy racing a concurrent 'git gc' on the source can end up with an inconsistent .git whose
```
a-ovchinnikov, line 53: I am sorry, I do not understand why is it still necessary, the previous commit seems to have turned gc off, should it be turned off elsewhere?

hermeto/core/utils.py
```diff
@@ -166,14 +167,34 @@ def _fast_copy(src: Path, dest: Path) -> int:
...
     """
     Recursively copy directory to another path.
 
     Use fast in-kernel copying (including reflink file system optimization) and fall back to
     regular copy if the former fails for some reason.
 
-    :raise FileExistsError: if the destination path already exists.
+    copytree first scans a directory and then copies each entry, so a process that mutates the tree
+    between those steps can make an entry disappear mid-copy. The usual cause is a concurrent
+    'git gc' on a source repo, which packs loose objects and refs and then removes the loose
```
a-ovchinnikov, line 189: This makes me wonder whether git initialization should be moved to after a copy, have you considered this?

hermeto/core/utils.py
derasdf: tenacity's great for retry-on-exception (like `Go._run`), but here I also retry when the copy *succeeds* but fails validation, and the two failure endings differ (re-raise the real error vs. raise a custom one). tenacity collapses both into `RetryError` (checked: `reraise` only re-raises the last exception, and the validation case has none), so I'd end up re-adding that branching after the decorator anyway. Kept the loop, but simplified it so there's a single `_attempt()` call and both give-up cases are inline, which I think was the thing the structure was hiding.

hermeto/core/utils.py
derasdf: Done

hermeto/core/utils.py
derasdf: Done

hermeto/core/resolver.py
derasdf: The gc-off only helps in the test because the fixture is what triggers the gc: it does `git add .` + one big `git commit`, which crosses the `gc.auto` threshold and spawns the detached `git gc`. Turning auto-gc off removes a trigger we create ourselves. In real runs hermeto never writes to the source (it only copytrees it), so it never triggers gc — the racing gc comes from the user's own git activity, outside our control. There's nothing of ours to switch off, and we won't mutate the user's `.git/config`. So the race can't be prevented, only retried through. And disabling auto-gc in the fixture doesn't drop coverage of the retry: that's exercised deterministically by the unit tests (which force the race by mocking copytree), not by hoping the IT repo races.

hermeto/core/utils.py
derasdf: If you mean init a fresh repo on the copy instead of copying the source `.git`: that would drop the origin, HEAD and history, but backends read those off the copy (gomod calls `get_repo_id()` on it for the SBOM), so a re-init loses the identity. If you mean the test fixture, its `init` already runs before the copy and the copy needs a populated `.git` to hit the race, so it can't move after either.

tests/integration/utils.py
```diff
@@ -89,6 +89,12 @@ def __init__(
 
         # main repo creation
         self.repo = GitRepo.init(repo_path, env=GIT_PRISTINE_ENV)
+        # Disable auto-gc: committing the whole scenario at once creates enough loose objects to
+        # cross the gc.auto threshold, and the detached 'git gc --auto' would then repack this repo
+        # while hermeto copies the source tree, racing the copy and making the suite flaky. Nothing
+        # else writes to this repo, so with auto-gc off it stays unchanged during the copy.
+        with self.repo.config_writer() as cw:
+            cw.set_value("gc", "auto", 0)
```
slimreaper35: Wouldn't be better to put it into the `GIT_PRISTINE_ENV` with other config options ?

derasdf: Fair, moved

hermeto/core/resolver.py
taylormadore: So outside of the issue with the integration tests that is fixed in the first commit, this issue wouldn't occur during the source copy unless the user were actively doing git operations in the source repo while Hermeto was executing a Request?

taylormadore (commented):
Has the source tree copy issue been seen outside of the issue with the tests?

hermeto/core/utils.py
a-ovchinnikov: I still don't get it why validator cannot raise an exception to signal to tenacity that it needs a re-run.

hermeto/core/utils.py
```diff
@@ -166,14 +167,34 @@ def _fast_copy(src: Path, dest: Path) -> int:
...
+# A concurrent 'git gc' on the source repo can make a copy fail or copy an inconsistent .git.
+# Housekeeping is short-lived, so a few retries are enough to land on a clean tree.
+_COPY_MAX_ATTEMPTS = 5
+_COPY_RETRY_BACKOFF = 0.1
+
+
+def copy_directory(
+    origin: Path,
+    destination: Path,
+    is_valid: Callable[[Path], bool] = lambda _: True,
```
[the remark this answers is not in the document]
a-ovchinnikov: @coderabbit this is noise again.

hermeto/core/resolver.py
derasdf: Yes, we never run git or touch `.git` on the source, we just copy it, so we never trigger gc ourselves. It'd only race if the user's own git activity (or a background auto-gc) hits the repo mid-request. Per the original investigation that can leave an unresolvable HEAD and make gomod resolve the wrong version, so we retry.

hermeto/core/utils.py
derasdf: Fair, updated with tenacity

PR #1765, opened 2026-09-08
Polishing Hermeto docs.

ajko-denai:
There were outdated links and information.

Deployment link in docs/bundler.md was outdated and poiting to dead site.
A typo "Dockerfile or Dockerfile" was fixed in multiple places.
Buildah warning about issue with "--network=none" was old and closed back in 2022. The whole block was removed.

### DIFF
--- a/docs/bundler.md
+++ b/docs/bundler.md
@@ -216,4 +216,4 @@ Dockerfile won't take effect, consider `BUNDLE_PATH` instead.
 
 [Bundler]: https://bundler.io
 [configuration options]: https://bundler.io/v2.5/man/bundle-config.1.html#DESCRIPTION
-[deployment mode]: https://www.bundler.cn/man/bundle-install.1.html#DEPLOYMENT-MODE
+[deployment mode]: https://guides.rubygems.org/command-reference/bundle-install#DEPLOYMENT-MODE
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -165,14 +165,14 @@ hermeto merge-sboms <hermeto_sbom_1.json> ... <hermeto_sbom_n.json> -o <merged_s
 
 After the pre-fetch and the above steps to inform the package manager(s) of the
 cache have been completed, it all needs to be wired up into a build. The primary
-use case for building these is within a Dockerfile or Dockerfile but the same
+use case for building these is within a Dockerfile but the same
 principles can be applied to other build strategies.
 
 #### Write the Dockerfile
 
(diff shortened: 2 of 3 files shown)


docs/usage.md
```diff
@@ -248,19 +248,14 @@ The absolute path to ./hermeto-output on your machine is (probably) not
 the path inside the container is eventually going to be.
 
 In order to run the build with network isolation, use the `--network=none`
-option. Note that this option only works if your podman/buildah version contains
-the fix for [buildah#4227][] (buildah >= 1.28). In older versions, a workaround
-could be to manually create an internal network (but you'll need root
-privileges): `sudo podman network create --internal isolated-network; sudo
-podman build --network isolated-network ...`.
+option.
```
a-ovchinnikov, line 251: WDYT about rewording this instead or dropping completely? There still could be a chance that someone uses an antiquated version of buildah for some reason. Something along the lines of
```
For buildah versions older than 1.28 one would need to manually create an internal network:
`sudo podman network create --internal isolated-network  
 sudo podman build --network isolated-network ...`.
```
Maybe put this into a footnote?

docs/bundler.md
a-ovchinnikov, line 1: _Commit message_: please adhere to qualified commit message structure. This requires providing a scope prefix to title, in this case it could be "docs:" or "chore:". Also when writing commit messages please try to adhere to 50/72 rule (50 is the recommended number of characters of the commit title, 72 is the recommended width of the commit body) -- going a bit above the suggested limits is fine, but going too far above will mess with tooling of those who prefer working through a terminal.

a-ovchinnikov (commented):
LGTM in general, please find two minor requests inline.

docs/usage.md
slimreaper35: FWIW I suggested this privately, since there is already buildah version 1.45.0. Yes, someone could use an ancient version of buildah, but I am not sure if we need to keep or maintain such documentation.

a-ovchinnikov: I definitely won't block on loosing this particular bit of information, that being said I would rather have it tuck under a footnote or collapsed section just in case (I fully agree that the chance that it will be needed by someone is extremely small, I just want the doc to be as water tight as possible).

docs/bundler.md
ajko-denai: Thank you, TIL. Fixed.

a-ovchinnikov (approved):

slimreaper35 (approved):

derasdf (approved):

PR #1770, opened 2026-09-12
docs: use POSIX dot instead of source in Dockerfile examples

Abhishek9639:
The pnpm example in `docs/pnpm.md` uses the `node:24` base image, where `/bin/sh` is dash. `RUN` executes via `/bin/sh -c` and dash has no `source` builtin, so the step failed before `pnpm install` ran:

```text
/bin/sh: 1: source: not found
Error: building at STEP "RUN source /tmp/hermeto.env && pnpm install": exit status 127
```

`source` is a bash extension, while the POSIX `.` is equivalent and works in dash, busybox ash and bash alike.

Following review feedback this now updates every Dockerfile example instead of only pnpm: `docs/usage.md`, `docs/gomod.md`, `docs/pip.md`, `docs/pnpm.md` and the `README.md` snippet. `docs/bundler.md` already used this form and is unchanged. Prose that named the `source` directive is reworded to match.

Verified `. /tmp/hermeto.env` on all three base images the docs use: `golang:1.19.2-alpine3.16` (busybox ash), `registry.access.redhat.com/ubi9/python-312` (bash) and `node:24` (dash).


tnevrlka (approved):
ACK, thank you

slimreaper35 (approved):

eskultety (changes requested):
> Other docs are unaffected their examples use Alpine or ubi9, where source is available.

Doesn't matter, that can change. Please update all docs as requested by one of the comments in this PR. Thank you for the patch!

Abhishek9639:
Thanks both updated it to cover all the examples rather than just pnpm.

`RUN source` → `RUN .` in `docs/usage.md`, `docs/gomod.md`, `docs/pip.md` and `docs/pnpm.md`, plus the inline snippet in `README.md`. `docs/bundler.md` already used `.`, so it's unchanged. I also reworded the few spots in `usage.md` where the prose referred to the `source` directive by name otherwise the text and the examples would disagree with each other.

I've dropped the "other docs are unaffected" line from the PR description as well; fair point, it was beside the point.

Verified `. /tmp/hermeto.env` on all three base images the docs actually use: `golang:1.19.2-alpine3.16` (busybox ash), `ubi9/python-312` (bash) and `node:24` (dash).

@eskultety would you mind taking another look?

eskultety (approved):

PR #1771, opened 2026-09-14
docs(cargo): copy .cargo and src in the Dockerfile example

Abhishek9639:
The hermetic build example in `docs/cargo.md` cannot work as written. `COPY` copies the *contents* of a source directory, not the directory itself, so `COPY Cargo.toml Cargo.lock .cargo .` lands the injected config at `/app/config.toml` instead of `/app/.cargo/config.toml`. Cargo never reads it, the vendored-sources replacement is skipped, and the build reaches for crates.io under `--network none`. The example also never copies the crate sources.

Reproduction, with `hermeto fetch-deps cargo` and `hermeto inject-files` already run:

```dockerfile
FROM docker.io/library/rust:latest
WORKDIR /app
COPY Cargo.toml Cargo.lock .cargo .
RUN cargo build --release
```

```text
no targets specified in the manifest
either src/lib.rs, src/main.rs, a [lib] section, or [[bin]] section must be present
Error: building at STEP "RUN cargo build --release": exit status 101
```

Adding `COPY src src` alone is not enough the `.cargo` flattening still breaks it:

```text
failed to load source for dependency `itoa`
unable to update registry `crates-io`
[6] Could not resolve hostname (Could not resolve host: index.crates.io)
Error: building at STEP "RUN cargo build --release": exit status 101
```

Gives `.cargo` an explicit destination and copies `src/`. With this the example builds under `--network none` (`Compiling hello-hermeto` → `Finished`), matching the bolded instruction at line 33 to copy `.cargo/config.toml` into the image.


eskultety (approved):
Somewhat a weird behaviour from `COPY`.

a-ovchinnikov (approved):

Abhishek9639:
Agreed, it catches people out though it is documented behaviour rather than a quirk: when the source is a directory, COPY copies its *contents*, not the directory itself, so `.cargo` gets merged into the destination instead of being recreated. With multiple sources the destination also has to be a directory, which is what collapsed all three paths into `/app` here.

The nasty part is that it fails silently: the file lands as `/app/config.toml`, Cargo simply finds no config, and you only notice when the build starts reaching for the network.

